PR #98731, opened 2025-11-02
driver :wifi: airoc: Fallback to user security on scan fail

### LINKED ISSUE #98955: AIROC Wi-Fi: Connection failure when scan fails to resolve security type for PSK connections
### Describe the bug

This issue reports a lack of robustness in the **airoc_mgmt_connect** function when a user attempts to connect using a Pre-Shared Key (PSK) but provides an explicit security type of **`WIFI_SECURITY_TYPE_NONE`**.

- **Problematic Behavior:** When `WIFI_SECURITY_TYPE_NONE` is provided with a PSK, the driver correctly attempts to **scan for the Access Point (AP)** to automatically determine the correct, specific security protocol (e.g., WPA2/WPA3).
  
- **Logic Flaw:** If this background scan **fails or does not find the AP**, the driver previously failed the entire connection attempt.
  
- **Impact:** The driver failed to fall back to the security parameters **originally provided by the user** (the PSK and `WIFI_SECURITY_TYPE_NONE`, which often implies the security is PSK-based and relies on the provided key). This results in an **unnecessary connection failure** due to a transient scan issue, even though the necessary credentials were provided.
  

**Area:** Drivers, Wi-Fi, AIROC, Network Management

### Regression

- [ ] This is a regression.

### Steps to reproduce

The bug is a logic error related to error handling after a security-type resolution attempt.

> **Write:**
> 
> 1. Configure an application to use the AIROC Wi-Fi driver.
>   
> 2. Attempt a connection using `airoc_mgmt_connect` where:
>   
>   - A **valid PSK** is provided.
>   - The security type is explicitly set to **`WIFI_SECURITY_TYPE_NONE`**.
> 3. **Manually induce a scan failure** (e.g., by temporarily disabling the AP or introducing interference) during the driver's internal security type resolution phase.
>   
> 4. The original code logic would return an error, failing the connection attempt, instead of proceeding to connect using the provided PSK. The fix ensures that if the resolved security remains **`WHD_SECURITY_UNKNOWN`**, the driver uses the user-provided parameters.
>

### Relevant log output

```shell

```

### Impact

Not sure

### Environment

- **OS:** Windows WSL
  
- **Toolchain:** Zephyr SDK
  
- **Driver:** AIROC Wi-Fi Management Driver

### Additional Context

Fix Proposal is presented in the following PR: #98731

kaabia:
Fixes <https://github.com/zephyrproject-rtos/zephyr/issues/98955>

In the `airoc_mgmt_connect` function, if the user provides a security type of `WIFI_SECURITY_TYPE_NONE` along with a PSK, the driver attempts to scan for the AP to determine the correct security protocol.

However, if this scan fails or does not find the AP, the driver would previously fail the connection attempt instead of using the security type originally provided by the user.

This change ensures that if the scan does not resolve the security type (`WHD_SECURITY_UNKNOWN`), the driver falls back to using the security type from the connection parameters. This makes the connection process more robust and resilient to scan failures.

### DIFF
--- a/drivers/wifi/infineon/airoc_wifi.c
+++ b/drivers/wifi/infineon/airoc_wifi.c
@@ -549,12 +549,19 @@ static int airoc_mgmt_scan(const struct device *dev, struct wifi_scan_params *pa
 	return 0;
 }
 
+static bool is_invalid_security(int security, uint8_t psk_length)
+{
+	return ((security == WIFI_SECURITY_TYPE_NONE) && (psk_length > 0));
+}
+
 static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_params *params)
 {
 	struct airoc_wifi_data *data = (struct airoc_wifi_data *)dev->data;
 	int ret = 0;
 	whd_scan_result_t scan_result;
 	whd_scan_result_t usr_result = {0};
+	/* Try to scan ssid to define security */
+	whd_scan_result_t tmp_result = {0};
 
 	if (k_sem_take(&data->sema_common, K_MSEC(AIROC_WIFI_WAIT_SEMA_MS)) != 0) {
 		return -EAGAIN;
@@ -574,13 +581,13 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
 
 	usr_result.SSID.length = params->ssid_length;
 	memcpy(usr_result.SSID.value, params->ssid, params->ssid_length);
+	usr_result.security = convert_zephyr_security_to_whd(params->security);
 
-	if ((params->security == WIFI_SECURITY_TYPE_NONE) && (params->psk_length > 0)) {
-		/* Try to scan ssid to define security */
+	if (is_invalid_security(params->security, params->psk_length)) {
 
 		if (whd_wifi_scan(airoc_sta_if, WHD_SCAN_TYPE_ACTIVE, WHD_BSS_TYPE_ANY, NULL, NULL,
 				  NULL, NULL, airoc_wifi_scan_cb_search, &scan_result,
-				  &(usr_result)) != WHD_SUCCESS) {
+				  &(tmp_result)) != WHD_SUCCESS) {
 			LOG_ERR("Failed start scan");
 			ret = -EAGAIN;
 			goto error;
@@ -593,8 +600,10 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
 			goto error;
 		}
 	} else {
-		/* Get security from user, convert it to  */
-		usr_result.security = convert_zephyr_security_to_whd(params->security);
+		/* Fallback to user input */
+		if (tmp_result.security == WHD_SECURITY_UNKNOWN) {
+			usr_result.security = tmp_result.security;
+		}
 	}
 
 	if (usr_result.security == WHD_SECURITY_UNKNOWN) {


drivers/wifi/infineon/airoc_wifi.c
```diff
@@ -591,8 +591,10 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
 			ret = -EAGAIN;
 			goto error;
 		}
-	} else {
-		/* Get security from user, convert it to  */
+	}
+
+	/* If security is still unknown, use user-provided security type */
+	if (usr_result.security == WHD_SECURITY_UNKNOWN) {
```
krish2718: Should we just reject this config altogether?  it's probably a config mistake giving PSK but not security type, no? e.g., Wi-Fi shell rejects this.

kaabia: Hi @krish2718 
Thanks for the insightful review. You've raised an excellent point about whether this configuration should be rejected entirely.

I agree that from a user-facing perspective, like the Wi-Fi shell, providing a PSK with WIFI_SECURITY_TYPE_NONE is a configuration error and should be flagged. The shell is the right place for such strict validation.

However, the goal of this patch is to enhance the robustness of the driver itself. A driver's API should be more resilient than a user utility. The original behavior—attempting to auto-detect security and then failing hard if the scan doesn't find the AP—can make the connection process brittle.

This patch introduces a fallback mechanism. The driver's logic becomes:

1. If security is ambiguous, try to auto-detect it by scanning.
2. If the scan fails to determine the security type, instead of giving up, fall back to using the security information the user originally provided.

This makes the driver more resilient to transient scan failures and allows it to attempt a connection with the information it has.

Based on your feedback, I've refined the patch to make this logic clearer and more robust. The updated version restores the original else block for clarity and adds a more explicit fallback check that only triggers if the scan fails. This ensures we don't inadvertently try to connect with WIFI_SECURITY_TYPE_NONE if that was the initial input that triggered the scan.

krish2718: Thanks @kaabia  for the explanation I understand the rationale, you are trying to improve the fallback scenario. But even at driver level I still think it should reject a request with incomplete info rather than assuming PSK first and then falling back to OPEN. 

Of course this discussion is outside the scope of your PR, and I don't have any objections to your fallback improvement approach. Just wanted to understand see if we can avoid this fallback altogether.

kaabia: hi @krish2718 

Thanks for the review and the insightful comments. I agree that, ideally, the driver shouldn't have to guess security settings.

To clarify one point, this change doesn't introduce a fallback from PSK to OPEN. It only uses the user-provided security type as a fallback if the scan fails to determine it. If the user provides a PSK but no security type (WIFI_SECURITY_TYPE_NONE), and the scan fails, the connection will still fail as it does today, since the driver can't safely make an assumption.

Regarding your question on how we could avoid this fallback altogether: I think that's a great goal for a future improvement. It would likely involve changes higher up the stack, for example, making the wifi shell command smarter. The shell could perform the scan itself to determine the security type before calling connect, ensuring the driver always receives a complete request.

For this PR, I've focused on making the existing driver logic more resilient. Thanks again for the discussion

krish2718: > To clarify one point, this change doesn't introduce a fallback from PSK to OPEN. It only uses the user-provided security type as a fallback if the scan fails to determine it. If the user provides a PSK but no security type (WIFI_SECURITY_TYPE_NONE), and the scan fails, the connection will still fail as it does today, since the driver can't safely make an assumption.

The entry point for the scan is `if ((params->security == WIFI_SECURITY_TYPE_NONE) && (params->psk_length > 0)) {`, so, use has chosen OPEN but also provided PSK. Before your PR, it will try and scan to figure out the security for the SSID and use it, else it will proceed with OPEN (user given). With your PR, you are adding a fallback if scan fails, but that fallback has guard `params->security != WIFI_SECURITY_TYPE_NONE`, so, scan itself won't be initiated in this case, no?

Also, in case of scan failure i.e., scan didn't trigger,  unsure if the callback is called, and even if scan is succeeded and callback is called, but SSID is not found the `user_data`  i.e., `usr_result.security` is not set. So, I would slightly refactor it like: A bit verbose but easier to read IMHO, WDYT?

```
diff --git a/drivers/wifi/infineon/airoc_wifi.c b/drivers/wifi/infineon/airoc_wifi.c
index 7604b82b66e..97a39104b1d 100644
--- a/drivers/wifi/infineon/airoc_wifi.c
+++ b/drivers/wifi/infineon/airoc_wifi.c
@@ -548,6 +548,11 @@ static int airoc_mgmt_scan(const struct device *dev, struct wifi_scan_params *pa
        return 0;
 }
 
+static bool is_invalid_security(int security)
+{
+       return ((security == WIFI_SECURITY_TYPE_NONE) && (psk_length > 0));
+}
+
 static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_params *params)
 {
        struct airoc_wifi_data *data = (struct airoc_wifi_data *)dev->data;
@@ -573,13 +578,15 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
 
        usr_result.SSID.length = params->ssid_length;
        memcpy(usr_result.SSID.value, params->ssid, params->ssid_length);
+       usr_result.security = convert_zephyr_security_to_whd(params->security);
 
-       if ((params->security == WIFI_SECURITY_TYPE_NONE) && (params->psk_length > 0)) {
+       if (is_invalid_security(params->security)) {
                /* Try to scan ssid to define security */
+               whd_scan_result_t tmp_result = {0};
 
                if (whd_wifi_scan(airoc_sta_if, WHD_SCAN_TYPE_ACTIVE, WHD_BSS_TYPE_ANY, NULL, NULL,
                                  NULL, NULL, airoc_wifi_scan_cb_search, &scan_result,
-                                 &(usr_result)) != WHD_SUCCESS) {
+                                 &(tmp_result)) != WHD_SUCCESS) {
                        LOG_ERR("Failed start scan");
                        ret = -EAGAIN;
                        goto error;
@@ -591,11 +598,14 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
                        ret = -EAGAIN;
                        goto error;
                }
-       } else {
-               /* Get security from user, convert it to  */
-               usr_result.security = convert_zephyr_security_to_whd(params->security);
+
+               /* Fallback to user input */
+               if (tmp_result.security == WHD_SECURITY_UNKNOWN) {
+                       usr_result.security = tmp_result.security;
+               }
        }
 
+
        if (usr_result.security == WHD_SECURITY_UNKNOWN) {
                ret = -EAGAIN;
                LOG_ERR("Could not scan device");
```

krish2718: > Regarding your question on how we could avoid this fallback altogether: I think that's a great goal for a future improvement. It would likely involve changes higher up the stack, for example, making the wifi shell command smarter. The shell could perform the scan itself to determine the security type before calling connect, ensuring the driver always receives a complete request.

> For this PR, I've focused on making the existing driver logic more resilient. Thanks again for the discussion

Agreed, as I said above we can take it up later.

kaabia: Thank you so much for you effort and clarifications
Patch applied !

krish2718 (dismissed):

drivers/wifi/infineon/airoc_wifi.c
```diff
@@ -592,8 +599,10 @@ static int airoc_mgmt_connect(const struct device *dev, struct wifi_connect_req_
 			goto error;
 		}
 	} else {
-		/* Get security from user, convert it to  */
-		usr_result.security = convert_zephyr_security_to_whd(params->security);
+		/* Fallback to user input */
+		if (tmp_result.security == WHD_SECURITY_UNKNOWN) {
```
jukkar, line 604: It looks like the `tmp_result` is defined in previous block and there would be a compile error, or did I read the code incorrectly?

jukkar (changes requested):
cc: @sreeramIfx, you seem to be Infineon platform maintainer but it seems that nobody from Infineon is reviewing or caring about the Airoc wifi driver. For example, I could not find any place in either `tests/` or `samples/` where the Airoc driver is enabled so that it could be compile tested. Could you or someone from Infineon take ownership of the Airoc driver?

@kaabia Could you adjust the commit subject to
`drivers: wifi: airoc: Fallback to user security on scan fail`

drivers/wifi/infineon/airoc_wifi.c
kaabia: You are abseloutely right, good catch !
Fixed

jukkar (approved):

sreeramIfx (approved):

PR #98732, opened 2025-11-02
fix(can): mcp2515: Guard TX interrupt handling for multi-buffer configs

kaabia:
Fixes <https://github.com/zephyrproject-rtos/zephyr/issues/98954>

The MCP2515 driver can be configured with 1, 2, or 3 transmit buffers via the `MCP2515_TX_CNT`  constant.

The interrupt handler checks for transmit complete interrupts for `TXB1` and `TXB2`. However, the code for these checks was not guarded by preprocessor conditionals. This meant the code would be compiled even if the driver was configured to use only one transmit buffer (`MCP2515_TX_CNT=1`), which is the default.

This change adds the appropriate `#if MCP2515_TX_CNT > 1` and `#if MCP2515_TX_CNT > 2` guards around the interrupt handling logic for the second and third transmit buffers, respectively. This ensures that the code is only included when the corresponding buffers are actually configured, improving code clarity and preventing compilation of unused logic.

### DIFF
--- a/drivers/can/can_mcp2515.c
+++ b/drivers/can/can_mcp2515.c
@@ -831,13 +831,17 @@ static void mcp2515_handle_interrupts(const struct device *dev)
 			mcp2515_tx_done(dev, 0, 0);
 		}
 
+#if MCP2515_TX_CNT > 1
 		if (canintf & MCP2515_CANINTF_TX1IF) {
 			mcp2515_tx_done(dev, 1, 0);
 		}
+#endif /* MCP2515_TX_CNT > 1 */
 
+#if MCP2515_TX_CNT > 2
 		if (canintf & MCP2515_CANINTF_TX2IF) {
 			mcp2515_tx_done(dev, 2, 0);
 		}
+#endif /* MCP2515_TX_CNT > 2 */
 
 		if (canintf & MCP2515_CANINTF_ERRIF) {
 			mcp2515_handle_errors(dev);


henrikbrixandersen (approved):

drivers/can/can_mcp2515.c
```diff
@@ -831,13 +831,17 @@ static void mcp2515_handle_interrupts(const struct device *dev)
 			mcp2515_tx_done(dev, 0, 0);
 		}
 
+#if MCP2515_TX_CNT > 1
```
henrikbrixandersen, line 834: Where is this set?

henrikbrixandersen (changes requested):

drivers/can/can_mcp2515.c
kaabia: In can_mcp2515.h file Line 16:

https://github.com/zephyrproject-rtos/zephyr/blob/e438b57a8caded26e90bc227edcf3329ece5241d/drivers/can/can_mcp2515.h#L16

henrikbrixandersen: That's not a Kconfig; it's a constant. Please rephrase the commit log and PR description accordingly.

kaabia: Done :)

henrikbrixandersen (approved):

str4t0m (approved):

PR #98733, opened 2025-11-02
lib: os: mpsc_pbuf: fix potential semaphore wait forever

fei315412-cmyk:
One thread calls mpsc_pbuf_alloc to produce data, which invokes add_skip_item and steps into k_sem_take.

Another thread calls mpsc_pbuf_claim to consume data. In this condition, mpsc_pbuf_claim has only small remaining space and needs to call rd_idx_inc to reserve space, but there is still no data available.

The consumer should call k_sem_give to wake mpsc_pbuf_alloc again, so the producer can allocate space and continue producing data.

Without this wake-up, the producer thread may wait forever in k_sem_take, leading to a deadlock situation.

### DIFF
--- a/lib/os/mpsc_pbuf.c
+++ b/lib/os/mpsc_pbuf.c
@@ -537,6 +537,7 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 {
 	union mpsc_pbuf_generic *item;
 	bool cont;
+	bool need_post = false;
 
 	do {
 		uint32_t a;
@@ -562,6 +563,7 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 				      idx_inc(buffer, buffer->tmp_rd_idx, inc);
 				rd_idx_inc(buffer, inc);
 				cont = true;
+				need_post = true;
 			} else {
 				item->hdr.busy = 1;
 				buffer->tmp_rd_idx =
@@ -575,6 +577,10 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 		}
 		k_spin_unlock(&buffer->lock, key);
 	} while (cont);
+
+	if (IS_ENABLED(CONFIG_MULTITHREADING) && need_post && item == NULL) {
+		k_sem_give(&buffer->sem);
+	}
 
 	return item;
 }
--- a/tests/lib/mpsc_pbuf/src/main.c
+++ b/tests/lib/mpsc_pbuf/src/main.c
@@ -1034,6 +1034,140 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
 	}
 }
 
+static uint32_t buf32_1[128];
+struct test_msg_data {
+	union mpsc_pbuf_generic hdr;
+	uint32_t buf_len; /*contain buf_len and buf*/
+	uint8_t buf[sizeof(buf32_1)];
+};
+struct test_msg_hnd {
+	struct mpsc_pbuf_buffer mpsc_buffer;
+	uint32_t product_max_cnt;
+	uint32_t product_cnt;
+	uint32_t consumer_cnt;
+};
+
+K_THREAD_STACK_DEFINE(t3_stack, 1024);
+static struct k_thread threads_t3;
+static k_tid_t tid3;
+
+static uint32_t test_mpsc_get_used_len(const union mpsc_pbuf_generic *packet)
+{
+	uint32_t size = 0;
+	struct test_msg_data *msg_data = NULL;
+
+	msg_data =  CONTAINER_OF(packet, struct test_msg_data, hdr);
+	size = msg_data->buf_len + sizeof(union mpsc_pbuf_generic);
+	size = ROUND_UP(size, sizeof(uint32_t)); /*return number  of uint32_t */
+	size = size / sizeof(uint32_t);
+	return size;
+}
+
+static void t_data_consumer_entry(void *p0, void *p1, void *p2)
+{
+	uint32_t read_cnt = 0;
+	bool wait = true;
+	const union mpsc_pbuf_generic *msg = NULL;
+	const struct test_msg_data *test_data = NULL;
+	struct test_msg_hnd *test_hnd = (struct test_msg_hnd *)p0;
+
+	while (1) {
+		if (msg == NULL) {
+			msg = mpsc_pbuf_claim(&test_hnd->mpsc_buffer);
+		}
+		test_data = (const struct test_msg_data *)msg;
+		if (test_data == NULL) {
+			continue;
+		}
+		if (test_data && test_hnd->mpsc_buffer.wr_idx == 0) {
+			wait = false;
+		}
+		if (wait == true) {
+			continue;
+		}
+		read_cnt++;
+		mpsc_pbuf_free(&test_hnd->mpsc_buffer, msg);
+		msg = NULL;
+		if (read_cnt == test_hnd->product_max_cnt) {
+			break;
+		}
+
+	}
+	test_hnd->consumer_cnt = read_cnt;
+}
+
+/* test mpsc_pbuf_alloc can get sem_take
+ * one thread product data, one thread consumer data
+ * requirement:
+ *      consumer slow process data
+ * step:
+ *  1:product data len is  0.75 of cfg.size
+ *  2:run product times
+ *
+ */
+ZTEST(log_buffer, test_sema_lock)
+{
+	struct test_msg_hnd  test_hnd = {};
+	struct test_msg_data  test_data;
+	struct test_msg_data  *item = NULL;
+	struct mpsc_pbuf_buffer_config  cfg;
+	uint32_t loop = 0;
+	size_t wlen = 0;
+	bool fist_wait = true;
+
+	if (CONFIG_SYS_CLOCK_TICKS_PER_SEC < 10000) {
+		ztest_test_skip();
+	}
+
+	cfg.buf = buf32_1;
+	cfg.size = ARRAY_SIZE(buf32_1);
+	cfg.get_wlen = test_mpsc_get_used_len;
+
+	mpsc_pbuf_init(&test_hnd.mpsc_buffer, &cfg);
+	test_hnd.product_max_cnt = 2;
+
+	tid3 = k_thread_create(&threads_t3, t3_stack, 1024,
+			t_data_consumer_entry,
+			&test_hnd, NULL, NULL,
+			10, 0, K_NO_WAIT);
+	k_thread_name_set(&threads_t3, "test_mpsc_consumer");
+	for (loop = 0; loop < test_hnd.product_max_cnt; loop++) {
+		if (loop == 0) {
+			test_data.buf_len = sizeof(buf32_1) / 2;
+		} else {
+			test_data.buf_len = sizeof(buf32_1) / 2 + sizeof(buf32_1) / 4;
+		}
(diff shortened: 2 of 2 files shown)


fei315412-cmyk:
@nordic-krch
run in qemu_x86  is good,    run qemu_x86_64/atom is bad, how to used gdb ?
west build -p -b qemu_x86_64/atom tests/lib/mpsc_pbuf -T libraries.mpsc_pbuf.concurrent
west build -t run

fei315412-cmyk:
image
this case easy find in native_sim and qemu_x86 and other platform sometimes happen.

fei315412-cmyk:
@kartben  manual trigger  Twister Status / all jobs . thanks

fei315412-cmyk:
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks
image
it means that i have not right to click  ?
@zephyrbot @kartben

fei315412-cmyk:
@nordic-krch @nashif @dcpleung  help me to  trigger  Twister Status / all jobs , thanks

nordic-krch:
@fei315412-cmyk sorry for delayed response. I'm currently overloaded with the internal release. I will try to look into that in next few days.

fei315412-cmyk:
@nordic-krch @nashif @dcpleung  for review .
  download test case to test.
  **trigger problem need two threds:**
      **produce thread:**
          1. produce thread  send first data more than 0.5 mpsc buffer size
          2. produce thread   send two  data size more than 0.5 mpsc buffer size + 4, and for to trigger mpsc_pbuf_alloc  
            and generate skip item(busy and invalid data) , and **k_sem_take** 
     **consumer thread** ：
         1. when produce thread  sem take , consumer thead begin to work.
         2.  first data will be consumer and  mpsc_pbuf_free to wakeup , however, the space  can not for 0.5 mpsc buffer size + 4,
            the produce thread will k_sem_take again
          3.  consumer thread will get null,  mpsc_pbuf_claim will execute  skip item(busy and invalid data ), not  execute  k_sem_give
  **result**: 
          produce thread step2  will be block, if not this commit

fei315412-cmyk:
@nordic-krch  @nashif @dcpleung   this commit was silence  more than  ten days.  have a slice time review it . thanks !

lib/os/mpsc_pbuf.c
```diff
@@ -576,6 +582,11 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 		k_spin_unlock(&buffer->lock, key);
 	} while (cont);
 
+#ifdef CONFIG_MULTITHREADING
+	if (need_post && item == NULL) {
+		k_sem_give(&buffer->sem);
+	}
+#endif
```
nordic-krch: Please use IS_ENABLED so that remaining ifdefs can be removed.

```suggestion
	if (IS_ENABLED(CONFIG_MULTITHREADING) && need_post && item == NULL) {
		k_sem_give(&buffer->sem);
	}
```

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+	mpsc_pbuf_init(&mpsc_buffer, &cfg);
+
+	tid3 = k_thread_create(&threads_t3, t3_stack, 1024,
+			t_data_consumer_entry,
+			&mpsc_buffer, NULL, NULL,
+			10, 0, K_NO_WAIT);
+	k_thread_name_set(&threads_t3, "test_mpsc_consumer");
+	for (loop = 0; loop < test_data_product_max_cnt; loop++) {
+		if (loop == 0) {
+			test_data.buf_len = 1600;
```
nordic-krch: Please calculate this length from buffers size. Same with `1800`.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+ * requirement:
+ *      consumer slow process data
+ * step:
+ *  1:product data len is  0.75 of cfg.size
+ *  2:run product times
+ *
+ */
+ZTEST(log_buffer, test_sema_lock)
+{
+	struct test_msg_data  test_data;
```
nordic-krch, line 1111: When I tried to run if on one of a real targets I got stack overflow because this data contains 2k buffer. Consider making it static or reducing buffer.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+		}
+		if (wait == true) {
+			continue;
+		}
+		read_cnt++;
+		mpsc_pbuf_free(buffer, msg);
+		msg = NULL;
+		if (read_cnt == test_data_product_max_cnt) {
+			break;
+		}
```
nordic-krch, line 1093: Does it need to be that complex? It seems that you want to read n buffers from the packet buffer.
Something like:
```suggestion
		msg = mpsc_pbuf_claim(buffer);
		if (msg != NULL) {
		  read_cnt++;
		  mpsc_pbuf_free(buffer, msg);
		  if (read_cnt == test_data_product_max_cnt) {
			conf = false;
		  }
	  }
```

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+{
+	struct test_msg_data *msg_data = NULL;
+
+	msg_data = CONTAINER_OF(packet, struct test_msg_data, hdr);
+
+	PRINT("msg data len = %d\n", msg_data->buf_len);
+	/*never drop*/
+}
+
+uint32_t
```
nordic-krch: `static` missing and redundant line break. 100 character is the limit.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+static struct mpsc_pbuf_buffer mpsc_buffer;
+
+static uint32_t test_data_product_max_cnt = 2;
+static uint32_t test_data_product_cnt;
+static uint32_t test_data_consumer_cnt;
+K_THREAD_STACK_DEFINE(t3_stack, 1024);
+static struct k_thread threads_t3;
+static k_tid_t tid3;
+
+void
```
nordic-krch: `static` missing and redundant line break. 100 character is the limit.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+
+	msg_data =  CONTAINER_OF(packet, struct test_msg_data, hdr);
+
+	size = msg_data->buf_len + sizeof(union mpsc_pbuf_generic);
+	size = ROUND_UP(size, sizeof(uint32_t)); /*return number  of uint32_t */
+	size = size / sizeof(uint32_t);
+	return size;
+}
+
+void t_data_consumer_entry(void *p0, void *p1, void *p2)
```
nordic-krch: static

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
 }
 
+static uint32_t buf32_1[512];
+struct test_msg_data {
+	union mpsc_pbuf_generic hdr;
+	uint32_t buf_len; /*contain buf_len and buf*/
+	uint8_t buf[sizeof(buf32_1)];
+};
+
+static struct mpsc_pbuf_buffer mpsc_buffer;
```
nordic-krch: can be moved to the test local variables as it is passed as argument to the consumer thread.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+
+static uint32_t test_data_product_max_cnt = 2;
```
nordic-krch: all those variables could be local test variables and max_cnt and consumer_cnt can be passed to the consumer thread by argument.

tests/lib/mpsc_pbuf/src/main.c
```diff
@@ -1034,6 +1034,151 @@ void start_threads(struct mpsc_pbuf_buffer *buffer)
...
+void
+test_mpsc_notify_drop(const struct mpsc_pbuf_buffer *buffer,
+		      const union mpsc_pbuf_generic *packet)
+{
+	struct test_msg_data *msg_data = NULL;
+
+	msg_data = CONTAINER_OF(packet, struct test_msg_data, hdr);
+
+	PRINT("msg data len = %d\n", msg_data->buf_len);
+	/*never drop*/
```
nordic-krch: `zassert_true(false);` to fail if that happens.

nordic-krch (changes requested):
Thanks for the fix. It is valid. My comments below.

tests/lib/mpsc_pbuf/src/main.c
fei315412-cmyk: this want to block produce threads  and  mpsc_pbuf_alloc call k_sem_take, so need to wait 
case for produce thread block by semaphore.

tests/lib/mpsc_pbuf/src/main.c
fei315412-cmyk: buffer size will reduce to 512.

tests/lib/mpsc_pbuf/src/main.c
fei315412-cmyk: local variables will stack over, so used global variables. 
now consumer used p0 argument

tests/lib/mpsc_pbuf/src/main.c
fei315412-cmyk: this only see callback  message. notice that can not discard message.

tests/lib/mpsc_pbuf/src/main.c
fei315412-cmyk: consumer thread will used pointer argumnet to resolve parament. 
do not suggest used local varibles. case for  dirty data . thanks

nordic-krch (approved):

lib/os/mpsc_pbuf.c
```diff
@@ -537,6 +537,9 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 {
 	union mpsc_pbuf_generic *item;
 	bool cont;
+#ifdef CONFIG_MULTITHREADING
```
nordic-krch: ifdef need to be removed.

lib/os/mpsc_pbuf.c
```diff
@@ -562,6 +565,9 @@ const union mpsc_pbuf_generic *mpsc_pbuf_claim(struct mpsc_pbuf_buffer *buffer)
 				      idx_inc(buffer, buffer->tmp_rd_idx, inc);
 				rd_idx_inc(buffer, inc);
 				cont = true;
+#ifdef CONFIG_MULTITHREADING
```
nordic-krch: ifdef need to be removed.

nordic-krch (changes requested):
Please split the work into 2 commits: test + fix in mpsc_pbuf

fei315412-cmyk:
@nordic-krch @dcpleung @nashif thanks for review.
please tigger Twister Status / all jobs , thanks
ammend again.
please  review again. thank.

nordic-krch (approved):

dcpleung (approved):

PR #98734, opened 2025-11-02
kernel: macro to approximate heap overhead

JordanYates:
Add a macro to approximate the heap size required by a heap in order for an allocation of N bytes to succeed. The intent is to provide a way for small heaps to more accurately specify their sizes.
Add a testcase that outputs the overhead of heaps declared multiple different ways, by printing the largest single allocation that can be made from the heap.
```
START - test_heap_overhead
    Base |  Raw | Z_HEAP_MIN_SIZE | Z_HEAP_MIN_SIZE_FOR
      64 |   12 |              60 |                  68
     128 |   68 |             124 |                 132
     256 |  196 |             244 |                 260
     512 |  444 |             500 |                 516
    1024 |  956 |            1004 |                1028
    2048 | 1972 |            2028 |                2052
```

Use `Z_HEAP_MIN_SIZE_FOR` on the system heap. This fixes allocations failing when there is only a single small user of the heap defining a symbol like the following, even when only allocating 16 bytes.
```
config HEAP_MEM_POOL_ADD_SIZE_{X}
        int
        default 64
```

The motivation for this PR is https://github.com/nrfconnect/sdk-nrf/pull/25117. When that symbol is the only heap user in the build, allocating a single default context fails, even though the declared size is 4x the size of the requested memory. Rather than increasing that Kconfig, which only solves the problem once, this should solve the problem everywhere.

### DIFF
--- a/include/zephyr/kernel.h
+++ b/include/zephyr/kernel.h
@@ -5949,6 +5949,42 @@ void k_heap_free(struct k_heap *h, void *mem) __attribute_nonnull(1);
 #define Z_HEAP_MIN_SIZE ((sizeof(void *) > 4) ? 56 : 44)
 #endif /* CONFIG_SYS_HEAP_RUNTIME_STATS */
 
+/* Size of `struct z_heap` */
+#define _Z_HEAP_SIZE                                                                               \
+	((4 * sizeof(uint32_t)) +                                                                  \
+	 (3 * (IS_ENABLED(CONFIG_SYS_HEAP_RUNTIME_STATS) ? sizeof(size_t) : 0)))
+
+/* Number of buckets required to store @a bytes */
+#define _Z_HEAP_NUM_BUCKETS(bytes) ((31 - __builtin_clz((bytes / 8) - 1)) + 1)
+
+/* Number of bytes consumed by buckets */
+#define _Z_HEAP_BUCKETS_SIZE(bytes) (_Z_HEAP_NUM_BUCKETS(bytes) * sizeof(uint32_t))
+
+/**
+ * @brief Minimum heap size required for allocating a given size
+ *
+ * Heaps store metadata at the start of the provided array, resulting in a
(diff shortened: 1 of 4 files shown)


npitre (commented):
I'm not comfortable with the naming of `Z_HEAP_OVERHEAD(bytes)`. This 
might suggest the returned value corresponds to the heap's overhead which 
is not true except for a single particular case.

The overhead depends a lot on the number of allocations and their sizes.
For example, if you fill up the heap with 4-byte allocations, the overhead
will be above 100%.

Maybe `Z_HEAP_MGMT_BYTES(alloc_size)` but still not very good.

Considering a macro that combines both the allocation size and the
"overhead" size: you could use `Z_HEAP_MIN_SIZE_FOR(alloc_size)` which 
I think is the least wrong name I can think of. So instead of:

```
K_HEAP_DEFINE(_system_heap, K_HEAP_MEM_POOL_SIZE + Z_HEAP_OVERHEAD(K_HEAP_MEM_POOL_SIZE));
```

you would have:

```
K_HEAP_DEFINE(_system_heap, Z_HEAP_MIN_SIZE_FOR(K_HEAP_MEM_POOL_SIZE));
```

JordanYates:
Updated with suggested name of `Z_HEAP_MIN_SIZE_FOR`

peter-mitsis (approved):

kartben:
given this will effectively cause increased RAM usage and can catch people off guard, should this be mentioned in the migration guide so that folks have an explanation and know they can slightly lower the size of their memory pool to go back to previous RAM usage if whatever value they had happened to work fine for their use case?

ceolin (approved):

JordanYates:
> @JordanYates any comments/thoughts on this? Thanks!

Personally I think increasing the RAM usage by a few dozen bytes on a new Zephyr release will probably be in the noise for most people. I can add something if you want though.

PR #98738, opened 2025-11-02
boards: st: stm32u5g9j_dk1: Enable STTS22H and add ambient-temp alias

CharlesDias:
Add STTS22H node at I2C3 with node label stts22h_temperature and add ambient-temp alias.

Testing with:
`west build -p -b stm32u5g9j_dk1 samples/sensor/stts22h/`

Related to:
- https://github.com/zephyrproject-rtos/zephyr/pull/98739

PR #98739, opened 2025-11-02
samples: sensor: stts22h: Minimal polling sample

CharlesDias:
Add a simple STTS22H sample that utilizes the Zephyr sensor API in polling mode to fetch and print the ambient temperature every 2 seconds periodically.

Testing with:
`west build -p -b stm32u5g9j_dk1 samples/sensor/stts22h/`

Related to:
- https://github.com/zephyrproject-rtos/zephyr/pull/98738


samples/sensor/stts22h/prj.conf
```diff
@@ -0,0 +1,13 @@
+CONFIG_STDOUT_CONSOLE=y
+CONFIG_I2C=y
```
JarmouniA:
```suggestion
```
Enabled by the driver Kconfig
https://github.com/zephyrproject-rtos/zephyr/blob/e438b57a8caded26e90bc227edcf3329ece5241d/drivers/sensor/st/stts22h/Kconfig#L11

samples/sensor/stts22h/README.rst
```diff
@@ -0,0 +1,49 @@
...
+References
+**********
+
+ - STTS22H: https://www.st.com/en/mems-and-sensors/stts22h.html
+
+Building and Running
+********************
+
+This project outputs STTS22H temperature data to the console. It requires a
+board that has an STTS22H connected over I2C and enabled in its devicetree.
```
JarmouniA:
```suggestion
This sample outputs STTS22H temperature data to the console.
```
Redundant with the above.

samples/sensor/stts22h/README.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Examples of supported boards include STM32U5G9J-DK1, where the sensor is
+enabled at I2C3 address 0x3F in the board DTS.
```
JarmouniA:
```suggestion
Supported boards: STM32U5G9J-DK1.
```

samples/sensor/stts22h/prj.conf
CharlesDias: Done

samples/sensor/stts22h/README.rst
CharlesDias: Done

samples/sensor/stts22h/README.rst
CharlesDias: Done

samples/sensor/stts22h/sample.yaml
```diff
@@ -0,0 +1,12 @@
+sample:
+  name: STTS22H temperature sensor
+tests:
+  sample.sensor.stts22h:
```
JarmouniA, line 4: please restore the compatible-based filter.

samples/sensor/stts22h/sample.yaml
```diff
@@ -0,0 +1,12 @@
...
+    depends_on:
+      - i2c
+    tags:
+      - sensors
+    integration_platforms:
```
JarmouniA, line 10: 1 integration platform is sufficient (this is not an allow list, won't prevent the sample from being built on other platforms that satisfy the filter).
https://docs.zephyrproject.org/latest/develop/test/twister.html#tests

samples/sensor/stts22h/sample.yaml
CharlesDias: Restored.

samples/sensor/stts22h/sample.yaml
CharlesDias: Thank you! I will take a closer look at the tests topic.

JarmouniA (dismissed):

MaureenHelm (changes requested):
Would prefer not to introduce a new driver-specific sensor sample unless there is a specific need, such as custom channels or attributes. Please have a look at using `samples/sensor/thermometer` instead.

JarmouniA:
> Would prefer not to introduce a new driver-specific sensor sample unless there is a specific need, such as custom channels or attributes. Please have a look at using `samples/sensor/thermometer` instead.

Agreed, didn't see that one.

PR #98741, opened 2025-11-02
Actuator device driver class

bjarki-andreasen:
Introduce the actuator device driver class.

This class is designed for devices which can displace an object, or apply force to an object, by a set amount, within a fixed range.

Examples are:

- A servo motor, rotating an arm to a set angle
- An open loop motor controller, applying a set torque to an axel
- A closed loop motor controller, rotating an axel by a set angular velocity
- A linear actuator, moving its arm to a set position.

The setpoint of the actuator is provided as a value from -1 to 1 in Q31 notation.

Controlling a servo is as simple as:
```
const struct device *servo = DEVICE_DT_GET(DT_ALIAS(servo));

int main(void)
{
        actuator_set_setpoint(servo, 0); /* Move to 0 rotation */
        k_msleep(1000); /* Give servo time to move */
        actuator_set_setpoint(servo, SYS_Q31_MILLI(-1000)); /* Move to most negative rotation */
        k_msleep(1000); /* Give servo time to move */
        actuator_set_setpoint(servo, SYS_Q31_MILLI(1000)); /* Move to most positive rotation */
}
```

Along with the driver class are shell commands, a fake driver, a pwm driven servo driver, and a set of helpers to make working with Q31 notation trivial.

The Q31 notation is chosen as:
* It matches the capabilities of an actuator, which works within a fixed range from -1 to 1, -1 being the max displacement/force in 1 direction, 0 being no displacement/force, and 1 being the max displacement/force in the other direction.
* It provides the highest resolution possible given a 32 bit data type
* Its unitless, making it portable between multiple actuator types
* It matches the sensor subsystem which also uses Q31 notation, and any potential fixed point control system subsystem would likely also use Q31 notation (like this one https://github.com/bjarki-andreasen/zephyr-virtual-board/tree/main/include/zvb/control_system which I plan to upstream as well)

We could consider adding callbacks for when an actuator reaches a setpoint, or reading its process_var if it has one, for now the DC motor controllers, servos and ESCs I have used don't have any such feedback.


drivers/actuator/Kconfig
```diff
@@ -0,0 +1,23 @@
...
+module-str = actuator
+source "subsys/logging/Kconfig.template.log_config"
+
+config ACTUATOR_INIT_PRIORITY
+	int "Actuator init priority"
+	default KERNEL_INIT_PRIORITY_DEVICE
+	help
+	  Actuator device driver initialization priority.
+
+rsource "Kconfig.fake"
```
faxe1008, line 21: Maybe keep sorted here as well

drivers/actuator/Kconfig.shell
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+config ACTUATOR_SHELL
+	bool "PWM driven servo device driver"
```
faxe1008: C&P typo :^)

drivers/actuator/actuator_shell.c
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+#include <zephyr/drivers/actuator.h>
+#include <zephyr/sys/q31.h>
+#include <zephyr/shell/shell.h>
+#include <zephyr/kernel.h>
+
+#include <stdlib.h>
+
+#define SETPOINT_MAX 1000
```
faxe1008, line 14: Would it maybe make sense to have those as a property on the actuator base binding?

EDIT:// hmm I see that you add the domain specific range properties to the actual driver binding. Maybe an api for getting these values would make sense? Something like `get_setpoint_range`

faxe1008:
A `build_all` test would also be nice I think.

drivers/actuator/actuator_shell.c
bjarki-andreasen: The setpoint range is always -1 to 1, that could be -10 degrees to 10 degrees, 0mm position to 30mm position, the max here is 1000 because the shell allows setting the setpoint in 1/1000 parts (milli), so setting 1000 is 1000/1000 == 1.

A sensor with higher precision could be set in 1/1000000 (micro) or even 1/1000000000 (nano), in which case the max value provided to the shell would be 1000000 or 1000000000 respectively, but for the shell, 1/1000 is probably high enough.

include/zephyr/sys/q31.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * @retval Value in Q31 notation
+ */
+#define SYS_Q31_NANO(_nano)									\
+	(											\
+		(_nano) == 0									\
+			? 0									\
+			: (_nano) < 0								\
+				? ((_nano) * 0x80000000LL) / 1000000000LL			\
+				: ((_nano) * 0x7FFFFFFFLL) / 1000000000LL			\
```
fabiobaltieri, line 42: would it make sense to use `if`s here to make it more readable? guess this form can be used in macros, just checdking though

fabiobaltieri (commented):
Cool, you could probably start by using this in `samples/basic/servo_motor/src/main.c`?

include/zephyr/sys/q31.h
bjarki-andreasen: Its for "macroability" so it can be used to initialize a Q31 in a driver

```
#define DRIVER_INST_DEFINE(inst)								\
												\
	static struct driver_data data##inst = {						\
		.pulse_ns = SYS_Q31_RANGE(0,							\
					  DT_INST_PROP(inst, pulse_min_ns),			\
					  DT_INST_PROP(inst, pulse_max_ns)),			\
	};											\
```
I considered inline functions as well, we can have those to, but the macro works in both cases

drivers/actuator/Kconfig.shell
bjarki-andreasen: fixed

drivers/actuator/Kconfig
bjarki-andreasen: now sorted

bjarki-andreasen:
> Cool, you could probably start by using this in `samples/basic/servo_motor/src/main.c`?

Looking at the servomotor sample, it is specifically for pwm driven servos, where a sample demoing the actuator API could control any actuator device, driven by any signal, maybe we could simply move the sample to samples/drivers/pwm and create a new sample for actuators? Trying to adapt the sample to use the actuator API changes it drastically to the point where it really is a new sample

dts/bindings/actuator/zephyr,servo-pwm.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+description: PWM driven servo device
+
+include:
+  - actuator-device.yaml
```
jilaypandya: Nit: `include: actuator-device.yaml`

include/zephyr/drivers/actuator.h
```diff
@@ -0,0 +1,51 @@
...
+#include <zephyr/dsp/types.h>
+#include <errno.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @cond INTERNAL_HIDDEN */
+
+typedef int (*actuator_api_set_setpoint)(const struct device *dev, q31_t setpoint);
```
jilaypandya:
```suggestion
typedef int (*actuator_api_set_setpoint_t)(const struct device *dev, q31_t setpoint);
```

include/zephyr/sys/q31.h
```diff
@@ -0,0 +1,139 @@
...
+#define SYS_Q31_DEC(_dec) \
+	SYS_Q31_MILLI((_dec) * 1000LL)
+
+/**
+ * @brief Scale value in Q31 notation (-1 to 1) to specified integer scale
+ *
+ * @details
+ *
+ * Examples:
+ *
```
jilaypandya, line 111: https://github.com/zephyrproject-rtos/zephyr/blob/73b6f8bd4a4bde38c8678d750793d02ae3f43c14/include/zephyr/drivers/sensor.h#L692

can we put the example in code block

tests/drivers/actuator/shell/testcase.yaml
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+tests:
+  drivers.actuator.shell:
+    platform_allow:
+      - native_sim
+      - native_sim/native/64
+    tags:
+      - drivers
+      - comparator
```
jilaypandya:
```suggestion
      - actuator
```

jilaypandya:
Really like the approach :-)

fabiobaltieri:
> Looking at the servomotor sample, it is specifically for pwm driven servos, where a sample demoing the actuator API could control any actuator device, driven by any signal, maybe we could simply move the sample to samples/drivers/pwm and create a new sample for actuators? Trying to adapt the sample to use the actuator API changes it drastically to the point where it really is a new sample

Well you could probably just drop it entirely once this is merged. Good stuff I was a bit puzzled by the very generic nature of the proposal but I think it's not a bad idea.

dts/bindings/actuator/actuator-device.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+include:
+  - base.yaml
```
rruuaanng:
```suggestion
include: [base.yaml]
```

dts/bindings/actuator/zephyr,actuator-fake.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+description: Fake actuator device
+
+compatible: zephyr,actuator-fake
+
+include:
+  - actuator-device.yaml
```
rruuaanng:
```suggestion
include: [actuator-device.yaml]
```

bjarki-andreasen:
@nandojve 

An actuator as defined in this PR as is a component which displaces an object, or applies a force, by a specific amount, within an absolute range. The actuator is an independent device which has its own internal control system. The actuator is an output.

An actuator can be part of a system, that system may be a control system, which defines its own input and output, where the actuator is an output, and some sensor is the input. For example, if a servo motor, which is an actuator, moves the aileron of an airplane, and a gyro is used to measure the roll rate, the servo does not stop being an actuator, nor is the servo is an external agent in this system, its very much at the heart of it.

Sensors are not actuators, sensors measure displacement or force, actuators apply displacement or force.

include/zephyr/sys/q31.h
bjarki-andreasen: The example is not pure C given the `->` and `@code(.c) ` will not let me add "nested" comments
```
/*
 * @code(.c)
 *         SYS_Q31_SCALE(SYS_Q31_DEC(1), 500)  /* 500 */ <- terminates comment here...
 */
```
so not sure how to do it prettier

drivers/actuator/actuator_servo_pwm.c
```diff
@@ -0,0 +1,122 @@
...
+
+	return pwm_set_pulse_dt(&dev_config->pwm_spec, dev_data->pulse_ns);
+}
+
+static int driver_pm_suspend(const struct device *dev)
+{
+	const struct driver_config *dev_config = dev->config;
+	int ret;
+
+	ret = pwm_set_pulse_dt(&dev_config->pwm_spec, 0);
```
jilaypandya, line 66: The last setpoint might not be retained in this case, or?

jilaypandya (approved):

drivers/actuator/actuator_servo_pwm.c
jilaypandya: When regarded independent of the system, the actuator in itself will retain the setpoint

bjarki-andreasen: it should be, stopping the pwm will disable the servo (at least the one I tested with), the setpoint which is set once the device is resumed is stored in `dev_data->pulse_ns` which is retained

include/zephyr/drivers/actuator.h
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @brief Set actuator setpoint
+ *
+ * @param dev Actuator device
+ * @param setpoint Setpoint in Q31 format
+ *
+ * @retval 0 Successful
+ * @retval -errno code Failure
+ */
+__syscall int actuator_set_setpoint(const struct device *dev, q31_t setpoint);
```
kartben, line 47: for actuators with feedback, I think _getting_ the setpoint would make sense (not sure that's called a "*set*point" in this case though!)

bjarki-andreasen: It would be the process variable, `process_var` or `pv` in short.

jilaypandya: The `actuator` with integrated feedback would be using some sort of potentiometer or hall sensor internally, or?. So the feedback would come actually from a `sensor` device. IMO we should limit the scope of the `actuator` to just `actuating`.

kartben: > The `actuator` with integrated feedback would be using some sort of potentiometer or hall sensor internally, or?. So the feedback would come actually from a `sensor` device. IMO we should limit the scope of the `actuator` to just `actuating`.

That's fair but that kind of requires people to then come up with MFD drivers and the like, which might be overkill. I do think getting the "position" should be part of the API, even if there will be cases where the way to get the position will be by reading a sensor (in which case maybe the DT node for an actuator could optionally reference a sensor device?).

Also if one were to be interested in implementing some kind of controller for the actuator, I think it would make sense that they would get both expected/set as well as actual "position" from the same device?

FWIW my comment was triggered by having played with smart control servos recently where it would _really_ feel overkill to have to come up with split actuator/sensor drivers. https://files.seeedstudio.com/wiki/robotics/Actuator/feetech/Communication_Protocol_Manual.pdf

mbolivar (approved):

PR #98742, opened 2025-11-02
samples: video: capture: fix index increment in caps display

### LINKED ISSUE #98744: drivers: video: wrong index increment in caps display
### Describe the bug

In the video capture sample, the list of frame interval is skipping entries, leading to wrong frame interval list being reported.

### Regression

- [x] This is a regression.

### Steps to reproduce

`west build --board mimxrt1170_evk/mimxrt1176/cm7 --shield nxp_btb44_ov5640 samples/drivers/video/capture`

or any other board that supports a sensor with multiple frame rates.

### Relevant log output

```shell
...
[00:00:00.000,000] <inf> main: - Supported frame intervals for the default format:
[ missing entries here ]
...
```

### Impact

### Environment

- OS: Linux
- Toolchain: Zephyr SDK 0.17.2
- Commit SHA: e438b57a8ca (`v4.3.0-rc2`)

### Additional Context

`video-sw-generator` cannot be used to reproduce the problem as supporting only a single framerate at index 0 which is correctly listed.

avolmat-st:
Since the i index is already being incremented as part of the for loop statement, the i++ done as part of the loop body should not be done since this lead to skip of entries of the table and moreover might lead to invalid access outside of the table.

### DIFF
--- a/samples/drivers/video/capture/src/main.c
+++ b/samples/drivers/video/capture/src/main.c
@@ -170,7 +170,6 @@ static int app_query_video_info(const struct device *const video_dev,
 			VIDEO_FOURCC_TO_STR(fcap->pixelformat),
 			fcap->width_min, fcap->width_max, fcap->width_step,
 			fcap->height_min, fcap->height_max, fcap->height_step);
-		i++;
 	}
 
 	/* Get default/native format */


josuah (approved):
Ouch! Thanks for fixing my bug!

JarmouniA:
Needs an accompanying bug report.

josuah:
On it...

ngphibang (approved):

PR #98749, opened 2025-11-02
lib: libc: common: source: signal: added raise function

Spago123:
Implemnted raise function by using using the k_oops kernel function to terminate the running thread, and logging to log the error to the debug interface

### DIFF
--- a/lib/libc/common/source/signal/signal.c
+++ b/lib/libc/common/source/signal/signal.c
@@ -5,6 +5,10 @@
 
 #include <errno.h>
 #include <signal.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(libc_raise, LOG_LEVEL_DBG);
 
 void (*signal(int signum, void (*handler)(int)))(int)
 {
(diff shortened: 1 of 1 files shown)


aescolar (commented):
@Spago123 note that nothing builds this file in tree.
What is your use case?

Spago123:
> @Spago123 note that nothing builds this file in tree. What is your use case?

Yup, I can see that nothing is building any of these common libc functions.
The use case is to pass a termination signal and stop the exucution of an thread with debug logging. In case of SIGTERM a graceful shutdown of the thread is required (the same as the libraray function).

keith-packard:
The C library should already be providing `signal` and `raise` implementations, so this shouldn't be needed except in places where the C library is broken. A relevant question is what each C library does in the `SIG_DFL` case. Zephyr should specify what that is and have each C library support code handle it.

In the case of picolibc, when a signal is set to `SIG_DFL`, the raise function calls `_exit(128+signal)`. The Zephyr implementation of that function for picolibc is not terribly useful; it calls `printf("exit\n");` and then spins in a loop. If you want that to cause the current thread to exit, you could provide your own implementation of that function or specify your own signal handlers.

Spago123:
> The C library should already be providing `signal` and `raise` implementations, so this shouldn't be needed except in places where the C library is broken. A relevant question is what each C library does in the `SIG_DFL` case. Zephyr should specify what that is and have each C library support code handle it.
> 
> In the case of picolibc, when a signal is set to `SIG_DFL`, the raise function calls `_exit(128+signal)`. The Zephyr implementation of that function for picolibc is not terribly useful; it calls `printf("exit\n");` and then spins in a loop. If you want that to cause the current thread to exit, you could provide your own implementation of that function or specify your own signal handlers.

@keith-packard that is a good point. This implementation can than be dropped, and the behaviour of picolibc can be checked and a thread terminatiom can be implementned there if needed.

keith-packard:
> @keith-packard that is a good point. This implementation can than be dropped, and the behaviour of picolibc can be checked and a thread terminatiom can be implementned there if needed.

That particular implementation is actually provided by Zephyr itself in the picolibc support code, so if we want _exit to do thread termination, that's the place to do it.

PR #98750, opened 2025-11-02
OV5640 sensor driver various improvements (cleanup / dvp polarities / frmival ..)

avolmat-st:
This series is still a WIP to perform various cleanup in the DVP context in order to integrate it within the overall driver and avoiding having too many dvp specific registers settings.
Among the modifications, addition of the DVP polarity control via the DT, support of the frmival apis, removal of all DVP specific ISP configuration ...

With the removal of dedicated ISP configuration done within the dvp init table, there might be difference in the image quality compared to previously but this should be address on the overall driver context rather than just the DVP side.

Still few bits to do, YUV support in DVP, more framerates / resolutions ..

Regarding YUV, it is possible to supply add the YUV in the caps for DVP, however it seems that this leads to UYVY instead of YUYV. I do not have the CSI variant of the OV5640 at hand now to confirm that in CSI mode the format is well YUYV and not as well UYVY. The difference is actually simply to set the control register to 0x30 or 0x3f but I'd expect this to not be interface specific hence I am wondering how it is currently in MIPI version.


drivers/video/ov5640.c
```diff
@@ -132,9 +137,12 @@ struct ov5640_config {
 	int bus_type;
 	int bus_width;
 	int data_shift;
+	bool pclk_sample;
+	bool hsync_active;
+	bool vsync_active;
```
josuah, line 154: Given these are not expected to change, how about bitfields to save some few bytes?
```suggestion
	uint8_t pclk_sample : 1;
	uint8_t hsync_active : 1;
	uint8_t vsync_active : 1;
```

drivers/video/ov5640.c
```diff
@@ -1434,6 +1379,9 @@ static int ov5640_init(const struct device *dev)
 		.bus_type = OV5640_EP_PROP_OR(n, bus_type, VIDEO_BUS_TYPE_CSI2_DPHY),              \
 		.bus_width = OV5640_EP_PROP_OR(n, bus_width, 10),                                  \
 		.data_shift = OV5640_EP_PROP_OR(n, data_shift, 0),                                 \
+		.pclk_sample = OV5640_EP_PROP_OR(n, pclk_sample, 0),                               \
+		.hsync_active = OV5640_EP_PROP_OR(n, hsync_active, 0),                             \
+		.vsync_active = OV5640_EP_PROP_OR(n, vsync_active, 0),                             \
```
josuah, line 1493: Maybe some day we have a `struct video_endpoint_dt_spec` to automate all of it? :)

drivers/video/ov5640.c
```diff
@@ -586,23 +586,23 @@ static const struct ov5640_mode_config dvp_modes[] = {
 		.height = 120,
 		.array_size_res_params = ARRAY_SIZE(dvp_160x120_res_params),
 		.res_params = dvp_160x120_res_params,
-		.max_frmrate = OV5640_60_FPS,
+		.max_frmrate = OV5640_30_FPS,
```
josuah: So there was actually a bug there?

josuah (dismissed):
This is great, this slowly turns the driver from a "big blocks of registers" to a "full-featured" version.

I did not get time to test it, but will try again on the next test batch as I get time for it.

drivers/video/ov5640.c
avolmat-st: Good idea. However maybe in such case we'd have to provide a macro allowing to indicate the default values is property is not given since the defaults are not part of video-interfaces.yaml, this would make the macro a bit heavy with lots of arguments ..
This would also be something to be used maybe only for drivers which really needs to tackle all/most of those parameters since such structure will embeds lots of csi / parallel properties which some driver might not actually need.

drivers/video/ov5640.c
avolmat-st: Yes, in the commit I fix this, there is actually no frmival handling for DVP, so max_frmrate is equal to def_frmrate as well. Tables for DVP based frmival are only introduced on the next commit and only 15 and 30fps is supported.

drivers/video/ov5640.c
josuah: Not critical/blocking so resolving this.

drivers/video/ov5640.c
josuah: Thanks for confirming, glad this got sorted out.

drivers/video/ov5640.c
```diff
@@ -710,12 +721,13 @@ static int ov5640_set_frmival(const struct device *dev, struct video_frmival *fr
...
-		{SC_PLL_CTRL1_REG, drv_data->cur_mode->mipi_frmrate_config[ind].pllCtrl1},
-		{SC_PLL_CTRL2_REG, drv_data->cur_mode->mipi_frmrate_config[ind].pllCtrl2},
+		{SC_PLL_CTRL1_REG, drv_data->cur_mode->frmrate_config[ind].pllCtrl1},
+		{SC_PLL_CTRL2_REG, drv_data->cur_mode->frmrate_config[ind].pllCtrl2},
 		{PCLK_PERIOD_REG, 0x0a}};
 
 	ret = video_write_cci_multiregs16(&cfg->i2c, frmrate_params, ARRAY_SIZE(frmrate_params));
-	ret |= video_modify_cci_reg(&cfg->i2c, OV5640_REG8(SC_PLL_CTRL0_REG), 0x0f, MIPI_BIT_MODE);
+	ret |= video_modify_cci_reg(&cfg->i2c, OV5640_REG8(SC_PLL_CTRL0_REG), 0x0f,
+				    ov5640_is_dvp(dev) ? 0x0a : MIPI_BIT_MODE);
```
ngphibang: nit: Maybe defining a macro for 0x0a (equivalent to MIPI_BIT_MODE) would be cleaner ?

ngphibang (dismissed):
Thanks alot for this clean-up and the implementation of the TODO "DVP frame rate setting". I tested CSI mode on RT1170, no regression found.

In the future, when you suport UYVY / YUYV mode, I can verify the format if needed.

drivers/video/ov5640.c
```diff
@@ -493,136 +479,165 @@ static const struct video_reg16 init_params_dvp[] = {
...
-static const struct video_reg16 dvp_480x272_res_params[] = {
-	{0x3800, 0x00}, {0x3801, 0x08}, {0x3802, 0x00}, {0x3803, 0x02}, {0x3804, 0x0a},
-	{0x3805, 0x37}, {0x3806, 0x07}, {0x3807, 0xa1}, {0x3808, 0x01}, {0x3809, 0xe0},
-	{0x380a, 0x01}, {0x380b, 0x10}, {0x380c, 0x06}, {0x380d, 0x14}, {0x380e, 0x03},
-	{0x380f, 0xe8}, {0x3810, 0x00}, {0x3811, 0x04}, {0x3812, 0x00}, {0x3813, 0x79},
-	{0x3814, 0x31}, {0x3815, 0x31}, {0x3820, 0x47}, {0x3821, 0x01}, {0x4602, 0x01},
-	{0x4603, 0xe0}, {0x4604, 0x01}, {0x4605, 0x10}};
+static const struct ov5640_frmrate_config dvp_frmrate_params[] = {
+	/* pixelrate is not used in DVP mode and 60fps is not supported */
+	{15, 0x21, 0x64, 0}, {30, 0x11, 0x64, 0}, {60, 0x11, 0x64, 0}};
```
josuah: Here 60 FPS is still in the list despite seeing "60 FPS is not supported", is this expected?

Maybe because it is used in the for loop below, just to avoid overflowing out of `ov5640_frame_rates`, with always 3 values:
```
for (i = 0; i < ARRAY_SIZE(ov5640_frame_rates); i++) {
```

josuah (commented):
Just letting you confirm but LGTM, thank you for the deduplication!

drivers/video/ov5640.c
avolmat-st: yes, that was the reason why. I'd probably put a new element in the structure to hold the array size and verify to not access to invalid index.

avolmat-st:
A future improvement necessary for this driver would be to be able to compute lowest acceptable clock settings in order to achieve requested resolution / framerate. This, combined with a new bindings such as max-pixel-clock or so could be used in order to indicate the maximum pixel rate achievable from a receiver point of view so that the sensor can try to achieve the requested settings. Something a bit similar to what we have already for link-frequency tables in case of CSI-2.
But, just dropping that here for info, something for a future PR.

ngphibang (dismissed):

drivers/video/ov5640.c
```diff
@@ -1365,6 +1417,16 @@ static int ov5640_init(const struct device *dev)
...
+		/* Set DVP polarity */
+		ret = video_write_cci_reg(&cfg->i2c, OV5640_REG8(POLARITY_CTRL_REG),
+					  (cfg->pclk_sample ? POLARITY_PCLK_HIGH : 0) |
+					  (cfg->hsync_active ? POLARITY_HSYNC_HIGH : 0) |
+					  (cfg->vsync_active ? 0 : POLARITY_VSYNC_LOW));
+		if (ret) {
+			LOG_ERR("Unable to set DVP polarity");
+			return -EIO;
+		}
 	} else {
```
josuah, line 1435: CI complains this function is too long, how about moving this to an `ov564o_init_dvp()`?

josuah (dismissed):

drivers/video/ov5640.c
josuah: It looks like it is treated as a warning, so might be possible to ignore. It makes sense to have long `_init()` function like a long list of function calls: not that complex IMHO.

josuah:
This is reviewed but @avolmat-st added a DNM, so maybe it needs some extra changes, removing stale label in the meantime.

drivers/video/ov5640.c
```diff
@@ -493,136 +479,165 @@ static const struct video_reg16 init_params_dvp[] = {
...
-static const struct video_reg16 dvp_480x272_res_params[] = {
-	{0x3800, 0x00}, {0x3801, 0x08}, {0x3802, 0x00}, {0x3803, 0x02}, {0x3804, 0x0a},
-	{0x3805, 0x37}, {0x3806, 0x07}, {0x3807, 0xa1}, {0x3808, 0x01}, {0x3809, 0xe0},
-	{0x380a, 0x01}, {0x380b, 0x10}, {0x380c, 0x06}, {0x380d, 0x14}, {0x380e, 0x03},
-	{0x380f, 0xe8}, {0x3810, 0x00}, {0x3811, 0x04}, {0x3812, 0x00}, {0x3813, 0x79},
-	{0x3814, 0x31}, {0x3815, 0x31}, {0x3820, 0x47}, {0x3821, 0x01}, {0x4602, 0x01},
-	{0x4603, 0xe0}, {0x4604, 0x01}, {0x4605, 0x10}};
+static const struct ov5640_frmrate_config dvp_frmrate_params[] = {
+	/* pixelrate is not used in DVP mode */
+	{15, 0x41, 0x60, 0}, {30, 0x21, 0x60, 0}, {60, 0x11, 0x60, 0}};
```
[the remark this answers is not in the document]
avolmat-st: This requires to measure or the pixelrate so I propose to leave this as a TODO and put some value for the time being.

drivers/video/ov5640.c
```diff
@@ -1433,6 +1483,9 @@ static int ov5640_init(const struct device *dev)
 		.bus_type = OV5640_EP_PROP_OR(n, bus_type, VIDEO_BUS_TYPE_CSI2_DPHY),              \
 		.bus_width = OV5640_EP_PROP_OR(n, bus_width, 10),                                  \
 		.data_shift = OV5640_EP_PROP_OR(n, data_shift, 0),                                 \
+		.pclk_sample = OV5640_EP_PROP_OR(n, pclk_sample, 0),                               \
+		.hsync_active = OV5640_EP_PROP_OR(n, hsync_active, 0),                             \
```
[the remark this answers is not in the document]
avolmat-st: All boards have been updated to embed those new properties

boards/shields/arducam_cu450_ov5640/arducam_cu450_ov5640_dvp.overlay
```diff
@@ -27,6 +27,9 @@
 			ov5640_ep_out: endpoint {
 				remote-endpoint-label = "arducam_ffc_40pin_ep_in";
 				bus-type = <VIDEO_BUS_TYPE_PARALLEL>;
+				hsync-active = <0>;
+				vsync-active = <0>;
```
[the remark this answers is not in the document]
avolmat-st: Indeed, both ov5640 side and platform side synchro signal should match. That said I put the new properties in the arducam_cu450_ov5640 shield so that same value as before is generated.
It would be great to give a try with this PR applied on  this platform and fix appropriately

drivers/video/ov5640.c
```diff
@@ -794,6 +805,29 @@ static int ov5640_set_fmt(const struct device *dev, struct video_format *fmt)
...
+			};
+
+			ret = video_write_cci_multiregs16(&cfg->i2c, pclk_divider,
+							  ARRAY_SIZE(pclk_divider));
+		}
+		if (ret) {
+			LOG_ERR("Unable to set resolution parameters");
+			return ret;
+		}
 	}
```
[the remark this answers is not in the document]
avolmat-st: Indeed, fixed.

avolmat-st:
Update pushed with 2 improvements based on Github review points.

avolmat-st:
PR simply rebased since it was based on a quite old sha1.

drivers/video/ov5640.c
```diff
@@ -493,136 +479,165 @@ static const struct video_reg16 init_params_dvp[] = {
...
-	{0x380f, 0xe4}, {0x3810, 0x00}, {0x3811, 0x10}, {0x3812, 0x00}, {0x3813, 0x04},
-	{0x3814, 0x31}, {0x3815, 0x31}, {0x3824, 0x04}, {0x460c, 0x20}};
-
-static const struct ov5640_mipi_frmrate_config mipi_hd_frmrate_params[] = {
+/*
+ * Helper macro allow to fill-in a struct video_reg16, avoiding usage of struct video_reg
+ * which would be most costly in code size in such case
+ */
+#define OV5640_REG16_DATA16_BE(addr, value)	\
+	{(addr), ((value) & 0xff00) >> 8}, {(addr) + 1, (value) & 0x00ff}
```
josuah, line 486: Is this really saving though?

16-bit address, 8-bit data, 8-bit padding,
16-bit address, 8-bit data, 8-bit padding,
=> 64-bit with `struct video_reg16`

bytes for storing one 16-bit register with `struct video_reg`
8-bit flags, 8-bit padding, 16-bit address,
16-bit padding, 16-bit data
=> 64-bit with `struct video_reg`

But there is still some saving due to the last two registers, so you win :)

drivers/video/ov5640.c
```diff
@@ -479,8 +466,7 @@ static const struct video_reg16 init_params_dvp[] = {
 	{0x568f, 0x22},
 	{0x5025, 0x00},
 	{0x3406, 0x00},
-	{0x3503, 0x00},
-	{0x3008, 0x02},
```
josuah, line 482: It was powering-up automatically (0x02), so now it waits set-stream: power savings for free!

drivers/video/ov5640.c
```diff
@@ -783,17 +794,42 @@ static int ov5640_set_fmt(const struct device *dev, struct video_format *fmt)
 
 	/* Set resolution */
 	for (i = 0; i < array_size_modes; i++) {
```
josuah, line 795: Can we get the offset through the array via `video_format_caps_offset()` to avoid looping manually? Maybe not for this PR though as this would require shuffling things.

Best keeping it as it is now for this PR maybe.

drivers/video/ov5640.c
```diff
@@ -783,17 +794,42 @@ static int ov5640_set_fmt(const struct device *dev, struct video_format *fmt)
...
-			if (ret) {
-				LOG_ERR("Unable to set resolution parameters");
-				return ret;
-			}
-
-			drv_data->cur_mode = &modes[i];
-			break;
+		if (fmt->width != modes[i].width || fmt->height != modes[i].height) {
+			continue;
 		}
```
josuah, line 798: Can the `continue;` be turned into a `break;` (also inverting the `if()` content) and then indentation further reduced below this block?

Removing the `break;` at the end of the loop too.

It is just about getting +0.01% readability.

josuah (approved):
Thank you for being so patient!

It is possible to specify all devicetree defaults in the bindings instead of `_OR()`, but I never know which one to use...

LGTM despite the comments (not things to change, just observing what you did).

nashif (approved):

PR #98751, opened 2025-11-02
scripts: west_commands: patch: fix handling of stdout and stderr from subprocess.run()

mshinjo:
Pass `capture_output=True` and `encoding="utf-8"` to `subprocess.run()` where appropriate, to correctly handle stdout and stderr from the process created by `subprocess.run()`.

### DIFF
--- a/scripts/west_commands/patch.py
+++ b/scripts/west_commands/patch.py
@@ -391,7 +391,7 @@ class Patch(WestCommand):
             self.dbg(f"patching {mod}... ", end="")
             apply_cmd += patch_path
             apply_cmd_list.extend([patch_path])
-            proc = subprocess.run(apply_cmd_list, cwd=mod_path)
+            proc = subprocess.run(apply_cmd_list, capture_output=True, cwd=mod_path)
             if proc.returncode:
                 self.dbg("FAIL")
                 self.err(proc.stderr)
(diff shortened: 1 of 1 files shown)


scripts/west_commands/patch.py
```diff
@@ -391,7 +391,7 @@ def apply(self, args, yml, dst_mods=None):
             self.dbg(f"patching {mod}... ", end="")
             apply_cmd += patch_path
             apply_cmd_list.extend([patch_path])
-            proc = subprocess.run(apply_cmd_list, cwd=mod_path)
+            proc = subprocess.run(apply_cmd_list, capture_output=True, cwd=mod_path, text=True)
```
carlescufi: I may be better to use `encoding="utf-8"` instead?

mshinjo: Thanks for the review! Yes, I agree that `encoding="utf-8"` is more predictable and consistent, I have made an update.

pdgendt (approved):

ljd42 (approved):

marc-hb (commented):
As always with hardcoding, some SUBPROC_ENCODING constant  at the top of the file (or any better name) would be nice.

Ideally, the same constant should probably apply to more than just one extension. But avoiding the duplication in just this file would already be progress.

PR #98752, opened 2025-11-02
drivers: mipi-dbi: Introduce BFLB DBI driver

VynDragon:
Adds a DBI driver for BL61x series.


arch/riscv/custom/thead/cache_xtheadcmo.c
```diff
@@ -83,7 +84,8 @@ int arch_icache_invd_range(void *addr_in, size_t size)
 		"fence\n"
 		"fence.i\n"
 	);
-	for (uintptr_t i = addr; i < addr + size; i += CONFIG_ICACHE_LINE_SIZE) {
+	for (uintptr_t i = addr; i < ROUND_UP(addr + size, CONFIG_DCACHE_LINE_SIZE);
```
josuah: it seems like it's a version before the `D` -> `I` fixup

drivers/mipi_dbi/Kconfig.bflb
```diff
@@ -0,0 +1,20 @@
...
+
+if MIPI_DBI_BFLB
+
+config MIPI_DBI_BFLB_ASSUME_HOLD_ON_CS
+	bool "Default to HOLD_ON_CS mode"
+	default y
+	help
+	  Pretend SPI_HOLD_ON_CS was passed by the display driver, as this is the expectation.
+
+endif # MIPI_DBI_NXP_LCDIC
```
josuah: There's some `NXP` still there
```suggestion
endif # MIPI_DBI_BFLB
```

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+
+static int mipi_dbi_bflb_trigger(const struct device *dev)
+{
+	volatile uint32_t tmp;
+	const struct mipi_dbi_bflb_config *config = dev->config;
+	k_timepoint_t end_timeout =
+		sys_timepoint_calc(K_MSEC(DBI_WAIT_TIMEOUT_MS));
+
+	while (mipi_dbi_bflb_bus_busy(dev) && !sys_timepoint_expired(end_timeout)) {
+	}
```
josuah, line 137: You could be interested in a `k_sleep(K_MSEC(1))` (or other duration) inside the while loop to allow the CPU do something else while it waits.

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+
+static int mipi_dbi_bflb_detrigger(const struct device *dev)
+{
+	volatile uint32_t tmp;
+	const struct mipi_dbi_bflb_config *config = dev->config;
+	k_timepoint_t end_timeout =
+		sys_timepoint_calc(K_MSEC(DBI_WAIT_TIMEOUT_MS));
+
+	while (mipi_dbi_bflb_bus_busy(dev) && !sys_timepoint_expired(end_timeout)) {
+	}
```
josuah, line 158: If you wanted a `k_sleep()` above, probably here too

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+	spi_mux = (dbi_divider & GLB_DBI_CLK_SEL_MSK) >> GLB_DBI_CLK_SEL_POS;
+	dbi_divider = (dbi_divider & GLB_DBI_CLK_DIV_MSK) >> GLB_DBI_CLK_DIV_POS;
+
+	if (spi_mux > 0) {
+		if (main_clock == BFLB_MAIN_CLOCK_RC32M
+			|| main_clock == BFLB_MAIN_CLOCK_PLL_RC32M) {
+			return BFLB_RC32M_FREQUENCY / (dbi_divider + 1);
+		}
+		clock_control_get_rate(clock_ctrl, (void *)BFLB_CLKID_CLK_CRYSTAL, &uclk);
+		return uclk / (dbi_divider + 1);
```
josuah, line 189: If wanting to simplify this, it is possible to move this code and define a new `XCLK` as single source for when `spi_mux > 0`:

screenshot_2025_1103_0052_42

josuah: I can make a PR for changing this after this is merged (i.e. in common for several peripherals).

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+	sys_write32(tmp, config->base + DBI_CONFIG_OFFSET);
+
+	return 0;
+}
+
+static uint32_t mipi_dbi_get_clk(void)
+{
+	uint32_t uclk;
+	uint32_t dbi_divider;
+	uint32_t spi_mux;
```
josuah: This looks like a typo, unless it's the same mux as SPI.

```suggestion
	uint32_t dbi_mux;
```

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+	mipi_dbi_bflb_trigger(dev);
+
+	return 0;
+}
+
+static int mipi_dbi_bflb_write_display(const struct device *dev,
+				       const struct mipi_dbi_config *dbi_config,
+				       const uint8_t *framebuf,
+				       struct display_buffer_descriptor *desc,
+				       enum display_pixel_format pixfmt)
```
josuah, line 520: Does it need to return `-ENOTSUP` or `-ENOSYS` when a pixel format is not supported or is it passing everything straight through unchanged (like if it was SPI)?

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+
+	if (len > 0) {
+		while (loaded < len) {
+			tmp = sys_read32(cfg->base + DBI_FIFO_CONFIG_1_OFFSET);
+			if ((tmp & DBI_TX_FIFO_CNT_MASK) > 0) {
+				tmp = 0;
+				j = 0;
+				for (; j < 4 && loaded + j < len; j++) {
+					tmp |= (buf[loaded + j]) << (j * 8);
+				}
```
josuah, line 422: I was thinking: let's use `sys_le32_to_cpu();` but nope(!) your version also handles the edge case when `len % 4 != 0`.

I was thinking: it writes extra 0x00 at the end when `len % 4 != 0` but nope(!) your version setup the length above already (`tmp |= (len - 1) << DBI_CR_DBI_DAT_BC_SHIFT`).

All good :heavy_check_mark:

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+				mipi_dbi_bflb_trigger(dev);
+				triggered = true;
+			}
+		}
+	} else {
+		mipi_dbi_bflb_trigger(dev);
+	}
+
+	do {
+		clock_bflb_settle();
```
josuah, line 436: likewise, a `k_sleep()` can be considered, or maybe it's better to be checking more aggressively.

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+	mipi_dbi_bflb_cs_control(dev, dbi_config, false, true);
+
+	if (mipi_dbi_has_pin(&config->dc_gpio)) {
+		ret = gpio_pin_set_dt(&config->dc_gpio, 0);
+		if (ret < 0) {
+			return ret;
+		}
+	}
+
+	k_sem_give(&data->lock);
```
josuah: There is a `k_sem_take()` but early `return xxx` before the `k_sem_give()`, was that expected so that the ISR does the `k_sem_give()` or rather a `goto out;` to getta outta here is better?

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+		return ret;
+	}
+
+	k_sleep(delay);
+
+	ret = gpio_pin_set_dt(&config->reset_gpio, 0);
+
+	k_sem_give(&data->lock);
+
+	return ret;
```
josuah: For this section, could be worth moving the `!mipi_dbi_has_pin()` above the `k_sem_take()` and a `goto out;` as well eventually.

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,773 @@
...
+		.dma.config.dma_slot = DT_INST_DMAS_CELL_BY_NAME(n, tx, trigsrc),		   \
+	};											   \
+												   \
+	DEVICE_DT_INST_DEFINE(n, mipi_dbi_bflb_init, NULL,					   \
+			&mipi_dbi_bflb_data_##n,						   \
+			&mipi_dbi_bflb_config_##n,						   \
+			POST_KERNEL,								   \
+			CONFIG_MIPI_DBI_INIT_PRIORITY,						   \
+			&mipi_dbi_bflb_driver_api);						   \
+	MIPI_DBI_BFLB_IRQ_HANDLER(n)
```
josuah: If you wanted you could have put this above where `MIPI_DBI_BFLB_IRQ_HANDLER_DECL()` is and then not need to any `MIPI_DBI_BFLB_IRQ_HANDLER_DECL()`.
But that's possibly on purpose as it is now, which of course works just as much.

josuah (dismissed):
Here are some runes you could be interested in checking.

[EDIT: I did mean to "comment" rather than "approve" as it's not undrafted yet

arch/riscv/custom/thead/cache_xtheadcmo.c
VynDragon: yes

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: it passes everything through here, there is not special support for RGB as the features for it in HW are not used at all (otherwise it would have a limit of size of 24-bit long number instead of 256).

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: > that expected so that the ISR does the k_sem_give() 

Yes, it's not a mutex for the same reason (sync lock and not just cross-thread lock)

VynDragon: Well except for the pin ret, that should do a give as well

arch/riscv/custom/thead/cache_xtheadcmo.c
VynDragon: done rebase fixed this

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: well not here cuz it's a do while

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/Kconfig.bflb
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
josuah: TIL semaphore given from ISR.

drivers/mipi_dbi/mipi_dbi_bflb.c
josuah: makes complete sense then, thanks

josuah (dismissed):
Ok, now `mipi_dbi_bflb_write_display()` is blocking with a mutex. No issue found in the modifications either.

Release 2025/11/14 (next Friday), after that merging can at last resume.

drivers/mipi_dbi/Kconfig.bflb
```diff
@@ -0,0 +1,20 @@
...
+	bool "Bouffalolab MIPI DBI driver"
+	default y
+	depends on DT_HAS_BFLB_DBI_ENABLED
+	select DMA
+	help
+	  Enable support for the Bouffalolab DBI controller
+
+if MIPI_DBI_BFLB
+
+config MIPI_DBI_BFLB_ASSUME_HOLD_ON_CS
```
danieldegrasse, line 14: Could we just fix the offending display driver to enable HOLD_ON_CS rather than using this workaround?

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+
+	tmp = sys_read32(config->base + DBI_CONFIG_OFFSET);
+
+	return (tmp & DBI_STS_DBI_BUS_BUSY) != 0;
+}
+
+static void mipi_dbi_bflb_clear_fifo(const struct device *dev)
+{
+	const struct mipi_dbi_bflb_config *config = dev->config;
+	volatile uint32_t tmp;
```
danieldegrasse, line 121: Does this need to be volatile?

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+
+	tmp = sys_read32(config->base + DBI_FIFO_CONFIG_0_OFFSET);
+	tmp |= DBI_TX_FIFO_CLR;
+	sys_write32(tmp, config->base + DBI_FIFO_CONFIG_0_OFFSET);
+}
+
+static int mipi_dbi_bflb_trigger(const struct device *dev)
+{
+	volatile uint32_t tmp;
```
danieldegrasse, line 130: Same question here, unsure if this needs to be volatile

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+	tmp = sys_read32(config->base + DBI_CONFIG_OFFSET);
+	tmp |= DBI_CR_DBI_EN;
+	sys_write32(tmp, config->base + DBI_CONFIG_OFFSET);
+
+	return 0;
+}
+
+static int mipi_dbi_bflb_detrigger(const struct device *dev)
+{
+	volatile uint32_t tmp;
```
danieldegrasse, line 151: Ditto-please remove volatile on these if possible

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+	const struct mipi_dbi_bflb_config *config = dev->config;
+	k_timepoint_t end_timeout =
+		sys_timepoint_calc(K_MSEC(DBI_WAIT_TIMEOUT_MS));
+
+	while (mipi_dbi_bflb_bus_busy(dev) && !sys_timepoint_expired(end_timeout)) {
+		k_usleep(1);
```
danieldegrasse, line 157: Why make the `k_usleep` call here? Is the goal to invoke the scheduler so other threads can run if needed? Might be better to just use `k_yield`.

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+
+	return uclk / (dbi_divider + 1);
+}
+
+static int mipi_dbi_bflb_configure_freqs(const struct device *dev,
+					 const struct mipi_dbi_config *config)
+{
+	const struct mipi_dbi_bflb_config *cfg = dev->config;
+	const struct device *clock_ctrl =  DEVICE_DT_GET_ANY(bflb_clock_controller);
+	volatile uint32_t tmp;
```
danieldegrasse, line 202:
```suggestion
	uint32_t tmp;
```

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+	switch (config->mode) {
+	case MIPI_DBI_MODE_SPI_3WIRE:
+		tmp |= DBI_MODE_C_3W << DBI_CR_DBI_SEL_SHIFT;
+		break;
+	case MIPI_DBI_MODE_SPI_4WIRE:
+		tmp |= DBI_MODE_C_4W << DBI_CR_DBI_SEL_SHIFT;
+		break;
+	case MIPI_DBI_MODE_8080_BUS_8_BIT:
+		tmp |= DBI_MODE_B_8B << DBI_CR_DBI_SEL_SHIFT;
+		return -ENOSYS;
```
danieldegrasse, line 284: Why return an error here if we are setting bits that look like we support the mode?

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+
+static int mipi_dbi_bflb_start_dma_write(const struct device *dev,
+				       const struct mipi_dbi_config *dbi_config,
+				       const uint8_t *buf, size_t len)
+{
+	const struct mipi_dbi_bflb_config *cfg = dev->config;
+	struct mipi_dbi_bflb_data *data = dev->data;
+	size_t len_todo = len > DBI_MAX_CMD_WRITE_LEN ? DBI_MAX_CMD_WRITE_LEN : len;
+	int ret;
+	volatile uint32_t tmp;
```
danieldegrasse, line 464:
```suggestion
	uint32_t tmp;
```

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+static int mipi_dbi_bflb_write_display(const struct device *dev,
+				       const struct mipi_dbi_config *dbi_config,
+				       const uint8_t *framebuf,
+				       struct display_buffer_descriptor *desc,
+				       enum display_pixel_format pixfmt)
+{
+	const struct mipi_dbi_bflb_config *cfg = dev->config;
+	struct mipi_dbi_bflb_data *data = dev->data;
+	int ret;
+	volatile uint32_t tmp;
```
danieldegrasse, line 525:
```suggestion
	uint32_t tmp;
```

drivers/mipi_dbi/mipi_dbi_bflb.c
```diff
@@ -0,0 +1,776 @@
...
+	.command_write = mipi_dbi_bflb_command_write,
+	.write_display = mipi_dbi_bflb_write_display,
+	.release = mipi_dbi_bflb_release,
+};
+
+static void mipi_dbi_bflb_isr(const struct device *dev)
+{
+	const struct mipi_dbi_bflb_config *config = dev->config;
+	struct mipi_dbi_bflb_data *data = dev->data;
+	volatile uint32_t tmp;
```
danieldegrasse, line 722:
```suggestion
	uint32_t tmp;
```

danieldegrasse (commented):
Can we enable this driver within a display sample or similar to make sure it builds in CI?

drivers/mipi_dbi/mipi_dbi_bflb.c
josuah: I was wondering if `k_yield()` would have any effect if the display thread is still highest priority.
Maybe it can still yield to other equal priority threads though.

drivers/mipi_dbi/Kconfig.bflb
VynDragon: Most drivers are offending, check the dbi driver for spi.

VynDragon: https://github.com/zephyrproject-rtos/zephyr/blob/9c1fbc8a86ad0f57190b6b84dd19bd88a189dc49/drivers/mipi_dbi/mipi_dbi_spi.c#L468

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: because the mode may be supported at some point, it isn't at the moment and i have nothing to test it.

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: I got tired of spending a very long time on edge-cases where gcc would swallow or reorder writes and read so all direct register variables access are now volatile by default.

drivers/mipi_dbi/mipi_dbi_bflb.c
JarmouniA: `k_sleep` would naturally result in another ready thread running. It's not the same as `k_busy_wait()`.
https://docs.zephyrproject.org/latest/kernel/services/scheduling/index.html#thread-sleeping

`k_yield` does not guarantee the rescheduling of the calling thread in a specific amount of time.

VynDragon: the goal is to leave some schedule time open, calculating expected transmit time was/is considered. What is the best solution here?

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: (and sys_write32 and sys_read32 are already volatile accesses, if you're worried this may cause performance issues or anything else, it's already not supposed to reorder or swallow those interactions.)

drivers/mipi_dbi/mipi_dbi_bflb.c
danieldegrasse: > k_yield does not guarantee the rescheduling of the calling thread in a specific amount of time.

This is a good point- however it seems like the goal here is to simply leave schedule time open for other high priority threads to run, right? Do we *want* the behavior of `k_sleep`, where lower priority cooperative threads could run until they yield? That seems like priority inversion

josuah: > Do we want the behavior of k_sleep, where lower priority cooperative threads could run until they yield?

I suppose so, for instance, for rendering the next frame (less timing-critical, lower priority) while the hardware is busy sending the previous frame.

Does it makes sense? Other approach recommended?

Thanks!

VynDragon: > Do we want the behavior of k_sleep, where lower priority cooperative threads could run until they yield?

Yes actually

drivers/mipi_dbi/mipi_dbi_bflb.c
danieldegrasse: Ok- can you add a comment explaining that to this code for anyone looking at it in the future?

dts/bindings/mipi-dbi/bflb,dbi.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2025 MASSDRIVER EI (massdriver.space)
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  DBI settings for BFLB MIPI DBI controller
```
JarmouniA:
```suggestion
  BFLB MIPI DBI controller
```

JarmouniA:
> > Can we enable this driver within a display sample or similar to make sure it builds in CI?
> 
> Not sure how to, otherwise out of pins for putting it board-side

You could at least ensure it is built by adding a node in `tests/drivers/build_all/display/app.overlay`, similar to `zephyr,mipi-dbi-spi`.

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: @danieldegrasse So, any thoughts on this?

dts/bindings/mipi-dbi/bflb,dbi.yaml
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: done

drivers/mipi_dbi/mipi_dbi_bflb.c
VynDragon: Added a expected wait time calculation to it sleeps for a accurate amount of time to ensure scheduling is done as good as possible, done.

tests/drivers/build_all/display/boards/ai_m62_12f.overlay
JarmouniA, line 1: Could you add the device node in `app.overlay` directly?

VynDragon: well i'd need to add a vnd,dma node, is that okay?

JarmouniA: > well i'd need to add a vnd,dma node, is that okay?

yes, we just want to build the driver.

VynDragon: Okay well that's not working, it needs to be built in the bflb 'environment' (include paths), it cant be built independent of platform like regular display drivers. Also shouldnt this be in a separate mipi-dbi build_all folder?

JarmouniA: > Okay well that's not working, it needs to be built in the bflb 'environment' (include paths), it cant be built independent of platform like regular display drivers.

You could add the necessary includes, no?

JarmouniA: > Also shouldnt this be in a separate mipi-dbi build_all folder?

There is already mipi-dbi-spi and mipi-dsi in there.

JarmouniA: Anyway, if it doesn't work or is complicated to implement, I'm fine with the board overlay and separate test scenario.

VynDragon: > There is already mipi-dbi-spi and mipi-dsi in there.

yes, for mipi dbi and mipi dsi *displays*, for mipi-dbi drivers there isnt a folder at all and for mipi-dsi there is only an api test

JarmouniA: > yes, for mipi dbi and mipi dsi displays,

It's for controllers as well.
https://github.com/zephyrproject-rtos/zephyr/blob/da9aa8fcad1a015afa7c07e19f0216a14eb44ae7/tests/drivers/build_all/display/app.overlay#L30
This covers mipi-dbi-spi controller driver for ex.

> for mipi-dsi there is only an api test

Build all is different from API testing, but yes, mipi-dsi controllers are missing here.

JarmouniA (dismissed):

JarmouniA (approved):

josuah (approved):
The test ran well:
https://github.com/zephyrproject-rtos/zephyr/actions/runs/19548205283/job/55972875159?pr=98752#step:13:831

danieldegrasse (approved):

nordic-krch (approved):

PR #98755, opened 2025-11-03
kernel: `Z_HEAP_MIN_SIZE` for runtime stats

### LINKED ISSUE #98754: kernel: `Z_HEAP_MIN_SIZE` incorrect for `CONFIG_SYS_HEAP_RUNTIME_STATS=y`
### Describe the bug

`Z_HEAP_MIN_SIZE` does not define a heap size that will successfully allocate a 1 byte request.

### Regression

- [ ] This is a regression.

### Steps to reproduce

1. `tests/lib/heap/src/main.c`
```
ZTEST(lib_heap, test_heap_min_size)
{
	struct sys_heap heap;
	void *mem;

	sys_heap_init(&heap, heapmem, Z_HEAP_MIN_SIZE);
	mem = sys_heap_alloc(&heap, 1);
	zassert_not_null(mem, "Could not allocate 1 byte from a Z_HEAP_MIN_SIZE heap");
	sys_heap_free(&heap, mem);
}
```
2. `west build -b native_sim/native -t run zephyr/tests/lib/heap -p`
3. `west build -b native_sim/native/64 -t run zephyr/tests/lib/heap -p`

### Relevant log output

```shell
START - test_heap_min_size
ASSERTION FAIL [chunk0_size + min_chunk_size(h) <= heap_sz] @ WEST_TOPDIR/zephyr/lib/heap/heap.c:573
        heap size is too small
@ WEST_TOPDIR/zephyr/lib/os/assert.c:43
[00:00:00.000,000] <err> os: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
[00:00:00.000,000] <err> os: Current thread: 0x477140 (test_heap_min_size)
[00:00:00.000,000] <err> os: Halting system
Exiting due to fatal error

START - test_heap_min_size
ASSERTION FAIL [chunk0_size + min_chunk_size(h) <= heap_sz] @ WEST_TOPDIR/zephyr/lib/heap/heap.c:573
        heap size is too small
@ WEST_TOPDIR/zephyr/lib/os/assert.c:43
[00:00:00.000,000] <err> os: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
[00:00:00.000,000] <err> os: Current thread: 0x80bdb00 (test_heap_min_size)
[00:00:00.000,000] <err> os: Halting system
Exiting due to fatal error
```

### Impact

### Environment

- Zephyr v4.3-rc2

### Additional Context

_No response_

JordanYates:
Increase `Z_HEAP_MIN_SIZE` values to successfully allocate a 1 byte chunk when `CONFIG_SYS_HEAP_RUNTIME_STATS=y`.

Validate that `Z_HEAP_MIN_SIZE` is the minimum required size.

Fixes #98754

### DIFF
--- a/include/zephyr/kernel.h
+++ b/include/zephyr/kernel.h
@@ -5930,10 +5930,14 @@ void *k_heap_realloc(struct k_heap *h, void *ptr, size_t bytes, k_timeout_t time
  */
 void k_heap_free(struct k_heap *h, void *mem) __attribute_nonnull(1);
 
-/* Hand-calculated minimum heap sizes needed to return a successful
- * 1-byte allocation.  See details in lib/os/heap.[ch]
+/* Minimum heap sizes needed to return a successful 1-byte allocation.
+ * Assumes a chunk aligned (8 byte) memory buffer.
  */
+#ifdef CONFIG_SYS_HEAP_RUNTIME_STATS
+#define Z_HEAP_MIN_SIZE ((sizeof(void *) > 4) ? 80 : 52)
+#else
 #define Z_HEAP_MIN_SIZE ((sizeof(void *) > 4) ? 56 : 44)
+#endif /* CONFIG_SYS_HEAP_RUNTIME_STATS */
 
 /**
  * @brief Define a static k_heap in the specified linker section
--- a/tests/lib/heap_min/CMakeLists.txt
+++ b/tests/lib/heap_min/CMakeLists.txt
@@ -1,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(heap_min)
+
+FILE(GLOB app_sources src/*.c)
+target_sources(app PRIVATE ${app_sources})
--- a/tests/lib/heap_min/src/main.c
+++ b/tests/lib/heap_min/src/main.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Embeint Inc
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <zephyr/kernel.h>
+#include <zephyr/ztest.h>
+#include <zephyr/sys/sys_heap.h>
+#include <inttypes.h>
+
+#include "assert.h"
+
+DEFINE_FFF_GLOBALS;
+
+/* Align the test memory to the heap chunk size */
+uint8_t heapmem[8192] __aligned(8);
+
+ZTEST(lib_heap_min, test_heap_min_size_assert)
+{
+	struct sys_heap heap;
+
+	expect_assert();
+	sys_heap_init(&heap, (void *)heapmem, Z_HEAP_MIN_SIZE - 1);
+	zassert_unreachable();
+}
+
+ZTEST(lib_heap_min, test_heap_min_size)
+{
+	struct sys_heap heap;
+	void *mem;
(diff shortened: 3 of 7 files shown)


npitre (approved):

tests/lib/heap_min/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(heap_min)
```
JarmouniA, line 5: heap_min can be confused with `tests/lib/min_heap` which is entirely a different thing.

tests/lib/heap_min/src/main.c
JarmouniA, line 1: Why not put this 2 testcases in `tests/lib/heap`?

JordanYates: The top of `tests/lib/heap` has a number of comments about large execution times and things that have been done to reduce that. I didn't want to double the duration of the test for no reason, as would be required to test both values of `CONFIG_SYS_HEAP_RUNTIME_STATS`.

tests/lib/heap_min/CMakeLists.txt
JordanYates: it could be yes

JarmouniA: Could you maybe put this test's folder inside `tests/lib/heap` and/or rename it to `heap_min_size`?

include/zephyr/kernel.h
```diff
@@ -5930,10 +5930,14 @@ void *k_heap_realloc(struct k_heap *h, void *ptr, size_t bytes, k_timeout_t time
...
-/* Hand-calculated minimum heap sizes needed to return a successful
- * 1-byte allocation.  See details in lib/os/heap.[ch]
+/* Minimum heap sizes needed to return a successful 1-byte allocation.
+ * Assumes a chunk aligned (8 byte) memory buffer.
  */
+#ifdef CONFIG_SYS_HEAP_RUNTIME_STATS
+#define Z_HEAP_MIN_SIZE ((sizeof(void *) > 4) ? 80 : 52)
+#else
 #define Z_HEAP_MIN_SIZE ((sizeof(void *) > 4) ? 56 : 44)
+#endif /* CONFIG_SYS_HEAP_RUNTIME_STATS */
```
peter-mitsis, line 5940: On 32-bit platforms I would expect an increase of 12-bytes, and on 64-bit platforms I would expect an increase of 24 bytes when SYS_HEAP_RUNTIME_STATS is enabled. I see 8 and 24 respectively.

Additionally, are we certain about the original minimal values?  If I follow things correctly, the size of  `k_heap` is also going to be dependent type of _wait_q_t (simple or scalable) as well as which spin lock features are enabled--and I am not yet certain if the original hand-calculated minimums have been kept up to date.

JordanYates: > Additionally, are we certain about the original minimal values?

That's the point of the new tests. They ensure that a 1 byte allocation works with `Z_HEAP_MIN_SIZE`, and fails with `Z_HEAP_MIN_SIZE - 1`

>  If I follow things correctly, the size of `k_heap`

The overhead is from the internal `z_heap`, not `k_heap`

peter-mitsis: I see where I went off track now.

peter-mitsis (approved):

PR #98759, opened 2025-11-03
drivers: watchdog: stm32 wwdg: remove floats to reduce footprint

### LINKED ISSUE #97111: stm32: watchdog: Get rid of floats to reduce footprint
### Summary

On `nucleo_f091rc`, WWDG drivers requires 4K more of flash than IWDG.

This is due to use of floats in computations in `wwdg_stm32_get_timeout` and `wwdg_stm32_convert_timeout`.

`wwdg_stm32_get_timeout` is only used for configuration check and there should be ways to get rid of it.
For `wwdg_stm32_convert_timeout` it should be possible to rework it in order to spare the use of float as it is done in IWDG driver (`iwdg_stm32_convert_timeout`)

### Alternatives

_No response_

### Additional Context

_No response_

jam-kt:
On STM boards without an FPU (like nucleo_f091rc) the WWDG driver requires 4K more of flash than the IWDG due to the use of floating point math. Fixed it by refactoring to use integer math. Reduced sample build ROM size from 24848 B to 21648 B on nucleo_f091rc.
Fixes zephyrproject-rtos/zephyr#97111

### DIFF
--- a/drivers/watchdog/wdt_wwdg_stm32.c
+++ b/drivers/watchdog/wdt_wwdg_stm32.c
@@ -114,10 +114,15 @@ static uint32_t wwdg_stm32_get_timeout(const struct device *dev,
 				       uint32_t prescaler_exp,
 				       uint32_t counter)
 {
-	uint32_t divider = WWDG_INTERNAL_DIVIDER * (1 << prescaler_exp);
-	float f_wwdg = (float)wwdg_stm32_get_pclk(dev) / divider;
+	uint32_t pclk = wwdg_stm32_get_pclk(dev);
+	uint32_t ticks = (counter & 0x3FU) + 1U;
+	uint64_t divider = WWDG_INTERNAL_DIVIDER << prescaler_exp;
+	uint64_t num = (uint64_t)ticks * divider * (uint64_t)USEC_PER_SEC;
+	uint64_t time_us64 = num / (uint64_t)pclk;
 
-	return USEC_PER_SEC * (((counter & 0x3F) + 1) / f_wwdg);
+	__ASSERT_NO_MSG(time_us64 <= UINT32_MAX);
+
+	return (uint32_t)time_us64;
 }
 
 /**
@@ -133,24 +138,20 @@ static void wwdg_stm32_convert_timeout(const struct device *dev,
 				       uint32_t *prescaler_exp,
 				       uint32_t *counter)
 {
-	uint32_t clock_freq = wwdg_stm32_get_pclk(dev);
+	uint32_t pclk = wwdg_stm32_get_pclk(dev);
+	uint32_t psc_exp;
+	uint32_t count;
 
-	/* Convert timeout to seconds. */
-	float timeout_s = (float)timeout / USEC_PER_SEC;
-	float wwdg_freq;
+	for (psc_exp = 0U; psc_exp <= WWDG_PRESCALER_EXPONENT_MAX; psc_exp++) {
+		uint32_t divider = WWDG_INTERNAL_DIVIDER << psc_exp;
+		uint64_t num = (uint64_t)timeout * (uint64_t)pclk;
+		uint64_t den = (uint64_t)USEC_PER_SEC * (uint64_t)divider;
 
-	*prescaler_exp = 0U;
-	*counter = 0;
+		count = (num / den) + WWDG_RESET_LIMIT;
 
-	for (*prescaler_exp = 0; *prescaler_exp <= WWDG_PRESCALER_EXPONENT_MAX;
-	     (*prescaler_exp)++) {
-		wwdg_freq = ((float)clock_freq) / WWDG_INTERNAL_DIVIDER
-			     / (1 << *prescaler_exp);
-		/* +1 to ceil the result, which may lose from truncation */
-		*counter = (uint32_t)(timeout_s * wwdg_freq + 1) - 1;
-		*counter += WWDG_RESET_LIMIT;
-
-		if (*counter <= WWDG_COUNTER_MAX) {
+		if (count <= WWDG_COUNTER_MAX) {
+			*counter = count;
+			*prescaler_exp = psc_exp;
 			return;
 		}
 	}


drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -114,10 +114,17 @@ static uint32_t wwdg_stm32_get_timeout(const struct device *dev,
 				       uint32_t prescaler_exp,
 				       uint32_t counter)
 {
-	uint32_t divider = WWDG_INTERNAL_DIVIDER * (1 << prescaler_exp);
-	float f_wwdg = (float)wwdg_stm32_get_pclk(dev) / divider;
+	uint32_t pclk = wwdg_stm32_get_pclk(dev);
+	uint32_t ticks = ((counter & 0x3FU) + 1U);
```
etienne-lms: Can drop the outer parenetheses.

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -114,10 +114,17 @@ static uint32_t wwdg_stm32_get_timeout(const struct device *dev,
...
+	uint64_t divider = WWDG_INTERNAL_DIVIDER * (1U << prescaler_exp);
 
-	return USEC_PER_SEC * (((counter & 0x3F) + 1) / f_wwdg);
+	/*
+	 * timeout formula reordered to increase precision without floats
+	 * time_us = floor(ticks * divider * USEC_PER_SEC / pclk).
+	 */
```
etienne-lms: Inline comment is not needed IMHO. The commit message already describes the change.

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -114,10 +114,17 @@ static uint32_t wwdg_stm32_get_timeout(const struct device *dev,
...
+	uint64_t num = (uint64_t)ticks * divider * (uint64_t)USEC_PER_SEC;
+
+	return (uint32_t)(num / (uint64_t)pclk);
```
etienne-lms: Add `__ASSERT_NO_MSG((num / (uint64_t)pclk) <= UINT32_MAX);` for consistency?

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -133,24 +140,23 @@ static void wwdg_stm32_convert_timeout(const struct device *dev,
...
+	uint32_t pclk = wwdg_stm32_get_pclk(dev);
+	uint32_t exp;
+	uint32_t count;
+
+	/* Required ticks = ceil(timeout * pclk / (USEC_PER_SEC * divider)) */
+	for (exp = 0U; exp <= WWDG_PRESCALER_EXPONENT_MAX; exp++) {
+		uint32_t divider = WWDG_INTERNAL_DIVIDER * (1U << exp);
+		uint64_t num = (uint64_t)timeout * (uint64_t)pclk;
+		uint64_t den = (uint64_t)USEC_PER_SEC * (uint64_t)divider;
+		uint64_t ticks = (num + den - 1U) / den;
```
etienne-lms: Why minus 1 here?

mathieuchopstm: `DIV_ROUND_UP(num, den)`

jam-kt: I was not aware of the macro, thank you.

I realized that my math behaves differently than the original when the number of ticks is an exact integer value.
In the original `*counter = (uint32_t)(timeout_s * wwdg_freq + 1) - 1;` if the tick value, `timeout_s * wwdg_freq`, is an integer value then the counter's final value is equal to `tick + WWDG_RESET_LIMIT`.
In mine, when num/den is an integer value, `DIV_ROUND_UP(num, den)` does not round up and the counter's final value is equal to `tick - 1 + WWDG_RESET_LIMIT`.

Would it be best to remove the ceiling division and have `count = (uint32_t)(num / den);` to match the original code's behaviour even if it deviates from the “ceil then −1” model?

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -114,10 +114,17 @@ static uint32_t wwdg_stm32_get_timeout(const struct device *dev,
 				       uint32_t prescaler_exp,
 				       uint32_t counter)
 {
-	uint32_t divider = WWDG_INTERNAL_DIVIDER * (1 << prescaler_exp);
-	float f_wwdg = (float)wwdg_stm32_get_pclk(dev) / divider;
+	uint32_t pclk = wwdg_stm32_get_pclk(dev);
+	uint32_t ticks = ((counter & 0x3FU) + 1U);
+	uint64_t divider = WWDG_INTERNAL_DIVIDER * (1U << prescaler_exp);
```
mathieuchopstm:
```suggestion
	uint64_t divider = WWDG_INTERNAL_DIVIDER << prescaler_exp;
```

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -133,24 +140,23 @@ static void wwdg_stm32_convert_timeout(const struct device *dev,
...
+		uint64_t num = (uint64_t)timeout * (uint64_t)pclk;
+		uint64_t den = (uint64_t)USEC_PER_SEC * (uint64_t)divider;
+		uint64_t ticks = (num + den - 1U) / den;
+
+		count = (uint32_t)ticks - 1U;
+		count += WWDG_RESET_LIMIT;
+
+		if (count <= WWDG_COUNTER_MAX) {
+			*counter = count;
+			*prescaler_exp = psc_exp;
```
mathieuchopstm: My analysis:
* `NSEC_PER_SEC` = 10^9 fits in 32 bits (so does `4 * NSEC_PER_SEC`)
* `PRESCALER_MAX` = (128 * 4096) = 524288 (2^19 ~= ½ * 10^6)
  * thus `4096 < prescaler < 524288` 
* `pclk` in [1 MHz; 1 GHz]

Thus
```
   1 MHz < pclk < 1 GHz (= NSEC_PER_SEC)
=> 10^-9 < (1/pclk) < 10^-6

=> 1 < (1/pclk) * NSEC_PER_SEC < 10^3
=> 1 < NSEC_PER_SEC / pclk < 10^3

=> 4096 < prescaler * NSEC_PER_SEC / pclk <= (2^19 * 10^3) < 10^9
```
and `uint32_t t_wdt = (NSEC_PER_SEC / pclk) * prescaler;` (number of ns per WDT tick) fits in 32 bits - I don't see the need for 64-bit math (which is much slower than 32-bit)

In my head, the following seems correct (but I may have missed something):
```c
#define MAX_TICKS_COUNT 0x40u /* = 0x7F - 0x40 + 1 */

	/* At prescaler=1; valid only up to pclk=4 GHz */
	uint32_t base_t_wdt = ((4 * NSEC_PER_SEC) / pclk) * (WWDG_INTERNAL_DIVIDER / 4);
	uint64_t timeout_ns = timeout * NSEC_PER_MSEC; //TODO: can be demoted to 32-bit if there is no valid timeout >4s

	for (uint32_t log2_psc = 0U; log2_psc <= WWDG_PRESCALER_EXPONENT_MAX; log2_psc++) {
		uint32_t t_wdt = base_t_wdt << log2_psc;
		uint64_t max_delay = (uint64_t)t_wdt * MAX_TICKS_COUNT;
		/* since MAX_TICKS_COUNT is a pow2, this is a cheap shift */

		if (max_delay < timeout_ns) {
			continue;
		}

		/* Guaranteed: 1 <= ticks <= MAX_TICKS_COUNT */
		*counter = DIV_ROUND_UP(timeout_ns, t_wdt);
		*prescaler_exp = log2_psc;
		return;
	}

	/* window overflow... */
```

This just requires the caller in `wwdg_stm32_install_timeout` to provide the raw `timeout` in `ms` (or in `ns`, but NOT `µs`).

(Note the clever way to support `pclk < 4 GHz` by changing order of operations: `t_wdt = NSEC_PER_SEC / pclk * prescaler = ((4 * NSEC_PER_SEC) / pclk) * (prescaler/4)` where `prescaler / 4` is free because it operates on constant `4096`)

mathieuchopstm: Also I would carry around the unmodified `counter` up until the call to `LL_WWDG_SetPrescaler` which would be
`LL_WWDG_SetPrescaler((counter + WWDG_RESET_LIMIT) << xxx)`

drivers/watchdog/wdt_wwdg_stm32.c
mathieuchopstm: I don't think there's a need for 64-bit math here either (To be reviewed).

But this function isn't even *needed* because `wwdg_stm32_convert_timeout` can already return an error when the requested timeout is out of range. (This function could be gated behind `LOG_LEVEL_xxx >= DBG`)

drivers/watchdog/wdt_wwdg_stm32.c
jam-kt: I believe it would need to be `uint32_t t_wdt = base_t_wdt << log2_psc;` in your example, is that correct?
Also I am unsure about changing the passed `timeout` from `µs` to another unit within the context of this PR.
If you would prefer to use the math you provided however I can push that instead.

drivers/watchdog/wdt_wwdg_stm32.c
jam-kt: I used 64 bit math here to preserve the original code's behaviour as closely as possible.
With typical values `ticks = 64`, `WWDG_INTERNAL_DIVIDER = 4096`, `prescaler_exp = 7` the expression
`uint64_t num = (uint64_t)ticks * divider * (uint64_t)USEC_PER_SEC;` would eval to `num = 64 * 524288 * 1,000,000 ` which is about 2^45

drivers/watchdog/wdt_wwdg_stm32.c
etienne-lms: In the previous implementation, `*counter = (uint32_t)(timeout_s * wwdg_freq + 1) - 1` gives the same result as `		*counter = (uint32_t)(timeout_s * wwdg_freq)`. I assume there is no expected 32bit overflow. Using `count = (uint32_t)(num / den) + WWDG_RESET_LIMIT` looks enough to me.

drivers/watchdog/wdt_wwdg_stm32.c
mathieuchopstm: You are correct, it would be `base_t_wdt << log2_psc` instead; I have updated the comment.

I don't see the issue with changing how the `timeout` is passed.

drivers/watchdog/wdt_wwdg_stm32.c
mathieuchopstm: Fine by me, though I'd prefer if this function because unused when debug logging isn't enabled.

drivers/watchdog/wdt_wwdg_stm32.c
mathieuchopstm: I'm not 100% confident in my math so I'd need to do additional verification, but at least checking for overflow before DIV computation and returning an error in case timeout is too long should definitely be integrated in PR.

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -225,17 +222,22 @@ static int wwdg_stm32_install_timeout(const struct device *dev,
...
 	calculated_timeout = wwdg_stm32_get_timeout(dev, prescaler_exp, counter);
 
 	LOG_DBG("prescaler: %d", (1 << prescaler_exp));
 	LOG_DBG("Desired WDT: %d us", timeout);
 	LOG_DBG("Set WDT:     %d us", calculated_timeout);
 
-	if (!(IS_WWDG_COUNTER(counter) &&
-	      IS_WWDG_TIMEOUT(timeout, calculated_timeout))) {
-		/* One of the parameters provided is invalid */
+	if (!(IS_WWDG_TIMEOUT(timeout, calculated_timeout))) {
```
etienne-lms: It looks strange this sanity test is only checked upon some log level.
`IS_WWDG_TIMEOUT()` seems to check some error margin ranges not checked by `IS_WWDG_COUNTER()`. I don't think this was tested only upon the log level value despite comment https://github.com/zephyrproject-rtos/zephyr/pull/98759#discussion_r2490822117.

etienne-lms (commented):
Change LGTM but regarding the added condition directive on log level.
@jam-kt, could you squash to 2 commits into a single one?

drivers/watchdog/wdt_wwdg_stm32.c
jam-kt: I've moved the `IS_WWDG_TIMEOUT()` check outside of the log level conditional while keeping the logging statements inside. Please let me know if anything else needs changing.

drivers/watchdog/wdt_wwdg_stm32.c
```diff
@@ -221,9 +222,11 @@ static int wwdg_stm32_install_timeout(const struct device *dev,
 	wwdg_stm32_convert_timeout(dev, timeout, &prescaler_exp, &counter);
 	calculated_timeout = wwdg_stm32_get_timeout(dev, prescaler_exp, counter);
 
+#if (LOG_LEVEL >= LOG_LEVEL_DBG)
 	LOG_DBG("prescaler: %d", (1 << prescaler_exp));
 	LOG_DBG("Desired WDT: %d us", timeout);
 	LOG_DBG("Set WDT:     %d us", calculated_timeout);
+#endif
```
etienne-lms: You don't need these directives. When log level is below debug level, `LOG_DBG()` macro already expands to no-operation.

mathieuchopstm: The `LOG_DBG` itself will be a no-op but not the call to `wwdg_stm32_get_timeout`, which is why I suggested the `#if`-blocking.

etienne-lms: `IS_WWDG_TIMEOUT()` performs an extra sanity test so worth to be checked even when log and debug assertions are off, IMHO.

mathieuchopstm: Which is why I was suggesting that `wwdg_stm32_convert_timeout` should return an error in case of overflow (and use that as "not valid" indicator rather than re-computed timeout): https://github.com/zephyrproject-rtos/zephyr/pull/98759#discussion_r2490828295

etienne-lms: I'm fine changing the the overflow test from an assertion into a error report.

That said `IS_WWDG_TIMEOUT()` checks that rounding in timeout computation does not make it deviate more than 10%. Is this sanity test worth to be preserved? I think it does.
A bit of details in commit message of 1c9041c3aa5cc642b6ae6baa4c67f9755fa35e70.

etienne-lms (approved):

mathieuchopstm (approved):

erwango (approved):
Thanks!

PR #98764, opened 2025-11-03
driver: wifi: esp32: add driver set_config interface

sylvioalves:
Allows set the device macaddress.

> net iface set_mac 1 aa:bb:cc:dd:ee:ff
> wifi connect -s ssid -p password -k 1


drivers/wifi/esp32/src/esp_wifi_drv.c
```diff
@@ -890,6 +890,32 @@ static int esp32_wifi_dev_init(const struct device *dev)
...
+		esp_wifi_set_mode(ESP32_WIFI_MODE_STA);
+		memcpy(dev_data->mac_addr, config->mac_address.addr, 6);
+		net_if_set_link_addr(esp32_wifi_iface, dev_data->mac_addr,
+				     sizeof(dev_data->mac_addr), NET_LINK_ETHERNET);
+
+		esp_err_t ret = esp_wifi_set_mac(WIFI_IF_STA, dev_data->mac_addr);
+
+		if (ret != ESP_OK) {
+			printk("Failed to set mac address: %d\n", ret);
+		}
```
jukkar: I think it would be better if we call
```
		memcpy(dev_data->mac_addr, config->mac_address.addr, 6);
		net_if_set_link_addr(esp32_wifi_iface, dev_data->mac_addr,
				     sizeof(dev_data->mac_addr), NET_LINK_ETHERNET);
```
after the `esp_wifi_set_mac()` has succeeded. Otherwise if the call fails, we have inconsistent mac address in upper net stack compared to esp32 driver.

sylvioalves: Hi, I will update/polish this once the user confirms it works on his side. Thanks for the suggestion.

sylvioalves: Well, let's update it now.

PR #98772, opened 2025-11-03
driver: sensor: st: implemented fifo watermark triggers for ism330dhcx

DenSandu:
implemented a new handler ```sensor_trigger_handler_with_data_t``` in order to have as ```@param``` a buffer of data that can be used by the [ocre-runtime](https://github.com/project-ocre/ocre-runtime) handler. The idea is that the runtime can then push the buffer to a queue of events that will be read by the container: in this way the runtime doesn't have to call the low-level driver functions (such as  ```ism330dhcx_fifo_sensor_tag_get```, ```ism330dhcx_fifo_out_raw_get```) to empty the fifo
Screenshot 2025-11-03 at 11 25 58

### DIFF
--- a/include/zephyr/drivers/sensor.h
+++ b/include/zephyr/drivers/sensor.h
@@ -890,7 +890,7 @@ static inline int sensor_trigger_set(const struct device *dev,
  */
 static inline int sensor_trigger_set_with_data(const struct device *dev,
 				     const struct sensor_trigger *trig,
-				     sensor_trigger_handler_with_data_t handler) // DEN
+				     sensor_trigger_handler_with_data_t handler)
 {
 	const struct sensor_driver_api *api =
 		(const struct sensor_driver_api *)dev->api;
--- a/include/zephyr/drivers/sensor.h
+++ b/include/zephyr/drivers/sensor.h
@@ -386,6 +386,12 @@ enum sensor_attribute {
 	 * Maximum value describing a sensor attribute type.
 	 */
 	SENSOR_ATTR_MAX = INT16_MAX,
+
+	/** FIFO watermark level */
+	SENSOR_ATTR_FIFO_WATERMARK,
+
+	/** FIFO operating mode*/
+	SENSOR_ATTR_FIFO_XL_BATCH,
 };
 
 /**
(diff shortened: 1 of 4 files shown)


avisconti:
@DenSandu can you possibly start looking at sonarqube issues above? Most of them should be very easy to address.
Then put also a milestone, which I guess might be v4.4.0

DenSandu:
@avisconti Thank you for the information. How can I set the milestone? I cannot modify it from the right panel

avisconti:
> @avisconti Thank you for the information. How can I set the milestone? I cannot modify it from the right panel

I just set v4.4.0.
Tomorrow I'll review

avisconti (changes requested):
I think you should squash all 4 commits together

include/zephyr/drivers/sensor.h
```diff
@@ -398,6 +404,21 @@ enum sensor_attribute {
...
+/**
+ * @typedef sensor_trigger_handler_with_data_t
+ * @brief Callback API upon firing of a trigger with raw data
+ *
+ * @param dev Pointer to the sensor device
+ * @param trigger The trigger
+ * @param raw_data Pointer to raw data associated with the trigger
+ * @param len Length of the raw data buffer
+ */
+typedef void (*sensor_trigger_handler_with_data_t)(const struct device *dev,
```
teburd, line 416: We already have a way of doing FIFO handling that doesn't require driver allocated buffers (on the stack or statically) but instead takes a buffer from the caller and returns it back to them.

This looks to be solving the same issue and changing the sensor API in a competing way.

teburd (changes requested):
To me this looks like it is trying to add another way to read sensor FIFOs from the already designed and agreed to method.

include/zephyr/drivers/sensor.h
DenSandu: https://docs.zephyrproject.org/latest/samples/sensor/stream_fifo/README.html#stream_fifo
I found this example using the RTIO framework. Is this the standard way of reading sensor FIFOs you are referring to?

PR #98777, opened 2025-11-03
dts: bindings: clock: nrf-auxpll: add out-div binding

Hi-Im-David:
It was previously possible to set output division setting to an invalid number. These bindings ensure a valid number is set of 0-8.


bjarki-andreasen (dismissed):

include/zephyr/dt-bindings/clock/nrf-auxpll.h
```diff
@@ -13,4 +13,14 @@
 #define NRF_AUXPLL_FREQ_DIV_AUDIO_48K  39845
 #define NRF_AUXPLL_FREQ_DIV_MAX        65535
 
+#define NRFX_AUXPLL_OUT_DIV_DISABLED 0
```
masz-nordic: Assert those values are equal to `NRF_AUXPLL_CTRL_OUTSEL_DIV_...`

Hi-Im-David: Added to auxpll driver with a new macro to make it a little cleaner

masz-nordic (approved):

lstnl (approved):

kartben (dismissed):
the change needs to be described in the migration guide please

Hi-Im-David:
> the change needs to be described in the migration guide please

This shouldn't actually affect any users of released products that use this driver (nRF54h20 and nRF9280) as the out-div is fixed on these products and should only be set in the base dtsi files (`nRF54h20.dtsi` and `nrf9280.dtsi`), not in a board or overlay file. The values currently in these files are correct, just not using the macros. However, this will be different for upcoming SOC's which allow you to configure the AUXPLL frequency, hence the reason for this PR.

PR #98778, opened 2025-11-03
Update zc143ac72mipi co5300 driver and display driver example

### LINKED ISSUE #98892: ZC143AC72MIPI shows green horizontal lines when using running the lvgl samples.
### Describe the bug

ZC143AC72MIPI shows green horizontal lines when using running the lvgl samples.

The target platform this was tested on is the mimxrt700_evk/mimxrt798s/cm33_cpu0

### Regression

- [ ] This is a regression.

### Steps to reproduce

west build -p always -b mimxrt700_evk/mimxrt798s/cm33_cpu0 --shield=zc143ac72mipi ./zephyr/samples/modules/lvgl/demos

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

KATE-WANG-NXP:
1. Add user configuration of buffer address alignment and update area pitch's alignment for display driver sample.
2. Add board specific configuration for mimxrt700_evk lvgl examples. The mimxrt700_evk uses DC8000 to drive all panels which requires a 64-byte align of buffer address and stride.
3. Update CO5300 panel driver
3.1 Fix wrong backlight pin in driver overlay
3.2 Remove the power-on pin configuration in code and binding, and add mipi display panel regulator in panel overlay instead. Set 'regulator-boot-on' to true means the power-on pin will be enabled during system boot.
3.3 Remove 'last_known_framebuffer' from panel data structure. It is not used anywhere
3.4 Fix bug in 'co5300_set_pixel_format' function.
3.5 Fix the issue that the panel does not support start coordinates and the width/height of the updated area being odd value.
       Solution: In panel driver, maintain a full screen-sized buffer, its address and pitch alignment is configurable in device tree and shall be compliant with the display controller's requirements. It can be placed in RAM or if the RAM space is not enough it can also be placed in other memory resion. When there is a frame update request, the updated area will be first filled to the buffer, if the area's size or coordinate is odd, adjust the value so the real updated area covers the requested updated area, then use this buffer to send pixel to panel, this can ensure the updated area's size and coordinate are always even.

### DIFF
--- a/boards/shields/zc143ac72mipi/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.overlay
+++ b/boards/shields/zc143ac72mipi/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.overlay
@@ -25,3 +25,9 @@
 &gpio3 {
 	status = "okay";
 };
+
+&co5300_zc143ac72mipi {
+	ext-ram = <&sram1>;
+	addr-align = <64>;
+	pitch-align = <64>;
+};
--- a/boards/shields/zc143ac72mipi/zc143ac72mipi.overlay
+++ b/boards/shields/zc143ac72mipi/zc143ac72mipi.overlay
@@ -8,6 +8,13 @@
 	chosen {
 		zephyr,display = &co5300_zc143ac72mipi;
 	};
+
+	en_mipi_display_zc143ac72mipi: enable-mipi-display {
+		compatible = "regulator-fixed";
+		regulator-name = "en_mipi_display";
+		enable-gpios = <&nxp_mipi_connector 32 GPIO_ACTIVE_HIGH>;
+		regulator-boot-on;
+	};
 };
 
 &zephyr_mipi_dsi {
@@ -20,9 +27,8 @@
 		compatible = "chipone,co5300";
 		reg = <0x0>;
 		reset-gpios = <&nxp_mipi_connector 21 GPIO_ACTIVE_HIGH>;
-		backlight-gpios = <&nxp_mipi_connector 34 GPIO_ACTIVE_HIGH>;
+		backlight-gpios = <&nxp_mipi_connector 0 GPIO_ACTIVE_HIGH>;
 		tear-effect-gpios = <&nxp_mipi_connector 22 GPIO_ACTIVE_HIGH>;
-		power-gpios = <&nxp_mipi_connector 32 GPIO_ACTIVE_HIGH>;
 		data-lanes = <1>;
 		width = <466>;
 		height = <466>;
--- a/drivers/display/display_co5300.c
+++ b/drivers/display/display_co5300.c
@@ -17,7 +17,7 @@ LOG_MODULE_REGISTER(co5300, CONFIG_DISPLAY_LOG_LEVEL);
 #include <zephyr/drivers/gpio.h>
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/kernel.h>
-
+#include <zephyr/linker/devicetree_regions.h>
 #include <fsl_lcdif.h>
 #include <fsl_mipi_dsi.h>
 
@@ -40,11 +40,13 @@ struct co5300_config {
 };
 
 struct co5300_data {
-	uint8_t *last_known_framebuffer;
 	uint8_t pixel_format;
 	uint8_t bytes_per_pixel;
 	struct gpio_callback tear_effect_gpio_cb;
 	struct k_sem tear_effect_sem;
+	/* Pointer to framebuffer */
+	uint8_t *frame_ptr;
+	uint32_t frame_pitch;
 };
 
 /* Organized as MIPI_CMD | SIZE OF MIPI PARAM | MIPI PARAM */
@@ -61,9 +63,6 @@ uint8_t lcm_init_cmds[] = {	0xFE, 0x1, 0x20,
 				0x63, 0x1, 0xFF,
 				0x2A, 0x4, 0x00, 0x06, 0x01, 0xD7,
 				0x2B, 0x4, 0x00, 0x00, 0x01, 0xD1};
-
-uint8_t pixel_format_bgr_cmds[] = {0x36, 0x1, 0x8};
-
 
 static void co5300_tear_effect_isr_handler(const struct device *gpio_dev,
 			struct gpio_callback *cb, uint32_t pins)
@@ -96,6 +95,53 @@ static int co5300_blanking_off(const struct device *dev)
 	return -ENOTSUP;
 }
 
+static void co5300_copy_and_adjust_coordinates(const struct device *dev,
+		const uint16_t x, const uint16_t y,
+		const struct display_buffer_descriptor *desc,
+		const void *buf,
+		uint16_t *local_x, uint16_t *local_y,
+		struct display_buffer_descriptor *local_desc)
+{
+	struct co5300_data *data = dev->data;
+	const uint8_t *src;
+	uint8_t *dst;
+
+	/* Copy the update area to the framebuffer */
+	src = buf;
+	dst = data->frame_ptr + (y * data->frame_pitch * data->bytes_per_pixel)
+		+ (x * data->bytes_per_pixel);
+	for (uint16_t row = 0; row < desc->height; row++) {
+		memcpy(dst, src, desc->width * data->bytes_per_pixel);
+		src += desc->pitch * data->bytes_per_pixel;
+		dst += data->frame_pitch * data->bytes_per_pixel;
+	}
+
+	/*
+	 * Initialize descriptor for local frame buffer.
+	 * The start coordinates and the width/height of the updated area
+	 * cannot be odd value for the panel. Adjust them to be even.
+	 */
+	if (x % 2 != 0) {
+		*local_x = x - 1;
+		local_desc->width = desc->width + 1U;
+	} else {
+		*local_x = x;
+		local_desc->width = desc->width;
+	}
+	if (y % 2 != 0) {
+		*local_y = y - 1;
+		local_desc->height = desc->height + 1U;
+	} else {
+		*local_y = y;
+		local_desc->height = desc->height;
+	}
+	local_desc->width = ROUND_UP(local_desc->width, 2U);
+	local_desc->height = ROUND_UP(local_desc->height, 2U);
+	local_desc->pitch = data->frame_pitch;
+	local_desc->frame_incomplete = desc->frame_incomplete;
+	local_desc->buf_size = local_desc->width * local_desc->height * data->bytes_per_pixel;
+}
+
 static int co5300_write(const struct device *dev,
 		const uint16_t x, const uint16_t y,
 		const struct display_buffer_descriptor *desc,
@@ -104,27 +150,46 @@ static int co5300_write(const struct device *dev,
 	const struct co5300_config *config = dev->config;
 	struct co5300_data *data = dev->data;
 	int ret;
-	uint16_t start_xpos;
-	uint16_t end_xpos;
-	uint16_t start_ypos;
-	uint16_t end_ypos;
-	const uint8_t *framebuffer_addr;
+	uint16_t start_pos;
+	uint16_t end_pos;
 	uint8_t cmd_params[4];
 	struct mipi_dsi_msg msg = {0};
 	uint32_t total_bytes_sent = 0U;
-	uint32_t framebuffer_size = 0U;
 	int bytes_written = 0;
+	const uint8_t *src;
+	uint32_t tx_size = 0U;
+	uint16_t local_x, local_y;
+	struct display_buffer_descriptor local_desc = {0};
 
-	LOG_DBG("WRITE:: W=%d, H=%d @%d,%d", desc->width, desc->height, x, y);
+	/* Check whether the updated area is outside of the panel frame. */
+	if ((x > config->panel_width) || (y > config->panel_height) ||
+			((x + desc->width) > config->panel_width) ||
+			((y + desc->height) > config->panel_height)) {
+		LOG_ERR("Update area outside panel dimensions");
+		return -EINVAL;
+	}
 
-	/* Set column address of target area */
+	/* Check whether the updated area is valid */
+	if (desc->width == 0 || desc->height == 0) {
+		LOG_ERR("The height/width of the update area cannot be 0");
+		return -EINVAL;
+	}
+
+	/* Copy data to framebuffer and adjust coordinates for even values */
+	co5300_copy_and_adjust_coordinates(dev, x, y, desc, buf, &local_x, &local_y, &local_desc);
+
+	/*
+	 * Set column address of target area. The circular panel actually starts
+	 * to show from row 6, row 0~5 are cut off physically. The actual display
+	 * area is row 6~472 and line 0~466. So adjust coordinates accordingly.
+	 */
 	/* First two bytes are starting X coordinate */
-	start_xpos = x;
-	sys_put_be16(start_xpos, &cmd_params[0]);
+	start_pos = local_x + 6U;
+	sys_put_be16(start_pos, &cmd_params[0]);
 
 	/* Second two bytes are ending X coordinate */
-	end_xpos = x + desc->width - 1;
-	sys_put_be16(end_xpos, &cmd_params[2]);
+	end_pos = local_x + local_desc.width + 5U;
+	sys_put_be16(end_pos, &cmd_params[2]);
 	ret = mipi_dsi_dcs_write(config->mipi_dsi, config->channel,
 				MIPI_DCS_SET_COLUMN_ADDRESS, cmd_params,
 				sizeof(cmd_params));
@@ -134,12 +199,12 @@ static int co5300_write(const struct device *dev,
 
 	/* Set page address of target area */
 	/* First two bytes are starting Y coordinate */
-	start_ypos = y;
-	sys_put_be16(start_ypos, &cmd_params[0]);
+	start_pos = local_y;
+	sys_put_be16(start_pos, &cmd_params[0]);
 
 	/* Second two bytes are ending Y coordinate */
-	end_ypos = y + desc->height - 1;
-	sys_put_be16(end_ypos, &cmd_params[2]);
+	end_pos = local_y + local_desc.height - 1;
+	sys_put_be16(end_pos, &cmd_params[2]);
 	ret = mipi_dsi_dcs_write(config->mipi_dsi, config->channel,
 				MIPI_DCS_SET_PAGE_ADDRESS, cmd_params,
 				sizeof(cmd_params));
@@ -155,34 +220,40 @@ static int co5300_write(const struct device *dev,
 		k_sem_take(&data->tear_effect_sem, K_FOREVER);
 	}
 
-
-	/* Start filling out the framebuffer */
-	framebuffer_addr = buf;
-	framebuffer_size = desc->width * desc->height * data->bytes_per_pixel;
+	/* Start memory write. */
+	/* The address and the total pixel size of the updated area. */
+	src = data->frame_ptr + (local_y * data->frame_pitch * data->bytes_per_pixel)
+		+ (local_x * data->bytes_per_pixel);
+	tx_size = local_desc.buf_size;
 
 	msg.type = MIPI_DSI_DCS_LONG_WRITE;
 	msg.flags = MCUX_DSI_2L_FB_DATA;
-	msg.user_data = (void *)desc;
+	msg.user_data = &local_desc;
 	msg.cmd = MIPI_DCS_WRITE_MEMORY_START;
 
-	while (framebuffer_size > 0) {
-		msg.tx_len = framebuffer_size;
-		msg.tx_buf = framebuffer_addr;
+	while (tx_size > 0) {
+		msg.tx_len = tx_size;
+		msg.tx_buf = src;
 		bytes_written = (int)mipi_dsi_transfer(config->mipi_dsi, config->channel, &msg);
 		if (bytes_written < 0) {
 			return bytes_written;
 		}
 
-		/* Advance source pointer and decrement remaining */
-		if (desc->pitch > desc->width) {
-			total_bytes_sent += bytes_written;
-			framebuffer_addr += bytes_written + total_bytes_sent /
-				(desc->width * data->bytes_per_pixel) *
-				((desc->pitch - desc->width) * data->bytes_per_pixel);
-		} else {
-			framebuffer_addr += bytes_written;
+		tx_size -= bytes_written;
+
+		if (tx_size == 0U) {
+			break;
 		}
-		framebuffer_size -= bytes_written;
+
+		/* Advance source pointer and decrement remaining */
+		if (local_desc.pitch > local_desc.width) {
+			total_bytes_sent += bytes_written;
+			src += bytes_written + total_bytes_sent /
+				(local_desc.width * data->bytes_per_pixel) *
+				((local_desc.pitch - local_desc.width) * data->bytes_per_pixel);
+		} else {
+			src += bytes_written;
+		}
 
 		/* All future commands should use WRITE_MEMORY_CONTINUE */
 		msg.cmd = MIPI_DCS_WRITE_MEMORY_CONTINUE;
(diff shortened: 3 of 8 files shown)


drivers/display/display_co5300.c
```diff
@@ -104,27 +100,34 @@ static int co5300_write(const struct device *dev,
...
 	LOG_DBG("WRITE:: W=%d, H=%d @%d,%d", desc->width, desc->height, x, y);
 
-	/* Set column address of target area */
+	/*
+	 * Set column address of target area. The circular panel actually starts
+	 * to show from row 6, row 0~5 are cut off physically. The actual display
+	 * area is row 6~472 and line 0~466. So adjust coordinates accordingly.
+	 */
 	/* First two bytes are starting X coordinate */
-	start_xpos = x;
```
EmilioCBen, line 122: Should we update the width and height of the panel in the shield overlay instead of adding this runtime calculations?

samples/drivers/display/src/main.c
```diff
@@ -241,8 +241,8 @@ int main(void)
 		rect_w = 16;
 		rect_h = 8;
 	} else {
-		rect_w = 2;
-		rect_h = 1;
+		rect_w = 4;
+		rect_h = 2;
 	}
```
mmahadevan108, line 247: Why is this changed?

samples/modules/lvgl/demos/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright 2025 NXP
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+CONFIG_LV_Z_MEM_POOL_SIZE=81920
+# RT700 uses DC8000 to drive panel which requires a 64 byte align of buffer address and stride.
+CONFIG_LV_DRAW_BUF_STRIDE_ALIGN=64
+CONFIG_LV_DRAW_BUF_ALIGN=64
```
mmahadevan108, line 10: If this is something needed for this panel, then I would suggest moving these to `boards/shields/zc143ac72mipi/Kconfig.defconfig` or `boards/shields/zc143ac72mipi/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf` (if board specific) and encapsulate this within `if LVGL   ....  endif`

boards/shields/zc143ac72mipi/zc143ac72mipi.overlay
```diff
@@ -8,6 +8,19 @@
...
+		regulator-name = "en_mipi_display";
+		enable-gpios = <&nxp_mipi_connector 32 GPIO_ACTIVE_HIGH>;
+		regulator-boot-on;
+	};
+
+	lvgl_pointer {
+		compatible = "zephyr,lvgl-pointer-input";
+		display = <&co5300_zc143ac72mipi>;
+		invert-y;
+	};
```
mmahadevan108: This seems unrelated to this change and related to touch support. Please delete this and include it in a separate PR that adds touch support.

drivers/display/display_co5300.c
```diff
@@ -173,6 +175,12 @@ static int co5300_write(const struct device *dev,
 			return bytes_written;
 		}
 
+		framebuffer_size -= bytes_written;
+
+		if (framebuffer_size == 0U) {
+			break;
+		}
+
```
uLipe, line 247: Is it guaranteed that a potential underflow will not occur here?

uLipe (commented):
I just have a question.

samples/drivers/display/Kconfig
```diff
@@ -0,0 +1,16 @@
+# Private config options for display sample
+
+# Copyright (c) 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+mainmenu "Display sample application"
+
+source "Kconfig.zephyr"
+
+config BUFFER_ADDR_ALIGN
```
JarmouniA:
```suggestion
config SAMPLE_BUFFER_ADDR_ALIGN
```

samples/drivers/display/Kconfig
```diff
@@ -0,0 +1,16 @@
...
+	int "The display buffer address alignment"
+	default 1
+
+config PITCH_ALIGN
```
JarmouniA:
```suggestion
config SAMPLE_PITCH_ALIGN
```

samples/drivers/display/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright 2025 NXP
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+CONFIG_BUFFER_ADDR_ALIGN=64
+CONFIG_PITCH_ALIGN=64
```
JarmouniA: Are these params dependent on the board or the display panel?

samples/modules/lvgl/demos/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf
JarmouniA: the LV mempool size is app-specific, so should stay here.

EmilioCBen:
Fixes #98892

samples/drivers/display/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf
KATE-WANG-NXP: The display controller on board

drivers/display/display_co5300.c
KATE-WANG-NXP: Hi I updated the driver to use an internal maintained buffer now so I think there will not be such issue.

samples/drivers/display/src/main.c
KATE-WANG-NXP: reverted

drivers/display/display_co5300.c
KATE-WANG-NXP: Hi, it is not the height and width, it is that to the co5300 controller the panel is 472x466 while actually the first 6 rows are cut off physically to make it a round panel, so if the higher app wants to update from 0,0, we need to pass 6,0 to the co5300 controller

drivers/display/Kconfig.co5300
```diff
@@ -1,10 +1,41 @@
...
 	select MIPI_DSI
 	depends on DT_HAS_CHIPONE_CO5300_ENABLED
 	help
 	  Enable driver for CO5300 display controller
+
+if CO5300
+
+config CO5300_EXTERNAL_FB_MEM
+	bool "Use external defined address for frame buffer"
+	default n
```
faxe1008: not needed for boolean

drivers/display/Kconfig.co5300
```diff
@@ -1,10 +1,41 @@
...
+	help
+	  Use external defined address for frame buffer, set to
+	  true when RAM is out of space to place the frame buffer externally
+
+if CO5300_EXTERNAL_FB_MEM
+
+config CO5300_EXTERNAL_FB_ADDR
+	hex "External frame buffer address"
+	help
```
faxe1008: Maybe move the if into here as a `depends on` for better discoverability in https://docs.zephyrproject.org/latest/kconfig.html

drivers/display/Kconfig.co5300
```diff
@@ -1,10 +1,41 @@
...
+	  The address of the externally defined frame buffer memory.
+	  Please ensure the address is compliant with the display controller's
+	  requirements and valid in the system memory map
+
+endif
+
+if !CO5300_EXTERNAL_FB_MEM
+
+config CO5300_FB_ADDR_ALIGN
+	int "Frame buffer address alignment"
```
faxe1008: Same here

drivers/display/display_co5300.c
```diff
@@ -104,27 +102,83 @@ static int co5300_write(const struct device *dev,
...
+	}
+	if (y % 2 != 0) {
+		local_y = y - 1;
+		local_desc.height = desc->height + 1U;
+	} else {
+		local_y = y;
+		local_desc.height = desc->height;
+	}
+	local_desc.width = ROUND_UP(local_desc.width, 2U);
+	local_desc.height = ROUND_UP(local_desc.height, 2U);
```
faxe1008: Are those even needed? Further up you make them even?

drivers/display/display_co5300.c
```diff
@@ -298,43 +363,27 @@ static int co5300_init(const struct device *dev)
 	}
 
 	/* Perform GPIO Reset */
-	if (config->power_gpios.port != NULL) {
-		ret = gpio_pin_configure_dt(&config->power_gpios, GPIO_OUTPUT_INACTIVE);
+	if (config->reset_gpios.port != NULL) {
```
faxe1008, line 364: This init is pretty long, the reset would be perfect to factor out into a seperate function.

drivers/display/display_co5300.c
KATE-WANG-NXP: Yes the update area start coordinate and the size shall all be even value, so if x = 1, width = 4, then x->0, width->5, and the width shall add another 1

drivers/display/display_co5300.c
```diff
@@ -230,33 +290,39 @@ static int co5300_set_pixel_format(const struct device *dev,
...
-
-		cmd_params = pixel_format_bgr_cmds[2];
 		data->pixel_format = MIPI_DSI_PIXFMT_RGB565;
-		cmd_params = MIPI_DCS_PIXEL_FORMAT_16BIT;
+		cmd_params[0] = MIPI_DCS_ADDRESS_MODE_BGR;
+		cmd_params[1] = MIPI_DCS_PIXEL_FORMAT_16BIT;
 		data->bytes_per_pixel = 2;
 		break;
 	case PIXEL_FORMAT_RGB_888:
 		data->pixel_format = MIPI_DSI_PIXFMT_RGB888;
```
faxe1008: Hmm usually the pixel format is set to the data only after the command succeeded otherwise you get inconsistent state.

samples/drivers/display/src/main.c
```diff
@@ -270,6 +270,8 @@ int main(void)
 		rect_w = capabilities.x_resolution;
 	}
 
+	rect_w = ROUND_UP(rect_w, CONFIG_SAMPLE_PITCH_ALIGN);
+
```
zejiang0jason, line 275: suggest change:
```c
size_t rect_pitch = ROUND_UP(rect_w, CONFIG_SAMPLE_PITCH_ALIGN);
size_t full_screen_pitch = ROUND_UP(capabilities.x_resolution, CONFIG_SAMPLE_PITCH_ALIGN);

buf_size = MAX(rect_pitch * rect_h, full_screen_pitch * h_step);
```

The variables `rect_pitch` and `full_screen_pitch` can be used later in `buf_desc` setting.

drivers/display/display_co5300.c
```diff
@@ -104,27 +102,83 @@ static int co5300_write(const struct device *dev,
...
+		local_y = y - 1;
+		local_desc.height = desc->height + 1U;
+	} else {
+		local_y = y;
+		local_desc.height = desc->height;
+	}
+	local_desc.width = ROUND_UP(local_desc.width, 2U);
+	local_desc.height = ROUND_UP(local_desc.height, 2U);
+	local_desc.width = desc->width + 1;
+	local_desc.height = desc->height + 1;
```
zejiang0jason: why assign the `local_desc.width` and `local_desc.height` again?

drivers/display/display_co5300.c
zejiang0jason: Is it zc143ac72 panel specific? If this controller is used by other panels, could this value be different?

drivers/display/display_co5300.c
```diff
@@ -104,27 +102,83 @@ static int co5300_write(const struct device *dev,
...
+	local_desc.pitch = data->frame_pitch;
+	local_desc.frame_incomplete = desc->frame_incomplete;
```
zejiang0jason: can we only send out whole dirty region, when the `frame_incomplete` is false?

drivers/display/display_co5300.c
```diff
@@ -155,34 +209,40 @@ static int co5300_write(const struct device *dev,
...
+	msg.user_data = &local_desc;
 	msg.cmd = MIPI_DCS_WRITE_MEMORY_START;
 
-	while (framebuffer_size > 0) {
-		msg.tx_len = framebuffer_size;
-		msg.tx_buf = framebuffer_addr;
+	while (tx_size > 0) {
+		msg.tx_len = tx_size;
+		msg.tx_buf = src;
 		bytes_written = (int)mipi_dsi_transfer(config->mipi_dsi, config->channel, &msg);
```
zejiang0jason, line 237: the `tx_size` is the whole data size, is the data to be sent continuous in memory? if not, can `mipi_dsi_transfer` support non-continuous data?

drivers/display/display_co5300.c
```diff
@@ -104,27 +102,83 @@ static int co5300_write(const struct device *dev,
...
+		local_y = y;
+		local_desc.height = desc->height;
+	}
+	local_desc.width = ROUND_UP(local_desc.width, 2U);
+	local_desc.height = ROUND_UP(local_desc.height, 2U);
+	local_desc.width = desc->width + 1;
+	local_desc.height = desc->height + 1;
+	local_desc.pitch = data->frame_pitch;
+	local_desc.frame_incomplete = desc->frame_incomplete;
+	local_desc.buf_size = local_desc.width * local_desc.height * data->bytes_per_pixel;
```
zejiang0jason: if possible, suggest make this area adjustment as a static function, currently this function is very large.

drivers/display/display_rm67162.c
JarmouniA, line 1: unrelated to this PR?

drivers/mipi_dbi/mipi_dbi_nxp_dcnano_lcdif.c
JarmouniA, line 1: unrelated to this PR?

drivers/display/display_co5300.c
```diff
@@ -446,6 +509,20 @@ static DEVICE_API(display, co5300_api) = {
...
 };
 
+#ifdef CONFIG_CO5300_EXTERNAL_FB_MEM
+#define CO5300_FRAMEBUFFER_DECL(node_id)
+#define CO5300_FRAMEBUFFER(node_id) (uint8_t *)CONFIG_CO5300_EXTERNAL_FB_ADDR
+#else
+/* CO5300 supports RGB565 and RGB888 pixel formats, so to ensure the buffer
+ * region is enough, use RGB888 which is 3Bpp to calculate the buffer size
+ */
+#define CO5300_FRAMEBUFFER_DECL(node_id) uint8_t __aligned(CONFIG_CO5300_FB_ADDR_ALIGN)		\
```
JarmouniA: I think this is not the right approach, ext memory buffers should be handled by the SoC's display controller, similar to
https://github.com/zephyrproject-rtos/zephyr/blob/f1c2ce4f4026bd5b212c6e4cc7d754aba69b189e/drivers/display/display_stm32_ltdc.c#L537

JarmouniA (changes requested):

drivers/display/display_co5300.c
KATE-WANG-NXP: @zejiang0jason Yes the driver display_co5300 is used by zc143ac72 only

drivers/display/display_co5300.c
KATE-WANG-NXP: sorry, typo

drivers/display/display_co5300.c
KATE-WANG-NXP: yes, it can if using lcdif

drivers/display/display_co5300.c
KATE-WANG-NXP: Yes that is what we do now in every update, the frame_incomplete param in the descriptor passed down from app level is needed and used by some of low level driver like lcdic or dbi_spi, we just preserve and pass down the param, it is not actually used.

drivers/display/display_rm67162.c
KATE-WANG-NXP: Move to another pr https://github.com/zephyrproject-rtos/zephyr/pull/99880

drivers/mipi_dbi/mipi_dbi_nxp_dcnano_lcdif.c
KATE-WANG-NXP: Move to another pr https://github.com/zephyrproject-rtos/zephyr/pull/99880

boards/shields/zc143ac72mipi/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.overlay
JarmouniA, line 1: Please move board DT changes to a separate commit.

drivers/display/display_co5300.c
```diff
@@ -446,6 +527,21 @@ static DEVICE_API(display, co5300_api) = {
 	.set_orientation = co5300_set_orientation,
 };
 
+#if DT_INST_NODE_HAS_PROP(0, ext_ram)
```
JarmouniA: This check should be instance-dependent, not always fixed to instance 0.

drivers/display/display_co5300.c
```diff
@@ -73,7 +72,6 @@ static void co5300_tear_effect_isr_handler(const struct device *gpio_dev,
 	k_sem_give(&data->tear_effect_sem);
 }
 
-
```
JarmouniA, line 76: Unrelated change.

drivers/display/display_co5300.c
```diff
@@ -230,37 +300,48 @@ static int co5300_set_pixel_format(const struct device *dev,
...
-		mipi_dsi_dcs_write(config->mipi_dsi, config->channel,
-			cmd_register, &cmd_params, cmd_param_size);
-
-		cmd_params = pixel_format_bgr_cmds[2];
-		data->pixel_format = MIPI_DSI_PIXFMT_RGB565;
-		cmd_params = MIPI_DCS_PIXEL_FORMAT_16BIT;
-		data->bytes_per_pixel = 2;
+		mipi_pixel_format = MIPI_DSI_PIXFMT_RGB565;
+		bytes_per_pixel = 2U;
+		cmd_params[0] = MIPI_DCS_ADDRESS_MODE_BGR;
```
JarmouniA, line 312: When running the Display sample, is the output conform to what is described in the readme (color-wise)?
https://docs.zephyrproject.org/latest/samples/drivers/display/README.html

dts/bindings/display/chipone,co5300.yaml
```diff
@@ -29,8 +29,22 @@ properties:
...
+    default: 1
+    description: |
+      The CO5300 maintains an internal frame buffer. Its address must be
+      compliant with the display controller's requirement.
+
+  ext-ram:
+    type: phandle
+    description: |
+      External SRAM/SDRAM in which the panel's internal frame buffer will
+      be stored. If not defined, internal RAM will be used.
```
JarmouniA:
```suggestion
      Secondary RAM in which the panel controller's frame buffer will
      be stored. If not defined, the 'zephyr,ram' chosen memory will be used.
```

JarmouniA:
```suggestion
      Secondary RAM in which the panel controller's frame buffer will
      be stored. If not defined, the 'zephyr,ram' chosen memory will be used.
```

dts/bindings/display/chipone,co5300.yaml
```diff
@@ -29,8 +29,22 @@ properties:
...
-  power-gpios:
-    type: phandle-array
+  pitch-align:
+    type: int
+    default: 1
     description: |
-      The RESETn pin is asserted to disable the controller causing a hard
-      reset. The controller receives this as an active-low signal.
+      The CO5300 maintains an internal frame buffer. Its stride must be
+      compliant with the display controller's requirement.
```
JarmouniA:
```suggestion
      The CO5300 driver maintains an internal frame buffer. This property
      serves for defining a stride for it that's compliant with the SoC's
      display controller's requirements.
```

drivers/display/display_co5300.c
```diff
@@ -446,6 +527,21 @@ static DEVICE_API(display, co5300_api) = {
...
+#define FRAME_BUFFER_SECTION
+#endif /* DT_INST_NODE_HAS_PROP(0, ext_ram) */
+
+#define CO5300_FRAMEBUFFER_DECL(node_id)							\
+	/* frame buffer aligned to cache line width for optimal cache flushing */		\
+	FRAME_BUFFER_SECTION static uint8_t __aligned(DT_INST_PROP(node_id, addr_align))	\
+			co5300_frame_buffer_##node_id[DT_INST_PROP(node_id, height) * 3U *	\
+			ROUND_UP(DT_INST_PROP(node_id, width),					\
+			DT_INST_PROP(node_id, pitch_align))]
+#define CO5300_FRAMEBUFFER(node_id) co5300_frame_buffer_##node_id
```
JarmouniA, line 542:
```suggestion

#define CO5300_FRAMEBUFFER(node_id) co5300_frame_buffer_##node_id
```

drivers/display/display_co5300.c
```diff
@@ -446,6 +527,21 @@ static DEVICE_API(display, co5300_api) = {
...
+#if DT_INST_NODE_HAS_PROP(0, ext_ram)
+#define FRAME_BUFFER_SECTION									\
+	Z_GENERIC_SECTION(LINKER_DT_NODE_REGION_NAME(DT_INST_PHANDLE(0, ext_ram)))
+#else
+#define FRAME_BUFFER_SECTION
+#endif /* DT_INST_NODE_HAS_PROP(0, ext_ram) */
+
+#define CO5300_FRAMEBUFFER_DECL(node_id)							\
+	/* frame buffer aligned to cache line width for optimal cache flushing */		\
+	FRAME_BUFFER_SECTION static uint8_t __aligned(DT_INST_PROP(node_id, addr_align))	\
```
JarmouniA: If the buffer's address alignment is just for cache friendliness, no need for the DT property, use the value of `CONFIG_DCACHE_LINE_SIZE`.

drivers/display/display_co5300.c
KATE-WANG-NXP: Yes, I tried to use instance but failed to find a reference, do you have an example on how to pass the instance in a #if?

drivers/display/display_co5300.c
KATE-WANG-NXP: No it is required by the SoC's display controller's, like described in [dts/bindings/display/chipone,co5300.yaml](https://github.com/zephyrproject-rtos/zephyr/pull/98778/files/4daf132acb7e4faf95f8f3335a7e63bf6c4d997b#diff-0eb21fa455ca3cde448d4705648191c39e8d335fe19a7097b76ef96001a96eac), eg. DCNano on RT700 requires a 64 byte alignment

drivers/display/display_co5300.c
KATE-WANG-NXP: Yes

drivers/display/display_co5300.c
JarmouniA: Use `COND_CODE_1` instead

drivers/display/display_co5300.c
JarmouniA: But that's not what the comment says.

KATE-WANG-NXP: Sorry typo

dts/bindings/display/chipone,co5300.yaml
```diff
@@ -29,8 +29,22 @@ properties:
...
+
+  addr-align:
+    type: int
+    default: 1
+    description: |
+      The CO5300 maintains an internal frame buffer. Its address must be
+      compliant with the display controller's requirement.
```
JarmouniA:
```suggestion
      The CO5300 driver maintains an internal frame buffer. This property
      serves for defining an alignment for its address that's compliant
      with the SoC's display controller's requirements.
```

JarmouniA (approved):

faxe1008 (approved):

zejiang0jason (approved):

mmahadevan108 (approved):

PR #98784, opened 2025-11-03
Blueototh: UUID: add string to uuid

chengkai15:
# Bluetooth: UUID: add string to UUID conversion function

## Summary
This PR adds a new `bt_str_to_uuid()` function to convert string representations to Bluetooth UUID structures, complementing the existing `bt_uuid_to_str()` function. This provides bidirectional conversion between UUID structures and their string representations.

## Changes
- **New API**: Add `bt_str_to_uuid()` function in `uuid.h` header
- **Implementation**: Implement string parsing for all UUID formats in `uuid.c`
- **Format Support**: Handle 16-bit, 32-bit, and 128-bit UUID string formats

## Technical Details
- **16-bit UUID**: Parses 4-character hexadecimal strings (e.g., "1800")
- **32-bit UUID**: Parses 8-character hexadecimal strings (e.g., "00001800")  
- **128-bit UUID**: Parses standard UUID format (e.g., "00001800-0000-1000-8000-00805f9b34fb")
- **Validation**: Returns `true` on success, `false` on invalid input format
- **Error Handling**: Comprehensive format validation with proper return codes

## Features
- **Bidirectional Conversion**: Complements existing `bt_uuid_to_str()` function
- **Flexible Input**: Supports all standard UUID string formats
- **Type Detection**: Automatically detects and sets appropriate UUID type (16/32/128-bit)
- **Robust Parsing**: Uses `sscanf` with exact format matching for reliable parsing

## Usage Example
```c
struct bt_uuid uuid;
const char *uuid_str = "1800";

if (bt_str_to_uuid(uuid_str, &uuid)) {
    // Successfully converted string to UUID
    // uuid.type will be BT_UUID_TYPE_16
    // BT_UUID_16(&uuid)->val will be 0x1800
}
```

## Testing
- Verified conversion of 16-bit, 32-bit, and 128-bit UUID strings
- Tested edge cases and invalid input formats
- Confirmed proper type detection and value assignment
- Validated round-trip conversion with `bt_uuid_to_str()`

## Impact
- **New Feature**: Adds missing string-to-UUID conversion capability
- **Backward Compatible**: No changes to existing APIs
- **Minimal Overhead**: Small footprint with comprehensive functionality
- **Utility Enhancement**: Useful for configuration, debugging, and dynamic UUID handling

This function fills an important gap in the Bluetooth UUID utilities, enabling applications to easily convert user-input or configuration strings into proper UUID structures for Bluetooth operations.

### DIFF
--- a/include/zephyr/bluetooth/uuid.h
+++ b/include/zephyr/bluetooth/uuid.h
@@ -5249,6 +5249,25 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
  */
 void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len);
 
+/** @brief Convert string to Bluetooth UUID.
+ *
+ *  Converts string to Bluetooth UUID. If the string UUID is 16 bit, the converted
+ *  bt_uuid will be bt_uuid_16 type. If the string UUID is 32 bit UUID, the converted
+ *  bt_uuid will be bt_uuid_32 type. If the string UUID is 128 bit UUID, the converted
+ *  bt_uuid will be bt_uuid_128 type.
+ *
+ *  @note The string can be in any format, 16-bit, 32-bit or 128-bit format. And the
+ *  UUID pointer should point to a bt_uuid large enough to hold the converted UUID.
+ *  For example: 128 bit format "00001101-0000-1000-8000-00805F9B34FB",
+ *  the converted bt_uuid will be bt_uuid_128 type.
+ *
+ *  @param str Pointer to string to convert
+ *  @param uuid Pointer where to put converted UUID
+ *
+ *  @return 0 if the string was converted successfully, otherwise negative error
+ */
+int bt_uuid_from_str(const char *str, struct bt_uuid *uuid);
+
 #ifdef __cplusplus
 }
 #endif
--- a/subsys/bluetooth/host/uuid.c
+++ b/subsys/bluetooth/host/uuid.c
@@ -1,6 +1,7 @@
 /* uuid.c - Bluetooth UUID handling */
 
 /*
+ * Copyright (c) 2025 Xiaomi Corporation
  * Copyright (c) 2015-2016 Intel Corporation
  *
  * SPDX-License-Identifier: Apache-2.0
@@ -9,6 +10,7 @@
 #include <errno.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <string.h>
 
 #include <zephyr/bluetooth/uuid.h>
@@ -128,5 +130,51 @@ void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len)
 	default:
 		(void)memset(str, 0, len);
 		return;
+	}
+}
+
+int bt_uuid_from_str(const char *str, struct bt_uuid *uuid)
+{
+	if (str == NULL || uuid == NULL) {
+		return -EINVAL;
+	}
+
+	switch (strlen(str)) {
+	case BT_UUID_SIZE_16 * 2: {
+		uint16_t *p = &BT_UUID_16(uuid)->val;
+
+		if (sscanf(str, "%04hx", p) != 1) {
+			return -EINVAL;
+		}
+
+		uuid->type = BT_UUID_TYPE_16;
+		return 0;
+	}
+	case BT_UUID_SIZE_32 * 2: {
+		uint32_t *p = &BT_UUID_32(uuid)->val;
+
+		if (sscanf(str, "%08x", p) != 1) {
+			return -EINVAL;
+		}
+
+		uuid->type = BT_UUID_TYPE_32;
+		return 0;
+	}
+	case BT_UUID_STR_LEN - 1: {
+		uint8_t *p = BT_UUID_128(uuid)->val;
+
+		if (sscanf(str,
+			   "%02hhx%02hhx%02hhx%02hhx-%02hhx%02hhx-%02hhx%02hhx"
+			   "-%02hhx%02hhx-%02hhx%02hhx%02hhx%02hhx%02hhx%02hhx",
+			   &p[15], &p[14], &p[13], &p[12], &p[11], &p[10], &p[9], &p[8], &p[7],
+			   &p[6], &p[5], &p[4], &p[3], &p[2], &p[1], &p[0]) != BT_UUID_SIZE_128) {
+			return -EINVAL;
+		}
+
+		uuid->type = BT_UUID_TYPE_128;
+		return 0;
+	}
+	default:
+		return -EINVAL;
 	}
 }
--- a/tests/bluetooth/uuid/src/test_bt_uuid_from_str.c
+++ b/tests/bluetooth/uuid/src/test_bt_uuid_from_str.c
@@ -1,0 +1,132 @@
+/* Copyright (c) 2025 Xiaomi Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+#include <zephyr/ztest.h>
+#include <zephyr/bluetooth/uuid.h>
+
+ZTEST_SUITE(bt_uuid_from_str, NULL, NULL, NULL, NULL, NULL);
+
+ZTEST(bt_uuid_from_str, test_uuid_from_str_16)
+{
+	struct bt_uuid_16 uuid = {0};
+	int ret;
+
+	ret = bt_uuid_from_str("180d", &uuid.uuid);
+	zassert_true(ret == 0, "Failed to parse 16-bit UUID");
+	zassert_true(bt_uuid_cmp(&uuid.uuid, BT_UUID_DECLARE_16(0x180d)) == 0,
+		     "Parsed UUID does not match expected 16-bit UUID");
+}
+
+ZTEST(bt_uuid_from_str, test_uuid_from_str_32)
+{
+	struct bt_uuid_32 uuid = {0};
+	int ret;
+
+	ret = bt_uuid_from_str("abcdef12", &uuid.uuid);
+	zassert_true(ret == 0, "Failed to parse 32-bit UUID");
+	zassert_true(bt_uuid_cmp(&uuid.uuid, BT_UUID_DECLARE_32(0xabcdef12)) == 0,
+		     "Parsed UUID does not match expected 32-bit UUID");
+}
+
+ZTEST(bt_uuid_from_str, test_uuid_from_str_128)
+{
+	struct bt_uuid_128 uuid = {0};
+	int ret;
+
+	ret = bt_uuid_from_str("00001101-0000-1000-8000-00805f9b34fb", &uuid.uuid);
+	zassert_true(ret == 0, "Failed to parse 128-bit UUID");
+	struct bt_uuid_128 expected =
+		BT_UUID_INIT_128(0xfb, 0x34, 0x9b, 0x5f, 0x80, 0x00, 0x00, 0x80, 0x00, 0x10, 0x00,
+				 0x00, 0x01, 0x11, 0x00, 0x00);
+	zassert_true(bt_uuid_cmp(&uuid.uuid, &expected.uuid) == 0,
+		     "Parsed UUID does not match expected 128-bit UUID");
+}
+
+ZTEST(bt_uuid_from_str, test_uuid_roundtrip_to_str_and_back_16)
+{
+	struct bt_uuid_16 u1 = BT_UUID_INIT_16(0x180d);
+	struct bt_uuid_16 uuid_tmp = {0};
+	char str[BT_UUID_STR_LEN];
+	int ret;
+
+	/* uuid -> str */
+	bt_uuid_to_str(&u1.uuid, str, sizeof(str));
+
+	/* str -> uuid */
+	ret = bt_uuid_from_str(str, &uuid_tmp.uuid);
+	zassert_true(ret == 0, "bt_uuid_from_str failed for 16-bit");
+
+	/* compare */
+	zassert_true(bt_uuid_cmp(&u1.uuid, &uuid_tmp.uuid) == 0, "Round-trip 16-bit UUID mismatch");
+}
+
+ZTEST(bt_uuid_from_str, test_uuid_roundtrip_to_str_and_back_32)
+{
+	struct bt_uuid_32 u1 = BT_UUID_INIT_32(0xabcdef12);
+	struct bt_uuid_32 uuid_tmp = {0};
+	char str[BT_UUID_STR_LEN];
+	int ret;
+
+	/* uuid -> str */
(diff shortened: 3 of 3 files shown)


subsys/bluetooth/host/uuid.c
```diff
@@ -130,3 +131,45 @@ void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len)
...
+			return false;
+		}
+
+		uuid->type = BT_UUID_TYPE_32;
+		return true;
+	}
+	case BT_UUID_STR_LEN - 1: {
+		uint8_t *p = BT_UUID_128(uuid)->val;
+
+		if (sscanf(str,
```
carlescufi, line 166: Another option to avoid the heavy `sscanf` would be to use `strtol` and a loop

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,18 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *
+ *  Converts string to Bluetooth UUID.
+ *  UUID can be in any format, 16-bit, 32-bit or 128-bit.
+ *
+ *  @param str pointer to string to convert
+ *  @param uuid pointer where to put converted UUID
+ *
+ *  @return true if the string was valid and the UUID was successfully created.
+ */
+bool bt_str_to_uuid(const char *str, struct bt_uuid *uuid);
```
Thalley: I feel like this should use the `bt_uuid` prefix, but 

```suggestion
bool bt_uuid_str_to_uuid(const char *str, struct bt_uuid *uuid);
```

Also sounds kinda off

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,18 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
 
+/** @brief Convert string to Bluetooth UUID.
+ *
+ *  Converts string to Bluetooth UUID.
+ *  UUID can be in any format, 16-bit, 32-bit or 128-bit.
+ *
+ *  @param str pointer to string to convert
+ *  @param uuid pointer where to put converted UUID
+ *
+ *  @return true if the string was valid and the UUID was successfully created.
```
Thalley: I would consider changing the return type to `int` to be similar to other functions that may return an error value. It would also allow us to return different values depending on why it would fail

subsys/bluetooth/host/uuid.c
```diff
@@ -130,3 +131,45 @@ void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len)
 		return;
 	}
 }
+
+bool bt_str_to_uuid(const char *str, struct bt_uuid *uuid)
+{
+	switch (strlen(str)) {
```
Thalley, line 142: Should we have NULL checks for `str` and `uuid`?

subsys/bluetooth/host/uuid.c
Thalley: I actually kind of like the idea of using sscanf like here; implementing and checking formatting with a loop seems like it would need quite a lot of checks to verify the formatting

Thalley:
Please expand https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/bluetooth/uuid to test this new function :)

include/zephyr/bluetooth/uuid.h
alwa-nordic: I suggest using `from` and also reversing the parameters. That way, we both get `bt_uuid` as a namespace, and we get `bt_uuid` as a `self` (to borrow terminology from Python).

```suggestion
bool bt_uuid_from_str(struct bt_uuid *uuid, const char *str);
```

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,18 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
  */
 void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len);
 
+/** @brief Convert string to Bluetooth UUID.
+ *
+ *  Converts string to Bluetooth UUID.
+ *  UUID can be in any format, 16-bit, 32-bit or 128-bit.
+ *
+ *  @param str pointer to string to convert
+ *  @param uuid pointer where to put converted UUID
```
alwa-nordic: The downcasts (`BT_UUID_16`, `BT_UUID_32`, `BT_UUID_128`) are not safe in the implementation. The documentation here should at least mention this, e.g. specify that `uuid` parameter is reinterpreted to one of `bt_uuid_16`, `bt_uuid_32`, `bt_uuid_128`, matching the string format.

But I would instead prefer three type-safe variants of this function, one for each `bt_uuid` subtype.

We can optionally have a C11 _Generic-based macro that automatically selects between the functions.

This makes even more sense considering that a 16-bit BT UUID is representable as an 128-bit UUID.

alwa-nordic (changes requested):

include/zephyr/bluetooth/uuid.h
alwa-nordic: We can also create a type compatible with all the subtypes and make a parsing function that selects the variant tag corresponding to the string format, or to the smallest type that can represent the BT UUID. Maybe one function for each of those options.

chengkai15: Thanks for  your suggestion. Perhaps this is a good practice to keep the function and its implementation consistent with `bt_uuid_to_str`. Use a single function that is compatible with all types and prompt the user to note the length of the output type.

and  help note updated

subsys/bluetooth/host/uuid.c
chengkai15: done

subsys/bluetooth/host/uuid.c
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

chengkai15:
> Please expand https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/bluetooth/uuid to test this new function :)

I add it with https://github.com/zephyrproject-rtos/zephyr/pull/98784/commits/4ed0d20c004a23ba6dd891911a0d2567f2cdc95f

However, I'm not sure how to set up the test environment.

alwa-nordic:
> I'm not sure how to set up the test environment.

Try this:

```bash
/zephyrproject$ uv venv --clear && uv pip install -r zephyr/scripts/requirements.txt && uv run -- bash -c 'source zephyr/zephyr-env.sh && west twister -ciT zephyr/tests/bluetooth/uuid/'
```

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
  */
 void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len);
 
+/** @brief Convert string to Bluetooth UUID.
+ *
+ *  Converts string to Bluetooth UUID. The string must be in the standard
+ *  UUID format, which could be 16 bit, 32 bit or 128 bit format. And the
```
Thalley: Might be a bit tricky to say "standard UUID format", as there are several ways to represent a UUID as a string. Bluetooth seemingly uses the 8-4-4-4-12 format from https://www.rfc-editor.org/rfc/rfc9562.html, so we could possibly refer to that RFC for the string format, or simply omit this part as `bt_uuid_to_str` also does.

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *  uuid pointer must point to a bt_uuid large enough to hold the converted
+ *  UUID. If the string uuid is 16 bit, the converted bt_uuid will be
+ *  bt_uuid_16 type. If the string uuid is 32 bt uuid, the converted bt_uuid
```
Thalley:
```suggestion
 *  bt_uuid_16 type. If the string uuid is 32 bit uuid, the converted bt_uuid
```

or did you mean to use `BT`/Bluetooth?

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
  */
 void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len);
 
+/** @brief Convert string to Bluetooth UUID.
+ *
+ *  Converts string to Bluetooth UUID. The string must be in the standard
+ *  UUID format, which could be 16 bit, 32 bit or 128 bit format. And the
+ *  uuid pointer must point to a bt_uuid large enough to hold the converted
+ *  UUID. If the string uuid is 16 bit, the converted bt_uuid will be
```
Thalley:
```suggestion
 *  UUID. If the string UUID is 16 bit, the converted bt_uuid will be
```

All uses of UUID should be in caps

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *  bt_uuid_16 type. If the string uuid is 32 bt uuid, the converted bt_uuid
+ *  will be bt_uuid_32 type. If the string uuid is 128 bt uuid, the converted
+ *  bt_uuid will be bt_uuid_128 type.
+ *
+ *  @note The string uuid must be in the standard UUID format, which could be
+ *  16 bit, 32 bit or 128 bit format.
```
Thalley: This seems like duplicate text of previous text

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *  For example: 128 bit format "00001101-0000-1000-8000-00805F9B34FB"
+ *
+ *  @param uuid pointer where to put converted UUID
```
Thalley:
```suggestion
 *  @param uuid Pointer where to put converted UUID
```

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *  @param str pointer to string to convert
```
Thalley:
```suggestion
 *  @param str Pointer to string to convert
```

include/zephyr/bluetooth/uuid.h
```diff
@@ -5249,6 +5249,27 @@ bool bt_uuid_create(struct bt_uuid *uuid, const uint8_t *data, uint8_t data_len)
...
+ *
+ *  @return 0 if the string was converted successfully, otherwise negative error
+ */
+int bt_uuid_from_str(struct bt_uuid *uuid, const char *str);
```
Thalley: Consider switching the order, so it's `input, output`


```suggestion
int bt_uuid_from_str(const char *str, struct bt_uuid *uuid);
```

tests/bluetooth/uuid/src/test_bt_uuid_from_str.c
```diff
@@ -0,0 +1,64 @@
+/* Copyright (c) 2025 Xiaomi Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+#include <zephyr/ztest.h>
+#include <zephyr/bluetooth/uuid.h>
+
+ZTEST_SUITE(bt_uuid_from_str, NULL, NULL, NULL, NULL, NULL);
```
Thalley, line 10: Consider adding one or more tests where you compare the output from both `bt_uuid_from_str` and `bt_uuid_to_str`

Basically if you have a UUID and put it through `bt_uuid_to_str` and then through `bt_uuid_from_str` you should get the same UUID, and vice versa

Thalley (changes requested):

include/zephyr/bluetooth/uuid.h
alwa-nordic: But you gave a thumbs-up on this: https://github.com/zephyrproject-rtos/zephyr/pull/98784#discussion_r2490125378

The reasoning was to prefer a `self`-like order. It's OK if we change our minds, but maybe we should write something in the style guide.

Thalley: Ah, missed that it used to be like that. I don't have a strong opinion about this, and I guess it's also a widely debated topic (`src, dst` or `dst, src`)

include/zephyr/bluetooth/uuid.h
chengkai15: fix

include/zephyr/bluetooth/uuid.h
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

include/zephyr/bluetooth/uuid.h
chengkai15: updated

tests/bluetooth/uuid/src/test_bt_uuid_from_str.c
chengkai15: updated

include/zephyr/bluetooth/uuid.h
alwa-nordic: This thread is not resolved. Any input from @jhedberg or other collaborators? Otherwise I decide as maintainer that there will be three safe functions instead.

subsys/bluetooth/host/uuid.c
```diff
@@ -130,3 +132,49 @@ void bt_uuid_to_str(const struct bt_uuid *uuid, char *str, size_t len)
 		return;
 	}
 }
+
+int bt_uuid_from_str(const char *str, struct bt_uuid *uuid)
```
jhedberg, line 136: Seems a bit inconsistent with the `bt_uuid_create()` API return type. Also, there doesn't really seem to exist any other category of failure than "I can't interpret this as a UUID".

PR #98791, opened 2025-11-03
drivers: sensor: ti: Add INA232 native support

joe-mer:
The INA232 is another device in the INA2XX family and is very similar to the already implemented INA230 and INA236. The main difference between the INA232 and the INA236 is that the INA232 does not have a Device ID register. Because of these similarities, it is implemented in the ina230.x files in the same way as the INA236. Modifications to the corresponding tests are included as well.

### DIFF
--- a/drivers/sensor/ti/ina2xx/ina230.c
+++ b/drivers/sensor/ti/ina2xx/ina230.c
@@ -18,10 +18,12 @@ LOG_MODULE_REGISTER(INA230, CONFIG_SENSOR_LOG_LEVEL);
 
 /** @brief The LSB value for the bus voltage register, in microvolts/LSB. */
 #define INA230_BUS_VOLTAGE_UV_LSB 1250U
+#define INA232_BUS_VOLTAGE_UV_LSB 1600U
 #define INA236_BUS_VOLTAGE_UV_LSB 1600U
 
 /** @brief The scaling for the power register. */
 #define INA230_POWER_SCALING 25
+#define INA232_POWER_SCALING 32
 #define INA236_POWER_SCALING 32
 
 INA2XX_REG_DEFINE(ina230_config, INA230_REG_CONFIG, 16);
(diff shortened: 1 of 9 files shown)


joe-mer:
I'm not quite sure how to properly address the failing check. As far as I can tell, it's unrelated to my changes. The code snippet where the error occurs appears to be in sync with the current main branch.

MaureenHelm (dismissed):

drivers/sensor/ti/ina2xx/ina230.c
```diff
@@ -198,67 +212,96 @@ static DEVICE_API(sensor, ina230_driver_api) = {
...
+				.cal = INA230_DT_CAL(inst),                                        \
+				.id_reg = NULL,                                                    \
+				.config_reg = &ina230_config,                                      \
+				.adc_config_reg = NULL,                                            \
+				.cal_reg = &ina230_cal,                                            \
+				.channels = &ina236_channels,                                      \
+			},                                                                         \
+		.uv_lsb = INA236_BUS_VOLTAGE_UV_LSB,                                               \
+		.power_scale = INA236_POWER_SCALING,                                               \
+		COND_CODE_1(DT_INST_NODE_HAS_PROP(inst, alert_gpios),                  \
```
bogdanovs: Since you edited whole block, can you align end-of-line here and everywhere

joe-mer: I can do that, but it will lead to "Test ClangFormat warnings" for three lines. At least it did while running CI on my local machine. For the other three lines, which dont align with the rest, CI runs without warning when aligned.

I want to mention that i didnt change the eol alignment myself, I just ran clang format to get rid of the warnings and this is what it came up with.

bogdanovs (approved):

jpanisbl (approved):

MaureenHelm (approved):

Issue #98530, opened 2025-10-29
spi: Allow for explicit description of which CS the cs-gpios correspond to.


decsny:
### Summary

Right now, cs-gpios property implicitly describes CS lines starting from CS line 0 and ascending one at a time. The enhancement request is to have some way to explicitly say which CS line each GPIO is supposed to correspond to. This is to allow for the mixing of using HW CS or SW CS for different devices on the SPI bus.

### Describe the solution you'd like

Example of current property is like this:
`cs-gpios = <&gpio0 11 GPIO_ACTIVE_LOW>, <&gpio0 5 GPIO_ACTIVE_LOW>;`
This says, CS0 is on GPIO port 0 pin 11, and CS1 is on GPIO port 0 pin 5.

What could be done is something like this:
`cs-gpios = <&gpio0 11 GPIO_ACTIVE_LOW>, <&gpio0 5 GPIO_ACTIVE_LOW>;`
`cs-gpios-nums = <1 3>;`
to say, CS1 is on GPIO port 0 pin 11, and CS3 is on GPIO port 0 pin 5.
And then CS0 and CS2 could be controlled by the native HW CS.
Or any other combination.

### Additional Context

https://github.com/zephyrproject-rtos/zephyr/issues/92542 is the original motivation of this request. Due to the SOC only has one native HW CS to use. But user wants to have more slaves controlled on the same bus by the same controller while still getting the benefit of that one HW CS for one of the slave that has stricter timing requirements.

matt-wood-ct:
This is well worth doing, helps me a lot and also should help with some complex overlay stacking issues

PR #98795, opened 2025-11-03
devicetree: spi: fix DT_SPI_DEV_HAS_CS_GPIOS macro

### LINKED ISSUE #98830: DT_SPI_DEV_HAS_CS_GPIOS macro is not working as documented
### Describe the bug

The presence of a cs-gpios property caused every SPI device to be configured to use GPIO CS, no matter if there was a valid entry in cs-gpios for the given child device or not.

This macro is 6 years old but we only rely on it since  https://github.com/zephyrproject-rtos/zephyr/commit/a60f93d742a48660882af5785fb4352f414ecd11 , that's why we only see this bug now.

It is a regression for the SPI device API.

### Regression

- [x] This is a regression.

### Steps to reproduce

Run a spi_loopback test with an overlay that have only 1 entry for cs-gpios  and 2 SPI child devices like:
```
&eusart1 {
	#address-cells = <1>;
	#size-cells = <0>;
	pinctrl-0 = <&eusart1_default>;
	pinctrl-names = "default";
	status = "okay";
	cs-gpios = <&gpioc 0 GPIO_ACTIVE_LOW>;
	slow@0 {
		compatible = "test-spi-loopback-slow";
		reg = <0>;
		spi-max-frequency = <500000>;
	};
	fast@1 {
		compatible = "test-spi-loopback-fast";
		reg = <1>;
		spi-max-frequency = <10000000>;
	};
};
```

You will got a BUS FAULT at runtime with a null pointer dereference.

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

Martinhoff-maker:
Fixes DT_SPI_DEV_HAS_CS_GPIOS behavior. 

The presence of a cs-gpios property caused every SPI device to be configured to use GPIO CS, no matter if there was a valid entry in cs-gpios for the given child device or not. This fix ensures that the implementation matches the documented behavior, where GPIO CS is only used if the cs-gpios property has an entry matching the reg address of the child device.

This restores the behavior of the devicetree init API prior to https://github.com/zephyrproject-rtos/zephyr/commit/a60f93d742a48660882af5785fb4352f414ecd11 , which accidentally introduced this issue by relying on the `DT_SPI_DEV_HAS_CS_GPIOS` macro, which has deviated from its documented behavior ever since it was originally introduced 6 years ago.

This problem has been discovered because of a null pointer dereference in the case where the unit address of a child node exceeded the size of cs-gpios.

Fixes #98830

### DIFF
--- a/include/zephyr/devicetree/spi.h
+++ b/include/zephyr/devicetree/spi.h
@@ -115,7 +115,8 @@ extern "C" {
  * @return 1 if spi_dev's bus node DT_BUS(spi_dev) has a chip select
  *         pin at index DT_REG_ADDR(spi_dev), 0 otherwise
  */
-#define DT_SPI_DEV_HAS_CS_GPIOS(spi_dev) DT_SPI_HAS_CS_GPIOS(DT_BUS(spi_dev))
+#define DT_SPI_DEV_HAS_CS_GPIOS(spi_dev)                                                           \
+	DT_PROP_HAS_IDX(DT_BUS(spi_dev), cs_gpios, DT_REG_ADDR_RAW(spi_dev))
 
 /**
  * @brief Get a SPI device's chip select GPIO controller's node identifier


decsny:
relates to https://github.com/zephyrproject-rtos/zephyr/issues/98530 , mentioning for tracking purposes

decsny (approved):
looks right to me

jhedberg:
@Martinhoff-maker could you open a matching bug report issue for this, and add "Fixes ..." to this PR description? From what I can see this is a fix for a regression/inconsistency that happened between the documented behavior and the actual behavor, and as such could still be a candidate for the next Zephyr release. @cfriedt FYI

jhedberg (approved):

jerome-pouiller (approved):

PR #98670, opened 2025-10-31
runners: stlink_gdbserver: various fixes

mathieuchopstm:
* make it possible to disable SWD mode
  * Arguments with `action='store_true', default=True` can't be "unset": not providing the argument uses the default value of True, and providing it sets the value to True!
  * Use `argparse.BooleanOptionalAction` type instead which creates a `--swd`/`--no-swd` argument pair
* make `--external-init` a runner argument instead of always providing it to the ST-LINK GDB Server default
   * https://github.com/zephyrproject-rtos/zephyr/commit/3cc9a843e8154612f71d249f79458d69b3a1edd9 made it always provided, but this breaks debugging on STM32N6 series

The only thing that really matters in patchset is basically reverting https://github.com/zephyrproject-rtos/zephyr/commit/3cc9a843e8154612f71d249f79458d69b3a1edd9 because it breaks debugging on STM32N6 series; if changes seem too big for current RC state, I'm fine with keeping this PR for later and sticking to a mere revert.

PR #98443, opened 2025-10-28
scripts: runners: generalize dry run

carlescufi:
- The first commit should be non-controversial, it just moves the argparse call to the abstract ZephyrBinaryRunner class
- The second commit is there to avoid duplicating logic in individual runners, reusing the infrastructure in core.py. The usage of the global `_DRY_RUN` should remain unaffected, but subclasses can now set `self.dry_run` and then use the base classe methods (`_log_cmd()`, `call()`, `check_call()` and `check_output()`) directly without having to implement the logic themselves

PR #81070, opened 2024-11-07
[RFC] scripts: ci: check_compliance: Add python lint/format check

pdgendt:
## Introduction

As the codebase and contributions keep growing, so does all the python tooling.
This PR is intended to make python scripts more uniform and safe.

### Proposed change

Add [ruff](https://docs.astral.sh/ruff/) as a linter and formatter for python files, and verify (new) files in CI.

## Detailed RFC

Some hightlights:

- `ruff` is a drop-in replacement for many popular python tools, and it is super fast.
- The formatter output is [PEP8](https://peps.python.org/pep-0008/) compliant and produces nearly identical output to [black](https://black.readthedocs.io/en/stable/)
- Can fix issues for you
- ...

Current enabled linter rules:

- [x] [flake8-bugbear](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b)
- [x] [pycodestyle](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w)
- [x] [pyflakes ](https://docs.astral.sh/ruff/rules/#pyflakes-f)
- [x] [isort](https://docs.astral.sh/ruff/rules/#isort-i) (black profile)
- [x] [flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim)
- [x] [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up)

Many more can be added/enabled but this is a solid baseline.

### Concerns

Existing files do not comply, so they are explicitly excluded. Ideally these are gradually converted to comply.

```sh
$ ruff check
Found 3708 errors.

$ ruff format --check
379 files would be reformatted, 34 files already formatted
```

Formatting is obviously opinionated, but I do think for python we should follow the popular kids, which are `black` and `ruff`.

PR #98798, opened 2025-11-03
runners: promote --no-load argument into common code and add support in stlink_gdbrunner

mathieuchopstm:
Promote the existing `--no-load` argument from OpenOCD/Intel Cyclone V runners into common runner code. This argument inhibits loading of image on target before a debug session, allowing `west debug` to effectively be used as a `west attach`-after-reset.

Also add support of this argument to the `stlink_gdbserver` runner.

### DIFF
--- a/scripts/west_commands/runners/stlink_gdbserver.py
+++ b/scripts/west_commands/runners/stlink_gdbserver.py
@@ -90,7 +90,9 @@ class STLinkGDBServerRunner(ZephyrBinaryRunner):
 
     @classmethod
     def capabilities(cls) -> RunnerCaps:
-        return RunnerCaps(commands={"attach", "debug", "debugserver"}, dev_id=True, extload=True)
+        return RunnerCaps(
+            commands={"attach", "debug", "debugserver"}, dev_id=True, extload=True, skip_load=True
+        )
 
     @classmethod
     def extload_help(cls) -> str:
@@ -128,6 +130,7 @@ class STLinkGDBServerRunner(ZephyrBinaryRunner):
             args.port_number,
             args.extload,
             args.external_init,
+            args.load,
         )
 
     def __init__(
@@ -139,6 +142,7 @@ class STLinkGDBServerRunner(ZephyrBinaryRunner):
         gdb_port: int,
         external_loader: str | None,
         external_init: bool,
+        load: bool,
     ):
         super().__init__(cfg)
         self.ensure_output('elf')
@@ -149,6 +153,7 @@ class STLinkGDBServerRunner(ZephyrBinaryRunner):
         self._ap_id = ap_id
         self._external_loader = external_loader
         self._do_external_init = external_init
+        self._load = load
 
     def do_run(self, command: str, **kwargs):
         if command in ["attach", "debug", "debugserver"]:
@@ -181,7 +186,8 @@ class STLinkGDBServerRunner(ZephyrBinaryRunner):
             gdbserver_cmd += ["--attach"]
         else:  # debug/debugserver
             gdbserver_cmd += ["--initialize-reset"]
-            gdb_args += ["-ex", f"load {elf_path}"]
+            if self._load:
+                gdb_args += ["-ex", f"load {elf_path}"]
 
         if self._stlink_serial:
             gdbserver_cmd += ["--serial-number", self._stlink_serial]
--- a/scripts/west_commands/runners/core.py
+++ b/scripts/west_commands/runners/core.py
@@ -304,6 +304,10 @@ class RunnerCaps:
 
     - rtt: whether the runner supports SEGGER RTT. This adds a --rtt-address
       option.
+
+    - skip_load: whether the runner supports the --load/--no-load option, which
+      allows skipping the load of image on target before starting a debug session
+      (this option only affects the 'debug' command)
     '''
 
     commands: set[str] = field(default_factory=lambda: set(_RUNNERCAPS_COMMANDS))
@@ -319,6 +323,7 @@ class RunnerCaps:
     rtt: bool = False  # This capability exists separately from the rtt command
                        # to allow other commands to use the rtt address
     dry_run: bool = False
+    skip_load: bool = False
 
     def __post_init__(self):
         if self.mult_dev_ids and not self.dev_id:
@@ -639,6 +644,12 @@ class ZephyrBinaryRunner(abc.ABC):
         parser.add_argument('--dry-run', action='store_true',
                             help=('''Print all the commands without actually
                             executing them''' if caps.dry_run else argparse.SUPPRESS))
+
+        # by default, 'west debug' is expected to flash before starting the session
+        parser.add_argument('--load', action=argparse.BooleanOptionalAction,
+                            help=("load image on target before 'west debug' session"
+                                  if caps.skip_load else argparse.SUPPRESS),
+                            default=True)
 
         # Runner-specific options.
         cls.do_add_parser(parser)
--- a/scripts/west_commands/runners/intel_cyclonev.py
+++ b/scripts/west_commands/runners/intel_cyclonev.py
@@ -28,7 +28,7 @@ class IntelCycloneVBinaryRunner(ZephyrBinaryRunner):
                  tcl_port=DEFAULT_OPENOCD_TCL_PORT,
                  telnet_port=DEFAULT_OPENOCD_TELNET_PORT,
                  gdb_port=DEFAULT_OPENOCD_GDB_PORT,
-                 gdb_init=None, no_load=False):
+                 gdb_init=None, load=True):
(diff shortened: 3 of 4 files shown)


marc-hb (commented):
The formatting fixes belong to a separate commit

https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#multiple-commits-on-a-single-pr

> This flag functions exactly like its identically-named counterpart in the OpenOCD runner: when present, the "load" command is not added to the GDB client's command line.

... duplication which points at the lack of a common function or class. You probably not interested in changing and testing the OpenOCD runner, but you can still create that common function / class and start using it only in `stlink_gdbserver.py`. Then it's ready for both existing and future runners to use.


EDIT: as a pure coincidence, here is a current example of moving generic code to generic places, away from product-specific code:
- #98443

scripts/west_commands/runners/stlink_gdbserver.py
```diff
@@ -110,10 +110,23 @@ def do_add_parser(cls, parser: argparse.ArgumentParser):
             help="Port number for GDB client",
         )
 
+        # Additional arguments
+        parser.add_argument(
+            "--no-load",
+            action='store_true',
+            help="Don't flash binary before debug (no GDB 'load' command)",
+        )
```
pdgendt: Turn this into a `--load/--no-load` option, consistent with `--rebuild/--no-rebuild` and others. Similar to `--swd` in https://github.com/zephyrproject-rtos/zephyr/pull/98670

```python
        # Additional arguments
        parser.add_argument(
            "--load",
            default=True,
            action=argparse.BooleanOptionalAction,
            help="Optionally flash binary before debug (GDB 'load' command)",
        )
```

This is to make the arguments more consistent and let the last passed argument win.

I also agree with @marc-hb, it would be nice to turn this into a runner capability with some common handling.

scripts/west_commands/runners/stlink_gdbserver.py
```diff
@@ -110,10 +110,23 @@ def do_add_parser(cls, parser: argparse.ArgumentParser):
...
     def do_create(cls, cfg: RunnerConfig, args: argparse.Namespace) -> "STLinkGDBServerRunner":
         return STLinkGDBServerRunner(
-            cfg, args.swd, args.apid, args.dev_id, args.port_number, args.extload
+            cfg,
+            args.swd,
+            args.apid,
+            args.dev_id,
+            args.port_number,
+            args.extload,
+            args.no_load,
```
mathieuchopstm: > The formatting fixes belong to a separate commit
> 
> https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#multiple-commits-on-a-single-pr

Are you referring to this? This isn't merely a "formatting" fix - `ruff` will yell at me because line is too long (arguably, I didn't check if it *does* yell at me here, but I know it will because another patch will be merged in the runner beforehand)

scripts/west_commands/runners/stlink_gdbserver.py
mathieuchopstm: Sounds reasonable. I'll have a look at how to make this global.

scripts/west_commands/runners/stlink_gdbserver.py
marc-hb: Yes, `ruff` is the formatting tool. It does not change any behavior.

The Python codebase is being "ruffified" but gradually to avoid massive git churn, see #81070 , https://github.com/zephyrproject-rtos/west/pull/865 , etc.

You can run `ruff` on this file and submit the change now; before you do any real work. This will get it out of the way.

https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits

mathieuchopstm: The problem is that this is *currently* accepted by `ruff`, but the commit adds a new argument which makes it go over the length limit and then `ruff` starts complaining. Running `ruff` beforehand will not change anything - the file is already compliant.

mathieuchopstm: Solved by rebase

mathieuchopstm:
Moved to common runner code as suggested by @marc-hb and @pdgendt (inspired by #98443).

Help message uses a generic wording (instead of "don't send GDB `load` command") as other runners may perform load-image-on-target before starting their GDB server - `--no-load` would apply for these too, if they support a way to inhibit the mechanism.

pdgendt (dismissed):
Nice, thanks!

scripts/west_commands/runners/stlink_gdbserver.py
marc-hb: ruff still not happy in https://github.com/zephyrproject-rtos/zephyr/actions/runs/19101453292/job/54574222071?pr=98798

pdgendt (dismissed):

mathieuchopstm:
Rebased to resolve conflict

pdgendt (dismissed):

marc-hb (commented):
I'm confused by this new `debug_load` capability: what happens when a future runner does _not_ to have the capability? Does that runner `-ex load` or not? Does it depend on the specific runner? Inconsistently...

- Developers could assume "loading" is a capability. No capability = no load command issued.
- But they could assume instead that _skipping_ the load is the capability. So, loading when no skip capability (similar to right now before this PR)

Generally speaking, having "two levels" of booleans adds complexity. So you could just drop the capability and force all runners to implement the option. It requires only one line of code in the runner. If some future runners can't support one of the two choices (cause they don't use `gdb`?), they could just fail when someone tries to unsupported choice.

Except... there isn't really a way to force runners to look at `args.load`, is there?

This is surprisingly more complicated that I thought.

mathieuchopstm:
> I'm confused by this new `debug_load` capability: what happens when a future runner does _not_ to have the capability? Does that runner `-ex load` or not? Does it depend on the specific runner? Inconsistently...

I do agree the capability name is not great. Maybe something like `noload` / `noload_debug` / `debug_without_load` would be more understandable?

> * Developers could assume "loading" is a capability. No capability = no load command issued.

Per "specification", `west debug` is *expected* to flash before starting the session:

>```
>$ west debug --help
>usage: west debug [-h] [-d DIR] [-r RUNNER] [--domain DOMAIN] [--skip-rebuild | --rebuild | --no-rebuild] [-H] [--board-dir DIR] [--gdb GDB] [--openocd OPENOCD]
>                   [--openocd-search DIR]
>
>Connect to the board, flash the program, and start a debugging session. Use "west attach" instead to attach a debugger without reflashing.
>```

Emphasis on ***`flash the program`***, hence the comment in my code:
> `# by default, 'west debug' is expected to flash before starting the session`

(to clarify again, the reason why `west flash --no-load` would be useful is that `west attach` merely starts a debugger, connects to and halts the target, whereas `west debug` also resets the target, breaks on entrypoint, ...)

> Generally speaking, having "two levels" of booleans adds complexity. So you could just drop the capability and force all runners to implement the option. It requires only one line of code in the runner. If some future runners can't support one of the two choices (cause they don't use `gdb`?), they could just fail when someone tries to unsupported choice.

What I had in mind was instead tools which would (*forcefully!*) download an image before starting the GDB server, or tools which don't use GDB at all. Looking at tree for runners which support `debug`, most do currently merely append `-ex load` to GDB commandline, but there are some exceptions:

- `Synopsys Metaware Debugger(mdb)`: seems to be its own thing entirely (i.e., not GDB)
https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/scripts/west_commands/runners/mdb.py#L75-L82
- `native`: executes native binary - `load`/`no-load` wouldn't apply
- `probe-rs`: uses an opaque `gdb` argument for which I can't find documentation :man_shrugging:
https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/scripts/west_commands/runners/probe_rs.py#L167-L187
- `spi_burn`: doesn't seem to download image when `west debug`'ing?
- `trace32`: seems to be its own thing?
https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/scripts/west_commands/runners/trace32.py#L131-L142

> Except... there isn't really a way to force runners to look at `args.load`, is there?

Yes, hence why it should be a capability if we follow what's already in tree.

marc-hb:
Or, you could call it `debug_skip_load` if you hate negative capabilities even more than I do.

```python
parser.add_argument('--load', action=argparse.BooleanOptionalAction,
        help=("load image on target before 'west debug' session"
       if caps.debug_skip_load else argparse.SUPPRESS),
            default=True)

pdgendt (approved):

marc-hb (approved):

PR #98800, opened 2025-11-03
Nordic: SoC: Add Kconfig options for soc init code

bjarki-andreasen:
Add Kconfig options which configure soc init code. These Kconfig options default to values found in the devicetree, while allowing users to overwrite them using Kconfig in cases where users prefer that over devicetree.

### DIFF
--- a/soc/nordic/common/Kconfig
+++ b/soc/nordic/common/Kconfig
@@ -76,3 +76,4 @@ endif # HAS_NORDIC_DMM
 
 rsource "vpr/Kconfig"
 rsource "uicr/Kconfig"
+rsource "Kconfig.clock_config"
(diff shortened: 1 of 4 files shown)


gmarull (changes requested):
what's the reason to add Kconfig "mirror" settings?

bjarki-andreasen:
> what's the reason to add Kconfig "mirror" settings?

Its a way to reuse the soc init code for applications which refuse to use the devicetree. I think we found better way of doing it, closing PR for now.

tejlmand:
Placing a comment just for reference in case someone later proposes similar idea.

I don't like mirroring dt in Kconfig, because i'm concerned it will turn into a mess in source code.
Some parts of source code using devicetree, other parts using the Kconfig mirror.

At some point some developers starts to overwrite the kconfig option because it makes their code work, but this overwrite is not reflected in the devicetree, and everything starts to bitrot.

PR #98804, opened 2025-11-03
boards: other: Default to RC source for ProMicro nRF52840

petejohanson:
The various ProMicro nRF52840 boards are frequently found to have BLE issues due to their external crystals not functioning well, so default to the internal RC source.

These boards end up getting used extensively with ZMK, and we *regularly* encounter support issues due to the crystals on them being broken/poorly tuned. So much so have a troubleshooting section specifically about this: https://zmk.dev/docs/troubleshooting/connection-issues#mitigating-a-faulty-oscillator

Recommend this is a saner default for these cheaper boards.

### DIFF
--- a/boards/others/promicro_nrf52840/promicro_nrf52840_nrf52840_defconfig
+++ b/boards/others/promicro_nrf52840/promicro_nrf52840_nrf52840_defconfig
@@ -9,6 +9,10 @@ CONFIG_HW_STACK_PROTECTION=y
 # Enable GPIO
 CONFIG_GPIO=y
 
+# Even though these boards have an external crystal, they are notoriously
+# faulty/poorly tuned, so use the internal RC by default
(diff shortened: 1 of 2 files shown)


carlescufi (approved):

cvinayak (approved):

nordicjm (approved):
Did you try changing the Kconfig for error to 500ppm? The issue could just be that they crystals have a far worse error rate but are still usable at 32KHz and will work for bluetooth with the correct ppm error Kconfig selected

petejohanson:
> Did you try changing the Kconfig for error to 500ppm? The issue could just be that they crystals have a far worse error rate but are still usable at 32KHz and will work for bluetooth with the correct ppm error Kconfig selected

We've not tried that as an approach, at least on the ZMK side. I'm happy to start suggesting that on our end for a while; I don't personally have any of those boards on hand to personally test.

PR #98807, opened 2025-11-03
cmake: compiler/xt-clang: do not warn about unused functions

dcpleung:
xt-clang sometimes complains about functions being unused though it is clearly being used. And GCC in Zephyr SDK does not complain about them. So skip the unused function warnings.

### DIFF
--- a/cmake/compiler/xt-clang/compiler_flags.cmake
+++ b/cmake/compiler/xt-clang/compiler_flags.cmake
@@ -28,3 +28,11 @@ set_compiler_property(PROPERTY warning_shadow_variables)
 check_set_compiler_property(APPEND PROPERTY warning_extended
                             -Wno-unknown-warning-option
 )
+
+# xt-clang sometimes complains about functions being unused
+# though it is clearly being used. And GCC in Zephyr SDK
+# does not complain about them. So skip the unused function
+# warnings.
+check_set_compiler_property(APPEND PROPERTY warning_extended
+                            -Wno-unused-function
+)


cmake/compiler/xt-clang/compiler_flags.cmake
```diff
@@ -28,3 +28,11 @@ set_compiler_property(PROPERTY warning_shadow_variables)
 check_set_compiler_property(APPEND PROPERTY warning_extended
                             -Wno-unknown-warning-option
 )
+
+# xt-clang sometimes complains about functions being unused
+# though it is clearly being used. And GCC in Zephyr SDK
```
tejlmand, line 33: to avoid this property being changed back anf forth, could you create an issue with related functions so that we can link to it.

Will also make it easier to understand scenarios where this might happen.
As example, perhaps if it occurs when using a specific pattern, then we could change code or only apply the flag on libraries where this false-positive warning is seen.

dcpleung: I haven't been able to figure out the pattern. Sometimes adding/removing code unrelated to the warnings would make them go away. I guess it might be some internal state tracking that has gone wrong.

Since we are using Zephyr SDK in CI, that should have enough coverage on the code in tree.

tejlmand: > Since we are using Zephyr SDK in CI, that should have enough coverage on the code in tree.

sure, for in-tree usage.

But what about downstream users of Zephyr using `xt-clang` ?
Those users are expected to have code out-of-tree of Zephyr, and if they are using xt-clang, then I wouldn't expect them to also build their code using Zephyr SDK.
So I assume those users would still like to have this warning enabled.

Therefore I would like to understand what is happening here.

Could this be related to `static inline` functions defined in header files ?
Is this happening on compilation or when using (xt-)clang as linter ?

> Sometimes adding/removing code unrelated to the warnings would make them go away.

Like adding / removing in the file reporting the error, or as in adding / removing code by creating a new file ?

PR #98808, opened 2025-11-03
tests: intel_adsp/smoke: cast to signed ints before abs()

dcpleung:
xt-clang complains about the clock ratio difference is unsigned when being fed to abs(), though GCC does not. So type cast that to signed integer before feeding to abs(). Clock ratio could be big enough to overflow signed 32-bit integer. So we first cast it to be a signed 64-bit integer before substraction.


tmleman (approved):

softwarecki (approved):

dcpleung:
Should be okay. It is not a blocker at this point. I can wait a bit.

nashif (approved):

PR #98811, opened 2025-11-03
llext: Fix ARMv8-M MPU memory region alignment

iabdalkader:
ARMv8-M and newer ARC MPUs use base/limit addressing with 32-byte granularity and do not require power-of-two sizing. Only ARMv7-M and older ARC MPUs have this restriction.

Check `CONFIG_MPU_REQUIRES_POWER_OF_TWO_ALIGNMENT` (set for ARMv7-M, older ARC) instead of `CONFIG_ARM_MPU` to avoid wasting memory on ARMv8-M systems.

This is something I noticed while looking at the llext code. If you don’t want to modify that critical code, feel free to close it.

### DIFF
--- a/subsys/llext/llext_mem.c
+++ b/subsys/llext/llext_mem.c
@@ -96,15 +96,19 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
 	 * program-accessible data (not to string tables, for example).
 	 */
 	if (region->sh_flags & SHF_ALLOC) {
-		if (IS_ENABLED(CONFIG_ARM_MPU) || IS_ENABLED(CONFIG_ARC_MPU)) {
-			/* On ARM with an MPU, regions must be sized and
-			 * aligned to the same power of two (larger than 32).
+		if (IS_ENABLED(CONFIG_MPU_REQUIRES_POWER_OF_TWO_ALIGNMENT)) {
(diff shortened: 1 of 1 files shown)


subsys/llext/llext_mem.c
pillo79, line 118: Nit: Sonar reported (correctly) that these two cases have equal handling. Please merge them into one.

pillo79 (commented):
Thanks, really didn't know such a flag existed. Huge improvement over the current situation 🥳 
 
LGTM, but I have no ARC experience. @laurenmurphyx64, any feedback? Thanks!

subsys/llext/llext_mem.c
iabdalkader: Yeah, that's intentional to be super explicit about the handling of each. Still merge them?

laurenmurphyx64 (approved):

teburd (approved):

subsys/llext/llext_mem.c
pillo79: IMHO they have a similar paged approach, so they should share the code. However PR has been already approved, so no need to! :slightly_smiling_face:

jhedberg:
Leaving this to 4.4 is perfectly fine with me. It was mainly the title that threw me off a bit ("Fix ...")

PR #97504, opened 2025-10-14
driver: flash_mspi_nor: Allow specific read/write IO modes and frequencies

utsavm9:
Consider these MSPI flash parts: [Gigadevice GD25LE64E](https://www.mouser.com/datasheet/2/870/gd25le64e_rev1_3_20220426-2581248.pdf?srsltid=AfmBOorNVcnYnluZdrfwc0T8omPY6eENgh54MhcyPPMGnzULMow4bohG), [Macronix MX25U6432F](https://www.mxic.com.tw/Lists/Datasheet/Attachments/9025/MX25U6432F,%201.8V,%2064Mb,%20v1.1.pdf), [Winbond W25Q64JW](https://www.mouser.com/datasheet/2/949/W25Q64JW_RevB_11042019-1760399.pdf?srsltid=AfmBOoqU8en17EXwRIWMS5knHo8qgqum1mYZ11-1KLpk4Y5CnxYXuwyQ). If only D0 and D1 lines are connected to these flash parts, then Dual IO 1-1-2 `DREAD 0x3B` and Single IO `PP 0x02` are the best read and write commands available to use. Notice that neither Dual IO `PP_1_1_2 0xA2` command is supported in any of these flash parts, and neither these flash parts support Dual Peripheral Interface (like QPI, OPI), so all general commands need to be in Single IO.

The driver currently provides no way to use Dual IO Read and Single IO for the rest of the commands currently, and would erroneously use Single IO PP command in Dual IO mode. This PR fixes and adds support for that.

## Target Board
nRF54H20 MSPI and flash driver to support [Gigadevice GD25LE64E](https://www.mouser.com/datasheet/2/870/gd25le64e_rev1_3_20220426-2581248.pdf?srsltid=AfmBOorNVcnYnluZdrfwc0T8omPY6eENgh54MhcyPPMGnzULMow4bohG), [Macronix MX25U6432F](https://www.mxic.com.tw/Lists/Datasheet/Attachments/9025/MX25U6432F,%201.8V,%2064Mb,%20v1.1.pdf) and [Winbond W25Q64JW](https://www.mouser.com/datasheet/2/949/W25Q64JW_RevB_11042019-1760399.pdf?srsltid=AfmBOoqU8en17EXwRIWMS5knHo8qgqum1mYZ11-1KLpk4Y5CnxYXuwyQ).

## Modifications
[...]

PR #98814, opened 2025-11-03
drivers: flash: flash_mspi_nor: support custom read frequency

danieldegrasse:
Many flash chips support a higher read frequency than other operations. Support specifying a custom read frequency to take advantage of the performance improvements available with this approach.


swift-tk:
It is already done here https://github.com/zephyrproject-rtos/zephyr/pull/97504.

danieldegrasse:
> It is already done here https://github.com/zephyrproject-rtos/zephyr/pull/97504.

Closing in favor of that PR, thanks

PR #98816, opened 2025-11-04
west: blobs: fix UnicodeDecodeError on Windows

### LINKED ISSUE #98520: hal_nxp has invalid key "click-through" in module.yml
### Describe the bug

hal_nxp/zephyr/module.yml contains an invalid key "click-through" causing the below issue when running `west build`. Not entirely sure when this was introduced but saw this on version included in west.yml from 4.2.0 and 4.3.0-rc1.

```
CMake Error at zephyr/zephyr/cmake/modules/zephyr_module.cmake:73 (message):
  validation.invalid

   --- All found errors ---

  ["Key 'click-through' was not defined.  Path: '/blobs/22'"]

  ERROR: Malformed "build" section in file:
/<some dir>/project/zephyr/modules/hal/nxp/zephyr/module.yml

  <SchemaError: error code 2: Schema validation failed:

   - Key 'click-through' was not defined. Path: '/blobs/22'.: Path: '/'>

Call Stack (most recent call first):
  zephyr/zephyr/cmake/modules/zephyr_default.cmake:133 (include)
  zephyr/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
  zephyr/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  CMakeLists.txt:5 (find_package)
```

### Regression

- [x] This is a regression.

### Steps to reproduce

build any project for any nxp board.

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Linux but would happen on any OS.
- zephyr-sdk-0.16.3
- 

### Additional Context

_No response_

dleach02:
Ensure the blobs command is trying to display the license file for click-through approval. Ensure that special characters doesn't cause a problem in Windows by opening the file with utf-8 encoding.

Fixes #98520

### DIFF
--- a/scripts/west_commands/blobs.py
+++ b/scripts/west_commands/blobs.py
@@ -186,7 +186,7 @@ class Blobs(WestCommand):
                     self.wrn('Skip fetching this blob.')
                     continue
 
-                with open(blob['license-abspath']) as license_file:
+                with open(blob['license-abspath'], encoding="utf-8") as license_file:
                     license_content = license_file.read()
                     print(license_content)
 


marc-hb:
Wow, UTF8 being hardcoded in two more places on the same day:
- https://github.com/zephyrproject-rtos/zephyr/pull/98751#pullrequestreview-3410319251

Plus many, many more.

It's hardcoded in fewer places but in https://github.com/zephyrproject-rtos/west too (I plead guilty)

Isn't there some Python place to hardcode it once for all rather than piecemeal like this? This will never end.

pdgendt (approved):

pdgendt:
> Isn't there some Python place to hardcode it once for all rather than piecemeal like this? This will never end.

Sadly, I don't think that's an option, from https://docs.python.org/3/library/functions.html#open

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used. See the [codecs](https://docs.python.org/3/library/codecs.html#module-codecs) module for the list of supported encodings.

But [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) is unaffected by [Python UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode). It will be the default from Python 3.15 onward, we mainly need a lot of patience.

carlescufi (approved):

PR #98817, opened 2025-11-04
tests: kernel: smp_abort: add delay to ensure circular dependency

### LINKED ISSUE #98819: tests: kernel: smp_abort timing race on ARMv8-R FVP
## Describe the bug

The `tests/kernel/smp_abort` test fails on ARMv8-R FVP 
(`fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp`) with threads continuing
execution after they should have been aborted. The test expects all
threads to be aborted through a circular dependency mechanism, but
one or more threads escape abortion and trigger assertion failures.

**Target platform:** ARM FVP Base RevC AEMv8-R (fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp)

**Root cause:** The test has a timing assumption that doesn't hold on 
platforms with large emulator/simulator quantum sizes. The test creates a 
circular abort scenario where each thread (in its ISR) aborts the next 
thread in sequence. However, if the quantum is large enough, one thread's
ISR may complete and return before the "previous" thread in the chain 
even calls `k_thread_abort()` on it, preventing the circular dependency 
from forming.

**What was tried:**
- Initially suspected kernel scheduler bugs
- Investigated memory barrier issues (ARMv8-R weak memory ordering)
- Examined shared `_thread_dummy` structure race conditions
- Analyzed `__attribute_const__` optimization on `z_smp_current_get()`
- All led to discovering this was a test timing issue, not a kernel bug

## Regression

- [ ] This is a regression.

No - test has always had this latent timing assumption

## Steps to reproduce

```bash
west build -p -b fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp tests/kernel/smp_abort/
west build -t run
```

Relevant log output

```
START - test_smp_thread_abort_deadlock
Thread 0 started
Thread 1 started
Thread 2 started
Thread 3 started

    Assertion failed at WEST_TOPDIR/zephyr/tests/kernel/smp_abort/src/main.c:53: thread_entry: (false is false)
Thread 0 did not abort!
```

Impact

Functional Limitation – Test fails on ARMv8-R FVP, but the kernel scheduler 
code is correct. The test works on QEMU and ARMv9-A FVP due to different
timing characteristics.

Environment

- OS: Linux
- Toolchain: Zephyr SDK
- Board: fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp
- Commit: a3aa51393b65

Additional Context

PR #98682 must be applied to get rid of the following:

```
ASSERTION FAIL [old_thread->switch_handle == ((void *)0)] @ WEST_TOPDIR/zephyr/kernel/sched.c:865
        old thread handle should be null.
```

This is however a separate issue unrelated to this one.

Fix: PR #98817 fixes this issue.

npitre:
The test_smp_thread_abort_deadlock test assumes all CPUs will
simultaneously establish a circular abort dependency (each thread
aborts the next in a ring). However, on platforms with large
emulator/simulator quantum sizes, one thread's ISR may complete
and return before another thread even calls `k_thread_abort()` on it,
preventing the circular dependency from forming.

This manifested on ARMv8-R FVP but not on QEMU or ARMv8-A/ARMv9-A FVP
due to different timing characteristics and quantum sizes.

Add a `k_busy_wait(100)` after `k_thread_abort()` to give all CPUs time
to call abort before any ISR completes.

Fixes #98819

### DIFF
--- a/tests/kernel/smp_abort/src/main.c
+++ b/tests/kernel/smp_abort/src/main.c
@@ -39,6 +39,15 @@ static void isr(const void *args)
 	}
 
 	k_thread_abort(var->target);    /* Abort thread on another CPU */
+
+	/*
+	 * Give other CPUs time to also call k_thread_abort() to establish
+	 * the circular dependency that this test is designed to exercise.
+	 * On platforms with large emulator/simulator quantum sizes, without
+	 * this delay one thread may complete its ISR and return before
+	 * another thread even calls k_thread_abort() on it.
+	 */
+	k_busy_wait(100);
 }
 
 static void thread_entry(void *p1, void *p2, void *p3)


wearyzen (approved):

peter-mitsis (approved):
If I follow the sequence of events correctly for where this delay is needed ...

1. T1, T2 and T3 are all in their ISRs. 
2. ISR T1 tries to abort T2, but must spin.
3. ISR T2 tries to abort T3, but first self-terminates T2 and then spins waiting for T3.
4. ISR T1 wins the sched lock, and is able to leave the ISR.
5. As ISR T1 the ISR it wins the sched lock again and goes back to being T1.
6. T1 saw that it had not yet been aborted.

PR #98838, opened 2025-11-04
Bluetooth: Classic: HFP: Terminate outgoing call without alerting

lylezhu2012:
If the outgoing call is not alerted, support the case to terminate the call when receiving the AT command `AT+CHUP`.


subsys/bluetooth/host/classic/hfp_ag.c
```diff
@@ -2358,6 +2358,8 @@ static int bt_hfp_ag_chup_handler(struct bt_hfp_ag *ag, struct net_buf *buf)
 			}
 		} else if (call_state == BT_HFP_CALL_ACTIVE) {
 			next_step = bt_hfp_ag_unit_call_terminate;
+		} else if (call_state == BT_HFP_CALL_OUTGOING) {
+			next_step = bt_hfp_ag_call_terminate;
 		}
```
MarkWangChinese: Does the call that is in state `BT_HFP_CALL_INCOMING` need to be terminated too?

MarkWangChinese (approved):

makeshi (approved):

PR #98849, opened 2025-11-04
drivers: usb: udc: mcux: force ehci speep as FS

MarkWangChinese:
When CONFIG_UDC_DRIVER_HIGH_SPEED_SUPPORT_ENABLED is disabled, force configure ehci controller to work as FS.

### DIFF
--- a/drivers/usb/udc/udc_mcux_ehci.c
+++ b/drivers/usb/udc/udc_mcux_ehci.c
@@ -708,6 +708,12 @@ static int udc_mcux_init(const struct device *dev)
 		return -ENOMEM;
 	}
 
+	if (!IS_ENABLED(CONFIG_UDC_DRIVER_HIGH_SPEED_SUPPORT_ENABLED)) {
+		USBHS_Type *usbBase = (USBHS_Type *)config->base;
+
+		usbBase->PORTSC1 |= USBHS_PORTSC1_PFSC_MASK;
+	}
+
 	/* enable USB interrupt */
 	config->irq_enable_func(dev);
 


MarkWangChinese:
Hi @tmon-nordic When forcing the HS controller to work as FS. User needs to set both `CONFIG_UDC_DRIVER_HIGH_SPEED_SUPPORT_ENABLED=n` and `CONFIG_USBD_MAX_SPEED_FULL=y`, right? Thanks in advance.

drivers/usb/udc/udc_mcux_ehci.c
```diff
@@ -708,6 +708,12 @@ static int udc_mcux_init(const struct device *dev)
 		return -ENOMEM;
 	}
 
+#ifndef CONFIG_UDC_DRIVER_HIGH_SPEED_SUPPORT_ENABLED
+	USBHS_Type *usbBase = (USBHS_Type *)config->base;
+
+	usbBase->PORTSC1 |= USBHS_PORTSC1_PFSC_MASK;
+#endif
+
```
tmon-nordic: It is preferred to use `if (!IS_ENABLED(CONFIG_UDC_DRIVER_HIGH_SPEED_SUPPORT_ENABLED))` which is optimized away at compile time if the path is not taken. The reason is to have code that always has proper syntax (with ifdef it is relatively easy to omit the path during refactor/rename).

MarkWangChinese: Done, thanks for the explanation.

tmon-nordic (approved):

dleach02 (approved):

PR #98411, opened 2025-10-28
Bluetooth: Host: Fix handling of adv reports when scanning for connection

zhongzhijie1:
🧩 Change Background

When calling:

```
int bt_conn_le_create(const bt_addr_le_t *peer,
		      const struct bt_conn_le_create_param *create_param,
		      const struct bt_le_conn_param *conn_param,
		      struct bt_conn **ret_conn)
```

Normally, a BLE connection will go through:

```
bt_conn_set_state(conn, BT_CONN_INITIATING);

err = bt_le_create_conn(conn);
if (err) {
	conn->err = 0;
	bt_conn_set_state(conn, BT_CONN_DISCONNECTED);
	bt_conn_unref(conn);

	/* Best-effort attempt to inform the scanner that the initiator stopped. */
	int scan_check_err = bt_le_scan_user_add(BT_LE_SCAN_USER_NONE);

	if (scan_check_err) {
		LOG_WRN("Error while updating the scanner (%d)", scan_check_err);
	}
	return err;
}
```

⚠️ Problem Case

After pairing and distributing IRK,
if the controller resolving list size = 0,
the next connection will follow the “host-based resolving” path:

```
if (bt_dev.le.rl_entries > bt_dev.le.rl_size) {
	/* Use host-based identity resolving. */
	bt_conn_set_state(conn, BT_CONN_SCAN_BEFORE_INITIATING);

	err = bt_le_scan_user_add(BT_LE_SCAN_USER_CONN);
	if (err) {
		bt_le_scan_user_remove(BT_LE_SCAN_USER_CONN);
		bt_conn_set_state(conn, BT_CONN_DISCONNECTED);
		bt_conn_unref(conn);

		return err;
	}

	*ret_conn = conn;
	return 0;
}
```

This means the Host will start a scan automatically and mark it with flag:

`BT_LE_SCAN_USER_CONN`

However, in le_adv_recv(), if the scan is not started by user (not BT_LE_SCAN_USER_EXPLICIT_SCAN),
the advertising report is ignored:

```
if (!atomic_test_bit(scan_state.scan_flags, BT_LE_SCAN_USER_EXPLICIT_SCAN)) {
	/* skip this adv report */
	return;
}

```

As a result, the connection logic below can never be triggered:

```
conn = bt_conn_lookup_state_le(BT_ID_DEFAULT, id_addr,
			       BT_CONN_SCAN_BEFORE_INITIATING);
if (!conn) {
	return;
}

[...]

PR #98859, opened 2025-11-04
Bluetooth: Host: Fix build error with observer-only config

jhedberg:
When building an observer-only build the check_pending_conn label would result in CI warnings/errors due to this only being a C23 feature:

scan.c:692:1: error: label at end of compound statement is a C23 extension

Turns out the #ifdefs are completely unnecessary, and the code can simply take advantage of IS_ENABLED(), which should get rid of the warning.

### DIFF
--- a/subsys/bluetooth/host/scan.c
+++ b/subsys/bluetooth/host/scan.c
@@ -538,7 +538,6 @@ int bt_le_scan_user_remove(enum bt_le_scan_user flag)
 	return scan_update();
 }
 
-#if defined(CONFIG_BT_CENTRAL)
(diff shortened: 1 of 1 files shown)


jhedberg:
This should fix a CI build error introduced by #98411: https://github.com/zephyrproject-rtos/zephyr/actions/runs/19072848093/job/54480145733

cfriedt (approved):

Thalley (approved):

PavelVPV (approved):

aescolar (approved):
Just for info: Note adding a `;` after the label would have also done it.

PR #97913, opened 2025-10-20
Bluetooth: Host: Run tx_processor on its own thread

alwa-nordic:
Move tx_processor off the system work queue to its own thread. Solves deadlock where `bt_hci_cmd_alloc()` blocks the system work queue.

PR #98861, opened 2025-11-04
bluetooth: host: Make le conn update as async

VinitMehta7492:
**Problem Statement**:

As per current implementation, bt_conn_le_conn_update API uses async transport [API](https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/subsys/bluetooth/host/conn.c#L4022).   

The implementation of sync API uses system work queue context and doesn't actually respect the HCI_CMD_TIMEOUT wait time at this [condition](https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/subsys/bluetooth/host/hci_core.c#L503) and returns immediately from semaphore causing false assert even though controller responds correctly to the HCI message with success status and before the semaphore is given.

Due to above behavior, application such as peripheral_gatt_write and cetral_gatt_write when tested against each other causes incorrect behavior when changing conn update parameter resulting in assert.

In order to overcome such scenarios particularly in LE conn update params which would require the controller or the LL to complete the conn update procedures before sending event made changes to use async API rather than sync.

### DIFF
--- a/subsys/bluetooth/host/conn.c
+++ b/subsys/bluetooth/host/conn.c
@@ -4019,7 +4019,7 @@ int bt_conn_le_conn_update(struct bt_conn *conn,
 	conn_update->conn_latency = sys_cpu_to_le16(param->latency);
 	conn_update->supervision_timeout = sys_cpu_to_le16(param->timeout);
 
-	return bt_hci_cmd_send_sync(BT_HCI_OP_LE_CONN_UPDATE, buf, NULL);
+	return bt_hci_cmd_send(BT_HCI_OP_LE_CONN_UPDATE, buf);
 }
 
 #if defined(CONFIG_BT_SMP) || defined(CONFIG_BT_CLASSIC)


alwa-nordic (dismissed):
I'm not convinced by the text in the PR description. In particular, I don't see how [the linked k_sem_take call](https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/subsys/bluetooth/host/hci_core.c#L503) can skip its timeout.

But this change is sensible because of the following call graph, which is a deadlock hazard.

```
bt_hci_cmd_send_sync
bt_conn_le_conn_update
le_conn_param_update_req
l2cap_recv
...
[BT RX Thread]
```

`le_conn_param_update_req` ignores errors from `bt_conn_le_conn_update`, so it also does not care about the return value from the Controller.

In fact, BT RX should not call `bt_hci_cmd_send()` either as that too is a deadlock hazard. But it's an improvement over `bt_hci_cmd_send_sync()`.

---

There is also the following code path, which does propagate the return value from the Controller out to public API. Fortunately, this is not documented as API.

```
bt_hci_cmd_send_sync
bt_conn_le_conn_update
send_conn_le_param_update
bt_conn_le_param_update
[Public API]
```

---

There's also a call path from `conn.c:deferred_work()` that reacts to the error value and clears `BT_CONN_PERIPHERAL_PARAM_AUTO_UPDATE`. This seems like it could be very slight behavioral change, where a the automatic connection parameter update will not trigger any more if we send a command, but the Controller responds with an error? I'm not sure whether this is intended or whether it matters. I'm inclined to say that we can ignore this. This code path was a bit hard to follow to see what exactly it does. @jhedberg, you recently looked at the automatic connection parameter update code. Did you dig this deep?

```
bt_hci_cmd_send_sync
bt_conn_le_conn_update
send_conn_le_param_update
bt_conn_le_param_update
deferred_work
[System Work Queue]
```

---

Overall, if our tests pass, I think this change should be accepted.

PavelVPV (changes requested):
@alwa-nordic , I see your point. What I doesn't look good here is that this change potentially masks a different issue. The description clearly says that this is caused by calling the `bt_conn_le_conn_update` on `sysworkq`, which can be related to the event pool size and not to the sync API call.

I'd prefer to have detailed explanation of the bug before changing behavior, because it can be some other command later that will trigger the assertion.

VinitMehta7492:
@alwa-nordic @PavelVPV Thank you for your review and comments.

The reason this change was done was considering the bt_conn_le_conn_update API being called from sysworkq context and [API](https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/subsys/bluetooth/host/hci_core.c#L503) returned without waiting for HCI_CMD_TIMEOUT value.

Please see the attached function call stack and sample app console logs before an assert is triggered:

image




**Sample App logs:**

[central_gatt_write.log](https://github.com/user-attachments/files/23416369/central_gatt_write.log)
[peripheral_gatt_write.log](https://github.com/user-attachments/files/23416370/peripheral_gatt_write.log)

In above scenario, I did confirm that the controller did respond to the command with success status. 

As per my understanding, this seems to be a synchronization issue when we call the sync API from SysworkQ context. Though, I also tried the suggestion of increasing the buffer as below with no change in the result:

- CONFIG_BT_BUF_EVT_RX_COUNT=20
- CONFIG_BT_BUF_CMD_TX_COUNT=10
- CONFIG_BT_BUF_EVT_RX_SIZE=255 

Let me know in case if more information is needed.

alwa-nordic:
Logs show `k_sem_take` `err` is `-11`. That corresponds with the following:

https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/include/zephyr/kernel.h#L3506-L3507

From your logs, it seems likely the timeout did not occur, since 10 seconds earlier is before `bt_enable()` was called. (It would be nice to have a log message showing when `bt_hci_cmd_send_sync` to confirm.) If the timeout did not occur then the semaphore was reset. I don't think that happens anywhere after it's created.

Are you sure you are adhering to API? Note that `buf` is required to be allocated with `bt_hci_cmd_alloc()`. And that buffer is moved into `bt_hci_cmd_send_sync()`. It cannot be reused.

https://github.com/zephyrproject-rtos/zephyr/blob/d054025fa39547198c73cce0a4fdacf3f2a72afa/include/zephyr/bluetooth/hci.h#L103-L105

PavelVPV:
Also, @VinitMehta7492 , can you confirm that `bt_hci_cmd_send_sync` is not called from ISR? I see `bt_uart_isr` in your backtrace. If UART is built with `CONFIG_UART_ASYNC_API`, `bt_hci_cmd_send_sync` is _probably_ called from `sysworkq`. But if UART driver is built with `CONFIG_UART_INTERRUPT_DRIVEN` and assertions are disabled (`CONFIG_ASSERT` is `n`, which is I guess by default), then this [assertion](https://github.com/zephyrproject-rtos/zephyr/blob/dbcfb77ce3d20784406d4d73a1de3f7f1bcea07c/kernel/sem.c#L136-L137) won't be triggered.

VinitMehta7492:
@PavelVPV, The [PR](https://github.com/zephyrproject-rtos/zephyr/pull/97913) fixes the problem.
However, I do see at times below logs. Seems a different issue though or might have exposed due to original issue being fixed

image

VinitMehta7492:
> > @PavelVPV, The [PR](https://github.com/zephyrproject-rtos/zephyr/pull/97913) seems not fixes the problem. I think it just delays it. There is one more observation:
> 
> @VinitMehta7492 , sorry for the late response. Is this bug still valid? If so, please create an issue to move this discussion there, I may give some suggestions there that you can try to debug this further.

@PavelVPV The final changes in the [PR](https://github.com/zephyrproject-rtos/zephyr/pull/97913) could fix the issue. Closing this thread now. Thanks.

PR #98868, opened 2025-11-04
boards: esp32_cam: remove unused SD card detect pin and fix LED polarity

### LINKED ISSUE #99140: esp32_cam board has invalid sdhc and led entries
### Describe the bug

`esp32_cam` board fails working when SDHC is used due to wrong board configuration. Its SDHC pinout should be tied to address 1 instead of 0.
Also, LED pin does not match its HW pinout operation model.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Build and flash: `west build -b esp32_cam/esp32/procpu samples/subsys/fs/fs_sample`

### Impact

### Environment

Linux, SDK 0.17.4, v4.3.0-rc3.

### Additional Context

_No response_

walidbadar:
The ESP32-CAM board does not connect the SD card detect pin, so the SDHC0_CD_GPIO21 pinmux configuration has been removed. This commit also updates the status LED to use active-low polarity for correct operation.

Screenshot from 2025-11-04 21-56-07

Screenshot from 2025-11-04 21-54-56

Reference: https://i0.wp.com/randomnerdtutorials.com/wp-content/uploads/2020/03/ESP32-CAM-AI-Thinker-schematic-diagram.png?quality=100&strip=all&ssl=1

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/99140

### DIFF
--- a/boards/aithinker/esp32_cam/esp32_cam-pinctrl.dtsi
+++ b/boards/aithinker/esp32_cam/esp32_cam-pinctrl.dtsi
@@ -43,12 +43,4 @@
 			output-low;
 		};
 	};
-
-	sdhc0_default: sdhc0_default {
-		group1 {
-			pinmux = <SDHC0_CD_GPIO21>;
-			bias-pull-up;
-			output-high;
-		};
-	};
 };
--- a/boards/aithinker/esp32_cam/esp32_cam_procpu.dts
+++ b/boards/aithinker/esp32_cam/esp32_cam_procpu.dts
@@ -19,14 +19,14 @@
 		led0 = &led0;
 		led1 = &led1;
 		watchdog0 = &wdt0;
-		sdhc0 = &sdhc0;
+		sdhc0 = &sdhc1;
 	};
 
 	leds {
 		compatible = "gpio-leds";
 
 		led0: led_0 {
-			gpios = <&gpio1 1 GPIO_ACTIVE_HIGH>;
+			gpios = <&gpio1 1 GPIO_ACTIVE_LOW>;
 			label = "Status LED";
 		};
 
@@ -43,7 +43,7 @@
 		zephyr,flash = &flash0;
 		zephyr,code-partition = &slot0_partition;
 		zephyr,bt-hci = &esp32_bt_hci;
-		zephyr,sdhc = &sdhc0;
+		zephyr,sdhc = &sdhc1;
 	};
 };
 
@@ -108,12 +108,9 @@
 };
 
 &sdhc {
-	sdhc0: sdhc@0 {
+	sdhc1: sdhc@1 {
 		status = "okay";
 
-		pinctrl-0 = <&sdhc0_default>;
-		pinctrl-names = "default";
-		power-delay-ms = <100>;
 		max-bus-freq = <52000000>;
 		bus-width = <4>;
 


boards/aithinker/esp32_cam/esp32_cam-pinctrl.dtsi
```diff
@@ -45,10 +45,6 @@
 	};
 
 	sdhc0_default: sdhc0_default {
```
sylvioalves: Perhaps you should remove the whole entry instead.

boards/aithinker/esp32_cam/esp32_cam_procpu.dts
```diff
@@ -108,7 +108,7 @@
 };
 
 &sdhc {
-	sdhc0: sdhc@0 {
```
sylvioalves, line 111: What is the reasoning for this address change?

walidbadar: there are some initialization error at runtime with sdhc driver when using `sdhc@0` changing to `sdhc@1` solved that problem also I confirmed it from `esp_wrover_kit_procpu.dts`. esp_wrover_kit is also using  `sdhc@1`.

https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/espressif/esp_wrover_kit/esp_wrover_kit_procpu.dts

boards/aithinker/esp32_cam/esp32_cam-pinctrl.dtsi
walidbadar: removing the whole entry would lead to compilation error as this config is used in `pinctrl-0 = <&sdhc0_default>` which is marked as required in espressif,esp32-sdhc-slot.yaml.

boards/aithinker/esp32_cam/esp32_cam_procpu.dts
sylvioalves: Ok, ESP32 has 2 shdc peripherals. Both `esp_wrover_kit` and `esp32_cam` use peripheral linked to slot 1, meaning it does need to be at address 1. So the change is correct.

boards/aithinker/esp32_cam/esp32_cam-pinctrl.dtsi
```diff
@@ -44,11 +44,7 @@
...
 	};
 
-	sdhc0_default: sdhc0_default {
-		group1 {
-			pinmux = <SDHC0_CD_GPIO21>;
-			bias-pull-up;
-			output-high;
-		};
+	sdhc1_default: sdhc1_default {
+		/* SD card detect pin is not connected on the ESP32-CAM board. */
```
sylvioalves: This is a bit odd. We should make SDHC detect pin not required. @raffarost will submit another PR fixing it in a couple hours.

walidbadar: image

sylvioalves: Yeah, I know build fails because the detect pin is required as of right now.

boards/aithinker/esp32_cam/esp32_cam_procpu.dts
```diff
@@ -108,10 +108,10 @@
...
+	sdhc1: sdhc@1 {
 		status = "okay";
 
-		pinctrl-0 = <&sdhc0_default>;
+		pinctrl-0 = <&sdhc1_default>;
 		pinctrl-names = "default";
 		power-delay-ms = <100>;
```
raffarost: you can remove this field too, as this board doesn't use a GPIO to control power to SD card

raffarost: `power-delay-ms` field is still there

sylvioalves (approved):

raffarost (approved):

PR #98871, opened 2025-11-04
net: lib: ocpp: fix misnamed State of Charge (SoC) measurand

### LINKED ISSUE #98870: OCPP state of charge cannot be reported
### Describe the bug

When an OCPP client tries to report `OCPP_OMM_CHARGING_PERCENT`, it is incorrectly being encoded as a measurand named "SoCState" which is not correct as per OCPP 1.6 specification. It should be "SoC"

See section "7.31. Measurand" of OCPP 1.6 standard.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Add OCPP_OMM_CHARGING_PERCENT reporting to in-tree OCPP sample. Notice OCPP server complaining about malformed message.

### Relevant log output

```shell

```

### Impact

### Environment

Zephyr 4.3 rc2

### Additional Context

_No response_

kartben:
As per OCPP 1.6, "State of charge of charging vehicle in percentage" is "Soc", not "SoCState".

Fixes: zephyrproject-rtos/zephyr#98870

### DIFF
--- a/subsys/net/lib/ocpp/core.c
+++ b/subsys/net/lib/ocpp/core.c
@@ -47,7 +47,7 @@ static struct {
 	FILL_METER_TABLE(OCPP_OMM_FAN_SPEED,
 			 "RPM", "rpm"),
 	FILL_METER_TABLE(OCPP_OMM_CHARGING_PERCENT,
-			 "SoCState", "Percent"),
+			 "SoC", "Percent"),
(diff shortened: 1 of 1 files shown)


rlubos (approved):

jukkar (approved):

ssekar15 (approved):
Thanks for fixing.

PR #98872, opened 2025-11-04
sca: enhancements for Polyspace tool

mbeckersys:
- fix for sysbuild
 - better output for prover mode
 - gracefully handle missing source files
 - better separation between kernel and user sources

Tested on main with various sample applications.

Signed-off-by: Martin Becker <mbecker@mathworks.com>


mbeckersys:
One check fails, but that is something unrelated to this PR. There seems to be a problem in Zephyr's test scripts.

cmake/sca/polyspace/sca.cmake
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
-# Copyright (c) 2024, The MathWorks, Inc.
-
+# Copyright (c) 2024-2025, The MathWorks, Inc.
+find_package(Zephyr)
```
tejlmand: why is this needed ?

`sca.cmake` is included as part of an outer `find_package(Zephyr)` call.

This also makes me wonder why the `include()` calls below are needed ?

cmake/sca/polyspace/sca.cmake
```diff
@@ -26,19 +26,20 @@ zephyr_get(POLYSPACE_OPTIONS)
 zephyr_get(POLYSPACE_OPTIONS_FILE)
 zephyr_get(POLYSPACE_PROG_NAME)
 zephyr_get(POLYSPACE_PROG_VERSION)
-message(TRACE "POLYSPACE_OPTIONS is ${POLYSPACE_OPTIONS}")
+message(NOTICE "POLYSPACE_OPTIONS is ${POLYSPACE_OPTIONS}")
```
tejlmand: having such messages as `TRACE` is fine, but we don't want to have every tool used in Zephyr to start printing all those options, as it will make it harder to notice the really important messages.

cmake/sca/polyspace/sca.cmake
```diff
@@ -26,19 +26,20 @@ zephyr_get(POLYSPACE_OPTIONS)
...
 
 # Get path and name of user application
-zephyr_get(APPLICATION_SOURCE_DIR)
-cmake_path(GET APPLICATION_SOURCE_DIR FILENAME APPLICATION_NAME)
-message(TRACE "APPLICATION_SOURCE_DIR is ${APPLICATION_SOURCE_DIR}")
-message(TRACE "APPLICATION_NAME is ${APPLICATION_NAME}")
+zephyr_get(APP_DIR)
+cmake_path(GET APP_DIR FILENAME APP_NAME)
+message(NOTICE "APP_DIR is ${APP_DIR}")
+message(NOTICE "APP_NAME is ${APP_NAME}")
```
tejlmand: same comment regarding `TRACE` vs `NOTICE`.
Also, source dir for current sample is already printed at start of CMake configure, so why again ?
Without sysbuild
```
$ west build -b native_sim samples/hello_world/
-- west build: making build dir /.../zephyr/build pristine
-- west build: generating a build system
Loading Zephyr default modules (Zephyr base).
-- Application: /.../zephyr/samples/hello_world
...
```

With sysbuild
```
$ west build -b native_sim samples/hello_world/
...
-- 
   *********************************
   * Running CMake for hello_world *
   *********************************

Loading Zephyr default modules (Zephyr base).
-- Application: /.../zephyr/samples/hello_world
...
```

cmake/sca/polyspace/sca.cmake
```diff
@@ -26,19 +26,20 @@ zephyr_get(POLYSPACE_OPTIONS)
...
-message(TRACE "POLYSPACE_OPTIONS is ${POLYSPACE_OPTIONS}")
+message(NOTICE "POLYSPACE_OPTIONS is ${POLYSPACE_OPTIONS}")
 
 
 # Get path and name of user application
-zephyr_get(APPLICATION_SOURCE_DIR)
-cmake_path(GET APPLICATION_SOURCE_DIR FILENAME APPLICATION_NAME)
-message(TRACE "APPLICATION_SOURCE_DIR is ${APPLICATION_SOURCE_DIR}")
-message(TRACE "APPLICATION_NAME is ${APPLICATION_NAME}")
+zephyr_get(APP_DIR)
```
tejlmand: why this change ?

`APPLICATION_SOURCE_DIR` is unique per sample, so in multi image build, such as hello_world with MCUboot, then that variable will point to the source of the image you're inside.

cmake/sca/polyspace/sca.cmake
```diff
@@ -87,11 +88,14 @@ set(POLYSPACE_OPTIONS_FILE_BASE ${POLYSPACE_RESULTS_DIR}/polyspace.psopts)
 add_custom_target(polyspace_configure ALL
   COMMAND ${POLYSPACE_CONFIGURE_EXE}
     -allow-overwrite
+    -allow-build-error
     -silent
     -prog ${POLYSPACE_PROG_NAME}
     -compilation-database ${CMAKE_BINARY_DIR}/compile_commands.json
     -output-options-file ${POLYSPACE_OPTIONS_FILE_BASE}
     ${POLYSPACE_CONFIGURE_OPTIONS}
+    || ${CMAKE_COMMAND} -E true # some targets don't have C/C++
```
tejlmand: what do you mean `# some targets don't have C/C++` ?

cmake/sca/polyspace/sca.cmake
```diff
@@ -133,10 +141,11 @@ add_dependencies(polyspace-results polyspace-analyze)
 
 add_custom_command(TARGET polyspace-results POST_BUILD
   COMMAND ${CMAKE_CURRENT_LIST_DIR}/polyspace-print-console.py
-  ${POLYSPACE_RESULTS_FILE}
+    ${POLYSPACE_RESULTS_FILE} || ${CMAKE_COMMAND} -E true
```
tejlmand: so the target can never fail.

How will users find out if `polyspace-print-console.py` is broken ?

tejlmand (changes requested):

cmake/sca/polyspace/sca.cmake
mbeckersys: This change was necessary to fix sysbuild. We need to separate each build target in SCA, otherwise we can't analyze. With the previous cmake file, we had build errors, since sysbuild seems to change this path intermittently. We want to be able to check both the bootloader and the app, and I didn't see a better way to do that.

cmake/sca/polyspace/sca.cmake
mbeckersys: For incremental builds of mcuboot, no sources are reported in the compilation database. In that case we allow to proceed to the next step. If we don't to that, then the build tops here and we will never see the application sources in incremental builds.

cmake/sca/polyspace/sca.cmake
mbeckersys: will fix

cmake/sca/polyspace/sca.cmake
mbeckersys: will fix

cmake/sca/polyspace/sca.cmake
mbeckersys: You are right, some are superfluous. Removed.

The remaining includes are necessary, because we use the board name and git information to construct a unique name for the analysis, that can be used for cross-referencing analysis results with CI jobs.

cmake/sca/polyspace/sca.cmake
tejlmand: no, sysbuild don't.
However, there is only one `APP_DIR` which points to main application, so if using sysbuild and enabling mcuboot, then APP_DIR points to source of the app and not source of mcuboot.

Perhaps creating an issue and describe the error seen, together with expected output before fixing anything.

At least that will make it easier to review the changes.

cmake/sca/polyspace/sca.cmake
tejlmand: > For incremental builds of mcuboot, no sources are reported in the compilation database.

That I have never experienced.
`compile_commands.json` should not depend on whether the build is incremental or not.

This sounds more like missing / wrong dependencies on the custom target which mean it's executed before the compile_commands.json has been generated.

That should be fixed by proper dependency handling and not by silently ignoring this situation.
Especially because this proposal will result in different results depending on how many build threads your build system might spawn.

Please provide a proper fix, or even better, an issue with a good description, including when this happen, like `ninja -j0` vs `ninja -j1` vs `make` (sometimes `ninja` and `make` file generators behave differently, and thereby revealing wrong / missing dependencies differently)

cmake/sca/polyspace/sca.cmake
tejlmand: but why is `include(board)` still there ?
CMake module `board` is included by `find_package(Zephyr)` so should already be there.

Note, please let the reviewer decide if the conversation is resolved.

cmake/sca/polyspace/sca.cmake
mbeckersys: Agreed, did a rewrite. Made the case explicit where results are empty, and letting cmake pick up any other errors/exceptions.

cmake/sca/polyspace/sca.cmake
mbeckersys: I don't think it is a bug in my cmake script, and I see no other way to fix it. Here are the details:
- I have double-checked the dependency to the compilation database (CCDB), it is correct
- we want to scan all cmake targets
- If there are no SCA-compatible entries in the CCDB (C/C++ for Polyspace), we can't do anything but skip
- this happens in two scenarios: 1) in incremental builds the CCDB only contains the changed files 2) pristine builds which genuinely do not contain any sources that the SCA tool supports

Nothing to fix here, IMHO.

cmake/sca/polyspace/sca.cmake
mbeckersys: Why is there a difference between include(git) and include(board)? without git it breaks, without board it works. I kept board for consistency, but I can as well remove it

cmake/sca/polyspace/sca.cmake
mbeckersys: Ah, thanks for that, this change was an actual bug. I thought that the variable name caused the problem, but it was actually zephyr_get(APPLICATION_SOURCE_DIR) that broke the build. If this is unintended, I can open an issue for that.

Switched back to APPLICATION_SOURCE_DIR, now it works as intended.

Is there a way to extract the target name? Right now I am using the last part of APPLICATION_SOURCE_DIR, which for the target mcuboot results in "zephyr".

cmake/sca/polyspace/sca.cmake
mbeckersys: The behavior "CCDB only shows changed files" has resolved itself, now that I made your suggested fix around APPLICATION_SOURCE_DIR.

cmake/sca/polyspace/sca.cmake
tejlmand: because `git` is just a CMake module which is not required by `find_package(Zephyr)` itself.
Therefore it's not loaded by Zephyr, but is loaded by features which might need `git`.

Zephyr itself can be built without `git`.

cmake/sca/polyspace/sca.cmake
tejlmand: > but it was actually zephyr_get(APPLICATION_SOURCE_DIR) that broke the build. If this is unintended, I can open an issue for that.

Fixed here: https://github.com/zephyrproject-rtos/zephyr/pull/99197

as I said :wink: 
> Perhaps creating an issue and describe the error seen, together with expected output before fixing anything.

cmake/sca/polyspace/sca.cmake
tejlmand:
```cmake
  || ${CMAKE_COMMAND} -E true 
```
is still present in code.

cmake/sca/polyspace/sca.cmake
```diff
@@ -105,10 +103,12 @@ add_custom_target(polyspace-analyze ALL
     -options-file ${POLYSPACE_OPTIONS_FILE_BASE}
     ${POLYSPACE_PROG_VERSION}
     ${POLYSPACE_OPTIONS_ZEPHYR}
+    ${POLYSPACE_OPTIONS_ZEPHYR_ONLYAPP}
     ${POLYSPACE_OPTIONS_FILE}
     ${POLYSPACE_OPTIONS}
-      || ${CMAKE_COMMAND} -E true # allow to continue processing results
+    || ${CMAKE_COMMAND} -E true # when there are no sources...no easy way to skip this target with cmake syntax
```
tejlmand, line 109: but there are sources.

as you have just experienced in https://github.com/zephyrproject-rtos/zephyr/pull/98872#discussion_r2511001317 then just trying to get CMake / build to proceed and not understading what actually happens is not fixing anything.

Instead it will lead to hard to understand / debug issues.

Now all errors raised by `${POLYSPACE_EXE}` will be silently ignored.

cmake/sca/polyspace/sca.cmake
```diff
@@ -118,8 +118,10 @@ add_custom_target(polyspace-results ALL
     -results-dir ${POLYSPACE_RESULTS_DIR}
     -output-name ${POLYSPACE_RESULTS_FILE}
     -format csv
-      || ${CMAKE_COMMAND} -E true # allow to continue processing results
+      || ${CMAKE_COMMAND} -E true # this just exports the results, skip if no results to continue with next target
```
tejlmand, line 121: same comment applies regarding failures.

share/sysbuild/image_configurations/BOOTLOADER_image_default.cmake
```diff
@@ -56,11 +56,11 @@ else()
...
-if(SB_CONFIG_MCUBOOT_MODE_RAM_LOAD_WITH_REVERT)
-  set_config_bool(${ZCMAKE_APPLICATION} CONFIG_BOOT_RAM_LOAD_REVERT y)
-else()
-  set_config_bool(${ZCMAKE_APPLICATION} CONFIG_BOOT_RAM_LOAD_REVERT n)
-endif()
+#if(SB_CONFIG_MCUBOOT_MODE_RAM_LOAD_WITH_REVERT)
+#  set_config_bool(${ZCMAKE_APPLICATION} CONFIG_BOOT_RAM_LOAD_REVERT y)
+#else()
+#  set_config_bool(${ZCMAKE_APPLICATION} CONFIG_BOOT_RAM_LOAD_REVERT n)
+#endif()
```
tejlmand, line 63: ?

tejlmand (changes requested):

PR #98514, opened 2025-10-29
Introduce mpipe (Multimedia Pipeline) subsystem

ngphibang:
In Zephyr today, multimedia applications—such as those involving video, audio, display, vision, and graphics—are typically implemented as simple, domain-specific sample applications. While these are sufficient for basic use cases, they quickly become inadequate when dealing with:

- Complex processing chains, e.g., multiple processing components between a camera and a display. 
- Cross-domain scenarios, e.g., an MPEG-DASH player handling video, audio, and subtitles streamed over a network, with dynamic resolution changes.

In such cases, application complexity increases significantly. Developers must manually manage buffer allocation, queuing and dequeuing for each component as well as synchronization between components across the pipeline. Moreover, similar functionality often needs to be reimplemented across projects, leading to duplicated effort. Applications also tend to require extensive customization for each use case and become fragile to even minor changes in requirements.

To address these challenges, this PR introduces mpipe (Multimedia Pipeline) subsystem — a lightweight multimedia framework designed specifically for Zephyr.

This PR only contains the core framework and tests. The utils, plugins and samples are in the following separate PRs (more can be added in the future):

- https://github.com/zephyrproject-rtos/zephyr/pull/114477
- https://github.com/zephyrproject-rtos/zephyr/pull/111526
- https://github.com/zephyrproject-rtos/zephyr/pull/108346
- https://github.com/zephyrproject-rtos/zephyr/pull/108345
- https://github.com/zephyrproject-rtos/zephyr/pull/108344
- https://github.com/zephyrproject-rtos/zephyr/pull/108889
- https://github.com/zephyrproject-rtos/zephyr/pull/112592

- https://github.com/zephyrproject-rtos/zephyr/pull/108347
- https://github.com/zephyrproject-rtos/zephyr/pull/100480
- https://github.com/zephyrproject-rtos/zephyr/pull/108348
- https://github.com/zephyrproject-rtos/zephyr/pull/111608
- https://github.com/zephyrproject-rtos/zephyr/pull/100673

**What mpipe does:**

- Decompose multimedia application into modular / built-in components called elements
- Elements can be connected together to form a customized processing chain called pipeline
- Elements can be easily extended / inherited to create customized elements (plugins)

**What mpipe offers:**

- Reusablity & maintainability
- Extentisiblity
- Facilitate the tasks of both application developers and libZMP plugin developers
- Support (very) complex multimedia application scenarios

[...]

Issue #97061, opened 2025-10-06
Transition to C17 as the minimum required C standard version


carlescufi:
### Summary

This task tracks the transition from C99 to C17 as the minimum required C standard version in order to compile Zephyr applications.

### Describe the solution you'd like

The following compilers have been verified to provide full C17 support:

 - [x] GCC (See [C language](https://gcc.gnu.org/onlinedocs/gcc/Standards.html#C-Language))
 - [x] Clang/LLVM (See [C Support in Clang](https://clang.llvm.org/c_status.html))
 - [x] IAR (See [this comment](https://github.com/zephyrproject-rtos/zephyr/issues/30105#issuecomment-2846636691))

nordicjm:
Looks like nanopb has problems with this @pdgendt https://github.com/zephyrproject-rtos/zephyr/pull/97076

pdgendt:
> Looks like nanopb has problems with this [@pdgendt](https://github.com/pdgendt) [#97076](https://github.com/zephyrproject-rtos/zephyr/pull/97076)

See my comment here: https://github.com/zephyrproject-rtos/zephyr/pull/97076#pullrequestreview-3308683715

nashif:
need data from mwdt and cadence based toolchains used by many platforms.

carlescufi:
> need data from mwdt and cadence based toolchains used by many platforms.

Who can provide that @nashif ?

nashif:
> > need data from mwdt and cadence based toolchains used by many platforms.
> 
> Who can provide that [@nashif](https://github.com/nashif) ?

@ruuddw @kv2019i  ^^

kv2019i:
>>> need data from mwdt and cadence based toolchains used by many platforms.

Sorry, late answer. In Intel adsp teams, for newer platforms we are using clang/llvm 10.x based Cadence (2022) toolchain and this should be ok for C17. For older platforms, we use old gcc-4.4.x based Cadence (2017) toolchain and this will obviously not have C17 as the toolchain is older than the language spec. We understand this very old and we cannot reasonably expect upstream to hold back on C17 because of this. We'll either have to drop these platforms, or investigate possibility for a toolchain upgrade. I can't speak for other DSP vendors using Zephyr -- I have only partial view on the toolchain versions that are used.. I see at least NXP and Mediatek folks are using newer Cadence toolchains based on the public CI workflows available.

ruuddw:
[...]

PR #98875, opened 2025-11-04
sys: util: Add SIGN, gcd and lcm

ngphibang:
Add some helpers to `util.h`

- SIGN macro
- GCD (Greatest Common Divisor)
- LCM (Least Common Multiple)

which are helpul for example, used in https://github.com/zephyrproject-rtos/zephyr/pull/98514

### DIFF
--- a/include/zephyr/sys/util.h
+++ b/include/zephyr/sys/util.h
@@ -1038,6 +1038,70 @@ static inline size_t sys_count_bits(const void *value, size_t len)
  */
 #define SIGN(x) (((x) > 0) - ((x) < 0))
 
+/**
+ * @brief Compute the Greatest Common Divisor (GCD) of two integers
+ * using the Euclidean algorithm.
+ *
+ * @param a First integer
+ * @param b Second integer
+ *
+ * @return The greatest common divisor of a and b, always returns an unsigned value.
+ *         If one of the parameters is 0, returns the absolute value of the other parameter.
+ */
+#define gcd(a, b) ((((__typeof__(a))-1) < 0) ? gcd_s(a, b) : gcd_u(a, b))
+
+static ALWAYS_INLINE uint32_t gcd_u(uint32_t a, uint32_t b)
+{
+	uint32_t c;
+
+	if (a == 0) {
+		return b;
+	}
+
+	if (b == 0) {
+		return a;
+	}
+
+	c = a % b;
+	while (c != 0) {
+		a = b;
+		b = c;
+		c = a % b;
+	}
+
+	return b;
+}
+
+static ALWAYS_INLINE uint32_t gcd_s(int32_t a, int32_t b)
+{
+	return gcd_u(a < 0 ? -(uint32_t)a : (uint32_t)a, b < 0 ? -(uint32_t)b : (uint32_t)b);
+}
+
+/**
+ * @brief Compute the Least Common Multiple (LCM) of two integers.
+ *
+ * @param a First integer
+ * @param b Second integer
+ *
+ * @retval The least common multiple of a and b.
+ * @retval 0 if either input is 0.
+ */
+#define lcm(a, b) ((((__typeof__(a))-1) < 0) ? lcm_s(a, b) : lcm_u(a, b))
+
+static ALWAYS_INLINE uint64_t lcm_u(uint32_t a, uint32_t b)
+{
+	if (a == 0 || b == 0) {
+		return 0;
+	}
+
+	return (uint64_t)(a / gcd_u(a, b)) * (uint64_t)b;
+}
+
+static ALWAYS_INLINE uint64_t lcm_s(int32_t a, int32_t b)
+{
+	return lcm_u(a < 0 ? -(uint32_t)a : (uint32_t)a, b < 0 ? -(uint32_t)b : (uint32_t)b);
+}
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/zephyr/sys/util.h
+++ b/include/zephyr/sys/util.h
@@ -1027,6 +1027,17 @@ static inline size_t sys_count_bits(const void *value, size_t len)
 	return cnt;
 }
 
+/**
+ * @brief Returns the sign of a number.
+ *
+ * @param x The input value to determine the sign
+ *
+ * @retval 1 if x is positive
+ * @retval -1 if x is negative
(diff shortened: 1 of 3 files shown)


include/zephyr/sys/util.h
```diff
@@ -1038,6 +1038,61 @@ static inline size_t sys_count_bits(const void *value, size_t len)
...
 #define SIGN(x) ((x > 0) - (x < 0))
 
+/**
+ * @brief Compute the Greatest Common Divisor (GCD) of two integers
+ * using the Euclidean algorithm.
+ *
+ * @param a First integer
+ * @param b Second integer
+ *
+ * @return The greatest common divisor of a and b, always returns a positive value.
```
peter-mitsis: This statement is not quite true. If both `a` and `b` are `INT_MIN`, then the result will be `INT_MIN` -- a negative value.

include/zephyr/sys/util.h
```diff
@@ -1027,6 +1027,17 @@ static inline size_t sys_count_bits(const void *value, size_t len)
...
+ *
+ * @param x The input value to determine the sign
+ *
+ * @retval 1 if x is positive
+ * @retval -1 if x is negative
+ * @retval 0 if x is zero
+ */
+#define SIGN(x) ((x > 0) - (x < 0))
+
 #ifdef __cplusplus
```
dbaluta, line 1105: Add paranthesis around (x) on the right side to avoid problems with complex expressions.

include/zephyr/sys/util.h
ngphibang: By mathematical definition, the greatest common divisor (GCD) of two integers is always a non-negative integer. But you are right, the old implementation didn't treat this edge case when both a and b are INT_MIN. We fixed this in the current push. Thanks !

include/zephyr/sys/util.h
ngphibang: yes, fixed.

include/zephyr/sys/util.h
```diff
@@ -1027,6 +1027,95 @@ static inline size_t sys_count_bits(const void *value, size_t len)
...
+ * @brief Compute the Greatest Common Divisor (GCD) of two integers
+ * using the Euclidean algorithm.
+ *
+ * @param a First integer
+ * @param b Second integer
+ *
+ * @return The greatest common divisor of a and b, always returns a positive value.
+ *         If one of the parameters is 0, returns the absolute value of the other parameter.
+ */
+#define gcd(a, b) \
```
josuah: Should it be `GCD` (uppercase) like in the issue description or lowercase on purpose (i.e. mimick a regular function since it uses _Generic so is type-sensitive)?

josuah (commented):
Good idea to split dependencies out of the main PR, which will hopefully make the main PR quicker to merge.

fabiobaltieri:
Yeah prob a good idea to do single evaluation macros whenever possible, for generics, this needs to wait for https://github.com/zephyrproject-rtos/zephyr/issues/97061

include/zephyr/sys/util.h
josuah: Not relevant given this https://github.com/zephyrproject-rtos/zephyr/pull/98875#issuecomment-3526295070 about MIN/MAX rework PR. Thanks!

josuah (dismissed):

include/zephyr/sys/util.h
```diff
@@ -1027,6 +1027,81 @@ static inline size_t sys_count_bits(const void *value, size_t len)
...
+		b = c;
+		c = a % b;
+	}
+
+	return b;
+}
+
+static ALWAYS_INLINE uint32_t gcd_s(int32_t a, int32_t b)
+{
+	return gcd_u(a < 0 ? -a : a, b < 0 ? -b : b);
```
mathieuchopstm: Isn't this undefined behavior if `a` or `b` is `INT32_MIN`?

trunghieulenxp: The negative of INT_MIN is always INT_MIN when using two's complement representation. However, this behavior may depend on the platform. Therefore, I think we could cast this to uint32_t before doing the absolute operation to avoid the overflow.

ngphibang: Fixed the overflow by INT_MIN.

fabiobaltieri (approved):

mathieuchopstm (approved):

include/zephyr/sys/util.h
mathieuchopstm: Seems correct. Per cppreference, casts perform an implicit conversion and:
> A value of any integer type can be implicitly converted to any other integer type. Except where covered by promotions and boolean conversions above, the rules are:
>
> [...] if the target type is unsigned, the value 2^`b`, where `b` is the number of value bits in the target type, is repeatedly subtracted or added to the source value until the result fits in the target type. In other words, unsigned integers implement modulo arithmetic.

> Unsigned integer arithmetic is always performed modulo 2^`n` where `n` is the number of bits in that particular integer. 
> [...]
> ```
> // Defined behavior since arithmetic is performed for unsigned integer.
> // Hence, the calculation is (-1) modulo (2 raised to n) = UINT_MAX, where n is
> // the number of bits of unsigned int. If unsigned int is 32-bit long, then this
> // gives (-1) modulo (2 raised to 32) = 4294967295
> ```

* `INT_MIN` = -2147483648
* thus `(uint32_t)INT_MIN` = -2147483648 + 4294967296 = 2147483648
* thus `-(uint32_t)INT_MIN` = -(2147483648) = 4294967296 - 2147483648 = 2147483648

josuah: For curiosity, musl-libc just went with `-a`
https://git.musl-libc.org/cgit/musl/tree/src/stdlib/abs.c

josuah (approved):

avolmat-st (approved):

include/zephyr/sys/util.h
mathieuchopstm: > For curiosity, musl-libc just went with `-a` https://git.musl-libc.org/cgit/musl/tree/src/stdlib/abs.c

According to [cppreference](https://en.cppreference.com/w/c/numeric/math/abs.html), `abs(INT_MIN)` is UB on 2's complement systems.

nashif (approved):

PR #98877, opened 2025-11-04
Opamp usability

KozhinovAlexander:
Followinmg changes were introduced:

[drivers: opamp: Kconfig: log module dependency](https://github.com/zephyrproject-rtos/zephyr/pull/98877/commits/2f7b551bd141c0341008082d5abd308402c0724b):
  use common log module dependency pattern in KConfig
  add LOG as module dependency for opamp driver

[dts: bindings: opamp: opamp-controller.yaml](https://github.com/zephyrproject-rtos/zephyr/pull/98877/commits/88802a0c16d945baee8188e5af2ac5dd20f4ca02):
  make functional mode required
  add functional modes description with circuit examples

[samples: drivers: opamp: output_measure: prj.conf](https://github.com/zephyrproject-rtos/zephyr/pull/98877/commits/90f3bf284b1f416179126ac79950e5ad2fcc9756):
  enable logging
  add log buffer size and debug options and comment them

[samples: drivers: opamp: output_measure](https://github.com/zephyrproject-rtos/zephyr/pull/98877/commits/3d940a05243d504c2053161fb9b8e5e4b459d1cb):
  show gain value at runtime as a starting
  use ADC calibration before conversion
  update README with sample gain prints


dts/bindings/opamp/opamp-controller.yaml
```diff
@@ -7,6 +7,7 @@ include:
 
 properties:
   functional-mode:
+    required: true
```
KozhinovAlexander, line 10: @decsny @ZhaoxiangJin The functional mode can be configured only in DTS thus shall be required from functional point of view. What do you think?
I need your input to know, how to proceed further, because if it will get required, some NXP board overlays shall be adjusted accordingly, otherwise twister tests will fail.

drivers/opamp/Kconfig
```diff
@@ -9,7 +10,9 @@ menuconfig OPAMP
 if OPAMP
 
 module = OPAMP
-module-str = opamp
+module-dep = LOG
+module-str = Log level for OPAMP driver
+module-help = Sets log level OPAMP driver
```
ZhaoxiangJin: `module-dep` and `module-help` is not used in `Kconfig.template.log_config`, Are they really needed in this file?

Change `module-str` from `opamp` to `Log level for OPAMP driver` is not make sense for me. `module-str` is used in `prompt "Max compiled-in log level for $(module-str)"`. If `module-str = Log level for OPAMP driver`, then will expand to `prompt "Max compiled-in log level for Log level for OPAMP driver"`

dts/bindings/opamp/opamp-controller.yaml
ZhaoxiangJin: it should be `required: true`, please update any NXP board overlays that need modification, thanks.

drivers/opamp/Kconfig
KozhinovAlexander: > `module-dep` and `module-help` is not used in `Kconfig.template.log_config`, Are they really needed in this file?
> 

looking at [drivers/net/KConfig](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/net/Kconfig#L158-L162) I see this entries.

> Change `module-str` from `opamp` to `Log level for OPAMP driver` is not make sense for me. `module-str` is used in `prompt "Max compiled-in log level for $(module-str)"`. If `module-str = Log level for OPAMP driver`, then will expand to `prompt "Max compiled-in log level for Log level for OPAMP driver"`

Thank you, it is done now.

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -24,6 +24,28 @@ uint8_t size = ARRAY_SIZE(gain);
...
+	[OPAMP_GAIN_7] = "7",
+	[OPAMP_GAIN_8] = "8",
+	[OPAMP_GAIN_15] = "15",
+	[OPAMP_GAIN_16] = "16",
+	[OPAMP_GAIN_31] = "31",
+	[OPAMP_GAIN_32] = "32",
+	[OPAMP_GAIN_33] = "33",
+	[OPAMP_GAIN_63] = "63",
+	[OPAMP_GAIN_64] = "64",
+	[OPAMP_GAIN_64 + 1] = "default",
```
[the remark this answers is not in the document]
KozhinovAlexander: Rejected: OPAMP_GAIN_DEFAULT is not in enum and is not planned to be added.

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -122,7 +147,8 @@ int main(void)
 			return 0;
 		}
 
-		printf("OPAMP output is: %d(mv)\n", sample_buffer);
+		printf("OPAMP output (@gain: %s) is: %d(mv)\n", opamp_gain2str[gain_value],
+		       sample_buffer);
```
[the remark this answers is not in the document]
KozhinovAlexander: Rejected: OPAMP_GAIN_DEFAULT is not in enum and is not planned to be added.

dts/bindings/opamp/opamp-controller.yaml
KozhinovAlexander: done

drivers/opamp/Kconfig
ZhaoxiangJin: The net Kconfig source [Kconfig.template.log_config.net](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/net/Kconfig#L162), and [module-dep](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/net/Kconfig.template.log_config.net#L10) is used in this kconfig template, so there is no problem.
But in OPAMP Kconfig, you are just source Kconfig.template.log_config, why we define unused Kconfig variable?

boards/nxp/frdm_mcxa156/frdm_mcxa156.dts
```diff
@@ -239,6 +239,7 @@ zephyr_mipi_dbi_parallel: &flexio0_lcd {
 	status = "okay";
 	pinctrl-0 = <&pinmux_opamp0>;
 	pinctrl-names = "default";
+	functional-mode = "follower";
```
ZhaoxiangJin: This doesn't seem to be the correct approach. The `functional-mode` shouldn't be placed in the board-level DTSI because some downstream might reuse `soc.dts` instead of `board.dts`, which could cause build issues. Could you please move the `functional-mode` from the NXP board overlay in the output_measure sample into the corresponding `soc.dts`? If this step is complicated, I can help with it.

KozhinovAlexander: Thank you. I will move it.

KozhinovAlexander: moved to soc level

dts/arm/nxp/nxp_lpc55S3x_common.dtsi
ZhaoxiangJin, line 1: You should keep commits bisectable, that is e79d40d1db74eb19fb77b414e2fade7afbb009e0 should before https://github.com/zephyrproject-rtos/zephyr/pull/98877/commits/7bd02a1a4d013c02cceac4e908c1496563e3044a

drivers/opamp/Kconfig
```diff
@@ -9,7 +10,9 @@ menuconfig OPAMP
 if OPAMP
 
 module = OPAMP
-module-str = opamp
+module-dep = LOG
+module-str = OPAMP driver
+module-help = Sets log level for OPAMP driver
```
ZhaoxiangJin: sorry, I still confused why `module-dep` and `module-help` are needed.

ZhaoxiangJin (changes requested):
1. commits not bisectable
2. not clear why `module-dep` and `module-help` are added, since they are not used IMO.

dts/arm/nxp/nxp_lpc55S3x_common.dtsi
KozhinovAlexander: done

drivers/opamp/Kconfig
KozhinovAlexander: both entries are emoved now

ZhaoxiangJin (commented):
no more change request, I will do test the next day, thanks for your effort.

KozhinovAlexander:
> no more change request, I will do test the next day, thanks for your effort.

Thank you for your review in a very fast manner 😃

KozhinovAlexander:
> when i build and run the sample base on your branch, i got:
> 
> ```
> *** Booting ZOPAMP output (@gain: 4) is: 1556(mv)
> ephyr OS buiOPAMP output (@gain: 2) is: 937(mv)
> ld v4.3.0-rc2OPAMP output (@gain: 1) is: 623(mv)
> -101-g65eb32acd9bb ***
> ```
> 
> Could you please help replace all printf calls in main.c with printk?

Changed to printk and rebased on top of newest main.

KozhinovAlexander:
> > > when i build and run the sample base on your branch, i got:
> > > ```
> > > *** Booting ZOPAMP output (@gain: 4) is: 1556(mv)
> > > ephyr OS buiOPAMP output (@gain: 2) is: 937(mv)
> > > ld v4.3.0-rc2OPAMP output (@gain: 1) is: 623(mv)
> > > -101-g65eb32acd9bb ***
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Could you please help replace all printf calls in main.c with printk?
> > 
> > 
> > Changed to printk and rebased on top of newest main.
> 
> It would be better to change all printf to printk in the main.c, could you please help to update?

done. all printf were changed to printk

KozhinovAlexander:
@ZhaoxiangJin please take a look

ZhaoxiangJin (dismissed):

drivers/opamp/Kconfig
```diff
@@ -9,7 +10,7 @@ menuconfig OPAMP
 if OPAMP
 
 module = OPAMP
-module-str = opamp
+module-str = OPAMP driver
```
kartben: given the original version is what 99.5% of the drivers use as a convention, I would not change it

ZhaoxiangJin: @KozhinovAlexander, it is better to keep the convention.

KozhinovAlexander: Sure, I'll do. Thank you both.

KozhinovAlexander: done. The KConfig has now no relevant changes, thus whole commit could be safely dropped, which was done too.

dts/bindings/opamp/opamp-controller.yaml
```diff
@@ -17,6 +18,94 @@ properties:
...
+                     ------|- \   |
+                           |   >--*-- Vout
+                     ------|+ /
+                     |     |/
+                     |
+      Vin2 --/\/\/\--*--/\/\/\-----
+               Rin        Rf      |
+                                 GND
+
+      e) standalone Mode
```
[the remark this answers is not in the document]
KozhinovAlexander: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/zephyrproject-rtos/zephyr/pull/98877#discussion_r2511908404)

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -24,31 +24,53 @@ uint8_t size = ARRAY_SIZE(gain);
 uint8_t size;
```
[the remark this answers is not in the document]
KozhinovAlexander: Can not be implemented:

```bash
4a2c32549a4e13db4263cb12ccaefbfc89aece2d:41: ERROR:GLOBAL_INITIALISERS: do not initialise globals to 0
#41: FILE: samples/drivers/opamp/output_measure/src/main.c:24:
+uint8_t size = 0;
```

dts/bindings/opamp/opamp-controller.yaml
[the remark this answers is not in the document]
kartben: I doubt this will work (especially w.r.t authorship/ sign off) but would be cool if it does :)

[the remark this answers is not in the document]
KozhinovAlexander: > I doubt this will work (especially w.r.t authorship/ sign off) but would be cool if it does :)

Let's see. 😃 Yesterday Claude was very succesfull for rebase on my machine. I had to rework all the commit messages anywasm but the commits split from one commit to multiple ones logically separated was good.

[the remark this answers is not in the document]
KozhinovAlexander: Did it manually in the end. Current copilot-github integration for reviews is still lacking well defind structure :) But the input from copilot is good, whus I will change to lower-case both words in section names.

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -97,32 +119,36 @@ int main(void)
...
 	};
 
 	ret = device_initialize(&adc_channel, opamp_dev, &sequence);
 	if (ret < 0) {
-		printf("Initialization failed with code (%d)\n", ret);
+		printk("Initialization failed with code (%d)\n", ret);
 		return 0;
 	}
 
+	enum opamp_gain gain_value = (enum opamp_gain)(OPAMP_GAIN_64 + 1); /* default gain */
```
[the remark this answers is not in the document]
KozhinovAlexander: This is the only solution otherwise enum elements in the opamp.h file have to be changed. Rejected.

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -24,31 +24,53 @@ uint8_t size = ARRAY_SIZE(gain);
...
+	[OPAMP_GAIN_7] = "7",
+	[OPAMP_GAIN_8] = "8",
+	[OPAMP_GAIN_15] = "15",
+	[OPAMP_GAIN_16] = "16",
+	[OPAMP_GAIN_31] = "31",
+	[OPAMP_GAIN_32] = "32",
+	[OPAMP_GAIN_33] = "33",
+	[OPAMP_GAIN_63] = "63",
+	[OPAMP_GAIN_64] = "64",
+	[OPAMP_GAIN_64 + 1] = "default",
```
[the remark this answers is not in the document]
KozhinovAlexander: requires change in the enum of the API. Thus can not be implemented. Rejected.

samples/drivers/opamp/output_measure/src/main.c
```diff
@@ -24,31 +24,53 @@ uint8_t size = ARRAY_SIZE(gain);
 uint8_t size;
 #endif
 
+const char *opamp_gain2str[] = {
```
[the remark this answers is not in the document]
KozhinovAlexander: Current implementation is applicable for sample.

dts/bindings/opamp/opamp-controller.yaml
```diff
@@ -17,6 +18,93 @@ properties:
     description: |
       Selects opamp functional mode.
 
+      a) inverting mode
```
etienne-lms: Maybe order in the same order used in the enumerated list above.

samples/drivers/opamp/output_measure/src/main.c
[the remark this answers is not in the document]
etienne-lms: This makes `enum opamp_gain` a value that is not part of the enumerated list. Some toolchain may complain.
Maybe consider:
```patch
+	const char *gain_str = "default";

	do {
#if OPAMP_SUPPORT_PROGRAMMABLE_GAIN
		/* For OPAMPs which support programmable gain, retrieve the gain array. */
		ret = opamp_set_gain(opamp_dev, gain[size - 1]);		
		if (ret < 0) {
			printk("OPAMP gain set failed with code (%d)\n", ret);
			return 0;
		}

+		gain_str = opamp_gain2str[gain[size - 1]];
		size--;
#endif
		ret = read_opamp_output(&adc_channel, &sequence, &sample_buffer);
		if (ret < 0) {
			printk("OPAMP output read failed with code (%d)\n", ret);
			return 0;
		}

-		printf("OPAMP output is: %d(mv)\n", sample_buffer);
+		printk("OPAMP output (@gain: %s) is: %d(mv)\n", gain_str, sample_buffer);

File: dts/bindings/opamp/opamp-controller.yaml (hunk 13 above)
Comment (KozhinovAlexander) [reply dts/bindings/opamp/opamp-controller.yaml]: Thnx. Done

File: samples/drivers/opamp/output_measure/src/main.c (hunk 10 above)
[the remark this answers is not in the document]
Comment (KozhinovAlexander) [reply samples/drivers/opamp/output_measure/src/main.c]: This is a good idea. Thnx. I've added changes you suggest.

File: dts/bindings/opamp/opamp-controller.yaml [hunk 14]
```diff
@@ -17,6 +18,93 @@ properties:
...
+
+      - The opamp mode is defined by external circuitry
+
+                |\
+      Vin- -----|- \
+                |   >---- Vout
+      Vin+ -----|+ /
+                |/
+
+      references:
```
etienne-lms:
```suggestion
      References:

File: samples/drivers/opamp/output_measure/src/main.c (hunk 12 above)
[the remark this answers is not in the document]
Comment (etienne-lms) [reply samples/drivers/opamp/output_measure/src/main.c]: IMHO, adding `static` here would be better, for consistency.

File: samples/drivers/opamp/output_measure/src/main.c [hunk 15]
```diff
@@ -97,32 +118,34 @@ int main(void)
...
 	};
 
 	ret = device_initialize(&adc_channel, opamp_dev, &sequence);
 	if (ret < 0) {
-		printf("Initialization failed with code (%d)\n", ret);
+		printk("Initialization failed with code (%d)\n", ret);
 		return 0;
 	}
 
+	const char *gain_str = "default";
```
etienne-lms, line 130: Could you add an empty line below to separate local variable definition from instructions?

etienne-lms (commented):
For commit "samples: drivers: opamp: output_measure".
Could you mention in the commit message that `printf()` is replaced with `printk()` (and preferably a word on the rational).
The commit message header line misses something, like:
```patch
-samples: drivers: opamp: output_measure
+samples: drivers: opamp: output_measure: enable calibration and log gain

File: samples/drivers/opamp/output_measure/src/main.c (hunk 15 above)
Comment (KozhinovAlexander) [reply samples/drivers/opamp/output_measure/src/main.c:130]: I see it as a cosmetic change, because style-checker does not throw any error. We could really make our reviews faster, if we avoid cosmetic changes 😉
UPD: I will change.

Comment (KozhinovAlexander) [reply samples/drivers/opamp/output_measure/src/main.c:130]: done

**Comment by KozhinovAlexander** [comment]
@etienne-lms Pleae take a look at your change-requests and resolve if they're modified correctly.

File: samples/drivers/opamp/output_measure/src/main.c [hunk 16]
```diff
@@ -24,31 +24,52 @@ uint8_t size = ARRAY_SIZE(gain);
 uint8_t size;
 #endif
 
+const static char *opamp_gain2str[] = {
```
etienne-lms, line 27: Could condition with `#if OPAMP_SUPPORT_PROGRAMMABLE_GAIN`.

etienne-lms (approved):
Non-blocking comment.

samples/drivers/opamp/output_measure/src/main.c
KozhinovAlexander: This change makes less sense for sample code, because samples are not necessarily expected to be optimized for memory footprint or runtime. The only advantage of your change request is that it would remove a lookup table containing 18 32-bit values (72 bytes) in case programmable gain is not supported. The table with 18 values is here: https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/opamp.h#L34-L53

KozhinovAlexander: Therefore I would prefer to drop this one.

ZhaoxiangJin (approved):

samples/drivers/opamp/output_measure/src/main.c
ZhaoxiangJin: move to #if OPAMP_SUPPORT_PROGRAMMABLE_GAIN condition is better, but up to you.

dts/bindings/opamp/opamp-controller.yaml
kartben: you'll want to document this change in the migration guide

PR #98881, opened 2025-11-04
pmci: mctp: Fix MCTP USB Packet Length Issue

gtossk2:
According to DSP0283 1.0.1, `Length` byte in MCTP over USB header fields should start from the `MCTP over USB Header` to the last byte in the `MCTP packet payload`.


eric-ocasio-nxp (approved):
Good catch!

teburd (approved):

nashif (approved):

PR #98889, opened 2025-11-05
Minor fix-up for NET timestamping

yangbolu1991:
Minor fix-up for NET timestamping.
Thanks.

### DIFF
--- a/include/zephyr/net/net_if.h
+++ b/include/zephyr/net/net_if.h
@@ -3151,12 +3151,12 @@ static inline bool net_if_is_dormant(struct net_if *iface)
 	return net_if_flag_is_set(iface, NET_IF_DORMANT);
 }
 
-#if defined(CONFIG_NET_PKT_TIMESTAMP) && defined(CONFIG_NET_NATIVE)
+#if defined(CONFIG_NET_PKT_TIMESTAMP_THREAD) || defined(__DOXYGEN__)
 /**
  * @typedef net_if_timestamp_callback_t
  * @brief Define callback that is called after a network packet
  *        has been timestamped.
- * @param "struct net_pkt *pkt" A pointer on a struct net_pkt which has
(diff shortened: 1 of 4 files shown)


include/zephyr/net/net_if.h
```diff
@@ -3151,7 +3151,7 @@ static inline bool net_if_is_dormant(struct net_if *iface)
 	return net_if_flag_is_set(iface, NET_IF_DORMANT);
 }
 
-#if defined(CONFIG_NET_PKT_TIMESTAMP) && defined(CONFIG_NET_NATIVE)
+#if defined(CONFIG_NET_PKT_TIMESTAMP_THREAD)
```
jukkar: Could you elaborate what is the issue here, the commit message is not explaining it.

yangbolu1991: These functions are implemented under `CONFIG_NET_PKT_TIMESTAMP_THREAD`, but are clarified under `CONFIG_NET_PKT_TIMESTAMP`.
https://github.com/zephyrproject-rtos/zephyr/blob/v4.3.0-rc2/subsys/net/ip/net_if.c#L6173

Header file here may misunderstand users, because enabling CONFIG_NET_PKT_TIMESTAMP is not enough to use these functions.

Thanks.

jukkar: Ok, thanks. Could you update also the commit message and add your clarifications there. Please omit the link as that is not needed in commit msg.

maass-hamburg: if you are already on it add `|| defined(__DOXYGEN__)`, so it's also shown in the doocs

yangbolu1991:
Added one fix-up for doc to address CI issue.
Thanks.

rlubos (approved):

jukkar (approved):

PR #98890, opened 2025-11-05
llext: option to exclude all default exported symbols

JordanYates:
Introduce a Kconfig option to exclude all symbols declared with `EXPORT_SYMBOL` or `EXPORT_SYMBOL_NAMED` from the application build. It is instead the applications responsibility to export exactly the symbols that it wants available for linking, done through two new macros, `EXPORT_APP_SYMBOL` and `EXPORT_APP_SYMBOL_NAMED`.

The reasoning behind this option is to reduce the size of the exported symbol array and speed up link times.
For one of my standard applications, a nRF54L15 connected to a LTE modem and running Bluetooth, enabling LLEXT increases the text size of the application from `548124` to `561828` bytes, over 13kB, from ~260 exported symbols (without `CONFIG_LLEXT_EXPORT_DEVICES`). This is obviously undesirable when it is known at development time that the extensions are going to be restricted to linking against a handful of symbols.

The names I've chosen (`LLEXT_EXPORT_DEFAULT_SYMBOLS`, `EXPORT_APP_SYMBOL`) probably aren't ideal, but the best I could come up with.

### DIFF
--- a/doc/services/llext/config.rst
+++ b/doc/services/llext/config.rst
@@ -166,6 +166,18 @@ significant as the number of exported symbols increases.
            forbidden to load an extension that was compiled with
            ``CONFIG_LLEXT_EXPORT_BUILTINS_BY_SLID=n``.
 
+Minimizing exporting symbols
+----------------------------
+
+To minimize the size of the exported symbol table, the main application can
+disable :kconfig:option:`CONFIG_LLEXT_EXPORT_DEFAULT_SYMBOLS`. This option
+causes :c:macro:`EXPORT_SYMBOL` and :c:macro:`EXPORT_SYMBOL_NAMED` to not
+export a symbol, which forces the symbol table to be empty by default.
+
+The main application is then responsible for explicitly exporting the symbols
+that are required by the extension with :c:macro:`EXPORT_APP_SYMBOL` and
+:c:macro:`EXPORT_APP_SYMBOL_NAMED`.
+
 EDK configuration
 -----------------
 
--- a/include/zephyr/llext/symbol.h
+++ b/include/zephyr/llext/symbol.h
@@ -161,13 +161,18 @@ struct llext_symtable {
  * Version of @ref EXPORT_SYMBOL that allows the user to specify a custom name
  * for the exported symbol.
  *
- * When @c CONFIG_LLEXT is not enabled, this macro is a no-op.
+ * When @c CONFIG_LLEXT or @c CONFIG_LLEXT_EXPORT_DEFAULT_SYMBOLS are not
+ * enabled, this macro is a no-op.
  *
  * @param sym_ident Symbol to export
  * @param sym_name Name associated with the symbol
  */
+#if defined(CONFIG_LLEXT_EXPORT_DEFAULT_SYMBOLS) || defined(LL_EXTENSION_BUILD)
 #define EXPORT_SYMBOL_NAMED(sym_ident, sym_name)				\
 	Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)
+#else
+#define EXPORT_SYMBOL_NAMED(sym_ident, sym_name)
+#endif
 
 /**
  * @brief Export a constant symbol from the current build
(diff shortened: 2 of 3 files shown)


teburd:
> Introduce a Kconfig option to exclude all symbols declared with `EXPORT_SYMBOL` or `EXPORT_SYMBOL_NAMED` from the application build. It is instead the applications responsibility to export exactly the symbols that it wants available for linking, done through two new macros, `EXPORT_APP_SYMBOL` and `EXPORT_APP_SYMBOL_NAMED`.
> 
> The reasoning behind this option is to reduce the size of the exported symbol array and speed up link times. For one of my standard applications, a nRF54L15 connected to a LTE modem and running Bluetooth, enabling LLEXT increases the text size of the application from `548124` to `561828` bytes, over 13kB, from ~260 exported symbols (without `CONFIG_LLEXT_EXPORT_DEVICES`). This is obviously undesirable when it is known at development time that the extensions are going to be restricted to linking against a handful of symbols.
> 
> The names I've chosen (`LLEXT_EXPORT_DEFAULT_SYMBOLS`, `EXPORT_APP_SYMBOL`) probably aren't ideal, but the best I could come up with.

Really we likely want collections of symbols for certain features to be Kconfig selectable kind of tied to the directory they are in. E.g. kernel symbols, posix symbols, lib C symbols, logging symbols, etc.

Not sure we want to collectively group all symbols Zephyr has like this. You could still always disable them all and export *only* what you want, but you'll likely find that you need a large amount of the lib C and logging symbols already at a minimum.

I'd accept a patch that does the above. I'm not willing to accept a patch that groups all symbols in Zephyr like this though I get the desire here.

On another note did you try using the hashed symbol table approach @mathieuchopstm created with SLID? This removes the string symbols and replaces them all with numeric IDs.

teburd:
There aren't actually that many direct exports in the tree, many come from the gen_syscalls.py script which should help automate most of this grouping perhaps. If you create a Kconfig to disable it entirely that would also be fine. It's aways done currently in Zephyr's CMakeLists.txt (grep for --syscall-exports-llext).

mathieuchopstm:
> Really we likely want collections of symbols for certain features to be Kconfig selectable kind of tied to the directory they are in. E.g. kernel symbols, posix symbols, lib C symbols, logging symbols, etc.
> 
> Not sure we want to collectively group all symbols Zephyr has like this. You could still always disable them all and export _only_ what you want, but you'll likely find that you need a large amount of the lib C and logging symbols already at a minimum.
> 
> I'd accept a patch that does the above. I'm not willing to accept a patch that groups all symbols in Zephyr like this though I get the desire here.

\+1 on this. I guess it could be something conceptually like:
```c
#define EXPORT_SYMBOL_IN_GROUP(group, sym) \
        IF_ENABLED(CONFIG_LLEXT_EXPORT_GROUP_ ##group, \
            EXPORT_SYMBOL(sym))
```
(Or rather, the opposite, `#define EXPORT_SYMBOL(sym) EXPORT_SYMBOL_IN_GROUP(DEFAULT, sym)`)

For the record, if an approach like this[1] is chosen, the "excluded" entries should not *actually* be discarded; instead, they should be placed in a dummy section which exists in the ELF but is not placed in any segment, and thus doesn't occupy space in actual flashed images (c.f. how SLID discard symbol names for how this can be done). This would allow scripts to process the ELF and display which groups are excluded, and which symbols they contain, which would make debugging "why is symbol X not exported?!"-type issues less painful.

[1] "don't emit symbol table entry if `<...>`" - that is only one solution to the problem here :slightly_smiling_face:

> On another note did you try using the hashed symbol table approach @mathieuchopstm created with SLID? This removes the string symbols and replaces them all with numeric IDs.

I would definitely try this. It's been long since I toyed with it ~~so I hope it's not broken~~ but the savings can be significant. In my very old tests, it reduced the *overall symbol table-related footprint* by ~75%, if my notes are to be trusted and still accurate: https://gist.github.com/mathieuchopstm/e06579bf45bf22b5fd60f1fe3734fd0d

JordanYates:
> Really we likely want collections of symbols for certain features to be Kconfig selectable ... you could still always disable them all and export only what you want

Disabling them all implies that there is also a way to disable the `DEFAULT` group that @mathieuchopstm suggested. The Kconfig name would obviously be different, but the end result would be practically the same. It does seem like a nice solution.

> but you'll likely find that you need a large amount of the lib C and logging symbols already at a minimum.

That depends entirely on the use case. LLEXT can still be useful with 0 exports from the base application.

> On another note did you try using the hashed symbol table approach @mathieuchopstm created with SLID?

This is not a usable solution for me. I am trying to build extensions that can run on arbitrary applications, and SLID is even more coupled to an individual application than the default mechanism (based on the required post-processing).

mathieuchopstm:
> > Really we likely want collections of symbols for certain features to be Kconfig selectable ... you could still always disable them all and export only what you want
> 
> Disabling them all implies that there is also a way to disable the `DEFAULT` group that @mathieuchopstm suggested. The Kconfig name would obviously be different, but the end result would be practically the same. It does seem like a nice solution.

FTR, my idea was that the `DEFAULT` group would only have application symbols as Zephyr itself would put all its exported symbols in groups. (exact details - groups names, granularity, etc - to be discussed but that's what sounds the most solid to me)

> > On another note did you try using the hashed symbol table approach @mathieuchopstm created with SLID?
> 
> This is not a usable solution for me. I am trying to build extensions that can run on arbitrary applications, and SLID is even more coupled to an individual application than the default mechanism (based on the required post-processing).

Can you elaborate on that? The SLID generation is deterministic - i.e. there is a 1:1(-ish*) mapping between symbol name - so I don't understand how it can create additional coupling. (Note that the determinism also implies that you can `slidgen` on an LLEXT ELF in a completely standalone fashion from the Zephyr ELF; obviously, this requires knowing the exported symbols' names, but *you should have them anyways* since how can you use exported symbols if you don't know the signature of functions you'll be calling? :slightly_smiling_face:)

(*1:1-ish because there can be collisions but (1) they are *very rare* and (2) `slidgen` will panic if it sees a collision)

PR #98891, opened 2025-11-05
zbus: helper to query channel message age

JordanYates:
Add a helper function to query how old the message in a channel is. Output is converted to milliseconds as `k_ticks` objects cannot be compared directly.

### DIFF
--- a/include/zephyr/zbus/zbus.h
+++ b/include/zephyr/zbus/zbus.h
@@ -855,6 +855,22 @@ static inline uint32_t zbus_chan_pub_stats_avg_period(const struct zbus_channel 
 	return k_uptime_get() / chan->data->publish_count;
 }
 
+/**
+ * @brief Get the age of a message in a channel
+ *
+ * @param chan The channel's reference.
+ *
+ * @retval UINT64_MAX if channel has never been published to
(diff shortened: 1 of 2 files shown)


include/zephyr/zbus/zbus.h
```diff
@@ -855,6 +855,22 @@ static inline uint32_t zbus_chan_pub_stats_avg_period(const struct zbus_channel
...
 
+/**
+ * @brief Get the age of data in a channel
+ *
+ * @param chan The channel's reference.
+ *
+ * @retval UINT64_MAX if channel has never been published to
+ * @retval age_ms Data age in milliseconds otherwise
+ */
+static inline uint64_t zbus_chan_pub_stats_data_age(const struct zbus_channel *chan)
```
rodrigopex: Thank you for the contribution. Could you please rename the function? Zbus's about channels, and we exchange messages over that. So, maybe `zbus_chan_pub_stats_msg_age` could be used.

rodrigopex (changes requested):

rodrigopex (approved):

pdgendt (approved):

PR #98894, opened 2025-11-05
drivers: crypto: sam: add Secure Hash Algorithm (SHA) support for Microchip sama7g5

TonyHan11:
Changes in this PR includes:
 - add SHA support for Microchip sama7g5 (dts, driver, tests)
 - reorder the nodes in `sama7g5.dtsi` by address ascending order


drivers/crypto/Kconfig
```diff
@@ -55,7 +55,7 @@ source "drivers/crypto/Kconfig.stm32"
 source "drivers/crypto/Kconfig.nrf_ecb"
 source "drivers/crypto/Kconfig.intel"
 source "drivers/crypto/Kconfig.npcx"
-source "drivers/crypto/Kconfig.xec"
+source "drivers/crypto/Kconfig.mchp"
```
nandojve: This commit must be the 2nd in the list.
It still a preparation for what is coming next.

dts/arm/microchip/sam/sama7g5.dtsi
```diff
@@ -180,6 +180,13 @@
 			interrupts = <GIC_SPI 70 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
 		};
 
+		sha: sha@e1814000 {
```
nandojve, line 267: You are adding an entry in the devicetree were a binding does not exists yet.
Extract the binding from ead9f9fe in a separate commit. Put that new commit before this.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,322 @@
...
+	enum hash_algo algo;
+	uint32_t sha_mr_algo;
+	uint8_t dgst_len;
+	uint8_t block_size;
+} mchp_sha_algo_cfgs[4] = {
+	{CRYPTO_HASH_ALGO_SHA224, SHA_MR_ALGO_SHA224, 224/8, 64},
+	{CRYPTO_HASH_ALGO_SHA256, SHA_MR_ALGO_SHA256, 256/8, 64},
+	{CRYPTO_HASH_ALGO_SHA384, SHA_MR_ALGO_SHA384, 384/8, 128},
+	{CRYPTO_HASH_ALGO_SHA512, SHA_MR_ALGO_SHA512, 512/8, 128},
+};
```
nandojve: Let's not mix struct definition with instantiation.
Define the instantiation separate.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,322 @@
...
+static struct k_sem mchp_sha_session_sem;
+
+static inline void mchp_sha_set_input(sha_registers_t *const regs, const unsigned char *data,
+				      int len)
+{
+	uint32_t i = 0, v;
+
+	while (len > 4) {
+		memcpy(&v, &data[i * 4], 4);
+		if (i < 16) {
```
nandojve: What is 4 and 16 ?
Let's define clear names for those to avoid thinking when reading the driver.
Fix in all driver similar entries.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,322 @@
...
+	int ret;
+
+	regs->SHA_CR = SHA_CR_SWRST_Msk;
+	regs->SHA_MR = algo_cfg->sha_mr_algo | SHA_MR_SMOD_AUTO_START | SHA_MR_PROCDLY_LONGEST;
+	regs->SHA_CR = SHA_CR_FIRST_Msk;
+	regs->SHA_MSR = len;
+	regs->SHA_BCR = len;
+
+	if (len == 0) {
+		regs->SHA_IDATAR[0] = 0x000000080;
```
nandojve: name the magic number 0x000000080

nandojve (changes requested):

nandojve (dismissed):

rruuaanng (dismissed):

ceolin (dismissed):

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,337 @@
...
+#include <zephyr/crypto/crypto.h>
+#include <zephyr/device.h>
+#include <zephyr/drivers/clock_control/mchp_sam_pmc.h>
+#include <zephyr/init.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/sys/util.h>
+
+#define MCHP_SHA_CAPS_SUPPORT (CAP_INPLACE_OPS | CAP_SEPARATE_IO_BUFS |		\
+			       CAP_SYNC_OPS | CAP_NO_IV_PREFIX)
```
valeriosetti: `CAP_NO_IV_PREFIX` is for block ciphers, not hash

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,337 @@
...
+static struct crypto_mchp_sha_session mchp_sha_sessions[2];
+static struct k_sem mchp_sha_session_sem;
+
+static inline void mchp_sha_set_input(sha_registers_t *const regs, const unsigned char *data,
+				      int len)
+{
+	uint32_t i = 0, v;
+
+	while (len > sizeof(uint32_t)) {
+		memcpy(&v, &data[i * sizeof(uint32_t)], sizeof(uint32_t));
```
valeriosetti: Perhaps the compiler is smart enough to simplify this, but isn't `memcpy` a bit overkilling to copy an `uint32_t`?
In case you change this, then there are other occurrences also on lines 76 and 91

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,337 @@
...
+
+#include <zephyr/crypto/crypto.h>
+#include <zephyr/device.h>
+#include <zephyr/drivers/clock_control/mchp_sam_pmc.h>
+#include <zephyr/init.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/atomic.h>
+#include <zephyr/sys/util.h>
+
+#define MCHP_SHA_CAPS_SUPPORT (CAP_INPLACE_OPS | CAP_SEPARATE_IO_BUFS |		\
```
valeriosetti: Since the driver reads data from `pkt->in_buf` and writes to `pkt->out_buf` I think that `CAP_INPLACE_OPS` [doesn't apply](https://github.com/zephyrproject-rtos/zephyr/blob/3daddb836f0b21bb1f2a2e9f9783c5a48c35f016/include/zephyr/crypto/hash.h#L98) in this case. But I might be wrong: @ceolin what do you think about this?

valeriosetti (changes requested):

drivers/crypto/crypto_mchp_sha_g1.c
TonyHan11: `CAP_NO_IV_PREFIX` removed.

drivers/crypto/crypto_mchp_sha_g1.c
TonyHan11: Replaced `memcpy`, thanks.

drivers/crypto/crypto_mchp_sha_g1.c
TonyHan11: Added support for `CAP_INPLACE_OPS`, the digest result will be stored to `pkt->in_buf` in this case.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,353 @@
...
+static inline void mchp_sha_set_input(sha_registers_t *const regs, const unsigned char *data,
+				      int len)
+{
+	uint32_t i = 0, v;
+
+	while (len >= sizeof(uint32_t)) {
+		v = data[i * sizeof(uint32_t) + 3];
+		v = (v << BITS_PER_BYTE) | data[i * sizeof(uint32_t) + 2];
+		v = (v << BITS_PER_BYTE) | data[i * sizeof(uint32_t) + 1];
+		v = (v << BITS_PER_BYTE) | data[i * sizeof(uint32_t) + 0];
```
valeriosetti: Here it seems to me that you are just filling an `uint32_t` with 4 `uint8_t`. I think you can get the same by declaring a new pointer like:
`uint32_t *data_u32 = (uint32_t *) data;`
at the beginning of the function and then just setting:
`v = data_u32[i]`
in the while loop below. Wdyt?
In case you need to change data endianness you can use function from `zephyr/include/zephyr/sys/byteorder.h` instead of reimplementing them.

The same holds also for `mchp_sha_get_output` below.

valeriosetti (changes requested):
Only one last minor improvement. All the rest is OK to me

valeriosetti (dismissed):

nandojve (dismissed):

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	if (len > 0) {
+		v = data[i * sizeof(uint32_t) + 0];
+		if (len > 1) {
+			v |= (uint32_t)data[i * sizeof(uint32_t) + 1] << 8;
+		}
+		if (len > 2) {
+			v |= (uint32_t)data[i * sizeof(uint32_t) + 2] << 16;
+		}
```
ceolin: couldn't you simply do:
```
v = 0;
memcpy(&v, &data[i * sizeof(uint32_t)], len);
```

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+static int mchp_sha_wait_data_rdy(sha_registers_t *const regs)
+{
+	uint32_t timeout = 1;
+
+	while ((regs->SHA_ISR & SHA_ISR_DATRDY_Msk) == 0) {
+		if (timeout-- == 0) {
+			LOG_ERR("MCHP SHA wait data ready timeout");
+			return -ETIMEDOUT;
+		}
+		k_sleep(K_USEC(5));
```
ceolin: I may be missing something but this looks a completely arbitrary value (and too small). k_sleep is quantized to ticks and the value that you will sleep is much likely way bigger than that

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+		}
+	}
+}
+
+static inline void mchp_sha_get_output(sha_registers_t *const regs, unsigned char *data, int len)
+{
+	uint32_t *data_u32 = (uint32_t *)data;
+	uint32_t i;
+
+	for (i = 0; i < (len / sizeof(uint32_t)) && i < ARRAY_SIZE(regs->SHA_IODATAR); i++) {
```
ceolin: This could be simplified for something like:
`for (i = 0; i < MIN((len / sizeof(uint32)t), ARRAY_SIZE(regs->SHA_IODATAR)); i++)`

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+	regs->SHA_BCR = len;
+
+	if (len == 0) {
+		/* For the empty message, automatic padding is not quired in this driver
+		 * (SHA_MSR.MSGSIZE and SHA_BCR.BYTCNT are configured to 0). The block to
+		 * be processed are the padded part (a one bit, then  '1' followed by zeor
+		 * bits) and then the message length.
+		 */
+		regs->SHA_IDATAR[0] = FIRST_WORD_4_PADDING_EMPTY_MSG;
+		for (int i = 1; i < block_size / sizeof(uint32_t); i++) {
```
ceolin, line 131: nitpick: most other places you are declaring the variable out of the for loop.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+		ret = mchp_sha_wait_data_rdy(regs);
+		if (ret < 0) {
+			return ret;
+		}
+
+		processed += block_size;
+	}
+
+out:
+	ret = mchp_sha_wait_data_rdy(regs);
```
ceolin, line 150: ins't this call redundant in several cases ? lines (145 and 156). AFAIU this is only needed when `len == 0`.

I am not familiar with the HAL, is it ok to call it twice without any change ?

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+		mchp_sha_set_input(regs, &data[processed], block_size);
+		ret = mchp_sha_wait_data_rdy(regs);
+		if (ret < 0) {
+			return ret;
+		}
+
+		processed += block_size;
+	}
+
+	if (len - processed) {
```
ceolin: you could make it a boolean expression (as recommended in the guideline - I think).

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+
+	if (session->algo_cfg == NULL) {
+		k_sem_give(&data->device_sem);
+		LOG_ERR("Unsupported algorithm");
+		return -ENOTSUP;
+	}
+
+	if (inplace_ops && pkt->in_len < session->algo_cfg->dgst_len) {
+		LOG_ERR("Insufficient in_buf for digest");
+		return -EINVAL;
```
ceolin, line 205: you are returning without giving the semaphore.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	ret = mchp_sha_process(cfg->regs, session->algo_cfg, pkt->in_buf, pkt->in_len,
+			       inplace_ops ? (uint8_t *)(uintptr_t)pkt->in_buf : pkt->out_buf);
```
ceolin, line 209: why are you casting `in_buf` to `uintptr_t` to (uint8_t *) ? It does not seem necessary this cast to `uintptr_t`

drivers/crypto/crypto_mchp_sha_g1.c
ceolin: you can simplify these multiple exit points just setting the error code to `ret` and jumping to end of the function before giving the semaphore.

drivers/crypto/crypto_mchp_sha_g1.c
```diff
@@ -0,0 +1,348 @@
...
+		return -ENOTSUP;
+	}
+
+	session = crypto_mchp_sha_get_unused_session();
+	if (session == NULL) {
+		LOG_ERR("No free session for now");
+		return -ENOSPC;
+	}
+	session->algo_cfg = algo_cfg;
+	session->in_use = true;
```
ceolin: already set in `crypto_mchp_sha_get_unused_session` with proper protection.

ceolin (changes requested):
Please take a look in the comments. The most important is about the semaphore usage in `mchp_sha_handler` and redundant call in `mchp_sha_process`. Other comments are minor things.

drivers/crypto/crypto_mchp_sha_g1.c
ceolin: I saw the comment about memcpy now. I think it is clear / simpler this way that multiple branches but it is your call which one you prefer.

drivers/crypto/crypto_mchp_sha_g1.c
TonyHan11: As the type of `pkt->in_buf` is `const uint8_t *`,  adding `uintptr_t` here for avoiding the `const` cast warnings.

TonyHan11:
Rebased by solving `soc/microchip/sam/sama7g5/soc.c` conflicts (renamed to `soc/microchip/sam/sama7/sama7g5/soc.c` in 0a9d1493a80).

drivers/crypto/crypto_mchp_sha_g1.c
ceolin: couldn't you simply `(uint8_t *)pkt-in_buf` ?

ceolin (approved):

nandojve (approved):

valeriosetti (approved):

PR #98895, opened 2025-11-05
Bluetooth: Classic: HFP: Improve codec configuration

lylezhu2012:
* Include: Bluetooth: Add HCI voice_setting definitions

Add macro `BT_HCI_VOICE_SETTINGS` to assemble the voice_setting

Add macro `BT_HCI_VOICE_SETTING_*_PREP` to set the fields for voice_setting.

Add macro `BT_HCI_VOICE_SETTING_*_GET` to get the value of the field from the voice_setting.

* Bluetooth: Classic: SCO: Set voice setting by upper layer

In current implementation, the voice setting of SCO is fixed value `BT_VOICE_CVSD_16BIT`. It is not aligned with the negotiated codec id of the HFP.

Add a filed `voice_setting` to `struct bt_sco_chan` that the voice setting can be configured by upper layer when providing the sco channel `struct bt_sco_chan`.

Use the `voice_setting` to setup SCO or accept SCO.

* Bluetooth: Classic: HFP_HF: Configure voice setting

Add a filed `active_codec_id` to `struct bt_hfp_hf` to keep the current codec ID. The default value is `BT_HFP_HF_CODEC_CVSD` when creating the SLC.

Add a filed `neg_codec_id` to `struct bt_hfp_hf` to flag the negotiated codec ID. When the AT+BCS command is executed without error, update the `active_codec_id` with `neg_codec_id`.

Pass the properly voice setting with the `active_codec_id` in function `bt_hfp_hf_sco_accept()`.

* Bluetooth: Classic: HFP_AG: Configure voice setting

Set the voice setting according to the selected codec ID.

* Bluetooth: Classic: SCO: Set connect param with default value

In current implementation, the values are fixed for the field `Max_Latency` and `Retransmission_Effort` of HCI command
`HCI_Setup_Synchronous_Connection` and `HCI_Accept_Synchronous_Connection_Request`

It cannot work properly with these fixed value. It is better to set it with default value. And it is determined by the controller.

Set the field `Max_Latency` and `Retransmission_Effort` of HCI command `HCI_Setup_Synchronous_Connection` and `HCI_Accept_Synchronous_Connection_Request` with default value.

* Bluetooth: Classic: SCO: Fix byte order issue

Change the byte order from CPU to LE16/LE32 for `accept SCO` and `setup SCO` HCI commands.


subsys/bluetooth/host/classic/hfp_ag.c
```diff
@@ -1475,6 +1475,33 @@ static void hfp_ag_sco_disconnected(struct bt_sco_chan *chan, uint8_t reason)
...
+		break;
+#if defined(CONFIG_BT_HFP_AG_CODEC_NEG)
+	case BT_HFP_AG_CODEC_MSBC:
+	case BT_HFP_AG_CODEC_LC3_SWB:
+		air_coding_fmt = BT_HCI_VOICE_SETTING_AIR_CODING_FMT_TRANSPARENT;
+		break;
+#endif /* CONFIG_BT_HFP_AG_CODEC_NEG */
+	default:
+		LOG_ERR("Unsupported codec ID %u", ag->selected_codec_id);
+		return -EINVAL;
```
makeshi, line 1494: -ENOTSUP？

lylezhu2012: I think it is a invalid parameter. So return error `-EINVAL`.

makeshi (approved):

MarkWangChinese (approved):

PR #98901, opened 2025-11-05
Bluetooth: Host: Add Shorter Connection Intervals

weeTike:
* **bluetooth: host: Add Shorter Connection Intervals support**
    This commit adds support for the Shorter Connection Intervals feature to
    the Bluetooth host.

* **bluetooth: host: Add compatibility functions for conn interval with SCI**
    Since Shorter Connection Intervals changes the unit that connection
    intervals can be represented in. It is therefore necessary to change how
    they are stored and represented.

* **bluetooth: host: shell: Add SCI shell commands**
    This adds support for Shorter Connection Interval commands in the bt
    shell.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -48,6 +48,9 @@ Bluetooth Host
 
 * :kconfig:option:`CONFIG_BT_SIGNING` has been deprecated.
 * :c:macro:`BT_GATT_CHRC_AUTH` has been deprecated.
+* :c:member:`bt_conn_le_info.interval` has been deprecated. Use
+  :c:member:`bt_conn_le_info.interval_us` instead. Note that the units have changed: ``interval``
+  was in units of 1.25 milliseconds, while ``interval_us`` is in microseconds.
 
 Networking
 **********
--- a/include/zephyr/bluetooth/conn.h
+++ b/include/zephyr/bluetooth/conn.h
@@ -1056,9 +1056,19 @@ struct bt_conn_le_info {
 	/** Connection interval in microseconds */
 	uint32_t interval_us;
 #if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
-	/** Connection interval in units of 1.25 ms */
-	uint16_t interval;
+	union {
+		/** @brief Connection interval in units of 1.25 ms
+		 *
+		 * @deprecated Use @ref bt_conn_le_info.interval_us instead
+		 */
+		__deprecated uint16_t interval;
+		/** @cond INTERNAL_HIDDEN */
+		/** Workaround for setting deprecated @ref bt_conn_le_info.interval */
+		uint16_t _interval;
+		/** @endcond */
+	};
 #endif /* !CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
+
 	uint16_t latency; /**< Connection peripheral latency */
 	uint16_t timeout; /**< Connection supervision timeout */
 
@@ -1096,7 +1106,7 @@ struct bt_conn_le_info {
  *
  *  Multiply by 125 to get microseconds.
  */
-#define BT_CONN_SCI_INTERVAL_TO_US(interval) ((interval) * BT_HCI_LE_SCI_INTERVAL_UNIT_US)
+#define BT_CONN_SCI_INTERVAL_TO_US(_interval) ((_interval) * BT_HCI_LE_SCI_INTERVAL_UNIT_US)
 
 /** BR/EDR Connection Info Structure */
 struct bt_conn_br_info {
--- a/include/zephyr/bluetooth/conn.h
+++ b/include/zephyr/bluetooth/conn.h
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
 	uint16_t supervision_timeout;
 };
 
+/** @brief Maximum Connection Interval Groups possible
+ *
+ * The practical maximum is 41 groups based on HCI event size constraints:
+ *     (HCI event max size - cmd_complete overhead - response fields) / sizeof(group)
+ *     (255 - 4 - 3) / 6 = 41 groups max
+ */
+#define BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS 41
+
+/** @brief Minimum supported connection interval group
+ *
+ * Each group represents an arithmetic sequence of supported connection intervals:
+ *   min, min + stride, min + 2 * stride, ..., min + n * stride (where min + n * stride <= max)
+ *
+ * Example: min = 10 (1250 us), max = 60 (7500 us), stride = 5 (625 us)
+ *   -> Supported: 1250 us, 1875 us, 2500 us, ..., 6875 us, 7500 us
+ */
+struct bt_conn_le_min_conn_interval_group {
+	/** @brief Lower bound of group interval range
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t min_125us;
+	/** @brief Upper bound of group interval range
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t max_125us;
+	/** @brief Increment between consecutive supported intervals
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_STRIDE_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t stride_125us;
+};
+
+/** Minimum supported connection interval information */
+struct bt_conn_le_min_conn_interval_info {
+	/** @brief Minimum supported connection interval
+	 *
+	 * Unit: microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_MIN_SUPP_CONN_INT_MIN_US - @ref BT_HCI_LE_MIN_SUPP_CONN_INT_MAX_US
+	 */
+	uint16_t min_supported_conn_interval_us;
+	/** @brief Number of interval groups.
+	 *
+	 * Range: 0 - @ref BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS
+	 *
+	 * If zero, the controller only supports Rounded ConnInterval Values (RCV).
+	 */
+	uint8_t num_groups;
+	/** @brief Array of supported connection interval groups.
+	 *
+	 * Multiple groups allow representing non-contiguous or differently-strided ranges.
+	 */
+	struct bt_conn_le_min_conn_interval_group groups[BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS];
+};
+
+/** Connection rate parameters for LE connections */
+struct bt_conn_le_conn_rate_param {
+	/** @brief Minimum connection interval
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t interval_min_125us;
+	/** @brief Maximum connection interval
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t interval_max_125us;
+	/** @brief Minimum subrate factor
+	 *
+	 * Range: @ref BT_HCI_LE_SUBRATE_FACTOR_MIN - @ref BT_HCI_LE_SUBRATE_FACTOR_MAX
+	 */
+	uint16_t subrate_min;
+	/** @brief Maximum subrate factor
+	 *
+	 * Range: @ref BT_HCI_LE_SUBRATE_FACTOR_MIN - @ref BT_HCI_LE_SUBRATE_FACTOR_MAX
+	 */
+	uint16_t subrate_max;
+	/** @brief Maximum Peripheral latency
+	 *
+	 * Unit: subrated connection intervals @ref bt_conn_le_conn_rate_changed.subrate_factor
+	 *
+	 * Range: @ref BT_HCI_LE_PERIPHERAL_LATENCY_MIN - @ref BT_HCI_LE_PERIPHERAL_LATENCY_MAX
+	 */
+	uint16_t max_latency;
+	/** @brief Minimum number of underlying connection events to remain active
+	 *  after a packet containing a Link Layer PDU with a non-zero Length
+	 *  field is sent or received.
+	 *
+	 * Range: @ref BT_HCI_LE_CONTINUATION_NUM_MIN - @ref BT_HCI_LE_CONTINUATION_NUM_MAX
+	 */
+	uint16_t continuation_number;
+	/** @brief Connection Supervision timeout
+	 *
+	 * Unit: 10 milliseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SUPERVISON_TIMEOUT_MIN - @ref BT_HCI_LE_SUPERVISON_TIMEOUT_MAX
+	 */
+	uint16_t supervision_timeout_10ms;
+	/** @brief Minimum length of connection event
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_CE_LEN_MIN_125US - @ref BT_HCI_LE_SCI_CE_LEN_MAX_125US
+	 */
+	uint16_t min_ce_len_125us;
+	/** @brief Maximum length of connection event
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_CE_LEN_MIN_125US - @ref BT_HCI_LE_SCI_CE_LEN_MAX_125US
+	 */
+	uint16_t max_ce_len_125us;
+};
+
+/** Updated connection rate parameters */
+struct bt_conn_le_conn_rate_changed {
+	/** Connection interval
+	 *
+	 * Unit: microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_US
+	 */
+	uint32_t interval_us;
+	/** Connection subrate factor
+	 *
+	 * Range: @ref BT_HCI_LE_SUBRATE_FACTOR_MIN - @ref BT_HCI_LE_SUBRATE_FACTOR_MAX
+	 */
+	uint16_t subrate_factor;
+	/** Peripheral latency
+	 *
+	 * Unit: subrated connection intervals @ref bt_conn_le_conn_rate_changed.subrate_factor
+	 *
+	 * Range: @ref BT_HCI_LE_PERIPHERAL_LATENCY_MIN - @ref BT_HCI_LE_PERIPHERAL_LATENCY_MAX
+	 */
+	uint16_t peripheral_latency;
+	/** Number of underlying connection events to remain active after
+	 *  a packet containing a Link Layer PDU with a non-zero Length
+	 *  field is sent or received.
+	 *
+	 * Range: @ref BT_HCI_LE_CONTINUATION_NUM_MIN - @ref BT_HCI_LE_CONTINUATION_NUM_MAX
+	 */
+	uint16_t continuation_number;
+	/** Connection Supervision timeout
+	 *
+	 * Unit: 10 milliseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SUPERVISON_TIMEOUT_MIN - @ref BT_HCI_LE_SUPERVISON_TIMEOUT_MAX
+	 */
+	uint16_t supervision_timeout_10ms;
+};
+
 /** Read all remote features complete callback params */
 struct bt_conn_le_read_all_remote_feat_complete {
 	/** @brief  HCI Status from LE Read All Remote Features Complete event.
@@ -890,7 +1053,12 @@ struct bt_conn_le_info {
 	const bt_addr_le_t *local;
 	/** Remote device address used during connection setup. */
 	const bt_addr_le_t *remote;
-	uint16_t interval; /**< Connection interval */
+	/** Connection interval in microseconds */
+	uint32_t interval_us;
+#if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
+	/** Connection interval in units of 1.25 ms */
+	uint16_t interval;
+#endif /* !CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
 	uint16_t latency; /**< Connection peripheral latency */
 	uint16_t timeout; /**< Connection supervision timeout */
 
@@ -923,6 +1091,12 @@ struct bt_conn_le_info {
  *  Multiply by 1250 to get microseconds.
  */
 #define BT_CONN_INTERVAL_TO_US(interval) ((interval) * 1250U)
+
+/** @brief Convert shorter connection interval to microseconds
+ *
+ *  Multiply by 125 to get microseconds.
+ */
+#define BT_CONN_SCI_INTERVAL_TO_US(interval) ((interval) * BT_HCI_LE_SCI_INTERVAL_UNIT_US)
 
 /** BR/EDR Connection Info Structure */
 struct bt_conn_br_info {
@@ -1328,6 +1502,74 @@ int bt_conn_le_subrate_set_defaults(const struct bt_conn_le_subrate_param *param
  */
 int bt_conn_le_subrate_request(struct bt_conn *conn,
 			       const struct bt_conn_le_subrate_param *params);
+
+/** @brief Read Minimum Supported Connection Interval Groups.
+ *
+ *  Read the minimum supported connection interval and supported interval
+ *  groups from the local controller. This information describes what the
+ *  local controller supports.
+ *
+ *  @sa bt_conn_le_read_min_conn_interval if only
+ *      @ref bt_conn_le_min_conn_interval_info.min_supported_conn_interval_us
+ *      is needed.
+ *
+ *  @kconfig_dep{CONFIG_BT_SHORTER_CONNECTION_INTERVALS}
+ *
+ *  @param info Pointer to structure to receive the minimum connection interval
+ *              information.
+ *
+ *  @return Zero on success or (negative) error code on failure.
+ */
+int bt_conn_le_read_min_conn_interval_groups(struct bt_conn_le_min_conn_interval_info *info);
+
+/** @brief Read Minimum Supported Connection Interval.
+ *
+ *  Read the minimum supported connection interval from the local controller.
+ *
+ *  @sa bt_conn_le_read_min_conn_interval_groups if groups are needed.
+ *
+ *  @kconfig_dep{CONFIG_BT_SHORTER_CONNECTION_INTERVALS}
+ *
+ *  @param min_interval_us Pointer to variable to receive the minimum connection interval
+ *                         in microseconds.
+ *
+ *  @return Zero on success or (negative) error code on failure.
+ */
+int bt_conn_le_read_min_conn_interval(uint16_t *min_interval_us);
+
+/** @brief Set Default Connection Rate Parameters.
+ *
+ *  Set default connection rate parameters to be used for future connections.
+ *  This command does not affect any existing connection.
+ *  Parameters set for specific connection will always have precedence.
+ *
+ *  @kconfig_dep{CONFIG_BT_SHORTER_CONNECTION_INTERVALS}
+ *
+ *  @param params Connection rate parameters.
+ *
+ *  @return Zero on success or (negative) error code on failure.
+ */
+int bt_conn_le_conn_rate_set_defaults(const struct bt_conn_le_conn_rate_param *params);
+
+/** @brief Request New Connection Rate Parameters.
+ *
+ *  Request a change to the connection parameters of a connection. This includes
+ *  Subrate parameters. This allows changing the connection interval to below the
+ *  Baseline ConnInterval Values (BCV) and with finer granularity, if supported.
+ *
+ *  Valid intervals of the local and peer controller should be known.
+ *  See @ref bt_conn_le_read_min_conn_interval_groups
+ *
+ *  @kconfig_dep{CONFIG_BT_SHORTER_CONNECTION_INTERVALS}
+ *
+ *  @param conn   @ref BT_CONN_TYPE_LE connection object.
+ *  @param params Connection rate parameters.
+ *
+ *  @return Zero on success or (negative) error code on failure.
+ *  @return -EINVAL @p conn is not a valid @ref BT_CONN_TYPE_LE connection.
+ */
+int bt_conn_le_conn_rate_request(struct bt_conn *conn,
+				 const struct bt_conn_le_conn_rate_param *params);
 
 /** @brief Read remote feature pages.
  *
@@ -2063,6 +2305,31 @@ struct bt_conn_cb {
 	void (*subrate_changed)(struct bt_conn *conn,
 				const struct bt_conn_le_subrate_changed *params);
 #endif /* CONFIG_BT_SUBRATING */
+
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
+	/** @brief LE Connection Rate Changed event.
+	 *
+	 *  This callback notifies the application that the connection rate
+	 *  parameters (including both connection interval and subrating)
+	 *  of the connection may have changed.
+	 *
+	 *  @param conn   Connection object.
+	 *  @param status HCI Status from LE Connection Rate Change event.
+	 *                Possible Status codes:
+	 *                - Success (0x00)
+	 *                - Unknown Connection Identifier (0x02)
+	 *                - Command Disallowed (0x0C)
+	 *                - Unsupported Feature or Parameter Value (0x11)
+	 *                - Invalid HCI Command Parameters (0x12)
+	 *                - Unsupported Remote Feature (0x1A)
+	 *                - Unsupported LL Parameter Value (0x20)
+	 *  @param params New connection rate parameters.
+	 *                The connection rate parameters will be NULL
+	 *                if @p status is not @ref BT_HCI_ERR_SUCCESS.
+	 */
+	void (*conn_rate_changed)(struct bt_conn *conn, uint8_t status,
+				  const struct bt_conn_le_conn_rate_changed *params);
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
 
 #if defined(CONFIG_BT_LE_EXTENDED_FEAT_SET)
 	/** @brief Read all remote features complete event.
--- a/subsys/bluetooth/audio/bap_broadcast_assistant.c
+++ b/subsys/bluetooth/audio/bap_broadcast_assistant.c
@@ -453,12 +453,12 @@ static void long_bap_read(struct bt_conn *conn, uint16_t handle)
 		if (err != 0) {
 			LOG_DBG("Failed to get conn info, use default interval");
 
-			conn_info.le.interval = BT_GAP_INIT_CONN_INT_MIN;
+			conn_info.le.interval_us = BT_CONN_INTERVAL_TO_US(BT_GAP_INIT_CONN_INT_MIN);
 		}
 
 		/* Wait a connection interval to retry */
 		err = k_work_reschedule(&inst->bap_read_work,
-					K_USEC(BT_CONN_INTERVAL_TO_US(conn_info.le.interval)));
+					K_USEC(conn_info.le.interval_us));
 		if (err < 0) {
 			LOG_DBG("Failed to reschedule read work: %d", err);
 			bap_long_read_reset(inst);
--- a/subsys/bluetooth/audio/bap_unicast_client.c
+++ b/subsys/bluetooth/audio/bap_unicast_client.c
@@ -1767,12 +1767,12 @@ static void long_ase_read(struct bt_bap_unicast_client_ep *client_ep)
 		if (err != 0) {
 			LOG_DBG("Failed to get conn info, use default interval");
 
-			conn_info.le.interval = BT_GAP_INIT_CONN_INT_MIN;
+			conn_info.le.interval_us = BT_CONN_INTERVAL_TO_US(BT_GAP_INIT_CONN_INT_MIN);
 		}
 
 		/* Wait a connection interval to retry */
 		err = k_work_reschedule(&client_ep->ase_read_work,
-					K_USEC(BT_CONN_INTERVAL_TO_US(conn_info.le.interval)));
+					K_USEC(conn_info.le.interval_us));
 		if (err < 0) {
 			LOG_DBG("Failed to reschedule ASE long read work: %d", err);
 		}
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -133,6 +133,9 @@ config BT_CTLR_CHANNEL_SOUNDING_SUPPORT
 config BT_CTLR_EXTENDED_FEAT_SET_SUPPORT
 	bool
 
+config BT_CTLR_SHORTER_CONNECTION_INTERVALS_SUPPORT
+	bool
+
 # Virtual option that all local LL implementations should select
 config HAS_BT_CTLR
 	bool
@@ -1183,6 +1186,17 @@ config BT_CTLR_EXTENDED_FEAT_SET
 	help
 	  Enable support for Bluetooth 6.0 LL Extended Feature Set
 	  in the Controller.
+
+config BT_CTLR_SHORTER_CONNECTION_INTERVALS
+	bool "Shorter Connection Intervals"
+	depends on BT_CTLR_SHORTER_CONNECTION_INTERVALS_SUPPORT
+	depends on BT_CTLR_EXTENDED_FEAT_SET
+	depends on BT_CTLR_SUBRATING
+	select BT_CTLR_SET_HOST_FEATURE
+	default y if BT_SHORTER_CONNECTION_INTERVALS
+	help
+	  Enable support for Bluetooth 6.2 Shorter Connection Intervals
+	  in the controller.
 
 rsource "Kconfig.df"
 rsource "Kconfig.ll_sw_split"
--- a/subsys/bluetooth/host/conn.c
+++ b/subsys/bluetooth/host/conn.c
@@ -2886,7 +2886,10 @@ int bt_conn_get_info(const struct bt_conn *conn, struct bt_conn_info *info)
 			info->le.local = &conn->le.resp_addr;
 			info->le.remote = &conn->le.init_addr;
 		}
-		info->le.interval = conn->le.interval;
+		info->le.interval_us = conn->le.interval_us;
+#if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+		info->le.interval = conn->le.interval_us / BT_HCI_LE_INTERVAL_UNIT_US;
+#endif
 		info->le.latency = conn->le.latency;
 		info->le.timeout = conn->le.timeout;
 #if defined(CONFIG_BT_USER_PHY_UPDATE)
@@ -3244,6 +3247,24 @@ void bt_conn_notify_subrate_change(struct bt_conn *conn,
 	}
 }
 
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+void bt_conn_notify_conn_rate_change(struct bt_conn *conn, uint8_t status,
+				     const struct bt_conn_le_conn_rate_changed *params)
+{
+	BT_CONN_CB_DYNAMIC_FOREACH(callback) {
+		if (callback->conn_rate_changed != NULL) {
+			callback->conn_rate_changed(conn, status, params);
+		}
+	}
+
+	STRUCT_SECTION_FOREACH(bt_conn_cb, cb) {
+		if (cb->conn_rate_changed != NULL) {
+			cb->conn_rate_changed(conn, status, params);
+		}
+	}
+}
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
+
 static bool le_subrate_common_params_valid(const struct bt_conn_le_subrate_param *param)
 {
 	/* All limits according to BT Core spec 5.4 [Vol 4, Part E, 7.8.123] */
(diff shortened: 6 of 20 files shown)


subsys/bluetooth/host/hci_core.c
```diff
@@ -2807,6 +2809,64 @@ void bt_hci_le_subrate_change_event(struct net_buf *buf)
...
+
+	conn = bt_conn_lookup_handle(sys_le16_to_cpu(evt->handle), BT_CONN_TYPE_LE);
+	if (!conn) {
+		LOG_ERR("Unknown conn handle 0x%04X for connection rate event",
+			sys_le16_to_cpu(evt->handle));
+		return;
+	}
+
+	if (evt->status == BT_HCI_ERR_SUCCESS) {
+		conn->le.interval_us = sys_le16_to_cpu(evt->conn_interval) * 125;
```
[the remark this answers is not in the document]
weeTike: This comment is not correct.

include/zephyr/bluetooth/conn.h
```diff
@@ -1329,6 +1451,67 @@ int bt_conn_le_subrate_set_defaults(const struct bt_conn_le_subrate_param *param
...
+ *  local controller supports.
+ *
+ *  @note To use this API @kconfig{CONFIG_BT_SHORTER_CONNECTION_INTERVALS} must be set.
+ *
+ *  @param info Pointer to structure to receive the minimum connection interval
+ *              information.
+ *
+ *  @return Zero on success or (negative) error code on failure.
+ */
+int bt_conn_le_read_min_conn_interval_groups(struct bt_conn_le_min_conn_interval_info *info);
```
Balaklaka: I think "action" should be at the end of the function call, <bt_conn_le_read_min_conn_interval_groups> should be  <bt_conn_le_min_conn_interval_groups_read> then.

include/zephyr/bluetooth/conn.h
```diff
@@ -1180,6 +1273,35 @@ enum bt_conn_auth_keypress {
 	BT_CONN_AUTH_KEYPRESS_ENTRY_COMPLETED = 0x04,
 };
 
+/** @brief Get the connection interval in microseconds
+ *
+ * @param conn_info pointer to a @ref bt_conn_info object
+ * @return The connection interval in microseconds
+ */
+static inline uint32_t bt_conn_le_get_interval_us(const struct bt_conn_info *conn_info)
```
Balaklaka: I think "action" should be at the end of the function call, but I see that there has not been consistence in this. There are more of this.

```suggestion
static inline uint32_t bt_conn_le_interval_us_get(const struct bt_conn_info *conn_info)
```

subsys/bluetooth/host/conn_internal.h
```diff
@@ -107,7 +107,11 @@ struct bt_conn_le {
 	bt_addr_le_t init_addr;
 	bt_addr_le_t resp_addr;
 
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+	uint32_t interval_us;
```
Balaklaka, line 111: Do we really need both interval_us and interval in conn_internal.h, could we just have interval_us instead.

subsys/bluetooth/host/conn_internal.h
```diff
@@ -405,6 +409,35 @@ static inline bool bt_conn_is_sco(const struct bt_conn *conn)
 #define bt_conn_is_sco(conn) (false)
 #endif
 
+/** @brief Set the connection interval in units of 1250 microseconds
+ *
+ * @param conn pointer to a connection object
+ * @param interval Connection interval in units of 1250 microseconds
+ */
+static inline void bt_conn_internal_le_set_interval_1250_us_units(struct bt_conn *conn,
```
Balaklaka: Is this according to naming? Do not need to say it is internal when it is in a file called conn_internal.h.
```suggestion
static inline void bt_conn_interval_1250_us_units_set(struct bt_conn *conn,
```

subsys/bluetooth/host/conn_internal.h
```diff
@@ -405,6 +409,35 @@ static inline bool bt_conn_is_sco(const struct bt_conn *conn)
...
+	conn->le.interval = interval;
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
+}
+
+/** @brief Get the connection interval in units of 1250 microseconds
+ *
+ * @param conn pointer to a connection object
+ * @return The connection interval in units of 1250 microseconds
+ */
+static inline uint16_t bt_conn_internal_le_get_interval_1250_us_units(struct bt_conn *conn)
```
Balaklaka:
```suggestion
static inline uint16_t bt_conn_interval_1250_us_units_get(struct bt_conn *conn)
```

subsys/bluetooth/host/hci_core.c
Balaklaka: 125 should be a defined constant, or someone will suddenly change this to 1250, as they thought it was a typo. The comment maybe wrong, but it high lighted a problem.

subsys/bluetooth/controller/Kconfig
```diff
@@ -1184,6 +1187,15 @@ config BT_CTLR_EXTENDED_FEAT_SET
 	  Enable support for Bluetooth 6.0 LL Extended Feature Set
 	  in the Controller.
 
+config BT_CTLR_SHORTER_CONNECTION_INTERVALS
+	bool "Shorter Connection Intervals"
```
tejlmand, line 1191: Based on comments here https://github.com/nrfconnect/sdk-nrf/pull/25459#discussion_r2502482142

why is this symbol having a prompt ?
If LL'ers are expected to select this symbol then it increases the risk of stuck-symbol syndrome: https://docs.zephyrproject.org/latest/build/kconfig/tips.html#stuck-symbols-in-menuconfig-and-guiconfig

tejlmand (changes requested):

subsys/bluetooth/controller/Kconfig
tejlmand: Nevermind, ref: https://github.com/nrfconnect/sdk-nrf/pull/25459#discussion_r2502598476

tejlmand (dismissed):

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+ * Each group represents an arithmetic sequence of supported connection intervals:
+ *   min, min+stride, min+2*stride, ..., min+n*stride (where min+n*stride <= max)
+ *
+ * Example: min=10 (1250 us), max=60 (7500 us), stride=5 (625 us)
+ *   -> Supported: 1250 us, 1875 us, 2500 us, ..., 6875 us, 7500 us
+ *
+ * All values in units of 125 microseconds.
+ */
+struct bt_conn_le_min_conn_interval_group {
+	/** Lower bound of group interval range */
```
alwa-nordic: Let's put the unit in the documentation of each field. This is nice when looking up the documentation for a symbol in an IDE.

include/zephyr/bluetooth/conn.h
```diff
@@ -890,7 +977,13 @@ struct bt_conn_le_info {
...
 	/** Remote device address used during connection setup. */
 	const bt_addr_le_t *remote;
-	uint16_t interval; /**< Connection interval */
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+	/**< Connection interval in microseconds */
+	uint32_t interval_us;
+#else
+	/**< Connection interval in units of 1.25 ms */
+	uint16_t interval;
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
```
alwa-nordic: The ifdef is a backwards-compatibility issue. Existing code similar to the the audio code modified in this PR should continue working until a deprecation period has elapsed. Let's discuss this.

I propose to instead deprecate `interval` in this PR. Provisionally, `struct bt_conn_le_info` will have both `interval` and `interval_us` and both should be set. `interval` will of course contain rounded values.

I would like features enabled by kconfigs to be additive. I don't want a struct field to disappear because a feature is enabled.

Which rounding is best to use for `interval`? Up, down or nearest? The audio code is using it for timeouts, so it would make sense to round up in this case. @Emil-Gydesen-Bose Any input on this w.r.t. compatibility? The audio code will be updated to use the new field, but maybe it's representative for out-of-tree code.

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	uint8_t num_groups;
+	/**
+	 * Array of supported connection interval groups.
+	 *
+	 * Multiple groups allow representing non-contiguous or differently-strided ranges.
+	 * If num_groups=0, only intervals in the Baseline ConnInterval Values (BCV) range
+	 * are supported.
+	 */
+	struct bt_conn_le_min_conn_interval_group groups[BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS];
+};
```
alwa-nordic: This struct is 248 bytes (+1 byte alignment). Will this increase our stack size requirement?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	uint16_t min;
+	/** Upper bound of group interval range */
+	uint16_t max;
+	/** Increment between consecutive supported intervals */
+	uint16_t stride;
+};
+
+/** Minimum supported connection interval information */
+struct bt_conn_le_min_conn_interval_info {
+	/** Minimum supported connection interval in units of 125 microseconds. */
```
alwa-nordic: We should try to be consistent about units. Is there a reason this field is in `125us` which does not apply to `bt_conn_le_info.interval_us`?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	uint16_t supervision_timeout;
+	/** Minimum length of connection event (N * 125 us). */
+	uint16_t min_ce_len;
+	/** Maximum length of connection event (N * 125 us). */
+	uint16_t max_ce_len;
+};
+
+/** Updated connection rate parameters */
+struct bt_conn_le_conn_rate_changed {
+	/** HCI Status from LE Connection Rate Change event. */
```
alwa-nordic: Please list the possible status codes. (And whether the field below and valid.)

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	uint16_t subrate_factor;
+	/** Peripheral latency in units of subrated connection intervals. */
+	uint16_t peripheral_latency;
+	/** Number of underlying connection events to remain active after
+	 *  a packet containing a Link Layer PDU with a non-zero Length
+	 *  field is sent or received.
+	 */
+	uint16_t continuation_number;
+	/** Connection Supervision timeout (N * 10 ms). */
+	uint16_t supervision_timeout;
```
alwa-nordic: I liked that you put the unit in the field name for `interval_us`. I propose we do this for all new fields that have units. So this becomes `supervision_timeout_10ms`.

@jhedberg, can we work this into the code style?

include/zephyr/bluetooth/conn.h
alwa-nordic, line 1533: This struct this function takes as argument is 248 bytes (+1 byte alignment). This could make the API a unwieldly to use in constrained systems. Will every caller be interested in all the info? Maybe we should have a method for obtaining just the `min_supported_conn_interval`? And a way to read out the data in parts?

I noticed `bt_conn_le_read_min_conn_interval()`<sup><small><i>(ninjaedit)</i></small></sup> after writing the above. Can we add something like `@sa bt_conn_le_read_min_conn_interval if you only want the minimum` to the documentation?

subsys/bluetooth/audio/bap_broadcast_assistant.c
```diff
@@ -453,12 +453,12 @@ static void long_bap_read(struct bt_conn *conn, uint16_t handle)
 		if (err != 0) {
 			LOG_DBG("Failed to get conn info, use default interval");
 
-			conn_info.le.interval = BT_GAP_INIT_CONN_INT_MIN;
+			bt_conn_le_set_interval_1250_us_units(&conn_info, BT_GAP_INIT_CONN_INT_MIN);
```
alwa-nordic: I don't want this function. Bootlegging application data in `struct bt_conn_info` is not reasonably part of the API, and we don't need to have a public method to do it.

alwa-nordic (changes requested):

include/zephyr/bluetooth/conn.h
weeTike: Did you see `bt_conn_le_read_min_conn_interval` below? That fetches min_supported_conn_interval only.

include/zephyr/bluetooth/conn.h
jhedberg: At least for HCI parameters we've tried to follow the exact naming in the spec, however I do see the benefit of having the unit in the name, so not opposed to it.

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3715,8 +3862,13 @@ static int cmd_info(const struct shell *sh, size_t argc, char *argv[])
 		print_le_addr("Local on-air", info.le.local);
 
 		shell_print(sh, "Interval: 0x%04x (%u us)",
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+			    info.le.interval_us / 125,
```
Balaklaka: Here 125 should be a const value, as it would be easy to think that this is a error and should have been 1250. I assume that 125 is correct? All the 125 values should be a defined value.

include/zephyr/bluetooth/conn.h
weeTike: I am not opposed to this, but it seems to me that if we make existing code work in this way it could have unintended side affects and lead to unexpected behaviour. In my opinion it is better to make people intentionally make their code compatible if using this kconfig, since it introduces a fundamental change to how connection timings work. I think code assuming old timings that is then roughly assimilated to new timings could easily break in hard to track down ways.

include/zephyr/bluetooth/conn.h
weeTike: 125 us is the base unit of SCI. HCI commands send and receive SCI timings in units of 125 us. I chose to use interval_us instead of in units of 125 because it is impartial to the base used and the timing represented is intuitive to understand.

include/zephyr/bluetooth/conn.h
alwa-nordic: Sorry, I meant to copy-paste `bt_conn_le_read_min_conn_interval` not `set_defaults`. The point about adding `@sa` in the documentation stands. What do you think about that?

weeTike: Sounds good

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3353,173 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+{
+	struct net_buf *rsp;
+	struct bt_hci_op_le_read_min_supported_conn_interval *rp;
+	int err;
+
+	if (info == NULL) {
+		return -EINVAL;
+	}
+
+	err = bt_hci_cmd_send_sync(BT_HCI_OP_LE_READ_MIN_SUPPORTED_CONN_INTERVAL, NULL,
```
alwa-nordic, line 3367: Future work: This implementation reads this value over HCI. But this value could also be conveyed using device tree when the Controller is known at compile time (`CONFIG_HAS_BT_CTLR`). I think this is worth looking into to save on code size and startup time and possibly RAM as well.

weeTike: This could be an optional idea, but the controller is free to calculate this value based on enabled features etc, so i dont think you can have a straight value in device tree. Could be calculated based on enabled kconfigs in some controllers.

include/zephyr/bluetooth/conn.h
alwa-nordic: The users will be made aware due to the deprecation warning.

But yeah, the risk of application breakage due to rounding the interval is difficult to characterize (because we don't know all the use-cases), so let's avoid the risk.

I propose:
 - `interval_us` field always exists.
 - `interval` is deprecated.
 - `interval` does *not exist* when `CONFIG_BT_SHORTER_CONNECTION_INTERVALS=y`.

This way, it's completely safe, and the feature becomes additive-only after two release cycles.

subsys/bluetooth/audio/bap_broadcast_assistant.c
alwa-nordic: If you agree, I suggest moving a copy of `bt_conn_le_set_interval_1250_us_units` into each of the audio files using it, and renaming it to not look like a public function from the Bluetooth API.

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
 	uint16_t supervision_timeout;
 };
 
+/** Most Connection Interval Groups possible
```
Thalley:
```suggestion
/** @brief Maximum Connection Interval Groups possible
```

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+ *
+ * The practical maximum is 41 groups based on HCI event size constraints:
+ *     (HCI event max size - cmd_complete overhead - response fields) / sizeof(group)
+ *     (255 - 4 - 3) / 6 = 41 groups max
+ */
+#define BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS 41
+
+/** Minimum supported connection interval group
```
Thalley:
```suggestion
/** @brief Minimum supported connection interval group
```

Suggest to add @brief when there's both a brief and elaborate texts

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+ *
+ * Each group represents an arithmetic sequence of supported connection intervals:
+ *   min, min+stride, min+2*stride, ..., min+n*stride (where min+n*stride <= max)
```
Thalley: Suggest to be consistent with spaces

```suggestion
 *   min, min + stride, min + 2 * stride, ..., min + n * stride (where min + n * stride <= max)
```

or

```suggestion
 *   min, min+stride, min+2*stride, ..., min+n*stride (where min+n*stride<=max)
```

Not sure if we have any guidelines for this in the documentation though (we require spaces around operators in code).

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+ *
+ * Example: min=10 (1250 us), max=60 (7500 us), stride=5 (625 us)
+ *   -> Supported: 1250 us, 1875 us, 2500 us, ..., 6875 us, 7500 us
+ *
+ * All values in units of 125 microseconds.
+ */
```
Thalley: I would define the unit per field. Using "All values" can be tricky and may be overlooked if the struct is extended with another field that isn't using that unit

include/zephyr/bluetooth/conn.h
Thalley: Suggest to add the unit as well as the minimum and maximum value of each field (where applicable)

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	/** Increment between consecutive supported intervals */
+	uint16_t stride;
+};
+
+/** Minimum supported connection interval information */
+struct bt_conn_le_min_conn_interval_info {
+	/** Minimum supported connection interval in units of 125 microseconds. */
+	uint8_t min_supported_conn_interval;
+	/** Number of interval groups. */
+	uint8_t num_groups;
```
Thalley: Ditto, where are the minimum and maximum values of these? 

Is `num_groups == 0` possible, or is it always non-0?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	 *
+	 * Multiple groups allow representing non-contiguous or differently-strided ranges.
+	 * If num_groups=0, only intervals in the Baseline ConnInterval Values (BCV) range
+	 * are supported.
+	 */
+	struct bt_conn_le_min_conn_interval_group groups[BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS];
+};
+
+/** Connection rate parameters for LE connections */
+struct bt_conn_le_conn_rate_param {
```
Thalley, line 319: Ditto, adding reference to the minimum and maximum values of each field would be nice

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	/** Minimum connection interval (N * 125 us). */
+	uint16_t interval_min;
+	/** Maximum connection interval (N * 125 us). */
+	uint16_t interval_max;
+	/** Minimum subrate factor. */
+	uint16_t subrate_min;
+	/** Maximum subrate factor. */
+	uint16_t subrate_max;
+	/** Maximum Peripheral latency in units of subrated connection intervals. */
+	uint16_t max_latency;
```
Thalley: Do we have a reference for this unit? Else suggest to write the unit

include/zephyr/bluetooth/conn.h
Thalley: I would actually omit the status. Presumably if the `status != BT_HCI_ERR_SUCCESS` then the rest of the values are invalid? In the callback, we could even just send `NULL` in case of error (assuming that is the case)

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+/** Updated connection rate parameters */
+struct bt_conn_le_conn_rate_changed {
+	/** HCI Status from LE Connection Rate Change event. */
+	uint8_t status;
+	/** Connection interval in units of 125 microseconds. */
+	uint16_t interval;
+	/** Connection subrate factor. */
+	uint16_t subrate_factor;
+	/** Peripheral latency in units of subrated connection intervals. */
+	uint16_t peripheral_latency;
```
Thalley: Ditto w.r.t. unit

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,93 @@ struct bt_conn_le_subrate_changed {
...
+	/** Number of underlying connection events to remain active after
+	 *  a packet containing a Link Layer PDU with a non-zero Length
+	 *  field is sent or received.
+	 */
+	uint16_t continuation_number;
+	/** Connection Supervision timeout (N * 10 ms). */
```
Thalley: Above you write `in units of 125 microseconds` and here you write `N * 10 ms`. Please be consistent

subsys/bluetooth/host/conn.c
alwa-nordic: Indeed, I'm suggesting that the Controller properties will be *computed* at compile time. This already somewhat works for single-image builds. It's just not made full use of. And I think the multi-image build system could be improved so it gains the `CONFIG_HAS_BT_CTLR` interface.

include/zephyr/bluetooth/conn.h
Thalley: I do like the idea of having the unit embedded in the field/variable name, and something I often do with local variables at least. In cases like this, where the documentation clearly states the unit, I'd be less inclined to provide the unit in the field. 

It works well if it's just `_ms`, `_us`, but with units like 1.25ms, it would probably be `_125us`, which may require us to use e.g. 32 bits for a variable that could otherwise be `uint16_t` since we have to scale to microseconds, as `_1.25ms` isn't a valid field name (I think?)

include/zephyr/bluetooth/conn.h
alwa-nordic: I'm not really seeing the difference to be honest. What if there comes a "ESCI" specification in 5-10 years? But, I think if we incorporate the units in the field names, the difference in units is emphasized enough and won't be confused, and I'm OK with how it is.

include/zephyr/bluetooth/conn.h
Thalley: I agree with the above. Given the many various units (0.625ms, 1.25ms, 10ms, etc.), there is a usability improvement of using microseconds everywhere IMO. 

However, we also need to consider the memory overhead of course. I think @cvinayak have previously pointed out a few times how the memory usage of Bluetooth in Zephyr, even for very simple builds, are continuously increasing, even without any new features added to said build.

include/zephyr/bluetooth/conn.h
```diff
@@ -890,7 +977,13 @@ struct bt_conn_le_info {
 	const bt_addr_le_t *local;
 	/** Remote device address used during connection setup. */
 	const bt_addr_le_t *remote;
-	uint16_t interval; /**< Connection interval */
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+	/**< Connection interval in microseconds */
+	uint32_t interval_us;
```
Thalley:
```suggestion
	/** Connection interval in microseconds */
	uint32_t interval_us;
```

include/zephyr/bluetooth/conn.h
Thalley: Based on https://github.com/zephyrproject-rtos/zephyr/pull/98901/files#r2503599661 I think this function will go away, right?

include/zephyr/bluetooth/conn.h
```diff
@@ -1180,6 +1273,35 @@ enum bt_conn_auth_keypress {
...
+	return conn_info->le.interval * 1250;
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
+}
+
+/** @brief Set the connection interval in units of 1250 microseconds
+ *
+ * @param conn_info pointer to a @ref bt_conn_info object
+ * @param interval Connection interval in units of 1250 microseconds
+ */
+static inline void bt_conn_le_set_interval_1250_us_units(struct bt_conn_info *conn_info,
```
Thalley: Ditto w.r.t. https://github.com/zephyrproject-rtos/zephyr/pull/98901/files#r2503599661

include/zephyr/bluetooth/conn.h
```diff
@@ -1329,6 +1451,67 @@ int bt_conn_le_subrate_set_defaults(const struct bt_conn_le_subrate_param *param
 int bt_conn_le_subrate_request(struct bt_conn *conn,
 			       const struct bt_conn_le_subrate_param *params);
 
+/** @brief Read Minimum Supported Connection Interval Groups.
+ *
+ *  Read the minimum supported connection interval and supported interval
+ *  groups from the local controller. This information describes what the
+ *  local controller supports.
+ *
+ *  @note To use this API @kconfig{CONFIG_BT_SHORTER_CONNECTION_INTERVALS} must be set.
```
Thalley:
```suggestion
 *  @kconfig_dep{CONFIG_BT_SHORTER_CONNECTION_INTERVALS}
```

Thalley: Applies multiple places in this PR

include/zephyr/bluetooth/conn.h
Thalley: I tend to agree with @alwa-nordic here. It would be nice to be consistent with units, and either use 125us or us, rather than a mix

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3353,173 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+	/* All limits according to BT Core spec 6.2 [Vol 4, Part E, 7.8.154] */
+
+	if (((param->interval_min < 0x0003) || (param->interval_min > 0x7D00)) ||
+	    ((param->interval_max < 0x0003) || (param->interval_max > 0x7D00)) ||
+	    (param->interval_min > param->interval_max)) {
+		return false;
+	}
+
+	if (param->subrate_min < 0x0001 || param->subrate_min > 0x01F4 ||
+	    param->subrate_max < 0x0001 || param->subrate_max > 0x01F4 ||
```
Thalley: Please define the minimum and maximum values from the HCI spec  in `hci_types.h` and use the definitions here, rather than having duplicate literal values

Thalley (changes requested):
Mostly only reviewed the header files

include/zephyr/bluetooth/conn.h
alwa-nordic: > However, we also need to consider the memory overhead of course. I think have previously pointed out a few times how the memory usage of Bluetooth in Zephyr, even for very simple builds, are continuously increasing, even without any new features added to said build.

We have the option to make the fields private and provide getters. The internal representation can then vary and be optimized.

Thalley: > We have the option to make the fields private and provide getters. The internal representation can then vary and be optimized.

That's a good point. It's only the values that are explicitly exposed via the API that will be affected, and in most cases those are likely stack allocated

include/zephyr/bluetooth/conn.h
weeTike: The status can be useful to the user to know what went wrong. If it is not zero, it gives an error code.

Thalley: I'm not suggesting to remove the error code, I'm suggesting to move the error code out of this struct, and then have the callback like 
```c
	void (*conn_rate_changed)(struct bt_conn *conn, uint8_t hci_status,
				  const struct bt_conn_le_conn_rate_changed *params);
```

and then in case of success we have 
```c
cb->conn_rate_changed(conn, 0, params);
```

and in case of error we have
```c
cb->conn_rate_changed(conn, err, NULL);
```

(with the assumption that `bt_conn_le_conn_rate_changed` won't contain any valid values if there's an error)

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,155 @@ struct bt_conn_le_subrate_changed {
...
+ *   min, min + stride, min + 2 * stride, ..., min + n * stride (where min + n * stride <= max)
+ *
+ * Example: min = 10 (1250 us), max = 60 (7500 us), stride = 5 (625 us)
+ *   -> Supported: 1250 us, 1875 us, 2500 us, ..., 6875 us, 7500 us
+ */
+struct bt_conn_le_min_conn_interval_group {
+	/** @brief Lower bound of group interval range
+	 *
+	 * Units: 125 microseconds
+	 * Range: 0x0003 - 0x7D00
```
Thalley: Thanks for adding the units and ranges in a human readable way. 

2 comments: 
1) I have a slight preference to refer to the min/max values by referencing the defines values (if they exist, i.e. `@ref BT_<whatever>` over `0x0003`. 

2) A newline here does not correspond to a newline in the built documentation (need empty line for that). So this currently renders as 

Image

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,155 @@ struct bt_conn_le_subrate_changed {
...
+ * Each group represents an arithmetic sequence of supported connection intervals:
+ *   min, min + stride, min + 2 * stride, ..., min + n * stride (where min + n * stride <= max)
+ *
+ * Example: min = 10 (1250 us), max = 60 (7500 us), stride = 5 (625 us)
+ *   -> Supported: 1250 us, 1875 us, 2500 us, ..., 6875 us, 7500 us
+ */
+struct bt_conn_le_min_conn_interval_group {
+	/** @brief Lower bound of group interval range
+	 *
+	 * Units: 125 microseconds
```
Thalley:
```suggestion
	 * Unit: 125 microseconds
```

Shouldn't it be singular in this context (ditto for other places)

include/zephyr/bluetooth/conn.h
```diff
@@ -2064,6 +2285,31 @@ struct bt_conn_cb {
 				const struct bt_conn_le_subrate_changed *params);
 #endif /* CONFIG_BT_SUBRATING */
 
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
```
Thalley:
```suggestion
#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
```

Need this part to enable Doxygen for conditionally compiled callbacks and fields

include/zephyr/bluetooth/conn.h
```diff
@@ -890,7 +1039,12 @@ struct bt_conn_le_info {
 	const bt_addr_le_t *local;
 	/** Remote device address used during connection setup. */
 	const bt_addr_le_t *remote;
-	uint16_t interval; /**< Connection interval */
+	/** Connection interval in microseconds */
+	uint32_t interval_us;
+#if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
```
Thalley:
```suggestion
#if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
```

Need this part to enable Doxygen for conditionally compiled callbacks and fields

subsys/bluetooth/audio/bap_broadcast_assistant.c
```diff
@@ -453,12 +453,13 @@ static void long_bap_read(struct bt_conn *conn, uint16_t handle)
 		if (err != 0) {
 			LOG_DBG("Failed to get conn info, use default interval");
 
-			conn_info.le.interval = BT_GAP_INIT_CONN_INT_MIN;
+			conn_info.le.interval_us =
+				BT_GAP_INIT_CONN_INT_MIN * BT_HCI_LE_INTERVAL_UNIT_US;
```
Thalley: Consider using `BT_CONN_INTERVAL_TO_US`

subsys/bluetooth/audio/bap_unicast_client.c
```diff
@@ -1767,12 +1767,13 @@ static void long_ase_read(struct bt_bap_unicast_client_ep *client_ep)
 		if (err != 0) {
 			LOG_DBG("Failed to get conn info, use default interval");
 
-			conn_info.le.interval = BT_GAP_INIT_CONN_INT_MIN;
+			conn_info.le.interval_us =
+				BT_GAP_INIT_CONN_INT_MIN * BT_HCI_LE_INTERVAL_UNIT_US;
```
Thalley: Consider using `BT_CONN_INTERVAL_TO_US`

subsys/bluetooth/audio/bap_broadcast_assistant.c
Thalley: This also makes me wonder if we should instead actually define a new `BT_GAP_INIT_CONN_INT_MIN_US` or similar. 

One issue here is that so far everything related to connection intervals in all of our APIs, have always expected connection interval to be a unit of 1.25ms. 
With the addition of `interval_us`, macros like `BT_GAP_INIT_CONN_INT_MIN` may be misleading, not just because the value of `BT_GAP_INIT_CONN_INT_MIN` is the "old" minimum (i.e. without `CONFIG_BT_SHORTER_CONNECTION_INTERVALS`), but also that `BT_GAP_INIT_CONN_INT_MIN` is in a 1.25ms unit, and thus doesn't work well for the the new `interval_us`.

subsys/bluetooth/host/shell/bt.c
```diff
@@ -1009,7 +1009,31 @@ void subrate_changed(struct bt_conn *conn,
 		bt_shell_print("Subrate change failed (HCI status 0x%02x)", params->status);
 	}
 }
-#endif
+
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+void conn_rate_changed(struct bt_conn *conn, uint8_t status,
```
Thalley:
```suggestion
static void conn_rate_changed(struct bt_conn *conn, uint8_t status,
```

subsys/bluetooth/host/shell/bt.c
```diff
@@ -1009,7 +1009,31 @@ void subrate_changed(struct bt_conn *conn,
...
+		       const struct bt_conn_le_conn_rate_changed *params)
+{
+	if (status == BT_HCI_ERR_SUCCESS) {
+		bt_shell_print("Connection rate parameters changed: "
+			       "Connection Interval: 0x%04x (%d us) "
```
Thalley:
```suggestion
			       "Connection Interval: %u us "
```

Does it make sense to print the value in hex, when it's now just a value in US? Previous unit (1.25ms) sort of made sense to have both as the actual value, and as converted to milliseconds.

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3388,131 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
-#endif
+
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+static int cmd_read_min_conn_interval_groups(const struct shell *sh, size_t argc, char *argv[])
+{
+	int err;
+	struct bt_conn_le_min_conn_interval_info info;
+
+	err = bt_conn_le_read_min_conn_interval_groups(&info);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3388,131 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
+		shell_print(
+			sh,
+			"  Group %d: min=0x%04x (%d us), max=0x%04x (%d us), stride=0x%04x (%d us)",
+			i,
+			info.groups[i].min_125us,
+			info.groups[i].min_125us * BT_HCI_LE_SCI_INTERVAL_UNIT_US,
+			info.groups[i].max_125us,
+			info.groups[i].max_125us * BT_HCI_LE_SCI_INTERVAL_UNIT_US,
+			info.groups[i].stride_125us,
+			info.groups[i].stride_125us * BT_HCI_LE_SCI_INTERVAL_UNIT_US);
```
Thalley: Would it make sense to have some simple 125us => us conversion macros like `BT_CONN_INTERVAL_TO_US`?

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3388,131 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
+	return 0;
+}
+
+static int cmd_read_min_conn_interval(const struct shell *sh, size_t argc, char *argv[])
+{
+	int err;
+	uint16_t min_interval_us;
+
+	err = bt_conn_le_read_min_conn_interval(&min_interval_us);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3388,131 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
+		shell_error(sh, "bt_conn_le_read_min_conn_interval returned error %d", err);
+		return -ENOEXEC;
+	}
+
+	shell_print(sh, "Minimum supported connection interval: %d us", min_interval_us);
```
Thalley:
```suggestion
	shell_print(sh, "Minimum supported connection interval: %u us", min_interval_us);
```

subsys/bluetooth/host/shell/bt.c
Thalley: This applies multiple places; our coding guidelines rule 85 require that we only provide boolean values to control statements (https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html)

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3388,131 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
+
+	for (size_t argn = 1; argn < argc; argn++) {
+		(void)shell_strtoul(argv[argn], 10, &err);
+
+		if (err) {
+			shell_help(sh);
+			shell_error(sh, "Could not parse input number %zu", argn);
+			return SHELL_CMD_HELP_PRINTED;
+		}
+	}
```
Thalley, line 3450: I understand the choice of doing this, and it is a simple way of testing if all the inputs are integers. 

However it doesn't check if the values are valid, i.e. below you may attempt to assign a 32-bit value to a 16-bit variable. 

Maybe we can extend `shell_strtoul` to take min/max values, or add a new (local) function that does this to ensure that the resulting value is within the expected range?

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3715,8 +3866,13 @@ static int cmd_info(const struct shell *sh, size_t argc, char *argv[])
...
 
 		shell_print(sh, "Interval: 0x%04x (%u us)",
-			    info.le.interval,
-			    BT_CONN_INTERVAL_TO_US(info.le.interval));
+#if defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS)
+			    info.le.interval_us / BT_HCI_LE_SCI_INTERVAL_UNIT_US,
+#else
+			    info.le.interval_us / BT_HCI_LE_INTERVAL_UNIT_US,
+#endif /* CONFIG_BT_SHORTER_CONNECTION_INTERVALS */
+			    info.le.interval_us);
```
Thalley: Please use `IS_ENABLED`, or consider just printing the value in US

subsys/bluetooth/host/conn.c
```diff
@@ -3243,6 +3246,24 @@ void bt_conn_notify_subrate_change(struct bt_conn *conn,
...
+				     const struct bt_conn_le_conn_rate_changed *params)
+{
+	BT_CONN_CB_DYNAMIC_FOREACH(callback) {
+		if (callback->conn_rate_changed) {
+			callback->conn_rate_changed(conn, status, params);
+		}
+	}
+
+	STRUCT_SECTION_FOREACH(bt_conn_cb, cb) {
+		if (cb->conn_rate_changed) {
```
Thalley:
```suggestion
		if (callback->conn_rate_changed != NULL) {
			callback->conn_rate_changed(conn, status, params);
		}
	}

	STRUCT_SECTION_FOREACH(bt_conn_cb, cb) {
		if (cb->conn_rate_changed != NULL) {
```

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+
+	info->min_supported_conn_interval_125us = rp->min_supported_conn_interval;
+	info->num_groups = rp->num_groups;
+
+	/* Copy groups up to the number allocated by the application */
+	for (uint8_t i = 0; i < rp->num_groups; i++) {
+		info->groups[i].min_125us = sys_le16_to_cpu(rp->groups[i].group_min);
+		info->groups[i].max_125us = sys_le16_to_cpu(rp->groups[i].group_max);
+		info->groups[i].stride_125us = sys_le16_to_cpu(rp->groups[i].group_stride);
+	}
```
Thalley, line 3391: Do we need to perform input validation here, in case that any of these values are outside the allowed value range?

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+
+	err = bt_hci_cmd_send_sync(BT_HCI_OP_LE_READ_MIN_SUPPORTED_CONN_INTERVAL, NULL,
+				   &rsp);
+	if (err) {
+		return err;
+	}
+
+	rp = (struct bt_hci_op_le_read_min_supported_conn_interval *)rsp->data;
+	*min_interval_us = rp->min_supported_conn_interval * BT_HCI_LE_SCI_INTERVAL_UNIT_US;
```
Thalley: Ditto, input validation?

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+
+	net_buf_unref(rsp);
+	return 0;
+}
+
+static bool le_conn_rate_common_params_valid(const struct bt_conn_le_conn_rate_param *param)
```
Thalley, line 3420: Consider using the `IN_RANGE` macro in this function

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+}
+
+int bt_conn_le_conn_rate_set_defaults(const struct bt_conn_le_conn_rate_param *params)
+{
+	struct bt_hci_op_le_set_default_rate_parameters *cp;
+	struct net_buf *buf;
+
+	if (!IS_ENABLED(CONFIG_BT_CENTRAL)) {
+		return -ENOTSUP;
+	}
```
Thalley: Considering guarding the entire function with `#if defined`, so that if `CONFIG_BT_CENTRAL=n` it will result in a compile-time error, instead of a runtime error, which makes it easier/faster to catch. Not sure if we have any coding guidelines on this matter though

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+
+	if (params == NULL || !le_conn_rate_common_params_valid(params)) {
+		return -EINVAL;
+	}
+
+	buf = bt_hci_cmd_alloc(K_FOREVER);
+	if (!buf) {
```
Thalley:
```suggestion
	if (buf == NULL) {
```

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+		LOG_DBG("Invalid connection type: %u for %p", conn->type, conn);
+		return -EINVAL;
+	}
+
+	if (params == NULL || !le_conn_rate_common_params_valid(params)) {
+		return -EINVAL;
+	}
+
+	buf = bt_hci_cmd_alloc(K_FOREVER);
+	if (!buf) {
```
Thalley:
```suggestion
	if (buf == NULL) {
```

subsys/bluetooth/host/conn.c
```diff
@@ -3331,6 +3352,177 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+	return bt_hci_cmd_send_sync(BT_HCI_OP_LE_SET_DEFAULT_RATE_PARAMETERS, buf, NULL);
+}
+
+int bt_conn_le_conn_rate_request(struct bt_conn *conn,
+				 const struct bt_conn_le_conn_rate_param *params)
+{
+	struct bt_hci_op_le_connection_rate_request *cp;
+	struct net_buf *buf;
+
+	if (!bt_conn_is_type(conn, BT_CONN_TYPE_LE)) {
```
Thalley: Consider `bt_conn_is_le` which was recently added

subsys/bluetooth/host/conn_internal.h
```diff
@@ -107,7 +107,7 @@ struct bt_conn_le {
 	bt_addr_le_t init_addr;
 	bt_addr_le_t resp_addr;
 
-	uint16_t interval;
+	uint32_t interval_us;
 	uint16_t interval_min;
 	uint16_t interval_max;
```
Thalley, line 113: Should we consider appending the unit of these interval values as well? 
Is it OK that they are not in microseconds when the actual interval is?

subsys/bluetooth/host/hci_core.c
```diff
@@ -2807,6 +2808,72 @@ void bt_hci_le_subrate_change_event(struct net_buf *buf)
...
+void bt_hci_le_conn_rate_change_event(struct net_buf *buf)
+{
+	struct bt_hci_evt_le_conn_rate_change *evt;
+	struct bt_conn_le_conn_rate_changed params;
+	struct bt_conn *conn;
+
+	evt = net_buf_pull_mem(buf, sizeof(*evt));
+
+	conn = bt_conn_lookup_handle(sys_le16_to_cpu(evt->handle), BT_CONN_TYPE_LE);
+	if (!conn) {
```
Thalley:
```suggestion
	if (conn == NULL) {
```

Thalley (changes requested):
Reviewed the source files this time as well. A couple of questions and suggestions, as well as a handful of minor coding style violations :)

subsys/bluetooth/host/conn.c
weeTike: I dont think this is the job of the host. The controller should only pass valid values (unless it violates spec) or the application should validate them.

Thalley: I think we discussed in the Zephyr BT working group at some point (@jhedberg @alwa-nordic or am I remembering wrong?). I don't recall what the conclusion was in terms of whether the host should consider the controller a 3rd party device (or 2nd party?), and thus should not trust it anymore than it should trust any other Bluetooth device. 

In any case, we certainly don't perform validation of all values from the controller today

subsys/bluetooth/host/shell/bt.c
weeTike: I think such a change would be outside of the scope of this pr. This seems to be a pattern a lot of other commands are using and should be good enough for now.

alwa-nordic (dismissed):

subsys/bluetooth/host/shell/bt.c
Thalley: I'm OK not considering this a blocker

doc/releases/migration-guide-4.4.rst
```diff
@@ -43,6 +43,11 @@ QSPI
 Bluetooth
 *********
 
+Bluetooth Host
+==============
+* :c:member:`bt_conn_info.interval` has been deprecated. Use :c:member:`bt_conn_info.interval_us`
+  instead.
```
Thalley: I think this is obvious to many of us, but perhaps we should make explicitly clear that the deprecated field and the new field are different units. 
Unfortunately I can imaging some users to simply do a search and replace for `interval`->`interval_is` not realizing that the units are significantly different

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
...
+	 * Unit: 125 microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_SCI_INTERVAL_MIN_125US - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
+	 */
+	uint16_t max_125us;
+	/** @brief Increment between consecutive supported intervals
+	 *
+	 * Unit: 125 microseconds
+	 *
+	 * Range: 0x0001 - @ref BT_HCI_LE_SCI_INTERVAL_MAX_125US
```
Thalley: Should there be a define for the minimum stride value as well, given that we have that for the minimum intervals?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
...
+	uint16_t stride_125us;
+};
+
+/** Minimum supported connection interval information */
+struct bt_conn_le_min_conn_interval_info {
+	/** @brief Minimum supported connection interval
+	 *
+	 * Unit: microseconds
+	 *
+	 * Range: 0x0177 - 0x1D4C
```
Thalley: Is there a reason to provide these as hex values, instead of decimal values?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
...
+	 * Unit: subrated connection intervals @ref bt_conn_le_conn_rate_changed.subrate_factor
+	 *
+	 * Range: 0x0000 - @ref BT_HCI_LE_PERIPHERAL_LATENCY_MAX
+	 */
+	uint16_t max_latency;
+	/** @brief Minimum number of underlying connection events to remain active
+	 *  after a packet containing a Link Layer PDU with a non-zero Length
+	 *  field is sent or received.
+	 *
+	 * Range: 0x0000 - @ref BT_HCI_LE_CONTINUATION_NUM_MAX
```
Thalley: Ditto with no minimum value define?

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
...
+	uint16_t max_ce_len_125us;
+};
+
+/** Updated connection rate parameters */
+struct bt_conn_le_conn_rate_changed {
+	/** Connection interval
+	 *
+	 * Unit: microseconds
+	 *
+	 * Range: 0x000177 - 0x3D0900
```
Thalley: Missing define? Should they be in decimals?

include/zephyr/bluetooth/conn.h
```diff
@@ -890,7 +1053,21 @@ struct bt_conn_le_info {
...
+	uint32_t interval_us;
+#if !defined(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) || defined(__DOXYGEN__)
+	union {
+		/** @brief Connection interval in units of 1.25 ms
+		 *
+		 * @deprecated Use @ref bt_conn_le_info.interval_us instead
+		 */
+		__deprecated uint16_t interval;
+		/** Workaround for setting deprecated @ref bt_conn_le_info.interval */
+		uint16_t _interval;
```
Thalley, line 1067:
```suggestion
		/** @cond INTERNAL_HIDDEN */
		/** Workaround for setting deprecated @ref bt_conn_le_info.interval */
		uint16_t _interval;
		/** @endcond */
```

I assume you don't want this (private?) field to show up in the documentation?

include/zephyr/bluetooth/conn.h
```diff
@@ -924,6 +1101,12 @@ struct bt_conn_le_info {
  */
 #define BT_CONN_INTERVAL_TO_US(interval) ((interval) * 1250U)
 
+/** @brief Convert shorter connection interval to microseconds
+ *
+ *  Multiply by 125 to get microseconds.
+ */
+#define BT_CONN_SCI_INTERVAL_TO_US(interval) ((interval) * BT_HCI_LE_SCI_INTERVAL_UNIT_US)
```
Thalley:
```suggestion
#define BT_CONN_SCI_INTERVAL_TO_US(_interval) ((_interval) * BT_HCI_LE_SCI_INTERVAL_UNIT_US)
```

Suggest to prepend with `_` to avoid conflicts with variables

include/zephyr/bluetooth/conn.h
```diff
@@ -2064,6 +2315,31 @@ struct bt_conn_cb {
...
+	 *  @param conn   Connection object.
+	 *  @param status HCI Status from LE Connection Rate Change event.
+	 *                Possible Status codes:
+	 *                  Success (0x00)
+	 *                  Unknown Connection Identifier (0x02)
+	 *                  Command Disallowed (0x0C)
+	 *                  Unsupported Feature or Parameter Value (0x11)
+	 *                  Invalid HCI Command Parameters (0x12)
+	 *                  Unsupported Remote Feature (0x1A)
+	 *                  Unsupported LL Parameter Value (0x20)
```
Thalley: I don't think this list of error codes renders the way it should: 

Image

(https://builds.zephyrproject.io/zephyr/pr/98901/docs/doxygen/html/structbt__conn__cb.html#a9cbc0e6b91fc2742c59815b676875235)

weeTike:
@jhedberg Would you be able to take a look please?

subsys/bluetooth/host/shell/bt.c
```diff
@@ -3361,7 +3387,130 @@ static int cmd_subrate_request(const struct shell *sh, size_t argc, char *argv[]
...
+	}
+
+	shell_print(sh, "Minimum supported connection interval: %u us",
+		    info.min_supported_conn_interval_us);
+	shell_print(sh, "Number of groups: %d", info.num_groups);
+
+	for (uint8_t i = 0; i < info.num_groups; i++) {
+		shell_print(
+			sh,
+			"  Group %d: min=0x%04x (%u us), max=0x%04x (%u us), stride=0x%04x (%u us)",
```
Thalley:
```suggestion
			"  Group %u: min=0x%04x (%u us), max=0x%04x (%u us), stride=0x%04x (%u us)",
```

subsys/bluetooth/host/conn.c
```diff
@@ -2003,30 +2003,30 @@ void bt_conn_notify_remote_info(struct bt_conn *conn)
 
 void bt_conn_notify_le_param_updated(struct bt_conn *conn)
 {
+	uint16_t interval = conn->le.interval_us / BT_HCI_LE_INTERVAL_UNIT_US;
```
Thalley: Should this be 
```suggestion
	uint16_t interval_1250us = conn->le.interval_us / BT_HCI_LE_INTERVAL_UNIT_US;
```

To follow the same naming used in the API fields?

subsys/bluetooth/host/conn.c
```diff
@@ -3332,6 +3353,182 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+	struct bt_hci_op_le_read_min_supported_conn_interval *rp;
+	int err;
+
+	if (info == NULL) {
+		return -EINVAL;
+	}
+
+	err = bt_hci_cmd_send_sync(BT_HCI_OP_LE_READ_MIN_SUPPORTED_CONN_INTERVAL, NULL,
+				   &rsp);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

subsys/bluetooth/host/conn.c
```diff
@@ -3332,6 +3353,182 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+	struct net_buf *rsp;
+	struct bt_hci_op_le_read_min_supported_conn_interval *rp;
+
+	if (min_interval_us == NULL) {
+		return -EINVAL;
+	}
+
+	err = bt_hci_cmd_send_sync(BT_HCI_OP_LE_READ_MIN_SUPPORTED_CONN_INTERVAL, NULL,
+				   &rsp);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

subsys/bluetooth/host/conn.c
```diff
@@ -3332,6 +3353,182 @@ int bt_conn_le_subrate_request(struct bt_conn *conn,
...
+	cp->supervision_timeout = sys_cpu_to_le16(params->supervision_timeout_10ms);
+	cp->min_ce_len = sys_cpu_to_le16(params->min_ce_len_125us);
+	cp->max_ce_len = sys_cpu_to_le16(params->max_ce_len_125us);
+
+	return bt_hci_cmd_send_sync(BT_HCI_OP_LE_SET_DEFAULT_RATE_PARAMETERS, buf, NULL);
+}
+#endif /* CONFIG_BT_CENTRAL */
+
+int bt_conn_le_conn_rate_request(struct bt_conn *conn,
+				 const struct bt_conn_le_conn_rate_param *params)
```
Thalley, line 3497: Is `bt_conn_le_conn_rate_set_defaults` only for centrals, but `bt_conn_le_conn_rate_request` for both central and peripheral?

subsys/bluetooth/host/hci_core.c
```diff
@@ -3885,6 +3961,14 @@ static int le_init(void)
 		}
 	}
 
+	if (IS_ENABLED(CONFIG_BT_SHORTER_CONNECTION_INTERVALS) &&
+	    BT_FEAT_LE_SHORTER_CONN_INTERVALS(bt_dev.le.features)) {
+		err = le_set_host_feature(BT_LE_FEAT_BIT_SHORTER_CONN_INTERVALS_HOST_SUPP, 1);
+		if (err) {
```
Thalley:
```suggestion
		if (err != 0) {
```

Thalley (dismissed):
Some minor comments left from me. 

I do have one concern that I think we need to address first (ideally with input from @jhedberg as well, so it's not just Nordic involved). This PR implements the shorter connection intervals, but to support this it also modifies the connection interval unit from 1.25ms to us (microseconds). I am not convince that this change is as trivial as this PR wants to make it, as `conn_interval`, `CONN_INTERVAL`, etc. everywhere in the code and API has always been in units of 1.25ms. If we now default to microseconds, instead of 1.25ms (as we do by deprecating the old value), then macros like `BT_GAP_MS_TO_CONN_INTERVAL`, `BT_LE_CONN_PARAM`, many `bt_conn` callbacks and functions, etc. all still use the terminology `interval`. I'm worried that we in this PR have 2 types of interval (one in 1.25ms units, and one in microseconds), and that converting to the new unit requires significantly more effort than what this PR does. 

Perhaps we should keep both interval fields temporarily, and then do a proper deprecation/switch to using microseconds?

subsys/bluetooth/host/conn.c
weeTike: Yes

include/zephyr/bluetooth/conn.h
```diff
@@ -252,6 +252,169 @@ struct bt_conn_le_subrate_changed {
...
+	/** @brief Minimum supported connection interval
+	 *
+	 * Unit: microseconds
+	 *
+	 * Range: @ref BT_HCI_LE_MIN_SUPP_CONN_INT_MIN_US - @ref BT_HCI_LE_MIN_SUPP_CONN_INT_MAX_US
+	 */
+	uint16_t min_supported_conn_interval_us;
+	/** @brief Number of interval groups.
+	 *
+	 * Range: 0 - 255
```
Thalley: Isn't the range more like `0 - BT_CONN_LE_MAX_CONN_INTERVAL_GROUPS`?

Balaklaka (approved):

alwa-nordic (dismissed):

alwa-nordic (dismissed):

include/zephyr/bluetooth/conn.h
asbjornsabo: 1ms25, as is often done for voltages? (3V3 for 3.3 volts)

jhedberg (dismissed):

Thalley (dismissed):
No blockers from me, and if we found anything to follow up on, we have plenty of time before the next release

KyraLengfeld (dismissed):

Thalley (dismissed):

alwa-nordic (approved):

jhedberg (approved):

KyraLengfeld (approved):

Thalley (approved):

PR #98902, opened 2025-11-05
drivers: audio: add driver for TI TAA3020 Audio ADC

M1cha:
The driver has basic support for recording audio and volume control. Impedance and line-level can be configured as well. The microphone is not supported, yet.


dts/bindings/audio/ti,taa3020.yaml
```diff
@@ -0,0 +1,67 @@
...
+properties:
+  gpio1-gpios:
+    type: phandle-array
+
+  areg-internal-en:
+    type: boolean
+    description: |
+      Enable the internal regulator. This property must be set, when the supply
+      voltage is above 1.8 V. Failing to enable this, might lead to damaging the
+      device!
```
dsseng: From the datasheet it seems like this depends not only on the supply voltage, but also on your design.

Even with VDDA == 1.8V it seems you may want to keep this disconnected and let the codec produce this voltage, and perhaps with 3.3V power you can also power the 1.8 rail from an external source.

This is what I got from briefly reading the public docs, you probably know this better

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
...
+	uint8_t val;
+	uint8_t in_ch_en = 0;
+
+	codec_set_sleep(dev, false);
+
+	if (config->areg_internal) {
+		codec_read_reg(dev, TAA3020_SLEEP_CFG, &val);
+		val |= TAA3020_SLEEP_CFG_AREG_SELECT;
+		codec_write_reg(dev, TAA3020_SLEEP_CFG, val);
+	}
```
dsseng: The datasheet suggests we should do a single write, both enabling 1v8 rail if using internal regulator and resuming. And this seems to make sense. Maybe this code should also be reorganized to send a single write to make sure the codec resumes properly in all configs (have you tested and validated the internal regulator mode?)

dsseng (commented):
Thanks! Looks great overall, some minor questions regarding matching the behaviour to the reference

dts/bindings/audio/ti,taa3020.yaml
rvoell-7lab: You could probably provide the 1,8 V externally and also connect a 3,3 V rail, but why? You could decide on one of the rails.

The datasheet says, when AVDD is less than 1.98 V, then short AVDD and AREG, when AVDD is above 2.7 V, then use the internal regulator. So i guess this description could be improved a bit.

Datasheet excerpt:

image

drivers/audio/taa3020.c
rvoell-7lab: I didn't found a section in the datasheet, where this is claimed. But the datasheet says, that in sleep mode one should send no other i2c transaction than waking up from sleep. I interpreted this in a way, that two transactions are necessary. One for wake up, and one for enabling the AREG regulator.

Relevant section from datasheet:

image

dts/bindings/audio/ti,taa3020.yaml
dsseng: I'm not an expert on the docs, but perhaps the binding should indicate you must enable the option when areg is not externally powered or something like this

drivers/audio/taa3020.c
dsseng: Take a look at the section `8.2.1.2.1 Example Device Register Configuration Script for EVM Setup`: I see a single write to SLEEP_CFG that both clears SLEEP_ENZ and sets I2C broadcast and AREG_SELECT. And this seems valid: tell the codec to both resume from sleep and enable regulator. This seems to correlate well with `when exiting sleep ... configure ... AREG`, likely meaning that you send a single request to both exit sleep and configure 1v8 at the same time.

You already do a transaction to resume, and I guess that falls under the condition stated

What power rail configuration do you use? Do you use the on-chip regulator? If so, it is at least tested and might be okay to do. There seems to be no code reference other than the scripts in the datasheet so I'd adhere to them if I were implementing this

M1cha: > Take a look at the section 8.2.1.2.1 Example Device Register Configuration Script for EVM Setup: I see a single write to SLEEP_CFG that both clears SLEEP_ENZ and sets I2C broadcast and AREG_SELECT.

We tested this and it works, thanks for the hint :+1:

dts/bindings/audio/ti,taa3020.yaml
M1cha: Since the datasheet is very unclear about this, I changed the sentence to simply point to the datasheet with a note that hardware can be damaged on wrong configuration.

glneo (dismissed):

drivers/audio/Kconfig.taa3020
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+config AUDIO_CODEC_TAA3020
+	bool "TI TAA3020 ADC support"
+	default y
+	depends on DT_HAS_TI_TAA3020_ENABLED
+	select I2C
+	depends on GPIO
+	depends on AUDIO
+	depends on AUDIO_CODEC
```
Alex9360: I think this AUDIO and AUDIO_CODEC dependency not needed by default it been like that only no need to add separate dependency.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
...
+	struct i2c_dt_spec bus;
+	const struct codec_channel_config *channels;
+	size_t num_channels;
+
+	bool areg_internal;
+};
+
+struct codec_channel_data {
+	bool configured;
+	audio_channel_t audio_channel;
```
Alex9360, line 40: For better memory alignment to reduce padding audio_channel_t (typically 4 bytes) should be placed before the bool fields

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
...
+struct codec_driver_data {
+	struct codec_channel_data channels[4];
+};
+
+static void codec_write_reg(const struct device *dev, uint8_t reg, uint8_t val)
+{
+	const struct codec_driver_config *const config = dev->config;
+
+	i2c_reg_write_byte_dt(&config->bus, reg, val);
+	LOG_DBG("%s WR REG:0x%02x VAL:0x%02x", dev->name, reg, val);
```
Alex9360: minor suggestion if it only has i2c interface then why do we need to have this wrapper function why don't you use raw i2c apis rather than this wrapper?
Apply to all wrapper.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
...
+{
+	uint8_t val = 0;
+	struct codec_driver_data *data = dev->data;
+
+	codec_read_reg(dev, TAA3020_ASI_OUT_CH_EN, &val);
+	if (enable) {
+		val |= TAA3020_ASI_OUT_CH_EN_CHANNEL(channel) & TAA3020_ASI_OUT_CH_EN_MASK;
+	} else {
+		val &= ~(TAA3020_ASI_OUT_CH_EN_CHANNEL(channel) & TAA3020_ASI_OUT_CH_EN_MASK);
+	}
```
Alex9360, line 86: even though TAA3020_ASI_OUT_CH_EN_CHANNEL(x) produces a single bit, the expression
val |= TAA3020_ASI_OUT_CH_EN_CHANNEL(channel) & TAA3020_ASI_OUT_CH_EN_MASK;
is unnecessary and confusing. The same applies to the disable path. The bitwise and with mask does no effect here. 

```
if (enable) {
    val |= TAA3020_ASI_OUT_CH_EN_CHANNEL(channel);
} else {
    val &= ~TAA3020_ASI_OUT_CH_EN_CHANNEL(channel);
}
```
This might help you.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
...
+	case TAA3020_CHANNEL_IN_TYPE_LINE:
+		val |= TAA3020_CHx_CFG0_INTYP_LINE;
+		break;
+	}
+
+	switch (chan_config->in_src) {
+	case TAA3020_CHANNEL_IN_SRC_ANALOG_DIFFERENTIAL:
+		val |= TAA3020_CHx_CFG0_INSRC_ANALOG_DIFFERENTIAL;
+		break;
+	case TAA3020_CHANNEL_IN_SRC_ANALOG_SIGNLE_ENDED:
```
Alex9360: typo "SIGNLE" should be "SINGLE"

drivers/audio/taa3020.c
M1cha: Currently, mostly for the debug log.

M1cha:
The last push contained the requested changes, the ones before that were just rebases.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,486 @@
+/*
+ * Copyright (c) 2025 sevenlab engineering GmbH
```
Alex9360: copyright year has to be change and also seems like format of the copyright is changed in zephyr pls refer #106836. apply to all files

Alex9360 (changes requested):

drivers/audio/taa3020.c
M1cha: thanks for the hint, fixed.

Alex9360 (approved):
LGTM. nit I think binding file also needs copyright verify once. also some of the CI failures related to default case in switch statements also needs to be fixed.

drivers/audio/taa3020.c
glneo: Drop the debug log stuff, then remove the wrapper function.

M1cha: Done

M1cha:
> LGTM. nit I think binding file also needs copyright verify once. also some of the CI failures related to default case in switch statements also needs to be fixed.

I disagree with sonarqube here, because adding default statements to switches which are intended to be complete would prevent getting a `-Wswitch` warning after adding new enum variants but forgetting to update the cases.

M1cha:
I addressed the comments and added the driver to build_all.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,478 @@
...
+	bool configured;
+	uint8_t volume;
+	bool muted;
+};
+
+struct codec_driver_data {
+	struct codec_channel_data channels[4];
+};
+
+static void codec_set_device_page(const struct device *dev, uint8_t page)
```
Alex9360: this page set function only called one time in the entire program also it just call i2c  write api so this can be also replaced by same i2c_write api no need this wrapper IMO.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,478 @@
...
+	val |= TAA3020_SLEEP_CFG_SLEEP_ENZ;
+
+	if (config->areg_internal) {
+		val |= TAA3020_SLEEP_CFG_AREG_SELECT;
+	} else {
+		val &= TAA3020_SLEEP_CFG_AREG_SELECT;
+	}
+
+	i2c_reg_write_byte_dt(&config->bus, TAA3020_SLEEP_CFG, val);
+	k_msleep(1);
```
Alex9360, line 366: how did you arrived this delay period? is it mentioned in data sheet?

Alex9360 (changes requested):

drivers/audio/taa3020.c
rvoell-7lab: Yes, it actually is.

image

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,478 @@
...
+	uint8_t val;
+	uint8_t in_ch_en = 0;
+
+	i2c_reg_read_byte_dt(&config->bus, TAA3020_SLEEP_CFG, &val);
+	val |= TAA3020_SLEEP_CFG_SLEEP_ENZ;
+
+	if (config->areg_internal) {
+		val |= TAA3020_SLEEP_CFG_AREG_SELECT;
+	} else {
+		val &= TAA3020_SLEEP_CFG_AREG_SELECT;
```
rvoell-7lab:
```suggestion
		val &= ~TAA3020_SLEEP_CFG_AREG_SELECT;
```

drivers/audio/taa3020.c
rvoell-7lab: The reason for having this function, is to have the code read what it actually does. The taa3020 has three register pages, of which the current driver only uses one, as it does not support all features yet.

One could argue with "YAGNI", but i personally like to have code which uses names functions describing what actually is done, even when not strictly needed.

That was the reason i added this function, but i'm fine when michael changes this :grinning:

M1cha: Usually I agree with @rvoell-7lab to make the code express intend and meaning through function names. In this case there's not much harm in removing the function, though because the enum variant `TAA3020_PAGE_CFG` makes this quite descriptive already.

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,478 @@
...
+{
+	uint8_t val = 0;
+	const struct codec_driver_config *const config = dev->config;
+	struct codec_driver_data *data = dev->data;
+
+	i2c_reg_read_byte_dt(&config->bus, TAA3020_ASI_OUT_CH_EN, &val);
+	if (enable) {
+		val |= TAA3020_ASI_OUT_CH_EN_CHANNEL(channel);
+	} else {
+		val &= TAA3020_ASI_OUT_CH_EN_CHANNEL(channel);
```
dsseng: Maybe you meant this, only clearing the masked bits?

```suggestion
		val &= ~TAA3020_ASI_OUT_CH_EN_CHANNEL(channel);
```

drivers/audio/taa3020.c
dsseng: Same here - not 100% sure, but seems like missing ~

drivers/audio/taa3020.c
```diff
@@ -0,0 +1,478 @@
...
+		err = codec_configure_channel(dev, &config->channels[i]);
+		if (err < 0) {
+			LOG_ERR("Failed to configure channel #%zu: %d", i, err);
+			return -EFAULT;
+		}
+
+		in_ch_en |= TAA3020_IN_CH_EN_CHANNEL(config->channels[i].channel);
+	}
+
+	in_ch_en &= TAA3020_IN_CH_EN_MASK;
```
rvoell-7lab: This masking operation is not needed and can be removed
```suggestion
```

M1cha:
I addressed the comments.

glneo (approved):

M1cha:
It would be great if some of the audio collaborators could take a look.

PR #98906, opened 2025-11-05
soc/ite/it8xxx2: make ARCH_IRQ_VECTOR_TABLE_ALIGN constant

### LINKED ISSUE #99000: Downstream regression on ITE platform due to change in alignment
### Describe the bug

https://github.com/zephyrproject-rtos/zephyr/commit/a825e014d858eb2badd600da6f3163e413b6bf3f caused a shit of the data section in many riscv platforms including ITE ones. Downstream we rely on the position of some fixed identifier that is placed immediately after the vector table, so the change there caused a regression.

[https://github.com/zephyrproject-rtos/zephyr/commit/a825e014d858eb2badd600da6f3163e413b6bf3f resto](https://github.com/zephyrproject-rtos/zephyr/pull/98906) restores the previous behavior. We mitigated by changing the setting downstream locally so keeping the severity low but I think this should go in stabilization.

### Regression

- [x] This is a regression.

### Steps to reproduce

n/a

### Relevant log output

It's a build failure in our application.

### Impact

### Environment

_No response_

### Additional Context

_No response_

RuibinChang:
The IQR vector table changed size since https://github.com/zephyrproject-rtos/zephyr/commit/a825e014d858eb2badd600da6f3163e413b6bf3f and it is
causing issues with downstream applications, for example,
data stored in bin specific field is shifted by
the ARCH_IRQ_VECTOR_TABLE_ALIGN, so I make
ARCH_IRQ_VECTOR_TABLE_ALIGN constant.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/99000

### DIFF
--- a/soc/ite/ec/it8xxx2/Kconfig
+++ b/soc/ite/ec/it8xxx2/Kconfig
@@ -187,6 +187,9 @@ config SOC_IT8XXX2_EXT_32K
 
 endchoice
 
+config ARCH_IRQ_VECTOR_TABLE_ALIGN
+	default 4
+
 config SOC_IT8XXX2_USE_ILM
 	bool
 	default y


fabiobaltieri:
Hi @RuibinChang, thanks a lot for the commit, could you update the message to mention that the vector table changed size since a825e014d85 and it is causing issues with downstream applications? The whole fwid story make sense only for chromeos so some more generic context would help. The bug number reference can go, it's not public anyway.

fabiobaltieri (approved):

keith-zephyr (approved):

PR #98911, opened 2025-11-05
soc: nordic: uicr: Move GEN_UICR options into the main Zephyr tree

SebastianBoe:
I initially added the GEN_UICR options to the gen_uicr image's Kconfig tree only to not pollute the main tree.

But there is a lot of useful help text in the GEN_UICR option's Kconfig file that I would like users to be able to read/reference from the docs.

To not increase the complexity of the Kconfig doc generator, we add the GEN_UICR options to the main tree and have them all be disabled for builds other than the gen_uicr image.

Being in the main tree has the added benefit of being recognzied by the compliance checker.


soc/nordic/common/uicr/gen_uicr/prj.conf
```diff
@@ -1 +1,6 @@
-# nothing here
+# Copyright (c) 2025 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+# Enable UICR generator options - this is automatically set when building
+# the gen_uicr image and enables the UICR generator configuration menu
+CONFIG_IS_GEN_UICR_IMAGE=y
```
jonathannilsen: Does this prj.conf file get overridden when we define a conf file at sysbuild/uicr.conf in a sample directory, or does that work as an overlay?

karstenkoenig (approved):

fabiobaltieri (approved):

hakonfam (approved):

nashif (approved):

Issue #97602, opened 2025-10-15
Fix ci issues reported for fvp_baser_aemv8r


wearyzen:
### Describe the bug

The list of issues is available here:
https://github.com/zephyrproject-rtos/zephyr/actions/runs/17800996278/job/50600943470?pr=95687
https://github.com/zephyrproject-rtos/zephyr/actions/runs/17800996278/job/50600943492?pr=95687
https://github.com/zephyrproject-rtos/zephyr/actions/runs/17800996278/job/50600943469?pr=95687

```
INFO    - 1) kernel.memory_protection.syscalls.kyield on fvp_baser_aemv8r/fvp_aemv8r_aarch32 failed (Testsuite failed)
INFO    - 2) kernel.memory_protection.syscalls.timeslicing on fvp_baser_aemv8r/fvp_aemv8r_aarch32 failed (Testsuite failed)
INFO    - 3) sample.bindesc on fvp_baser_aemv8r/fvp_aemv8r_aarch32 failed (rc=2)
INFO    - 4) arch.common.stack_unwind.arm64 on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp error (Build failure - undefined reference to `arch_page_phys_get')
INFO    - 5) arch.common.stack_unwind.symtab on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp error (Build failure - undefined reference to `arch_page_phys_get')
INFO    - 6) integration.kernel on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp failed (rc=2)
INFO    - 7) ipc.icmsg_pbuf on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp failed (rc=2)
INFO    - 8) kernel.smp_abort on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp failed (rc=2)
INFO    - 9) kernel.workqueue on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp failed (rc=2)
INFO    - 10) kernel.workqueue.work_timeout on fvp_baser_aemv8r/fvp_aemv8r_aarch64/smp failed (rc=2)
```

### Regression

- [ ] This is a regression.

### Relevant log output

```shell

```

### Impact

Functional Limitation – Some features not working as expected, but system usable.

nashif:
is this still an issue? where can this be obvserved? what is the plan?

wearyzen:
[...]

PR #98912, opened 2025-11-05
boards: fix: ci failures for Arm FVP boards

wearyzen:
Related issue #97602


npitre (commented):
This is a mixed bag of unrelated things.

For example, the stackwalk filter is a legitimate generic fix that deserves
its own PR.

As to the slowness with all those huge timeout multipliers, did you consider
`-C cache_state_modelled=0` as fvp_base_revc_2xaemv8a does?

And 8KB for the idle thread stacksize is a bit silly. Do you really need that?

wearyzen:
> This is a mixed bag of unrelated things.
> 
> For example, the stackwalk filter is a legitimate generic fix that deserves its own PR.
> 
> As to the slowness with all those huge timeout multipliers, did you consider `-C cache_state_modelled=0` as fvp_base_revc_2xaemv8a does?
> 
> And 8KB for the idle thread stacksize is a bit silly. Do you really need that?

I have a local change with `-C cache_state_modelled=0` and few other to avoid the stacksize but they don't fix all the issues so I went for the current approach with which I can unblock some tests for this board. The issues will be fixed eventually when I get some more bandwidth for this but for now I would like to unblock PR 95687.

npitre:
You should ask someone in power to override the CI failure for merging 
PR 95687. A one-line change like this shouldn't be gated on pre-existing 
CI issues, and butchering a bunch of tests for this small change is not 
the way to go either.

wearyzen:
> You should ask someone in power to override the CI failure for merging PR 95687. A one-line change like this shouldn't be gated on pre-existing CI issues, and butchering a bunch of tests for this small change is not the way to go either.

I am not sure what I would is recommended in this case... whether to mask few tests in ci (not sure how) or add them to the ignore list of that board. @kartben any suggestions?


BTW the current ci failure will be fixed once https://github.com/zephyrproject-rtos/zephyr/pull/98817 gets merged.

wearyzen:
@npitre I have raised https://github.com/zephyrproject-rtos/zephyr/pull/99014 and https://github.com/zephyrproject-rtos/zephyr/pull/99017 to keep them separate from this FVP PR. The commits are still added to this PR to confirm that CI passes but I'll removed the last 2 commits once those PRs are merged.

Apart from the last 2 commits I think all the other commits are related to fixing tests on FVP boards so I hope this can help push this forward.

ithinuel (approved):

wearyzen:
> @wearyzen could you update commit [c78fbae](https://github.com/zephyrproject-rtos/zephyr/commit/c78fbae3262d44cc01477c91191f9b8802642acc) message with a link to the issue(s) please? This should not dismiss existing reviews :)

done.

kartben (approved):

wearyzen:
> @wearyzen thanks! Unless I misunderstood, the commit is not really fixing the issue but just referring to where you are tracking the fact that tests will need to be re-enabled at some point, so i guess "Fixes ..." is the wrong term here, but eh no big deal I guess

I've changed it to "Related issue" for now but let me know if there is something better to update or keep in mind for the next time.

PR #98929, opened 2025-11-05
dts: arm: st: u0: deprecate sram0 node label

etienne-lms:
Deprecate sram0 node label from STM32U0xx SoC series DTSI files since it duplicates existing sram1 label that better matches SoCs reference manual wording. Update board DTSI files accordingly.

This prevents confusion for 2 nodes labels defining the same RAM but defined at different places.

Factorize definitions of `sram0` node (embedded SRAM1) and `sram2` node (embedded SRAM2) in STM32U0xx SoCs series. U031xx all have 8kB of SRAM1 and 4kB of SRAM2. U073xx and U083xx all have 32kB of SRAM1 and 8kB of SRAM2.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -34,6 +34,10 @@ Kernel
 Boards
 ******
 
+* Node label ``sram0`` is deprecated in STM32U0xx SoCs in favor to label ``sram1`` that better
+  match the wording used to refer to this internal memory in the SoCs reference manuals and
+  litterature. The label is still defined but will be removed in the future.
+
 Device Drivers and Devicetree
 *****************************
 
--- a/dts/arm/st/u0/stm32u031X4.dtsi
+++ b/dts/arm/st/u0/stm32u031X4.dtsi
@@ -6,14 +6,8 @@
 
 #include <st/u0/stm32u031.dtsi>
 
-/ {
-	soc {
-		flash-controller@40022000 {
-			flash0: flash@8000000 {
-				reg = <0x08000000 DT_SIZE_K(16)>;
-			};
-		};
-	};
+&flash0 {
+	reg = <0x08000000 DT_SIZE_K(16)>;
 };
 
 &sram1 {
--- a/dts/arm/st/u0/stm32u031X4.dtsi
+++ b/dts/arm/st/u0/stm32u031X4.dtsi
@@ -7,18 +7,6 @@
 #include <st/u0/stm32u031.dtsi>
 
 / {
-	sram1: memory@20000000 {
-		compatible = "zephyr,memory-region", "mmio-sram";
-		reg = <0x20000000 DT_SIZE_K(8)>;
-		zephyr,memory-region = "SRAM1";
-	};
-
-	sram2: memory@20080000 {
-		compatible = "zephyr,memory-region", "mmio-sram";
-		reg = <0x20080000 DT_SIZE_K(4)>;
-		zephyr,memory-region = "SRAM2";
-	};
-
 	soc {
 		flash-controller@40022000 {
 			flash0: flash@8000000 {
@@ -26,4 +14,12 @@
 			};
 		};
 	};
+};
+
+&sram1 {
+	reg = <0x20000000 DT_SIZE_K(8)>;
+};
+
+&sram2 {
+	reg = <0x20080000 DT_SIZE_K(4)>;
 };
--- a/dts/arm/st/u0/stm32u073.dtsi
+++ b/dts/arm/st/u0/stm32u073.dtsi
@@ -72,16 +72,6 @@
 		};
 	};
 
-	sram1: memory@20000000 {
-		compatible = "zephyr,memory-region", "mmio-sram";
-		zephyr,memory-region = "SRAM1";
-	};
-
-	sram2: memory@20008000 {
-		compatible = "zephyr,memory-region", "mmio-sram";
-		zephyr,memory-region = "SRAM2";
-	};
-
 	usb_fs_phy: usbphy {
 		compatible = "usb-nop-xceiv";
 		#phy-cells = <0>;
--- a/boards/st/nucleo_u031r8/nucleo_u031r8.dts
+++ b/boards/st/nucleo_u031r8/nucleo_u031r8.dts
@@ -20,7 +20,7 @@
 	chosen {
 		zephyr,console = &usart2;
 		zephyr,shell-uart = &usart2;
-		zephyr,sram = &sram0;
+		zephyr,sram = &sram1;
 		zephyr,flash = &flash0;
 	};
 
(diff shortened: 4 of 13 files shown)


dts/arm/st/u0/stm32u031.dtsi
```diff
@@ -7,6 +7,18 @@
 #include <st/u0/stm32u0.dtsi>
 
 / {
+	sram0: memory@20000000 {
+		compatible = "zephyr,memory-region", "mmio-sram";
+		reg = <0x20000000 DT_SIZE_K(8)>;
+		zephyr,memory-region = "SRAM1";
```
dsseng: DT label doesn't match the region name, this is likely to confuse people who debug this

etienne-lms: Ok, i'll restore the original name to match the node label.

dsseng: they should probably match the bank names in the SoC RM, which seem to be sram1 and sram2

dts/arm/st/u0/stm32u073.dtsi
```diff
@@ -72,13 +72,15 @@
 		};
 	};
 
-	sram1: memory@20000000 {
+	sram0: memory@20000000 {
```
JarmouniA: mismatch with region name.

etienne-lms: Prior this change, `sram0` label was defined for node `/soc/memory@20000000` in **stm32u0.dtsi** and `sram1` label was defined for the very same node in various **stm32u0??*.dtsi** files. Label `sram0` is used in board and samples/tests overlay files whilke `sram1` label is not used anywhere (unless I missed something).

This is why I would like the remove `sram1` label to prevent confusion.
An alternative can be to remove `sram0` label, keep only `sram1` label and update board DTS files to use the latter. The sole issue will be with **tests/drivers/memc/ram/** for which will neeed to disable`sram1` for these SoCs.

JarmouniA: Change the region name as well to SRAM0 (or am I missing something?)

etienne-lms: Yes, i'll do, as per https://github.com/zephyrproject-rtos/zephyr/pull/98929#discussion_r2494939135.

dts/arm/st/u0/stm32u031.dtsi
etienne-lms: Using better names would be orth be considered on all STM32 SoCs in a consistent way.
For now,I plan to restore the current SoCs way where `sram0` label is the default Zephyr main SRAM label, and i'll use `"SRAM0"` as region name.

dts/arm/st/u0/stm32u073.dtsi
etienne-lms: Hmm... I'd rather change my mind.

Would you (and @dsseng) be okay to rather deprecate `sram0` for these STMU0xx SoCs in favor to `sram1` label (and region naming). It would better match the SoC ref man.

Actually that's conservative since current implementation (before my changes). It would only explicitly state that `sram0` node label is deprecated. I would still define this label, for compliance with out-of-tree boards, but will be able to fully remove it after a few release tags, maybe for v4.6 LTS or next v5.x tags.

dsseng: > Would you (and @dsseng) be okay to rather deprecate `sram0` for these STMU0xx SoCs in favor to `sram1` label (and region naming). It would better match the SoC ref man.

I believe this is a wise idea. sram0 isn't a special name, and you should be able to plug any RAM node in using the chosen properties. So I'm for naming the device as it is named in the SoC docs, thus sram1, and deprecating old one.

Deprecation process feels appropriate here in my opinion (the devices seem to have been merged in 4.0, so there are likely downstream consumers that might rely on the node names), but @JarmouniA should certainly know more on this

JarmouniA: > Would you (and @dsseng) be okay to rather deprecate sram0 for these STMU0xx SoCs in favor to sram1 label (and region naming). It would better match the SoC ref man.

Makes sense.
My objection was only on the mismatch between label and region name, didn't look at the ref man.

etienne-lms:
I've updated the change according to https://github.com/zephyrproject-rtos/zephyr/pull/98929#discussion_r2503930511.

dsseng (approved):

dts/arm/st/u0/stm32u031X4.dtsi
```diff
@@ -7,18 +7,6 @@
...
-		reg = <0x20000000 DT_SIZE_K(8)>;
-		zephyr,memory-region = "SRAM1";
-	};
-
-	sram2: memory@20080000 {
-		compatible = "zephyr,memory-region", "mmio-sram";
-		reg = <0x20080000 DT_SIZE_K(4)>;
-		zephyr,memory-region = "SRAM2";
-	};
-
```
etienne-lms, line 21: @erwango, I wonder if factorizing SRAMs size info in **stm32u0??.dtsi** sub-series files is a good idea. Maybe it is more convenient to have both flash and SRAM sizes all defined together in **stm32u??X?.dtsi**, for consistency with the other STM32 SoCs.
Based on feedback on this P-R related to STM32U0xx, I plan to propose equivalent changes in other SoCs to better name SRAM nodes accrdong to reference manual wording.

gautierg-st (approved):

dts/arm/st/u0/stm32u031X4.dtsi
```diff
@@ -27,3 +15,11 @@
...
 	};
 };
+
+&sram1 {
+	reg = <0x20000000 DT_SIZE_K(8)>;
+};
+
+&sram2 {
+	reg = <0x20080000 DT_SIZE_K(4)>;
+};
```
erwango: In st .dtsi files, we generally don't overwrite nodes using `&node` references but directly "in place", so it would be:
```
/ {
	sram1: memory@20000000 {
		reg = <0x20000000 DT_SIZE_K(8)>;
	};
       [...]
```
I think, this should be kept for consistency and better readability of the .dtsi nodes (as it preserves "hierarchy": for instance we see that sram, is not part of node soc/, which is not the case using &node reference.).

erwango (changes requested):

doc/releases/migration-guide-4.4.rst
```diff
@@ -34,6 +34,10 @@ Kernel
 Boards
 ******
 
+* Node label ``sram0`` is deprecated in STM32U0xx SoCs in favor to label ``sram1`` that better
+  match the wording used to refer to this internal memory in the SoCs reference manuals and
+  litterature. The label is still defined but will be removed in the future.
```
mathieuchopstm, line 39: Typo/error fixes + reword proposal:
```suggestion
* Node label ``sram0`` is deprecated in STM32U0 SoCs, replaced by ``sram1`` that better
  matches the internal memory's name found in Reference Manuals and litterature. The
  ``sram0`` label is still defined for compatibility, but will be removed in the future.
```

FRASTM (approved):

PR #98937, opened 2025-11-05
twister: Fix serial open for software devices that do not immediately appear

VisaJE:
* Testing on linux, --flash-before argument did not work to wait for jlink upload to complete. Instead Serial port opening throws file not found.
* The fix ensures that the code waits for the serial port to appear before attempting to connect.

### DIFF
--- a/scripts/pylib/twister/twisterlib/handlers.py
+++ b/scripts/pylib/twister/twisterlib/handlers.py
@@ -27,6 +27,7 @@ from pathlib import Path
 from queue import Empty, Queue
 
 import psutil
+from serial.tools import list_ports
 from twisterlib.environment import ZEPHYR_BASE, strip_ansi_sequences
 from twisterlib.error import TwisterException
 from twisterlib.platform import Platform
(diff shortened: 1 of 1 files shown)


scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -662,14 +663,17 @@ def _create_serial_connection(self, dut, serial_device, hardware_baud,
...
+                port=None if dut.flash_before else serial_device,
                 baudrate=hardware_baud,
                 parity=serial.PARITY_NONE,
                 stopbits=serial.STOPBITS_ONE,
                 bytesize=serial.EIGHTBITS,
                 # the worst case of no serial input
-                timeout=max(flash_timeout, self.get_test_timeout())
+                timeout=max(flash_timeout, self.get_test_timeout()),
             )
+            ser.port = serial_device
```
hakehuang: is this needed?

VisaJE: No

VisaJE:
Hm thanks i'm new to the authoring system

VisaJE:
Huh I wonder if the while loop introduces a lot of debug logs in some system :confused:

hakehuang (approved):

nashif (approved):

PR #97943, opened 2025-10-20
lib: smf: rename share_parent() to is_descendant_of()

vlad-kulikov:
Match behavior: returns true when target_state equals or
is an ancestor of test_state.

PR #98940, opened 2025-11-05
Smf/add vlad kulikov maintainer

vlad-kulikov:
This is a follow-up to @keith-zephyr’s note that SMF is looking for a new
maintainer:
https://github.com/zephyrproject-rtos/zephyr/pull/97943#issuecomment-3439589314

I realize this is coming shortly after I was added as a collaborator, but
the request is only because there is an active need right now and I’m
already spending most of my time in SMF: submitting source-level 
changes, sending SMF-related RFCs and making performance-oriented
tweaks. If there wasn’t an open search, I would have waited longer 
before proposing this.

I would be happy to take on the SMF maintainer duties so the 
subsystem has someone who is active in it.

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -4775,10 +4775,10 @@ State machine framework:
   status: maintained
   maintainers:
     - sambhurst
+    - vlad-kulikov
   collaborators:
     - keith-zephyr
     - glenn-andrews
-    - vlad-kulikov
   files:
     - doc/services/smf/
     - include/zephyr/smf.h


nashif (dismissed):

keith-zephyr (dismissed):

nashif (approved):

vlad-kulikov:
> something wrong with the branch, can you please rebase?

Thanks, I see you already rebased the branch.

keith-zephyr (approved):

PR #98945, opened 2025-11-05
serial: uart_bitbang: fix incorrect callback device

### LINKED ISSUE #98229: uart_bitbang: uart_bitbang_tx_counter_top_interrupt uses wrong device
### Describe the bug

The uart bitbang callbacks (e.g. uart_bitbang_rx_counter_top_interrupt, uart_bitbang_tx_counter_top_interrupt, …) probably use the wrong device when calling nested callbacks.

For example, `uart_bitbang_rx_counter_top_interrupt`  uses dev instead of uart_dev. 

```
#ifdef CONFIG_UART_INTERRUPT_DRIVEN
				if ((data->user_cb) && (data->irq & UART_BITBANG_IRQ_PE)) {
					data->user_cb(dev, data->user_data);
				}
#endif /* CONFIG_UART_INTERRUPT_DRIVEN */
```

This causes `uart_irq_update(uart)` to always returning a value <1, because uart_irq_update is called on a counter-device instead of a uart one.
As a result, data received by the bit-banging driver is not processed correctly in the uart modem backend. This prevents, the use of the gnss subsystem (and probably of all things that make use of the uart interrupt) in combination with the bitbang driver. 

### Regression

- [ ] This is a regression.

### Steps to reproduce

Build GNSS example with activated bitbang driver on a T-Beam board (v1.2). 

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Windows
- Zephyr OS build: v4.2.0-6166-g8a72d776c5af

### Additional Context

_No response_

dcpleung:
When doing the callback, it should pass it the UART device pointer. So fix that.

Fixes #98229

### DIFF
--- a/drivers/serial/uart_bitbang.c
+++ b/drivers/serial/uart_bitbang.c
@@ -187,7 +187,7 @@ static void uart_bitbang_rx_counter_top_interrupt(const struct device *dev, void
 
 #ifdef CONFIG_UART_INTERRUPT_DRIVEN
 				if ((data->user_cb) && (data->irq & UART_BITBANG_IRQ_PE)) {
-					data->user_cb(dev, data->user_data);
+					data->user_cb(uart_dev, data->user_data);
 				}
 #endif /* CONFIG_UART_INTERRUPT_DRIVEN */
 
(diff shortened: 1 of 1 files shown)


nashif (approved):

joelguittet:
I was not using the dev in my case so I have not seen the problem when developing this driver, but I understand the issue with the modem, so this looks good to me !

joelguittet (approved):

edersondisouza (approved):

PR #98946, opened 2025-11-05
drivers: sdhc: esp32: Updates and fixes

raffarost:
Adjust driver so that pinctrl property is not required in device tree, for boards which only use IOCTRL SDHC (fixed) pins. Remove card init functions from driver init, as `set_io()` function is meant to power cycle the card (if pin is used) and initialize card clock and bit width. Remove unnecessary timeout from `handle_idle_state_events()`.

### DIFF
--- a/drivers/sdhc/sdhc_esp32.c
+++ b/drivers/sdhc/sdhc_esp32.c
@@ -186,12 +186,7 @@ static int handle_idle_state_events(struct sdhc_esp32_data *data)
 	 */
 	struct sdmmc_event evt;
 
-	int64_t yield_delay_us = 100 * 1000; /* initially 100ms */
-	int64_t t0 = esp_timer_get_time();
-	int64_t t1 = 0;
-
 	while (sdmmc_host_wait_for_event(data, 0, &evt) == 0) {
-
 		if (evt.sdmmc_status & SDMMC_INTMASK_CD) {
 			LOG_DBG("card detect event");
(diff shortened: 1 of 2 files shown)


wmrsouza (dismissed):

sylvioalves (dismissed):

drivers/sdhc/sdhc_esp32.c
```diff
@@ -1306,11 +1306,14 @@ static int sdhc_esp32_init(const struct device *dev)
 	configure_pin_iomux(cfg->d2_pin);
 	configure_pin_iomux(cfg->d3_pin);
 
-	ret = pinctrl_apply_state(cfg->pcfg, PINCTRL_STATE_DEFAULT);
+	/* GPIO matrix routed pins */
+	if (cfg->pcfg) {
```
pdgendt:
```suggestion
	if (cfg->pcfg != NULL) {
```

wmrsouza (dismissed):

sylvioalves (dismissed):

sylvioalves:
@jhedberg targeting this to v4.3.0 in order to get #98868 fixed in time. I get if it is too late though.

uLipe (approved):

wmrsouza (approved):

marekmatej (approved):

sylvioalves (approved):

PR #98962, opened 2025-11-06
Fix lp mspm0g3519 pinctrl dtsi

lsd-he360:
The LP MSPM0G3519 board has additional pins that cannot be configured with entries defined in `mspm0g1x0x_g3x0x-pinctrl.dtsi`. The HAL was updated to include pinctrl definitions for TI gx51x family of MCUs.


lsd-he360:
I tested the changes by running the echo bot sample.

echo-bot

PR #98973, opened 2025-11-06
stm32n6_dk: tests: drivers:  fix some test regressions  

djiatsaf-st:
This PR fixes regressions encountered in the following test/sample driver scenarios : 

- samples/drivers/spi_flash/sample.drivers.spi.flash
- samples/subsys/usb/uvc/sample.subsys.usb.uvc.encoder.jpeg
- tests/drivers/disk/disk_access/drivers.disk.stm32_sdhc
- tests/drivers/gpio/gpio_api_1pin/drivers.gpio.1pin
- tests/lib/cmsis_dsp/transform/libraries.cmsis_dsp.transform.cq31
- tests/lib/cmsis_dsp/transform/libraries.cmsis_dsp.transform.rq31.fpu
- tests/kernel/mem_protect/protection/kernel.memory_protection.protection
-  tests/application_development/ram_context_for_isr/application_development.ram_context_for_isr


tests/application_development/ram_context_for_isr/testcase.yaml
```diff
@@ -21,3 +21,4 @@ tests:
       - frdm_kl25z/mkl25z4
       - mcx_n9xx_evk/mcxn947/cpu0/qspi
       - frdm_k32l2b3/k32l2b31a
+      - stm32n6570_dk/stm32n657xx/sb
```
mathieuchopstm, line 26: Add a message about why N6 is excluded. Also, maybe this test should depend on `CONFIG_XIP` instead?

etienne-lms (commented):
For commit "tests: drivers: gpio: api_1pin: boards: update led0 on stm32n6570_dk":
- could you mention in the commit message that LED1 is connect from a circuitry than prevents GPIO state from triggering a interrupt while LED2 is directly wired to the GPIO pin, hence the change.
- I noted that the board DTS file sets `red_led_1` as `GPIO_ACTIVE_HIGH` while it's `GPIO_ACTIVE_HIGH` (when toggling, the stat is indeed inverted. I think such a fix would make sense in your series.

tests/drivers/gpio/gpio_api_1pin/boards/stm32n6570_dk_stm32n657xx_sb.overlay
```diff
@@ -0,0 +1,5 @@
+/ {
```
etienne-lms, line 6: Suggestion:
```c
/*
 * SPDX-FileCopyrightText: Copyright (c) 2025 The Zephyr Contributors.
 */
```
Ditto for the other new files added by this series.

boards/st/stm32n6570_dk/twister.yaml
```diff
@@ -1,8 +1,8 @@
 name: STM32N6570 Discovery Kit
 type: mcu
 arch: arm
-ram: 1024
-flash: 1024
+ram: 128
+flash: 512
```
etienne-lms, line 5: This needs to be decreased as well to pass tests?
*stm32n6570_dk/stm32n657xx* has 1536kByte of Zephyr flash (sysbluild/MCUboot).
*stm32n6570_dk/stm32n657xx/fsbl* has 128MB of  Zephyr flash.
*stm32n6570_dk/stm32n657xx/sb* has 128MB of  Zephyr flash.

boards/st/stm32n6570_dk/twister.yaml
```diff
@@ -1,8 +1,8 @@
 name: STM32N6570 Discovery Kit
 type: mcu
 arch: arm
-ram: 1024
-flash: 1024
+ram: 128
```
etienne-lms, line 4: *stm32n6570_dk/stm32n657xx* has 1MByt of Zephyr RAM.
*stm32n6570_dk/stm32n657xx/fsbl* has 511kB of  Zephyr RAM.
*stm32n6570_dk/stm32n657xx/sb* has 511kB of  Zephyr RAM.

I would rather suggest to keep 1024 here and change below:
```patch
variants:
  stm32n6570_dk/stm32n657xx:
    twister: false
  stm32n6570_dk/stm32n657xx/fsbl:
    twister: false
  stm32n6570_dk/stm32n657xx/sb:
    ram: 511
    toolchain:
      - zephyr
      - gnuarmemb

**Review by etienne-lms — COMMENTED** [review]
Comments for commit "boards: st: reduce the size of ram and flash on stm32n6570_dk".

File: drivers/flash/Kconfig.stm32_xspi [hunk 5]
```diff
@@ -19,7 +19,7 @@ config FLASH_STM32_XSPI
 	select PINCTRL
 	select DMA if $(DT_STM32_XSPI_HAS_DMA)
 	select USE_STM32_HAL_DMA if $(DT_STM32_XSPI_HAS_DMA)
-	select USE_STM32_HAL_DMA_EX if SOC_SERIES_STM32H5X && \
+	select USE_STM32_HAL_DMA_EX if (SOC_SERIES_STM32H5X || SOC_SERIES_STM32N6X) && \
```
etienne-lms, line 22: IMHO this change would deserve a specific commit instead of being part of commit
"samples: drivers: spi_flash: exclude stm32n6570_dk from CI testing".

boards/st/stm32n6570_dk/twister.yaml
erwango: What would be the interest of keeping ram/flash here is they're not supposed to be used ?
That would would add confusion, IMO.

boards/st/stm32n6570_dk/twister.yaml
erwango: `stm32n6570_dk/stm32n657xx/sb` is quite specific as it is non XIP, and hence available 511K of RAM should be able to swallow (flash + ram) tests expectations. So, unless twister is able to take it into account (or until twister is able to), we're forced to approximations.
Having:
```
flash: 512
ram: 128
```
would let the possibility to run most of test for which expectations are flash:512/ram:128.
Taking into account that 512 generally means: "more than 256/less than 1024 and same for 128."

This won't work in all cases of course, but I think this is a good starting point that we can refine later on.
@djiatsaf-st have you been able to test the result ?

tests/application_development/ram_context_for_isr/testcase.yaml
djiatsaf-st: From my point of view, we can leave it as is. 
Perhaps that will save us some unforeseen problems.

tests/drivers/gpio/gpio_api_1pin/boards/stm32n6570_dk_stm32n657xx_sb.overlay
djiatsaf-st: done

drivers/flash/Kconfig.stm32_xspi
djiatsaf-st: updated

boards/st/stm32n6570_dk/stm32n6570_dk_common.dtsi
```diff
@@ -61,7 +61,7 @@
 		};
 
 		red_led_1: led_2 {
-			gpios = <&gpiog 10 GPIO_ACTIVE_HIGH>;
+			gpios = <&gpiog 10 GPIO_ACTIVE_LOW>;
```
etienne-lms, line 63: IMHO, a dedicated commit would be better. (either for one to find and pick if needed)
Not blocking.

samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf
erwango, line 1: Since that's a sample that could be tested in various configurations, should it be applied to all variants?
If yes, you should probably propose a soc/stm32n657xx.conf

boards/st/stm32n6570_dk/twister.yaml
djiatsaf-st: > @djiatsaf-st have you been able to test the result ?

Yes, nothing unusual or strange was detected.

erwango (changes requested):

samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf
djiatsaf-st: I tested on all variants (fsbl, sb, application) and they work fine without the need for extra configs using `west`.
Increasing the main stack size is only necessary when testing with `twister` (which is possible only on the sb variant) .  
From my point of view, there is no need to add a soc/stm32n657xx.conf file.

samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf
```diff
@@ -0,0 +1 @@
+CONFIG_MAIN_STACK_SIZE=4096
```
jfischer-no: >Increase the size of the main stack to avoid a stack overflow error
when running the test.

It is a platform agnostic sample, what test do you mean?

There should not be trivial soc/board overlays.
@josuah Should the main stack size of the UVC sample be increased in prj.conf?

jfischer-no (dismissed):

samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf
djiatsaf-st: > It is a platform agnostic sample, what test do you mean?
> 
sample: `samples/subsys/usb/uvc`

Concerned test scenario : `sample.subsys.usb.uvc.encoder.jpeg`

josuah: On my end I have not done any fine-tuning of the stack size.

The [default `1024`](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_MAIN_STACK_SIZE) may be too low and increasing it to 2048 or 4096 directly in `prj.conf` makes sense if this helps reducing the number of per-board configuration.

boards/st/stm32n6570_dk/twister.yaml
etienne-lms: I find it a bit confusing I would rather see such number of the related variant section, but fine as-is.

etienne-lms (commented):
Nitpicking: typo in a commit message header line:
```patch
-tests: drivers: disk: dissk_ access: change sdmmc bus-with on stm32n6570_dk
+tests: drivers: disk: disk_ access: change sdmmc bus-with on stm32n6570_dk

File: samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf (hunk 7 above)
Comment (djiatsaf-st) [reply samples/subsys/usb/uvc/boards/stm32n6570_dk_stm32n657xx_sb.conf]: Updated

**Review by etienne-lms — DISMISSED** [review]

**Review by tbursztyka — DISMISSED** [review]

**Review by josuah — DISMISSED** [review]
LGTM for the UVC part, thanks

File: tests/kernel/mem_protect/protection/boards/stm32n6570_dk_stm32n657xx_sb.conf [hunk 8]
```diff
@@ -0,0 +1 @@
+CONFIG_USERSPACE=y
```
erwango: I don't think this is the correct fix.
If Protection test is designed to run w/o USERSPACE, then we should review mpu_regions settings dependencies.

Let's remove this commit from the series in order not to block the other fixes.

erwango (changes requested):

tests/kernel/mem_protect/protection/boards/stm32n6570_dk_stm32n657xx_sb.conf
djiatsaf-st: Noted
I've removed the commit.

erwango (approved):

FRASTM (approved):

etienne-lms (approved):

PR #98982, opened 2025-11-06
Bluetooth: Classic: HFP: Compatible old version

lylezhu2012:
* Bluetooth: Classic: HFP_HF: Compatible old version AG

In current implementation, the AG with old version cannot be supported properly. Such as, if the AG is version 0.96, the AT command `AT+BRSF` should not be sent. And if the AG is no newer than version 1.5, the `Codec Negotiation` should be unsupported.

Compatible old version with the following changes,

Discover the AG SDP record to get the profile version and AG features. If the SDP discovery is failed, break the RFCOMM DLC
connection.

If the AG version is v0.96, do not send AT command `AT+BRSF`.

Create SCO connection directly if the AG is not newer than version 1.5 in the function `bt_hfp_hf_audio_connect()`.

* Bluetooth: Classic: HFP_AG: Compatible old version HF

In current implementation, the HF with old version cannot be supported properly. Such as, if the AG is version 0.96, the AT command `AT+BRSF` will not be sent. It causes the AG cannot know the features of HF.

Compatible old version with the following changes, 

Discover the HF SDP record to get the profile version and HF features.
If the SDP discovery is failed, break the RFCOMM DLC connection.

If the AG version is v0.96, update the HF features according to the discovered HF features.

* Bluetooth: Classic: HFP_HF: Register SCO connect callback

There is an issue that the SCO connect cannot be un-referenced by HFP HF when the SCO connection is broken if the SCO connect is not created by HFP HF.

Register SCO connect change callback. And un-reference the SCO connect in SCO disconnected callback.

### DIFF
--- a/subsys/bluetooth/host/classic/hfp_hf.c
+++ b/subsys/bluetooth/host/classic/hfp_hf.c
@@ -4474,6 +4474,31 @@ static int bt_hfp_hf_sco_accept(const struct bt_sco_accept_info *info,
 	return 0;
 }
 
+static void hf_sco_connected(struct bt_conn *conn, uint8_t err)
+{
+	ARG_UNUSED(conn);
+	ARG_UNUSED(err);
+}
+
+static void hf_sco_disconnected(struct bt_conn *conn, uint8_t reason)
+{
+	ARG_UNUSED(reason);
+
+	__ASSERT(conn != NULL, "Invalid SCO conn");
+
+	ARRAY_FOR_EACH(bt_hfp_hf_pool, i) {
+		if (bt_hfp_hf_pool[i].sco_conn == conn) {
+			bt_conn_unref(bt_hfp_hf_pool[i].sco_conn);
+			bt_hfp_hf_pool[i].sco_conn = NULL;
+		}
+	}
+}
+
+static struct bt_sco_conn_cb hf_sco_conn_cb = {
+	.connected = hf_sco_connected,
+	.disconnected = hf_sco_disconnected
+};
+
 static void hfp_hf_init(void)
 {
 	static struct bt_rfcomm_server chan = {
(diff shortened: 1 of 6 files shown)


subsys/bluetooth/host/classic/hfp_hf.c
```diff
@@ -3409,23 +3442,120 @@ static int bcc_finish(struct at_client *hf_at, enum at_result result,
...
+	err = hfp_hf_set_voice_setting(hf);
+	if (err < 0) {
+		LOG_ERR("Failed to set voice setting");
+		return err;
+	}
+
+	hf->sco_conn = bt_conn_create_sco(&hf->acl->br.dst, &hf->chan);
+	if (hf->sco_conn == NULL) {
+		LOG_ERR("Failed to create SCO");
+		return -ENOMEM;
```
makeshi, line 3530: looks return -ENOTCONN more make sense, here failure not memory problem.

subsys/bluetooth/host/classic/hfp_hf.c
```diff
@@ -4129,16 +4259,106 @@ static void bt_hf_work(struct k_work *work)
...
+		goto failed;
+	}
+
+	err = bt_sdp_get_profile_version(result->resp_buf, BT_SDP_HANDSFREE_SVCLASS,
+					 &hf->ag_sdp_version);
+	if (err != 0) {
+		LOG_ERR("Failed to get AG profile version");
+		goto failed;
+	}
+	err = bt_sdp_get_features(result->resp_buf, &hf->ag_sdp_features);
```
makeshi, line 4314: missing a empty line

subsys/bluetooth/host/classic/hfp_hf.c
lylezhu2012: The code block aims to create the connection. But it is failed to create connection. Usually it is caused by no memory instead of not connection.

subsys/bluetooth/host/classic/hfp_hf.c
lylezhu2012: Fixed.

makeshi (dismissed):

MarkWangChinese (dismissed):

makeshi (approved):

MarkWangChinese (approved):

PR #98986, opened 2025-11-06
Add mcxw23 power management support

zejiang0jason:
Enabled modes:
      idle: SLEEP
      suspend: DEEP-SLEEP
      standby: POWER-DOWN with CPU retention
    
    OS Time Base: OSTIMER with fro1m clock source

### DIFF
--- a/boards/nxp/frdm_mcxw23/doc/index.rst
+++ b/boards/nxp/frdm_mcxw23/doc/index.rst
@@ -198,6 +198,14 @@ should see the following message in the terminal:
    *** Booting Zephyr OS build v4.2.0-2105-g48f2ffda26de ***
    Hello World! frdm_mcxw23/mcxw236
 
+Power Management
+================
+
+When Power Management is enabled :kconfig:option:`CONFIG_PM`, OSTIMER is used as
+OS tick timer.
+
+Limitation: Wakeup pin can't be used as wakeup source in Standby mode.
+
 .. include:: ../../common/board-footer.rst.inc
 
 .. _MCXW23 SoC Website:
--- a/dts/arm/nxp/nxp_mcxw23x_common.dtsi
+++ b/dts/arm/nxp/nxp_mcxw23x_common.dtsi
@@ -32,12 +32,47 @@
 			reg = <0>;
 			#address-cells = <1>;
 			#size-cells = <1>;
+			cpu-power-states = <&idle &suspend &standby>;
 
 			mpu: mpu@e000ed90 {
 				compatible = "arm,armv8m-mpu";
 				reg = <0xe000ed90 0x40>;
 			};
 		};
+
+		power-states {
+			/* Idle mode maps to Sleep mode. */
+			idle: idle {
+				compatible = "zephyr,power-state";
+				power-state-name = "runtime-idle";
+				min-residency-us = <60>;
+				exit-latency-us = <16>;
+				zephyr,pm-device-disabled;
+			};
+
+			/* Suspend mode maps to Deep Sleep mode. */
+			suspend: suspend {
+				compatible = "zephyr,power-state";
+				power-state-name = "suspend-to-idle";
+				min-residency-us = <1300>;
+				exit-latency-us = <749>;
+			};
+
+			/* Standby mode maps to Power Down mode with CPU retention . */
+			standby: standby {
+				compatible = "zephyr,power-state";
+				power-state-name = "standby";
+				min-residency-us = <2000>;
+				exit-latency-us = <1345>;
+			};
+		};
+	};
+
+	/* For the resource which will be off and lose content. */
+	standby_off_domain: standby-off-domain {
+		compatible = "power-domain-soc-state-change";
+		#power-domain-cells = <0>;
+		onoff-power-states = <&standby>;
 	};
 
 	sysclk: system-clock {
@@ -175,6 +210,7 @@
 		mode = <0>;
 		input = <0>;
 		prescale = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 	};
 
 	ctimer1: ctimer@9000 {
@@ -187,6 +223,7 @@
 		mode = <0>;
 		input = <0>;
 		prescale = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 	};
 
 	ctimer2: ctimer@28000 {
@@ -199,6 +236,7 @@
 		mode = <0>;
 		input = <0>;
 		prescale = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 	};
 
 	ctimer3: ctimer@29000 {
@@ -211,6 +249,7 @@
 		mode = <0>;
 		input = <0>;
 		prescale = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 	};
 
 	ctimer4: ctimer@2a000 {
@@ -223,6 +262,7 @@
 		mode = <0>;
 		input = <0>;
 		prescale = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 	};
 
 	sc_timer: pwm@85000 {
@@ -232,6 +272,7 @@
 		status = "disabled";
 		clocks = <&syscon MCUX_SCTIMER_CLK>;
 		prescaler = <2>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 		#pwm-cells = <3>;
 	};
 
@@ -251,6 +292,7 @@
 		resets = <&reset NXP_SYSCON_RESET(1, 11)>;
 		dmas = <&dma0 4 &dma0 5>;
 		dma-names = "rx", "tx";
+		zephyr,disabling-power-states = <&suspend &standby>;
 		status = "disabled";
 	};
 
@@ -262,6 +304,8 @@
 		resets = <&reset NXP_SYSCON_RESET(1, 12)>;
 		dmas = <&dma0 6 &dma0 7>;
 		dma-names = "rx", "tx";
+		zephyr,disabling-power-states = <&suspend &standby>;
+		power-domains = <&standby_off_domain>;
 		status = "disabled";
 	};
 
@@ -273,6 +317,8 @@
 		resets = <&reset NXP_SYSCON_RESET(1, 13)>;
 		dmas = <&dma0 8 &dma0 9>;
 		dma-names = "rx", "tx";
+		zephyr,disabling-power-states = <&suspend &standby>;
+		power-domains = <&standby_off_domain>;
 		status = "disabled";
 	};
 
@@ -288,6 +334,8 @@
 		reg = <0x82000 0x1000>;
 		interrupts = <1 1>;
 		dma-channels = <23>;
+		zephyr,disabling-power-states = <&suspend &standby>;
+		power-domains = <&standby_off_domain>;
 		status = "disabled";
 		#dma-cells = <1>;
 	};
@@ -297,6 +345,8 @@
 		reg = <0xa7000 0x1000>;
 		interrupts = <58 1>;
 		dma-channels = <10>;
+		zephyr,disabling-power-states = <&suspend &standby>;
+		power-domains = <&standby_off_domain>;
 		status = "disabled";
 		#dma-cells = <1>;
 	};
@@ -329,6 +379,7 @@
 		resets = <&reset NXP_SYSCON_RESET(1, 0)>;
 		#address-cells = <1>;
 		#size-cells = <0>;
+		zephyr,disabling-power-states = <&suspend &standby>;
 
 		mrt0_channel0: mrt0_channel@0 {
 			compatible = "nxp,mrt-channel";
--- a/soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
+++ b/soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
@@ -8,7 +8,8 @@ if SOC_SERIES_MCXW2XX
 # In case we need to switch to OS timer, then
 # define CONFIG_MCUX_OS_TIMER=y
 configdefault MCUX_OS_TIMER
-	default n
+	default y if PM
+	default n if !PM
 
 config CORTEX_M_SYSTICK
 	default n if MCUX_OS_TIMER
@@ -18,9 +19,14 @@ config NUM_IRQS
 
 DT_SYSCLK_PATH := $(dt_nodelabel_path,sysclk)
 
-config SYS_CLOCK_HW_CYCLES_PER_SEC
-	default 1000000 if MCUX_OS_TIMER
+configdefault SYS_CLOCK_HW_CYCLES_PER_SEC
+	default 32768   if MCUX_OS_TIMER && PM
+	default 1000000 if MCUX_OS_TIMER && !PM
 	default $(dt_node_int_prop_int,$(DT_SYSCLK_PATH),clock-frequency) if CORTEX_M_SYSTICK
+
+configdefault SYS_CLOCK_TICKS_PER_SEC
+	default 1024 if MCUX_OS_TIMER && PM
+	default 1000 if MCUX_OS_TIMER && !PM
 
 # Set to the minimal size of data which can be written.
 config FLASH_FILL_BUFFER_SIZE
@@ -44,5 +50,36 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
 	default 2048
 
 endif # BT
+
+if PM
+
+# Enable PM_DEVICE for Suspend mode and Standby mode,
+# because some devices can't work in these modes.
+configdefault PM_DEVICE
+	default y if "$(dt_nodelabel_enabled,standby)" || "$(dt_nodelabel_enabled,suspend)"
+
+configdefault PM_POLICY_DEVICE_CONSTRAINTS
+	default y if PM_DEVICE
+
+configdefault IDLE_STACK_SIZE
+	default 640
+
+configdefault TICKLESS_KERNEL
+	default y
+
+if BT
+choice PM_POLICY
+	default PM_POLICY_CUSTOM
+endchoice
+endif # BT
+
+endif # PM
+
+if PM_DEVICE
+
+configdefault POWER_DOMAIN
+	default y
+
+endif # PM_DEVICE
 
 endif # SOC_SERIES_MCXW2XX
--- a/soc/nxp/mcx/mcxw/mcxw2xx/power.c
+++ b/soc/nxp/mcx/mcxw/mcxw2xx/power.c
@@ -5,24 +5,223 @@
  */
 
 #include <zephyr/kernel.h>
+#include <zephyr/init.h>
 #include <zephyr/pm/pm.h>
+#include <zephyr/drivers/gpio.h>
+#include <zephyr/drivers/interrupt_controller/nxp_pint.h>
+#include <fsl_power.h>
+#include <zephyr/logging/log.h>
 
-void pm_state_set(enum pm_state state, uint8_t id)
+#if CONFIG_PM_POLICY_CUSTOM
+#include <zephyr/pm/policy.h>
+#endif /* CONFIG_PM_POLICY_CUSTOM */
+
+#ifdef CONFIG_BT
+#include "ll_intf.h"
+#endif /* CONFIG_BT */
+
+LOG_MODULE_DECLARE(soc, CONFIG_SOC_LOG_LEVEL);
+
+/* Wakeup pin. */
+#if CONFIG_GPIO && DT_NODE_EXISTS(DT_NODELABEL(btn_wk))
+#define WAKEUP_PIN_ENABLE 1
+#else
+#define WAKEUP_PIN_ENABLE 0
+#endif
+
+#if WAKEUP_PIN_ENABLE
+#define WAKEUP_BUTTON_NODE DT_NODELABEL(btn_wk)
+static const struct gpio_dt_spec wakeup_pin_dt = GPIO_DT_SPEC_GET(WAKEUP_BUTTON_NODE, gpios);
+#endif /* WAKEUP_PIN_ENABLE */
+
+#if defined(CONFIG_BT) && !defined(CONFIG_PM_POLICY_CUSTOM)
+#error Select CONFIG_PM_POLICY_CUSTOM when CONFIG_BT is selected
+#endif
+
+#if CONFIG_PM_POLICY_CUSTOM
+__weak const struct pm_state_info *pm_policy_next_state(uint8_t cpu, int32_t ticks)
+{
+	uint8_t num_cpu_states;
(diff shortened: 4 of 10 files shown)


soc/nxp/mcx/mcxw/mcxw2xx/power.c
```diff
@@ -6,17 +6,58 @@
...
+}
 
 void pm_state_set(enum pm_state state, uint8_t id)
 {
 	ARG_UNUSED(id);
+	uint32_t exclude_from_pd;
+	uint64_t wakeup_sources;
+
+	/* Set PRIMASK */
+	__disable_irq();
```
Albort12138, line 149: In idle() function, *arch_irq_lock()* already invoked to control global interrupts.

ZhaoxiangJin: If mcxw23 arch is `ARMV7_M`/`ARMV8_M `mainline, `__disable_irq();` is required here.

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
```diff
@@ -4,6 +4,9 @@
 
 if SOC_SERIES_MCXW2XX
 
+config MCUX_OS_TIMER
```
ZhaoxiangJin: If PM enabled, ostimer will be the system timer? The OS timer clock frequency is 1MHz, that means the system in a low performance at active mode.
Ideally, the best approach is to use a combination of systick and ostimer, with ostimer used in low-power mode and systick used in active mode. see https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.cortex_m_systick#L46

drivers/timer/mcux_os_timer.c
```diff
@@ -179,12 +179,13 @@ static uint32_t mcux_lpc_ostick_compensate_system_timer(void)
...
-	if (IS_ENABLED(CONFIG_MCUX_OS_TIMER_PM_POWERED_OFF)) {
-		/* Reset the OS Timer to a known state */
-		RESET_PeripheralReset(kOSEVENT_TIMER_RST_SHIFT_RSTn);
-		/* Reactivate os_timer for cases where it loses its state */
-		OSTIMER_Init(base);
-	}
+
+#if defined(CONFIG_MCUX_OS_TIMER_PM_POWERED_OFF)
+	/* Reset the OS Timer to a known state */
+	RESET_PeripheralReset(kOSEVENT_TIMER_RST_SHIFT_RSTn);
```
ZhaoxiangJin: Outside this PR scope: calling the reset API would be the most appropriate approach here.

ZhaoxiangJin:
It is better to enable [power_mgmt_soc](https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/subsys/pm/power_mgmt_soc) test

drivers/timer/mcux_os_timer.c
zejiang0jason: yes, that is the right way.

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
zejiang0jason: good suggestion. I am not clear is, the OS tick frequency is defined by: `CONFIG_SYS_CLOCK_TICKS_PER_SEC`, even the OSTIMER input clock frequency is slow, the OS tick can also be quick (100Hz ~ 1000Hz). What is the reason that the system performance will be low?

soc/nxp/mcx/mcxw/mcxw2xx/power.c
```diff
@@ -6,17 +6,58 @@
...
+
+static void pm_get_lowpower_resource_list(uint32_t *exclude_from_pd, uint64_t *wakeup_sources)
+{
+	*exclude_from_pd = kLOWPOWERCFG_DCDC | kLOWPOWERCFG_BOD1;
+	*wakeup_sources  = 0;
+
+#ifdef CONFIG_BT
+	/* BT needs 32kHz clock. */
+	*exclude_from_pd |= (kLOWPOWERCFG_XTAL32K | kLOWPOWERCFG_BLE_WUP);
+	*wakeup_sources  |= (kWAKEUP_BLE_LL | kWAKEUP_BLE_SLP_TMR);
```
yeaissa: Adding kWAKEUP_BLE_LL to the wakeup_sources wont allow you to enter power down. The driver will throw kStatus_InvalidArgument error for this, please check
https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/devices/MCX/MCXW/MCXW236/drivers/fsl_power.c#L1572
You can remove those wakeup_sources as we don't need them.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
yeaissa, line 1: This won't be working for BLE. The time the system can go to sleep is derived from the OS idle time and the BLEController idle time.
Here the state selected by the default policy is only considering the OS idle time, exit, min-residency latencies. 
A new policy shall be implemented to consider the controller IDLE time, and select the appropriate state depending on that.

yeaissa (commented):
Lets discuss this together, I see some missing points that needs to be considered for BLE.

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
ZhaoxiangJin: When SysTick is configured as the system clock, `SYS_CLOCK_HW_CYCLES_PER_SEC`is 32000000; when OSTimer is configured as the system clock, it is 1000000.

zejiang0jason: `CONFIG_SYS_CLOCK_TICKS_PER_SEC` is 100, `SYS_CLOCK_HW_CYCLES_PER_SEC` is the clock source frequency.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
Albort12138: you mean arch_irq_lock() do not support ARMV7_M/ARMV8_M?

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
ZhaoxiangJin: You are correct, `SYS_CLOCK_TICKS_PER_SEC `defines the tick frequency, and 1MHz is sufficient for a 100Hz(non-tickless)/10000Hz(tickless) tick frequency. Different `SYS_CLOCK_HW_CYCLES_PER_SEC `values only affect the timing resolution of hardware cycles, which in turn affects applications that require high-precision timestamps, but do not affect performance.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
zejiang0jason: thanks, updated, considering BLE when select the next state.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
ZhaoxiangJin: For `ARMV7_M`/`ARMV8_M` mainline, the `arch_irq_lock()` implementation is  `__set_BASEPRI_MAX(_EXC_IRQ_DEFAULT_PRIO);` while `ARMV6_M`/`ARMV8_M `baseline is `__disable_irq();`. So for  `ARMV7_M`/`ARMV8_M` mainline, we need `__disable_irq();` to set PRIMASK before configuring BASEPRI to prevent interruption before wake-up, just like [this](https://github.com/zephyrproject-rtos/zephyr/blob/main/arch/arm/core/cortex_m/cpu_idle.c#L81).

soc/nxp/mcx/mcxw/mcxw2xx/power.c
zejiang0jason: Hi @yeaissa , do you mean both `kWAKEUP_BLE_LL` and `kWAKEUP_BLE_SLP_TMR` are not needed?

soc/nxp/mcx/mcxw/mcxw2xx/power.c
```diff
@@ -6,17 +6,139 @@
...
+		return NULL;
+	}
+#endif
+
+#ifdef CONFIG_BT
+	if (bt_is_ready())
+	{
+		uint32_t remaining_time;
+		ble_stat_t stat;
+		stat = ll_intf_le_get_remaining_time_for_next_event_unsafe(&remaining_time);
```
yeaissa: Please Use BLEController_GetRemainingTimeForNextEventUnsafe instead.

dts/arm/nxp/nxp_mcxw23x_common.dtsi
```diff
@@ -32,12 +32,55 @@
...
+				exit-latency-us = <16>;
+				zephyr,pm-device-disabled;
+			};
+
+			/* Suspend mode maps to Deep Sleep mode. */
+			suspend: suspend {
+				compatible = "zephyr,power-state";
+				power-state-name = "suspend-to-idle";
+				min-residency-us = <1300>;
+				exit-latency-us = <400>;
```
yeaissa: I question this value, on SDK code this is set to 749.
https://github.com/nxp-mcuxpresso/mcux-devices-mcx/blob/main/MCXW/MCXW236/drivers/power_manager_mcxw23/device/fsl_pm_device.c#L31

yeaissa: Can we check the other states too

dts/arm/nxp/nxp_mcxw23x_common.dtsi
```diff
@@ -32,12 +32,55 @@
...
+			idle: idle {
+				compatible = "zephyr,power-state";
+				power-state-name = "runtime-idle";
+				min-residency-us = <60>;
+				exit-latency-us = <16>;
+				zephyr,pm-device-disabled;
+			};
+
+			/* Suspend mode maps to Deep Sleep mode. */
+			suspend: suspend {
```
yeaissa, line 54: Deep sleep shall point to standby (This is the case for other NXP SoCs).
mcux_os_timer driver only supports the standby state in case of low power/wake up timer. please check:
https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/mcux_os_timer.c#L97
Since we shall support other states that needs os_timer wakeup, we may need to update the driver by adding the higher states that requires this feature. @mmahadevan108 any idea?

dts/arm/nxp/nxp_mcxw23x_common.dtsi
yeaissa, line 1: I see os_timer state is still set to `disabled`, please remove.
I noticed that this init was missing and the timer was not behaving as expected:
https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/nxp/mcx/mcxw/mcxw2xx/soc.c#L100

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
yeaissa, line 1: CONFIG_TICKLESS_MODE is disabled by default, please enable, otherwise the system will wakeup on each tick.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
yeaissa: Thanks

dts/arm/nxp/nxp_mcxw23x_common.dtsi
zejiang0jason: The code is added because ostimer can't work in standby mode for some SOCs, but it work in MCXW23 deep sleep and power down mode, so that code is not necessary.

soc/nxp/mcx/mcxw/mcxw2xx/power.c
Albort12138: In power driver, DisableGlobalIRQ() already used, so maybe not need to invoke __disableIRQ()

dts/arm/nxp/nxp_mcxw23x_common.dtsi
zejiang0jason: because of this setting: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.mcux_os#L8, once os_timer is enabled in device tree, it will be set to `y` in kconfig, and be used as OS tick. I think it is better that only use it as OS tick in when power management is used, so my proposed way is only enable os_timer using dts overlay in specific project.

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
zejiang0jason: enabled in [soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig](https://github.com/zephyrproject-rtos/zephyr/pull/98986/files#diff-32fc40b35cec6400cb7a1a8edd4bd32c88838d8e78e8bdf3b0fc3ec6fe1e0244) when PM is used.

drivers/timer/mcux_os_timer.c
```diff
@@ -142,12 +143,12 @@ static uint32_t mcux_lpc_ostick_set_counter_timeout(int32_t curr_timeout)
 	}
 	counter_running = true;
 
-	if (IS_ENABLED(CONFIG_MCUX_OS_TIMER_PM_POWERED_OFF)) {
-		/* Capture the current timer value for cases where it loses its state
-		 * in low power modes.
-		 */
-		cyc_sys_compensated += OSTIMER_GetCurrentTimerValue(base);
-	}
+#if defined(CONFIG_MCUX_OS_TIMER_PM_POWERED_OFF)
```
ZhaoxiangJin: I'd like to know why the code needs to change?

drivers/dma/dma_mcux_lpc.c
```diff
@@ -948,6 +976,69 @@ static int dma_mcux_lpc_get_attribute(const struct device *dev, uint32_t type, u
 	return 0;
 }
 
+#if CONFIG_PM
```
ZhaoxiangJin: should be `#if CONFIG_PM_DEVICE` instead `#if CONFIG_PM`

drivers/interrupt_controller/intc_nxp_pint.c
```diff
@@ -175,6 +175,28 @@ void nxp_pint_pin_unset_callback(uint8_t pin)
 	pint_irq_cfg[slot].callback = NULL;
 }
 
+/**
+ * @brief Get PINT slot index the pin is allocated to
+ *
+ * @param pin: The pin to get the PINT slot index for
+ * @return The allocated slot index, if not allocated, return -1
+ */
```
ZhaoxiangJin: same doxygen comments (compare with the .h) can remove.

drivers/interrupt_controller/intc_nxp_pint.c
```diff
@@ -175,6 +175,28 @@ void nxp_pint_pin_unset_callback(uint8_t pin)
...
+int nxp_pint_pin_get_slot_index(uint8_t pin)
+{
+	int slot;
+
+	if (pin > ARRAY_SIZE(pin_pint_id)) {
+		return -1;
```
ZhaoxiangJin: `return -EINVAL;` is better maybe.

boards/nxp/frdm_mcxw23/doc/index.rst
```diff
@@ -198,6 +198,21 @@ should see the following message in the terminal:
    *** Booting Zephyr OS build v4.2.0-2105-g48f2ffda26de ***
    Hello World! frdm_mcxw23/mcxw236
 
+Power Management
+================
+
+When Power Management is enabled, OSTIMER is used as OS tick timer.
```
ZhaoxiangJin: what about: When Power Management :kconfig:option:`CONFIG_PM` is enabled. add kconfig option makes the information clearer.

drivers/timer/mcux_os_timer.c
zejiang0jason: I checked the usage of `IS_ENABLED`, the original implementation is the right way, will revert back.

zejiang0jason: Moved the ostimer change to a separate PR: https://github.com/zephyrproject-rtos/zephyr/pull/100134

boards/nxp/frdm_mcxw23/doc/index.rst
```diff
@@ -198,6 +198,21 @@ should see the following message in the terminal:
    *** Booting Zephyr OS build v4.2.0-2105-g48f2ffda26de ***
    Hello World! frdm_mcxw23/mcxw236
 
+Power Management
```
yeaissa, line 201: I'm not ok with this, since it will require adding the overlay on each app that will run with PM. I expect to have everything I need once I enable CONFIG_PM.
To avoid this lets set the `os_timer` status to `okay` by default in the nxp_mcxw23x_common.dtsi file, and change the MCUX_OS_TIMER redefinition in Kconfig.defconfig to:

```
 config MCUX_OS_TIMER
-       default y if PM
+       default n if !PM
```

This will allow to run on the CORTEX_M_SYSTICK when PM is not defined and switch to OS_TIMER when PM is enabled.

soc/nxp/mcx/mcxw/mcxw2xx/Kconfig.defconfig
```diff
@@ -4,6 +4,13 @@
 
 if SOC_SERIES_MCXW2XX
 
+# MCXW23 uses SYSTICK timer as the kernel timer by default.
+# In case we need to switch to OS timer, then
+# define CONFIG_MCUX_OS_TIMER=y
+config MCUX_OS_TIMER
```
ZhaoxiangJin: should be `configdefault MCUX_OS_TIMER`, for the other redefined Kconfig options in this file, corresponding updates should be made as well.

zejiang0jason: done here: https://github.com/zephyrproject-rtos/zephyr/pull/100270

yeaissa (approved):

soc/nxp/mcx/mcxw/mcxw2xx/power.c
zejiang0jason: it is better to keep Zephyr code completed, no matter how hal_nxp has done.

drivers/dma/dma_mcux_lpc.c
```diff
@@ -948,6 +976,69 @@ static int dma_mcux_lpc_get_attribute(const struct device *dev, uint32_t type, u
...
+	for (uint32_t channel = 0; channel < config->num_of_channels; channel++) {
+		if (dma_data->channel_index[channel] != -1) {
+			p_channel_data = &dma_data->channel_data[virtual_channel];
+
+			p_channel_data->backup_reg.xfercfg = dma_base->CHANNEL[channel].XFERCFG;
+			p_channel_data->backup_reg.cfg     = dma_base->CHANNEL[channel].CFG;
+			virtual_channel++;
+		}
+	}
+}
```
JimmyChen-NXP: To restore/backup the register context, is it a special case for the DMA peripheral or it is common case for all peripherals? If it apply for all other peripherals, it looks need either a common functionality or it is really necessary?

dts/arm/nxp/nxp_mcxw23x_common.dtsi
```diff
@@ -32,12 +32,47 @@
...
+				power-state-name = "standby";
+				min-residency-us = <2000>;
+				exit-latency-us = <1345>;
+			};
+		};
+	};
+
+	/* For the resource which will be off and lose content. */
+	standby_off_domain: standby-off-domain {
+		compatible = "power-domain-soc-state-change";
```
Albort12138, line 73: Please note that power_domain_soc_state_change is likely only useful when PM_DEVICE_SYSTEM_MANAGED is enabled.
Reference https://github.com/zephyrproject-rtos/zephyr/blob/enable-cmc-power-domain/drivers/power_domain/power_domain_soc_state_change.c#L50
This function relies on knowing the next allowed system state. However, when PM_DEVICE_SYSTEM_MANAGED is disabled, the next state after a power domain is suspended could still be active, which makes the logic less reliable in that scenario.

zejiang0jason: > when PM_DEVICE_SYSTEM_MANAGED is disabled, the next state after a power domain is suspended could still be active

Could you please share more details?

dts/arm/nxp/nxp_mcxw23x_common.dtsi
```diff
@@ -32,12 +32,47 @@
...
+				min-residency-us = <60>;
+				exit-latency-us = <16>;
+				zephyr,pm-device-disabled;
+			};
+
+			/* Suspend mode maps to Deep Sleep mode. */
+			suspend: suspend {
+				compatible = "zephyr,power-state";
+				power-state-name = "suspend-to-idle";
+				min-residency-us = <1300>;
```
mmahadevan108, line 57: How did we get the residency and exit-latency?

soc/nxp/mcx/mcxw/mcxw2xx/poweroff.c
```diff
@@ -0,0 +1,34 @@
...
+#endif /* GPIO_ACTIVE_LOW */
+
+#else
+#define POWEROFF_WAKEUP kWAKEUP_PIN_DISABLE
+#endif /* CONFIG_GPIO && DT_NODE_EXISTS(DT_NODELABEL(btn_wk)) */
+
+void z_sys_poweroff(void)
+{
+	POWER_EnterPowerOff(0, POWEROFF_WAKEUP);
+
```
mmahadevan108, line 32: How was this tested?

dts/arm/nxp/nxp_mcxw23x_common.dtsi
zejiang0jason: from MCUX SDK code and also chip datasheet, and choose the larger one from them:
https://github.com/nxp-mcuxpresso/mcux-devices-mcx/blob/main/MCXW/MCXW236/drivers/power_manager_mcxw23/device/fsl_pm_device.c
image

soc/nxp/mcx/mcxw/mcxw2xx/poweroff.c
zejiang0jason: yes, test code:
```c
#include <stdio.h>
#include <zephyr/sys/poweroff.h>

int main(void)
{
    printf("Hello World! %s\n", CONFIG_BOARD_TARGET);

    sys_poweroff();

    return 0;
}
```
build command: `west build -p always samples/hello_world/ -b frdm_mcxw23 -DCONFIG_POWEROFF=y`.

press SW5, the chip is wakeup from poweroff, it reset and print "hello world" again.

mmahadevan108 (approved):

ZhaoxiangJin (approved):

PR #98988, opened 2025-11-06
net: zperf: udp upload clock compensate

fengming-ye:
When clock accuracy is 1KHz or lower, udp_upload rate has much lower accuracy. eg. 20M rate has 60M actual throughput. It's because any packet_duration less than 1000us will be treated as 1 tick. In this case rate cannot control throughput.

Add clock compensate when time clock is lower than 1KHz to fix it. In every 10 ticks or longer delta time, calculate packets sent actual number and expected number, and then compensate ticks. And add compensate ticks to delay ticks count.

If we cannot compensate a minus ticks as delay cannot be negative, add it to next compensate period.
If compensate sum ticks is lower than -1000, it means current rate is higher than low level capability. So no need to compensate in remaining traffic.

### DIFF
--- a/subsys/net/lib/zperf/zperf_udp_uploader.c
+++ b/subsys/net/lib/zperf/zperf_udp_uploader.c
@@ -155,6 +155,77 @@ static inline int zperf_upload_fin(int sock,
 	return 0;
 }
 
+#define USECS_PER_TICK (Z_HZ_us / Z_HZ_ticks)
+#if (USECS_PER_TICK >= 1000)
+#define ZPERF_UDP_UPLOAD_CLOCK_COMPENSATE
+#endif
+
+#ifdef ZPERF_UDP_UPLOAD_CLOCK_COMPENSATE
+struct compensate_ctx {
+	int period;
+	int64_t period_start;
+	int packet_duration_us;
+	int actual_pkts;
+	int compensate;
+};
+
+/**
+ * Add compensate to packet delay when time clock accuracy is lower than 1kHz.
+ * After given period, compare actual sent packets with expected sent packets,
+ * and get summary compensate ticks.
+ * Then try to compensate in this loop.
+ * If delay is not enough as it cannot be less than 0, pile up compensate ticks.
+ */
+static int cal_compensate_delay(struct compensate_ctx *ctx, int64_t loop_time, int delay)
+{
+	int64_t delta_time;
+	int expected_pkts;
+	int compensate_pkts;
+	int compensate_ticks;
+	int compensate_delay = delay;
+
+	if (ctx->period == 0) {
+		return delay;
+	}
+
+	if (ctx->period_start == -1) {
+		ctx->period_start = loop_time;
+		ctx->actual_pkts = 0;
+	}
+
+	ctx->actual_pkts++;
+	delta_time = loop_time - ctx->period_start;
+
+	if (delta_time < ctx->period) {
+		return delay;
+	}
+
+	/* calculate compensate ticks and maintain it during whole traffic */
+	expected_pkts = delta_time * USECS_PER_TICK / ctx->packet_duration_us;
+	compensate_pkts = ctx->actual_pkts - expected_pkts;
+	compensate_ticks = compensate_pkts * ctx->packet_duration_us / USECS_PER_TICK;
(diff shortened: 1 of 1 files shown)


subsys/net/lib/zperf/zperf_udp_uploader.c
```diff
@@ -268,7 +316,7 @@ static int udp_upload(int sock, int port,
 		if (IS_ENABLED(CONFIG_NET_ZPERF_LOG_LEVEL_DBG)) {
 			if (print_time >= loop_time) {
 				NET_DBG("nb_packets=%u\tdelay=%u\tadjust=%d",
-					nb_packets, (unsigned int)delay,
+					nb_packets, (unsigned int)compensate_delay,
```
rlubos, line 360: I don't think there's a need for yet another `compensate_delay` variable, just update the `delay` value if needed. There'd be no need for the `else` branches above then.

subsys/net/lib/zperf/zperf_udp_uploader.c
```diff
@@ -222,6 +235,41 @@ static int udp_upload(int sock, int port,
 			delay = 0U; /* delay should never be negative */
 		}
 
+		/* compensate delay ticks when clock accuracy is lower than 10KHz */
+		if (tick_us >= 1000 && period) {
+			actual_pkts++;
```
rlubos: I think this compensation functionality should be encapsulated into a helper function, `udp_upload()` is already pretty lengthy, implementing this compensation algorithm inline just blurs it even more. I'd suggest something like:
```c
if (tick_us >= ZPERF_COMPENSATION_THRESHOLD) {
	delay = compensate_delay(&compensate_ctx, loop_time, delay, ... /* whatever else's needed */).
}
```
Where all the details regarding delay compensation would be encapsulated within the funtcion. `compensate_ctx` would be a struct containig any data that needs to remain persistent between loop iterations.

fengming-ye: Wrapped newly added logic to a function.
But I cannot merge compensate veriable to "delay". Because delay accumulates feedback from every loop to get next packet send time. And it works well except for clock accuracy issue. I didn;t come up with an easy way to merge them without breaking this logic.

subsys/net/lib/zperf/zperf_udp_uploader.c
```diff
@@ -222,6 +300,14 @@ static int udp_upload(int sock, int port,
 			delay = 0U; /* delay should never be negative */
 		}
 
+		/* add clock compensate to packet delay when clock accuracy is lower than 1KHz */
+#ifdef ZPERF_UDP_UPLOAD_CLOCK_COMPENSATE
+		compensate_delay = cal_compensate_delay(loop_time, (int)packet_duration_us,
+			(int)delay, &period_start, &period, &compensate_ticks);
```
rlubos: Please fix the argument alignment:
```suggestion
		compensate_delay = cal_compensate_delay(loop_time, (int)packet_duration_us,
						        (int)delay, &period_start, &period, 
						        &compensate_ticks);
```

subsys/net/lib/zperf/zperf_udp_uploader.c
```diff
@@ -155,6 +155,77 @@ static inline int zperf_upload_fin(int sock,
...
+ * If delay is not enough as it cannot be less than 0, pile up compensate ticks.
+ */
+static int cal_compensate_delay(int64_t loop_time,
+				int packet_duration_us,
+				int delay,
+				int64_t *period_start,
+				int *period,
+				int *compensate)
+{
+	static int actual_pkts;
```
rlubos: There shouldn't be static variables in functions like this, if two UDP uploads ran in parallel they'd interfere with each other. The variable should be defined in the `udp_upload()` context, and passed as a parameter to the function just like others, that's why I originally suggested to group all session-related data in a `struct compensate_context` that'd be declared within `udp_upload()` and passed as a parameter (via pointer) to `cal_compensate_delay()` instead of a long list of individual variables.

fengming-ye: updated as required

rlubos (approved):

jukkar (approved):

PR #98990, opened 2025-11-06
llext: export symbols by groups

JordanYates:
Add the concept of "groups" to exported symbols, that allows including or excluding those groups from the exported symbol list through Kconfig.

Alternative to #98890

### DIFF
--- a/doc/services/llext/config.rst
+++ b/doc/services/llext/config.rst
@@ -126,6 +126,44 @@ LLEXT subsystem to optimize memory footprint in this case.
            information on this topic is available on GitHub issue `#75341
            <https://github.com/zephyrproject-rtos/zephyr/issues/75341>`_.
 
+.. _llext_symbol_groups:
+
+Symbol Groups
+-------------
+
+All LLEXT symbols belong to a group, with the inclusion of each group in the
+exported symbol table controlled by a corresponding Kconfig symbol. Exporting
+a symbol as part of a group is done with the :c:macro:`EXPORT_GROUP_SYMBOL`
+and :c:macro:`EXPORT_GROUP_SYMBOL_NAMED` macros. For example the following
+exports the symbol ``memcpy`` as part of the ``LIBC`` group:
+
+.. code:: c
+
+   EXPORT_GROUP_SYMBOL(LIBC, memcpy);
+
+Group names are arbitrary, but they must be all uppercase. For each group
+used in C code, there **MUST** be a corresponding Kconfig symbol of the form:
+
+.. code::
+
+   config LLEXT_EXPORT_SYMBOL_GROUP_{GROUP_NAME}
+      bool "Export all symbols from the {GROUP_NAME} group"
+
+The default group for symbols (those declared with :c:macro:`EXPORT_SYMBOL`
+or :c:macro:`EXPORT_SYMBOL_NAMED`) is the ``UNASSIGNED`` group. As per the
+above rules, the inclusion of this group is controlled by
+:kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_UNASSIGNED`.
+
+The groups currently defined by Zephyr are:
+
+.. csv-table:: Zephyr LLEXT symbol groups
+  :header: Group Name, Kconfig Symbol, Description
+
+  ``UNASSIGNED``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_UNASSIGNED`, Symbols without an explicit group
+  ``SYSCALL``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_SYSCALL`, Zephyr kernel system calls
+  ``LIBC``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_LIBC`, C standard library functions (:c:func:`memcpy` etc)
+  ``DEVICE``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_DEVICE`, Devicetree devices
+
 .. _llext_kconfig_slid:
 
 Using SLID for symbol lookups
--- a/include/zephyr/llext/symbol.h
+++ b/include/zephyr/llext/symbol.h
@@ -8,6 +8,7 @@
 #define ZEPHYR_LLEXT_SYMBOL_H
 
 #include <zephyr/sys/iterable_sections.h>
+#include <zephyr/sys/util_macro.h>
 #include <zephyr/toolchain.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -155,19 +156,33 @@ struct llext_symtable {
 #endif
 /** @endcond */
 
+#if defined(LL_EXTENSION_BUILD)
+/* Extension build: No Kconfig conditionals */
+#define Z_EXPORT_SYMBOL_NAMED_IN_GROUP(group, sym_ident, sym_name)		\
+	Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)
+#else
+/* LLEXT application: Export if relevant group is enabled */
+#define Z_EXPORT_SYMBOL_NAMED_IN_GROUP(group, sym_ident, sym_name)		\
+	IF_ENABLED(CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_ ## group,			\
+		(Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)))
+#endif /* defined(LL_EXTENSION_BUILD) */
+
 /**
  * @brief Export a constant symbol from the current build with a custom name
  *
  * Version of @ref EXPORT_SYMBOL that allows the user to specify a custom name
- * for the exported symbol.
+ * for the exported symbol. The symbol is part of the UNASSIGNED group.
  *
  * When @c CONFIG_LLEXT is not enabled, this macro is a no-op.
+ *
+ * When @c CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_UNASSIGNED  is not enabled, this
+ * macro is a no-op.
  *
  * @param sym_ident Symbol to export
  * @param sym_name Name associated with the symbol
  */
 #define EXPORT_SYMBOL_NAMED(sym_ident, sym_name)				\
-	Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)
+	Z_EXPORT_SYMBOL_NAMED_IN_GROUP(UNASSIGNED, sym_ident, sym_name)
 
 /**
  * @brief Export a constant symbol from the current build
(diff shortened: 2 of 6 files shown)


subsys/llext/llext_export.c
pillo79, line 68: These should be in a TOOLCHAIN group maybe?

subsys/llext/Kconfig
```diff
@@ -10,6 +10,30 @@ menuconfig LLEXT
 
 if LLEXT
 
+config LLEXT_EXPORT_SYMBOL_GROUP_DEFAULT
```
pillo79: Nit: I suggest we move these at the end, grouped in an additional `menu "Exported symbol groups"` (or something like that).

subsys/llext/Kconfig
```diff
@@ -10,6 +10,36 @@ menuconfig LLEXT
 
 if LLEXT
 
+config LLEXT_EXPORT_DEFAULT
```
pillo79: Nit: IMHO reduces confusion with `SYMBOL_GROUP_DEFAULT`
```suggestion
config LLEXT_EXPORT_DEFAULT_GROUPS
```

pillo79 (commented):
Love this idea, backwards compatible and modular/extensible in the future. Will also use the same in the Arduino module, so +2 from me! 🙂
A few suggestions below.

subsys/llext/llext_export.c
JordanYates: Is there a Kconfig symbol that can be used to check when these are available?

subsys/llext/Kconfig
JordanYates: Yea the naming is not great. Maybe using `ZEPHYR` as the default group instead of `DEFAULT`?

teburd (dismissed):
Thank you for this, love the idea. The "DEFAULT" naming is a bit odd though I tend to agree with others, not sure I have a better answer though so we can move on.

subsys/llext/Kconfig
pillo79: Personally I have no issue with the "DEFAULT" group name; I was just confused by the "default groups" config looking like _the default group_ itself. 
If you fancy changing `SYMBOL_GROUP_DEFAULT`, to make the difference even more clear, I would probably vote for `SYMBOL_GROUP_NONE`. The plain truth :grin:

subsys/llext/llext_export.c
pillo79: I don't use that toolchain, so I don't really know. TBH also leaving them in the common group has merit - they seem to be required by that toolchain. Ignore this comment :+1:

include/zephyr/llext/symbol.h
```diff
@@ -155,33 +156,86 @@ struct llext_symtable {
...
+ *
+ * Version of @ref EXPORT_SYMBOL_NAMED that allows placing the symbol in a
+ * custom group.
+ *
+ * When @c CONFIG_LLEXT is not enabled, this macro is a no-op.
+ *
+ * When @c CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_{group} is not enabled, this macro
+ * is a no-op.
+ *
+ * @param group Group of the symbol
```
mathieuchopstm: Reword suggestion + add note about uppercase requirement
```suggestion
 * @param group Group in which symbol is exported (must be uppercase)
```

include/zephyr/llext/symbol.h
```diff
@@ -155,33 +156,86 @@ struct llext_symtable {
...
+ *
+ * Version of @ref EXPORT_SYMBOL that allows placing the symbol in a
+ * custom group.
+ *
+ * When @c CONFIG_LLEXT is not enabled, this macro is a no-op.
+ *
+ * When @c CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_{group} is not enabled, this macro
+ * is a no-op.
+ *
+ * @param group Group of the symbol
```
mathieuchopstm:
```suggestion
 * @param group Group in which symbol is exported (must be uppercase)
```

subsys/llext/Kconfig
```diff
@@ -154,6 +155,40 @@ config LLEXT_EXPERIMENTAL
 	  Include support for LLEXT experimental and unstable functionality that
 	  has a very high likelihood to change in the future.
 
+menu "Exported symbol groups"
+
+config LLEXT_EXPORT_DEFAULT_GROUPS
```
mathieuchopstm, line 160: Tiny nit: we may want to reserve a prefix for Zephyr groups (e.g., "application is not allowed to create groups starting with `Z_`")

But just documenting the OS-reserved groups in docs is fine by me too. (It's *slightly* less future proof, but I doubt the feature will be used for custom groups by applications much tbf)

mathieuchopstm (commented):
Minor details but LGTM overall.

Maybe the documentation should be extended? 🙂

Especially since my understanding is that *by default groups are not exported*, so an application trying to create custom groups is responsible for creating the corresponding Kconfig symbols beforehand or it'll have a nasty surprise.

For the default group... how about `UNASSIGNED`? (thinking about the documentation: "if a symbol is not explicitly assigned to a group, it will be placed in the `UNASSIGNED` group") I do agree `DEFAULT` is kinda bad

include/zephyr/llext/symbol.h
```diff
@@ -155,33 +156,86 @@ struct llext_symtable {
...
 
+#if defined(LL_EXTENSION_BUILD)
+/* Extension build: No Kconfig conditionals */
+#define Z_EXPORT_SYMBOL_NAMED_IN_GROUP(group, sym_ident, sym_name)		\
+	Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)
+#else
+/* LLEXT application: Export if relevant group is enabled */
+#define Z_EXPORT_SYMBOL_NAMED_IN_GROUP(group, sym_ident, sym_name)		\
+	IF_ENABLED(CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_ ## group,			\
+		(Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)))
```
mathieuchopstm, line 167: Actually, one remark I made in [this comment](https://github.com/zephyrproject-rtos/zephyr/pull/98890#issuecomment-3492343266):
```suggestion
	COND_CODE_1(CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_ ## group,			\
		(Z_EXPORT_SYMBOL_NAMED(sym_ident, sym_name)),                   \
		(z_export_symbol_in_discarded_symbol_table(...))
```

But I won't block for this since it's more of a QoL feature - we can create an improvement issue to track this when PR is merged, and I'd be fine with implementing it myself (gives me a good excuse to get my hands dirty in LLEXT again... reminds me there's a cleanup I've been meaning to do since long but haven't gotten around to 🙂).

JordanYates: That feature has quite a bit to it, since it would also need python tooling etc to examine the output files. I was going to leave it for others who would find value in the feature to implement it :)

pillo79 (approved):

teburd (approved):

tbursztyka (approved):

doc/services/llext/config.rst
```diff
@@ -126,6 +126,44 @@ LLEXT subsystem to optimize memory footprint in this case.
...
+exported symbol table controlled by a corresponding Kconfig symbol. Exporting
+a symbol as part of a group is done with the :c:macro:`EXPORT_GROUP_SYMBOL`
+and :c:macro:`EXPORT_GROUP_SYMBOL_NAMED` macros. For example the following
+exports the symbol ``memcpy`` as part of the ``LIBC`` group:
+
+.. code:: c
+
+   EXPORT_GROUP_SYMBOL(LIBC, memcpy);
+
+Group names are arbitrary, but they must be all uppercase. For each group
```
mathieuchopstm, line 144: Nit:
```suggestion
Group names are arbitrary but they must be all uppercase. For each group
```

doc/services/llext/config.rst
```diff
@@ -126,6 +126,44 @@ LLEXT subsystem to optimize memory footprint in this case.
...
+used in C code, there **MUST** be a corresponding Kconfig symbol of the form:
+
+.. code::
+
+   config LLEXT_EXPORT_SYMBOL_GROUP_{GROUP_NAME}
+      bool "Export all symbols from the {GROUP_NAME} group"
+
```
mathieuchopstm, line 151: I'd add a
```suggestion
.. warning::

  The build system does not detect when a group is used in C code but has no
  corresponding Kconfig symbol. All symbols part of such groups will be silently
  dropped from the export table, as if the Kconfig symbol *was* defined but had
  been assigned value `n`.
```
to really get the point across in big scary red box 🙂

doc/services/llext/config.rst
```diff
@@ -126,6 +126,44 @@ LLEXT subsystem to optimize memory footprint in this case.
...
+The groups currently defined by Zephyr are:
+
+.. csv-table:: Zephyr LLEXT symbol groups
+  :header: Group Name, Kconfig Symbol, Description
+
+  ``UNASSIGNED``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_UNASSIGNED`, Symbols without an explicit group
+  ``SYSCALL``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_SYSCALL`, Zephyr kernel system calls
+  ``LIBC``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_LIBC`, C standard library functions (:c:func:`memcpy` etc)
+  ``DEVICE``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_DEVICE`, Devicetree devices
+
```
mathieuchopstm, line 166: Maybe:

```suggestion

.. note::

  Except for the ``UNASSIGNED`` group, these groups are reserved for internal
  usage by Zephyr; applications must not place any symbol in these groups.
```

doc/services/llext/config.rst
```diff
@@ -126,6 +126,44 @@ LLEXT subsystem to optimize memory footprint in this case.
...
+The groups currently defined by Zephyr are:
+
+.. csv-table:: Zephyr LLEXT symbol groups
+  :header: Group Name, Kconfig Symbol, Description
+
+  ``UNASSIGNED``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_UNASSIGNED`, Symbols without an explicit group
+  ``SYSCALL``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_SYSCALL`, Zephyr kernel system calls
+  ``LIBC``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_LIBC`, C standard library functions (:c:func:`memcpy` etc)
+  ``DEVICE``, :kconfig:option:`CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_DEVICE`, Devicetree devices
```
mathieuchopstm, line 165: Super pedantic nit: you can define devices that do not correspond to Devicetree nodes 🙂 (i.e., `DEVICE_DEFINE(...)`)

I'm fine with the current wording though; feel free to keep as is, or update it for more accuracy (e.g., `Zephyr device objects`).

mathieuchopstm (commented):
LGTM in current state; some suggestions:

PR #99010, opened 2025-11-06
net: lwm2m: get cache free slots

jvermillard:
Add function to query LwM2M cache free slots on a given path.

This is usefull to detect if the buffer start to be full and flush the content using lwm2m_send

### DIFF
--- a/include/zephyr/net/lwm2m.h
+++ b/include/zephyr/net/lwm2m.h
@@ -29,6 +29,7 @@
 
 #include <time.h>
 #include <zephyr/kernel.h>
+#include <zephyr/types.h>
 #include <zephyr/sys/mutex.h>
 #include <zephyr/net/coap.h>
 #include <zephyr/net/lwm2m_path.h>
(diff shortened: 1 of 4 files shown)


include/zephyr/net/lwm2m.h
```diff
@@ -1619,6 +1620,15 @@ int lwm2m_enable_cache(const struct lwm2m_obj_path *path, struct lwm2m_time_seri
...
 			   lwm2m_cache_filter_cb_t filter_cb);
 
+/**
+ * @brief Return number of free slots in a cached resource.
+ *
+ * @param path LwM2M path to the cached resource.
+ *
+ * @return Number of free slots, or negative errno code in case of error.
+ */
+ssize_t lwm2m_cache_free_slots_get(const struct lwm2m_obj_path *path);
```
rlubos: Theoretically `ssize_t` only guarantees to be capable of storing `-1` as a negative value, so if the function is supposed to return negative error codes it should return regular int.

subsys/net/lib/lwm2m/lwm2m_registry.h
```diff
@@ -238,5 +238,6 @@ bool lwm2m_cache_write(struct lwm2m_time_series_resource *cache_entry,
 bool lwm2m_cache_read(struct lwm2m_time_series_resource *cache_entry,
 		      struct lwm2m_time_series_elem *buf);
 size_t lwm2m_cache_size(const struct lwm2m_time_series_resource *cache_entry);
+ssize_t lwm2m_cache_free_slots_get(const struct lwm2m_obj_path *path);
```
rlubos: The function is already declared in the public header, so I don't think there's a need to redeclare it in a private one.

rlubos (commented):
The doc build issue seems to be inherited from main, the PR will need a rebase once fixed.

include/zephyr/net/lwm2m.h
jvermillard: done

subsys/net/lib/lwm2m/lwm2m_registry.h
jvermillard: done

rlubos (approved):

SeppoTakalo (approved):

jukkar (approved):

PR #99012, opened 2025-11-06
stm32: use generic LL headers everywhere

mathieuchopstm:
A few places are still including series-specific LL headers despite the existence of generic (series-agnostic) `stm32_ll_PPP` variants.

Replace all such includes with the generic header: `s/#include <stm32[^_]_ll_(.+)\.h>/#include <stm32_ll_$1.h>`. This also allows removal of include guards in a few places.

While at it, also remove CMSIS include which is already covered by `<soc.h>` in a few places too. (A second pass could be required; this is not the main objective of this PR)


erwango (approved):
Good clean up!

etienne-lms (approved):

PR #98140, opened 2025-10-23
drivers: syscon: Preparation for NVMEM usage

pdgendt:
This PR has the necessary bits to make the SYSCON API usable in NVMEM. Some of the parts

- Add a `syscon_get_reg_width` API function to determine how many bytes are read
- Add shell commands for easy testing
- Implement i.MX OCOTP SYSCON driver

This work relates to https://github.com/zephyrproject-rtos/zephyr/pull/96598 as the OCOTP SYSCON driver will be used as an NVMEM backend. This allows to rework the following with NVMEM:

https://github.com/zephyrproject-rtos/zephyr/blob/8aba2e3f3aa87deb9c8606ef54fddca1d936d031/drivers/ethernet/eth_nxp_enet.c#L646-L673

PR #99015, opened 2025-11-06
drivers: syscon: Add shell commands

pdgendt:
Add shell commands to use the SYSCON driver API with SYSCON devices.

Extracted from https://github.com/zephyrproject-rtos/zephyr/pull/98140

### DIFF
--- a/drivers/syscon/syscon_shell.c
+++ b/drivers/syscon/syscon_shell.c
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2025 Basalte bv
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file
+ * @brief SYSCON shell commands.
+ */
+
+#include <zephyr/shell/shell.h>
+#include <zephyr/drivers/syscon.h>
+
+static int cmd_base(const struct shell *sh, size_t argc, char **argv)
+{
+	const struct device *dev;
+	uintptr_t base;
+	int ret;
+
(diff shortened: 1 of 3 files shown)


gmarull (approved):

drivers/syscon/syscon_shell.c
```diff
@@ -0,0 +1,155 @@
...
+	entry->help = NULL;
+	entry->subcmd = NULL;
+}
+
+SHELL_DYNAMIC_CMD_CREATE(dsub_device_name, device_name_get);
+
+SHELL_STATIC_SUBCMD_SET_CREATE(syscon_cmds,
+	SHELL_CMD_ARG(base, &dsub_device_name,
+		      SHELL_HELP("Get the SYSCON device base address", "<device>"),
+		      cmd_base, 2, 0),
```
henrikbrixandersen, line 142: Is this really useful? Typically this would be part of the devicetree node name.

henrikbrixandersen (approved):

drivers/syscon/syscon_shell.c
pdgendt: It is part of the API, didn't want to discriminate any of the API functions. Also if some custom/out-of-tree driver actually needs logic for this, then devicetree might not be sufficient.

henrikbrixandersen: Thanks, makes sense to me.

fabiobaltieri (approved):

PR #98324, opened 2025-10-27
include: zephyr: APIs doc: fix Doxygen \retval command usage

JarmouniA:
Fix how Doxygen special command \retval is used in APIs doc text.

\retval: Starts a description for a function's return value with name `<return value>`, which needs to be attached and to not be a word that is part of the sentence, and each return value should have its own \retval.
Otherwise \retval is replaced with \return.

See https://www.doxygen.nl/manual/commands.html#cmdreturn
https://www.doxygen.nl/manual/commands.html#cmdretval
https://docs.rtems.org/docs/main/eng/coding-doxygen.html 

Doc preview https://builds.zephyrproject.io/zephyr/pr/98324/docs/doxygen/html/topics.html

PR #99016, opened 2025-11-06
drivers: syscon: Cleanup some doxygen return commands

pdgendt:
Documenting return values with doxygen should use the retval tags.

Extracted from https://github.com/zephyrproject-rtos/zephyr/pull/98140


JarmouniA (approved):
You could maybe help review #98324.

kartben (approved):
I think leading "If..." and "When..." should be removed, but in the absence of clear guidelines this can be considered a nit ;)

PR #99017, opened 2025-11-06
tests: fix test failure for smp targets

wearyzen:
The test is expected to run on single cpu so set max number of cpus to 1
so that it doesn't fail smp targets like fvp_baser_aemv8r in ci.


npitre:
I just tried this test on fvp_base_revc_2xaem/v9a/smp with
`CONFIG_MP_MAX_NUM_CPUS=4` and it executed successfully.

npitre:
Hmmm... Looking at the code I agree it makes more sense to restrict it 
to 1 CPU, especially when a benchmark is involved.

For the record, FVP_Base_RevC-2xAEMvA with -C cache_state_modelled=1 
simply hangs when CONFIG_MP_MAX_NUM_CPUS=4 and the test passes when 
CONFIG_MP_MAX_NUM_CPUS=1.

npitre (approved):

nashif (approved):

PR #99018, opened 2025-11-06
tests: Bluetooth: Tester: Fix NULL pointer dereference

sjanc:
PA sync can be terminated before BIG sync is stopped. This results in ps_sync being NULL and thus crash in stop callback. Simply store address needed by BTP events and don't rely on pa_sync.

fixes https://github.com/zephyrproject-rtos/zephyr/issues/99045


Thalley (approved):

jhedberg:
Linked bug report, please, so we can still get this in

jhedberg (approved):

jhedberg:
> This one fix crash in tester :)

Still needs a bug report, though :)

sjanc:
https://github.com/zephyrproject-rtos/zephyr/issues/99045

PR #99020, opened 2025-11-06
Bluetooth: ISO: Move iso dereference after assertion

sjanc:
Make sure assertion is true before dereferencing chan->iso. In extreme case compiler could skip assert check (asume it is always false) if address was already dereferenced.

### DIFF
--- a/subsys/bluetooth/host/iso.c
+++ b/subsys/bluetooth/host/iso.c
@@ -454,7 +454,7 @@ void bt_iso_connected(struct bt_conn *iso)
 
 static void bt_iso_chan_disconnected(struct bt_iso_chan *chan, uint8_t reason)
 {
-	const uint8_t conn_type = chan->iso->iso.info.type;
+	uint8_t conn_type;
 	struct net_buf *buf;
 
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/host/iso.c
```diff
@@ -454,13 +454,15 @@ void bt_iso_connected(struct bt_conn *iso)
...
-	const uint8_t conn_type = chan->iso->iso.info.type;
+	uint8_t conn_type;
 	struct net_buf *buf;
 
 	LOG_DBG("%p, reason 0x%02x", chan, reason);
 
 	__ASSERT(chan->iso != NULL, "NULL conn for iso chan %p", chan);
 
+	conn_type = chan->iso->iso.info.type;
+
```
Thalley: Since this is no longer `const`, I guess we could/should move the assignment closer to its first use (line 485)

Thalley (dismissed):
Change is fine with one minor suggestion

Thalley (approved):

jhedberg (approved):

PR #99021, opened 2025-11-06
drivers: udc_dwc2: avoid register read on disabled controller

jfischer-no:
The VBUS bounces when the USB connector is plugged in. This can lead to events VBUS removed and Suspended occurring in that order. With hibernation support enabled, hibernation request, as result of the suspend interrupt, will be delegated to the driver thread. Once the driver thread is scheduled to process hibernation request, the controller may be already disabled and controller/phy clocks be off. On nRF54LM20 this leads to CPU crash and a hang. To avoid this happening, the call to disable interrupts should be a first act in udc_disable(). Also, check if the UDC is still enabled before hibernation.

### DIFF
--- a/drivers/usb/udc/udc_dwc2.c
+++ b/drivers/usb/udc/udc_dwc2.c
@@ -2338,6 +2338,7 @@ static int udc_dwc2_disable(const struct device *dev)
 	}
 
 	config->irq_disable_func(dev);
+	cancel_hibernation_request(priv);
 
 	if (priv->hibernated) {
 		dwc2_exit_hibernation(dev, false, true);


josuah (dismissed):
If I understood it, disabling IRQs earlier reduces chances of IRQs making register I/O while DWC2 is un-powered. This makes sense to me.

drivers/usb/udc/udc_dwc2.c
```diff
@@ -3419,7 +3418,11 @@ static ALWAYS_INLINE void dwc2_thread_handler(void *const arg)
...
 		if (prev & BIT(DWC2_DRV_EVT_ENTER_HIBERNATION)) {
-			dwc2_enter_hibernation(dev);
+			if (udc_is_enabled(dev)) {
+				dwc2_enter_hibernation(dev);
+			} else {
+				LOG_INF("UDC is disabled, ignore hibernation request");
+			}
 		}
 
 		config->irq_enable_func(dev);
```
tmon-nordic: If the interrupts were disabled in `udc_dwc2_disable()`, then this handler will enable the interrupts which is undesired.

`udc_dwc2_disable()` is called with driver mutex locked, so I think it should be enough to call `cancel_hibernation_request()` inside `udc_dwc2_disable()`.

jfischer-no: Fixed, thanks.

drivers/usb/udc/udc_dwc2.c
```diff
@@ -2337,8 +2339,6 @@ static int udc_dwc2_disable(const struct device *dev)
 		return -EIO;
 	}
 
-	config->irq_disable_func(dev);
-
```
tmon-nordic, line 2364: Is there need to move the IRQ disable around? I think canceling hibernation request would make it better directly above `if (priv->hibernated) {` line.

jfischer-no: Yes, see commit message. udc_dwc2_ep_deactivate() may take relative long time to process, additionally up to 100 milliseconds in k_event_wait(). Making it the first act makes the issue not reproducible on my test devices. I think canceling hibernation event should be just after the interrupt disabling call, it looks more logical to me.

tmon-nordic (approved):

drivers/usb/udc/udc_dwc2.c
tmon-nordic: Actually, disabling interrupt before calling `udc_ep_disable_internal(dev, USB_CONTROL_EP_IN)` will force the function to timeout if endpoint was armed because it relies on `EPDISBLD` being received. Therefore this is wrong to disable the interrupts immediately on `udc_dwc2_disable()` entry.

A side note is that the control out endpoint cannot be disabled but the code does not really anticipate it and therefore it does timeout. But this is really a separate issue.

For the IN endpoint there is also problem that PHY clock has to be running in order to receive `EPDISBLD` interrupt. If the PHY clock gets lost then the timeout is really due to the lacking PHY clock. But that should be the only reason for timeout, and not the "disable interrupts earlier" (which is a broken logic).

The real problem this PR is addressing is entering hibernation after `udc_dwc2_disable()`. For this the real solution is to call `cancel_hibernation_request()`.

tmon-nordic (changes requested):
`config->irq_disable_func()` cannot be called before `udc_ep_disable_internal(dev, USB_CONTROL_EP_IN))`

drivers/usb/udc/udc_dwc2.c
jfischer-no: > The real problem this PR is addressing is entering hibernation after `udc_dwc2_disable()`. For this the real solution is to call `cancel_hibernation_request()`.

That is not enough. The suspended interrupt can be raised just after `cancel_hibernation_request()`. I can even reproduce it with my hardware. I changed it to disable only USBSUSP.

drivers/usb/udc/udc_dwc2.c
```diff
@@ -2325,6 +2325,8 @@ static int udc_dwc2_disable(const struct device *dev)
 	struct net_buf *buf;
 	int err;
 
+	sys_clear_bits((mem_addr_t)&base->gintmsk, USB_DWC2_GINTSTS_USBSUSP);
+	cancel_hibernation_request(priv);
```
tmon-nordic: This doesn't make much sense and is race prone. The interrupt for sure can happen before GINTMSK.USBSUSP bit is masked.

All you have to do is to just move `cancel_hibernation_request(priv);` below `config->irq_disable_func(dev);`. If there is a failure path possible with just `cancel_hibernation_request(priv);` below `config->irq_disable_func(dev);` then please write it.

jfischer-no: > This doesn't make much sense and is race prone. The interrupt for sure can happen before GINTMSK.USBSUSP bit is masked.

No it is not. Yes the interrupt can happen before USBSUSP bit is masked. But the goal is to avoid USBSUSP interrupt raising and k_event_post called when VBUS bounces during the connector is plugged in. If the USBSUSP is masked how it is done in this commit then the probability of the interrupt occurring is very low. I was unable to observe it.

> All you have to do is to just move `cancel_hibernation_request(priv);` below `config->irq_disable_func(dev);`. If there is a failure path possible with just `cancel_hibernation_request(priv);` below `config->irq_disable_func(dev);` then please write it.

It is not a "failure path", it is driver thread unnecessarily waked-up to waste cycles and discover that BIT(DWC2_DRV_EVT_ENTER_HIBERNATION) is not set.

tmon-nordic: Disabling control OUT endpoint is wrong anyway due to DWC2 IP design (DOEPCTL0 EPDis is read only), but this is out-of-scope here.

Please put the `cancel_hibernation_request(priv);` below `config->irq_disable_func(dev);` without any interrupt masking. This will solve the issue.

tmon-nordic (approved):

carlescufi (approved):

PR #99022, opened 2025-11-06
Drivers: SPI: Add new  SPI  module to PG23, PG28

silabs-vifisch:
Our goal is to extend usart capabilities to help users to use additional SPI device aout of the box.

signed-off by: silabs-vifisch <viktor.fisch@silabs.com>

### DIFF
--- a/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
+++ b/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
@@ -21,6 +21,20 @@
 		};
 	};
 
+	usart0_default: usart0_default {
+		group0 {
+			pins = <USART0_TX_PD11>, <EUSART1_SCLK_PD13>;
+			drive-push-pull;
+			output-high;
+		};
+
+		group1 {
+			pins = <USART0_RX_PD12>;
+			input-enable;
+			silabs,input-filter;
+		};
+	};
+
 	i2c0_default: i2c0_default {
 		group0 {
 			pins = <I2C0_SCL_PD10>, <I2C0_SDA_PD9>;
--- a/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
+++ b/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
@@ -23,7 +23,7 @@
 
 	usart0_default: usart0_default {
 		group0 {
-			pins = <USART0_TX_PD11>, <EUSART1_SCLK_PD13>;
+			pins = <USART0_TX_PD11>;
 			drive-push-pull;
 			output-high;
 		};
--- a/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
+++ b/boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
@@ -168,6 +168,18 @@
 	status = "okay";
 };
 
+&usart0 {
+	compatible = "silabs,usart-spi";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <4000000>;
+	pinctrl-0 = <&usart0_default>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	cs-gpios = <&gpiod 14 GPIO_ACTIVE_LOW>;
+};
+
 &i2c0 {
 	pinctrl-0 = <&i2c0_default>;
 	pinctrl-names = "default";


jhedberg:
@silabs-vifisch you need to fix your git author and signed-off-by information to be of the format `Firstname Lastname <email>`. Also, please just use `git commit -s` to generate the signed-off-by line, since it looks like you've manually created it (as yours starts with a lower-case letter)

boards/silabs/dev_kits/pg23_pk2504a/pg23_pk2504a.dts
```diff
@@ -158,6 +158,18 @@
...
 
+&eusart1 {
+	compatible = "silabs,eusart-spi";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <4000000>;
+	pinctrl-0 = <&eusart1_default>;
+	pinctrl-names = "default";
+	status = "okay";
+
```
jerome-pouiller: No empty line. Finish with `status = "okay";`. 

@jhedberg, I don't remember exactly the Zephyr policy when it comes to enabling a bus without any peripheral attached to it. Should we keep it `"disabled"`?

jerome-pouiller (commented):
`silabs` has to be mentioned somewhere in the commit title. Usually, we prefix with `dts: arm: silabs: `

Fix typo in `aout of the box`.

boards/silabs/dev_kits/pg23_pk2504a/pg23_pk2504a.dts
jhedberg: TBH, I don't know if there is a policy, or if there is what it is.

boards/silabs/dev_kits/pg23_pk2504a/pg23_pk2504a-pinctrl.dtsi
```diff
@@ -21,6 +21,20 @@
 		};
 	};
 
+	eusart1_default: eusart1_default {
+		group0 {
+			pins = <EUSART1_TX_PC3>, <EUSART1_SCLK_PC1>;
```
Martinhoff-maker: On pg23_pk2504a, i see that PC0, PC1, PC2 and are linked to lcd segment so I think this pinout is not right. I suggest to go on breakout pad pin.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
```diff
@@ -21,6 +21,20 @@
 		};
 	};
 
+	eusart0_default: eusart0_default {
+		group0 {
+			pins = <EUSART0_TX_PB3>, <EUSART0_SCLK_PB1>;
```
Martinhoff-maker: same comment.

boards/silabs/dev_kits/pg23_pk2504a/pg23_pk2504a-pinctrl.dtsi
silabs-vifisch: I think it is not really a problem, this is an alternate function, LCD is not used.
Any other pin has an alternate function, e.g. A5,  GPIO.EM4WU0, GPIO.TRACEDATA1, LESENSE.LESENSE_EN_2 and LCD.LCD_SEG_11 so it can be used for LESENSE if LCD is not used.

Martinhoff-maker: I agree that we have alternate function, but since we are on board level, those pin are physically connected to the lcd segment and nothing else, how will you use the eusart then ?

jerome-pouiller: I agree with @Martinhoff-maker, how the user is expected to connect to PC0, PC1, PC2 and PC3? From the schematic, it seems he has to unsolder the LCD controller, am I wrong?

silabs-vifisch: I think it is enough jut to disable the LCD. Actually on port is even routed to the breakout pad too, so it can be used from outside as well.

Martinhoff-maker: image

I don't see any connection with breakout pad ? where do you see it ? have you a link ?

silabs-vifisch: This picture is for pg23 board. The problem with this that all pins are reserved, not connected or may used by LCD if it is enabled, only B4-B6,  are available but for SPI we need 4 pins. So I have not much choice.
Concerning pg28 I reconfigured SPI pin out I've already committed.

asmellby: If the board doesn't provide SPI functionality, there shouldn't be SPI configuration in the board devicetree file.

silabs-vifisch: Actually we can configure SPI if LCD is not used.

jerome-pouiller: Keep in mind these overlay are not about the chip, they are about the board. It only expose the device th users can use with this board. Maybe some other boards/designs are more relevant to test SPI on PG23?

About PG28, indeed PD11-PD13 is fine (especially since the pin are explicitly assign to SPI in the schematics). However, you have to remove the use of the pins PB1-PB3.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
```diff
@@ -21,6 +21,20 @@
 		};
 	};
 
+	usart0_default: usart0_default {
+		group0 {
+			pins = <USART0_TX_PD11>, <EUSART1_SCLK_PD13>;
```
jerome-pouiller: `EUSART1` ?

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
```diff
@@ -168,6 +168,18 @@
 	status = "okay";
 };
 
+&usart0 {
+	compatible = "silabs,usart-spi";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	clock-frequency = <4000000>;
+	pinctrl-0 = <&eusart0_default>;
```
jerome-pouiller: eusart0 ?

samples/drivers/spi_bitbang/src/main.c
jerome-pouiller, line 1: You probably didn't want to commit these changes.

jerome-pouiller (commented):
Issue with PG23 is not yet fixed.

Split commits in an atomic way (for this PR, either one commit for PG28 and one commit for PG23, either just one commit for all changes)

BTW, please check your changes before to send new version of your PR (or convert it to a draft).

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
silabs-vifisch: eusart1 is is already defined in the .dts as an uart, it is probably used by another module. This is the reason why I decided  to introduce another usart module dedicated to SPI. I looked for available pin and find PD11-PD13 as we agreed above as it is assigned to SPI. So I looked which usart module can be routed to that pins, and found out his was uasrt0.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
silabs-vifisch: eusart0 become unnecessary as uart0 is used instead. This is the reason its removal.

samples/drivers/spi_bitbang/src/main.c
silabs-vifisch: So these changes seem to be reasonable.
The only thing I need to remove is spi_bitbang.c as it got into the commit by accident I didn't want to commit that at all.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
jerome-pouiller: That's not my point.

`eusart0_default` is not defined. So, I wonder if this code has been compiled once.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
jerome-pouiller: That's not my point.

`EUSART1_SCLK_PD13` assigns `PD13` to `eusart1` while you obviously want to use `usart0`.

silabs-vifisch: Yes, usart0 is tended to be used.
Actually eusart1 defaults to PD7 and 8 in pg28 .dts file.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a.dts
silabs-vifisch: usart0_default is defined as usart0 is used, it compiled this way.

jhedberg:
@silabs-vifisch please improve the commit message a bit more. It's not really describing the change properly IMO. It's also good to answer the question "why?" in the commit message, as that may not be clear to everyone.

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
jerome-pouiller: Don't you mean USART0_CLK_PD13 rather than EUSART1_SCLK_PD13?

boards/silabs/dev_kits/pg28_pk2506a/pg28_pk2506a-pinctrl.dtsi
jerome-pouiller, line 1: About the commit log:
  - Remove leading space
  - I don't see the link with bluetooth or samples. Generally, the prefix reflects the main directory you change.
  - The title is too vague

PR #99023, opened 2025-11-06
linker: llext: move #ifdef CONFIG_LLEXT guard inside common file

mathieuchopstm:
LLEXT-related sections should only be added to the linker script when the associated `CONFIG_LLEXT` is enabled. This has been done by checking for this Kconfig symbol in every linker file, but this creates a lot of boilerplate for no good reason.

Use the much simpler solution: move the check inside the common linker file and remove existing `#ifdef CONFIG_LLEXT` checks in all linker files.

(Bonus point: this doesn't break out-of-tree users which will just be checking twice)

### DIFF
--- a/boards/qemu/x86/qemu_x86_tiny.ld
+++ b/boards/qemu/x86/qemu_x86_tiny.ld
@@ -253,10 +253,7 @@ SECTIONS
 	{
 
 #include <zephyr/linker/rel-sections.ld>
-
-#ifdef CONFIG_LLEXT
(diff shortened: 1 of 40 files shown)


iuliana-prodan (approved):

teburd (approved):
Nice cleanup

wearyzen (approved):

marekmatej (approved):

nashif (approved):

wmrsouza (approved):

PR #99024, opened 2025-11-06
stm32: tests: drivers: flash: fix regressions on stm32h7x serie after cache management enable by default

djiatsaf-st:
This commit https://github.com/zephyrproject-rtos/zephyr/commit/d5ead50fc69daa682e18cfe89d4ea22d0fcd6445 enable cache management by default on Cortex-M7 of the STM32H7x. 

This PR addresses two issues observed on STM32H7 after this addition. 

- **Fix flash data coherency after write**
     Flash writes on STM32H7 require explicit cache flushing to ensure data coherency.

- **Prevent Bus Fault when unlocking blocked flash registers**
     Blocking access to flash option/control registers may cause imprecise Bus Faults when attempting to unlock 
     them immediately afterward. 
    No regressions encountered on other platforms with this fix.

### DIFF
--- a/drivers/flash/flash_stm32h7x.c
+++ b/drivers/flash/flash_stm32h7x.c
@@ -796,6 +796,9 @@ static int flash_stm32h7_write(const struct device *dev, off_t offset, const voi
 		rc = rc2;
 	}
 
+#ifdef CONFIG_DCACHE
+	flash_stm32h7_flush_caches(dev, offset, len);
+#endif
 	flash_stm32_sem_give(dev);
 
 	return rc;
--- a/tests/drivers/flash/stm32/src/main.c
+++ b/tests/drivers/flash/stm32/src/main.c
@@ -278,6 +278,16 @@ ZTEST(flash_stm32, test_stm32_block_registers)
 	TC_PRINT("Try to unlock blocked OPT\n");
 	__set_FAULTMASK(1);
 	flash_stm32_option_bytes_lock(flash_dev, false);
+
+	/* Ensure the Imprecise Bus Fault caused by the illegal
+	 * access is seen now while BusFault is still masked by
+	 * triggering a Context synchronization event. This is
+	 * notably required on series such as STM32H7 when Icache
+	 * is enabled - the Imprecise Bus Fault is reported after
+	 * we unmask BusFault and triggers a kernel panic.
+	 */
+	barrier_isync_fence_full();
+
 	/* Clear Bus Fault pending bit */
 	SCB->SHCSR &= ~SCB_SHCSR_BUSFAULTPENDED_Msk;
 	barrier_dsync_fence_full();


drivers/flash/flash_stm32h7x.c
```diff
@@ -797,6 +797,11 @@ static int flash_stm32h7_write(const struct device *dev, off_t offset, const voi
 	}
 
 	flash_stm32_sem_give(dev);
+#if defined(CONFIG_CACHE_MANAGEMENT) && defined(CONFIG_DCACHE)
```
JarmouniA:
```suggestion
#ifdef CONFIG_DCACHE
```

JarmouniA (dismissed):

drivers/flash/flash_stm32h7x.c
djiatsaf-st: updated

tests/drivers/flash/stm32/src/main.c
```diff
@@ -278,6 +278,12 @@ ZTEST(flash_stm32, test_stm32_block_registers)
 	TC_PRINT("Try to unlock blocked OPT\n");
 	__set_FAULTMASK(1);
 	flash_stm32_option_bytes_lock(flash_dev, false);
+
+	/* Ensure instruction pipeline is flushed to prevent imprecise
+	 * Bus Fault on STM32H7 with cache enabled
+	 */
```
mathieuchopstm: If my understanding is correct:
```suggestion
	/* Ensure the Imprecise Bus Fault caused by the illegal
	 * access is seen now while BusFault is still masked by
	 * triggering a Context synchronization event. This is
	 * notably required on series such as STM32H7 when Icache
	 * is enabled - the Imprecise Bus Fault is reported after
	 * we unmask BusFault and triggers a kernel panic.
	 */
```

tests/drivers/flash/stm32/src/main.c
```diff
@@ -278,6 +278,12 @@ ZTEST(flash_stm32, test_stm32_block_registers)
...
+	__ISB();
```
mathieuchopstm:
```suggestion
	barrier_isync_fence_full();
```

drivers/flash/flash_stm32h7x.c
```diff
@@ -797,6 +797,11 @@ static int flash_stm32h7_write(const struct device *dev, off_t offset, const voi
 	}
 
 	flash_stm32_sem_give(dev);
+#ifdef CONFIG_DCACHE
+	off_t flush_offset = ROUND_DOWN(offset, FLASH_SECTOR_SIZE);
+	size_t flush_len = ROUND_UP(offset + len - 1, FLASH_SECTOR_SIZE) - flush_offset;
+	flash_stm32h7_flush_caches(dev, flush_offset, flush_len);
```
etienne-lms: Should cache be flushed before semaphore is given back?
Maybe even before flash_stm32_write_range() is called?

djiatsaf-st: > Should cache be flushed before semaphore is given back? 

Yes, it works.

djiatsaf-st: For flash_stm32_write_range(), it's too early, you need to write it first, from what I've observed.

etienne-lms: Ok , indeed, it maes sense since here we invalidate the related cache lines.

That said, since `flash_stm32h7_cr_lock(..., true)` need to wait flash returns to idle state, I would suggest to call **~~`flash_stm32_write_range()`~~ `flash_stm32h7_flush_caches()`** before `flash_stm32h7_cr_lock()` to optimize (a little bit) the write sequence wait states.

**(edited) fixed function label**

etienne-lms: IMHO, we should rename `flash_stm32h7_flush_caches()` into `flash_stm32h7_invalidate_caches()`. Maybe in a later P-R.

Could you change the commit message, e.g.
```patch
-drivers: flash: stm32_h7: flush data cache after flash write
+drivers: flash: stm32_h7: invalidate data cache after flash write
 
 STM32H7 platforms now enable cache management by default.
-To ensure data coherency after flash writes, added cache flush
-logic using flash_stm32h7_flush_caches() for the affected flash
-region.
+To ensure data coherency after flash writes, invalidate cache lines
+related to the region written.
 This prevents stale data and ensures proper memory visibility.
 
 Signed-off-by: ...

**Review by etienne-lms — DISMISSED** [review]
~~My *Approve* feedback since my review comments are not blocking.~~

File: drivers/flash/flash_stm32h7x.c (hunk 4 above)
Comment (etienne-lms) [reply drivers/flash/flash_stm32h7x.c]: > Ok , indeed, it maes sense since here we invalidate the related cache lines.

Hmm... actually I'm not that sure this is really nice.

If `offset` and `length` are not aligned on a cache line (which `flash_stm32_valid_range()` allows), this could legitimate invalidate data in cache. Not really expected. Maybe the cache line should be *cleaned* before the write operation and *invalidated* after the write operation.

Comment (etienne-lms) [reply drivers/flash/flash_stm32h7x.c]: Sorry for the noise with the several comments above.
I tried to summarize below.

**1. Which address range to invalidate?**

Here the invalidate address range should match the written range, not the sector in which we write.
Something like:
```c
	const off_t quantum = FLASH_STM32_WRITE_BLOCK_SIZE;
	off_t flush_offset = ROUND_DOWN(offset, FLASH_STM32_WRITE_BLOCK_SIZE);
	size_t flush_len = ROUND_UP(offset + len - 1, FLASH_STM32_WRITE_BLOCK_SIZE) -
			   flush_offset;

	flash_stm32h7_flush_caches(dev, flush_offset, flush_len);
```

That said, supported targets have `FLASH_STM32_WRITE_BLOCK_SIZE` of 16byte or 32byte and 32byte cache lines, therefore, since `flash_stm32h7_flush_caches()` already extends target address range to cache line size (see the CMSIS function), this could be simplified to simply:
```c
	flash_stm32h7_flush_caches(dev, offset, len);
```

**2. Case written address range is not aligned on cache line size**

Sorry, the below comment made no sense. There can't be pending data in cache to be cleaned.
I remove the comment...
~~On STM32H7Axx/STM32H7Bxx, `FLASH_NB_32BITWORD_IN_FLASHWORD` is 4 hence we can write to 16 byte aligned data, not 32 byte (which match the cache Cortex-M7 line). Hence, for these devices, I think prior writing to the flash we should clean possible cached data outside written range.~~

~~Something like this:~~
(removed)

djiatsaf-st: updated
Thanks

> ```c
> 	flash_stm32h7_flush_caches(dev, offset, len);
> ```
>

gautierg-st (approved):

mathieuchopstm (approved):

etienne-lms (approved):

drivers/flash/flash_stm32h7x.c
```diff
@@ -796,6 +796,9 @@ static int flash_stm32h7_write(const struct device *dev, off_t offset, const voi
 		rc = rc2;
 	}
 
+#ifdef CONFIG_DCACHE
+	flash_stm32h7_flush_caches(dev, offset, len);
```
erwango, line 800: Should it be extended to other series ?

erwango (approved):

PR #99025, opened 2025-11-06
tests: bluetooth: tester: Fix VCS tests

Frodevan:
Zephyr automatically sets the persisted flag for VCS if the volume is changed after the service is registered. As some of these PTS tests require a set volume with the flag cleared, the initial volume needs to be set before registering VCS.

This PR is associated with https://github.com/auto-pts/auto-pts/pull/1603

### DIFF
--- a/tests/bluetooth/tester/src/audio/btp/btp_vcs.h
+++ b/tests/bluetooth/tester/src/audio/btp/btp_vcs.h
@@ -8,6 +8,8 @@
 
 #include <stdint.h>
 
+#define BTP_VCS_REGISTER_FLAG_MUTED		BIT(0)
+
 #define BTP_VCS_READ_SUPPORTED_COMMANDS		0x01
 struct btp_vcs_read_supported_commands_rp {
 	uint8_t data[0];
@@ -22,3 +24,10 @@ struct btp_vcs_set_vol_cmd {
 #define BTP_VCS_VOL_DOWN			0x04
 #define BTP_VCS_MUTE				0x05
 #define BTP_VCS_UNMUTE				0x06
+
+#define BTP_VCS_REGISTER			0x07
+struct btp_vcs_register_cmd {
+	uint8_t step;
+	uint8_t flags;
+	uint8_t volume;
+} __packed;
--- a/tests/bluetooth/tester/src/audio/btp_vcp.c
+++ b/tests/bluetooth/tester/src/audio/btp_vcp.c
@@ -140,6 +140,43 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
 	return BTP_STATUS_SUCCESS;
 }
 
+static void set_register_params(uint8_t gain_mode, uint8_t step,
+				bool muted, uint8_t volume);
+
+static uint8_t register_vcs(const void *cmd, uint16_t cmd_len,
+			     void *rsp, uint16_t *rsp_len)
+{
+	static bool vcs_registered_flag;
+	const struct btp_vcs_register_cmd *cp = cmd;
+	int err;
+
+	LOG_DBG("Registering VCS");
+
+	if (vcs_registered_flag) {
+		return BTP_STATUS_FAILED;
+	}
+
+	bool muted = (cp->flags & BTP_VCS_REGISTER_FLAG_MUTED) != 0;
+
+	set_register_params(BT_AICS_MODE_MANUAL, cp->step, muted, cp->volume);
+
+	err = bt_vcp_vol_rend_register(&vcp_register_param);
+	if (err != 0) {
+		return BTP_STATUS_FAILED;
+	}
+
+	err = bt_vcp_vol_rend_included_get(&included);
+	if (err != 0) {
+		return BTP_STATUS_FAILED;
+	}
+
+	aics_server_instance.aics_cnt = included.aics_cnt;
+	aics_server_instance.aics = included.aics;
+	vcs_registered_flag = true;
+
+	return BTP_STATUS_SUCCESS;
+}
+
 static void vcs_state_cb(struct bt_conn *conn, int err, uint8_t volume, uint8_t mute)
 {
 	LOG_DBG("VCP state cb err (%d)", err);
@@ -186,6 +223,11 @@ static const struct btp_handler vcs_handlers[] = {
 		.opcode = BTP_VCS_UNMUTE,
 		.expect_len = 0,
 		.func = unmute,
+	},
+	{
+		.opcode = BTP_VCS_REGISTER,
+		.expect_len = sizeof(struct btp_vcs_register_cmd),
+		.func = register_vcs,
 	},
 };
 
@@ -463,7 +505,8 @@ struct bt_aics_cb aics_server_cb = {
 };
 
 /* General profile handling */
-static void set_register_params(uint8_t gain_mode)
+static void set_register_params(uint8_t gain_mode, uint8_t step,
+				bool muted, uint8_t volume)
 {
 	static char input_desc[CONFIG_BT_VCP_VOL_REND_AICS_INSTANCE_COUNT]
 			      [BT_AICS_MAX_INPUT_DESCRIPTION_SIZE];
@@ -495,31 +538,18 @@ static void set_register_params(uint8_t gain_mode)
 		vcp_register_param.aics_param[i].cb = &aics_server_cb;
 	}
 
-	vcp_register_param.step = 1;
-	vcp_register_param.mute = BT_VCP_STATE_UNMUTED;
-	vcp_register_param.volume = 100;
+	vcp_register_param.step = step;
+	if (muted) {
+		vcp_register_param.mute = BT_VCP_STATE_MUTED;
+	} else {
+		vcp_register_param.mute = BT_VCP_STATE_UNMUTED;
+	}
+	vcp_register_param.volume = volume;
 	vcp_register_param.cb = &vcs_cb;
 }
 
 uint8_t tester_init_vcs(void)
 {
-	int err;
-
-	set_register_params(BT_AICS_MODE_MANUAL);
-
-	err = bt_vcp_vol_rend_register(&vcp_register_param);
-	if (err) {
-		return BTP_STATUS_FAILED;
-	}
-
-	err = bt_vcp_vol_rend_included_get(&included);
-	if (err) {
-		return BTP_STATUS_FAILED;
-	}
-
-	aics_server_instance.aics_cnt = included.aics_cnt;
-	aics_server_instance.aics = included.aics;
-
 	tester_register_command_handlers(BTP_SERVICE_ID_VCS, vcs_handlers,
 					 ARRAY_SIZE(vcs_handlers));
 
--- a/tests/bsim/bluetooth/tester/src/bsim_btp.c
+++ b/tests/bsim/bluetooth/tester/src/bsim_btp.c
@@ -873,6 +873,8 @@ static bool is_valid_vcs_packet_len(const struct btp_hdr *hdr, struct net_buf_si
 		return buf_simple->len == 0U;
 	case BTP_VCS_UNMUTE:
 		return buf_simple->len == 0U;
+	case BTP_VCS_REGISTER:
+		return buf_simple->len == 0U;
 
 	/* no events */
 	default:
--- a/tests/bsim/bluetooth/tester/src/bsim_btp.h
+++ b/tests/bsim/bluetooth/tester/src/bsim_btp.h
@@ -563,6 +563,27 @@ static inline void bsim_btp_wait_for_tmap_discovery_complete(void)
 	net_buf_unref(buf);
 }
 
+static inline void bsim_btp_vcs_register(uint8_t step, uint8_t flags, uint8_t vol)
+{
+	struct btp_vcs_register_cmd *cmd;
+	struct btp_hdr *cmd_hdr;
+
+	NET_BUF_SIMPLE_DEFINE(cmd_buffer, BTP_MTU);
+
+	cmd_hdr = net_buf_simple_add(&cmd_buffer, sizeof(*cmd_hdr));
+	cmd_hdr->service = BTP_SERVICE_ID_VCS;
+	cmd_hdr->opcode = BTP_VCS_REGISTER;
+	cmd_hdr->index = BTP_INDEX;
+	cmd = net_buf_simple_add(&cmd_buffer, sizeof(*cmd));
+	cmd->step = step;
+	cmd->flags = flags;
+	cmd->volume = vol;
+
+	cmd_hdr->len = sys_cpu_to_le16(cmd_buffer.len - sizeof(*cmd_hdr));
+
+	bsim_btp_send_to_tester(cmd_buffer.data, cmd_buffer.len);
+}
+
 static inline void bsim_btp_vcp_discover(const bt_addr_le_t *address)
 {
 	struct btp_vcp_discover_cmd *cmd;
--- a/tests/bsim/bluetooth/tester/src/audio/vcp_peripheral.c
+++ b/tests/bsim/bluetooth/tester/src/audio/vcp_peripheral.c
@@ -35,6 +35,7 @@ static void test_vcp_peripheral(void)
 	bsim_btp_core_register(BTP_SERVICE_ID_AICS);
 
 	bsim_btp_gap_set_discoverable(BTP_GAP_GENERAL_DISCOVERABLE);
+	bsim_btp_vcs_register(1U, 0U, 100U);
 	bsim_btp_gap_start_advertising(0U, 0U, NULL, BT_HCI_OWN_ADDR_PUBLIC);
 	bsim_btp_wait_for_gap_device_connected(&remote_addr);
 	bt_addr_le_to_str(&remote_addr, addr_str, sizeof(addr_str));


tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -85,8 +87,12 @@ static uint8_t set_volume(const void *cmd, uint16_t cmd_len,
...
 	LOG_DBG("Set volume 0x%02x", cp->volume);
 
-	if (bt_vcp_vol_rend_set_vol(cp->volume) != 0) {
-		return BTP_STATUS_FAILED;
+	if (vcs_registered_flag) {
+		if (bt_vcp_vol_rend_set_vol(cp->volume) != 0) {
+			return BTP_STATUS_FAILED;
+		}
+	} else {
+		vcp_register_param.volume = cp->volume;
```
Thalley: Suggest to reject the command when VCS isn't registered instead

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +146,32 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
...
+		if (err) {
+			return BTP_STATUS_FAILED;
+		}
+
+		aics_server_instance.aics_cnt = included.aics_cnt;
+		aics_server_instance.aics = included.aics;
+		vcs_registered_flag = true;
+	}
+
+	return BTP_STATUS_SUCCESS;
```
Thalley, line 177: Suggest make this work similar to `tester_init_vcs` w.r.t. if it is called twice, it fails the second time.

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
```diff
@@ -22,3 +22,4 @@ struct btp_vcs_set_vol_cmd {
 #define BTP_VCS_VOL_DOWN			0x04
 #define BTP_VCS_MUTE				0x05
 #define BTP_VCS_UNMUTE				0x06
+#define BTP_VCS_REGISTER			0x07
```
Thalley, line 28: I would suggest to make this command contain the parameters (basically every value in `struct bt_vcp_vol_rend_register_param` except the callbacks), so that it's fully controlled by the upper tester / AutoPTS, instead of relying on some defaults set by the BT Tester. This allows AutoPTS to maintain the same values and behavior between different DUTs.

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -67,6 +67,8 @@ struct service_handles {
 struct service_handles chrc_handles;
 
 /* Volume Control Service */
+static int vcs_registered_flag;
```
Thalley:
```suggestion
static bool vcs_registered_flag;
```

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -503,22 +540,9 @@ static void set_register_params(uint8_t gain_mode)
...
-	}
-
-	err = bt_vcp_vol_rend_included_get(&included);
-	if (err) {
-		return BTP_STATUS_FAILED;
-	}
-
-	aics_server_instance.aics_cnt = included.aics_cnt;
-	aics_server_instance.aics = included.aics;
+	vcs_registered_flag = false;
```
Thalley: Given that `vcs_registered_flag` is `static`, this should already be false at this point

```suggestion
```

tests/bluetooth/tester/src/audio/btp_vcp.c
Thalley: This can be done by an early check of 

```c
if (vcs_registered_flag) {
    return BTP_STATUS_FAILED;
}
```

Thalley (changes requested):
I like the idea overall, as it can provide addition flexibility (required by some VCS tests)

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Doing it that way will depend on the other suggestion regarding ` struct bt_vcp_vol_rend_register_param`. I'll see how far I get with that, but it will be cleaner to pipe the initial settings through that yes.

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
Frodevan: I agree, and I am currently working on this.

Frodevan: btp_vcs_register_cmd has now been updated to carry the step size, as well as initial state of mute and volume.

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
```diff
@@ -22,3 +22,10 @@ struct btp_vcs_set_vol_cmd {
 #define BTP_VCS_VOL_DOWN			0x04
 #define BTP_VCS_MUTE				0x05
 #define BTP_VCS_UNMUTE				0x06
+
+#define BTP_VCS_REGISTER			0x07
+struct btp_vcs_register_cmd {
+	uint8_t step;
+	bool mute;
+	uint8_t volume;
+} __packed;
```
Thalley, line 33: This is good for VCS, but I wonder if we need to consider the AICS and VOCS register parameters as well, while we are at it? 

In the case that we need/want to do something similar to VCS for the included AICS or VOCS services, I'd rather not need to add a `_v2` version of this, or to add `set_aics_register_param` commands to do this. 

Do you think it would be too much effort to add the AICS/VOCS parameters as part of this, or would it make sense to do it now while we are adding this command?

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
```diff
@@ -22,3 +22,10 @@ struct btp_vcs_set_vol_cmd {
 #define BTP_VCS_VOL_DOWN			0x04
 #define BTP_VCS_MUTE				0x05
 #define BTP_VCS_UNMUTE				0x06
+
+#define BTP_VCS_REGISTER			0x07
+struct btp_vcs_register_cmd {
+	uint8_t step;
+	bool mute;
```
Thalley: IIRC, we don't use `bool` in BTP; suggest to change this to a `uint8_t`

```suggestion
	uint8_t mute;
```

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -495,31 +537,18 @@ static void set_register_params(uint8_t gain_mode)
...
 
-	vcp_register_param.step = 1;
-	vcp_register_param.mute = BT_VCP_STATE_UNMUTED;
-	vcp_register_param.volume = 100;
+	vcp_register_param.step = step;
+	if (mute) {
+		vcp_register_param.mute = BT_VCP_STATE_MUTED;
+	} else {
+		vcp_register_param.mute = BT_VCP_STATE_UNMUTED;
+	}
```
Thalley: `mute` should either be a `bool`, or we should just do

```suggestion
	vcp_register_param.mute = mute;
```

tests/bsim/bluetooth/tester/src/bsim_btp.c
```diff
@@ -873,6 +873,8 @@ static bool is_valid_vcs_packet_len(const struct btp_hdr *hdr, struct net_buf_si
 		return buf_simple->len == 0U;
 	case BTP_VCS_UNMUTE:
 		return buf_simple->len == 0U;
+	case BTP_VCS_REGISTER:
+		return buf_simple->len == 0U;
```
Thalley:
```suggestion
		return buf_simple->len == sizeof(struct btp_vcs_register_cmd);
```

tests/bsim/bluetooth/tester/src/bsim_btp.h
```diff
@@ -563,6 +563,27 @@ static inline void bsim_btp_wait_for_tmap_discovery_complete(void)
...
+	cmd_hdr = net_buf_simple_add(&cmd_buffer, sizeof(*cmd_hdr));
+	cmd_hdr->service = BTP_SERVICE_ID_VCS;
+	cmd_hdr->opcode = BTP_VCS_REGISTER;
+	cmd_hdr->index = BTP_INDEX;
+	cmd = net_buf_simple_add(&cmd_buffer, sizeof(*cmd));
+	cmd->step = step;
+	cmd->mute = mute;
+	cmd->volume = vol;
+
+	cmd_hdr->len = cmd_buffer.len - sizeof(*cmd_hdr);
```
Thalley:
```suggestion
	cmd_hdr->len = sys_cpu_to_le16(cmd_buffer.len - sizeof(*cmd_hdr));
```

I know `sys_cpu_to_le16` is missing from quite a lot of these (and arguably not required given that BSIM is currently always little-endian), but we should add it for correctness

Thalley (changes requested):
Overall looks good. I think we need to consider whether to also include parameters for AICS and VOCS

tests/bsim/bluetooth/tester/src/bsim_btp.c
Frodevan: If you look at line 866 and 867, BTP_VCS_SET_VOL is also returning == 0U. Both of these BTP commands contains a struct of parameters, so I am not sure how this is expected to return.

Thalley: It is supposed to return whether `buf_simple->len` matches the size of the command :) 

I think it's a bug for `BTP_VCS_SET_VOL` as well, and looking at the existing BSIM tests, we don't seem to use that command, so that's why the bug has gone unnoticed I think

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
Frodevan: In my case, it would be a lot of work. For someone more experienced with Zephyr and handling of nested structs of variable sizes, it would probably be a lot easier.

Another issue I see here, however, is that some of the AICS and VOCS settings may be very dependent on KConfig. This can lead to a potential desync between the actual bttester build vs. the assumed settings in AutoPTS, if AutoPTS is to set all the settings.

Thalley: > Another issue I see here, however, is that some of the AICS and VOCS settings may be very dependent on KConfig. This can lead to a potential desync between the actual bttester build vs. the assumed settings in AutoPTS, if AutoPTS is to set all the settings.

I think that argument can be said about any feature guarded by a Kconfig option (which is most of them) :) 

Would like to get input from @sjanc here w.r.t. whether just registering VCS explicitly like this is OK, or if we should go all the way and include AICS and VOCS

tests/bsim/bluetooth/tester/src/bsim_btp.h
Frodevan: Fixed

tests/bsim/bluetooth/tester/src/bsim_btp.c
Frodevan: Changed

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Changed

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
Frodevan: Changed

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +142,40 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
...
+	const struct btp_vcs_register_cmd *cp = cmd;
+	int err;
+
+	LOG_DBG("Registering VCS");
+
+	if (vcs_registered_flag) {
+		return BTP_STATUS_FAILED;
+	}
+
+	set_register_params(BT_AICS_MODE_MANUAL, cp->step, cp->mute, cp->volume);
```
Thalley: I know it seems kind of foolish to have the `cp->mute` be a `uint8_t` and then want to use it as a `bool`, but that's just how BTP is designed :) 

I suggest the following to "cast" the `cp->mute` to a `bool` 

```suggestion
	set_register_params(BT_AICS_MODE_MANUAL, cp->step, cp->mute != 0, cp->volume);
```

Thalley: Alternative have `set_register_params` take the parameter as either `BT_VCP_STATE_MUTED` or `BT_VCP_STATE_UNMUTED` and use it directly as suggested by https://github.com/zephyrproject-rtos/zephyr/pull/99025/files#r2510685936

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +142,40 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
...
+
+	err = bt_vcp_vol_rend_register(&vcp_register_param);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +142,40 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
...
+		return BTP_STATUS_FAILED;
+	}
+
+	err = bt_vcp_vol_rend_included_get(&included);
+	if (err) {
```
Thalley:
```suggestion
	if (err != 0) {
```

Thalley (commented):
Minor comments, but nothing that is really blocking. Will await reply from @sjanc on the design before approving

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: changed

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: changed

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: changed, that being said almost all other instances in the file use _if (err) {_

Thalley: True, but they probably got in before our coding guidelines required the above :)

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -67,6 +67,8 @@ struct service_handles {
 struct service_handles chrc_handles;
 
 /* Volume Control Service */
+static bool vcs_registered_flag;
```
Thalley: We can actually move this variable to the `register_vcs` function

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +142,40 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
 	return BTP_STATUS_SUCCESS;
 }
 
+static void set_register_params(uint8_t gain_mode, uint8_t step,
+				bool mute, uint8_t volume);
+
+static uint8_t register_vcs(const void *cmd, uint16_t cmd_len,
+			  void *rsp, uint16_t *rsp_len)
```
Thalley:
```suggestion
static uint8_t register_vcs(const void *cmd, uint16_t cmd_len,
			    void *rsp, uint16_t *rsp_len)
```

Indentation was off (tabs = 8 spaces)

Thalley (commented):
Minor comments, otherwise LGTM

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
sjanc: I'm actually fine with adding new command in future if needed (it is relatively easy to handle this in autopts,  ie call X if IUT supports X and assume implicit action if it doesn't)

however I'd make mute a flag as mentioned in autopts related PR

Frodevan: The mute setting has been converted to a flag now

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
```diff
@@ -8,6 +8,8 @@
 
 #include <stdint.h>
 
+#define BTP_VCS_REGISTER_FLAG_MUTED			0x01
```
Thalley:
```suggestion
#define BTP_VCS_REGISTER_FLAG_MUTED		BIT(0)
```

Ditto with using `BIT`

tests/bluetooth/tester/src/audio/btp_vcp.c
```diff
@@ -140,6 +140,42 @@ static uint8_t unmute(const void *cmd, uint16_t cmd_len,
...
+	int err;
+
+	LOG_DBG("Registering VCS");
+
+	if (vcs_registered_flag) {
+		return BTP_STATUS_FAILED;
+	}
+
+	set_register_params(BT_AICS_MODE_MANUAL, cp->step,
+			    (cp->flags | BTP_VCS_REGISTER_FLAG_MUTED != 0), cp->volume);
```
Thalley:
```suggestion

	set_register_params(BT_AICS_MODE_MANUAL, cp->step,
			    ((cp->flags & BTP_VCS_REGISTER_FLAG_MUTED) != 0), cp->volume);
```

Given the complexity, you probably want to move this to a local variable


```suggestion
        muted = (cp->flags & BTP_VCS_REGISTER_FLAG_MUTED) != 0;
        
	set_register_params(BT_AICS_MODE_MANUAL, cp->step, mute, cp->volume);
```

Alternative we can also send the flags to `set_register_params` and handle it there

Thalley (changes requested):

tests/bluetooth/tester/src/audio/btp/btp_vcs.h
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_vcp.c
Frodevan: This is easier to read yes, changed.

Also I vote to keep it here, since this function has the responsibility of the BTP command unpacking.

Thalley (dismissed):

Thalley (dismissed):

Thalley (approved):

asbjornsabo (approved):
LGTM

sjanc:
hmm this seems to break some AICS and VOCS tests

PR #99029, opened 2025-11-06
shields: nxp: lcd_par_s035: fix gt911 touch initialization

### LINKED ISSUE #98876: shields: NXP: LCD-PAR-S035: touchscreen may not respond
### Describe the bug

A timing issue can cause some LCD-PAR-S035 shields to be unresponsive to touch events.  Writing to the display is not impacted.  Because this issue is caused by marginal timing, it is not seen on all shields.  Specific shields suffer from the issue more than others, and on some shields the issue may be intermittent.  This is a software issue, and is not related to the revision of the shield.  The issue can also be seen on multiple base-boards, and has been reported on the FRDM-MCXN947 and FRDM-RW612.  When a shield is stuck in this state, the GT911 touch controller does not ACK its target address on the I2C bus.  The shield I tested showing this issue is so sensitive to the timing that adding a test probe to the INT pin avoided the issue and touch started working.  The issue was originally reported to me on v4.2.0, but I still see the issue with v4.3.0-rc2

The root cause is related to two signals on the GT911 touch controller: INT and Reset.  Normally, INT is an output from the touch IC to the host to interrupt and indicate it detected a touch event.  But during reset, the GT911 uses INT as an input to configure the I2C target address.  To use the expected I2C target address on this shield, the INT signal must remain low before and after the Reset signal is deasserted.  But this Reset signal is also shared with the display controller on the shield.  Both the GT911 and ST7796S drivers drive this Reset signal.  And when the ST7796S driver deasserts Reset, the required INT timing is not always achieved.  The result is GT911 no longer ACKs the I2C target address.

This captures the signals on a bad shield that does not respond to touch.  The INT signal is not included here because probing the INT signal hides the issue.  The GT911 touch controller is initialized first.  The first Reset pulse is driven by the GT911 driver.  Then the GT911 driver has multiple I2C transfers to read the GT911 firmware and write it back.  Those transfers all succeed.  The GT911 driver init completes, and next the ST7796S display driver initializes.  It asserts Reset again for 100ms, which is what causes this issue.  Afterwards, the GT911 NAKs all read/writes to its I2C address.  The GT911 driver attempts to read the GT911 every 10ms and repeats this forever.
Image

This capture zooms in on a single I2C write command after the ST7796S init, and the GT911 no longer ACKs the host.
Image

Adding a probe to the INT signal on J5 pin12 hides the issue, and touch starts working.  This capture shows that the INT signal is hight when the ST7796S driver asserts Reset.  And the timing required for the INT signal with Reset is not met.
Image

### Regression

- [ ] This is a regression.

### Steps to reproduce

Not every shield shows the issue.  I tested with the FRDM-MCXN947 base-board and the LVGL demo on v4.3.0-rc2 with Zephyr SDK v0.17.4 using this build command:
```
west build -b frdm_mcxn947//cpu0 --shield lcd_par_s035_8080 samples/modules/lvgl/demos/ --pristine
```

### Relevant log output

```shell

```

### Impact

Intermittent – Occurs occasionally; hard to reproduce.

### Environment

_No response_

### Additional Context

I have a hack that prevents the issue, but it is a quick fix to demonstrate how to manage the INT signal on this shield.  With this [touch_LCD-PAR-S035_hack.patch](https://github.com/user-attachments/files/23344665/touch_LCD-PAR-S035_hack.patch) the INT signal remains as an output driving low through both Reset pulses.  The touch works as expected.  This patch was applied to v4.3.0-rc2.
Image

DerekSnell:
Fixes #98876 

Defers init of GT911 touch, MIPI DBI, and display device drivers.  `lcd_par_s035_init()` in shield.c now initializes these to control the GT911 INT_GPIO state during the Reset sequence.

- Fixes issue with this shield caused by shared Reset signal between the GT911 and display driver.
- Removes `reset-gpios` property from `gt911` node.  The GT911 driver does not need to assert the Reset signal because the display driver asserts Reset before the GT911 driver initializes.

Tested this fix on the following:
- LCD_PAR_S035 shield with FRDM_MCXN947
- LCD_PAR_S035 shield with FRDM_RW612

This capture is from the LCD_PAR_S035 with the FRDM_RW612 and this change applied.  The GT911 INT_GPIO is driven low first, then the LCDIC display driver asserts Reset and INT_GPIO remains low through the init process.
image

### DIFF
--- a/boards/shields/lcd_par_s035/CMakeLists.txt
+++ b/boards/shields/lcd_par_s035/CMakeLists.txt
@@ -1,0 +1,8 @@
+#
+# Copyright 2025 NXP
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+zephyr_library()
+zephyr_library_sources(shield.c)
--- a/boards/shields/lcd_par_s035/Kconfig.defconfig
+++ b/boards/shields/lcd_par_s035/Kconfig.defconfig
@@ -28,13 +28,4 @@ config INPUT_GT911_INTERRUPT
 
 endif # LVGL
 
-if INPUT
-
-# GT911 driver drives reset pin low, so it needs to initialize before
-# the display driver but after the MIPI DBI driver
-config INPUT_INIT_PRIORITY
-	default 82
-
-endif # INPUT
-
 endif # SHIELD_LCD_PAR_S035
--- a/boards/shields/lcd_par_s035/Kconfig.shield
+++ b/boards/shields/lcd_par_s035/Kconfig.shield
@@ -4,3 +4,10 @@
 config SHIELD_LCD_PAR_S035
 	def_bool $(shields_list_contains,lcd_par_s035_spi) ||\
 		$(shields_list_contains,lcd_par_s035_8080)
+
+config SHIELD_LCD_PAR_S035_INIT_PRIORITY
+	int "LCD_PAR_S035 shield init priority"
+	default 85
+	depends on SHIELD_LCD_PAR_S035
+	help
+	  SYSINIT() priority to initialize the shield
--- a/boards/shields/lcd_par_s035/lcd_par_s035_8080.overlay
+++ b/boards/shields/lcd_par_s035/lcd_par_s035_8080.overlay
@@ -12,6 +12,10 @@
 		zephyr,touch = &gt911_lcd_par_s035;
 	};
 
+	aliases {
+		mipi-dbi = &zephyr_mipi_dbi_parallel;
+	};
+
 	lvgl_pointer {
 		compatible = "zephyr,lvgl-pointer-input";
 		input = <&gt911_lcd_par_s035>;
@@ -20,6 +24,9 @@
 	};
 };
 
+/* Initialization of touch, MIPI DBI and display devices is deferred to
+ * lcd_par_s035_init() in shield.c
+ */
 &nxp_8080_touch_panel_i2c {
 	status = "okay";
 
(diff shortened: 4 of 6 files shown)


drivers/input/input_gt911.c
```diff
@@ -434,6 +434,38 @@ static int gt911_init(const struct device *dev)
 	return 0;
 }
 
+#if CONFIG_INPUT_GT911_PREPARE_INT_GPIO
```
JarmouniA:
```suggestion
#ifdef CONFIG_INPUT_GT911_PREPARE_INT_GPIO
```

JarmouniA (dismissed):
Can't you [deffer the initialization](https://docs.zephyrproject.org/latest/kernel/drivers/index.html#deferred-initialization) of display & touch controllers and do the setup in the right order in the board's (or shield's, if that's a thing) `board.c`? 
I really do not like seeing code in the touch controller driver that actually has nothing to do with it, and it's solving a problem created by the shield.

decsny:
> Can't you [deffer the initialization](https://docs.zephyrproject.org/latest/kernel/drivers/index.html#deferred-initialization) of display & touch controllers and do the setup in the right order in the board's (or shield's, if that's a thing) `board.c`? I really do not like seeing code in the touch controller driver that actually has nothing to do with it, and it's solving a problem created by the shield.

I don't see how deferred initialization fixes this. Whether you defer or not you still have to have the same patch @DerekSnell submitted here.

JarmouniA (approved):

dleach02 (approved):

mmahadevan108 (approved):

jhedberg (approved):

cfriedt (approved):

PR #99037, opened 2025-11-07
drivers: firmware: scmi: Introduce basic system power protocol

yongxu-wang15:
Add SCMI System Power Domain Protocol support with:

- System power state set operations
- Protocol version/attributes query operations
- Standard SCMI power states (shutdown, cold/warm reset, suspend)
- Graceful and forceful power transitions
- Optional NXP vendor-specific power states
- Device tree binding for "arm,scmi-system"

This enables system-wide power management through standardized SCMI interfaces for reboot and power control operations.

### DIFF
--- a/drivers/firmware/scmi/CMakeLists.txt
+++ b/drivers/firmware/scmi/CMakeLists.txt
@@ -11,6 +11,7 @@ zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SHMEM shmem.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_CLK_HELPERS clk.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_PINCTRL_HELPERS pinctrl.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_POWER_DOMAIN_HELPERS power.c)
+zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_REBOOT reboot.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SYSTEM system.c)
 
 add_subdirectory_ifdef(CONFIG_ARM_SCMI nxp)
--- a/drivers/firmware/scmi/CMakeLists.txt
+++ b/drivers/firmware/scmi/CMakeLists.txt
@@ -3,7 +3,7 @@
 zephyr_library()
 
 # SCMI core files
-zephyr_library_sources_ifdef(CONFIG_ARM_SCMI core.c)
+zephyr_library_sources_ifdef(CONFIG_ARM_SCMI core.c common.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_MAILBOX_TRANSPORT mailbox.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SHMEM shmem.c)
 
--- a/drivers/firmware/scmi/CMakeLists.txt
+++ b/drivers/firmware/scmi/CMakeLists.txt
@@ -11,5 +11,6 @@ zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SHMEM shmem.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_CLK_HELPERS clk.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_PINCTRL_HELPERS pinctrl.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_POWER_DOMAIN_HELPERS power.c)
+zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SYSTEM system.c)
 
 add_subdirectory_ifdef(CONFIG_ARM_SCMI nxp)
--- a/drivers/firmware/scmi/Kconfig
+++ b/drivers/firmware/scmi/Kconfig
@@ -48,6 +48,13 @@ config ARM_SCMI_SHMEM_INIT_PRIORITY
 	help
 	  SCMI SHMEM driver device initialization priority.
 
+config ARM_SCMI_SYSTEM
+	bool "SCMI system power protocol"
+	default y
+	depends on DT_HAS_ARM_SCMI_SYSTEM_ENABLED
+	help
+	  Enable support for SCMI system protocol functions.
+
 config ARM_SCMI_TRANSPORT_HAS_STATIC_CHANNELS
 	bool "Transport layer has static channels"
 	help
--- a/drivers/firmware/scmi/Kconfig
+++ b/drivers/firmware/scmi/Kconfig
@@ -35,6 +35,14 @@ config ARM_SCMI_POWER_DOMAIN_HELPERS
 	help
 	  Enable support for SCMI power domain protocol helper functions.
 
+config ARM_SCMI_REBOOT
+	bool "SCMI system reboot support"
+	depends on ARM_SCMI_SYSTEM && REBOOT
+	default y
+	help
+	  Enable system reboot functionality using SCMI system power protocol.
+	  This provides sys_arch_reboot() implementation for platforms with SCMI.
+
 config ARM_SCMI_SHMEM
 	bool "SCMI shared memory (SHMEM) driver"
 	default y
--- a/drivers/firmware/scmi/common.c
+++ b/drivers/firmware/scmi/common.c
@@ -124,7 +124,7 @@ int scmi_protocol_message_attributes_get(struct scmi_protocol *proto,
 	reply.len = sizeof(reply_buffer);
 	reply.content = &reply_buffer;
 
-	ret = scmi_send_message(proto, &msg, &reply, k_is_pre_kernel());
+	ret = scmi_send_message(proto, &msg, &reply, true);
 	if (ret < 0) {
 		return ret;
 	}
--- a/drivers/firmware/scmi/common.c
+++ b/drivers/firmware/scmi/common.c
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2025 NXP
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file
+ * @brief SCMI Common Protocol Commands Implementation
+ *
+ * This file contains the implementation of the SCMI commands that are
+ * common to all protocols (generic or vendor-specific) as listed in
+ * ARM SCMI Specification such as v4.0 (DEN0056F), Section 3.2.2 "Base protocol Commands".
+ *
+ * The following common commands are implemented:
+ * - PROTOCOL_VERSION (0x0): Query protocol version
+ * - PROTOCOL_ATTRIBUTES (0x1): Get protocol-specific attributes
+ * - MESSAGE_ATTRIBUTES (0x2): Query message capabilities
+ * - NEGOTIATE_PROTOCOL_VERSION (0x10): Negotiate protocol version support
+ *
+ * These commands provide standardized interfaces that can be reused across
+ * different SCMI protocol implementations, ensuring consistency and reducing
+ * code duplication.
+ *
+ * Reference: ARM System Control and Management Interface Platform Design Document
+ * Version 4.0, Document number: DEN0056F
+ * Available at: https://developer.arm.com/documentation/den0056/latest
+ */
+
+#include <string.h>
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/firmware/scmi/protocol.h>
+
+struct scmi_protocol_version_reply {
+	int32_t status;
+	uint32_t version;
+};
+
+struct scmi_protocol_attributes_reply {
+	int32_t status;
+	uint32_t attributes;
+};
+
+struct scmi_protocol_message_attributes_reply {
+	int32_t status;
+	uint32_t attributes;
+};
+
+int scmi_protocol_get_version(struct scmi_protocol *proto, uint32_t *version)
+{
+	struct scmi_protocol_version_reply reply_buffer;
+	struct scmi_message msg, reply;
+	int ret;
+
+	if (!proto || !version) {
+		return -EINVAL;
+	}
+
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_PROTOCOL_VERSION, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = 0;
+	msg.content = NULL;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	ret = scmi_send_message(proto, &msg, &reply, k_is_pre_kernel());
+	if (ret < 0) {
+		return ret;
+	}
+
+	*version = reply_buffer.version;
+
+	return scmi_status_to_errno(reply_buffer.status);
+}
+
+int scmi_protocol_attributes_get(struct scmi_protocol *proto, uint32_t *attributes)
+{
+	struct scmi_protocol_attributes_reply reply_buffer;
+	struct scmi_message msg, reply;
+	int ret;
+
+	if (!proto || !attributes) {
+		return -EINVAL;
+	}
+
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_PROTOCOL_ATTRIBUTES, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = 0;
+	msg.content = NULL;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	ret = scmi_send_message(proto, &msg, &reply, k_is_pre_kernel());
+	if (ret < 0) {
+		return ret;
+	}
+
+	*attributes = reply_buffer.attributes;
+
+	return scmi_status_to_errno(reply_buffer.status);
+}
+
+int scmi_protocol_message_attributes_get(struct scmi_protocol *proto,
+					uint32_t message_id, uint32_t *attributes)
+{
+	struct scmi_protocol_message_attributes_reply reply_buffer;
+	struct scmi_message msg, reply;
+	int ret;
+
+	if (!proto || !attributes) {
+		return -EINVAL;
+	}
+
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_MESSAGE_ATTRIBUTES, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = sizeof(message_id);
+	msg.content = &message_id;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	ret = scmi_send_message(proto, &msg, &reply, k_is_pre_kernel());
+	if (ret < 0) {
+		return ret;
+	}
+
+	*attributes = reply_buffer.attributes;
+
+	return scmi_status_to_errno(reply_buffer.status);
+}
+
+int scmi_protocol_version_negotiate(struct scmi_protocol *proto, uint32_t version)
+{
+	struct scmi_message msg, reply;
+	int32_t status;
+	int ret;
+
+	if (!proto) {
+		return -EINVAL;
+	}
+
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_NEGOTIATE_PROTOCOL_VERSION, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = sizeof(version);
+	msg.content = &version;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(status);
+	reply.content = &status;
+
+	ret = scmi_send_message(proto, &msg, &reply, k_is_pre_kernel());
+	if (ret < 0) {
+		return ret;
+	}
+
+	return scmi_status_to_errno(status);
+}
--- a/drivers/firmware/scmi/core.c
+++ b/drivers/firmware/scmi/core.c
@@ -215,6 +215,45 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
 	}
 }
 
+static int scmi_core_protocol_negotiate(struct scmi_protocol *proto)
+
+{
+	uint32_t agent_version, platform_version;
+	int ret;
+
+	if (!proto) {
+		return -EINVAL;
+	}
+
+	agent_version = proto->version;
+
+	if (!agent_version) {
+		LOG_ERR("Protocol 0x%X: Agent version not specified", proto->id);
+		return -EINVAL;
+	}
+
+	ret = scmi_protocol_get_version(proto, &platform_version);
+	if (ret < 0) {
+		LOG_ERR("Protocol 0x%X: Failed to get platform version: %d",
+			proto->id, ret);
+		return ret;
+	}
+
+	if (platform_version > agent_version) {
+		ret = scmi_protocol_version_negotiate(proto, agent_version);
+		if (ret < 0) {
+			LOG_WRN("Protocol 0x%X: Negotiation failed (%d). "
+				"Platform v0x%08x does not support downgrade to agent v0x%08x",
+				proto->id, ret, platform_version, agent_version);
+		}
+	}
+
+	LOG_INF("Using protocol 0x%X: agent version 0x%08x, platform version 0x%08x",
+			proto->id, agent_version, platform_version);
+
+	return 0;
+}
+
 static int scmi_core_protocol_setup(const struct device *transport)
 {
 	int ret;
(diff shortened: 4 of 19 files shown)


drivers/firmware/scmi/reboot.c
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if defined(CONFIG_ARM_SCMI_REBOOT)
+
+#include <zephyr/drivers/firmware/scmi/system.h>
+#include <string.h>
+
+#if defined(CONFIG_REBOOT)
```
dsseng: `config ARM_SCMI_REBOOT` `depends on ARM_SCMI_SYSTEM && REBOOT`

Thus, this #if is always selected. I believe it should be safe to remove these conditions

drivers/firmware/scmi/reboot.c
```diff
@@ -0,0 +1,86 @@
...
+	uint32_t version, mesg_attr;
+
+	/* Query SCMI system protocol version */
+	ret = scmi_system_protocol_version(&version);
+	if (ret < 0) {
+		LOG_WRN("Failed to query SCMI system protocol version: %d", ret);
+		return 0;
+	}
+
+	LOG_INF("SCMI system protocol version: 0x%08x", version);
```
dsseng: Should we maybe check the version and negotiate if it's not in the supported range? 3.4.2.2 part of the spec says that talking a version other than reported one is a best-effort approach.

But probably the protocol is going to stay backwards compatible, so this can likely be avoided for now (given Zephyr integrators have close control on TF-A version being used in the system)

dsseng (commented):
Thanks for sharing this code not only between SoCs but also generalizing it!

drivers/firmware/scmi/reboot.c
yongxu-wang15: yes, it can be removed now, done

drivers/firmware/scmi/reboot.c
yongxu-wang15: yes, its needed to add some version check and negotiate in protocol init phase, have added this flow in newly commit, please help to review it, thanks

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,47 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+	int ret;
+
+	if (!proto) {
+		return -EINVAL;
+	}
+
+	protocol_version = scmi_protocol_get_supported_version(proto);
+
+	if (protocol_version) {
+		ret = scmi_protocol_version_get(proto, &agent_version);
```
dbaluta: reduce one level of indentation here by doing:

```
if (!protocol_version(
       return 0
```

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,47 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+		if (ret < 0) {
+			LOG_ERR("Protocol 0x%X: Failed to get agent version: %d",
+				proto->id, ret);
```
dbaluta: Here you have two function names that are not consistent:
-> scmi_protocol_get_supported_version
-> scmi_protocol_version_get

Maybe second one could be `scmi_protocol_get_version`?

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,47 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+					proto->id, protocol_version);
+			} else {
+				LOG_ERR("Protocol 0x%X: Negotiation failed (%d)", proto->id, ret);
+				return ret;
+			}
+		} else {
+			LOG_INF("Protocol 0x%X: Platform version 0x%08x is compatible",
+				proto->id, agent_version);
+		}
+	}
```
dbaluta, line 249: I think the INFO could be summarized at the end of the function and avoid all log_inf sprinkled all over the code. 

E.g LOG_INF("Using protocol %0x%X: platform version %0x: agent version %0x")

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	uint32_t attributes;
+};
+
+struct scmi_protocol_message_attributes_reply {
+	int32_t status;
+	uint32_t attributes;
+};
+
+int scmi_protocol_version_get(struct scmi_protocol *proto, uint32_t *version)
+{
```
dbaluta, line 50: Please add comments on top of each function explaining what it does, input params, expected output params.

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,47 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+
+			ret = scmi_protocol_version_negotiate(proto, protocol_version);
+			if (ret == 0) {
+				LOG_INF("Protocol 0x%X: Successfully negotiated to version 0x%08x",
+					proto->id, protocol_version);
+			} else {
+				LOG_ERR("Protocol 0x%X: Negotiation failed (%d)", proto->id, ret);
+				return ret;
+			}
+		} else {
```
dsseng: The case when our version is higher than the platform's exposed version (firmware is too old) seems to falsely fall into the compatible category.

drivers/firmware/scmi/core.c
yongxu-wang15: sure, used it now

drivers/firmware/scmi/core.c
yongxu-wang15: okay, updated into scmi_protocol_get_version, thanks

drivers/firmware/scmi/core.c
yongxu-wang15: sure, removed some duplicated LOG in there, only keep LOG_INF in last and LG_WRN if negotiate fail

drivers/firmware/scmi/common.c
yongxu-wang15: protocol.h file have added these api define description, plase check it in that, thanks

drivers/firmware/scmi/core.c
yongxu-wang15: removed this flow, only keep LOG_INF in last to record two side protocol version, thanks

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	msg.len = 0;
+	msg.content = NULL;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	use_polling = k_is_pre_kernel();
+
+	ret = scmi_send_message(proto, &msg, &reply, use_polling);
```
LaurentiuM1234: can just use `k_is_pre_kernel()` directly here instead of `use_polling`?

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	if (ret < 0) {
+		return ret;
+	}
+
+	*version = reply_buffer.version;
```
LaurentiuM1234, line 73: space here, please

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_PROTOCOL_ATTRIBUTES, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = 0;
+	msg.content = NULL;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	use_polling = k_is_pre_kernel();
```
LaurentiuM1234: ditto

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+
+	ret = scmi_send_message(proto, &msg, &reply, use_polling);
+	if (ret < 0) {
+		return ret;
+	}
+
+	*attributes = reply_buffer.attributes;
```
LaurentiuM1234, line 102: ditto

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	msg.len = sizeof(message_id);
+	msg.content = &message_id;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(reply_buffer);
+	reply.content = &reply_buffer;
+
+	use_polling = k_is_pre_kernel();
+
+	ret = scmi_send_message(proto, &msg, &reply, use_polling);
```
LaurentiuM1234: ditto

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	if (ret < 0) {
+		return ret;
+	}
+
+	*attributes = reply_buffer.attributes;
```
LaurentiuM1234, line 132: ditto

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	msg.hdr = SCMI_MESSAGE_HDR_MAKE(SCMI_MSG_NEGOTIATE_PROTOCOL_VERSION, SCMI_COMMAND,
+			proto->id, 0x0);
+	msg.len = sizeof(desired_version);
+	msg.content = &desired_version;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(status);
+	reply.content = &status;
+
+	use_polling = k_is_pre_kernel();
```
LaurentiuM1234: ditto

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
+/*
+ * Copyright 2025 NXP
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
LaurentiuM1234, line 4: maybe it would be a good idea to also add some sort of description here?

something like: this file contains the implementation of the SCMI commands that are common to all protocols (generic or vendor) as listed in SCMI spec. "insert version you've used as reference here", "insert section from SCMI spec which describes these common commands".

this way, we can clarify what is supposed to go in this file and what's not

drivers/firmware/scmi/common.c
```diff
@@ -0,0 +1,148 @@
...
+	ret = scmi_send_message(proto, &msg, &reply, use_polling);
+	if (ret < 0) {
+		return ret;
+	}
+
+	*attributes = reply_buffer.attributes;
+	return scmi_status_to_errno(reply_buffer.status);
+}
+
+int scmi_protocol_version_negotiate(struct scmi_protocol *proto, uint32_t desired_version)
```
LaurentiuM1234: suggestion: change `desired_version` to `version` to align with the parameter name described in the spec? 

I would expect the function's description to be enough to indicate that this parameter is what protocol version we want to use

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -126,4 +136,46 @@ int scmi_send_message(struct scmi_protocol *proto,
 		      struct scmi_message *msg, struct scmi_message *reply,
 		      bool use_polling);
 
+/**
+ * @brief Get protocol version
+ *
+ * @param proto Protocol instance
+ * @param version Pointer to store protocol version
+ * @retval 0 if successful
```
LaurentiuM1234, line 147: space between param documentation and return value doc?

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -126,4 +136,46 @@ int scmi_send_message(struct scmi_protocol *proto,
...
+ * @retval negative errno if failure
+ */
+int scmi_protocol_get_version(struct scmi_protocol *proto, uint32_t *version);
+
+/**
+ * @brief Get protocol attributes
+ *
+ * @param proto Protocol instance
+ * @param attributes Pointer to store protocol attributes
+ * @retval 0 if successful
```
LaurentiuM1234, line 158: ditto

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -126,4 +136,46 @@ int scmi_send_message(struct scmi_protocol *proto,
...
+ */
+int scmi_protocol_attributes_get(struct scmi_protocol *proto, uint32_t *attributes);
+
+/**
+ * @brief Get protocol message attributes
+ *
+ * @param proto Protocol instance
+ * @param message_id Message ID to query
+ * @param attributes Pointer to store message attributes
+ * @retval 0 if successful
```
LaurentiuM1234, line 170: ditto

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -126,4 +136,46 @@ int scmi_send_message(struct scmi_protocol *proto,
...
+ */
+int scmi_protocol_message_attributes_get(struct scmi_protocol *proto,
+					uint32_t message_id, uint32_t *attributes);
+
+/**
+ * @brief Negotiate protocol version
+ *
+ * @param proto Protocol instance
+ * @param desired_version Desired protocol version
+ * @retval 0 if successful
```
LaurentiuM1234, line 182: ditto

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -76,6 +76,17 @@ enum scmi_common_cmd {
 	SCMI_MSG_NEGOTIATE_PROTOCOL_VERSION = 0x10,
 };
 
+/**
+ * @struct scmi_protocol_data
+ * @brief Common protocol data structure with version info
+ */
+struct scmi_protocol_data {
```
LaurentiuM1234: I believe this is broken at the moment?

You want to force all protocol drivers to encapsulate their private data using this structure and yet we still have protocols such as the clock protocol which doesn't yet use this structure.

Have you tested the protocol negotiation for clock control?

My suggestion would be to incorporate the protocol version directly into `struct scmi_protocol` and leave the `data` filed as-is w/ no restriction regarding the format. This would also mean modifying the protocol macros to add the version parameter but I think the change should be trivial?

If you do this, please make sure you do so in a separate commit, which only handles the addition of the `version` parameter to `struct scmi_protocol` and the aforementioned macros. To fill in the protocol versions for clock and pin control supported by Zephyr, please use SCMI spec. v3.2 (DEN0056E). This was the spec. version used when the SCMI support was initially introduced.

include/zephyr/drivers/firmware/scmi/protocol.h
```diff
@@ -178,4 +189,23 @@ int scmi_protocol_message_attributes_get(struct scmi_protocol *proto,
...
 int scmi_protocol_version_negotiate(struct scmi_protocol *proto, uint32_t desired_version);
 
+/**
+ * @brief Get protocol version from data field
+ *
+ * @param proto pointer to SCMI protocol
+ * @retval 0 if successful
+ * @retval negative errno if failure
+ */
+static inline uint32_t scmi_protocol_get_supported_version(struct scmi_protocol *proto)
```
LaurentiuM1234: don't see much use to this helper? just access "proto->version" directly if you go for the aforementioned suggested change.

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,41 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+		return -EINVAL;
+	}
+
+	protocol_version = scmi_protocol_get_supported_version(proto);
+
+	if (!protocol_version) {
+		return 0;
+	}
+
+	ret = scmi_protocol_get_version(proto, &agent_version);
```
LaurentiuM1234: you're querying the platform not the agent here

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,41 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+static int scmi_core_protocol_negotiate(struct scmi_protocol *proto)
+{
+	uint32_t agent_version, protocol_version;
+	int ret;
+
+	if (!proto) {
+		return -EINVAL;
+	}
+
+	protocol_version = scmi_protocol_get_supported_version(proto);
```
LaurentiuM1234: this would be the agent

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,41 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+
+	if (!protocol_version) {
```
LaurentiuM1234: should a NULL protocol version be treated as a successful protocol negotiation? intuitively, I'd say NULL would mean the protocol driver writer forgot to fill in the protocol version?

drivers/firmware/scmi/core.c
```diff
@@ -215,6 +215,41 @@ int scmi_send_message(struct scmi_protocol *proto, struct scmi_message *msg,
...
+	if (ret < 0) {
+		LOG_ERR("Protocol 0x%X: Failed to get agent version: %d",
+			proto->id, ret);
+		return ret;
+	}
+
+	if (agent_version > protocol_version) {
+		ret = scmi_protocol_version_negotiate(proto, protocol_version);
+		if (ret < 0) {
+			LOG_WRN("Protocol 0x%X: Negotiation failed (%d)", proto->id, ret);
```
LaurentiuM1234: I would also add a warning regarding the meaning of the failed negotiation (i.e. platform protocol version might not be backward compatible).

For reference, this is how Linux did it: https://elixir.bootlin.com/linux/v6.17.8/source/drivers/firmware/arm_scmi/driver.c (line 2187)

drivers/firmware/scmi/Kconfig
```diff
@@ -35,6 +35,13 @@ config ARM_SCMI_POWER_DOMAIN_HELPERS
 	help
 	  Enable support for SCMI power domain protocol helper functions.
 
+config ARM_SCMI_SYSTEM
```
LaurentiuM1234: keep alphabetically ordered

include/zephyr/drivers/firmware/scmi/nxp/system.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @name NXP Vendor System Power States
+ * @{
+ */
+#define SCMI_SYSTEM_POWER_STATE_NXP_WAKE           0x80000000U
```
LaurentiuM1234: suggestion: `SCMI_NXP_SYSTEM_`

drivers/firmware/scmi/reboot.c
```diff
@@ -0,0 +1,70 @@
...
+#include <zephyr/sys/reboot.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(scmi_reboot, CONFIG_PM_LOG_LEVEL);
+
+static int scmi_reboot_handler(int type)
+{
+	struct scmi_system_power_state_config cfg;
+	int ret;
+
```
LaurentiuM1234, line 19: should you not check if the reboot type is supported by the platform here? I see you check for both COLD and WARM reboots in `init()` but you don't seem to be doing anything with that information other than some logging?

drivers/firmware/scmi/common.c
yongxu-wang15: sure, its enough expectly scmi_protocol_message_attributes_get which used in reboot context, irq is diabled in system reboot level, so that apply poll mode for it in newly commit

drivers/firmware/scmi/common.c
yongxu-wang15: okay, added more description now

drivers/firmware/scmi/common.c
yongxu-wang15: sure, added it

drivers/firmware/scmi/common.c
yongxu-wang15: done

drivers/firmware/scmi/common.c
yongxu-wang15: use poll mode in there as it will be used in reboot context

drivers/firmware/scmi/common.c
yongxu-wang15: done

drivers/firmware/scmi/reboot.c
yongxu-wang15: okay, removed LOG INF in init and go to check if platform is supported warm reboot before send this command

include/zephyr/drivers/firmware/scmi/protocol.h
yongxu-wang15: sure, its should be add space, done

include/zephyr/drivers/firmware/scmi/protocol.h
yongxu-wang15: sure

include/zephyr/drivers/firmware/scmi/protocol.h
yongxu-wang15: sure, updated

include/zephyr/drivers/firmware/scmi/protocol.h
yongxu-wang15: added it now

drivers/firmware/scmi/Kconfig
yongxu-wang15: yes, updated it

drivers/firmware/scmi/core.c
yongxu-wang15: okay, add more warning log in there

drivers/firmware/scmi/core.c
yongxu-wang15: okay, add LOG_ERR in there

drivers/firmware/scmi/common.c
yongxu-wang15: sure, used version  now

include/zephyr/drivers/firmware/scmi/protocol.h
yongxu-wang15: thanks suggestion, its more better flow, I have updated it from your comments, and add all scmi protocol version

yongxu-wang15:
Rebased to fix file confict, thanks

drivers/firmware/scmi/reboot.c
```diff
@@ -0,0 +1,59 @@
...
+{
+	struct scmi_system_power_state_config cfg;
+	int ret;
+	uint32_t mesg_attr;
+
+	cfg.flags = SCMI_SYSTEM_POWER_FLAG_FORCEFUL;
+
+	switch (type) {
+	case SYS_REBOOT_WARM:
+		ret = scmi_system_protocol_message_attributes(SCMI_SYSTEM_MSG_POWER_STATE_SET,
```
LaurentiuM1234, line 24: should you not do this check for COLD reboot as well?

drivers/firmware/scmi/CMakeLists.txt
```diff
@@ -12,5 +12,6 @@ zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_CLK_HELPERS clk.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_PINCTRL_HELPERS pinctrl.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_POWER_DOMAIN_HELPERS power.c)
 zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_SYSTEM system.c)
+zephyr_library_sources_ifdef(CONFIG_ARM_SCMI_REBOOT reboot.c)
```
LaurentiuM1234, line 14: keep sorted alphabetically?

drivers/firmware/scmi/system.c
```diff
@@ -0,0 +1,76 @@
...
+	msg.len = sizeof(*cfg);
+	msg.content = cfg;
+
+	reply.hdr = msg.hdr;
+	reply.len = sizeof(status);
+	reply.content = &status;
+
+	use_polling = k_is_pre_kernel();
+
+	ret = scmi_send_message(proto, &msg, &reply, use_polling);
```
LaurentiuM1234, line 70: use `k_is_pre_kernel()` directly?

LaurentiuM1234 (commented):
Looking good. Left some final nits below.

drivers/firmware/scmi/reboot.c
yongxu-wang15: as scmi doc not define this cold reboot bit to determine it now, it just had warm reboot and suspend bit, so not added it in this commit to follow it

drivers/firmware/scmi/CMakeLists.txt
yongxu-wang15: sure, done

drivers/firmware/scmi/system.c
yongxu-wang15: ok, updated

LaurentiuM1234 (approved):
LGTM, thank you for your contribution!

yongxu-wang15:
ping: @dbaluta @dsseng, hi can you help review this pr again? I think the comments left before have been resolved, thanks

dsseng (commented):
No extra comments on my side, all good. However I do not know any other relevant reviewers to add, sorry

iuliana-prodan (approved):

PR #99038, opened 2025-11-07
pm: shell: add warm cold system reboot commands

yongxu-wang15:
Add 'pm cpu warm' and 'pm cpu cold' shell commands to trigger system-wide reboot operations through sys_reboot() interface.

### DIFF
--- a/subsys/pm/pm_cpu_shell.c
+++ b/subsys/pm/pm_cpu_shell.c
@@ -8,6 +8,7 @@
 #include <zephyr/shell/shell.h>
 #include <zephyr/pm/pm.h>
 #include <zephyr/pm/policy.h>
+#include <zephyr/sys/reboot.h>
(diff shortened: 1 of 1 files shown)


pdgendt (commented):
How is this different from `kernel reboot warm|cold`?

yongxu-wang15:
> How is this different from `kernel reboot warm|cold`?


zephyr kernel not has precise document about the difference between warm and cold, I think its should be conneted with specific soc type, https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/sys/reboot.h

bjarki-andreasen (changes requested):
sys_reboot is not covered by the PM subsystem. PM may effectively reboot if s2ram or similar states are entered, but sys_reboot is not necessarily used in these cases.

PR #99043, opened 2025-11-07
boards: renesas: add support for Renesas RA MCK-RA4T1

thenguyenyf:
- Add support for MCK-RA4T1 board
- Updated samples, tested to run with supported drivers on this board


soc/renesas/ra/ra4t1/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2025 Renesas Electronics Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+zephyr_include_directories(.)
+
+zephyr_sources(
+  soc.c
+)
```
nordicjm: nit. this can go on one line

soc/renesas/ra/ra4t1/soc.c
```diff
@@ -0,0 +1,76 @@
...
+#include <zephyr/irq.h>
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
+
+#include "bsp_cfg.h"
+#include <bsp_api.h>
+
+uint32_t SystemCoreClock BSP_SECTION_EARLY_INIT;
+
+volatile uint32_t g_protect_pfswe_counter BSP_SECTION_EARLY_INIT;
```
nordicjm, line 27: why is this not static (line 27)?

soc/renesas/ra/ra4t1/soc.c
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * This needs to be run from the very beginning.
+ */
+void soc_early_init_hook(void)
+{
+	extern volatile uint16_t g_protect_counters[];
+
+	for (uint32_t i = 0; i < 5; i++) {
+		g_protect_counters[i] = 0;
+	}
```
nordicjm, line 45: no idea what this is, doesn't appear to be used anywhere else in whole PR
```suggestion
```

soc/renesas/ra/ra4t1/soc.c
```diff
@@ -0,0 +1,76 @@
...
+
+	/* Enable protection using PRCR register. */
+	R_BSP_RegisterProtectEnable(BSP_REG_PROTECT_SAR);
+#endif
+
+	SystemCoreClock = BSP_MOCO_HZ;
+	g_protect_pfswe_counter = 0;
+
+#ifdef CONFIG_RUNTIME_NMI
+	for (uint32_t i = 0; i < 16; i++) {
```
nordicjm, line 70: @wearyzen can you have a review of this and check this is OK?

boards/renesas/mck_ra4t1/doc/index.rst
```diff
@@ -0,0 +1,127 @@
...
+==================
+
+.. zephyr:board-supported-hw::
+
+Programming and Debugging
+*************************
+
+.. zephyr:board-supported-runners::
+
+Applications for the ``mck_ra4t1`` board configuration can be built, flashed, and debugged in the usual way. See
```
nordicjm: -configuration

boards/renesas/mck_ra4t1/doc/index.rst
```diff
@@ -0,0 +1,127 @@
...
+
+1. Connect to J-Link OB via USB port to host PC
+
+2. Make sure J-Link OB jumper is in default configuration as describe in `MCB-RA4T1 - User's Manual`_
+
+3. Execute west command
+
+	.. code-block:: console
+
+		west flash -r jlink
```
nordicjm: indent looks very wrong, also why is this needed when the above hello world bit does this already?

tests/drivers/can/api/boards/mck_ra4t1.overlay
henrikbrixandersen, line 1: This should  be added on the board level, not under a test-specific overlay.

tests/drivers/can/timing/boards/mck_ra4t1.overlay
henrikbrixandersen, line 1: This should  be added on the board level, not under a test-specific overlay.

henrikbrixandersen (dismissed):
Many of the sample overlays looks like they ought to be added on the board level instead.

dts/arm/renesas/ra/ra4/r7fa4t1bx_48.dtsi
khoa-nguyen-18, line 1: I don’t think it’s necessary to create a separate `.dtsi` file for this type. From my perspective, to keep it as simple as possible, we only need three layers for the SoC: the common `.dtsi` → flash type `.dtsi` → specific `.dtsi`.

soc/renesas/ra/ra4t1/soc.c
```diff
@@ -0,0 +1,76 @@
...
+#include <zephyr/init.h>
+#include <zephyr/kernel.h>
+#include <zephyr/arch/cpu.h>
+#include <cmsis_core.h>
+#include <zephyr/arch/arm/nmi.h>
+#include <zephyr/irq.h>
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
+
+#include "bsp_cfg.h"
```
khoa-nguyen-18:
```suggestion
#include <bsp_cfg.h>
```

soc/renesas/ra/ra4t1/soc.c
```diff
@@ -0,0 +1,76 @@
...
+#include <bsp_api.h>
```
khoa-nguyen-18:
```suggestion
#include "soc.h"
```

boards/renesas/mck_ra4t1/mck_ra4t1.dts
```diff
@@ -0,0 +1,87 @@
...
+	model = "Renesas MCK-RA4T1";
+	compatible = "renesas,ra4t1", "renesas,ra";
+
+	chosen {
+		zephyr,sram = &sram0;
+		zephyr,flash = &flash0;
+		zephyr,console = &uart9;
+		zephyr,shell-uart = &uart9;
+		zephyr,entropy = &trng;
+		zephyr,flash-controller = &flash1;
```
khoa-nguyen-18:
```suggestion
		zephyr,flash-controller = &flash;
```

tests/drivers/can/api/boards/mck_ra4t1.overlay
thenguyenyf: This board does not equipe a CAN transceiver, so I don't enable CANFD node on board by default.

tests/drivers/can/timing/boards/mck_ra4t1.overlay
thenguyenyf: Same as above

soc/renesas/ra/ra4t1/soc.c
thenguyenyf: This is a re-declaration for a global context used in another place so it cannot be added with the static keyword.

tests/drivers/can/api/boards/mck_ra4t1.overlay
henrikbrixandersen: I don't see a reason for enabling the CAN tests, then?

soc/renesas/ra/ra4t1/soc.c
thenguyenyf: As above explanation, this is a global context that used in another place, and it should be properly assigned an initial value. We are considering moving it into more suitable place. But right now, for this board support, I think we should keep it as it.

nordicjm: I ctrl+f'd through the PR and found no other instance of usage, so where it is used? Why does it not work using the zephyr driver model?

dts/arm/renesas/ra/ra4/r7fa4t1bx_48.dtsi
thenguyenyf: By software view, the grouping by pin num is fit with describe in HWM for all part numbering, that:
- Feature set: specific how many HWIP was added on this PN
- Pin num: specific which GPIO port and which pin existed on this PN
- Flash size: specific its code flash size on this PN

So, for RA4T1 this is the ideal way to optimize maintenance effort and avoid duplicate definitions for each PN, that usually meet in case of the ports/pins are identical for all PNs belonging to the product with the same pin count.

boards/renesas/mck_ra4t1/mck_ra4t1.dts
thenguyenyf: Not sure if it's a bug/issue, but all RA4 device with `renesas,ra-flash-hp-controller` have the same definition. If it's correct, should we send a PR to correct it for all other boards?

soc/renesas/ra/ra4t1/soc.c
thenguyenyf: It is used for a BSP function placed in HAL: `R_BSP_RegisterProtectEnable`

soc/renesas/ra/ra4t1/CMakeLists.txt
thenguyenyf: I fixed it

boards/renesas/mck_ra4t1/doc/index.rst
thenguyenyf: I fixed it

boards/renesas/mck_ra4t1/doc/index.rst
thenguyenyf: Updated

tests/drivers/can/api/boards/mck_ra4t1.overlay
thenguyenyf: After consideration, I have removed these tests as the CAN controller is not enabled on the board. Please check it.

soc/renesas/ra/ra4t1/soc.c
thenguyenyf: I fixed it

soc/renesas/ra/ra4t1/soc.c
thenguyenyf: I fixed it

boards/renesas/mck_ra4t1/mck_ra4t1.dts
thenguyenyf: It is fixed

dts/arm/renesas/ra/ra4/r7fa4t1bx_48.dtsi
khoa-nguyen-18: Agree, but I have some comments:
- The name of this pin-number layer should not include the feature (e.g., `r7fa4t1xx...dtsi`), because we just need a specific pin count file that defines for a feature set variant. In some SoCs with multiple feature sets, this allows the file to be reused.
- The name of this pin-number layer should not use the raw pin number, because some SoCs may have different Part number but have the same pin count, such as the `RA6M5`.
- I have a concern about defining the pin-number layer in `RA8D1` and `RA8D2`, since unlike the other SoCs, they have multiple separate I/O Port numbering schemes. Could you please check this?


image

image

thenguyenyf: > The name of this pin-number layer should not include the feature (e.g., r7fa4t1xx...dtsi), because we just need a specific pin count file that defines for a feature set variant. In some SoCs with multiple feature sets, this allows the file to be reused.

In case of r7fa4t1, there is only 1 feature set. So, with me, it makes enough sense to keep only 1 r7fa4t1bxxx... for this SoC support.

> The name of this pin-number layer should not use the raw pin number, because some SoCs may have different Part number but have the same pin count, such as the RA6M5.

This is not an issue, when the difference is the way this SoC package and its electrical characteristic which is not a concern in software view, only for PCB layout designer. The functionality is completely {pin count} x {feature set} specific for most of RA devices. For some exception, it could be {pin count} x {feature set} x ... (e.g: {package type})

> I have a concern about defining the pin-number layer in RA8D1 and RA8D2, since unlike the other SoCs, they have multiple separate I/O Port numbering schemes. Could you please check this?

I see. The device node grouping on RA8D1 and RA8D2 may have some difference, but we can easily add small change by overlay/delete some node in each own {pn}.dtsi or {feature set}.dtsi. I believe this will have no significant impact.

khoa-nguyen-18 (approved):
The support looks good to me

soc/renesas/ra/ra4t1/soc.c
wearyzen: g_bsp_group_irq_sources is a Renesas concept and they seem to be handling group irqs in a common NMI handler defined in renesas hal. I don't see any issue from arch point of view with this.

thaoluonguw (approved):
LGTM!

KhiemNguyenT (approved):

PR #99047, opened 2025-11-07
tests: can: api: remove userspace test for m2l31x

ccli8:
This target has limited number of MPU regions (4) and cannot support userspace.


henrikbrixandersen (changes requested):
This doesn't scale - many tests have optional userspace context testing.

Have you considered an approach similar to https://github.com/zephyrproject-rtos/zephyr/blob/c102a8e30e164a90a0f9dfc6ffc0ef529d0bac15/boards/nxp/twr_ke18f/Kconfig.defconfig#L12 ?

PR #99055, opened 2025-11-07
CAN: MCAN: Implement CAN_MODE_ONE_SHOT

bjarki-andreasen:
Implement CAN_MODE_ONE_SHOT

Notes:
Need to find a good way to test one shot


henrikbrixandersen:
I'd suggest reaching out to the subsystem maintainer prior to embarking on something like this. 

Adding the one-shot mode to M_CAN has been attempted a number of times, but each of them failed to meet the CAN API contracts. The current code in this PR is a hack with high complexity for a little used, optional feature. As such, I'm very reluctant to accept it. 

See also https://github.com/zephyrproject-rtos/zephyr/issues/79780

PR #99057, opened 2025-11-07
Enable lpcac for LPC55S36

Holt-Sun:
Enable LPCAC driver in LPC55S36.


dleach02 (dismissed):

drivers/cache/Kconfig.nxp_lpcac_n4a
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+config CACHE_NXP_LPCAC_N4A
+	bool "NXP external cache driver for nxp lpcac n4a controller"
```
ZhaoxiangJin: Is N4A an abbreviation? It would be best to write the full name of N4A here.

modules/hal_nxp/mcux/Kconfig.mcux
```diff
@@ -1,7 +1,7 @@
 # MCUXpresso SDK
 
 # Copyright (c) 2016, Freescale Semiconductor, Inc.
-# Copyright 2024 NXP
+# Copyright 2024, 2025 NXP
```
ZhaoxiangJin: 2024-2025

drivers/cache/cache_nxp_lpcac_n4a.c
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/cache.h>
+#include <zephyr/logging/log.h>
+#include <fsl_cache_lpcac.h>
+
+LOG_MODULE_REGISTER(cache_nxp_lpcac_n4a, CONFIG_CACHE_LOG_LEVEL);
```
ZhaoxiangJin: log is not used in this file, maybe can be removed?

ZhaoxiangJin (dismissed):

drivers/cache/cache_nxp_lpcac_n4a.c
Holt-Sun: Removed.

drivers/cache/Kconfig.nxp_lpcac_n4a
Holt-Sun: Good catch. It should be syscon_lpcac. Updated.

dleach02 (dismissed):

JiafeiPan (dismissed):

drivers/cache/cache_nxp_syscon_lpcac.c
```diff
@@ -0,0 +1,58 @@
...
+int cache_instr_flush_and_invd_all(void)
+{
+	return -ENOTSUP;
+}
+
+int cache_instr_flush_range(void *addr, size_t size)
+{
+	ARG_UNUSED(addr);
+	ARG_UNUSED(size);
+	return -ENOTSUP;
```
butok, line 43: Should it really return an error, or maybe just do nothing and return 0?
The question applies to all "dummy" functions.

Holt-Sun: It should return an error.
The public API explicitly documents -ENOTSUP for “not supported”. A 0 return value would falsely report success and could hide logic errors in callers that rely on range maintenance after dynamic code generation, flash writes, or relocation.

Albort12138: How about introducing a new Kconfig option to indicate whether the cache supports operations based on an address and size?

Holt-Sun: Done, please review.

Albort12138 (approved):

ZhaoxiangJin (dismissed):
the 4.4 release notes should be updated I guess.

include/zephyr/cache.h
```diff
@@ -306,7 +306,7 @@ static ALWAYS_INLINE int z_impl_sys_cache_data_invd_range(void *addr, size_t siz
  */
 static ALWAYS_INLINE int sys_cache_instr_invd_range(void *addr, size_t size)
 {
-#if defined(CONFIG_CACHE_MANAGEMENT) && defined(CONFIG_ICACHE)
+#if defined(CONFIG_CACHE_MANAGEMENT) && defined(CONFIG_ICACHE) && defined(CONFIG_ICACHE_RANGE_OPS)
 	return cache_instr_invd_range(addr, size);
 #endif
```
JiafeiPan, line 311: I don't think it is needed to introduce a new Kconfig item "CONFIG_ICACHE_RANGE_OPS", there are already some drivers which also don't support such operation, but they defined a simple function just return -ENOTSUP (for example: [cache_andes.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/cache/cache_andes.c#L486)
And the header file of cache: include/zephyr/drivers/cache.h provides all the definitions of all these APIs, maybe there will be no building issue although the driver don't implement all of them, but I think we need to align the header file and driver.

drivers/cache/cache_nxp_syscon_lpcac.c
Holt-Sun: @Albort12138 , since Zephyr already exposes the cache API unconditionally. A new Kconfig adds fragmentation and build complexity without improving user experience or code size meaningfully. I reverted the patch to add CONFIG_ICACHE_RANGE_OPS to keep it simple.
If there are real needs, I will do the update in later commits.

JiafeiPan (dismissed):

Albort12138 (approved):

Albort12138 (approved):

ZhaoxiangJin (dismissed):

zejiang0jason (approved):

ZhaoxiangJin (approved):

PR #99071, opened 2025-11-07
arch: arm: core: mpu: configure ITCM region if zephyr,itcm is set

PetervdPerk-NXP:
Right now arm_mpu_regions only sets the flash and ram region in the mpu. However when using zephyr,itcm you also want to specify the the ITCM region.

This PR specifies the ITCM region if the zephyr,itcm is selected.

### DIFF
--- a/arch/arm/core/mpu/arm_mpu_regions.c
+++ b/arch/arm/core/mpu/arm_mpu_regions.c
@@ -22,7 +22,19 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 #endif
 
+#if DT_REG_SIZE(DT_CHOSEN(zephyr_itcm)) > 0 &&                                                     \
+	!DT_SAME_NODE(DT_CHOSEN(zephyr_itcm), DT_CHOSEN(zephyr_flash))
 	/* Region 1 */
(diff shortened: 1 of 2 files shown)


wearyzen (changes requested):
ITCM can have an mpu region set if its DT node has `zephyr,memory-attr = <DT_MEM_ARM_MPU_FLASH>;` 
e.g. is https://github.com/zephyrproject-rtos/zephyr/blob/da9aa8fcad1a015afa7c07e19f0216a14eb44ae7/dts/arm/nxp/nxp_rt118x_cm33.dtsi#L14.

I see a lot of boards with `zephyr,itcm` chosen and since MPU regions are limited I suspect this change could break a few board of them reporting mpu init failure. Those boards could have itcm as part of their custom mpu_region table or have the above attribute set, either way this change could impact many boards that might not want it.

JarmouniA:
> I initially thought that the memory region name zephyr,memory-region = "ITCM"; would be enough for the MPU configuration. However, it seems that zephyr,memory-attr = <DT_MEM_ARM(ATTR_MPU_FLASH)>; must also be specified.

Zephyr memory region (a linker section) is different from MPU region, and the 1st does not imply the 2nd.

PR #99072, opened 2025-11-07
boards: arduino: portenta_h7: Add JLink runner.

iabdalkader:
Adds JLink runner configuration for both Cortex-M7 and Cortex-M4 cores on the Arduino Portenta H7.


pillo79 (dismissed):

iabdalkader:
> I am not sure making j-link the default runner is intentional/desirable?

Well, at least this way it has some debugger configured, I happen to use jlink so it's my main interest. Do you want me to like add more runners? For example, openocd? I'd have to copy the "support" dir from giga though.

kartben (approved):

pillo79 (approved):

PR #99073, opened 2025-11-07
 dts: arm: imx95: specify ITCM zephyr,memory-region

PetervdPerk-NXP:
When using `zephyr,itcm` on a imx95 based board the linkage may fail due to the absence of a defined ITCM region. This issue can be resolved by specifying the ITCM region using `zephyr,memory-region`.

### DIFF
--- a/dts/arm/nxp/nxp_imx95_m7.dtsi
+++ b/dts/arm/nxp/nxp_imx95_m7.dtsi
@@ -105,8 +105,9 @@
 
 	soc {
 		itcm: itcm@0 {
-			compatible = "nxp,imx-itcm";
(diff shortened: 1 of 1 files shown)


dts/arm/nxp/nxp_imx95_m7.dtsi
```diff
@@ -105,8 +105,9 @@
```
iuliana-prodan, line 105: Please reformulate the commit message, is unclear _"When using zephyr,itcm linker would because it couldn't the region."_

iuliana-prodan (approved):

JiafeiPan (approved):

PR #99077, opened 2025-11-07
drivers: digipot: add generic interface and AD528x driver support

jocelynmass:
This patch introduces a generic digital potentiometer (digipot) driver interface in Zephyr along with an implementation for the Analog Devices AD528x series.

Key changes:
	•	Added a digipot driver API to provide a unified interface for controlling digital potentiometers.
	•	Implemented support for AD5280 and AD5282 devices, including RDAC read/write operations over I2C.
	•	Added Kconfig and device tree bindings to configure and instantiate AD528x devices.
	•	Provides basic initialization, wiper position set/get functions, and integration with Zephyr’s device model.

This enables Zephyr applications to use digital potentiometers in a standardized way, making it easier to add support for additional digipot devices in the future.

Tested on:
	•	STM32 platform using I2C devices configured via device tree.


carlescufi:
Hi @jocelynmass. Thanks for your contribution. Since this introduces a new API, it needs to go through the Architecture WG meeting. Agendas are emailed to the [devel@ mailing list](https://lists.zephyrproject.org/g/devel/messages), and additional details can be found here:
https://github.com/zephyrproject-rtos/zephyr/wiki/Architecture-Working-Group

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * @brief Shutdown the digipot wiper channel
+ *
+ * @param dev Pointer to the digipot device
+ * @param channel digipot RDAC channel
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_shutdown(const struct device *dev, uint8_t channel)
```
henrikbrixandersen: I am not sure if "shutdown" is a commonly supported operation on digital potentiometers? What is the intention?

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,134 @@
...
+ * @brief Set the digipot wiper position
+ *
+ * @param dev Pointer to the digipot device
+ * @param channel digipot RDAC channel
+ * @param pos digipot wiper position
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_set_wiper(const struct device *dev, uint8_t channel, uint16_t pos)
```
henrikbrixandersen: Many digital potentiometers only support "stepping" the wiper in either direction. While this API can be emulated by resetting the wiper during driver initialisation and keeping an internal position count, I wonder if this API is generic - or too specific to this particular device?

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,134 @@
...
+ * @brief Get the digipot wiper position
+ *
+ * @param dev Pointer to the digipot device
+ * @param channel digipot RDAC channel
+ * @param pos digipot wiper position
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_get_wiper(const struct device *dev, uint8_t channel, uint16_t *pos)
```
henrikbrixandersen: Same comment as above re: stepping vs. absolute positions.

henrikbrixandersen (commented):
Perhaps we don't need to include `_wiper` in all of these API function names? What other operations do you foresee on a digital potentiometer that do not operate on the wiper?

henrikbrixandersen (changes requested):
See comments above.

carlescufi:
> Thanks for the heads up. Unfortunately, I have a conflict at that time today. I’ll watch the recording.

@jocelynmass unfortunately we did not have the time to cover this topic today, so I will schedule it for next Tuesday.

include/zephyr/drivers/digipot.h
jocelynmass: Thanks for the comment. 
Most commonly used digital potentiometers (AD5282, MCP41xxx, MCP4251, AD525x, DS3502, etc.) support absolute wiper positioning and readback, so I chose an API that exposes the absolute set/get operations directly.
You're right that a subset of devices only provide relative increment/decrement commands. These devices could still be supported by resetting the wiper at init and maintaining an internal software position, but that might not be ideal for all drivers.
If we later need to support purely step-based digipots, we can extend the API with optional `step_up()` / step_down()` operations, or add a capability flag.
For now, the absolute API matches the majority of hardware and keeps the common use case clean.

include/zephyr/drivers/digipot.h
jocelynmass: Good question, you're right that "shutdown" is not universally supported across all digital potentiometers.
Some digipot families (e.g. Microchip MCP41xxx/42xxx and a few ADI parts) expose a hardware shutdown or Hi-Z mode, with the goal of reducing the power consumption.
I included the shutdown API because several commonly used devices support it, but I'm fine making it optional or capability-based if that makes the API more generic. We could add a feature flag (e.g. DIGIPOT_CAP_SHUTDOWN)
or remove shutdown entirely if it's considered too device-specific.
Let me know what direction you'd prefer, I can update the API accordingly.

matt-wood-ct:
Have you had any thoughts on my request for NV digipot coverage?
I'd be keen to see some way to store and recall the NV value separate from the volatile value

jocelynmass:
> MCP41U83

My apologies for not addressing your question earlier. I chose to wait for the outcome of the architecture meeting before adding further API functions, to ensure that the design evolves consistently with the broader feedback.

jocelynmass:
> Also, @jocelynmass could you sign-off-by your commits please? See [our commit message guidelines](https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-message-guidelines) to find out how to format your commit messages, and at [our contribution workflow](https://docs.zephyrproject.org/latest/contribute/guidelines.html#contribution-workflow) to understand how to update your Pull Request.

Unfortunately, I’m unable to attend the meeting as I have recurring conflicts during the Architecture call times.
I have updated the branch accordingly and added signed-off-by tags to all commits.

include/zephyr/drivers/digipot.h
henrikbrixandersen: Not sure what data you have backing up that these are among the most commonly used digital potentiometers, but there's also a different type of digital potentiometer on the market, which is simply controlled via discrete signals (e.g. https://www.analog.com/en/products/max5455.html).

I think it is worth to design an API which caters for both. For the latter type, resetting the wiper position is often not an option, as they can be pre-positioned during calibration etc.

include/zephyr/drivers/digipot.h
henrikbrixandersen: > ... with the goal of reducing the power consumption.

This sounds more like this specific driver should implement power management support?

include/zephyr/drivers/digipot.h
jocelynmass: My initial focus was on serial-controlled digipots (i2c/spi), which are also the only type integrated with linux. 
If the intent is for this API to cover both absolute-position and step-only devices, I can add optional step_up() / step_down() functions.

henrikbrixandersen: Okay, let's try without the step APIs for now and go with absolute wiper position APIs.

mbolivar (changes requested):
Please squash your fixup commit as needed

include/zephyr/drivers/digipot.h
jocelynmass: You’re right, I’m still getting familiar with Zephyr and its overall architecture. After reviewing the documentation, this approach makes complete sense. I’ll remove the shutdown API from the digipot driver for now and plan to add proper power management support in a follow-up PR.

henrikbrixandersen:
> Got it, I will change this later today

@jocelynmass Will you be returning to this?

jocelynmass:
> > Got it, I will change this later today
> 
> @jocelynmass Will you be returning to this?

Sorry for the delay, I got sidetracked with other stuff. I’m planning to finish the changes before the end of this week.

drivers/digipot/Kconfig.ad528x
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+config DIGIPOT_AD528X
+	bool "Analog Devices AD528x digipot driver"
+	default y
+	select I2C
+	depends on DT_HAS_ADI_AD5280_ENABLED \
+		|| DT_HAS_ADI_AD5282_ENABLED
+	help
+	  Enable the driver for the Analog Devices AD528x digipot
```
ttmut, line 13:
```suggestion
	  Enable the driver for the Analog Devices AD528x digipot

```
Missing newline?

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+DT_INST_FOREACH_STATUS_OKAY(INST_AD5280)
+#endif
+#undef DT_DRV_COMPAT
+
+/* Instantiate for AD5282 compatibles (if any) */
+#define DT_DRV_COMPAT adi_ad5282
+#if DT_HAS_COMPAT_STATUS_OKAY(DT_DRV_COMPAT)
+DT_INST_FOREACH_STATUS_OKAY(INST_AD5282)
+#endif
+#undef DT_DRV_COMPAT
```
ttmut, line 155:
```suggestion
#undef DT_DRV_COMPAT

```
Missing newline.

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+ * Copyright (c) 2025 Jocelyn Masserot <jocemass@gmail.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/i2c.h>
+#include <zephyr/drivers/digipot.h>
+#include <zephyr/logging/log.h>
```
ttmut, line 11:
```suggestion
```
I guess this can be removed since the driver does not use any logging utilities.

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+
+
+#define AD528X_RDAC_SELECT(x)	FIELD_PREP(BIT(7), x)
+#define AD528X_RDAC_RESET		BIT(6)
+#define AD528X_RDAC_SHUTDN		BIT(5)
+#define AD528X_RDAC_POS_NB		256
```
ttmut, line 17:
```suggestion
#define AD528X_RDAC_SELECT(x) FIELD_PREP(BIT(7), x)
#define AD528X_RDAC_RESET     BIT(6)
#define AD528X_RDAC_SHUTDN    BIT(5)
#define AD528X_RDAC_POS_NB    256
```

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+/* Helper that instantiates a driver for AD5282 (2 RDACs) */
+#define INST_AD5282(index)												\
+	static const struct ad528x_config config_ad5282_##index = {			\
+		.bus = I2C_DT_SPEC_INST_GET(index),								\
+		.rdac_nb = 2,													\
+	};																	\
+	DEVICE_DT_INST_DEFINE(index, ad528x_init, NULL, NULL,				\
+							&config_ad5282_##index,						\
+							POST_KERNEL, CONFIG_DIGIPOT_INIT_PRIORITY,	\
+							&ad528x_driver_api);
```
ttmut, line 141:
```suggestion
/* Helper that instantiates a driver for AD5280 (1 RDAC) */
#define INST_AD5280(index)                                                                         \
	static const struct ad528x_config config_ad5280_##index = {                                \
		.bus = I2C_DT_SPEC_INST_GET(index),                                                \
		.rdac_nb = 1,                                                                      \
	};                                                                                         \
	DEVICE_DT_INST_DEFINE(index, ad528x_init, NULL, NULL, &config_ad5280_##index, POST_KERNEL, \
			      CONFIG_DIGIPOT_INIT_PRIORITY, &ad528x_driver_api);

/* Helper that instantiates a driver for AD5282 (2 RDACs) */
#define INST_AD5282(index)                                                                         \
	static const struct ad528x_config config_ad5282_##index = {                                \
		.bus = I2C_DT_SPEC_INST_GET(index),                                                \
		.rdac_nb = 2,                                                                      \
	};                                                                                         \
	DEVICE_DT_INST_DEFINE(index, ad528x_init, NULL, NULL, &config_ad5282_##index, POST_KERNEL, \
			      CONFIG_DIGIPOT_INIT_PRIORITY, &ad528x_driver_api);
```
Not a dealbreaker but the escape characters look a bit disorganized.

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,109 @@
...
+ * @param channel digipot RDAC channel
+ * @param pos digipot wiper position
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_set(const struct device *dev, uint8_t channel, uint16_t pos)
+{
+    const struct digipot_driver_api *api = (const struct digipot_driver_api *)dev->api;
+    return api->set_position(dev, channel, pos);
```
ttmut, line 71:
```suggestion
	return DEVICE_API_GET(digipot, dev)->set_position(dev, channel, pos);
```

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,109 @@
...
+ * @param channel digipot RDAC channel
+ * @param pos digipot wiper position
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_get(const struct device *dev, uint8_t channel, uint16_t *pos)
+{
+    const struct digipot_driver_api *api = (const struct digipot_driver_api *)dev->api;
+    return api->get_position(dev, channel, pos);
```
ttmut, line 87:
```suggestion
	return DEVICE_API_GET(digipot, dev)->get_position(dev, channel, pos);
```

include/zephyr/drivers/digipot.h
```diff
@@ -0,0 +1,109 @@
...
+ * @param dev Pointer to the digipot device
+ * @param channel digipot RDAC channel
+ *
+ * @retval 0 if successful
+ * @retval < 0 if setting wiper failed
+ */
+static inline int digipot_reset(const struct device *dev, uint8_t channel)
+{
+    const struct digipot_driver_api *api = (const struct digipot_driver_api *)dev->api;
+    return api->reset_position(dev, channel);
```
ttmut, line 102:
```suggestion
	return DEVICE_API_GET(digipot, dev)->reset_position(dev, channel);
```

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+
+	return ad528x_write(dev, cmd, position);
+}
+
+static int ad528x_wiper_get(const struct device *dev, uint8_t channel, uint16_t *position)
+{
+	const struct ad528x_config *config = dev->config;
+	uint8_t cmd = 0;
+
+	if (channel > (config->rdac_nb-1)) {
```
ttmut, line 80:
```suggestion
	if (channel > (config->rdac_nb - 1)) {
```

drivers/digipot/digipot_ad528x.c
```diff
@@ -0,0 +1,155 @@
...
+
+	return ad528x_read(dev, cmd, position);
+}
+
+static int ad528x_wiper_reset(const struct device *dev, uint8_t channel)
+{
+	const struct ad528x_config *config = dev->config;
+	uint8_t cmd = 0;
+
+	if (channel > (config->rdac_nb-1)) {
```
ttmut, line 94:
```suggestion
	if (channel > (config->rdac_nb - 1)) {
```

ttmut (changes requested):
Adding a few comments. Thanks!

mbolivar (approved):
+1 for the DT bindings only, did not review the rest.

drivers/digipot/digipot_ad528x.c
ttmut: Ok, maybe it is a rendering issue.

PR #99080, opened 2025-11-07
driver: gpio: Add sleep delay in gpio isr

elmo9999:
When EC wakes by GPIO, it may fall back to SUSPEND_TO_IDLE before the host or peripherals have a chance to take further action.

Add a lightweight delay-sleep mechanism to keep system awake for as short period after a GPIO wake event.

### DIFF
--- a/drivers/gpio/gpio_rts5912.c
+++ b/drivers/gpio/gpio_rts5912.c
@@ -15,6 +15,9 @@
 #include "zephyr/drivers/gpio/gpio_utils.h"
 #include <zephyr/logging/log.h>
 #include <zephyr/dt-bindings/gpio/realtek-gpio.h>
+#include <zephyr/pm/pm.h>
+#include <zephyr/pm/policy.h>
+#include <zephyr/sys/atomic.h>
 
 #include <reg/reg_gpio.h>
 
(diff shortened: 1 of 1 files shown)


JasonLin-RealTek (dismissed):

drivers/gpio/gpio_rts5912.c
```diff
@@ -33,6 +36,22 @@ struct gpio_rts5912_data {
 	sys_slist_t callbacks;
 };
 
+#if defined(CONFIG_PM)
+#define GPIO_EXPIRED_TIMEOUT_MS 1000
```
keith-zephyr: 1 second seems long.  This will block sleep mode for at least 1 full second every time the CPU wakes up.

drivers/gpio/gpio_rts5912.c
```diff
@@ -455,6 +474,14 @@ static void gpio_rts5912_isr(const void *arg)
 	if (gcr[pin] & GPIO_GCR_INTSTS_Msk) {
 		gcr[pin] |= GPIO_GCR_INTSTS_Msk;
 
+#if defined(CONFIG_PM)
+		if (atomic_cas(&gpio_wake_hold, 0, 1)) {
```
keith-zephyr, line 483: Please add a comment explaining why you're blocking SUSPEND_TO_IDLE here.

elmo9999: done.

drivers/gpio/gpio_rts5912.c
elmo9999: From lid open (gpio isr) to S0->S0ix, it take ~1s in my unit, so I set 1000 ms delay.

JasonLin-RealTek (approved):

keith-zephyr (approved):

PR #99086, opened 2025-11-07
doc: extensions: domain: fix inventory on incremental builds

gmarull:
Generated objects.inv would not contain e.g. "zephyr:board" entry on incremental builds, making it impossible to use Intersphinx on other projects when rebuilding the Zephyr docset.

It can be verified by inspecting inventory like this:

```python
from sphinx.util.inventory import InventoryFile
InventoryFile.loads(open("objects.inv", "rb").read(), uri="")
inv.data.keys()
```


carlescufi (approved):

jhedberg (approved):

kartben (approved):
Not sure I would recommend doing incremental builds when boards get updated but this PR can do no harm in the meantime :)

PR #99091, opened 2025-11-07
 dts: nxp: mcxn94x: Add SoC compatible

pkoscik:
~~Harmonize board compats between `mcx_n9xx_evk/mcxn947/cpu0(/qspi)` and `mcx_n9xx_evk/mcxn947/cpu1` variants.~~

Add SoC-level compatible string to the nxp_mcxn94x_common DT file.

### DIFF
--- a/dts/arm/nxp/nxp_mcxn94x_common.dtsi
+++ b/dts/arm/nxp/nxp_mcxn94x_common.dtsi
@@ -10,6 +10,10 @@
 	chosen {
 		zephyr,entropy = &trng;
 	};
+
+	soc {
+		compatible = "nxp,mcxn94x", "nxp,mcx", "simple-bus";
+	};
 };
 
 &peripheral {


boards/nxp/mcx_nx4x_evk/mcx_n9xx_evk_mcxn947_cpu0.dtsi
```diff
@@ -14,7 +14,9 @@
 #include "mcx_nx4x_evk_cpu0.dtsi"
 
 / {
-	aliases{
+	compatible = "nxp,mcxn947", "nxp,mcx";
```
butok: Just thinking. Should it be part of SOC dts?

pkoscik: Good idea, couple of notes

* `mcx_n9xx_evk_mcxn947_cpu*.dtsi` are the only `947` specific files - at least, according to filename. 

* There is [nxp_mcxn94x_common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/2da42ad250cc9533763af67c5743542f89a881e4/dts/arm/nxp/nxp_mcxn94x_common.dtsi), which (according to filename) seems to be a generic description for the `94x` series.
  * This file is currently only included in the `nxp_mcxn94x(_ns).dtsi` DTs, that are later included in the `mcx_n9xx_evk_mcxn947_cpu*.dtsi` and `mcx_n9xx_evk.dtsi` files.
  * Since `NXP MCX-N9XX-EVK` ships with `MCX-N947` - currently this file is only used to describe `947` based devices.
  * But filename suggests that this file should stay generic

I can prepare a patch to add `nxp,mcxn94x` to the `nxp_mcxn94x_common.dtsi`. This would also work for me.

butok: > Good idea

Better to wait for other people's opinions ;)

decsny: this compatible does not seem to actually matter at all, so I have no opinion.

decsny: Also, if we want to be pedantic and bikeshed over this, the root node represents the whole system, not just the SOC. So it should be named after the board.

dleach02: I agree with @butok should be down in the SOC

butok: > I agree with @butok should be down in the SOC

It should be done for all NXP boards ;)

butok: If to look at other vendors, it's based on
1. Board name
```
model = "STMicroelectronics STM32F401RE-NUCLEO board";
compatible = "st,stm32f401re-nucleo";
```
```
compatible = "nordic,nrf54lm20dk_nrf54lm20a-cpuapp";
model = "Nordic nRF54LM20 DK nRF54LM20A Application MCU";
```
2. both board and all SOCs
```
model = "Atmel SAM V71 Xplained Ultra board";
compatible = "atmel,sam_v71_xult", "atmel,samv71q21", "atmel,samv71";
```
```
model = "Arduino Nano Matter";
compatible = "arduino,arduino_nano_matter", "silabs,mgm240sd22vna";
```
I guess the 2nd option is more complete. 

BUT I do not know how and where it's used, or if it's just for informational needs. Do you know something about it?

pkoscik: Thanks for the input! 


> Just thinking. Should it be part of SOC dts?

> I agree with @butok should be down in the SOC

> Also, if we want to be pedantic and bikeshed over this, the root node represents the whole system, not just the SOC. So it should be named after the board.

Good points.

I'll prepare a patch to set this compat at the SoC level.

I assume  https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/nxp/nxp_mcxn94x.dtsi would be the place for it.

 
> If to look at other vendors, it's based on [...]

I think the cleanest way is to have both a Board (root) level and SoC level compatible fields. For example PTAL at [a DT for nucleo_l152re](https://zephyr-dashboard.renode.io/zephyr/bcc4689cd01620c1a5408986085343e656ec5a71/nucleo_l152re/hello_world/hello_world.dts):

```dts
/ {
	...
	model = "STMicroelectronics STM32L152RE-NUCLEO board"; /* in zephyr/boards/st/nucleo_l152re/nucleo_l152re.dts:15 */
	compatible = "st,stm32l152re-nucleo";                  /* in zephyr/boards/st/nucleo_l152re/nucleo_l152re.dts:16 */
	...

	soc {
		...
		compatible = "st,stm32l152",
		             "st,stm32l1",
		             "simple-bus";    /* in zephyr/dts/arm/st/l1/stm32l152.dtsi:15 */
		...
	};
};
```

> BUT I do not know how and where it's used, or if it's just for informational needs. Do you know something about it?

The compatible fields (for all nodes, not just board/SoC nodes) are used in Renode to generate platforms for Zephyr targets. Then we are able to simulate them at scale as you can see in [Renode Zephyr Dashboard](https://zephyr-dashboard.renode.io/). But the informational aspect is also important.

> It should be done for all NXP boards ;)

I agree, but I believe this would be outside of the scope of this PR

butok: It makes sense.

butok (approved):

mmahadevan108 (approved):

Issue #75965, opened 2024-07-16
WS2812 SPI first LED misbehaving


matt-wood-ct:
**Describe the bug**

When driving a strip of WS2812B LEDs from the IMXRT1040 SOC using the `WS2812_SPI` driver the first LED is not controlled

**To Reproduce**
Setup a board based on an IMXRT series process with an overlay which configures the `ws2812-spi` driver, example configuration:
```
&lpspi1 {
    led_strip: ws2812@0 {
        compatible = "worldsemi,ws2812-spi";

        /* SPI */
        reg = <0>; /* ignored, but necessary for SPI bindings */
        spi-max-frequency = <6000000>;
        reset-delay = <70>;

        /* WS2812 */
        chain-length = <6>; /* arbitrary; change at will */
        spi-one-frame = <0xF8>;
        spi-zero-frame = <0x60>;
        color-mapping = <LED_COLOR_ID_GREEN
                         LED_COLOR_ID_RED
                         LED_COLOR_ID_BLUE>;
    };
};
```
Run the LED strip driver sample.

See LEDs sequencing through colours except the first LED remains off.

**Expected behavior**
All LEDs including the first LED should go through the intended colour sequence.

**Impact**
In tree patch created, can bring the driver out of tree until fix is implemented/adopted

**Logs and console output**
N/A

**Environment (please complete the following information):**

 - OS: Windows
 - Toolchain Zephyr SDK 0.16.3
 - Zephyr 64f41309d104b161b1960bcc20e2ff6bd58c6355

**Additional context**

I have investigated the problem and create a patch which can resolve the problem in my case, but I am unclear why a similar feature is not already in tree, unless I am missing something... which I probably am 😄

What I found was that the SPI MOSI had LED data shifted out as expected but at the start of the burst the transition was not graceful so the LEDs appear to ignore the first LED worth of data. Below is a logic analyser capture of the line activity:  
![image](https://github.com/user-attachments/assets/956fb3f3-dba8-40ba-9f34-7fd08a6ffdac)

The fix I discovered was to add some bus low time at the start (80uS) to trigger a reset on the LED. After this reset the burst of data can be sent and all LEDs behave well. I achived this by adding a head an tail of zeros onto the SPI data which is sent on `ws2812_strip_update_rgb`, in my case 64 byte of zero which translated to around 85uS of low on paper and ~82uS in practice on my hardware. This yields a bus behaviour which looks like this:  
[...]

PR #99092, opened 2025-11-07
drivers: led_strip: ws2812_spi: Ensure MOSI is low during reset

arthur-mmlw:
Some platforms, such as STM32, do not guarantee that the MOSI line is held low when the SPI peripheral is disabled. This can prevent WS2812-based LEDs from resetting correctly, as they require a low signal for a specific duration to trigger a reset.

This commit introduces a Kconfig option to actively drive the MOSI line low during the reset period by sending zero-byte frames via SPI. This ensures the reset condition is reliably met across all hardware platforms.

### DIFF
--- a/drivers/led_strip/ws2812_spi.c
+++ b/drivers/led_strip/ws2812_spi.c
@@ -126,18 +126,40 @@ static int ws2812_strip_update_rgb(const struct device *dev,
 	const size_t total_bits = num_pixels * cfg->num_colors *
 				  BITS_PER_COLOR_CHANNEL * bits_per_symbol;
 	const size_t buf_len = DIV_ROUND_UP(total_bits, SPI_FRAME_BITS);
-	struct spi_buf buf = {
-		.buf = cfg->px_buf,
-		.len = buf_len,
-	};
+	struct spi_buf buffers[COND_CODE_0(CONFIG_WS2812_STRIP_SPI_RESET_DELAY_SEND_NOP, (1), (3))];
+	struct spi_buf *buf = &buffers[0];
 	const struct spi_buf_set tx = {
-		.buffers = &buf,
-		.count = 1
+		.buffers = buffers,
+		.count = ARRAY_SIZE(buffers),
 	};
+	uint32_t reset_delay_cycles;
 	uint8_t *px_buf = cfg->px_buf;
 	uint8_t bit_mask = BIT(SPI_FRAME_BITS - 1);
 	size_t i;
 	int rc;
+
+	if (IS_ENABLED(CONFIG_WS2812_STRIP_SPI_RESET_DELAY_SEND_NOP)) {
+		/*
+		 * Convert the reset delay from micro seconds to a number of cycles of the SPI
+		 * clock. The bus frequency is a maximum and may not be the actual frequency used.
+		 * In that case, the reset delay will be longer than expected but that should not be
+		 * an issue.
+		 */
+		reset_delay_cycles = cfg->bus.config.frequency * cfg->reset_delay / USEC_PER_SEC;
+
+		/*
+		 * Define dummy buffers that will send zeroes on the MOSI line for the duration of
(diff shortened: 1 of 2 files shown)


simonguinot (changes requested):
Hi @arthur-mmlw,

Thanks for this patch. But it is not the first time it is proposed. Please have a look at the links I attached in a previous comment.
 
And since my point of view didn't change, my comment is still the same:  https://github.com/zephyrproject-rtos/zephyr/issues/75965#issuecomment-2231414364.

arthur-mmlw:
@simonguinot I understand your point and I agree with you that adding a lot of padding bytes before and after the data frame is a waste of memory.

However, I'm not using more memory in this patch. Instead, I'm relying on the fact that a `NULL` pointer is interpreted as a NOP frame in the SPI driver to send zeroes. _This effectively transmits zeroes without allocating or copying data._

I've checked this behavior in the drivers `spi_bitbang.c`, `spi_ll_stm32.c`, `spi_psoc6.c` and `spi_rpi_pico_pio.c`.

I haven’t checked every SPI driver, and for some, it's unclear from the code or comments whether dummy/NOP NULL buffers are supported.

simonguinot:
> @simonguinot I understand your point and I agree with you that adding a lot of padding bytes before and after the data frame is a waste of memory.
> 
> However, I'm not using more memory in this patch. Instead, I'm relying on the fact that a `NULL` pointer is interpreted as a NOP frame in the SPI driver to send zeroes. _This effectively transmits zeroes without allocating or copying data._

Seen.
 
> I've checked this behavior in the drivers `spi_bitbang.c`, `spi_ll_stm32.c`, `spi_psoc6.c` and `spi_rpi_pico_pio.c`.
> 
> I haven’t checked every SPI driver, and for some, it's unclear from the code or comments whether dummy/NOP NULL buffers are supported.

It seems to me that the behavior with NULL buffers depends on SPI driver and controller. So it might work in some cases but not in others...

The method you propose is different but my point is still the same: it is not up to the LED driver to handle the SPI MOSI signal at idle state. It should be controlled via the SPI subsystem. And if it can't be done, then there are plenty of alternatives (UART, I2S, GPIO).

That said, I don't have a strong opinion on this subject and I will wait to see what others think.

drivers/led_strip/ws2812_spi.c
```diff
@@ -126,13 +126,31 @@ static int ws2812_strip_update_rgb(const struct device *dev,
...
+		 */
+		buffers[0].buf = NULL;
+		buffers[0].len = reset_delay_cycles / 8;
+		buf = &buffers[1];
+		buffers[2].buf = NULL;
+		buffers[2].len = reset_delay_cycles / 8;
+	}
+
+	buf->buf = cfg->px_buf;
+	buf->len = buf_len;
```
simonguinot, line 162: Please don't mix code with variables definitions. Variables must be defined at the beginning of the function.

arthur-mmlw: Fixed!

simonguinot (changes requested):
Still NAK for me

Strooom:
@simonguinot wrote : 

> And if it can't be done, then there are plenty of alternatives (UART, I2S, GPIO).

in fact for STM32 this is not the case : the I2S and GPIO drivers for led-strips are not available.

Most STM32 do have I2S peripherals, but they are not (yet) in the SOC DTS, eg STM32WLE5
For GPIO driver, it seems only to be available for nRF SOC.

So for the time being it seems SPI is the only way for STM32 and so it's useful to resolve this MOSI idle state issue.

simonguinot:
> So for the time being it seems SPI is the only way for STM32 and so it's useful to resolve this MOSI idle state issue.

Sure, but not in a generic LED driver.

Note that you may have another option soon: https://github.com/zephyrproject-rtos/zephyr/pull/106336

PR #94590, opened 2025-08-17
USB Host: integrate class API [2: helpers]

josuah:
Main thread:

- https://github.com/zephyrproject-rtos/zephyr/issues/95661

Downstream:

- https://github.com/zephyrproject-rtos/zephyr/pull/94591
- https://github.com/zephyrproject-rtos/zephyr/pull/99334
- https://github.com/zephyrproject-rtos/zephyr/pull/99173
- https://github.com/zephyrproject-rtos/zephyr/pull/99097
- https://github.com/zephyrproject-rtos/zephyr/pull/99775
- https://github.com/zephyrproject-rtos/zephyr/pull/100703

It contains all the commits of the previous PR, and in addition:

- Wrapper for the class API
- Split the `struct usbh_class_data` and `usbh_class_api`
- Add utilities to register classes as a linked list (NXP)
- Add a filter logic to match a device class to host class instances (NXP)

This also contains functions imported from #94085 by @AidenHu.

So far, this was only tested with this command to make sure it builds and run, but not in hardware yet:
```
west build -t run -b native_sim samples/subsys/usb/shell/ \
-DEXTRA_CONF_FILE=virtual.conf \
-DDTC_OVERLAY_FILE=virtual.overlay \
-DEXTRA_CONF_FILE=device_and_host_prj.conf
```

The PR #94085 got then completely rebased on top of this, available at [`josuah:enable_usb_host_video_class_api2`](https://github.com/josuah/zephyr/tree/enable_usb_host_video_class_api2), which is meant as a drop-in replacement for #94085.

It was also built (but not run) with this command:

```
west build -b rd_rw612_bga -S video-sw-generator samples/drivers/video/capture
```

All tests are done on top of https://github.com/zephyrproject-rtos/zephyr/compare/main...josuah:zephyr:pr_usb_host_class_api2

Issue #95661, opened 2025-09-08
Planned USB host support work


jfischer-no:
### Problem Description

There is a problem.

### Proposed Change (Summary)

Nothing.

### Proposed Change (Detailed)

**Important**
_This is a meta issue for maintainers and developers of relevant subsystems to track progress on USB support. Please do not ask general questions or make comments about USB support here. Please do not start a metaphysical discussion about the reasons for the existence of the host stack._

### Problem Description

Follow up on #42066, focusing on host support. For the USB device support see #95649.

Support for USB hosts in Zephyr is ramping up, but it will take some time to reach the same level as support for devices. **It is not ready for productive use. We cannot say when it will be ready, nor can we specify an exact date for the implementation of feature XYZ.** Please note that all APIs are experimental and may change at any time. Bear this in mind when working on new drivers or class drivers.

There are basic tests for both USB device and host support, which run in CI on the PR basis. We also have experimental USB/IP support, which is primarily used for exporting USB devices. There should not be any regressions, as these would also affect users of the USB device stack.

## Drivers

If you start working on the new driver, please use [MAX3421E or virtual driver](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/usb/uhc/uhc_max3421e.c) as a reference point and template for the new driver. Please follow the [coding style guidelines](https://docs.zephyrproject.org/latest/contribute/style/code.html) and keep the style consistent with the USB drivers/subsystem code. Please note that clangformat does not format code according to the coding style guidelines, so you will need to fix it afterwards.

  - [ ] New driver for controller foobar ()

## Stack

The stack is in an experimental, early-development state. Ongoing work:

  - [ ] Class (Driver) API (@josuah https://github.com/zephyrproject-rtos/zephyr/pull/94590)

## USB Class/Function driver

There is no support for any classes or functions in the Zephyr tree. However, there is ongoing work and draft PRs on some functions. There are also no examples available yet that can be used as templates. The plan is to first add a basic test function as a counterpart to the device's loopback function, which could then be used for testing and as a template for the new class driver implementations. If you want to start work on the new class/function driver, this is the right issue to discuss. Please follow the coding style guidelines and keep the style consistent with the USB drivers/subsystem code. Please note that clangformat does not format code according to the coding style guidelines, so you will need to fix it afterwards.

  - [ ] Implementation of USB classes support
      - [ ] testusb function support as loopback counterpart
[...]

PR #99097, opened 2025-11-07
subsys: usb: host: Add USB Host CDC ECM Class Support for Ethernet

santhosh-c-c:
USB Host: Add CDC ECM class support

This PR introduces the USB CDC ECM (Ethernet Control Model) class support 
to the Zephyr USB host subsystem, enabling Ethernet functionality for 
USB hosts.

Changes since draft:
- Rebased on merged dependencies #94590
- Subsys: Optimized High-Speed device initialization by defaulting 
  bMaxPacketSize0 to 64 per USB 2.0 spec (85df60a)

Dependency:
- #106802 (usb: host: fix HS device enumeration and interface probing) carries equivalent and additional USBH core fixes needed for HS device interop. The bMaxPacketSize0 commit in this PR overlaps with the first commit of 106802 and will be dropped from this PR once 106802 merges.

Signed-off-by: Santhosh Charles <santhosh@linumiz.com>

### DIFF
--- a/subsys/usb/host/class/Kconfig.cdc_ecm_host
+++ b/subsys/usb/host/class/Kconfig.cdc_ecm_host
@@ -1,0 +1,19 @@
+# Copyright (c) 2026 Linumiz
+# SPDX-License-Identifier: Apache-2.0
+
+config USBH_CDC_ECM
+	bool "USB host CDC ECM class driver"
+	depends on USB_HOST_STACK
+	depends on NET_L2_ETHERNET
+	help
+	  USB host class driver for CDC Ethernet Control Model (ECM) devices.
+
+if USBH_CDC_ECM
+
+module = USBH_CDC_ECM
+module-str = usbh cdc_ecm
+default-count = 1
+source "subsys/logging/Kconfig.template.log_config"
+source "subsys/usb/common/Kconfig.template.instances_count"
+
+endif # USBH_CDC_ECM
--- a/subsys/usb/host/class/usbh_cdc_ecm.c
+++ b/subsys/usb/host/class/usbh_cdc_ecm.c
@@ -1,0 +1,788 @@
+/*
+ * Copyright (c) 2026 Linumiz
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/net/ethernet.h>
+#include <zephyr/net/net_if.h>
+
+#include <zephyr/drivers/usb/uhc.h>
+#include <zephyr/usb/usbh.h>
+#include <zephyr/usb/usb_ch9.h>
+#include <zephyr/usb/class/usb_cdc.h>
+
+#include "usbh_class.h"
+#include "usbh_device.h"
+#include "usbh_desc.h"
+#include "usbh_ch9.h"
+
+LOG_MODULE_REGISTER(usbh_cdc_ecm, CONFIG_USBH_CDC_ECM_LOG_LEVEL);
+
+#define ECM_CTRL_MASK			BIT(0)
+#define ECM_FUNC_MASK			BIT(1)
+#define ECM_INTR_IN_EP_MASK		BIT(2)
+#define ECM_BULK_IN_EP_MASK		BIT(3)
+#define ECM_BULK_OUT_EP_MASK		BIT(4)
+#define ECM_DATA_MASK			BIT(5)
+#define ECM_UNION_MASK			BIT(6)
+#define ECM_MASK_ALL			GENMASK(6, 0)
+
+#define CDC_ECM_SEND_TIMEOUT_MS		K_MSEC(1000)
+#define CDC_ECM_ETH_PKT_FILTER_ALL	0x000F
+#define CDC_ECM_ETH_MAX_FRAME_SIZE	(NET_ETH_MTU + sizeof(struct net_eth_hdr))
+#define CDC_ECM_LANGID_EN_US		0x0409
+
+#define ECM_MAC_ADDR_LEN		NET_ETH_ADDR_LEN
+
+struct cdc_ecm_notification {
+	union {
+		uint8_t bmRequestType;
+		struct usb_req_type_field RequestType;
+	};
+	uint8_t bNotificationType;
+	uint16_t wValue;
+	uint16_t wIndex;
+	uint16_t wLength;
+} __packed;
+
+enum cdc_ecm_state {
+	CDC_ECM_DISCONNECTED,
+	CDC_ECM_CONNECTED,
+	CDC_ECM_CONFIGURED,
+	CDC_ECM_SUSPENDED,
+};
+
+/*
+ * USB string descriptor for iMACAddress:
+ * 2 byte header + 12 UTF-16LE hex characters = 26 bytes
+ */
+#define ECM_MAC_STR_DESC_LEN	(2 + ECM_MAC_ADDR_LEN * 2 * sizeof(uint16_t))
+
+struct usbh_cdc_ecm_data {
+	struct net_if *iface;
+	uint8_t mac_addr[ECM_MAC_ADDR_LEN];
+	enum cdc_ecm_state state;
+
+	struct usb_device *udev;
+	uint16_t bulk_mps;
+	uint16_t intr_mps;
+	uint8_t ctrl_iface;
+	uint8_t data_iface;
+	uint8_t bulk_in_ep;
+	uint8_t bulk_out_ep;
+	uint8_t intr_in_ep;
+	uint8_t imac_str_idx;
+
+	struct k_mutex tx_mutex;
+	struct net_stats_eth stats;
+};
+
+static int cdc_ecm_start_rx(struct usbh_cdc_ecm_data *priv);
+static int cdc_ecm_start_intr(struct usbh_cdc_ecm_data *priv);
+
+static void cleanup_xfer(struct usb_device *udev, struct uhc_transfer *xfer)
+{
+	if (xfer->buf) {
+		usbh_xfer_buf_free(udev, xfer->buf);
+	}
+
+	usbh_xfer_free(udev, xfer);
+}
+
+static int submit_xfer(struct usbh_cdc_ecm_data *priv, uint8_t ep,
+		       usbh_udev_cb_t cb, size_t buf_size,
+		       struct net_pkt *pkt)
+{
+	struct uhc_transfer *xfer;
+	struct net_buf *buf;
+	int ret;
+
+	xfer = usbh_xfer_alloc_with_buf(priv->udev, ep, cb, priv, buf_size);
+	if (!xfer) {
+		return -ENOMEM;
+	}
+
+	buf = xfer->buf;
+
+	if (pkt && buf_size > 0) {
+		if (net_pkt_read(pkt, buf->data, buf_size) != 0) {
+			cleanup_xfer(priv->udev, xfer);
+			return -EIO;
+		}
+
+		net_buf_add(buf, buf_size);
+	}
+
+	ret = usbh_xfer_enqueue(priv->udev, xfer);
+	if (ret < 0) {
+		cleanup_xfer(priv->udev, xfer);
+		return ret;
+	}
+
+	return 0;
+}
+
+static int cdc_ecm_intr_in_cb(struct usb_device *udev,
+			      struct uhc_transfer *xfer)
+{
+	struct usbh_cdc_ecm_data *priv = xfer->priv;
+	struct cdc_ecm_notification *notify;
+	int ret;
+
+	if (xfer->err != 0) {
+		LOG_DBG("Interrupt IN error: %d", xfer->err);
+		goto resubmit;
+	}
+
+	if (xfer->buf == NULL ||
+	    xfer->buf->len < sizeof(struct cdc_ecm_notification)) {
+		goto resubmit;
+	}
+
+	notify = (struct cdc_ecm_notification *)xfer->buf->data;
+
+	if (notify->bNotificationType == USB_CDC_NETWORK_CONNECTION) {
+		if (sys_le16_to_cpu(notify->wValue) != 0) {
+			net_if_carrier_on(priv->iface);
+		} else {
+			net_if_carrier_off(priv->iface);
+		}
+	}
+
+resubmit:
+	cleanup_xfer(udev, xfer);
+
+	if (priv->state == CDC_ECM_CONFIGURED) {
+		ret = cdc_ecm_start_intr(priv);
+		if (ret != 0) {
+			LOG_ERR("Failed to resubmit interrupt IN: %d", ret);
+		}
+	}
+
+	return 0;
+}
+
+static int cdc_ecm_bulk_in_cb(struct usb_device *udev,
+			      struct uhc_transfer *xfer)
+{
+	struct usbh_cdc_ecm_data *priv = xfer->priv;
+	struct net_pkt *pkt;
+	int ret;
+
+	if (xfer->err != 0) {
+		LOG_DBG("Bulk IN error: %d", xfer->err);
+		priv->stats.errors.rx++;
+		goto resubmit;
+	}
+
+	if (xfer->buf == NULL || xfer->buf->len == 0) {
+		goto resubmit;
+	}
+
+	pkt = net_pkt_rx_alloc_with_buffer(priv->iface, xfer->buf->len,
+					   AF_UNSPEC, 0, K_NO_WAIT);
+	if (pkt == NULL) {
+		LOG_DBG("Failed to allocate RX net_pkt");
+		priv->stats.errors.rx++;
+		goto resubmit;
+	}
+
+	if (net_pkt_write(pkt, xfer->buf->data, xfer->buf->len) < 0) {
+		net_pkt_unref(pkt);
+		priv->stats.errors.rx++;
+		goto resubmit;
+	}
+
+	priv->stats.bytes.received += xfer->buf->len;
+	priv->stats.pkts.rx++;
+
+	if (net_recv_data(priv->iface, pkt) < 0) {
+		net_pkt_unref(pkt);
+		priv->stats.errors.rx++;
+	}
+
+resubmit:
+	cleanup_xfer(udev, xfer);
+
+	if (priv->state == CDC_ECM_CONFIGURED) {
+		ret = cdc_ecm_start_rx(priv);
+		if (ret != 0) {
+			LOG_ERR("Failed to resubmit bulk IN: %d", ret);
+		}
+	}
+
+	return 0;
+}
+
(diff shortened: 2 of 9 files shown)


santhosh-c-c:
> > @MarkWangChinese is this related to your ongoing effort for implementing ECM?
> 
> Thanks @josuah
> 
> Hi @santhosh-c-c As the comments [#95661 (comment)](https://github.com/zephyrproject-rtos/zephyr/issues/95661#issuecomment-3490731892), our side (NXP) is trying to enable the USB Host ECM too, our plan is creating the formal pr before the end of December, we will do basic function test too. Could you tell me your plan? let's align to meet both our requirements. Thanks.

Hello @MarkWangChinese

Our code has been tested on the hardware (MIMXRT1064-EVK), and we are currently waiting for the dependency to be merged first, namely this PR: https://github.com/zephyrproject-rtos/zephyr/pull/94590. Once that merge is complete, we will proceed with merging our ECM host class support.

An internal review is ongoing and we are looking for any features that should be added. This internal review should not take more than a week. If you would like, we can discuss any additional features that need to be added so we can align with your requirements.

carlescufi:
@santhosh-c-c dependencies should now be merged. Please take out of draft and rebase on top of `main`.

santhosh-c-c:
@josuah @carlescufi @MarkWangChinese 
Ping for initial review, Thanks!

santhosh-c-c:
Hello @jfischer-no @josuah @MarkWangChinese Pinging this PR for review as discussed in the other thread https://github.com/zephyrproject-rtos/zephyr/issues/95661#issuecomment-4273996267. I'm standing by to address any comments or requested changes to help get this core support merged.

josuah (commented):
It seems like this PR is smaller, simpler which makes it easier to integrate.

With https://github.com/zephyrproject-rtos/zephyr/pull/100289 already going on, this means that some deduplication needs to happen, and using the current PR as a basis for https://github.com/zephyrproject-rtos/zephyr/pull/100289 is not feasible given the difference of structure.

Do you think it would be possible to refactor this PR to make it closer to #100289 while also applying your subjective improvements?

For instance, every time something is imported with changes applied, also propose the changes in #100289 as a review.

josuah:
It is possible to merge some commits from this PR outside of CDC ECM context.

It is possible to integrate some commits from https://github.com/zephyrproject-rtos/zephyr/pull/100289 into this current PR.

This directly contributes to progress on CDC ECM host support.

Thank you for the efforts and time on this.

santhosh-c-c:
> It seems like this PR is smaller, simpler which makes it easier to integrate.
> 
> With #100289 already going on, this means that some deduplication needs to happen, and using the current PR as a basis for #100289 is not feasible given the difference of structure.
> 
> Do you think it would be possible to refactor this PR to make it closer to #100289 while also applying your subjective improvements?
> 
> For instance, every time something is imported with changes applied, also propose the changes in #100289 as a review.

Thanks @josuah for the input. A few clarifications

1. On refactoring this PR closer to #100289: the structural difference is intentional. This PR keeps the implementation minimal and focused on the core ECM data path. #100289 layers significantly more on top (multicast filter, packet filter bitmap, hw stats, link speed, async xfer queue with cleanup list). Refactoring this PR to mirror that structure would essentially turn it into a duplicate of #100289 and defeat the purpose of having a small, reviewable base.

2. On commits outside the CDC ECM context: agreed. The `usb: host: set default HS bMaxPacketSize0 to 64` commit from this PR is a generic USBH core fix unrelated to ECM and is already being addressed independently in #106802 (along with two other HS enumeration/probing fixes from the same testing effort). I will drop that commit from this PR once #106802 lands.

3. On integrating commits from #100289 into this PR: this is the part of your suggestion I want to act on directly. The two non-driver commits in #100289 are useful regardless of which PR lands the ECM driver,  the `usb: class: cdc: Add ECM related macros` header additions, and the `usb: host: ch9: Add string descriptor related request helper function` ch9/desc helpers. I plan to pull these in (with credit) and replace the inline string parsing in this PR. That brings the two PRs structurally closer without merging the feature scope.

4. On the overall direction: my preference remains landing a minimal core first (this PR), with #100289 rebasing the additional features on top. This keeps each PR focused and easier to review but I would like @jfischer-no's call on which path is preferred before I proceed.

@jfischer-no could you weigh in on the preferred merge strategy for ECM host support?

subsys/usb/host/class/Kconfig.cdc_ecm_host
```diff
@@ -0,0 +1,25 @@
...
+	depends on NET_L2_ETHERNET
+	help
+	  USB host class driver for CDC Ethernet Control Model (ECM) devices.
+
+if USBH_CDC_ECM
+
+config USBH_CDC_ECM_INSTANCES_COUNT
+	int "Number of CDC ECM class instances"
+	default 1
+	range 1 4
```
tmon-nordic: Where does the upper limit of 4 comes from?

santhosh-c-c: The number 4 is an arbitrary choice not tied to any spec.

tmon-nordic: What about changing it to e.g. `range 1 $(UINT8_MAX)`?

santhosh-c-c: Will change it to `range 1 $(UINT8_MAX)`

subsys/usb/common/ecm.c
josuah, line 1: This is put in common but does not modify the device side class to use it,
is it intended for a future PR?

If merging it in this current form is planned, maybe this can stay in `usbh_cdc_ecm.c` and only move to `common` when/if some merging between device and host-side implementation is done?

subsys/usb/host/class/usbh_cdc_ecm.c
```diff
@@ -0,0 +1,753 @@
...
+	cleanup_xfer(udev, xfer);
+
+	if (priv->state == CDC_ECM_CONFIGURED) {
+		ret = cdc_ecm_start_rx(priv);
+		if (ret) {
+			LOG_ERR("Failed to resubmit bulk IN: %d", ret);
+		}
+	}
+
+	return ret;
```
josuah: Here, the `ret` value is not preserved from the body of the function in case `priv->state == CDC_ECM_CONFIGURED` is true. The body of the function never modifies it from 0.

It might be possible to:

- Turn `ret = 0;` into `ret;`
- `return 0;` at the end of the function

subsys/usb/host/class/usbh_cdc_ecm.c
```diff
@@ -0,0 +1,753 @@
...
+	priv->state = CDC_ECM_CONNECTED;
+
+	ret = cdc_ecm_parse_descriptors(priv, ctrl_iface);
+	if (ret) {
+		LOG_ERR("Failed to parse ECM descriptors");
+		goto err;
+	}
+
+	ret = cdc_ecm_fetch_mac(priv);
+	if (ret) {
```
josuah: Minor style element.

You might be asked during further review to preserve all `if()` expression as boolean. (point 85 [here](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html) FYI)

Typically, `if (ret != 0)`.

subsys/usb/host/class/usbh_cdc_ecm.c
```diff
@@ -0,0 +1,753 @@
...
+	uint8_t bulk_out_ep;
+	uint8_t intr_in_ep;
+	uint8_t imac_str_idx;
+
+	struct k_mutex tx_mutex;
+	struct net_stats_eth stats;
+};
+
+static int cdc_ecm_start_rx(struct usbh_cdc_ecm_data *priv);
+static int cdc_ecm_start_intr(struct usbh_cdc_ecm_data *priv);
```
josuah, line 84: There does not seem to be circular dependencies, do you think it would be possible to reorder the functions so that this is not needed?

subsys/usb/host/class/usbh_cdc_ecm.c
```diff
@@ -0,0 +1,753 @@
...
+	uint8_t ecm_mask = 0;
+	bool in_ecm_function = false;
+
+	desc = (const void *)((const uint8_t *)cfg + cfg->bLength);
+	desc_end = (void *)((const uint8_t *)cfg + cfg->wTotalLength);
+
+	while ((const void *)desc < desc_end && desc != NULL) {
+		switch (desc->bDescriptorType) {
+		case USB_DESC_INTERFACE:
+			if_desc = (const struct usb_if_descriptor *)desc;
```
josuah: Maybe it was meant to have the same exit condition as below in case the next USB function does not have an IAD?

```c
		case USB_DESC_INTERFACE_ASSOC:
			if (in_ecm_function) {
				goto done;
			}
```

josuah (commented):
Thank you for clarifying.

Would you be interested in helping #100289 rebase on top of this? For instance by indicating the parts that would need to be changed?

subsys/usb/common/ecm.c
santhosh-c-c: The shared header was speculative, wrapped the contents directly into `usbh_cdc_ecm.c`

subsys/usb/host/class/usbh_cdc_ecm.c
santhosh-c-c: There's a cycle: `cdc_ecm_start_{rx,intr}` call `submit_xfer` with the `cdc_ecm_{rx,intr}_in_cb` callbacks, and the callbacks resubmit by calling `cdc_ecm_start_{rx,intr}`.

santhosh-c-c:
> Would you be interested in helping #100289 rebase on top of this? For instance by indicating the parts that would need to be changed?

Happy to help @josuah Here is how I see the migration mapping. Splitting into phases by dependency.

**Phase A: Independent PRs (both drivers benefit):**
* **ECM Macros:** Extract the `usb_cdc.h` additions from #100289 and land as a standalone PR.
* **CH9 Helpers:** Extract the string descriptor helpers (`usbh_req_desc_str`, `usbh_desc_str_utf16le_to_ascii`, etc.) from #100289. Once merged, this PR (#99097) will rebase to use them in `cdc_ecm_fetch_mac`, replacing the manual UTF-16LE parsing.

**Phase B: Merge Minimal ECM Core (#99097):**
* Lands with probe, single-pass descriptor parsing, MAC fetch, intr-in carrier tracking, bulk in/out, and initial `SET_ETHERNET_PACKET_FILTER`.

**Phase C: Incremental Feature PRs (ported from #100289):**
To keep reviews manageable, #100289’s driver commit becomes follow-up PRs on top of #99097:
1.  Switch to memslab-pooled `xfer cb_priv` + `queued_xfers` list.
2.  Dynamic packet filter, Multicast filter, Hardware statistics.
3.  `CONNECTION_SPEED_CHANGE` notification + dynamic link speed.
4.  Concurrent TX queue (`sema` + `DATA_TX_CONCURRENT_NUM`).
5.  Disconnect xfer cleanup (drain `queued_xfers`).

**Reconciliation Details:**
* **Descriptor Parser:** #100289 strictly requires `HEADER_FUNC_DESC`. This PR is permissive (handles real-world devices missing the HEADER). I lean toward keeping it permissive, but open to discussion.
* **Iface Bring-up:** #100289 correctly defers `usbh_device_interface_set` until `eth_iface` start (admin-up). I currently do it at probe, but plan to absorb NXP's approach here before the rebase.
* **TX Path:** Addressed in Phase C (4) migrating from my mutex single-xfer to NXP's sema concurrent model for better throughput.

@jfischer-no @MarkWangChinese Let me know your thoughts.

josuah:
> 1. Switch to memslab-pooled xfer cb_priv + queued_xfers list.

I have the impression that this can be avoided altogether.

The thing is: when removing a class from the stack, all associated transfers also need to be removed, and the way to do this is to call dequeue() on each, but there is no way to get the list of all the transfers that are queued to a particular class.

This could be done at the stack level maybe, as then it is a lot less effort for each and every class (implemented with an array for UVC for instance).

https://github.com/zephyrproject-rtos/zephyr/blob/44853cf97498ecbdf941bb3648f04333c54b4bc2/subsys/usb/host/class/usbh_uvc.c#L116

Without this, transfers de-allocation upon class removal is not being handled?

subsys/usb/host/class/usbh_cdc_ecm.c
josuah: I see now, thank you

MarkWangChinese:
> > 1. Switch to memslab-pooled xfer cb_priv + queued_xfers list.
> 
> I have the impression that this can be avoided altogether.
> 
> The thing is: when removing a class from the stack, all associated transfers also need to be removed, and the way to do this is to call dequeue() on each, but there is no way to get the list of all the transfers that are queued to a particular class.
> 
> This could be done at the stack level maybe, as then it is a lot less effort for each and every class (implemented with an array for UVC for instance).
> 
> https://github.com/zephyrproject-rtos/zephyr/blob/44853cf97498ecbdf941bb3648f04333c54b4bc2/subsys/usb/host/class/usbh_uvc.c#L116
> 
> Without this, transfers de-allocation upon class removal is not being handled?

Hi @josuah For disconnecting, the controller will callback all the xfers that are enqueued. For the case of video: stooping the stream https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/host/class/usbh_uvc.c#L3022. I think it is the controller driver interface issue, we needs one controller interface api to release the resource of "dev + endpoint" if it is no longer used, then the related usb bandwidth can be released. If needed, we can talk here https://github.com/zephyrproject-rtos/zephyr/issues/101159. Thanks.

santhosh-c-c:
> Hi @josuah For disconnecting, the controller will callback all the xfers that are enqueued. For the case of video: stooping the stream https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/host/class/usbh_uvc.c#L3022. I think it is the controller driver interface issue, we needs one controller interface api to release the resource of "dev + endpoint" if it is no longer used, then the related usb bandwidth can be released. If needed, we can talk here #101159. Thanks.

@josuah - Based on Mark's explanation, does the host stack actually guarantee that `removed()` runs before the callbacks `(EP_REQUEST)` for cancelled transfers? Since they use different message queues `(usbh_bus_msgq and usbh_msgq)` processed by separate threads, I don't see anything ensuring that specific order.

josuah:
I think this is not guaranteed: there is a thread for endpoint events and other events, they are split in different queues there:

https://github.com/zephyrproject-rtos/zephyr/blob/603fa4818e8faedec88364d2eb3f6c44eb52d763/subsys/usb/host/usbh_core.c#L33-L45

That suggests full concurrency, and events such as "remove this device and all the classes" might need to go through mechanism like "atomically remove all endpoints from all queues and ask drivers they processed/errored-out the ongoing buffer.

santhosh-c-c:
> I think this is not guaranteed: there is a thread for endpoint events and other events, they are split in different queues there:

Thanks for the clarification @josuah

santhosh-c-c:
@jfischer-no @tmon-nordic @josuah
Ping for review, Thanks!

subsys/usb/host/class/usbh_cdc_ecm.c
josuah: I think there is a helper function that combines all three in one which you may be able to use.

josuah, line 450: https://github.com/zephyrproject-rtos/zephyr/pull/102941 seems to introduce string descriptor fetching, maybe this is useful to this PR as well?

We might expect that not only CDC ECM will need string fetching, so introducing it as a stack feature would makes sense?

subsys/usb/host/class/Kconfig.cdc_ecm_host
josuah: In order to configure the number of instances, you might be able to automate it using this snippet (combining it with the `log_config` below)

```
module = USBH_CDC_ECM
module-str = "usbh cdc_ecm"
default-count = 1
source "subsys/logging/Kconfig.template.log_config"
source "subsys/usb/common/Kconfig.template.instances_count"
```

subsys/usb/host/class/usbh_cdc_ecm.c
GeorgeCGV: nit: 
```suggestion
#define ECM_MAC_ADDR_LEN	NET_ETH_ADDR_LEN
```
from already included `zephyr/net/ethernet.h`

GeorgeCGV: Maybe worth making `hex_nibble` return an error (_i.e. `-EINVAL`_) upon an invalid value instead of `0` to protect against malicious devices?

GeorgeCGV: nit:
```suggestion
#define CDC_ECM_ETH_MAX_FRAME_SIZE	(NET_ETH_MTU + sizeof(struct net_eth_hdr))
```
from already included `zephyr/net/ethernet.h`.

Ideally, the host shall rely on `wMaxSegmentSize` (_from a device descriptor_).

santhosh-c-c:
@josuah @GeorgeCGV - Addressed the review comments and rebased on main.

Changes:

- Added a small preparatory commit introducing usbh_xfer_alloc_with_buf() in usbh_device.h, mirroring usbh_xfer_alloc() on top of the existing uhc_xfer_alloc_with_buf(), so the class driver no longer allocates the transfer and buffer separately.
- MAC address retrieval now uses the string descriptor request helpers from #102941 (usbh_req_desc_str() + usbh_desc_str_to_ascii() + hex2bin()). That PR has not merged yet, so it is carried here as a cherry-picked commit, I will drop it once #102941 lands. Depends-on: #102941.
- Instantiation is now Kconfig-count based (CONFIG_USBH_CDC_ECM_INSTANCES_COUNT + LISTIFY) instead of a devicetree compatible.
- Remaining style/naming comments addressed.

Please take another look when you get a chance

subsys/usb/host/class/usbh_cdc_ecm.c
```diff
@@ -0,0 +1,788 @@
...
+static int cdc_ecm_send(const struct device *dev, struct net_pkt *pkt)
+{
+	struct usbh_cdc_ecm_data *priv = dev->data;
+	size_t total_len;
+	int ret;
+
+	total_len = net_pkt_get_len(pkt);
+	if (total_len == 0 || total_len > CDC_ECM_ETH_MAX_FRAME_SIZE) {
+		return -EMSGSIZE;
+	}
```
GeorgeCGV, line 711: Any plans to parse `wMaxSegmentSize` from the descriptor and use it instead of `CDC_ECM_ETH_MAX_FRAME_SIZE`; or better `MIN(CDC_ECM_ETH_MAX_FRAME_SIZE, <wMaxSegmentSize>)`?

santhosh-c-c: ACK, will update this!

GeorgeCGV: Just discovered #100289. Marked it as resolved since I’m not sure which PR the maintainers will pick.

PR #99103, opened 2025-11-07
usb: host: update udev->addr after setting the address

josuah:
Build:
```
git remote add josuah https://github.com/josuah/zephyr/
git fetch josuah
git checkout da1f3ee6f95fbecb502ed823c202d8aca9426d7e

west build -d build_nrf54lm20 -b nrf54lm20dk/nrf54lm20a/cpuapp samples/subsys/usb/shell/ \
-DEXTRA_CONF_FILE=host_prj.conf -DCONFIG_USB_DEVICE_STACK_NEXT=n
```
Before:
```
uart:~$ usbh init
uart:~$ usbh enable
uart:~$ usbh device address 1 2
host: New device address is 2
uart:~$ usbh device list 
1
uart:~$ usbh device descriptor device 1
host: Failed to request device descriptor
uart:~$ usbh device descriptor device 2
host: No USB device with address 2
```
After:
```
uart:~$ usbh init
uart:~$ usbh enable
uart:~$ usbh device address 1 2
host: New device address is 2
uart:~$ usbh device list 
2
uart:~$ usbh device descriptor device 2
bLength			18
bDescriptorType		1
bcdUSB			200
bDeviceClass		0
bDeviceSubClass		0
bDeviceProtocol		0
bMaxPacketSize0		64
idVendor		bda
idProduct		8176
bcdDevice		200
iManufacturer		1
iProduct		2
iSerial			3
bNumConfigurations	1
```

### DIFF
--- a/subsys/usb/host/usbh_device.c
+++ b/subsys/usb/host/usbh_device.c
@@ -447,6 +447,27 @@ error:
 	return err;
 }
 
+int usbh_device_set_address(struct usb_device *const udev, const uint8_t new_addr)
+{
+	int err;
+
+	err = usbh_req_set_address(udev, new_addr);
+	if (err) {
+		LOG_ERR("Failed to set device address to 0x%02x", new_addr);
+		return err;
+	}
+
+	udev->addr = new_addr;
+
+	if (new_addr == 0) {
+		udev->state = USB_STATE_DEFAULT;
+	} else {
+		udev->state = USB_STATE_ADDRESSED;
+	}
+
+	return 0;
+}
+
 int usbh_device_init(struct usb_device *const udev)
 {
 	struct usbh_context *const uhs_ctx = udev->ctx;
@@ -505,16 +526,10 @@ int usbh_device_init(struct usb_device *const udev)
 		goto error;
 	}
 
-	err = usbh_req_set_address(udev, new_addr);
+	err = usbh_device_set_address(udev, new_addr);
 	if (err) {
-		LOG_ERR("Failed to set device address");
-		udev->addr = 0;
-
 		goto error;
 	}
-
-	udev->addr = new_addr;
-	udev->state = USB_STATE_ADDRESSED;
 
 	LOG_INF("New device with address %u state %u", udev->addr, udev->state);
 
--- a/subsys/usb/host/usbh_device.h
+++ b/subsys/usb/host/usbh_device.h
@@ -36,6 +36,9 @@ int usbh_device_interface_set(struct usb_device *const udev,
 			      const uint8_t iface, const uint8_t alt,
 			      const bool dry);
 
+/* Set USB device address */
+int usbh_device_set_address(struct usb_device *const udev, const uint8_t num);
+
 /* Wrappers around to avoid glue UHC calls. */
 static inline struct uhc_transfer *usbh_xfer_alloc(struct usb_device *udev,
 						   const uint8_t ep,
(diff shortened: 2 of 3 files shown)


subsys/usb/host/usbh_ch9.c
```diff
@@ -185,8 +185,16 @@ int usbh_req_set_address(struct usb_device *const udev,
...
 	const uint8_t bRequest = USB_SREQ_SET_ADDRESS;
+	int ret;
+
+	ret = usbh_req_setup(udev, bmRequestType, bRequest, addr, 0, 0, NULL);
+	if (ret != 0) {
+		return ret;
+	}
 
-	return usbh_req_setup(udev, bmRequestType, bRequest, addr, 0, 0, NULL);
+	udev->addr = addr;
```
tmon-nordic: The mismatch is only in https://github.com/zephyrproject-rtos/zephyr/blob/91b1b84eee8159ce9706e8e0a3826cea04a09170/subsys/usb/host/usbh_shell.c#L564-L569

The actual USB host stack updates device address https://github.com/zephyrproject-rtos/zephyr/blob/91b1b84eee8159ce9706e8e0a3826cea04a09170/subsys/usb/host/usbh_device.c#L508-L517

The question is really about what should be updated by `usbh_req_set_address()` and what is left to the caller.

josuah: I cannot see a situation where `usbh_req_set_address()` change the address but `udev` should not be updated...

I am tempted to include it all in `usbh_req_set_address()` which would mirror the behavior of `usbh_device_set_configuration()`:

https://github.com/zephyrproject-rtos/zephyr/blob/3cf7b202780d788976e4bf3e46724e63d23f6bcc/subsys/usb/host/usbh_device.c#L441-L442

Let's try this. Thank you!

josuah: Ah, it is layered, it needs a matching `usbh_device_set_address()` calling `usbh_req_set_address()`.

subsys/usb/host/usbh_device.c
```diff
@@ -447,6 +447,22 @@ int usbh_device_set_configuration(struct usb_device *const udev, const uint8_t n
...
+	int err;
+
+	err = usbh_req_set_address(udev, new_addr);
+	if (err) {
+		LOG_ERR("Failed to set device address to 0x%02x", new_addr);
+		return err;
+	}
+
+	udev->addr = new_addr;
+	udev->state = USB_STATE_ADDRESSED;
```
tmon-nordic: It is `USB_STATE_ADDRESSED` only when `new_addr` is non-zero. If `new_addr` is zero, then it is `USB_STATE_DEFAULT`.

josuah: I applied your recommendation. I have one question if I may ask:

screenshot_2025_1120_0255_00

The diagram from Ch. 9 puts an arrow from "ADDRESSED" to "DEFAULT" via the RESET event only.

Just for confirmation: Should the function return an error in case `set_address(0)` is called?

It makes sense to me as you suggested, however.

tmon-nordic: 9.4.6 Set Address states following
image

so it is somewhat conflicting that response to SetAddress() with a value of 0 is undefined and that if in Address state, when device gets 0 it should go to Default state.

So if anything, it `set_address(0)` would:
  * transition to Default state if device is in Address state,
  * remain in Default state if it is already in Default state,
  * for any other state than Address and Default, return error (without even sending the request to device) because the behavior is not specified.

jfischer-no (approved):

carlescufi (approved):
LGTM

PR #98884, opened 2025-11-05
support file-groups and improve review requests for maintainer file changes

nashif:
- **scripts: get_maintainer: support file groups**
- **scripts: set_assignee.py: Support file groups**
- **MAINTAINERS file: add documentation for file groups**
- **MAINTAINERS file: add description for tests**
- **ci: assigner: merge maintainer check into assigner workflow**
- **scripts: move set_assignee.py into scripts/ci**
- **scripts: set_assignee: request review from maintainers of changed areas**

PR #99104, opened 2025-11-07
Nxp file groups

decsny:
this is an initial pass of using file groups to help consolidate NXP areas and add more specific reviewers and also limiting "total reviewers" added per PR to NXP files if they are not relevant. more file groups will probably be added in the future especially in the drivers area.

- [x] #98884

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -3491,14 +3491,13 @@ NXP Platform Drivers:
     - decsny
     - manuargue
     - dbaluta
-    - Raymond0225
     - Holt-Sun
+    - zejiang0jason
   files-regex:
     - ^drivers/.*nxp.*
     - ^drivers/.*mcux.*
     - drivers\/.*[_\/]+lpc[_\.\d]+.*c
   files:
-    - drivers/*/*imx*
     - drivers/*/*mcux*.c
     - drivers/*/*.mcux
     - drivers/*/*.nxp
@@ -3509,7 +3508,6 @@ NXP Platform Drivers:
     - include/zephyr/dt-bindings/*/*nxp*
     - include/zephyr/dt-bindings/*/*mcux*
     - include/zephyr/dt-bindings/inputmux/
-    - include/zephyr/dt-bindings/rdc/
     - include/zephyr/drivers/*/*nxp*
     - include/zephyr/drivers/*/*nxp*/
     - include/zephyr/drivers/*/*mcux*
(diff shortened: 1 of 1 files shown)


MAINTAINERS.yml
```diff
@@ -3501,33 +3500,23 @@ NXP Platform Drivers:
...
-    - drivers/usb/
   files-regex-exclude:
     - .*s32.*
+  file-groups:
+    - name: NXP USB
+      collaborators:
+        - mmahadevan108
+        - MarkWangChinese
+      files:
+        - drivers/usb/
```
dleach02, line 3528: so does the previous filters above get the initial priority to force it into the NXP drivers area then this sub-filter will select these collaborators?

MAINTAINERS.yml
```diff
@@ -3573,11 +3597,76 @@ NXP Platforms (MCU):
...
+        - EmilioCBen
+      files:
+        - boards/nxp/lpc*/
+        - soc/nxp/lpc/
+        - dts/arm/nxp/nxp_lpc*
+    - name: NXP MCU Xtensa
+      collaborators:
+        - iuliana-prodan
+        - TomasBarakNXP
+        - VitekST
```
dleach02, line 3662: I think Xtensa should stay as a top level similar to the NXP MPU bucket to ensure the proper assignee and collaborators.

MAINTAINERS.yml
```diff
@@ -3640,31 +3715,20 @@ NXP Platforms (S32):
...
+        - boards/nxp/ucans32k1sic/
   labels:
     - "platform: NXP S32"
     - "platform: NXP"
   description: NXP S32 platforms and S32-specific drivers
 
-NXP Platforms (Xtensa):
-  status: maintained
-  maintainers:
-    - dbaluta
```
dleach02, line 3675: I think it is better to have this bucket stay the way it is. This will ensure that dbaluta is the assignee for any xtensa based PR.

decsny: What I have seen happening already with current way is that all the xtensa files overlap completely with other areas (MPU or MCU, usually one at a time), the PRs usually touch non-xtensa files and I have not seen @dbaluta as assignee on Xtensa PRs due to this, the MCU or MPU maintainer gets 100% match but xtensa gets less than 100% match, so @dbaluta not assigned anyways, but maybe he can  correct me.

dleach02:
@decsny please rebase and take this PR out of draft status so we can move forward

MAINTAINERS.yml
decsny: we would have to exclude the xtensa core files from the other areas (MCU and MPU) to get him assigned on those PRs I think

dleach02 (approved):

nashif (approved):

stephanosio (approved):

manuargue (approved):

PR #99108, opened 2025-11-07
Adding NiceNano / ali nRF52840 boards suport

ICantMakeThings:
not sure if this will be pulled as it requires being pulled in 2 other repos.

more specifically these 2 pr's:

https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/857
https://github.com/platformio/platform-nordicnrf52/pull/211

the repo is from [here](https://github.com/ICantMakeThings/Nicenano-NRF52-Supermini-PlatformIO-Support)

### DIFF
--- a/boards/arm/nicenano/Kconfig
+++ b/boards/arm/nicenano/Kconfig
@@ -1,0 +1,10 @@
+# nRF52840 DK NRF52840 board configuration
+
+# Copyright (c) 2016 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+config BOARD_ENABLE_DCDC
+	bool "Enable DCDC mode"
+	select SOC_DCDC_NRF52X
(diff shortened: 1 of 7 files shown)


ICantMakeThings:
oh mad sorry, i think this is zephyr exclusive as i dont recall it being in platformio, should i close the pr?

nordicjm (changes requested):
we do not accept or support boards in hwmv1 format

ICantMakeThings:
now I'm not sure, as promicro_nrf52840 already exists, should I close the PR? or change the defines to support a newer hwm version and this will get merged? no point to update it if it won't even get pulled.

nordicjm:
If the board you are submitting already exists in tree then yes, close PR, if it is a different board or variant then it can be submitted but this PR would need to be updated to be in hwmv2 format

PR #99110, opened 2025-11-07
scripts: west_commands: Add 'new' command

wkhadgar:
Add a 'new' west command as helper for Zephyr projects, drivers and modules.

The project creation is the default state, generating a minimal configuration for a Zephyr app. The module creation consists of a simple template to be completed by the user. Driver creation follows the `example_application` format, producing or appending the created driver skeleton into the `drivers` and `dts/bindings` directories.

This west command aims to ease developers load when starting new projects, or porting existing code into drivers/modules that operate with west and Zephyr.

### DIFF
--- a/scripts/west-commands.yml
+++ b/scripts/west-commands.yml
@@ -99,3 +99,8 @@ west-commands:
       - name: gtags
         class: Gtags
         help: create a GNU global tags file for the current workspace
+  - file: scripts/west_commands/new.py
+    commands:
+      - name: new
+        class: New
+        help: create new zephyr project, module or driver template
--- a/scripts/west_commands/new.py
+++ b/scripts/west_commands/new.py
@@ -1,0 +1,288 @@
+# Copyright (c) 2025 Paulo Santos (pauloxrms@gmail.com).
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""west "new" command"""
+
+import argparse
+import textwrap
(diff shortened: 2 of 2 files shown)


henrikbrixandersen (changes requested):
I think adding something like this requires more discussion. For instance, the code is embedded in a Python script - and it doesn't even follow the established naming of structs and macros. 

To be honest, I don't see much value in generating boilerplate code this simple.

nordicjm (changes requested):
I agree with Brix, and notwithstanding that the filenames and symbol names this seems to use seem completely arbitrary and not matching what is used in-tree

wkhadgar:
> the filenames and symbol names this seems to use seem completely arbitrary and not matching what is used in-tree

I based the naming and structuring on the current `example_application`, though it may not be up to date? 

I would happily change the naming to the best possible! Also, the only out-of-pocket thing I had to make was the `_drv` postfix of the driver implementation, since you only provide the driver class name, and at least one driver implementation should be given to the user to work with.

Other option would be templating not a whole driver class, but only a driver implementation instead, giving to the tool in which class you're planting your implemention... or something like that? 

I guess maybe this could be another option, so the user may choose to implement only a class and then implement a driver for that class with its own name? Looks better to me since it modularizes better... Not sure if you agree @henrikbrixandersen @nordicjm

PR #99117, opened 2025-11-08
samples: net: qos: Fix technical datarace

ClaCodes:
There was technically a datarace on the global 'simulated work time' variable. To avoid that, transport the information along as part of the net-packet.

(Note: it was unlikely an issue, because it was only written, when readers should not have been active, that is why I say `technically`)


samples/net/qos/ethernet/src/main.c
```diff
@@ -123,7 +121,21 @@ static void net_iface_init(struct net_if *iface)
...
 static int net_if_fake_send(const struct device *dev, struct net_pkt *pkt)
 {
-	LOG_INF("sending service %u, pkt %p", net_pkt_ll_proto_type(pkt), pkt);
+	int ok;
+	uint32_t simulated_work_time;
+
+	net_pkt_cursor_init(pkt);
+	net_pkt_set_overwrite(pkt, true);
+
+	/* this sample does not support VLAN */
```
jukkar, line 130: We could enforce this by adding a Kconfig file for the sample and adding suitable depends on `!NET_VLAN`  entry.

PR #99120, opened 2025-11-08
devicetree: Add DT_CHILD_BY_UNIT_ADDR_INT

pdgendt:
Allow fetching child node identifiers by unit address.

It also modifies the `ADC_DT_SPEC_STRUCT` macro to not use `FOREACH` macro expansion, which doesn't play nice with `COND_CODE_X` and other macro loops.

### DIFF
--- a/doc/releases/release-notes-4.4.rst
+++ b/doc/releases/release-notes-4.4.rst
@@ -155,6 +155,11 @@ New Samples
   Same as above, this will also be recomputed at the time of the release.
  Just link the sample, further details go in the sample documentation itself.
 
+DeviceTree
+**********
+
+* :c:macro:`DT_CHILD_BY_UNIT_ADDR_INT`
+* :c:macro:`DT_INST_CHILD_BY_UNIT_ADDR_INT`
 
 Libraries / Subsystems
 **********************
--- a/include/zephyr/devicetree.h
+++ b/include/zephyr/devicetree.h
@@ -436,6 +436,41 @@
 #define DT_CHILD(node_id, child) UTIL_CAT(node_id, DT_S_PREFIX(child))
 
 /**
+ * @brief Get a node identifier for a child node with a matching unit address
+ *
+ * @note Only works for children with unique integer unit addresses.
+ *
+ * Example devicetree fragment:
+ *
+ * @code{.dts}
+ *     / {
+ *             soc-label: soc {
+ *                     serial1: serial@40001000 {
+ *                             status = "okay";
+ *                             current-speed = <115200>;
+ *                             ...
+ *                     };
+ *             };
+ *     };
+ * @endcode
+ *
+ * Example usage with DT_PROP() to get the status of the
+ * `serial@40001000` node:
+ *
+ * @code{.c}
+ *     #define SOC_NODE DT_NODELABEL(soc_label)
+ *     DT_PROP(DT_CHILD_BY_UNIT_ADDR_INT(SOC_NODE, 1073745920), status) // "okay"
+ * @endcode
+ *
+ * @param node_id node identifier
+ * @param addr Integer unit address for the child node.
+ *
+ * @return node identifier for the child node with the specified unit address
+ */
+#define DT_CHILD_BY_UNIT_ADDR_INT(node_id, addr) \
+	DT_CAT3(node_id, _CHILD_UNIT_ADDR_INT_, addr)
+
+/**
  * @brief Get a node identifier for a status `okay` node with a compatible
  *
  * Use this if you want to get an arbitrary enabled node with a given
@@ -4065,6 +4100,21 @@
  */
 #define DT_INST_CHILD(inst, child) \
 	DT_CHILD(DT_DRV_INST(inst), child)
+
+/**
+ * @brief Get a node identifier for a child node with a matching unit address of DT_DRV_INST(inst)
+ *
+ * @note Only works for children with unique integer unit addresses.
+ *
+ * @param inst instance number
+ * @param addr Integer unit address for the child node.
+ *
+ * @return node identifier for the child node with the specified unit address
+ *
+ * @see DT_CHILD_BY_UNIT_ADDR_INT
+ */
+#define DT_INST_CHILD_BY_UNIT_ADDR_INT(inst, addr) \
+	DT_CHILD_BY_UNIT_ADDR_INT(DT_DRV_INST(inst), addr)
 
 /**
  * @brief Get the number of child nodes of a given node
--- a/scripts/dts/gen_defines.py
+++ b/scripts/dts/gen_defines.py
@@ -484,6 +484,19 @@ def write_children(node: edtlib.Node) -> None:
 
     out_dt_define(f"{node.z_path_id}_CHILD_NUM_STATUS_OKAY", ok_nodes_num)
 
+    child_unit_addrs = {}
+    for child in node.children.values():
+        # Provide a way to query child nodes
+        if (addr := child.unit_addr) is not None:
+            child_unit_addrs.setdefault(addr, []).append(child)
+
+    for addr, children in child_unit_addrs.items():
+        if len(children) != 1:
+            # Duplicate unit addresses for different children, skip
+            continue
+
+        out_dt_define(f"{node.z_path_id}_CHILD_UNIT_ADDR_INT_{addr}", f"DT_{children[0].z_path_id}")
+
     out_dt_define(f"{node.z_path_id}_FOREACH_CHILD(fn)",
             " ".join(f"fn(DT_{child.z_path_id})" for child in
                 node.children.values()))
--- a/tests/lib/devicetree/api/src/main.c
+++ b/tests/lib/devicetree/api/src/main.c
@@ -2514,6 +2514,14 @@ ZTEST(devicetree_api, test_children)
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_a), val), 0, "");
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_b), val), 1, "");
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_c), val), 2, "");
+
+	zassert_equal(DT_PROP(DT_CHILD_BY_UNIT_ADDR_INT(DT_NODELABEL(test_children), 10), val), 0);
+	zassert_equal(DT_PROP(DT_CHILD_BY_UNIT_ADDR_INT(DT_NODELABEL(test_children), 11), val), 1);
+	zassert_equal(DT_PROP(DT_CHILD_BY_UNIT_ADDR_INT(DT_NODELABEL(test_children), 12), val), 2);
+
+	zassert_equal(DT_PROP(DT_INST_CHILD_BY_UNIT_ADDR_INT(0, 10), val), 0);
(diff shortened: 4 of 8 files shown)


tests/lib/devicetree/api/src/main.c
```diff
@@ -2514,6 +2514,26 @@ ZTEST(devicetree_api, test_children)
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_a), val), 0, "");
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_b), val), 1, "");
 	zassert_equal(DT_PROP(DT_INST_CHILD(0, child_c), val), 2, "");
+
+	zassert_equal(DT_PROP(DT_CHILD_BY_REG_ADDR(DT_NODELABEL(test_children), 0, 0xa),
```
rruuaanng: It seems that no scene with idx `1` has been added ：）

pdgendt: Added tests for idx `1`.

doc/releases/release-notes-4.4.rst
```diff
@@ -96,6 +96,10 @@ New Samples
   Same as above, this will also be recomputed at the time of the release.
  Just link the sample, further details go in the sample documentation itself.
 
+DeviceTree
+**********
+
+* Added :c:macro:`DT_CHILD_BY_REG_ADDR` and :c:macro:`DT_INST_CHILD_BY_REG_ADDR`.
```
rruuaanng:
```suggestion
:c:macro:`DT_CHILD_BY_REG_ADDR`
:c:macro:`DT_INST_CHILD_BY_REG_ADDR`
```

scripts/dts/gen_defines.py
```diff
@@ -484,6 +484,22 @@ def write_children(node: edtlib.Node) -> None:
...
+                child_reg_addrs.setdefault((idx, reg.addr), []).append(child)
+
+    for (idx, addr), children in child_reg_addrs.items():
+        if len(children) > 1:
+            # Duplicate addresses for different children, skip
+            continue
+
+        # Decimal and hexadecimal variants
+        out_dt_define(f"{node.z_path_id}_CHILD_REG_IDX_{idx}_ADDR_{addr}", f"DT_{children[0].z_path_id}")
+        out_dt_define(f"{node.z_path_id}_CHILD_REG_IDX_{idx}_ADDR_{hex(addr)}", f"DT_{children[0].z_path_id}")
```
rruuaanng:
```suggestion
        if len(children) <= 1:
            # Each child node must have a unique address
            out_dt_define(f"{node.z_path_id}_CHILD_REG_IDX_{idx}_ADDR_{addr}", f"DT_{children[0].z_path_id}")
            out_dt_define(f"{node.z_path_id}_CHILD_REG_IDX_{idx}_ADDR_{hex(addr)}", f"DT_{children[0].z_path_id}")
```
It might be better to avoid control flow statements ;)
(Sorry for making you push so many times.)

rruuaanng (dismissed):

scripts/dts/gen_defines.py
pdgendt: It's a single `for`-loop statement, I prefer the "early return/break/continue" over an additional indentation level. If anything it could be clearer to have `!= 1` instead.

include/zephyr/devicetree.h
```diff
@@ -435,6 +435,41 @@
...
+ * @code{.dts}
+ *     / {
+ *             soc-label: soc {
+ *                     serial1: serial@40001000 {
+ *                             status = "okay";
+ *                             current-speed = <115200>;
+ *                             ...
+ *                     };
+ *             };
+ *     };
```
decsny: in the description you say it's by reg address, but I don't see a reg property here in the example, unless you mean to refer to the *unit address*, in which case the name of the macro is also wrong

decsny (dismissed):

include/zephyr/devicetree.h
```diff
@@ -435,6 +435,41 @@
...
+ * @endcode
+ *
+ *
+ * @param node_id node identifier
+ * @param idx Register address index for the child node.
+ * @param addr Register address for the child node. Can be hexadecimal (prefix with 0x) or decimal
+ *
+ * @return node identifier for the child node with the reg address at a specified index
+ */
+#define DT_CHILD_BY_REG_ADDR(node_id, idx, addr) \
```
decsny: if we are going to do this, then i suggest you make a _BY_IDX version , and make this one do index 0. and also could be useful to add a _BY_NAME version.

henrikbrixandersen: Note that resources can be shared between multiple nodes (`reg` property addresses do not need to be unique across nodes).

include/zephyr/devicetree.h
pdgendt: I've added a proper `_BY_UNIT_ADDR` variant, which is needed to fix the adc dt spec. I could drop the `_BY_REG_ADDR`, but it doesn't harm I guess?

henrikbrixandersen: Not sure it makes sense either given https://github.com/zephyrproject-rtos/zephyr/pull/99120#discussion_r2519478326

pdgendt:
> > Allow fetching child node identifiers by reg address.
> 
> Can you elaborate a bit on the use case, please?

I've added a commit that replaces the `struct adc_dt_spec` instantiation macro, that used to have a special `FOREACH..IS_EQ` check. This check doesn't work for cases where drivers need to instantiate it as part of a `DT_FOREACH_COMPAT` macro (double `FOREACH` macro).

include/zephyr/devicetree.h
pdgendt: Updated the PR

include/zephyr/devicetree.h
```diff
@@ -435,6 +435,41 @@
...
+ * @endcode
+ *
+ * Example usage with DT_PROP() to get the status of the
+ * `serial@40001000` node:
+ *
+ * @code{.c}
+ *     #define SOC_NODE DT_NODELABEL(soc_label)
+ *     DT_PROP(DT_CHILD_BY_UNIT_ADDR(SOC_NODE, 0x40001000), status) // "okay"
+ * @endcode
+ *
```
mathieuchopstm, line 464: I'd suggest adding a `@note` somewhere that the macro doesn't work if multiple children with the same unit address exist. I don't think it's very common, but I do think there are some such situations in tree, and it's better to document it than have people get nasty surprises.

include/zephyr/devicetree.h
```diff
@@ -435,6 +435,41 @@
...
+ *
+ * @param node_id node identifier
+ * @param addr Unit address for the child node. Can be hexadecimal (prefix with 0x) or decimal
```
mathieuchopstm: Possibly, this should indicate the hex must be in lowercase

mathieuchopstm (commented):
LGTM otherwise

include/zephyr/devicetree.h
```diff
@@ -4066,6 +4102,19 @@
 #define DT_INST_CHILD(inst, child) \
 	DT_CHILD(DT_DRV_INST(inst), child)
 
+/**
+ * @brief Get a node identifier for a child node by a unit address of DT_DRV_INST(inst)
+ *
+ * @param inst instance number
+ * @param addr Unit address for the child node. Can be hexadecimal (prefix with 0x) or decimal
```
mathieuchopstm: Either "see `DT_CHILD_BY_UNIT_ADDR` for restrictions" or:

```suggestion
 * @param addr Unit address for the child node. Can be decimal or
 *             hexadecimal (lowercase and prefixed with 0x)
```

mathieuchopstm (dismissed):

pdgendt:
Ping @decsny @mbolivar

rruuaanng (dismissed):

include/zephyr/devicetree.h
decsny: I'm not sure how that's related to what  I said, let me clarify

_BY_IDX would get the reg value of a certain index in the reg property, and _BY_NAME would do the same but instead of using index number it uses a name from reg-names

decsny: Maybe I misunderstood the intention of the PR based on what the code was before, now I see it's changed to completely something else to get the node based on the unit address

pdgendt: I wasn't aware at first the unit address could be different from the first reg address.

As I need to get the child for a unit address, I've updated the PR to reflect that.

mbolivar (changes requested):
I'm not comfortable with this as-is. Although the "usual" situation is that unit addresses are hex numbers, there are some important bus-specific counterexamples, such as for pcie devices.

Upstream linux has over 1000 such cases where there are commas as separators in unit addresses, based on a random commit from Linus's tree on my laptop: https://gist.github.com/mbolivar/1ef0f7a37a1fcb5792e7ba660b5a86a3

This is perfectly legitimate -- see DTSpec v0.4 2.2.1 Node Names for the reference to table 2.1 which allows commas, periods, plus signs, etc in unit addresses, on a bus-specific basis.

Can you think of another way to accomplish the same thing that doesn't take us into deep water, spec-wise? Otherwise, I'm happy to continue the current approach if you're willing to put in the effort to come up with a robust approach that we won't have to keep hacking on over time, and which fully supports the spec's permissions on unit addresses.

pdgendt:
@mbolivar I've dropped the hexadecimal variant to simplify the macro handling, with the limitations stated in the description.

I think this makes it unambiguous how a child node can be constructed from a given parent and integer unit address. I don't think we can cover all spec cases, as the python handling does not support non-integer unit addresses in the current implementation.

pdgendt:
Note the build issues in https://github.com/zephyrproject-rtos/zephyr/pull/100207 which are fixed by this PR.

mathieuchopstm (dismissed):
Seems fair to me, considering that Zephyr is currently unable to cope with non-integer unit addresses and the macros now document that they only work on integer unit addresses; I am leaving final decision to maintainer though.

(The ability to get a child by address is useful in quite a lot of situations, so it'd be nice to (dare I say: *finally*) have a macro that can do it, even if limited in scope)

rruuaanng (dismissed):

scripts/dts/gen_defines.py
```diff
@@ -484,6 +484,19 @@ def write_children(node: edtlib.Node) -> None:
 
     out_dt_define(f"{node.z_path_id}_CHILD_NUM_STATUS_OKAY", ok_nodes_num)
 
+    child_unit_addrs = {}
+    for child in node.children.values():
+        # Provide a way to query child nodes
+        if (addr := child.unit_addr) is not None:
+            child_unit_addrs.setdefault(addr, []).append(child)
```
mbolivar, line 491: Huh -- I didn't know about this dict method. Interesting alternative to collections.defaultdict.

mbolivar (changes requested):
@pdgendt how about renaming this to `DT_CHILD_BY_UNIT_ADDR_INT` to give us some wiggle room in the future?

If we ever extend the library to return the unit address as a string, I would want to be able to implement a zephyr macro DT_CHILD_BY_UNIT_ADDR() that would accept the lowercase-and-underscores version of the unit address instead of this. That feels more general and unlikely to result in collisions in practice as well.

henrikbrixandersen (approved):

mbolivar (approved):

mathieuchopstm (approved):

PR #99123, opened 2025-11-08
usb: device_next: hid: add a suspend callback

fabiobaltieri:
Add an optional suspend callback to let a downstream hid driver be notified of HID suspend and resume events.

### DIFF
--- a/include/zephyr/usb/class/usbd_hid.h
+++ b/include/zephyr/usb/class/usbd_hid.h
@@ -104,6 +104,12 @@ struct hid_device_ops {
 	void (*iface_ready)(const struct device *dev, const bool ready);
 
 	/**
+	 * This callback is called when the HID interface is suspended by the
+	 * host. This callback is optional.
+	 */
+	void (*iface_suspended)(const struct device *dev, const bool suspended);
+
+	/**
 	 * This callback is called for the HID Get Report request to get a
 	 * feature, input, or output report, which is specified by the argument
 	 * type. If there is no report ID in the report descriptor, the id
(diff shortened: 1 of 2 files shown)


tmon-nordic (approved):
I like this approach better than relying on USB notifications in the application, because here the call is synchronous and therefore the state never gets out-of-sync.

include/zephyr/usb/class/usbd_hid.h
```diff
@@ -103,6 +103,12 @@ struct hid_device_ops {
 	 */
 	void (*iface_ready)(const struct device *dev, const bool ready);
 
+	/**
+	 * This callback is called when the HID interface is suspended by the
+	 * host. This callback is optional.
+	 */
+	void (*iface_suspended)(const struct device *dev, const bool suspended);
+
```
jfischer-no, line 111: An interface cannot be suspended, and there is nothing like that in the HID specification. To get notifications when a USB device is suspended, please use USBD_MSG_SUSPEND and USBD_MSG_RESUME. There is no reason to have additional callbacks in each function/class implementation.

jfischer-no (changes requested):

include/zephyr/usb/class/usbd_hid.h
fabiobaltieri: > An interface cannot be suspended, and there is nothing like that in the HID specification. To get notifications when a USB device is suspended, please use USBD_MSG_SUSPEND and USBD_MSG_RESUME. There is no reason to have additional callbacks in each function/class implementation.

Ok, looked more into my application and yeah actually I managed to get it working with those two messages, the culprit may or may not have been a `if (!usbd_can_detect_vbus(usbd_ctx)) {return;}` in my code that was causing all messages to be ignored. :-)

Nevermind then thanks for the pointer, guess then these placeholder callback could go? Anyway closing this down.

PR #99124, opened 2025-11-08
kernel: Add Kconfig option to disable LTO for kernel sources

GTLin08:
Some SoCs require kernel code to be placed in RAM, which makes link-time optimization (LTO) unsuitable for these files.
Disabling LTO allows the affected code to be linked as separate objects and placed in specific memory regions.
Running kernel code from RAM can improve execution performance, especially for timing-critical routines or context switch paths.

Another commit is to enable `KERNEL_NO_LTO `when `SOC_IT8XXX2_KERNEL_IN_RAM `and `LTO `are enabled on the IT8XXX2 chip.


soc/ite/ec/it8xxx2/kernel_no_lto.cmake
```diff
@@ -0,0 +1,38 @@
...
+# Prepare the code snippet to inject
+message(STATUS "[no-LTO] Injecting no-LTO configuration into kernel/CMakeLists.txt...")
+set(INJECT_CODE "${MARKER}
+set_source_files_properties(busy_wait.c device.c kheap.c idle.c mutex.c sem.c work.c
+                            thread.c sched.c timeout.c time.c poll.c events.c banner.c
+                            system_work_q.c
+                            PROPERTIES COMPILE_FLAGS \"-fno-lto -g\")")
+
+# Append the injected code to the end of the file
+file(APPEND ${KERNEL_CMAKE} "${INJECT_CODE}")
```
fabiobaltieri: uh, why is this needed? The whole append/remove thing feels a bit hacky to be honest, hope we can find a different solution, you just need the `-fno-lto -g` in the build command line for the files above right?

GTLin08: The reason is that setting `-fno-lto` elsewhere (e.g. board-level CMake) didn’t work likely because of build order issues. Kernel sources were still built with LTO, so I had to apply it directly here to make sure those files compile without LTO and can reside in RAM.

fabiobaltieri: ok so this is basically touching the source file to somehow convince cmake to build it again with the new flag?

GTLin08: Yes, that's correct, it temporarily modifies the kernel CMake file to ensure those specific sources are rebuilt with `-fno-lto`. After the build configuration is complete, the injected changes are automatically cleaned up to restore the original file state.

fabiobaltieri: Got it. I don't really think it's a good idea to try and modify non generated files from the build script, @tejlmand do you have a better approach to suggest?

Would we not be able to get away with adding a Kconfig flag and the options in the kernel `CMakeLists.txt` file directly? If this SoC has a scenario where having a kernel specific option to selectively not do LTO it may be handy for other devices too.

keith-zephyr: Instead of modifying the kernel's CMake file, what about creating a separate CMake file that defines the no-LTO  and other desired properties?

You could have the kernel CMake file do a conditional include based on a CONFIG option being set.

GTLin08: Thanks @fabiobaltieri and @keith-zephyr for the helpful suggestions.
I’ve updated the implementation accordingly. It now adds a Kconfig option to control the behavior, and the logic is handled inside kernel/CMakeLists.txt without modifying it from external scripts. Additionally, I’ve added a separate `kernel_no_lto.cmake` file to define the no-LTO settings.

kernel/Kconfig
```diff
@@ -1138,6 +1138,27 @@ endif # BOOTARGS
 
 endmenu
 
+config KERNEL_NO_LTO
+	bool "Build selected kernel core files without LTO"
+	help
```
fabiobaltieri, line 1172: guess this should `depends on LTO`?

fabiobaltieri (dismissed):

kernel/Kconfig
GTLin08: Done.

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,44 @@
...
+endif()
+
+# Apply -fno-lto to all C source files, except for initialization
+# and error handling modules (e.g. init.c, errno.c, fatal.c) and
+# CONFIG_KERNEL_LTO_ALLOWLIST
+foreach(src ${KERNEL_SRCS})
+  if(src MATCHES "\\.c$" AND
+     NOT src MATCHES "errno\\.c$" AND
+     NOT src MATCHES "fatal\\.c$" AND
+     NOT src MATCHES "init\\.c$")
```
keith-zephyr: Should these files be added to the LTO_ALLOWLIST as the default settings?

GTLin08: Agreed, I’ve added them to the default LTO_ALLOWLIST. Thanks!

GTLin08:
@fabiobaltieri Please revisit, thanks!

fabiobaltieri (dismissed):
@tejlmand does this look like an acceptable approach?

kernel/Kconfig
```diff
@@ -1138,6 +1138,28 @@ endif # BOOTARGS
...
+	  Running kernel code from RAM can improve execution performance, especially
+	  for timing-critical routines or context switch paths.
+
+if KERNEL_NO_LTO
+config KERNEL_LTO_ALLOWLIST
+	string "List of kernel source files to keep LTO"
+	help
+	  List of kernel source filenames that should retain LTO even when
+	  CONFIG_KERNEL_NO_LTO is enabled.
+	  Example: CONFIG_KERNEL_LTO_ALLOWLIST="init.c errno.c fatal.c"
```
keith-zephyr: Did you want to make these files the default setting?  Otherwise you could expand the help text to suggest this list as a good default as these features don't benefit from the performance gain of running in RAM.

GTLin08: If we set default "init.c errno.c fatal.c" for CONFIG_KERNEL_LTO_ALLOWLIST, users would still need to override the entire string when adding more files. Kconfig does not support appending to the default value, so it cannot be extended incrementally.

keith-zephyr: Acknowledged. As a follow up PR, you could enable CONFIG_KERNEL_NO_LTO on the ITE eval board so that there is a reference implementation for users to copy.

keith-zephyr (dismissed):

kernel/Kconfig
```diff
@@ -1138,6 +1138,28 @@ endif # BOOTARGS
 
 endmenu
 
+config KERNEL_NO_LTO
+	bool "Build selected kernel core files without LTO"
+	depends on LTO
```
JarmouniA, line 1170: This should depend on XIP as well?

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,42 @@
+# SPDX-License-Identifier: Apache-2.0
```
JarmouniA: missing copyright notice

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,42 @@
...
+endif()
+
+# Split allowlist string into list
+if(DEFINED CONFIG_KERNEL_LTO_ALLOWLIST)
+  separate_arguments(LTO_ALLOWLIST NATIVE_COMMAND "${CONFIG_KERNEL_LTO_ALLOWLIST}")
+endif()
+
+# Apply -fno-lto to all C source files, except for some initialization files
+# (e.g. init.c, errno.c, fatal.c) and those that are less critical to be
+# placed in RAM. These files can be excluded from -fno-lto by using
```
JarmouniA:
```suggestion
# executed in RAM. These files can be excluded from -fno-lto by using
```

kernel/Kconfig
JarmouniA: Also, are you using Zephyr Code Relocation to place kernel in RAM?

kernel/kernel_no_lto.cmake
GTLin08: Done.

kernel/kernel_no_lto.cmake
GTLin08: Done.

kernel/Kconfig
GTLin08: > This should depend on XIP as well?

I've added the dependency on XIP as suggested.


> Also, are you using Zephyr Code Relocation to place kernel in RAM?

I'm not using Zephyr's generic code relocation framework. We relocate the kernel-related libraries into RAM using a platform-specific RAM region defined in our SoC linker script, so disabling LTO ensures the kernel objects keep their sections and can be placed correctly in RAM.

kernel/Kconfig
GTLin08: Sure, once this PR is merged, I’ll create a follow-up PR to enable.

fabiobaltieri (dismissed):

keith-zephyr (dismissed):

GTLin08:
@tejlmand, Any feedback would be greatly appreciated.

kernel/Kconfig
```diff
@@ -1138,6 +1138,29 @@ endif # BOOTARGS
...
+	  Some SoCs require kernel code to be placed in RAM, which makes link-time
+	  optimization (LTO) unsuitable for these files (-fno-lto). Disabling LTO
+	  allows the affected code to be linked as separate objects and placed in
+	  specific memory regions.
+
+	  Running kernel code from RAM can improve execution performance, especially
+	  for timing-critical routines or context switch paths.
+
+if KERNEL_NO_LTO
+config KERNEL_LTO_ALLOWLIST
```
nordicjm: this should be done in CMake using functions not Kconfig

GTLin08: Thanks for the suggestion. I’ve removed the `KERNEL_LTO_ALLOWLIST` Kconfig option and switched to using a CMake function as recommended.

GTLin08:
Rebase.

fabiobaltieri (dismissed):

kernel/Kconfig
```diff
@@ -1159,6 +1159,19 @@ endif # BOOTARGS
 
 endmenu
 
+config KERNEL_NO_LTO
+	bool "Build selected kernel core files without LTO"
```
tejlmand: why is this having a prompt ?

This seems to be something the SoC decides and not the user.

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,50 @@
...
+if(TARGET kernel)
+  get_property(KERNEL_SRCS TARGET kernel PROPERTY SOURCES)
+else()
+  message(WARNING "[no-LTO] kernel target not found, skipping")
+  return()
+endif()
+
+# Check if there is a permitted kernel file that can enable LTO
+if(COMMAND kernel_lto_allow_list)
+  kernel_lto_allow_list(ADDITIONAL_ALLOWLIST)
```
tejlmand: couple of issues with this.

1. This is not tested anywhere in the tree.
2. This is vulnerable to CMake configure ordering, how is it ensured that users implement the function `kernel_lto_allow_list` before this CMake code executes ?

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2025 ITE Corporation. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Optional script to disable LTO for kernel files
+
+message(STATUS "[no-LTO] Building kernel files without LTO")
```
tejlmand: please remove.
```suggestion
```

kernel/kernel_no_lto.cmake
```diff
@@ -0,0 +1,50 @@
...
+        break()
+      endif()
+    endforeach()
+
+    if(NOT skip)
+      set_source_files_properties(${src} PROPERTIES COMPILE_FLAGS "-fno-lto -g")
+    endif()
+
+  endif()
+endforeach()
```
tejlmand: No, if we're going to allow SoCs to disable LTO for kernel, then this should be done on a per-library basis, not this granularity.

You're also introducing new hidden (undocumented) settings, like `ADDITIONAL_ALLOWLIST`, which furthermore are process order dependent.

When disabling lto on kernel library level, then this still allows projects to make any additional adjustments they like using `set_source_files_properties()` in their CMakeLists.txt file.

tejlmand (changes requested):
Disabling LTO on kernel sources should be on library basis, not this level of granularity.

kernel/Kconfig
GTLin08: Remove the prompt.

kernel/kernel_no_lto.cmake
GTLin08: Instead of calling `kernel_lto_allow_list()` I switched to a Kconfig solution. This makes the allowlist explicit, avoids ordering issues, and allows each board to define its own list through Kconfig.

kernel/kernel_no_lto.cmake
GTLin08: Done.

kernel/kernel_no_lto.cmake
GTLin08: I tried following the suggested direction of disabling LTO at the library level rather than granularity. However, based on the experiments I ran, LTO still appears to be applied to the kernel even after attempting to disable it.

My test was:  `target_compile_options(kernel PRIVATE -fno-lto)`
However, after compilation, the kernel-related `.c` files still appear to be optimized with LTO.

To investigate further, I checked where -flto is injected in the Zephyr CMake files:
```
zephyr/cmake/compiler/gcc/compiler_flags.cmake:
  set_compiler_property(PROPERTY optimization_lto -flto=auto)
zephyr/cmake/linker/ld/linker_flags.cmake:
  set_property(TARGET linker PROPERTY lto_arguments -flto=auto -fno-ipa-sra -ffunction-sections -fdata-sections)
```
Since `-flto=auto` is added through compiler properties and linker arguments at a higher level, target-level flags do not seem to override these settings in my tests. So far, the only approach that reliably disables LTO for specific kernel files is applying -fno-lto directly via `set_source_files_properties()` on a per–source-file basis.

GTLin08: @tejlmand, Any feedback would be greatly appreciated.

tejlmand: > My test was: `target_compile_options(kernel PRIVATE -fno-lto)` However, after compilation, the kernel-related `.c` files still appear to be optimized with LTO.

yes, because you hit the CMake design principle of an interface lib compiler option to take precedence.

In this case what you can do is to allow a lib to overrule zephyr_interface.
In `<zephyr>/CMakeLists.txt` you can do:
```
    set(genex_tgt_lto "$<TARGET_PROPERTY:LTO>")
    set(genex_lto "$<TARGET_PROPERTY:compiler,optimization_lto>")
    zephyr_compile_options("$<$<OR:$<STREQUAL:${genex_tgt_lto},>,${genex_tgt_lto}>:${genex_lto}>")
```
instead of (line 315 and 318):
```
    zephyr_compile_options($<TARGET_PROPERTY:compiler,optimization_lto>)
```

this will allow targets to disable (or more precisely no apply) lto flags like this:
```
# kernel/CMakeLists.txt
...
if(CONFIG_KERNEL_NO_LTO)
  set_target_properties(kernel PROPERTIES LTO 0)
endif()
```

Doing this also allows you to remove `kernel_no_lto.cmake` file.

Small note regarding the genex in the proposal. The use of `$<STREQUAL:${genex_tgt_lto},>` is to allow a library to not set the LTO property. An unset property defaults to the empty string, so by comparing LTO property against the empty allows us to default unset property to `1`.

GTLin08: > Small note regarding the genex in the proposal. The use of `$<STREQUAL:${genex_tgt_lto},>` is to allow a library to not set the LTO property. An unset property defaults to the empty string, so by comparing LTO property against the empty allows us to default unset property to `1`.

Thanks for the clarification and for walking me through both the generator expression logic and the changes in `<zephyr>/CMakeLists.txt`. The explanation of the design rationale behind the LTO handling was very helpful and made the overall intent much clearer.
 
I’ve followed your suggestion and removed `kernel_no_lto.cmake`. The kernel LTO control is now handled at the library level.

> Doing this also allows you to remove `kernel_no_lto.cmake` file.

The original `kernel_no_lto.cmake` has been renamed to `kernel_allow_lto_files.cmake`, retaining only the allowlist mechanism. The main reason for keeping this is that these files are considered less critical to be executed in RAM, and placing them in the RAM code region provides limited benefit. In addition, some projects are already tight on available RAM, so excluding such files helps reduce RAM pressure.

The `KERNEL_LTO_ALLOWLIST` configuration is kept to provide flexibility, allowing projects to explicitly select which kernel source files should still use LTO when appropriate.

tejlmand: > The KERNEL_LTO_ALLOWLIST configuration is kept to provide flexibility, allowing projects to explicitly select which kernel source files should still use LTO when appropriate.

The problem with that is that the design is broken in different ways.

Directly on the fact that `KERNEL_NO_LTO` is promptless (as it should be), but `KERNEL_LTO_ALLOWLIST` is having a prompt and is intended afaict for the sample to adjust in its `prj.conf`.

Issue with that is that the sample cannot control `KERNEL_NO_LTO` and thus risks making an invalid setting of `KERNEL_LTO_ALLOWLIST` which depends on `KERNEL_NO_LTO`.

Secondly, setting a list of files like this in Kconfig moves control of source code from CMake to Kconfig.
Kconfig should be used for enabling / disabling build related settings, not lists of source files.

Not to mention the risk of list handling (list separator) from Kconfig -> CMake, or passing the list on command line.
There is no mentioning as to whether normal CMake list separator `;` is to be used, or comma, or ....

Therefore, fine-grained control of compile flags for source files should still be in CMake, which a sample can still do using:
```
if(CONFIG_KERNEL_NO_LTO)
  set_source_files_properties(<files> TARGET_DIRECTORY kernel PROPERTIES COMPILE_FLAGS $<TARGET_PROPERTY:compiler,optimization_lto>)
endif()
```

Therefore I still believe the files allow list is broken and should be removed, and instead let the advanced users in need of such fine-grained control resort to plain CMake.

GTLin08: I understand the concerns you mentioned regarding the design, especially the dependency on a promptless option and the mixing of source files control into Kconfig. `KERNEL_LTO_ALLOWLIST` has been removed, and any per-file LTO control is now handled in project level CMake using `set_source_files_properties()`.

Thanks for pointing out these issues and for the guidance on keeping build logic.

JarmouniA: > and any per-file LTO control is now handled in project level CMake using set_source_files_properties().

Can you add a note about this somewhere in the kernel docs?

tejlmand: > > and any per-file LTO control is now handled in project level CMake using set_source_files_properties().
>
> Can you add a note about this somewhere in the kernel docs?

while it's always nice with good documentation, then the possibility of fine-grained control of source file properties using `set_source_files_properties()` is not special to the kernel lib but can basically be used on any source file in Zephyr where a downstream project has special requirements.

It also requires special care by the project not to accidentally apply LTO to certain file, as the general comment for this PR states:
> Some SoCs require kernel code to be placed in RAM, which makes link-time optimization (LTO) unsuitable for these files.

So if adding a note in the docs then care must be taken on how to phrase it.

tejlmand (approved):

fabiobaltieri (approved):

PR #98406, opened 2025-10-28
tests: bluetooth: tester: Fix bis_sink_sync

alexsven:
- Allows the tester to match on either ID or addr
- Previously required that both needed to match, but Auto-PTS is sending the wrong address

PR #99125, opened 2025-11-08
scripts: checkpatch: fix missing blank line detection after declarations

### LINKED ISSUE #98976: checkpatch: Missing line after declaration not enforced if line is removed
### Describe the bug

If a user attempts to add something like 

```c
void foo(void)
{
	int a = 0;
	if (a == 0) {
		return;
	}
}
```

then `checkpatch.pl` would return an error like

```
WARNING : Test Checkpatch warning: 
LINE_SPACING: Missing a blank line after declarations
```

as it should be
```c
void foo(void)
{
	int a = 0;

	if (a == 0) {
		return;
	}
}
```

However if the existing code is 
```c
void foo(void)
{
	int a = 0;

	if (a == 0) {
		return;
	}
}
```

and a commit removes the newline so that it becomes

```c
void foo(void)
{
	int a = 0;
	if (a == 0) {
		return;
	}
}
```

Then the missing newline is not captured by checkpatch.pl, and it passes CI. 

### Regression

- [ ] This is a regression.

### Steps to reproduce

N/A

### Relevant log output

```shell
N/A
```

### Impact

### Environment

8b2c75a2fbe4eef38a99f0a050d9d1d268d1bb69

### Additional Context

It has been seen in a few PRs (e.g. https://github.com/zephyrproject-rtos/zephyr/pull/98406), so it's valid issue, but it's fairly harmless.

vignesh-aerlync:
Fix checkpatch not warning when blank line after declaration
is removed. Updated regexes to also handle context lines,
ensuring consistent detection.
    
Fixes: #98976

### DIFF
--- a/scripts/checkpatch.pl
+++ b/scripts/checkpatch.pl
@@ -1,11 +1,13 @@
 #!/usr/bin/env perl
 # SPDX-License-Identifier: GPL-2.0
+# SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
 #
 # (c) 2001, Dave Jones. (the file handling bit)
 # (c) 2005, Joel Schopp <jschopp@austin.ibm.com> (the ugly bit)
 # (c) 2007,2008, Andy Whitcroft <apw@uk.ibm.com> (new conditions, test suite)
 # (c) 2008-2010 Andy Whitcroft <apw@canonical.com>
 # (c) 2010-2018 Joe Perches <joe@perches.com>
+# (c) 2025 Aerlync Labs Inc
 
 use strict;
 use warnings;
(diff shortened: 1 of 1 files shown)


Thalley (dismissed):
Does not seem to fix the issue

Thalley (dismissed):
The PR seems to fix the issue (after doing proper testing). Not sure whether the changes are valid though, so someone who is more familiar with this should review

vignesh-aerlync:
> @vignesh-aerlync I tried cherry-picking the offending commit from #98406 that caused me to create the reference issue, and then cherry-picked your commit here, but checkpatch.pl still does not fail. How did you test your changes?


Thanks for pointing that out . I tested in https://github.com/zephyrproject-rtos/zephyr/pull/98406  ,my patch missed handling that particular case you mentioned. I’ll update my change accordingly.

scripts/checkpatch.pl
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env perl
-# SPDX-License-Identifier: GPL-2.0
+# SPDX-License-Identifier: Apache-2.0
```
keith-zephyr: You shouldn't drop the original license.

Thalley (dismissed):
I can approve based on that it fixes the issue that I raised. Not sure if the changes in the file are as expected (especially in terms of the license-related changes).

scripts/checkpatch.pl
keith-zephyr: You also cannot add a new license without approval from the original copyright owner(s). 
This file is explicitly called out as being allowed because it is not linked into the firmware.
https://docs.zephyrproject.org/latest/LICENSING.html#zephyr-licensing

It's permitted to extend the copyright, but in general you should never change the license type.

keith-zephyr (changes requested):

keith-zephyr (approved):

Thalley (approved):

PR #99127, opened 2025-11-08
drivers: ethernet: lan9250: add support for a random mac address

ckhardin:
Extend the lan9250 driver to support using a local administered unicast random mac address during init. This follows the device tree settings for zephyr_random_mac_address from other ethernet drivers for the added support.

### DIFF
--- a/drivers/ethernet/eth_lan9250.c
+++ b/drivers/ethernet/eth_lan9250.c
@@ -678,17 +678,49 @@ static int lan9250_init(const struct device *dev)
 		return -EINVAL;
 	}
 
-	if (gpio_pin_configure_dt(&config->interrupt, GPIO_INPUT)) {
+	ret = gpio_pin_configure_dt(&config->interrupt, GPIO_INPUT);
+	if (ret < 0) {
 		LOG_ERR("Unable to configure GPIO pin %u", config->interrupt.pin);
-		return -EINVAL;
+		return ret;
 	}
 
-	gpio_init_callback(&(context->gpio_cb), lan9250_gpio_callback, BIT(config->interrupt.pin));
-	if (gpio_add_callback(config->interrupt.port, &(context->gpio_cb))) {
-		return -EINVAL;
+	gpio_init_callback(&(context->gpio_cb), lan9250_gpio_callback,
+			   BIT(config->interrupt.pin));
+	ret = gpio_add_callback(config->interrupt.port, &context->gpio_cb);
+	if (ret < 0) {
+		LOG_ERR("Unable to add GPIO callback %u", config->interrupt.pin);
+		return ret;
 	}
 
-	gpio_pin_interrupt_configure_dt(&config->interrupt, GPIO_INT_EDGE_TO_ACTIVE);
+	ret = gpio_pin_interrupt_configure_dt(&config->interrupt,
+					      GPIO_INT_EDGE_TO_ACTIVE);
+	if (ret < 0) {
+		LOG_ERR("Unable to enable GPIO INT %u", config->interrupt.pin);
+		return ret;
+	}
+
+	if (config->reset.port != NULL) {
+		if (!gpio_is_ready_dt(&config->reset)) {
+			LOG_ERR("GPIO port %s not ready", config->reset.port->name);
+			return -EINVAL;
+		}
+
+		ret = gpio_pin_configure_dt(&config->reset, GPIO_OUTPUT_INACTIVE);
+		if (ret < 0) {
+			LOG_ERR("Unable to configure GPIO pin %u", config->reset.pin);
+			return ret;
+		}
+
+		/* See Section 19.6.3 from the LAN9250 Data Sheet
+		 *
+		 * trstia is 200 microseconds min (use 250 us)
+		 * tcfg is 15 milliseconds min (use 20 ms for after reset)
+		 */
+		gpio_pin_set_dt(&config->reset, 1);
+		k_usleep(250);
+		gpio_pin_set_dt(&config->reset, 0);
+		k_msleep(20);
+	}
 
 	/* Reset and wait for ready on the LAN9250 SPI device */
 	ret = lan9250_sw_reset(dev);
@@ -720,6 +752,7 @@ static int lan9250_init(const struct device *dev)
 	static const struct lan9250_config lan9250_##inst##_config = {                             \
 		.spi = SPI_DT_SPEC_INST_GET(inst, SPI_WORD_SET(8)),                                \
 		.interrupt = GPIO_DT_SPEC_INST_GET(inst, int_gpios),                               \
+		.reset = GPIO_DT_SPEC_INST_GET_OR(inst, reset_gpios, {0}),                         \
 		.timeout = CONFIG_ETH_LAN9250_BUF_ALLOC_TIMEOUT,                                   \
 		.mac_cfg = NET_ETH_MAC_DT_INST_CONFIG_INIT(inst),                                  \
 	};                                                                                         \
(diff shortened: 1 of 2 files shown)


drivers/ethernet/eth_lan9250.c
```diff
@@ -690,13 +690,34 @@ static int lan9250_init(const struct device *dev)
...
 	ret = lan9250_sw_reset(dev);
 	if (ret < 0) {
 		LOG_ERR("Reset failed");
 		return ret;
 	}
 	lan9250_configure(dev);
+
+	/* use a random MAC address if requested in the device tree */
+	if (config->random_mac) {
+		sys_rand_get(context->mac_address, sizeof(context->mac_address));
```
dsseng: Please fix a missing include as outlined by the CI log

drivers/ethernet/eth_lan9250.c
```diff
@@ -690,13 +690,34 @@ static int lan9250_init(const struct device *dev)
...
+	if (config->reset.port) {
+		if (!gpio_is_ready_dt(&config->reset)) {
+			LOG_ERR("GPIO port %s not ready", config->reset.port->name);
+			return -EINVAL;
+		}
+		if (gpio_pin_configure_dt(&config->reset, GPIO_OUTPUT)) {
+			LOG_ERR("Unable to configure GPIO pin %u", config->reset.pin);
+			return -EINVAL;
+		}
+		gpio_pin_set_dt(&config->reset, 0);
```
dsseng, line 721: Should the driver first assert the reset and then release it to make sure the controller is reset?

dsseng (commented):
Thank you! Please find my suggestions below

drivers/ethernet/eth_lan9250.c
ckhardin: problem of porting commits from internal branches to external ... added the random.h

drivers/ethernet/eth_lan9250.c
ckhardin: assumption in these design is that a pull-up is on the line and the init just de-asserts reset - check the other drivers with a reset-gpio because the pin-mux will be an input and so the line is just pulled up externally and then driven low by the driver.

this matches the patten for other device trees for the gpio outside of chips that have to actually reset via a gpio and not thru a register set.

dsseng:
> interesting - not sure how the msg wait all stack issues relate to this change - so, need to see if that failure is upstream or in this PR alone?

Looks like a flaky test, unlikely to be caused by this PR. Someone with the CI access rights will restart the failed twister run and let you know if it is somehow relevant

drivers/ethernet/eth_lan9250.c
maass-hamburg: as this is a spi ethernet mac, that could be connected to any soc and any gpio, we shouldn't assume behavior, instead it should be done explicit. Also use the timing values from the datasheet, that the pin is asserted long enough.

maass-hamburg (changes requested):

drivers/ethernet/eth_lan9250.c
```diff
@@ -690,13 +691,42 @@ static int lan9250_init(const struct device *dev)
 
 	gpio_pin_interrupt_configure_dt(&config->interrupt, GPIO_INT_EDGE_TO_ACTIVE);
 
+	if (config->reset.port) {
```
maass-hamburg:
```suggestion
	if (config->reset.port != NULL) {
```
the condition has to be a boolean. See coding guidelines
rule 85
```
The controlling expression of an if statement and the controlling expression of an iteration-statement shall have essentially Boolean type
```

maass-hamburg (changes requested):

drivers/ethernet/eth_lan9250.c
ckhardin: Not all the ethernet drivers are updated to the coding rules - this is a cut and paste from eth_w5500.c, so updated it for this patch.

ckhardin:
@maass-hamburg and @dsseng - is there anything else needed? the changes have been made for the style and the reset gpio code was added to this as well.

maass-hamburg:
Please do it like in https://github.com/zephyrproject-rtos/zephyr/pull/100620

drivers/ethernet/eth_lan9250.c
```diff
@@ -697,6 +729,8 @@ static int lan9250_init(const struct device *dev)
 		return ret;
 	}
 	lan9250_configure(dev);
+
+	(void)net_eth_mac_load(&config->mac_cfg, context->mac_address);
```
dsseng, line 733: should this maybe retrieve and handle the return code? here we likely want to abort the initialization sequence if we cannot read the MAC (e.g. missing nvmem cell or a similar issue)

dsseng (commented):
Thanks, looks good

drivers/ethernet/eth_lan9250.c
maass-hamburg: As the user can still set the mac later via the api, so it is recoverable.

dsseng: Okay, may be safe to ignore then. Or maybe initialize to a random address if failed to load the permanent one?

drivers/ethernet/eth_lan9250.c
```diff
@@ -678,17 +678,49 @@ static int lan9250_init(const struct device *dev)
...
+			LOG_ERR("Unable to configure GPIO pin %u", config->reset.pin);
+			return ret;
+		}
+
+		/* See Section 19.6.3 from the LAN9250 Data Sheet
+		 *
+		 * trstia is 200 microseconds min (use 250 us)
+		 * tcfg is 15 milliseconds min (use 20 ms for after reset)
+		 */
+		gpio_pin_set_dt(&config->reset, 1);
```
pdgendt: Nit: you can combine this in a single call

```c
ret = gpio_pin_configure_dt(&config->reset, GPIO_OUTPUT_ACTIVE);
```

pdgendt (dismissed):

drivers/ethernet/eth_lan9250.c
maass-hamburg: At the other drivers we ignore it too for now, so it should be fine

ckhardin: if the others change - then this would follow them, don't think this driver should break ground since it is not the most actively developed.

drivers/ethernet/eth_lan9250.c
ckhardin: following the code in the other drivers - for consistency - they configured inactive and then explicitly run the reset.

From eth_adin2111.c
```
                ret = gpio_pin_configure_dt(&cfg->reset, GPIO_OUTPUT_INACTIVE);
                if (ret < 0) {
                        LOG_ERR("Failed to configure reset GPIO, %d", ret);
                        return ret;
                }

                /* perform hard reset */
                /* assert pin low for 16 µs (10 µs min) */
                gpio_pin_set_dt(&cfg->reset, 1);
                k_busy_wait(16U);
                /* deassert and wait for 90 ms (max) for clocks stabilisation */
                gpio_pin_set_dt(&cfg->reset, 0);
                k_msleep(ADIN2111_HW_BOOT_DELAY_MS);
```

ckhardin:
@maass-hamburg if there are any other changes needed please comment so this can get in, trying to get our fork synced so I don't have to keep patching between them.

maass-hamburg (approved):

pdgendt (approved):
You can omit the initialization in the runtime for
```c
.mac_address = DT_INST_PROP_OR(inst, local_mac_address, {0}), \
```

PR #99132, opened 2025-11-09
[RFC] Support for libsrtp

seyednasermoravej:
This contribution is to add support for SRTP protocol. The original library named libSRTP is developed by Cisco [here](https://github.com/cisco/libsrtp).
Dependencies for this module are:
[libSRTP](https://github.com/cisco/libsrtp)
[mbedTLS](https://github.com/Mbed-TLS/mbedtls)
The mbedTLS integration is made by linking its library to the libSRTP. The libSRTP can use different encryption libraries including openSSL, NSS, mbedTLS, etc. As mbedTLS is available for Zephyr RTOS, I used this library.
The original libSRTP needed just a little modification to be used in Zephyr RTOS. Just adding a file [`zephyr/module.yml` with content `cmake-ext: True` and `Kconfig-ext: True`](https://github.com/seyednasermoravej/libsrtp/blob/add-support-for-Zephyr-RTOS/zephyr/module.yml) inside it. I pushed the modified repository into my GitHub account.
To test the result, for simplicity, I modified [rtpw](https://github.com/cisco/libsrtp/blob/main/test/rtpw.c) example. To test the sample, I used a PC as a sender/receiver and a Zephyr application as a receiver/sender. The original rtpw sample provided by Cisco only sends/receives text instead of Audio/Image. So, I did the same.
How can I integrate it in Zephyr RTOS as an internal or external module?


seyednasermoravej:
> as it uses mbedtls, please make sure that is uses the new psa crypto api, as zephyr is soon switching to https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-4.0.0 where support for the legacy mbedtls api was dropped.

Thank you. I'll check that.

josuah:
Following the [guidelines](https://docs.zephyrproject.org/latest/contribute/external.html#submission-and-review-process), here is the module submission issue with all the needed info:

- https://github.com/zephyrproject-rtos/zephyr/issues/99437

samples/modules/libsrtp/rtpw/Kconfig
```diff
@@ -0,0 +1,44 @@
...
+	  Build the application as an RTP receiver.
+
+endchoice
+
+config NET_CONFIG_PEER_IPV4_ADDR
+	string
+	default "10.42.0.1" if NET_SAMPLE_SENDER
+	default "0.0.0.0" if NET_SAMPLE_RECEIVER
+	help
+	  Peer IPv4 address used by the RTPW example.
```
jukkar: This Kconfig symbol is already defined in net config library so you should not add it here.
You could create overlay files for this sample for both send and receiver roles if you want to allow user to select the value for the IP address.

samples/modules/libsrtp/rtpw/prj.conf
```diff
@@ -0,0 +1,52 @@
+#network
+CONFIG_NETWORKING=y
+CONFIG_NET_IPV4=y
+CONFIG_NET_IPV6=n
+CONFIG_NET_DHCPV4=n
+CONFIG_NET_ARP=y
+CONFIG_NET_CONFIG_SETTINGS=y
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="10.42.0.2"
```
jukkar: Please use the IP address range we have for samples use so "192.0.2.1" here. The user can always override this when compiling the sample.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,169 @@
...
+	sock = socket(PF_INET, SOCK_DGRAM, IPPROTO_UDP);
+	if (sock < 0) {
+		LOG_ERR("couldn't open socket.");
+		return 0;
+	}
+
+	memset(&name, 0, sizeof(struct sockaddr_in));
+	name.sin_addr = rcvr_addr;
+	name.sin_family = PF_INET;
+	name.sin_port = htons(CONFIG_NET_SAMPLE_SRTP_SERVER_PORT);
```
jukkar: Would you be able to add IPv6 support here too? We have been trying to support also IPv6 in net samples if possible, and it is a very simple thing to add as the rtp library seems to be only interested in the socket value.

seyednasermoravej: Unfortunately, libSRTP doesn't support IPv6.

samples/modules/libsrtp/rtpw/prj.conf
```diff
@@ -0,0 +1,54 @@
...
+CONFIG_NET_MGMT_EVENT=y
+CONFIG_SLIP_STATISTICS=n
+CONFIG_NET_SOCKETS=y
+
+CONFIG_ENTROPY_GENERATOR=y
+CONFIG_TEST_RANDOM_GENERATOR=y
+CONFIG_NET_LOG=y
+CONFIG_NET_SHELL=y
+CONFIG_POSIX_API=y
+CONFIG_POSIX_NETWORKING=y
```
josuah: As you pointed out, from the issue linked above, you would need to replace this by [`CONFIG_POSIX_AEP_CHOICE_PSE53]`](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_POSIX_AEP_CHOICE_PSE53) which should be enough to select the other things like `POSIX_NETWORKING` and `POSIX_DEVICE_IO`.

modules/libsrtp/Kconfig
```diff
@@ -0,0 +1,18 @@
...
+	select MBEDTLS_CIPHER_AES_ENABLED
+	select MBEDTLS_CIPHER_CAMELLIA_ENABLED
+	select MBEDTLS_CIPHER_DES_ENABLED
+	select MBEDTLS_CIPHER_CHACHA20_ENABLED
+	select MBEDTLS_CIPHER_CCM_ENABLED
+	select MBEDTLS_CIPHER_GCM_ENABLED
+	select MBEDTLS_CIPHER_MODE_XTS_ENABLED
+	select MBEDTLS_CIPHER_MODE_CBC_ENABLED
+	select MBEDTLS_CIPHER_MODE_CTR_ENABLED
+
```
josuah, line 17: If I understand the link above correctly this mean you need to add

```suggestion
	depends on POSIX_AEP_REALTIME_DEDICATED
```

[`CONFIG_POSIX_AEP_REALTIME_DEDICATED`](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_POSIX_AEP_REALTIME_DEDICATED)

josuah (commented):
@seyednasermoravej
Like you pointed out in the chat, there seems to be some change coming WRT POSIX-compatibility:

- https://github.com/zephyrproject-rtos/zephyr/pull/97855

Good catch!

samples/modules/libsrtp/rtpw/prj.conf
seyednasermoravej: According to [POSIX Option and Option Group Details](https://docs.zephyrproject.org/latest/services/portability/posix/aep/index.html) you're right. As `CONFIG_POSIX_AEP_CHOICE_PSE53` selects [`CONFIG_POSIX_DEVICE_IO`](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_POSIX_DEVICE_IO) and [CONFIG_POSIX_AEP_REALTIME_DEDICATED](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_POSIX_AEP_REALTIME_DEDICATED). Also the [CONFIG_POSIX_NETWORKING](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_POSIX_NETWORKING) is selected by `CONFIG_POSIX_AEP_REALTIME_DEDICATED` itself. But by just enabling `CONFIG_POSIX_AEP_CHOICE_PSE53=y` I faced this warning.
`warning: implicit declaration of function 'close'; did you mean 'fclose'? [-Wimplicit-function-declaration]`
 ` 160 |         ret = close(sock);`
  `    |               ^~~~~`
  `    |               fclose`

maass-hamburg (changes requested):
You have to port libsrtp to use the PSA api, as stated in https://github.com/zephyrproject-rtos/zephyr/issues/97660 The mbedTLS legacy api is already removed in the newest Version of mbedTLS and zephyr is going to switch soon. Adding a library that only supports the old api is going to make the switch more difficult.

samples/modules/libsrtp/rtpw/src/main.c
rlubos: Hmm, but is it really a limitation of `srtp` library or just that `rtp.c` module? `srtp` API itself doesn't seem to bother with any data transmission at all from what I see. And that `rtp.c` module indeed uses hardcoded IPv4 address structs:
https://github.com/seyednasermoravej/libsrtp/blob/163cb8ada1f0e7f0e3184f615d082f2ab40e6784/test/rtp.h#L89
However, it's quite surprising they put it in `test\`, makes me wonder if it was even intended to use externally... Anyway, it just seems to be a simple wrapper over a socket, so maybe we could have something like this implemented on the Zephyr side.

josuah:
Perhaps the correct model is to submit it as external first, and then in the future reconsider depending on how things evolve.

@seyednasermoravej do you think this would work well this way around?

samples/modules/libsrtp/rtpw/src/main.c
seyednasermoravej: Your right, IPv4 is hardcoded in the following files: `rtp.c`, `rtp_decoder.c`, `rtp.h`, and `rtpw.c` files. When I wanted to port libSRTP I had two options. To copy `rtpw.c` related example files into Zephyr folders (in `zephyrproject/zephyr/modules/libsrtp` or `zephyrproject/samples/modules/libsrtp/`) or use existing files from the `test/` folder. I decide the second option to make this PR as small as possible. When @jukkar asked me for IPv6, I modified everything to support for IPv6 but I faced the compile error. So, I conclude it's libsrtp limitation, I forgot maybe the cause of this limitation is files in the `test/` folder not the libSRTP itself. Now as we know the root of limitation, I can modify the related files to support for IPv6.

seyednasermoravej:
> Perhaps the correct model is to submit it as external first, and then in the future reconsider depending on how things evolve.
> 
> @seyednasermoravej do you think this would work well this way around?

Yes, it's a good idea.

seyednasermoravej:
Thanks to everyone who has been involved in this discussion so far — I really appreciate all your feedback. I’m trying to make sure I fully understand the best direction going forward, so I’d like to summarize the current situation and ask a few clarifying questions.

We seem to have two possible approaches:

- Implement SRTP natively in Zephyr, as suggested by @rlubos, or

- Port libSRTP as an external module, as suggested by @josuah and @nashif.

Whether SRTP ends up internal or external, there are a few open issues:

1. **IPv6 support**, requested by @jukkar and @rlubos.
This would require Zephyr to have its own rtp.h and rtp.c (unless I submit a PR to libSRTP upstream and they accept modifications for IPv6).

2. **Migration to Mbed TLS 4.0 / TF-PSA Crypto 1.0**, raised by @maass-hamburg.
This affects several files (`aes_gcm.h`, `aes_gcm_mbedtls.c`, `aes_icm_ext.h`, `aes_icm_mbedtls.c`, and `hmac_mbedtls.c`), and I’m currently working on these changes.

Given these limitations and our previous discussion, I would appreciate guidance on the following points:

1. Should I focus on a native SRTP implementation for Zephyr, or continue porting the existing libSRTP code? (@rlubos)

2. If the libSRTP port is the right direction, was @josuah’s comment sufficient to keep it as an internal module, or should I adjust the code to use it as an external module instead?

3. Is IPv6 support required for this PR, or can it be added later after submitting an upstream PR to libSRTP? (@jukkar @rlubos)

4. Is migration to Mbed TLS 4.0 / TF-PSA Crypto 1.0 mandatory for this PR, or should it wait until Zephyr itself moves to these versions? (@maass-hamburg)

Thank you all again for the support and guidance.

jukkar:
> * Is IPv6 support required for this PR, or can it be added later after submitting an upstream PR to libSRTP? (@jukkar @rlubos)

IPv6 is not mandatory at first stage but highly recommended in the future, so a PR for this can be submitted later.

carlescufi:
Architecture WG meeting:

- This is a protocol for video streaming, very popular on the internet
- @seyednasermoravej explains that we require the actual implementation file as well as headers and new crypto
- @josuah is following this effort and would like to see this in Zephyr
- @nashif questions whether this actually belongs in Zephyr if it's only a sample that is going to use it, and not a dependency to subsystem-level code

josuah:
This shew-up quickly in the TSC meeting, here is what I reported hoping this reflects the current situation well:

- "how" to integrate it still being worked on, implementation details
- This can then be moved from **In Progress** to **Todo** again, decision will be easier then

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,171 @@
...
+		k_msleep(500);
+	}
+
+	rtp_sender_deinit_srtp(snd);
+	rtp_sender_dealloc(snd);
+
+#elif defined(CONFIG_NET_SAMPLE_RECEIVER)
+	rtp_receiver_t rcvr;
+
+	if (bind(sock, (struct sockaddr *)&name, sizeof(name)) < 0) {
```
josuah: Just noticing this: it might be possible to use the usual Zephyr error handling here:

```
ret = bind();
if (ret < 0) {
    ...
}
```

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,171 @@
...
+		close(sock);
```
josuah: We can go with `close()` -> `zsock_close()` based on prior feedback.

samples/modules/libsrtp/rtpw/src/main.c
josuah: We can go with `bind()` -> `zsock_bind()` based on prior feedback.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,171 @@
...
+	if (status) {
+		LOG_ERR("srtp initialization failed with error code %d.", status);
+		return 0;
+	}
+
+	/* set address */
+	LOG_INF("peer IPv4 address: %s.", CONFIG_NET_CONFIG_PEER_IPV4_ADDR);
+	LOG_INF("my IPv4 address: %s.", CONFIG_NET_CONFIG_MY_IPV4_ADDR);
+
+	if (inet_pton(AF_INET, CONFIG_NET_CONFIG_PEER_IPV4_ADDR, &rcvr_addr) != 1) {
```
josuah: It might be possible to convert to Zephyr-style error handling, as well as use the `zsock_` version

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,171 @@
...
+		LOG_ERR("Invalid IPv4 address: %s.", CONFIG_NET_CONFIG_PEER_IPV4_ADDR);
+		return 0;
+	}
+
+	/* open socket */
+	sock = socket(PF_INET, SOCK_DGRAM, IPPROTO_UDP);
```
josuah: It might be possible to use the `zsock_` version

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+    srtp_ctx_t *srtp_ctx;
+    struct sockaddr_in addr; /* reciever's address */
+
+#ifdef CONFIG_NET_SAMPLE_SENDER
+	uint16_t seq_host;
+    uint32_t ts_host;
+#endif
+} rtp_ctx_t;
+
+srtp_err_status_t rtp_init(rtp_ctx_t *rtp,
```
josuah: Here, the type `rtp_ctx_t *rtp` can be replaced by `struct rtp_msg`:

```suggestion
srtp_err_status_t rtp_init(struct rtp_msg *message,
```

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+    rtp_msg_t message;
+    struct k_mutex lock;
+    srtp_ctx_t *srtp_ctx;
+    struct sockaddr_in addr; /* reciever's address */
+
+#ifdef CONFIG_NET_SAMPLE_SENDER
+	uint16_t seq_host;
+    uint32_t ts_host;
+#endif
+} rtp_ctx_t;
```
josuah: It is possible to remove this type altogether, as it is used a single time and only in `main()`, the variables can be used directly instead of stored in this struct.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+{
+    k_mutex_lock(&sender->lock, K_FOREVER);
+
+    sender->seq_host++;
+    sender->ts_host++;
+
+    sender->message.header.seq = sys_cpu_to_be16(sender->seq_host);
+    sender->message.header.ts  = sys_cpu_to_be32(sender->ts_host);
+
+    k_mutex_unlock(&sender->lock);
```
josuah: No need for locking once the content is put directly in `main()`

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+ssize_t rtp_sendto(rtp_ctx_t *sender, const void *msg, size_t len);
+static inline void rtp_update_header(rtp_ctx_t *sender)
+{
+    k_mutex_lock(&sender->lock, K_FOREVER);
+
+    sender->seq_host++;
+    sender->ts_host++;
+
+    sender->message.header.seq = sys_cpu_to_be16(sender->seq_host);
+    sender->message.header.ts  = sys_cpu_to_be32(sender->ts_host);
```
josuah: This content can be put directly in `rtp_sendto()`, and then further moved to `main()`.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+    }
+    octets_sent =
+        zsock_sendto(sender->socket_fd, (void *)&sender->message, pkt_len, 0,
+               (struct sockaddr *)&sender->addr, sizeof(struct sockaddr_in));
+
+    if (octets_sent != pkt_len) {
+        LOG_ERR("error: couldn't send message %s", (char *)msg);
+    }
+
+    return octets_sent;
```
josuah: All of this content can be moved to `main()`.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+    rtp->message.header.p = 0;
+    rtp->message.header.x = 0;
+    rtp->message.header.cc = 0;
+
+    /* set other stuff */
+    rtp->socket_fd = sock_fd;
+    rtp->addr = addr;
+
+    return srtp_err_status_ok;
+}
```
josuah: All of the content of this function can be moved to `main()` directly.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+    if (stat) {
+        LOG_ERR("error: srtp unprotection failed with code %d%s\n",
+                stat,
+                stat == srtp_err_status_replay_fail ? " (replay check failed)"
+                : stat == srtp_err_status_auth_fail ? " (auth check failed)"
+                                                    : "");
+        return -1;
+    }
+    strncpy(msg, receiver->message.body, octets_recvd);
+
```
josuah: All of this content can be moved directly to `main()`.

samples/modules/libsrtp/rtpw/src/main.c
josuah: Now that `rtp.c` is merged into `main.c`, it might be possible to support IPv6 in this sample.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,309 @@
...
+	/* check that hex string is the right length */
+	if (key_len != policy.rtp.cipher_key_len) {
+		LOG_ERR("wrong number of digits in key (should be %d digits, found %d).",
+			2 * policy.rtp.cipher_key_len, 2 * key_len);
+		return 0;
+	}
+
+	/* initialize sender's rtp and srtp contexts */
+	rtp_ctx_t rtp = { 0 };
```
josuah: As seen above, this can all be turned into individual variables, as the struct is only used once here.

josuah (commented):
Here the code was just copy-pasted from libSRTP into this sample,
what can be done now is to further remove everything not essential for it to work:

- Remove all extra struct definition except `struct rtp_msg`
- Convert `rtp_msg_t` to `struct sample_rtp_msg` (without `typedef`).
- Remove all functions and call everything from `main()` directly
- Remove all variables not needed, for instance `socket_fd` in `rtp_ctx_t` is already there as `sock_fd` in `main()`.

This should reduce the sample quite much.

Then, once done and working again, it becomes possible to make CI pass and adapt everything to Zephyr coding style, and this should be ready.

This would be a very crude, very manual way to fill the content of the RTP data, but automating this is the duty of https://github.com/zephyrproject-rtos/zephyr/pull/104603 which will complement this nicely.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,287 @@
...
+	bind_addr.sin6_port = sys_cpu_to_be16((CONFIG_NET_SAMPLE_SRTP_SERVER_PORT));
+	bind_addr.sin6_addr = in6addr_any;
+	ret = zsock_bind(sock_fd, (struct sockaddr *)&bind_addr, sizeof(bind_addr));
+#else
+	struct sockaddr_in bind_addr = {0};
+
+	bind_addr.sin_family = AF_INET;
+	bind_addr.sin_port = sys_cpu_to_be16((CONFIG_NET_SAMPLE_SRTP_SERVER_PORT));
+	bind_addr.sin_addr.s_addr = INADDR_ANY;
+	ret = zsock_bind(sock_fd, (struct sockaddr *)&bind_addr, sizeof(bind_addr));
```
josuah, line 260: This line is the same, could be put below the `#ifdef` eventually:

```c
	ret = zsock_bind(sock_fd, (struct sockaddr *)&bind_addr, sizeof(bind_addr));
```

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,287 @@
...
+		size_t octets_recvd;
+		srtp_err_status_t stat;
+
+		octets_recvd = zsock_recvfrom(sock_fd, (void *)&message, word_len, 0,
+					      (struct sockaddr *)NULL, 0);
+
+		if (octets_recvd < 0) {
+			word_len = 0;
+			return -EMSGSIZE;
+		}
```
josuah: This is a POSIX function, so you might be able to do `return -errno;`

I think "octets" is only used in libsrtp? (British English despite Cisco standing for San-FranCISCO), "bytes" sounds more familiar to Zephyr.

modules/libsrtp/Kconfig
josuah: Is this still relevant?

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,287 @@
...
+	message.header.version = 2;
+	message.header.p = 0;
+	message.header.x = 0;
+	message.header.cc = 0;
+
+	status = srtp_create(&srtp_ctx, &policy);
+	if (status) {
+		LOG_ERR("srtp_create() failed with code %d.", status);
+		return -EINVAL;
+	}
```
josuah: Now that we have everything in `main()` and reduced to minimum, it becomes easier to split the code again.

We have:

```c
main() {
    srtp_create(&ctx);

#ifdef CONFIG_NET_SAMPLE_SENDER
    /* sender code */
#elif defined(CONFIG_NET_SAMPLE_RECEIVER)
    /* receiver code */
#endif

    srtp_dealloc(ctx);
}
```

How about this:

```c
__maybe_unused app_sender(*ctx) {
    /* sender code */
}

__maybe_unused app_receiver(*ctx) {
    /* sender code */
}

main() {
    srtp_create(&ctx);

#ifdef CONFIG_NET_SAMPLE_SENDER
    app_sender(&ctx);
#elif defined(CONFIG_NET_SAMPLE_RECEIVER)
    app_receiver(&ctx);
#endif

    srtp_dealloc(ctx);
}
```

Currently, the code calls `return -ERROR;`, which skips the `srtp_dealloc();` upon error.

By having separate functions for `app_sender()` and `app_receiver()`, this means you can keep all the current code with `return`, and handle the error in `main()` which will not `return ret;` but instead log the error and continue with the de-initialization.

This way this demonstrates correct error handling for users who copy-paste this sample.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,287 @@
...
+				LOG_ERR("error: srtp protection failed with code %d\n", stat);
+				return -EINVAL;
+			}
+			octets_sent = zsock_sendto(sock_fd, (void *)&message, pkt_len, 0,
+						   (struct sockaddr *)&addr, addr_len);
+
+			if (octets_sent != pkt_len) {
+				LOG_ERR("error: couldn't send message %s", (char *)word);
+			}
+			LOG_INF("sending word: %s", word);
```
josuah, line 207: It sometimes helps with proof-reading to keep the error handling and the function call in one block:

```suggestion

			/* apply srtp */
			stat = srtp_protect(srtp_ctx, (uint8_t *)&message.header, msg_len,
					    (uint8_t *)&message.header, &pkt_len, 0);
			if (stat) {
				LOG_ERR("error: srtp protection failed with code %d\n", stat);
				return -EINVAL;
			}

			bytes_sent = zsock_sendto(sock_fd, (void *)&message, pkt_len, 0,
						  (struct sockaddr *)&addr, addr_len);
			if (bytes_sent != pkt_len) {
				LOG_ERR("error: couldn't send message %s", (char *)word);
			}
```

samples/modules/libsrtp/rtpw/prj_receiver.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_NET_CONFIG_PEER_IPV4_ADDR="0.0.0.0"
```
josuah, line 2: You probably meant to also add `CONFIG_NET_SAMPLE_RECEIVER=y` here?

samples/modules/libsrtp/rtpw/prj_sender.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_NET_CONFIG_PEER_IPV4_ADDR="192.0.2.2"
```
josuah, line 2: You probably meant to also add `CONFIG_NET_SAMPLE_SENDER=y` here?

samples/modules/libsrtp/rtpw/src/main.c
josuah: Done as per latest commit, folding this thread so that a new one can be opened if needed.

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+   Send and receive data using (S)RTP protocol.
+
+Overveiw
+********
+
+A simple demo of using (S)RTP protocol to send and receive data over network. The sample could be sender or receiver.
+
+To test (S)RTP via a Linux machine, you should build the libsrtp.
+
+To test the sender while PC is receiver, select ``NET_SAMPLE_SENDER`` in Kconfig file. Next open a terminal and enter command:
```
josuah: I think it is possible to use this instead of \`\`NET_SAMPLE_SENDER\`\`:

```
:kconfig:option:`CONFIG_NET_SAMPLE_SENDER`
```

Here and everywhere, then it turns into a link.

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
+.. zephyr:code-sample:: rtpw
+   :name: SRTP client and server
+
+   Send and receive data using (S)RTP protocol.
+
+Overveiw
+********
+
+A simple demo of using (S)RTP protocol to send and receive data over network. The sample could be sender or receiver.
```
josuah: You can say "SRTP" instead of "(S)RTP" I think? As this sample only supports SRTP.

samples/modules/libsrtp/rtpw/README.rst
josuah: Small typo fix: "enter command" -> "enter this command"

You can also wrapping all the doc text to 100 characters wide max.

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
+.. zephyr:code-sample:: rtpw
+   :name: SRTP client and server
+
+   Send and receive data using (S)RTP protocol.
```
josuah: Suggestion:
"using (S)RTP protocol" -> "using the SRTP protocol using libsrtp"

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Overveiw
+********
+
+A simple demo of using (S)RTP protocol to send and receive data over network. The sample could be sender or receiver.
+
+To test (S)RTP via a Linux machine, you should build the libsrtp.
```
josuah:
```suggestion
To test SRTP via a Linux machine, you should build the `libsrtp`_ test program.
```

Then at the bottom of the file, adding a newline then this to create a link:

```rst
.. _libsrtp: https://github.com/cisco/libsrtp/
```

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+   [00:00:01.779,000] <inf> rtpw_sample: my IPv4 address: 192.0.2.1.
+   [00:00:18.010,000] <inf> rtpw_sample: receiving word: abducing
+   [00:00:18.510,000] <inf> rtpw_sample: receiving word: acidheads
+   [00:00:19.010,000] <inf> rtpw_sample: receiving word: acidness
+   [00:00:19.510,000] <inf> rtpw_sample: receiving word: actons
+
+Configuration
+*************
+
+The sender/receiver type and the SRTP key can be changed via Kconfig file.
```
josuah: suggestion:
"changed via Kconfig file" -> "changed via Kconfig"

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+You can use this application on a supported board
+
+Build the http-client sample application like this:
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/modules/libsrtp/rtpw
+   :board: <board to use>
+   :conf: <config file to use>
+   :goals: build
+   :compact:
```
josuah: Maybe this can be removed given there are the two examples given below?

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Sender mode
+===========
+
+Select the sender mode sample after setting ``NET_SAMPLE_SENDER``
+in the Kconfig file and building the project with the
+``overlay-sender.conf`` overlay file enabled using these commands:
```
josuah: Samples are expected to be run in every situation without changing any source.
If `CONFIG_NET_SAMPLE_SENDER=y` is already set in `overlay-sendoer.conf`, no more need to change `prj.conf`. :)

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/modules/libsrtp/rtpw
+   :board: frdm_mcxn947/mcxn947/cpu0
+   :goals: build
+   :compact:
+
+An alternative way after setting ``NET_SAMPLE_SENDER``
+in the Kconfig file is to specify ``-DEXTRA_CONF_FILE=overlay-sender.conf`` when
+running ``west build`` or ``cmake``.
```
josuah: How about showing the complete command, then possible to remove the extra text?

```suggestion

.. zephyr-app-commands::
   :zephyr-app: samples/modules/libsrtp/rtpw
   :board: frdm_mcxn947/mcxn947/cpu0
   :conf: "prj.conf overlay-sender.conf"
   :goals: build
   :compact:
```

I think this will generate exactly the command we need.

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Receiver mode
+=============
+
+Select the receiver mode sample after setting ``NET_SAMPLE_RECEIVER``
+in the Kconfig file and building the project with the
+``overlay-receiver.conf`` overlay file enabled using these commands:
```
josuah: As above:

```suggestion
Select the receiver mode by building the project with the
``overlay-receiver.conf`` overlay file enabled using this command:
```

samples/modules/libsrtp/rtpw/README.rst
```diff
@@ -0,0 +1,144 @@
...
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/modules/libsrtp/rtpw
+   :board: frdm_mcxn947/mcxn947/cpu0
+   :goals: build
+   :compact:
+
+An alternative way after setting ``NET_SAMPLE_RECEIVER``
+in the Kconfig file is to specify ``-DEXTRA_CONF_FILE=overlay-receiver.conf`` when
+running ``west build`` or ``cmake``.
```
josuah:
```suggestion
.. zephyr-app-commands::
   :zephyr-app: samples/modules/libsrtp/rtpw
   :board: frdm_mcxn947/mcxn947/cpu0
   :conf: "prj.conf overlay-receiver.conf"
   :goals: build
   :compact:
```

samples/modules/libsrtp/libsrtp.rst
```diff
@@ -0,0 +1,5 @@
+.. zephyr:code-sample-category:: libsrtp
+   :name: libSRTP
+   :show-listing:
+
+   These samples demonstrate how to use (S)RTP protocol in Zephyr.
```
josuah:
```suggestion
   These samples demonstrate how to use the SRTP protocol in Zephyr.
```

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	/* set address */
+#if defined(CONFIG_NET_IPV6)
+	struct sockaddr_in6 *addr6 = (struct sockaddr_in6 *)&addr;
+
+	LOG_INF("Peer IPv6 address: %s.", CONFIG_NET_CONFIG_PEER_IPV6_ADDR);
+	LOG_INF("My IPv6 address: %s.", CONFIG_NET_CONFIG_MY_IPV6_ADDR);
+
+	if (inet_pton(AF_INET6, CONFIG_NET_CONFIG_PEER_IPV6_ADDR, &addr6->sin6_addr) != 1) {
```
josuah: I just noticed there is a `zsock_` version for it:

```suggestion
	if (zsock_inet_pton(AF_INET6, CONFIG_NET_CONFIG_PEER_IPV6_ADDR, &addr6->sin6_addr) != 1) {
```

josuah (commented):
I think we are almost there.

We are still unsure if this needs to go upstream in Zephyr or if this goes in a downstream project, but either way this will help to complete these changes.

Thanks again for the modifications!

pdgendt:
> We are still unsure if this needs to go upstream in Zephyr or if this goes in a downstream project, but either way this will help to complete these changes.

I don't think it belongs in upstream Zephyr (in the current state). It's basically only a sample, without any glue code or dependency on Zephyr itself. It should be an external module that can be added to https://docs.zephyrproject.org/latest/develop/manifest/index.html#external-projects-modules

If the intention is to get WebRTC as a next step, that could be a different story, but not as is.

modules/libsrtp/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright 2026 Sayed Naser Moravej
+# SPDX-License-Identifier: Apache-2.0
+
+if(CONFIG_LIBSRTP)
+
+set(EXTRA_CONF_FILE
+    "${CMAKE_CURRENT_LIST_DIR}/libsrtp_psa.conf"
```
josuah: Is this file missing?

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,477 @@
...
+	/* check that hex string is the right length */
+	if (net_sample_key_len != input_key_len) {
+		LOG_ERR("wrong number of digits in key (should be %d digits, found %d).",
+			2 * input_key_len, 2 * net_sample_key_len);
+		return -EBADMSG;
+	}
+
+	/* initialize sender's rtp and srtp contexts */
+	rtp_msg_t message;
+	srtp_ctx_t *srtp_ctx;
```
josuah: It seems like all this code is exactly the same between sender and receiver.

How about de-duplicating it to reduce maintenance effort?

Using the original approach of using `#ifdef` just for the parts that change between receiver and sender would work well.

Maybe you can move all of this into `main()` and then move everything *below* to `app_sender()`/`app_receiver()`?

Or any other way for deduplicating this of course, just suggesting one way here, maybe there are others.

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,477 @@
...
+		LOG_ERR("Failed to close socket");
+		return ret;
+	}
+
+	status = srtp_shutdown();
+	if (status) {
+		LOG_ERR("srtp shutdown failed with error code %d.", status);
+		return -ECANCELED;
+	}
+	return 0;
```
josuah: Likewise, this could be moved to `main()` to deduplicate this code between sender and receiver part.

josuah (commented):
Thanks for the integrtion of PSA APIs and rebasing on the latest bugfixes.

I see some very small tuning to do here if interested.

samples/modules/libsrtp/rtpw/prj_receiver.conf
josuah, line 1: When using [`-D FILE_SUFFIX=<something>`](https://docs.zephyrproject.org/latest/develop/application/index.html) option, Zephyr will use `app_<something>.overlay` with an underscore `_` instead of dash `-` after `app`.

Let's imitate this pattern and rename the two files?

`overlay-receiver.conf` -> `prj_receiver.conf`
`overlay-sender.conf` -> `prj_sender.conf`

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,367 @@
...
+	int sock_fd, ret;
+	struct sockaddr_storage addr;
+	socklen_t addr_len;
+	srtp_policy_t policy;
+	srtp_err_status_t status;
+	size_t key_size = 128;
+	size_t tag_size = 8;
+	srtp_ctx_t *srtp_ctx = NULL;
+
+	memset(&addr, 0, sizeof(addr));
```
josuah: You can remove this by initializing `addr` like this if you want.

```
	struct sockaddr_storage addr = {};
```

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,367 @@
...
+uint32_t ssrc = 0xdeadbeefu; /* ssrc value hardcoded for now */
+
+__maybe_unused int app_sender(srtp_ctx_t *srtp_ctx, srtp_policy_t policy,
+	size_t key_len, int sock_fd, struct sockaddr_storage addr,
+	socklen_t addr_len, size_t salt_len)
+{
+
+	/* initialize sender's rtp and srtp contexts */
+	int ret = 0;
+	rtp_msg_t message;
```
josuah: This is actually large (1500 bytes), this is probably why `CONFIG_MAIN_STACK_SIZE` had to be set to `4096`.

I would put this one on a global variable, or eventually declaring a sys_heap and allocating it.

Then the main stack size can be reduced to maybe 2048 bytes?

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,367 @@
...
+	if (status) {
+		LOG_ERR("srtp_create() failed with code %d.", status);
+		return -EINVAL;
+	}
+
+
+	size_t word_len;
+	struct k_mutex lock;
+
+	k_mutex_init(&lock);
```
josuah: Is a lock really needed? As this is a single-threaded application.

josuah (commented):
I think this PR is in good state now.

It can be a good base for writing an RTP integration and libMP integration when these get ready.

Thank you for all the changes!

samples/modules/libsrtp/rtpw/src/main.c
seyednasermoravej: Yes, I reduced it to 2048. It worked. I also tested 1500 but it overflowed

josuah: <details><summary>nevermind, I just saw your edits now, all good thank you!</summary>

I think here the problem is not about the stack size itself, but about `rtp_msg_t`.

If we look above, it contains a 1500 bytes buffer:

```c
#define RTP_MAX_BUF_LEN 1500

typedef struct {
	srtp_hdr_t header;
	char body[RTP_MAX_BUF_LEN];    << here
} rtp_msg_t;
```

So we allocate 1500 bytes of stack wherever this buffer is present.

Example solutions for this: allocate it with i.e. `k_malloc()`, or easier in this case, make it a global variable as well, the goal being to keep the sample as simple as possible.

Not hugely important but it helps with reducing chances of stack overflow.

</details>

samples/modules/libsrtp/rtpw/src/main.c
```diff
@@ -0,0 +1,358 @@
...
+
+/*
+ * RTP_MAX_BUF_LEN defines the largest RTP packet
+ */
+#define RTP_MAX_BUF_LEN 1500
+
+typedef struct {
+	srtp_hdr_t header;
+	char body[RTP_MAX_BUF_LEN];
+} rtp_msg_t;
```
josuah: Zephyr, like Linux, seems to avoids typedefs as described here:

https://www.kernel.org/doc/html/latest/process/coding-style.html#typedefs

So maybe `struct rtp_msg` is good?

carlescufi:
@seyednasermoravej is this PR now using the RTP implementation that was recently merged?

seyednasermoravej:
> @seyednasermoravej is this PR now using the RTP implementation that was recently merged?

No, it's based on the Cisco RTP sample. 
I'm working on another branch to add support for the SRTP that is on top of this current PR and the RTP that was recently merged.

seyednasermoravej:
I opened a PR to add security on top of the [native RTP](https://github.com/zephyrproject-rtos/zephyr/pull/104603) implementation [here](https://github.com/zephyrproject-rtos/zephyr/pull/118267). 
@carlescufi

PR #99139, opened 2025-11-10
drivers/espi: it8xxx2: Add a config to share h2ram pool space

GTLin08:
The h2ram is an independent 4KB section. With default settings of CONFIG_ESPI_PERIPHERAL_HOST_CMD_PARAM_PORT_NUM and CONFIG_ESPI_PERIPHERAL_ACPI_SHM_REGION_PORT_NUM,
There is a 2KB gap that is unused. This change was made to make unused gap can be reused by other modules to reduce overall RAM usage.

### DIFF
--- a/drivers/espi/espi_it8xxx2.c
+++ b/drivers/espi/espi_it8xxx2.c
@@ -948,8 +948,14 @@ static const struct ec2i_t pmc2_settings[] = {
 #endif
 #endif
 
-static uint8_t h2ram_pool[MAX(H2RAM_ACPI_SHM_MAX, H2RAM_EC_HOST_CMD_MAX)]
-					__attribute__((section(".h2ram_pool")));
+#ifdef CONFIG_ESPI_IT8XXX2_H2RAM_SHARED
+#define H2RAM_STORAGE
+#else
+#define H2RAM_STORAGE static
+#endif
+
+H2RAM_STORAGE uint8_t h2ram_pool[MAX(H2RAM_ACPI_SHM_MAX, H2RAM_EC_HOST_CMD_MAX)]
+	__attribute__((section(".h2ram_pool")));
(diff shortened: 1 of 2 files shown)


drivers/espi/espi_it8xxx2.c
```diff
@@ -948,8 +948,12 @@ static const struct ec2i_t pmc2_settings[] = {
 #endif
 #endif
 
+#ifdef CONFIG_ESPI_IT8XXX2_H2RAM_SHARED
+uint8_t h2ram_pool[MAX(H2RAM_ACPI_SHM_MAX, H2RAM_EC_HOST_CMD_MAX)]
+#else
 static uint8_t h2ram_pool[MAX(H2RAM_ACPI_SHM_MAX, H2RAM_EC_HOST_CMD_MAX)]
-					__attribute__((section(".h2ram_pool")));
+#endif
+	__attribute__((section(".h2ram_pool")));
```
fabiobaltieri, line 958: how about using an intermediary macro for the static? something like

```c
#ifdef CONFIG_ESPI_IT8XXX2_H2RAM_SHARED
#define H2RAM_STORAGE
#else
#define H2RAM_STORAGE static
#endif

H2RAM_STORAGE uint8_t h2ram_pool[MAX(H2RAM_ACPI_SHM_MAX, H2RAM_EC_HOST_CMD_MAX)]
 					__attribute__((section(".h2ram_pool")));
```

just to avoid repeating the allocation since there's some logic in there

GTLin08: Done.

fabiobaltieri (approved):

Dino-Li (approved):

keith-zephyr (approved):

PR #99143, opened 2025-11-10
manifest: Update hal_adi revision

hfakkiz:
Update hal_adi revision to get HAL fix commits. One of the fix commits addresses an issue where the MAX32650 system clock frequency value was not set correctly, which could lead to improper operation of time-dependent functions.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -144,7 +144,7 @@ manifest:
       groups:
         - fs
     - name: hal_adi
-      revision: eeb155f7382343438114605963ae64436cc53434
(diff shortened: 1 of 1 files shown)


petejohanson-adi:
Can we actually bump directly to 4a189d5 please? Just got added, for a SYS driver fix for the USN functionality used by the hwinfo driver.

petejohanson-adi (approved):

MaureenHelm (approved):

PR #99147, opened 2025-11-10
drivers: timer: nrf_grtc: Decouple clock source from CLOCK_CONTROL

bjarki-andreasen:
The GRTC timer, typically used as sys clock on newer nordic chips,
is currently tightly coupled to the CLOCK_CONTROL_NRF drivers though
not being dependent on it. The GRTC and its device driver is
independent from CLOCK_CONTROL, its clock requirements are managed
by hardware, based on its clock source selection.

This commit moves the clock source selection to the GRTC driver, and
removes the hard coupling to the CLOCK_CONTROL drivers. To preserve
backwards compatibility, if CLOCK_CONTROL_NRF_K32SRC_RC is selected,
GRTC will default to LFLPRC.

### DIFF
--- a/drivers/timer/Kconfig.nrf_grtc
+++ b/drivers/timer/Kconfig.nrf_grtc
@@ -56,4 +56,29 @@ config NRF_GRTC_TIMER_AUTO_KEEP_ALIVE
 	  This feature prevents the SYSCOUNTER from sleeping when any core is in
 	  active state.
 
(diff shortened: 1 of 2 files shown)


nika-nordic (approved):

anangl (approved):

bjarki-andreasen:
@andyross You are assigned to this one, could you take a look? or reassign maybe?

PR #99152, opened 2025-11-10
Bluetooth: Host: Fix redundant ACL Rx buffer count

cvinayak:
Fix redundant ACL Rx buffers introduced when increasing HCI event buffer count.


cvinayak:
https://github.com/copilot/tasks/pull/PR_kwDOA5AKIc6yftj_?session_id=b545d136-354f-4c3d-816a-d9058420cc31

PR #99154, opened 2025-11-10
drivers: video: stm32-dcmi: correct set_frmival handling

avolmat-st:
Correct set_frmival in order to avoid having un-optimized frmival selection due to rounded values.
Computation has been done in usec, leading to incorrect frmival selection due to value being rounded internally. As an example, code was selecting 1/60 from sensor then 4 time frame drop by DCMI instead of selecting directly 1/15 from sensor.
Use msec instead to hide those rounding issue and avoid as well 64bit variables.
This also put the first video_frmival_nsec outside of the loop to avoid having to do the processing everytime.

### DIFF
--- a/drivers/video/video_stm32_dcmi.c
+++ b/drivers/video/video_stm32_dcmi.c
@@ -375,9 +375,11 @@ static int video_stm32_dcmi_set_frmival(const struct device *dev, struct video_f
 		.format = &data->fmt,
 	};
 	struct video_frmival best_sensor_frmival;
-	uint64_t best_diff_nsec = INT32_MAX;
-	uint64_t diff_nsec = 0, a, b;
+	uint32_t best_diff_us = INT32_MAX;
+	uint32_t diff_us = 0, a, b;
 	int best_capture_rate = 1;
+
+	a = video_frmival_nsec(frmival) / USEC_PER_MSEC;
 
 	/*
 	 * Try to figure out a frameinterval setting allow to reach as close as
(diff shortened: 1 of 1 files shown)


mjs513:
Will add to @KurtE's comment.  Tested with the HM0360 as well and again seems to have fixed the issue we were seeing.

drivers/video/video_stm32_dcmi.c
```diff
@@ -392,9 +394,8 @@ static int video_stm32_dcmi_set_frmival(const struct device *dev, struct video_f
 		fie.discrete.numerator = frmival->numerator;
 		fie.discrete.denominator = frmival->denominator * capture_rate;
 
-		a = video_frmival_nsec(&fie.discrete);
 		video_closest_frmival(config->sensor_dev, &fie);
-		b = video_frmival_nsec(&fie.discrete);
+		b = video_frmival_nsec(&fie.discrete) * capture_rate / USEC_PER_MSEC;
```
etienne-lms, line 398: The computation seems to change versus original code.
You get the diff versus original interval in `diff_nsec`, not the previously compute interval.
I may miss something (how `video_closest_frmival()` behaves regarding denominator mult by `capture_rate`) but I'd suggest:
```patch
-	a = video_frmival_nsec(frmival) / USEC_PER_MSEC;
+	b = video_frmival_nsec(frmival) / USEC_PER_MSEC;
 
 	for (int capture_rate = 1; capture_rate <= STM32_DCMI_MAX_FRAME_DROP; capture_rate *= 2) {
 		(...)
 
+		a = b;
 		video_closest_frmival(config->sensor_dev, &fie);
-		b = video_frmival_nsec(&fie.discrete) * capture_rate / USEC_PER_MSEC;
+		b = video_frmival_nsec(&fie.discrete) / USEC_PER_MSEC;

File: drivers/video/video_stm32_dcmi.c [hunk 2]
```diff
@@ -375,10 +375,12 @@ static int video_stm32_dcmi_set_frmival(const struct device *dev, struct video_f
 		.format = &data->fmt,
 	};
 	struct video_frmival best_sensor_frmival;
-	uint64_t best_diff_nsec = INT32_MAX;
-	uint64_t diff_nsec = 0, a, b;
+	uint32_t best_diff_nsec = INT32_MAX;
+	uint32_t diff_nsec = 0, a, b;
```
etienne-lms: Since values are computed in µs, I'd suggest the rename the variables (`best_diff_us`, `diff_us`) to prevent confusion.

drivers/video/video_stm32_dcmi.c
avolmat-st: Yes, change is on purpose. The original interval is the interval requested by the application, aka the one we try to reach as close as possible, that's the reason why it is now only computed once.
Then the loop is here to iterate over all capture_rate provided by the DCMI IP. The DCMI is capable of dropping frames, so for example if it drops 1 frame out of 2, the interval will be multiplied by two.
Since this drop is introduced by the DCMI, the code requests an updated interface to the sensor (basically denominator multiplied by the drop rate), and see how close it can reach it.
In order to properly compare with the originally computed interval (the requested one), it is necessary to multiply the computed frame interval (video_frmival_nsec) by the capture_rate (drop introduced by the DCMI) since the video_frmival_nsec is computed using the frmival given by the sensor (aka not taking into account the drop from the DCMI).

ngphibang (approved):
LGTM

erwango (approved):

PR #99157, opened 2025-11-10
soc: stm32f7: add power management support

RoyAc6:
Adds low-power mode support for STM32F7 series
Implements basic sleep and stop modes integrated
with the power management subsystem, similar to
the STM32F4 implementation.

Added Mode:
STOP low-power mode, exposed as Zephyr PM state
"suspend-to-idle".

Tested on: NUCLEO-F722ZE board.
Signed-off-by: Roy Jamil <roy.jamil@ac6.fr>

### DIFF
--- a/dts/arm/st/f7/stm32f7.dtsi
+++ b/dts/arm/st/f7/stm32f7.dtsi
@@ -27,6 +27,7 @@
 	chosen {
 		zephyr,entropy = &rng;
 		zephyr,flash-controller = &flash;
+		zephyr,cortex-m-idle-timer = &rtc;
 	};
 
 	cpus {
@@ -37,12 +38,22 @@
 			device_type = "cpu";
 			compatible = "arm,cortex-m7";
 			reg = <0>;
+			cpu-power-states = <&stop>;
 			#address-cells = <1>;
 			#size-cells = <1>;
 
 			mpu: mpu@e000ed90 {
 				compatible = "arm,armv7m-mpu";
 				reg = <0xe000ed90 0x40>;
+			};
+		};
+
+		power-states {
+			stop: stop {
+				compatible = "zephyr,power-state";
+				power-state-name = "suspend-to-idle";
+				min-residency-us = <400>;
+				exit-latency-us = <300>;
 			};
 		};
 	};
--- a/soc/st/stm32/stm32f7x/power.c
+++ b/soc/st/stm32/stm32f7x/power.c
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2025 Ac6
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/pm/pm.h>
+#include <clock_control/clock_stm32_ll_common.h>
+#include <soc.h>
+
+#include <stm32_ll_bus.h>
+#include <stm32_ll_cortex.h>
+#include <stm32_ll_pwr.h>
+
+LOG_MODULE_DECLARE(soc, CONFIG_SOC_LOG_LEVEL);
+
+void pm_state_set(enum pm_state state, uint8_t substate_id)
+{
+	ARG_UNUSED(substate_id);
+
+	switch (state) {
+	case PM_STATE_SUSPEND_TO_IDLE:
+		/* Prepare STOP mode with low-power regulator */
+		LL_LPM_DisableEventOnPend();
+		LL_PWR_ClearFlag_WU1();
+		LL_PWR_ClearFlag_WU2();
+		LL_PWR_ClearFlag_WU3();
+		LL_PWR_ClearFlag_WU4();
+		LL_PWR_ClearFlag_WU5();
+		LL_PWR_ClearFlag_WU6();
+
+		/* Use low-power regulator in STOP to reduce consumption */
+		LL_PWR_SetPowerMode(LL_PWR_MODE_STOP_LPREGU);
+		LL_LPM_EnableDeepSleep();
+
+		__WFI();
+
+		break;
+	default:
+		LOG_DBG("Unsupported power state %u", state);
+		break;
+	}
+}
+
+void pm_state_exit_post_ops(enum pm_state state, uint8_t substate_id)
+{
+	ARG_UNUSED(substate_id);
+
+	switch (state) {
+	case PM_STATE_SUSPEND_TO_IDLE:
+		/* Back to normal sleep and restore clocks after STOP */
+		LL_LPM_DisableSleepOnExit();
+		LL_LPM_EnableSleep();
+
+		/* Restore the clock setup after STOP mode */
+		stm32_clock_control_init(NULL);
+		break;
+
+	default:
+		LOG_DBG("Unsupported power substate-id %u", state);
+		break;
+	}
+
+	/* System is now active, re-enable interrupts disabled on entry */
+	irq_unlock(0);
+}
--- a/soc/st/stm32/stm32f7x/soc.c
+++ b/soc/st/stm32/stm32f7x/soc.c
@@ -16,6 +16,7 @@
 #include <soc.h>
 
 #include <cmsis_core.h>
+#include <stm32_ll_bus.h>
 #include <stm32_ll_system.h>
 
 /**
@@ -37,4 +38,7 @@ void soc_early_init_hook(void)
 	/* Update CMSIS SystemCoreClock variable (HCLK) */
 	/* At reset, system core clock is set to 16 MHz from HSI */
 	SystemCoreClock = 16000000;
+
+	/* Ensure PWR peripheral clock is enabled on APB1 */
+	LL_APB1_GRP1_EnableClock(LL_APB1_GRP1_PERIPH_PWR);
 }
--- a/soc/st/stm32/stm32f7x/CMakeLists.txt
+++ b/soc/st/stm32/stm32f7x/CMakeLists.txt
@@ -8,3 +8,7 @@ zephyr_sources(
 zephyr_include_directories(.)
 
 set(SOC_LINKER_SCRIPT ${ZEPHYR_BASE}/include/zephyr/arch/arm/cortex_m/scripts/linker.ld CACHE INTERNAL "")
+
+zephyr_sources_ifdef(CONFIG_PM
+  power.c
+  )
--- a/soc/st/stm32/stm32f7x/Kconfig
+++ b/soc/st/stm32/stm32f7x/Kconfig
@@ -14,4 +14,5 @@ config SOC_SERIES_STM32F7X
 	select HAS_STM32_FLASH_PREFETCH
 	select CPU_HAS_ARM_MPU
 	select HAS_SWO
+	select HAS_PM
 	select SOC_EARLY_INIT_HOOK


soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /* System is now active, re-enable interrupts disabled on entry */
+    irq_unlock(0);
+}
+
+void stm32_power_init(void)
+{
+    /* Ensure PWR peripheral clock is enabled on APB1 */
+    LL_APB1_GRP1_EnableClock(LL_APB1_GRP1_PERIPH_PWR);
```
JarmouniA: Please move this to `soc_early_init_hook()`, PWR is needed for things (Zephyr Poweroff support for ex.) other than PM.

RoyAc6: I've moved `LL_APB1_GRP1_EnableClock(LL_APB1_GRP1_PERIPH_PWR);` directly into `soc_early_init_hook()`.  

It might also be worth reviewing and cleaning up the following STM32 SoC files:
- STM32F4x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32f4x/power.c#L78  
- STM32L4x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32l4x/power.c#L117  
- STM32L5x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32l5x/power.c#L101  
- STM32H5x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32h5x/power.c#L71  
- STM32G4x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32g4x/power.c#L82  
- STM32U5x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32u5x/power.c#L173  
- STM32WBAx: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32wba/power.c#L210  
- STM32G0x: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32g0x/power.c#L84  

Do you want me to do it ?

JarmouniA: I think that's out-of-scope for this PR.

soc/st/stm32/stm32f7x/soc.c
```diff
@@ -37,4 +37,9 @@ void soc_early_init_hook(void)
 	/* Update CMSIS SystemCoreClock variable (HCLK) */
 	/* At reset, system core clock is set to 16 MHz from HSI */
 	SystemCoreClock = 16000000;
+
+#if defined(CONFIG_PM)
```
JarmouniA:
```suggestion
```

soc/st/stm32/stm32f7x/power.c
RoyAc6: Yes, I mean in another PR

erwango: We have a PR in preparation to clean PWR clock handling globally, so don't bother too much.

dts/arm/st/f7/stm32f7.dtsi
```diff
@@ -38,13 +39,24 @@
...
 			reg = <0>;
+			cpu-power-states = <&stop>;
 			#address-cells = <1>;
 			#size-cells = <1>;
 
 			mpu: mpu@e000ed90 {
 				compatible = "arm,armv7m-mpu";
 				reg = <0xe000ed90 0x40>;
 			};
+			};
```
erwango: This is suspicious...

dts/arm/st/f7/stm32f7.dtsi
```diff
@@ -38,13 +39,24 @@
...
+
+
+			power-states {
+			stop: stop {
+				compatible = "zephyr,power-state";
+				power-state-name = "suspend-to-idle";
+				min-residency-us = <400>;
+				exit-latency-us = <300>;
+			};
```
erwango: Indentation issue + missing `};`

soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,77 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <clock_control/clock_stm32_ll_common.h>
+#include <soc.h>
+
+#include <stm32f7xx_ll_bus.h>
+#include <stm32f7xx_ll_cortex.h>
+#include <stm32f7xx_ll_pwr.h>
+#include <stm32f7xx.h>
```
erwango: See https://github.com/zephyrproject-rtos/zephyr/pull/99012

soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,77 @@
...
+
+#include <zephyr/drivers/clock_control/stm32_clock_control.h>
+#include <zephyr/drivers/counter.h>
+#include <zephyr/drivers/interrupt_controller/gic.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/pm/pm.h>
+#include <zephyr/init.h>
```
etienne-lms: Suggestion
```patch
-#include <zephyr/drivers/clock_control/stm32_clock_control.h>
-#include <zephyr/drivers/counter.h>
-#include <zephyr/drivers/interrupt_controller/gic.h>
 #include <zephyr/kernel.h>
 #include <zephyr/pm/pm.h>
-#include <zephyr/init.h>

File: dts/arm/st/f7/stm32f7.dtsi (hunk 4 above)
Comment (kylebonnici) [reply dts/arm/st/f7/stm32f7.dtsi]: @erwango  If there are missing }; the dts-linter would complain so I do not believe we have any missing };

Comment (etienne-lms) [reply dts/arm/st/f7/stm32f7.dtsi]: It's rather the indentation that is fuzzy. Indeed compiling this DTS is valid. But CI DTS linter compliance test obviously complains.

Comment (RoyAc6) [reply dts/arm/st/f7/stm32f7.dtsi]: It's an indentation issue. I'm not sure how it ended up like this!

File: soc/st/stm32/stm32f7x/power.c (hunk 5 above)
Comment (RoyAc6) [reply soc/st/stm32/stm32f7x/power.c]: replaced with 
```
#include <stm32_ll_bus.h>
#include <stm32_ll_cortex.h>
#include <stm32_ll_pwr.h>
```

File: soc/st/stm32/stm32f7x/power.c (hunk 6 above)
Comment (RoyAc6) [reply soc/st/stm32/stm32f7x/power.c]: done

File: soc/st/stm32/stm32f7x/power.c [hunk 7]
```diff
@@ -0,0 +1,71 @@
...
+#include <soc.h>
+
+#include <stm32_ll_bus.h>
+#include <stm32_ll_cortex.h>
+#include <stm32_ll_pwr.h>
+
+LOG_MODULE_DECLARE(soc, CONFIG_SOC_LOG_LEVEL);
+
+BUILD_ASSERT(DT_SAME_NODE(DT_CHOSEN(zephyr_cortex_m_idle_timer), DT_NODELABEL(rtc)),
+		"STM32Fx series needs RTC as an additional IDLE timer for power management");
```
mathieuchopstm:
```suggestion
		"STM32F7x series needs RTC as an additional IDLE timer for power management");
```
\+ two remarks:
* Have you checked that this doesn't explode if the `DT_CHOSEN` doesn't exist?
* Most STM32F7 SoCs also have a LPTIM so this check is not sufficient

I'd recommend just getting rid of the check: user is responsible for configuring their device properly, ***and we are doing the proper configuration by default*** so there's little error risk.

soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,71 @@
...
+		LL_PWR_ClearFlag_WU3();
+		LL_PWR_ClearFlag_WU4();
+		LL_PWR_ClearFlag_WU5();
+		LL_PWR_ClearFlag_WU6();
+
+		/* Use low-power regulator in STOP to reduce consumption */
+		LL_PWR_SetPowerMode(LL_PWR_MODE_STOP_LPREGU);
+		LL_LPM_EnableDeepSleep();
+
+		k_cpu_idle();
```
mathieuchopstm: calling `k_cpu_idle()` here is illegal (it unlocks IRQs)
```suggestion
		__WFI();
```

mathieuchopstm (dismissed):
NAK for `k_cpu_idle` usage

erwango (changes requested):
Please enable in samples/boards/st/power_mgmt/wkup_pins/boards

Also, can you check if some Kconfig configuration wouldn't be missing ?
See https://github.com/zephyrproject-rtos/zephyr/pull/100246

etienne-lms (commented):
Could you squash the 3 commits into a single one? I think you can discard the `Co-authored-by` tags related to suggestions provided during the P-R review, they are not needed.

soc/st/stm32/stm32f7x/power.c
RoyAc6: > I'd recommend just getting rid of the check: user is responsible for configuring their device properly, _**and we are doing the proper configuration by default**_ so there's little error risk.

done

soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,68 @@
...
+#include <zephyr/pm/pm.h>
+#include <clock_control/clock_stm32_ll_common.h>
+#include <soc.h>
+
+#include <stm32_ll_bus.h>
+#include <stm32_ll_cortex.h>
+#include <stm32_ll_pwr.h>
+
+LOG_MODULE_DECLARE(soc, CONFIG_SOC_LOG_LEVEL);
+
```
etienne-lms: Suggestion, for consistency and build-time check of RTC use (see stm32f*
```c
BUILD_ASSERT(DT_SAME_NODE(DT_CHOSEN(zephyr_cortex_m_idle_timer), DT_NODELABEL(rtc)),
	     "STM32Fx series needs RTC as an additional IDLE timer for power management");

File: soc/st/stm32/stm32f7x/power.c [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+void pm_state_set(enum pm_state state, uint8_t substate_id)
+{
+	ARG_UNUSED(substate_id);
+
+	switch (state) {
+	case PM_STATE_SUSPEND_TO_IDLE:
+		/* Prepare STOP mode with low-power regulator */
+		LL_LPM_DisableEventOnPend();
```
mathieuchopstm, line 26: This has no effect when WFI is used, and Zephyr expects this bit to remain at 1. Leave it untouched.
```suggestion
```
(yes, I looked at tree and we do that everywhere... but it's still wrong, other implementations should be fixed)

soc/st/stm32/stm32f7x/power.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void pm_state_exit_post_ops(enum pm_state state, uint8_t substate_id)
+{
+	ARG_UNUSED(substate_id);
+
+	switch (state) {
+	case PM_STATE_SUSPEND_TO_IDLE:
+		/* Back to normal sleep and restore clocks after STOP */
+		LL_LPM_DisableSleepOnExit();
```
mathieuchopstm, line 54: This bit was not set in the first place, no point clearing it.
```suggestion
```

etienne-lms: @mathieuchopstm, many STM32 SoCs **power.c** file do call this routine despite `LL_LPM_EnableSleepOnExit()` is never called. Should these calls be removed to avoid confusion?

mathieuchopstm: Yes but that's for another (cleanup) PR. I'm aware of this issue, we have lots of stuff to revisit...

RoyAc6:
@etienne-lms I tried it after with your suggestions, I'm still having some issues to have it working. I'm not sure what I'm missing, I'll give it a try on a different board

PR #99163, opened 2025-11-10
drivers: i2c: i2c_xilinx_axi: switch to DEVICE_MMIO API

mausys:
Switch the Xilinx AXI I2C device driver to use the DEVICE_MMIO_... macros instead of directly using the physical base address from the device tree. This allows the driver to work on SoCs with an MMU, such as the Zynq-7000.


henrikbrixandersen (approved):

neeliajay:
I validated the MMIO macros change on actual hardware (Kintex KCU105 and ZynqMP).
I ran EEPROM tests over the board's AXI-I2C interface and observed the expected behavior.
The driver now derives the I2C base address using MMIO macros rather than using a raw address taken directly from the device tree.

michalsimek (approved):

PR #99166, opened 2025-11-10
display: ensure samples check display api func return value + ltdc read/write rectangle check

avolmat-st:
Following PR #98719, add proper display API function return value checks in all samples in order to fail if an error happen.

Moreover, within the LTDC driver, ensure that the rectangle given (x,y / width-height) properly fit into the framebuffer in order xto avoid OOB access.

Such check is currently done within the LTDC only, however I think that some of those checks should be done within the API functions themself to ensure a first level of proper parameter validity. In order to have proper behavior of all display drivers, some level of framework level checks can be done, removing amount of duplicated code between drivers.
I will prepare a PR for that purpose and propose it.

### DIFF
--- a/samples/drivers/display/src/main.c
+++ b/samples/drivers/display/src/main.c
@@ -222,6 +222,7 @@ int main(void)
 	struct display_buffer_descriptor buf_desc;
 	size_t buf_size = 0;
 	fill_buffer fill_buffer_fnc = NULL;
+	int ret;
 
 	display_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_display));
 	if (!device_is_ready(display_dev)) {
@@ -362,7 +363,15 @@ int main(void)
 		if ((capabilities.y_resolution - idx) < h_step) {
 			buf_desc.height = (capabilities.y_resolution - idx);
 		}
-		display_write(display_dev, 0, idx, &buf_desc, buf);
+		ret = display_write(display_dev, 0, idx, &buf_desc, buf);
+		if (ret < 0) {
+			LOG_ERR("Failed to write to display (error %d)", ret);
+#ifdef CONFIG_ARCH_POSIX
+			posix_exit_main(1);
+#else
+			return 0;
+#endif
+		}
 	}
 
 	buf_desc.pitch = rect_w;
@@ -372,12 +381,28 @@ int main(void)
 	fill_buffer_fnc(TOP_LEFT, 0, buf, buf_size);
 	x = 0;
 	y = 0;
-	display_write(display_dev, x, y, &buf_desc, buf);
+	ret = display_write(display_dev, x, y, &buf_desc, buf);
+	if (ret < 0) {
+		LOG_ERR("Failed to write to display (error %d)", ret);
+#ifdef CONFIG_ARCH_POSIX
+		posix_exit_main(1);
+#else
+		return 0;
+#endif
+	}
 
 	fill_buffer_fnc(TOP_RIGHT, 0, buf, buf_size);
 	x = capabilities.x_resolution - rect_w;
 	y = 0;
-	display_write(display_dev, x, y, &buf_desc, buf);
+	ret = display_write(display_dev, x, y, &buf_desc, buf);
+	if (ret < 0) {
+		LOG_ERR("Failed to write to display (error %d)", ret);
+#ifdef CONFIG_ARCH_POSIX
+		posix_exit_main(1);
+#else
+		return 0;
+#endif
+	}
 
 	/*
 	 * This is the last write of the frame, so turn this off.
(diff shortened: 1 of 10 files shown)


samples/modules/lvgl/demos/src/main.c
```diff
@@ -61,7 +62,12 @@ int main(void)
 #endif
 	lvgl_unlock();
 
-	display_blanking_off(display_dev);
+	ret = display_blanking_off(display_dev);
+	if (ret < 0 && ret != -ENOSYS) {
+		LOG_ERR("Failed to turn blanking off (error %d)", ret);
+		return 0;
```
faxe1008, line 68: return ret

avolmat-st: Actually I have been wondering about what to return and couldn't really find a common behavior between samples. For example few lines above, if the device is not ready, gives an error message but also return 0.

In some other sample, a non 0 value is returned but most common behavior seems to return 0.

In this PR I tried as much as possible to stick to what is done within each sample even if I feel it weird to return 0 on error. I would be ok to return ret in case of error in fact but, this has to be done commonly I think with other samples.

uLipe (dismissed):

samples/modules/lvgl/demos/src/main.c
JarmouniA: Samples should return 0.

samples/modules/lvgl/screen_transparency/src/main.c
```diff
@@ -56,7 +57,11 @@ int main(void)
 	initialize_gui();
 
 	lv_timer_handler();
-	display_blanking_off(display_dev);
+	ret = display_blanking_off(display_dev);
+	if (ret < 0 && ret != -ENOSYS) {
+		LOG_ERR("Failed to turn blanking off (error %d)", ret);
+		return ret;
```
JarmouniA:
```suggestion
		return 0;
```

avolmat-st: Thanks. fixed here and another occurence as well.

samples/drivers/display/src/main.c
```diff
@@ -389,9 +414,25 @@ int main(void)
...
 	y = capabilities.y_resolution - rect_h;
-	display_write(display_dev, x, y, &buf_desc, buf);
+	ret = display_write(display_dev, x, y, &buf_desc, buf);
+	if (ret < 0) {
+		LOG_ERR("Failed to write to display (error %d)", ret);
+#ifdef CONFIG_ARCH_POSIX
+		posix_exit_main(1);
+#else
+		return 0;
+#endif
```
josuah, line 424: How about `return SAMPLE_EXIT_FAILURE;` with an ifdef like this?

```c
#ifdef CONFIG_ARCH_POSIX
#define SAMPLE_EXIT_FAILURE 1
#else
#define SAMPLE_EXIT_FAILURE 0
#endif
```

JarmouniA (approved):

samples/drivers/display/src/main.c
ngphibang: May I know if there is a reason that in this sample returning 0 or 1 will depend on `CONFIG_ARCH_POSIX` while in other samples, we always return 0 ?

avolmat-st: I am guessing a bit here. In "normal" cases, main is getting called from the kernel/init.c code, aka

```c
#ifdef CONFIG_BOOTARGS
        extern int main(int, char **);
        extern char **prepare_main_args(int *argc);

        int argc = 0;
        char **argv = prepare_main_args(&argc);
        (void)main(argc, argv);
#else
        extern int main(void);

        (void)main();
#endif /* CONFIG_BOOTARGS */
```

So here in fact return value isn't much used apparently.

In case of POSIX mode (I suspect this is only for the native_sim (?)), a non zero return value can be used to be able to return back the error code and do some testing on the proper behavior of the sample run via native_sim ?

avolmat-st: @JarmouniA, could you be more explicit about the fact that samples should return 0 (cf https://github.com/zephyrproject-rtos/zephyr/pull/99166#discussion_r2513248849) ?

JarmouniA: The application's return value is unused (outside native_sim).

fabiobaltieri (approved):

samples/drivers/display/src/main.c
josuah: So it actually needs the `posix_exit_main()` call if wanting to return the error from `native_sim`. Thank you for clarifying.

danieldegrasse (approved):

uLipe (approved):

avolmat-st:
@faxe1008 could you have a look at this again ?

vlad-kulikov (approved):

faxe1008 (approved):

PR #99167, opened 2025-11-10
drivers: tmag5273: Add mag gain correction property

marekmaskarinec:
Adds `mag-gain-correction` DT property, which sets the mag gain correction value.

It was already possible to select the channel for the gain correction `ch-mag-gain-correction`, but there is no property to choose the correction value itself.

### DIFF
--- a/dts/bindings/sensor/ti,tmag5273.yaml
+++ b/dts/bindings/sensor/ti,tmag5273.yaml
@@ -134,6 +134,18 @@ properties:
 
       Only active if angle-magnitude-calculation is active.
 
+  mag-gain-correction:
+    type: int
+    default: 0
+    description: |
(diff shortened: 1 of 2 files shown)


dts/bindings/sensor/ti,tmag5273.yaml
```diff
@@ -134,6 +134,16 @@ properties:
 
       Only active if angle-magnitude-calculation is active.
 
+  mag-gain-correction:
+    type: int
+    default: 0
```
MaureenHelm, line 139: Need to explain in the description why the default value was selected
https://docs.zephyrproject.org/latest/build/dts/bindings-upstream.html#dt-bindings-default-rules

MaureenHelm (approved):

jpanisbl (approved):

Issue #5501, opened 2017-12-28
Deprecate qemu.cmake in favour of a qemu runner backend


SebastianBoe:
Hi,

I suspect we want to eventually rename/alias 'flash' to 'run'. And treat qemu emulation as just
another runner backend. I believe this would make it easier for us to maintain the runner and qemu
infrastructure, and would make it easier for us to keep the user-interface for flashing and running
consistent.

A qemu backend would also allow us to move all the symlink/fifo/pipehandling out of qemu.cmake and
into a python runner backend.

It is not clear how difficult it would be to do this.

SebastianBoe:
@mbolivar : Did you already intend to move in this direction? I see you have added a dummy qemu.py backend already.

mbolivar:
@SebastianBoe Happy new year!

That was a copy of a dummy qemu.sh that was already there when I started working on the runner backends. I created a dummy Python script just so people would have a place to start if they want to continue this work. I have no plans to do it.

SebastianBoe:
Happy new year :)

Do you think it might make sense to incorporate running qemu into the runner infrastructure?

mbolivar:
I definitely think it makes sense. It's just another way to run the code, after all. I haven't tried the new "native" port, but I think that would make sense too -- keeping interfaces uniform is good usability.

mbolivar:
Upgrading the priority since I've received several questions about this recently.

nashif:
see #35366

zephyrbot:
Hi @nashif,

This issue, marked as an Enhancement, was opened a while ago and did not get any traction. Please confirm the issue is correctly assigned and re-assign it otherwise.

Please take a moment to review if the issue is still relevant to the project.  If it is, please provide feedback and direction on how to move forward. If it is not, has already been addressed, is a duplicate, or is no longer relevant, please close it with a short comment explaining the reason.


@SebastianBoe you are also encouraged to help moving this issue forward by providing additional information and confirming this request/issue is still relevant to you.

Thanks!

nordicjm:
Seems like this was implemented years ago?

mbolivar:
> Seems like this was implemented years ago?

no; https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/west_commands/runners/qemu.py has always been a stub waiting for a hero to step up and fill it in

PR #99170, opened 2025-11-10
runners: add and improve QEMU runner (Python-based replacement for qemu.cmake)

ritesh006:
This PR introduces a Python-based QEMU runner (`scripts/west_commands/runners/qemu.py`)
as part of the effort to deprecate the legacy `cmake/emu/qemu.cmake`.

Key points:
- Implements a ZephyrBinaryRunner subclass for QEMU
- Supports --qemu-binary, --qemu-arg, and serial backends (stdio, fifo, pty)
- Provides RunnerCaps(run=True) for consistent `west run` integration
- Cleans up temporary FIFOs safely after execution
- Adds a deprecation shim in cmake/emu/qemu.cmake

Tested on:
- `qemu_x86` board running `samples/hello_world` (successful boot and output)
- Zephyr SDK 0.16.x with QEMU 8.2.2

This addresses part of [zephyrproject-rtos/zephyr#5501](https://github.com/zephyrproject-rtos/zephyr/issues/5501)  
and sets the foundation for full migration from CMake to Python-based QEMU runner.

### DIFF
--- a/cmake/emu/qemu.cmake
+++ b/cmake/emu/qemu.cmake
@@ -473,3 +473,10 @@ foreach(target ${qemu_targets})
     add_dependencies(${target} qemu_nvme_disk qemu_kernel_target)
   endif()
 endforeach()
+
+# cmake/emu/qemu.cmake
+message(WARNING "qemu.cmake is deprecated. QEMU support is moving to a Python runner (scripts/west_commands/runners/qemu.py). "
+                "Please prefer 'west build -b <qemu_board> && west build -t run' or use the 'qemu' runner with 'west flash -r qemu'.")
+
+# Optionally, export a small compatibility variable or call the old implementation if still present.
+# If you want to preserve exact behavior, call the legacy implementation here, otherwise keep this as a shim.
--- a/scripts/west_commands/runners/qemu.py
+++ b/scripts/west_commands/runners/qemu.py
@@ -1,31 +1,141 @@
-# Copyright (c) 2017 Linaro Limited.
-#
+# Copyright (c) 2024 Ritesh Kudkelwar
 # SPDX-License-Identifier: Apache-2.0
 
-'''Runner stub for QEMU.'''
+"""QEMU runner implementation."""
 
-from runners.core import RunnerCaps, ZephyrBinaryRunner
+import contextlib
+import logging
+import os
+import shlex
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
+
+from runners.core import RunnerConfig, ZephyrBinaryRunner
+
+log = logging.getLogger(__name__)
 
 
-class QemuBinaryRunner(ZephyrBinaryRunner):
-    '''Place-holder for QEMU runner customizations.'''
+# Runner metadata
+class QemuRunner(ZephyrBinaryRunner):
+    name = "qemu"
+    description = "Run Zephyr ELF using QEMU (starter implementation)"
 
     @classmethod
-    def name(cls):
-        return 'qemu'
+    def add_parser(cls, parser):
+        super().add_parser(parser)
+        parser.add_argument(
+            "--qemu-binary",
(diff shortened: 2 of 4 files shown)


pdgendt:
Please fix compliance and CI issues, also since you considerately modified the file, remove it from the ruff excludes file and format it using `ruff`.

https://github.com/zephyrproject-rtos/zephyr/blob/13daf245a801a0b3d79a67221e4b599fd1cb3ad9/.ruff-excludes.toml#L1415

cmake/emu/qemu.cmake
```diff
@@ -99,12 +99,12 @@ endif()
 if(CONFIG_QEMU_ICOUNT)
   if(CONFIG_QEMU_ICOUNT_SLEEP)
     list(APPEND QEMU_FLAGS
-	  -icount shift=${CONFIG_QEMU_ICOUNT_SHIFT},align=off,sleep=on
-	  -rtc clock=vm)
+    -icount shift=${CONFIG_QEMU_ICOUNT_SHIFT},align=off,sleep=on
```
nordicjm: these changes are not valid with this alignment

nordicjm (changes requested):

scripts/west_commands/runners/qemu.py
```diff
@@ -1,31 +1,183 @@
...
+        """
+        Entry point invoked when user runs "west flash/run -r qemu" (or similar).
+        This should prepare FIFOs/serial, build command line, and spawn QEMU.
+        """
+        # 1) Ensure qemu binary
+        qemu_bin = kwargs.get("qemu_binary") or self._guess_qemu_binary()
+        if not qemu_bin or not shutil.which(qemu_bin):
+            raise RuntimeError(
+                "QEMU binary not found: set --qemu-binary or install qemu-system-* on PATH"
+            )
```
pdgendt: Use `self.require(qemu_bin)` instead?

scripts/west_commands/runners/qemu.py
```diff
@@ -1,31 +1,183 @@
...
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
 
+from runners.core import RunnerConfig, ZephyrBinaryRunner
 
-class QemuBinaryRunner(ZephyrBinaryRunner):
-    '''Place-holder for QEMU runner customizations.'''
+log = logging.getLogger(__name__)
```
pdgendt, line 17: Runners already have a logger, as `self.logger`.

scripts/west_commands/runners/qemu.py
```diff
@@ -1,31 +1,183 @@
...
+        # Minimal defaults. Extend per-board: memory, machine type etc.
+        # For common x86 qemu_x86: use -M pc -m 512M -nographic
+        board = (self.cfg.board or "").lower() if hasattr(self.cfg, "board") else ""
+        if "qemu_x86" in board or "qemu_x86" in str(self.cfg.board_dir):
+            return ["-M", "pc", "-m", "512", "-nographic"]
+        # cortex-m3 qemu example:
+        if "cortex_m3" in board or "qemu_cortex_m3" in str(self.cfg.board_dir):
+            return ["-M", "lm3s6965evb", "-nographic"]
+        # fall back to minimal no-graphic
+        return ["-nographic"]
```
pdgendt: This should not be done here, but instead passed as arguments from the boards. Search `board_runner_args` for examples.

pdgendt (changes requested):

scripts/west_commands/runners/qemu.py
ritesh006: done

beriberikix:
Agree it's probably a separate use case!

scripts/west_commands/runners/qemu.py
pdgendt: This comment is not resolved yet.

PR #99171, opened 2025-11-10
stm32: configure power supply through Device Tree

mathieuchopstm:
Replace the existing Kconfig-based mechanism to select the power supply with a Devicetree-based approach, inspired by the STM32WB0 implementation.

Commits could be reorganized: currently, they're split by series, but also squash SoC DTSI and board DTS changes together - I could mark properties as not required initially then perform updates separately, and make required in a last-pass commit if that sounds cleaner.


dts/arm/st/h7/stm32h7.dtsi
```diff
@@ -164,6 +164,12 @@
 			};
 		};
 
+		pwr: power@58024800 {
+			compatible = "st,stm32h7-pwr";
+			reg = <0x58024800 0x400>;
+			status = "disabled";
```
etienne-lms: It seems boards needs to enable `pwr` node unless what a property is mandated and used from a disabled DT node, which while functional seems awkward.

dts/arm/st/u5/stm32u5.dtsi
```diff
@@ -888,10 +888,17 @@
...
 		pwr: power@46020800 {
-			compatible = "st,stm32-pwr";
+			compatible = "st,stm32-dualreg-pwr", "st,stm32-pwr";
 			reg = <0x46020800 0x400>; /* PWR register bank */
 			status = "disabled";
 
+			/*
+			 * After reset, the LDO regulator is selected as
+			 * it works in all hardware configurations; set
+			 * default in SoC DTSI to reduce churn in boards.
```
etienne-lms, line 898: `burden` instead of `churn`?

boards/arduino/portenta_h7/arduino_portenta_h7-common.dtsi
```diff
@@ -51,6 +51,11 @@
 	d3ppre = <2>;
 };
 
+&pwr {
+	power-supply = "smps-ext-ldo";
```
etienne-lms: `smps-ldo`?

mathieuchopstm: You're right, it was `CONFIG_POWER_SUPPLY_SMPS_1V8_SUPPLIES_LDO` - will fix

dts/arm/st/h7/stm32h7.dtsi
mathieuchopstm: Enabling `pwr` node brings the wake-up pin code for which support is not implemented here.

I do agree that is it a bit awkward but I think we have other such patterns in tree.

(Another solution is to implement wake-up pins support first, then do the PWR transition...)

mathieuchopstm: Although now I realize that having the node as `disabled` makes `required: true` useless. Since *both* `status = "okay";` and `compatible = "st,stm32-pwr"` is required for wake-up pins, I'll drop the explicit disable

etienne-lms:
As CI build notes, there are a few board DTS files that need to be updated with `&pwr { power-supply = "ldo"; };`, all embedding an STM32H7xx SoC:
*boards/ruiside/art_pi/art_pi.dts*
*boards/st/nucleo_h723zg/nucleo_h723zg.dts*
*boards/st/nucleo_h743zi/nucleo_h743zi.dts*
*boards/st/nucleo_h753zi/nucleo_h753zi.dts*
*boards/st/stm32h735g_disco/stm32h735g_disco.dts*
*boards/witte/linum/linum.dts*
*boards/weact/mini_stm32h743/mini_stm32h743.dts*
*boards/weact/mini_stm32h7b0/mini_stm32h7b0.dts*
*boards/fanke/fk7b0m1_vbt6/fk7b0m1_vbt6.dts*
*boards/fanke/fk750m1_vbt6/fk750m1_vbt6.dts*
*boards/fanke/fk723m1_zgt6/fk723m1_zgt6.dts*
*boards/fanke/fk743m5_xih6/fk743m5_xih6.dts*

dts/arm/st/wba/stm32wba.dtsi
```diff
@@ -165,7 +165,7 @@
 		};
 
 		pwr: power@46020800 {
-			compatible = "st,stm32-pwr";
+			compatible = "st,stm32-dualreg-pwr", "st,stm32-pwr";
```
asm5878, line 168: Why we need to add in explicit way "st,stm32-pwr"? For what I have understood "st.stm32-dualreg-pwr" is an extension of it.

bjarki-andreasen (dismissed):
Cool

etienne-lms (dismissed):
@mathieuchopstm, could you add a message body to commit "migration-guide: 4.4: add entry about STM32 power configuration"?
Otherwise LAGTM.

dts/arm/st/u5/stm32u5.dtsi
```diff
@@ -888,10 +888,17 @@
...
+			 */
+			power-supply = "ldo";
```
erwango, line 900: Is it different on WBA ?

erwango (commented):
Otherwise LGTM

etienne-lms (dismissed):

dts/arm/st/u5/stm32u5.dtsi
mathieuchopstm: I *think* it should be the same. The reason I set a default for U5 only is that basically all U5 boards in tree rely on the old `default POWER_SUPPLY_LDO`, and there are many of them, whereas there are only 2~3 WBA boards *and* they're all SMPS (which *cannot* be set as default, because it doesn't work on e.g. WBA52)

asm5878: For WBA we should have ldo by default, but the nucleo boards can be configured to use SMPS.

erwango (dismissed):

asm5878 (dismissed):

dts/arm/st/wba/stm32wba.dtsi
mathieuchopstm: `st,stm32-pwr` is required for the wake-up pins minidriver:
https://github.com/zephyrproject-rtos/zephyr/blob/f03c105c12c150f914ab0658b4e285dff91463a0/soc/st/stm32/Kconfig#L72-L76

mathieuchopstm:
Rebased + added a `STM32` header, similar to what was found in previous migration guides

erwango (approved):

gautierg-st (approved):

doc/releases/migration-guide-4.4.rst
```diff
@@ -40,6 +40,32 @@ QSPI
...
+=====
+
+* STM32 power supply configuration is now performed using Devicetree properties.
+  New bindings :dtcompatible:`st,stm32h7-pwr`, :dtcompatible:`st,stm32h7rs-pwr`
+  and :dtcompatible:`st,stm32-dualreg-pwr` have been introduced, and all Kconfig
+  symbols related to power supply configuration have been removed:
+
+  * ``CONFIG_POWER_SUPPLY_LDO``
+
+  * ``CONFIG_POWER_SUPPLY_DIRECT_SMPS``,
```
etienne-lms, line 53: Nitpicking; comma here and at line 57 but not for the other configs?
Also, I think you could drop the empty lines between lines 51 and 67.

I wondered if this enumeration should use `:kconfig:option:` prefix. But maybe not applicable since the config symbols have been removed.

etienne-lms (commented):
Non-blocking comments.

doc/releases/migration-guide-4.4.rst
mathieuchopstm: Comma is a good catch - stray remains from rework...

Indeed, `:kconfig:option:` can't be used here since the options no longer exist.

PR #94504, opened 2025-08-14
USB Host: integrate class API [1: types]

josuah:
Downstream:
- https://github.com/zephyrproject-rtos/zephyr/pull/94590

This is a very small, incremental change to help downstream pull requests use a shared base.

I was thinking of doing several such small pull requests with just a few commits, to progressively integrate the code of other pull requests:

- https://github.com/zephyrproject-rtos/zephyr/pull/94266
- https://github.com/zephyrproject-rtos/zephyr/pull/94085

Here is how I tested it:

```
west build -t run -b native_sim samples/subsys/usb/shell/ \
-DEXTRA_CONF_FILE=virtual.conf \
-DDTC_OVERLAY_FILE=virtual.overlay \
-DEXTRA_CONF_FILE=device_and_host_prj.conf
```

Then in another terminal, connect to the console, such as using `picocom /dev/pts/0`.

PR #99173, opened 2025-11-10
subsys: usb: host: Add USB Host Vendor specific serial class support

Girinandha-M:
Dependency:

* https://github.com/zephyrproject-rtos/zephyr/pull/106802

Upstream Reference:

* https://github.com/zephyrproject-rtos/zephyr/pull/94504

subsys: usb: host: add vendor-specific serial class

Add USB vendor-specific serial host class driver that matches
devices by VID, PID.

This is not a standard CDC-ACM driver - it's for vendor-specific
serial devices that use bulk endpoints instead of CDC protocol.

Signed-off-by: Girinandha Manivelpandiyan <girinandha@linumiz.com>

### DIFF
--- a/subsys/usb/host/class/usbh_vendor_serial.c
+++ b/subsys/usb/host/class/usbh_vendor_serial.c
@@ -1,0 +1,553 @@
+/*
+ * Copyright (c) 2026 Linumiz
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/kernel.h>
+#include <zephyr/usb/usbh.h>
+#include <zephyr/usb/usb_ch9.h>
+#include <zephyr/drivers/uart.h>
+#include <zephyr/drivers/usb/uhc.h>
+#include <zephyr/logging/log.h>
+
+#include "usbh_class.h"
+#include "usbh_device.h"
+#include "usbh_desc.h"
+
+LOG_MODULE_REGISTER(usbh_vendor_serial, CONFIG_USBH_VENDOR_SERIAL_LOG_LEVEL);
+
+#define BUF_CURRENT		0
+#define BUF_NEXT		1
+
+struct vendor_serial_data {
+	struct usb_device *udev;
+	const struct device *dev;
+
+	uint8_t bulk_in_ep;
+	uint8_t bulk_out_ep;
+	uint16_t bulk_in_mps;
+
+	uart_callback_t uart_cb;
+	void *uart_cb_data;
+
+	uint8_t *cur_buf;
+	uint8_t *next_buf;
+	size_t cur_len;
+	size_t cur_off;
+	size_t next_len;
+
+	const uint8_t *tx_buf;
+	size_t tx_len;
+
+	struct k_mutex lock;
+};
+
+static void notify_uart(struct vendor_serial_data *data,
+			struct uart_event *evt)
+{
+	if (data->uart_cb != NULL) {
+		data->uart_cb(data->dev, evt, data->uart_cb_data);
+	}
+}
+
+static void evt_rx_rdy(struct vendor_serial_data *data, size_t len)
+{
+	struct uart_event evt = {
+		.type = UART_RX_RDY,
+		.data.rx.buf = data->cur_buf,
+		.data.rx.offset = data->cur_off,
+		.data.rx.len = len,
+	};
+
+	data->cur_off += len;
+	notify_uart(data, &evt);
+}
+
+static void evt_rx_disabled(struct vendor_serial_data *data)
+{
+	struct uart_event evt = {.type = UART_RX_DISABLED};
+
+	data->cur_buf = NULL;
+	data->cur_len = 0;
+	data->cur_off = 0;
+	notify_uart(data, &evt);
+}
+
+static void evt_rx_buf_request(struct vendor_serial_data *data)
+{
+	struct uart_event evt = {.type = UART_RX_BUF_REQUEST};
+
+	notify_uart(data, &evt);
+}
+
+static void evt_rx_buf_release(struct vendor_serial_data *data, int which)
+{
+	struct uart_event evt = {.type = UART_RX_BUF_RELEASED};
+
+	if (which == BUF_NEXT) {
+		if (data->next_buf == NULL) {
+			return;
+		}
+		evt.data.rx_buf.buf = data->next_buf;
+		data->next_buf = NULL;
+		data->next_len = 0;
+	} else {
+		if (data->cur_buf == NULL) {
+			return;
+		}
+		evt.data.rx_buf.buf = data->cur_buf;
+		data->cur_buf = NULL;
+		data->cur_len = 0;
+		data->cur_off = 0;
+	}
+
+	notify_uart(data, &evt);
+}
+
+static int rx_cb(struct usb_device *udev, struct uhc_transfer *xfer);
+static int tx_cb(struct usb_device *udev, struct uhc_transfer *xfer);
+
+static int submit_rx(struct vendor_serial_data *data, const size_t len)
+{
+	struct uhc_transfer *xfer;
+	struct net_buf *nbuf;
+	int ret;
+
+	xfer = usbh_xfer_alloc(data->udev, data->bulk_in_ep, rx_cb, data);
+	if (xfer == NULL) {
+		return -ENOMEM;
+	}
+
+	nbuf = usbh_xfer_buf_alloc(data->udev, len);
+	if (nbuf == NULL) {
+		ret = -ENOMEM;
+		goto free_xfer;
+	}
+
+	ret = usbh_xfer_buf_add(data->udev, xfer, nbuf);
+	if (ret != 0) {
+		goto free_buf;
+	}
+
+	ret = usbh_xfer_enqueue(data->udev, xfer);
+	if (ret != 0) {
+		goto free_buf;
+	}
+
+	return 0;
+
+free_buf:
+	usbh_xfer_buf_free(data->udev, nbuf);
+free_xfer:
+	usbh_xfer_free(data->udev, xfer);
+
+	return ret;
+}
+
+static int submit_tx(struct vendor_serial_data *data,
+		     const uint8_t *const buf, const size_t len)
+{
+	struct uhc_transfer *xfer;
+	struct net_buf *nbuf;
+	int ret;
+
+	xfer = usbh_xfer_alloc(data->udev, data->bulk_out_ep, tx_cb, data);
+	if (xfer == NULL) {
+		return -ENOMEM;
+	}
+
+	nbuf = usbh_xfer_buf_alloc(data->udev, len);
+	if (nbuf == NULL) {
+		ret = -ENOMEM;
+		goto free_xfer;
+	}
+
+	net_buf_add_mem(nbuf, buf, len);
+
+	ret = usbh_xfer_buf_add(data->udev, xfer, nbuf);
+	if (ret != 0) {
+		goto free_buf;
+	}
+
+	ret = usbh_xfer_enqueue(data->udev, xfer);
+	if (ret != 0) {
+		goto free_buf;
+	}
+
+	return 0;
+
+free_buf:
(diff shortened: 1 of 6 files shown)


josuah:
Hello @Girinandha-M, the fixes from https://github.com/zephyrproject-rtos/zephyr/pull/99334 got integrated into https://github.com/zephyrproject-rtos/zephyr/pull/94590.

If you wished to simplify your PR you may be able to switch to https://github.com/zephyrproject-rtos/zephyr/pull/94590.

subsys/usb/host/class/usbh_serial.c
josuah: While this is a simple and efficient API, there is an existing API in Zephyr for serial lines, which for instance USB CDC ACM supports, and which might be needed in the final version of this driver.

@jfischer-no @tmon-nordic 
Should this be handled as a UART async API? As UART IRQ-based API? Custom one like now?

I think the goal is to get modem used, so both would work, I suppose something like this on top of it:
https://github.com/zephyrproject-rtos/zephyr/tree/3cf7b202780d788976e4bf3e46724e63d23f6bcc/drivers/modem

tmon-nordic: Implementing UART API would probably be the easiest solution to get higher-level things working on top of such USB serial port. I am not sure what design sacrifices would have to be made to support UART API though.

Girinandha-M: Thanks for the feedback — agreed on reusing the existing Zephyr serial/UART APIs.

The async UART API was chosen specifically to match modem-style usage (non-blocking TX/RX)

Please let me know if there are concerns with using the async UART API

carlescufi:
@Girinandha-M the dependencies should be merged now. Can you take this out of draft and rebase on top of `main`?

dts/bindings/usb/zephyr,usbh-serial.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2026 Linumiz
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+ USB Host Vendor class serial support.
+
+compatible: "zephyr,usbh-serial"
+
+include: base.yaml
```
jfischer-no: Devicetree should not be used for class driver instantiation.

include/zephyr/usb/class/usb_serial.h
```diff
@@ -0,0 +1,64 @@
+/*
```
jfischer-no: What is the purpose of this header file and why is it exposed to all users?

subsys/usb/host/class/Kconfig.serial
```diff
@@ -0,0 +1,27 @@
+#
+# Copyright (c) 2026 Linumiz
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+config USBH_SERIAL
+	bool "USB Host Vendor-Specific Serial Support"
```
jfischer-no: What vendor? Is it intended to touch this code every time a new "Vendor" needs to be supported? It does not make sense to have this code in the tree parallel to CDC ACM.

jfischer-no (changes requested):

subsys/usb/host/class/Kconfig.serial
josuah: From discussions on Discord, this is the counterpart of this Linux driver:

https://github.com/torvalds/linux/blob/72c395024dac5e215136cbff793455f065603b06/drivers/usb/serial/usb-serial.c#L45-L50

With various vendor-specific extras, some device not having any vendor specifics listed there:

https://github.com/torvalds/linux/blob/72c395024dac5e215136cbff793455f065603b06/drivers/usb/serial/usb-serial-simple.c#L35-L127

jfischer-no: I guess you mean imitation not counterpart. I do not think it is something to imitate or use as reference.

Girinandha-M: @jfischer-no @josuah Thanks for your feedback.

My goal with this change was to support the Quectel EG916Q modem, which exposes a vendor-specific bulk interface used for AT/PPP communication.

I initially looked at how Linux handles similar devices via the option driver and attempted a more generic “vendor serial” class approach to potentially support multiple modems in the future.

However, I understand the concern about introducing a parallel framework to CDC ACM and potentially imitating Linux’s usb-serial model.

For Zephyr, would it be preferable to:

1. Implement this as a device-specific driver (matching VID/PID) under a modem driver instead of a generic vendor class?
2. Integrate this functionality differently within the existing USB host class structure?
or is there any other different approach should i look into ?

josuah: Link for convenience: [`./drivers/usb/serial/option.c`](https://github.com/torvalds/linux/blob/ca4ee40bf13dbd3a4be3b40a00c33a1153d487e5/drivers/usb/serial/option.c)

Girinandha-M:
Rebased on main, reworked based on feedback:

- Dropped DT bindings, using DEVICE_DEFINE + LISTIFY (same pattern as UVC)
- No public header, everything internal
- UART async API for modem/networking subsystem compatibility
- VID/PID filter matching with extensible quirk table
Tested on imxrt1064_evk + Quectel EG916Q: enumeration, class probe, AT command TX all working

Found and fixed a upstream HS enumeration bugs in usbh_device.c during testing (separate commit).

subsys/usb/host/class/usbh_vendor_serial.c
```diff
@@ -0,0 +1,580 @@
...
+#define QUECTEL_VENDOR_ID	0x2c7c
+#define QUECTEL_EG916Q_PID	0x6007
+
+#define BUF_CURRENT		0
+#define BUF_NEXT		1
+
+struct vendor_serial_quirk {
+	uint16_t vid;
+	uint16_t pid;
+	uint8_t at_iface;
```
josuah: Is it really AT interface? This driver is expected to be a generic serial, so network-specific vocabulary might be best taken away from the shared bits?

subsys/usb/host/class/usbh_vendor_serial.c
```diff
@@ -0,0 +1,580 @@
...
+
+	memcpy(nbuf->data, buf, len);
+	net_buf_add(nbuf, len);
+
+	ret = usbh_xfer_buf_add(data->udev, xfer, nbuf);
+	if (ret) {
+		usbh_xfer_buf_free(data->udev, nbuf);
+		usbh_xfer_free(data->udev, xfer);
+		return ret;
+	}
```
josuah, line 169: Grouping several feedback points by taking inspiration of other code in USB codebase, that can be propagated elsewhere as well:

```suggestion
static int submit_tx(cons struct device *const dev,
		     const uint8_t *const buf, const size_t len)
{
	struct vendor_serial_data *const data = dev->data;
	struct uhc_transfer *xfer;
	struct net_buf *nbuf;
	int ret;

	xfer = usbh_xfer_alloc(data->udev, data->bulk_out_ep, tx_cb, buf);
	if (xfer == NULL) {
		return -ENOMEM;
	}

	nbuf = usbh_xfer_buf_alloc(data->udev, len);
	if (nbuf == NULL) {
		ret = -ENOMEM;
		goto some_label;
	}

	net_buf_add_mem(nbuf, buf, len);

	ret = usbh_xfer_buf_add(data->udev, xfer, nbuf);
	if (ret != 0) {
		goto some_label;
	}
```

- [`net_buf`](https://docs.zephyrproject.org/latest/doxygen/html/group__net__buf.html) typically has utilities for most data movement
- const everywhere possible
- use `const struct device *dev` and `struct usbh_class_data *c_data` as first parameter everywhere, and dispatch the `data` etc. parameters when needed (using `*const` too), possibly because it helps
- use explicit `if (... == NULL)` and `if (... == 0)` to help with MISRA
- use centralized exit using `goto` for repeated error handling

This is just my interpretation of other code, hopefully it is accurate and you are not asked to roll back any of it... feel free to discuss any point too.

subsys/usb/host/class/usbh_vendor_serial.c
```diff
@@ -0,0 +1,580 @@
...
+};
+
+static const struct vendor_serial_quirk quirk_table[] = {
+	{
+		.vid = QUECTEL_VENDOR_ID,
+		.pid = QUECTEL_EG916Q_PID,
+		.at_iface = 2,
+		.desc = "Quectel EG916Q-GL",
+	},
+};
```
josuah: Did you have to skip the first interface? Would an heuristic that seeks the first interface with a pair of BULK endpoints work?

`.desc` sounds great for debugging, but risks to grow a huge table of strings as more devices are added, and bloat flash usage.

How about just logging the VID:PID? Impacts debuggability but still reasonable.

josuah (commented):
Here is some very incomplete review, in the meantime that a more complete review can be done.

subsys/usb/host/class/usbh_vendor_serial.c
Girinandha-M: The EG916Q exposes 4 vendor-class interfaces (0-3), all with identical bInterfaceClass=0xff, bInterfaceSubClass=0x00, bInterfaceProtocol=0x00. 
Interface 0 is diagnostics, 1 is NMEA, 2 is AT, 3 is PPP, but descriptors are indistinguishable. A heuristic picking the first bulk pair would claim interface 0 (diagnostics), which is wrong.

The current design uses a three-way probe:
   - Quirk match (VID+PID+iface) - claim specific interface.
   - Known device, unlisted interface - skip (prevents heuristic from grabbing the wrong port).
   - Unknown device - heuristic, first bulk pair selected.

This keeps the quirk table small (only devices with ambiguous interfaces need entries) while unknown single-interface vendor devices work with zero configuration.

subsys/usb/host/usbh_device.c
```diff
@@ -561,31 +561,31 @@ int usbh_device_init(struct usb_device *const udev)
 		goto error;
 	}
 
-	err = usbh_req_desc_dev(udev, sizeof(udev->dev_desc), &udev->dev_desc);
+	err = alloc_device_address(udev, &new_addr);
```
tmon-nordic, line 575: While this is OK from the interoperability point of view (allowing to work with broken devices), what exactly does "matching USB 2.0 enumeration sequence" refer to?

tmon-nordic (commented):
~~I think the first three commits are worthwhile to have in main even before there is agreement on how to proceed with "usb: host: add USB Serial support". Can you move them to separate PR?~~ Ah you already posted them separately in #106802

subsys/usb/host/class/usbh_vendor_serial.c
josuah: > 0 is diagnostics, 1 is NMEA, 2 is AT, 3 is PPP

So there should be 4 serial interfaces rather than just one?

As this is not described as a driver for AT protocol anywhere.

It might be an user choice to decide which interface to access and for which purpose?

Someone might want to access the other kinds of command interfaces some day.

This means there would need one `dev` per serial line, each with its own serial `dev->api`.

So far, `const struct device *` for USB hosts are statically allocated in advance.

This suggests you would need another LISTIFY to generate the `const struct device *`.

In addition to the existing LISTIFY, which would only generate the `struct usbh_class_data *c_data`.

And then some dynamic logic would allow to map the interfaces to use depending on each device. Such as some kind of filtering. Maybe giving a flag to each interface (i.e. AT, NMEA, PPP, DEBUG), and the user could have a callback to take a decision of skipping or binding a particular endpoint pair?

The goal being removing the policy out of the drivers, but instead just fully document each hardware connected and expose the information to users.

danielepa-telit:
Hey, thanks for this PR: it's very interesting and it would also be useful for supporting Telit modems and, in general, Qualcomm based USB modems.

Hopefully, in the near future I will give it a try.

josuah:
Removing "stale" label as it seems to just wait review

Girinandha-M:
@jfischer-no @josuah @tmon-nordic  Ping for Review. Thanks

Girinandha-M:
@jfischer-no @josuah, ping for reivew.

PR #99185, opened 2025-11-11
drivers: input: vs1838b: Add noise filter for IR signals

atakandemirtas:
Adds logic to shift out the first edge if the initial three edges do not match a leading burst, improving noise resilience in the IR input signal processing.

### DIFF
--- a/drivers/input/input_vs1838b.c
+++ b/drivers/input/input_vs1838b.c
@@ -315,6 +315,15 @@ static void vs1838b_input_callback(struct device const *dev, struct gpio_callbac
 		data->edges_ticks[data->edges_count++] = tick;
 	}
 
+	/* If the first 3 edges do not match a leading burst,
+	 * shift left the edges_ticks to get rid of leading noises.
+	 */
+	if ((data->edges_count == 3) && !detect_leading_burst(data->edges_ticks)) {
+		data->edges_ticks[0] = data->edges_ticks[1];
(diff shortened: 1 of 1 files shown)


fabiobaltieri:
@atakandemirtas note that this also fails compliance check, those needs to be fixed https://github.com/zephyrproject-rtos/zephyr/actions/runs/19257472142/job/55065977108?pr=99185

bastienjauny:
The change looks good, the only issue is the compliance checks, as @fabiobaltieri already pointed out. Looks like it's just some trailing whitespaces in the comment block.

fabiobaltieri (approved):

pdgendt (approved):

drivers/input/input_vs1838b.c
```diff
@@ -315,6 +315,15 @@ static void vs1838b_input_callback(struct device const *dev, struct gpio_callbac
 		data->edges_ticks[data->edges_count++] = tick;
 	}
 
+	/* If the first 3 edges do not match a leading burst,
+	 * shift left the edges_ticks to get rid of leading noises.
+	 */
+	if ((data->edges_count == 3) && !detect_leading_burst(data->edges_ticks)) {
```
faxe1008, line 321: Maybe replace 3 with `(NEC_LEAD_PULSE_EDGE_WIDTH + 1)` to make clear where the 3 comes from.

faxe1008 (approved):

PR #99188, opened 2025-11-11
cmake: modules: version: Check that required fields are present

nordicjm:
Will throw an error showing what the issue with the VERSION file is if a required field is missing


cmake/modules/version.cmake
```diff
@@ -51,33 +51,54 @@ foreach(type file IN ZIP_LISTS VERSION_TYPE VERSION_FILE)
   file(READ ${file} ver)
   set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS ${file})
 
+  set(${type}_VERSION_MAJOR_LENGTH 0)
+  set(${type}_VERSION_MINOR_LENGTH 0)
+  set(${type}_PATCHLEVEL_LENGTH 0)
+  set(${type}_VERSION_TWEAK_LENGTH 0)
```
tejlmand: why this complexity (also code below) ?

```suggestion
```

cmake/modules/version.cmake
```diff
@@ -51,33 +51,54 @@ foreach(type file IN ZIP_LISTS VERSION_TYPE VERSION_FILE)
...
+    string(LENGTH ${${type}_VERSION_TWEAK} ${type}_VERSION_TWEAK_LENGTH)
+  endif()
+
   string(REGEX MATCH "EXTRAVERSION = ([a-z0-9\.\-]*)" _ ${ver})
   set(${type}_VERSION_EXTRA ${CMAKE_MATCH_1})
 
   # Validate all version fields fit in a single byte
-  if(${type}_VERSION_MAJOR GREATER 255)
-    message(FATAL_ERROR "VERSION_MAJOR must be in the range 0-255 (Current ${${type}_VERSION_MAJOR})")
+  if(${type}_VERSION_MAJOR_LENGTH EQUAL 0 OR ${type}_VERSION_MAJOR GREATER 255)
```
tejlmand: If field is not defined, then `CMAKE_MATCH_1` will be empty and thus the `<type>_VERSION_x` is not defined.

So this should be sufficient:
```suggestion
  if(NOT DEFINED ${type}_VERSION_MAJOR OR ${type}_VERSION_MAJOR GREATER 255)
```

and avoid all the `LENGTH` handling above.

Comment applies to below code lines also.

cmake/modules/version.cmake
```diff
@@ -51,33 +51,54 @@ foreach(type file IN ZIP_LISTS VERSION_TYPE VERSION_FILE)
...
+  set(${type}_PATCHLEVEL_LENGTH 0)
+  set(${type}_VERSION_TWEAK_LENGTH 0)
+
   string(REGEX MATCH "VERSION_MAJOR = ([0-9]*)" _ ${ver})
   set(${type}_VERSION_MAJOR ${CMAKE_MATCH_1})
 
+  if(DEFINED ${type}_VERSION_MAJOR)
+    string(LENGTH ${${type}_VERSION_MAJOR} ${type}_VERSION_MAJOR_LENGTH)
+  endif()
+
```
tejlmand:
```suggestion
```

tejlmand (changes requested):

cmake/modules/version.cmake
nordicjm: Fixed

cmake/modules/version.cmake
nordicjm: Fixed

cmake/modules/version.cmake
nordicjm: Fixed

tejlmand (approved):

pdgendt (approved):

PR #99199, opened 2025-11-11
boards: nordic: nrf54h20dk: Add reserved partition for IronSide SE

57300:
Add a self-documenting devicetree node in the default memory map file, to communicate how much memory is reserved for this firmware.

### DIFF
--- a/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
+++ b/boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
@@ -131,6 +131,10 @@
 		#size-cells = <1>;
 		ranges;
 
+		ironside_se_reserved: partition@0 {
+			reg = <0x0 DT_SIZE_K(192)>;
+		};
+
 		cpuapp_boot_partition: partition@30000 {
 			compatible = "zephyr,mapped-partition";
 			reg = <0x30000 DT_SIZE_K(64)>;


boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
```diff
@@ -131,6 +131,10 @@
 		#address-cells = <1>;
 		#size-cells = <1>;
 
+		ironside_se_reserved: partition@0 {
+			reg = <0x0 DT_SIZE_K(192)>;
+		};
```
bjarki-andreasen, line 136: If its reserved, it should be in `reserved-memory` node, not in mram partitions which are expected to be used by the application.

bjarki-andreasen (changes requested):

boards/nordic/nrf54h20dk/nrf54h20dk_nrf54h20-memory_map.dtsi
57300: Hmm, `/reserved-memory` also contains nodes used by applications on this platform, but perhaps we haven't been using that node as intended. DTSpec is not clear on this IMO.

Maybe I should refactor the memory map again before applying your suggestion. It's possible now that we've mostly gotten rid of the legacy SDFW-specific abstractions that couldn't fit with common devicetree practice.

Alternatively, I could replace this node with a comment. I think placing this information close to the MRAM nodes would aid readability anyway. The point of this PR is to make this memory block visible to customers.

bjarki-andreasen: Looking past `reserved-memory`, if  we add a comment stating that the `ironside_se_reserved` partition is reserved and can not be changed or moved, I can agree to that, I would even remove the nodelabel, it will not be used or referenced anyway (which is implicit with the `reserved-memory` subnodes)

nordicjm: Reserved memory is for special hardware features, that would be the likes of DMA for RAM, not NVM partitions, so declaring as a partition is correct

bjarki-andreasen: > Reserved memory is for special hardware features, that would be the likes of DMA for RAM, not NVM partitions, so declaring as a partition is correct

I would describe secure firmware as a special hardware feature in this case, devicetree spec states "The operating system shall exclude reserved memory from normal usage", which seems very fitting.

I don't have a super strong opinion on it, but that is my thinking.

bjarki-andreasen (dismissed):

masz-nordic (dismissed):

PR #99203, opened 2025-11-11
boards: arduino: giga_r1: Fix partition size.

iabdalkader:
With the current config, a debug loader overwrites the sketch.

The correct partition layout is (offset, size):
- 0x00000 - 0x40000 (256KB): Bootloader
- 0x40000 - 0xa0000 (640KB): llext loader
- 0xe0000 - 0x20000 (128KB): sketch


pillo79 (dismissed):
That is indeed correct. We could say it's a feature though! 😅

iabdalkader:
This patch is no longer needed, we're going to keep the sizes as they are, and carve out sketch space in overlays when it's needed.

PR #97448, opened 2025-10-13
STM32: Change/fix memory sizes in stm32h7rs device tree files

TD-JBL:
## Changes

1. SOC's DT: The h7rs option bytes default setting configures the AXI SRAM1-4 as one contiguous memory, sram0 can therefore be increased from 128k to 456k (3x128k + 72k).
2. Set the ram size in boards/st/nucleo_h7s3l8 and stm32h7s78_dk .yaml files to 456k. 
3. Board Nucleo_H7S3L8: Setting the correct size of ext_memory node to 32MB, as the board uses a Macronix MX25UW25645GXDI00 with 256MBits. 

## Testing status
* Access to whole 456k of sram0 checked with custom app.

PR #99206, opened 2025-11-11
boards: st: stm32h7s78_dk: remove unnecessary CONFIG_SYS_HEAP_AUTO in both defconfigs

bklaric1:
# Changes
Remove the CONFIG_SYS_HEAP_AUTO=y config from both defconfig files, since it is no longer needed.

It was necessary when sram0 was smaller than 262136 bytes, as stated [here](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_STM32_APP_IN_EXT_FLASH).  
But since #97448 increased the sram0 size to 456kB, this option is selected by default.

# Tests
Tested as following:
1. Commented out `CONFIG_SYS_HEAP_AUTO=y` in `stm32h7s78_dk_stm32h7s7xx_ext_flash_app_defconfig`
2. Built and flashed a sample (dumb_http_server, since I was already testing something else with it)
3. Checked `\build\zephyr\.config` and confirmed that CONFIG_SYS_HEAP_AUTO=y is enabled by default

### DIFF
--- a/boards/st/stm32h7s78_dk/stm32h7s78_dk_defconfig
+++ b/boards/st/stm32h7s78_dk/stm32h7s78_dk_defconfig
@@ -18,10 +18,3 @@ CONFIG_UART_CONSOLE=y
 
 # enable GPIO
 CONFIG_GPIO=y
-
-# SRAM0 being small (128KB), SYS_HEAP_SMALL_ONLY
-# is automatically selected, leading to not being
-# able to tackle HEAP on the large PSRAM. Force it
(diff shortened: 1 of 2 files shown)


erwango (approved):
Thanks.
Btw, this symbol is enabled automatically when needed, so hard setting it wasn't required anyway:
```
	default SYS_HEAP_SMALL_ONLY if (SRAM_SIZE <= 256)
```

mathieuchopstm (approved):

PR #98646, opened 2025-10-31
drivers: gnss: allow same callback

PedroSAndre:
[drivers: gnss: allow same callback](https://github.com/zephyrproject-rtos/zephyr/commit/4a285165d58c192017499836befb211182d32282) 

The current implementation of GNSS_DATA_CALLBACK_DEFINE and
GNSS_SATELLITES_CALLBACK_DEFINE only use the provided `_callback` to
uniquely identify the entry that will be created. This implies that the
same callback cannot be registered twice, even if the device it is being
attached to is different.

In order to allow the same callback to be registered several times, the
ideal would be to also use `_dev` in combination with `_callback` to
uniquely identify the generated entry. But this would require getting the
node_id of the device, hence breaking the current API. In order to avoid
that, use __COUNTER__ to avoid collisions.

[gnss: rtk: fix multiple instances of same device](https://github.com/zephyrproject-rtos/zephyr/commit/7eea2f77b618f455cdffeaf64157ed4f46926ecb) 

When multiple instances of the same type of RTK enabled GNSS device are
instantiated on the same board, there will be a collision with their RTK
data callbacks. The u-blox F9P module is a clear example of that, where the
following compilation error is reported:

     error: redefinition of '_gnss_rtk_data_callback__f9p_rtk_data_cb'

Similar to the solution used at https://github.com/zephyrproject-rtos/zephyr/commit/d36a0fcb9503f6e79a8844cfd12daeb225d24a5a (drivers: gnss: allow same
callback), use `__COUNTER__` instead of `_callback` in the generated entry
to avoid collisions.

PR #99208, opened 2025-11-11
drivers: gnss: introduce node based callback macros

PedroSAndre:
[drivers: gnss: introduce node based callback macros](https://github.com/zephyrproject-rtos/zephyr/commit/11472e7918d07840f0570d314288f473a60dd85e) 

The current GNSS_DATA_CALLBACK_DEFINE, GNSS_SATELLITES_CALLBACK_DEFINE and GNSS_RTK_DATA_CALLBACK_DEFINE macros only utilize `_callback` when naming the respective callback. This means that if the same callback function is used more than once, there will be naming conflicts.

In order to allow the same callback function to be used several times for specific gnss devices, introduce GNSS_DT_DATA_CALLBACK_DEFINE, GNSS_DT_SATELLITES_CALLBACK_DEFINE and GNSS_DT_RTK_DATA_CALLBACK_DEFINE macros that take a device tree node identifier instead of a device reference. This makes it possible to name the callback using both `node_id` and `_callback`. Such will uniquely identify any combination of gnss device and callback.

[drivers: gnss: f9p: fix multiple rtk enabled instances](https://github.com/zephyrproject-rtos/zephyr/commit/5ed368722cbae283b78ed7eaf642cda4a7e1f323) 

The U-blox F9P driver uses the GNSS_RTK_DATA_CALLBACK_DEFINE macro to register a callback for when RTK data is available. This is an issue when multiple instances of the F9P are available and RTK is enabled since this macro leads to a naming collision. The following compilation error is reported:

     error: redefinition of '_gnss_rtk_data_callback__f9p_rtk_data_cb'

This is because GNSS_RTK_DATA_CALLBACK_DEFINE only uses `_callback` to identify the callback being registered. As a fix, use the recently introduced GNSS_DT_RTK_DATA_CALLBACK_DEFINE instead, which takes into account the `node_id` while naming the callback.

This PR is a continuation of #98646.

Signed-off-by: Pedro André <pedro@sentrytechnologies.ai>

### DIFF
--- a/include/zephyr/drivers/gnss.h
+++ b/include/zephyr/drivers/gnss.h
@@ -430,8 +430,17 @@ static inline int z_impl_gnss_get_latest_timepulse(const struct device *dev,
 		.dev = _dev,                                                                    \
 		.callback = _callback,                                                          \
 	}
+
+#define GNSS_DT_DATA_CALLBACK_DEFINE(_node_id, _callback)                                          \
+	static const STRUCT_SECTION_ITERABLE(                                                      \
+		gnss_data_callback,                                                                \
+		_CONCAT_4(_gnss_data_callback_, DT_DEP_ORD(_node_id), _, _callback)) = {           \
+		.dev = DEVICE_DT_GET(_node_id),                                                    \
+		.callback = _callback,                                                             \
+	}
 #else
 #define GNSS_DATA_CALLBACK_DEFINE(_dev, _callback)
+#define GNSS_DT_DATA_CALLBACK_DEFINE(_node_id, _callback)
 #endif
 
 /**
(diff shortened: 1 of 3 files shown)


include/zephyr/drivers/gnss.h
```diff
@@ -430,8 +430,17 @@ static inline int z_impl_gnss_get_latest_timepulse(const struct device *dev,
 		.dev = _dev,                                                                    \
 		.callback = _callback,                                                          \
 	}
+
+#define GNSS_DT_DATA_CALLBACK_DEFINE(node_id, _callback)                                           \
+	static const STRUCT_SECTION_ITERABLE(                                                      \
+		gnss_data_callback,                                                                \
+		_CONCAT(_CONCAT(_gnss_data_callback_, node_id), _CONCAT(_, _callback))) = {        \
```
bjarki-andreasen: Use DT_DEP_ORD() to get a unique integer from the node_id, this produces shorter names as the node_id can be very long:
```
_CONCAT(_CONCAT(_gnss_data_callback_, DT_DEP_ORD(node_id)), _CONCAT(_, _callback)))
```
Then use `_CONCAT_4()` instead of multiple `_CONCAT`
```
_CONCAT_4(_gnss_data_callback_, DT_DEP_ORD(node_id), _, _callback)
```
Lastly, using `_node_id` is a bit more consitent with `_callback`, though this one is entirely up to preference, many DT APIs take `node_id` without the `_`

bjarki-andreasen (commented):
Really cool

bjarki-andreasen (approved):
Spotless

fabiobaltieri (approved):

PR #99210, opened 2025-11-11
boards: arduino: add ADC support for the Nano Matter

silabs-bozont:
This PR adds ADC support for the Arduino Nano Matter on its analog pins A0-A7.

### DIFF
--- a/boards/arduino/nano_matter/arduino_nano_matter.dts
+++ b/boards/arduino/nano_matter/arduino_nano_matter.dts
@@ -8,6 +8,7 @@
 /dts-v1/;
 #include <silabs/xg24/mgm240sd22vna.dtsi>
 #include <zephyr/dt-bindings/input/input-event-codes.h>
+#include <dt-bindings/adc/silabs-adc.h>
 #include <zephyr/dt-bindings/pwm/pwm.h>
 #include "arduino_nano_matter-pinctrl.dtsi"
 #include "arduino_nano_matter_connector.dtsi"
@@ -36,6 +37,7 @@
 		pwm-led2 = &blue_pwm_led;
 		sw0 = &button0;
 		watchdog0 = &wdog0;
+		adc0 = &adc0;
 	};
 
 	leds {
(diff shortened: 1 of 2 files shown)


boards/arduino/nano_matter/arduino_nano_matter.dts
```diff
@@ -84,6 +86,17 @@
...
+	zephyr,user {
+		io-channels = <&adc0 0>,
+			      <&adc0 1>,
+			      <&adc0 2>,
+			      <&adc0 3>,
+			      <&adc0 4>,
+			      <&adc0 5>,
+			      <&adc0 6>,
+			      <&adc0 7>;
+	};
```
pillo79, line 99: I guess this is Arduino-specific. It should be moved to the variant overlay in https://github.com/arduino/ArduinoCore-zephyr/pull/255 (or a separate PR).

pillo79 (changes requested):
LGTM and thanks, just a minor fix request!

boards/arduino/nano_matter/arduino_nano_matter.dts
silabs-bozont: I've included it here so users can get an enumeration of io-channels even outside of the Arduino Core, because the hardware is the same. Also other Silabs boards include these like [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/silabs/radio_boards/xg26/xg26_rb41xxa.dtsi#L139). If you have a strong opinion against including this here, I'm open to debate :)

Edit: I don't think `io-channels` is Arduino specific - it has many occurrences in this repo too.

pillo79: Oh, I thought having board-defined `zephyr,user` was _verboten_ in Zephyr. Nevermind then! :slightly_smiling_face:

pillo79 (approved):

facchinm (approved):

pennam (approved):

iabdalkader (approved):

PR #99213, opened 2025-11-11
usb: device_next: cdc_acm: Use a separate buffer pool for OUT

ndreys:
CDC ACM code will try to eagerly consume buffers from the buffer pool as long as there some data in it's TX FIFO. Since both IN and OUT transfers share the same buffer pool this make the code vulnerable to the "noisy neighbor" problem, where a influx of TX/IN traffic can completely deplete the buffer pool and cause a buffer allocation failure in cdc_acm_rx_fifo_handler(). Since there is no retry mechanism on buffer allocation failure there we end up stuck with USB controller NAK-ing every OUT transfer.

To avoid this problem, let's just split the buffer pools into IN/OUT. This way we know that an allocation failure in cdc_acm_rx_fifo_handler() implies that there are queued IN transfers which will re-arm the DMA in USB peripheral once they complete.

cc: @jfischer-no

### DIFF
--- a/subsys/usb/device_next/class/usbd_cdc_acm.c
+++ b/subsys/usb/device_next/class/usbd_cdc_acm.c
@@ -128,16 +128,28 @@ struct cdc_acm_uart_data {
 };
 
 static void cdc_acm_irq_rx_enable(const struct device *dev);
+static uint8_t cdc_acm_get_bulk_in(struct usbd_class_data *const c_data);
 
-#if CONFIG_USBD_CDC_ACM_BUF_POOL
 UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
-		    DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) * 2,
+		    DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) *
+		    (1 + IS_ENABLED(CONFIG_USBD_CDC_ACM_BUF_POOL)),
 		    USBD_MAX_BULK_MPS, sizeof(struct udc_buf_info), NULL);
 
 static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
 					 const uint8_t ep)
 {
-	ARG_UNUSED(c_data);
+	struct net_buf_pool *pool;
+
+	if (!IS_ENABLED(CONFIG_USBD_CDC_ACM_BUF_POOL) &&
+	    ep == cdc_acm_get_bulk_in(c_data)) {
+		/*
+		 * The required buffer is 64 bytes per instance on a
+		 * full-speed device. Use common (UDC) buffer, as this
+		 * results in a smaller footprint.
+		 */
+		return usbd_ep_buf_alloc(c_data, ep, USBD_MAX_BULK_MPS);
+	}
+
 	struct net_buf *buf = NULL;
 	struct udc_buf_info *bi;
 
@@ -151,17 +163,6 @@ static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
 
 	return buf;
 }
-#else
-/*
- * The required buffer is 128 bytes per instance on a full-speed device. Use
- * common (UDC) buffer, as this results in a smaller footprint.
- */
-static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
-{
-	return usbd_ep_buf_alloc(c_data, ep, USBD_MAX_BULK_MPS);
-}
-#endif /* CONFIG_USBD_CDC_ACM_BUF_POOL */
 
 #if CONFIG_USBD_CDC_ACM_WORKQUEUE
 static struct k_work_q cdc_acm_work_q;
@@ -710,6 +711,11 @@ static void cdc_acm_rx_fifo_handler(struct k_work *work)
 	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data));
 	if (buf == NULL) {
 	        atomic_clear_bit(&data->state, CDC_ACM_RX_FIFO_BUSY);
+		/*
+		 * If we fail to allocate buffer here that means there
+		 * are enqueued IN netbufs that will continue the RX
+		 * dataflow, so it is OK for us to bail out.
+		 */
 		return;
 	}
 
--- a/subsys/usb/device_next/class/usbd_cdc_acm.c
+++ b/subsys/usb/device_next/class/usbd_cdc_acm.c
@@ -709,6 +709,7 @@ static void cdc_acm_rx_fifo_handler(struct k_work *work)
 
 	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data));
 	if (buf == NULL) {
+	        atomic_clear_bit(&data->state, CDC_ACM_RX_FIFO_BUSY);
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -128,20 +128,19 @@ struct cdc_acm_uart_data {
...
-#if CONFIG_USBD_CDC_ACM_BUF_POOL
-UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
+UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool_out,
 		    DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) * 2,
 		    USBD_MAX_BULK_MPS, sizeof(struct udc_buf_info), NULL);
 
-static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
+static struct net_buf *__cdc_acm_buf_alloc(struct net_buf_pool *pool,
+					   const uint8_t ep)
```
josuah: `__` as a prefix (two underscores) is often used by toolchains and all other functions use `cdc_acm_` as prefix.

How about `cdc_acm_buf_alloc_from_pool()` or equivalent of your choosing?

subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -709,6 +733,11 @@ static void cdc_acm_rx_fifo_handler(struct k_work *work)
 
 	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data));
 	if (buf == NULL) {
+		/*
+		 * If we fail to allocate buffer here that means there
+		 * are enqueued IN netbufs that will continue the RX
+		 * dataflow, so it is OK for us to bail out.
+		 */
```
tmon-nordic, line 718: There are two buffers per instance while the class only allocates one at a time per direction. There is no double buffering. How is it possible to observe the issue you mention in commit message?

subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -159,7 +179,11 @@ static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
 static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
 					 const uint8_t ep)
 {
-	return usbd_ep_buf_alloc(c_data, ep, USBD_MAX_BULK_MPS);
+	if (ep == cdc_acm_get_bulk_out(c_data)) {
+		return __cdc_acm_buf_alloc(&cdc_acm_ep_pool_out, ep);
```
josuah: When `CONFIG_USBD_CDC_ACM_BUF_POOL` is disabled, then `cdc_acm_ep_pool_out` is still used.
Is this the intended behavior?

Curious why this function needed to change as opposed in the `#else` part of that ifdef as well.

josuah (commented):
Some feedback about the implementation.

However, it might make more sense to complete the discussion on this thread first.

- https://github.com/zephyrproject-rtos/zephyr/pull/99213/files#r2517323683

subsys/usb/device_next/class/usbd_cdc_acm.c
ndreys: Hmm, I originally developed this patch on a Zephyr version that doesn't have `CDC_ACM_TX_FIFO_BUSY` and has an eager FIFO filling behavior via `cdc_acm_irq_cb_handler` rescheduling itself in
```
	if (data->rx_fifo.altered) {
		LOG_DBG("rx fifo altered, submit work");
		cdc_acm_work_submit(&data->rx_fifo_work);
	}

	if (data->tx_fifo.altered) {
		LOG_DBG("tx fifo altered, submit work");
		cdc_acm_work_schedule(&data->tx_fifo_work, K_NO_WAIT);
	}

	if (atomic_test_bit(&data->state, CDC_ACM_IRQ_RX_ENABLED) &&
	    !ring_buf_is_empty(data->rx_fifo.rb)) {
		LOG_DBG("rx irq pending, submit irq_cb_work");
		cdc_acm_work_submit(&data->irq_cb_work);
	}

	if (atomic_test_bit(&data->state, CDC_ACM_IRQ_TX_ENABLED) &&
	    ring_buf_space_get(data->tx_fifo.rb)) {
		LOG_DBG("tx irq pending, submit irq_cb_work");
		cdc_acm_work_submit(&data->irq_cb_work);
	}
```
so my original issue probably got fixed by 5cbd1f6c39002410a814700cb75dcce189bb5295 ("usb: device_next: cdc_acm: Prevent polling for buffer in TX"). 

With that being said, it seems like the latest version of `main` will still be exposed to the same kind of problem with `CONFIG_USBD_CDC_ACM_BUF_POOL=n`, right? The pool is now shared with the rest of the USB stack, so per instance allocation gating doesn't seem to give us any strong guarantees, no?


Tangentially related, but while I have you on the line:
```
	if (atomic_test_and_set_bit(&data->state, CDC_ACM_RX_FIFO_BUSY)) {
		LOG_WRN("RX transfer already in progress");
		return;
	}

	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data));
	if (buf == NULL) {
		return;
	}

	/* Shrink the buffer size if operating on a full speed bus */
	buf->size = MIN(cdc_acm_get_bulk_mps(c_data), buf->size);

	ret = usbd_ep_enqueue(c_data, buf);
	if (ret) {
		LOG_ERR("Failed to enqueue net_buf for 0x%02x",
			cdc_acm_get_bulk_out(c_data));
		net_buf_unref(buf);
	}
```

should we not be clearing `CDC_ACM_RX_FIFO_BUSY` in case of failures here?

subsys/usb/device_next/class/usbd_cdc_acm.c
ndreys: Sure, trivial to rename

subsys/usb/device_next/class/usbd_cdc_acm.c
ndreys: > When CONFIG_USBD_CDC_ACM_BUF_POOL is disabled, then cdc_acm_ep_pool_out is still used.
Is this the intended behavior?

Yep.

> Curious why this function needed to change as opposed in the #else part of that ifdef as well.

You'd have the same problem with a global pool as well, I think. You can have some other consumer of the buffers from that pool, not necessarily even just a TX direction of this same instance, take up all available resources leaving without a buffer to re-arm USB OUT DMA at a critical time.

subsys/usb/device_next/class/usbd_cdc_acm.c
tmon-nordic: With `CONFIG_USBD_CDC_ACM_BUF_POOL=n` there is a potential problem if `CONFIG_UDC_BUF_COUNT` is low, especially if there are also other classes.

`CDC_ACM_RX_FIFO_BUSY` should be cleared on failures as otherwise CDC ACM instance won't be able to receive anything past first error.

ndreys: OK, I reworked the branch to include the fix for clearing RX busy bit and adjusted the code to fix only the case of the shared UDC buffer. With the latter in place I don't know if that whole optimization is really worth it anymore (it does save 64 bytes per instance, but :shrug: )  @jfischer-no do you have an opinion on this? Maybe we can just revert the commit that added using the UDC buffer as a simpler solution?

subsys/usb/device_next/class/usbd_cdc_acm.c
josuah: > You'd have the same problem with a global pool as well, I think.

I assume that users disable the local pool (`CONFIG_USBD_CDC_ACM_BUF_POOL=n`) to reduce the memory footprint if i.e. they only need for occasional debug logs.

```diff
-#if CONFIG_USBD_CDC_ACM_BUF_POOL
 UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
```

With this change, Kconfig promises "we won't define a local buffer pool" and the next line unconditionally defines a buffer pool.

Or maybe there's something I missed...

Thanks.

PR #99225, opened 2025-11-12
arch: indirect executable RAM symbol

JordanYates:
Instead of marking RAM regions as executable when `CONFIG_XIP=n`, add a dedicated symbol to control the behaviour. XIP is the most obvious case that requires this, but LLEXT for example can also load code into RAM that is then executed.

The addition of `XIP_NOT` forces `EXECUTABLE_RAM` when `XIP` is disabled in order to prevent users from accidently turning it off.

Intended to solve usage faults when using LLEXT on platforms with an MPU or MMU, since currently any attempt to execute a loaded function will fault. Whether this is the right place or not to solve the problem I don't know, but it does solve it for `mps2/an385`.

### DIFF
--- a/arch/Kconfig
+++ b/arch/Kconfig
@@ -241,6 +241,20 @@ config XIP
 	  supply a linker command file when building your image. Enabling this
 	  option increases both the code and data footprint of the image.
 
+config XIP_NOT
+	bool
+	default y
+	depends on !XIP
+	select EXECUTABLE_RAM if (MMU || MPU)
+	help
+	  Helper symbol that forces EXECUTABLE_RAM if !XIP.
+
+config EXECUTABLE_RAM
+	bool "Configure RAM to be executable"
+	depends on (MMU || MPU)
+	default y if LLEXT
+	help
+	  Configure RAM to be executable. Changes the attributes of REGION_RAM_ATTR.
 
 if ARC || ARM || ARM64 || X86 || RISCV || RX || ARCH_POSIX
 
--- a/include/zephyr/arch/arm/mpu/arm_mpu_v7m.h
+++ b/include/zephyr/arch/arm/mpu/arm_mpu_v7m.h
@@ -111,13 +111,12 @@
 
 /* Some helper defines for common regions */
 
-/* On Cortex-M, we can only set the XN bit when CONFIG_XIP=y. When
- * CONFIG_XIP=n, the entire image will be linked to SRAM, so we need to keep
- * the SRAM region XN bit clear or the application code will not be executable.
- */
+/* Executable flag for RAM */
+#define RAM_EXEC COND_CODE_1(CONFIG_EXECUTABLE_RAM, (0), (NOT_EXEC))
+
 #define REGION_RAM_ATTR(size)                                                                      \
 	{(NORMAL_OUTER_INNER_WRITE_BACK_WRITE_READ_ALLOCATE_NON_SHAREABLE |                        \
-	  IF_ENABLED(CONFIG_XIP, (MPU_RASR_XN_Msk |)) size | P_RW_U_NA_Msk)}
+	  RAM_EXEC | size | P_RW_U_NA_Msk)}
 #define REGION_RAM_NOCACHE_ATTR(size)                                                              \
 	{(NORMAL_OUTER_INNER_NON_CACHEABLE_NON_SHAREABLE | MPU_RASR_XN_Msk | size | P_RW_U_NA_Msk)}
 #if defined(CONFIG_MPU_ALLOW_FLASH_WRITE)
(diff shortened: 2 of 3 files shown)


arch/Kconfig
```diff
@@ -241,6 +241,19 @@ config XIP
 	  supply a linker command file when building your image. Enabling this
 	  option increases both the code and data footprint of the image.
 
+config XIP_NOT
+	bool
+	default y if !XIP
```
JarmouniA:
```suggestion
	default y
	depends on !XIP
```

arch/Kconfig
```diff
@@ -241,6 +241,20 @@ config XIP
 	  supply a linker command file when building your image. Enabling this
 	  option increases both the code and data footprint of the image.
 
+config XIP_NOT
```
nashif, line 244: that is a weird way to name a kconfig symbol, where is this being used?

JordanYates: I know the name is weird. The point of the symbol is to force `CONFIG_EXECUTABLE_RAM=y` if `XIP=n`, since its not possible to do something like:
```
config XIP
    select EXECUTABLE_RAM if !XIP
```
Implementing it as a default on `EXECUTABLE_RAM` would allow the user to disable `EXECUTABLE_RAM` while `XIP` is disabled, which would be an invalid configuration.

Happy to take suggestions on naming or a better way to achieve the goal (choice symbol maybe?)

JordanYates:
> Actually, MPU/MMU should be supported with USERSPACE set as well (because the memory partition APIs are hidden by that). But thanks for enlarging the use cases! 👍

I wouldn't know, I haven't compiled an application with `USERSPACE=y` in 5 years of using Zephyr, and I have no plans to start now :).

> Not sure how to implement it, but I would very much prefer EXECUTABLE_RAM to be force-set for non MPU/MMU targets, so that we could immediately know from that symbol's absence that there are restrictions in place.

At a high level, I don't believe there is any way to reliably know at build time whether running functions loaded from LLEXT is possible. It will work if the allocated RAM is marked as executable (or not marked as no-execute), but that depends on where the memory is being allocated from and any custom configuration done on the memory.

This PR is taking a hammer to the problem by configuring the entire RAM space as executable if `LLEXT` is enabled, but there could easily be more nuanced solutions (that only configure the system heap as executable for example). `EXECUTABLE_RAM` wouldn't be set then, but LLEXT functions would still work.

I'm not suggesting this PR is the best long-term solution, but its the simplest I could come up with.

arch/Kconfig
```diff
@@ -241,6 +241,20 @@ config XIP
...
+	default y
+	depends on !XIP
+	select EXECUTABLE_RAM if (MMU || MPU)
+	help
+	  Helper symbol that forces EXECUTABLE_RAM if !XIP.
+
+config EXECUTABLE_RAM
+	bool "Configure RAM to be executable"
+	depends on (MMU || MPU)
+	default y if LLEXT
```
wearyzen, line 255: could we just make this `depend on !XIP` and avoid having the XIP_NOT?

teburd:
> > Any reason to do it this way rather than enabling k_mem_partitions without usermode?
> 
> The memory partition API only appears to be available when `USERSPACE=y`

That's true today but doesn't have to be true tomorrow. Arguably the memory partitions were only needed by usermode previously, but if you have kernel mode extensions this is an incredibly useful feature to enable while not fully enabling userspace.

> 
> https://github.com/zephyrproject-rtos/zephyr/blob/69630923751ba21209abd1a59b84492551962760/include/zephyr/app_memory/mem_domain.h#L27
> 
> That might be the correct approach (I've never used `k_mem_partition` so I don't know), but its not immediately available as a solution.
> 
> > This saves you from making all ram read, write, and execute which likely isn't really what you wanted. But I could be wrong!
> 
> You are correct, it is not ideal. But it is preferable to LLEXT being unusable for functions :)
> 
> edit: I guess the ideal solution would be a Kconfig symbol that controls whether the system heap should be executable or not. I'm not sure how hard that would be though.

I would far rather see k_mem_partition usable with userspace=n, which when I had looked at it previously was quite possible but required some work.

teburd:
Alternative to the userspace like mpu control, defining custom MPU regions and memory in DT is workable as I've seen now.

JordanYates:
> Alternative to the userspace like mpu control, defining custom MPU regions and memory in DT is workable as I've seen now.

Not sure on the meaning here, you've got a better solution or this one is now fine?

wearyzen (commented):
I just saw that my comment was never published. Could we have something like :
```
config EXECUTABLE_RAM
	bool "Configure RAM to be executable" if XIP && !LLEXT
	depends on MMU || MPU
	default y if !XIP || LLEXT
```

Issue #98368, opened 2025-10-28
Zephyr Module Registry for community modules


beriberikix:
### Problem Description

Currently, _discovering_ & integrating third-party modules (such as communication stacks, custom device drivers, or utility libraries) into a Zephyr project is a entirely manual, multi-step process:

- The developer must discover the module through external, often fragmented, channels (blogs, forums, searching external Git providers).
- They must then manually look up the module's Git repository URL and the specific Git revision/tag they wish to use.
- Finally, they must manually edit their project's `west.yml` file to add a new project entry with the correct parameters.

This friction acts as a barrier to entry for new users and impedes the adoption of Zephyr/community efforts. There is no official, unified way to browse or search for compatible modules, leading to duplicated efforts and fragmented discovery. This is a deficiency compared to modern package ecosystems (like Rust's Cargo or Python's pip) and what led me to start https://github.com/golioth/awesome-zephyr-rtos.

### Proposed Change (Summary)

I propose the creation and adoption of two related components:

- Zephyr Module Registry: A new, lightweight Git repository that serves as a centralized, community-driven index of third-party modules. It contains only metadata files, not source code.
- west extension for discovery: A new command-line extension for the west tool that interacts with the registry. It provides commands to search, view, and install modules directly into a project's west.yml with a single, standardized command, such as `west <command-name> install <module-name>`.

This change will establish a standardized, low-friction, and easily discoverable method for managing external Zephyr dependencies.

### Proposed Change (Detailed)

The following is a concrete implementation of this abstract concept. I created a proof-of-concept for both parts of this system to kick off the discussion and test the technical waters:

- Registry Example: https://github.com/beriberikix/west-modules-registry
- Extension Example: https://github.com/beriberikix/wmr-discover-extension

The main point here is the functional model, not a final design. I'm asking the community how to best implement this feature based on this working example.

The functional model needs two key things:

1. Zephyr Module Registry Structure

The module registry would be a new, super lightweight Git repo that acts only as an index as not to create duplication of source code. This is heavily inspired by the simplicity of Brew for macOS and Linux.

* Content: It stores module definitions as `zmr.yml` metadata files.
* Versioning: It uses a straightforward path structure to handle versions and the default `latest` tag:

```
<module-name>/
├── <version-tag>/zmr.yml
└── latest/zmr.yml
```

* Metadata: The `zmr.yml` file is the heart of it, containing the description, groups, and the exact `west` snippet needed for a seamless installation.

[...]

PR #98219, opened 2025-10-24
doc/develop/manifest/external: add Zephelin

fkokosinski:
Introduce documentation for using Zephelin (Zephyr Profiling Library) and Zephelin Trace Viewer as an external module in Zephyr.

Zephelin repository: https://github.com/antmicro/zephelin
Zephelin documentation: https://antmicro.github.io/zephelin/
Zephelin Trace Viewer repository: https://github.com/antmicro/zephelin-trace-viewer
Zephelin Trace Viewer web application: https://antmicro.github.io/zephelin-trace-viewer

PR #99227, opened 2025-11-12
samples: subsys: display: Add gui_guider sample

ZongchunYu:


JarmouniA (changes requested):
We already have LVGL demos in `samples/modules/lvgl/demos`. You should probably submit this to upstream lvgl demos repo.

faxe1008:
@ZongchunYu I get the idea behind this PR, but this is imo the wrong solution. Issues that I see:

- The sample does not do anything useful out of the box. It does not "show you how to do X".
- It introduces an intree depedency of a GUI editor into Zephyr. If anything related to the editor breaks, the burden of fixing it and maintaining backwards compatibility is pushed into Zephyr.

What would be more promising I think would be to have a zephyr module on your end and deal with it there.
There is upcoming support for a west templating mechanism which you could use for your sample application:
https://github.com/zephyrproject-rtos/zephyr/pull/99110

In regards to discoverability of your project:
We are happy to add documentation and a link to your Zephyr module: https://github.com/zephyrproject-rtos/zephyr/pull/98219
Also we have this proposal cooking where your model would also find it's place: https://github.com/zephyrproject-rtos/zephyr/issues/98368

ZongchunYu:
> @ZongchunYu I get the idea behind this PR, but this is imo the wrong solution. Issues that I see:
> 
> * The sample does not do anything useful out of the box. It does not "show you how to do X".
> * It introduces an intree depedency of a GUI editor into Zephyr. If anything related to the editor breaks, the burden of fixing it and maintaining backwards compatibility is pushed into Zephyr.
> 
> What would be more promising I think would be to have a zephyr module on your end and deal with it there. There is upcoming support for a west templating mechanism which you could use for your sample application: #99110
> 
> In regards to discoverability of your project: We are happy to add documentation and a link to your Zephyr module: #98219 Also we have this proposal cooking where your model would also find it's place: #98368

Hello faxe1008,

Thanks for your suggestion!  I will check how to add this zephyr module.

PR #99228, opened 2025-11-12
Bluetooth: Classic: HFP_AG: Get default indicator value

lylezhu2012:
In current implementation, the default indicator value cannot be set when building the SLC. And it causes the incorrect indicator value is notified by AG.

Add a callback `get_indicator_value` to get the default indicator value from the application.

Add shell command `indicator_value` to set the default indicator value. The set indicator value will be passed to HFP AG when callback `get_indicator_value` is notified.

### DIFF
--- a/include/zephyr/bluetooth/classic/hfp_ag.h
+++ b/include/zephyr/bluetooth/classic/hfp_ag.h
@@ -147,6 +147,41 @@ struct bt_hfp_ag_cb {
 	 */
 	void (*sco_disconnected)(struct bt_conn *sco_conn, uint8_t reason);
 
+	/** Get indicator values callback
+	 *
+	 *  If this callback is provided it will be called whenever the AG needs to provide current
+	 *  indicator values to the HF.
+	 *  This typically occurs when the HF sends `AT+CIND?` command to query the current status
+	 *  of AG indicators.
+	 *
+	 *  The application should populate the indicator values through the provided pointers. All
+	 *  indicator values should be set according to the current status of the AG.
+	 *
+	 *  @param ag HFP AG object.
+	 *  @param service Pointer to store service availability indicator value.
+	 *                 0 = service is not available, 1 = service is available.
+	 *  @param strength Pointer to store signal strength indicator value.
+	 *                Range: 0-5 (0 = no signal, 5 = maximum signal).
+	 *  @param roam Pointer to store roaming status indicator value.
+	 *              0 = not roaming, 1 = roaming.
+	 *  @param battery Pointer to store battery level indicator value.
+	 *                 Range: 0-5 (0 = battery exhausted, 5 = battery full).
+	 *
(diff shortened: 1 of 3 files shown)


subsys/bluetooth/host/classic/shell/hfp.c
```diff
@@ -2032,11 +2086,16 @@ static int cmd_ag_hf_indicator(const struct shell *sh, size_t argc, char **argv)
 	"<[R-ready][S-send][P-processing]> "        \
 	"<id> <type> <operation> <text string>"
 
+#define HELP_AG_INDICATOR_VALUE \
+	"[<service availability 0-1> <signal strength 0-5> " \
+	"<roaming status 0-1> <battery level 0-5>]"
```
makeshi, line 2115: form the above code logic, 
	if (argc != 5) {
		shell_help(sh);
		return SHELL_CMD_HELP_PRINTED;
	}
There four agrs more like mandatory arguments

include/zephyr/bluetooth/classic/hfp_ag.h
```diff
@@ -147,6 +147,41 @@ struct bt_hfp_ag_cb {
...
+	 *        zero, the value of these all indicators will be set to 0 by default. And the
+	 *        specific can be set and notified by calling the dedicated function. Such as
+	 *        `service availability indicator value` can be set by calling the function
+	 *        `bt_hfp_ag_service_availability()`. The `signal strength value` can be set
+	 *        by calling `bt_hfp_ag_signal_strength()`, and so on.
+	 *
+	 *  @return 0 in case of success or negative value in case of error.
+	 */
+	int (*get_indicator_value)(struct bt_hfp_ag *ag, uint8_t *service, uint8_t *strength,
+				   uint8_t *roam, uint8_t *battery);
```
makeshi, line 183: Some places use SIGNAL for strength. Do we need to align them?

subsys/bluetooth/host/classic/shell/hfp.c
lylezhu2012: argc could be 1 or 5 in function `cmd_ag_indicator_value()`,

```
	if (argc == 1) {
		has_ag_indicator_value = false;
		return 0;
	}

	if (argc != 5) {
		shell_help(sh);
		return SHELL_CMD_HELP_PRINTED;
	}
```

include/zephyr/bluetooth/classic/hfp_ag.h
lylezhu2012: The signal is reserved word of zephyr. It can not be used as the variable name.

MarkWangChinese (dismissed):

makeshi (dismissed):

makeshi (approved):

MarkWangChinese (approved):

PR #99230, opened 2025-11-12
logging: log_output: Extend API with optional get_available function

nordic-krch:
Add optional `log_output_get_available` function which returns amount of data that can be currently accepted by the backend. Extend RTT backend to implement that functionality and allow getting amount of space in the write buffer.

Use that functionality in dictionary mode to ensure that only whole messages are written to the log output.

### DIFF
--- a/subsys/logging/backends/log_backend_rtt.c
+++ b/subsys/logging/backends/log_backend_rtt.c
@@ -282,7 +282,16 @@ static int data_out(uint8_t *data, size_t length, void *ctx)
 	return logging_func(data, length, ctx);
 }
 
-LOG_OUTPUT_DEFINE(log_output_rtt, data_out, char_buf, sizeof(char_buf));
+static int get_available(void *ctx)
+{
+	ARG_UNUSED(ctx);
+
+	return SEGGER_RTT_GetAvailWriteSpace(CONFIG_LOG_BACKEND_RTT_BUFFER);
+}
+
+LOG_OUTPUT_EXT_DEFINE(log_output_rtt, data_out,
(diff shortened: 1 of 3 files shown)


subsys/logging/backends/log_backend_rtt.c
```diff
@@ -282,7 +282,16 @@ static int data_out(uint8_t *data, size_t length, void *ctx)
 	return logging_func(data, length, ctx);
 }
 
-LOG_OUTPUT_DEFINE(log_output_rtt, data_out, char_buf, sizeof(char_buf));
+static int get_available(void *ctx)
+{
+	ARG_UNUSED(ctx);
+
```
ksstms, line 288: Add something like this so we don't break overwrite mode?
```c
	if (IS_ENABLED(CONFIG_LOG_BACKEND_RTT_MODE_OVERWRITE)) {
		return _SEGGER_RTT.aUp[CONFIG_LOG_BACKEND_RTT_BUFFER].SizeOfBuffer;
	}
```

subsys/logging/backends/log_backend_rtt.c
```diff
@@ -282,7 +282,16 @@ static int data_out(uint8_t *data, size_t length, void *ctx)
...
+	return SEGGER_RTT_GetAvailWriteSpace(CONFIG_LOG_BACKEND_RTT_BUFFER);
+}
+
+LOG_OUTPUT_EXT_DEFINE(log_output_rtt, data_out,
+		IS_ENABLED(CONFIG_LOG_BACKEND_RTT_OUTPUT_DICTIONARY_HEX) ? get_available : NULL,
```
ksstms, line 293: Should be `CONFIG_LOG_BACKEND_RTT_OUTPUT_DICTIONARY` so it works with the normal binary format too.

PR #99232, opened 2025-11-12
STM32 RNG: NIST SP800-90B support

gautierg-st:
In order to comply to NIST SP800-90B, the STM32 RNG should set its registers as documented in Application Note AN4230 (Table 3). To that effect, some values have been modified to match the table, and some have been added.

### DIFF
--- a/drivers/entropy/entropy_stm32.c
+++ b/drivers/entropy/entropy_stm32.c
@@ -247,22 +247,18 @@ static void configure_rng(void)
 
 #if DT_INST_NODE_HAS_PROP(0, health_test_config)
 #if DT_INST_NODE_HAS_PROP(0, health_test_magic)
+		/* On certain series, a magic value must be written first as
+		 * health configuration before the actual configuration value.
+		 */
 		LL_RNG_SetHealthConfig(rng, DT_INST_PROP(0, health_test_magic));
 #endif /* health_test_magic */
 		LL_RNG_SetHealthConfig(rng, desired_htcr);
 #endif /* health_test_config */
 
-#if defined(CONFIG_SOC_SERIES_STM32L4X)
-		LL_RNG_ResetConditioningResetBit(rng);
-		/* Wait for conditioning reset process to be completed */
-		while (LL_RNG_IsResetConditioningBitSet(rng) == 1) {
-		}
-#else
 		LL_RNG_DisableCondReset(rng);
 		/* Wait for conditioning reset process to be completed */
 		while (LL_RNG_IsEnabledCondReset(rng) == 1) {
 		}
-#endif /* CONFIG_SOC_SERIES_STM32L4X */
 	}
 #endif /* STM32_CONDRST_SUPPORT */
 
@@ -312,23 +308,14 @@ static int recover_seed_error(RNG_TypeDef *rng)
 {
 	uint32_t count_timeout = 0;
 
-#if defined(CONFIG_SOC_SERIES_STM32L4X)
-		LL_RNG_SetConditioningResetBit(rng);
-		LL_RNG_ResetConditioningResetBit(rng);
-#else
-		LL_RNG_EnableCondReset(rng);
-		LL_RNG_DisableCondReset(rng);
-#endif /* CONFIG_SOC_SERIES_STM32L4X */
+	LL_RNG_EnableCondReset(rng);
+	LL_RNG_DisableCondReset(rng);
 
 	/* When reset process is done cond reset bit is read 0
 	 * This typically takes: 2 AHB clock cycles + 2 RNG clock cycles.
 	 */
 
-#if defined(CONFIG_SOC_SERIES_STM32L4X)
-	while (LL_RNG_IsResetConditioningBitSet(rng) ||
-#else
 	while (LL_RNG_IsEnabledCondReset(rng) ||
-#endif /* CONFIG_SOC_SERIES_STM32L4X */
 		ll_rng_is_active_seis(rng) ||
 		ll_rng_is_active_secs(rng)) {
 		count_timeout++;
--- a/drivers/entropy/entropy_stm32.c
+++ b/drivers/entropy/entropy_stm32.c
@@ -218,8 +218,10 @@ static void configure_rng(void)
 #ifdef STM32_CONDRST_SUPPORT
 	uint32_t desired_nist_cfg = DT_INST_PROP_OR(0, nist_config, 0U);
 	uint32_t desired_htcr = DT_INST_PROP_OR(0, health_test_config, 0U);
+	uint32_t desired_nscr = DT_INST_PROP_OR(0, noise_source_control, 0U);
 	uint32_t cur_nist_cfg = 0U;
 	uint32_t cur_htcr = 0U;
+	uint32_t cur_nscr = 0U;
 
 #if DT_INST_NODE_HAS_PROP(0, nist_config)
 	/*
@@ -242,7 +244,12 @@ static void configure_rng(void)
 	cur_htcr = LL_RNG_GetHealthConfig(rng);
 #endif /* health_test_config */
 
-	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr) {
+#if DT_INST_NODE_HAS_PROP(0, noise_source_control)
+	cur_nscr = LL_RNG_GetNoiseConfig(rng);
+#endif /* noise_source_control */
+
+	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr ||
+	    cur_nscr != desired_nscr) {
 		stm32_reg_modify_bits(&rng->CR, cur_nist_cfg, desired_nist_cfg | RNG_CR_CONDRST);
 
 #if DT_INST_NODE_HAS_PROP(0, health_test_config)
@@ -254,6 +261,10 @@ static void configure_rng(void)
 #endif /* health_test_magic */
 		LL_RNG_SetHealthConfig(rng, desired_htcr);
 #endif /* health_test_config */
+
+#if DT_INST_NODE_HAS_PROP(0, noise_source_control)
+		LL_RNG_SetNoiseConfig(rng, DT_INST_PROP(0, noise_source_control));
+#endif /* noise_source_control */
 
 		LL_RNG_DisableCondReset(rng);
 		/* Wait for conditioning reset process to be completed */
--- a/dts/arm/st/h7rs/stm32h7rs.dtsi
+++ b/dts/arm/st/h7rs/stm32h7rs.dtsi
@@ -917,6 +917,8 @@
 			reg = <0x48020000 0x400>;
 			clocks = <&rcc STM32_CLOCK(AHB3, 0)>;
 			interrupts = <37 0>;
+			nist-config = <0xf00d00>;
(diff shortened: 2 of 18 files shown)


drivers/entropy/entropy_stm32.c
```diff
@@ -224,22 +226,36 @@ static void configure_rng(void)
 	cur_htcr = LL_RNG_GetHealthConfig(rng);
 #endif /* health_test_config */
 
-	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr) {
+	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr ||
+	    cur_nscr != desired_nscr) {
 		stm32_reg_modify_bits(&rng->CR, cur_nist_cfg, desired_nist_cfg | RNG_CR_CONDRST);
 
+#if defined(CONFIG_SOC_SERIES_STM32L4X)
```
mathieuchopstm: To be sure we clean up this mess in due time:
```suggestion
/*
 * TODO: this is needed only due to mismatch of STM32L4 HAL with others.
 * Remove this special case once STM32CubeL4 v1.18.2 or later has been
 * integrated in hal_stm32 (see also ST Internal Reference 207828).
 */
#if defined(CONFIG_SOC_SERIES_STM32L4X)
```

drivers/entropy/entropy_stm32.c
```diff
@@ -224,22 +226,36 @@ static void configure_rng(void)
 	cur_htcr = LL_RNG_GetHealthConfig(rng);
```
mathieuchopstm, line 244: Outside PR scope: I'm surprised this builds on L4

drivers/entropy/entropy_stm32.c
```diff
@@ -224,22 +226,36 @@ static void configure_rng(void)
...
+		LL_RNG_SetHealthconfiguration(rng, desired_htcr);
 #endif /* health_test_config */
+#endif /* RNG_VER_3_2 || RNG_VER_3_1 */
 
-#if defined(CONFIG_SOC_SERIES_STM32L4X)
 		LL_RNG_ResetConditioningResetBit(rng);
 		/* Wait for conditioning reset process to be completed */
 		while (LL_RNG_IsResetConditioningBitSet(rng) == 1) {
 		}
 #else
```
mathieuchopstm:
```suggestion
#else /* CONFIG_SOC_SERIES_STM32L4X */
```

drivers/entropy/entropy_stm32.c
```diff
@@ -224,22 +226,36 @@ static void configure_rng(void)
...
+#if DT_INST_NODE_HAS_PROP(0, health_test_config)
+#if DT_INST_NODE_HAS_PROP(0, health_test_magic)
+		LL_RNG_SetHealthConfig(rng, DT_INST_PROP(0, health_test_magic));
```
mathieuchopstm, line 260: Nit, maybe outside PR scope: I'd add a comment because that's quite confusing at first glance
```suggestion
		/* On certain series, a magic value must be written first as
		 * health configuration before the actual configuration value.
		 */
		LL_RNG_SetHealthConfig(rng, DT_INST_PROP(0, health_test_magic));
```

dts/arm/st/l4/stm32l4r5.dtsi
```diff
@@ -41,5 +41,11 @@
 				STM32_DMA_MEM_INC | STM32_DMA_PERIPH_8BITS | STM32_DMA_MEM_32BITS |
 				STM32_DMA_PRIORITY_HIGH) STM32_DMA_FIFO_1_4>;
 		};
+
+		rng: rng@50060800 {
+			/delete-property/ nist-config;
+			/delete-property/ health-test-magic;
+			/delete-property/ health-test-config;
+		};
```
mathieuchopstm, line 50: Could use a comment explaining why the values are removed.

dts/arm/st/wba/stm32wba.dtsi
mathieuchopstm, line 1: Lacking `noise-source-control = <0x0FFF>`?

gautierg-st: Yes, and so does the H7R/S series. But these two series lack the NSCR field in the RNG_Typedef definition (as well as the LL functions to set/get the value). See internal tickets 221082 and 221083.

drivers/entropy/entropy_stm32.c
gautierg-st: Done

drivers/entropy/entropy_stm32.c
gautierg-st: You're correct, I've updated with the proper function.

drivers/entropy/entropy_stm32.c
gautierg-st: Done

drivers/entropy/entropy_stm32.c
gautierg-st: Done

dts/arm/st/l4/stm32l4r5.dtsi
gautierg-st: Done

dts/arm/st/wba/stm32wba.dtsi
gautierg-st: Since the noise field is not yet available for configuration, I've used Configuration C (from the RefMan) values instead of NIST ones.

mathieuchopstm: A note/comment should be added in the DTSI to clarify why it was done this way; and also, such that the "correct" values can be used once the fields are available in HAL.

dts/arm/st/l4/stm32l4r5.dtsi
```diff
@@ -41,5 +41,12 @@
 				STM32_DMA_MEM_INC | STM32_DMA_PERIPH_8BITS | STM32_DMA_MEM_32BITS |
 				STM32_DMA_PRIORITY_HIGH) STM32_DMA_FIFO_1_4>;
 		};
+
+		rng: rng@50060800 {
+			/* These properties are only available on L4P/Q, not on L4R/S */
```
mathieuchopstm: Proposal:
```suggestion
			/* L4R/S RNG differs from L4P/Q; these properties are not available */
```

mathieuchopstm (commented):
Otherwise LGTM

dts/arm/st/l4/stm32l4r5.dtsi
gautierg-st: Done

drivers/entropy/entropy_stm32.c
```diff
@@ -220,26 +222,51 @@ static void configure_rng(void)
...
+#if defined(CONFIG_SOC_SERIES_STM32L4X)
+	cur_htcr = LL_RNG_GetHealthconfiguration(rng);
+#else /* CONFIG_SOC_SERIES_STM32L4X */
 	cur_htcr = LL_RNG_GetHealthConfig(rng);
+#endif /* CONFIG_SOC_SERIES_STM32L4X */
 #endif /* health_test_config */
 
-	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr) {
+	if (cur_nist_cfg != desired_nist_cfg || cur_htcr != desired_htcr ||
+	    cur_nscr != desired_nscr) {
```
etienne-lms, line 252: `cur_nscr` is previously read. Should be, for consistency?

etienne-lms (commented):
I think commit "drivers: entropy: stm32: add noise source control support"
would be worth to be split in 2 commits: first the health test part for L4 series, then the noise source configuration.

mathieuchopstm (dismissed):

etienne-lms (commented):
For commit "dts: arm: st: fix rng nist compliant configs",
I suggest to explicit the referred application note:
```patch
 (...)
-registers as documented in Application Note AN4230 (Table 3).
+registers as documented in Application Note AN4230 (Rev 12, Table 3) [1].
 (...)
 
+Link: https://www.st.com/resource/en/application_note/dm00073853-stm32-microcontroller-random-number-generation-validation-using-the-nist-statistical-test-suite-stmicroelectronics.pdf [1]
 Signed-off-by: ...
```

drivers/entropy/entropy_stm32.c
gautierg-st: Yes it should be read beforehand. Done

drivers/entropy/entropy_stm32.c
```diff
@@ -220,26 +220,46 @@ static void configure_rng(void)
...
 
+/* TODO: this is needed only due to mismatch of STM32L4 HAL with others.
+ * Remove this special case once STM32CubeL4 v1.18.2 or later has been
+ * integrated in hal_stm32 (see also ST Internal Reference 207828).
+ */
 #if DT_INST_NODE_HAS_PROP(0, health_test_config)
+#if defined(CONFIG_SOC_SERIES_STM32L4X)
+	cur_htcr = LL_RNG_GetHealthconfiguration(rng);
+#else /* CONFIG_SOC_SERIES_STM32L4X */
 	cur_htcr = LL_RNG_GetHealthConfig(rng);
```
erwango: If we're certain this will be fixed in L4 driver, it would be cleaner to put a patch in hal_stm32 module and clean up the code here.
Once the hal is updated this will be transparent for the driver.

erwango (changes requested):
Otherwise LGTM

drivers/entropy/entropy_stm32.c
gautierg-st: It is fixed in latest L4 HAL release 1.13.6 released a week ago (see https://github.com/STMicroelectronics/stm32l4xx-hal-driver/blob/047a8610c7dc0dcfdb4cd63b86cd3a2cbd3e5899/Inc/stm32l4xx_ll_rng.h#L678).
So I think it's easier to update this HAL and then I'll update this PR.

erwango (dismissed):

dts/arm/st/h7rs/stm32h7rs.dtsi
```diff
@@ -917,6 +917,8 @@
 			reg = <0x48020000 0x400>;
 			clocks = <&rcc STM32_CLOCK(AHB3, 0)>;
 			interrupts = <37 0>;
+			nist-config = <0xf00d>;
+			health-test-config = <0xaac7>;
```
etienne-lms, line 921: Should be `nist-config = <0xf00d00>;`, the implementation does not left shift the value provided by the DT property.

etienne-lms (commented):
Did you discard the WBA6x refman config C: `nist-config = <0x820f00>;` in **stm32wba65.dtsi**? It still mentioned in the commit message and would apply.

Note also that **stm32wba.dtsi** sets `nist-config = <0xf00d>;` but the value should be `0xf00d00` (as is recommended in WBA5x refman, config C in its refman) since the implementation does not left shift the DT property value.

dts/arm/st/h7rs/stm32h7rs.dtsi
gautierg-st: Done. And updated WBA as well.

etienne-lms (approved):

erwango (approved):

PR #99233, opened 2025-11-12
 drivers: spi: sam0: Add DMA support for synchronous transfers

RamyaI73943:
The SAM0 SPI driver previously supported DMA only for asynchronous operations, which required higher-level drivers needing blocking DMA transfers to implement workarounds. This patch extends DMA support to the synchronous transfer path, enabling efficient blocking SPI transactions without driver-specific hacks.

With this change, both synchronous and asynchronous SPI operations can leverage DMA for improved performance and reduced CPU overhead.

### DIFF
--- a/drivers/spi/Kconfig.sam0
+++ b/drivers/spi/Kconfig.sam0
@@ -8,6 +8,14 @@ config SPI_SAM0
 	default y
 	depends on DT_HAS_ATMEL_SAM0_SPI_ENABLED
 	select PINCTRL
-	select DMA if SPI_ASYNC
 	help
 	  Enable support for the SAM0 SERCOM SPI driver.
+
+config SPI_SAM0_DMA
+	bool "Atmel SAM0 dma support"
+	default y
+	depends on SPI_SAM0
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ATMEL_SAM0_SPI),dmas)
+	depends on DMA
+	help
+	  Enable DMA-based data transfers for the Atmel SAM0 SPI driver.
--- a/drivers/spi/spi_sam0.c
+++ b/drivers/spi/spi_sam0.c
@@ -36,7 +36,7 @@ struct spi_sam0_config {
 	uint32_t gclk_gen;
 	uint16_t gclk_id;
 
-#ifdef CONFIG_SPI_ASYNC
+#ifdef CONFIG_SPI_SAM0_DMA
 	const struct device *dma_dev;
 	uint8_t tx_dma_request;
 	uint8_t tx_dma_channel;
@@ -48,7 +48,7 @@ struct spi_sam0_config {
 /* Device run time data */
 struct spi_sam0_data {
 	struct spi_context ctx;
-#ifdef CONFIG_SPI_ASYNC
+#ifdef CONFIG_SPI_SAM0_DMA
 	const struct device *dev;
 	uint32_t dma_segment_len;
 #endif
@@ -372,57 +372,7 @@ static bool spi_sam0_is_regular(const struct spi_buf_set *tx_bufs,
 	return true;
 }
 
-static int spi_sam0_transceive(const struct device *dev,
-			       const struct spi_config *config,
-			       const struct spi_buf_set *tx_bufs,
-			       const struct spi_buf_set *rx_bufs)
-{
-	const struct spi_sam0_config *cfg = dev->config;
-	struct spi_sam0_data *data = dev->data;
-	SercomSpi *regs = cfg->regs;
-	int err;
-
-	spi_context_lock(&data->ctx, false, NULL, NULL, config);
-
-	err = spi_sam0_configure(dev, config);
-	if (err != 0) {
-		goto done;
-	}
-
-	spi_context_cs_control(&data->ctx, true);
-
-	/* This driver special cases the common send only, receive
-	 * only, and transmit then receive operations.	This special
-	 * casing is 4x faster than the spi_context() routines
-	 * and allows the transmit and receive to be interleaved.
-	 */
-	if (spi_sam0_is_regular(tx_bufs, rx_bufs)) {
-		spi_sam0_fast_transceive(dev, config, tx_bufs, rx_bufs);
-	} else {
-		spi_context_buffers_setup(&data->ctx, tx_bufs, rx_bufs, 1);
-
-		do {
-			spi_sam0_shift_master(regs, data);
-		} while (spi_sam0_transfer_ongoing(data));
-	}
-
-	spi_context_cs_control(&data->ctx, false);
-
-done:
-	spi_context_release(&data->ctx, err);
-	return err;
-}
-
-static int spi_sam0_transceive_sync(const struct device *dev,
-				    const struct spi_config *config,
-				    const struct spi_buf_set *tx_bufs,
-				    const struct spi_buf_set *rx_bufs)
-{
-	return spi_sam0_transceive(dev, config, tx_bufs, rx_bufs);
-}
-
-#ifdef CONFIG_SPI_ASYNC
-
+#ifdef CONFIG_SPI_SAM0_DMA
 static void spi_sam0_dma_rx_done(const struct device *dma_dev, void *arg,
 				 uint32_t id, int error_code);
 
@@ -599,28 +549,22 @@ static void spi_sam0_dma_rx_done(const struct device *dma_dev, void *arg,
 		return;
 	}
 }
+#endif /* CONFIG_SPI_SAM0_DMA */
 
-
-static int spi_sam0_transceive_async(const struct device *dev,
-				     const struct spi_config *config,
-				     const struct spi_buf_set *tx_bufs,
-				     const struct spi_buf_set *rx_bufs,
-				     spi_callback_t cb,
-				     void *userdata)
+static int spi_sam0_transceive(const struct device *dev,
+			       const struct spi_config *config,
+			       const struct spi_buf_set *tx_bufs,
+			       const struct spi_buf_set *rx_bufs,
+			       bool asynchronous,
+			       spi_callback_t cb,
+			       void *userdata)
 {
 	const struct spi_sam0_config *cfg = dev->config;
 	struct spi_sam0_data *data = dev->data;
+	SercomSpi *regs = cfg->regs;
 	int retval;
 
-	/*
-	 * Transmit clocks the output and we use receive to determine when
-	 * the transmit is done, so we always need both
-	 */
-	if (cfg->tx_dma_channel == 0xFF || cfg->rx_dma_channel == 0xFF) {
-		return -ENOTSUP;
-	}
-
-	spi_context_lock(&data->ctx, true, cb, userdata, config);
+	spi_context_lock(&data->ctx, asynchronous, cb, userdata, config);
 
 	retval = spi_sam0_configure(dev, config);
 	if (retval != 0) {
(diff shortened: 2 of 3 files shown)


drivers/spi/spi_sam0.c
```diff
@@ -637,20 +594,64 @@ static int spi_sam0_transceive_async(const struct device *dev,
...
+	} else {
+		spi_context_buffers_setup(&data->ctx, tx_bufs, rx_bufs, 1);
+
+		do {
+			spi_sam0_shift_master(regs, data);
+		} while (spi_sam0_transfer_ongoing(data));
+	}
 
+	spi_context_cs_control(&data->ctx, false);
+#endif /* CONFIG_SPI_SAM0_DMA */
```
maass-hamburg: we still want this as a fallback, when `CONFIG_SPI_SAM0_DMA=y`,
but for this instance `cfg->tx_dma_channel == 0xFF || cfg->rx_dma_channel == 0xFF` is true. It could be that one instance supports dma and the other don't.

maass-hamburg (changes requested):

drivers/spi/spi_sam0.c
RamyaI73943: > we still want this as a fallback, when `CONFIG_SPI_SAM0_DMA=y`, but for this instance `cfg->tx_dma_channel == 0xFF || cfg->rx_dma_channel == 0xFF` is true. It could be that one instance supports dma and the other don't.

Okay. Updated the PR. Kindly check it.

drivers/spi/spi_sam0.c
```diff
@@ -629,28 +573,79 @@ static int spi_sam0_transceive_async(const struct device *dev,
...
-	if (retval != 0) {
-		goto err_cs;
+		spi_sam0_dma_advance_segment(dev);
+		retval = spi_sam0_dma_advance_buffers(dev);
+		if (retval != 0) {
+			goto err_cs;
+		}
+
+		/* Wait for DMA completion signaled by spi_sam0_dma_rx_done() */
+		retval = spi_context_wait_for_completion(&data->ctx);
```
maass-hamburg:
```suggestion
		if (retval != 0) {
			dma_stop(cfg->dma_dev, cfg->tx_dma_channel);
			dma_stop(cfg->dma_dev, cfg->rx_dma_channel);

			spi_context_cs_control(&data->ctx, false);
		} else {

			/* Wait for DMA completion signaled by spi_sam0_dma_rx_done() */
			retval = spi_context_wait_for_completion(&data->ctx);
		}
```

then we can remove the
```
rr_cs:
	dma_stop(cfg->dma_dev, cfg->tx_dma_channel);
	dma_stop(cfg->dma_dev, cfg->rx_dma_channel);

	spi_context_cs_control(&data->ctx, false);
```
further down

drivers/spi/spi_sam0.c
```diff
@@ -629,28 +573,79 @@ static int spi_sam0_transceive_async(const struct device *dev,
...
+
+		spi_context_release(&data->ctx, retval);
+
+		return retval;
 	}
```
maass-hamburg, line 604:
```suggestion
	} else (asynchronous) {
		retval = -ENOTSUP;
		spi_context_cs_control(&data->ctx, false);
		goto err_unlock;
	}
```
need to make sure to fail async when dma is not available on this instance

maass-hamburg (changes requested):

drivers/spi/spi_sam0.c
RamyaI73943: @maass-hamburg Okay

drivers/spi/spi_sam0.c
```diff
@@ -692,7 +685,7 @@ static int spi_sam0_init(const struct device *dev)
 		return err;
 	}
 
-#ifdef CONFIG_SPI_ASYNC
+#ifdef CONFIG_SPI_SAM0_DMA
 	if (!device_is_ready(cfg->dma_dev)) {
 		return -ENODEV;
 	}
```
maass-hamburg: shouldn't fail here too, dma_dev could not be set for one instance, that doesn't need it

RamyaI73943: > shouldn't fail here too, dma_dev could not be set for one instance, that doesn't need it

@maass-hamburg Okay. But we should check whether device_is_ready() for the instance that is defined dma. Right?
So I would suggest the below change.
```
if (cfg->dma_dev && !device_is_ready(cfg->dma_dev)) {
        return -ENODEV;
}
```
What do you think about this?

maass-hamburg: Yes but I would prefer a explicit check for NULL
```
if ((cfg->dma_dev != NULL) && !device_is_ready(cfg->dma_dev)) {
```

RamyaI73943: > Yes but I would prefer a explicit check for NULL
> 
> ```
> if ((cfg->dma_dev != NULL) && !device_is_ready(cfg->dma_dev)) {
> ```

@maass-hamburg Okay

drivers/spi/spi_sam0.c
```diff
@@ -629,28 +573,77 @@ static int spi_sam0_transceive_async(const struct device *dev,
...
+
+			spi_context_cs_control(&data->ctx, false);
+		} else {
+			/* Wait for DMA completion signaled by spi_sam0_dma_rx_done() */
+			retval = spi_context_wait_for_completion(&data->ctx);
+		}
+		spi_context_release(&data->ctx, retval);
+
+		return retval;
+	} else if (asynchronous) {
```
maass-hamburg:
```suggestion
	}

	if (asynchronous) {
```

first if ends with a return, so `else if` can be replaced with a regular `if`.

also makes the ci happy

maass-hamburg (dismissed):

drivers/spi/Kconfig.sam0
```diff
@@ -8,6 +8,14 @@ config SPI_SAM0
...
-	select DMA if SPI_ASYNC
 	help
 	  Enable support for the SAM0 SERCOM SPI driver.
+
+config SPI_SAM0_DMA
+	bool "Atmel SAM0 dma support"
+	default y
+	depends on SPI_SAM0
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ATMEL_SAM0_SPI),dmas)
+	select DMA
```
nandojve: Let's use depends on DMA. This means that user must select this option in their project to use DMA.

nandojve (changes requested):

drivers/spi/Kconfig.sam0
maass-hamburg: This would be a breaking change as
`select DMA if SPI_ASYNC`
was there before. 

Also I think the default should be the variant, that is the fastest.
The user can still use the slower other variant, by deactivating this option. Polling should not be the default if other ways are available, as polling will lead to a higher cpu load.

RamyaI73943:
Hi @nandojve, @ArunMCHP, @decsny, @pdgendt, @stephanosio, @tbursztyka , @teburd could you please take a look at this PR when you have a chance? It has one approval and just needs a second review to merge. Thanks!

drivers/spi/Kconfig.sam0
nandojve: I see.

A solution to this could be add a migration note informing users about what they need to do in case of build errors, which is define all the DMA lines for the peripherals that already have DMA support, not only SPI if this happens.

nandojve (changes requested):

drivers/spi/Kconfig.sam0
RamyaI73943: > I see.
> 
> A solution to this could be add a migration note informing users about what they need to do in case of build errors, which is define all the DMA lines for the peripherals that already have DMA support, not only SPI if this happens.

@maass-hamburg Shall we change it to using "depends on DMA". What do you think about this?

maass-hamburg: In zephyr we normally use `select` for dependencies of drivers, so that they are automatically enabled, so we shouldn't deviate from that general rule here. Or what else is a argument against that? Also I don't want more overlays for like spi based ethernet chips. The fastest option should be the default.

RamyaI73943: @maass-hamburg Okay

RamyaI73943: @nandojve Do you have a chance to reconsider your review based on @maass-hamburg comments. What do you think about this?

nandojve: @RamyaI73943 ,

`normally` is not equal to `always` and most of time normally is equal to wrong.

The dependency must be there to this work. This is why a `depends on` and not the contrary.

You need to understand that this is an `optional` configuration. It is already by `default=y` then if all dependencies met it will be auto-selected. You don't need to `force` the usage to broke all the boards that don't include the DMA driver enabled or don't have DMA at all. Why do you possible want that ?

Did you realize that samd20 does not even have a DMA peripheral ?

RamyaI73943: @maass-hamburg, @nandojve 
I understand and appreciate both of your perspectives.

From my understanding, the `select DMA` approach prioritizes keeping the fastest implementation enabled by default without requiring additional user configuration, while the `depends on DMA` approach ensures that the option is only available when the DMA subsystem and supporting hardware are actually present, helping maintain portability across all SAM0 variants.

At this stage, I am not confident in making the final decision between these two approaches. I would therefore appreciate both of your guidance on which method you would prefer me to follow, and I will update the patch accordingly once we have a clear direction.

maass-hamburg: > You need to understand that this is an `optional` configuration. It is already by `default=y` then if all dependencies met it will be auto-selected. You don't need to `force` the usage to broke all the boards that don't include the DMA driver enabled or don't have DMA at all. Why do you possible want that ?

If `config DMA` is enabled (by selecting it), the according dma driver should be enabled by default, if at least one instance is enabled in the dt.

For platforms that don't have a dma, it should be catched by `depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ATMEL_SAM0_SPI),dmas)`. `config DMA` should not be used as a protection. It can be enabled at any time and says nothing about the real availability of dma. Same as `depends on I2C` will neither ensure, that the board supports I2C at all or that the needed i2c controller is enabled.

And yes it is the whole point i want it enabled by default, if all requirements are fulfilled. If possible it should use the faster way by default. What is the advantage of not doing it? Is is the code size?

I for my part, don't want to see that for spi based ethernet controllers board specific configs are added to the networking samples, because _We need to enable DMA here to increase performance_ especially for the iperf sample.

nandojve: Unconditionally `select DMA` is not acceptable in this driver.

RamyaI73943: @nandojve 
@maass-hamburg 
 Since the PR is still open, could you please review the discussion and help decide on the preferred approach so we can move forward?

nandojve: If you can show a real example that this driver works `with DMA` in one SPI port and `without DMA` in a second port I can accept it.

tbursztyka (dismissed):
logic seems to handle both case dma being configured or not. From my side it looks sound, though I have no hw for testing it.

teburd (dismissed):

RamyaI73943:
This pull request has been approved by three reviewers. Please let me know if any further action is required on my part.

drivers/spi/Kconfig.sam0
nandojve: ^ Still pending

nandojve, line 19: This still pending

nandojve (changes requested):

drivers/spi/Kconfig.sam0
RamyaI73943: @nandojve

**1. Test Setup**

**Node 1 — SAM E54 Xplained Pro +  Zperf**

I have validated the mixed DMA/non-DMA SPI instance behavior on real hardware. Below are the test setup details and results.

| SPI Instance | Header | Connected Device | DMA Configuration | SERCOM |
|---|---|---|---|---|
| SPI Port 1 | EXT1 | LAN865x MAC-PHY | DMA **enabled** | SERCOM 4 |
| SPI Port 2 | EXT2 | LAN865x MAC-PHY | DMA **disabled** |SERCOM 6 |

**Node 2 — Raspberry Pi 4 + iperf** 

Two EVB-LAN8670-USB devices connected respectively to the two LAN865x devices on Node 1

I am running `west build -b same54_xpro samples/net/zperf `sample application test.


**2. Validation Summary**

The test was performed under both `select DMA` and `depends on DMA` Kconfig configurations. In both cases, the observed behavior was identical and consistent with the expected per-instance DMA fallback logic:

**SPI Port 1 (DMA enabled):** Achieved maximum throughput, confirming the DMA transfer path is correctly exercised.
**SPI Port 2 (DMA disabled):** Operated correctly over the polling fallback path, with no errors or instability.

This confirms that both SPI instances co-exist correctly — one utilizing DMA and the other gracefully falling back to the non-DMA path — as requested 



**3. Identified Gaps and fixes:**
                                                                                                                                                                                                                                                                                Two fixes allow SPI instances with and without dmas in devicetree to coexist correctly on the same SoC.
                                                                                                                                                                                                                   
  Fix 1: Guard `SPI_SAM0_DMA_CHANNELS` macro with `IF_ENABLED`

Wrap the DMA field assignments with `IF_ENABLED(DT_INST_NODE_HAS_PROP(n, dmas), (...))` so that dma_dev and channel fields are only populated when the dmas property is present. Without this, instances lacking
dmas would cause a build failure or leave dma_dev in an undefined state.

  Fix 2: Add `cfg->dma_dev != NULL` guard in `spi_sam0_transceive()`

The existing channel check (!= 0xFF) is insufficient — zero-initialised channels have value 0, not 0xFF, so a NULL dma_dev instance could still enter the DMA path and crash. Adding cfg->dma_dev != NULL makes
the guard unambiguous.

I will update the PR with above fixes soon.
**Note:** These fixes are needed under both `select DMA` and `depends on DMA` Kconfig configurations.


**4. Achieved bandwidth results**
| DMA | RX | TX | 
|---|---|---|
| Enabled | 6.76 Mbps | 6.81 Mbps | 
| Disabled | 3.91 Mbps | 4.44 Mbps |

RamyaI73943:
@maass-hamburg , @teburd , @tbursztyka I am updating this PR with two fixes. Please refer the comment for more details. https://github.com/zephyrproject-rtos/zephyr/pull/99233#discussion_r3093104768

drivers/spi/Kconfig.sam0
nandojve, line 19: Now this line can be removed.
```diff
-    select DMA
```
If the devicetree enable DMA you got what you needed.
Remember, DMA should be selected by a Kconfig that should be always `default y` and `depends on` a devicetree node.
So, this driver now works 100% independently from DMA. If user select it, then you get max performance.
You can document the behavior in the https://github.com/zephyrproject-rtos/zephyr/blob/7bbe9da0892c2c3bc2cdefbc5e60b941f9a36da4/dts/bindings/spi/atmel%2Csam0-spi.yaml#L5

Check this as example:
https://github.com/zephyrproject-rtos/zephyr/blob/7bbe9da0892c2c3bc2cdefbc5e60b941f9a36da4/dts/bindings/usb/atmel%2Csam-udp.yaml#L21-L40

nandojve (changes requested):

drivers/spi/Kconfig.sam0
RamyaI73943: > Now this line can be removed.

Thank you for the comments. Before pushing the update, I want to confirm my understanding. So now your suggestion is to remove `select DMA` and add `depends on DMA`. Right?

nandojve: Correct, now driver if fully compatible with/without DMA.
If the system offer DMA, the driver should be able to auto select best option.
The DMA always should be select by the user somehow.
By default, if the DMA node is enabled at devicetree it auto selct DMA in the DMA Kconfig session. User still can disable by devicetree or force by CONFIG_DMA=n at prj.conf

maass-hamburg: Just to be clear `CONFIG_DMA` only activatets the driver area it doesn't mean that we are going to use DMA in any way. I could activate it on a system that doesn't have any DMA and it would be fine. If a enabled `CONFIG_DMA` would lead to a problem on that platform, that is a bug.

nandojve:
Hi @RamyaI73943 ,

CI must be green : )

nandojve (approved):

RamyaI73943:
@nandojve Thanks for approving the PR. Could you please resolve the conversations opened by you and approve it once again?

tbursztyka (approved):

nandojve (approved):

PR #99238, opened 2025-11-12
arch: arm: mpu: add configurable cache policies for armv7m and v8

lucien-nxp:
Add Kconfig options for FLASH and SRAM cache policy selection:

- CONFIG_ARM_MPU_SRAM_WRITE_THROUGH: enables Write-Through cache policy for SRAM regions instead of default Write-Back

Includes corresponding MPU attribute macros for ARMv7-M and ARMv8-M architectures. Maintains backward compatibility with existing configurations.

Validation performed on NXP RT1180 platform on arm_mpu_wt case


arch/arm/core/mpu/Kconfig
```diff
@@ -79,6 +79,20 @@ config ARM_MPU_PXN
 	  unprivileged mode and executing such region from privileged mode will
 	  result in a Memory Management fault.
 
+config ARM_MPU_FLASH_WRITE_BACK
+	bool "Use Write-Back cache policy for FLASH regions"
+	default n
```
JarmouniA:
```suggestion
```

arch/arm/core/mpu/Kconfig
```diff
@@ -79,6 +79,20 @@ config ARM_MPU_PXN
...
+	help
+	  When enabled, FLASH regions will use Write-Back cache policy
+	  instead of the default Write-Through policy.
+
+config ARM_MPU_SRAM_WRITE_THROUGH
+	bool "Use Write-Through cache policy for SRAM regions"
+	default n
```
JarmouniA:
```suggestion
```

arch/arm/core/mpu/arm_mpu_regions.c
```diff
@@ -9,27 +9,39 @@
 
 #include <zephyr/arch/arm/mpu/arm_mpu_mem_cfg.h>
 
+#ifdef CONFIG_ARM_MPU_FLASH_WRITE_BACK
+#define FLASH_REGION_ATTR REGION_FLASH_WB_ATTR
```
JarmouniA:
```suggestion
#define ARM_MPU_FLASH_REGION_ATTR REGION_FLASH_WB_ATTR
```

arch/arm/core/mpu/arm_mpu_regions.c
```diff
@@ -9,27 +9,39 @@
...
+#else
+#define FLASH_REGION_ATTR REGION_FLASH_ATTR
+#endif
+
+#ifdef CONFIG_ARM_MPU_SRAM_WRITE_THROUGH
+#define SRAM_REGION_ATTR REGION_RAM_WT_ATTR
```
JarmouniA:
```suggestion
#define ARM_MPU_SRAM_REGION_ATTR REGION_RAM_WT_ATTR
```

arch/arm/core/mpu/Kconfig
lucien-nxp: Updated.

arch/arm/core/mpu/Kconfig
lucien-nxp: Updated.

arch/arm/core/mpu/arm_mpu_regions.c
lucien-nxp: Updated.

arch/arm/core/mpu/arm_mpu_regions.c
lucien-nxp: Updated.

hakehuang:
> Current approach does not modify any default configurations. If users have specific requirements for cache policies, they can change them through overlay or hard coding methods to avoid the need for multiple regression tests to cover the supported cases.

I would suggest add one test for this, maybe mpu_test with WT for mimxrt1180

tests/arch/arm/arm_mpu_wt/src/main.c
```diff
@@ -0,0 +1,526 @@
...
+	uint32_t read_value = *test_addr;
+
+	zassert_equal(read_value, test_value,
+		      "User mode access failed: expected 0x%08x, got 0x%08x",
+		      test_value, read_value);
+
+	/* Test multiple patterns in user mode */
+	for (int i = 0; i < TEST_PATTERN_COUNT; i++) {
+		*test_addr = test_patterns[i];
+		barrier_dsync_fence_full();
```
hakehuang: I do not thing it is good if we already have a (WT)cache system, and we need resync cache every time.

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm33.overlay
```diff
@@ -0,0 +1,11 @@
+/*
```
hakehuang, line 1: need demo how to use mem_attr to set the cache policy

lucien-nxp: I have added CONFIG_ARM_MPU_SRAM_WRITE_THROUGH in prj.config default. Not all platforms will set the MPU attribute in DTS, and the attributes of the sram/flash nodes will be configured by the array in mpu_region.c based on the existing framework of Zephyr. Therefore, this case only tests the wt attribute of the sdram node.

tests/arch/arm/arm_mpu_wt/src/main.c
lucien-nxp: Agreed. I'll update this with any other changes if there's more feedback.

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm33.overlay
hakehuang: the problem is that, in many user case, we do not need to configure the whole sdram as WT, WB would be much prefer.

lucien-nxp: Shouldn't the addition be just a simple test case? I‘m not sure we need to be concerned about the specific usage scenarios of the users. This case is only for the default verification of the zephyr, sram node configuration into WT mode and the related tests. Moreover, in my patch, I have already provided REGION_RAM_WT_ATTR/DT_MEM_ARM_MPU_RAM_WT, which can be configured by the users in the DT.

hakehuang: > Shouldn't the addition be just a simple test case? I‘m not sure we need to be concerned about the specific usage scenarios of the users. This case is only for the default verification of the zephyr, sram node configuration into WT mode and the related tests. Moreover, in my patch, I have already provided REGION_RAM_WT_ATTR/DT_MEM_ARM_MPU_RAM_WT, which can be configured by the users in the DT.

You provide, but you have no sample, this is no complete work. so I suggest do it, and make this complete

include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h
```diff
@@ -40,6 +41,7 @@
 #define DT_MEM_ARM_MPU_EXTMEM      DT_MEM_ARM(ATTR_MPU_EXTMEM)
 #define DT_MEM_ARM_MPU_RAM_PXN     DT_MEM_ARM(ATTR_MPU_RAM_PXN)
 #define DT_MEM_ARM_MPU_DEVICE      DT_MEM_ARM(ATTR_MPU_DEVICE)
+#define DT_MEM_ARM_MPU_RAM_WT      DT_MEM_ARM(ATTR_MPU_RAM_WT)
```
wearyzen, line 44: could you rename this to something that matches with the attributes mentioned [here](https://developer.arm.com/documentation/107565/0101/Memory-protection/Attribute-indirection/Memory-Attribute-Indirection-Register--MAIR0-MAIR1-?lang=en)?
either DT_MEM_ARM_MPU_RAM_OUTER_WT_NONTRANSIENT or DT_MEM_ARM_MPU_RAM_OUTER_WT_TRANSIENT
or DT_MEM_ARM_MPU_RAM_INNER_WT_NONTRANSIENT or DT_MEM_ARM_MPU_RAM_INNER_WT_TRANSIENT

lucien-nxp: "memory-attr-arm.h": I understand this file that a more abstract layer of expression can cover all architectures, but the link you provided only includes the ARMv8M architecture. For ARMv7, some attributes cannot be configured. I think it may not align with the original intention behind the design of this floor, this layer mainly provides some common and optional configurations, so that users do not need to worry about the architecture issues.

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm33.overlay
lucien-nxp: I have added code to support DT configuration. CM33 and CM7 use different ways to configuration memory with WT attribution.

include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h
wearyzen: You are right.
The problem is that this layer has I think 12 bits for attributes and with current design it gives us 12 options, 8 of which would have been taken after this PR, so I was thinking of redesigning it as per the link I shared.
The naming suggestion was to avoid any breaking change for the macro users but I think I can figure that out later. I am ok with this name.

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm7.config
```diff
@@ -0,0 +1 @@
+CONFIG_ARM_MPU_SRAM_WRITE_THROUGH=y
```
wearyzen: I think this can be moved to either prj.conf or testcase.yaml instead of defining for each board.

wearyzen (commented):
I think the test could be made simpler.
Important part is to check if memory attributes are set correctly and if WT works as expected (either by volatile without a sync or a dma) but I am not sure if subtests like userspace etc. are adding any value here.
Also, does this run in Zephyr ci since I don't see any qemu/fvp boards as allowed? if not then should we even have this test?

arch/arm/core/mpu/Kconfig
```diff
@@ -79,6 +79,12 @@ config ARM_MPU_PXN
 	  unprivileged mode and executing such region from privileged mode will
 	  result in a Memory Management fault.
 
+config ARM_MPU_SRAM_WRITE_THROUGH
+	bool "Use Write-Through cache policy for SRAM regions"
+	help
```
butok, line 84: Should it be enabled by default?
Just asking ;)

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm7.config
lucien-nxp: Hake suggest we need also to provide a solution for some regions defined by DT, `CONFIG_ARM_MPU_SRAM_WRITE_THROUGH` only work for zephyr,sram node. So I set `CONFIG_ARM_MPU_SRAM_WRITE_THROUGH` in [mimxrt1180_evk_mimxrt1189_cm7.config](https://github.com/zephyrproject-rtos/zephyr/pull/99238/files/2b3b60eb3c28610e289b15ddfb825f46beabe72f#diff-e74fbd22e2bb0570c163e9097ffa658f99ff90fb9381ca460e823fa7f7cc1675), and show DT mpu settings for cm33 core

arch/arm/core/mpu/Kconfig
lucien-nxp: No, we only provide another option for sram node mpu settings if needed like some middleware's requirement. Actually, WRITE_THROUGH way will reduce performance, but for some multi-bus applications, developers don't have to worry about the maintenance of the cache.

lucien-nxp:
> Important part is to check if memory attributes are set correctly and if WT works as expected (either by volatile without a sync or a dma) but I am not sure if subtests like userspace etc. are adding any value here.

1. I agree the userspace and stack tests don't add significant value for WT verification, now I only keep three test unit to test cache configuration.
2. I have added qemu_cortex_m7 in testcase.yaml.

hakehuang (dismissed):
board testing pass in v4.3.0-2109-gaffd08466091

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm7.config
wearyzen: as per this: https://github.com/zephyrproject-rtos/zephyr/pull/99238/changes#diff-707133542f0760fc2fcd52acd7a24bc7a6eb2c03da27cb3c544ef1998586987eR11

the testcase runs only on the 2 allowed platforms so, moving the config to testcase.yml like below shouldn't change the current behavior.
```
    extra_configs:
      - CONFIG_ARM_MPU_SRAM_WRITE_THROUGH=y
```

tests/arch/arm/arm_mpu_wt/src/main.c
```diff
@@ -0,0 +1,548 @@
...
+#endif
+
+	LOG_INF("DMA transfer completed successfully");
+	return 0;
+}
+
+/**
+ * @brief Test basic memory region read/write functionality
+ */
+ZTEST(arm_mpu_wt, test_sram_basic_access)
```
wearyzen: I think this (writing and reading test pattern) is kind of covered as part of the other 2 testcases so can be removed as well.

tests/arch/arm/arm_mpu_wt/boards/mimxrt1180_evk_mimxrt1189_cm7.config
lucien-nxp: Agreed. Thank you, I have added it.

tests/arch/arm/arm_mpu_wt/src/main.c
lucien-nxp: Thank you, I have removed this test unit.

wearyzen (dismissed):

tests/arch/arm/arm_mpu_wt/src/main.c
```diff
@@ -0,0 +1,518 @@
...
+#define TEST_MEMORY_NODE DT_ALIAS(test_memory)
+#define TEST_MEMORY_BASE DT_REG_ADDR(TEST_MEMORY_NODE)
+#define TEST_MEMORY_REGION_SIZE DT_REG_SIZE(TEST_MEMORY_NODE)
+#define USE_CUSTOM_MEMORY 1
+#else
+#define USE_CUSTOM_MEMORY 0
+#endif
+
+static const uint32_t test_patterns[] = {
+	0x12345678, 0xDEADBEEF, 0xCAFEBABE, 0x87654321,
```
JarmouniA: As indicated by SonarQube,
"Add a "U" suffix to this constant, to make them explicitly "unsigned"."

tests/arch/arm/arm_mpu_wt/testcase.yaml
```diff
@@ -0,0 +1,15 @@
...
+  arch.arm.mpu.write_through:
+    tags:
+      - arm
+      - mpu
+      - cache
+      - dma
+    arch_allow:
+      - arm
+    filter: CONFIG_ARM_MPU
+    platform_allow:
```
JarmouniA:
```suggestion
    integration_platforms:
```

tests/arch/arm/arm_mpu_wt/testcase.yaml
```diff
@@ -0,0 +1,15 @@
+tests:
+  arch.arm.mpu.write_through:
+    tags:
+      - arm
+      - mpu
+      - cache
+      - dma
+    arch_allow:
+      - arm
+    filter: CONFIG_ARM_MPU and dt_nodelabel_enabled("test-dma")
```
JarmouniA:
```suggestion
    filter: CONFIG_ARM_MPU and dt_alias_exists("test-dma")
```
test-dma is an alias not a node label.
https://docs.zephyrproject.org/latest/develop/test/twister.html#expressions

JarmouniA (changes requested):

tests/arch/arm/arm_mpu_wt/testcase.yaml
lucien-nxp: Yes, thank you for reminding. I have corrected it.

JarmouniA (dismissed):

JarmouniA (approved):

wearyzen (approved):

PR #99243, opened 2025-11-12
More for lm20 ns

nordic-piks:
Add required overlay for execution at ns.


nordic-segl (approved):

tests/drivers/i2s/i2s_api/boards/nrf54lm20dk_nrf54lm20a_cpuapp_ns.overlay
nordic-krch, line 1: What about reusing content from `nrf54lm20dk_nrf54lm20a_cpuapp.overlay` to avoid duplication. Maybe common file that is included by both? Same for other file.

nordic-piks: Done, thanks.

anangl (approved):

kl-cruz (approved):

nika-nordic (approved):

PR #99251, opened 2025-11-12
boards: evk_iris_w1_rw612 flash config fix

3rang:
Fixed flash_config.c to properly handle two different board variants; previously both used the same configuration.
No major impact.

RFC [#99236.](https://github.com/zephyrproject-rtos/zephyr/discussions/99236)


boards/u-blox/ubx_evk_iris_w1/fidelix_flash_config.c
```diff
@@ -1,6 +1,8 @@
 /*
  * Copyright (c) 2021-2024 NXP
  *
+ * Copyright (c) 2025 u-blox AG
+ *
```
nordicjm, line 5: this file has not changed?

boards/u-blox/ubx_evk_iris_w1/revision.cmake
```diff
@@ -2,13 +2,13 @@
...
+  set(BOARD_REVISION fidelix CACHE STRING "Board revision")
 endif()
 
 # Validate revision
-if(NOT BOARD_REVISION STREQUAL "macronix" AND NOT BOARD_REVISION STREQUAL "fidelex")
+if(NOT BOARD_REVISION STREQUAL "macronix" AND NOT BOARD_REVISION STREQUAL "fidelix")
   message(FATAL_ERROR
     "Invalid BOARD_REVISION: ${BOARD_REVISION}\n"
-    "Must be one of: macronix, fidelex"
+    "Must be one of: macronix, fidelix"
```
nordicjm, line 14: should add the old one to `boards/deprecated.cmake` so it doesn't break existing users

boards/u-blox/ubx_evk_iris_w1/fidelix_flash_config.c
3rang: No changes to the file, only the spelling corrected.

boards/u-blox/ubx_evk_iris_w1/revision.cmake
3rang: only 2 variant of flash, and here spelling fixed.

nordicjm: sure, but the old name is in 2 zephyr releases so existing users would be broken by this change, therefore the old one should be added to the deprecated file so it automatically updates and users are warned

3rang: Makes sense. I’ll add ````deprecated.cmake```` ; it won’t break older releases.

boards/u-blox/ubx_evk_iris_w1/board.yml
```diff
@@ -4,9 +4,10 @@ board:
   vendor: u-blox
   revision:
     format: custom
-    default: fidelex
+    default: fidelix
     revisions:
       - name: "macronix"
-      - name: "fidelex"
+      - name: "fidelix"
+      - name: "fidelex"  # Deprecated: typo, redirects to fidelix
```
nordicjm:
```suggestion
```

boards/u-blox/ubx_evk_iris_w1/deprecated.cmake
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2025 u-blox AG
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Handle typo in board revision name
+if(BOARD_REVISION STREQUAL "fidelex")
```
nordicjm: this is not correct

boards/u-blox/ubx_evk_iris_w1/revision.cmake
nordicjm: this is not correct, look at `boards/deprecated.cmake` and put the old name there and the new name

nordicjm (changes requested):

boards/u-blox/ubx_evk_iris_w1/board.yml
3rang: Corrected.

boards/u-blox/ubx_evk_iris_w1/deprecated.cmake
3rang: updated in main ``boards/deprecated.cmake``

boards/deprecated.cmake
```diff
@@ -64,3 +64,6 @@ set(esp32s3_devkitm/esp32s3/procpu_DEPRECATED
 set(esp32s3_devkitm/esp32s3/appcpu_DEPRECATED
     esp32s3_devkitc/esp32s3/appcpu
 )
+set(ubx_evk_iris_w1_fidelex_DEPRECATED
+    ubx_evk_iris_w1@fidelix
```
nordicjm:
```suggestion
    ubx_evk_iris_w1@fidelix/rw612
```

boards/deprecated.cmake
```diff
@@ -64,3 +64,6 @@ set(esp32s3_devkitm/esp32s3/procpu_DEPRECATED
 set(esp32s3_devkitm/esp32s3/appcpu_DEPRECATED
     esp32s3_devkitc/esp32s3/appcpu
 )
+set(ubx_evk_iris_w1_fidelex_DEPRECATED
```
nordicjm:
```suggestion
set(ubx_evk_iris_w1@fidelex/rw612_DEPRECATED
```
please check this actually works

3rang: with suggested ```ubx_evk_iris_w1_fidelex/rw612_DEPRECATED```  will work, if we build with ```ubx_evk_iris_w1_fidelex/rw612``` but not with ```ubx_evk_iris_w1_fidelex``` .

do you think we must add qualifiers ?

3rang: ok , let me sync with standard , will update as you suggested.

3rang:
@nordicjm  , could you check my latest updates .

nordicjm (approved):

boards/u-blox/ubx_evk_iris_w1/fidelix_flash_config.c
kartben: claiming copyright over a file rename is a bit odd but oh well

kartben (approved):

PR #99254, opened 2025-11-12
xtensa: mmu: ptables: Set PPN in region_map_update to fix memory mapping

softwarecki:
Set Physical Page Number in L2 entry in the `region_map_update` function to ensure correct mapping for memory that was not previously mapped based on `xtensa_soc_mmu_ranges`.

When a new memory domain is created, the `arch_mem_domain_init` function allocates new ptables using the `dup_table` function. This function only copies original L2 entries created from `xtensa_soc_mmu_ranges` with `OPTION_SAVE_ATTRS`. Entries added later by `arch_mem_map` are replaced with `XTENSA_MMU_PTE_L2_ILLEGAL` and have PPN set to zero.

When adding a new partition to a memory domain, the `update_region` function is used. It sets new access attributes in the L2 table without setting the PPN value. As a result, it does not work correctly for memory that was not previously mapped based on `xtensa_soc_mmu_ranges`.

Extras: Add macro `XTENSA_MMU_PTE_PPN_GET` to simplify retrieval of the physical page number assigned to a page table entry.

### DIFF
--- a/arch/xtensa/core/ptables.c
+++ b/arch/xtensa/core/ptables.c
@@ -1227,6 +1227,7 @@ static void region_map_update(uint32_t *l1_table, uintptr_t start,
 		}
 
 		pte = l2_table[l2_pos];
+		pte = PTE_PPN_SET(pte, start + offset);
 
 		if ((option & OPTION_RESTORE_ATTRS) == OPTION_RESTORE_ATTRS) {
 			new_attrs = PTE_BCKUP_ATTR_GET(pte);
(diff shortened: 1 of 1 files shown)


serhiy-katsyuba-intel (approved):

dcpleung (dismissed):

tmleman (dismissed):

dcpleung (approved):

tmleman (approved):
Re-approving.

abonislawski (approved):

kartben:
> @dcpleung @nashif Can we merge this? It has been in review for quite a while and CI is all green.

it needed a second approval which only came 50 minutes ago :)

PR #99258, opened 2025-11-12
drivers: dma: stm32u5: Configure burst lengths

LinderPi:
The source and destination burst lengths can be configured with a value from 1 to 64. As the appropriate values are available in the DMA configuration structure provided by Zephyr, these can be set during driver configuration. Unfortunately, the STM32 DMA LL API does currently not provide the minimum and maximum values.

### DIFF
--- a/drivers/dma/dma_stm32u5.c
+++ b/drivers/dma/dma_stm32u5.c
@@ -25,6 +25,7 @@ LOG_MODULE_REGISTER(dma_stm32, CONFIG_DMA_LOG_LEVEL);
 #define DT_DRV_COMPAT st_stm32u5_dma
 
 #define STM32U5_DMA_LINKED_LIST_NODE_SIZE (2)
+#define STM32U5_DMA_MAX_BURST_LENGTH      (64) /* Maximum number of beats in a burst */
 
 static const uint32_t table_src_size[] = {
 	LL_DMA_SRC_DATAWIDTH_BYTE,
@@ -403,6 +404,44 @@ static int dma_stm32_configure(const struct device *dev,
 		LOG_ERR("source and dest unit size error, %d",
 			config->source_data_size);
 		return -EINVAL;
+	}
+
+	if ((config->source_burst_length % config->source_data_size) != 0) {
+		LOG_ERR("Source burst length %d is not aligned to source data size %d",
+			config->source_burst_length, config->source_data_size);
+		return -EINVAL;
+	}
+
+	if ((config->dest_burst_length % config->dest_data_size) != 0) {
+		LOG_ERR("Destination burst length %d is not aligned to destination data size %d",
+			config->dest_burst_length, config->dest_data_size);
+		return -EINVAL;
+	}
+
+	uint32_t burst_beats = config->source_burst_length / config->source_data_size;
+
+	if (burst_beats > STM32U5_DMA_MAX_BURST_LENGTH) {
+		LOG_ERR("Source burst length %d is invalid", config->source_burst_length);
+		return -EINVAL;
+	} else if (burst_beats > 0) {
+		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id), burst_beats);
+	} else {
+		/* Default HW behavior (upon reset) is a single beat burst */
+		LOG_WRN("Accepting source burst length 0 for backwards compatibility");
+		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id), 1U);
+	}
+
+	burst_beats = config->dest_burst_length / config->dest_data_size;
+
+	if (burst_beats > STM32U5_DMA_MAX_BURST_LENGTH) {
+		LOG_ERR("Destination burst length %d is invalid", config->dest_burst_length);
+		return -EINVAL;
+	} else if (burst_beats > 0) {
+		LL_DMA_SetDestBurstLength(dma, dma_stm32_id_to_stream(id), burst_beats);
+	} else {
+		/* Default HW behavior (upon reset) is a single beat burst */
+		LOG_WRN("Accepting destination burst length 0 for backwards compatibility");
+		LL_DMA_SetDestBurstLength(dma, dma_stm32_id_to_stream(id), 1U);
 	}
 
 	stream->busy		= true;


drivers/dma/dma_stm32u5.c
```diff
@@ -484,6 +484,16 @@ static int dma_stm32_configure(const struct device *dev,
 
 	LL_DMA_SetBlkDataLength(dma, dma_stm32_id_to_stream(id), config->head_block->block_size);
 
+	if (config->source_burst_length > 0 && config->source_burst_length <= 64) {
+		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id),
+					 config->source_burst_length);
+	}
```
etienne-lms: Error out if burst length > 64?
Or maybe a warning trace message that burst length was not applied.

LinderPi: Good point. I added warning log messages and rebased the branch.

erwango: @LinderPi Thanks for sharing this. Is that fixing performances issues that you noticed ? If not, please open an issue as well with your observations.

LinderPi: Not entirely. I need to further investigate it and will open an issue 👍

drivers/dma/dma_stm32u5.c
mathieuchopstm, line 1: I would place the check (and error if invalid) in `dma_stm32_configure`, as done with other parameters - somewhere around here:
https://github.com/zephyrproject-rtos/zephyr/blob/8cdce672767153a70ea8d18c231427790db0c2aa/drivers/dma/dma_stm32u5.c#L388-L393

LinderPi: I somewhat agree. However, I suppose this breaks existing applications that did not set the burst lengths. If this is desired, I change it.

mathieuchopstm: Why would it break applications? Just accept `0` as a valid (but ignored) value, since it ought to be the default value.

LinderPi: You are right. I implemented this and left the warning message if the value is ignored.

drivers/dma/dma_stm32u5.c
```diff
@@ -405,6 +405,26 @@ static int dma_stm32_configure(const struct device *dev,
...
 	}
 
+	if (config->source_burst_length > 64) {
+		LOG_ERR("Source burst length %d is invalid", config->source_burst_length);
+		return -EINVAL;
+	} else if (config->source_burst_length > 0) {
+		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id),
+					 config->source_burst_length);
+	} else {
+		LOG_WRN("Source burst length %d is not applied", config->source_burst_length);
```
mathieuchopstm: 1. We still need to `SetSrcBurstLength` because it may have been set to a different value by a previous call
2. Reword proposal
```suggestion
		/* Default HW behavior (upon reset) is single beat burst */
		LOG_WRN("Accepting source_burst_length=0 for backwards compatibility", config->source_burst_length);
		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id), 1U);
```

drivers/dma/dma_stm32u5.c
```diff
@@ -405,6 +405,26 @@ static int dma_stm32_configure(const struct device *dev,
...
+	}
+
+	if (config->dest_burst_length > 64) {
+		LOG_ERR("Destination burst length %d is invalid", config->dest_burst_length);
+		return -EINVAL;
+	} else if (config->dest_burst_length > 0) {
+		LL_DMA_SetDestBurstLength(dma, dma_stm32_id_to_stream(id),
+					  config->dest_burst_length);
+	} else {
+		LOG_WRN("Destination burst length %d is not applied", config->dest_burst_length);
```
mathieuchopstm: 1. We still need to `SetSrcBurstLength` because it may have been set to a different value by a previous call
2. Reword proposal
```suggestion
		/* Default HW behavior (upon reset) is single beat burst */
		LOG_WRN("Accepting dest_burst_length=0 for backwards compatibility", config->source_burst_length);
		LL_DMA_SetDestBurstLength(dma, dma_stm32_id_to_stream(id), 1U);
```

drivers/dma/dma_stm32u5.c
```diff
@@ -405,6 +405,26 @@ static int dma_stm32_configure(const struct device *dev,
 		return -EINVAL;
 	}
 
+	if (config->source_burst_length > 64) {
+		LOG_ERR("Source burst length %d is invalid", config->source_burst_length);
+		return -EINVAL;
+	} else if (config->source_burst_length > 0) {
+		LL_DMA_SetSrcBurstLength(dma, dma_stm32_id_to_stream(id),
+					 config->source_burst_length);
```
mathieuchopstm: The STM32 burst length is in number of "beats" (i.e., transfer unit size = `config->source_data_size`) whereas the API's burst length is in number of *bytes*. This should be taken into account with something like:

```c
if ((burst_beat_num % config->source_data_size) != 0) {
   LOG_ERR("Source burst length %d not aligned to source data size %d", /* ... */);
   return -EINVAL;
}

uint32_t burst_beat_num = config->source_burst_length / config->source_data_size;
if (burst_beat_num > 64) {
   /* error */
} else if (burst_beat_num > 0) {
  /* ... */
} else {
  /* default len=1 */
}
```

Same remark for `dest`.

(It may be worth to `LOG2(xxx_data_size)` beforehand for checks/computations)

drivers/dma/dma_stm32u5.c
```diff
@@ -405,6 +405,26 @@ static int dma_stm32_configure(const struct device *dev,
 		return -EINVAL;
 	}
 
+	if (config->source_burst_length > 64) {
```
mathieuchopstm: Define a constant for this:
```c
#define STM32U5_DMA_MAX_BURST_LEN 64 /* Max number of beats in a burst */
```
```suggestion
	if (config->source_burst_length > STM32U5_DMA_MAX_BURST_LEN) {
```

mathieuchopstm (commented):
also

drivers/dma/dma_stm32u5.c
LinderPi: Good point, I was not aware of this. Tested it on a STM32H573.

mathieuchopstm (approved):

erwango (approved):

PR #99260, opened 2025-11-12
drivers: gpio: nxp: fix gpio_mcux_lpc_configure()

### LINKED ISSUE #99255: nxp: lpc55: gpio_port_get_raw() returns wrong value for GPIO output pins
### Describe the bug

 gpio_port_get_raw() returns wrong value for a GPIO pin configured as output for LPC platforms: lpcxpresso55s06/16/28/69 

### Steps to reproduce

Simplest way to reproduce the issue:
- Add a call to gpio_pin_get_raw() inside the while loop of \zephyr\samples\basic\blinky\main.c. For example:
```
        int pin_value = gpio_pin_get_raw(led.port, led.pin); 
        printk ("Led pin raw level: %d\n", pin_value);
```
- And gpio_pin_get_raw() always returns 0 no matter if the GPIO LED is on or off.

NOTE: This is an LPC-specific issue as other tested platforms do not have this problem.

### Environment

Commit: dbcfb77ce3d20784406d4d73a1de3f7f1bcea07c

butok:
- Enables Digital mode for both input and output GPIO pins, in gpio_mcux_lpc_configure().
- Fixes https://github.com/zephyrproject-rtos/zephyr/issues/99255

### DIFF
--- a/drivers/gpio/gpio_mcux_lpc.c
+++ b/drivers/gpio/gpio_mcux_lpc.c
@@ -104,10 +104,9 @@ static int gpio_mcux_lpc_configure(const struct device *dev, gpio_pin_t pin,
 		*pinconfig |= IOCON_PIO_OD_MASK;
 	}
 
-	if ((flags & GPIO_INPUT) != 0) {
-		/* Set DIGIMODE bit */
-		*pinconfig |= IOCON_PIO_DIGIMODE_MASK;
-	}
+	/* Set DIGIMODE bit */
+	*pinconfig |= IOCON_PIO_DIGIMODE_MASK;
+
 	/* Select GPIO mux for this pin (func 0 is always GPIO) */
 	*pinconfig &= ~(IOCON_PIO_FUNC_MASK);
 #endif


drivers/gpio/gpio_mcux_lpc.c
```diff
@@ -104,10 +104,9 @@ static int gpio_mcux_lpc_configure(const struct device *dev, gpio_pin_t pin,
 		*pinconfig |= IOCON_PIO_OD_MASK;
 	}
 
-	if ((flags & GPIO_INPUT) != 0) {
-		/* Set DIGIMODE bit */
-		*pinconfig |= IOCON_PIO_DIGIMODE_MASK;
-	}
+	/* Set DIGIMODE bit */
+	*pinconfig |= IOCON_PIO_DIGIMODE_MASK;
+
```
decsny, line 109: this check was clearly intentional, so just removing it without justification is suspicious. do you have any insight into why it was there in the first place.

decsny (dismissed):

drivers/gpio/gpio_mcux_lpc.c
dleach02: this came from @danieldegrasse when he initially added pinmuxing control to gpio driver. His commit doesn't go into any additional details as to why (f8828a443404bd7ef0b7a)

danieldegrasse: If I had to guess, I made this check based on the MCUX config tool- maybe it only sets DIGIMODE when the pin is configured as a digital input? I'm not sure though

In theory you should only need to readback the pin value when it is set as an input, but from the LPC55S69 RM I don't see any reason to keep DIGIMODE clear here- you need to clear it when using analog functions, but those are selected via pinctrl.

RM suggests DIGIMODE should generally be set, so I think this is a valid change

image

Will leave it to NXP folks to decide though

butok: > this check was clearly intentional, so just removing it without justification is suspicious. do you have any insight into why it was there in the first place.

Digital mode must be set for GPIO Operations, as you need to read or write digital logic levels and enable digital input buffer.
Forgetting to enable DIGIMODE for digital functions may cause unreliable digital readings or communication failures. It was reproduced by https://github.com/zephyrproject-rtos/zephyr/issues/99255

danieldegrasse (approved):

dleach02 (approved):

drivers/gpio/gpio_mcux_lpc.c
HyperionCoding: > > this check was clearly intentional, so just removing it without justification is suspicious. do you have any insight into why it was there in the first place.
> 
> Digital mode must be set for GPIO Operations, as you need to read or write digital logic levels and enable digital input buffer. Forgetting to enable DIGIMODE for digital functions may cause unreliable digital readings or communication failures. It was reproduced by #99255

The real reason was incorrect way to read output bits. The user manual says the output should be read from SET registers:

image

Setting digital mode always on needs more justification and I believe it should not be used for outputs. This small note hints it should be on only for inputs:

image

In multiple other places the DIGIMODE is talked like it is synonymous to input mode.

A related but separate issue is that lpc pinctrl_soc.h files set digimode on simply if analog mode is not selected. That should be removed too.

@danieldegrasse

butok: > I believe it should not be used for outputs. 

We need to get a 'real' value on a GPIO pin pad, no matter if it's an input or output pin.

PR #99276, opened 2025-11-12
Describe display formats and various fixes (BGR565->RGB565X, RGB888, SDL driver)

avolmat-st:
Running the display sample on STM32 in RGB888 color leads to having red rectangle and blue rectangle swapped.
(RGB565 and ARGB8888 are correct).

Within the sample, the RGB888 red 0x00FF0000u is written into memory as
```c
        for (size_t idx = 0; idx < buf_size; idx += 3) {
                *(buf + idx + 0) = color >> 16;
                *(buf + idx + 1) = color >> 8;
                *(buf + idx + 2) = color >> 0;
        }
```
That is address 0: R, address 1: G then address 2: B

Having checked within LVGL but also other formats such as Linux DRM via https://github.com/afrantzis/pixel-format-guide tools, RGB888 is represented in memory as:

```
python3 -m pfg describe DRM_FORMAT_RGB888
Format:               DRM_FORMAT_RGB888
Component data type:  UNORM
Described as:         Bytes in memory
Memory little-endian: 0                1                2
                      M              L M              L M              L
                      B₇B₆B₅B₄B₃B₂B₁B₀ G₇G₆G₅G₄G₃G₂G₁G₀ R₇R₆R₅R₄R₃R₂R₁R₀
Memory big-endian:    0                1                2
                      M              L M              L M              L
                      B₇B₆B₅B₄B₃B₂B₁B₀ G₇G₆G₅G₄G₃G₂G₁G₀ R₇R₆R₅R₄R₃R₂R₁R₀
```

and for ARGB8888

```
python3 -m pfg describe DRM_FORMAT_ARGB8888
Format:               DRM_FORMAT_ARGB8888
Component data type:  UNORM
Described as:         Bytes in memory
Memory little-endian: 0                1                2                3
                      M              L M              L M              L M              L
                      B₇B₆B₅B₄B₃B₂B₁B₀ G₇G₆G₅G₄G₃G₂G₁G₀ R₇R₆R₅R₄R₃R₂R₁R₀ A₇A₆A₅A₄A₃A₂A₁A₀
Memory big-endian:    0                1                2                3
                      M              L M              L M              L M              L
                      B₇B₆B₅B₄B₃B₂B₁B₀ G₇G₆G₅G₄G₃G₂G₁G₀ R₇R₆R₅R₄R₃R₂R₁R₀ A₇A₆A₅A₄A₃A₂A₁A₀
```

In the display sample, the ARGB8888 corresponds well to this format, however not RGB888.

I think this is only a display sample issue since LVGL based program works well in RGB888 so the issue is only the data prepared within the display sample.
I've tested this on STM32 / LTDC based platform and this work well. It would be also great if it could be tested on some other platforms supporting RGB888 on display.

This PR corrects the RGB888 data within the sample and also add a description within the display header to avoid such future misunderstanding of the data ordering.


ngphibang:
Tested on native sim and RT1170, both works correctly. And logically, I think it should be fixed like that to have consistency between RGB565, RGB888 and ARGB8888 where the R channel is in the highest byte.

However, in Linux kernel doc, the `V4L2_PIX_FMT_RGB24` fomat is contradictory with this PR, it seems the doc is wrong ?
https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/pixfmt-packed-rgb.html

ngphibang (changes requested):
Need to be reconsidered as works wrongly on native sim

avolmat-st:
Let me update this PR with overall check of formats as well as SDL display mapping.

boards/lilygo/twatch_s3/Kconfig.defconfig
```diff
@@ -4,7 +4,7 @@
 if DISPLAY
 
 choice ST7789V_PIXEL_FORMAT
-	default ST7789V_BGR565
+	default ST7789V_RGB565X
```
faxe1008, line 7: Hmm not sold on the name  `RGBX`. Maybe something involving the endianess? The X does not carry any meaning

faxe1008 (commented):
This also needs a entry in the migration guide.

boards/lilygo/twatch_s3/Kconfig.defconfig
avolmat-st: I based this name on the format V4L2_PIX_FMT_RGB565X. This could also maybe be named RGB565_BE, aka the Big Endian version of RGB565. This latter name is used in libcamera for example in Linux.
Maybe this is better, what do you think ?

avolmat-st: In fact I totally forgot about it :D but this is also the name being used for such format on the video side:

```c
/**
 * 5 red bits [15:11], 6 green bits [10:5], 5 blue bits [4:0].
 * This 16-bit integer is then packed in big endian format over two bytes:
 *
 * @code{.unparsed}
 *   15.....8 7......0
 * | RrrrrGgg gggBbbbb | ...
 * @endcode
 */
#define VIDEO_PIX_FMT_RGB565X VIDEO_FOURCC('R', 'G', 'B', 'R')
```

It might be great to try to keep same naming overall.

avolmat-st: @ngphibang, @josuah,

trying to define a name on the display side to fix BGR565 pixel format (on display / panel side), which is in fact RGB565 in big endian ordering.
This is actually the same thing as what is currently exposed as VIDEO_PIX_FMT_RGB565X on video side, which is based on the V4L2_PIX_FMT_RGB565X.
Considering that in a further step, merge of display / video format is underway, what is your opinion about whether keep RGB565X (aka keeping same naming as V4L2) or adopting something like RGB565_BE instead maybe ?

Of course we tend to like keeping similar approach to existing frameworks / format descriptions. But at the same time differences will arise since for example RGB888 (v4l2) vs BGR888 (drm) should be avoided in order to make it easier to understand on the overall Zephyr code base so, I have the feeling that diverging a bit from the V4L2 / DRM spec is also ok if this help in understanding (and if we describe it properly)

josuah: > adopting something like RGB565_BE instead

If some naming leads to bugs, it sounds like enough justification to rename it to avoid that to keep happening.

Bugs are present through a lot of implementations for this particular format, which shows that the original naming is not optimal.

A comment saying "this is the same as V4L2 format XYZ" will help connecting things together.

boards/lilygo/twatch_s3/Kconfig.defconfig
```diff
@@ -4,7 +4,7 @@
 if DISPLAY
 
 choice ST7789V_PIXEL_FORMAT
-	default ST7789V_BGR565
```
JarmouniA, line 7: Pixel format symbols specific to a controller should not exist in the 1st place, should be configured in DT with standard display formats like the others. Working on a PR to fix this.

drivers/display/Kconfig.sdl
```diff
@@ -34,8 +34,8 @@ choice SDL_DISPLAY_DEFAULT_PIXEL_FORMAT
 	config SDL_DISPLAY_DEFAULT_PIXEL_FORMAT_RGB_565
 		bool "RGB 565"
 
-	config SDL_DISPLAY_DEFAULT_PIXEL_FORMAT_BGR_565
-		bool "BGR 565"
+	config SDL_DISPLAY_DEFAULT_PIXEL_FORMAT_RGB_565X
```
ngphibang: Here and line 34, should it be RGB565(X) without underscore to be coherent with the definition and other enums ?

include/zephyr/drivers/display.h
```diff
@@ -62,16 +62,16 @@ enum display_pixel_format {
 	 */
 	PIXEL_FORMAT_RGB_565		= BIT(4),
```
ngphibang: Could we take the chance to modify also the definition to avoid ambiguity ? i.e., drop "For example, in little-endian machine:" as it makes people think that in BE machine the organization will be different ... while the correct definition specifies that the byte 0 (lowest address) is always `gggBbbbb` and the byte 1 is always `RrrrrGgg` despite of LE or BE machine. I think the definition in `video.h` should also be modified a bit but it is anyways less confused than here.

drivers/mipi_dsi/Kconfig.mcux
```diff
@@ -23,7 +23,7 @@ config MIPI_DSI_MCUX_2L_SMARTDMA
 	depends on DMA_MCUX_SMARTDMA
 	help
 	  Use SMARTDMA. This accelerator will automatically
-	  convert RGB565 input data to BGR565 (little endian to big endian),
+	  convert RGB565 input data to RGB565X (little endian to big endian),
```
ngphibang: Could we drop `(little endian to big endian)` to avoid ambiguity ?

drivers/display/display_sdl.c
```diff
@@ -286,23 +318,31 @@ static void sdl_display_write_rgb565(uint8_t *disp_buf,
...
+ * SDL_PIXELFORMAT_BGRA32	Bbbbbbbb Gggggggg Rrrrrrrr Ffffffff
+ */
 static void sdl_display_write_rgb565x(uint8_t *disp_buf,
 				      const struct display_buffer_descriptor *desc, const void *buf)
 {
 	uint32_t w_idx;
 	uint32_t h_idx;
 	uint32_t pixel;
 	const uint16_t *pix_ptr;
+	uint16_t rgb565;
```
ngphibang, line 345: `uint16_t rgb565x;` to be clearer ?

drivers/display/display_sdl.c
```diff
@@ -236,15 +252,24 @@ static void sdl_display_write_rgb888(uint8_t *disp_buf,
...
 		}
 	}
 }
 
+/*
+ * Convert from			Byte 0   Byte 1   Byte 2   Byte 3
+ *				7......0 15.....8 23....16 31....24
+ * PIXEL_FORMAT_AL_88		Gggggggg Aaaaaaaa Gggggggg Aaaaaaaa
+ * into
+ * SDL_PIXELFORMAT_BGRA32	Bbbbbbbb Gggggggg Rrrrrrrr Ffffffff
```
ngphibang: `SDL_PIXELFORMAT_BGRA32	Bbbbbbbb Gggggggg Rrrrrrrr Aaaaaaaa`

drivers/display/display_sdl.c
```diff
@@ -311,15 +351,27 @@ static void sdl_display_write_rgb565x(uint8_t *disp_buf,
...
 		}
 	}
 }
 
+/*
+ * Convert from			Byte 0   Byte 1   Byte 2   Byte 3
+ *				7......0 15.....8 23....16 31....24
+ * PIXEL_FORMAT_MONO(01/10)	MMMMMMMM MMMMMMMM MMMMMMMM MMMMMMMM
+ * into
+ * SDL_PIXELFORMAT_BGRA32	Bbbbbbbb Gggggggg Rrrrrrrr Ffffffff
```
ngphibang, line 373: `SDL_PIXELFORMAT_BGRA32	MMMMMMMM MMMMMMMM MMMMMMMM Ffffffff` ?

drivers/display/display_sdl.c
```diff
@@ -343,12 +395,22 @@ static void sdl_display_write_mono(uint8_t *disp_buf, const struct display_buffe
...
 
+/*
+ * Convert from			Byte 0   Byte 1   Byte 2   Byte 3
+ *				7......0 15.....8 23....16 31....24
+ * PIXEL_FORMAT_L_8		Gggggggg Gggggggg Gggggggg Gggggggg
+ * into
+ * SDL_PIXELFORMAT_BGRA32	Bbbbbbbb Gggggggg Rrrrrrrr Ffffffff
+ *
+ * G value is applied to all R/G/B value for BGRA32
+ */
```
ngphibang, line 413: I think this comment is targeted read_l8(), for write_l8() it should be rewritten.

boards/lilygo/twatch_s3/Kconfig.defconfig
ngphibang: I think RGB565X naming is fine as X here stand for (eXchange a.k.a swapped) and we don't need to deviate from v4l2. I think we should avoid LE/BE in the name and also in the documentation as it is confused. In fact, the (RGB565 / BGR565 ...) formats are independent of LE or BE machine (byte 0 and byte 1 are always the same on both LE/BE machines).

ngphibang (commented):
The renaming BGR565 to RGB565X is correct as it is actually a byte swap not channel swap. And this does not change the logic (just the renaming) so don't need to test on all platforms I think. BTW, the 1st commit title and body should be rewritten to be coherent with the naming (RGB565X not RGB565_BE).

The actual (logic) fix is only in the sdl display driver (last commit).

I retested anyways on native sim and RT1170, it works fine. Thanks.

boards/lilygo/twatch_s3/Kconfig.defconfig
avolmat-st: Alright so, shall we keep the RGB565X naming ? I am fine with it.

boards/lilygo/twatch_s3/Kconfig.defconfig
avolmat-st: Should we try to split the issues here. Can we first fix the current issue addressed by this PR and then this ST7789V_PIXEL_FORMAT issue can be reworked in another PR that you would prepare and propose ?
Here regarding this ST7789V / boards etc this is simply a name replacement.

drivers/display/display_sdl.c
avolmat-st: Actually at this line I think it is better to keep the name rgb565 since as the comment says, swap is introduced on purpose so that same handling as for rgb565 can be applied. So what is stored into this variable is actually rgb565 and not rgb565x.

drivers/display/display_sdl.c
avolmat-st: The comment was first intended to repeat how the data should be generated actually. Into the BGRA32, that's B G R then forced FF which will be written (B / G and R being generated from the monochrome value)

boards/lilygo/twatch_s3/Kconfig.defconfig
JarmouniA: > Should we try to split the issues here

Yes that's fine.

ngphibang:
Even with the white background, it takes that long to fill the whole screen with white colors. Maybe because the eLCDIF driver allocates double buffers on the heap and does the memcpy (when the framebuffer size is not of the full display resolution 720p) ?

JarmouniA (dismissed):

ngphibang (dismissed):
LGTM. Thanks !

sylvioalves (changes requested):
I know that display driver is not stable yet, but there are changes in `display.h` that breaks current API.

At least I think migration notes update would be needed:
```
Display
=======

* The ``PIXEL_FORMAT_BGR_565`` pixel format has been renamed to
  :c:macro:`PIXEL_FORMAT_RGB_565X` to correctly reflect that it is a
  byte-swapped version of RGB_565, not a channel-swapped format.
  Applications using ``PIXEL_FORMAT_BGR_565`` must update to use
  ``PIXEL_FORMAT_RGB_565X``. (:github:`99276`)

* The devicetree macro ``PANEL_PIXEL_FORMAT_BGR_565`` has been renamed to
  ``PANEL_PIXEL_FORMAT_RGB_565X``. (:github:`99276`)

* The Kconfig options ``SDL_DISPLAY_DEFAULT_PIXEL_FORMAT_BGR_565`` and
  ``ST7789V_BGR565`` have been renamed to
  ``SDL_DISPLAY_DEFAULT_PIXEL_FORMAT_RGB_565X`` and ``ST7789V_RGB565X``
  respectively. (:github:`99276`)
```

JarmouniA (approved):

ngphibang (approved):

sylvioalves (approved):

samples/drivers/display/src/main.c
```diff
@@ -160,8 +161,29 @@ static void fill_buffer_mono(enum corner corner, uint8_t grey,
...
 {
+	uint8_t color;
+
+	switch (corner) {
+	case TOP_LEFT:
+		color = 0x00u;
+		break;
+	case TOP_RIGHT:
+		/* Use 0xE0 since 0xFF leads to drawing white corner on white back-ground */
+		color = 0xE0u;
```
VynDragon, line 172: I hadn't noticed that change, what is the rationale for choosing 0xE0, or even thinking about 0xFF? This should be 0xAA and the next one 0x55, or 0x3F and 0x7F, given one is going to be 0x0 and the rest of the screen 0xFF / this flipped

PR #99277, opened 2025-11-12
stm32/ltdc/dsi: correct properly use the pixel-format properties from ltdc node and mipi-dsi panel

avolmat-st:
So far the pixel-format properties from within the ltdc node is being set (mandatory in the yaml) but never used within the driver.
This is described in the yaml as the format used by the panel of the display-controller.
As a matter of fact, currently the LTDC controller only support RGB888 as output format of the controller (not to be confused with the format of data within the framebuffers, aka formats of the input layers of the LTDC).

The two first commits of this PR correct this, by checking that the pixel-format is well RGB888 and correct all formats written in dts.

Last commit correct the way MIPI-DSI is configured. Currently, the DSI format was set by looking at the LTDC format (which is in fact LTDC layer format) in order to set the ColorCoding of the MIPI-DSI. This is not correct since the ColorCoding has to be based on the format supported (and set) to the DSI panel.

### DIFF
--- a/drivers/display/display_stm32_ltdc.c
+++ b/drivers/display/display_stm32_ltdc.c
@@ -573,6 +573,11 @@ static DEVICE_API(display, stm32_ltdc_display_api) = {
 		frame_buffer_##inst[CONFIG_STM32_LTDC_FB_NUM * STM32_LTDC_FRAME_BUFFER_LEN(inst)];
 #endif
 
+/* LTDC supports RGB888 and RGB666 for output however only RGB_888 is supported for now */
+#if DT_INST_PROP(0, pixel_format) != PANEL_PIXEL_FORMAT_RGB_888
+#error "Only RGB_888 is supported as a LTDC output (aka panel or mipi-dsi input format)"
+#endif
+
 #define STM32_LTDC_DEVICE(inst)									\
 	STM32_LTDC_FRAME_BUFFER_DEFINE(inst);                       \
 	STM32_LTDC_DEVICE_PINCTRL_INIT(inst);							\
--- a/dts/bindings/display/lcd-controller.yaml
+++ b/dts/bindings/display/lcd-controller.yaml
@@ -10,5 +10,6 @@ properties:
     type: int
     required: true
     description: |
-      Initial Pixel format for panel attached to this controller.
-      See dt-bindings/display/panel.h for a list
+      Initial Pixel format of this controller's output data, which is sent
+      to a panel directly or to an adapter.
+      See dt-bindings/display/panel.h for a list of pixel format.
--- a/boards/ruiside/art_pi/art_pi.dts
+++ b/boards/ruiside/art_pi/art_pi.dts
@@ -195,7 +195,7 @@
 	clocks = <&rcc STM32_CLOCK(APB3, 3)>, <&rcc STM32_SRC_PLL3_R NO_SEL>;
 	width = <800>;
 	height = <480>;
-	pixel-format = <PANEL_PIXEL_FORMAT_RGB_565>;
+	pixel-format = <PANEL_PIXEL_FORMAT_RGB_888>;
 	def-back-color-red = <0X00>;
 	def-back-color-green = <0X00>;
 	def-back-color-blue = <0X00>;
--- a/boards/shields/arduino_giga_display_shield/boards/arduino_giga_r1_m7.overlay
+++ b/boards/shields/arduino_giga_display_shield/boards/arduino_giga_r1_m7.overlay
@@ -30,7 +30,7 @@
 	def-back-color-red = <0>;
 	def-back-color-green = <0>;
 	def-back-color-blue = <0>;
-	pixel-format = <PANEL_PIXEL_FORMAT_RGB_565>;
+	pixel-format = <PANEL_PIXEL_FORMAT_RGB_888>;
 
 	disp-on-gpios = <&gpioc 6 GPIO_ACTIVE_HIGH>;
 	bl-ctrl-gpios = <&gpiob 12 GPIO_ACTIVE_HIGH>;
--- a/boards/shields/st_lcd_dsi_mb1835/st_lcd_dsi_mb1835.overlay
+++ b/boards/shields/st_lcd_dsi_mb1835/st_lcd_dsi_mb1835.overlay
@@ -89,7 +89,7 @@
 	status = "okay";
 	width = <480>;
 	height = <480>;
-	pixel-format = <PANEL_PIXEL_FORMAT_ARGB_8888>;
+	pixel-format = <PANEL_PIXEL_FORMAT_RGB_888>;
 
 	bl-ctrl-gpios = <&dsi_lcd_qsh_030 53 GPIO_ACTIVE_HIGH>;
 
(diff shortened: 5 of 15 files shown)


danieldegrasse (commented):
I guess the panel internally converts the input data (regardless of format) to RGB888? Then the MIPI-DSI block can convert it to RGB-565 or similar?

Seems fine provided that the `display_get_capabilities` API is reporting the expected input buffer format

drivers/display/display_stm32_ltdc.c
```diff
@@ -561,6 +561,11 @@ static DEVICE_API(display, stm32_ltdc_display_api) = {
 		frame_buffer_##inst[CONFIG_STM32_LTDC_FB_NUM * STM32_LTDC_FRAME_BUFFER_LEN(inst)];
 #endif
 
+/* LTDC supports RGB888 and RGB666 for output however only RGB_888 is supported for now */
+#if DT_INST_PROP(0, pixel_format) != PANEL_PIXEL_FORMAT_RGB_888
+#error "Only RGB_888 is supported as a LTDC output (aka panel or mipi-dsi input format)"
+#endif
```
etienne-lms, line 579: Should this source file also assert `CONFIG_STM32_LTDC_RGB888` is  enabled?

By the way, lines 48 to 56 seem wrong, unless I missed something:
```patch
-#if CONFIG_STM32_LTDC_ARGB8888
+#if defined(CONFIG_STM32_LTDC_ARGB8888)
 #define STM32_LTDC_INIT_PIXEL_SIZE	4u
 #define STM32_LTDC_INIT_PIXEL_FORMAT	LTDC_PIXEL_FORMAT_ARGB8888
 #define DISPLAY_INIT_PIXEL_FORMAT	PIXEL_FORMAT_ARGB_8888
-#elif CONFIG_STM32_LTDC_RGB888
+#elif defined(CONFIG_STM32_LTDC_RGB888)
 #define STM32_LTDC_INIT_PIXEL_SIZE	3u
 #define STM32_LTDC_INIT_PIXEL_FORMAT	LTDC_PIXEL_FORMAT_RGB888
 #define DISPLAY_INIT_PIXEL_FORMAT	PIXEL_FORMAT_RGB_888
-#elif CONFIG_STM32_LTDC_RGB565
+#elif defined(CONFIG_STM32_LTDC_RGB565)
 # define STM32_LTDC_INIT_PIXEL_SIZE	2u
 #define STM32_LTDC_INIT_PIXEL_FORMAT	LTDC_PIXEL_FORMAT_RGB565
 #define DISPLAY_INIT_PIXEL_FORMAT	PIXEL_FORMAT_RGB_565
 #else
 #error "Invalid LTDC pixel format chosen"
 #endif

**Review by etienne-lms — COMMENTED** [review]
Commit "display: stm32-ltdc: clarify pixel-format property of lcd_controller"
should come after commit "boards: stm32: correct LTDC panel pixel-format to RGB888"
in your series unless what build would fail on that commit.

File: drivers/display/display_stm32_ltdc.c (hunk 1 above)
Comment (avolmat-st) [reply drivers/display/display_stm32_ltdc.c:579]: PANEL_PIXEL_FORMAT_RGB888 is related to the DSI format, while CONFIG_STM32_LTDC_* refer to the input format of the LTDC hence they can be different.

**Comment by JarmouniA** [comment]
> As a matter of fact, currently the LTDC controller only support RGB888 as output format of the controller (not to be confused with the format of data within the framebuffers, aka formats of the input layers of the LTDC).

This should be clarified in the description of the `pixel-format` property in the LTDC binding, because the base description is not that.
https://github.com/zephyrproject-rtos/zephyr/blob/9463d9a51d9cb1094bf98ef437a39850a7b5705d/dts/bindings/display/lcd-controller.yaml#L13

**Comment by JarmouniA** [comment]
> to me this is related to the pixel sent to the panel and not the data given to the controller itself.

I agree

> In the modification I am doing, I am using this pixel-format property for the format of data sent to the panel, as written in the YAML.

Okey, but in the case where LTDC is sending data to DSI block instead of the panel, the format set for LTDC can differ from the one set for the DSI device (panel controller), right? So the default binding description is not quite accurate in that case.

File: dts/bindings/display/lcd-controller.yaml [hunk 2]
```diff
@@ -10,5 +10,6 @@ properties:
     type: int
     required: true
     description: |
-      Initial Pixel format for panel attached to this controller.
-      See dt-bindings/display/panel.h for a list
+      Initial Pixel format for panel or adapter attached to this controller.
```
JarmouniA:
```suggestion
      Initial Pixel format of this controller's output data, which is sent to a panel directly or to an adapter.
```

dts/bindings/display/lcd-controller.yaml
```diff
@@ -10,5 +10,6 @@ properties:
...
+      This can be directly a panel but can also be an interface converter.
```
JarmouniA:
```suggestion
```

JarmouniA (approved):

etienne-lms (approved):

akscram:
@avolmat-st, I ran into the assert implemented in this PR. This assert checks that `pixel-format` has to be set to `PANEL_PIXEL_FORMAT_RGB_888`.

I have a display module with the ST7789VW display controller. This controller is connected to SPI and LTDC through DPI parallel RGB interface. It's configured through SPI during its initialization phase to work through parallel RGB interface. The RGB interface is driven by LTDC. The display controller and LTDC both have to be configured in the RGB 565 pixel format mode. This means that LTDC actually outputs RGB 565 through the DPI RGB interface.

The initial configured which I tried is the following (non-related options are omitted):
```ini
# Display
CONFIG_DISPLAY=y

#MIPI
CONFIG_MIPI_DBI_SPI_3WIRE=n
CONFIG_ST7789V_RGB565=y
```
And the following in DTS (non-related parts are omitted):
```c
/ {
    mipi_dbi {
        compatible = "zephyr,mipi-dbi-spi";
        spi-dev = <&spi1>;
        status = "okay";
        spi-dev = <&spi1>;
        ...
        st7789vw: st7789vw@0 {
            compatible = "sitronix,st7789v";
            status = "okay";
            mipi-mode = "MIPI_DBI_MODE_SPI_4WIRE";
            ...
        };
};
&ltdc {
    ...
    status = "okay";
    display-controller = <&st7789vw>;
    width = <240>;
    height = <320>;
    pixel-format = <PANEL_PIXEL_FORMAT_RGB_565>;
    ...
 };
```
However, a build failed with the following assert:
```
zephyr/drivers/display/display_stm32_ltdc.c:585:2: error: #error "Only RGB_888 is supported as a LTDC output (aka panel or mipi-dsi input format)"
```
This assert is confusing because the output format in my case is actually RGB 565 and LTDC supports this format through the RGB interface.

I had to either remove the assert, which turned to be working well, or I had to add the following configuration option:
```
CONFIG_STM32_LTDC_RGB565=y
```
And set `pixel-format` in my board's DTS to `<PANEL_PIXEL_FORMAT_RGB_888>` which is confusing because it contradicts to the actual output format. This assert seems to be not-needed at least when LTDC outputs through DPI RGB interface. It could be needed when LTDC outputs to MIPI-DSI host - this I don't know. What do you think?

PR #99284, opened 2025-11-13
arch: arm64: mmu: Call k_panic() when translation tables exhausted

kedareswararao:
When CONFIG_MAX_XLAT_TABLES is too small and new_table() cannot allocate
a translation table, the system must halt rather than continue with
undefined behavior.

This change ensures k_panic() is called after reporting the error,
preventing the system from proceeding when it runs out of translation
tables. Additionally, adds printk() fallback for configurations where
CONFIG_LOG is disabled to ensure the error is always visible.

### DIFF
--- a/arch/arm64/core/mmu.c
+++ b/arch/arm64/core/mmu.c
@@ -54,7 +54,15 @@ static uint64_t *new_table(void)
 		}
 	}
 
-	LOG_ERR("CONFIG_MAX_XLAT_TABLES, too small");
+#if defined(CONFIG_LOG)
+	LOG_ERR("CONFIG_MAX_XLAT_TABLES is too small");
+#else
+	printk("ERROR: CONFIG_MAX_XLAT_TABLES is too small\n");
+#endif
+
+	/* Unfortunately many code paths are not ready for failure */
+	k_panic();
+
 	return NULL;
 }
 


kedareswararao:
> It seems you only take care of static memory mapping but don't consider the dynamic mmu mapping (for example device_map in the driver), if yes, we could report the error during building by using `BUILD_ASSERT()`, it will be more useful to find the issue without any debugging.

Thanks @JiafeiPan for the review 

Thanks for the suggestion. I looked into the possibility of using BUILD_ASSERT() to validate CONFIG_MAX_XLAT_TABLES, but in this case it isn’t feasible below are my observations please correct if my understanding is wrong. 

BUILD_ASSERT() can only validate static, compile-time information. However, the number of translation tables required depends on runtime-only factors such as region alignment, fragmentation, and how the MMU code splits device-tree regions during set_mapping(). These details are not available at build time, so a static assertion would either give false confidence or block valid configurations.

The current approach performs validation at early boot, before MMU initialization, using the actual DT-derived regions and configuration. This catches issues reliably and prevents the silent crashes observed when table exhaustion happens later.

Motivation for this pull request:
On VersalNet APU, the failure occurs even without any device_map() usage. The platform has 20+ devices, and the final number of tables depends on runtime alignment and splits performed by the MMU layer. With the default CONFIG_MAX_XLAT_TABLES=12, this results in insufficient tables and early faults. Since the exact table count cannot be computed statically, runtime validation is necessary.

To improve robustness, I am planning to extend the logic that reserves tables for potential dynamic mappings by adding more granular, config-based heuristics. This will better cover platforms with variable peripheral layouts while still relying on runtime verification for correctness.

something like below 
#if defined(CONFIG_PCIE) || defined(CONFIG_ETH_E1000) || \
    defined(CONFIG_NVME) || defined(CONFIG_VIRTUALIZATION)
       dynamic_mapping_reserve = 5;  /* Reserve for runtime device_map() calls */
#else
       dynamic_mapping_reserve = 2;  /* Minimal reserve for potential drivers */
#endif

update the estimated_min accordingly 

 int estimated_min = base_overhead + (zephyr_regions / 2) +
                           (platform_regions / 2) + userspace_overhead +
                           dynamic_mapping_reserve;

Let me know if this direction looks good, and I can update the PR accordingly.

Regards,
Kedar.

npitre (commented):
I agree that silent crashes aren't nice.
But this is a bit too much guessing work for my taste.

What about this instead:

```
diff --git a/arch/arm64/core/mmu.c b/arch/arm64/core/mmu.c
index 0e3fbb27e60..e5edc6140a1 100644
--- a/arch/arm64/core/mmu.c
+++ b/arch/arm64/core/mmu.c
@@ -54,7 +54,15 @@ static uint64_t *new_table(void)
 		}
 	}
 
+#if defined(CONFIG_LOG)
 	LOG_ERR("CONFIG_MAX_XLAT_TABLES, too small");
+#else
+	printk("ERROR: CONFIG_MAX_XLAT_TABLES too small\n");
+#endif
+
+	/* Unfortunately many code paths are not ready for failure */
+	k_panic();
+
 	return NULL;
 }
 
```

npitre (dismissed):

arch/arm64/core/mmu.c
```diff
@@ -54,7 +54,15 @@ static uint64_t *new_table(void)
 		}
 	}
 
+#if defined(CONFIG_LOG)
 	LOG_ERR("CONFIG_MAX_XLAT_TABLES, too small");
+#else
+	printk("ERROR: CONFIG_MAX_XLAT_TABLES too small\n");
```
JiafeiPan: Unify the log both for LOG_ERR and printk to be "CONFIG_MAX_XLAT_TABLES is too small"?

kedareswararao: Thanks for the review sure will update

npitre (approved):

JiafeiPan (approved):

nashif (approved):

PR #97731, opened 2025-10-16
stm32: tests: drivers : fix mpu faults by enabling SRAM clocks and configuring MPU regions on stm32h7rsx

djiatsaf-st:
This pull request addresses multiple MPU-related faults observed during runtime tests on the `STM32H7RSX` 
series by introducing the following changes:

- Enable MPU regions for SRAM1 and SRAM2 access 
- Enables AHB2 peripheral clocks for SRAM1 and SRAM2

PR #96134, opened 2025-09-17
enable ethernet support on the stm32mp135f-dk

arnopo:
This Pull reuest enable the support of the ethernet 1 instance

Hardware details:
- the PHY is a LAN8742A with an Hardware reset based on a GPIO.
- the clock reference is an externl crystal running at 25 Mhz
- 16K of non cacheable memory in the  SYSRAM is used to store ethernet DMA Buffers and descriptors 
 
 Samples tested:
 - samples/net/ipv4_autoconf
 - samples/net/dhcpv4_client
 
 Known current limitation:
 - Ethernet2 instance not implemented 
 - sample/net/zperf not build
 
 Depends on:

- https://github.com/zephyrproject-rtos/hal_stm32/pull/320
- https://github.com/zephyrproject-rtos/zephyr/pull/96575

PR #99296, opened 2025-11-13
boards: st: stm32h7s78_dk: add Ethernet support

### LINKED ISSUE #99164: STM32H7S78-DK Ethernet support
### Is your feature request related to a problem? Please describe.

I wanted to implement Ethernet support for the STM32H7S78-DK board, which under Zephyr at the moment doesn't officially support Ethernet.

I based my work on the STM32H7S3L8 board (with adapted pins) and successfully tested Ethernet using the [dumb_http_server](https://github.com/zephyrproject-rtos/zephyr/issues/2927) sample.

Before opening a PR, I’d like feedback on one implementation detail.

The setup for Ethernet DMA in `mpu_regions.c` depends on `sram2`:
https://github.com/zephyrproject-rtos/zephyr/blob/91b1b84eee8159ce9706e8e0a3826cea04a09170/soc/st/stm32/stm32h7rsx/mpu_regions.c#L33-L42
But `sram2` is currently disabled, so the MPU regions are not created unless the user manually enables it:
https://github.com/zephyrproject-rtos/zephyr/blob/91b1b84eee8159ce9706e8e0a3826cea04a09170/dts/arm/st/h7rs/stm32h7rs.dtsi#L67-L77

There are two different ways of solving this, at least known to me at this moment:

1. Enabling `sram2` in [stm32h7s78_dk-common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi) and reducing the `sram0` by 16KB:
```
&sram0 {
	reg = <0x24000000 DT_SIZE_K(440)>;
};

...

&sram2 {
	status = "okay";
	zephyr,memory-attr = <DT_MEM_ARM(ATTR_MPU_RAM)>;
};
```

2. Replacing `sram2` with `sram1` here (or creating a fallback, if `sram2` is not available):
https://github.com/zephyrproject-rtos/zephyr/blob/91b1b84eee8159ce9706e8e0a3826cea04a09170/soc/st/stm32/stm32h7rsx/mpu_regions.c#L33-L42

I think this ties, at least partially, to this issue: #98866, since I needed to enable `CONFIG_SPEED_OPTIMIZATIONS=y` in order to avoid MPU FAULT.

If this is a good way to implement Ethernet support for this board, I will gladly open a PR with these and other needed changes to enable Ethernet.

Thanks in advance.

### Describe alternatives you've considered

_No response_

bklaric1:
## Changes

- Added `mac` and `mdio` configuration to `stm32h7s78_dk-common.dtsi` and enabled them. The RMII Ethernet pin configuration follows this schematic from ST: [MB1736](https://www.st.com/resource/en/schematic_pack/mb1736-h7s7l8-d01-schematic.pdf).
- ~Added fallback option for definition of `sram_eth_node` in `mpu_regions.c` and `sections.ld`. This definition earlier defaulted to `sram2`, but now has `sram1` as fallback option.~ ~Defined the SRAM region for Ethernet descriptor and buffer
using `memory-regions = <...>` in the `mac` node (as discussed) and added the updated definition of `sram_eth_node` to `mpu_regions.c` and `sections.ld`. Also added the needed property in `dts/bindings/ethernet/st,stm32h7-ethernet.yaml`.~
  - ~Added `memory-regions = <...>` to `boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts`, since they use the same `mpu_regions.c` and `sections.ld`.~
- Updated `stm32h7s78_dk.yaml` ~and `stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml`~ to include the Ethernet tag.
- Updated `index.rst` to include the Ethernet feature and documented the steps needed in order to use Ethernet for this board.
- Added `tests/drivers/memc/ram/boards/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.overlay` and updated `tests/drivers/memc/ram/boards/stm32h7s78_dk.overlay` since Ethernet drivers fully consume `sram1` node and thus fail the test without the additions.
- ~Added `#memory-region-cells = <0>;` to all possible memory regions (`sram1` and `sram2` nodes) in `dts/arm/st/h7rs/stm32h7rs.dtsi`~

## Tests

I tested these changes with the following examples:

- [dumb_http_server](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/net/sockets/dumb_http_server)
- [http_server](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/net/sockets/http_server)

I could ping the board after flashing both samples and also open the website in the browser. Everything worked as expected.

## Notes

I opened #99164 in order to discuss how to implement the fallback for `sram2` and whether this approach is correct. Taking a look at the discussion should answer some questions about the implementation.

Closes #99164

### DIFF
--- a/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
+++ b/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
@@ -11,6 +11,11 @@
 #include "zephyr/dt-bindings/display/panel.h"
 #include <zephyr/dt-bindings/input/input-event-codes.h>
 
+/*
+ * WARNING:
+ * JP6 must be in PC1 position when using Ethernet.
+ */
+
 / {
 	chosen {
 		zephyr,console = &uart4;
@@ -226,6 +231,31 @@
 	 * reference: DS14359 Rev 4 page 134
 	 */
 	status = "disabled";
+};
+
+&mac {
+	pinctrl-0 = <&eth_rmii_ref_clk_pd7
+		     &eth_rmii_crs_dv_pa7
+		     &eth_rmii_rxd0_pc4
+		     &eth_rmii_rxd1_pc5
+		     &eth_rmii_tx_en_pg11
+		     &eth_rmii_txd0_pb0
+		     &eth_rmii_txd1_pb1>;
+	pinctrl-names = "default";
+	phy-connection-type = "rmii";
+	phy-handle = <&eth_phy>;
+	status = "okay";
+};
+
+&mdio {
+	pinctrl-0 = <&eth_mdio_pa2 &eth_mdc_pc1>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	eth_phy: ethernet-phy@0 {
+		compatible = "ethernet-phy";
+		reg = <0x00>;
+	};
 };
 
 &xspi1 {
--- a/boards/st/stm32h7s78_dk/doc/index.rst
+++ b/boards/st/stm32h7s78_dk/doc/index.rst
@@ -181,6 +181,7 @@ Default Zephyr Peripheral Mapping:
 - XSPI1 NCS/DQS0/DQS1/CLK/IO: PO0/PO2/PO3/PO4/PP0..15
 - I2C1 SCL/SDA: PB6/PB9
 - FDCAN1 RX/TX : PB8/PB9
+- ETH: PA2/PA7/PB0/PB1/PC1/PC4/PC5/PD7/PG11
 
 System Clock
 ------------
@@ -207,6 +208,12 @@ FDCAN
 STM32H7S78-DK Discovery board has 2 FDCAN bus interfaces.
 FDCAN1 is configured but not enabled by default. To enable it, make sure
 that ``i2c1`` is disabled, since they share the PB9 pin.
+
+Ethernet
+--------
+
+In order to use Ethernet on STM32H7S78-DK, you need to set the ``JP6`` jumper
+to PC1 position on the back side of the board.
 
 Programming and Debugging
 *************************
--- a/boards/st/stm32h7s78_dk/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml
+++ b/boards/st/stm32h7s78_dk/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml
@@ -17,4 +17,5 @@ supported:
   - octospi
   - usbd
   - memc
+  - netif:eth
 vendor: st
--- a/tests/drivers/memc/ram/boards/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.overlay
+++ b/tests/drivers/memc/ram/boards/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.overlay
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 STMicroelectronics
+ * Copyright (c) 2025 Benjamin Klaric <benjamin.klaric01@gmail.com>
  *
  * SPDX-License-Identifier: Apache-2.0
  */


bklaric1:
The #99505 should be resolved first before proceeding with the merge (if it even gets to that point), since the changes to `mpu_regions.c` might become obsolete.

soc/st/stm32/stm32h7rsx/mpu_regions.c
```diff
@@ -31,7 +31,13 @@ static const struct arm_mpu_region mpu_regions[] = {
 	MPU_REGION_ENTRY("ID", 0x08FFF800, REGION_FLASH_ATTR(REGION_2K)),
 
 #if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-#define sram_eth_node DT_NODELABEL(sram2)
+
+#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(sram2))
+#define sram_eth_node	DT_NODELABEL(sram2)
+#else
+#define sram_eth_node	DT_NODELABEL(sram1)
+#endif
```
etienne-lms: I wonder is an explicit `memory-region = <...>` in the `mac` node would be better to tell which SRAM region is expected to hold ethernet buffers.

bklaric1: This indeed seems like a better way to do it, since the user can edit this in `stm32h7s78_dk-common.dtsi` directly and not in `\soc\st\stm32\stm32h7rsx\mpu_regions.c` and `sections.ld`, which reduces complexity of changing this.

I implemented and tested it as following, with the [http_server sample](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/net/sockets/http_server):

- Removed the referenced addition in `mpu_regions.c` as well as in `sections.ld`

- [stm32h7s78_dk-common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi):
  ```
  ...
  &mac {
	  pinctrl-0 = <&eth_rmii_ref_clk_pd7
			  &eth_rmii_crs_dv_pa7
			  &eth_rmii_rxd0_pc4
			  &eth_rmii_rxd1_pc5
			  &eth_rmii_tx_en_pg11
			  &eth_rmii_txd0_pb0
			  &eth_rmii_txd1_pb1>;
	  pinctrl-names = "default";
	  phy-connection-type = "rmii";
	  phy-handle = <&eth_phy>;
	  memory-region = <&sram1>; <------------- added
	  status = "okay";
  };
  ...
  ```

- [st,stm32h7-ethernet.yaml](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/ethernet/st%2Cstm32h7-ethernet.yaml):
  ```
  ...
  properties:
    memory-region:
      type: phandle
      required: true
      description: |
        Reference to SRAM node used for Ethernet DMA buffer and descriptor.
  ```
  Note that this addition is neccessary, since without it, I get the following error during build process:
  ```
  devicetree error: 'memory-region' appears in /soc/ethernet@40028000/ethernet in  
  <path_to_project>/build/zephyr/zephyr.dts.pre, but is not declared in 'properties:' in  
  <path_to>/zephyrproject/zephyr/dts/bindings\ethernet\st,stm32h7-ethernet.yaml
  ```

After adding both changes, the sample builds and flashes without errors and I can ping the board afterwards.

Should I add a new commit with this change then?

etienne-lms: I think it would be nice.
Note you likely need also to add `#memory-region-cells = <0>` in the referred memory node.
This is expected by Zephyr DT parsing scripts. See for example this discussion thread: https://github.com/zephyrproject-rtos/zephyr/pull/96134#discussion_r2448844120.

bklaric1: Thanks for the note and the discussion link. I will implement this and make a new commit tomorrow.

bklaric1: @etienne-lms I looked at this again, and it seems there are two different ways to do this:

1. Like in #96134:
  Define a memory area in `sram` for ethernet buffer and descriptor like this:
  https://github.com/zephyrproject-rtos/zephyr/blob/0c9d8b7a8e3b56c4255eafb667b26bcd7e757ac3/boards/st/stm32mp135f_dk/stm32mp135f_dk.dts#L83-L88
  and pass it with `memory-region = <...>` as the memory region for `mac`:
  https://github.com/zephyrproject-rtos/zephyr/blob/0c9d8b7a8e3b56c4255eafb667b26bcd7e757ac3/boards/st/stm32mp135f_dk/stm32mp135f_dk.dts#L300

2. Like I did it in the comment above:
  Pass the memory region directly:
    ```
    memory-region = <&sram1>;
    ````
    and define it inside `mpu_regions.c` and `sections.ld` like this:
    ```
    #define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_region)
    ```
Both require the addition of a property in [st,stm32h7-ethernet.yaml](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/ethernet/st%2Cstm32h7-ethernet.yaml) like this:
```
properties:
  memory-region:
    type: phandle <-------- only needed for 2. method
    required: true
    description: |
      Reference to SRAM node used for Ethernet DMA buffer and descriptor.
```

The 2. makes the changing, for example, `sram1` to `sram2` for ethernet buffer and descriptor much easier.

Also note that I could only get the 2. method to work for the stm32h7s78dk board.

If you could let me know which method is preferred, I can change my PR accordingly. Thanks.

bklaric1:
Regarding the failed checks: 

- compliance check fail (intendation and trailing whitespace) will be fixed with the next commit
- twister check error wasn't caused by changes that my commit introduced as far as I can say. The error output:
  ```
  /opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: zephyr/zephyr_pre0.elf section `SRAM1' will not fit in region `SRAM1'
  /opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: region `SRAM1' overflowed by 256 bytes
  ```
  states that the `SRAM1` region is overflowed. But the comment in the `mpu_regions.c` file states the following:
  https://github.com/zephyrproject-rtos/zephyr/blob/f1c2ce4f4026bd5b212c6e4cc7d754aba69b189e/soc/st/stm32/stm32h7rsx/mpu_regions.c#L39

So I think this isn't relevant to the PR, but I might be wrong.

Note: next commit will also add `- netif:eth` to [stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml), since I forgot to add that.

soc/st/stm32/stm32h7rsx/mpu_regions.c
etienne-lms: Case 2. looks  ok to me.

bklaric1: I pushed a commit with this implemented. Thanks for the insight.

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
```diff
@@ -12,3 +12,10 @@ description: |
 compatible: "st,stm32h7-ethernet"
 
 include: st,stm32-ethernet-common.yaml
+
+properties:
+  memory-region:
+    type: phandle
+    required: true
```
etienne-lms, line 18: I think the property should not be required. Other `"st,stm32h7-ethernet"` compatible nodes do not require a dedicated memory region: boards based on STM32N6, STM32H563 and STM32H7xx.

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
```diff
@@ -211,15 +216,16 @@
...
+		     &eth_rmii_crs_dv_pa7
+			 &eth_rmii_rxd0_pc4
+			 &eth_rmii_rxd1_pc5
+			 &eth_rmii_tx_en_pg11
+			 &eth_rmii_txd0_pb0
+			 &eth_rmii_txd1_pb1>;
 	pinctrl-names = "default";
 	phy-connection-type = "rmii";
 	phy-handle = <&eth_phy>;
+	memory-region = <&sram1>;
```
etienne-lms: Could you also add `memory-region = <&sram1>` in **boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts** that would need it.

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
```diff
@@ -211,15 +216,16 @@
...
-			&eth_rmii_rxd1_pc5
-			&eth_rmii_tx_en_pg11
-			&eth_rmii_txd0_pb0
-			&eth_rmii_txd1_pb1>;
+		     &eth_rmii_crs_dv_pa7
+			 &eth_rmii_rxd0_pc4
+			 &eth_rmii_rxd1_pc5
+			 &eth_rmii_tx_en_pg11
+			 &eth_rmii_txd0_pb0
+			 &eth_rmii_txd1_pb1>;
```
etienne-lms: Suggestion: either:
```c
	pinctrl-0 = <&eth_rmii_ref_clk_pd7
		     &eth_rmii_crs_dv_pa7
		     &eth_rmii_rxd0_pc4
		     &eth_rmii_rxd1_pc5
		     &eth_rmii_tx_en_pg11
		     &eth_rmii_txd0_pb0
		     &eth_rmii_txd1_pb1>;
```
or
```c
	pinctrl-0 = <&eth_rmii_ref_clk_pd7 &eth_rmii_crs_dv_pa7 &eth_rmii_rxd0_pc4
		     &eth_rmii_rxd1_pc5 &eth_rmii_tx_en_pg11 &eth_rmii_txd0_pb0
		     &eth_rmii_txd1_pb1>;

**Review by etienne-lms — COMMENTED** [review]
Could you squash to 2 commits?

File: boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi (hunk 4 above)
Comment (bklaric1) [reply boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi]: My bad. I forgot to apply the patch from the last CI check. Will be done in the next commit.

File: dts/bindings/ethernet/st,stm32h7-ethernet.yaml (hunk 2 above)
Comment (bklaric1) [reply dts/bindings/ethernet/st,stm32h7-ethernet.yaml:18]: As stated in the comment [above](https://github.com/zephyrproject-rtos/zephyr/pull/99296#discussion_r2538197249):
> [st,stm32h7-ethernet.yaml](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/ethernet/st%2Cstm32h7-ethernet.yaml):
> ```
>   ...
>   properties:
>     memory-region:
>       type: phandle
>       required: true
>       description: |
>         Reference to SRAM node used for Ethernet DMA buffer and descriptor.
>   ```
>   Note that this addition is neccessary, since without it, I get the following error during build process:
>   ```
>   devicetree error: 'memory-region' appears in /soc/ethernet@40028000/ethernet in  
>   <path_to_project>/build/zephyr/zephyr.dts.pre, but is not declared in 'properties:' in  
>   <path_to>/zephyrproject/zephyr/dts/bindings\ethernet\st,stm32h7-ethernet.yaml
>   ```
> 
> After adding both changes, the sample builds and flashes without errors and I can ping the board afterwards.

I couldn't get it to work without adding the property. Even the 1. method fails without it (even if it fails after building).

I agree that other boards don't require the property to use Ethernet.

I took a look at [stm32n6570_dk](https://github.com/zephyrproject-rtos/zephyr/tree/main/boards/st/stm32n6570_dk). It doesn't use the property, as you stated, but it also doesn't define `sram_eth_node` in [mpu_regions.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32n6x/mpu_regions.c) and [mpu_regions.ld](https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/st/stm32/stm32n6x/mpu_regions.ld), as the stm32h7s78_dk defines it.

Maybe I am missing something here?

File: boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi (hunk 3 above)
Comment (bklaric1) [reply boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi]: Maybe it is better to make a separate PR for this? So to not mix the s3 and s7 boards?
If not, I will add it in my next commit.

Comment (etienne-lms) [reply boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi]: Your change makes stm32h7rs SoCs to require a `memory-region` property to get the ethernet buffers RAM. nucleo_h7s3l8 board belongs to these SoC series. If not updated, it will fail. It should be updated by a commit before **soc/st/stm32/stm32h7rsx/mpu_regions.c** is updated, or in the very same commit.

File: dts/bindings/ethernet/st,stm32h7-ethernet.yaml (hunk 2 above)
Comment (etienne-lms) [reply dts/bindings/ethernet/st,stm32h7-ethernet.yaml:18]: It is the `required` directive that should be removed.

Comment (bklaric1) [reply dts/bindings/ethernet/st,stm32h7-ethernet.yaml:18]: Ohh. My bad, I misunderstood the sentance. I will remove it in the next commit.

File: boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi (hunk 3 above)
Comment (bklaric1) [reply boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi]: Got it. The upcoming commit will add this.

**Comment by bklaric1** [comment]
Squashed upcoming commit into 2. commit and added the requested changes.

**Comment by bklaric1** [comment]
Regarding the failed [Compliance Check](https://github.com/zephyrproject-rtos/zephyr/actions/runs/19698687522/job/56429386896?pr=99296): I don't see the trailing whitespace that is being reported, neither locally nor in GitHub Web Editor, so I can't really fix this.

**Comment by bklaric1** [comment]
@pdgendt I edited the first commit and removed the trailing whitespace there. Hopefully this fixes the CI check error.

**Comment by bklaric1** [comment]
I think the latter is the better choice here, because we don't want to introduce possible problems with skipping this test for `sram1` node.

I added this change as a third commit. I thought, since it adds changes to unrelated file (to ethernet feature addition) that it should be a separate commit. If wanted, I can squash it into the second commit.

**Comment by etienne-lms** [comment]
I forgot to ask a question when starting to review this P-R.:
Why do you want to use `sram1` node for ethernet access instead of `sram2` or done by the nucleo_h7s3l8 board (also stm32h7rs SoC based)? Maybe I missed something.

Digging into the implementation, I wonder if my suggestion to use a `memory-region` DT property is the right way. It may falsely suggest that a board can change this property to locate the ethernet buffers elsewhere. To do so, a bit more changes are needed. Current implementation rely on explicit mapping definition in the SoC specific **mpu_region.c** file.
I've sketched something to see how to handle that in a more flexible way. I'll create an RFC P-R to share.

**Comment by bklaric1** [comment]
I just set it to `sram1` node, since it is enabled by default. I can change it to `sram2` if that is what is prefered.

Yes, both ways this is being done, current implementation and my implementation in `mpu_regions.c` and `sections.ld`, are explicitly mapping ethernet drivers to either `sram1` or `sram2`. 

> I opened #99164 in order to discuss how to implement the fallback for `sram2` and whether this approach is correct. Taking a look at the discussion should answer some questions about the implementation.

As stated here, I myself am unsure if this approach is correct. 

In my opinion, defining what memory region is going to be used for ethernet drivers should be either:

- kept in `mpu_regions.c` and `sections.ld` but then an introduction of a fallback is needed, since by default the `sram2` node is disabled for this board, or explicitly tell the user that in order to use ethernet, the `sram2` node must be enabled (I originally spent quite a while to figure out that it's mapped to `sram2` but `sram2` is disabled by default &rarr; confusing for the user)
- kept like in the current implementation, but tell the user with an inline comment, that `sram1`/`sram2` node is being fully consumed by the ethernet drivers.

> The #99505 should be resolved first before proceeding with the merge (if it even gets to that point), since the changes to `mpu_regions.c` might become obsolete.

In any case, I think this should be addressed first, as using either `sram1` or `sram2` for ethernet drivers might introduce some unwanted problems in the long run:

In any case, I am open to shifting this PR to what your PR agrees on.

**Comment by bklaric1** [comment]
Added the requested change and squashed into one commit (with a long commit message).

File: soc/st/stm32/stm32h7rsx/mpu_regions.c [hunk 5]
```diff
@@ -31,7 +31,13 @@ static const struct arm_mpu_region mpu_regions[] = {
 	MPU_REGION_ENTRY("ID", 0x08FFF800, REGION_FLASH_ATTR(REGION_2K)),
 
 #if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-#define sram_eth_node DT_NODELABEL(sram2)
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_region)
+
+#if DT_NODE_HAS_COMPAT(sram_eth_node, zephyr_sram)
+#error "Invalid Ethernet memory-region: cannot use zephyr,sram. \
+Use a SRAM region like sram1 or sram2 for Ethernet."
```
etienne-lms:
```suggestion
Remove property memory-region from mac node to locate ethernet buffers in system RAM."

Comment (etienne-lms) [inline soc/st/stm32/stm32h7rsx/mpu_regions.c]:
```suggestion
BUILD_ASSERT(sram_eth_node != DT_CHOSEN(zephyr_sram)
	     "Remove property memory-region from 'mac' node to locate ethernet buffers in system RAM.")

File: soc/st/stm32/stm32h7rsx/mpu_regions.c [hunk 6]
```diff
@@ -31,7 +31,13 @@ static const struct arm_mpu_region mpu_regions[] = {
 	MPU_REGION_ENTRY("ID", 0x08FFF800, REGION_FLASH_ATTR(REGION_2K)),
 
 #if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-#define sram_eth_node DT_NODELABEL(sram2)
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_region)
```
etienne-lms:
```suggestion
#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac)) &&
    DT_NODE_HAS_PROP(DT_NODELABEL(mac), memory-region)
#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_region)
```

Ditto in **sections.ld**.

bklaric1:
@etienne-lms Amended the last commit to include the requested changes.

Question: do I need to add a (second) Copyright line when I edit an existing file or only when I create a new file?

soc/st/stm32/stm32h7rsx/mpu_regions.c
bklaric1: @etienne-lms Shouldn't `memory-region` be written as `memory_region`? I think this is what causes the twister-build (9) to fail.

soc/st/stm32/stm32h7rsx/mpu_regions.c
bklaric1: @etienne-lms Isn't the assert message exact opposite of what we want to say? The `memory-region` shouldn't be set to `zepyhr,sram` and removing it from `mac` causes the condition to be false and definition of `sram_eth_node` to be skipped, causing the following error:
```
/opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: warning: orphan section `.eth_stm32_desc' from `zephyr/drivers/ethernet/libdrivers__ethernet.a(eth_stm32_hal_common.c.obj)' being placed in section `.eth_stm32_desc'
/opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: warning: orphan section `.eth_stm32_buf' from `zephyr/drivers/ethernet/libdrivers__ethernet.a(eth_stm32_hal_common.c.obj)' being placed in section `.eth_stm32_buf'
```

Correct me if I'm wrong

bklaric1: @etienne-lms The `BUILD_ASSERT` as you wrote it, and any similar variation I tried (replace `sram_eth_node` with `DT_PHANDLE(DT_NODELABEL(mac), memory_region)`, check with `DT_NODE_HAS_COMPAT()` instead) don't pass the twister-builds. They fail with the following error:
```
In file included from /__w/zephyr/zephyr/include/zephyr/toolchain.h:52,
                 from /__w/zephyr/zephyr/include/zephyr/sys/util.h:18,
                 from /__w/zephyr/zephyr/include/zephyr/sys/list_gen.h:12,
                 from /__w/zephyr/zephyr/include/zephyr/sys/slist.h:25,
                 from /__w/zephyr/zephyr/soc/st/stm32/stm32h7rsx/mpu_regions.c:8:
/__w/zephyr/zephyr/include/zephyr/toolchain/gcc.h:87:36: error: expected expression before '_Static_assert'
   87 | #define BUILD_ASSERT(EXPR, MSG...) _Static_assert((EXPR), "" MSG)
      |                                    ^~~~~~~~~~~~~~
/__w/zephyr/zephyr/soc/st/stm32/stm32h7rsx/mpu_regions.c:37:1: note: in expansion of macro 'BUILD_ASSERT'
   37 | BUILD_ASSERT(sram_eth_node != DT_CHOSEN(zephyr_sram),
      | ^~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```
The suggested check with `#error ...`, that you suggested I replace with `BUILD_ASSERT`, worked.
I would appreciate if you could could review this again.

soc/st/stm32/stm32h7rsx/mpu_regions.c
bklaric1: Changing `memory-region` to `memory_region` **fixed** the problem where the condition was false and definition of `sram_eth_node` was skipped, causing the following error:
```
/opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: warning: orphan section `.eth_stm32_desc' from `zephyr/drivers/ethernet/libdrivers__ethernet.a(eth_stm32_hal_common.c.obj)' being placed in section `.eth_stm32_desc'
/opt/toolchains/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: warning: orphan section `.eth_stm32_buf' from `zephyr/drivers/ethernet/libdrivers__ethernet.a(eth_stm32_hal_common.c.obj)' being placed in section `.eth_stm32_buf'
```

soc/st/stm32/stm32h7rsx/mpu_regions.c
mathieuchopstm: Use `DT_SAME_NODE(a, b)` to check if two nodes are equivalent.

bklaric1: Thanks for the suggestion. I amended my last commit to include this. Let's see if it passes the twister-builds now.

bklaric1: @mathieuchopstm doesn't seem to fix the error reported by twister-builds (5) (6) (8) (9) (10).

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
```diff
@@ -12,3 +12,9 @@ description: |
 compatible: "st,stm32h7-ethernet"
 
 include: st,stm32-ethernet-common.yaml
+
+properties:
+  memory-region:
```
maass-hamburg: maybe have it consistent between the stm ethernet macs, and do it like in dts/bindings/ethernet/st,stm32mp13-ethernet.yaml and use `memory-regions` (with an `s`)

bklaric1: Missed the `s`. Will update.
But if we are trying to be consistent, shouldn't the property also include the `required: true` as in `st,stm32mp13-ethernet.yaml`?
https://github.com/zephyrproject-rtos/zephyr/blob/a1970ebd2f428dc472d505d7c7e5eee4774ee7b0/dts/bindings/ethernet/st%2Cstm32mp13-ethernet.yaml#L16-L22

maass-hamburg: Probably not, the `soc/st/stm32/stm32h7x` doesn't use it and we still want to be able to have it on the main ram by not setting it.
Implied by
```
BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
	     "Remove property memory-region from 'mac' node "
	     "to locate ethernet buffers in system RAM.")
```

But that's something the ST people (@etienne-lms )should decide

bklaric1: Added the `s` across the commit.

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
```diff
@@ -12,3 +12,9 @@ description: |
 compatible: "st,stm32h7-ethernet"
 
 include: st,stm32-ethernet-common.yaml
```
maass-hamburg: add
```
include:
  - "st,stm32-ethernet-common.yaml"
  - "memory-region.yaml"
```

and remove `type: phandle`

soc/st/stm32/stm32h7rsx/mpu_regions.c
```diff
@@ -30,8 +30,14 @@ static const struct arm_mpu_region mpu_regions[] = {
...
 
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-#define sram_eth_node DT_NODELABEL(sram2)
+#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac)) && \
+	DT_NODE_HAS_PROP(DT_NODELABEL(mac), memory_regions)
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
+
+BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
+	     "Remove property memory-regions from 'mac' node "
+	     "to locate ethernet buffers in system RAM.")
```
maass-hamburg: you could do
```
#if DT_NODE_HAS_STATUS_OKAY(sram_eth_node) && !DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram))
```
here and in the `sections.ld` file to get rid of the build assert

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
bklaric1: If I do that, then I get the following error:
```
devicetree error: <Node /memory@30000000 in /zephyr/dts/arm/st/h7rs/stm32h7rs.dtsi:60> lacks #memory-region-cells
```
That line corresponds to `sram1`:
https://github.com/zephyrproject-rtos/zephyr/blob/923ece1eea7d303d06b617bb85159832c6b9c12e/dts/arm/st/h7rs/stm32h7rs.dtsi#L60-L65

The memory region stm32mp series uses for Ethernet is custom defined region and not predefined region as `sram1` or `sram2`. And in that custom defined region the attribute `#memory-region-cells` is defined:
https://github.com/zephyrproject-rtos/zephyr/blob/923ece1eea7d303d06b617bb85159832c6b9c12e/boards/st/stm32mp135f_dk/stm32mp135f_dk.dts#L78-L83

What do you suggest here?
Note that I tried to get Ethernet working with custom region but wasn't successful.

soc/st/stm32/stm32h7rsx/mpu_regions.c
bklaric1: Removing assert and replacing it with a similar check works. For example, I already had this:
```
#if DT_NODE_HAS_COMPAT(sram_eth_node, zephyr_sram)
#error "Invalid Ethernet memory-region: cannot use zephyr,sram. \
Use a SRAM region like sram1 or sram2 for Ethernet."
```
but @etienne-lms suggested `BUILD_ASSERT` instead.

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
maass-hamburg: add `#memory-region-cells = <0>;` to all possible memory regions?

soc/st/stm32/stm32h7rsx/mpu_regions.c
maass-hamburg: then leave it

bklaric1: Okay. But what do I do about the failing twister-builds caused by the `BUILD_ASSERT`?

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
bklaric1: Adding it fixes the mentioned error and doesn't break anything else with this build (http_server sample). I added it to the `sram0`, `sram1` and `sram2` nodes.
However, I can't guarantee that this addition doesn't break something else unrelated to Ethernet. I can add it to the commit and see if it triggers any twister-build errors. Would that work?

soc/st/stm32/stm32h7rsx/mpu_regions.c
maass-hamburg: apply:
```
diff --git a/soc/st/stm32/stm32h7rsx/mpu_regions.c b/soc/st/stm32/stm32h7rsx/mpu_regions.c
index 5384d8bb2a6..f28fdf8b5b4 100644
--- a/soc/st/stm32/stm32h7rsx/mpu_regions.c
+++ b/soc/st/stm32/stm32h7rsx/mpu_regions.c
@@ -34,10 +34,6 @@ static const struct arm_mpu_region mpu_regions[] = {
 	DT_NODE_HAS_PROP(DT_NODELABEL(mac), memory_regions)
 #define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
 
-BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
-	     "Remove property memory-regions from 'mac' node "
-	     "to locate ethernet buffers in system RAM.")
-
 #if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
 	/* Region 5 - Ethernet DMA buffer RAM */
 	MPU_REGION_ENTRY("SRAM_ETH_BUF", DT_REG_ADDR(sram_eth_node),
@@ -48,6 +44,12 @@ BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
 #endif
 };
 
+#ifdef sram_eth_node
+	BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
+	     "Remove property memory-regions from 'mac' node "
+	     "to locate ethernet buffers in system RAM.");
+#endif
+
 const struct arm_mpu_config mpu_config = {
 	.num_regions = ARRAY_SIZE(mpu_regions),
 	.mpu_regions = mpu_regions,

```

you can't do a BUILD_ASSERT inside a 
`static const struct arm_mpu_region mpu_regions[] = {`

bklaric1: Done.

maass-hamburg (dismissed):

etienne-lms:
> Question: do I need to add a (second) Copyright line when I edit an existing file or only when I create a new file

You don't need to add a copyright line at source/header files entry when you modify them to have your copyright implicitly taken into account. What drives the copyright owners list is the Git reversion history, which tracks all contributions, not the file entry inline comment. 
You are free to add your copyright line, but it's not handy when each contributor adds its own line, especially for small modifications.

soc/st/stm32/stm32h7rsx/mpu_regions.c
```diff
@@ -42,6 +44,12 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 };
 
+#ifdef sram_eth_node
+	BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
```
etienne-lms: Indentation
```suggestion
BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),

File: dts/arm/st/h7rs/stm32h7rs.dtsi [hunk 11]
```diff
@@ -54,13 +54,15 @@
 		reg = <0x24000000 DT_SIZE_K(456)>;
 		compatible = "zephyr,memory-region", "mmio-sram";
 		zephyr,memory-region = "SRAM0";
+		#memory-region-cells = <0>;
```
etienne-lms: Can be removed. We don't expect one to reference `&sram0` as a `memory-region` phandle.

dts/arm/st/h7rs/stm32h7rs.dtsi
```diff
@@ -54,13 +54,15 @@
...
 	};
 
 	/* System data RAM accessible over AHB bus: SRAM1 in D2 domain */
 	sram1: memory@30000000 {
 		reg = <0x30000000 DT_SIZE_K(16)>;
 		compatible = "zephyr,memory-region", "mmio-sram";
 		zephyr,memory-region = "SRAM1";
+		#memory-region-cells = <0>;
 		zephyr,memory-attr = <DT_MEM_ARM(ATTR_MPU_RAM)>;
```
etienne-lms, line 65: From the change made in https://github.com/zephyrproject-rtos/zephyr/pull/97731, won't `sram1` MPU attributes here now conflicts with the one defined by **mpu_regions.c**?

boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
```diff
@@ -181,6 +181,7 @@
 	pinctrl-names = "default";
 	phy-connection-type = "rmii";
 	phy-handle = <&eth_phy>;
+	memory-regions = <&sram1>;
```
etienne-lms: Maybe use `sram2` to preserve to previous configuration?
Otherwise, the commit message should tell that h7rs based device ethnet buffers are now in `sram1` (AXI SRAM1) by default instead of `sram2` (AXI SRAM2).

TD-JBL: sram1 and sram2 are AHB, sram0 is AXI sram (called sram 1-4 in the datasheet/rm)

soc/st/stm32/stm32h7rsx/mpu_regions.c
bklaric1: Done.

dts/arm/st/h7rs/stm32h7rs.dtsi
bklaric1: Done.

boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
bklaric1: `sram2` is disabled by default. That is why I opted to use `sram1`.
I can switch to using `sram2` but that requires the node to be enabled by default. Otherwise an introduction of a fallback is needed (to `sram1`).
What should I do here?
I will update the commit message in any case. Thanks @TD-JBL for clarification.

dts/arm/st/h7rs/stm32h7rs.dtsi
bklaric1: That is partially what I was discussing with @TD-JBL in #99164.
I can only confirm that I get no errors during build process and Ethernet works as expected with either `sram1` or `sram2`.

boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
TD-JBL: Yes the sram2 is disabled to have users enable it on purpose:
1. Use as normal ram: enabled it in your board and set `zephyr,memory-attr = <DT_MEM_ARM(ATTR_MPU_RAM)>`
2. Use it as dma buffer for ethernet, enable it in your board and let `mpu_regions.c` configure the mpu-region for it.
I don't mind changing this behaviour, maybe have it already enabled... not sure whats be best way

dts/arm/st/h7rs/stm32h7rs.dtsi
TD-JBL: > From the change made in #97731, won't `sram1` MPU attributes here now conflicts with the one defined by **mpu_regions.c**?

Yes it will create 2 overlaying mpu regions for `sram1` and the one from `zephyr,memory-attr` has the higher priority and should cause problems with caching on the dma region, but for whatever reason it seems to work.

boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
maass-hamburg: I don't think that preserving behavior is that important for a dev board

erwango (changes requested):
See [comment for equivalent change on H7](https://github.com/zephyrproject-rtos/zephyr/pull/100910#pullrequestreview-3578364989).
Please apply the same for H7RS (configure memory in .dtsi files)

bklaric1:
@erwango I am going to deal with the requested changes suggested in #100910 first and then apply them to this PR.

bklaric1:
@erwango I removed the changes from this PR that have already been added by #101920 and adapted this PR to correspond to what was discussed in #101920.
Do I also need to rebase on main?

maass-hamburg (dismissed):

boards/st/stm32h7s78_dk/stm32h7s78_dk.yaml
```diff
@@ -16,4 +16,5 @@ supported:
   - octospi
   - usbd
   - memc
+  - netif:eth
```
erwango: Setting this will make net samples compiled in CI on this target by default.
Is it possible to build/run net samples on default `stm32h7s78_dk` target with its 64K of flash ?
If not, then its' preferred to avoid defining it and let tests happen on `ext_flash_app` variant

erwango (changes requested):
Thanks!

One comment, otherwise LGTM

boards/st/stm32h7s78_dk/stm32h7s78_dk.yaml
bklaric1: My mistake. Net samples can't run in internal flash; tested the `http_sample` and this is the flash usage: 
```
           FLASH:      369192 B        60 MB      0.59%
```
I will remove it.

juickar (approved):

erwango (approved):

PR #99301, opened 2025-11-13
docs: bindings-syntax: re-add note about `title` being optional

mathieuchopstm:
When originally introduced in ee17657ad3290465e8255c8c9da7efbd5343f789, `title` was described as optional, but this information was lost when the documentation was reworked in 3a2f839d459ae20597f1bc4af0e78d35a68b5dd4, even though the underlying behavior in `edtlib` is unchanged:
https://github.com/zephyrproject-rtos/zephyr/blob/9ef73c88f6bc4721277933f57f893f81989beb68/scripts/dts/python-devicetree/src/devicetree/edtlib.py#L187-L189

Add a note to specify that `title` is still optional (but recommended) and indicate that `description` should be short when `title` is not used.

### DIFF
--- a/doc/build/dts/bindings-syntax.rst
+++ b/doc/build/dts/bindings-syntax.rst
@@ -73,9 +73,9 @@ These keys are explained in the following sections.
 Title
 *****
 
-Short description of the bound device, typically the hardware model. It should
-typically be on the format "Vendor Family Model". If acronyms are used, they
-should be spelled out in parentheses. The naming should stay as close to the
+An *optional*, short description of the bound device, typically the hardware model.
+It should typically be of the format "Vendor Family Model". If acronyms are used,
+they should be spelled out in parentheses. The naming should stay as close to the
 vendor datasheet as possible.
 
 Titles should not exceed 100 characters. The description field should be used


jhedberg (dismissed):

doc/build/dts/bindings-syntax.rst
```diff
@@ -86,6 +86,10 @@ be used in the title, everything is a binding.
 
    title: Acme Foo UART (Universal Asynchronous Receiver/Transmitter)
 
+.. note::
+  The ``title`` key is recommended, but optional. If omitted, the ``description``
+  key is used as binding title instead, so it should be short in such situations.
```
kartben: that is not correct. If the title is None, it is None :) But in some situations (e.g. documentation) a "makeshift" title may be extracted from the _first sentence_ of the description

kartben (changes requested):

doc/build/dts/bindings-syntax.rst
mathieuchopstm: Does this sound better?
```suggestion
  The ``title`` key is recommended, but optional. If not provided, note that the
  first line of the ``description`` key may be used as a replacement for it in
  certain situations.
```
(I `s/sentense/line`'ed since I think the documentation parser behaves like that? but I haven't checked)

Or should I keep it simple and simply
```suggestion
  The ``title`` key is optional.
```

kartben: @mathieuchopstm It's definitely the first sentence (https://github.com/zephyrproject-rtos/zephyr/blob/77d461c4c0cefda2efb73f3de7bfe301b8e22762/doc/_scripts/gen_boards_catalog.py#L39) ...which also means that it can sometimes get it really wrong if there are periods in odd places etc.

And yes, simpler is better I think. How about "Short description of the bound device, ..." --> "An optional short description ..."

kartben (approved):

rruuaanng (approved):
:)

kartben (approved):

mathieuchopstm:
@mbolivar PTAL or re-assign the PR

mbolivar (approved):

PR #99304, opened 2025-11-13
drivers: regulator: add stm32 vrefbuf driver

tpambor:
This PR introduces a driver for the VREFBUF regulator available on STM32 microcontrollers. It also updates the device tree to include VREFBUF support for the STM32H5 and STM32U5 series.

Fixes #97249


drivers/regulator/regulator_stm32_vrefbuf.c
mathieuchopstm, line 1: Please use `LL_VREFBUF_xxx()` APIs from `#include <stm32_ll_system.h>` (I have annotated this all across the file)

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	struct regulator_common_data common;
+};
+
+struct regulator_stm32_vrefbuf_config {
+	struct regulator_common_config common;
+	VREFBUF_TypeDef *base;
+	const struct stm32_pclken pclken[1];
+	const struct reset_dt_spec reset;
+	bool vrefp_output_enable;
+	const struct regulator_stm32_vrefbuf_voltage *ref_voltages;
```
mathieuchopstm: For packing:
```suggestion
	const struct regulator_stm32_vrefbuf_voltage *ref_voltages;
	bool vrefp_output_enable;
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	uint8_t ref_voltage_count;
+};
+
+static int regulator_stm32_vrefbuf_enable(const struct device *dev)
+{
+	const struct regulator_stm32_vrefbuf_config *config = dev->config;
+
+	config->base->CSR = config->base->CSR | VREFBUF_CSR_ENVR;
```
mathieuchopstm:
```suggestion
	LL_VREFBUF_Enable();
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+
+	return 0;
+}
+
+static int regulator_stm32_vrefbuf_disable(const struct device *dev)
+{
+	const struct regulator_stm32_vrefbuf_config *config = dev->config;
+
+	config->base->CSR = config->base->CSR & ~VREFBUF_CSR_ENVR;
```
mathieuchopstm:
```suggestion
	LL_VREFBUF_Disable();
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	int32_t best_voltage = INT32_MAX;
+
+	for (uint8_t i = 0; i < config->ref_voltage_count; i++) {
+		if (!IN_RANGE(config->ref_voltages[i].uv, min_uv, max_uv) ||
+		    config->ref_voltages[i].uv >= best_voltage) {
+			continue;
+		}
+
+		best_voltage = config->ref_voltages[i].uv;
+		best_vrs = config->ref_voltages[i].vrs;
```
mathieuchopstm: Missing a `break;`?

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if (best_voltage == INT32_MAX) {
+		return -EINVAL;
+	}
```
mathieuchopstm, line 95: Add a `bool found_setting;` instead, set to `true` in the `for` loop and checked here.
Then remove initial values of `best_voltage`/`best_vrs`.

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+
+	uint8_t current_vrs = FIELD_GET(VREFBUF_CSR_VRS, config->base->CSR);
+
+	if (best_vrs == current_vrs) {
+		return 0;
+	}
+
+	config->base->CSR =
+		(config->base->CSR & ~VREFBUF_CSR_VRS) | FIELD_PREP(VREFBUF_CSR_VRS, best_vrs);
```
mathieuchopstm: 1. Is there a reason for checking that `current == best`? If the hardware doesn't care about the value being written if it's the same, don't bother and just write unconditionally...
2. Use `stm32_reg_modify_bits()` from `#include <stm32_bitops.h>`

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+		return -ENODEV;
+	}
+
+	if (reset_line_deassert_dt(&config->reset) != 0) {
+		LOG_ERR("Could not deassert reset line");
+		return -EIO;
+	}
+
+	if (config->vrefp_output_enable) {
+		config->base->CSR = config->base->CSR & ~VREFBUF_CSR_HIZ;
```
mathieuchopstm:
```suggestion
		LL_VREFBUF_DisableHIZ();
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	} else {
+		config->base->CSR = config->base->CSR | VREFBUF_CSR_HIZ;
```
mathieuchopstm:
```suggestion
		LL_VREFBUF_EnableHIZ();
```

mathieuchopstm (commented):
Pass 1

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+#include <zephyr/drivers/reset.h>
+#include <zephyr/drivers/clock_control/stm32_clock_control.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(regulator_stm32_vrefbuf, CONFIG_REGULATOR_LOG_LEVEL);
+
+struct regulator_stm32_vrefbuf_voltage {
+	uint8_t vrs;
+	uint32_t uv;
+};
```
mathieuchopstm, line 19: Step 1: change structure
```suggestion
struct regulator_stm32_vrefbuf_voltage {
	uint32_t vrs; /* LL_VREFBUF_VOLTAGE_SCALE_<X> */
	int32_t uv;   /* VREFBUF output in µV with this scale */
};
```

drivers/regulator/regulator_stm32_vrefbuf.c
mathieuchopstm: Step 3: use LL functions
```suggestion
	LL_VREFBUF_SetVoltageScaling(vrs);
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+		(config->base->CSR & ~VREFBUF_CSR_VRS) | FIELD_PREP(VREFBUF_CSR_VRS, best_vrs);
+
+	return 0;
+}
+
+static int regulator_stm32_vrefbuf_get_voltage(const struct device *dev, int32_t *volt_uv)
+{
+	const struct regulator_stm32_vrefbuf_config *config = dev->config;
+
+	uint8_t vrs = FIELD_GET(VREFBUF_CSR_VRS, config->base->CSR);
```
mathieuchopstm: Step 3: use LL functions
```suggestion
	uint32_t vrs = LL_VREFBUF_GetVoltageScaling();
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	uint32_t uv;
+};
+
+struct regulator_stm32_vrefbuf_data {
+	struct regulator_common_data common;
+};
+
+struct regulator_stm32_vrefbuf_config {
+	struct regulator_common_config common;
+	VREFBUF_TypeDef *base;
```
mathieuchopstm: Step 4: drop the no-longer-needed base address
```suggestion
```

dts/bindings/regulator/st,stm32-vrefbuf.yaml
```diff
@@ -0,0 +1,50 @@
...
+      First element of each array entry is the Voltage Reference Scale (VRS)
+      value, the second element is the corresponding voltage in microvolts.
+      As example:
+        ref-voltages = <0 2500000>, <1 2048000>, <2 1800000>;
+      Will define 3 reference voltages:
+        - Voltage Reference Scale (VRS) 0: 2.5V
+        - Voltage Reference Scale (VRS) 1: 2.048V
+        - Voltage Reference Scale (VRS) 2: 1.8V
+      Voltage Reference Scales and the corresponding voltages can be found
+      in the datasheet of the specific STM32 device.
```
mathieuchopstm: Step 2: update binding and instantiation macro to consume only voltage
```suggestion
  ref-voltages:
    type: array
    required: true
    description: |
      Reference voltages supported by the VREFBUF regulator.
      
      Each entry contains the VREFBUF output voltage when the corresponding
      voltage scale is selected. For example, the first array entry (index 0)
      corresponds to the output voltage when Scale 0 is selected.
      
      Internally, "Scale <x>" is mapped to LL_VREFBUF_VOLTAGE_SCALE<x> which
      should correspond to the voltage called VREFBUF<x> in Reference Manual.
      Refer to your product's Reference Manual, Datasheet and LL headers for
      more details and to obtain the corresponding voltage value.
      
      For example:
        ref-voltages = <2500000>, <2048000>, <1800000>;
      indicates that 3 reference voltages are supported:
        - SCALE0: 2.5V
        - SCALE1: 2.048V
        - SCALE2: 1.8V
```

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,195 @@
...
+	.set_voltage = regulator_stm32_vrefbuf_set_voltage,
+	.get_voltage = regulator_stm32_vrefbuf_get_voltage,
+};
+
+#define VREFBUF_VOLTAGE_ELEM(node_id, prop, idx)                                                   \
+	COND_CODE_1(DT_PROP_HAS_IDX(node_id, prop, UTIL_X2(idx)),                                  \
+	({		                                                                           \
+		.vrs = DT_PROP_BY_IDX(node_id, prop, UTIL_X2(idx)),		                   \
+		.uv = DT_PROP_BY_IDX(node_id, prop, UTIL_INC(UTIL_X2(idx)))	                   \
+	},), ())
```
mathieuchopstm: Step 2: update binding and instantiation macro to consume only voltage
```suggestion
#define VREFBUF_VOLTAGE_ELEM(node_id, prop, idx)                                                   \
	({		                                                                           \
		.vrs = CONCAT(LL_VREFBUF_VOLTAGE_SCALE_, idx),		                   \
		.uv = DT_PROP_BY_IDX(node_id, prop, idx)	                   \
	},
```

dts/arm/st/h5/stm32h523.dtsi
mathieuchopstm, line 1: Step 5: update DTSI

drivers/regulator/regulator_stm32_vrefbuf.c
tpambor: The minimum voltage within the range specified by the caller is selected here. The voltage values are not consistently ordered across all chip families. For example, the STM32H5 series uses SCALE0 = 2.5 V, SCALE1 = 2.048 V, and SCALE2 = 1.8 V, whereas the STM32U5 series uses SCALE0 = 1.5 V, SCALE1 = 1.8 V, and so on. Therefore, we iterate over all entries to determine the lowest possible voltage within the range, as it may not be the first entry that falls within the range.

mathieuchopstm: Good catch.

drivers/regulator/regulator_stm32_vrefbuf.c
```diff
@@ -0,0 +1,187 @@
...
+}
+
+static int regulator_stm32_vrefbuf_set_voltage(const struct device *dev, int32_t min_uv,
+					       int32_t max_uv)
+{
+	const struct regulator_stm32_vrefbuf_config *config = dev->config;
+	uint32_t best_vrs = 0;
+	int32_t best_voltage = INT32_MAX;
+
+	/* choose minimum possible voltage in range provided by the caller */
```
mathieuchopstm: At least:
```suggestion
	/* search closest voltage above minimum requested by the caller */
```

But I'd suggest the following to remove ambiguity
```suggestion
	/*
	 * Search closest voltage above minimum requested by the caller
	 * among all voltages we support. Note that we must check all
	 * configurations, as the order is not consistent across series
	 * (i.e., whether VREFBUF0 < VREFBUF1 or VREFBUF0 > VREFBUF1).
	 */
```

dts/bindings/regulator/st,stm32-vrefbuf.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - name: base.yaml
+  - name: reset-device.yaml
+  - name: regulator.yaml
+    property-allowlist:
+      - regulator-name
+      - regulator-init-microvolt
+      - regulator-min-microvolt
+      - regulator-max-microvolt
+      - regulator-boot-on
+      - regulator-always-on
```
mathieuchopstm, line 18: Q: are these consumed by the regulator subsystem and invisible to the driver?

mathieuchopstm (commented):
Besides DTSI values (and organization) I haven't checked, LGTM.

dts/bindings/regulator/st,stm32-vrefbuf.yaml
tpambor: Yes, there are consumed here https://github.com/zephyrproject-rtos/zephyr/blob/7987614b5449371bb6a0cd353ecb103c349c9b47/include/zephyr/drivers/regulator.h#L197-L228

and handled by the regulator subsystem, e.g.
https://github.com/zephyrproject-rtos/zephyr/blob/7987614b5449371bb6a0cd353ecb103c349c9b47/drivers/regulator/regulator_common.c#L230-L246

drivers/regulator/regulator_stm32_vrefbuf.c
etienne-lms: I think the previous wording was more accurate. We get the lowest voltarge that match the request level range.
```
	/* Select the minimum possible voltage inside range provided by the caller.
	 * Note that we must check all configurations since the order is not
	 * consistent across series (i.e., whether VREFBUF0 < VREFBUF1 or
	 * VREFBUF0 > VREFBUF1).
	 */

File: drivers/regulator/regulator_stm32_vrefbuf.c [hunk 16]
```diff
@@ -0,0 +1,187 @@
...
+                                                                                                   \
+	static const struct regulator_stm32_vrefbuf_voltage ref_voltages_##inst[] = {              \
+		DT_FOREACH_PROP_ELEM(DT_DRV_INST(inst), ref_voltages, VREFBUF_VOLTAGE_ELEM)        \
+	};                                                                                         \
+                                                                                                   \
+	static const struct regulator_stm32_vrefbuf_config config_##inst = {                       \
+		.common = REGULATOR_DT_INST_COMMON_CONFIG_INIT(inst),                              \
+		.pclken = STM32_DT_INST_CLOCKS(inst),                                              \
+		.reset = RESET_DT_SPEC_GET(DT_DRV_INST(inst)),                                     \
+		.vrefp_output_enable = DT_INST_PROP_OR(inst, vrefp_output_enable, false),          \
```
etienne-lms: `.vrefp_output_enable = DT_INST_PROP(inst, vrefp_output_enable),` is enough.

etienne-lms (commented):
LGTM. Values for H5 and U5 are consistent.
Minor comments.

drivers/regulator/regulator_stm32_vrefbuf.c
etienne-lms: Hmm... reading back your comment @mathieuchopstm, `closest voltage above minimum` sounds quite like `minimum possible voltage`. I've read it too fast.. Your suggestion was fine ::confused::.

mathieuchopstm: By the way, my suggestion comes from checking other drivers which also select the closest-but-above voltage.

(well, strictly speaking, it is `>=` rather than `>`, so I guess this would be more correct:)
```suggestion
	/*
	 * Search among supported voltages for the closest one at least equal
	 * to minimum requested by caller. Note that all configurations must
	 * be checked as the range order is not consistent across series
	 * (i.e., whether VREFBUF0 < VREFBUF1 or VREFBUF0 > VREFBUF1).
	 */
```

dts/bindings/regulator/st,stm32-vrefbuf.yaml
```diff
@@ -0,0 +1,56 @@
...
+      indicates that 3 reference voltages are supported:
+        - SCALE0: 2.5V
+        - SCALE1: 2.048V
+        - SCALE2: 1.8V
+
+  vrefp-output-enable:
+    type: boolean
+    description: |
+      If true, enables the VREF+ output pin. Otherwise, the VREF+ pin is
+      in high-impedance state. Default is false.
```
erwango: Review the if true / Default if false that are inherant to boolean.

Prefer: `Enables the VREF+ output pin. VREF+ pin is in high-impedance state otherwise`

erwango (commented):
Nitpick

drivers/regulator/regulator_stm32_vrefbuf.c
etienne-lms: My 2 cents: a bit simpler to read as
```c
	for (uint8_t i = 0; i < config->ref_voltage_count; i++) {
		if (IN_RANGE(config->ref_voltages[i].uv, min_uv, max_uv) &&
		    config->ref_voltages[i].uv < best_voltage) {
			best_voltage = config->ref_voltages[i].uv;
			best_vrs = config->ref_voltages[i].vrs;
		}
	}
```

etienne-lms (dismissed):

mathieuchopstm (dismissed):

drivers/regulator/regulator_stm32_vrefbuf.c
mathieuchopstm: I agree your proposal is more readable, but non-blocking.

mathieuchopstm (dismissed):

tpambor:
@nordic-auko @seov-nordic Could you have a look?

etienne-lms (dismissed):

mathieuchopstm (approved):

kartben:
@nordic-auko @seov-nordic please review or re-assign to @erwango as stm32 platform maintainer.

seov-nordic (approved):

PR #99305, opened 2025-11-13
drivers: timer: nrf_grtc_timer: Add system_clock_disable implementation

adamkondraciuk:
Add implementation of `sys_clock_disable` function for GRTC timer.
Also extended the system off samples by adding an option to disable
the system clock


nordic-piks (approved):

nordic-segl (approved):

bjarki-andreasen (changes requested):
Driver should select SYSTEM_TIMER_HAS_DISABLE_SUPPORT, see https://github.com/zephyrproject-rtos/zephyr/blob/5805685c326b8d7b6385c4ad910b22c9c762adcb/drivers/timer/Kconfig.nrf_rtc#L12
for example.

bjarki-andreasen (approved):

nordic-krch (approved):

PR #99306, opened 2025-11-13
drivers: can: mcux: flexcan: add dynamic memory allocation for MB

### LINKED ISSUE #92798: can: can_mcux_flexcan.c allows a mailbox overrun resulting in an imprecise data bus fault
### Describe the bug

I'm on commit hash: 

can_mcux_flexcan.c lacks proper bounds checking to ensure that send calls don't overrun outside of the area allocated by the flexcan peripheral for message buffers.

On line 737 of can_mcux_flexcan.c inside of mcux_flexcan_send, there is this section of code:
```
for (alloc = 0; alloc < MCUX_FLEXCAN_MAX_TX; alloc++) {
	if (!atomic_test_and_set_bit(data->tx_allocs, alloc)) {
		break;
	}
}

data->tx_cbs[alloc].function = callback;
data->tx_cbs[alloc].arg = user_data;
xfer.mbIdx = ALLOC_IDX_TO_TXMB_IDX(alloc);
```

As you can see, the driver simply iterates from 0 to `MCUX_FLEXCAN_MAX_TX`. `MCUX_FLEXCAN_MAX_TX` is set by the following:
```
#define MCUX_FLEXCAN_MAX_TX (MCUX_FLEXCAN_MAX_MB - MCUX_FLEXCAN_MAX_RX)
```
```
#define MCUX_FLEXCAN_MAX_MB FSL_FEATURE_FLEXCAN_HAS_MESSAGE_BUFFER_MAX_NUMBERn(0)
```
```
#define FSL_FEATURE_FLEXCAN_HAS_MESSAGE_BUFFER_MAX_NUMBERn(x) (64)
```
```
#define MCUX_FLEXCAN_MAX_RX (CONFIG_CAN_MAX_FILTER + RX_START_IDX)
```

However, if the peripheral is configured to use 64 byte message buffers, a maximum of 64 buffers is wrong.

From the datasheet, the max message buffer would be 13:

Image

Additionally, even if this max were set correctly, it doesn't seem great to just assume we will have an open buffer. In the case of no open buffer, this code just assumes the last one as the buffer to use. Really, this code should error out if it can't find an open buffer.

Regardless, `alloc` ends up being set incorrectly and eventually triggers a data bus error within the `FLEXCAN_SetFDTxMbConfig` call when it tries to access memory outside of the message buffer space.

In my case, this is always when it hits mailbox 14. It's probably unlikely to hit this if your CAN bus is operating normally as you won't reach the higher mailbox numbers. However, when the CAN peripheral is in an error state, the mailboxes just stack up until you overflow.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Enable CAN FD on the can_mcux_flexcan.c driver, put the CAN bus into an error state so that mailboxes are not being cleared (not strictly necessary, but makes it easier to hit this case), send messages until mailboxes overflow.

Edit: I'm not sure the error state is actually a part of this or would actually work. I think I'm overflowing just through normal operation.

### Relevant log output

```shell

```

### Impact

### Environment

Commit: 0285cf4235cec506e0c7d4e1f9e63d2de5fca783
Device: Custom IMXRT1064 board

### Additional Context

_No response_

william-tang914:
This change replaces static memory allocation for FlexCAN message buffers with dynamic allocation using a per-instance heap. The motivation is to support FlexCAN instances with varying message buffer counts and CAN FD configurations.
For CAN FD instances, message buffer count is adjusted based on 64-byte payload size (7 buffers per RAM block vs 32 for 8-byte payload).

Key changes:
  - Add CAN_MCUX_FLEXCAN_HEAP_SIZE Kconfig option (default 2048 bytes) to configure heap size for message buffer allocation
  - Replace compile-time MCUX_FLEXCAN_MAX_MB/RX/TX macros with runtime calculation in mcux_flexcan_get_mb_config()
  - Add dynamic allocation of rx_allocs, tx_allocs, rx_cbs, and tx_cbs arrays using k_heap during driver initialization
  - Update all array bounds checks and loop limits to use runtime values stored in driver data structure
  - Remove Kconfig symbol `CAN_MAX_MB` beacuse such information is added to device tree.
  - Remove range limitation in Kconfig symbol `CAN_MAX_FILTER` because such validation is added in mcux_flexcan_get_mb_config()

 Fixes #92798

### DIFF
--- a/boards/nxp/mr_canhubk3/doc/index.rst
+++ b/boards/nxp/mr_canhubk3/doc/index.rst
@@ -143,11 +143,8 @@ flexcan5         | PTC11  | PTC11_CAN0_RX  P22/P23
    and support maximum 32 message buffers for concurrent active instances with 8 bytes
    payload. We need to pay attention to configuration options:
 
-   1. :kconfig:option:`CONFIG_CAN_MAX_MB` must be less or equal than the
+   1. :kconfig:option:`CONFIG_CAN_MAX_FILTER` must be less or equal than the
       maximum number of message buffers that is according to the table below.
-
-   2. :kconfig:option:`CONFIG_CAN_MAX_FILTER` must be less or equal than
-      :kconfig:option:`CONFIG_CAN_MAX_MB`.
 
 ===============  ==========  ================  ================
 Devicetree node  Payload     Hardware support  Software support
--- a/drivers/can/Kconfig.mcux
+++ b/drivers/can/Kconfig.mcux
@@ -28,29 +28,18 @@ config CAN_MCUX_FLEXCAN_WAIT_TIMEOUT
 	  Maximum number of wait loop iterations for the MCUX FlexCAN HAL when entering/leaving
 	  freeze mode.
 
-config CAN_MAX_MB
-	int "Maximum number of message buffers for concurrent active instances"
-	default 16
-	depends on SOC_SERIES_S32K3 || SOC_SERIES_S32K1 || SOC_SERIES_S32ZE
-	range 1 96 if SOC_SERIES_S32K3
-	range 1 32 if SOC_SERIES_S32K1 && !SOC_S32K142W && !SOC_S32K144W
-	range 1 64 if SOC_S32K142W || SOC_S32K144W
-	range 1 128 if SOC_SERIES_S32ZE
-	help
-	  Defines maximum number of message buffers for concurrent active instances.
-
 config CAN_MAX_FILTER
 	int "Maximum number of concurrent active RX filters"
 	default 5
-	range 1 15 if SOC_SERIES_KINETIS_KE1XF || SOC_SERIES_KINETIS_K6X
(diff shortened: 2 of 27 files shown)


henrikbrixandersen (changes requested):
Dynamic allocation is not the solution here. This can be derived at build timer per instance (as is done in other drivers as well, see e.g. https://github.com/zephyrproject-rtos/zephyr/blob/3cf7b202780d788976e4bf3e46724e63d23f6bcc/include/zephyr/drivers/can/can_mcan.h#L1216).

On a side note, I do have a plan for a Zephyr-native FlexCAN driver, but the more workarounds and `#ifdefs` that are added, the more I realise that the NXP HAL hides way too many differences in the IP implementations covered under the market name "FlexCAN", which are very, very difficult to capture in a single driver.

william-tang914:
Hi @henrikbrixandersen , following your advice, I create another draft PR.
https://github.com/zephyrproject-rtos/zephyr/pull/99483

PR #99313, opened 2025-11-13
doc: kconfig: document "named choices" as a language extension

mathieuchopstm:
When creating a `choice` block, it is possible to associate a name to the choice, resulting in a so-called *named choice* being created. This allows modification of the choice after its initial declaration - for example, its default value can be overridden from another Kconfig file.

This feature originates from Linux and is supported by Kconfiglib; however, since commit [c83f020973bc72d9eec65474d8c47495191aef20](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/Documentation/kbuild/kconfig-language.rst?h=v6.18-rc5&id=c83f020973bc72d9eec65474d8c47495191aef20) (first found in release v6.9), Linux no longer supports this feature. Critically, this means that the feature is no longer covered by the [Linux Kconfig Language documentation](https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html), even though it is supported and used in Zephyr.

Add "named choices" to the Kconfig language extensions documentation page since it has de facto become an extension after its support was dropped from Linux.

### DIFF
--- a/doc/build/kconfig/extensions.rst
+++ b/doc/build/kconfig/extensions.rst
@@ -126,3 +126,35 @@ which includes some Kconfig extensions:
 - ``def_int``, ``def_hex``, and ``def_string`` keywords are available,
   analogous to ``def_bool``. These set the type and add a ``default`` at the
   same time.
+
+- A symbol name can be associated to ``choice`` groups using the ``choice <symbol>`` syntax.
+  Such choices, called *named choices*, can be modified from places other than their initial
+  definition. For example, the following statements define the named choice ``FOOBAR``:
+
+  .. code-block:: kconfig
+
+    choice FOOBAR
+	    prompt "Example choice"
+	    default BAR
+
+    config FOO
+	    bool "Foo"
+
+    config BAR
+	    bool "Bar"
+
+    endchoice
+
+  The following statements could then be used, in a ``Kconfig.defconfig`` file for example,
+  to override the default option of choice ``FOOBAR``:
+
+  .. code-block:: kconfig
+
+    # Note how "prompt" is not present here
+    choice FOOBAR
+        default FOO
+    endchoice
+
+  .. note::
+    The *named choices* feature originates from Linux, but it is no longer supported in Linux
+    since kernel release 6.9, and has thus become a Kconfiglib language extension.


doc/build/kconfig/extensions.rst
```diff
@@ -126,3 +126,35 @@ which includes some Kconfig extensions:
...
   analogous to ``def_bool``. These set the type and add a ``default`` at the
   same time.
+
+- A symbol name can be associated to ``choice`` groups using the ``choice <symbol>`` syntax.
+  Such choices, called *named choices*, can be modified by other places after their initial
+  definition. For example, if ``choice BOOT_SIGNATURE_TYPE`` was initially defined as:
+
+  .. code-block:: kconfig
+
+    choice BOOT_SIGNATURE_TYPE
```
tejlmand: I think we can present this using `FOO` and `BAR` as we do with other examples.

Like:
```
choice FOO_CHOICE
    default FOO

    config FOO
	    bool "Foo"

    config BAR
	    bool "Bar"

endchoice
```

and then describe that the default can be changed.

doc/build/kconfig/extensions.rst
```diff
@@ -126,3 +126,35 @@ which includes some Kconfig extensions:
...
+
+    config BOOT_SIGNATURE_TYPE_NONE
+	    bool "No signature; use only hash check"
+
+    config BOOT_SIGNATURE_TYPE_RSA
+	    bool "RSA signatures"
+
+    endchoice
+
+  Since this ``choice`` is named, it is possible to modify it after its initial definition.
```
tejlmand: > after its initial definition.

Not correct. A `defconfig` file is sourced BEFORE the definition of the choice entries.
This also follows from
> Kconfig uses the first default with a satisfied condition, where an empty condition corresponds to if y (is always satisfied).

https://docs.zephyrproject.org/latest/build/kconfig/setting.html#configuring-invisible-kconfig-symbols

doc/build/kconfig/extensions.rst
```diff
@@ -126,3 +126,35 @@ which includes some Kconfig extensions:
...
+  For example, the following statements can be used to override the default option:
```
tejlmand:
```suggestion
  For example, the following statements can be used to override the default option, such as in a ``Kconfig.defconfig`` file:
```

tejlmand (changes requested):
Thanks, some minor proposals for improving further.

doc/build/kconfig/extensions.rst
mathieuchopstm: Right, originally I wrote `modify it from a place other than its initial definition`, but does *initial definition* even make sense? I'm not sure about what wording to use, exactly.

mathieuchopstm: This sentence has been reworded but there's still:
```
Such choices, called *named choices*, can be modified by other places after their initial definition.
```

What do you think of
```
Such choices, called *named choices*, can be modified from places other than their initial definition.
```

mbolivar (dismissed):

doc/build/kconfig/extensions.rst
mathieuchopstm: Done.

doc/build/kconfig/extensions.rst
mathieuchopstm: Done.

tejlmand (dismissed):

doc/build/kconfig/extensions.rst
tejlmand: yes, that would be a better as it includes both the option of changing the default (done before, such as in defconfig), and after, like extending a choice with extra options in a Zephyr module.

mathieuchopstm: Change done.

tejlmand (approved):

mbolivar (approved):

nashif (approved):

PR #99314, opened 2025-11-13
samples: openthread: border_router: Add NAT64 functionality

Cristib05:
This PR aims to add NAT64 functionality in OpenThread Border Router sample application.


modules/openthread/platform/infra_if.c
```diff
@@ -257,3 +277,113 @@ void infra_if_stop_icmp6_listener(void)
...
+
+exit:
+	return error;
+}
+
+static int icmpv4_handle_echo_reply(struct net_icmp_ctx *ctx,
+				    struct net_pkt *pkt,
+				    struct net_icmp_ip_hdr *hdr,
+				    struct net_icmp_hdr *icmp_hdr,
+				    void *user_data)
```
rlubos: Please use `ARG_UNUSED()` macro on unused arguments.

modules/openthread/platform/infra_if.c
```diff
@@ -257,3 +277,113 @@ void infra_if_stop_icmp6_listener(void)
 	(void)net_icmp_cleanup_ctx(&rs_ctx);
 	(void)net_icmp_cleanup_ctx(&na_ctx);
 }
+
+#if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR)
+otError infra_if_nat64_init(struct in_addr addr)
```
rlubos: `addr` argument doesn't seem to be used, what's the purpose here?

subsys/net/l2/openthread/openthread_border_router.c
```diff
@@ -607,3 +621,59 @@ static void openthread_border_router_add_route_to_multicast_groups(void)
...
+{
+	otInstance *instance = openthread_get_default_instance();
+	struct net_if_ipv4 *ipv4 = ail_iface_ptr->config.ip.ipv4;
+	struct in_addr ipv4_addr = {0};
+	struct in_addr ipv4_def_route = {0};
+	bool translator_state = false;
+	struct net_if_addr *unicast;
+	otIp4Cidr cidr;
+
+	ARRAY_FOR_EACH(ipv4->unicast, i) {
```
rlubos: We have `net_if_ipv4_get_global_addr()` helper that would return first encountered IPv4 global address found on interface, perhaps it could be used instead?

modules/openthread/platform/infra_if.c
```diff
@@ -257,3 +277,113 @@ void infra_if_stop_icmp6_listener(void)
...
+			openthread_border_router_deallocate_message((void *)req);
+		}
+	}
+}
+
+otError infra_if_create_raw_socket(void)
+{
+	otError error = OT_ERROR_NONE;
+
+	raw_send_sock = zsock_socket(AF_INET, SOCK_RAW, IPPROTO_RAW);
```
rlubos: I'm a bit puzzled here, `raw_send_sock` doesn't seem to be used anywhere? And the error check below verifies `raw_udp_sock` instead (which is created elsewhere)?

rlubos (commented):
Nice, thanks for the effort

modules/openthread/platform/infra_if.c
```diff
@@ -125,7 +145,7 @@ otError infra_if_init(otInstance *instance, struct net_if *ail_iface)
 	ail_iface_index = (uint32_t)net_if_get_by_iface(ail_iface_ptr);
 	net_ipv6_addr_create_ll_allrouters_mcast(&addr);
 	VerifyOrExit(net_ipv6_mld_join(ail_iface, &addr) == 0, error = OT_ERROR_FAILED);
-
+	sockfd_udp[0].fd = -1;
```
jukkar: Shouldn't we loop through `sockfd_udp` list here?

modules/openthread/platform/infra_if.c
```diff
@@ -32,6 +38,20 @@ static struct net_icmp_ctx na_ctx;
 
 static void infra_if_handle_backbone_icmp6(struct otbr_msg_ctx *msg_ctx_ptr);
 static void handle_ra_from_ot(const uint8_t *buffer, uint16_t buffer_length);
+#if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR)
+#define MAX_SERVICES 1
```
jukkar, line 42: Perhaps Kconfig configurable?

modules/openthread/platform/infra_if.c
```diff
@@ -257,3 +277,113 @@ void infra_if_stop_icmp6_listener(void)
 	(void)net_icmp_cleanup_ctx(&rs_ctx);
 	(void)net_icmp_cleanup_ctx(&na_ctx);
 }
+
+#if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR)
+otError infra_if_nat64_init(struct in_addr addr)
+{
+	otError error = OT_ERROR_NONE;
+	static struct net_icmp_ctx ctx;
+	struct sockaddr anyaddr = { .sa_family = AF_INET};
```
jukkar: The type should be `struct sockaddr_in`. Note that you should never instantiate variable with type `struct sockaddr` as it is only meant to do casts. If you need anonymous sockaddr, then you can create a variable of `struct sockaddr_storage` type.

subsys/net/l2/openthread/openthread_border_router.c
```diff
@@ -175,6 +187,7 @@ static int openthread_stop_border_router_services(struct net_if *ot_iface,
 		border_agent_deinit();
 		infra_if_stop_icmp6_listener();
 		udp_plat_deinit();
+		openthread_border_router_stop_nat64_service();
```
jukkar, line 193: Did not fully follow the defines what are available, but does this require ifdef guards here?

subsys/net/l2/openthread/openthread_border_router.c
```diff
@@ -267,6 +280,7 @@ static void ail_ipv4_address_event_handler(struct net_mgmt_event_callback *cb, u
 	}
 
 	mdns_plat_monitor_interface(iface);
+	openthread_border_router_start_nat64_service();
```
jukkar, line 288: ditto

modules/openthread/platform/infra_if.c
Cristib05: I tried to see if I could use only one socket for both sending and receiving, it didn't work at first, then I had to do a context switch from this and I have completely forgot...

modules/openthread/platform/infra_if.c
Cristib05: I am planning to do this in another PR, where I would do Kconfigs for all BR platform modules.

modules/openthread/platform/infra_if.c
Cristib05: Thanks!

modules/openthread/platform/infra_if.c
rlubos: What didn't work exactly, I think it should be perfectly fine to send/receive over a single IP socket. I mean two sockets are fine too, but it'd be good to get to the bottom of this to save some resources.

modules/openthread/platform/infra_if.c
```diff
@@ -257,3 +280,122 @@ void infra_if_stop_icmp6_listener(void)
...
+
+exit:
+	return error;
+}
+
+otError infra_if_send_raw_message(uint8_t *buf, uint16_t len)
+{
+	otError error = OT_ERROR_NONE;
+
+	VerifyOrExit(zsock_send(raw_send_sock, buf, len, 0) == 0,
```
rlubos: I don't think this check is correct? `zsock_send()` should return the number of bytes set, not 0.

subsys/net/l2/openthread/openthread_border_router.c
```diff
@@ -607,3 +627,52 @@ static void openthread_border_router_add_route_to_multicast_groups(void)
...
+static void openthread_border_router_start_nat64_service(void)
+{
+	otInstance *instance = openthread_get_default_instance();
+	struct in_addr *ipv4_addr = NULL;
+	struct in_addr ipv4_def_route = {0};
+	bool translator_state = false;
+	otIp4Cidr cidr;
+
+	ipv4_addr = net_if_ipv4_get_global_addr(ail_iface_ptr, NET_ADDR_PREFERRED);
+
```
rlubos: nit: Could remove the empty line here.

rlubos (approved):

jukkar (approved):

PR #94603, opened 2025-08-18
scripts: zephyr_module: accept empty env variables `ZEPHYR_EXTRA_MODULES` and `EXTRA_ZEPHYR_MODULES`

thorsten-klein:
Proceed if the environment variables `ZEPHYR_EXTRA_MODULES` or `EXTRA_ZEPHYR_MODULES` are set to an empty string. In this case. handle same way as if the variable was unset.

This is helpful, when temporarily omitting the values set to environment variable, e.g.

``` bash
export ZEPHYR_EXTRA_MODULES=$(realpath my-modules)
west build app # build with extra modules
ZEPHYR_EXTRA_MODULES= west build app # temporarily omit extra modules
```

PR #99317, opened 2025-11-13
scripts: zephyr_module: ignore empty paths in extra modules env variables

thorsten-klein:
Successor of #94603.

Ignore empty paths in environment variable `ZEPHYR_EXTRA_MODULES` or `EXTRA_ZEPHYR_MODULES`.
This can happen easily when you append paths to env variables, e.g. in bash
```
ZEPHYR_EXTRA_MODULES+=";/my/module"
```
#### Actual:
If the variable was empty, it results in `;/my/module`.
Currently, this string resolves to `[Path(), Path('/my/module')]`, which is same as `[Path('.'), Path('/my/module')]`.
As you can see, `Path('.')` is wrong.

#### Expected:
`zephyr_module.py` should ignore empty paths.

### DIFF
--- a/scripts/zephyr_module.py
+++ b/scripts/zephyr_module.py
@@ -772,7 +772,7 @@ def parse_modules(zephyr_base, manifest=None, west_projs=None, modules=None,
             extra_module = os.environ.get(var, None)
             if not extra_module:
                 continue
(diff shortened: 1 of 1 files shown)


pdgendt (approved):
Take a look at the `signed-off-by` compliance issue.

tejlmand (approved):

PR #99318, opened 2025-11-13
tests: bluetooth: tester: Fix CSIS/SR/SP/BV-07-C

Frodevan:
This particular test-case requires bttester to respond with an OOB procedure SIRK only error. To enable this response, the BTP command to set SIRK mode has been extended with a new selection.

Associated AutoPTS PR: https://github.com/auto-pts/auto-pts/pull/1613

### DIFF
--- a/tests/bluetooth/tester/src/audio/btp/btp_csis.h
+++ b/tests/bluetooth/tester/src/audio/btp/btp_csis.h
@@ -12,6 +12,10 @@
 #include <zephyr/bluetooth/audio/csip.h>
 
 /* CSIS commands */
+#define BTP_CSIS_SIRK_TYPE_PLAINTEXT		0x00
+#define BTP_CSIS_SIRK_TYPE_ENCRYPTED		0x01
+#define BTP_CSIS_SIRK_TYPE_OOB_ONLY		0x02
+
 #define BTP_CSIS_READ_SUPPORTED_COMMANDS	0x01
 struct btp_csis_read_supported_commands_rp {
 	uint8_t data[0];
@@ -33,7 +37,7 @@ struct btp_csis_get_member_rsi_rp {
 	uint8_t rsi[BT_CSIP_RSI_SIZE];
 } __packed;
 
-#define BTP_CSIS_ENC_SIRK_TYPE			0x04
-struct btp_csis_sirk_type_cmd {
-	uint8_t encrypted;
+#define BTP_CSIS_SET_SIRK_TYPE			0x04
+struct btp_csis_sirk_set_type_cmd {
+	uint8_t type;
 } __packed;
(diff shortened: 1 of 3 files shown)


tests/bluetooth/tester/src/audio/btp/btp_csis.h
```diff
@@ -33,7 +33,7 @@ struct btp_csis_get_member_rsi_rp {
 	uint8_t rsi[BT_CSIP_RSI_SIZE];
 } __packed;
 
-#define BTP_CSIS_ENC_SIRK_TYPE			0x04
+#define BTP_CSIS_SIRK_TYPE			0x04
 struct btp_csis_sirk_type_cmd {
```
Thalley: Consider naming both to use the name used by BTP which is "Set SIRK type"


```suggestion
#define BTP_CSIS_SET_SIRK_TYPE			0x04
struct btp_csis_set_sirk_type_cmd {
```

tests/bluetooth/tester/src/audio/btp_csis.c
```diff
@@ -69,9 +69,23 @@ static uint8_t csis_sirk_type(const void *cmd, uint16_t cmd_len, void *rsp, uint
...
+	switch (cp->type) {
+		case 0:
+			LOG_DBG("SIRK type: plain text");
+			sirk_read_response = BT_CSIP_READ_SIRK_REQ_RSP_ACCEPT;
+			break;
+		case 1:
+			LOG_DBG("SIRK type: encrypted");
+			sirk_read_response = BT_CSIP_READ_SIRK_REQ_RSP_ACCEPT_ENC;
+			break;
+		case 2:
```
Thalley: Instead of magic numbers, we should define the possible values (both here and in AutoPTS)

tests/bluetooth/tester/src/audio/btp_csis.c
```diff
@@ -132,6 +142,8 @@ uint8_t tester_init_csis(void)
 	};
 	int err = bt_csip_set_member_register(&register_params, &csis_svc_inst);
 
+	sirk_read_response = BT_CSIP_READ_SIRK_REQ_RSP_REJECT;
```
Thalley: Why are we defaulting to rejecting? 

IMO it probably makes more sense to default to `BT_CSIP_READ_SIRK_REQ_RSP_ACCEPT` as that's the most commonly used response type

Thalley (changes requested):

tests/bluetooth/tester/src/audio/btp/btp_csis.h
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_csis.c
Frodevan: Done

tests/bluetooth/tester/src/audio/btp_csis.c
Frodevan: It was just a thought that the test-case definition might want to actively make a choice of the type of sirk to use, but I also see now that this may cause confusion for anyone who are to implement new test-cases so it's now changed back to default to plaintext sirk.

sjanc (approved):

Thalley (approved):

PR #99321, opened 2025-11-13
STM32 SPI: Extend data size support

gautierg-st:
Until now the STM32 SPI driver only supported 8 and 16-bit data sizes, even though some SoCs supports more sizes than that.

This PR extends the possible data sizes that can be used, ranging from 4 to 32-bit depending on SoC and SPI instances.


dts/bindings/spi/st,stm32h7-spi.yaml
```diff
@@ -32,3 +32,9 @@ properties:
   fifo-enable:
     type: boolean
     description: Enable the SPI FIFO usage for performance improvement.
+
+  limited-instance:
+    type: boolean
+    description: |
+      Indicates if the SPI instance is a limited-features instance.
```
erwango: Does it comes with other limitations ?
Otherwise I think we should somehow explicit that this is a data width limitation.

drivers/spi/spi_ll_stm32.c
```diff
@@ -67,6 +67,55 @@ LOG_MODULE_REGISTER(spi_ll_stm32);
...
 #endif /* CONFIG_SOC_SERIES_STM32MP1X */
 
+/* Several cases on the supported data width
+ * - For non-H7 compatible:
+ *   - Some series only support 8 and 16-bit
+ *   - Some series support 4 to 16-bit for all instances
+ * - For H7 compatible:
+ *   - All series have at least one instance supporting 4 to 32-bit
+ *   - Some series have instances supporting only 4 to 16-bit
+ *   - Some series have instances supporting only 8 and 16-bit
```
erwango: it would be nice that these cases appear more clearly from device tree.

drivers/spi/spi_stm32.c
```diff
@@ -67,6 +67,55 @@ LOG_MODULE_REGISTER(spi_ll_stm32);
...
+#define STM32_SPI_LIMITED_DATAWIDTH_MAX	16
+#else
+#define STM32_SPI_DATAWIDTH_MAX	16
+#endif
+
+#define STM32_SPI_DATAWIDTH(i, _)								\
+	_CONCAT_2(LL_SPI_DATAWIDTH_, UTIL_INC(UTIL_INC(UTIL_INC(UTIL_INC(i)))), BIT)
+static const uint32_t table_datawidth[] = {
+	LISTIFY(UTIL_DEC(UTIL_DEC(UTIL_DEC(STM32_SPI_DATAWIDTH_MAX))), STM32_SPI_DATAWIDTH, (,))
+};
```
erwango, line 102: A comment on what this maros is trying to achieve would be welcome

drivers/spi/spi_stm32.c
```diff
@@ -410,35 +484,43 @@ static void spi_stm32_shift_m(const struct spi_stm32_config *cfg,
...
 			UNALIGNED_PUT(rx_frame, (uint8_t *)data->ctx.rx_buf);
-			spi_context_update_rx(&data->ctx, 1, 1);
-		} else {
+		} else if (frame_size == 2U) {
 			rx_frame = LL_SPI_ReceiveData16(spi);
 			UNALIGNED_PUT(rx_frame, (uint16_t *)data->ctx.rx_buf);
-			spi_context_update_rx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		} else {
+			rx_frame = LL_SPI_ReceiveData32(spi);
```
erwango, line 534: rx_frame is `uint16_t`

drivers/spi/spi_stm32.c
```diff
@@ -410,35 +484,43 @@ static void spi_stm32_shift_m(const struct spi_stm32_config *cfg,
...
 			tx_frame = UNALIGNED_GET((uint16_t *)(data->ctx.tx_buf));
 			LL_SPI_TransmitData16(spi, tx_frame);
-			spi_context_update_tx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		} else {
+			tx_frame = UNALIGNED_GET((uint32_t *)(data->ctx.tx_buf));
+			LL_SPI_TransmitData32(spi, tx_frame);
+#endif
 		}
+		spi_context_update_tx(&data->ctx, frame_size, 1);
```
erwango, line 518: `tx_frame` is `uint16_t`

drivers/spi/spi_ll_stm32.c
```diff
@@ -99,6 +148,19 @@ static void spi_stm32_pm_policy_state_lock_put(const struct device *dev)
 	}
 }
 
+static uint32_t bits2bytes(spi_operation_t operation)
```
erwango: uint8_t ?

See comment son usage below

drivers/spi/spi_stm32.c
```diff
@@ -1151,11 +1255,7 @@ static int transceive(const struct device *dev,
 	}
 
 	/* Set buffers info */
-	if (SPI_WORD_SIZE_GET(config->operation) == 8) {
-		spi_context_buffers_setup(&data->ctx, tx_bufs, rx_bufs, 1);
-	} else {
-		spi_context_buffers_setup(&data->ctx, tx_bufs, rx_bufs, 2);
-	}
+	spi_context_buffers_setup(&data->ctx, tx_bufs, rx_bufs, bits2bytes(config->operation));
```
erwango, line 1270: `spi_context_buffers_setup` takes `uint8_t` as last arg.

drivers/spi/spi_stm32.h
```diff
@@ -97,13 +98,17 @@ static inline uint32_t ll_func_dma_get_reg_addr(SPI_TypeDef *spi, uint32_t locat
...
+	} else {
+		return LL_SPI_IsActiveFlag_EOT(spi) == 0;
+	}
 #else
 	/* the SPI Tx empty and busy flags are needed */
-	return (LL_SPI_IsActiveFlag_TXE(spi) &&
-		!LL_SPI_IsActiveFlag_BSY(spi));
-#endif /* LL_SPI_SR_TXC */
+	return (!LL_SPI_IsActiveFlag_TXE(spi) ||
+		LL_SPI_IsActiveFlag_BSY(spi));
```
erwango, line 110: Dedicated commit required ?

erwango (changes requested):

drivers/spi/spi_stm32.c
mathieuchopstm: Possibly more readable implementation:
```c
#define N_PLUS_4(n, _) UTIL_INC(UTIL_INC(UTIL_INC(UTIL_INC(n))))

/* Generates a list from 4 to max_width: "4, 5, ..., max_width" */
#define DATA_WIDTHS(max_width) \
    LISTIFY(UTIL_DEC(UTIL_DEC(UTIL_DEC(max_width))), N_PLUS_4, (,))

#define STM32_SPI_DATAWIDTH(n) CONCAT(LL_SPI_DATAWIDTH_, n, BIT)
static const uint32_t table_datawidth[] = {
    FOR_EACH(STM32_SPI_DATAWIDTH, (,), DATA_WIDTHS(STM32_SPI_DATAWIDTH_MAX))
};
```

dts/bindings/spi/st,stm32h7-spi.yaml
gautierg-st: It's also smaller FIFO capacity and reduced transfer size. From U5 RM:
image

The "limited-but-not-so-much" instances found on H7 for example don't have the size limitations:
image

The limited size should probably be checked depending on the instances. But this can be done in another PR.

drivers/spi/spi_ll_stm32.c
gautierg-st: So instead of having a boolean property, make it a string enum:
- "full-32-bit"
- "full-16-bit"
- "limited"

And set it for all series instead of just "h7-compatible" ?

dts/bindings/spi/st,stm32-spi-common.yaml
```diff
@@ -21,3 +21,16 @@ properties:
...
+    required: true
+    enum:
+      - "full-32-bit"
+      - "full-16-bit"
+      - "limited-16-bit"
+    description: |
+      Indicates the data width supported by the SPI instance.
+      - full-32-bit: supports data width from 4 to 32 bits
+      - full-16-bit: supports data width from 4 to 16 bits
+      - limited-16-bit: supports only 8 and 16 bits data width
```
erwango: I'd like to bikeshed on bit on this. What about more explicits:
```
      - full-4-to-32-bit: supports data widths from 4 to 32 bits
      - full-4-to-16-bit: supports data widths from 4 to 16 bits
      - limited-8-16-bit: supports only 8 and 16 bits data widths
```
or more simple:
```
      - 4-to-32-bit: supports data width from 4 to 32 bits
      - 4-to-16-bit: supports data width from 4 to 16 bits
      - 8-16-bit: supports only 8 and 16 bits data width
```
(also not sure if we don't need `s` to `bit` )

erwango (changes requested):
Best part of the review remains :-D
Otherwise LGTM

dts/bindings/spi/st,stm32-spi-common.yaml
gautierg-st: The full/limited terms are explicitly used in the RefMan of H7-compatible SPI instances. So that way, it is immediately identifiable, and they imply some limitations for some instances.
On the other hand, these terms are not used for other series. For each given series, all instances are identical, so there is no need in the RefMan to include those terms. But I think they can be extrapolated for these series (though it doesn't come with all the features of the H7-compatible).
All in all, I'm rather in favor of your first proposition (bonus point: all `:` are nicely aligned :D )

> (also not sure if we don't need `s` to `bit` )

AFAIK, we talk about 32-bit words (32-bit is an adjective, so no `s`), or words of 32 bits (here bits is a noun so it takes an `s`). So I'd say both are fine, it depends on your perspective.

erwango: ok, go for first proposal then !

gautierg-st: Done

erwango (dismissed):

erwango (approved):

drivers/spi/spi_stm32.c
```diff
@@ -302,43 +378,55 @@ static int spi_dma_move_buffers(const struct device *dev, size_t len)
...
-		spi_context_update_tx(&data->ctx, 1, 1);
-	} else {
+	} else if (frame_size == 2U) {
 		if (spi_context_tx_buf_on(&data->ctx)) {
 			tx_frame = UNALIGNED_GET((uint16_t *)(data->ctx.tx_buf));
 		}
 		LL_SPI_TransmitData16(spi, tx_frame);
-		spi_context_update_tx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	} else {
```
mathieuchopstm, line 395: Nit: SonarQube would prefer (and I think this looks better too)
```suggestion
	} else {
#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
```

drivers/spi/spi_stm32.c
```diff
@@ -302,43 +378,55 @@ static int spi_dma_move_buffers(const struct device *dev, size_t len)
...
-		spi_context_update_rx(&data->ctx, 1, 1);
-	} else {
+	} else if (frame_size == 2U) {
 		rx_frame = LL_SPI_ReceiveData16(spi);
 		if (spi_context_rx_buf_on(&data->ctx)) {
 			UNALIGNED_PUT(rx_frame, (uint16_t *)data->ctx.rx_buf);
 		}
-		spi_context_update_rx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	} else {
```
mathieuchopstm, line 422: Ditto:
```suggestion
	} else {
#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
```

drivers/spi/spi_stm32.c
```diff
@@ -410,35 +498,44 @@ static void spi_stm32_shift_m(const struct spi_stm32_config *cfg,
...
 			tx_frame = UNALIGNED_GET((uint8_t *)(data->ctx.tx_buf));
 			LL_SPI_TransmitData8(spi, tx_frame);
-			spi_context_update_tx(&data->ctx, 1, 1);
-		} else {
+		} else if (frame_size == 2U) {
 			tx_frame = UNALIGNED_GET((uint16_t *)(data->ctx.tx_buf));
 			LL_SPI_TransmitData16(spi, tx_frame);
-			spi_context_update_tx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		} else {
```
mathieuchopstm, line 513: Ditto:
```suggestion
		} else {
#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
```

drivers/spi/spi_stm32.c
```diff
@@ -410,35 +498,44 @@ static void spi_stm32_shift_m(const struct spi_stm32_config *cfg,
...
 			rx_frame = LL_SPI_ReceiveData8(spi);
 			UNALIGNED_PUT(rx_frame, (uint8_t *)data->ctx.rx_buf);
-			spi_context_update_rx(&data->ctx, 1, 1);
-		} else {
+		} else if (frame_size == 2U) {
 			rx_frame = LL_SPI_ReceiveData16(spi);
 			UNALIGNED_PUT(rx_frame, (uint16_t *)data->ctx.rx_buf);
-			spi_context_update_rx(&data->ctx, 2, 1);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		} else {
```
mathieuchopstm, line 533: Ditto:
```suggestion
		} else {
#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
```

drivers/spi/spi_stm32.c
```diff
@@ -551,7 +648,7 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 				      const uint8_t *tx_buf, uint8_t *rx_buf, uint32_t buf_len)
 {
 	struct spi_stm32_data *data = dev->data;
-	uint32_t size = buf_len / (SPI_WORD_SIZE_GET(config->operation) / BITS_PER_BYTE);
+	uint32_t size = buf_len / bits2bytes(config->operation);
```
mathieuchopstm, line 651: I was going to suggest that `bits2bytes` should return LOG2() (i.e., 0/1/2 instead of 1/2/4) and do:
```suggestion
	uint32_t size = buf_len >> bits2bytes(config->operation);
```
But GCC seems clever enough to [do this optimization automatically](https://godbolt.org/z/6PPq8MvYn), so I guess it's better to leave as-is.

drivers/spi/spi_stm32.c
```diff
@@ -1431,12 +1539,22 @@ static int transceive_dma(const struct device *dev,
 #endif /* SPI_SR_FTLVL */
 
 #ifdef CONFIG_SPI_STM32_ERRATA_BUSY
-		WAIT_FOR(ll_func_spi_dma_busy(spi) != 0,
+		WAIT_FOR(!ll_func_spi_dma_busy(spi),
```
mathieuchopstm, line 1542: Consider:
```suggestion
		(void)WAIT_FOR(!ll_func_spi_dma_busy(spi),
```
to make it clear that this is waiting for not busy *or* timeout

mathieuchopstm (approved):
LGTM. A few nits:

tbursztyka (approved):

PR #99325, opened 2025-11-13
drivers: eeprom: mb85rsxx: Adding support for new product ID

### LINKED ISSUE #99322: Add support for MB85RSMTA EEPROM IC
### Summary

Currently, the driver for the MB85RSxx EEPROM chip only supports the MB85RSMT chip. This chip is being replaced by the [MB85RSMTA](https://www.mouser.ca/datasheet/3/3800/1/MB85RS2MTA_DS6v0_E.pdf) chip - while functionally the same, this new chip has a different product ID (page 9 of the linked document), which is causing the device initialization to fail.

Add support for the new product ID of the updated IC.

### Alternatives

_No response_

### Additional Context

_No response_

siddhantmodi:
Fixes #99322 

Adds support for the 0x40 product ID needed by the MB85RSMTA version of the chip.

### DIFF
--- a/drivers/eeprom/eeprom_mb85rsxx.c
+++ b/drivers/eeprom/eeprom_mb85rsxx.c
@@ -45,9 +45,10 @@ LOG_MODULE_REGISTER(mb85rsxx, CONFIG_EEPROM_LOG_LEVEL);
  * MB85RSXX product ID (2 bytes); first byte provides memory size, so let's use a mask later when
  * checking it
  */
-#define EEPROM_MB85RSXX_PROD_ID		0x20U
-#define EEPROM_MB85RSXX_PROD_ID2	0x03U
-#define EEPROM_MB85RSXX_PROD_MASK	GENMASK(7, 5)
+#define EEPROM_MB85RSXX_PROD_ID1_MB85RS2MT	0x20U
+#define EEPROM_MB85RSXX_PROD_ID1_MB85RS2MTA	0x48U
+#define EEPROM_MB85RSXX_PROD_ID2	        0x03U
+#define EEPROM_MB85RSXX_PROD_MASK	        GENMASK(7, 5)
 
 struct eeprom_mb85rsxx_config {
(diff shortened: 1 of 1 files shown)


drivers/eeprom/eeprom_mb85rsxx.c
```diff
@@ -45,7 +45,8 @@ LOG_MODULE_REGISTER(mb85rsxx, CONFIG_EEPROM_LOG_LEVEL);
  * MB85RSXX product ID (2 bytes); first byte provides memory size, so let's use a mask later when
  * checking it
  */
-#define EEPROM_MB85RSXX_PROD_ID		0x20U
+#define EEPROM_MB85RSXX_PROD_ID_XX	0x20U
+#define EEPROM_MB85RSXX_PROD_ID_XXX	0x40U
```
henrikbrixandersen: Could we please pick some more descriptive definition names? What does `XX` and `XXX` mean?

henrikbrixandersen (changes requested):

drivers/eeprom/eeprom_mb85rsxx.c
sidmodi-nexera: Pushed a change with explicit IC names for the product IDs. This list can get pretty long as Fujitsu/Ramxeed have many products that can use this driver but have different product IDs.

I have only added the ID for the specific chip I am using at the moment as that is the only chip I have verified to be fully compatible with the existing driver. While most of the others **seem** to be compatible, they would need more vetting, so I propose that those are added on a per-use basis by developers that need to use them. Thoughts?

henrikbrixandersen: Sounds like a sound approach, thanks.

henrikbrixandersen (dismissed):

henrikbrixandersen (approved):

fabiobaltieri (approved):

PR #99326, opened 2025-11-13
drivers: video: Add pwdn pin support on ov2640

yunsmall:
The current ov2640 driver does not support pwdn pin support, so if there is a separate pin to control pwdn, there will be problems. It is necessary to add `zephyr,deferred-init` tags in the device tree to manually initialize, which is inconvenient. Tested on dnesp32s3 from alientek, it works. I also formatted the code with clangd.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -445,6 +445,9 @@ Video
   :kconfig:option:`CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE` which represent the
   size in byte allocated for the whole video buffer pool.
 
+* The :dtcompatible:`ovti,ov2640` reset pin handling has been corrected, resulting in an inverted
+  active level compared to before, to match the active level expected by the sensor.
+
 .. zephyr-keep-sorted-stop
 
 Bluetooth
--- a/drivers/video/ov2640.c
+++ b/drivers/video/ov2640.c
@@ -20,143 +20,143 @@
 LOG_MODULE_REGISTER(video_ov2640, CONFIG_VIDEO_LOG_LEVEL);
 
 /* DSP register bank FF=0x00*/
-#define QS                  0x44
-#define HSIZE               0x51
-#define VSIZE               0x52
-#define XOFFL               0x53
-#define YOFFL               0x54
-#define VHYX                0x55
-#define TEST                0x57
-#define ZMOW                0x5A
-#define ZMOH                0x5B
-#define ZMHH                0x5C
-#define BPADDR              0x7C
-#define BPDATA              0x7D
-#define SIZEL               0x8C
-#define HSIZE8              0xC0
-#define VSIZE8              0xC1
-#define CTRL1               0xC3
+#define QS     0x44
+#define HSIZE  0x51
+#define VSIZE  0x52
+#define XOFFL  0x53
+#define YOFFL  0x54
+#define VHYX   0x55
+#define TEST   0x57
+#define ZMOW   0x5A
+#define ZMOH   0x5B
+#define ZMHH   0x5C
+#define BPADDR 0x7C
+#define BPDATA 0x7D
+#define SIZEL  0x8C
+#define HSIZE8 0xC0
+#define VSIZE8 0xC1
+#define CTRL1  0xC3
 
-#define CTRLI               0x50
-#define CTRLI_LP_DP         0x80
+#define CTRLI       0x50
+#define CTRLI_LP_DP 0x80
 
-#define CTRL0               0xC2
-#define CTRL0_YUV422        0x08
-#define CTRL0_YUV_EN        0x04
-#define CTRL0_RGB_EN        0x02
+#define CTRL0        0xC2
+#define CTRL0_YUV422 0x08
+#define CTRL0_YUV_EN 0x04
+#define CTRL0_RGB_EN 0x02
 
-#define CTRL2               0x86
-#define CTRL2_DCW_EN        0x20
-#define CTRL2_SDE_EN        0x10
-#define CTRL2_UV_ADJ_EN     0x08
-#define CTRL2_UV_AVG_EN     0x04
-#define CTRL2_CMX_EN        0x01
+#define CTRL2           0x86
+#define CTRL2_DCW_EN    0x20
+#define CTRL2_SDE_EN    0x10
+#define CTRL2_UV_ADJ_EN 0x08
+#define CTRL2_UV_AVG_EN 0x04
+#define CTRL2_CMX_EN    0x01
 
-#define CTRL3               0x87
-#define CTRL3_BPC_EN        0x80
-#define CTRL3_WPC_EN        0x40
-#define R_DVP_SP            0xD3
-#define R_DVP_SP_AUTO_MODE  0x80
+#define CTRL3              0x87
+#define CTRL3_BPC_EN       0x80
+#define CTRL3_WPC_EN       0x40
+#define R_DVP_SP           0xD3
+#define R_DVP_SP_AUTO_MODE 0x80
 
-#define R_BYPASS                0x05
-#define R_BYPASS_DSP_EN         0x00
-#define R_BYPASS_DSP_BYPAS      0x01
+#define R_BYPASS           0x05
+#define R_BYPASS_DSP_EN    0x00
+#define R_BYPASS_DSP_BYPAS 0x01
 
-#define IMAGE_MODE              0xDA
-#define IMAGE_MODE_JPEG_EN      0x10
-#define IMAGE_MODE_RGB565       0x08
+#define IMAGE_MODE         0xDA
+#define IMAGE_MODE_JPEG_EN 0x10
+#define IMAGE_MODE_RGB565  0x08
 
-#define RESET                   0xE0
-#define RESET_JPEG              0x10
-#define RESET_DVP               0x04
+#define RESET      0xE0
+#define RESET_JPEG 0x10
+#define RESET_DVP  0x04
 
-#define MC_BIST                 0xF9
-#define MC_BIST_RESET           0x80
-#define MC_BIST_BOOT_ROM_SEL    0x40
+#define MC_BIST              0xF9
+#define MC_BIST_RESET        0x80
+#define MC_BIST_BOOT_ROM_SEL 0x40
 
-#define BANK_SEL                0xFF
-#define BANK_SEL_DSP            0x00
-#define BANK_SEL_SENSOR         0x01
+#define BANK_SEL        0xFF
+#define BANK_SEL_DSP    0x00
+#define BANK_SEL_SENSOR 0x01
 
 /* Sensor register bank FF=0x01*/
-#define COM1                0x03
-#define REG_PID             0x0A
-#define REG_PID_VAL         0x26
-#define REG_VER             0x0B
-#define REG_VER_VAL         0x42
-#define AEC                 0x10
-#define CLKRC               0x11
-#define COM10               0x15
-#define HSTART              0x17
-#define HSTOP               0x18
-#define VSTART              0x19
-#define VSTOP               0x1A
-#define AEW                 0x24
-#define AEB                 0x25
-#define ARCOM2              0x34
-#define FLL                 0x46
-#define FLH                 0x47
-#define COM19               0x48
-#define ZOOMS               0x49
-#define BD50                0x4F
-#define BD60                0x50
-#define REG5D               0x5D
-#define REG5E               0x5E
-#define REG5F               0x5F
-#define REG60               0x60
-#define HISTO_LOW           0x61
-#define HISTO_HIGH          0x62
+#define COM1        0x03
+#define REG_PID     0x0A
+#define REG_PID_VAL 0x26
+#define REG_VER     0x0B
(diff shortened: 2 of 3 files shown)


avolmat-st (changes requested):
Hello @yunsmall and thank you for this modification.

There is too many modifications at once in a single commit in order to be able to review it. I understand that there is not only the addition of the pwdn pin but also you've probably ran a tool to automatically fix codying style of the driver.

Looking the modifications in this commit, it seems to be necessary at some place, but in such case it should be done in a dedicated commit with the codying style fixes alone are done since doing several things at once in a commit doesn't allow us to review the real modifications.

Could you update your PR by splitting this into several commits, for ex:
1st commit: addition of the pwdn pin support (hence I think modification in ov2640.c and yaml file)
2nd commit: fix of major codying styles if necessary

The shield / dts for using this sensor on the dnesp32s3 is not part of Zephyr ? If so, I think it should also be updated to add the pwdn-pin now that this is available and allow the sensor to be used.

Thanks a lot for your help.

yunsmall:
@avolmat-st Thank you for your response for my PR. I did what you suggested to me, which is splitting the commit into two separated commits.
The board dnesp32s3 is an esp32s3 board released by Alientek, which is not a part of Zephyr. And I am also not a member of Alientek, so I suppose that the dts of this board should be added by their team.
Much thanks for this wonderful project.

drivers/video/ov2640.c
```diff
@@ -1008,9 +1016,17 @@ static int ov2640_init(const struct device *dev)
...
 
-#if DT_INST_NODE_HAS_PROP(0, reset_gpios)
-	const struct ov2640_config *cfg = dev->config;
+#if DT_INST_NODE_HAS_PROP(0, pwdn_gpios)
+	ret = gpio_pin_configure_dt(&cfg->pwdn_gpio, GPIO_OUTPUT_ACTIVE);
+	if (ret) {
+		return ret;
+	}
 
+	gpio_pin_set_dt(&cfg->pwdn_gpio, 0);
```
avolmat-st: Here, the pwdn pin is configured as output and set to active. But right after the gpio_pin_set_dt function is setting it to 0 (aka INACTIVE). Same thing can thus be done directly by simply calling

```c
ret = gpio_pin_configure_dt(&cfg->pwdn_gpio, GPIO_OUTPUT_INACTIVE);
```
and not having to call **gpio_pin_set_dt** at all.

I've seen that this is done same way for the reset pin actually, but this is quite useless call as well in fact.

drivers/video/ov2640.c
```diff
@@ -1008,9 +1016,17 @@ static int ov2640_init(const struct device *dev)
 		.height = SVGA_VSIZE,
 	};
 
-#if DT_INST_NODE_HAS_PROP(0, reset_gpios)
-	const struct ov2640_config *cfg = dev->config;
+#if DT_INST_NODE_HAS_PROP(0, pwdn_gpios)
+	ret = gpio_pin_configure_dt(&cfg->pwdn_gpio, GPIO_OUTPUT_ACTIVE);
+	if (ret) {
```
avolmat-st: Trend is now to ensure that if statement are explicit boolean. We do not change all code for the time being but at least try to enforce this for added / modified code. Could you thus apply this here.
```
if (ret < 0) {
```

avolmat-st (changes requested):
Thanks for the update and sorry for the delay of the review.

Small comments.

Could you also rework the commit messages of your 2 commits. You can have a look at Zephyr Contribution page for information about how to write commit messages:

https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-message-guidelines

Thanks a lot.

drivers/video/ov2640.c
yunsmall: Applied. But I thought that the redundant function calls for reset pin maybe can improve readability.

drivers/video/ov2640.c
yunsmall: Applied.

drivers/video/ov2640.c
yunsmall: 🤔Oh, the separated function call for reset pin is necessary, because reset pin is triggered at low level, so you'd better configure it to active then set it to 0 and then 1.

drivers/video/ov2640.c
```diff
@@ -20,418 +20,414 @@
...
 
 #define NUM_SATURATION_LEVELS (5)
 static const uint8_t saturation_regs[NUM_SATURATION_LEVELS + 1][5] = {
-	{ BPADDR, BPDATA, BPADDR, BPDATA, BPDATA },
-	{ 0x00, 0x02, 0x03, 0x28, 0x28 }, /* -2 */
-	{ 0x00, 0x02, 0x03, 0x38, 0x38 }, /* -1 */
-	{ 0x00, 0x02, 0x03, 0x48, 0x48 }, /*  0 */
-	{ 0x00, 0x02, 0x03, 0x58, 0x58 }, /* +1 */
-	{ 0x00, 0x02, 0x03, 0x58, 0x58 }, /* +2 */
+	{BPADDR, BPDATA, BPADDR, BPDATA, BPDATA}, {0x00, 0x02, 0x03, 0x28, 0x28}, /* -2 */
```
josuah: It looks like clang-format was not able to make something useful out of the first line, you might need to add a comment like `/* header */` on the line with `BPADDR`.

drivers/video/ov2640.c
```diff
@@ -463,16 +459,14 @@ struct ov2640_data {
...
-		.height_step = 0 \
-	}
+#define OV2640_VIDEO_FORMAT_CAP(width, height, format)                                             \
+	{.pixelformat = (format),                                                                  \
+	 .width_min = (width),                                                                     \
+	 .width_max = (width),                                                                     \
+	 .height_min = (height),                                                                   \
+	 .height_max = (height),                                                                   \
+	 .width_step = 0,                                                                          \
+	 .height_step = 0}
```
josuah: Here, `clang-format` did not wrap the braces in the conventional style.
This can be controlled by adding a trailing `,` after the `0` on the last line, before the `}`, and re-applying clang-format again.

drivers/video/ov2640.c
```diff
@@ -1077,8 +1095,14 @@ static int ov2640_init_0(const struct device *dev)
 	}
 #endif
 
-	uint32_t i2c_cfg = I2C_MODE_CONTROLLER |
-					I2C_SPEED_SET(I2C_SPEED_STANDARD);
+#if DT_INST_NODE_HAS_PROP(0, pwdn_gpios)
```
josuah, line 1091: The problem with this is that it only checks for the instance `0` (the first one).
If there are 2 sensors and only the 2nd has `pwdn_gpios`, then this would not work.

You might need something like this:

```suggestion
#if DT_ANY_INST_HAS_PROP_STATUS_OKAY(pwdn_gpios)
	if (cfg->pwdn_gpios.port != NULL) {
```

The `#if` helps making `cfg->pwdn_gpios` compile.

The `if ()` helps detecting the per-instance presence/absence of a GPIO port.

You might need to replicate this for each occurence of `DT_INST_NODE_HAS_PROP(0, pwdn_gpios)`.

Would that be ok?
Thank you!

josuah (commented):
About the commit message: would it be possible to wrap the text to 75 characters like described [here](https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-message-body)? I think the CI bot will complain otherwise...

Thank you for the improvements!

drivers/video/ov2640.c
avolmat-st: I didn't mentioned it since anyway the whole driver doesn't support multi instance for the time being maybe that would be more for a dedicated commit to make this driver multi-instance capable ? (not only pwdn but also reset, config/data structures etc.

drivers/video/ov2640.c
avolmat-st: Ok maybe not correct the reset pin handling now (or at least in the same commit as the one adding the pwdn pin handling) but I think this actually highlight a probably incorrect configuration of the reset pin in the DT. What the driver does is:

```c
#if DT_INST_NODE_HAS_PROP(0, reset_gpios)
        const struct ov2640_config *cfg = dev->config;

        ret = gpio_pin_configure_dt(&cfg->reset_gpio, GPIO_OUTPUT_ACTIVE);
        if (ret) {
                return ret;
        }

        gpio_pin_set_dt(&cfg->reset_gpio, 0);
        k_sleep(K_MSEC(1));
        gpio_pin_set_dt(&cfg->reset_gpio, 1);
        k_sleep(K_MSEC(1));
#endif
```

Meaning, this configure the reset GPIO at it's active state (active state is actually what has been written in the device-tree via ACTIVE_HIGH or ACTIVE_LOW). Then this set it to INACTIVE, then again to ACTIVE.
You mentioned that the reset is actually ACTIVE_LOW (triggered at low level).
In this driver, it make much more sense to say that we perform the reset (aka make it ACTIVE, then make it INACTIVE). So it should be

```c
#if DT_INST_NODE_HAS_PROP(0, reset_gpios)
        const struct ov2640_config *cfg = dev->config;

        ret = gpio_pin_configure_dt(&cfg->reset_gpio, GPIO_OUTPUT_ACTIVE);
        if (ret) {
                return ret;
        }
        k_sleep(K_MSEC(1));
        gpio_pin_set_dt(&cfg->reset_gpio, 0);
        k_sleep(K_MSEC(1));
#endif
```

and ensure that in the device-tree, the GPIO used for reset-gpio is marked as ACTIVE_LOW. (I think it might be written as ACTIVE_HIGH currently)

drivers/video/ov2640.c
yunsmall: So should I not change this? This driver does indeed use the object instantiated by DEVICE_DT_INST_DEFINE(0,...) at the end of the file, indicating that this driver only supports a single device.

drivers/video/ov2640.c
yunsmall: Actually I marked `reset-gpio` as `GPIO_ACTIVE_HIGH` in my dts file, otherwise my ov2640 will not be initialized properly, although in binding file it says
> The RESETn pin is asserted to disable the sensor causing a hard 
reset.  The sensor receives this as an active-low signal.

drivers/video/ov2640.c
josuah: Good point, then that does not matter for now, it is possible to make the improvement in a new PR.
Thanks!

doc/releases/migration-guide-4.4.rst
```diff
@@ -160,14 +160,17 @@ Stepper
 USB
 ===
 
-  * :dtcompatible:`maxim,max3421e_spi` has been renamed to :dtcompatible:`maxim,max3421e-spi`.
+* :dtcompatible:`maxim,max3421e_spi` has been renamed to :dtcompatible:`maxim,max3421e-spi`.
```
avolmat-st: I think it is better to avoid from making modifications in unrelated part of the migration-guide and only add the entry related to the modification introduced in this PR only.

doc/releases/migration-guide-4.4.rst
```diff
@@ -160,14 +160,17 @@ Stepper
...
 
 .. zephyr-keep-sorted-stop
 
 Video
-===
+=====
```
avolmat-st, line 441: Here as well, no need to modify the section title

doc/releases/migration-guide-4.4.rst
```diff
@@ -160,14 +160,17 @@ Stepper
...
 
 * CONFIG_VIDEO_OV7670 is now gone and replaced by CONFIG_VIDEO_OV767X.  This allows supporting both the OV7670 and 0V7675.
+* For :dtcompatible:`ovti,ov2640`, the initialization error of the reset pin has been corrected.
+  If your device tree includes property ``reset-gpios``, please ensure that the configuration of
+  this pin includes ``GPIO_ACTIVE_LOW``. If the configuration includes ``GPIO_ACTIVE_HIGH``, please correct it.
```
avolmat-st: What about simply

```suggestion
The :dtcompatible:`ovti,ov2640` reset pin handling has been corrected, resulting in an inverted active level compared to before, to match the active level expected by the sensor.
```

avolmat-st (changes requested):
Thanks a lot for the added fix and addition in the migration guide. Looks good to me.

I made very small comments.

Also, regarding the 1st commit log, what about simply mentioning something like

```
drivers: video: ov2640: Add pwdn pin support on ov2640.

Add handling for the power-down (pwdn) pin of the OV2640 sensor.

Signed-off-by: yunsmall <yun_small@163.com>
```

doc/releases/migration-guide-4.4.rst
yunsmall: Sorry, it's my fault. I saw that there were formatting issues with these things and couldn't resist fixing them.😂

avolmat-st (approved):
LGTM, Thanks a lot for this work. Approved.

avolmat-st (dismissed):
LGTM.

josuah (dismissed):

avolmat-st:
@yunsmall, could you rebase this PR in order to fix conflict and be able to merge it ?

yunsmall:
OK.

josuah (approved):

avolmat-st (approved):
Thanks for the update. LGTM.

PR #99332, opened 2025-11-13
drivers: serial: ns16550: support loopback mode and interrupt API

danieldegrasse:
This PR includes several fixes for the ns16550 uart driver:

- Add support for loopback mode in ns16550 uart driver, by setting LOOP
bit in MDC register.
- correctly support the uart interrupt API for IP variants that feature the USR register. The base ns16550 IP spec does not seem to be able to support the interrupt API correctly, since it should be possible to call `uart_irq_pending` multiple times after calling `uart_irq_update` (as far as I can tell from reading the UART API)
- support odd parity in the driver

Signed-off-by: Daniel DeGrasse <ddegrasse@tenstorrent.com>

### DIFF
--- a/drivers/serial/uart_ns16550.c
+++ b/drivers/serial/uart_ns16550.c
@@ -245,6 +245,11 @@ BUILD_ASSERT(IS_ENABLED(CONFIG_PCIE), "NS16550(s) in DT need CONFIG_PCIE");
 #define MSR_RI 0x40   /* complement of ring signal */
 #define MSR_DCD 0x80  /* complement of dcd */
 
+/* constants for uart status register */
+#define USR_TFNF 0x02 /* Transmit FIFO Not Full */
+#define USR_TFE 0x04 /* Transmit FIFO Empty */
+#define USR_RFNE 0x08 /* Receive FIFO Not Empty */
+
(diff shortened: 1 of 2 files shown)


drivers/serial/uart_ns16550.c
```diff
@@ -1255,7 +1284,18 @@ static int uart_ns16550_irq_is_pending(const struct device *dev)
 	struct uart_ns16550_dev_data *data = dev->data;
 	k_spinlock_key_t key = k_spin_lock(&data->lock);
 
+#ifdef CONFIG_UART_NS16550_DW8250_DW_APB
```
danieldegrasse, line 1287: The reason for this change is that the `uart_elementary` test calls `uart_irq_update` once, then calls `uart_irq_pending`. From my read of the UART API this is allowed, but this usage will break the current implementation of `uart_irq_pending` since the UART driver does not update the ISR status outside of `uart_irq_pending`.

Not sure if the `uart_elementary` use is correct though, because other implementations like the shell call `uart_irq_tx_ready` once per call of `uart_irq_update`.

danieldegrasse:
@dcpleung PTAL

dcpleung (approved):

maass-hamburg (approved):

PR #99336, opened 2025-11-13
arch: arm: Add non-cacheable shared memory protection region

MaureenHelm:
Adds support for defining a memory region with non-cacheable, sharable memory attributes in the devicetree and configuring a corresponding memory protection region in the arm mpu.


include/zephyr/arch/arm/mpu/arm_mpu_v8.h
```diff
@@ -271,6 +271,14 @@
 		IF_ENABLED(CONFIG_ARM_MPU_PXN, (.pxn = PRIV_EXEC_NEVER,))			   \
 	}
 
+#define REGION_RAM_NOCACHE_OUT_SHARABLE_ATTR(base, size)                                           \
+	{                                                                                          \
+		.rbar = NOT_EXEC | P_RW_U_RW_Msk | OUTER_SHAREABLE_Msk, /* AP, XN, SH */           \
```
JarmouniA, line 276: Shareable non-cached memory does not make much sense to me, shareability is strongly related to cache coherency
https://developer.arm.com/documentation/dui1095/a/The-Cortex-M23-Processor/Memory-model/Memory-regions--types--and-attributes

wearyzen: Normal memory can be shareable or Non-shareable and Normal memory can be cached or non-cached (https://developer.arm.com/documentation/107565/0101/Memory-protection/Attribute-indirection/Memory-Attribute-Indirection-Register--MAIR0-MAIR1-?lang=en) so this should be ok.
I am worried about the number of bits we are running out of in include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h.
I think we'll have to redefine them in a way to allow all configurations mentioned in https://developer.arm.com/documentation/107565/0101/Memory-protection/Attribute-indirection/Memory-Attribute-Indirection-Register--MAIR0-MAIR1-?lang=en but I can do that in a different PR.

include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h
```diff
@@ -40,6 +41,7 @@
 #define DT_MEM_ARM_MPU_EXTMEM      DT_MEM_ARM(ATTR_MPU_EXTMEM)
 #define DT_MEM_ARM_MPU_RAM_PXN     DT_MEM_ARM(ATTR_MPU_RAM_PXN)
 #define DT_MEM_ARM_MPU_DEVICE      DT_MEM_ARM(ATTR_MPU_DEVICE)
+#define DT_MEM_ARM_MPU_RAM_NOCACHE_OUT_SHARABLE DT_MEM_ARM(ATTR_MPU_RAM_NOCACHE_OUT_SHARABLE)
```
wearyzen, line 44:
```suggestion
#define DT_MEM_ARM_MPU_RAM_OUTER_NOCACHE_SHARABLE DT_MEM_ARM(ATTR_MPU_RAM_OUTER_NOCACHE_SHARABLE)
```

include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h
```diff
@@ -31,6 +31,7 @@
 #define ATTR_MPU_EXTMEM      BIT(5)
 #define ATTR_MPU_RAM_PXN     BIT(6)
 #define ATTR_MPU_DEVICE      BIT(7)
+#define ATTR_MPU_RAM_NOCACHE_OUT_SHARABLE BIT(8)
```
wearyzen, line 34:
```suggestion
#define ATTR_MPU_RAM_OUTER_NOCACHE_SHARABLE BIT(8)
```

include/zephyr/arch/arm/mpu/arm_mpu_v8.h
wearyzen: REGION_IO_ATTR  looks more relevant in this context so I would prefer https://github.com/zephyrproject-rtos/zephyr/pull/100349 over this PR.

include/zephyr/dt-bindings/memory-attr/memory-attr-arm.h
mathieuchopstm, line 1:
```
 * This is legacy and it should NOT be extended further. If new MPU region
 * types must be added, these must rely on the generic memory attributes.
```

Though I do agree this means basically reworking the *entire* implementation of ARM MPU... (I have ideas stashed somewhere, I could share them if someone feels like doing it, or attempt to free some bandwidth and have a look)

mathieuchopstm (changes requested):

PR #99350, opened 2025-11-14
mctp: host/endpoint UART sample fixes

edersondisouza:
Some fixes are functional, but there are also some cosmetic ones. More details on each commit.


teburd (approved):
Thanks for fixing these up!

dcpleung (approved):

PR #99360, opened 2025-11-14
Add counter alarm callback

ZhaoxiangJin:
When low-power mode timer is enabled, a counter alarm is set on the idle timer before entering low-power mode. The counter API requires a non-NULL callback function for the alarm. This PR adds a stub callback function to satisfy the API requirement.

### DIFF
--- a/drivers/timer/Kconfig.cortex_m_systick
+++ b/drivers/timer/Kconfig.cortex_m_systick
@@ -97,7 +97,7 @@ endif # PM
 endchoice # CORTEX_M_SYSTICK_LPM_TIMER
 
 config CORTEX_M_SYSTICK_RESET_BY_LPM
-	bool
+	bool "SysTick reset by low-power mode"
 	depends on !CORTEX_M_SYSTICK_LPM_TIMER_NONE
 	help
 	  Some SoCs provide one or more low-power mode which, as part of their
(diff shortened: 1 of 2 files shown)


drivers/timer/Kconfig.cortex_m_systick
```diff
@@ -97,7 +97,7 @@ endif # PM
 endchoice # CORTEX_M_SYSTICK_LPM_TIMER
 
 config CORTEX_M_SYSTICK_RESET_BY_LPM
-	bool
+	bool "SysTick reset by low-power mode"
```
mathieuchopstm, line 100: Any reason for making this option visible? My rationale was that it is a fixed(ish*) hardware parameter so there's no reason for the user to modify it.

(\*After thinking a bit, I guess one may argue that if *some* but not *all* low-power modes kill SysTick, one may want to select this only if these particular low-power modes are enabled - but that seems like more trouble than its worth)

mathieuchopstm (approved):

drivers/timer/Kconfig.cortex_m_systick
ZhaoxiangJin: yes, I'm enabling PM on an MCU, the Systick is only reset in the deepest power mode; it's retained in other sleep modes.

Holt-Sun (approved):

wearyzen (approved):

PR #99377, opened 2025-11-14
soc: nordic: common: nrf_sys_event: Add API for registering an event

nordic-krch:
Add API for indicating that a deterministic interrupt will occur at certain point in time in the future. Implementation is ensuring that there will be no latency due to NVM memory waking up. There are 2 ways of ensuring that:
 - setting low latency power mode in RRAMC (higher power consumption
    in idle (not available in non-secure build)
 - using PPI and GRTC to trigger the RRAMC wake up task right before
      an expected interrupt

Module has a pool of GRTC channels and dynamically allocates and frees those channels when events are registered and unregistered.

If GRTC channel is not available then algorithm falls back to power mode setting (in secure build).

API offers registering an event using relative and absolute timing.
API can be used from Zero Latency interrupts.

# Performance

Registering a relative event takes ~110 cycles when GRTC is used and ~45 cycles when RRAMC power mode is changed. Unregistering is ~40 cycles. 

# nrf_sys_event vs PM events API

API looks similar to PM events API and in future it might be used by PM but we want to keep it as standalone API due to:
- microsecond precision that is required (PM events work with ticks)
- ability to use it without PM
- ability to use it in no multi-threading applications

~~`DNM` as it is based on #99568.~~


bjarki-andreasen (dismissed):
This is very specific to nordic hardware so it makes sense to keep it in soc, zephyr's services work in ticks, and with threads, anything that requires the level of resolution and direct use of specific hardware will not be supported abstractly.

bjarki-andreasen (approved):

kl-cruz (approved):

PR #99384, opened 2025-11-14
Bluetooth: CSIP: Only allow rank changes with set size changes

Thalley:
The rank cannot be notified and the function allowed for just updated to the rank. The CSIS spec does not mention the usecase, nor support, for changing the rank of a device. The spec only seems to allow for dynamic changes to the set size (and only during that can the rank be set).


Thalley:
CI issue is unrelated to this change

asbjornsabo (approved):
LGTM

larsgk (approved):

PR #99385, opened 2025-11-14
cmake: modules: boards: Fix board moved check

nordicjm:
Fixes checking for board folders that have moved by checking if the BOARD_DIR variable is already set (i.e. loaded from cache)

### DIFF
--- a/cmake/modules/boards.cmake
+++ b/cmake/modules/boards.cmake
@@ -141,7 +141,7 @@ Hints:
   endif()
 endforeach()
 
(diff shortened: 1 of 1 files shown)


pdgendt (approved):

tejlmand (approved):

henrikbrixandersen (approved):
Thanks, saved me some debugging.

jhedberg:
Would it make sense to backport this to v4.3-branch? (if so, please add the appropriate label)

PR #99388, opened 2025-11-14
kernel: kconfig: Disable DEVICE_DEINIT_SUPPORT by default

nordicjm:
This Kconfig, which it itself admits is for a "very specific case" was set to default as yes, this includes extra code in drivers with this functionality and increases driver struct size for cases where this function isn't needed (i.e. all because it's enabled by default), therefore change it to be opt-in rather than opt-out

Depends on https://github.com/zephyrproject-rtos/zephyr/pull/99391


nashif (dismissed):

peter-mitsis:
I'm ok with setting this as disabled by default. It just looks like there are some drivers that will need to be tidied up as there are now instances of functions being defined but no longer used.

tmleman (dismissed):

bjarki-andreasen (dismissed):

VynDragon (approved):

peter-mitsis (approved):

PR #99394, opened 2025-11-14
tests: smf: use BIT() for flat state flags

vlad-kulikov:
Replace manual `1 << n` flag definitions in the flat SMF test with the common `BIT(n)` macro. This aligns the test with other SMF test.


glenn-andrews (approved):
LGTM

faxe1008 (approved):

keith-zephyr (approved):

PR #99398, opened 2025-11-14
arch: common: move lib/acpi into arch/common/

nashif:
acpi is not really library code based on the new definition of what
should go into lib/. Move acpi into arch/x86/ where it is being used
exclusively.

Signed-off-by: Anas Nashif <anas.nashif@intel.com>

see https://github.com/zephyrproject-rtos/zephyr/issues/99250#issuecomment-3527963623


edersondisouza (changes requested):
On commit message: arch/x86 -> arch/common

edersondisouza (approved):

henrikbrixandersen (approved):

PR #99408, opened 2025-11-14
tests: Bluetooth: Tester: Fix issues with ISO tests

Thalley:
Fixes some issues with the ISO and Audio tests for the BT Tester, as data was dropped due to low buffers and slow speed.


cvinayak (approved):

tests/bluetooth/tester/prj.conf
```diff
@@ -47,6 +47,8 @@ CONFIG_BT_ISO_BROADCASTER=y
 CONFIG_BT_ISO_SYNC_RECEIVER=y
 CONFIG_BT_ISO_CENTRAL=y
 CONFIG_BT_ISO_PERIPHERAL=y
+CONFIG_BT_ISO_RX_BUF_COUNT=4
+CONFIG_BT_ISO_TX_BUF_COUNT=4
```
sjanc, line 51: those are already set to those value in overlay-le-audio.conf
I'd suggest to remove them from there if already set here

Thalley: This is for the GAP/ISO tests. The LE Audio overlay isn't used by those tests

sjanc: yes, but prj.conf is always included so no need to have those also in leaudio overlay

Thalley: That is a fair point; as long as LE Audio don't need to to modify the values, we could omit them from there. 

Does that mean that you'd prefer the overlay to not contain any values already set by `prj.conf`? If so, then I suggest we do a followup for that, as today there are a significant amount of duplicated values

sjanc: fair enough :)

sjanc (approved):

PR #99413, opened 2025-11-14
LLEXT: add discarded symbol groups inspector

mathieuchopstm:
Follow-up to #98990 (c.f. https://github.com/zephyrproject-rtos/zephyr/pull/98990#discussion_r2499663423)

When the LLEXT subsystem detects that a symbol is marked for export from the main image, but the symbol group in which the export belongs is not enabled, add a marker in a special section in addition to not placing the symbol in the final export table.

Add a post-build script which consumes the special section as post-build to dump information about all discarded symbols in a build artifact, and also check for common errors (missing Kconfig symbol / improperly named groups).

While at it, rework LLEXT scripts organization a little bit - maybe a dedicated `scripts/llext` directory is starting to make sense though... I also wanted to create a `${PROJECT_BINARY_DIR}/llext` directory for all artifacts, but I'm not super confident about how to ensure CMake will create said directory, so the report is in artifacts root for now (like SLID listings).

There's also a bit more clean-up I found along the way (namely: a single `llext_discarded_strtab` could replace the two sections `llext_discarded_exports_strtab`+`llext_exports_strtab`, and `read_str` could possibly be shared between scripts), but that can always come later.

Example output on `west build samples/subsys/llext/modules`:
<details>

```
This file lists all symbols that have been marked as LLEXT exports
but will not be added to the export table because their symbol group
has not been enabled. Refer to LLEXT Documentation for more details.

====================================================================================================
                                            Group DEVICE                                            
====================================================================================================
Symbol Name                                                  | Address
--------------------------------------------------------------------------------
__device_dts_ord_36                                          | 0800744c
__device_dts_ord_39                                          | 0800748c
__device_dts_ord_47                                          | 0800736c
__device_dts_ord_57                                          | 0800738c
__device_dts_ord_61                                          | 080074ac
__device_dts_ord_64                                          | 080074cc
__device_dts_ord_72                                          | 0800746c
__device_dts_ord_9                                           | 0800734c
__device_dts_ord_90                                          | 0800742c
__device_dts_ord_91                                          | 0800740c
__device_dts_ord_92                                          | 080073ec
__device_dts_ord_93                                          | 080073cc
__device_dts_ord_94                                          | 080073ac
```

</details>

Example build log + output when a group Kconfig is missing:
<details>

```
...
           FLASH:       39728 B       512 KB      7.58%
             RAM:        7752 B        96 KB      7.89%
           SRAM0:          0 GB        96 KB      0.00%
        IDT_LIST:          0 GB        32 KB      0.00%
Generating files from /local/zprj/builds/default/zephyr/zephyr.elf for board: nucleo_f401re
WARNING: Could not find Kconfig symbol CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_NONEXISTENT
```
```
====================================================================================================
                                         Group NONEXISTENT                                          
Could not find Kconfig symbol CONFIG_LLEXT_EXPORT_SYMBOL_GROUP_NONEXISTENT:
symbol group will never be included in the LLEXT export table!
====================================================================================================
Symbol Name                                                  | Address
--------------------------------------------------------------------------------
example_fn                                                   | 080054e3
```

</details>

### DIFF
--- a/scripts/build/llext_elf_toolkit.py
+++ b/scripts/build/llext_elf_toolkit.py
@@ -7,10 +7,10 @@
 ELF utilities shared by LLEXT scripts
 """
 
+import struct
+
 import elftools.elf.elffile
 import elftools.elf.sections
-
-import struct
 
 
 def get_target_specific_structure(struct_desc: str, ptr_size: int, endianness: str):
@@ -27,7 +27,7 @@ def get_target_specific_structure(struct_desc: str, ptr_size: int, endianness: s
         ptr_size: Pointer size of target, in bytes (4/8)
         endianness: Endianness of target ('little'/'big')
     """
-    assert ptr_size == 4 or ptr_size == 8
+    assert ptr_size in (4, 8)
     assert endianness in ['little', 'big']
 
     endspec = "<" if endianness == 'little' else ">"
--- a/scripts/build/llext_elf_toolkit.py
+++ b/scripts/build/llext_elf_toolkit.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2025 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+ELF utilities shared by LLEXT scripts
+"""
+
+import elftools.elf.elffile
+import elftools.elf.sections
+
+import struct
+
+
+def get_target_specific_structure(struct_desc: str, ptr_size: int, endianness: str):
+    """
+    Obtain target-specific 'struct.Struct' from structure description string.
+    Internally, pointer-sized elements ('P') is replaced with an integer field
+    of the proper size, and the endianness indicator is prepended automatically.
+
+    Parameters:
+        struct_template:
+            Python 'struct' descriptor, with 'P' for pointer-sized fields
+            and without an endianness indicator
+
+        ptr_size: Pointer size of target, in bytes (4/8)
+        endianness: Endianness of target ('little'/'big')
+    """
+    assert ptr_size == 4 or ptr_size == 8
+    assert endianness in ['little', 'big']
+
+    endspec = "<" if endianness == 'little' else ">"
+    if ptr_size == 4:
+        ptrspec = "I"
+    elif ptr_size == 8:
+        ptrspec = "Q"
+
+    return struct.Struct(endspec + struct_desc.replace("P", ptrspec))
+
+
+# ELF Shdr flag applied to the export table section, to indicate
+# the section has already been prepared by this script. This is
+# mostly a security measure to prevent the script from running
+# twice on the same ELF file, which can result in catastrophic
+# failures if SLID-based linking is enabled (in this case, the
+# preparation process is destructive).
+#
+# This flag is part of the SHF_MASKOS mask, of which all bits
+# are "reserved for operating system-specific semantics".
+# See: https://refspecs.linuxbase.org/elf/gabi4+/ch4.sheader.html
+SHF_LLEXT_PREPARATION_DONE = 0x08000000
+
+
+class SectionDescriptor:
+    """ELF Section descriptor
+
+    This is a wrapper class around pyelftools' "Section" object.
+
+    Attributes:
+        name: Section name
+        size: Section size
+        flags: Section flags
+        offset: File offset to section
+        shdr_index: Section header index
+        shdr_offset: File offset to section header
+        section: pyelftools Section object
+    """
+
+    def __init__(self, elffile: elftools.elf.elffile.ELFFile, section_name: str):
+        self.name = section_name
+        self.section = elffile.get_section_by_name(section_name)
+        if not isinstance(self.section, elftools.elf.sections.Section):
+            raise KeyError(f"section {section_name} not found")
+
+        self.shdr_index = elffile.get_section_index(section_name)
+        self.shdr_offset = elffile['e_shoff'] + self.shdr_index * elffile['e_shentsize']
+        self.size = self.section['sh_size']
+        self.flags = self.section['sh_flags']
+        self.offset = self.section['sh_offset']
--- a/scripts/build/llext_inspect_discarded_groups.py
+++ b/scripts/build/llext_inspect_discarded_groups.py
@@ -1,0 +1,255 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2025 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Inspects Zephyr build artifacts (ELF, plus Kconfig if available)
+to display information about symbols placed in the discard exports
+table due to their symbol group not being enabled.
+"""
+
+import argparse
+import logging
+import re
+import sys
+from pathlib import Path
+
+import llext_elf_toolkit
+from elftools.elf.elffile import ELFFile
+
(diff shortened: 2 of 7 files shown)


scripts/build/llext_inspect_discarded_groups.py
```diff
@@ -0,0 +1,263 @@
...
+            elif value == "m":
+                value = 2
+            elif value[0] == value[-1] == '"':
+                value = value.strip('"')
+            else:
+                value = int(value, base=0)
+
+            symbols[key] = value
+
+    return symbols
```
mathieuchopstm: Originally, I thought about using `kconfiglib` like a good citizen, but it's much less trivial to use than I hoped... c.f.: https://github.com/zephyrproject-rtos/zephyr/blob/9c1fbc8a86ad0f57190b6b84dd19bd88a189dc49/cmake/modules/kconfig.cmake#L138-L163

mathieuchopstm:
> Does this result in symbols that are otherwise unreferenced still being included in the build due to the `addr` reference in `Z_LLEXT_DISCARD_STRTAB`? e.g does `function_used_nowhere` now exist in the final `.hex`?
> 
> ```
> static int function_used_nowhere(int x)
> {
>    return 2* x;
> }
> 
> LLEXT_GROUP_SYMBOL(DISABLED, function_used_nowhere);
> ```

Good question; I did not think about it, but I would assume that yes. I'll take a closer look and drop the `addr` if so since it isn't *that* important.

scripts/build/llext_elf_toolkit.py
```diff
@@ -0,0 +1,80 @@
...
+import struct
+
+import elftools.elf.elffile
+import elftools.elf.sections
+
+
+def get_target_specific_structure(struct_desc: str, ptr_size: int, endianness: str):
+    """
+    Obtain target-specific 'struct.Struct' from structure description string.
+    Internally, pointer-sized elements ('P') is remplaced with an integer field
```
teburd:
```suggestion
    Internally, pointer-sized elements ('P') is replaced with an integer field
```

JordanYates (changes requested):
Linking in the `z_llext_discarded_const_symbol` values should be gated behind a Kconfig. One reason to disable groups was to reduce ROM consumption, but this debugging feature adds back in all the discarded function names in ROM unless I am mistaken.

mathieuchopstm:
> Linking in the `z_llext_discarded_const_symbol` values should be gated behind a Kconfig. One reason to disable groups was to reduce ROM consumption, but this debugging feature adds back in all the discarded function names in ROM unless I am mistaken.

This is partially correct and partially incorrect. It is true that the strings will now be found ***in the ELF file***. However, the discarded export table (and corresponding strtab) are placed in a special output section with type `INFO` - per GNU ld documentation:
> `INFO` [...] The section should be marked as not allocatable, so that no memory is allocated for the section when the program is run.

In practice, this means that the section is discarded when the ELF is flattened to `.hex`/`.bin`/etc (this is how SLID works too, by the way):

<details>

```
$ wb samples/subsys/llext/modules/ -b nucleo_f429zi -p
...
$ cd BUILDDIR/zephyr
$ cat llext_discarded_groups.txt 
This file lists all symbols that have been marked as LLEXT exports
but will not be added to the export table because their symbol group
has not been enabled. Refer to LLEXT Documentation for more details.

====================================================================================================
                                            Group DEVICE                                            
====================================================================================================
Symbol Name
--------------------------------------------------------------------------------
__device_dts_ord_122
__device_dts_ord_123
__device_dts_ord_124
__device_dts_ord_125
__device_dts_ord_126
__device_dts_ord_127
__device_dts_ord_128
__device_dts_ord_37
__device_dts_ord_40
__device_dts_ord_48
__device_dts_ord_57
__device_dts_ord_62
__device_dts_ord_72
__device_dts_ord_80
__device_dts_ord_83
__device_dts_ord_9

$ strings zephyr.bin | grep __device_dts_ord_122
$ strings zephyr.bin | grep hello_world                    <--- this is exported by sample
hello_world
```

</details>

But for the sake of it, I have done two builds with and without the commits and both result in identical linker size report (feel free to test on your side too):
<details>

```
[172/172] Linking C executable zephyr/zephyr.elf
           FLASH:       40748 B         2 MB      1.94%
             RAM:        7816 B       192 KB      3.98%
           SRAM0:          0 GB       192 KB      0.00%
            DTCM:          0 GB        64 KB      0.00%
        IDT_LIST:          0 GB        32 KB      0.00%
Generating files from /local/zprj/builds/default/zephyr/zephyr.elf for board: nucleo_f429zi
(zprj) Z~/$ gitutil-pop-head-commit 
llext: add discarded symbol groups inspector

When the LLEXT subsystem detects that a symbol is marked for export from
the main image, but the symbol group in which the export belongs is not
enabled, add a marker in a special section in addition to not placing the
symbol in the final export table.

Add a post-build script which consumes the special section as post-build to
dump information about all discarded symbols in a build artifact, and also
check for common errors (missing Kconfig symbol / improperly named groups).

Signed-off-by: Mathieu Choplain <mathieu.choplain-ext@st.com>

(zprj) Z~/$ gitutil-pop-head-commit 
scripts: build: llext: move ELF utilities to dedicated script

Move ELF-related utilities from the "llext_prepare_exptab" script to a
dedicated "llext_elf_toolkit" which can be reused by other LLEXT scripts.
While at it, also improve documentation of class SectionDescriptor.

Signed-off-by: Mathieu Choplain <mathieu.choplain-ext@st.com>

$ <delete all staged files>
$ wb samples/subsys/llext/modules/ -b nucleo_f429zi -p
...
[172/172] Linking C executable zephyr/zephyr.elf
           FLASH:       40748 B         2 MB      1.94%
             RAM:        7816 B       192 KB      3.98%
           SRAM0:          0 GB       192 KB      0.00%
            DTCM:          0 GB        64 KB      0.00%
        IDT_LIST:          0 GB        32 KB      0.00%
```

</details>

mathieuchopstm:
@JordanYates could you take a look again please?

scripts/build/llext_inspect_discarded_groups.py
teburd, line 1: This is quite unfortunate that we need yet another 250 loc python script to do this :(

teburd (dismissed):

scripts/build/llext_inspect_discarded_groups.py
mathieuchopstm: The core is really around 100~150 LoC - quite a few things could be factorized out and shared with other scripts (e.g., the dotconfig parser is ~50 LoC, `_init_log` is 15 LoC, the `read_str` helper is 15 LoC...). That's not specific to LLEXT scripts as a matter of fact.

Maybe there should be some effort to go around all scripts and see what can be done smarter, but that's a outside the PR's scope. I did try to align with existing scripts such that, if a cleanup-and-refactoring effort does occur, the script should be "as easy as others" to migrate.

scripts/build/llext_elf_toolkit.py
mathieuchopstm: Good catch! Fixed.

mathieuchopstm:
removed an outdated comment

teburd (approved):

scripts/build/llext_elf_toolkit.py
kartben: is -> are, also :)

JordanYates (approved):

PR #99415, opened 2025-11-14
west: blobs: list/clean `--allow-regex` filter

noahp:
Enable filtering the blobs selected by `west blobs list` and `west blobs clean` with a regex pattern, just like `west blobs fetch`.

Support this when checking for blob existence in `drivers/wifi` for esp32, so now it's possible to build for a selected esp32 family SOC (for example `esp32c6`) with only the blobs for that chip downloaded:

```bash
west blobs clean hal_espressif  # delete all hal_espressif blobs
west blobs fetch hal_espressif --allow-regex '.*esp32c6.*'
west build -b esp32c6_devkitc/esp32c6/hpcore zephyr/samples/net/wifi/shell
```

### DIFF
--- a/cmake/modules/extensions.cmake
+++ b/cmake/modules/extensions.cmake
@@ -1816,10 +1816,11 @@ endfunction()
 # fatal error message is printed (depends on REQUIRED flag).
 #
 # Usage:
-#   zephyr_blobs_verify(<MODULE module|FILES file [files...]> [REQUIRED] [REGEX])
+#   zephyr_blobs_verify(<MODULE module|FILES file [files...]> [REQUIRED] [REGEX regex_pattern])
 #
 # Example:
 # zephyr_blobs_verify(MODULE my_module REQUIRED) # verify all blobs in my_module and fail on error
+# zephyr_blobs_verify(MODULE my_module REGEX ".*esp32c6.*" REQUIRED) # verify only esp32c6 blobs
 # zephyr_blobs_verify(FILES img/file.bin)        # verify a single file and print on error
 function(zephyr_blobs_verify)
   cmake_parse_arguments(BLOBS_VERIFY "REQUIRED" "MODULE;REGEX" "FILES" ${ARGN})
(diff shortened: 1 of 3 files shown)


drivers/wifi/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ if(CONFIG_BUILD_ONLY_NO_BLOBS)
 else()
 
 if(CONFIG_DT_HAS_ESPRESSIF_ESP32_WIFI_ENABLED)
-  zephyr_blobs_verify(MODULE hal_espressif REQUIRED)
+message(WARNING "Verifying presence of Espressif ESP32 Wi-Fi binary blobs.")
```
tejlmand, line 16: leftover debug message ?
```suggestion
```

cmake/modules/extensions.cmake
```diff
@@ -1816,13 +1816,13 @@ endfunction()
...
 # Usage:
-#   zephyr_blobs_verify(<MODULE module|FILES file [files...]> [REQUIRED])
+#   zephyr_blobs_verify(<MODULE module|FILES file [files...]> [REQUIRED] [REGEX])
 #
 # Example:
 # zephyr_blobs_verify(MODULE my_module REQUIRED) # verify all blobs in my_module and fail on error
 # zephyr_blobs_verify(FILES img/file.bin)        # verify a single file and print on error
 function(zephyr_blobs_verify)
-  cmake_parse_arguments(BLOBS_VERIFY "REQUIRED" "MODULE" "FILES" ${ARGN})
+  cmake_parse_arguments(BLOBS_VERIFY "REQUIRED" "MODULE;REGEX" "FILES" ${ARGN})
```
tejlmand, line 1826: main issue I see with this is that this `REGEX` is a python regex, not a CMake `REGEX`.

Having Python REGEX in a CMake function is something which I can't approve cause this with eventual result in developers not being able to know which regex dialect they should write.

tejlmand (changes requested):
REGEX'es as function arguments in CMake should be the CMake regex dialect, not the python dialect.

PR #99417, opened 2025-11-14
blinky sample on Pimoroni Pico Plus 2W 

casaroli:
This adds the W variant of this board, which has the LED connected to the wifi controller.

- Add the GPIO control to the wifi driver
- Add the Pico Plus 2W board configuration variant
- Adjust the configuration of the LED in the DTS
- Adjust the blinky sample to enable wifi and network for the W variant

This board LED is connected to the wifi controller. For this reason we remove this from the `gpio` test suite and disable the two relevant tests that require GPIO (midi, led).

If this PR gets merged, I plan on making another PR to enable blinky on the rpi_pico2 W board as well as the mcuboot configuration for the pico_plus2


samples/subsys/usb/midi/sample.yaml
```diff
@@ -9,3 +9,5 @@ tests:
       type: one_line
       regex:
         - "USB device support enabled"
+    platform_exclude:
+      - pico_plus2/rp2350b/m33/w
```
jfischer-no: No. Instead remove `usbd` from [boards/pimoroni/pico_plus2/pico_plus2_rp2350b_m33_w.yaml](https://github.com/zephyrproject-rtos/zephyr/pull/99417/commits/166fedb93eab28e026aef802a6b4d66a8d472c94#diff-94eed868c8b2a1b3dc91d872457c20acd94fe0edfd2ecad99ea3e508742c5437), https://github.com/zephyrproject-rtos/zephyr/pull/99417/commits/166fedb93eab28e026aef802a6b4d66a8d472c94.

jfischer-no (changes requested):

samples/basic/blinky/boards/pico_plus2_rp2350b_m33_w.conf
```diff
@@ -0,0 +1,5 @@
+# the LED is connected to the wifi controller
+CONFIG_NETWORKING=y
+CONFIG_WIFI=y
+CONFIG_MAIN_STACK_SIZE=2048
+CONFIG_TEST_RANDOM_GENERATOR=y
```
jfischer-no, line 5: I do not think it makes sense to enable NETWORKING and WIFI for a sample like `basic/blinky`. This sample should be as simple as possible and work out of the box.

drivers/wifi/infineon/airoc_wifi.c
```diff
@@ -954,6 +953,100 @@ static const struct net_wifi_mgmt_offload airoc_api = {
 	.wifi_mgmt_api = &airoc_wifi_mgmt,
 };
 
+#if DT_NODE_EXISTS(DT_INST_CHILD(0, gpio))
```
soburi, line 956: The GPIO driver part should be in drivers/gpio.
You'll probably also need to create an MFD device.

soburi (changes requested):

samples/basic/blinky/boards/pico_plus2_rp2350b_m33_w.conf
casaroli: I agree, but in this case the only LED on the board is connected to the wifi module, which requires the loading and initialization of two firmwares (on every boot). How do you suggest we proceed?

drivers/wifi/infineon/airoc_wifi.c
casaroli: Thank you for the comment. I just learned about MFD devices. I will investigate them.

samples/basic/blinky/boards/pico_plus2_rp2350b_m33_w.conf
jfischer-no: Then, there are no LEDs supported by default, and this board cannot be used with this sample. And that is fine.

Issue #99358, opened 2025-11-14
st_lcd_dsi_mb1835 shield on stm32u5g9j_dk1 board overrides heap memory pool


phoddie:
The configuration for the `st_lcd_dsi_mb1835` shield includes the following fragment for the `stm32u5g9j_dk1` board:

https://github.com/zephyrproject-rtos/zephyr/blob/0b93c2f8eec5dd4c48f57765ed7ef0d6fa0cade6/boards/shields/st_lcd_dsi_mb1835/boards/stm32u5g9j_dk1.conf#L1-L2

This forces the memory heap, specified by `CONFIG_HEAP_MEM_POOL_SIZE`,  to 64 KB, overriding any value requested by the application. My understanding (based on the [documentation](https://docs.zephyrproject.org/latest/kernel/memory_management/heap.html#defining-the-heap-memory-pool)) is that if a board needs to increase the memory partition, it should do so using `HEAP_MEM_POOL_ADD_SIZE_*`. However, it looks like the memory needs of the driver in the heap are modest (apart from frame buffers, which are outside the this heap) so perhaps no memory configuration is necessary.

CC: @CharlesDias

JarmouniA:
mempool size shouldn't be specified at the shield level, it's app specific, please send a PR if you can.

PR #96177, opened 2025-09-17
shields: stm32: avoid LTDC format and HEAP size settings in shields conf

avolmat-st:
The LTDC pixel format and HEAP size is a matter of the application rather than the configuration of the shield. Remove such config from the stm32 related (LTDC) display shields and as an example, move them into the display sample app.

PR #99418, opened 2025-11-14
CONFIG_HEAP_MEM_POOL_SIZE is for apps, not shields

phoddie:
Please see #99358.


erwango (dismissed):
I don't disagree on the aim of this PR, but it would need to explain why it is safe to remove this config if not other changes are made to ensure use cases remain functional.

phoddie:
> but it would need to explain why it is safe to remove this config if not other changes are made to ensure use cases remain functional

Unfortunately, without access to all of the relevant hardware, I cannot prove that. I have the `st_lcd_dsi_mb1835`  and it runs fine with a smaller heap. Reviewing the source code of the driver, it appears to only do a single dynamic allocation, and that is of a full frame buffer which would already fail with the configured heap size.

Looking from the other perspective, these shields are unexpectedly overriding a setting that is for applications (as per @JarmouniA). If they  need more RAM, there is a separate way for them to do that. It was a significant time-sink for me to track down why my configured value for `CONFIG_HEAP_MEM_POOL_SIZE` was being ignored. So, there is real harm from these mistakes in the shields. If an application depends on the values provided by the shields, that is an application bug.

JarmouniA (approved):

erwango:
> There is no proof it is breaking any in-tree samples that were tested with these shields and known to be functional prior to this change. Also, there is no Zephyr requirement for every sample to work with every board/shield.

I assume these settings were there for a reason and not set randomly, so removing them will have most likely an impact.

> Also, there is no Zephyr requirement for every sample to work with every board/shield.

And yet we're spending 20% of our time fixing issues reported on samples.

phoddie:
@CharlesDias – Thank you! It looks like PR #96177 does overlap the PR for `st_b_lcd40_dsi1_mb1166`. I'm happy to remove that from here. This PR makes the same change for `st_Icd_dsi_mb1835` and `rk055hdmipi4ma0` which are not covered by that PR, as far as I can see. @erwango – do you still object to the changes on the grounds that they might break an unidentified sample?

boards/shields/rk055hdmipi4ma0/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf
```diff
@@ -1,5 +1,4 @@
 CONFIG_MCUX_DCNANO_LCDIF_EXTERNAL_FB_MEM=y
 CONFIG_MCUX_DCNANO_LCDIF_EXTERNAL_FB_ADDR=0x60000000
 CONFIG_MCUX_DCNANO_LCDIF_MAINTAIN_CACHE=n
-CONFIG_HEAP_MEM_POOL_SIZE=40000
```
KATE-WANG-NXP, line 4: Please rebase and update, this files content has already been updated, there is no such need now.
https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/shields/rk055hdmipi4ma0/boards/mimxrt700_evk_mimxrt798s_cm33_cpu0.conf#L1

KATE-WANG-NXP (changes requested):

PR #97295, opened 2025-10-09
dts: ra8: fix syntax error in r7fa8XXXX.dtsi files

gurland:
The "status" property was defined after the pllX subnodes in pll node, which violates the Devicetree Specification v0.4, section 6.3:

"Nodes may contain property definitions and/or child node definitions. If both are present, properties shall come before child nodes."

This caused the Device Tree Compiler error: "Properties must precede subnodes. Unable to parse input tree"

This commit moves the "status" property to precede all pll subnodes, fixing the syntax error and ensuring compliance with the Devicetree Specification.

PR #99423, opened 2025-11-14
dts: renesas: fix syntax error in multiple DTS/DTSI files

gurland:
Properties in `bclk` nodes were defined after the bclkout subnode, `disk` subnode was defined before properties in a few boards DTS files which violates the Devicetree Specification v0.4, section 6.3:

"Nodes may contain property definitions and/or child node definitions. If both are present, properties shall come before child nodes."

This caused the Device Tree Compiler error: "Properties must precede subnodes. Unable to parse input tree"

This commit is simmilar to https://github.com/zephyrproject-rtos/zephyr/pull/97295 and moves nested nodes after properties, fixing those syntax errors and ensuring compliance with the Devicetree Specification.


khoa-nguyen-18 (approved):

thenguyenyf (approved):

quytranpzz (approved):

khoa-nguyen-18:
Whenever you have time, it would be appreciated if @KhiemNguyenT and @soburi could take a look

soburi (approved):

PR #99428, opened 2025-11-15
drivers: ethernet: xmc4xxx: Fixes the checksum of bridged packets

mrjimenez:
Due to checksum offloading, the checksum field of a packet must be zero before sending it to packet to the ethernet controller, otherwise the computed checksum will be incorrect.

If the packet is bridged, it already has the correct checksum in place, which will cause the automatically computed checksum to incorrectly become zero.

In this patch we give bridged packets the special attention they need.

### DIFF
--- a/drivers/ethernet/eth_xmc4xxx.c
+++ b/drivers/ethernet/eth_xmc4xxx.c
@@ -228,6 +228,138 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
 	regs->RECEIVE_POLL_DEMAND = 0U;
 }
 
+static inline void pkt_hexdump(struct net_pkt *pkt, const char *str)
+{
+	struct net_buf *buf = pkt->buffer;
+
+	while (buf) {
+		LOG_HEXDUMP_DBG(buf->data, buf->len, str);
+		buf = buf->frags;
+	}
+}
+
+/* Clears the checksum field of an ICMPv4 packet. */
+static int eth_xmc4xxx_clear_checksum_icmpv4(struct net_pkt *pkt)
+{
+	NET_PKT_DATA_ACCESS_DEFINE(icmpv4_access, struct net_icmp_hdr);
+	struct net_icmp_hdr *icmp_hdr;
+
+	if (IS_ENABLED(CONFIG_NET_IPV4_HDR_OPTIONS)) {
+		if (net_pkt_skip(pkt, net_pkt_ipv4_opts_len(pkt))) {
+			return -ENOBUFS;
+		}
+	}
+
+	icmp_hdr = (struct net_icmp_hdr *)net_pkt_get_data(pkt, &icmpv4_access);
+	if (!icmp_hdr) {
+		return -ENOBUFS;
+	}
+
+	icmp_hdr->chksum = 0;
+	net_pkt_set_chksum_done(pkt, true);
+
+	return net_pkt_set_data(pkt, &icmpv4_access);
+}
+
+/* Clears the checksum value in packets that will go through checksum
+ * offloading.
+ */
+static int eth_xmc4xxx_clear_checksum(struct net_pkt *pkt)
+{
+	int ret = 0;
+	uint16_t p_type = 0xFFFF;
+	uint8_t next_header = 0xFF;
+	struct net_eth_hdr *eth_hdr;
+	struct net_pkt_cursor backup;
+	bool overwrite;
+	const char *msg = NULL;
+	const char *msg_does_not_apply = "<does not apply>";
+	const char *p_type_str = msg_does_not_apply;
+	const char *next_header_str = msg_does_not_apply;
+	const char *msg_error_skipping_pkt = "Error skipping pkt";
+
+	overwrite = net_pkt_is_being_overwritten(pkt);
+	net_pkt_cursor_backup(pkt, &backup);
+	net_pkt_cursor_init(pkt);
+	net_pkt_set_overwrite(pkt, true);
+	eth_hdr = NET_ETH_HDR(pkt);
+	ret = net_pkt_skip(pkt, sizeof(struct net_eth_hdr));
+	if (ret) {
+		msg = msg_error_skipping_pkt;
+		goto end;
+	}
+	p_type = ntohs(eth_hdr->type);
+	switch (p_type) {
+	case NET_ETH_PTYPE_IP: {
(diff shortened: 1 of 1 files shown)


teburd (dismissed):

drivers/ethernet/eth_xmc4xxx.c
```diff
@@ -228,6 +228,124 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
...
+		break;
+	}
+
+end:
+	net_pkt_cursor_restore(pkt, &backup);
+	net_pkt_set_overwrite(pkt, overwrite);
+	if (msg) {
+		LOG_DBG("p_type = 0x%04x (%s), next_header = 0x%02x (%s)\n", p_type, p_type_str,
+			next_header, next_header_str);
+		pkt_hexdump(pkt, msg);
```
kartben, line 358: I don't think you want to unconditionally call `pkt_hexdump`?

drivers/ethernet/eth_xmc4xxx.c
```diff
@@ -228,6 +228,124 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
...
+}
+
+/* Clears the checksum value in packets that will go through checksum
+ * offloading.
+ */
+static int eth_xmc4xxx_clear_checksum(struct net_pkt *pkt)
+{
+	int ret = 0;
+	uint16_t p_type;
+	uint8_t next_header = -1;
```
kartben: assigning -1 to an unsigned int is a bit... odd? Maybe you want `0xFF` or something ...

drivers/ethernet/eth_xmc4xxx.c
```diff
@@ -228,6 +228,124 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
...
+	const char *msg = NULL;
+	const char *p_type_str;
+	const char *next_header_str = "<does not apply>";
+
+	overwrite = net_pkt_is_being_overwritten(pkt);
+	net_pkt_cursor_backup(pkt, &backup);
+	net_pkt_cursor_init(pkt);
+	net_pkt_set_overwrite(pkt, true);
+	eth_hdr = NET_ETH_HDR(pkt);
+	net_pkt_skip(pkt, sizeof(struct net_eth_hdr));
```
kartben: return value shouldn't be ignored - this might error out, and is one of our coding guidelines too. Applies everywhere

https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#:~:text=The%20value%20returned%20by%20a%20function%20having%20non%2Dvoid%20return%20type%20shall%20be%20used

drivers/ethernet/eth_xmc4xxx.c
```diff
@@ -228,6 +228,124 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
...
+		next_header_str = "Unknown";
+		msg = "Unknown next header";
+		break;
+	}
+
+end:
+	net_pkt_cursor_restore(pkt, &backup);
+	net_pkt_set_overwrite(pkt, overwrite);
+	if (msg) {
+		LOG_DBG("p_type = 0x%04x (%s), next_header = 0x%02x (%s)\n", p_type, p_type_str,
```
kartben:
```suggestion
		LOG_DBG("p_type = 0x%04x (%s), next_header = 0x%02x (%s)", p_type, p_type_str,
```

drivers/ethernet/eth_xmc4xxx.c
```diff
@@ -228,6 +228,124 @@ static inline void eth_xmc4xxx_trigger_dma_rx(ETH_GLOBAL_TypeDef *regs)
...
+
+/* Clears the checksum value in packets that will go through checksum
+ * offloading.
+ */
+static int eth_xmc4xxx_clear_checksum(struct net_pkt *pkt)
+{
+	int ret = 0;
+	uint16_t p_type;
+	uint8_t next_header = -1;
+	struct net_eth_hdr *eth_hdr = NET_ETH_HDR(pkt);
```
kartben:
```suggestion
	struct net_eth_hdr *eth_hdr;
```

kartben (dismissed):

drivers/ethernet/eth_xmc4xxx.c
teburd: I had the same thought, but seeing as its inlined and the only thing in the loop in pkt_hexdump was a LOG_HEXDUMP_DBG my suspicion is it goes away without debug level logging as a NOP.

mrjimenez: > I had the same thought, but seeing as its inlined and the only thing in the loop in pkt_hexdump was a LOG_HEXDUMP_DBG my suspicion is it goes away without debug level logging as a NOP.

Yes, exactly. But now that you mention it, the function is not inlined, it is just static, that would be a nice addition.

Anyway, I agree it looks strange at first glance. Things that could be done:

- Write a macro with a suffix like `_DBG`. I am not very fond of macros, unless they are strictly needed, I don't like that one.
- Put the loop of `pkt_hexdump()` directly in the function and not call an inlined function. On the other hand,  it is more reusable if it is a function.
- Put a `_dbg` suffix in the function name.

Suggestions?

drivers/ethernet/eth_xmc4xxx.c
mrjimenez: Not totally unheard of... There are some (few) uses in macros, struct initializations and assignments in the code base.

I don't mind using `0xFF`, but Just out of curiosity, would `uint8_t next_header = (uint8_t) -1;` be acceptable?

drivers/ethernet/eth_xmc4xxx.c
mrjimenez: Sure. Sorry, my bad. Will fix it.

drivers/ethernet/eth_xmc4xxx.c
mrjimenez: Ok, will add to the next commit.

drivers/ethernet/eth_xmc4xxx.c
mrjimenez: Ok, will add to the next commit.

mrjimenez:
Should I mark those two conversations as resolved? Sorry, I am new here, I don't know exactly what the dynamics is.

teburd (approved):

jukkar (approved):

PR #99429, opened 2025-11-15
drivers: ethernet: stm32hal: Fixes source address control

mrjimenez:
The default configuration for all STM32 microcontrollers in HAL is
to replace the ehternet MAC address with the address configured in
replace address zero. In "modules/hal", Grep for:

macDefaultConf.SourceAddrControl = ETH_SOURCEADDRESS_REPLACE_ADDR0;

But this is a bad thing for bridging. Since Zephyr ethernet LL code
always writes the source address, this patch has no impact on normal
operation. And has the benefit of maintaining the source MAC address of
bridged packets.

### DIFF
--- a/drivers/ethernet/eth_stm32_hal_v2.c
+++ b/drivers/ethernet/eth_stm32_hal_v2.c
@@ -679,6 +679,14 @@ void eth_stm32_set_mac_config(const struct device *dev, struct phy_link_state *s
 #if DT_HAS_COMPAT_STATUS_OKAY(st_stm32n6_ethernet)
 	mac_config.PortSelect = PHY_LINK_IS_SPEED_1000M(state->speed) ? DISABLE : ENABLE;
 #endif
+
+	/* Always disable hardware source address replacement.
+	 * Zephyr network stack sets the source MAC address and
+	 * therefore hardware replacement should not be enabled,
+	 * since it may affect bridging applications, for example.
+	 */
+	mac_config.SourceAddrControl = ETH_SOURCEADDRESS_DISABLE;
+
 	hal_ret = HAL_ETH_SetMACConfig(heth, &mac_config);
 	if (hal_ret != HAL_OK) {
 		LOG_ERR("HAL_ETH_SetMACConfig failed: %d", hal_ret);


drivers/ethernet/eth_stm32_hal_v1.c
```diff
@@ -239,6 +239,11 @@ void eth_stm32_set_mac_config(const struct device *dev, struct phy_link_state *s
 
 	heth->Init.Speed = PHY_LINK_IS_SPEED_100M(state->speed) ? ETH_SPEED_100M : ETH_SPEED_10M;
 
+	/* The default configuration is ETH_SOURCEADDRESS_REPLACE_ADDR0,
+	 * which is bad for bridging.
+	 */
+	mac_config.SourceAddrControl = ETH_SOURCEADDRESS_DISABLE;
```
juickar: mac_config is not defined in the v1 driver and `ETH_SOURCEADDRESS_DISABLE` is not defined for the series that uses it (F1X and F2X)

juickar (changes requested):

drivers/ethernet/eth_stm32_hal_v1.c
mrjimenez: Ok, I have removed these lines from v1.

drivers/ethernet/eth_stm32_hal_v2.c
```diff
@@ -679,6 +679,12 @@ void eth_stm32_set_mac_config(const struct device *dev, struct phy_link_state *s
 #if DT_HAS_COMPAT_STATUS_OKAY(st_stm32n6_ethernet)
 	mac_config.PortSelect = PHY_LINK_IS_SPEED_1000M(state->speed) ? DISABLE : ENABLE;
 #endif
+
+	/* The default configuration is ETH_SOURCEADDRESS_REPLACE_ADDR0,
+	 * which is bad for bridging.
+	 */
+	mac_config.SourceAddrControl = ETH_SOURCEADDRESS_DISABLE;
```
juickar, line 688: Maybe we should put this under `#if defined(CONFIG_NET_ETHERNET_BRIDGE)` to make sure it will have no impact on non-bridging scenarios

mrjimenez: I disagree. We don't need CONFIG_NET_ETHERNET_BRIDGE to perform bridging. Which, by the way, is the case in my application.

Also, since the code that fills the source address is independent of the architecture, what this option does in practice is to overwrite a field that was already filled. In other words, this option performs extra work, with results that are not always what you want or would expect. It was not easy to spot that:

1. The source addres of some packets were being "magically" changed causing failure.
2. That different boards using exactly the same Zephyr software behave differently (i.e. actually worked).
3. That the cause was some magic optimization trick in that particular hardware.

IMHO, this option is bad, unless you are agressively optimizing code, which would imply using L2 code specific for this family of processors, which is not the case in Zephyr.

maass-hamburg: maybe we should only set it in `case ETHERNET_CONFIG_TYPE_PROMISC_MODE` of `eth_stm32_hal_set_config` then?

mrjimenez: It would still be just rewriting something that was already written before by the L2 layer. I can't foresee all the possible use cases, but still I think that this kind of optimization does not fit this type of generic network stack. It could make a difference in an optimized product with an assembly coded network stack.

It is a really bad feeling when someone tells you: "works with all the other boards, except for the one you want to use, we have no idea why, so accept it and move on."

maass-hamburg: If it is all this change does is, that the mac address in the ethernet packet will no longer be "magically" changed, it seems fine. I don't see the reason why the ethernet controller should change the mac address of outgoig packets by default. The mac, that the networking subsystem had put there shouldn't be changed.

juickar: Okay, fair enough let's disable it by default but could you change the comment on top to:

```
	/* Always disable hardware source address replacement.
	 * Zephyr network stack sets the source MAC address and
	 * therefore hardware replacement should not be enabled.
	 */
```
or something similar

mrjimenez: Ok, just did that.

maass-hamburg (approved):

juickar (approved):

erwango (approved):

PR #99430, opened 2025-11-15
boards: beagle: pocketbeagle_2: doc: Add DFU instructions

Ayush1325:
- Add instructions for DFU on a53 target
- Useful for rapid development.

### DIFF
--- a/boards/beagle/pocketbeagle_2/doc/index.rst
+++ b/boards/beagle/pocketbeagle_2/doc/index.rst
@@ -95,6 +95,9 @@ Flashing
 A53 Core
 ========
 
+Using SD Card
+-------------
+
 The testing requires the binary to be copied to the BOOT partition in SD card.
 
 To test the A53 core, we build the :zephyr:code-sample:`hello_world` sample with the following command.
@@ -117,6 +120,38 @@ We now copy this binary onto the SD card in the :file:`/boot/` directory and nam
 The SD card can now be used for booting.
 
 The binary will run and print Hello world to the debug port.
+
+DFU
+---
+
+While doing rapid development, the SD Card workflow can become cumbersome. Since AM62x platform
+supports DFU over USB, we can use it to boot Zephyr applications, without ever having an SD Card
+plugged in. For more in-depth instructions, check out the `blog post <https://www.beagleboard.org/blog/2025-11-14-pocketbeagle-2-zephyr-dfu>`_.
+
+1. Download the DFU u-boot binaries. Using `v2025.10-am62-pocketbeagle2-11.02.04 <https://github.com/beagleboard/u-boot-pocketbeagle2/releases/download/v2025.10-am62-pocketbeagle2-11.02.04/>`_ here.
+
+.. code-block:: console
+
+   wget https://github.com/beagleboard/u-boot-pocketbeagle2/releases/download/v2025.10-am62-pocketbeagle2-11.02.04/tiboot3-usbdfu.bin
+   wget https://github.com/beagleboard/u-boot-pocketbeagle2/releases/download/v2025.10-am62-pocketbeagle2-11.02.04/tispl.bin
+   wget https://github.com/beagleboard/u-boot-pocketbeagle2/releases/download/v2025.10-am62-pocketbeagle2-11.02.04/u-boot-zephyrdfu.img
+
+2. Install `dfu-util <https://dfu-util.sourceforge.net/>`_ CLI.
+
+3. Plug in PocketBeagle 2 and check if it is available.
+
+.. code-block:: console
+
+   dfu-util -l
+
+4. Flash. The application path is assumed to be :file:`build/zephyr/zephyr.bin` here.
+
+.. code-block:: console
+
+   dfu-util -R -a bootloader -D ./tiboot3-usbdfu.bin
+   dfu-util -R -a tispl.bin -D ./tispl.bin
+   dfu-util -R -a u-boot.img -D ./u-boot-zephyrdfu.img
+   dfu-util -R -a zephyr.bin -D ./build/zephyr/zephyr.bin
 
 M4F Core
 ========


boards/beagle/pocketbeagle_2/doc/index.rst
```diff
@@ -118,6 +121,38 @@ The SD card can now be used for booting.
 
 The binary will run and print Hello world to the debug port.
 
+DFU
+---
+
+While doing rapid development, the SD Card workflow can become cumbersome. Since AM62x platform
+supports DFU over USB, we can use it to boot zephyr applications, without ever having an SD Card
+plugged in. For more in-depth instructions, checkout the `blog post <https://www.beagleboard.org/blog/2025-11-14-pocketbeagle-2-zephyr-dfu>`_.
```
dsseng:
```suggestion
supports DFU over USB, we can use it to boot Zephyr applications, without ever having an SD Card
plugged in. For more in-depth instructions, check out the `blog post <https://www.beagleboard.org/blog/2025-11-14-pocketbeagle-2-zephyr-dfu>`_.
```

dsseng (approved):
🆒

boards/beagle/pocketbeagle_2/doc/index.rst
Ayush1325: Fixed

boards/beagle/pocketbeagle_2/doc/index.rst
```diff
@@ -118,6 +121,38 @@ The SD card can now be used for booting.
...
 The binary will run and print Hello world to the debug port.
 
+DFU
+---
+
+While doing rapid development, the SD Card workflow can become cumbersome. Since AM62x platform
+supports DFU over USB, we can use it to boot Zephyr applications, without ever having an SD Card
+plugged in. For more in-depth instructions, check out the `blog post <https://www.beagleboard.org/blog/2025-11-14-pocketbeagle-2-zephyr-dfu>`_.
+
+1. Download the DFU u-boot binaries. Using `v2025.10-am62-pocketbeagle2-11.02.04 <https://github.com/beagleboard/u-boot-pocketbeagle2/releases/download/v2025.10-am62-pocketbeagle2-11.02.04/>`_ here.
```
jadonk, line 131: Github is good, but I think we are not doing the right thing by relying on Github for long-term hosting of binary artifacts. Can you add what is required to https://www.beagleboard.org/distros? It is probably best to zip the 3 files into one download.

boards/beagle/pocketbeagle_2/doc/index.rst
```diff
@@ -118,6 +121,38 @@ The SD card can now be used for booting.
...
+
+   dfu-util -l
+
+4. Flash. The application path is assumed to be :file:`build/zephyr/zephyr.bin` here.
+
+.. code-block:: console
+
+   dfu-util -R -a bootloader -D ./tiboot3-usbdfu.bin
+   dfu-util -R -a tispl.bin -D ./tispl.bin
+   dfu-util -R -a u-boot.img -D ./u-boot-zephyrdfu.img
```
jadonk, line 153: I think `tiboot3-usbdfu.bin` has all the necessary TIFS code. I think `tispl.bin` includes the DDR configuration. Why is `u-boot-zephyrdfu.img` needed?

jadonk (commented):
I'm not particularly comfortable depending on Github for long-term binary releases.

boards/beagle/pocketbeagle_2/doc/index.rst
Ayush1325: u-boot-zephyrdfu.img is basically just our normal u-boot + bootcommand: https://github.com/beagleboard/u-boot-pocketbeagle2/blob/66a114eee5021b90d375d63ce258ec43131636e6/build_u-boot.sh#L191

Normally, one would be required to type some commands in u-boot shell to put u-boot in ram dfu mode. However, by using a custom bootcommand to do that automatically allows bypassing need for any manual input.

Ayush1325:
> I'm not particularly comfortable depending on Github for long-term binary releases.

Well, since we do not have any way to automatically push to distro page, I am reluctant to use that for things I would like to update regularly.

I am fine with using GitLab releases, but from what I remember, Robert started mirroring u-boot builders to GitHub to offload the traffic from our GitLab servers. While I can't comment on speed in US, at least in India, downloading from GitHub releases is significantly faster than our self-hosted GitLab (it's noticeable even on small files due to latency).

boards/beagle/pocketbeagle_2/doc/index.rst
```diff
@@ -118,6 +121,38 @@ The SD card can now be used for booting.
...
+
+.. code-block:: console
+
+   dfu-util -l
+
+4. Flash. The application path is assumed to be :file:`build/zephyr/zephyr.bin` here.
+
+.. code-block:: console
+
+   dfu-util -R -a bootloader -D ./tiboot3-usbdfu.bin
```
glneo, line 151: So with this flow we have to DFU all the U-Boot parts every time, which takes time and extra steps. Plus it uses custom/pre-built U-Boot builds which we might want to avoid in the future.

Here is what I do for rapid DFU/Zephyr development. Simply use the SD card instruction as before, but instead of copying the `zephyr.bin` image onto the card each time, update the uEnv.txt file to the following:
```
dfu_alt_info=zephyr.bin ram 0x80200000 0x10000000
uenvcmd=dfu 0 ram 0; dcache flush; icache flush; dcache off; icache off; go 0x82000000
```

Now all one needs to do is boot at before (no need to change into DFU more or anything) and on the host:
```
dfu-util -R -a zephyr.bin -D ./build/zephyr/zephyr.bin
```

Ayush1325: That sounds nice. I will create a PR to document that here.

However, the setup I am describing here is specifically for use without a storage media. It originated in the [beagle forums](https://forum.beagleboard.org/t/pocketbeagle-2-boot-with-snagboot/41236).

glneo: BTW, you can also skip the A53 SPL/U-Boot stages also. Instead of packaging the A53 SPL in `tispl.bin` you replace that with your `zephyr.bin`. The flow ends up:
```
R5 SPL (`tiboot3.bin`) -> TF-A -> OP-TEE -> Zephyr (`tispl.bin`)
```
This takes some additional work in crafting the `tispl.bin` (still needs binman from U-Boot project, maybe some of that can be pulled into Zephyr someday), so out-of-scope for now, but it does work.

boards/beagle/pocketbeagle_2/doc/index.rst
glneo: If you are trying to skip manually typing commands into the U-Boot shell, why not just skip the U-Boot stage? Instead of modifying `u-boot.img` you could modify `tispl.bin` to accept `zephyr.bin` as its next stage instead of needing to go though the `u-boot.img` stage.

glneo (approved):

boards/beagle/pocketbeagle_2/doc/index.rst
jadonk: @Ayush1325 can you provide at least an alternate download via https://www.beagleboard.org/distros?

boards/beagle/pocketbeagle_2/doc/index.rst
jadonk: @Ayush1325 can you just skip u-boot?

jadonk (changes requested):

PR #99440, opened 2025-11-15
treewide: adjust CMakeLists.txt files indentation

josuah:
The rules are interpolated from what there is on the docs:

https://docs.zephyrproject.org/latest/contribute/style/cmake.html#commands-and-syntax

```cmake
# Good:
target_sources(my_target PRIVATE
  src/file1.cpp
  src/file2.cpp
)
```

And what seemed to be common on in-tree, in particular for `if()` and `messages()`:

- Long lines have continuation lines use 2 spaces
  ```
  if(long line ...
    continuation line...
  ```
  ```
  function(arg1 arg2...
    continuation line...
  ```
- When the first line is filled with content, the closing parenthesis is
  at the end of the last continuation line
  ```
  if(long line filled until the end ...
    continuation line...)
  ```
  ```
  message(long line filled until the end...
    continuation line...)
  ```
- When the first line is empty or only containing a distinct label, such
  as an UPPERCASE label) the closing parenthesis is on its own line,
  with same indentation level as the first line.
  ```
  zephyr_library_sources(
    something_a.c
    something_b.c
  )
  ```
  ```
  zephyr_library_sources_ifdef(CONFIG_SOMETHING
    something_a.c
    something_b.c
  )
  ```
- Call to message() are converted from \ for escaping newlines to a list
  of "... " strings, with an empty space at the end if separation between
  words is needed.
  ```
  message("long line... Hello "
    "cmake! continuation line...")
  ```

If anything seems wrong choice, let me know of what to apply and I'd be glad to apply it again tree-wide.

If the scope of modifications is wrong (i.e. should do all `CMakeLists.txt` and `.cmake`), then let me know and I would change the scope for this PR.

I am keeping it as a draft for now to avoid triggering the bot.


nashif (dismissed):

jhedberg (dismissed):

Thalley (dismissed):

cmake/modules/FindGnuLd.cmake
```diff
@@ -43,9 +43,11 @@ if(EXISTS "${GNULD_LINKER}")
 endif()
 
 # See if the compiler has a preferred linker
-execute_process(COMMAND ${CMAKE_C_COMPILER} --print-prog-name=ld.bfd
-                OUTPUT_VARIABLE GNULD_LINKER
-                OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND
+  ${CMAKE_C_COMPILER} --print-prog-name=ld.bfd
```
tejlmand: Why is this line wrapped ?
```
COMMAND ${CMAKE_C_COMPILER} --print-prog-name=ld.bfd
```

on a single line makes perfect sense, because `COMMAND` is the argument name to the function, and the following are values to the `COMMAND` argument.

cmake/modules/FindLlvmLld.cmake
```diff
@@ -20,9 +20,12 @@
 include(FindPackageHandleStandardArgs)
 
 # See if the compiler has a preferred linker
-execute_process(COMMAND ${CMAKE_C_COMPILER} --print-prog-name=ld.lld
-                OUTPUT_VARIABLE LLVMLLD_LINKER
-                OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(
+  COMMAND
+  ${CMAKE_C_COMPILER} --print-prog-name=ld.lld
```
tejlmand: same regarding `COMMAND` arguments and values following.

nordicjm (dismissed):
Big improvement, thanks

cmake/linker/ld/target_relocation.cmake
```diff
@@ -13,30 +13,31 @@ macro(toolchain_ld_relocation)
...
-    $<$<BOOL:${CMAKE_VERBOSE_MAKEFILE}>:--verbose>
-    -d ${APPLICATION_BINARY_DIR}
-    -i ${DICT_FILE}
-    -o ${MEM_RELOCATION_LD}
-    -s ${MEM_RELOCATION_SRAM_DATA_LD}
-    -b ${MEM_RELOCATION_SRAM_BSS_LD}
-    -c ${MEM_RELOCATION_CODE}
-    --default_ram_region ${MEM_REGION_DEFAULT_RAM}
+      ${PYTHON_EXECUTABLE}
+      ${ZEPHYR_BASE}/scripts/build/gen_relocate_app.py
```
tejlmand, line 28: this is not really consistent with what we do here https://github.com/zephyrproject-rtos/zephyr/blob/5864af4fc89ce5894467238d645f2bb10bcd5ed7/cmake/linker/iar/target.cmake#L47-L61

where all lines with arguments are wrapped and values indented one more level, when on argument and its values goes beyond the line length.

nashif (dismissed):

cmake/modules/FindZephyr-sdk.cmake
```diff
@@ -55,14 +55,14 @@ endif()
 # 2) No toolchain specified == Default to Zephyr toolchain
 # Until we completely deprecate it
 if(("zephyr" STREQUAL ${ZEPHYR_TOOLCHAIN_VARIANT}) OR
-   (NOT DEFINED ZEPHYR_TOOLCHAIN_VARIANT) OR
-   (DEFINED ZEPHYR_SDK_INSTALL_DIR) OR
-   (Zephyr-sdk_FIND_REQUIRED))
+  (NOT DEFINED ZEPHYR_TOOLCHAIN_VARIANT) OR
+  (DEFINED ZEPHYR_SDK_INSTALL_DIR) OR
+  (Zephyr-sdk_FIND_REQUIRED))
```
tejlmand:
```suggestion
  (Zephyr-sdk_FIND_REQUIRED)
)
```

drivers/adc/CMakeLists.txt
```diff
@@ -4,76 +4,79 @@ zephyr_syscall_header(${ZEPHYR_BASE}/include/zephyr/drivers/adc.h)
...
+zephyr_library_sources_ifdef(CONFIG_USERSPACE adc_handlers.c)
+
+# zephyr-keep-sorted-start
+zephyr_library_sources_ifdef(CONFIG_ADC_AD405X adc_ad405x.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_AD4114 adc_ad4114.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_AD4130 adc_ad4130.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_AD4170 adc_ad4170.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_AD559X adc_ad559x.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_AD7124 adc_ad7124.c)
+zephyr_library_sources_ifdef(CONFIG_ADC_ADS1112 adc_ads1112.c)
```
tejlmand, line 18: I know the current 2nd column alignment was broken due to varying length on different lines.

However, I personally find this hard to read, and aligning files on second column does help readability, but the risk is of course that all lines are changed if a new and longer `CONFIG_x` var is added.

Should this be discussed further or is there consensus regarding this ?

Note, only placing this comment here, though we have more occurrences.

samples/net/sockets/echo_server/CMakeLists.txt
```diff
@@ -6,8 +6,8 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(sockets_echo_server)
 
 if(CONFIG_NET_SOCKETS_SOCKOPT_TLS AND
-   CONFIG_MBEDTLS_KEY_EXCHANGE_PSK_ENABLED AND
-   (CONFIG_NET_SAMPLE_PSK_HEADER_FILE STREQUAL "dummy_psk.h"))
+  CONFIG_MBEDTLS_KEY_EXCHANGE_PSK_ENABLED AND
+  (CONFIG_NET_SAMPLE_PSK_HEADER_FILE STREQUAL "dummy_psk.h"))
```
tejlmand:
```suggestion
  (CONFIG_NET_SAMPLE_PSK_HEADER_FILE STREQUAL "dummy_psk.h")
)
```

samples/net/sockets/http_server/CMakeLists.txt
```diff
@@ -50,16 +50,18 @@ target_link_libraries(app PRIVATE zephyr_interface zephyr)
 
 zephyr_linker_sources(SECTIONS sections-rom.ld)
 zephyr_linker_section_ifdef(CONFIG_NET_SAMPLE_HTTPS_SERVICE NAME
-				http_resource_desc_test_https_service
-				KVMA RAM_REGION GROUP RODATA_REGION)
+  http_resource_desc_test_https_service
```
tejlmand: when fixing this, then `NAME` argument should go with its value.
```
zephyr_linker_section_ifdef(CONFIG_NET_SAMPLE_HTTPS_SERVICE
  NAME http_resource_desc_test_https_service
  KVMA RAM_REGION GROUP RODATA_REGION
)
```

samples/net/sockets/http_server/CMakeLists.txt
```diff
@@ -50,16 +50,18 @@ target_link_libraries(app PRIVATE zephyr_interface zephyr)
...
+  KVMA RAM_REGION GROUP RODATA_REGION
+)
 zephyr_linker_section_ifdef(CONFIG_NET_SAMPLE_HTTP_SERVICE NAME
-				http_resource_desc_test_http_service
-				KVMA RAM_REGION GROUP RODATA_REGION)
+  http_resource_desc_test_http_service
```
tejlmand: same comment regarding `NAME <value>`.

tests/bluetooth/host/conn/mocks/CMakeLists.txt
```diff
@@ -3,32 +3,32 @@
...
 target_link_options(mocks PUBLIC
-	"SHELL:-T ${ZEPHYR_BASE}/tests/bluetooth/host/conn/mocks/mock-sections.ld"
+  "SHELL:-T ${ZEPHYR_BASE}/tests/bluetooth/host/conn/mocks/mock-sections.ld"
 )
 
 target_compile_options(mocks
-	PRIVATE
-	-DCONFIG_BT_ID_MAX=1
+  PRIVATE
+  -DCONFIG_BT_ID_MAX=1
```
tejlmand: `PRIVATE` goes with its value, and may even go to one line with `mocks` as it is short enough.
```
target_compile_options(mocks PRIVATE -DCONFIG_BT_ID_MAX=1)
```

tests/bluetooth/mesh/brg/CMakeLists.txt
```diff
@@ -6,18 +6,21 @@ project(bluetooth_mesh_brg)
 
 FILE(GLOB app_sources src/*.c)
 target_sources(app
-	PRIVATE
-	${app_sources}
-	${ZEPHYR_BASE}/subsys/bluetooth/mesh/brg_cfg.c)
+  PRIVATE
+  ${app_sources}
+  ${ZEPHYR_BASE}/subsys/bluetooth/mesh/brg_cfg.c
```
tejlmand: to be consistent, then the sources are value to `PRIVATE` so should be indented if using the principle from earlier and breaks the 100 char limit.
```
  PRIVATE
    ${app_sources}
    ${ZEPHYR_BASE}/subsys/bluetooth/mesh/brg_cfg.c
```

tests/bluetooth/mesh/brg/CMakeLists.txt
```diff
@@ -6,18 +6,21 @@ project(bluetooth_mesh_brg)
...
+)
 
 target_include_directories(app
-	PRIVATE
-	${ZEPHYR_BASE}/subsys/bluetooth/mesh
-	${ZEPHYR_MBEDTLS_MODULE_DIR}/include)
+  PRIVATE
+  ${ZEPHYR_BASE}/subsys/bluetooth/mesh
+  ${ZEPHYR_MBEDTLS_MODULE_DIR}/include
```
tejlmand: same `<ARG> <value>` indentation needed.

tests/bluetooth/mesh/brg/CMakeLists.txt
```diff
@@ -6,18 +6,21 @@ project(bluetooth_mesh_brg)
...
-	PRIVATE
-	-DCONFIG_BT_SETTINGS
-	-DCONFIG_BT_MESH_BRG_CFG_SRV
-	-DCONFIG_BT_MESH_BRG_TABLE_ITEMS_MAX=16
-	-DCONFIG_PSA_CRYPTO_PROVIDER_MBEDTLS)
+  PRIVATE
+  -DCONFIG_BT_SETTINGS
+  -DCONFIG_BT_MESH_BRG_CFG_SRV
+  -DCONFIG_BT_MESH_BRG_TABLE_ITEMS_MAX=16
+  -DCONFIG_PSA_CRYPTO_PROVIDER_MBEDTLS
```
tejlmand: same `<ARG> <value>` indentation needed.

tejlmand (changes requested):
Great work, thanks.

We need to decide on how we actually want indentation when facing long `<ARG> <val1> <val2> <val3> ...` constructs.

Generally in this PR it appears we want:
```
function(
  <ARG>
    <val1>
    <val2>
    <val3>
  <ANOTHER_ARG>
    <val_n>
```

however, it's not done consistently, namely there is still a lot of construct of the form:
```
target_sources(app PRIVATE
  src/main.c
  src/test_past.c
)
```
where the proper form would be:
```
target_sources(app
  PRIVATE
    src/main.c
    src/test_past.c
)
```

I would be fine allowing the first version, just to get this PR merged without constant rebases, as it improve the current state a lot, but I think the discusion on form is important, especially so we can ensure consistency.

Also so onther oberservations in PR.

But great work, and surprisingly few wrongs considered the size of the PR.

tejlmand (dismissed):
approved.

Observations to be followed up.

Thalley (dismissed):

drivers/adc/CMakeLists.txt
josuah: The rationale is adding new files requires modifying all the other lines for this to work.
Alternative would be picking an alignment column large enough for most and let longer file names overflow.
Any preference?
I'm 100% ok with reverting this part altogether, or applying whichever alignment level there was.

cmake/modules/FindGnuLd.cmake
josuah: I misunderstood CMake indentation where the a particular importance was given to the first arguments.
Adjusting it to preserve grouping of `UPPERCASE var_associated`

Should there be alignemnt to the parenthesis (like C code?) or alignment should always be 2 spaces?

josuah: I tried to follow the example and give the first keyword more importance, updating this to the other style.

samples/net/sockets/http_server/CMakeLists.txt
josuah: I wasn't sure about that one. thanks

cmake/linker/ld/target_relocation.cmake
josuah: It helps identifying the expected style, applying this through, thanks.
Will  try to apply this.

cmake/modules/FindZephyr-sdk.cmake
josuah: Same for other `if()` that span over multiple lines?

nashif (approved):

jhedberg (approved):

jhedberg (approved):

PR #99444, opened 2025-11-15
bindings: pwm-leds: Include base.yaml

ndreys:
Include base.yaml in order to support zephyr,deferred-init property.

### DIFF
--- a/dts/bindings/led/pwm-leds.yaml
+++ b/dts/bindings/led/pwm-leds.yaml
@@ -1,5 +1,6 @@
 # Copyright (c) 2018, Linaro Limited
 # SPDX-License-Identifier: Apache-2.0
+include: base.yaml
 
 description: |
   Group of PWM-controlled LEDs.


dts/bindings/led/pwm-leds.yaml
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) 2018, Linaro Limited
 # SPDX-License-Identifier: Apache-2.0
+include: base.yaml
```
simonguinot, line 3: Hi @ndreys,

Thanks for this patch.

What problem are you trying to solve by allowing to defer the initialization of a PWM LED device ?

jeppenodgaard (commented):
Based on the [dts how-to](https://docs.zephyrproject.org/latest/build/dts/howtos.html#write-device-drivers-using-devicetree-apis), it looks like `base.yaml` should always be included.

You probably have a better idea @simonguinot, but I would suggest adding `base.yaml` to `led-node.yaml` which should (it is currently not) be included by the driver binding.

dts/bindings/led/pwm-leds.yaml
ndreys: I have a single image that is used on multiple hardware variants where only some of all declared PWM LED devices are physically present on the board. We use hardware ID queried at the boot time and enable only those LEDs there are there.

thedjnK (approved):

fabiobaltieri (approved):

soburi (approved):

dts/bindings/led/pwm-leds.yaml
simonguinot: OK. So you intend to use the `zephyr,deferred-init` property at the child node level. In that case, shouldn't we instead include `base.yaml` at the `child-binding` level ?

ndreys: Child node? I'm not sure I follow what you're saying at all. 

I have a node that looks like
````
	some-led {
		compatible = "pwm-leds";
		status = "okay";
		zephyr,deferred-init;

````
all I want to do is to be able to include that last line. What's the concern with the patch as is?

**Comment by simonguinot** [comment]
> Based on the [dts how-to](https://docs.zephyrproject.org/latest/build/dts/howtos.html#write-device-drivers-using-devicetree-apis), it looks like `base.yaml` should always be included.
> 
> You probably have a better idea @simonguinot, but I would suggest adding `base.yaml` to `led-node.yaml` which should (it is currently not) be included by the driver binding.

That's the case actually, but `label` is the only allowed property. And yes, you're probably right. It would make sense to make the `zephyr,deferred-init` property available for all types of LED child nodes, not just PWM LEDs. But `base.yaml` also provides a ton of properties that are irrelevant for a LED node.

File: dts/bindings/led/pwm-leds.yaml (hunk 1 above)
Comment (simonguinot) [reply dts/bindings/led/pwm-leds.yaml:3]: > Child node? I'm not sure I follow what you're saying at all.

I think I got confused by the proposition of @jeppenodgaard, to include `base.yaml` in `led-node.yml`.

> 
> I have a node that looks like
> 
> ```
> 	some-led {
> 		compatible = "pwm-leds";
> 		status = "okay";
> 		zephyr,deferred-init;
> ```

OK, so you split your PWM LEDs in several blocks: 

```
pwmleds_group0 {
        compatible = "pwm-leds";
        zephyr,deferred-init;

        pwmled0: pwmled_0 {
                ...;
        };

        pwmled1: pwmled_1 {
                ...;
        };
...
};

pwmleds_group1 {
        compatible = "pwm-leds";
        zephyr,deferred-init;
        ...
}
```

And depending on the board on which the application is running, you call `device_init` on the available PWM LED groups. Is that correct ?

> 
> all I want to do is to be able to include that last line. What's the concern with the patch as is?

Probably none. I just need to understand the use case, in order to ensure the patch is correct.

ndreys: > And depending on the board on which the application is running, you call device_init on the available PWM LED groups. Is that correct ?

Yeah, that app code know to try to find the group that's enabled and use it as the LED.

simonguinot (approved):
LGTM !

PR #99445, opened 2025-11-15
lsm6dsv16x: Switch the driver to use level triggered IRQ

ndreys:
The benefit of using edge triggered interrupt is that they allow one to write handlers that don't require disabling/enabling of the interrupts. This driver donesn't leverage that and, in fact, by using edge trigger it opens up a race condition window between the read of the status register and re-enablement of the IRQ. During that window and IRQ can arrive and be lost stalling the data fetching completely.

### DIFF
--- a/drivers/sensor/st/lsm6dsv16x/lsm6dsv16x_trigger.c
+++ b/drivers/sensor/st/lsm6dsv16x/lsm6dsv16x_trigger.c
@@ -273,7 +273,7 @@ static void lsm6dsv16x_handle_interrupt(const struct device *dev)
 
 	if (!ON_I3C_BUS(cfg) || (I3C_INT_PIN(cfg))) {
 		ret = gpio_pin_interrupt_configure_dt(lsm6dsv16x->drdy_gpio,
-						GPIO_INT_EDGE_TO_ACTIVE);
+						GPIO_INT_LEVEL_ACTIVE);
 		if (ret < 0) {
 			LOG_ERR("%s: Not able to configure pin_int", dev->name);
 		}
@@ -488,5 +488,5 @@ int lsm6dsv16x_init_interrupt(const struct device *dev)
 #endif
 
 	return gpio_pin_interrupt_configure_dt(lsm6dsv16x->drdy_gpio,
-					       GPIO_INT_EDGE_TO_ACTIVE);
+					       GPIO_INT_LEVEL_ACTIVE);
 }


avisconti:
@ndreys shouldn't you change all the references to GPIO_INT_EDGE_TO_ACTIVE? There are many also in lsm6dsv16x_rtio_stream.c. Moreover, I think that we may add a property in the DT defaulting to INT_EDGE_TO_ACTIVE that one can change. Like st,lis2dh does:

```
  int1-gpio-config:
    type: int
    default: 0
    description: |
      Select the interrupt configuration for INT1 gpio.

      The default of 0 is the most common situation to avoid multiple interrupts
      to be triggered by same event.

      - 0 # LIS2DH_DT_GPIO_INT_EDGE_BOTH
      - 1 # LIS2DH_DT_GPIO_INT_EDGE_RISING
      - 2 # LIS2DH_DT_GPIO_INT_EDGE_FALLING
      - 3 # LIS2DH_DT_GPIO_INT_LEVEL_HIGH
      - 4 # LIS2DH_DT_GPIO_INT_LEVEL_LOW

    enum: [0, 1, 2, 3, 4]
```

What do you think? I think would be better.

avisconti (changes requested):
consider to apply suggestions

ndreys:
> @ndreys shouldn't you change all the references to GPIO_INT_EDGE_TO_ACTIVE? There are many also in lsm6dsv16x_rtio_stream.c. 

Yeah, I noticed that a bit ago too. My patch pre-dates the addition of RTIO support, so it doesn't have that change. I'm in the process of migration my codebase from trigger callbacks to RTIO streaming, so I'll update this patch once I'm finished.

> Moreover, I think that we may add a property in the DT defaulting to INT_EDGE_TO_ACTIVE that one can change. Like st,lis2dh does:
> 
> ```
>   int1-gpio-config:
>     type: int
>     default: 0
>     description: |
>       Select the interrupt configuration for INT1 gpio.
> 
>       The default of 0 is the most common situation to avoid multiple interrupts
>       to be triggered by same event.
> 
>       - 0 # LIS2DH_DT_GPIO_INT_EDGE_BOTH
>       - 1 # LIS2DH_DT_GPIO_INT_EDGE_RISING
>       - 2 # LIS2DH_DT_GPIO_INT_EDGE_FALLING
>       - 3 # LIS2DH_DT_GPIO_INT_LEVEL_HIGH
>       - 4 # LIS2DH_DT_GPIO_INT_LEVEL_LOW
> 
>     enum: [0, 1, 2, 3, 4]
> ```
> 
> What do you think? I think would be better.

Well my concern is that interrupt handlers for edge triggered interrupts need to be written differently than level triggered handlers. For edge triggered IRQs we'd either have to stop disabling/enabling the IRQ or add an extra status register check after each re-enablement to avoid a race between IRQ enablement and and edge of the IRQ signal. 🤷 all could be done but it's way more code, edge triggered interrupts aren't supported natively by a number of GPIO controllers, so I'm not sure I see the use-case for edge triggered IRQs or allowing this configurability. Maybe if there was a platform where GPIO block supported edge IRQ but not level IRQs?

ndreys:
> Not all parts support level triggers. I haven’t seen one that doesn’t support edge triggers

Do you have an example of a part that doesn't support level IRQs but supports edge detection? Just from HW implementation perspective I'd imagine level IRQs to be simpler circuit that is implemented first and the edge detection is implemented on top of it.

For an example of a part that doesn't support edge detection natively, look at nRF52 GPIO IP block. It doesn't have native edge detection capability and to have that you either need to connect the IP block to a separate IP block that has a limited number of channels or rely on the software edge detection fallback that is implemented in nrf HAL layer.

avisconti:
> @avisconti since it looks like ST parts do not support level triggered IRQs, the fix as is might break using ST's IMUs on ST parts which seems like something that you guys might not want.

Mmmh, maybe we can configure it in DT.
See https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/sensor/st,lis2dh-common.yaml>
There are properties such as int1-config and int2-config.

And the lis2dh driver configure interrupt accordingly

EDIT:
See that I was suggesting exactly the same some time ago here:
https://github.com/zephyrproject-rtos/zephyr/pull/99445#issuecomment-3611900487

Any doubt I have missed?

avisconti:
I prepared #109823 which may put in obsolescence this PR.

avisconti:
Made obsolete by https://github.com/zephyrproject-rtos/zephyr/pull/109823

PR #99447, opened 2025-11-15
usb: hid: Introduce HID_USAGE_PAGE16()

ndreys:
Introduce HID_USAGE_PAGE16() in order to be able to declare vendor specific usage pages which require 16-bit numbers.

### DIFF
--- a/include/zephyr/usb/class/hid.h
+++ b/include/zephyr/usb/class/hid.h
@@ -280,6 +280,15 @@ extern "C" {
 	HID_ITEM(HID_ITEM_TAG_USAGE_PAGE, HID_ITEM_TYPE_GLOBAL, 1), page
 
 /**
+ * @brief Define HID Usage Page item with the data length of two bytes.
+ *
+ * @param page Usage Page
+ * @return     HID Usage Page item
+ */
+#define HID_USAGE_PAGE16(page)		\
+	HID_ITEM(HID_ITEM_TAG_USAGE_PAGE, HID_ITEM_TYPE_GLOBAL, 2), (uint8_t)page, (page >> 8)
+
+/**
  * @brief Define HID Logical Minimum item with the data length of one byte.
  *
  * For usage examples, see @ref HID_MOUSE_REPORT_DESC(),


include/zephyr/usb/class/hid.h
```diff
@@ -267,17 +267,17 @@ extern "C" {
...
 	HID_ITEM(HID_ITEM_TAG_COLLECTION_END, HID_ITEM_TYPE_MAIN, 0)
 
+#define HID_USAGE_PAGE(page)		\
+	HID_ITEM(HID_ITEM_TAG_USAGE_PAGE, HID_ITEM_TYPE_GLOBAL, 1), page
+
 /**
- * @brief Define HID Usage Page item.
- *
- * For usage examples, see @ref HID_MOUSE_REPORT_DESC(),
- * @ref HID_KEYBOARD_REPORT_DESC()
```
tmon-nordic, line 274: Why are you removing the description from HID_USAGE_PAGE?

ndreys: Ugh, didn't mean to do that, will fix.

ndreys: Should be fixed.

jfischer-no (approved):

include/zephyr/usb/class/hid.h
```diff
@@ -279,6 +279,15 @@ extern "C" {
...
 
+/**
+ * @brief Define HID Usage Page item with the data length of two bytes.
+ *
+ * @param page Usage Page
+ * @return     HID Usage Page item
+ */
+#define HID_USAGE_PAGE16(page)		\
+	HID_ITEM(HID_ITEM_TAG_USAGE_PAGE, HID_ITEM_TYPE_GLOBAL, 2), (uint8_t)page, (page >> 8)
+
```
josuah, line 290: Would that be an `HID_USAGE_ID()` then?

> **Usage Page** *000001nn*
> Unsigned integer specifying the current Usage Page. Since a usage are 32 bit values, Usage Page items can be used to conserve space in a report descriptor by setting the high order 16 bits of a subsequent usages. Any usage that follows which ~~is~~ defines 16 bits or less **is interpreted as a Usage ID** and concatenated with the Usage Page to form a 32 bit Usage.

josuah (commented):
I have a doubt, but could be because I am new to USB HID...

include/zephyr/usb/class/hid.h
ndreys: No, what this paragraph is saying is that you can save space in your report descriptor by specifying top 16-bits of a Usage (which is 32 bits) once and then only changing lower 16-bits with a 16-bit Usage tag. With _the latter_ being interpreted as "usage ID".  So something like

```
USAGE_PAGE(0xdead)
...
USAGE(0xf00d)   # -> full usage 0xdeadf00d

...

USAGE(0xd00d) # -> full usage 0xdeadd000d
```

etc.

josuah: I didn't see the difference with `HID_USAGE16` below but it's `HID_ITEM_TYPE_GLOBAL` vs `_LOCAL`.

Thanks, makes more sense now.

josuah (approved):

PR #99448, opened 2025-11-15
usb: usbd: Partially unconstify .ptr in struct usbd_desc_node

ndreys:
It is useful to be able to set .ptr to point to a string obtained at runtime, so let's remove the second const qualifier so that the pointer itself can be adjusted.

### DIFF
--- a/include/zephyr/usb/usbd.h
+++ b/include/zephyr/usb/usbd.h
@@ -162,7 +162,7 @@ struct usbd_desc_node {
 		struct usbd_bos_desc_data bos;
 	};
 	/** Opaque pointer to a descriptor payload */
-	const void *const ptr;
+	const void *ptr;
 	/** Descriptor size in bytes */
 	uint8_t bLength;
 	/** Descriptor type */


josuah:
Could we know a bit more about the use-case?
This would confirm that what you were trying to do is not already supported in a different way, or that maybe it needs some dedicated API.
Thanks!

josuah:
Sorry for such a long wait.

Do you think this would work to use HWINFO as back-end for the serial number? The application can do a simple HWINFO driver which is just defining a global function.

https://github.com/zephyrproject-rtos/zephyr/blob/6182bc08c9d72e9e9aed8f5cf05a406fbfd25dd8/subsys/usb/device_next/usbd_ch9.c#L563-L577

Or did that not work in your situation?

aaron-ets:
I also have the same patch locally.  Similarly, we want to set the serial number to a vendor serial number that is set in flash and not a random hwinfo one.  In addition, we use the same mcu in multiple products so we need to set the product string at runtime after determining the board.

ndreys:
> Sorry for such a long wait.
> 
> Do you think this would work to use HWINFO as back-end for the serial number? The application can do a simple HWINFO driver which is just defining a global function.
> 
> https://github.com/zephyrproject-rtos/zephyr/blob/6182bc08c9d72e9e9aed8f5cf05a406fbfd25dd8/subsys/usb/device_next/usbd_ch9.c#L563-L577
> 
> Or did that not work in your situation?

Unfortunately not. `hwinfo_get_device_id()` implies that device ID is a a long hex number and my serial number are alphanumeric strings. If I plug them in via that route they are going to get mangled by
```
static ssize_t get_sn_from_hwid(uint8_t sn[static USBD_SN_ASCII7_LENGTH])
{
	static const char hex[] = "0123456789ABCDEF";
	uint8_t hwid[USBD_SN_ASCII7_LENGTH / 2U];
	ssize_t hwid_len = -ENOSYS;

	if (IS_ENABLED(CONFIG_HWINFO)) {
		hwid_len = hwinfo_get_device_id(hwid, sizeof(hwid));
	}

	if (hwid_len < 0) {
		if (hwid_len == -ENOSYS) {
			LOG_ERR("HWINFO not implemented or enabled");
		}

		return hwid_len;
	}

	for (ssize_t i = 0; i < MIN(hwid_len, sizeof(hwid)); i++) {
		sn[i * 2] = hex[hwid[i] >> 4];
		sn[i * 2 + 1] = hex[hwid[i] & 0xF];
	}

	return MIN(hwid_len, sizeof(hwid)) * 2;
}
```

:shrug: Maybe one option could be to extend HWINFO API to include something like `hwinfo_get_device_serial_number()`? There's definitely use-case for that in the BT subsystem. Although I don't know if that's going to raise any objections from the HWINFO maintainers.

josuah (dismissed):
(trying to move this forward)

Both options seems to work really.

If wanting to go with the road of modifying the pointer, though, I anticipate that modifying the "opaque pointer" directly is not going to be the recommended way. For instance, the comment still implies "opaque pointer, update at your own risk".

This might need at least a wrapper function to access it, i.e. put in `include/zephyr/usb/usbd.h`, that makes sure to update the `bLength`, set `bDescriptorType`.

jfischer-no:
> I'm setting up device serial to contain the contents of a user (read "at manufacturing time") programmable flash. We don't know if that data is valid at compile time, so that's a string that is a result of a function call. Basically I'm doing
> 
> ```
>     static struct usbd_desc_node sn = {
>         .str = {
>             .utype = USBD_DUT_STRING_SERIAL_NUMBER,
>             .ascii7 = true,
>         },
>         .bLength = USB_STRING_DESCRIPTOR_LENGTH(...),
>         .bDescriptorType = USB_DESC_STRING,
>     };
> 
>     sn.ptr = unit_serial_get();
> ```
> 
> older versions of `USBD_DESC_STRING_DEFINE()` didn't have `ascii_##d_name` as const, so I was originally just filling that, but it got constified recently, so I can't use that macro anymore.

```diff
+#define USBD_DESC_STRING_RAM_DEFINE(d_name, d_string, d_utype)			\
+	static uint8_t ascii_##d_name[sizeof(d_string)] = d_string;		\
+	static struct usbd_desc_node d_name = {					\
+		.str = {							\
+			.utype = d_utype,					\
+			.ascii7 = true,						\
+		},								\
+		.ptr = &ascii_##d_name,						\
+		.bLength = USB_STRING_DESCRIPTOR_LENGTH(d_string),		\
+		.bDescriptorType = USB_DESC_STRING,				\
+	}
+
+USBD_DESC_STRING_RAM_DEFINE(sn, "12345", USBD_DUT_STRING_SERIAL_NUMBER);
```
@ndreys I am not sure that we need to modify the pointer type. Would a macro that places the ASCII7 array in RAM resolve your issue? As you mentioned above, your application could overwrite the array's default content when needed. I opened https://github.com/zephyrproject-rtos/zephyr/pull/107208 as there could be other potential use cases.

aaron-ets:
It seems to me like the zephyr-ey thing to do would be to have the macros for compile time initialization and `usbd_desc_serial_number_initialize(struct usbd_desc_node *desc, const char *sn);` to initialize at runtime, for each string descriptor?

ndreys:
> > I'm setting up device serial to contain the contents of a user (read "at manufacturing time") programmable flash. We don't know if that data is valid at compile time, so that's a string that is a result of a function call. Basically I'm doing
> > ```
> >     static struct usbd_desc_node sn = {
> >         .str = {
> >             .utype = USBD_DUT_STRING_SERIAL_NUMBER,
> >             .ascii7 = true,
> >         },
> >         .bLength = USB_STRING_DESCRIPTOR_LENGTH(...),
> >         .bDescriptorType = USB_DESC_STRING,
> >     };
> > 
> >     sn.ptr = unit_serial_get();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > older versions of `USBD_DESC_STRING_DEFINE()` didn't have `ascii_##d_name` as const, so I was originally just filling that, but it got constified recently, so I can't use that macro anymore.
> 
> ```diff
> +#define USBD_DESC_STRING_RAM_DEFINE(d_name, d_string, d_utype)			\
> +	static uint8_t ascii_##d_name[sizeof(d_string)] = d_string;		\
> +	static struct usbd_desc_node d_name = {					\
> +		.str = {							\
> +			.utype = d_utype,					\
> +			.ascii7 = true,						\
> +		},								\
> +		.ptr = &ascii_##d_name,						\
> +		.bLength = USB_STRING_DESCRIPTOR_LENGTH(d_string),		\
> +		.bDescriptorType = USB_DESC_STRING,				\
> +	}
> +
> +USBD_DESC_STRING_RAM_DEFINE(sn, "12345", USBD_DUT_STRING_SERIAL_NUMBER);
> ```
> 
> @ndreys I am not sure that we need to modify the pointer type. Would a macro that places the ASCII7 array in RAM resolve your issue? As you mentioned above, your application could overwrite the array's default content when needed. I opened #107208 as there could be other potential use cases.

Well the advantage of changing the pointer type is that I don't have to store a copy of the serial number in RAM just for the sake of fully constifying that pointer :shrug: . Not the end of the world, but it seems that if we are thinking of a new API here might as well go for more "efficient" option. 

Yeah, `USBD_DESC_STRING_RAM_DEFINE()`, would codify the previous behavior of the `USBD_DESC_STRING_DEFINE()` so that could work.

jfischer-no:
> Well the advantage of changing the pointer type is that I don't have to store a copy of the serial number in RAM just for the sake of fully constifying that pointer 🤷 . Not the end of the world, but it seems that if we are thinking of a new API here might as well go for more "efficient" option.

I may have missed something, but you always need an array to store it. The built-in optional use of HWINFO is not that different.

ndreys:
> > Well the advantage of changing the pointer type is that I don't have to store a copy of the serial number in RAM just for the sake of fully constifying that pointer 🤷 . Not the end of the world, but it seems that if we are thinking of a new API here might as well go for more "efficient" option.
> 
> I may have missed something, but you always need an array to store it. 

This was in response to you saying "I am not sure that we need to modify the pointer type".

```
static struct usbd_desc_node sn = {
        .str = {
            .utype = USBD_DUT_STRING_SERIAL_NUMBER,
            .ascii7 = true,
        },
        .bLength = USB_STRING_DESCRIPTOR_LENGTH(...),
        .bDescriptorType = USB_DESC_STRING,
    };

    sn.ptr = unit_serial_get();
```

vs.

```
USBD_DESC_STRING_RAM_DEFINE(foo, "SOME PLACEHOLDER",  USBD_DUT_STRING_SERIAL_NUMBER);
strcpy(ascii_foo,   unit_serial_get());
```

All the rest of the USB stack behavior being equal, if I can modify where `sn.ptr` points to, I don't need to have `ascii_foo` or copy data from Flash into it in the former compared to the latter.

> The built-in optional use of HWINFO is not that different.

Built-in HWINFO doesn't work for me for a different reason, but yes, sure, it also does require some superfluous data movement, which isn't great either.

jfischer-no:
> All the rest of the USB stack behavior being equal, if I can modify where `sn.ptr` points to, I don't need to have `ascii_foo` or copy data from Flash into it in the former compared to the latter.


Okay, your serial number, in ASCII7 format and compliant with USB spec is already in the flash? So why you cannot define your own application macro like 

```diff
+#define USBD_DESC_STRING_CUSTOM_DEFINE(d_name, d_array, d_len)			\
+	static struct usbd_desc_node d_name = {					\
+		.str = {							\
+			.utype = USBD_DUT_STRING_SERIAL_NUMBER,			\
+			.ascii7 = true,						\
+		},								\
+		.ptr = &d_array,						\
+		.bLength = d_len * 2,						\
+		.bDescriptorType = USB_DESC_STRING,				\
+	}
+
+static const uint8_t my_ascii_sn[] = "AABBCCDD";
+USBD_DESC_STRING_CUSTOM_DEFINE(test_sn, my_ascii_sn, sizeof(my_ascii_sn));
```
 
 and just use it?

ndreys:
> > All the rest of the USB stack behavior being equal, if I can modify where `sn.ptr` points to, I don't need to have `ascii_foo` or copy data from Flash into it in the former compared to the latter.
> 
> Okay, your serial number, in ASCII7 format and compliant with USB spec is already in the flash? So why you cannot define your own application macro like
> 
> ```diff
> +#define USBD_DESC_STRING_CUSTOM_DEFINE(d_name, d_array, d_len)			\
> +	static struct usbd_desc_node d_name = {					\
> +		.str = {							\
> +			.utype = USBD_DUT_STRING_SERIAL_NUMBER,			\
> +			.ascii7 = true,						\
> +		},								\
> +		.ptr = &d_array,						\
> +		.bLength = d_len * 2,						\
> +		.bDescriptorType = USB_DESC_STRING,				\
> +	}
> +
> +static const uint8_t my_ascii_sn[] = "AABBCCDD";
> +USBD_DESC_STRING_CUSTOM_DEFINE(test_sn, my_ascii_sn, sizeof(my_ascii_sn));
> ```
> 
> and just use it?

Because the value of `.ptr` isn't known to me at the compile time. I get it as a result of a function call that happens at the run-time(`unit_serial_get()` in our hypothetical examples above).

jfischer-no:
> Because the value of `.ptr` isn't known to me at the compile time. I get it as a result of a function call that happens at the run-time(`unit_serial_get()` in our hypothetical examples above).

¯|_(ツ)_/¯ I do not think we should provide solutions for every possible peculiarity.

ndreys:
> > Because the value of `.ptr` isn't known to me at the compile time. I get it as a result of a function call that happens at the run-time(`unit_serial_get()` in our hypothetical examples above).
> 
> ¯|_(ツ)_/¯ I do not think we should provide solutions for every possible peculiarity.

LOL, OK

aaron-ets:
> ¯|_(ツ)_/¯ I do not think we should provide solutions for every possible peculiarity.

What is the downside of removing the const?  I'm surprised you consider this peculiar, it seems common practice to store a serial number in flash or an i2c eeprom that wouldn't be available at compile time.

jfischer-no:
> > ¯|_(ツ)_/¯ I do not think we should provide solutions for every possible peculiarity.
> 
> What is the downside of removing the const? I'm surprised you consider this peculiar, it seems common practice to store a serial number in flash or an i2c eeprom that wouldn't be available at compile time.

The downside is that it is no longer const. Josuah and I suggested a few different, common, portable, and more maintainable ways to archive what you want. Therefore, I do not think your changes make sense.

jfischer-no (changes requested):
See https://github.com/zephyrproject-rtos/zephyr/pull/99448#issuecomment-4240118341

PR #99449, opened 2025-11-15
RFC: i2c: target: Add HID over I2C support

ndreys:
Add support of target side of things of HID over I2C protocol. The code in this patch allows the creation of a basic input device that can be probed by Linux and queried for input reports. This is a minimal viable implementation and a lot of basic features (get/set report for example) are missing.

NOTE: This is a very rough first cut of the code and I'd expect some changes for sure. I just want to publish this sooner to get some feedback.


ndreys:
Not stale, please remove the label. I should be able to get back to this in next couple of weeks. The feedback has been pretty sparse though. E.g. I'd imagine people having some opinions about how to make now USB specific HID bits transport independent. Is there a better venue to bring this one to attention?

teburd:
Please do get CI passing, the feature would be great to see in Zephyr

PR #95588, opened 2025-09-06
sensor: current_amp: optional high-range gain fallback

JordanYates:
Add the option to specify an alternate ADC gain value to use if the initial measurement saturates the range. This enables higher data resolutions when the values are small compared to the maximum signal values, while still supporting the maximum.

As a concrete example, measuring charge currents from a small solar panel (0 - 50mA), while also supporting high USB charge currents (up to 1A).

PR #99454, opened 2025-11-15
drivers: current_sense_amplifier: fix zero-current-voltage binding

### LINKED ISSUE #99453: sensor: current-amp: zero-current-offset no longer works
### Describe the bug

#87988 introduced a zero-current-offset, useful for hall sensors.
#95588 then rewrote the conversion algorithm from milliamps to microamps and forgot to include said zero-current-offset, breaking all peripherals that use said offset.

For my device, I use a CC6920BSO-5A, which [has a sensitivity of 264 mA/V](https://www.lcsc.com/datasheet/C2880430.pdf).

I use the following device tree node:
```dts
    curr_sens: curr_sens {
        compatible = "current-sense-amplifier";
        io-channels = <&adc1 0>;
        sense-resistor-milli-ohms = <1>;
        sense-gain-mult = <264>;
        zero-current-voltage-mv = <1650>;
    };
```

I consume about 1.1A and get the following debug output:
```none
[00:00:05.420,000] <inf> current_amp: 1667/4095, 1343041 uV, current:5087276 uA
u
```

I then patched `current_sense_amplifier_scale_ua_dt` and added:
```c
temp = temp - 1000 * (int64_t)(spec->zero_current_voltage_mv);
```

... and now it works:
```none
[00:00:21.368,000] <inf> current_amp: 1664/4095, 1340625 uV, current:-1171875 uA
```

### Regression

- [x] This is a regression.

### Steps to reproduce

Not that easy, I don't know of a dev board with a built in hall sensor. It's very trivial to reproduce logically, the offset calculation in `current_sense_amplifier_scale_dt` is simply missing in `current_sense_amplifier_scale_ua_dt`.

### Relevant log output

```shell

```

### Impact

### Environment

Zephyr v4.3.0

### Additional Context

_No response_

Finomnis:
Fixes #99453.

It seems the new microvolt/microamp rework (#95588) simply forgot to implement zero current voltage offsets.

### DIFF
--- a/include/zephyr/drivers/adc/current_sense_amplifier.h
+++ b/include/zephyr/drivers/adc/current_sense_amplifier.h
@@ -88,6 +88,7 @@ current_sense_amplifier_scale_ua_dt(const struct current_sense_amplifier_dt_spec
 	 *          (INT32_MAX * 1000) * UINT16_MAX <= INT64_MAX
 	 *                                   ~2**57 <= 2**63
 	 */
+	temp = temp - 1000 * (int64_t)(spec->zero_current_voltage_mv);
 	int64_t scaled = temp * 1000 * spec->sense_gain_div;
 	/* Perform final divisions */
 	return scaled / spec->sense_gain_mult / spec->sense_milli_ohms;


JordanYates (approved):

Finomnis:
@JordanYates how did multiple reviewers miss that? Things like this make me kind of uneasy about the code quality of Zephyr...

I mean, to visually compare the new implementation with the previous implementation is the least a reviewer should do, and imo it is quite hard to miss...

JordanYates:
> @JordanYates how did multiple reviewers miss that?

Presumably because the problem was not with lines of code removed from the PR, the difference can only be caught it you compare line by line the swapped out function. For my part, the logic was written before the offset binding PR was opened.

> Things like this make me kind of uneasy about the code quality of Zephyr...

A niche feature of a niche driver broke on one release, its not ideal but it happens. If you can find an embedded framework that supports as much as Zephyr does with better testing coverage, I'm sure we'd all like to hear about it.

> I mean, to visually compare the new implementation with the previous implementation is the least a reviewer should do, and imo it is quite hard to miss...

People have limited time, I absolutely do not blame any reviewer for missing this. The great part about Zephyr is that anyone can contribute. Since the problem was obvious to you, it would be great if you could help us proactively review PRs.

anangl (approved):

PR #99455, opened 2025-11-15
samples: ocpp: add native_sim config

walidbadar:
-     Disable DHCPv4 and use static network configuration for native_sim board.


samples/net/ocpp/boards/native_sim.conf
```diff
@@ -0,0 +1,7 @@
+# Network address config
+CONFIG_NET_DHCPV4=n
+CONFIG_NET_CONFIG_SETTINGS=y
+CONFIG_NET_CONFIG_NEED_IPV4=y
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.0.2.1"
```
ssekar15, line 5: no hard coded value please

samples/net/ocpp/src/main.c
```diff
@@ -39,6 +39,11 @@ static int ocpp_get_time_from_sntp(void)
 	struct timespec tv;
 	int ret;
 
+	if (strlen(CONFIG_NET_SAMPLE_SNTP_SERVER) < 1) {
```
ssekar15: run time error will be get down in `sntp_init`.

ssekar15 (changes requested):

samples/net/ocpp/boards/native_sim.conf
jukkar: This IP address space is reserved for documentation and we are using it in all the network samples, so these are OK to be used here.

samples/net/ocpp/src/main.c
walidbadar: No, `ocpp_get_time_from_sntp` would `return -EINVAL` before getting to `sntp_init`.

ssekar15: the change is not needed, if CONFIG_NET_SAMPLE_SNTP_SERVER not configured which will endup with run time error in sntp_init.

samples/net/ocpp/src/main.c
```diff
@@ -298,14 +303,22 @@ int main(void)
...
 	ret = ocpp_getaddrinfo(CONFIG_NET_SAMPLE_OCPP_SERVER, CONFIG_NET_SAMPLE_OCPP_PORT, &ip);
 	if (ret < 0) {
 		return ret;
 	}
 
 	csi.cs_ip = ip;
 
-	ocpp_get_time_from_sntp();
+	ret = ocpp_get_time_from_sntp();
+	if (ret < 0) {
```
ssekar15: should not return even sntp unreachable.. e.g system time is maintained in RTC or some other source.

samples/net/ocpp/src/main.c
```diff
@@ -298,14 +303,22 @@ int main(void)
 
 	wait_for_network();
 
+	if (strlen(CONFIG_NET_SAMPLE_OCPP_SERVER) < 1) {
```
ssekar15: ditto, ocpp_getaddrinfo endup runtime error.

samples/net/ocpp/boards/native_sim.conf
ssekar15: noted, good to know.

ssekar15 (changes requested):

samples/net/ocpp/src/main.c
walidbadar: the reason for this logic was to warn the user that ocpp server address was not configured, there could also be other reason why `getaddrinfo` might get fail.

jukkar: We could avoid the call to strlen() here
 ```
if (CONFIG_NET_SAMPLE_OCPP_SERVER[0] == '\0') {
```

ssekar15 (approved):

jukkar (approved):

PR #67280, opened 2024-01-07
boards: m5stack_core2: Add possibility to control 5V_BUS signal

MrMarteng:
This commit fixes [#67277](https://github.com/zephyrproject-rtos/zephyr/issues/67277) and hence adds possibilityto properly utilize grove extension port on M5Stack Core2 module.

5V_BUS can be enabled via regulator bus_5v. By default this regulator is disabled.

PR #99466, opened 2025-11-16
Add 5V power regulators for external outputs on CoreS3/S3-SE

xingrz:
This PR models all GPIO-controlled 5V rails on the M5Stack CoreS3/CoreS3-SE as fixed regulators: `bus_5v`, `bus_out_reg`, and `usb_otg_reg`.

The `bus_5v` regulator matches the equivalent node present on the M5Stack Core2 (#67280).

The `bus_out_reg` and `usb_otg_reg` regulators expose the two load switches that control the direction of 5V power between the USB-C port, the battery, and the M-Bus/Grove connectors, as summarized below:

| `BUS_OUT_EN` | `USB_OTG_EN` | Battery   | USB-OTG | M-Bus/Grove |
|--------------|--------------|-----------|---------|-------------|
| 0            | 0            | Charging  | Input   | Input       |
| 1            | 0            | Charging  | Input   | Output      |
| 0            | 1            | Charging  | Output  | Input       |
| 1            | 1            | Discharge | Output  | Output      |

The `bus_out_reg` regulator is enabled by default to match the behavior of existing M5Stack boards. `usb_otg_reg` remains disabled by default, as the ESP32-S3 does not currently support USB host mode.

### DIFF
--- a/boards/m5stack/m5stack_cores3/Kconfig.defconfig
+++ b/boards/m5stack/m5stack_cores3/Kconfig.defconfig
@@ -24,19 +24,36 @@ endif # LVGL
 
 # Takes good care of initialization priorities
 
+if REGULATOR_FIXED
+
+# Needs to init after MFD_AW9523B (80), since bus_5v uses aw9523b_gpio.
+config REGULATOR_FIXED_INIT_PRIORITY
+	default 81
+
+endif # REGULATOR_FIXED
+
 if REGULATOR_AXP192_AXP2101
+
+# Needs to init before the display (83), since AXP2101 powers it.
 config REGULATOR_AXP192_AXP2101_INIT_PRIORITY
 	default 81
+
 endif # REGULATOR_AXP192_AXP2101
 
 if MIPI_DBI
+
+# Needs to init after MFD_AW9523B (80), since reset-gpios uses aw9523b_gpio on this board.
 config MIPI_DBI_INIT_PRIORITY
 	default 82
+
 endif # MIPI_DBI
 
 if DISPLAY
+
+# Needs to init after REGULATOR_AXP192_AXP2101 (81) and MIPI_DBI (82).
 config DISPLAY_INIT_PRIORITY
 	default 83
+
 endif # DISPLAY
 
 endif # BOARD_M5STACK_CORES3_ESP32S3_PROCPU || BOARD_M5STACK_CORES3_ESP32S3_PROCPU_SE
--- a/boards/m5stack/m5stack_cores3/m5stack_cores3_procpu_common.dtsi
+++ b/boards/m5stack/m5stack_cores3/m5stack_cores3_procpu_common.dtsi
@@ -69,6 +69,19 @@
 		regulator-name = "bus_5v";
 		enable-gpios = <&aw9523b_gpio 15 GPIO_ACTIVE_HIGH>;
 	};
+
+	bus_out_reg: bus_out {
+		compatible = "regulator-fixed";
+		regulator-name = "bus_out";
+		regulator-boot-on;
+		enable-gpios = <&aw9523b_gpio 1 GPIO_ACTIVE_HIGH>;
+	};
+
+	usb_otg_reg: usb_otg {
+		compatible = "regulator-fixed";
+		regulator-name = "usb_otg";
+		enable-gpios = <&aw9523b_gpio 5 GPIO_ACTIVE_HIGH>;
+	};
 };
 
 &usb_serial {
--- a/boards/m5stack/m5stack_cores3/m5stack_cores3_procpu_common.dtsi
+++ b/boards/m5stack/m5stack_cores3/m5stack_cores3_procpu_common.dtsi
@@ -63,6 +63,12 @@
 			rotation = <0>;
 		};
 	};
+
+	bus_5v: bus_5v {
+		compatible = "regulator-fixed";
+		regulator-name = "bus_5v";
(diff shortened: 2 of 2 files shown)


boards/m5stack/m5stack_cores3/Kconfig.defconfig
```diff
@@ -24,6 +24,11 @@ endif # LVGL
 
 # Takes good care of initialization priorities
 
+if REGULATOR_FIXED
+config REGULATOR_FIXED_INIT_PRIORITY
```
nordicjm, line 30: not properly formatted https://docs.zephyrproject.org/latest/contribute/style/kconfig.html

nordicjm (changes requested):
Also, shouldn't the bus just have a reference to the regulator itself? So that when it is enabled the PM code will power on the regulator, and power it off if the device is suspended? @fabiobaltieri ?

xingrz:
> shouldn't the bus just have a reference to the regulator itself? So that when it is enabled the PM code will power on the regulator, and power it off if the device is suspended?

The bus can also be externally powered. For example the [Base LAN PoE v1.2](https://docs.m5stack.com/en/base/lan_poe_v12) feeds 5V back into pin 28 of the bus.

xingrz:
> I’m planning to model those shields as MFDs

For those interested in the overall approach, here is a WIP branch: https://github.com/xingrz/zephyr/compare/cores3-power/pr...m5-bytebutton/dev.

I'm still gathering more Units to verify the design, and will submit a follow-up PR for the Unit support once it’s ready.

boards/m5stack/m5stack_cores3/Kconfig.defconfig
```diff
@@ -24,6 +24,11 @@ endif # LVGL
...
+	default 81
```
fabiobaltieri, line 31: why do you have to change this? may be worth a comment

xingrz: The `bus_5v` regulator depends on `aw9523b_gpio` for its enable pin, and that GPIO comes from the `aw9523b` MFD. The regulator must therefore init after the MFD (priority 80), hence the choice of 81.

fabiobaltieri: ok thanks, bit of a priority mess going on here, would be nice to clean it up but that's well beyond the scope of this

boards/m5stack/m5stack_cores3/Kconfig.defconfig
```diff
@@ -24,17 +24,26 @@ endif # LVGL
...
+endif # REGULATOR_FIXED
+
 if REGULATOR_AXP192_AXP2101
+# Needs to init before the display (83), since AXP2101 powers it.
 config REGULATOR_AXP192_AXP2101_INIT_PRIORITY
 	default 81
 endif # REGULATOR_AXP192_AXP2101
 
 if MIPI_DBI
+# Needs to init after MFD_AW9523B (80), since reset-gpios uses aw9523b_gpio on this board.
```
fabiobaltieri, line 45: Oh I see this impacts multiple drivers, would it not be easier to change the aw9523b_gpio priority instead and leave the others alone?

xingrz: `aw9523b_gpio` already initializes at priority 40, which is not late. The issue is that it relies on a `k_sem` from the `aw9523b` MFD, so anything using `aw9523b_gpio` must be initialized after the MFD itself. https://github.com/zephyrproject-rtos/zephyr/blob/5973b553ed94bc9ffea35319e08ee618bd9ba7e6/drivers/gpio/gpio_aw9523b.c#L96

fabiobaltieri (dismissed):

boards/m5stack/m5stack_cores3/Kconfig.defconfig
nordicjm: not fixed

xingrz: You mean an empty line after `if` and before `endif`? Oh I see.

nordicjm (dismissed):

fabiobaltieri (dismissed):

snippets/m5stack/cores3-bus-out-en/snippet.yml
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2025 Chen Xingyu <hi@xingrz.me>
+# SPDX-License-Identifier: Apache-2.0
+
+name: cores3-bus-out-en
```
tejlmand: snippets should be more general purpose and not this board specific.

Could this just be an overlay for the board itself or perhaps a board variant target ?
same comment applies to other snippet.

tejlmand (changes requested):

snippets/m5stack/cores3-bus-out-en/snippet.yml
xingrz: That makes sense. How about this: keep BUS_OUT enabled by default (matching the behavior on Core2 and other M5Stack boards) and keep USB_OTG disabled for now since ESP32-S3 doesn't support USB host yet?

tejlmand: sounds ok to me if that means the general recommendations are still followed by the board:
https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#general-recommendations

tejlmand (approved):

nashif (approved):

PR #99467, opened 2025-11-16
drivers: watchdog: siwx91x: Fix pause in sleep

### LINKED ISSUE #103242: Watchdog timer on Siwx91x devices is not on pause during deep sleep
### Describe the bug

On Siwx91x devices, the watchdog timers designed to pause during deep-sleep and resume after wakeup.
An incorrect check in wdt_setup was preventing this intended behavior.
### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

smalae:
- The watchdog timer on Siwx91x devices is designed to pause during deep-sleep and resume after wakeup.
However, an incorrect check in`wdt_setup` was preventing this intended behavior. Remove the check, restoring the correct pause-in-sleep functionality.

- Add a configuration file for the siwx917_rb4338a and siwx917_rb4342a boards to run the test_wdt_enable_wait_mode
test with power management enabled.

### DIFF
--- a/tests/drivers/watchdog/wdt_basic_api/testcase.yaml
+++ b/tests/drivers/watchdog/wdt_basic_api/testcase.yaml
@@ -162,3 +162,9 @@ tests:
     platform_allow:
       - frdm_mcxw23
       - mcxw23_evk
+  drivers.watchdog.pm_enabled:
+    platform_allow:
+      - siwx917_rb4338a
+    extra_configs:
+      - CONFIG_PM=y
+      - CONFIG_PM_DEVICE=y
--- a/drivers/watchdog/wdt_silabs_siwx91x.c
+++ b/drivers/watchdog/wdt_silabs_siwx91x.c
@@ -149,10 +149,6 @@ static int siwx91x_wdt_setup(const struct device *dev, uint8_t options)
 		return -ENOTSUP;
 	}
 
-	if (options & (WDT_OPT_PAUSE_IN_SLEEP)) {
-		return -ENOTSUP;
-	}
-
 	RSI_WWDT_ConfigSysRstTimer(config->reg, data->delay_reset);
 	RSI_WWDT_ConfigIntrTimer(config->reg, data->delay_irq);
 


jerome-pouiller (commented):
Globally, the Zephyr API tests do not enable PM. However, this scenario is indeed relevant, you can declare a new entry in testcase.yaml with `platform_allow: siwx917_rb4338a` and  `extra_configs: CONFIG_PM=y`.

BTW, it is not required to test the same hardware block twice. So, you just have to declare `rb4342`.

smalae:
> Globally, the Zephyr API tests do not enable PM. However, this scenario is indeed relevant, you can declare a new entry in testcase.yaml with `platform_allow: siwx917_rb4338a` and `extra_configs: CONFIG_PM=y`.
> 
> BTW, it is not required to test the same hardware block twice. So, you just have to declare `rb4342`.

Done

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
```diff
@@ -36,6 +36,8 @@ tests:
       - raytac_an54lq_db_15/nrf54l15/cpuapp/ns
       - frdm_mcxw23
       - mcxw23_evk
+      - siwx917_rb4338a
+      - siwx917_rb4342a
```
Martinhoff-maker: Since you have removed the overlay, you need to remove that board from "platform allow" too in both testcase.

smalae: I want to reuse the same overlay for both 4338a and 4342a boards. This particular test case is executed only for the 917 boards, as they have been explicitly excluded from other tests.

jerome-pouiller: hmm... I am not sure to understand. If `PM_DEVICE` is not set, watchdog still work, no? So we want to enable `drivers.watchdog` and `drivers.watchdog.wdt_basic_test_suite.pm_enabled`, no? I believe, rhere is no reason to blacklist `siwx917_rb4338a` from this test.

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
```diff
@@ -162,3 +164,11 @@ tests:
...
       - frdm_mcxw23
       - mcxw23_evk
+  drivers.watchdog.wdt_basic_test_suite.pm_enabled:
+    platform_allow:
+      - siwx917_rb4338a
+      - siwx917_rb4342a
+    extra_configs:
+      - CONFIG_PM=y
+      - CONFIG_PM_DEVICE=y
+    extra_args: DTC_OVERLAY_FILE="boards/siwx917_rb4338a.overlay"
```
jerome-pouiller: The idea to avoid to run unit test on two board with the same hardware block. So, I suggest to only keep `siwx917_rb4338a` in the list of `platform_allow` and remove the `DTC_OVERLAY_FILE="boards/siwx917_rb4338a.overlay"` since it will be automatic.

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
smalae: The entire test is skipped if CONFIG_PM is not enabled. When drivers.watchdog is enabled, the result is effectively no output.

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
smalae: Done

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
smalae: Added back siwx917_rb4338a to drivers.watchdog test

jerome-pouiller (approved):

smalae:
> @smalae I think the second commit body is now not correct.

Updated

Martinhoff-maker (approved):

asmellby (approved):

PR #99477, opened 2025-11-16
cmake: arm: Fix Cortex-R5 FPU configuration logic

pvinci:
The current logic for Cortex-R5 FPU flags is inverted - it adds +nofp when FPU IS enabled, which is backwards.

Fixed logic:
- If FPU is disabled: add +nofp suffix
- If FPU is enabled without double precision: add +nofp.dp suffix
- If FPU is enabled with double precision: use plain cortex-r5

This fixes builds for Cortex-R5F and other Cortex-R5F parts with VFPv3-D16.


cmake/gcc-m-cpu.cmake
```diff
@@ -52,13 +52,13 @@ if("${ARCH}" STREQUAL "arm")
     endif()
   elseif(CONFIG_CPU_CORTEX_R5)
     set(GCC_M_CPU cortex-r5)
-    if(CONFIG_FPU AND CONFIG_CPU_HAS_VFP)
-      if(NOT CONFIG_VFP_FEATURE_DOUBLE_PRECISION)
-        set(GCC_M_CPU ${GCC_M_CPU}+nofp.dp)
-      endif()
-    else()
+    # Fixed logic: Only add +nofp when FPU is explicitly disabled
+    if(NOT CONFIG_FPU OR NOT CONFIG_CPU_HAS_FPU)
```
tejlmand, line 56: `CONFIG_FPU` depends on `CPU_HAS_FPU`, so it can only be `y` if `CONFIG_CPU_HAS_FPU=y` else it will always be `n`, so should be sufficient to test for `if(NOT CONFIG_FPU)`

cmake/gcc-m-cpu.cmake
```diff
@@ -52,13 +52,13 @@ if("${ARCH}" STREQUAL "arm")
...
       set(GCC_M_CPU ${GCC_M_CPU}+nofp)
+    elseif(CONFIG_FPU AND CONFIG_CPU_HAS_FPU AND NOT CONFIG_VFP_FEATURE_DOUBLE_PRECISION)
```
tejlmand, line 58: `CONFIG_FPU` depends on `CPU_HAS_FPU`, so the `AND CONFIG_CPU_HAS_FPU` can be removed.
```suggestion
    elseif(CONFIG_FPU AND NOT CONFIG_VFP_FEATURE_DOUBLE_PRECISION)
```

cmake/gcc-m-cpu.cmake
```diff
@@ -52,13 +52,13 @@ if("${ARCH}" STREQUAL "arm")
     endif()
   elseif(CONFIG_CPU_CORTEX_R5)
     set(GCC_M_CPU cortex-r5)
-    if(CONFIG_FPU AND CONFIG_CPU_HAS_VFP)
-      if(NOT CONFIG_VFP_FEATURE_DOUBLE_PRECISION)
-        set(GCC_M_CPU ${GCC_M_CPU}+nofp.dp)
-      endif()
-    else()
+    # Fixed logic: Only add +nofp when FPU is explicitly disabled
```
tejlmand, line 55: How will `Fixed logic` comment help someone understanding the code when they look at this code with the reference point of this PR ?

Please rewrite to something which is also helpful in the future or remove it.

cmake/gcc-m-cpu.cmake
```diff
@@ -52,13 +52,13 @@ if("${ARCH}" STREQUAL "arm")
...
+    if(NOT CONFIG_FPU OR NOT CONFIG_CPU_HAS_FPU)
       set(GCC_M_CPU ${GCC_M_CPU}+nofp)
+    elseif(CONFIG_FPU AND CONFIG_CPU_HAS_FPU AND NOT CONFIG_VFP_FEATURE_DOUBLE_PRECISION)
+      set(GCC_M_CPU ${GCC_M_CPU}+nofp.dp)
     endif()
+    # For TDA4VM with full VFP: use plain cortex-r5 (no suffix)
```
tejlmand, line 61: There is no `else`, so this comment seems at wrong location.

tejlmand (changes requested):
Thanks for the fix, but please take a look at the code comments cause I don't think they will make sense to someone else looking at the code in n-years from now if they don't have this PR as a reference point.

PR #99478, opened 2025-11-16
arch: arm: Add missing MPU header include

pvinci:
Add missing ARM MPU header include when CONFIG_ARM_MPU is enabled. This fixes compilation errors when using MPU on ARM platforms.


wearyzen (changes requested):
could you add more details on what fails? IIRC there are MPU tests enabled in CI and running without any compilation error so is there a regression or a specific test that needs this change?

PR #99481, opened 2025-11-17
drivers: can: numaker: fix m55m1x core clock

ccli8:
On M55M1 series, the clock module index is virtual to stay compatible with 32-bit only cell value in devicetree. Its real value of being 64-bit integer needs to acquire via lookup into mapping table.

### DIFF
--- a/drivers/can/can_numaker.c
+++ b/drivers/can/can_numaker.c
@@ -16,6 +16,10 @@
 #include <soc.h>
 #include <NuMicro.h>
 
+#if defined(CONFIG_SOC_SERIES_M55M1X)
+#include <zephyr/dt-bindings/clock/numaker_m55m1x_clock.h>
+#endif
+
 LOG_MODULE_REGISTER(can_numaker, CONFIG_CAN_LOG_LEVEL);
 
 /* Implementation notes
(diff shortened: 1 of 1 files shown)


drivers/can/can_numaker.c
```diff
@@ -42,11 +46,27 @@ static int can_numaker_get_core_clock(const struct device *dev, uint32_t *rate)
...
 	uint32_t clkdiv_divider;
+	__typeof__(CANFD0_MODULE) clk_modidx_real;
+
+#if defined(CONFIG_SOC_SERIES_M55M1X)
+	switch (config->clk_modidx) {
+	case NUMAKER_CANFD0_MODULE:
+		clk_modidx_real = CANFD0_MODULE;
+		break;
+	case NUMAKER_CANFD1_MODULE:
+		clk_modidx_real = CANFD1_MODULE;
```
ssekar15, line 57: missing `break;` ? or else add `__fallthrough;`

ssekar15 (changes requested):

drivers/can/can_numaker.c
ccli8: `break` is missing, thanks

ssekar15 (approved):

henrikbrixandersen (approved):

str4t0m (approved):

cyliangtw (approved):

PR #99483, opened 2025-11-17
drivers: can: mcux: flexcan: move to per-instance message buffer configuration

### LINKED ISSUE #92798: can: can_mcux_flexcan.c allows a mailbox overrun resulting in an imprecise data bus fault

william-tang914:
This change refactors the FlexCAN driver to use per-instance message buffer configuration instead of global Kconfig options. The driver now calculates message buffer allocation at compile time based on device tree properties for each FlexCAN instance.
    
Key changes:
- Remove global CAN_MAX_MB Kconfig option and associated range constraint
- Add max_mb, rx_mb, and tx_mb fields to mcux_flexcan_config structure
- Move callback arrays and atomic allocation bitmaps from static global arrays to per-instance static arrays generated by macros
- Update data structure to use pointers to per-instance arrays instead of fixed-size embedded arrays
- Add compile-time validation to ensure sufficient message buffers are available for required RX filters
    
This approach allows different FlexCAN instances to have different message buffer configurations based on their hardware capabilities and device tree settings, improving memory efficiency and flexibility. The change maintains backward compatibility while enabling better resource utilization across different SoC variants.
    
Fixes #92798

### DIFF
--- a/boards/nxp/mr_canhubk3/doc/index.rst
+++ b/boards/nxp/mr_canhubk3/doc/index.rst
@@ -143,11 +143,8 @@ flexcan5         | PTC11  | PTC11_CAN0_RX  P22/P23
    and support maximum 32 message buffers for concurrent active instances with 8 bytes
    payload. We need to pay attention to configuration options:
 
-   1. :kconfig:option:`CONFIG_CAN_MCUX_FLEXCAN_MAX_MB` must be less or equal than the
+   1. :kconfig:option:`CONFIG_CAN_MCUX_FLEXCAN_MAX_FILTERS` must be less than the
       maximum number of message buffers that is according to the table below.
-
-   2. :kconfig:option:`CONFIG_CAN_MCUX_FLEXCAN_MAX_FILTERS` must be less or equal than
-      :kconfig:option:`CONFIG_CAN_MCUX_FLEXCAN_MAX_MB`.
 
 ===============  ==========  ================  ================
 Devicetree node  Payload     Hardware support  Software support
--- a/drivers/can/Kconfig.mcux
+++ b/drivers/can/Kconfig.mcux
@@ -28,29 +28,15 @@ config CAN_MCUX_FLEXCAN_WAIT_TIMEOUT
 	  Maximum number of wait loop iterations for the MCUX FlexCAN HAL when entering/leaving
 	  freeze mode.
 
-config CAN_MCUX_FLEXCAN_MAX_MB
-	int "Maximum number of message buffers for concurrent active instances"
-	default 16
-	depends on SOC_SERIES_S32K3 || SOC_SERIES_S32K1 || SOC_SERIES_S32ZE
-	range 1 96 if SOC_SERIES_S32K3
-	range 1 32 if SOC_SERIES_S32K1 && !SOC_S32K142W && !SOC_S32K144W
-	range 1 64 if SOC_S32K142W || SOC_S32K144W
-	range 1 128 if SOC_SERIES_S32ZE
-	help
-	  Defines maximum number of message buffers for concurrent active instances.
-
 config CAN_MCUX_FLEXCAN_MAX_FILTERS
 	int "Maximum number of concurrent active RX filters"
+	default 5 if CAN_MCUX_FLEXCAN_FD
 	default 13
-	range 1 15 if SOC_SERIES_KINETIS_KE1XF || SOC_SERIES_KINETIS_K6X
-	range 1 13 if (SOC_SERIES_IMXRT10XX || SOC_SERIES_IMXRT11XX) && CAN_MCUX_FLEXCAN_FD
-	range 1 63 if SOC_SERIES_IMXRT10XX || SOC_SERIES_IMXRT11XX
-	range 1 96 if SOC_SERIES_S32K3
-	range 1 32 if SOC_SERIES_S32K1 && !SOC_S32K142W && !SOC_S32K144W
-	range 1 64 if SOC_S32K142W || SOC_S32K144W
-	range 1 128 if SOC_SERIES_S32ZE
 	help
 	  Defines maximum number of concurrent active RX filters
+	  The maximum value depends on the number of available message
+	  buffers for each FlexCAN instance. CAN FD mode uses larger
+	  message buffers, resulting in fewer available filters.
 
 endif # CAN_MCUX_FLEXCAN
 
--- a/drivers/can/can_mcux_flexcan.c
+++ b/drivers/can/can_mcux_flexcan.c
@@ -33,20 +33,6 @@ LOG_MODULE_REGISTER(can_mcux_flexcan, CONFIG_CAN_LOG_LEVEL);
 #define RX_START_IDX 0
 #endif
 
-/* The maximum number of message buffers for concurrent active instances */
-#ifdef CONFIG_CAN_MCUX_FLEXCAN_MAX_MB
-#define MCUX_FLEXCAN_MAX_MB CONFIG_CAN_MCUX_FLEXCAN_MAX_MB
-#else
-#define MCUX_FLEXCAN_MAX_MB FSL_FEATURE_FLEXCAN_HAS_MESSAGE_BUFFER_MAX_NUMBERn(0)
-#endif
-
-/*
- * RX message buffers (filters) will take up the first N message
- * buffers. The rest are available for TX use.
- */
-#define MCUX_FLEXCAN_MAX_RX (CONFIG_CAN_MCUX_FLEXCAN_MAX_FILTERS + RX_START_IDX)
-#define MCUX_FLEXCAN_MAX_TX (MCUX_FLEXCAN_MAX_MB - MCUX_FLEXCAN_MAX_RX)
-
 /*
  * Convert from RX message buffer index to allocated filter ID and
  * vice versa.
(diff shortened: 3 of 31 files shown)


drivers/can/can_mcux_flexcan.c
```diff
@@ -1432,20 +1425,54 @@ static DEVICE_API(can, mcux_flexcan_fd_driver_api) = {
...
 									\
 	static void mcux_flexcan_irq_config_##id(const struct device *dev); \
 	static void mcux_flexcan_irq_enable_##id(void); \
 	static void mcux_flexcan_irq_disable_##id(void); \
 									\
+	static struct mcux_flexcan_rx_callback flexcan_rx_cbs_##id	\
+			[FLEXCAN_INST_RX_MB(id)] = {0};			\
+									\
+	static struct mcux_flexcan_tx_callback flexcan_tx_cbs_##id	\
+			[FLEXCAN_INST_TX_MB(id)] = {0};			\
```
zejiang0jason, line 1512: just query, do we need to consider the case that `FLEXCAN_INST_RX_MB(id)` or `FLEXCAN_INST_TX_MB(id)` is 0?

william-tang914: In previous, Kconfig symbol CAN_MAX_FILTER is confined to be greater than 0, but there was no constraint that MCUX_FLEXCAN_MAX_TX not be 0.
If MCUX_FLEXCAN_MAX_TX is 0, compiler will report following warning 
```
zephyrproject/zephyr/drivers/can/can_mcux_flexcan.c:1041:32: warning: array subscript 'alloc' is outside the bounds of an interior zero-length array 'struct mcux_flexcan_tx_callback[0]' [-Wzero-length-bounds]
 1041 |         function = data->tx_cbs[alloc].function;
```
For safety, I think it would be better to confine `FLEXCAN_INST_TX_MB(id)` and `FLEXCAN_INST_RX_MB(id)` greater than 0.

zejiang0jason (dismissed):

boards/nxp/mr_canhubk3/doc/index.rst
```diff
@@ -143,12 +143,9 @@ flexcan5         | PTC11  | PTC11_CAN0_RX  P22/P23
    and support maximum 32 message buffers for concurrent active instances with 8 bytes
    payload. We need to pay attention to configuration options:
 
-   1. :kconfig:option:`CONFIG_CAN_MAX_MB` must be less or equal than the
+   1. :kconfig:option:`CONFIG_CAN_MAX_FILTER` must be less or equal than the
```
henrikbrixandersen: `CONFIG_CAN_MAX_FILTER` no longer exists.

dts/bindings/can/nxp,flexcan.yaml
```diff
@@ -39,3 +40,8 @@ properties:
     type: int
     required: true
     description: CAN engine clock source
+
+  nxp,max-mb:
```
henrikbrixandersen: There's no need to prefix this with a vendor prefix ("nxp") as the binding is specific to that vendor.

I suggest naming this property `number-of-mb` to match the IP options listed on https://www.nxp.com/products/nxp-product-information/ip-block-licensing/flexcan-controller:FLEXCAN-CONTROLLER - and since it conveys its purpose better (the IP instance has a fixed number of MBs, which is what we are trying to describe here).

dts/bindings/can/nxp,flexcan.yaml
```diff
@@ -39,3 +40,8 @@ properties:
...
+    type: int
+    required: true
+    description: Maximum number of 8-byte payload message buffers FlexCAN instance supported
```
henrikbrixandersen: Please adjust the description accordingly.

dts/bindings/can/nxp,flexcan-fd.yaml
```diff
@@ -27,3 +29,9 @@ compatible: "nxp,flexcan-fd"
 
 include: ["nxp,flexcan.yaml", "can-fd-controller.yaml", "pinctrl-device.yaml",
           "nxp,rdc-policy.yaml"]
+
+properties:
+  nxp,max-mb-fd:
+    type: int
+    required: true
+    description: Maximum number of 64-byte payload message buffers FlexCAN FD instance supported
```
henrikbrixandersen: Same applies here.

drivers/can/Kconfig.mcux
```diff
@@ -28,27 +28,9 @@ config CAN_MCUX_FLEXCAN_WAIT_TIMEOUT
...
-	default 16
-	depends on SOC_SERIES_S32K3 || SOC_SERIES_S32K1 || SOC_SERIES_S32ZE
-	range 1 96 if SOC_SERIES_S32K3
-	range 1 32 if SOC_SERIES_S32K1 && !SOC_S32K142W && !SOC_S32K144W
-	range 1 64 if SOC_S32K142W || SOC_S32K144W
-	range 1 128 if SOC_SERIES_S32ZE
-	help
-	  Defines maximum number of message buffers for concurrent active instances.
-
 config CAN_MAX_FILTER
```
henrikbrixandersen: This needs rebasing.

drivers/can/can_mcux_flexcan.c
```diff
@@ -1478,20 +1425,56 @@ static DEVICE_API(can, mcux_flexcan_fd_driver_api) = {
...
+	static ATOMIC_DEFINE(flexcan_tx_allocs_##id, FLEXCAN_INST_TX_MB(id));	\
+									\
 	static const struct mcux_flexcan_config mcux_flexcan_config_##id = { \
 		DEVICE_MMIO_NAMED_ROM_INIT(flexcan_mmio, DT_DRV_INST(id)),	\
 		.common = CAN_DT_DRIVER_CONFIG_INST_GET(id, 0, FLEXCAN_MAX_BITRATE(id)), \
 		.clock_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR(id)),	\
 		.clock_subsys = (clock_control_subsys_t)		\
 			DT_INST_CLOCKS_CELL(id, name),			\
 		.clk_source = DT_INST_PROP(id, clk_source),		\
+		.max_mb = FLEXCAN_INST_MAX_MB(id),			\
```
henrikbrixandersen: Please reflect the updated DT naming in the C code as well.

henrikbrixandersen (changes requested):

dts/bindings/can/nxp,flexcan.yaml
william-tang914: Thank you for the suggestion. I didn’t even know such a webpage existed.

henrikbrixandersen (approved):

william-tang914:
I reproduced original issue https://github.com/zephyrproject-rtos/zephyr/issues/92798 by connect only one node to CAN bus. Now code code will be blocked by TX semaphore if there is no TX MB. So I think original issue shall be fixed.

https://github.com/nxp-upstream/zephyr/blob/8936a0bbcf3b6bad2d1ca8d102c068bad4f56642/drivers/can/can_mcux_flexcan.c#L747-L749

manuargue (approved):

PR #99489, opened 2025-11-17
boards: beagle: beagleconnect_freedom: Add rev C5 support

Ayush1325:
There seem to be a decent number of C5 boards around. So add support for it while keeping C7 as the default version.

The main difference between C7 and C5 is that C5 uses CC1352P instead of CC1352P7


boards/beagle/beagleconnect_freedom/revision.cmake
```diff
@@ -0,0 +1,20 @@
...
+
+if("${BOARD_QUALIFIERS}" MATCHES "/cc1352p7$" AND "${BOARD_REVISION}" STREQUAL "C5")
+  set(ACTIVE_BOARD_REVISION "C7")
+  message(WARNING "C5 only supports CC1352P. Using C7 instead.")
+endif()
+
+if("${BOARD_QUALIFIERS}" MATCHES "/cc1352p$" AND "${BOARD_REVISION}" STREQUAL "C7")
+  set(ACTIVE_BOARD_REVISION "C5")
+  message(WARNING "C7 only supports CC1352P7. Using C5 instead.")
+endif()
```
nordicjm: this should check from the first character until either the end of the line **or** until a `/`, this code prevents users extending a board out of tree, and it should also not be changing the revision, if there is no revision provided then there can be a default set but do not override what the user has put, throw an error

nordicjm (changes requested):

boards/beagle/beagleconnect_freedom/revision.cmake
Ayush1325: So, throwing the error breaks CI since there was no way I could find to tie an soc to a specific revision.
It was discussed here: https://github.com/zephyrproject-rtos/zephyr/pull/95065

nordicjm: It uses the twister files you have, if there is no .yaml for a board target then twister will not try to use it, and you need to have the revisions in the .yaml files for the twister board name. You should probably not set a default revision anyway since it isn't valid for one of the SoCs, people should provide the correct revision

Ayush1325: > It uses the twister files you have, if there is no .yaml for a board target then twister will not try to use it, and you need to have the revisions in the .yaml files for the twister board name. You should probably not set a default revision anyway since it isn't valid for one of the SoCs, people should provide the correct revision

Do you mean this file `boards/beagle/beagleconnect_freedom/beagleconnect_freedom_cc1352p7.yaml`? It can have a revisions field?

nordicjm: Yes, rename and change like so: https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/nordic/nrf9160dk/nrf9160dk_nrf9160_0_7_0.yaml

Ayush1325: @nordicjm I think I might be missing something. I cannot seem to use the names with the revision in `platform_allow` or `exclude` lists.

nordicjm: you need to have the twister .yaml files, it uses the names exactly as they are from those

Ayush1325: @nordicjm Can you check the PR now. It seems like I can add variants to `board.yml`.
Also, you also mentioned the following:

> this code prevents users extending a board out of tree

Do you mean I should remove the check that the provided revision is either C5 or C7? Or is that fine?

boards/beagle/beagleconnect_freedom/board.yml
```diff
@@ -3,4 +3,17 @@ board:
...
+    exact: true
+    default: "C7"
+    revisions:
+      - name: "C5"
+      - name: "C7"
+  variants:
+    - name: beagleconnect_freedom@C5
+      qualifier: "cc1352p"
+    - name: beagleconnect_freedom@C7
+      qualifier: "cc1352p7"
```
nordicjm: err this is not what variants are used for no
```suggestion
```

boards/beagle/beagleconnect_freedom/board.yml
```diff
@@ -3,4 +3,17 @@ board:
   full_name: BeagleConnect Freedom
   vendor: beagle
   socs:
-  - name: cc1352p7
+    - name: cc1352p
+    - name: cc1352p7
+  revision:
+    format: custom
+    exact: true
+    default: "C7"
```
nordicjm:
```suggestion
```

boards/beagle/beagleconnect_freedom/revision.cmake
```diff
@@ -0,0 +1,18 @@
+set(BOARD_REVISIONS "C5" "C7")
+
+# If BOARD_REVISION not set, use the default revision
+if(NOT DEFINED BOARD_REVISION)
+  set(BOARD_REVISION ${LIST_BOARD_REVISION_DEFAULT})
+endif()
+
```
nordicjm:
```suggestion
```

boards/beagle/beagleconnect_freedom/revision.cmake
```diff
@@ -0,0 +1,18 @@
...
+if(NOT BOARD_REVISION IN_LIST BOARD_REVISIONS)
+  message(FATAL_ERROR "${BOARD_REVISION} is not a valid revision for BeagleConnect Freedom. Accepted revisions: ${BOARD_REVISIONS}")
+endif()
+
+if("${BOARD_QUALIFIERS}" MATCHES "^/cc1352p7$" AND "${BOARD_REVISION}" STREQUAL "C5")
+  message(FATAL_ERROR "C5 only supports CC1352P. Using C7 instead.")
```
nordicjm: *use, as for below

boards/beagle/beagleconnect_freedom/board.yml
Ayush1325: Well, I have tried adding a `twister.yaml` file. However, twister does still try building for wrong variants. So not sure what I am doing incorrectly.

nordicjm: No not a `twister.yaml` file, twister yaml files, so `beagleconnect_freedom_cc1352p_C5.yaml` with the specifiers and details for that then `beagleconnect_freedom_cc1352p7_C7` with the specifiers and details for that one

Ayush1325: According to twister internals, `*.yaml` other than `twister.yaml` are categorized as legecy files: https://github.com/zephyrproject-rtos/zephyr/blob/f1c2ce4f4026bd5b212c6e4cc7d754aba69b189e/scripts/pylib/twister/twisterlib/platform.py#L224

I am pretty sure I tried what you are mentioning here, and it did not work, but let me push that again.

nordicjm: @PerMac if you cannot fit the behaviour here into `twister.yaml` files then please remove the comment about multiple twister files being legacy and work towards that because `twister.yaml` files would then absolutely not fit what is required.
@Ayush1325 use multiple twister files

Ayush1325: Ok, I stand corrected. It works. Will create a PR to do the same with PocketBeagle 2 revisions.

boards/beagle/beagleconnect_freedom/board.yml
Ayush1325: Done

boards/beagle/beagleconnect_freedom/revision.cmake
Ayush1325: Done.

boards/beagle/beagleconnect_freedom/revision.cmake
Ayush1325: Done

nordicjm (dismissed):

jadonk:
I don't see any updates to the documentation.

Is there any change in building? Can I still do `west build -b beagleconnect_freedom .` and get a C7 build?

Are you sure the terminology is "CC1352P" and not "CC1352P1"? I know ti.com seems to have it under "cc1352p", but that seems to be before they introduced the others.

Looking around for a family overview largely comes up empty, but I did find a [hardware architecture section](https://software-dl.ti.com/simplelink/esd/simplelink_cc13xx_cc26xx_sdk/7.10.01.24/exports/docs/thread/html/cc13xx_cc26xx/hardware-architecture.html) in the SDK docs:

*    For CC13x1 and CC26x1 devices the system core is ARM Cortex M4
*    For CC13x2 and CC26x2 devices the system core is ARM Cortex M4F
*    For CC13x4 and CC26x4 devices the system core is ARM Cortex M33

...

Device | Flash (kB) | SRAM (kB)
-- | -- | --
CC13x1x3 or CC26x1x3 | 352 | 40
CC13x2x1 or CC26x2x1 | 352 | 80
CC13x2x7 or CC26x2x7 | 704 | 144
CC13x4x10 or CC26x4x10 | 1024 | 256

I don't see in either of those sections any use of the CC1352P without the CC1352P1.

Ayush1325:
> Are you sure the terminology is "CC1352P" and not "CC1352P1"? I know ti.com seems to have it under "cc1352p", but that seems to be before they introduced the others.

Well, TI certainly seems to be only using CC1352P and CC1352P7 in most of the public docs. Additionally, zephyr already has the soc added as CC1352P (and CC1352R): https://github.com/zephyrproject-rtos/zephyr/tree/main/soc/ti/simplelink. So any name change now would involve tree-wide changes to drivers launchxl board.
 
> Looking around for a family overview largely comes up empty, but I did find a [hardware architecture section](https://software-dl.ti.com/simplelink/esd/simplelink_cc13xx_cc26xx_sdk/7.10.01.24/exports/docs/thread/html/cc13xx_cc26xx/hardware-architecture.html) in the SDK docs:
> 
>     * For CC13x1 and CC26x1 devices the system core is ARM Cortex M4
> 
>     * For CC13x2 and CC26x2 devices the system core is ARM Cortex M4F
> 
>     * For CC13x4 and CC26x4 devices the system core is ARM Cortex M33
> 
> 
> ...
> Device 	Flash (kB) 	SRAM (kB)
> CC13x1x3 or CC26x1x3 	352 	40
> CC13x2x1 or CC26x2x1 	352 	80
> CC13x2x7 or CC26x2x7 	704 	144
> CC13x4x10 or CC26x4x10 	1024 	256
> 
> I don't see in either of those sections any use of the CC1352P without the CC1352P1.

The product page listss only CC1352P: https://www.ti.com/product/CC1352P. Also in alternatives, it lists CC1352P7, CC1352R (not R1) and CC1312R.

Ayush1325:
> I don't see any updates to the documentation.
> 
> Is there any change in building? Can I still do `west build -b beagleconnect_freedom .` and get a C7 build?

No, you cannot do `west build -b beagleconnect_freedom .`. The full board name is required now. So `beagleconnect_freedom@C7/cc1352p7`.
With that said, the docs currently do not have any build instructions: https://docs.zephyrproject.org/latest/boards/beagle/beagleconnect_freedom/doc/index.html
I can add some details regarding the C1 rev in the docs in this PR if required. However, I think it would be better to do a major PR to update and improve docs separately, since currently the docs are pretty much empty.

nordicjm (dismissed):

kartben (dismissed):

boards/beagle/beagleconnect_freedom/doc/index.rst
```diff
@@ -20,6 +20,7 @@ The CC1352P7 wireless MCU has a 48 MHz Arm Cortex-M4F SoC and a Bluetooth Low En
 The board also features a TI MSP430F5503 microcontroller used as a USB-to-serial bridge and
 GPIO expander.
 
+(NOTE: Rev C5 used TI CC1352P MCU and is no longer available. Rev C7 uses TI CC1352P7.)
```
jadonk: This is a bit misleading. C5 was only used for pre-production boards. If you bought a BeagleConnect Freedom, it would be revision C7 or later. It isn't just that C5 is no longer available--it was never made available for sale.

boards/beagle/beagleconnect_freedom/beagleconnect_freedom_cc1352p7_C7.yaml
```diff
@@ -1,5 +1,5 @@
-identifier: beagleconnect_freedom
-name: BeagleConnect Freedom
+identifier: beagleconnect_freedom@C7/cc1352p7
+name: BeagleConnect Freedom C7
```
jadonk, line 2: Do boards typically have their revision in their name?

boards/beagle/beagleconnect_freedom/board.yml
```diff
@@ -3,4 +3,12 @@ board:
...
```
jadonk: I'm looking at this in the summary of changes, so we are good with including a default of "C7"? If so, I'll approve. If it breaks building via `west build -b beagleconnect_freedom`, I'm going to reject. I want to have "C5" support in mainline, but it isn't worth breaking default builds.

jadonk (commented):
I just want to make sure builds of `west build -b beagleconnect_freedom .` continue to work for rev C7, then I'll approve.

NOTE: the defconfig assignments are based on SoC, not board revision, but the capacitor array delta is a board-level change, not based on the SoC. This is probably fine, but could cause some confusion for someone copying the design without understanding the way the two board revisions are populated. C5 has capacitors around the 48MHz crystal and C7 does not.

boards/beagle/beagleconnect_freedom/board.yml
Ayush1325: > I'm looking at this in the summary of changes, so we are good with including a default of "C7"? If so, I'll approve. If it breaks building via `west build -b beagleconnect_freedom`, I'm going to reject. I want to have "C5" support in mainline, but it isn't worth breaking default builds.

Well, I can do some magic in `revision.cmake` to make just beagleconnect_freedom work, but not sure if that would be ok to upstream. @nordicjm Any thoughts?

@jadonk Another thing to note is that if the bcf msp430 support is ever added to zephyr, that will also break the current `beagleconnect_freedom` board name. Multi-soc boards require the soc qualifier.

boards/beagle/beagleconnect_freedom/beagleconnect_freedom_cc1352p7_C7.yaml
Ayush1325: They should if the revision signifies a soc change. After all c5 with cc1352p7 or c7 with cc1352p does not make sense.

boards/beagle/beagleconnect_freedom/board.yml
nordicjm: This is not going to work, it is clearly listed on the hwmv2 page if you have multiple SoCs you **must** provide the SoC, so no, board name required, SoC name required, revision required

boards/beagle/beagleconnect_freedom/board.yml
nordicjm: this is still here

boards/beagle/beagleconnect_freedom/revision.cmake
```diff
@@ -1,5 +1,9 @@
 set(BOARD_REVISIONS "C5" "C7")
 
+if(NOT DEFINED BOARD_REVISION)
+  set(BOARD_REVISION ${LIST_BOARD_REVISION_DEFAULT})
+endif()
+
```
nordicjm:
```suggestion
```

nordicjm (changes requested):

boards/beagle/beagleconnect_freedom/revision.cmake
Ayush1325: I would at least like to have support for `beagleconnect_freedom/cc1352p5` Requiring `beagleconnect_freedom@C7/cc1352p7` for board name seems rather verbose specially when that is supposed to be the only revision available at retail now.

nordicjm (changes requested):
Not sure why my review was re-requested when the issues are still here

nordicjm (dismissed):

boards/beagle/beagleconnect_freedom/doc/index.rst
jadonk: Why not fix this comment?

jadonk (changes requested):
I hate that this breaks the default board name, but I guess that is the way it is. Seems adding the MSP430 will also force more specification for builds. I'm OK with it, if you fix the comment to actually make it clear C5 was a pre-production revision, I'll approve.

boards/beagle/beagleconnect_freedom/board.yml
```diff
@@ -3,4 +3,11 @@ board:
...
   socs:
-  - name: cc1352p7
+    - name: cc1352p
+    - name: cc1352p7
+  revision:
+    format: custom
+    exact: true
+    revisions:
+      - name: "C5"
+      - name: "C7"
```
nordicjm, line 13:
```suggestion
```

Ayush1325: Removing this causes unknown platform errors:
```
 ERROR   - platform_allow in sample.net.sockets.echo_client.802154.subg - unrecognized platform - beagleconnect_freedom@C5/cc1352p
```

https://github.com/zephyrproject-rtos/zephyr/actions/runs/25036288977/job/73328678331?pr=99489

jadonk (dismissed):

jadonk (approved):

boards/beagle/beagleconnect_freedom/board.yml
Ayush1325: cc @nordicjm

glneo (approved):

PR #99499, opened 2025-11-17
Define number of platform socket services using Kconfig

Cristib05:
Provide support to define number of socket services using Kconfig.


modules/openthread/Kconfig
```diff
@@ -394,13 +394,38 @@ config OPENTHREAD_ZEPHYR_BORDER_ROUTER_MODEL_NAME
...
+	  This configuration defines the maximum number of pollable mDNS sockets
+	  to be used within platform mDNS module.
+
+config OPENTHREAD_ZEPHYR_BORDER_ROUTER_TREL_SERVICES
+	int "The maximum number of pollable TREL sockets to be registered in platform code."
+	depends on OPENTHREAD_ZEPHYR_BORDER_ROUTER
+	default 1
+	help
+	  This configuration defines the maximum number of pollable mDNS sockets
+	  to be used within platform mDNS module.
```
jukkar: Is this help text correct, it looks identical as the mDNS one?

modules/openthread/platform/dhcp6_pd.c
```diff
@@ -17,9 +17,8 @@
 
 #define DHCPV6_SERVER_PORT 547
 #define DHCPV6_CLIENT_PORT 546
-#define DHCPV6_PD_CLIENT_NUM_SERVICES 1
```
jukkar, line 20: Why not keep the define, that way there are smaller number of changes, the code is easier to read (because the config option is quite long) and the lines below will be shorter too so something like
```
#define DHCPV6_PD_CLIENT_NUM_SERVICES CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_DHCP6_PD_SERVICES
```
same comment to other changes.

Cristib05: Have I misunderstood your proposal from NAT64 PR? https://github.com/zephyrproject-rtos/zephyr/pull/99314#discussion_r2526192700

jukkar: No, your proposal is good, I was just commenting the implementation. For example you already have
```
#define MAX_SERVICES   1
```
why not have here instead
```
#define MAX_SERVICES CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_MAX_MDNS_SERVICES
```

Cristib05: oh, ok, I think I've understood now. Thank you!

modules/openthread/Kconfig
```diff
@@ -394,13 +394,38 @@ config OPENTHREAD_ZEPHYR_BORDER_ROUTER_MODEL_NAME
 	  This value represents Border Agent's product model.
 
 config OPENTHREAD_ZEPHYR_BORDER_ROUTER_MAX_UDP_SERVICES
-	int "The maximum number of pollable UDP sockets to be registered in platform UDP."
+	int "The maximum number of pollable UDP sockets to be registered in platform code."
 	depends on OPENTHREAD_ZEPHYR_BORDER_ROUTER
```
rlubos: This config is already behind `if OPENTHREAD_ZEPHYR_BORDER_ROUTER` so `depends on OPENTHREAD_ZEPHYR_BORDER_ROUTER` is redundant, could be removed. Same for other configs below.

rlubos (approved):

pdgendt (approved):

jukkar (approved):

Issue #95500, opened 2025-09-05
dwc2 usb txff assigned to EP using txff_num


jean12332:
### Is your feature request related to a problem? Please describe.

When I tesetd zephyr/samples/subsys/usb/mass EP1 ep_enable fail  in udc_dwc.c beacause of TXFF.  Due to our IC design limitations,  EP1 (mps 512B) fixed use of TXFF1(depth is 16*4B) in udc_dwc.c,  but our TXFF1 is not large enough.  So we need to choose other TXFFs, such as TXFF2. According to the DWC2 databook, this way is allowed and more flexible to use. Moreover, not all EPs are bidirectional. For example, EP5 is out-only  and EP6 is in-only, when EP6 is enable in sequence, TXFF5 can be used. 

### Describe the solution you'd like

more flexible way of txff config

### Describe alternatives you've considered

When the TXFFn.depth assigned to EPn is not met, search for other available TXFFs in order and maintain the corresponding relationship by the new txffnum variable associated with DIEPCTLn.txfnum filed.

PR #95723, opened 2025-09-09
uhc: dwc2: vendor quirks for nRF54LM20

josuah:
- https://github.com/zephyrproject-rtos/zephyr/issues/102931

Dependencies:

- https://github.com/zephyrproject-rtos/zephyr/pull/102967

For testing, provide power externally using an USB OTG adapter with external power supply.

PR #94766, opened 2025-08-20
drivers: udc: add helper to estimate FIFO and largest endpoint MPS

jfischer-no:
To configure FIFOs properly, UDC controllers with internal endpoint FIFO
memory may require information such as the largest endpoint MPS, the
number of endpoints, and the estimated RX and TX FIFO sizes.

PR #99510, opened 2025-11-17
dts: bindings: usb: Specify fifo sizes in devicetree

tmon-nordic:
Add g-rx-fifo-size, g-np-tx-fifo-size and g-tx-fifo-size required properties to snps,dwc2 compatible. Property names are the same as used in Linux because Zephyr aims for devicetree source compatibility with other operating systems.

Specifying fifo sizes in devicetree greatly reduces necessary driver complexity and allows application developer to adjust the sizes if necessary to best utilize underlying hardware.

Implements #96206

### DIFF
--- a/drivers/usb/udc/udc_dwc2.c
+++ b/drivers/usb/udc/udc_dwc2.c
@@ -47,30 +47,8 @@ enum dwc2_drv_event_type {
 	DWC2_DRV_EVT_DISABLE,
 };
 
-/* Minimum RX FIFO size in 32-bit words considering the largest used OUT packet
- * of 512 bytes. The value must be adjusted according to the number of OUT
- * endpoints.
- */
-#define UDC_DWC2_GRXFSIZ_FS_DEFAULT	(15U + 512U/4U)
-/* Default Rx FIFO size in 32-bit words calculated to support High-Speed with:
- *   * 1 control endpoint in Completer/Buffer DMA mode: 13 locations
- *   * Global OUT NAK: 1 location
- *   * Space for 3 * 1024 packets: ((1024/4) + 1) * 3 = 774 locations
- * Driver adds 2 locations for each OUT endpoint to this value.
- */
-#define UDC_DWC2_GRXFSIZ_HS_DEFAULT	(13 + 1 + 774)
-
-/* TX FIFO0 depth in 32-bit words (used by control IN endpoint)
- * Try 2 * bMaxPacketSize0 to allow simultaneous operation with a fallback to
- * whatever is available when 2 * bMaxPacketSize0 is not possible.
- */
-#define UDC_DWC2_FIFO0_DEPTH		(2 * 16U)
-
 /* Get Data FIFO access register */
 #define UDC_DWC2_EP_FIFO(base, idx)	((mem_addr_t)base + 0x1000 * (idx + 1))
-
-/* Percentage limit of how much SPRAM can be allocated for RxFIFO */
-#define MAX_RXFIFO_GDFIFO_PERCENTAGE 25
 
 static void udc_dwc2_ep_disable(const struct device *dev,
 				struct udc_ep_config *const cfg,
@@ -239,41 +217,6 @@ static void dwc2_flush_tx_fifo(const struct device *dev, const uint8_t fnum)
 	sys_write32(grstctl, grstctl_reg);
 	while (sys_read32(grstctl_reg) & USB_DWC2_GRSTCTL_TXFFLSH) {
 	}
-}
-
-/* Return TX FIFOi depth in 32-bit words (i = f_idx + 1) */
-static uint32_t dwc2_get_txfdep(const struct device *dev, const uint32_t f_idx)
-{
-	struct usb_dwc2_reg *const base = dwc2_get_base(dev);
-	uint32_t dieptxf;
-
-	dieptxf = sys_read32((mem_addr_t)&base->dieptxf[f_idx]);
-
-	return usb_dwc2_get_dieptxf_inepntxfdep(dieptxf);
-}
-
-/* Return TX FIFOi address (i = f_idx + 1) */
-static uint32_t dwc2_get_txfaddr(const struct device *dev, const uint32_t f_idx)
-{
-	struct usb_dwc2_reg *const base = dwc2_get_base(dev);
-	uint32_t dieptxf;
-
-	dieptxf = sys_read32((mem_addr_t)&base->dieptxf[f_idx]);
-
-	return  usb_dwc2_get_dieptxf_inepntxfstaddr(dieptxf);
-}
-
-/* Set TX FIFOi address and depth (i = f_idx + 1) */
-static void dwc2_set_txf(const struct device *dev, const uint32_t f_idx,
-			 const uint32_t dep, const uint32_t addr)
-{
-	struct usb_dwc2_reg *const base = dwc2_get_base(dev);
-	uint32_t dieptxf;
-
-	dieptxf = usb_dwc2_set_dieptxf_inepntxfdep(dep) |
-		  usb_dwc2_set_dieptxf_inepntxfstaddr(addr);
-
-	sys_write32(dieptxf, (mem_addr_t)&base->dieptxf[f_idx]);
 }
 
 /* Enable/disable endpoint interrupt */
@@ -1074,24 +1017,6 @@ static void request_hibernation(struct udc_dwc2_data *const priv)
 	}
 }
 
-static void dwc2_unset_unused_fifo(const struct device *dev)
-{
-	struct udc_dwc2_data *const priv = udc_get_private(dev);
-	struct udc_ep_config *tmp;
-
-	for (uint8_t i = priv->ineps - 1U; i > 0; i--) {
-		tmp = udc_get_ep_cfg(dev, i | USB_EP_DIR_IN);
-
-		if (tmp->stat.enabled && (priv->txf_set & BIT(i))) {
-			return;
-		}
-
-		if (!tmp->stat.enabled && (priv->txf_set & BIT(i))) {
-			priv->txf_set &= ~BIT(i);
-		}
-	}
-}
-
 /*
  * In dedicated FIFO mode there are i (i = 1 ... ineps - 1) FIFO size registers,
  * e.g. DIEPTXF1, DIEPTXF2, ... DIEPTXF4. When dynfifosizing is enabled,
@@ -1101,16 +1026,11 @@ static int dwc2_set_dedicated_fifo(const struct device *dev,
 				   struct udc_ep_config *const cfg,
 				   uint32_t *const diepctl)
 {
-	struct udc_dwc2_data *const priv = udc_get_private(dev);
-	uint8_t ep_idx = USB_EP_GET_IDX(cfg->addr);
+	struct usb_dwc2_reg *const base = dwc2_get_base(dev);
 	const uint32_t addnl = USB_MPS_ADDITIONAL_TRANSACTIONS(cfg->mps);
 	uint32_t reqdep;
-	uint32_t txfaddr;
-	uint32_t txfdep;
-	uint32_t tmp;
-
-	/* Keep everything but FIFO number */
-	tmp = *diepctl & ~USB_DWC2_DEPCTL_TXFNUM_MASK;
+	uint16_t start, depth;
+	uint8_t txfnum;
 
 	reqdep = DIV_ROUND_UP(udc_mps_ep_size(cfg), 4U);
 	if (dwc2_in_buffer_dma_mode(dev)) {
@@ -1120,63 +1040,29 @@ static int dwc2_set_dedicated_fifo(const struct device *dev,
 		reqdep *= (1 + addnl);
 	}
 
-	if (priv->dynfifosizing) {
-		if (priv->txf_set & ~BIT_MASK(ep_idx)) {
-			dwc2_unset_unused_fifo(dev);
-		}
+	/* TxFIFOs are preassigned, just verify it can be used */
+	txfnum = usb_dwc2_get_depctl_txfnum(*diepctl);
+	if (txfnum == 0) {
+		uint32_t gnptxfsiz = sys_read32((mem_addr_t)&base->gnptxfsiz);
 
-		if ((ep_idx - 1) != 0U) {
-			txfaddr = dwc2_get_txfdep(dev, ep_idx - 2) +
-				  dwc2_get_txfaddr(dev, ep_idx - 2);
-		} else {
-			txfaddr = priv->rxfifo_depth +
-				MIN(UDC_DWC2_FIFO0_DEPTH, priv->max_txfifo_depth[0]);
-		}
-
-		if (priv->txf_set & BIT(ep_idx)) {
-			uint32_t curaddr;
-
-			curaddr = dwc2_get_txfaddr(dev, ep_idx - 1);
-			txfdep = dwc2_get_txfdep(dev, ep_idx - 1);
-			if (txfaddr != curaddr || reqdep > txfdep) {
-				LOG_ERR("FIFO%u cannot be reused, new addr 0x%04x depth %u",
-					ep_idx, txfaddr, reqdep);
-				return -ENOMEM;
-			}
-		} else {
-			txfdep = reqdep;
-		}
-
-		/* Make sure to not set TxFIFO greater than hardware allows */
-		if (txfdep > priv->max_txfifo_depth[ep_idx]) {
-			return -ENOMEM;
-		}
-
-		/* Do not allocate TxFIFO outside the SPRAM */
-		if (txfaddr + txfdep > priv->dfifodepth) {
-			return -ENOMEM;
-		}
-
-		/* Set FIFO depth (32-bit words) and address */
-		dwc2_set_txf(dev, ep_idx - 1, txfdep, txfaddr);
+		start = usb_dwc2_get_gnptxfsiz_nptxfstaddr(gnptxfsiz);
+		depth = usb_dwc2_get_gnptxfsiz_nptxfdep(gnptxfsiz);
 	} else {
-		txfdep = dwc2_get_txfdep(dev, ep_idx - 1);
-		txfaddr = dwc2_get_txfaddr(dev, ep_idx - 1);
+		uint32_t dieptxf = sys_read32((mem_addr_t)&base->dieptxf[txfnum - 1U]);
 
-		if (reqdep > txfdep) {
-			return -ENOMEM;
-		}
-
-		LOG_DBG("Reuse FIFO%u addr 0x%08x depth %u", ep_idx, txfaddr, txfdep);
+		start = usb_dwc2_get_dieptxf_inepntxfstaddr(dieptxf);
+		depth = usb_dwc2_get_dieptxf_inepntxfdep(dieptxf);
 	}
 
-	/* Assign FIFO to the IN endpoint */
-	*diepctl = tmp | usb_dwc2_set_depctl_txfnum(ep_idx);
-	priv->txf_set |= BIT(ep_idx);
-	dwc2_flush_tx_fifo(dev, ep_idx);
+	if (reqdep > depth) {
+		return -ENOMEM;
+	}
 
-	LOG_INF("Set FIFO%u (ep 0x%02x) addr 0x%04x depth %u size %u",
-		ep_idx, cfg->addr, txfaddr, txfdep, dwc2_ftx_avail(dev, ep_idx));
+	dwc2_flush_tx_fifo(dev, txfnum);
+
+	LOG_INF("Set FIFO%u (ep 0x%02x) addr 0x%04x depth %u avail %u",
+		txfnum, cfg->addr, start, depth,
+		dwc2_ftx_avail(dev, USB_EP_GET_IDX(cfg->addr)));
 
 	return 0;
 }
@@ -1300,33 +1186,6 @@ static int udc_dwc2_ep_activate(const struct device *dev,
 		LOG_DBG("DIEPTXF%u %08x DIEPCTL%u %08x",
 			i, sys_read32((mem_addr_t)&base->dieptxf[i - 1U]), i, dxepctl);
 	}
-
-	return 0;
-}
-
-static int dwc2_unset_dedicated_fifo(const struct device *dev,
-				     struct udc_ep_config *const cfg,
-				     uint32_t *const diepctl)
-{
-	struct udc_dwc2_data *const priv = udc_get_private(dev);
-	uint8_t ep_idx = USB_EP_GET_IDX(cfg->addr);
-
-	/* Clear FIFO number field */
-	*diepctl &= ~USB_DWC2_DEPCTL_TXFNUM_MASK;
-
-	if (priv->dynfifosizing) {
-		uint16_t higher_mask = ~BIT_MASK(ep_idx + 1);
-
-		if (priv->txf_set & higher_mask) {
-			LOG_WRN("Some of the FIFOs higher than %u are set, %x",
-				ep_idx, priv->txf_set & higher_mask);
-			return 0;
-		}
-
-		dwc2_set_txf(dev, ep_idx - 1, 0, 0);
-	}
-
-	priv->txf_set &= ~BIT(ep_idx);
 
 	return 0;
 }
@@ -1477,11 +1336,6 @@ static int udc_dwc2_ep_deactivate(const struct device *dev,
 	dxepctl = sys_read32(dxepctl_reg);
 	LOG_DBG("Disable ep 0x%02x DxEPCTL%u %x", cfg->addr, ep_idx, dxepctl);
 	dxepctl &= ~USB_DWC2_DEPCTL_USBACTEP;
-
-	if (USB_EP_DIR_IS_IN(cfg->addr) && udc_mps_ep_size(cfg) != 0U &&
-	    ep_idx != 0U) {
-		dwc2_unset_dedicated_fifo(dev, cfg, &dxepctl);
-	}
 
 	sys_write32(dxepctl, dxepctl_reg);
 	dwc2_set_epint(dev, cfg, false);
@@ -1725,6 +1579,7 @@ static int dwc2_core_soft_reset(const struct device *dev)
 
 static int udc_dwc2_init_controller(const struct device *dev)
 {
+	const struct udc_dwc2_config *const config = dev->config;
 	struct udc_dwc2_data *const priv = udc_get_private(dev);
 	struct usb_dwc2_reg *const base = dwc2_get_base(dev);
 	mem_addr_t grxfsiz_reg = (mem_addr_t)&base->grxfsiz;
@@ -1742,6 +1597,8 @@ static int udc_dwc2_init_controller(const struct device *dev)
 	uint32_t ghwcfg3;
 	uint32_t ghwcfg4;
 	uint32_t val;
+	uint16_t dfifodepth;
+	uint8_t txfnum;
 	int ret;
 	bool hs_phy;
 
@@ -1812,8 +1669,8 @@ static int udc_dwc2_init_controller(const struct device *dev)
 		usb_dwc2_get_ghwcfg2_otgarch(ghwcfg2),
 		usb_dwc2_get_ghwcfg2_otgmode(ghwcfg2));
 
-	priv->dfifodepth = usb_dwc2_get_ghwcfg3_dfifodepth(ghwcfg3);
-	LOG_DBG("DFIFO depth (DFIFODEPTH) %u bytes", priv->dfifodepth * 4);
+	dfifodepth = usb_dwc2_get_ghwcfg3_dfifodepth(ghwcfg3);
+	LOG_DBG("DFIFO depth (DFIFODEPTH) %u bytes", dfifodepth * 4);
 
 	priv->max_pktcnt = GHWCFG3_PKTCOUNT(usb_dwc2_get_ghwcfg3_pktsizewidth(ghwcfg3));
 	priv->max_xfersize = GHWCFG3_XFERSIZE(usb_dwc2_get_ghwcfg3_xfersizewidth(ghwcfg3));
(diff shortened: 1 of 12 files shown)


dts/arm/intel_socfpga_std/socfpga.dtsi
tmon-nordic, line 237: @nashif I found fifo information in [Cyclone V HPS Register Address Map and Definitions](https://www.intel.com/content/www/us/en/programmable/hps/cyclone-v/hps.html#sfo1410069396316.html) but there the usb0 is at 0xFFB00000 and not 0xFFB30000. Is it a bug in zephyr devicetree or am I looking at wrong place?

tmon-nordic:
@jean12332 can you verify whether this also fixes #95500?

raffarost:
> @raffarost Can you provide the devicetree values from esp32-s3?

hi @tmon-nordic,
does this look reasonable? Values for GRXFSIZ and GNPTXFSIZ are the ones currently in use.
```
Section         Words
----------------------
RX FIFO         64
EP0 TX (NP)     32
EP1 TX          40
EP2 TX          40
EP3 TX          40
EP4 TX          40
----------------------
Total           256  (GDFIFOCFG)
```
@sylvioalves

tmon-nordic:
> > @raffarost Can you provide the devicetree values from esp32-s3?
> 
> hi @tmon-nordic, does this look reasonable? Values for GRXFSIZ and GNPTXFSIZ are the ones currently in use.
> 
> ```
> Section         Words
> ----------------------
> RX FIFO         64
> EP0 TX (NP)     32
> EP1 TX          40
> EP2 TX          40
> EP3 TX          40
> EP4 TX          40
> ----------------------
> Total           256  (GDFIFOCFG)
> ```
> 
> @sylvioalves

What is actual GDFIFOCFG register value? The upper 16 bits show how many locations are available for fifos, the lower 16 bits show total SPRAM size. When DMA is used, some locations are used for storing DMA pointers (this is where the difference between the upper 16 bits and lower 16 bits come from).

GHWCFG3 upper 16 bits report available DFIFO Depth (after reset should be the same as upper 16 bits from GDFIFOCFG register).

drivers/usb/udc/udc_dwc2.c
```diff
@@ -3543,13 +3426,26 @@ static const struct udc_api udc_dwc2_api = {
 	static struct udc_ep_config ep_cfg_out[DT_INST_PROP(n, num_out_eps)];	\
 	static struct udc_ep_config ep_cfg_in[DT_INST_PROP(n, num_in_eps)];	\
 										\
+	BUILD_ASSERT(DT_INST_PROP_LEN(n, g_tx_fifo_size) ==			\
+		     (DT_INST_PROP(n, ghwcfg4) & USB_DWC2_GHWCFG4_INEPS_MASK) >>\
```
josuah: I've started to use macros for extracting things out of `ghwcfg#` here if that helps, maybe I can put them on `drivers/usb/common/usb_dwc2_hw.h` as a separate PR.

https://github.com/zephyrproject-rtos/zephyr/pull/95723/files#diff-a5c1ac2b1e9f1d28cb96d97ecee24e3b15ab543cf1f0d4cfbe650471bae2c02aR168-R193

```c
#define UHC_DWC2_CHAN_REG(base, chan_idx) \
	((struct usb_dwc2_host_chan *)(((mem_addr_t)(base)) + 0x500UL + ((chan_idx) * 0x20UL)))

#define UHC_DWC2_FIFODEPTH(config) \
	((uint32_t)(((config)->ghwcfg3 & USB_DWC2_GHWCFG3_DFIFODEPTH_MASK) >> \
		    USB_DWC2_GHWCFG3_DFIFODEPTH_POS))

#define UHC_DWC2_NUMHSTCHNL(config) \
	((uint32_t)(((config)->ghwcfg2 & USB_DWC2_GHWCFG2_NUMHSTCHNL_MASK) >> \
		    USB_DWC2_GHWCFG2_NUMHSTCHNL_POS))

#define UHC_DWC2_OTGARCH(config) \
	((uint32_t)(((config)->ghwcfg2 & USB_DWC2_GHWCFG2_OTGARCH_MASK) >> \
		    USB_DWC2_GHWCFG2_OTGARCH_POS))

#define UHC_DWC2_HSPHYTYPE(config) \
	((uint32_t)(((config)->ghwcfg2 & USB_DWC2_GHWCFG2_HSPHYTYPE_MASK) >> \
		    USB_DWC2_GHWCFG2_HSPHYTYPE_POS))

#define UHC_DWC2_FSPHYTYPE(config) \
	((uint32_t)(((config)->ghwcfg2 & USB_DWC2_GHWCFG2_FSPHYTYPE_MASK) >> \
		    USB_DWC2_GHWCFG2_FSPHYTYPE_POS))

#define UHC_DWC2_PHYDATAWIDTH(config) \
	((uint32_t)(((config)->ghwcfg4 & USB_DWC2_GHWCFG4_PHYDATAWIDTH_MASK) >> \
		    USB_DWC2_GHWCFG4_PHYDATAWIDTH_POS))
```

tmon-nordic: Your macros rely on there being ghwcfg members inside config structure. For BUILD_ASSERT it would be necessary to have a macro that takes in just the ghwcfg value directly.

While we do have the getters and setters for register fields, these are really functions - which implies that they cannot be used in BUILD_ASSERT.

I am not sure in how many places the macro would be used. If you are about to move to `drivers/usb/common/usb_dwc2_hw.h` then I think it owuld make sense to have the macro take ghwcfg member and not config (and not include UHC prefix because at least OTGARCH, HSPHYTYPE, FSPHYTYPE and PHYDATAWIDTH are not exclusive for host mode).

josuah: At the end it looks compact enough to just use `FIELD_GET()` directly, so I went with it at least for now:
```
FIELD_GET(USB_DWC2_GHWCFG3_DFIFODEPTH_MASK, ghwcfg3)                                                 
FIELD_GET(USB_DWC2_GHWCFG2_NUMHSTCHNL_MASK, ghwcfg2)                                                 
FIELD_GET(USB_DWC2_GHWCFG2_OTGARCH_MASK, ghwcfg2)                                                    
FIELD_GET(USB_DWC2_GHWCFG2_HSPHYTYPE, ghwcfg2)                                                       
FIELD_GET(USB_DWC2_GHWCFG2_FSPHYTYPE_MASK, ghwcfg2)                                                  
FIELD_GET(USB_DWC2_GHWCFG4_PHYDATAWIDTH_MASK, ghwcfg4)                                               
```

dts/bindings/usb/snps,dwc2.yaml
```diff
@@ -51,3 +51,22 @@ properties:
     description: |
       Value of the GHWCFG4 register. It is used to determine available endpoint
       types during driver pre-initialization.
+
+  g-rx-fifo-size:
+    type: int
+    required: true
+    description: |
+      Rx Data FIFO Depth. Must not exceed GRXFSIZ.RxFDep power-on value.
```
sylvioalves, line 61: Where `RxFDep` is used?

tmon-nordic: The value provided in devicetree is validated in ASSERT. `usb_dwc2_get_grxfsiz` is returning GRXFSIZ.RxFDep value.

Devicetree is expected to contain proper value, and if it does not then ASSERT is the right way to notify user (developer who writes/modified devicetree `snps,dwc2` entry).

sylvioalves (dismissed):

josuah (dismissed):

josuah:
This same strategy should help for host mode as well :+1:

sylvioalves (dismissed):

josuah (dismissed):

dts/bindings/usb/snps,dwc2.yaml
```diff
@@ -51,3 +51,22 @@ properties:
...
+    required: true
+    description: |
+      IN Endpoint 0 FIFO Depth. Must not exceed GNPTXFSIZ.NPTXFDep power-on value.
+
+  g-tx-fifo-size:
+    type: array
+    required: true
+    description: |
+      IN Endpoint Transmit fifo depths. Values must not exceed their respective
+      DIEPTXFn.INEPnTxFDep power-on values.
```
josuah, line 78: Given the calculation for host mode might be slightly different, would it make sense to give it a `d-` and `h-` prefix instead of `g-`?

josuah: `g-tx-fifo-size` is possibly a typo for `d-tx-fifo-size`? The register is named *D*IEPTXFn. Or possibly it was a choice.

tmon-nordic: The names are the same as in Linux
https://github.com/torvalds/linux/blob/0257f64bdac7fdca30fa3cae0df8b9ecbec7733a/Documentation/devicetree/bindings/usb/dwc2.yaml#L145-L155

due to Zephyr design goals
https://github.com/zephyrproject-rtos/zephyr/blob/81f27d13d16b74b98c80281144440ac8cfd5b781/doc/build/dts/design.rst?plain=1#L51-L58

josuah: And you said it in the first few lines of your message.

> Property names are the same as used in Linux

I should have read better. Thank you for confirming!

jfischer-no (changes requested):
>Dynamically allocating fifo sizes leads to very complex code that yields
subpar results.

For sure not, https://github.com/zephyrproject-rtos/zephyr/pull/94766 is not complex, and provides the necessary FIFO sizes based on actual USB device configuration. Using devicetree would not scale, and we have already far too many overlays in the tree.

tmon-nordic:
> > Dynamically allocating fifo sizes leads to very complex code that yields
> > subpar results.
> 
> For sure not, #94766 is not complex, and provides the necessary FIFO sizes based on actual USB device configuration. 

It is not complex, but it does not allow the users to take control over the allocations. One problem is for example deciding whether to have at least two wMaxPacketSize available for bulk endpoints (this does have significant impact on MSC performance).

There is a reason why reference Synopsys DWC2 driver (the one in Linux mainline) does use these very devicetree entries. I feel it is not worth it to try to come up with smarter code. However, if you want to go that path - do propose something that gives actual control to the application developer. There are already multiple problems with endpoint assignemnt logic in Zephyr (see https://github.com/zephyrproject-rtos/zephyr/pull/102041#discussion_r2930016478) and trying to come up with magic code for allocating fifos is not making the task any simpler.

dts/arm/intel_socfpga_std/socfpga.dtsi
```diff
@@ -238,6 +238,11 @@
 			ghwcfg1 = <0x00000000>;
 			ghwcfg2 = <0x208ffc90>;
 			ghwcfg4 = <0xfe0f0020>;
+			/* 8064 locations with full flexibility */
```
josuah, line 241: Should that also be a devicetree property (total number of locations) so that it is possible to have build-time asserts?

tmon-nordic: Linux doesn't have spram size as devicetree property, and I don't see a clear advantage of having one. I think just having asserts for sanity-checking in debug builds is all that is necessary.

josuah: Hmm yes I agree

dts/bindings/usb/snps,dwc2.yaml
josuah: If I understand correctly, users will sometimes want to optimize their application by giving more FIFO memory to their high-performance TX endpoint (IN).

For this they need to first compile their application, investigate the descriptors to see which endpoint the host allocated, then come back to devicetree and select the high-performance endpoint and give it more FIFO size (i.e. `2 * wMaxPacketSize`).

Ideally, users would want to have an API to say "TX endpoint of this class instance is high-bandwidth", but class instance is a runtime API, not a hardware detail.

In the meantime, some "app note" style comment describing this process could be interesting for users willing to improve their USB perf?

josuah:
I wonder if it would make sense to have some verification of which endpoint qualifies as "high-bandwidth" based on FIFO sizes? So that in the future, an API to say "this class instance needs a high-performance TX endpoint".

https://github.com/zephyrproject-rtos/zephyr/blob/d0402d79b885bb3ac1afe19ecc3992436439d2d7/include/zephyr/drivers/usb/udc.h#L83-L84

Maybe for a future PR though as seems unrelated.

I do not have in-depth insights of how UDC DWC2 works but did not find anything missing from what I could grasp.

tmon-nordic:
> I wonder if it would make sense to have some verification of which endpoint qualifies as "high-bandwidth" based on FIFO sizes?

This PR already does this, see `high_bandwidth = data->caps.hs && fifo_bytes >= 2 * mps;`

josuah:
My bad, I completely missed it!

Currently, it looks like the FIFO allocation is done directly inside `udc_dwc2_enable()`, however, endpoints capabilities are filled inside `dwc2_driver_preinit()`.

But in order to know which endpoints might be able to support high-bandwidth transfers, it is required to already know the allocated sizes.

So this means that endpoint allocation needs to be done before `dwc2_driver_preinit()`.

- Either fill an array with the heuristic decision (instead of applying it directly)
- Either having this data from devicetree like proposed here

In both situation, we end-up with a `fifo_sizes[]` array that contains the list of allocated FIFO sizes.

So in order to just get this bugfix (high-bandwidth endpoints are not only those that are high-speed capable):

```diff
-			config->ep_cfg_in[n].caps.high_bandwidth = data->caps.hs;
+			config->ep_cfg_in[n].caps.high_bandwidth = high_bandwidth;
```

Then this would require to keep most of this PR new introduced code, and remain only to discuss how to fill the `fifo_sizes[]` array?

Would that mean that this PR can be split in a first part that fills `fifo_sizes[]` (which might not require as much discussion) and a second part that decides how to fill this array (from devicetree or at runtime).

But I am not sure this helps to split the PR, sorry... Then nothing to change.

tmon-nordic:
> So in order to just get this bugfix (high-bandwidth endpoints are not only those that are high-speed capable):
> 
> ```diff
> -			config->ep_cfg_in[n].caps.high_bandwidth = data->caps.hs;
> +			config->ep_cfg_in[n].caps.high_bandwidth = high_bandwidth;
> ```

This is already done in this PR, so no change necessary within this PR context.

> Would that mean that this PR can be split in a first part that fills `fifo_sizes[]` (which might not require as much discussion) and a second part that decides how to fill this array (from devicetree or at runtime).
> 
> But I am not sure this helps to split the PR, sorry... Then nothing to change.

I don't think it helps to split this PR. I think going with the devicetree approach first is safe. If someone comes up with a brilliant idea of dynamic allocation that "just-works" then removing the devicetree entries would be trivial. The opposite way (adding new devicetree entries) is so much more complicated (we get to keep updating this PR every time new devicetree entry comes).

Currently the endpoint allocation is so bad that I am sure that this fixed devicetree allocations is better, easier to maintain and gives more flexibility to actual end-user.

josuah (dismissed):
I have the impression that on `main`, high-speed endpoints that do not have enough FIFO memory are still marked as high-bandwidth. Some SoCs do not have enough FIFO locations for having all HS endpoints as high-bandwidth.

> If someone comes up with a brilliant idea of dynamic allocation that "just-works" then removing the devicetree entries would be trivial.

As I understand, this PR first refactors the code to (a) configure the hardware FIFO sizes using values from a C array `fifo_sizes[]`, (b) fill it from devicetree, and (c) also use this `fifo_sizes[]` to correctly set `config->ep_cfg_{in,out}[n].caps.high_bandwidth`.

And this has the effect of (1) requiring devicetree configuration, (2) improving the FIFO size allocation, and also (3) fixing a bug where high-speed endpoints bug described above.

Even if wanting a heuristic-based allocation scheme, in order to fix high-bandwidth endpoints, we do definitely need (a) and (c).

This means that the current PR you propose also makes it easier to completely implement heuristics heuristics, because (a) and (c) would need to be done anyway in order to fix (3), and because devicetree-based config (2) allows user to experiment and report what worked best: collect field data out of which design a more versatile heuristics.

The only alternative I see is to directly introduce a heuristics trying to guess users most likely use-case but still keep (a) and (c), and I it is up to you and @jfischer-no to decide ofc. Here is just what I could understand.

josuah (dismissed):

drivers/usb/udc/udc_dwc2.c
```diff
@@ -1120,63 +1040,19 @@ static int dwc2_set_dedicated_fifo(const struct device *dev,
...
-		} else {
-			txfaddr = priv->rxfifo_depth +
-				MIN(UDC_DWC2_FIFO0_DEPTH, priv->max_txfifo_depth[0]);
-		}
-
-		if (priv->txf_set & BIT(ep_idx)) {
-			uint32_t curaddr;
+	/* TxFIFOs are preassigned, just verify it can be used */
+	txfnum = usb_dwc2_get_depctl_txfnum(*diepctl);
+	dieptxf = sys_read32((mem_addr_t)&base->dieptxf[txfnum]);
```
sylvioalves: Is it off by one? `[txfnum - 1]`

drivers/usb/udc/udc_dwc2.c
```diff
@@ -1120,63 +1040,19 @@ static int dwc2_set_dedicated_fifo(const struct device *dev,
...
-	/* Assign FIFO to the IN endpoint */
-	*diepctl = tmp | usb_dwc2_set_depctl_txfnum(ep_idx);
-	priv->txf_set |= BIT(ep_idx);
-	dwc2_flush_tx_fifo(dev, ep_idx);
+	dwc2_flush_tx_fifo(dev, txfnum);
 
 	LOG_INF("Set FIFO%u (ep 0x%02x) addr 0x%04x depth %u size %u",
-		ep_idx, cfg->addr, txfaddr, txfdep, dwc2_ftx_avail(dev, ep_idx));
+		txfnum, cfg->addr, usb_dwc2_get_dieptxf_inepntxfstaddr(dieptxf),
+		usb_dwc2_get_dieptxf_inepntxfdep(dieptxf), dwc2_ftx_avail(dev, txfnum));
```
sylvioalves: should it be indexed by `USB_EP_GET_IDX(cfg->addr)` instead of `txfnum` here?

drivers/usb/udc/udc_dwc2.c
```diff
@@ -3307,13 +3195,26 @@ static const struct udc_api udc_dwc2_api = {
 	static struct udc_ep_config ep_cfg_out[DT_INST_PROP(n, num_out_eps)];	\
 	static struct udc_ep_config ep_cfg_in[DT_INST_PROP(n, num_in_eps)];	\
 										\
+	BUILD_ASSERT(DT_INST_PROP_LEN(n, g_tx_fifo_size) ==			\
```
sylvioalves, line 3208: What if you add another BUILD_ASSERT for `g_np_tx_fifo_size) * 4 >= 8`? Well, that would only be needed if na fifo-size is < 2, which I don't think should happen?

drivers/usb/udc/udc_dwc2.c
tmon-nordic: Yes, you are right. This is off by one. Besides using `[txfnum - 1]`, there needs to be a if case for `txfnum == 0` that should check against `GNPTXFSIZ` `NPTXFDep`.

drivers/usb/udc/udc_dwc2.c
tmon-nordic: The `"FIFO%u"` should be indexed with `txfnum` because we refer to TxFIFO here, which does not necessarily need to match the IN endpoint number. `txfnum` valid range is always `[0; n]` where `n` is the number of TxFIFOs, while the only known thing about IN endpoints is that IN endpoint 0 will be there. Any other IN endpoint availability depends on the core configuration, and there is no requirement for any consecutiveness - for example `{0, 3, 7, 8, 10}` could be a valid set (if it was configured like this prior to silicon tapeout).

`dwc2_ftx_avail(dev, txfnum)` is wrongly indexed with `txfnum`. It definitely should be indexed with `USB_EP_GET_IDX(cfg->addr)` because otherwise we can reference non-existing registers.

drivers/usb/udc/udc_dwc2.c
tmon-nordic: Actually `GNPTXFSIZ` `NPTXFDep` and every `DIEPTXFn` `INEPnTxFDep` minimum value is 16, maximum 32768. I think it would be better to specify it in `snps,dwc2.yaml` so the validation is done even before build.

sylvioalves (dismissed):

mmahadevan108:
cc @MarkWangChinese
Hi Mark, Can you please review this PR.

josuah (dismissed):

drivers/usb/udc/udc_dwc2.c
```diff
@@ -2195,6 +2080,15 @@ static int dwc2_driver_preinit(const struct device *dev)
...
 	LOG_DBG("Number of IN endpoints (INEPS + 1) %u", ineps);
 
+	/* RxFIFO has to be able to hold at least:
+	 *   * 1 control endpoint in Completer/Buffer DMA mode: 13 locations
+	 *   * Global OUT NAK: 1 location
+	 * If in addition to above TxFIFO also has a space for at least:
+	 *   * Space for 2 * 1024 packets: ((1024/4) + 1) * 2 = 514 locations
+	 * then consider OUT endpoints to be High-Bandwidth capable.
+	 */
+	high_bandwidth = data->caps.hs && config->fifo_sizes[0] >= 528;
```
MarkWangChinese, line 2090: Does it need to check `priv->dynfifosizing`? I don't have much experience with the DWC2 controller, but from the code logic, `config->fifo_sizes` appears to be valid only when `priv->dynfifosizing` is `true`. Other than that, the changes look good to me.

tmon-nordic: `config->fifo_sizes` is always valid. Without `priv->dynfifosizing` it only means that the fifo sizes described in devicetree are completely fixed and they must match hardware configuration. With `priv->dynfifosizing` the values can be lowered compared to reset defaults (in most cases, the values actually must be lowered because otherwise they span out of SPRAM).

The only thing different between `priv->dynfifosizing` true and false, is that when true there are actual asserts that validate the value, and when false there are no asserts. We don't currently have any controller without dynamic fifo sizing in the tree, so I would prefer to not add such asserts blindly.

maass-hamburg (dismissed):

maass-hamburg:
dt parts are fine

sylvioalves (dismissed):

tmon-nordic:
@jfischer-no Let me try to narrow this down, because I think we are talking past each other on what the two PRs do.

To be accurate about one thing first: #94766 does address the alternate setting hole. Sizing `reqdep` from `cfg->m_mps` instead of the mps of the currently selected setting stops that particular failure, so I withdraw the suggestion that your PR ignores it.

What it costs to get there is a new field in the public `struct udc_ep_config` and another in `struct udc_ep_stat`, carried by every UDC driver, a changed `udc_ep_try_config()`, and a new requirement that the stack claim every endpoint of every configuration before `udc_enable()`. That is generic API surface spent so that the dwc2 allocator survives one case it would otherwise get wrong, and the allocator is otherwise unchanged. I would rather remove the reason the driver has to know the largest mps across all alternate settings than add a field to tell it.

What it does not change is that the endpoint capabilities still do not describe the FIFOs, that the allocation policy still takes the minimum, and that the TxFIFO number is still the endpoint number. Those are the parts I do not see a way to reconcile with a heuristic.

**`caps.high_bandwidth`, `caps.bulk` and `caps.mps` still do not reflect the FIFO depth.** `caps.high_bandwidth` is `data->caps.hs` for every IN endpoint, so `ep_check_config()` accepts a high-bandwidth endpoint on an endpoint whose FIFO cannot be sized for it, and `dwc2_set_dedicated_fifo()` reports `-ENOMEM` when it is enabled. That is not a recoverable error: `assign_ep_addr()` runs once in `usbd_init()` and its result is already in the descriptors the host has read, so the failed enable propagates to `ctrl_xfer_stall` and the device STALLs SET_CONFIGURATION or SET_INTERFACE. It enumerates, advertises a configuration it cannot support, and then refuses it.

With the capabilities describing the actual depths, none of that happens: `assign_ep_addr()` walks the endpoints and simply does not pick one that cannot do the job, at init time, before any descriptor is published. If no endpoint can do it, that surfaces as an init failure with a clear reason instead of a mystery enumeration problem in the field.

I do not think a heuristic can get here, and not because of its complexity. `udc_ep_claim_config()` is called from `assign_ep_addr()`, and #94766 derives the depths in `udc_dwc2_init_controller()` during `udc_enable()` - after assignment has finished. So the input to the heuristic, which endpoints the stack wants, is only known after the decision that depends on its output, how deep each FIFO is. Devicetree values are available in `dwc2_driver_preinit()`, before any of it. @josuah set out the same ordering problem on 24 June, noting that the capabilities are filled in `dwc2_driver_preinit()` while the sizes are only known later, and I agree with his reading of it. I do not think you have replied to that comment.

**Bulk IN endpoints get exactly one wMaxPacketSize of TxFIFO.** `reqdep *= MIN(2, (1 + addnl))` is a multiplication by 1 for bulk, since `addnl` is 0, so a 512-byte bulk IN endpoint gets 128 locations whether it is sized from `mps` or from `m_mps`. Under DMA that prevents the next packet from being fetched while the current one is still on the bus. On nRF54H20, MSC reaches 15.6 MB/s with 128 locations and 36.9 MB/s with 256, with otherwise identical software.

**Allocating the minimum has no upside.** Unused SPRAM has no alternative use - there is nothing else in the controller that can spend those locations - while a TxFIFO that is one packet deep costs throughput as above. So "allocate what the endpoint strictly needs" is not a conservative choice, it is simply a worse one, and the gap is a factor of two on a common class. Fixing it inside an allocator means deciding how much beyond the minimum each endpoint should get, which is where the guessing starts: 2 x MPS is right for bulk under DMA, pointless for a low-rate interrupt endpoint, and on parts where everything does not fit it has to come out of somewhere.

I did try the cheap version. #105631 doubles the TxFIFO for bulk endpoints in five lines and does deliver the throughput. @josuah asked in review whether a device could then run out of FIFO space sooner, and the honest answer is yes: on a part that cannot afford two packets for every bulk endpoint, an unconditional doubling turns a configuration that works today into one that fails. Making it safe inside the allocator means knowing the whole configuration before the first FIFO is programmed, which is the ordering problem above again. As a per-board decision in devicetree it has neither difficulty, which is why I think #105631 is a bandaid and this is the fix.

**The TxFIFO number is still derived from the endpoint index.** `dwc2_set_dedicated_fifo()` does `dwc2_set_txf(dev, ep_idx - 1, ...)` and `priv->txf_set |= BIT(ep_idx)`. TxFIFOs are always contiguous, but GHWCFG1 may describe non-contiguous IN endpoint numbers, so the mapping is simply wrong on such configurations, and the start address is then derived from a FIFO that was never programmed.

This is not hypothetical, and it is not only ours: #95500 is exactly this report, from someone else's silicon. Both halves of it are worth re-reading, because they are the two things I have been trying to argue:

> EP1 (mps 512B) fixed use of TXFF1 (depth is 16*4B) in udc_dwc.c, but our TXFF1 is not large enough. So we need to choose other TXFFs, such as TXFF2.

> not all EPs are bidirectional. For example, EP5 is out-only and EP6 is in-only, when EP6 is enable in sequence, TXFF5 can be used.

So on that part TXFF1 is capped at 16 locations while the endpoint it is hardwired to needs 512, and the IN endpoint numbers are not contiguous. The reporter's own suggestion was to "maintain the corresponding relationship by the new txffnum variable associated with DIEPCTLn.txfnum filed", which is what this PR does, resolved once at init from GHWCFG1 rather than searched at runtime. The issue was closed as stale without being addressed.

That issue does not name the silicon, but the layout is not unusual and not hard to find. RTL8730E is a part anyone can buy, and its datasheet describes six endpoints: endpoint 0 and endpoint 5 bidirectional, endpoints 1 and 3 IN only, endpoints 2 and 4 OUT only. That is four IN endpoints and therefore three dedicated TxFIFOs. Using the endpoint index as the FIFO number, endpoint 3 gets DIEPTXF3 and takes its start address from DIEPTXF2, which the driver never programmed, so those locations are lost and the address is whatever the reset value happens to be; endpoint 5 is refused outright. Walking GHWCFG1 instead gives TXFNUM 0, 1, 2 and 3 to endpoints 0, 1, 3 and 5, which is what the hardware expects.

**On the hardware constraints.** Configured depths must not exceed their power-on values, and those bounds are per FIFO - the 16 location TXFF1 in #95500 is one example, and on XMC4500 the GNPTXFSIZ power-on value is 16, so the control IN FIFO can never hold more than 64 bytes regardless of what an allocator would like to give it. Usable FIFO RAM on that part ends at EPInfoBaseAddr 0x27a, 634 locations, of which the RxFIFO alone wants 282. The power-on values can also sum well above the available SPRAM: on nRF54H20 each of the seven DIEPTXFn can be up to 512, but there are only 3050 locations, so at most two of them can be 512 and something has to give. Starting addresses are constrained too, so shrinking one TxFIFO does not necessarily make those locations usable by another.

None of that is impossible to encode in a heuristic, but the result is a set of per-part rules that describe silicon, which is what devicetree is for. It is also the reason the mainline Linux driver takes these three values from devicetree, under the same property names, and it gives an application a way to say "put the deep FIFOs on these endpoints" - which an allocator running inside `.ep_enable` cannot offer, and it gives us somewhere to decide thresholding later instead of having the driver guess.

**These two PRs are not competing for the same fix.** With the depths fixed there is nothing for the driver to get wrong about alternate settings and no reason for it to be told the largest mps across them, so that part of #94766 becomes unnecessary here rather than contradicted. If a heuristic later covers the policy questions above, dropping the three devicetree properties is a small patch; the reverse direction is considerably more work, and every new `snps,dwc2` node that lands meanwhile is one more node to add depths for.

So the two questions I would like an answer to:

- how would an allocator populate the endpoint capabilities, given that it only learns what the stack wants after the stack has already used those capabilities to assign endpoints?
- And is there a version of it that gives bulk IN endpoints more than the minimum and spends the SPRAM that is otherwise left idle?

If you have answers to both, I would rather review that than keep this open. If not, I would ask you to reconsider the change request.

drivers/usb/udc/udc_dwc2.c
```diff
@@ -1917,62 +1769,88 @@ static int udc_dwc2_init_controller(const struct device *dev)
...
-		 * on some target.
+		uint32_t start_addr;
+
+		/* Rx FIFO always starts at address 0 */
+		__ASSERT_EVAL(,
+			val = usb_dwc2_get_grxfsiz(sys_read32(grxfsiz_reg)),
+			val >= config->fifo_sizes[0],
+			"DT g-rx-fifo-size %d exceeds power-on value %d",
+			config->fifo_sizes[0], val);
+		sys_write32(usb_dwc2_set_grxfsiz(config->fifo_sizes[0]), grxfsiz_reg);
```
jfischer-no, line 1781: It should always check DT values and not silently write to grxfsiz register. udc_init() and udc_dwc2_init_controller() are expected to fail if something is not correct.

```c
		/* Rx FIFO always starts at address 0 */
		val = usb_dwc2_get_grxfsiz(sys_read32(grxfsiz_reg));
		if (config->fifo_sizes[0] > val) {
			LOG_ERR("DT g-rx-fifo-size %d exceeds power-on value %d",
				config->fifo_sizes[0], val);
			return -EINVAL;
		}

		sys_write32(usb_dwc2_set_grxfsiz(config->fifo_sizes[0]), grxfsiz_reg);
```

Similar should be applied to the lines L:1785,  L:1800,  L:1812.

dts/bindings/usb/snps,dwc2.yaml
```diff
@@ -51,3 +51,28 @@ properties:
     description: |
       Value of the GHWCFG4 register. It is used to determine available endpoint
       types during driver pre-initialization.
+
+  g-rx-fifo-size:
+    type: int
+    min: 16
+    max: 32768
+    required: true
```
jfischer-no, line 59: (Just a note) Properties are not used if DYNFIFOSIZING is not enabled in CHWCFG2, but it looks like all the DWC2 users in the tree have DYNFIFOSIZING enabled.
In the Linux kernel, they are optional.

jfischer-no (changes requested):

drivers/usb/udc/udc_dwc2.c
tmon-nordic: Changed to runtime checks.

dts/bindings/usb/snps,dwc2.yaml
tmon-nordic: Linux does default to `g_rx_fifo_size = 2048`, `g_np_tx_fifo_size = 1024` and rest TxFIFOs to average depth. Then the defaults are clamped to be not larger than the power-on reset values.

This automatic calculation would for sure break on Zephyr in-tree targets that relied on `MAX_RXFIFO_GDFIFO_PERCENTAGE`. Another problem with the defaults is that the values are not known until the registers can be accessed, which brings back the ep_cfg mps problem.

I think it is reasonable to make them required for the time being. Alternative would be to make them not required, but fail `udc_dwc2` build if the devicetree values are missing.

PR #99511, opened 2025-11-17
Bluetooth: BAP: Missing assignment to aggregated_bis_syncs

Thalley:
When using the add_src and mod_src APIs the
aggregated_bis_syncs variable was only ever set to 0 and never updated, thus making all calls to
bis_syncs_unique_or_no_pref to be true.


asbjornsabo (approved):
LGTM

larsgk (approved):

PR #99512, opened 2025-11-17
modules: openthread: fix doxygen for openthread_set_nat64_receive_cb

kartben:
Fix include guards and better document the kconfig dependency

### DIFF
--- a/modules/openthread/include/openthread.h
+++ b/modules/openthread/include/openthread.h
@@ -133,6 +133,7 @@ int openthread_stop(void);
  */
 void openthread_set_receive_cb(openthread_receive_cb cb, void *context);
 
+#if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR) || defined(__DOXYGEN__)
 /**
  * @brief Set the additional callback for receiving NAT64 translated packets.
  *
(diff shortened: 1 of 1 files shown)


pdgendt (approved):

nashif (approved):

rlubos (approved):

carlescufi (approved):

jhedberg (approved):

modules/openthread/include/openthread.h
```diff
@@ -133,17 +133,19 @@ int openthread_stop(void);
  */
 void openthread_set_receive_cb(openthread_receive_cb cb, void *context);
 
+#if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR) || defined(__DOXYGEN__)
```
Thalley, line 136: As per https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-1-conditional-compilation we shouldn't guard function declarations as all - Could we just remove it?

alxelax: Could improvement be done as a separate PR instead? Seems CI is broken due to this.

Thalley: Absolutely :) That's also why I didn't make it as change request; better get CI happy first :D

PR #99513, opened 2025-11-17
arch: arm: cortex_m: Allow VTOR to be relocated to TCM

PetervdPerk-NXP:
This change allows you to move the interrupt vector table to ITCM or DTCM instead of SRAM, which provides deterministic, single-cycle access, unlike multi-cycle SRAM reads that may incur additional wait states. This reduces exception entry latency and benefits real-time and interrupt-driven workloads.

Both ITCM and DTCM are supported for vector table placement. ITCM is accessed via the CPU instruction bus and typically provides the lowest possible interrupt latency, making it the preferred location when available. DTCM, while on the data bus, still offers significantly faster and more predictable access compared to SRAM and serves as a suitable alternative where ITCM is unavailable or reserved for hot code.

### DIFF
--- a/arch/arm/Kconfig
+++ b/arch/arm/Kconfig
@@ -190,4 +190,43 @@ config HAS_SWO
 	help
 	  When enabled, indicates that SoC has an SWO output
 
+DT_CHOSEN_Z_DTCM := zephyr,dtcm
+DT_CHOSEN_Z_ITCM := zephyr,itcm
+
+choice
+	prompt "Vector table memory location"
+	depends on SRAM_VECTOR_TABLE
+	default ARM_VECTOR_TABLE_SRAM
+
+config ARM_VECTOR_TABLE_SRAM
+	bool "Place the vector table in DT Chosen SRAM instead of DT Chosen Flash"
+	help
+	  When executing in place (XiP), selecting this option will result in the
+	  interrupt vector table being relocated from DT 'zephyr,flash' chosen
+	  memory to DT 'zephyr,sram' chosen memory.
+
+config ARM_VECTOR_TABLE_DTCM
+	bool "Place the vector table in DT Chosen DTCM instead of DT Chosen Flash"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_DTCM))
+	help
+	  When executing in place (XiP), selecting this option will result in the
+	  interrupt vector table being relocated from DT 'zephyr,flash' chosen
+	  memory to DT 'zephyr,dtcm' chosen memory. While the vector table is
+	  instruction-fetched during exception entry, a DTCM option is provided
+	  for systems where ITCM is unavailable. DTCM still offers low-latency,
+	  deterministic access compared to normal RAM, but is not the optimal
+	  location for instruction fetch performance.
+
+config ARM_VECTOR_TABLE_ITCM
+	bool "Place the vector table in DT Chosen ITCM instead of DT Chosen Flash"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_ITCM))
+	help
+	  When executing in place (XiP), selecting this option will result in the
+	  interrupt vector table being relocated from DT 'zephyr,flash' chosen
+	  memory to DT 'zephyr,itcm' chosen memory. ITCM provides single-cycle,
+	  deterministic instruction fetches via the CPU instruction bus, it offers
+	  the lowest interrupt latency and is the preferred location when available.
+
+endchoice
+
 endmenu
--- a/tests/application_development/vector_table_relocation/src/main.c
+++ b/tests/application_development/vector_table_relocation/src/main.c
@@ -11,6 +11,20 @@
 #include <zephyr/kernel.h>
 #include <zephyr/sys/printk.h>
 #include <zephyr/ztest.h>
+#include <zephyr/devicetree.h>
+
+#if defined(CONFIG_ARM_VECTOR_TABLE_DTCM)
+#define DTCM_NODE    DT_CHOSEN(zephyr_dtcm)
+#define SRAM_VT_BASE DT_REG_ADDR(DTCM_NODE)
+#define SRAM_VT_SIZE DT_REG_SIZE(DTCM_NODE)
+#elif defined(CONFIG_ARM_VECTOR_TABLE_ITCM)
+#define ITCM_NODE    DT_CHOSEN(zephyr_itcm)
+#define SRAM_VT_BASE DT_REG_ADDR(ITCM_NODE)
+#define SRAM_VT_SIZE DT_REG_SIZE(ITCM_NODE)
+#else
+#define SRAM_VT_BASE CONFIG_SRAM_BASE_ADDRESS
+#define SRAM_VT_SIZE (CONFIG_SRAM_SIZE * 1024U)
+#endif
 
 #ifdef SCB_VTOR_TBLBASE_Msk
 #define VTOR_MASK (SCB_VTOR_TBLBASE_Msk | SCB_VTOR_TBLOFF_Msk)
@@ -54,8 +68,7 @@ ZTEST(vector_table_relocation, test_vector_table_in_ram)
 	volatile uint32_t vtor_address = SCB->VTOR & VTOR_MASK;
 
 	printf("VTOR address: 0x%x\n", vtor_address);
-	zassert_true(vtor_address >= CONFIG_SRAM_BASE_ADDRESS &&
-			     vtor_address <= CONFIG_SRAM_BASE_ADDRESS + CONFIG_SRAM_SIZE * 1024U,
+	zassert_true(vtor_address >= SRAM_VT_BASE && vtor_address <= SRAM_VT_BASE + SRAM_VT_SIZE,
 		     "Vector table is not in RAM! Address: 0x%x", vtor_address);
 }
 
--- a/tests/application_development/vector_table_relocation/testcase.yaml
+++ b/tests/application_development/vector_table_relocation/testcase.yaml
@@ -11,3 +11,15 @@ tests:
       - mps3/corstone310/fvp
       - mps4/corstone315/fvp
       - mps4/corstone320/fvp
+
+  application_development.vector_table_relocation.dtcm:
+    arch_allow: arm
+    filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,dtcm")
+    extra_configs:
+      - CONFIG_ARM_VECTOR_TABLE_DTCM=y
+
+  application_development.vector_table_relocation.itcm:
+    arch_allow: arm
+    filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,itcm")
+    extra_configs:
+      - CONFIG_ARM_VECTOR_TABLE_ITCM=y
--- a/arch/arm/core/cortex_m/ram_vector_table.ld
+++ b/arch/arm/core/cortex_m/ram_vector_table.ld
@@ -28,5 +28,11 @@ SECTION_PROLOGUE(.sram_vt,,)
 	. += _vector_end - _vector_start;
 	MPU_ALIGN(_sram_vector_size);
 	_sram_vector_end = .;
+#if defined(CONFIG_ARM_VECTOR_TABLE_ITCM)
+} GROUP_DATA_LINK_IN(ITCM, ROMABLE_REGION)
+#elif defined(CONFIG_ARM_VECTOR_TABLE_DTCM)
+} GROUP_DATA_LINK_IN(DTCM, ROMABLE_REGION)
+#else
 } GROUP_DATA_LINK_IN(RAMABLE_REGION, ROMABLE_REGION)
+#endif
 _sram_vector_size = _sram_vector_end - _sram_vector_start;


arch/Kconfig
```diff
@@ -621,6 +621,26 @@ config SRAM_VECTOR_TABLE
 	  When XiP is enabled, this option will result in the vector table being
 	  relocated from Flash to SRAM.
 
+DT_CHOSEN_Z_DTCM := zephyr,dtcm
```
dcpleung: There is no `zephyr,dtcm`, but `arm,dtcm`.

Since this seems to be ARM specific at this point, maybe move the kconfig in `arch/arm/Kconfig`?

PetervdPerk-NXP: Moved to `arch/arm/Kconfig`.

There's actually `zephyr,dtcm` it's from the zephyr-specific-chosen-nodes see https://docs.zephyrproject.org/latest/build/dts/api/api.html#zephyr-specific-chosen-nodes.

dcpleung: Ah... okay. I just grep-ed in `dts/bindings`.

dcpleung (dismissed):

wearyzen:
Could you refer to the MPS3 boards which already use ITCM and DTCM without these extra configs?
https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/arm/mps3/mps3_corstone300_fvp.dts#L22

Another example where MPS3 board selects SRAM for flash and code but still uses ITCM for vector table:
https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/arm/mps3/mps3_corstone310_fvp.dts#L22
https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/arm/mps3/Kconfig.defconfig#L43

PetervdPerk-NXP:
> Could you refer to the MPS3 boards which already use ITCM and DTCM without these extra configs?
> 
> https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/arm/mps3/mps3_corstone300_fvp.dts#L22

In that example the MPS3 uses DTCM as primary ram setting `zephyr,sram` to `DTCM`, then this doesn't apply because if `CONFIG_SRAM_VECTOR_TABLE` gets used then `RAMABLE_REGION` is is equal to DTCM memory.

This is why we depend on `$(dt_chosen_enabled,$(DT_CHOSEN_Z_DTCM))` if `zephyr,dtcm` or `zephyr,itcm` is set then we can oly relocate the vector table to DTCM or ITCM.

In the case for example the `vmu_rt1170` the primary ram is `&ocram` and has seperate DTCM and ITCM entries.
https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/nxp/vmu_rt1170/vmu_rt1170_mimxrt1176_cm7.dts#L29

In this case
https://github.com/zephyrproject-rtos/zephyr/blob/1380739f2569abebae8a05b6517866dffcb692c7/boards/arm/mps3/Kconfig.defconfig#L43
The MPS3 seems to fully executing from ITCM so no XIP from flash but then `CONFIG_SRAM_VECTOR_TABLE` wouldn't be applicable at all anyhow.

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	depends on SRAM_VECTOR_TABLE
+	default VECTOR_TABLE_SRAM
+
+config VECTOR_TABLE_SRAM
+	bool "SRAM"
+	help
+	  Place the vector table in SRAM.
+
+config VECTOR_TABLE_DTCM
+	bool "DTCM"
```
wearyzen:
```suggestion
	bool "Place the vector table in DTCM instead of SRAM"
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	help
+	  Place the interrupt vector table in Data Tightly Coupled Memory (DTCM).
+
+	  While the vector table is instruction-fetched during exception entry, a DTCM
+	  option is provided for systems where ITCM is unavailable. DTCM still offers
+	  low-latency, deterministic access compared to normal RAM, but is not the
+	  optimal location for instruction fetch performance.
+
+config VECTOR_TABLE_ITCM
+	bool "ITCM"
```
wearyzen:
```suggestion
	bool "Place the vector table in ITCM instead of SRAM"
```

wearyzen (changes requested):

PetervdPerk-NXP:
@wearyzen Changed the Kconfig and added test variants for ITCM and DTCM.

wearyzen (dismissed):

tests/application_development/vector_table_relocation/testcase.yaml
```diff
@@ -11,3 +11,15 @@ tests:
       - mps3/corstone310/fvp
       - mps4/corstone315/fvp
       - mps4/corstone320/fvp
+
+  application_development.vector_table_relocation.dtcm:
+    arch_allow: arm
+    filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,dtcm")
```
wearyzen, line 17: I don't see any of the ci jobs running this test, is there a board that satisfies this condition and runs in ci?

PetervdPerk-NXP: I've tested locally with twister specifying board vmu_rt1170. But any target that has `zephyr,dtcm` should work. Any idea how to figure out that easily? Or should we add the vmu_rt1170 target? If so, how is that done?

wearyzen: Does it run on qemu if not we can try to find a board that's runs in either qemu or fvp and has itcm/dtcm. Let me see if I can find something and get back to this.

PetervdPerk-NXP: It looks like there's isn't any QEMU coverage for cortex-m7 which is typically a core that has TCM support.
Also to my surprise there isn't QEMU coverage at all for ARMv8-M, cores like cortex-m33/m55/m85 although we've got plenty of hardware in zephyr that has that.

nordicjm (dismissed):

tests/application_development/vector_table_relocation/testcase.yaml
wearyzen: We have support for Cortex-m55 and Cortex-m85 with Arm's FVP simulator and they run in ci but they have CONFIG_ROMSTART_RELOCATION_ROM=y so can't be used here.
But I found [this](https://github.com/zephyrproject-rtos/zephyr/blob/0c9d8b7a8e3b56c4255eafb667b26bcd7e757ac3/tests/application_development/vector_table_relocation/src/main.c#L57) line in the test which checks for the vector table to be in sram region and not in dtcm/itcm so I am wondering how the test passed for you locally.
I think you need to modify the test to check the ITCM and DTCM range as well.

PetervdPerk-NXP: My apologies yesterday evening I only tested compilation didn't have access to hardware.
I've modified to regioncheck to work with ITCM and DTCM.

Test result on real hardware
```
Device testing on:

| Platform                  | ID   | Serial device   |
|---------------------------|------|-----------------|
| vmu_rt1170/mimxrt1176/cm7 |      | /dev/ttyUSB0    |

INFO    - JOBS: 22
INFO    - Adding tasks to the queue...
INFO    - Added initial list of jobs to queue
INFO    - 1/4 vmu_rt1170/mimxrt1176/cm7 application_development.vector_table_relocation.dtcm PASSED (device 3.591s <zephyr>)
INFO    - 2/4 vmu_rt1170/mimxrt1176/cm7 application_development.ram_context_for_isr        PASSED (device 3.614s <zephyr>)
INFO    - 3/4 vmu_rt1170/mimxrt1176/cm7 application_development.vector_table_relocation.arm PASSED (device 3.490s <zephyr>)
INFO    - 4/4 vmu_rt1170/mimxrt1176/cm7 application_development.vector_table_relocation.itcm PASSED (device 3.560s <zephyr>)

INFO    - 2889 test scenarios (2677 configurations) selected, 2673 configurations filtered (2673 by static filter, 0 at runtime).
INFO    - 4 of 4 executed test configurations passed (100.00%), 0 built (not run), 0 failed, 0 errored, with no warnings in 66.96 seconds.
INFO    - 4 of 4 executed test cases passed (100.00%) on 1 out of total 1268 platforms (0.08%).
INFO    - 4 test configurations executed on platforms, 0 test configurations were only built.

Hardware distribution summary:

| Board                     | ID   |   Counter |   Failures |
|---------------------------|------|-----------|------------|
| vmu_rt1170/mimxrt1176/cm7 |      |         4 |          0 |
INFO    - Saving reports...
```

tests/application_development/vector_table_relocation/src/main.c
```diff
@@ -11,6 +11,19 @@
...
+#define DTCM_NODE DT_CHOSEN(zephyr_dtcm)
+#define DTCM_BASE DT_REG_ADDR(DTCM_NODE)
+#define DTCM_SIZE DT_REG_SIZE(DTCM_NODE)
+#endif
+
+#ifdef CONFIG_VECTOR_TABLE_ITCM
+#define ITCM_NODE DT_CHOSEN(zephyr_itcm)
+#define ITCM_BASE DT_REG_ADDR(ITCM_NODE)
+#define ITCM_SIZE DT_REG_SIZE(ITCM_NODE)
+#endif
```
wearyzen:
```suggestion
#if defined(CONFIG_VECTOR_TABLE_DTCM)
#define DTCM_NODE DT_CHOSEN(zephyr_dtcm)
#define SRAM_VT_BASE DT_REG_ADDR(DTCM_NODE)
#define SRAM_VT_SIZE DT_REG_SIZE(DTCM_NODE)
#elif defined(CONFIG_VECTOR_TABLE_ITCM)
#define ITCM_NODE DT_CHOSEN(zephyr_itcm)
#define SRAM_VT_BASE DT_REG_ADDR(ITCM_NODE)
#define SRAM_VT_SIZE DT_REG_SIZE(ITCM_NODE)
#else
#define SRAM_VT_BASE CONFIG_SRAM_BASE_ADDRESS
#define SRAM_VT_SIZE (CONFIG_SRAM_SIZE * 1024U)
#endif
```

tests/application_development/vector_table_relocation/src/main.c
```diff
@@ -54,9 +67,19 @@ ZTEST(vector_table_relocation, test_vector_table_in_ram)
...
+				"Vector table is not in DTCM! Address: 0x%x", vtor_address);
+#elif defined(CONFIG_VECTOR_TABLE_ITCM)
+	zassert_true(vtor_address >= ITCM_BASE &&
+				vtor_address <= ITCM_BASE + ITCM_SIZE,
+				"Vector table is not in ITCM! Address: 0x%x", vtor_address);
+#else
 	zassert_true(vtor_address >= CONFIG_SRAM_BASE_ADDRESS &&
 			     vtor_address <= CONFIG_SRAM_BASE_ADDRESS + CONFIG_SRAM_SIZE * 1024U,
 		     "Vector table is not in RAM! Address: 0x%x", vtor_address);
+#endif
```
wearyzen:
```suggestion
zassert_true(vtor_address >= SRAM_VT_BASE &&
vtor_address <= SRAM_VT_BASE+ SRAM_VT_SIZE,
"Vector table is not in RAM! Address: 0x%x", vtor_address);
```

wearyzen (changes requested):

tests/application_development/vector_table_relocation/testcase.yaml
wearyzen: I played around with MPS3 and MPS4 and got MPS3 to run the ITCM test with an overlay but DTCM failed for some reason I am not sure yet. I can try to modify one of these boards (MPS4 probably) later but for now I think your test results look fine.
I've added a comment to avoid extra ifdefs, otherwise LGTM

PetervdPerk-NXP: Thanks for the suggestions, I've cleaned up the code and reran the test, all okay from side.

wearyzen (dismissed):

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+
+choice
+	prompt "Vector table memory location"
+	depends on SRAM_VECTOR_TABLE
+	default VECTOR_TABLE_SRAM
+
+config VECTOR_TABLE_SRAM
+	bool "Place the vector table in SRAM"
+	help
+	  Place the vector table in SRAM.
```
JarmouniA:
```suggestion
	  Place the vector table in DT Chosen SRAM.
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+
+config VECTOR_TABLE_DTCM
+	bool "Place the vector table in DTCM instead of SRAM"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_DTCM))
+	help
+	  Place the interrupt vector table in Data Tightly Coupled Memory (DTCM).
```
JarmouniA:
```suggestion
	  Place the interrupt vector table in DT Chosen DTCM (Data Tightly Coupled Memory).
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	  While the vector table is instruction-fetched during exception entry, a DTCM
+	  option is provided for systems where ITCM is unavailable. DTCM still offers
+	  low-latency, deterministic access compared to normal RAM, but is not the
+	  optimal location for instruction fetch performance.
+
+config VECTOR_TABLE_ITCM
+	bool "Place the vector table in ITCM instead of SRAM"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_ITCM))
+	help
+	  Place the interrupt vector table in Instruction Tightly Coupled Memory (ITCM).
```
JarmouniA:
```suggestion
	  Place the interrupt vector table in DT Chosen ITCM (Instruction Tightly Coupled Memory).
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	depends on SRAM_VECTOR_TABLE
+	default VECTOR_TABLE_SRAM
+
+config VECTOR_TABLE_SRAM
+	bool "Place the vector table in SRAM"
+	help
+	  Place the vector table in SRAM.
+
+config VECTOR_TABLE_DTCM
+	bool "Place the vector table in DTCM instead of SRAM"
```
JarmouniA:
```suggestion
	bool "Place the vector table in DTCM instead of DT Chosen Flash"
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+DT_CHOSEN_Z_DTCM := zephyr,dtcm
+DT_CHOSEN_Z_ITCM := zephyr,itcm
+
+choice
+	prompt "Vector table memory location"
+	depends on SRAM_VECTOR_TABLE
+	default VECTOR_TABLE_SRAM
+
+config VECTOR_TABLE_SRAM
+	bool "Place the vector table in SRAM"
```
JarmouniA:
```suggestion
	bool "Place the vector table in DT Chosen SRAM"
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	help
+	  Place the interrupt vector table in Data Tightly Coupled Memory (DTCM).
+
+	  While the vector table is instruction-fetched during exception entry, a DTCM
+	  option is provided for systems where ITCM is unavailable. DTCM still offers
+	  low-latency, deterministic access compared to normal RAM, but is not the
+	  optimal location for instruction fetch performance.
+
+config VECTOR_TABLE_ITCM
+	bool "Place the vector table in ITCM instead of SRAM"
```
JarmouniA:
```suggestion
	bool "Place the vector table in ITCM instead of DT Chosen Flash"
```

tests/application_development/vector_table_relocation/testcase.yaml
JarmouniA, line 1: Please move the excluded platforms in the 1st test scenario to a common section so they apply to all test scenarios.

JarmouniA (changes requested):

tests/application_development/vector_table_relocation/testcase.yaml
PetervdPerk-NXP: Technically the excluded platforms don't apply to the ITCM and DTCM testcases because simply they don't have chosen defined and thus get filtered out already. Do we really want to have it in common ?

JarmouniA (commented):
The 1st commit message is not accurate, we are not relocating the table from SRAM to I/DTCM, rather from Flash, because `SRAM_VECTOR_TABLE` depends on XIP.
https://github.com/zephyrproject-rtos/zephyr/blob/b5744c3158c2c0b12ff0e78d680fa653e0cfe3e1/arch/Kconfig#L609
Maybe this should be clarified in the Kconfig options help as well.

tests/application_development/vector_table_relocation/testcase.yaml
JarmouniA: Ah yes you are right !

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
 
+DT_CHOSEN_Z_DTCM := zephyr,dtcm
+DT_CHOSEN_Z_ITCM := zephyr,itcm
+
+choice
+	prompt "Vector table memory location"
+	depends on SRAM_VECTOR_TABLE
+	default VECTOR_TABLE_SRAM
+
+config VECTOR_TABLE_SRAM
```
JarmouniA:
```suggestion
config ARM_VECTOR_TABLE_SRAM
```
So that it is not confused with `SRAM_VECTOR_TABLE`.

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	bool "Place the vector table in SRAM"
+	help
+	  Place the vector table in SRAM.
+
+config VECTOR_TABLE_DTCM
```
JarmouniA:
```suggestion
config ARM_VECTOR_TABLE_DTCM
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,40 @@ config HAS_SWO
...
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_DTCM))
+	help
+	  Place the interrupt vector table in Data Tightly Coupled Memory (DTCM).
+
+	  While the vector table is instruction-fetched during exception entry, a DTCM
+	  option is provided for systems where ITCM is unavailable. DTCM still offers
+	  low-latency, deterministic access compared to normal RAM, but is not the
+	  optimal location for instruction fetch performance.
+
+config VECTOR_TABLE_ITCM
```
JarmouniA:
```suggestion
config ARM_VECTOR_TABLE_ITCM
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,41 @@ config HAS_SWO
...
+choice
+	prompt "Vector table memory location"
+	depends on SRAM_VECTOR_TABLE
+	default ARM_VECTOR_TABLE_SRAM
+
+config ARM_VECTOR_TABLE_SRAM
+	bool "Place the vector table in DT Chosen SRAM instead of DT Chosen Flash"
+	help
+	  Place the interrupt vector table in the SRAM region specified in the
+	  Device Tree’s chosen node after copying it from flash
```
JarmouniA:
```suggestion
	  When executing in place (XiP), selecting this option will result in
	  the interrupt vector table being relocated from DT 'zephyr,flash' chosen memory
	  to DT 'zephyr,sram' chosen memory. 
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,41 @@ config HAS_SWO
...
+
+config ARM_VECTOR_TABLE_DTCM
+	bool "Place the vector table in DT Chosen DTCM instead of DT Chosen Flash"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_DTCM))
+	help
+	  Place the interrupt vector table in DT Chosen DTCM (Data Tightly Coupled Memory).
```
JarmouniA:
```suggestion
	  When executing in place (XiP), selecting this option will result in
	  the interrupt vector table being relocated from DT 'zephyr,flash' chosen memory
	  to DT 'zephyr,dtcm' chosen memory. 
```

arch/arm/Kconfig
```diff
@@ -190,4 +190,41 @@ config HAS_SWO
...
+	  While the vector table is instruction-fetched during exception entry, a DTCM
+	  option is provided for systems where ITCM is unavailable. DTCM still offers
+	  low-latency, deterministic access compared to normal RAM, but is not the
+	  optimal location for instruction fetch performance.
+
+config ARM_VECTOR_TABLE_ITCM
+	bool "Place the vector table in DT Chosen ITCM instead of DT Chosen Flash"
+	depends on $(dt_chosen_enabled,$(DT_CHOSEN_Z_ITCM))
+	help
+	  Place the interrupt vector table in DT Chosen ITCM (Instruction Tightly Coupled Memory).
```
JarmouniA:
```suggestion
	  When executing in place (XiP), selecting this option will result in
	  the interrupt vector table being relocated from DT 'zephyr,flash' chosen memory
	  to DT 'zephyr,itcm' chosen memory. 
```

JarmouniA (approved):

wearyzen (approved):

PR #99514, opened 2025-11-17
bluetooth: audio: scan_delegator: Fix validation

alexsven:
- Fixes bug where bis_sync_requests were wrongfully validated.
  - Two BISes across subgroups should not have the same bits set
- Remove use of internal->bis_syncs before it's updated in mod_src.

### DIFF
--- a/subsys/bluetooth/audio/bap_scan_delegator.c
+++ b/subsys/bluetooth/audio/bap_scan_delegator.c
@@ -148,7 +148,7 @@ static bool bis_syncs_unique_or_no_pref(uint32_t requested_bis_syncs,
 		return true;
 	}
 
-	return (requested_bis_syncs & aggregated_bis_syncs) != 0U;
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/audio/bap_scan_delegator.c
```diff
@@ -148,7 +148,7 @@ static bool bis_syncs_unique_or_no_pref(uint32_t requested_bis_syncs,
 		return true;
 	}
 
-	return (requested_bis_syncs & aggregated_bis_syncs) != 0U;
+	return ((requested_bis_syncs & aggregated_bis_syncs) == 0U);
```
Thalley:
```suggestion
	return (requested_bis_syncs & aggregated_bis_syncs) == 0U;
```

The extra parenthesis aren't needed, but won't block on that

Thalley (dismissed):

Thalley (approved):

asbjornsabo (approved):
LGTM

PR #99515, opened 2025-11-17
samples: net: ocpp: add websocket path configuration option

kartben:
Make the WebSocket path configurable since having it hardcoded in the sample is not practical.

### DIFF
--- a/samples/net/ocpp/README.rst
+++ b/samples/net/ocpp/README.rst
@@ -26,6 +26,21 @@ Requirements
 - SteVe Demo Server (<https://github.com/steve-community/steve/blob/master/README.md>)
 - LAN for testing purposes (Ethernet)
 
+Configuration
+*************
+
+The sample application supports several configuration options that can be set
+via Kconfig:
+
+- ``CONFIG_NET_SAMPLE_OCPP_SERVER``: OCPP central system server IP address
+
+- ``CONFIG_NET_SAMPLE_OCPP_PORT``: OCPP central system server port number
+
+- ``CONFIG_NET_SAMPLE_OCPP_WS_PATH``: WebSocket URL path.
+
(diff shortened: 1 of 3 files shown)


jukkar (approved):

samples/net/ocpp/README.rst
```diff
@@ -26,6 +26,21 @@ Requirements
 - SteVe Demo Server (<https://github.com/steve-community/steve/blob/master/README.md>)
 - LAN for testing purposes (Ethernet)
 
+Configuration
+*************
+
+The sample application supports several configuration options that can be set
+via Kconfig:
+
+- ``CONFIG_NET_SAMPLE_OCPP_SERVER``: OCPP central system server IP address
```
ssekar15, line 35: s/server IP address/ server URL or IP address

ssekar15 (changes requested):
thanks looks good, nit

samples/net/ocpp/README.rst
kartben: Unless I am mistaken I don't think the sample enables DNS resolver, so this was intentionally referring to IP addresses since hostnames won't work

ssekar15: Indeed sample uses `getaddrinfo` DNS API to resolve IP address from URL

kartben: > Indeed sample uses `getaddrinfo` DNS API to resolve IP address from URL

yes, and unless CONFIG_DNS_RESOLVER is enabled this doesn't work for anything but a string representing an IP address :)

kartben: @ssekar15 please revisit unless you think I missed something

ssekar15: okay

ssekar15 (approved):

PR #99520, opened 2025-11-17
riscv: pmp: Add support for unlocked global PMP entries

fsammoura1980:
Adds the Kconfig option `PMP_NO_LOCK_GLOBAL` to remove the PMP Lock bit usage for global entries. The global entry is an internal detail of the driver implementation and should not be reflected in the user interface. This allows the application to dynamically reconfigure the PMP entries without requiring hard reset. This is essential for firmware that performs an RO-to-RW jump. By keeping these system entries unlocked, higher-privileged M-mode code can dynamically reconfigure memory permissions during the secure handover process, which is not possible if the entries are permanently locked during early boot.

Adds four new test cases to validate the PMP stack guard protection when `CONFIG_PMP_NO_LOCK_GLOBAL` is enabled (i.e., when PMP entries are not globally locked).

### DIFF
--- a/arch/riscv/Kconfig
+++ b/arch/riscv/Kconfig
@@ -410,6 +410,14 @@ config PMP_GRANULARITY
 	  (ie 4, 8, 16, ...), but if neither TOR mode nor NA4 mode is
 	  supported, the minimum granularity is 8.
 
+config PMP_NO_LOCK_GLOBAL
+	bool "Do not lock the global PMP entries"
+	select PMP_KERNEL_MODE_DYNAMIC
+	help
+	  Configure the PMP entries as unlocked (L=0) to implement PMP relative
+	  features. This allows application to dynamically reconfigure PMP
+	  entries without requiring hard reset.
+
 endif #RISCV_PMP
 
 config PMP_STACK_GUARD
--- a/arch/riscv/core/pmp.c
+++ b/arch/riscv/core/pmp.c
@@ -521,14 +521,15 @@ void z_riscv_pmp_init(void)
 	 * Use a PMP slot to make region (starting at address 0x0) inaccessible
 	 * for detecting null pointer dereferencing.
 	 */
-	set_pmp_entry(&index, PMP_NONE | PMP_L,
+	set_pmp_entry(&index, PMP_NONE | COND_CODE_1(CONFIG_PMP_NO_LOCK_GLOBAL, (0x0), (PMP_L)),
 		      0,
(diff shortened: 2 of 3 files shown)


arch/riscv/core/pmp.c
```diff
@@ -540,18 +540,18 @@ void z_riscv_pmp_init(void)
 	 * addresses inaccessible. This will never change so we do it here
 	 * and lock it too.
```
jimmyzhe: Could you update the comment here? Everything else LGTM. Thanks.

fsammoura1980: Done. I modified the comment.

jimmyzhe: Thanks for the detailed comment.

The purpose of CONFIG_PMP_NO_LOCK is already described in the Kconfig help text.
For the ISR stack guard comment here, I think we can simply remove the lock related word.
```
        /*
         * Set the stack guard for this CPU's IRQ stack by making the bottom
-        * addresses inaccessible. This will never change so we do it here
-        * and lock it too.
+        * addresses inaccessible. This will never change so we do it here.
         */
```

fsammoura1980: ok. agreed. I modified it per your recommendation.

jimmyzhe (dismissed):

arch/riscv/Kconfig
```diff
@@ -410,6 +410,14 @@ config PMP_GRANULARITY
 	  (ie 4, 8, 16, ...), but if neither TOR mode nor NA4 mode is
 	  supported, the minimum granularity is 8.
 
+config PMP_NO_LOCK
+	bool "Do not use PMP Lock bit"
```
fkokosinski: IMO this prompt is a bit misleading and could benefit from being changed to something more verbose like:
> Do not lock the stackguard PMP entries

fsammoura1980: I changed it to `Do not lock the global PMP entries`

fkokosinski: Sounds good, thanks! Could you change the symbol name as well? To something like `PMP_NO_LOCK_GLOBAL` as well? Because `PMP_NO_LOCK` sounds a bit like we're not supporting the L bit at all.

fsammoura1980: Done. Modified it to `PMP_NO_LOCK_GLOBAL`

fkokosinski: Thx!

fkokosinski (approved):

jimmyzhe (approved):

PR #99523, opened 2025-11-17
tests: riscv: Add unit tests for riscv_pmp_decode_region

fsammoura1980:
This commit introduces a suite of unit tests for the `pmp_decode_region` function using the Ztest framework. The tests validate the correct calculation of start and end addresses for various RISC-V PMP (Physical Memory Protection) entry configurations.

The following scenarios are covered:
- TOR (Top of Range) mode for index 0.
- TOR (Top of Range) mode for index > 0.
- NA4 (Naturally Aligned Four-byte) mode.
- NAPOT (Naturally Aligned Power-of-Two) mode.
- Default behavior for a disabled PMP entry.

These tests ensure the PMP region decoding logic is accurate across different addressing modes.

### DIFF
--- a/tests/arch/riscv/pmp/pmp-decode-region/CMakeLists.txt
+++ b/tests/arch/riscv/pmp/pmp-decode-region/CMakeLists.txt
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(riscv_pmp)
+
+FILE(GLOB app_sources src/*.c)
+target_sources(app PRIVATE ${app_sources})
+
+target_include_directories(app PRIVATE
+  ${ZEPHYR_BASE}/kernel/include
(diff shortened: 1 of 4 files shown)


tests/arch/riscv/pmp/mem-attr-entries/src/main.c
```diff
@@ -124,4 +124,86 @@ ZTEST(riscv_pmp_memattr_entries, test_dt_pmp_perm_conversion)
 		      PMP_R | PMP_W | PMP_X, result);
 }
 
+ZTEST(riscv_pmp_memattr_entries, test_pmp_tor_index_0)
```
fkokosinski: Thanks a lot for the tests! They don't seem to be related to the `mem-attr` property per se, so I'd prefer them to be moved to a separate test under `tests/arch/riscv/pmp`.

fkokosinski (changes requested):

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
fsammoura1980: Done

fkokosinski (approved):

jimmyzhe (approved):

PR #99530, opened 2025-11-17
riscv: pmp: Add API to change region permissions at runtime

fsammoura1980:
Introduce function `z_riscv_pmp_change_permissions` provides a mechanism to modify the Read, Write, and Execute (R/W/X) permissions of an existing PMP region based on its memory attribute index.

This function locates the PMP entry that corresponds to a region defined in the device tree via memory attributes (as managed by `mem_attr_get_regions`) and updates the PMP configuration register (`pmpcfg`) to reflect the new permissions.

It first iterates through all configured PMP entries, decodes their address and size, and attempts to find a full match for the specified region's start address and size. Once the matching PMP entry is found, its R, W, and X bits are updated based on the provided 'perm' mask, and the change is committed to the PMP configuration registers.

This commit implements a new unit test suite to validate the `z_riscv_pmp_change_permissions` function.

### DIFF
--- a/arch/riscv/core/pmp.c
+++ b/arch/riscv/core/pmp.c
@@ -58,6 +58,8 @@ LOG_MODULE_REGISTER(mpu);
 
 #define PMP_NONE 0
 
+#define PMP_PERM_MASK (PMP_R | PMP_W | PMP_X)
+
 /**
  * @brief Decodes PMP configuration and address registers into a memory region's
  * start/end addresses.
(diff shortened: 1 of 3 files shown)


arch/riscv/core/pmp.c
```diff
@@ -504,6 +506,82 @@ static unsigned int global_pmp_end_index[MODE_TOTAL];
 static unsigned long mem_attr_pmp_addr[CONFIG_PMP_SLOTS];
 #endif
 
+#ifdef CONFIG_MEM_ATTR
+int z_riscv_pmp_change_permissions(size_t region_idx, uint8_t perm)
```
fkokosinski, line 510: should we allow this symbol to set the L bit as well?

fsammoura1980: We do not allow locked entries for MEM_ATTR. We only[ allow R/W/X](https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/zephyr/main/arch/riscv/include/pmp.h;l=14#:~:text=13-,14,-15). As such, I do not believe allowing the L bit is acceptable here.

fkokosinski: ack

fkokosinski (dismissed):

npitre (commented):
A few comments:

- Don't use `ZTEST_USER()` here. None of this should ever be usable from
  user (unprivileged) threads.

- You shouldn't use `csr_write()` when a nice abstraction already exists for
  you with extra validation checks. Please consider
  `write_pmp_entries(entry_index, entry_index+1, false, pmp_addr, pmp_cfg, ARRAY_SIZE(pmp_addr))`.

npitre (commented):
Why the `MSTATUS_MPRV` addition? Looks like a possible cut-and-paste from
elsewhere. Please justify.

fsammoura1980:
> Look for `arch_irq_lock()` usage. You should justify it with a comment by mentioning that, unlike the other functions, this is not invoked from an atomic context.

Thanks! I modified the code and added a comment.

npitre:
> Thanks! I modified the code and added a comment.

You forgot to unlock interrupts if `entry_index == -1`.

npitre (dismissed):

fkokosinski (dismissed):

include/zephyr/arch/riscv/pmp.h
```diff
@@ -11,6 +11,33 @@
 
 #ifdef CONFIG_RISCV_PMP
 
+#ifdef CONFIG_MEM_ATTR
```
kartben: no guard here please - "[Function declarations that are available only when the option is enabled should be provided unconditionally](https://docs.zephyrproject.org/latest/develop/api/design_guidelines.html#:~:text=Function%20declarations%20that%20are%20available%20only%20when%20the%20option%20is%20enabled%20should%20be%20provided%20unconditionally)."

include/zephyr/arch/riscv/pmp.h
```diff
@@ -11,6 +11,33 @@
...
+ * the device tree via memory attributes (as managed by mem_attr_get_regions)
+ * and updates the PMP configuration register (pmpcfg) to reflect the new
+ * permissions.
+ *
+ * It first iterates through all configured PMP entries, decodes their address
+ * and size, and attempts to find a full match for the specified region's start
+ * address and size. Once the matching PMP entry is found, its R, W, and X
+ * bits are updated based on the provided 'perm' mask, and the change is
+ * committed to the PMP configuration registers.
+ *
```
kartben, line 27: if you want to express the fact it is only available when the config option is enabled:

```suggestion
 *
 * @kconfig_dep{CONFIG_MEM_ATTR}
```

kartben (dismissed):

include/zephyr/arch/riscv/pmp.h
fsammoura1980: Thanks for the feedback. Your comment has been resolved

include/zephyr/arch/riscv/pmp.h
fsammoura1980: Thank you for the feedback. I incorporated your comment.

npitre (approved):

fkokosinski (approved):

PR #99531, opened 2025-11-17
west: sign/rimage: use rimage config to find SOF directory

dcpleung:
After removal of SOF from manifest, if rimage is in path, the build would fail because the gen-uuid-reg.py cannot be found anymore. To fix this, use the rimage config path to find the SOF directory since the rimage config path is a directory under SOF. This would allow rimage to again sign images to run on hardware.

### DIFF
--- a/boards/intel/adsp/doc/intel_adsp_generic.rst
+++ b/boards/intel/adsp/doc/intel_adsp_generic.rst
@@ -60,22 +60,26 @@ Signing Tool
 As firmware binary signing is mandatory on Intel products from Skylake onwards,
 you will also need to set up the SOF rimage signing tool and key.
 
-.. code-block:: shell
+Clone the SOF repository and follow the instructions in the rimage :file:`README.md`
+to build the tool on your system.
 
-   cd zephyrproject
-   west config manifest.project-filter -- +sof
-   west update
-   cd modules/audio/sof/tools/rimage
-
-Follow the instructions in the rimage :file:`README.md` to build the tool on
-your system. You can either copy the executable to a directory in your PATH or
+You can either copy the executable to a directory in your PATH or
 use ``west config rimage.path /path/to/rimage-build/rimage``; see more details
 in the output of ``west sign -h``. Running directly from the build directory
 makes you less likely to use an obsolete rimage version by mistake.
 
-Platform-specific configuration files are located in the ``rimage/config/``
-subdirectory. For a different configuration directory you can use:
-``west config build.cmake-args -- -DRIMAGE_CONFIG_PATH=/path/to/my/rimage/config``.
+The west signing script will also need to know the path to the SOF directory.
+Depending on whether you need to use twister:
+
+* If you only need to build with ``west build``:
+
+  * Platform-specific configuration files are located in the ``rimage/config/``
+    subdirectory. For a different configuration directory you can use:
+    ``west config build.cmake-args -- -DRIMAGE_CONFIG_PATH=/path/to/my/rimage/config``.
+
+* If you also need to use twister to build and run on hardware:
+
+  * Environment variable ``SOF_SRC_DIR`` must be set to point to the SOF directory.
 
 
 Xtensa Toolchain (Optional)
--- a/scripts/west_commands/sign.py
+++ b/scripts/west_commands/sign.py
@@ -423,6 +423,75 @@ class ImgtoolSigner(Signer):
 
 class RimageSigner(Signer):
 
+    def check_if_sof_dir(self, maybe_sof_dir):
+        '''
+        Check if a pathlib.Path() object is the root of the SOF tree.
+
+        Returns a Path object if it is indeed the SOF tree, or None if not.
+
+        This checks if the file "versions.json" exists under the path, as
+        this file contains the SOF versioning information.
+        '''
+        versions_json_file = maybe_sof_dir / 'versions.json'
+        if versions_json_file.exists():
+            return maybe_sof_dir
+
+        return None
+
+    def get_sof_src_dir(self):
+        '''
+        Find the SOF tree.
+
+        Returns a Path object if SOF tree is found, or None if not.
+        '''
+        sof_src_dir = None
+
+        # Try to find SOF directory via path to rimage config files pointed to
+        # via CMake cached variable "RIMAGE_CONFIG_PATH" if it exists.
+        # The config files are inside <sof top>/tools/tools/rimage/config in
+        # the SOF tree.
+        conf_dir = self.cmake_cache.get('RIMAGE_CONFIG_PATH')
+        if conf_dir:
+            # Config path is <sof top>/tools/rimage/config/
+            # So need to go up 3 levels to get SOF directory.
+            maybe_sof_dir = pathlib.Path(conf_dir).parent.parent.parent
+
+            # Make sure this is the actual SOF directory as
+            # RIMAGE_CONFIG_PATH may point to somewhere else.
+            sof_src_dir = self.check_if_sof_dir(maybe_sof_dir)
+
+        # Try to find SOF if it is included via manifest as a project
+        # under the main manifest (e.g. included via a sub-manifest).
+        if not sof_src_dir:
+            try:
+                sof_proj = self.command.manifest.get_projects(['sof'], allow_paths=False)
+                sof_src_dir = pathlib.Path(sof_proj[0].abspath)
+
+                # Since SOF is pulled in as a project, we assume it is
+                # the correct one as the manifest has to be modified
+                # manually.
+            except ValueError:
+                pass
+
+        # Try to find SOF as SOF is the top level manifest via west init.
+        if not sof_src_dir:
+            maybe_sof_dir = pathlib.Path(manifest.manifest_path()).parent
+
+            # Make sure this is the actual SOF directory
+            # as the top level manifest may not be SOF.
+            sof_src_dir = self.check_if_sof_dir(maybe_sof_dir)
+
+        # If the above all failed to find the SOF tree, see if the path is specified
+        # via environment variable "SOF_SRC_DIR" as last resort.
+        if not sof_src_dir and os.getenv("SOF_SRC_DIR"):
+            maybe_sof_dir = pathlib.Path(os.getenv("SOF_SRC_DIR"))
+
+            # Make sure this is the actual SOF directory
+            # as the top level manifest may not be SOF.
+            sof_src_dir = self.check_if_sof_dir(maybe_sof_dir)
+
+        return sof_src_dir
+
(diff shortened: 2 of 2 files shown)


marc-hb (commented):
This looks like a minor hack but still a hack: why is sof.git not in any manifest? 

You can combine SOF with any moving Zephyr version of your choice using `sof/west.yml` + a "submanifest". Example:

https://github.com/thesofproject/sof/commit/f77741740093a25e6b14d6861b1f824a25eb0d5e

This was not theoretical, this submanifest example was used "for real" in CI for a long time.

I regularly get lost in `west` "import" flexibility, yet none is suitable here? Just wondering about the use case. I suspect using an "out-of-manifest" Zephyr module causes other complications beyond this one?

scripts/west_commands/sign.py
```diff
@@ -554,11 +554,14 @@ def sign(self, command, build_dir, build_conf, formats):
...
-        try:
-            sof_proj = command.manifest.get_projects(['sof'], allow_paths=False)
-            sof_src_dir = pathlib.Path(sof_proj[0].abspath)
-        except ValueError: # sof is the manifest
-            sof_src_dir = pathlib.Path(manifest.manifest_path()).parent
+        if os.getenv("SOF_SRC_DIR"):
+            sof_src_dir = pathlib.Path(os.getenv("SOF_SRC_DIR"))
+        else:
+            try:
+                sof_proj = command.manifest.get_projects(['sof'], allow_paths=False)
```
marc-hb: Since 1a780f933e1da387a this is dead code, no point keeping it.

EDIT: I mean dead code unless sof/west.yml downloads... SOF recursively :-D

EDIT2: correction: or, someone combines zephyr/west.yml with some submanifest that downloads SOF. Looks like I got lost again in `west` infinite possibilities...

marc-hb (commented):
I just remembered there is already `west config rimage.path ...` and also `west sign --tool-path ...`.

Considering rimage has been merged back into sof a long time ago, maybe `sof_src_dir` could be automagically inferred from one of these when not found in the manifest, avoiding one extra configuration item.

If one new configuration item can really not be avoided, then `west config ...` is more flexible, IMHO more convenient and a new environment variable and more consistent with where `rimage` configuration already is.

My 2 cents.

marc-hb (commented):
No issue with `cmake_cache.get('RIMAGE_CONFIG_PATH')`, but the logic still does not feel right. For a very long time this code has been:

```python
try:
  to find SOF project in [zephyr/]west.yml
  ...
except: # not found, so assume that the manifest is sof/west.yml      
   ...
```

Now, SOF has been removed from zephyr/west.yml. But instead of changing the `try` clause, this PR changes the `except` clause?!

After this PR, the code looks like:

```python
try:
  to find something that will never be found! # changed location, unchanged code
  ...
except:
  do something that might work in both zephyr/west.yml and sof/west.yml cases?
```

This PR might work because your new `except` clause might work in both cases. But it's weird and confusing now.

BTW don't attempt to preserve the `try: except:` at all costs. An if/then/else might be enough now.

dcpleung:
Change the discovery logic a bit to cover possible scenarios.

scripts/west_commands/sign.py
```diff
@@ -554,13 +558,56 @@ def sign(self, command, build_dir, build_conf, formats):
...
+        # Try to find SOF directory via path to rimage config files
+        conf_dir = self.cmake_cache.get('RIMAGE_CONFIG_PATH')
+        if conf_dir:
+            # Config path is <sof top>/tools/rimage/config/
+            # So need to go up 3 levels to get SOF directory.
+            maybe_sof_dir = pathlib.Path(conf_dir).parent.parent.parent
+
+            # Make sure this is the actual SOF directory as
+            # RIMAGE_CONFIG_PATH may point to somewhere else.
+            versions_json_file = maybe_sof_dir / 'versions.json'
```
marc-hb: SonarCloud is not far off: this entire `versions_json_file.exists()` test is repeated 3 times. This is calling for some new test function.

Speaking of functions, the length of this PR (assuming it is justified) is now calling for some new `def get_sof_dir()` function too that get just `return` as soon as it finds something.

marc-hb (commented):
Wow, this increases the number of ways to find the SOF git clone to four now! But we have only two different, known situations to deal with: `zephyr/west.yml` and `sof/west.yml`? Four different solutions for two situations? 

This `rimage` area has alway been difficult to maintain because of the too many different combinations and possibilities; merely reading the (already too long) `sign()` method is enough to get the corresponding headache: can't touch anything because "you never know". Ideally, that `sign` method should support fewer variations, not more. In the meantime, each way to find the SOF git clone should at least tell in a source comment what specific situation it is supposed to deal with.


BTW do you know how having `rimage` in the path affects https://github.com/zephyrproject-rtos/zephyr/issues/99497#issuecomment-3541662097 ?

> Twister does not take west config variable so we still need a way to specific where SOF is, which is via environment variable SOF_SRC_DIR.

I don't get this: why does twister even need to know about `west config`? Both the environment variables and the `west config` approaches have the same advantage of "cutting through" the layer of indirections _without them even knowing about it_.
It's not like `sof_src_dir` is something dynamic that depends on the platform or test case.

dcpleung:
The main reason for needing to find the SOF directory is simply the need to run the UUID generation script.

1. Having the SOF pulled in via sub-manifest has been working fine before we remove SOF from the the manifest. We are still doing this in our internal CI for testing so this method needs to stay.
2. The SOF main repo, on the other hand, is pulling Zephyr as a project. So we need to use the root manifest path to find the SOF tree.
3. Using `west config` to specify `RIMAGE_IMAGE_PATH` is mentioned in the documentation. So we have to keep this method to avoid breaking someone's workflow.
4. However, using `west config` to specify the config path does not work with twister. That config is never applied during twister builds. So we need another way to specify where to find SOF tree to enable running twister to build and test on hardware.

Whether rimage is in the path or specified via `west config rimage.path`, the results are the same: unable to run the UUID generation script, and thus failing the signing process.

boards/intel/adsp/doc/intel_adsp_generic.rst
```diff
@@ -60,22 +60,26 @@ Signing Tool
...
-   cd zephyrproject
-   west config manifest.project-filter -- +sof
-   west update
-   cd modules/audio/sof/tools/rimage
+Clone the SOF repository and follow the instructions in the rimage :file:`README.md`
+to build the tool on your system.
 
-Follow the instructions in the rimage :file:`README.md` to build the tool on
-your system. You can either copy the executable to a directory in your PATH or
+You can either copy the executable to a directory in your PATH or
```
marc-hb, line 66: This section overlaps and partially duplicates `doc/develop/west/sign.rst`. I don't have an opinion on which one is the best place, but signing should consolidated in a single place out of these two. The other place should just refer to the first page. The number of configuration possibilities is already too high; the last thing needed is having 70% of them documented in one place and some other 60% in another. I think this has been part of the confusion.

scripts/west_commands/sign.py
```diff
@@ -423,6 +423,75 @@ def edt_flash_params(cmd, flash):
...
+                sof_src_dir = pathlib.Path(sof_proj[0].abspath)
+
+                # Since SOF is pulled in as a project, we assume it is
+                # the correct one as the manifest has to be modified
+                # manually.
+            except ValueError:
+                pass
+
+        # Try to find SOF as SOF is the top level manifest via west init.
+        if not sof_src_dir:
```
marc-hb, line 477: Now that you have a `get_sof_src_dir()` function, you can `return` as soon as you find something suitable and drop all the `if not sof_src_dir`

scripts/west_commands/sign.py
```diff
@@ -423,6 +423,75 @@ def edt_flash_params(cmd, flash):
...
+
+        # Try to find SOF directory via path to rimage config files pointed to
+        # via CMake cached variable "RIMAGE_CONFIG_PATH" if it exists.
+        # The config files are inside <sof top>/tools/tools/rimage/config in
+        # the SOF tree.
+        conf_dir = self.cmake_cache.get('RIMAGE_CONFIG_PATH')
+        if conf_dir:
+            # Config path is <sof top>/tools/rimage/config/
+            # So need to go up 3 levels to get SOF directory.
+            maybe_sof_dir = pathlib.Path(conf_dir).parent.parent.parent
```
marc-hb, line 457: Looking at how `RIMAGE_CONFIG_PATH` is used and documented, it is clearly intended  as a way to use _alternative_ `.toml` files, away from the default ones found in SOF. Typically: for internal products still embargoed. So, (ab)using `RIMAGE_CONFIG_PATH` to find SOF does not look like a good idea.  With this code, we have both RIMAGE_CONFIG_PATH falling back on sof_src_dir, and - the other way round -  sof_src_dir falling back on RIMAGE_CONFIG_PATH?! First come first serve?

If anything, the precedence of this should be last or close to last.

marc-hb (commented):
Thanks for the list! I think these rationales should be recorded in the documentation and briefly mentioned in the source code so this maintenance nightmare does not spiral even more out of control than it already does. I mean not just documenting how things can be used in the present but also _rationales_ / reasons why things exist, together with some history. 

EDIT: also, please make it clearer in which use cases _only_ rimage is used and SOF is _not_ used. Now that they are back together in the same repo it's easy to mix up the use cases (I have been confused).

> 4. However, using west config to specify the config path does not work with twister. That config is never applied during twister builds. 

Sorry I don't understand what "applied" means and I still don't see what a new envvar method brings to the table something that could not already be done before.

Before running twister, whatever or whoever runs twister can run: `west config rimage.path /path/to/rimage` on their system once for all. Then, this sign.py file will find rimage whether twister is involved or not - twister and CMake do not matter there. Adding a new environment variable seems to merely duplicate this approach in a less transparent way (cause envvars are more volatile and harder to trace and troubleshoot than configuration files) while adding even more combinatorial explosion.



> The main reason for needing to find the SOF directory is simply the need to run the UUID generation script.

That's the most recent reason but searching for "sof_src_dir" shows older ones:
- for the default .toml location
- for the default key location
- for running CPP on .toml files



By the way this text from @andyross from commit 00417b36bc7e is still present:

```
+# The short version is that zephyr.ri can only realistically be
+# exercised from a SOF build and it doesn't belong here in Zephyr;
+# rimage is a SOF tool.  Signing audio firmware is only done for SOF
+# firmware and not general Zephyr apps for the same hardware. This
+# should live in SOF where it doesn't have to be duplicated for every
+# device and where it won't be forced to communicate via side channels.
```

It seems to directly contradict this PR and should probably be corrected.

dcpleung:
Use https://github.com/zephyrproject-rtos/zephyr/pull/103777 instead.

PR #99542, opened 2025-11-18
drivers: rcc: Improve the RCC definition for the sf32lb platform

HalfSweet:
Improve the RCC definition for the sf32lb platform.
All RCC configurations in sf32lb have been completed, with the following additions:
- sys-clk-src
- peri-clk-src
- mpi1/2-clk-src
- usb-clk-src
- usb-div

### DIFF
--- a/drivers/clock_control/clock_control_sf32lb_rcc.c
+++ b/drivers/clock_control/clock_control_sf32lb_rcc.c
@@ -5,6 +5,7 @@
 
 #define DT_DRV_COMPAT sifli_sf32lb_rcc_clk
 
+#include <stdbool.h>
 #include <stdint.h>
 
 #include <zephyr/arch/cpu.h>
@@ -13,6 +14,7 @@
 #include <zephyr/drivers/clock_control/sf32lb.h>
 #include <zephyr/dt-bindings/clock/sf32lb-clocks-common.h>
 #include <zephyr/dt-bindings/clock/sf32lb52x-clocks.h>
+#include <zephyr/sys/util.h>
 #include <zephyr/toolchain.h>
 
 #include <register.h>
@@ -23,14 +25,9 @@
 
 #define HPSYS_RCC_CSR    offsetof(HPSYS_RCC_TypeDef, CSR)
 #define HPSYS_RCC_CFGR   offsetof(HPSYS_RCC_TypeDef, CFGR)
+#define HPSYS_RCC_USBCR  offsetof(HPSYS_RCC_TypeDef, USBCR)
 #define HPSYS_RCC_DLL1CR offsetof(HPSYS_RCC_TypeDef, DLL1CR)
 #define HPSYS_RCC_DLL2CR offsetof(HPSYS_RCC_TypeDef, DLL2CR)
-
-#define HPSYS_RCC_CSR_SEL_SYS_CLK_HXT48  FIELD_PREP(HPSYS_RCC_CSR_SEL_SYS_Msk, 1U)
-#define HPSYS_RCC_CSR_SEL_SYS_CLK_DLL1   FIELD_PREP(HPSYS_RCC_CSR_SEL_SYS_Msk, 3U)
-#define HPSYS_RCC_CSR_SEL_MPI1_CLK_DLL2  FIELD_PREP(HPSYS_RCC_CSR_SEL_MPI1_Msk, 2U)
-#define HPSYS_RCC_CSR_SEL_MPI2_CLK_DLL2  FIELD_PREP(HPSYS_RCC_CSR_SEL_MPI2_Msk, 2U)
-#define HPSYS_RCC_CSR_SEL_PERI_CLK_HXT48 FIELD_PREP(HPSYS_RCC_CSR_SEL_PERI_Msk, 1U)
 
 #define HPSYS_RCC_DLLXCR_EN          HPSYS_RCC_DLL1CR_EN
 #define HPSYS_RCC_DLLXCR_STG_Msk     HPSYS_RCC_DLL1CR_STG_Msk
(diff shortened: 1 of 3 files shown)


ck-telecom (dismissed):

ck-telecom:
```
drivers: clock_control: sf3232lb_rcc: Add RCC configuration driver
Add RCC configuration driver.

Signed-off-by: Haoran Jiang <halfsweet@halfsweet.cn>
```
fix typo and commit format, also rebase and resovle conflicts.

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,54 @@ properties:
...
+      ``peri`` if not provided.
+
+  sifli,mpi2-clk-src:
+    type: string
+    enum: ["dll1", "dll2", "peri"]
+    default: "peri"
+    description: |
+      Source used for the MPI2/QSPI clock tree. Selecting ``dll1`` or ``dll2``
+      requires the corresponding DLL child node to be enabled. Defaults to
+      ``peri`` if not provided.
```
gmarull: defaults need to be justified (e.g. boot time default) per bindings rules

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,54 @@ properties:
...
+      not provided.
+
+  sifli,mpi1-clk-src:
+    type: string
+    enum: ["dll1", "dll2", "peri"]
+    default: "peri"
+    description: |
+      Source used for the MPI1/QSPI clock tree. Selecting ``dll1`` or ``dll2``
+      requires the corresponding DLL child node to be enabled. Defaults to
+      ``peri`` if not provided.
```
gmarull: same

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,54 @@ properties:
...
+    description: |
+      Source used for the HPSYS system clock. When set to ``dll1`` the DLL1 child
+      node must be enabled. Defaults to ``hxt48`` if not provided.
+
+  sifli,peri-clk-src:
+    type: string
+    enum: ["hrc48", "hxt48"]
+    description: |
+      Source used for the HPSYS peripheral clock domain. Defaults to ``hxt48`` if
+      not provided.
```
gmarull: same, default actually not set

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,54 @@ properties:
     description: |
       Divider for PCLK2 (0-7). fPCLK2 = fHCLK / 2^PDIV2.
 
+  sifli,sys-clk-src:
+    type: string
+    enum: ["hrc48", "hxt48", "lpclk", "dll1"]
+    description: |
+      Source used for the HPSYS system clock. When set to ``dll1`` the DLL1 child
+      node must be enabled. Defaults to ``hxt48`` if not provided.
```
gmarull: same, default not set?

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,54 @@ properties:
...
+      requires the DLL2 child node to be enabled. Defaults to ``sysclk`` if not
+      provided.
+
+  sifli,usb-div:
+    type: int
+    default: 4
+    description: |
+      Divider applied to the selected USB clock source. The divided clock must
+      be 60 MHz. For example, when ``clk_dll2`` runs at 240 MHz the divider must
+      be ``4``.
```
gmarull: Default should not be set if its value depends on how other things are configured.

gmarull: Also, why isn't this computed automatically if target clock freq is known?

drivers/clock_control/clock_control_sf32lb_rcc.c
```diff
@@ -23,29 +25,157 @@
...
+enum sf32lb_mpi_clk_idx {
+	SF32LB_MPI_CLK_IDX_DLL1,
+	SF32LB_MPI_CLK_IDX_DLL2,
+	SF32LB_MPI_CLK_IDX_PERI,
+};
+
+enum sf32lb_usb_clk_idx {
+	SF32LB_USB_CLK_IDX_SYSCLK,
+	SF32LB_USB_CLK_IDX_DLL2,
+};
```
gmarull, line 71: since this seems to mimic DT enums, you may consider using this approach (so everything resides on a DT header): https://wiki.st.com/stm32mpu/wiki/Clock_device_tree_configuration (ref `st,clksrc`). I think it's better.

gmarull (commented):
good job on extending this, see my comments

gmarull:
note, for future this is also needed:

image

It's weird it's controlled by RTC/PMUC, maybe RTC/PMUC need a clock control sub-node, and a separate driver can manage that, and then nodes like RTC would have `clocks = <&rcc_clk XXX_ENABLE_CLOCK>, <&pmuc_clk XXX_ACTUAL_CLOCK>`?

drivers/clock_control/clock_control_sf32lb_rcc.c
HalfSweet: I have a slightly different perspective on this matter... While listing all data in the header file does significantly simplify the driver code, it makes editing the device tree less intuitive and prone to errors (such as misconfiguring the corresponding macro definitions).
Is there a recommended best practice for handling this scenario in Zephyr?

gmarull: I don’t think there’s really a standard here, but it’s a common thing to look into other projects using devicetree to see how they have solved the same or similar problems.

HalfSweet: After internal discussion, we concluded that using strings as enumerations would be preferable
I adjusted the order of the enumeration values so that each enum matches the final register setting value, making the driver code easier to read

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
HalfSweet: done. all settings have been assigned default values and corresponding descriptions.

ck-telecom (approved):

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
```diff
@@ -65,5 +65,56 @@ properties:
...
+      requires the DLL2 child node to be enabled. Defaults to ``sysclk`` if not
+      provided (reset default).
+
+  sifli,usb-div:
+    type: int
+    default: 4
+    description: |
+      Divider applied to the selected USB clock source. The divided clock must
+      be 60 MHz. For example, when ``clk_dll2`` runs at 240 MHz the divider must
+      be ``4`` (reset default).
```
gmarull, line 117: as said, providing default here may not be ideal if clk_dll2 is configured to other frequencies.

HalfSweet: Calculating the USB clock source and frequency division during compilation in C is simply too complex. Perhaps the correct allocation of the clock tree should be ensured by the person writing the device tree—when they enable the USB module, the corresponding frequency division should already be calculated. Of course, if it's not enabled, nothing happens. The default value of 4 here is merely because that's the default value after startup.

gmarull (approved):

dts/bindings/clock/sifli,sf32lb-rcc-clk.yaml
gmarull: fair enough

PR #99546, opened 2025-11-18
ci: compliance: Lower log level for non-diff issues in DTS linter

### LINKED ISSUE #99299: Compliance: dts-linter: only check dts diff
### Summary

The newly added dts-linter may analyze the entire DTS content, which is valid since [dts-linter](https://github.com/kylebonnici/dts-linter) created by @kylebonnici requires full context for accurate diagnostics. However, this behavior can block PRs that touch problematic DTS files whose fixes haven't been merged yet.

Analyze the DTS diff and linter result, then report only the issues found in the diff.

yvesll:
Fixes #99299
For issues reported by dts-linter, if they are not in current COMMIT RANGE, reduce the log level to warning.

### DIFF
--- a/scripts/ci/check_compliance.py
+++ b/scripts/ci/check_compliance.py
@@ -116,6 +116,28 @@ def get_files(filter=None, paths=None):
     return files
 
 
+def get_modified_lines(path):
+    """
+    Get the set of modified line numbers for a specific file in the COMMIT_RANGE.
+    :param path: Path to the file to check for modifications
+    :return: Set of modified line numbers in the file
+    """
+    out = git('diff', '-U0', COMMIT_RANGE, '--', path)
+    lines = out.splitlines()
+    modified_lines = set()
+
+    for line in lines:
+        if not line.startswith('@@'):
+            continue
+        match = re.match(r'@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@', line)
+        if match:
+            start = int(match.group(1))
+            count = int(match.group(2)) if match.group(2) else 1
+            modified_lines.update(range(start, start + count))
+
+    return modified_lines
+
+
 def get_module_setting_root(root, settings_file):
     """
     Parse the Zephyr module generated settings file given by 'settings_file'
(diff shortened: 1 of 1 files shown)


yvesll:
If multiple PRs modify a DTS file with compliance issues, one PR needs to handle the cleanup, which will block the others. After the cleanup PR is merged, the remaining PRs must resolve conflicts if they also applied fixes.
Several of my PRs with longer review cycles have encountered this problem

yvesll:
> One point I would like to bring up is that the dts-linter will not format a file with syntax issues. Hence we have to be mindful that if file is broken and this issue seams to be out of the commit range, then then this may lead one to believe that no formatting is needed + file will remain broken.

Yeah, two of my open PRs failed due to this syntax error before 😄 , that's why I open this topic.
```c
	quadspi_ardb: @68000000 {
		compatible = "nxp,";
		reg = <0x68000000 0x21c>;
		status = "disabled";
	};
```

scripts/ci/check_compliance.py
```diff
@@ -606,7 +636,7 @@ def run(self):
         if not self.ensure_npx():
             self.skip(
                 'dts-linter not installed. To run this check, '
-                'install Node.js and then run [npm ci] command inside ZEPHYR_BASE'
+                'install Node.js and then run [npm --prefix ./scripts/ci ci] command inside ZEPHYR_BASE'
```
yvesll: At least the hint command shall be fixed. `npm ci` failed for me as no `package.json` under `ZEPHYR_BASE`.

yvesll: Move this to #100334

fabiobaltieri:
> Yes, I have do a global check on main branch, here is a summary:
> 
> * Total directories checked: 233
> * Total DTS/DTSI files checked: 2107
> * **Directories with errors: 15**
> * **Files with errors: 408**

Ok but then how about we get in the habit of keeping everything in sync every time the syntax rules are changed? It really should be VERY rare once the ruleset is stable, I'd imagine it'll only take another two or three rounds to get to something stable.

The problem with introducing these "force only on change code" check is that they enforce making files inconsistent between themselves, if it was optional one could ignore it and at least keep the file consistent, if they are per whole file they get very annoying but again the styling is consistent, in this situation one must do a partial change, and can't even use the automation to do it because the automation works on the whole file, it's just the worse of both worlds.

PR #99548, opened 2025-11-18
drivers: input: Support invert x/y in rel mode

petejohanson:
Despite the datasheet stating otherwise, the Pinnacle based trackpads can have their X/Y data inverted when operating in relative mode, so set those configs in all modes during init.

### DIFF
--- a/drivers/input/input_pinnacle.c
+++ b/drivers/input/input_pinnacle.c
@@ -819,13 +819,16 @@ static int pinnacle_init(const struct device *dev)
 	value = PINNACLE_FEED_CONFIG1_FEED_ENABLE;
 	if (!config->relative_mode) {
 		value |= PINNACLE_FEED_CONFIG1_DATA_MODE_ABSOLUTE;
-		if (config->invert_x) {
-			value |= PINNACLE_FEED_CONFIG1_X_INVERT;
-		}
-		if (config->invert_y) {
-			value |= PINNACLE_FEED_CONFIG1_Y_INVERT;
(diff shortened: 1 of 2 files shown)


fabiobaltieri (dismissed):
Not too enthusiastic about adding a dependency to a feature that is documented as unsupported, who knows if there's some firmware version where it doesn't, though these have been on the market for a long time, maybe the feature was off the early version and it's fine to have it now.

drivers/input/input_pinnacle.c
```diff
@@ -819,13 +819,16 @@ static int pinnacle_init(const struct device *dev)
...
 		value |= PINNACLE_FEED_CONFIG1_DATA_MODE_ABSOLUTE;
-		if (config->invert_x) {
-			value |= PINNACLE_FEED_CONFIG1_X_INVERT;
-		}
-		if (config->invert_y) {
-			value |= PINNACLE_FEED_CONFIG1_Y_INVERT;
-		}
 	}
+
+	/* Datasheet states these are absolute only, but the also work in rel mode */
```
akscram: Typo: the -> they

akscram (approved):
We should also update descriptions of `invert-x` and `invert-y` properties defined in `dts/bindigs/input/cirque,pinnacle-common.yaml` because they mention only the absolute mode right now. Other than that, looks good! Thanks

fabiobaltieri (approved):

faxe1008 (approved):

PR #99549, opened 2025-11-18
drivers: mipi_dbi: add mipi dbi driver for sf32lb

ck-telecom:
add mipi dbi driver for sf32lb(spi & 8080)


rabbitsaviola:
@ck-telecom could you please add label `platform: SF32LB` for tracking?

VynDragon:
Hello, I think you will need to add a test build to a board or do like I did for BL61x if you dont have a board with space to put it in.

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -106,6 +106,15 @@
 			status = "disabled";
 		};
 
+		lcdc: lcdc@50008000 {
+			compatible = "sifli,sf32lb-mipi-dbi";
```
gmarull: lcdc is not strictly a mipi-dbi peripheral (it supports other interfaces as JDI), so this DT representation does not look correct. Maybe leave compatible unset at dts level, and let board configure it, depending on the connected display (see mpi1/2 nodes).

drivers/mipi_dbi/mipi_dbi_sf32lb.c
```diff
@@ -0,0 +1,537 @@
...
+	te_conf |= FIELD_PREP(LCD_IF_TE_CONF_FMARK_MODE_Msk, 1) |
+		   FIELD_PREP(LCD_IF_TE_CONF_ENABLE_Msk, 1);
+
+	sys_write32(delay_us, config->base + TE_CONF2);
+	sys_write32(te_conf, config->base + TE_CONF);
+
+	return 0;
+}
+
+static const struct mipi_dbi_driver_api dbi_sf32lb_api = {
```
JarmouniA:
```suggestion
static DEVICE_API(mipi_dbi, dbi_sf32lb_api) = {
```

JarmouniA (commented):
Needs to be built in CI, similar to eaf743e48b4347563d7dd2620eee75d1899e0276

tests/drivers/build_all/display/testcase.yaml
```diff
@@ -12,6 +12,7 @@ tests:
       - native_sim/native/64
     platform_exclude:
       - ai_m62_12f_kit
+      - sf32lb52_devkit_lcd
```
JarmouniA:
```suggestion
```

JarmouniA (commented):
Last commit removing Bluetooth testing needs further justification.

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
```diff
@@ -0,0 +1,40 @@
...
+				 <PA08_LCDC1_8080_DIO1>,
+				 <PA37_LCDC1_8080_DIO2>,
+				 <PA39_LCDC1_8080_DIO3>,
+				 <PA40_LCDC1_8080_DIO4>,
+				 <PA41_LCDC1_8080_DIO5>,
+				 <PA42_LCDC1_8080_DIO6>,
+				 <PA43_LCDC1_8080_DIO7>,
+				 <PA02_LCDC1_8080_TE>;
+			input-enable;
+		};
```
gmarull, line 23: shield?

boards/sifli/sf32lb52_devkit_lcd/sf32lb52_devkit_lcd.yaml
```diff
@@ -14,4 +14,5 @@ supported:
   - gpio
   - watchdog
   - bluetooth
+  - display
```
gmarull: not true, please add a shield

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,14 @@
 			};
 		};
 
+		lcdc: lcdc@50008000 {
+			reg = <0x50008000 0x1000>;
```
gmarull, line 125: add comment about missing compatible, see eg MPI

dts/bindings/mipi-dbi/sifli,sf32lb-mipi-dbi.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2025, Qingsong Gou <gouqs@hotmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+description: SF32LB MIPI DBI controller
+
+compatible: "sifli,sf32lb-mipi-dbi"
```
gmarull: `sifli,sf32lb-lcdc-mipi-dbi`?

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
ck-telecom: currently using a private board to test this driver, and interface not compatible with board sf32lb52_devkit_lcd

boards/sifli/sf32lb52_devkit_lcd/sf32lb52_devkit_lcd.yaml
ck-telecom: private board, no official shield available for 8080 interface

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
gmarull: then it cannot be part of upstream...

ck-telecom: The code for test build has removed

JarmouniA: > shield?

We just want to build the MIPI-DBI controller driver in CI, not test it or test a specific display controller (MIPI-DBI device), so doing it without a shield is fine in my opinion.

boards/sifli/sf32lb52_devkit_lcd/sf32lb52_devkit_lcd.yaml
JarmouniA: Yes please remove the tag, it's reserved for boards with built-in displays.

VynDragon: How will the build happen if it's not tagged?

JarmouniA: > How will the build happen if it's not tagged?

Not sure what you're asking.

VynDragon: Will the build_all trigger for this driver without the display tag on the board supposed to be used for the driver build test? Eg like the bl61x one: https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/aithinker/ai_m62_12f_kit/ai_m62_12f_kit.yaml ? Or is the bl61x one is doing it wrong (it has no display on the board either).

JarmouniA: > Will the build_all trigger for this driver without the display tag on the board supposed to be used for the driver build test?

They are unrelated. See the description of `Supported` in https://docs.zephyrproject.org/latest/develop/test/twister.html#board-configuration

> Or is the bl61x one is doing it wrong (it has no display on the board either).

Should be removed from that board as well.

JarmouniA (commented):
Please restore the Build All test conf. See https://github.com/zephyrproject-rtos/zephyr/pull/99549#discussion_r2631967946

boards/sifli/sf32lb52_devkit_lcd/sf32lb52_devkit_lcd.yaml
JarmouniA: > How will the build happen if it's not tagged?

Yeah, now I see what you mean, wasn't aware the test depends on the `display` tag
https://github.com/zephyrproject-rtos/zephyr/blob/64efa7dcb3eb2a119cbc5832b69fd263aa1d4d67/tests/drivers/build_all/display/testcase.yaml#L6

It should be moved inside `drivers.display.build.default` test scenario.

VynDragon: Aight, will look into it.

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,18 @@
 			};
 		};
 
+		lcdc: lcdc@50008000 {
+			/*
+			 * configure compatible depending on interface type, choices:
+			 * - sifli,sf32lb-lcdc-mipi-dbi
```
JarmouniA: If there is only one choice why not set it!?

tests/drivers/build_all/display/testcase.yaml
```diff
@@ -20,3 +20,7 @@ tests:
   drivers.display.build.bflb_dbi:
     platform_allow:
       - ai_m62_12f_kit
+
+  drivers.display.build.sf32lb_dbi:
+    platform_allow:
+      - sf32lb52_devkit_lcd
```
JarmouniA, line 20: needs an accompanying overlay that enables the controller, otherwise its driver cannot be built.

dts/arm/sifli/sf32lb52x.dtsi
ck-telecom: JDI interface is not support now, start from downstream

dts/arm/sifli/sf32lb52x.dtsi
ck-telecom: @JarmouniA changed from here

dts/arm/sifli/sf32lb52x.dtsi
JarmouniA: JDI?

ck-telecom: > lcdc is not strictly a mipi-dbi peripheral (it supports other interfaces as JDI), so this DT representation does not look correct. Maybe leave compatible unset at dts level, and let board configure it, depending on the connected display (see mpi1/2 nodes).

JarmouniA: I'm not aware of any display interface named JDI, what I know is that JDI is a display vendor.
If the controller supports multiple protocols, make a binding for the base device, with no protocol-specific properties, then mipi-dbi node will be its child node in DT. The DBI driver will then get the base info like clocks/reg from the parent node.

JarmouniA: #99863 is doing a similar thing.
https://github.com/zephyrproject-rtos/zephyr/blob/0282ccd311f59a7f0df2e686d3768e3d08d31c6e/dts/xtensa/espressif/esp32s3/esp32s3_common.dtsi#L395

ck-telecom: If define a parent node, acctually there is no functional driver for this compatible, just a place holder, determine at board level should be fine.

JarmouniA: > acctually there is no functional driver for this compatible

That's not an issue.

> determine at board level should be fine.

No, having multiple exclusive compatibles for the same device node is not fine.

JarmouniA (dismissed):

dts/bindings/mipi-dbi/sifli,sf32lb-lcdc-mipi-dbi.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2025, Qingsong Gou <gouqs@hotmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+description: SF32LB MIPI DBI controller
```
JarmouniA: Could you mention here that a node with this compatible shall be a child node of a "sifli,sf32lb-lcdc" compatible node?

dts/bindings/mipi-dbi/sifli,sf32lb-lcdc-mipi-dbi.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2025, Qingsong Gou <gouqs@hotmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  SF32LB LCDC MIPI DBI controller, this node with this compatible shall be a
```
JarmouniA:
```suggestion
  SF32LB LCDC MIPI DBI controller.
  A node with this compatible shall be a
```

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
```diff
@@ -0,0 +1,35 @@
...
+	};
+};
+
+&lcdc {
+	status = "okay";
+
+	mipi_dbi: mipi_dbi {
```
JarmouniA:
```suggestion
	mipi_dbi: mipi-dbi {
```

JarmouniA (dismissed):

JarmouniA (commented):
Nits: 1s commit msg: replace "node" with "binding", 2nd commit msg: replace "bingdings" with "binding".

drivers/mipi_dbi/mipi_dbi_sf32lb.c
```diff
@@ -0,0 +1,537 @@
...
+
+struct dbi_sf32lb_data {
+	const struct mipi_dbi_config *active_config;
+};
+
+static inline void wait_busy(const struct device *dev)
+{
+	const struct dbi_sf32lb_config *config = dev->config;
+
+	while (sys_test_bit(config->base + LCD_SINGLE, LCD_IF_LCD_SINGLE_LCD_BUSY_Pos) ||
```
VynDragon, line 48: Nit: Would recommend some kind of timeout system in case something somewhere gets stuck.

VynDragon (dismissed):

JarmouniA (dismissed):

drivers/mipi_dbi/mipi_dbi_sf32lb.c
ck-telecom: Will provide another PR fix this

dts/arm/sifli/sf32lb52x.dtsi
gmarull: > > acctually there is no functional driver for this compatible
> 
> That's not an issue.
> 
> > determine at board level should be fine.
> 
> No, having multiple exclusive compatibles for the same device node is not fine.

This approach has been used multiple times in other SoCs for peripherals that can work in multiple different modes, but only one at a time.

gmarull: > I'm not aware of any display interface named JDI, what I know is that JDI is a display vendor. If the controller supports multiple protocols, make a binding for the base device, with no protocol-specific properties, then mipi-dbi node will be its child node in DT. The DBI driver will then get the base info like clocks/reg from the parent node.

This interface is used in multiple MiP displays (initially designed by JDI), now mainly Sharp.

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,20 @@
...
+			reg = <0x50008000 0x1000>;
+			#address-cells = <1>;
+			#size-cells = <0>;
+			clocks = <&rcc_clk SF32LB52X_CLOCK_LCDC1>;
+			status = "disabled";
+
+			mipi-dbi {
+				compatible = "sifli,sf32lb-lcdc-mipi-dbi";
+				status = "disabled";
+			};
```
gmarull, line 132: please, use original proposal. This device cannot work on multiple modes in parallel.

dts/arm/sifli/sf32lb52x.dtsi
JarmouniA: > This approach has been used multiple times in other SoCs for peripherals that can work in multiple different modes, but only one at a time.

There is really no need for it. With the child-node approach, everything can be defined in DT, the user does not have to figure out how to configure things on its own. And the n-drivers for the same device can be made exclusive at the Kconfig level.

dts/arm/sifli/sf32lb52x.dtsi
JarmouniA: What determines which protocol (child node device) is being used is which driver is enabled, and rn we only have one. When there are multiple, we will make them exclusive at the Kconfig level.

gmarull (changes requested):
Please, update DT part

dts/arm/sifli/sf32lb52x.dtsi
danieldegrasse: If we are going to have multiple protocols supported, why wouldn't we do it with multiple compatibles for the parent device? This solution has been used many times in Zephyr, and it kind of seems like the right approach here (if I'm following the issue right)

gmarull: To my understanding, fixing compatible to one of the IP "modes of operation" is not a bad DT practice. Spec says:

image

`define the specific programming model for the device`

JarmouniA: @gmarull You still haven't answered my question, what's the alleged other mode of operation for this IP?
Also, compatibles go with properties (and DT bus in the case of MIPI-DBI), so it's not like we just swap compatibles and we are done.

JarmouniA: > If we are going to have multiple protocols supported, why wouldn't we do it with multiple compatibles for the parent device? This solution has been used many times in Zephyr, and it kind of seems like the right approach here (if I'm following the issue right)

@danieldegrasse Please see my comment https://github.com/zephyrproject-rtos/zephyr/pull/99549#issuecomment-3682910573

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,18 @@
 			};
 		};
 
+		lcdc: lcdc@50008000 {
+			/*
+			 * configure compatible depending on memory type, choices:
+			 * - sifli,sf32lb-lcdc-mipi-dbi
```
JarmouniA: There is only one operating mode supported for this device currently, so this approach is not justified. Also, compatibles go with properties, and a DT bus is the case of MIPI-DBI, so they cannot be swapped.

JarmouniA (changes requested):

dts/arm/sifli/sf32lb52x.dtsi
gmarull: > @gmarull You still haven't answered my question, what's the alleged other mode of operation for this IP? Also, compatibles go with properties (and DT bus in the case of MIPI-DBI), so it's not like we just swap compatibles and we are done.

It can drive multiple display types: DBI/8080, SPI/DSPI/QSPI, and JDI.

Another option would be to have the LCDC node, then, a single child node (added at board level, not SoC) which specifies which PHY is used:

image

We could even have a common driver for the non-PHY parts I guess this way.

JarmouniA: > it can drive multiple display types: DBI/8080, SPI

DBI/8080, SPI are all modes of MIPI-DBI.
https://github.com/zephyrproject-rtos/zephyr/blob/4f2ee86cd44863a9527b91485cbef01d4c711e4d/include/zephyr/drivers/mipi_dbi.h#L12

> and JDI.

No display panel controller that supports it exists currently in Zephyr AFAIK.

> Another option would be to have the LCDC node, then, a single child node (added at board level, not SoC) which specifies which PHY is used.

That would be fine.

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,23 @@
...
+			compatible = "sifli,sf32lb-lcdc";
+			reg = <0x50008000 0x1000>;
+			#address-cells = <1>;
+			#size-cells = <0>;
+			clocks = <&rcc_clk SF32LB52X_CLOCK_LCDC1>;
+			status = "disabled";
+
+			mipi-dbi {
+				/*
+				* configure compatible depending on memory type, choices:
```
JarmouniA:
```suggestion
				* configure compatible depending on the MIPI-DBI mode, choices:
```

JarmouniA (dismissed):

JarmouniA (dismissed):

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -120,6 +120,23 @@
...
+			#address-cells = <1>;
+			#size-cells = <0>;
+			clocks = <&rcc_clk SF32LB52X_CLOCK_LCDC1>;
+			status = "disabled";
+
+			mipi-dbi {
+				/*
+				 * configure compatible depending on the MIPI-DBI mode, choices:
+				 * - sifli,sf32lb-lcdc-mipi-dbi
+				 */
```
gmarull: this doesn't make sense: node is explicitly for mipi-dbi, so compatible can be set

drivers/mipi_dbi/mipi_dbi_sf32lb.c
```diff
@@ -0,0 +1,537 @@
...
+
+		while (data_len > 0) {
+			uint32_t v, l;
+
+			/* Convert 0xAA,0xBB,0xCC ->  0x00AABBCC */
+			for (v = 0, l = 0; l < 4; l++) {
+				v = (v << 8) | (*buf);
+				data_len--;
+				buf++;
+			}
```
gmarull, line 440: byteops.h APIs?

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
```diff
@@ -0,0 +1,38 @@
...
+				 <PA02_LCDC1_8080_TE>;
+			input-enable;
+		};
+	};
+};
+
+&lcdc {
+	status = "okay";
+
+	mipi_dbi: mipi-dbi {
```
gmarull, line 30: since this is now using a child to select mode of operation, please, make sure to add a build assert on the driver that forces to only have a single child node enabled, something like:

```
BUILD_ASSERT((DT_CHILD_NUM_STATUS_OKAY(DT_INST_PARENT(n)) == 1), "LCDC only supports one operating mode");
```

dts/arm/sifli/sf32lb52x.dtsi
JarmouniA: There is mipi-dbi-spi too.

drivers/mipi_dbi/mipi_dbi_sf32lb.c
ck-telecom: This could be more flexible no matter how many bytes left aligned with 4 bytes.

JarmouniA (dismissed):

ck-telecom:
@gmarull @danieldegrasse please approve if you have no comment.

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
```diff
@@ -0,0 +1,37 @@
...
+		status = "okay";
+		#address-cells = <1>;
+		#size-cells = <0>;
```
gmarull: why is this needed?

gmarull (commented):
lgtm, one nitpick

tests/drivers/build_all/display/boards/sf32lb52_devkit_lcd.overlay
ck-telecom: parent node declare `address-cells` and `size-cells`, and seems like `partitions`, missing this will cause building error
```
&mpi2 {
	compatible = "sifli,sf32lb-mpi-qspi-nor";
	dmas = <&dmac 0 SF32LB52X_DMA_REQ_MPI2 SF32LB_DMA_PL_MEDIUM>;
	sifli,psclr = <0>;
	sifli,lines = <4>;
	status = "okay";

	py25q128ha: flash@0 {
		compatible = "puya,py25q128ha", "jedec,qspi-nor";
		reg = <0x0>;
		size = <DT_SIZE_M(128)>;
		quad-enable-requirements = "S2B1v1";

		partitions {
			compatible = "fixed-partitions";
			#address-cells = <1>;
			#size-cells = <1>;

			ptable: partition@0 {
				label = "ptable";
				reg = <0x0 DT_SIZE_K(64)>;
			};

			code: partition@10000 {
				label = "code";
				reg = <0x10000 DT_SIZE_K(16320)>;
			};
		};
	};
};
```

gmarull: this is only required if children nodes have address/size, looks like not the case here?

ck-telecom: If so, I will remove the parent node `address-cell` and `size-cell` like others nodes

ck-telecom: [https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/mipi-dbi/mipi-dbi-controller.yaml](url) says require ture, so will added again

gmarull: @danieldegrasse can you clarify why is this needed in base bindings?

gmarull (approved):

JarmouniA (approved):

fabiobaltieri:
reassigning, does not change subsystem files, only platform ones

PR #92318, opened 2025-06-27
ci: .github: workflows: add IAR workflow

kartben:
Similar to the clang workflow that runs against native_sim, but for IAR Arm on qemu_cortex_m3.

PR #99556, opened 2025-11-18
kernel: thread: mark return undefined in z_thread_entry using DWARF

PetervdPerk-NXP:
Add DWARF hint for Cortex-M targets to handle z_thread_entry correctly in debuggers. This function starts a new thread and never returns. Use `.cfi_undefined lr` so DWARF-based unwinding does not rely on LR. Without this, unwinding may follow a bogus return address, leading to invalid memory reads and potential bus faults during backtrace.

Situation before this patch
image

Situation after this patch
image

As you can the debugger doesn't try to read at address 0xAAAAAAAA, which in the case of vmu_rt1170 would lead to a fault.

### DIFF
--- a/lib/os/thread_entry.c
+++ b/lib/os/thread_entry.c
@@ -12,6 +12,7 @@
  */
 
 #include <zephyr/kernel.h>
+#include <zephyr/arch/cfi.h>
 #ifdef CONFIG_CURRENT_THREAD_USE_TLS
 #include <zephyr/random/random.h>
 
@@ -35,6 +36,17 @@ extern Z_THREAD_LOCAL volatile uintptr_t __stack_chk_guard;
 FUNC_NORETURN void z_thread_entry(k_thread_entry_t entry,
 				 void *p1, void *p2, void *p3)
 {
+/*
+ * Inform the unwinder that the current return address is undefined.
+ *
+ * This is typically used at points in the code where execution does not
+ * return like a normal function (for example, thread entry routines).
+ * Marking the return address as undefined prevents stack unwinding or
+ * backtrace tools from following a bogus return address, which could
+ * otherwise lead to reading invalid memory and cause faults during
+ * unwinding or debugging.
+ */
+	ARCH_CFI_UNDEFINED_RETURN_ADDRESS();
 #ifdef CONFIG_CURRENT_THREAD_USE_TLS
 	z_tls_current = k_sched_current_thread_query();
 #endif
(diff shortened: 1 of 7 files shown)


lib/os/thread_entry.c
```diff
@@ -35,6 +35,19 @@ extern Z_THREAD_LOCAL volatile uintptr_t __stack_chk_guard;
 FUNC_NORETURN void z_thread_entry(k_thread_entry_t entry,
 				 void *p1, void *p2, void *p3)
 {
+#ifdef CONFIG_CPU_CORTEX_M
```
peter-mitsis: Given your explanation, it would not surprise me if the underlying idea would apply to other architectures. As a result, it may be worthwhile to abstract this out to something like ...
   arch_cfi_undefined_return_address();
... and define that on a per-architecture basis.

PetervdPerk-NXP: Sure we could make prototype that each arch could implement.
A quick google suggest the following for RISC-V and x86-64 but I don't have the hardware to test it.
```
RISC-V .cfi_undefined ra
x86-64 .cfi_undefined rip
```
Should I implement those other architectures as well, or just stick to Cortex-M for now?

PetervdPerk-NXP: @peter-mitsis Implemented `ARCH_CFI_UNDEFINED_RETURN_ADDRESS()` with architecture specific registers.

peter-mitsis (approved):

ycsin:
Looks like a bug fix to me, can we backport this?

bperseghetti (approved):

nashif (approved):

PetervdPerk-NXP:
> This breaks toolchain not supporting `.cfi_undefined` (e.g. IAR). Please wait for fix before backporting this.

It looks like we still don’t have IAR CI coverage; otherwise, this issue would likely have been caught. In fact, #92318 probably would have detected it.

PetervdPerk-NXP:
> > > This breaks toolchain not supporting `.cfi_undefined` (e.g. IAR). Please wait for fix before backporting this.
> > 
> > 
> > It looks like we still don’t have IAR CI coverage; otherwise, this issue would likely have been caught. In fact, #92318 probably would have detected it.
> 
> I am aware :). The current way when we catch it in internal CI is not optimal, but it is how it is. If you want an IAR license token for yourself I can help you (though I think NXP already has this).

I could make a draft PR with

```
#if defined(__GNUC__) || defined(__clang__)
#define CFI_SUPPORTED 1
#else
#define CFI_SUPPORTED 0
#endif
```

If you could run it through a IAR teststuite to confirm the regression is gone then it's fine by me.

PR #99557, opened 2025-11-18
Bluetooth: AVRCP: Fix PTS test case failures in Bluetooth AVRCP

makeshi:
- Validate PDU ID in vendor dependent command handler; return INVALID_COMMAND for unsupported IDs to improve error handling
- Mask invalid bit in absolute volume values to avoid out-of-range errors and ensure compatibility with PTS.

### DIFF
--- a/subsys/bluetooth/host/classic/avrcp.c
+++ b/subsys/bluetooth/host/classic/avrcp.c
@@ -1355,10 +1355,8 @@ static int process_set_absolute_volume_rsp(struct bt_avrcp *avrcp, uint8_t tid,
 		return BT_AVRCP_STATUS_INVALID_PARAMETER;
 	}
 	absolute_volume = net_buf_pull_u8(buf);
-	if (absolute_volume > BT_AVRCP_MAX_ABSOLUTE_VOLUME) {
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/host/classic/avrcp.c
```diff
@@ -2331,6 +2331,16 @@ static inline uint8_t get_cmd_min_len_by_pdu(uint8_t pdu_id)
...
+static bool is_valid_cmd_pdu_id(uint8_t pdu_id)
+{
+	for (size_t i = 0; i < ARRAY_SIZE(cmd_vendor_handlers); i++) {
+		if (cmd_vendor_handlers[i].pdu_id == pdu_id) {
+			return true;
+		}
+	}
+	return false;
+}
+
```
MarkWangChinese: Is it OK to move both `is_valid_cmd_pdu_id` and `get_cmd_type_by_pdu` to `handle_vendor_pdu` because `handle_vendor_pdu` loops the hander array?

makeshi: yes, good suggestion, updated.

MarkWangChinese (approved):

lylezhu2012 (approved):

PR #99571, opened 2025-11-18
Bluetooth: Host: Legacy passkey entry 6.2 update

HaavardRei:
As of Core v6.2, the passkey entry pairing method for legacy pairing does no longer grant authenticated MITM protection. This commit updates `smp.c` accordingly to not grant the authenticated states when using legacy passkey entry pairing.

Adds a check to make sure that keys that have been stored persistently adheres to these changes. Keys that have been generated using the legacy passkey entry pairing method will thus be cleared when attempting to restore keys from storage.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -51,6 +51,10 @@ Bluetooth Host
 * :c:member:`bt_conn_le_info.interval` has been deprecated. Use
   :c:member:`bt_conn_le_info.interval_us` instead. Note that the units have changed: ``interval``
   was in units of 1.25 milliseconds, while ``interval_us`` is in microseconds.
+* Legacy Bluetooth LE pairing using the passkey entry method no longer grants authenticated (MITM)
+  protection as of the Bluetooth Core Specification v6.2. Stored bonds that were generated using
+  this method will be downgraded to unauthenticated when loaded from persistent storage, resulting
+  in a lower security level.
 
 Networking
 **********
--- a/subsys/bluetooth/host/keys.c
+++ b/subsys/bluetooth/host/keys.c
@@ -427,6 +427,18 @@ static int keys_set(const char *name, size_t len_rd, settings_read_cb read_cb,
 		memcpy(keys->storage_start, val, len);
 	}
 
+	/* As of Core v6.2, authenticated keys are only valid for OOB or LE SC pairing
+	 * methods. This check ensures that keys are valid if a device is updated from a
+	 * previous version that did not enforce this requirement.
+	 */
+	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
+	    !((keys->flags & BT_KEYS_OOB) || (keys->flags & BT_KEYS_SC))) {
+		LOG_WRN("The keys for %s are downgraded to unauthenticated as they no longer meet "
+			"authentication requirements",
+			bt_addr_le_str(&addr));
+		keys->flags &= ~BT_KEYS_AUTHENTICATED;
+	}
+
 	LOG_DBG("Successfully restored keys for %s", bt_addr_le_str(&addr));
 #if defined(CONFIG_BT_KEYS_OVERWRITE_OLDEST)
 	if (aging_counter_val < keys->aging_counter) {
--- a/subsys/bluetooth/host/smp.c
+++ b/subsys/bluetooth/host/smp.c
@@ -667,7 +667,9 @@ static bool update_keys_check(struct bt_smp *smp, struct bt_keys *keys)
 	}
 
 	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
-	     smp->method == JUST_WORKS) {
+	    ((smp->method == JUST_WORKS) ||
+	     (!atomic_test_bit(smp->flags, SMP_FLAG_SC) &&
+	      (smp->method == PASSKEY_DISPLAY || smp->method == PASSKEY_INPUT)))) {
 		return false;
 	}
 
@@ -2492,11 +2494,12 @@ static uint8_t legacy_request_tk(struct bt_smp *smp)
 	 * Fail if we have keys that are stronger than keys that will be
 	 * distributed in new pairing. This is to avoid replacing authenticated
 	 * keys with unauthenticated ones.
-	  */
+	 */
 	keys = bt_keys_find_addr(conn->id, &conn->le.dst);
 	if (keys && (keys->flags & BT_KEYS_AUTHENTICATED) &&
-	    smp->method == JUST_WORKS) {
-		LOG_ERR("JustWorks failed, authenticated keys present");
+	    (smp->method == JUST_WORKS || smp->method == PASSKEY_DISPLAY ||
+	     smp->method == PASSKEY_INPUT)) {
+		LOG_ERR("Pairing failed, authenticated keys present");
 		return BT_SMP_ERR_UNSPECIFIED;
 	}
 
@@ -2984,7 +2987,9 @@ static uint8_t remote_sec_level_reachable(struct bt_smp *smp)
 		}
 		__fallthrough;
 	case BT_SECURITY_L3:
-		if (smp->method == JUST_WORKS) {
+		if (smp->method == JUST_WORKS ||
+		    (!atomic_test_bit(smp->flags, SMP_FLAG_SC) &&
+		     (smp->method == PASSKEY_DISPLAY || smp->method == PASSKEY_INPUT))) {
 			return BT_SMP_ERR_AUTH_REQUIREMENTS;
 		}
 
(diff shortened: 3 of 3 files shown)


PavelVPV (dismissed):

jhedberg (dismissed):

alwa-nordic (dismissed):

jhedberg:
@HaavardRei there's a merge conflict, i.e. you need to rebase

HaavardRei:
Added a DNM because after some local testing of the clearing functionality in keys.c, I realized the check for authenticated keys needed to be moved after the `memcpy(keys->storage_start, val, len);` in order for the `keys->flags` field to actually be loaded. Fixed this now and checked that a stored bond using legacy passkey entry is cleared upon update to this commit.

PavelVPV (dismissed):

subsys/bluetooth/host/smp.c
```diff
@@ -2492,11 +2494,12 @@ static uint8_t legacy_request_tk(struct bt_smp *smp)
...
 	 * distributed in new pairing. This is to avoid replacing authenticated
 	 * keys with unauthenticated ones.
-	  */
+	 */
 	keys = bt_keys_find_addr(conn->id, &conn->le.dst);
 	if (keys && (keys->flags & BT_KEYS_AUTHENTICATED) &&
-	    smp->method == JUST_WORKS) {
-		LOG_ERR("JustWorks failed, authenticated keys present");
+	    (smp->method == JUST_WORKS || smp->method == PASSKEY_DISPLAY ||
+	     smp->method == PASSKEY_INPUT)) {
```
alwa-nordic, line 2501: See https://github.com/zephyrproject-rtos/zephyr/pull/99571/files#r2544855587

subsys/bluetooth/host/smp.c
```diff
@@ -667,7 +667,9 @@ static bool update_keys_check(struct bt_smp *smp, struct bt_keys *keys)
 	}
 
 	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
-	     smp->method == JUST_WORKS) {
+	    ((smp->method == JUST_WORKS) ||
+	     (!atomic_test_bit(smp->flags, SMP_FLAG_SC) &&
+	      (smp->method == PASSKEY_DISPLAY || smp->method == PASSKEY_INPUT)))) {
```
alwa-nordic, line 672: This is something to discuss quickly before we proceed. This will allow overwriting a legacy passkey pairing with a legacy just-works pairing. It may be the right thing to do, but it is a weakening of security.

@seko-nordic Can you comment on whether this is appropriate?

The legacy passkey method is no longer considered as sufficiently authenticated for MITM defense, but it could still be considered stronger than passkey.

We also know that the peer device previously had the IO capabilities to do a passkey method, so it's suspicious that it would no longer have it.

If we consider firmware downgrades, the fact that bond overwrites don't generate a 'unpaired' event to the application could be a security risk. @seko-nordic Are frimware downgrades admissible in our thread model?

I'll NAK this PR so this is not merged before we discuss it. But it's not necessarily wrong.

jhedberg (dismissed):

doc/releases/migration-guide-4.4.rst
```diff
@@ -51,6 +51,10 @@ Bluetooth Host
 * :c:member:`bt_conn_le_info.interval` has been deprecated. Use
   :c:member:`bt_conn_le_info.interval_us` instead. Note that the units have changed: ``interval``
   was in units of 1.25 milliseconds, while ``interval_us`` is in microseconds.
+* Legacy Bluetooth LE pairing using the passkey entry method no longer grants authenticated (MITM)
+  protection as of the Bluetooth Core Specification v6.2. Keys stored persistently that were
+  generated using this method will no longer be valid. A check is added for this when loading keys
+  from persistent storage, in which case the keys will be cleared.
```
alwa-nordic: I think "bond" is the a more common term than "key" for this. It's also worth warning that bonds will be deleted silently.

```suggestion
* Legacy Bluetooth LE pairing using the passkey entry method no longer grants authenticated (MITM)
  protection as of the Bluetooth Core Specification v6.2. Bonds stored persistently that were
  generated using this method will no longer be valid. A check is added for this when loading bonds
  from persistent storage, in which case the bonds will be silently deleted without notifying the
  application through `bt_conn_auth_info_cb.bond_deleted`.
```

subsys/bluetooth/host/keys.c
```diff
@@ -427,6 +427,19 @@ static int keys_set(const char *name, size_t len_rd, settings_read_cb read_cb,
...
+	/* As of Core v6.2, authenticated keys are only valid for OOB or LE SC pairing
+	 * methods. This check ensures that keys are valid if a device is updated from a
+	 * previous version that did not enforce this requirement.
+	 */
+	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
+	    !((keys->flags & BT_KEYS_OOB) || (keys->flags & BT_KEYS_SC))) {
+		LOG_ERR("Authenticated keys for %s do not meet requirements",
+			bt_addr_le_str(&addr));
+		bt_keys_clear(keys);
+
```
alwa-nordic: I think we should invoke `bt_conn_auth_info_cb.bond_deleted` to notify the application, in case the application needs to clean up something that's tied to the existence of the bond.

subsys/bluetooth/host/keys.c
```diff
@@ -427,6 +427,19 @@ static int keys_set(const char *name, size_t len_rd, settings_read_cb read_cb,
...
 
+	/* As of Core v6.2, authenticated keys are only valid for OOB or LE SC pairing
+	 * methods. This check ensures that keys are valid if a device is updated from a
+	 * previous version that did not enforce this requirement.
+	 */
+	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
+	    !((keys->flags & BT_KEYS_OOB) || (keys->flags & BT_KEYS_SC))) {
+		LOG_ERR("Authenticated keys for %s do not meet requirements",
+			bt_addr_le_str(&addr));
+		bt_keys_clear(keys);
```
alwa-nordic: Is it safe to call `settings_delete` in a settings set handler? Nothing to the contrary is documented, so it should be?

(Not a NAK comment)

alwa-nordic (changes requested):

subsys/bluetooth/host/smp.c
seko-nordic: Intend here is to not allow overwriting of auth keys with passkey, do I read this correctly?
You saying that as a side effect it will allow overwriting passkey with JW, is it correct? May be we should remove this side effect.

subsys/bluetooth/host/keys.c
HaavardRei: Think so, anyway there is a call to bt_keys_clear a few lines above that seemingly hasn't given any problems.

subsys/bluetooth/host/keys.c
HaavardRei: Like this? https://github.com/zephyrproject-rtos/zephyr/blob/4a0e6d48c9726723a08ee2f6f497a7c53965da40/subsys/bluetooth/host/hci_core.c#L2179-L2186

subsys/bluetooth/host/smp.c
HaavardRei: Isn't the purpose of this check to - if we already have authenticated keys - disallow JW pairing and now legacy passkey entry pairing as this results in lower security than the existing bond?
I might be interpreting it wrong, but I don't see how this allows for a legacy JW pairing to overwrite a legacy passkey entry pairing.

doc/releases/migration-guide-4.4.rst
HaavardRei: I don't think it's good documentation to explicitly state that "we don't call this callback here". Can rather have the callback :) 
(and you do get log_err-s during setting loading)

subsys/bluetooth/host/smp.c
alwa-nordic: You are right. I misread the code. The check is stricter now. Sorry about that.

doc/releases/migration-guide-4.4.rst
alwa-nordic: I agree, it's better to have the callback.

subsys/bluetooth/host/keys.c
alwa-nordic: I think it's better to call `bt_unpair()` here. That's public API.

HaavardRei:
> why not just treat those keys as unauthenticated (ie clear BT_KEYS_AUTHENTICATED flag) instead of deleting bond?

@sjanc This is a fair point. Previously I thought it'd be "more secure" to outright remove the bond, but the way the specification is formed it is possible to have a bond between a Core 6.1 device and a 6.2 device where (using passkey legacy entry pairing) the devices get respectively security levels 3 and 2.

subsys/bluetooth/host/keys.c
```diff
@@ -427,6 +427,18 @@ static int keys_set(const char *name, size_t len_rd, settings_read_cb read_cb,
...
 	}
 
+	/* As of Core v6.2, authenticated keys are only valid for OOB or LE SC pairing
+	 * methods. This check ensures that keys are valid if a device is updated from a
+	 * previous version that did not enforce this requirement.
+	 */
+	if ((keys->flags & BT_KEYS_AUTHENTICATED) &&
+	    !((keys->flags & BT_KEYS_OOB) || (keys->flags & BT_KEYS_SC))) {
+		LOG_WRN("The keys for %s are downgraded to unauthenticated as they no longer meet "
+			"authentication requirements",
```
HaavardRei, line 437: This became a long warning (not sure if there is a "limit" to log message length), proposals to shorten it is welcome :)

doc/releases/migration-guide-4.4.rst
HaavardRei: Changed to bonds, for the later part after discussion the key authentication is downgraded to non-auth with a warning.

subsys/bluetooth/host/keys.c
HaavardRei: No longer relevant after downgrading change.

HaavardRei:
Behavior changed to no longer clearing the bond, but removing the authentication flag on settings restore. (This is not stored persistently, so future reboots will print the same warning regarding downgrading)

alwa-nordic (approved):

jhedberg (approved):

PR #99574, opened 2025-11-18
drivers: dai: add size parameter to dai_config_set API

abonislawski:
- Add size parameter to dai_config_set API
    Add explicit bespoke_cfg_size parameter to the dai_config_set()
    function and its underlying driver API to improve configuration
    validation and security.
- Add support for SSP_GTW_DMA_CONFIG_ID TLV type
    Add handling for SSP_GTW_DMA_CONFIG_ID (0x1000) TLV type in SSP driver's
    auxiliary data parsing functions. This TLV type is explicitly ignored as
    it does not require any processing by the driver.
- Fix SSP blob v3.0 TLV parsing
    Fix incorrect TLV (Type-Length-Value) data length calculation for
    SSP configuration blob version 3.0. The blob30->size field does not
    include auxiliary TLV data appended after the main structure, leading
    to incorrect parsing boundaries.

### DIFF
--- a/drivers/dai/intel/alh/alh.c
+++ b/drivers/dai/intel/alh/alh.c
@@ -124,7 +124,7 @@ static int dai_alh_config_get(const struct device *dev, struct dai_config *cfg,
 }
 
 static int dai_alh_config_set(const struct device *dev, const struct dai_config *cfg,
-				  const void *bespoke_cfg)
+				  const void *bespoke_cfg, size_t size)
 {
 	struct dai_intel_alh *dp = (struct dai_intel_alh *)dev->data;
 
--- a/drivers/dai/intel/ssp/ssp.c
+++ b/drivers/dai/intel/ssp/ssp.c
@@ -1875,6 +1875,9 @@ static int dai_ssp_check_aux_data(struct ssp_intel_aux_tlv *aux_tlv, int aux_len
 #else
 		return 0;
 #endif
+	case SSP_GTW_DMA_CONFIG_ID:
+		/* ignored */
+		return 0;
 	default:
 		LOG_ERR("undefined aux data type %u", aux_tlv->type);
 		return -EINVAL;
@@ -1999,6 +2002,9 @@ static int dai_ssp_parse_tlv(struct dai_intel_ssp *dp, const uint8_t *aux_ptr, s
 #endif
 			LOG_INF("link clock_source %u", link->clock_source);
 #endif
+			break;
+		case SSP_GTW_DMA_CONFIG_ID:
+			/* ignored */
 			break;
 		default:
 			LOG_ERR("undefined aux data type %u", aux_tlv->type);
--- a/drivers/dai/intel/ssp/ssp.c
+++ b/drivers/dai/intel/ssp/ssp.c
@@ -2018,7 +2018,7 @@ static int dai_ssp_parse_tlv(struct dai_intel_ssp *dp, const uint8_t *aux_ptr, s
 	return 0;
 }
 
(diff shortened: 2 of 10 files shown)


drivers/dai/intel/alh/alh.c
```diff
@@ -124,7 +124,7 @@ static int dai_alh_config_get(const struct device *dev, struct dai_config *cfg,
 }
 
 static int dai_alh_config_set(const struct device *dev, const struct dai_config *cfg,
-				  const void *bespoke_cfg)
+				  const void *bespoke_cfg, size_t size)
```
kv2019i, line 127: Btw, @nashif now the dai.h reached unstable (from previous experimental status), can you confirm we can still make such changes with backward compatibility layers? 

I'm working on another similar change (to get_properties) and was wondering whether I need to provide compatibility to old, and provide the new interface as an extension. But eading the process docs, this would seem not needed for "unstable" interfaces. Of course, we should never change unless we really have to, but I think both this PR (and my upcoming) have solid reasons for the change.

drivers/dai/intel/ssp/ssp.c
```diff
@@ -2032,7 +2032,7 @@ static int dai_ssp_parse_aux_data(struct dai_intel_ssp *dp, const void *spec_con
 		aux_len = cfg_len - pre_aux_len;
 		aux_ptr = (uint8_t *)blob15 + pre_aux_len;
 	} else if (blob30->version == SSP_BLOB_VER_3_0) {
-		cfg_len = blob30->size;
+		cfg_len = size;
```
kv2019i, line 2035: Oh boy, it is unfortunate proper TLV encoding is not used here (especially as there's a size field already), but it is what it is. 

I do like the "size" addition nevertheless as it helps with using dai.h from user-space ('spec_config' can be safely copied without having to parse its details).

The addition does create new type of error conditions. E.g. what if "size < blob15->size" or what if "size < blob30_size"? It seems these are ok (e.g. aux_len check on L2045), but please confirm you checked these.

kv2019i (dismissed):
The flag-day need for SOF is not ideal, but change itself looks good.

One alternative is to add dai_config_set_v2 API that takes size and move users incrementally to the new interface.

drivers/dai/intel/ssp/ssp.c
abonislawski: Unfortunately TLVs are not part of blob30 so blob30->size wont cover them.
aux_len check on L2045 validates such case with incorrect size configuration.

drivers/dai/intel/alh/alh.c
abonislawski: I checked API usage and looks like there is only one place in SOF so the transition should be really smooth.

On the other hand I have also considered new bespoke_cfg_size field inside dai_config struct and in that case we could maintain compatibility.

kv2019i: @abonislawski this also aligns better with the "dai_config_update()" op that already has the size field.

kv2019i (approved):

Holt-Sun (approved):

tmleman (approved):

include/zephyr/drivers/dai.h
```diff
@@ -315,7 +315,7 @@ __subsystem struct dai_driver_api {
 	int (*probe)(const struct device *dev);
 	int (*remove)(const struct device *dev);
 	int (*config_set)(const struct device *dev, const struct dai_config *cfg,
-			  const void *bespoke_cfg);
+			  const void *bespoke_cfg, size_t size);
```
kartben, line 318: i know the API is techinically "unstable" but I would recommend mentioning the API change and required course of action to update in the migration guide

PR #99576, opened 2025-11-18
tests: bluetooth: tester: fix model rx handler

alxelax:
Commit fixes the model rx handler.
Model can receive maximum access payload.
Test should be able fit maximum access payload and its own header for serial communication.


PavelVPV (commented):
I guess the check on line 5340 is still not valid and should include `sizeof(*ev)`: `payload_len + sizeof(*ev)`.

PavelVPV (dismissed):

PavelVPV (approved):

sjanc (approved):

vlad-kulikov (approved):

PR #99591, opened 2025-11-18
Support multiple devices to connect with usb host if usb hub is used

AidenHu:
When USB hub is used, maybe multiple devices are attached, so need to consider about this.

### DIFF
--- a/subsys/usb/host/usbh_core.c
+++ b/subsys/usb/host/usbh_core.c
@@ -46,6 +46,7 @@ static int usbh_event_carrier(const struct device *dev,
 static void dev_connected_handler(struct usbh_context *const ctx,
 				  const struct uhc_event *const event)
 {
+	int err;
 	struct usb_device *udev;
 
 	LOG_DBG("Device connected event");
@@ -66,13 +67,10 @@ static void dev_connected_handler(struct usbh_context *const ctx,
 
 	k_mutex_lock(&ctx->mutex, K_FOREVER);
 	sys_dlist_append(&ctx->udevs, &udev->node);
-
-	if (ctx->root == NULL) {
-		ctx->root = udev;
-	}
 	k_mutex_unlock(&ctx->mutex);
 
-	if (usbh_device_init(udev)) {
+	err = usbh_device_init(udev);
+	if (ret != 0) {
 		LOG_ERR("Failed to reset new USB device");
 		sys_dlist_remove(&udev->node);
 	}
@@ -80,13 +78,17 @@ static void dev_connected_handler(struct usbh_context *const ctx,
 
 static void dev_removed_handler(struct usbh_context *const ctx)
 {
-	if (ctx->root != NULL) {
-		usbh_device_free(ctx->root);
-		ctx->root = NULL;
+	struct usb_device *udev = NULL;
+
+	udev = usbh_device_get_root(ctx);
+
+	if (NULL != udev) {
+		usbh_device_free(udev);
 		LOG_DBG("Device removed");
 	} else {
 		LOG_DBG("Spurious device removed event");
 	}
+	/* TODO: handle remove for all of classes for the unattached device */
 }
 
 static int discard_ep_request(struct usbh_context *const ctx,
--- a/subsys/usb/host/usbh_core.c
+++ b/subsys/usb/host/usbh_core.c
@@ -46,30 +46,35 @@ static int usbh_event_carrier(const struct device *dev,
 static void dev_connected_handler(struct usbh_context *const ctx,
 				  const struct uhc_event *const event)
 {
+	struct usb_device *udev;
 
 	LOG_DBG("Device connected event");
-	if (ctx->root != NULL) {
-		LOG_ERR("Device already connected");
-		usbh_device_free(ctx->root);
-		ctx->root = NULL;
-	}
 
-	ctx->root = usbh_device_alloc(ctx);
-	if (ctx->root == NULL) {
+	udev = usbh_device_alloc(ctx);
+	if (udev == NULL) {
 		LOG_ERR("Failed allocate new device");
 		return;
 	}
 
-	ctx->root->state = USB_STATE_DEFAULT;
+	udev->state = USB_STATE_DEFAULT;
 
 	if (event->type == UHC_EVT_DEV_CONNECTED_HS) {
-		ctx->root->speed = USB_SPEED_SPEED_HS;
+		udev->speed = USB_SPEED_SPEED_HS;
 	} else {
-		ctx->root->speed = USB_SPEED_SPEED_FS;
+		udev->speed = USB_SPEED_SPEED_FS;
 	}
 
-	if (usbh_device_init(ctx->root)) {
+	k_mutex_lock(&ctx->mutex, K_FOREVER);
+	sys_dlist_append(&ctx->udevs, &udev->node);
+
+	if (ctx->root == NULL) {
+		ctx->root = udev;
+	}
+	k_mutex_unlock(&ctx->mutex);
+
+	if (usbh_device_init(udev)) {
 		LOG_ERR("Failed to reset new USB device");
+		sys_dlist_remove(&udev->node);
 	}
 }
 
--- a/subsys/usb/host/usbh_device.c
+++ b/subsys/usb/host/usbh_device.c
@@ -447,6 +447,26 @@ error:
 	return err;
 }
 
+struct usb_device *usbh_device_get_root(struct usbh_context *const ctx)
+{
+	sys_dnode_t *node;
+
+	node = sys_dlist_peek_head(&ctx->udevs);
+	if (node == NULL) {
+		/* No devices in the list */
+		return NULL;
+	}
+
+	/* Get the usb_device structure from the node */
+	return CONTAINER_OF(node, struct usb_device, node);
+}
+
+bool usbh_device_is_root(struct usbh_context *const ctx,
+			struct usb_device *const udev)
+{
+	return sys_dlist_peek_head(&ctx->udevs) == &udev->node;
+}
+
 int usbh_device_init(struct usb_device *const udev)
 {
 	struct usbh_context *const uhs_ctx = udev->ctx;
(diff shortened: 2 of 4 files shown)


subsys/usb/host/usbh_device.c
```diff
@@ -464,11 +466,19 @@ int usbh_device_init(struct usb_device *const udev)
...
 
-	/* FIXME: The port to which the device is connected should be reset. */
-	err = uhc_bus_reset(uhs_ctx->dev);
-	if (err) {
-		LOG_ERR("Failed to signal bus reset");
-		return err;
+	k_mutex_lock(&uhs_ctx->mutex, K_FOREVER);
+	SYS_DLIST_FOR_EACH_NODE(&uhs_ctx->udevs, node) {
+		device_count++;
+	}
```
josuah: I think you can use `device_count = sys_dlist_len(&uhs_ctx->udevs);` in this case.

[doc](https://docs.zephyrproject.org/latest/doxygen/html/group__doubly-linked-list__apis.html#ga6797f138ad66cfb5ee7918bb12869eac)

But let's discuss on https://github.com/zephyrproject-rtos/zephyr/pull/99591/files#r2538708506 first?

subsys/usb/host/usbh_core.c
```diff
@@ -46,30 +46,35 @@ static int usbh_event_carrier(const struct device *dev,
...
 	} else {
-		ctx->root->speed = USB_SPEED_SPEED_FS;
+		udev->speed = USB_SPEED_SPEED_FS;
+	}
+
+	k_mutex_lock(&ctx->mutex, K_FOREVER);
+	sys_dlist_append(&ctx->udevs, &udev->node);
+	
+	if (ctx->root == NULL) {
+		ctx->root = udev;
```
josuah: It is expected that `ctx->root == sys_dlist_peek_head(&ctx->udevs)` always.
If not, there is some bug in the host class code.

@AidenHu @jfischer-no
What do you think about replacing `ctx->root` by `usbh_device_get_root(ctx)` and add a `usbh_device_is_root(ctx, udev)` to store the root device only once (as the first element of `ctx->udevs`)?

This way there is no need to make sure that `ctx->root` and first of `ctx->udevs` are in sync.

josuah (commented):
Thanks. Some discussion about it below.

subsys/usb/host/usbh_device.c
```diff
@@ -464,11 +466,19 @@ int usbh_device_init(struct usb_device *const udev)
...
+	k_mutex_unlock(&uhs_ctx->mutex);
+
+	/* Only reset bus if this is the root device. */
+	if (device_count == 1U) {
+		err = uhc_bus_reset(uhs_ctx->dev);
+		if (err) {
+			LOG_ERR("Failed to signal bus reset");
+			return err;
+		}
```
tmon-nordic, line 499: What about PORT_RESET hub class feature? Where this should be handled? At high-level Linux solves the issue by having a software-only hub at the top level.

subsys/usb/host/usbh_device.c
AidenHu: @

> I think you can use `device_count = sys_dlist_len(&uhs_ctx->udevs);` in this case.
> 
> [doc](https://docs.zephyrproject.org/latest/doxygen/html/group__doubly-linked-list__apis.html#ga6797f138ad66cfb5ee7918bb12869eac)
> 
> But let's discuss on https://github.com/zephyrproject-rtos/zephyr/pull/99591/files#r2538708506 first?

@josuah  agree, already done for sys_dlist_len

subsys/usb/host/usbh_core.c
AidenHu: @josuah 
I have checked (run the code in my project)` ctx->root == sys_dlist_peek_head(&ctx->udevs) `is yes.
Yes, agree to provide these two APIs. This will avoid the potential risk. I have updated and provided the function implementation.

subsys/usb/host/usbh_device.c
AidenHu: > What about PORT_RESET hub class feature? Where this should be handled? At high-level Linux solves the issue by having a software-only hub at the top level.

Hi @tmon-nordic
The PORT_RESET of HUB is handled by a set feature request. This is handled in hub class layer. Actually I have finished the first version USB hub class and tested in my local project. Recently I will created one PR about it.

subsys/usb/host/usbh_device.c
```diff
@@ -447,11 +447,42 @@ int usbh_device_set_configuration(struct usb_device *const udev, const uint8_t n
...
+int usbh_device_is_root(struct usbh_context *const ctx,
+			struct usb_device *const udev)
+{
+	if (sys_dlist_peek_head(&ctx->udevs) == &udev->node) {
+		return 1;
+	}
+	else {
+		return 0;
+	}
+}
```
josuah: For functions with `_is_` in the name, you could be interested in returning `bool` and then you can return the expression directly if you wish:

```c
bool usbh_device_is_root(struct usbh_context *const ctx,
			struct usb_device *const udev)
{
	return sys_dlist_peek_head(&ctx->udevs) == &udev->node;
}
```

AidenHu: Yes, thanks @josuah. Have updated it.

josuah:
Thank you @AidenHu for the modifications. As soon as the CI tests above all pass, and review is complete, this PR can be merged.

screenshot_2025_1120_0245_57

It is possible to read more about it [here](https://docs.zephyrproject.org/latest/contribute/guidelines.html#continuous-integration) ("Continuous Integration") if curious.

subsys/usb/host/usbh_core.c
```diff
@@ -46,42 +46,47 @@ static int usbh_event_carrier(const struct device *dev,
...
-		ctx->root->speed = USB_SPEED_SPEED_FS;
+		udev->speed = USB_SPEED_SPEED_FS;
 	}
 
-	if (usbh_device_init(ctx->root)) {
+	k_mutex_lock(&ctx->mutex, K_FOREVER);
+	sys_dlist_append(&ctx->udevs, &udev->node);
+	k_mutex_unlock(&ctx->mutex);
+
+	if (usbh_device_init(udev)) {
```
josuah: For functions that return errors, the most frequent style is this way:
```suggestion
	ret = usbh_device_init(udev);
	if (ret != 0) {
```
It is not really listed anywhere in the coding style, but it is how the code looks like in USB and most places.

subsys/usb/host/usbh_core.c
```diff
@@ -46,42 +46,47 @@ static int usbh_event_carrier(const struct device *dev,
...
 static void dev_removed_handler(struct usbh_context *const ctx)
 {
-	if (ctx->root != NULL) {
-		usbh_device_free(ctx->root);
-		ctx->root = NULL;
+	struct usb_device *udev = NULL;
+
+	udev = usbh_device_get_root(ctx);
+
+	if (udev) {
```
josuah: Comparing explicitly against `NULL` helps with MISRA compliance.

See rule 85 here:
https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#coding-guidelines

This is not strictly applied, as I remember, but it can also help keeping the code a bit more homogene.

subsys/usb/host/usbh_device.c
```diff
@@ -447,11 +447,41 @@ int usbh_device_set_configuration(struct usb_device *const udev, const uint8_t n
...
+	}
+
+	/* Get the usb_device structure from the node */
+	return CONTAINER_OF(node, struct usb_device, node);
+}
+
+bool usbh_device_is_root(struct usbh_context *const ctx,
+			struct usb_device *const udev)
+{
+	if (ctx == NULL || udev == NULL) {
```
josuah: I think `ctx` or `udev` is always guaranteed to be `!= NULL` and might not need verification.

subsys/usb/host/usbh_device.c
```diff
@@ -447,11 +447,41 @@ int usbh_device_set_configuration(struct usb_device *const udev, const uint8_t n
 	return err;
 }
 
+struct usb_device *usbh_device_get_root(struct usbh_context *const ctx)
+{
+	sys_dnode_t *node;
+
+	if (ctx == NULL) {
+		return NULL;
+	}
```
josuah: I think `ctx` is always guaranteed to be `!= NULL`, and might not need verification.

subsys/usb/host/usbh_core.c
AidenHu: Thanks, fixed it.

subsys/usb/host/usbh_core.c
AidenHu: > Comparing explicitly against `NULL` helps with MISRA compliance.
> 
> See rule 85 here: https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#coding-guidelines
> 
> This is not strictly applied, as I remember, but it can also help keeping the code a bit more homogene.

Thanks, fixed it as the rule.

subsys/usb/host/usbh_device.c
AidenHu: > I think `ctx` or `udev` is always guaranteed to be `!= NULL` and might not need verification.

Thanks, fixed.

subsys/usb/host/usbh_device.c
AidenHu: > I think `ctx` is always guaranteed to be `!= NULL`, and might not need verification.

Thanks, fixed.

subsys/usb/host/usbh_core.c
```diff
@@ -46,42 +46,49 @@ static int usbh_event_carrier(const struct device *dev,
...
+		udev->speed = USB_SPEED_SPEED_FS;
 	}
 
-	if (usbh_device_init(ctx->root)) {
+	k_mutex_lock(&ctx->mutex, K_FOREVER);
+	sys_dlist_append(&ctx->udevs, &udev->node);
+	k_mutex_unlock(&ctx->mutex);
+
+	err = usbh_device_init(udev);
+	if (ret != 0) {
```
GeorgeCGV, line 73:
```suggestion
	if (err != 0) {
```

AidenHu: > 

Hi @GeorgeCGV 

All of fix have been moved to [usb host connect disconnect#100072](https://github.com/zephyrproject-rtos/zephyr/pull/100072). This point you mentioned is also fixed. 
Please ignore this PR.  [usb host connect disconnect#100072](https://github.com/zephyrproject-rtos/zephyr/pull/100072) will be the last PR we need to view and merged to main.

PR #99592, opened 2025-11-18
susbsys/mgmt/mcumgr/grp: add role field to parameters response

nvlsianpu:
Added optional `inst_role` field to MCUmgr Parameters response. Field is intended to express MCUmgr instance role on the device. Interpretation of the value is Application specific.

Purpose this additional field might be to express application specific: capabilities, requirements and cooperation schemes.
For instance I'm going to use this for distinguish MCUmgr instance in the application `A` from another MCumgr instance in the application `B` and also from MCumgr instance in the loader (ref.: SB_CONFIG_MCUBOOT_MODE_FIRMWARE_UPDATER=y).


de-nordic (approved):

nordicjm:
Generally an application specific value would go to an application specific (custom) MCUmgr group, the group IDs under 64 are for global standardised groups and commands, why should this be added to the global OS group there when it's completely application specific - and likely to have conflicts out in the field with other devices? There's also no documentation on this and no test for this.

PR #99599, opened 2025-11-18
Clean a few sys_getopt global status variable uses after #98845

aescolar:
sys_getopt has now its global status variables prefixed with sys_getopt_
Let's fix the few uses in tree that were forgotten in #98845. 
Note using the old names worked as before as Zephyr's POSIX_C_LIB_EXT keeps a replica of these variables without the prefix, but this is not something one can count on, as in general the C library provided getopt is not going to have any relationship to sys_getopt()

And since sys_getopt was split from POSIX_C_LIB_EXT the CRC_SHELL component only needs to enable GETOPT and not the whole POSIX_C_LIB_EXT.

Fixes #95477

And an extra commit removing a now unnecessary include of unistd.h from subsys/net/l2/wifi/wifi_shell.c (since we use sys_getopt and not the posix_api getopt shim)

### DIFF
--- a/doc/services/shell/index.rst
+++ b/doc/services/shell/index.rst
@@ -702,7 +702,7 @@ An example non-thread safe usage:
   while ((char c = sys_getopt(argc, argv, "abhc:")) != -1) {
         switch (c) {
         case 'c':
-                cvalue = optarg;
+                cvalue = sys_getopt_optarg;
                 break;
         default:
                 break;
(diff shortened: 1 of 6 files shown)


jukkar (dismissed):

jukkar (approved):

jakub-uC (approved):

cfriedt (changes requested):
Just a temporary nak so that the maintainer has time to review.

cfriedt (approved):

PR #99600, opened 2025-11-18
net: http_server: Support reason phrase for chunked encoded response

### LINKED ISSUE #99722: net: http_server: No suuport for reason-phrase for chunked encoded response
### Describe the bug

When using the Zephyr HTTP server, I’ve observed that for dynamic resources, the response status line is sent as HTTP/1.1 200 without a reason-phrase (e.g., "OK"). For static resources, the correct format HTTP/1.1 200 OK is used.

Expected behavior:
According to RFC 9110, the reason-phrase in the status line is optional. However, many HTTP clients, proxies, and test suites expect the traditional, human-readable reason-phrase to be present for compatibility. For best interoperability and consistency, it would be preferable if all Zephyr HTTP responses included the reason-phrase.

### Regression

- [ ] This is a regression.

### Steps to reproduce

This can be easily seen on Wireshark capture of HTTP transaction.

### Relevant log output

```shell

```

### Impact

### Environment

-OS: Linux
Version: 3.7

### Additional Context

_No response_

kfirous:
Add support for including HTTP reason phrases in chunked transfer-encoded responses

Fixes #99722

### DIFF
--- a/subsys/net/lib/http/Kconfig
+++ b/subsys/net/lib/http/Kconfig
@@ -231,6 +231,15 @@ config HTTP_SERVER_STATIC_FS_RESPONSE_SIZE
 	  Please note that it is allocated on the stack of the HTTP server thread,
 	  so CONFIG_HTTP_SERVER_STACK_SIZE has to be sufficiently large.
 
+config HTTP_SERVER_COMPLETE_STATUS_PHRASES
+	bool "Complete HTTP status reason phrases"
+	help
+	  Include reason phrases for HTTP status codes in server responses.
+	  When disabled, HTTP responses will not include reason phrases,
+	  reducing code size. Note that reason phrases are optional per
+	  HTTP specification and their absence does not affect protocol
+	  compliance or client behavior.
+
 endif
 
 # Hidden option to avoid having multiple individual options that are ORed together
--- a/subsys/net/lib/http/http_server_http1.c
+++ b/subsys/net/lib/http/http_server_http1.c
@@ -196,8 +196,43 @@ static int handle_http1_static_resource(
 	ret; })
 
 #define RESPONSE_TEMPLATE_DYNAMIC_PART1                                                            \
-	"HTTP/1.1 %d\r\n"                                                                          \
+	"HTTP/1.1 %d%s%s\r\n"                                                                      \
 	"Transfer-Encoding: chunked\r\n"
+
+static const char *http_status_str(enum http_status status)
+{
+#if defined(CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES)
+	switch (status) {
+	case HTTP_200_OK:
+		return "OK";
+	case HTTP_201_CREATED:
+		return "Created";
+	case HTTP_204_NO_CONTENT:
+		return "No Content";
+	case HTTP_400_BAD_REQUEST:
+		return "Bad Request";
+	case HTTP_401_UNAUTHORIZED:
+		return "Unauthorized";
+	case HTTP_403_FORBIDDEN:
+		return "Forbidden";
+	case HTTP_404_NOT_FOUND:
+		return "Not Found";
+	case HTTP_405_METHOD_NOT_ALLOWED:
+		return "Method Not Allowed";
+	case HTTP_500_INTERNAL_SERVER_ERROR:
+		return "Internal Server Error";
+	default:
+		return "";
+	}
+#endif
+	return "";
+}
+
+#if defined(CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES)
+#define REASON_PHRASE_MAX_LENGTH sizeof("Internal Server Error")
+#else
+#define REASON_PHRASE_MAX_LENGTH 0
+#endif
 
 static int http1_send_headers(struct http_client_ctx *client, enum http_status status,
 			      const struct http_header *headers, size_t header_count,
@@ -205,7 +240,8 @@ static int http1_send_headers(struct http_client_ctx *client, enum http_status s
 {
 	int ret;
 	bool content_type_sent = false;
-	char http_response[MAX(sizeof(RESPONSE_TEMPLATE_DYNAMIC_PART1) + sizeof("xxx"),
+	char http_response[MAX(sizeof(RESPONSE_TEMPLATE_DYNAMIC_PART1) + sizeof("xxx") +
+			       REASON_PHRASE_MAX_LENGTH,
 			       CONFIG_HTTP_SERVER_MAX_HEADER_LEN + 2)];
 
 	if (status < HTTP_100_CONTINUE || status > HTTP_511_NETWORK_AUTHENTICATION_REQUIRED) {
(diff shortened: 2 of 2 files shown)


rlubos (commented):
Thanks for the PR! Generally no objections, but since reason phrase is optional in the response, I think we should make this feature configurable with Kconfig? It seems to be quite costly memory-wise. Compared to main, on `nucleo_h723zg` it consumes extra ~2,5 kB of FLASH.

subsys/net/lib/http/http_server_http1.c
```diff
@@ -196,9 +196,148 @@ static int handle_http1_static_resource(
 	ret; })
 
 #define RESPONSE_TEMPLATE_DYNAMIC_PART1                                                            \
-	"HTTP/1.1 %d\r\n"                                                                          \
+	"HTTP/1.1 %d %s\r\n"                                                                       \
 	"Transfer-Encoding: chunked\r\n"
 
+static const char *http_status_reason_phrase(enum http_status status)
```
carlescufi:
```suggestion
static const char *http_status_str(enum http_status status)
```

subsys/net/lib/http/http_server_http1.c
```diff
@@ -219,7 +358,8 @@ static int http1_send_headers(struct http_client_ctx *client, enum http_status s
 	}
 
 	/* Send response code and transfer encoding */
-	snprintk(http_response, sizeof(http_response), RESPONSE_TEMPLATE_DYNAMIC_PART1, status);
+	snprintk(http_response, sizeof(http_response), RESPONSE_TEMPLATE_DYNAMIC_PART1, status,
+		 http_status_reason_phrase(status));
```
carlescufi:
```suggestion
		 http_status_str(status));
```

subsys/net/lib/http/http_server_http1.c
kfirous: Thanks! Fixed

subsys/net/lib/http/http_server_http1.c
kfirous: Thanks! Fixed

subsys/net/lib/http/Kconfig
```diff
@@ -231,6 +231,16 @@ config HTTP_SERVER_STATIC_FS_RESPONSE_SIZE
 	  Please note that it is allocated on the stack of the HTTP server thread,
 	  so CONFIG_HTTP_SERVER_STACK_SIZE has to be sufficiently large.
 
+config HTTP_SERVER_COMPLETE_STATUS_PHRASES
+	bool "Complete HTTP status reason phrases"
+	default n
```
rlubos: Please remove `default n`, it's implicit.

subsys/net/lib/http/http_server_http1.c
```diff
@@ -196,16 +196,45 @@ static int handle_http1_static_resource(
...
+}
+
 static int http1_send_headers(struct http_client_ctx *client, enum http_status status,
 			      const struct http_header *headers, size_t header_count,
 			      struct http_resource_detail_dynamic *dynamic_detail)
 {
 	int ret;
 	bool content_type_sent = false;
-	char http_response[MAX(sizeof(RESPONSE_TEMPLATE_DYNAMIC_PART1) + sizeof("xxx"),
+	char http_response[MAX(sizeof(RESPONSE_TEMPLATE_DYNAMIC_PART1) + 64,
```
jukkar: We try to avoid having magic constants that do not really explain the purpose of the value.
So here I would suggest something like this
```
#if defined(CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES)
#define MAX_STATUS_STR_LEN 64
#else
#define MAX_STATUS_STR_LEN 0
#endif

char http_response[MAX(sizeof(RESPONSE_TEMPLATE_DYNAMIC_PART1) + sizeof("xxx") +
                   MAX_STATUS_STR_LEN),
```
so I would leave the sizeof() that tries to tell that the status code is 3 chars long.

subsys/net/lib/http/http_server_http1.c
```diff
@@ -196,16 +196,45 @@ static int handle_http1_static_resource(
 	ret; })
 
 #define RESPONSE_TEMPLATE_DYNAMIC_PART1                                                            \
-	"HTTP/1.1 %d\r\n"                                                                          \
+	"HTTP/1.1 %d %s\r\n"                                                                       \
```
jukkar: In order to avoid confusion in the HTTP parser that receives this data, I would not place empty char after the status code (happens if `CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES` is not set).
So better would be
```
"HTTP/1.1 %d%s%s\r\n"
```

subsys/net/lib/http/http_server_http1.c
rlubos: No status string is close to 64 bytes, no need to allocate that much. I'd suggest to define a symbol that represents the maximum length of a status string, and just add it to the response length, something like:
```c
#if defined(CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES)
#define REASON_PHRASE_MAX_LENGTH sizeof("Internal Server Error")
#else
#define REASON_PHRASE_MAX_LENGTH 0
#endif   
```

subsys/net/lib/http/http_server_http1.c
```diff
@@ -219,7 +248,8 @@ static int http1_send_headers(struct http_client_ctx *client, enum http_status s
 	}
 
 	/* Send response code and transfer encoding */
-	snprintk(http_response, sizeof(http_response), RESPONSE_TEMPLATE_DYNAMIC_PART1, status);
+	snprintk(http_response, sizeof(http_response), RESPONSE_TEMPLATE_DYNAMIC_PART1, status,
+		 http_status_str(status));
```
jukkar, line 260: and then here add extra space if `CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES` is enabled like this
```
IS_ENABLED(CONFIG_HTTP_SERVER_COMPLETE_STATUS_PHRASES) ? " " : "", http_status_str(status));
```

subsys/net/lib/http/Kconfig
jukkar: The "default n" can be omitted because options are "n" by default so we do not add these lines to Kconfig files.

subsys/net/lib/http/http_server_http1.c
rlubos: So now we're adding a dummy space after the status code if the reason phrases are disabled, that doesn't sound right. 

I see two options here, either replace the space inbetween code and string with `%s` and let the `snprintk()` below specify a space or an empty string based on configuration, or simply have two templates and use them in `snprintk()` conditionally.

subsys/net/lib/http/http_server_http1.c
kfirous: Done

subsys/net/lib/http/Kconfig
kfirous: Removed

subsys/net/lib/http/http_server_http1.c
kfirous: Went for the first option as @jukkar suggested.

rlubos (approved):
Thanks!

jukkar (approved):

PR #99604, opened 2025-11-18
net: Disable namespace compatibility support for tests 

jukkar:
Do not enable network compatibility mode flag for network tests so that we will get better coverage of the wrong network APIs used.


pdgendt (commented):
The order of the commits is wrong for bisectability, you need to disable the Kconfig symbol off as the last commit.

pdgendt (dismissed):

aescolar (approved):

rlubos (approved):

pdgendt (approved):

ssekar15 (approved):

SeppoTakalo (approved):

PR #94478, opened 2025-08-13
serial: uart_native_pty: IRQ support

tpambor:
Add support for the interrupt-driven API. Interrupts are emulated using a polling thread.

This was merged in a previous version as #93957 but caused problems, see #94425 and reverted in #94426.

The issue stemmed from false RX interrupts being triggered when `select` indicated that data was available to read, but the subsequent read operation failed. In the updated implementation, a dedicated polling thread now feeds incoming data into a ring buffer. Interrupts are no longer triggered directly by select; instead, the ring buffer is used to signal valid RX interrupts, ensuring more reliable and accurate handling.

(Locally) this is now passing `samples/sensor/sensor_shell` and `tests/drivers/can/host`, which previously failed in #94425.

PR #99606, opened 2025-11-18
[backport-4.1] manual backport of commit 7c0e4aecace9287bcd0508e991672a732fc95ddc

kartben:
This is to fix a test that works fine on a local computer but not in CI. This is currently preventing CI on 4.1 branch from being green.

I have confirmed in an earlier push that a dummy change to the samples/sensor/sensor_shell/pytest/test_sensor_shell.py file triggers the test ... and it fails. The second push with the cherrypicked commit makes CI happy.

### DIFF
--- a/samples/sensor/sensor_shell/pytest/test_sensor_shell.py
+++ b/samples/sensor/sensor_shell/pytest/test_sensor_shell.py
@@ -11,6 +11,7 @@ logger = logging.getLogger(__name__)
 def test_sensor_shell_info(shell: Shell):
     logger.info('send "sensor info" command')
 
+    shell.wait_for_prompt()
     lines = shell.exec_command('sensor info')
     assert any(['device name: sensor@0' in line for line in lines]), 'expected response not found'
     assert any(['device name: sensor@1' in line for line in lines]), 'expected response not found'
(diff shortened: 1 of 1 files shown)


JarmouniA:
> Maybe you are already aware of this, but in case you aren't, the commit in question was reverted in #94426, a new version of the fix was merged in #94478.

Ah sorry, discard that, only df1e6f5290f1d60c02a71ba1872ab0521e1f6ba1 was reverted.

tpambor:
The problem was in the `uart_native_pty` driver which didn't correctly receive incoming data. This was fixed then in #94478. 7c0e4aecace9287bcd0508e991672a732fc95ddc wasn't reverted as it is not causing any harm, but it helped with stability/synchronization so that with it the test did actually pass with the first buggy version of the uart driver. With the fixed driver the test passed also without 7c0e4aecace9287bcd0508e991672a732fc95ddc. Errors were different like
```
DEBUG    twister_harness.helpers.shell:shell.py:49 Got prompt
DEBUG    twister_harness.device.device_adapter:device_adapter.py:150 #: uart:~$ ean show can
DEBUG    twister_harness.device.device_adapter:device_adapter.py:150 #: ean: command not found
```
and the test here does not fail for native_sim, so that I think that this is probably a different issue here.

fabiobaltieri (approved):

PR #99612, opened 2025-11-18
Bluetooth: BAP: Fix bad check for ASE state for CIS connect

Thalley:
A CIS may be connected in either the QoS Configured state or the enabling state. The QoS Configured state is the earliest state it is allowed, due to it being the first state where the CIS_ID and ASE_ID are paired.
The enabling state is the "last" state it is allowed, as if the ASE is in the streaming, disabling or releasing state we should not allow/expect a CIS connection to happen.


larsgk (approved):

koffes (approved):

asbjornsabo (approved):
LGTM

PR #35, opened 2017-05-02
ext: simplelink: Restructure CC3220SDK as SimpleLink SDK

GAnthony:
The SimpleLink SDK actually encapsulates a family of devices,
of which cc32xx is just one device.  Other devices can fit
under this SDK directory structure.

This expansion will also allow the import of the WiFi
host driver and its driver porting layer in the future,
inserted at the correct levels in the SimpleLink SDK
directory heirarchy.

Follow the URL (ending in "#directory-structure") referenced in
ext/hal/ti/simplelink/README which explains the SDK structure,
and where devices fit in.

Jira: ZEP-1958

Change-Id: I16515d3e3779de0d55d3b1b8e25029609d1f66c3
Signed-off-by: Gil Pitney <gil.pitney@linaro.org>

---
This change is [Reviewable](https://reviewable.io/reviews/zephyrproject-rtos/zephyr/35)

PR #99620, opened 2025-11-19
drivers: usb: udc: Add UDPHS driver

MCHP-MPU-Solutions-SHA:
Add driver support for Sama7g5 USB UDC (UDPHS).

Many thanks.
Xing.


drivers/usb/udc/Kconfig.mchp
```diff
@@ -0,0 +1,47 @@
...
+	default 8
+	help
+	  Device controller driver thread priority.
+
+config UDC_MCHP_SAM_BURST_LOCK
+	bool "Lock bus access for maximum bandwidth"
+	default n
+	help
+	  USB packets system data bursts will be locked for
+	  maximum optimization of the bus bandwidth usage.
```
tmon-nordic, line 40: I have no idea what this really means. From code it seems that you set some bit. Is it essentially affecting DMA operation/memory access arbitrage?

MCHP-MPU-Solutions-SHA: Hi, this is a chip related feature, for more details please refer to the [datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MPU32/ProductDocuments/DataSheets/SAMA7G5-Series-Data-Sheet-DS60001765.pdf), see bit BURST_LCK in register UDPHS_DMACONTROL.
This will allow the UDPHS DMA channel has max burst length when accessing the bus.

And I made a switch for this feature in the Kconfig, customer can enable it for higher USB bandwidth under heavy system load.

dts/arm/microchip/sam/sama7g5.dtsi
```diff
@@ -711,5 +717,31 @@
 				#size-cells = <0>;
 			};
 		};
+
+		udphsa: usbd@200000 {
```
nandojve: Convention is order by address.

dts/bindings/usb/microchip,sam-udc.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+description: Microchip SAM UDPHS USB Device High Speed Port
+
+compatible: "microchip,sam-udphs"
+
+include: [usb-ep.yaml, pinctrl-device.yaml]
```
nandojve: https://github.com/zephyrproject-rtos/zephyr/blob/db0c34d836aeb79ddbdff0b219499009eebc59d9/dts/bindings/ethernet/atmel%2Cgmac-common.yaml#L5-L7

```diff
-include: [usb-ep.yaml, pinctrl-device.yaml]
+include:
+    - name: usb-ep.yaml
+    - name: pinctrl-device.yaml
```

soc/microchip/sam/sama7/sama7g5/soc.c
```diff
@@ -65,6 +65,9 @@ static const struct arm_mmu_region mmu_regions[] = {
 		   (MMU_REGION_FLAT_ENTRY("pwm", PWM_BASE_ADDRESS, 0x500,
 					  MT_STRONGLY_ORDERED | MPERM_R | MPERM_W),))
 
+	MMU_REGION_FLAT_ENTRY("rstc", RSTC_BASE_ADDRESS, 0x10,
```
nandojve, line 89: by convention we are enabling this before devicetree binding. Move this commit to 2nd position.

Note: Adjust the commit message body to be close to the 72 character per line.

drivers/usb/udc/Kconfig.mchp_sam
```diff
@@ -0,0 +1,47 @@
...
+
+config UDC_MCHP_SAM_THREAD_PRI
+	int "UDC controller driver thread priority"
+	default 8
+	help
+	  Device controller driver thread priority.
+
+config UDC_MCHP_SAM_BURST_LOCK
+	bool "Lock bus access for maximum bandwidth"
+	default n
```
nandojve: I think the _default n_ can be dropped because it is the default. See [redundant-defaults](https://docs.zephyrproject.org/latest/build/kconfig/tips.html#redundant-defaults). Drop all redundant defaults.

drivers/usb/udc/udc_mchp_udphs_g1.c
```diff
@@ -0,0 +1,1611 @@
...
+	uint8_t nr_banks:2;
+	uint8_t can_dma:1;
+	uint8_t high_bw:1;
+	uint8_t ep_size:4;
+#define SZ_64   6  /* 1 << 6  */
+#define SZ_512  9  /* 1 << 9  */
+#define SZ_1024 10 /* 1 << 10 */
+};
+
+static const struct udphs_ep_desc sam_ep_desc[UDPHS_EPT_NUMBER] = {
```
nandojve, line 99: By this table the driver will only expose HS, no FS. Is that right ?

nandojve (changes requested):
Add a report from `sample/subsys/usb/testusb` HS. FS is it should work like that too.
A full report usually tests from 0 up to 29.

dts/arm/microchip/sam/sama7g5.dtsi
MCHP-MPU-Solutions-SHA: The devices in sama7g5.dtsi have been listed in alphabetical order, so I put the udphs device at the end of it.
Do you still need me order the udphs devices by space address?

dts/bindings/usb/microchip,sam-udc.yaml
MCHP-MPU-Solutions-SHA: Updated.

soc/microchip/sam/sama7/sama7g5/soc.c
MCHP-MPU-Solutions-SHA: Commits order has been adjusted.
Commit message body has been adjusted.

drivers/usb/udc/Kconfig.mchp_sam
MCHP-MPU-Solutions-SHA: All redundant defaults have been removed.

drivers/usb/udc/udc_mchp_udphs_g1.c
MCHP-MPU-Solutions-SHA: No, this table exposed all endpoints for HS and FS.
These endpoints can be used in  HS mode and FS mode.

MCHP-MPU-Solutions-SHA:
> Add a report from `sample/subsys/usb/testusb` HS. FS is it should work like that too. A full report usually tests from 0 up to 29.

Hi Gerson,
I DO got issue when testing with testusb (in linux/tools/usb).
Please share it if you have any comment?

PC log:
// Linux 202 6.14.0-35-generic #35~24.04.1-Ubuntu
///////////////////////////////////
user@202:~/repo/linux/tools/usb$ sudo ./testusb -D /dev/bus/usb/001/103 
./testusb: /dev/bus/usb/001/103 may see only control tests
high speed      /dev/bus/usb/001/103    0
/dev/bus/usb/001/103 test 0,    0.000009 secs
/dev/bus/usb/001/103 test 9,    3.001323 secs
--- halt here ---
//////////////////////////////////

ARM log:
/////////////////////////////////
haha setup: 80 06 00 04 00 00 09 00 <-- this is the setup package from PC
haha desc_type 4 <-- this is the request type from PC, and the type "4" doesn't support by Zephyr, check [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/usb/usb_ch9.h#L96) 
haha errno=-134
Set halt ep80
////////////////////////////////
Zephyr return "-ENOTSUP (-134)" in [here](github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/device_next/usbd_ch9.c#L839)

I also checked the "Table 9-4" in the "usb_20.pdf", yes the type 4 is "Reserved for future use".
I'm not a USB expert, why PC Host send an unexpected desc_type to me?

Many thanks.
Xing.

drivers/usb/udc/Kconfig.mchp_sam
```diff
@@ -0,0 +1,45 @@
...
+	bool "Lock bus access for maximum bandwidth"
+	help
+	  USB packets system data bursts will be locked for
+	  maximum optimization of the bus bandwidth usage.
+
+config UDC_MCHP_SAM_RAW_DEBUG
+	bool "UDC controller driver raw debug"
+	depends on LOG_MODE_IMMEDIATE
+	help
+	  Enable driver raw debugging for short and real-time log outputs.
```
jfischer-no: Remove this option. No idea what you are meaning with "raw debugging" or "real-time log outputs", but there is no reason to overwrite LOG_DBG with printk() neither when using LOG_MODE_IMMEDIATE nor without it, or using LOG_MODE_IMMEDIATE.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
+/*
+ * Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
jfischer-no: `SPDX-FileCopyrightText: Copyright Microchip Technology Inc. and its subsidiaries`

https://docs.zephyrproject.org/latest/contribute/guidelines.html#copyright-and-license-notices

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/drivers/gpio.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(udc_mchp_sam, CONFIG_UDC_DRIVER_LOG_LEVEL);
+
+#ifdef CONFIG_UDC_MCHP_SAM_RAW_DEBUG
+#undef LOG_DBG
+#define LOG_DBG(fmt, args...) printk(fmt "\n", ##args)
+#endif
```
jfischer-no: Remove these lines.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+
+#define ENABLE	1
+#define DISABLE	0
+
+#define SETUP_PKT_SIZE	sizeof(struct usb_setup_packet)
+
+#define EP(addr)	USB_EP_GET_IDX(addr)
```
jfischer-no: I do not see any benefits having these macros, please remove them and use 1, 0, true, false, sizeof(struct usb_setup_packet), sizeof(*setup).

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+#define DMA(addr)	(EP(addr) - 1)
+#define MAX_DMA_LEN	((UDPHS_DMACONTROL_BUFF_LENGTH_Msk >> UDPHS_DMACONTROL_BUFF_LENGTH_Pos) + 1)
+#define BYTE_COUNT(x)	(((x) & UDPHS_EPTSTA_BYTE_COUNT_Msk) >> UDPHS_EPTSTA_BYTE_COUNT_Pos)
+#define BUSY_BANKS(x)	(((x) & UDPHS_EPTSTA_BUSY_BANK_STA_Msk) >> UDPHS_EPTSTA_BUSY_BANK_STA_Pos)
+#define BUFF_COUNT(x)	(((x) & UDPHS_DMASTATUS_BUFF_COUNT_Msk) >> UDPHS_DMASTATUS_BUFF_COUNT_Pos)
```
jfischer-no: Prefix them with UDC_SAM, for those that are too long, please use following style
```c
#define UDC_SAM_MAX_DMA_LEN	\
	((UDPHS_DMACONTROL_BUFF_LENGTH_Msk >> UDPHS_DMACONTROL_BUFF_LENGTH_Pos) + 1)
```

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	struct k_event events;
+	atomic_t xfer_new;
+	atomic_t xfer_finished;
+
+	struct gpio_callback vbus_cb;
+	const struct gpio_dt_spec *vbus_gpio;
+	uint8_t vbus_state;
+
+	enum udc_bus_speed speed;
+	char setup[SETUP_PKT_SIZE];
```
jfischer-no: `uint8_t setup[SETUP_PKT_SIZE];`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 9 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 10 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 11 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 12 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 13 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 14 */
+	{ .nr_banks = 1, .ep_size = SZ_512 },                              /* ep 15 */
+};
+
+static inline udphs_registers_t *base_reg(const struct udc_sam_config *config)
```
jfischer-no: If that is really needed, it should use `const struct device *const dev` as parameter and not `const struct udc_sam_config *config`. Same for all the shorthands/helpers below.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static inline int ep_size(const struct udc_sam_config *config, const uint8_t ep)
+{
+	const struct udphs_ep_desc *ep_desc = &config->ep_desc[ep];
+
+	return 1 << ep_desc->ep_size;
+}
+
+static inline int ep_can_dma(const struct udc_sam_config *config, const uint8_t ep)
```
jfischer-no: `_can_` or `_is_` should return bool type, for example `static inline bool ep_can_dma()`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	const struct udphs_ep_desc *ep_desc = &config->ep_desc[ep];
+
+	return ep_desc->nr_banks;
+}
+
+static inline int ep_size(const struct udc_sam_config *config, const uint8_t ep)
+{
+	const struct udphs_ep_desc *ep_desc = &config->ep_desc[ep];
+
+	return 1 << ep_desc->ep_size;
```
jfischer-no: `return BIT(ep_desc->ep_size);`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	udphs_dma_registers_t *dma;
+	int i;
+
+	udphs->UDPHS_CTRL = UDPHS_CTRL_DETACH_Msk;
+	udphs->UDPHS_CTRL |= UDPHS_CTRL_EN_UDPHS_Msk;
+	udphs->UDPHS_IEN = UDPHS_IEN_ENDRESET_Msk;
+	udphs->UDPHS_CLRINT = UDPHS_CLRINT_Msk;
+	udphs->UDPHS_EPTRST = UDPHS_EPTRST_EPT__Msk;
+
+	for (i = 0; i < UDPHS_DMA_NUMBER; i++) {
```
jfischer-no: `for (int i = 0; i < UDPHS_DMA_NUMBER; i++) {`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+static void udphs_pullup_en(udphs_registers_t *const udphs)
+{
+	udphs->UDPHS_CTRL |= UDPHS_CTRL_PULLD_DIS_Msk;
+	udphs->UDPHS_CTRL &= ~UDPHS_CTRL_DETACH_Msk;
+}
+
+#define SPEED_NORMAL		0
+#define SPEED_FORCE_HIGH	2
+#define SPEED_FORCE_FULL	3
+static void udphs_speed_mode(udphs_registers_t *const udphs, int mode)
```
jfischer-no: static void udphs_speed_mode(udphs_registers_t *const udphs, uint32_t mode)

drivers/usb/udc/udc_mchp_udphs_g1.c
```diff
@@ -0,0 +1,1611 @@
...
+	udphs->UDPHS_CTRL &= ~UDPHS_CTRL_DETACH_Msk;
+}
+
+#define SPEED_NORMAL		0
+#define SPEED_FORCE_HIGH	2
+#define SPEED_FORCE_FULL	3
+static void udphs_speed_mode(udphs_registers_t *const udphs, int mode)
+{
+	udphs->UDPHS_TST &= ~UDPHS_TST_SPEED_CFG_Msk;
+	udphs->UDPHS_TST |= UDPHS_TST_SPEED_CFG(mode);
```
jfischer-no, line 232: Please implement test_mode, it is required for the high speed devices.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	udphs->UDPHS_TST |= UDPHS_TST_SPEED_CFG(mode);
+}
+
+static void udphs_set_address(udphs_registers_t *const udphs, unsigned char addr)
+{
+	udphs->UDPHS_CTRL &= ~UDPHS_CTRL_DEV_ADDR_Msk;
+	if (addr != 0) {
+		udphs->UDPHS_CTRL |= UDPHS_CTRL_FADDR_EN_Msk |
+				     UDPHS_CTRL_DEV_ADDR(addr);
+	}
```
jfischer-no: Move these lines to udc_sam_set_address()

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static void udphs_send_wakeup(udphs_registers_t *const udphs)
+{
+	udphs->UDPHS_CTRL |= UDPHS_CTRL_REWAKEUP_Msk;
```
jfischer-no: Move it to udc_sam_host_wakeup()

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static void udphs_reset_ep(udphs_registers_t *const udphs, int ep)
```
jfischer-no: `ep` is usually endpoint address, perhaps idx fits better

`static void udphs_reset_ep(udphs_registers_t *const udphs, uint8_t idx)`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static void udphs_reset_ep_all(udphs_registers_t *const udphs)
+{
+	for (int i = 0; i < UDPHS_EPT_NUMBER; i++) {
+		udphs_reset_ep(udphs, i);
+	}
+}
+
+static void udphs_clear_ep_status(udphs_registers_t *const udphs, int ep)
```
jfischer-no: static void udphs_clear_ep_status(udphs_registers_t *const udphs, uint8_t idx)

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+{
+	udphs->UDPHS_EPTRST = UDPHS_EPTRST_EPT_0_Msk << ep;
+}
+
+static int udphs_config_ep(udphs_registers_t *const udphs, int ep, struct udphs_ep_config *cfg)
```
jfischer-no: There is no need to artificially separate `udc_sam` and `udphs_`. `int ep` parameter likely wrong, called from udc_sam_ep_enable() the argument `ep` is uint8_t type, and can be 0x81 for example. Please move these lines to udc_sam_ep_enable() and get rid of `struct udphs_ep_config`.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	udphs->UDPHS_IEN |= UDPHS_IEN_EPT_0_Msk << ep;
+
+	return 0;
+}
+
+static void udphs_rx_setup(udphs_registers_t *const udphs)
+{
+	udphs_ept_registers_t *const ept =  &udphs->UDPHS_EPT[0];
+
+	ept->UDPHS_EPTCTLENB = UDPHS_EPTCTLENB_RX_SETUP_Msk;
```
jfischer-no: Move to sam_rx_setup_pkt()

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+}
+
+static void udphs_ep_fifo_in(udphs_ept_registers_t *const ept, int enable)
```
jfischer-no: bool enable

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+{
+	if (enable) {
+		ept->UDPHS_EPTCTLENB = UDPHS_EPTCTLENB_TXRDY_Msk;
+	} else {
+		ept->UDPHS_EPTCTLDIS = UDPHS_EPTCTLDIS_TXRDY_Msk;
+	}
+}
+
+static void udphs_ep_fifo_out(udphs_ept_registers_t *const ept, int enable)
```
jfischer-no: bool enable

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+static void udphs_ep_set_halt(udphs_ept_registers_t *const ept)
+{
+	ept->UDPHS_EPTSETSTA = UDPHS_EPTSETSTA_FRCESTALL_Msk;
+}
+
+static void udphs_ep_clear_halt(udphs_ept_registers_t *const ept)
+{
+	ept->UDPHS_EPTCLRSTA = UDPHS_EPTCLRSTA_TOGGLESQ_Msk |
+			       UDPHS_EPTCLRSTA_FRCESTALL_Msk;
+}
```
jfischer-no: move to udc_sam_ep_set_halt/udc_sam_ep_clear_halt

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+	uint8_t ep = EP(ep_cfg->addr);
+	udphs_ept_registers_t *const ept = ep_reg(config, ep);
+	unsigned int lock_key;
+
+	LOG_DBG("Prep IN ep%02x %u %s %s", ep_cfg->addr, buf->len,
+					   ep_can_dma(config, ep) ? "dma" : "fifo",
+					   udc_ep_buf_has_zlp(buf) ? "zlp" : "");
+
+	if (buf->len == 0) {
+		udc_ep_buf_set_zlp(buf);
```
jfischer-no: Why? Drivers should not use it.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1611 @@
...
+		lock_key = irq_lock();
+		udphs_ep_fifo_in(ept, ENABLE);
+	}
+
+	irq_unlock(lock_key);
+
+	return 0;
+}
+
+static void udc_ep0_internal(const struct device *dev, int enable)
```
jfischer-no: `static void udc_ep0_internal(const struct device *dev, const bool enable)`

jfischer-no (changes requested):

nandojve:
Hi Xing,

> Is it mandatory to pass all tests?
Yes, you need to fix them. There is no point to introduce a driver that we already know that have problems.

drivers/usb/udc/Kconfig.mchp_sam
MCHP-MPU-Solutions-SHA: Hi Johann,

Thank you very much for the detailed review.
This feature has been removed.

Many thanks.
Xing.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: These helpers will be kept.
Parameters of them has been updated.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_udphs_g1.c
MCHP-MPU-Solutions-SHA: Yes, I will implement it.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: May I keep these lines in udphs_ function?
The idea is put all register accessing related codes in the udphs_ prefixed function, and the udphs_ functions works like HAL interface function.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: May I keep these lines in udphs_ function?
The idea is put all register accessing related codes in the udphs_ prefixed function, and the udphs_ functions works like HAL interface function.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Understand, all ep and idx variables have been reorganized.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Please may I keep these lines in udphs_ function?
The idea is put all register accessing related codes in the udphs_ prefixed function, and the udphs_ functions works like HAL interface function.

 All ep and idx variables have been reorganized.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Please may I keep these lines in udphs_ function?
The idea is put all register accessing related codes in the udphs_ prefixed function, and the udphs_ functions works like HAL interface function.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Please may I keep these lines in udphs_ function?
The idea is put all register accessing related codes in the udphs_ prefixed function, and the udphs_ functions works like HAL interface function.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Fixed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: I use "bi->zlp" to notify handle_ep_irq() to send a ZLP when an IN packet with "buf->len == 0" comes up. 
(I don't anticipate that "buf->len == 0" and "bi->zlp == true" will come up at the same time in an IN packet.) 
Or should I create a new variable to use for the notification?

nandojve:
needs rebase

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: No, there is no need to replicate HAL madness in a native driver. It is like artificially worsening it.

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: I did not get it, why should a driver do something like that? There is nothing wrong with zero length IN transfer, ZLP flag is set by the stack in special cases like for  control transfer handling.

MCHP-MPU-Solutions-SHA:
> usbd_handle_ctrl_xfer

I just rebase this PR, please help to check if issue can be reproduced with CDC NCM.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Yes, it's weird to set the ZLP flag in a UDC controller driver, I have remove these code, and added the xfer_zero flag for internal use in the driver.

boards/microchip/sam/sama7g54_ek/sama7g54_ek.dts
```diff
@@ -418,3 +428,17 @@
...
 };
+
+zephyr_udc0: &udphsa {
+	vbus-gpios = <&piod 11 GPIO_ACTIVE_HIGH>;
+	pinctrl-names = "default";
+	pinctrl-0 = <&pinctrl_usba_vbus_det>;
+	status = "okay";
+};
+
+&udphsb {
```
nandojve: Should we define the

`zephyr_udc1: &udphsb {` ?

CC: @jfischer-no

nandojve (changes requested):
Hi Xing,

Thank you for the changes. From my side just minor.

boards/microchip/sam/sama7g54_ek/sama7g54_ek.dts
MCHP-MPU-Solutions-SHA: Yes, DTS has been updated.

cssedev:
> > I see some errors in the log when using the driver with CDC NCM. I do not have another board to check if this is a USB stack error or a driver error, so it may not be related to this PR.
> > `net_buf: usbd_handle_ctrl_xfer():1227: buf 0x60322258 double free`
> 
> Issue should be caused by this code: https://github.com/MCHP-MPU-Solutions-SHA/zephyr/blob/sama7g5_udphs/drivers/usb/udc/udc_mchp_sam.c#L646 And has been fixed in the latest PR.
> 
> Please check and feedback. Many thanks for your report.

The PR seems to work as expected during initial testing on top of 23892b0, with the error no longer present.

dts/bindings/usb/microchip,sam-udc.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
```
nandojve: take a note to update the years in all files.

```diff
-# Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
+# Copyright (C) 2025-2026 Microchip Technology Inc. and its subsidiaries
```

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1653 @@
...
+
+	ctrl = UDPHS_DMACONTROL_BUFF_LENGTH(req->len) |
+	       UDPHS_DMACONTROL_END_BUFFIT_Msk |
+	       UDPHS_DMACONTROL_END_B_EN_Msk |
+	       UDPHS_DMACONTROL_CHANN_ENB_Msk;
+	if (!req->is_in) {
+		ctrl |= UDPHS_DMACONTROL_END_TR_IT_Msk |
+			UDPHS_DMACONTROL_END_TR_EN_Msk;
+	}
+#ifdef CONFIG_UDC_MCHP_SAM_BURST_LOCK
```
nandojve: empty line before/after `#ifdef/#endif`

nandojve (changes requested):

dts/bindings/usb/microchip,sam-udc.yaml
MCHP-MPU-Solutions-SHA: Updated, many thanks.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Updated, many thanks.

nandojve: Change from
```diff
	}
+
#ifdef CONFIG_UDC_MCHP_SAM_BURST_LOCK
	ctrl |= UDPHS_DMACONTROL_BURST_LCK_Msk;
#endif
+
	(void)dma->UDPHS_DMASTATUS;
```

to

```diff
	}
+
	if (CONFIG_UDC_MCHP_SAM_BURST_LOCK) {
		ctrl |= UDPHS_DMACONTROL_BURST_LCK_Msk;
	}
+
	(void)dma->UDPHS_DMASTATUS;
```

Same as in Linux, Kconfig symbols can be used like this which makes code clearly and readable.

boards/microchip/sam/sama7g54_ek/sama7g54_ek.yaml
```diff
@@ -18,4 +18,5 @@ supported:
   - sdhc
   - shell
   - uart
+  - udc
```
nandojve: https://github.com/zephyrproject-rtos/zephyr/blob/b55a618c64f2d1e6e3865befe4ce86409bd139ec/boards/atmel/sam0/same54_xpro/same54_xpro.yaml#L26

nandojve (changes requested):

drivers/usb/udc/Kconfig.mchp
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+config UDC_MCHP_SAM
+	bool "Microchip SAM UDPHS USB Device High Speed Port"
+	default y
+	depends on DT_HAS_MICROCHIP_SAM_UDPHS_ENABLED
+	select CACHE_MANAGEMENT
```
JarmouniA:
```suggestion
	depends on CACHE_MANAGEMENT
```
Should be selected by SoC.

soc/microchip/sam/sama7/sama7g5/soc.c
```diff
@@ -102,6 +108,18 @@ static const struct arm_mmu_region mmu_regions[] = {
 		   (MMU_REGION_FLAT_ENTRY("trng", TRNG_BASE_ADDRESS, 0x100,
 					  MT_STRONGLY_ORDERED | MPERM_R | MPERM_W),))
 
+	IF_ENABLED(DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(udphsa)),
```
JarmouniA, line 117: check by compatible

soc/microchip/sam/sama7/sama7g5/soc.c
```diff
@@ -102,6 +108,18 @@ static const struct arm_mmu_region mmu_regions[] = {
...
+		(MMU_REGION_FLAT_ENTRY("udphsa", UDPHSA_BASE_ADDRESS, 0x400,
+				       MT_STRONGLY_ORDERED | MPERM_R | MPERM_W),
+		 MMU_REGION_FLAT_ENTRY("udphsa_ram", UDPHS_RAMA_ADDR, 0x100000,
+				       MT_STRONGLY_ORDERED | MPERM_R | MPERM_W),))
+
+	IF_ENABLED(DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(udphsb)),
```
JarmouniA, line 127: ditto

soc/microchip/sam/sama7/sama7g5/soc.c
```diff
@@ -185,4 +203,23 @@ void soc_early_init_hook(void)
...
+
+	/* Enable USB PHYs clock */
+	PMC_REGS->PMC_XTALF = PMC_XTALF_XTALF_F24M;
+	RSTC_REGS->RSTC_GRSTR |= RSTC_GRSTR_USB_RST1_Msk |
+				 RSTC_GRSTR_USB_RST2_Msk |
+				 RSTC_GRSTR_USB_RST3_Msk;
+	SFR_REGS->SFR_UTMI0R[0] |= SFR_UTMI0R_COMMONONN_Msk;
+	SFR_REGS->SFR_UTMI0R[1] |= SFR_UTMI0R_COMMONONN_Msk;
+	SFR_REGS->SFR_UTMI0R[2] |= SFR_UTMI0R_COMMONONN_Msk;
+	if (DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(udphsa))) {
```
JarmouniA, line 229: ditto

soc/microchip/sam/sama7/sama7g5/soc.c
```diff
@@ -185,4 +203,23 @@ void soc_early_init_hook(void)
...
+		SFR_REGS->SFR_UTMI0R[0] |= SFR_UTMI0R_VBUS_Msk;
+		SFR_REGS->SFR_UTMI0R[0] &= ~SFR_UTMI0R_COMMONONN_Msk;
+		RSTC_REGS->RSTC_GRSTR &= ~RSTC_GRSTR_USB_RST1_Msk;
+	}
+	if (DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(udphsb))) {
```
JarmouniA, line 234: ditto

JarmouniA (dismissed):

boards/microchip/sam/sama7g54_ek/sama7g54_ek.yaml
MCHP-MPU-Solutions-SHA: > https://github.com/zephyrproject-rtos/zephyr/blob/b55a618c64f2d1e6e3865befe4ce86409bd139ec/boards/atmel/sam0/same54_xpro/same54_xpro.yaml#L26

Fixed, many thanks.

drivers/usb/udc/Kconfig.mchp
MCHP-MPU-Solutions-SHA: > Should be selected by SoC.

Fixed, many thanks.

boards/microchip/sam/sama7g54_ek/sama7g54_ek.yaml
MCHP-MPU-Solutions-SHA: Hi,
I have rebase this PR.
For the Kconfig symbol with bool type, if the value is false (default n), then I can NOT find it in:
build/zephyr/include/generated/zephyr/autoconf.h

Is there a misunderstanding here? I would appreciate your guidance.
Many thanks.

soc/microchip/sam/sama7/sama7g5/soc.c
MCHP-MPU-Solutions-SHA: Hi,
will DT_HAS_COMPAT_STATUS_OKAY is recommended?

The compatible is the driver name, but we have two device controllers will share the same driver, and these two controllers have different register addresses.
(In sama7g5.dtsi file you will find two device will use this driver, they are "udphsa" and "udphsb")
With DT_HAS_COMPAT_STATUS_OKAY, How to identify which controller's register address need to be mapped?

With DT_NODE_HAS_STATUS_OKAY, we can identify it by the node's label name (device name).

Is there a misunderstanding here? I would appreciate your guidance.
Many thanks.

JarmouniA: > will DT_HAS_COMPAT_STATUS_OKAY is recommended?

Yes

> The compatible is the driver name

No, represents the hw device independently from the driver used https://docs.zephyrproject.org/latest/build/dts/bindings-syntax.html#compatible

> but we have two device controllers will share the same driver, and these two controllers have different register addresses.
(In sama7g5.dtsi file you will find two device will use this driver, they are "udphsa" and "udphsb")
With DT_HAS_COMPAT_STATUS_OKAY, How to identify which controller's register address need to be mapped?

In this instance, the address should come from the DT device node directly instead of using HAL macros.

JarmouniA: For clocks setup below, I wonder why are they enabled in SoC code instead of device driver, and why the info of which clock to enable is not coming from DT.
But if that's a deeper problem that requires a larger fix beyond the scope of the PR, you can keep the use of node labels.

MCHP-MPU-Solutions-SHA: Many thanks for comments.
The Reg address and size has been aligned with DTS.
I have to keep use of node labels, because need to identify which device node is enabled in the DTS. But please share it if there is more comment for it.

For the clock setup below are due to a defect in the clock/PMC driver. Once the clock/PMC driver update is complete, these GCLK codes will be removed.

nandojve (dismissed):
@jfischer-no , @JarmouniA 

Is there anything open from your side?

dts/arm/microchip/sam/sama7/sama7g5/sama7g5.dtsi
```diff
@@ -44,6 +50,32 @@
...
+		udphsb: usbd@300000 {
+			compatible = "microchip,sam-udphs";
+			reg = <0x00300000 0x100000
+			       0xe0818000 0x400>;
+			interrupt-parent = <&gic>;
+			interrupts = <GIC_SPI 105 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
+			clocks = <&pmc PMC_TYPE_PERIPHERAL 105>, <&usb_clk>;
+			clock-names = "pclk", "hclk";
+			maximum-speed = "high-speed";
+			num-bidir-endpoints = <16>;
```
jfischer-no, line 75: If the number of endpoints always 16, then do not use `num-bidir-endpoints` property and do not include `usb-ep.yaml`

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+	.shutdown = udc_sam_shutdown,
+	.set_address = udc_sam_set_address,
+	.host_wakeup = udc_sam_host_wakeup,
+	.ep_enable = udc_sam_ep_enable,
+	.ep_disable = udc_sam_ep_disable,
+	.ep_set_halt = udc_sam_ep_set_halt,
+	.ep_clear_halt = udc_sam_ep_clear_halt,
+	.ep_enqueue = udc_sam_ep_enqueue,
+	.ep_dequeue = udc_sam_ep_dequeue,
+};
```
jfischer-no, line 1533: test_mode implementation is missed, https://github.com/zephyrproject-rtos/zephyr/pull/99620/commits/4e90b4be5fa581716e589c04e825a1c308f1d5c3#r2560534330

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: ^ not addressed.

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: ^ not addressed.

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: ^ not addressed.

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: ^ not addressed.

drivers/usb/udc/udc_mchp_sam.c
jfischer-no: ^ not addressed.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+		/*
+		 * The END_TR_ST interrupt comes from the last transfmission,
+		 * just ignore it.
+		 */
+		return 0;
+	}
+
+	if ((status & ctrl) & (UDPHS_DMASTATUS_END_BF_ST_Msk | UDPHS_DMASTATUS_END_TR_ST_Msk)) {
+		if (ctrl & UDPHS_DMACONTROL_END_TR_EN_Msk) { /* Is OUT */
+			if (!atomic_test_bit(&priv->xfer_running,
```
jfischer-no: This code can be simplified. Invert the first condition, exit there. Move OUT/IN handling to own functions.

drivers/usb/udc/udc_mchp_udphs_g1.c
```diff
@@ -0,0 +1,1652 @@
...
+
+		if (status & UDPHS_INTSTA_SPEED_Msk) {
+			priv->speed = UDC_BUS_SPEED_HS;
+		} else {
+			priv->speed = UDC_BUS_SPEED_FS;
+		}
+
+		udc_ep0_internal(dev, 1);
+		udc_submit_event(dev, UDC_EVT_RESET, 0);
+	}
```
jfischer-no, line 951: SOF interrupts are missed.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+	data->caps.out_ack = true;
+	data->caps.mps0 = UDC_MPS0_64;
+	if (config->vbus_gpio.port) {
+		data->caps.can_detect_vbus = true;
+	}
+	if (config->speed_idx == UDC_BUS_SPEED_HS) {
+		data->caps.hs = true;
+	}
+
+	for (i = 0; i < UDPHS_EPT_NUMBER / 2; i++) {
```
jfischer-no: How I understood, there are 16 "hw endpoint blocks".
Each of these "hw" endpoints can be IN or OUT?
Or are there 8 dedicated IN endpoints and 8 dedicated OUT endpoints?
Or are there 16 "hw" endpoints that can handle up to 16 IN and 16 OUT endpoints?

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+	if (evt & BIT(UDC_THREAD_EVT_SETUP)) {
+		k_event_clear(&priv->events, BIT(UDC_THREAD_EVT_SETUP));
+
+		ret = udc_handle_evt_setup(dev, priv->setup);
+		if (ret) {
+			udc_submit_event(dev, UDC_EVT_ERROR, ret);
+		}
+	}
+
+	if (udc_ctrl_stage_is_setup(dev)) {
```
jfischer-no: That does not seem to be right. Why do you need to always check it?

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+
+			ret = udc_ctrl_feed_dout(dev, 0);
+			if (ret == -ENOMEM) {
+				udc_submit_ep_event(dev, buf, ret);
+			} else {
+				return ret;
+			}
+		}
+
+		if ((udc_ctrl_stage_is_setup(dev)) && (buf->data)) {
```
jfischer-no: What happens here?

drivers/usb/udc/udc_common.c
```diff
@@ -924,6 +924,13 @@ int udc_ctrl_submit_status(const struct device *dev,
 	return udc_submit_ep_event(dev, buf, 0);
 }
 
+bool udc_ctrl_stage_is_setup(const struct device *dev)
```
jfischer-no: I looked how you used it in the driver and it looks a bit fishy. I doubt this function is needed. You could move it to your driver for now. This part is likely to be reworked anyway.

jfischer-no (changes requested):

nandojve:
Hi @MCHP-MPU-Solutions-SHA ,

Could you rebase and address the messages?
Now, you need adapt the driver to due to the #103493 and #100876

MCHP-MPU-Solutions-SHA:
> Hi @MCHP-MPU-Solutions-SHA ,
> 
> Could you rebase and address the messages? Now, you need adapt the driver to due to the #103493 and #100876

Hi Gerson,

This PR has been rebase and re-fine for #103493 and #100876.
Please help to review again.
Many thanks.

dts/arm/microchip/sam/sama7/sama7g5/sama7g5.dtsi
MCHP-MPU-Solutions-SHA: This cannot be guaranteed (it is not always 16); different IP instances on the same chip may have different EP values.
I have add num_of_eps variable in udc_sam_config for taking over the "num-bidir-endpoints" from the DTS.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: test_mode function has been implemented.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: These HAL function has been removed.

dma_irq handler function has been updated.

drivers/usb/udc/udc_mchp_sam.c
```diff
@@ -0,0 +1,1652 @@
...
+
+		if (status & UDPHS_INTSTA_SPEED_Msk) {
+			priv->speed = UDC_BUS_SPEED_HS;
+		} else {
+			priv->speed = UDC_BUS_SPEED_FS;
+		}
+
+		udc_ep0_internal(dev, 1);
+		udc_submit_event(dev, UDC_EVT_RESET, 0);
+	}
```
MCHP-MPU-Solutions-SHA: SOF event report function has been added.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: In the last commit, the Setup package cannot be processed in time, so I disable the Setup package interrupt until the current one is finished.

With this rebase, issue cannot be reproduced anymore, and this code has been removed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: caps.out_ack has bee set to "true" in this UDC driver, in the last commit, the udc_ctrl_update_stage() won't help to free the allocated buffer, then I have to free it manually. 

With this rebase, issue cannot be reproduced anymore, and this code has been removed.

drivers/usb/udc/udc_common.c
MCHP-MPU-Solutions-SHA: This code has been removed.

drivers/usb/udc/udc_mchp_sam.c
MCHP-MPU-Solutions-SHA: Yes, I DO want to discuss about these weird codes.
There are 16 EPs in an UDPHS instance:
EP0 --> control endpoint, Bidirectional (IN and OUT)
EP1 ~ EP15 --> data endpoint, Unidirectional (IN or OUT), must be configured as IN or OUT before using.
(May I say that the EP0 is full-duplex and EP1 ~ EP15 is half-duplex?)

For EP1 ~ EP15, once an EP has been configured as IN, then it can not be used as OUT at the same time, the reverse is also true.
So I have to register these EPs in this way.

Please what does the "num-bidir-endpoints" mean in the usb-ep.yaml?
Does it mean the number of bidirectional EPs?
Or it means the number of EPs which can be configured as IN or OUT?

Many thanks.

MCHP-MPU-Solutions-SHA:
Hi @jfischer-no,

Thank you very much for the very good reference driver udc_sam0.c, it has greatly accelerated our progress.
This PR has bee rebase, please help to review again.
Many thanks.

dts/bindings/usb/microchip,sam-udphs.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2025-2026 Microchip Technology Inc. and its subsidiaries
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+description: Microchip SAM UDPHS USB Device High Speed Port
```
nandojve: Add examples entry:

https://github.com/zephyrproject-rtos/zephyr/blob/7bbe9da0892c2c3bc2cdefbc5e60b941f9a36da4/dts/bindings/usb/atmel%2Csam-udp.yaml#L21-L40

nandojve (changes requested):
Is the the same UDPHS from Arduino Due?
If that is the case, could you add the entries for that board too?

dts/bindings/usb/microchip,sam-udphs.yaml
MCHP-MPU-Solutions-SHA: Example for dts file has been added.

MCHP-MPU-Solutions-SHA:
> Is the the same UDPHS from Arduino Due? If that is the case, could you add the entries for that board too?
I google it and found that SAM3X8E was used on this board, then downloaded the SAM3X8E datasheet from microchip.com.
In the datasheet, I couldn't found a USB controller with name UDPHS.
And UOTGHS was used by SAM3X8E, but its registers are not compatible with UDPHS.

Please correct me if there is issue.

nandojve (dismissed):
LGTM!

MCHP-MPU-Solutions-SHA:
@jfischer-no @tmon-nordic
Hi, could you please help to review this PR?

dts/bindings/usb/microchip,sam-udphs.yaml
```diff
@@ -0,0 +1,46 @@
...
+  interrupts:
+    required: true
+
+  clocks:
+    required: true
+
+  dma-disable:
+    type: boolean
+    description: |
+      Force disable DMA function, data will be transferred via FIFO.
```
tmon-nordic: Is this configuration option, or can there be hardware that cannot use DMA?

tmon-nordic (commented):
Needs to be rebased to fix conflicts.

dts/bindings/usb/microchip,sam-udphs.yaml
MCHP-MPU-Solutions-SHA: This is a software configuration that allows customers to force the disabling of the DMA feature (which is enabled by default).
Currently, all IP instances support DMA feature.

MCHP-MPU-Solutions-SHA:
> Needs to be rebased to fix conflicts.

PR has been rebase.

dts/bindings/usb/microchip,sam-udphs.yaml
tmon-nordic: Should this be then converted to Kconfig symbol?

MCHP-MPU-Solutions-SHA: Of course, we can convert it to a Kconfig symbol, but doing so would make it a global option for all instances on the chip.
Put it in the dts file that allows each instance can ON/OFF it flexible.

Please let me know if you want to put it into the Kconfig file.

tmon-nordic: This is the ever-going topic about what should be part of devicetree and what should not be. If you think that Kconfig disadvantage of being global is not a big deal, then convert it to Kconfig symbol so it is clear that this is user configuration and not hardware feature.

nandojve: @tmon-nordic ,

I don't think using a global Kconfig works here. This is a selection between DMA or FIFO usage per instance.
I'll suggest to @MCHP-MPU-Solutions-SHA define not a boolean for this, instead an enum with string type. This allows a default value but user can switch per instance between the default and the another method.

MCHP-MPU-Solutions-SHA: Many thanks for this comment, so the general rule is all features and options related to Hardware (device) will be placed in DTS, while all fetures and options related to Software (driver) will be placed in Kconfig.

The "disable-dma" is a Software feature and I have put it into the Kconfig.mchp.

MCHP-MPU-Solutions-SHA: @nandojve Sorry, I didn't refresh this web page, so I didn't see your comment right away.
I added the “disable-dma” option because the DMA feature had been problematic (Cache Coherency Issues) right up until the last UDC stack update ([#103493](https://github.com/zephyrproject-rtos/zephyr/pull/103493) and [#100876](https://github.com/zephyrproject-rtos/zephyr/pull/100876)).
After updating the UDC stack, the issue disappeared without any modifications to the SAM UDC driver. To prevent this issue from recurring, I added this switch (disable-dma) to the MCHP SAM UDC driver to disable the DMA functionality within the driver.
Once the DMA function of SAM UDC driver doesn't work stable (due to a specific UDC stack update), then the customer can disable it via "disable-dma" and continue to use the FIFO function as a workaround.

It's OK for me to make it as a global switch for the SAM UDC driver.
Please advice if there is anything.
Many thanks.

drivers/usb/udc/Kconfig.mchp
```diff
@@ -0,0 +1,43 @@
...
+config UDC_MCHP_SAM_THREAD_PRI
+	int "UDC controller driver thread priority"
+	default 8
+	help
+	  Device controller driver thread priority.
+
+config UDC_MCHP_SAM_DISABLE_DMA
+	bool "Disable DMA transfer"
+	help
+	  Force disable DMA function, data will be transferred via FIFO.
```
tmon-nordic: Kconfig is more about enabling options, having a option that needs to be enabled to effectively disable is a bit strange.

DWC2 has `config UDC_DWC2_DMA` with `default y`. I think same approach can be used here. The driver internal details are fine to stay with `dma_disable`.

MCHP-MPU-Solutions-SHA: Updated, the driver also has been updated align with Kconfig.

nandojve:
fix compliance checks

MCHP-MPU-Solutions-SHA:
> fix compliance checks

Issue has been fixed.

nandojve (dismissed):

nandojve:
Ping @jfischer-no , @tmon-nordic

tmon-nordic (dismissed):

MCHP-MPU-Solutions-SHA:
@jfischer-no Could you please help to review this case?
All required changes has been updated / fixed.

nandojve:
What still missing here?

nandojve:
Hi @jfischer-no ,
Could you revisit this PR? Is still something missing ?

MCHP-MPU-Solutions-SHA:
Hi, @tmon-nordic,
Could you please help to review and merge this PR?
Many thanks.

nandojve (approved):

nandojve:
@tmon-nordic ,

Could you re-approve?

NhMchp (approved):

jfischer-no (approved):

PR #99623, opened 2025-11-19
drivers: sensor: st: i3g4250d: Migrate SPI support to stmemsc from custom implementation

adityaganesh:
This removes the SPI custom implementation and uses stmemsc which makes portability better and makes it easier to add support for i2c which is not added yet for this sensor. Below is the screenshot of com port of stm32f411e_disco on which this was tested
Screenshot 2025-11-18 203330


adityaganesh:
Hello @avisconti were you able to review this PR . Let me know if there are any issues that needs to be fixed

drivers/sensor/st/i3g4250d/i3g4250d.h
```diff
@@ -12,23 +12,38 @@
...
+
+
+#if I3G4250D_ANY_INST_ON_BUS_STATUS_OKAY(spi)
+#include <zephyr/drivers/spi.h>
+#endif /* I3G4250D_ANY_INST_ON_BUS_STATUS_OKAY(spi) */
+
 struct i3g4250d_device_config {
-	struct spi_dt_spec spi;
+	stmdev_ctx_t ctx;
+	#if I3G4250D_ANY_INST_ON_BUS_STATUS_OKAY(spi)
```
avisconti: I think you need to remove this one and keep only the other inside the union. This because if in the future we require to add i2c then it will work just adding the i2c part

avisconti (changes requested):
Just a small modification. Rest seems good.
Thanks!

adityaganesh:
Hey @avisconti I have made the changes, and I am attaching a screenshot below of the console logs after the changes were made 
Screenshot 2026-01-17 192118

adityaganesh:
Hey @avisconti I have squashed both the commits into a single commit . Let me know if there is something else that needs to be done

avisconti (approved):

MaureenHelm (approved):

PR #99627, opened 2025-11-19
arch: arm: disable userspace for m2l31x

ccli8:
This SoC has a limited number of only 4 MPU regions and cannot support USERSPACE related tests and samples. Disable this feature to avoid runtime error which usually reports nothing silently.


ccli8:
To support a few cases where 4 MPU regions can still support USERSPACE program,, cancel this PR.

ithinuel:
Note that if this is causing issues for testing in CI, a valid solution is to have the affected platform exclude/filter those tests.

IIRC there are a few ways to achieve this via filters, tags, platform_exclude etc.

PR #99631, opened 2025-11-19
dts: arm: nxp: disabled gpio node for imx95 m7

yongxu-wang15:
disabled gpio node in imx95 m7 which not owner gpio privilege in defualt system manager config, the status should be set as okay in specific case test instead of nxp_imx95_m7.dtsi

### DIFF
--- a/dts/arm/nxp/nxp_imx95_m7.dtsi
+++ b/dts/arm/nxp/nxp_imx95_m7.dtsi
@@ -132,6 +132,7 @@
 			gpio-controller;
 			#gpio-cells = <2>;
 			ngpios = <16>;
+			status = "disabled";
 		};
 
 		gpio2: gpio@43810000 {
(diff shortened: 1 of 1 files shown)


dts/arm/nxp/nxp_imx95_m7.dtsi
```diff
@@ -132,6 +132,7 @@
 			gpio-controller;
```
dbaluta, line 132: Use imperative mood for the subject. disabled -> disable.

Otherwise LGTM.

There is also a question if there is any dts that relies on this being already enabled but I think this change is correct and the respective dtses should be fixed instead.

yongxu-wang15: sure, updated commit description now, and I found this issue in zephyr/tests/kernel/timer/timer_behavior usecase, the case open CONFIG_GPIO=y, and as the default dts enabled gpio1 node, so it will induce run fail issue, so disabled  it in dtsi file, thanks

dbaluta (approved):

yangbolu1991 (approved):

JiafeiPan (approved):

PR #99637, opened 2025-11-19
drivers: misc: stm32n6_axisram: fix clock config info

etienne-lms:
Correct clock config info recently modified that should use STM32_DT_INST_CLOCK_INFO_BY_NAME() macro insteead of STM32_CLOCK_INFO_BY_NAME() macro.


mathieuchopstm (approved):

etienne-lms:
I've append**ed** a commit for a build-all test covering stm32n6_axisram driver.
(thanks @erwango for the suggestion)

erwango (approved):

tests/drivers/build_all/misc/prj.conf
```diff
@@ -0,0 +1,2 @@
+CONFIG_TEST=y
+CONFIG_TEST_USERSPACE=y
```
etienne-lms, line 2: I'm not sure these configs are needed or of any interest, at least regarding what's currently built among the misc drivers.

erwango: Indeed,I was wondering why these were there. It doesn't do harm I guess, but could also be removed.

gautierg-st (approved):

PR #99638, opened 2025-11-19
soc: sifli: sf32: Add SF32LB52 platform ftab generation

HalfSweet:
Using the `gen_ftab.py` script, we can retrieve the current image distribution information from the device tree and generate the corresponding `ftab.bin` file. This eliminates the need to burn an additional `ftab.bin` externally before running the device

### DIFF
--- a/scripts/west_commands/runners/sftool.py
+++ b/scripts/west_commands/runners/sftool.py
@@ -1,8 +1,10 @@
 # Copyright (c) 2025 Core Devices LLC
+# Copyright (c) 2025 SiFli Technologies(Nanjing) Co., Ltd
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
 import shlex
+import subprocess
 
 from runners.core import RunnerCaps, RunnerConfig, ZephyrBinaryRunner
 
@@ -18,6 +20,7 @@ class SftoolRunner(ZephyrBinaryRunner):
         erase: bool,
         dt_flash: bool,
         tool_opt: list[str],
+        flash_files: list[str],
     ) -> None:
         super().__init__(cfg)
 
@@ -25,6 +28,7 @@ class SftoolRunner(ZephyrBinaryRunner):
         self._port = port
         self._erase = erase
         self._dt_flash = dt_flash
+        self._flash_files = flash_files
 
         self._tool_opt: list[str] = []
         for opts in [shlex.split(opt) for opt in tool_opt]:
@@ -53,6 +57,13 @@ class SftoolRunner(ZephyrBinaryRunner):
             required=True,
             help="Serial port device, e.g. /dev/ttyUSB0",
         )
+        parser.add_argument(
+            "--flash-file",
+            dest="flash_files",
+            action="append",
+            default=[],
+            help="Additional file@address entries that must be flashed before the Zephyr image",
+        )
 
     @classmethod
     def do_create(cls, cfg: RunnerConfig, args: argparse.Namespace) -> "SftoolRunner":
@@ -63,6 +74,7 @@ class SftoolRunner(ZephyrBinaryRunner):
             erase=args.erase,
             dt_flash=args.dt_flash,
             tool_opt=args.tool_opt,
+            flash_files=args.flash_files,
         )
 
     def do_run(self, command: str, **kwargs):
@@ -71,16 +83,25 @@ class SftoolRunner(ZephyrBinaryRunner):
         cmd = [sftool, "--chip", self._chip, "--port", self._port]
         cmd += self._tool_opt
 
-        if self._erase:
-            self.check_call(cmd + ["erase_flash"])
+        flash_targets: list[str] = []
+        flash_targets.extend(self._flash_files)
 
         if self.cfg.file:
-            self.check_call(cmd + ["write_flash", self.cfg.file])
+            flash_targets.append(self.cfg.file)
         else:
             if self.cfg.bin_file and self._dt_flash:
                 addr = self.flash_address_from_build_conf(self.build_conf)
-                self.check_call(cmd + ["write_flash", f"{self.cfg.bin_file}@0x{addr:08x}"])
+                flash_targets.append(f"{self.cfg.bin_file}@0x{addr:08x}")
             elif self.cfg.hex_file:
-                self.check_call(cmd + ["write_flash", self.cfg.hex_file])
+                flash_targets.append(self.cfg.hex_file)
             else:
                 raise RuntimeError("No file available for flashing")
+
+        write_args = ["write_flash"]
+        if self._erase:
+            write_args.append("-e")
+
+        full_cmd = cmd + write_args + flash_targets
+        self._log_cmd(full_cmd)
+        if not self.dry_run:
+            subprocess.run(full_cmd, check=True)
--- a/soc/sifli/sf32/gen_ftab.py
+++ b/soc/sifli/sf32/gen_ftab.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2025 SiFli Technologies(Nanjing) Co., Ltd
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Generate the SF32LB52 platform ftab image from devicetree information.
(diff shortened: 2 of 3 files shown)


scripts/west_commands/runners/sftool.py
```diff
@@ -74,6 +85,9 @@ def do_run(self, command: str, **kwargs):
         if self._erase:
             self.check_call(cmd + ["erase_flash"])
 
+        for extra in self._flash_files:
+            self.check_call(cmd + ["write_flash", extra])
+
         if self.cfg.file:
             self.check_call(cmd + ["write_flash", self.cfg.file])
```
gmarull: wouldn't it be more efficient to write all files in a single invocation? IIRC sftool supports that

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
```diff
@@ -1,7 +1,43 @@
...
+  COMMENT "Generating SF32LB52x ftab image"
+  VERBATIM
+)
+
+add_custom_target(sf32lb52x_ftab DEPENDS ${SF32LB52X_FTAB_BIN})
+add_dependencies(zephyr_interface sf32lb52x_ftab)
+
+string(MAKE_C_IDENTIFIER sftool SF32LB52X_SFTOOL_RUNNER_ID)
+set_property(GLOBAL APPEND PROPERTY BOARD_RUNNER_ARGS_${SF32LB52X_SFTOOL_RUNNER_ID}
+  --flash-file=${SF32LB52X_FTAB_BIN}@${SF32LB52X_PTABLE_ADDR_HEX})
```
gmarull: wouldn't it be easier for sftool runner to check for file existence, and append it to the list of files to flash if found?

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
```diff
@@ -1,7 +1,43 @@
...
+
+dt_chosen(SF32LB52X_CODE_NODE PROPERTY "zephyr,code-partition")
+dt_reg_addr(SF32LB52X_CODE_OFFSET PATH ${SF32LB52X_CODE_NODE})
+math(EXPR SF32LB52X_CODE_START "${SF32LB52X_FLASH_BASE} + ${SF32LB52X_CODE_OFFSET}")
+
+dt_nodelabel(SF32LB52X_PTABLE_NODE NODELABEL ptable)
+dt_reg_addr(SF32LB52X_PTABLE_OFFSET PATH ${SF32LB52X_PTABLE_NODE})
+math(EXPR SF32LB52X_PTABLE_ADDR "${SF32LB52X_FLASH_BASE} + ${SF32LB52X_PTABLE_OFFSET}")
+math(EXPR SF32LB52X_PTABLE_ADDR_HEX_INT "${SF32LB52X_PTABLE_ADDR}" OUTPUT_FORMAT HEXADECIMAL)
+string(REPLACE "0X" "0x" SF32LB52X_PTABLE_ADDR_HEX ${SF32LB52X_PTABLE_ADDR_HEX_INT})
```
gmarull: question: why is this needed given gen_ftab can directly parse edt pickle file? runner has also access to edt (some use it AFAIK)

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
HalfSweet: Do you mean that when sftool runner runs, it searches the build directory to see if the sf32lb52x_ftab.bin file exists?
However, this raises an issue: we don't know which address the ftab should be flashed to. If the runner were to retrieve device tree information, wouldn't that create excessive coupling? I prefer sftool runner to maintain only core functionality, with additional files generated during the build process passed in via standardized interfaces—rather than having the runner guess which files exist

scripts/west_commands/runners/sftool.py
HalfSweet: You're right, I'll merge it

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
HalfSweet: I will attempt to keep all device tree processing within gen_ftab

scripts/west_commands/runners/sftool.py
HalfSweet: done.

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
gmarull: Fair point. Could gen ftab generate a hex instead, so we do not have to care about addresses in CMake?

HalfSweet: It seems like it might not be that easy to achieve without introducing any third-party libraries... I'll give it a shot.

gmarull: Zephyr default reqs include IntelHex ;-)

HalfSweet: done.

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
```diff
@@ -1,7 +1,33 @@
...
+add_custom_command(
+  OUTPUT ${SF32LB52X_FTAB_HEX}
+  DEPENDS ${EDT_PICKLE} ${SF32LB52X_FTAB_SCRIPT}
+  COMMAND ${SF32LB52X_FTAB_GEN_COMMAND}
+  COMMENT "Generating SF32LB52x ftab hex image"
+  VERBATIM
+)
+
+add_custom_target(sf32lb52x_ftab DEPENDS ${SF32LB52X_FTAB_HEX})
+add_dependencies(zephyr_interface sf32lb52x_ftab)
```
gmarull: I wonder if using `POST_BUILD` would be better here, @nordicjm @tejlmand can you please provide some input?

gmarull (dismissed):
lgtm, let's also wait for build system experts to provide some input as well

soc/sifli/sf32/gen_ftab.py
```diff
@@ -0,0 +1,210 @@
...
+import struct
+import sys
+from pathlib import Path
+
+from intelhex import IntelHex
+
+
+def _zephyr_base() -> Path:
+    try:
+        return Path(os.environ["ZEPHYR_BASE"]).resolve()
```
nordicjm, line 26: ~~don't do this (which is completely wrong from 5+ years ago), look at python files in `scripts` to see how to get this~~ apparently this is how it is being done much to my annoyance

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
nordicjm: would suggest using `extra_post_build_commands` - can see it being used here https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/mcuboot.cmake#L199

scripts/west_commands/runners/sftool.py
```diff
@@ -74,13 +85,18 @@ def do_run(self, command: str, **kwargs):
...
                 addr = self.flash_address_from_build_conf(self.build_conf)
-                self.check_call(cmd + ["write_flash", f"{self.cfg.bin_file}@0x{addr:08x}"])
+                flash_targets.append(f"{self.cfg.bin_file}@0x{addr:08x}")
             elif self.cfg.hex_file:
-                self.check_call(cmd + ["write_flash", self.cfg.hex_file])
+                flash_targets.append(self.cfg.hex_file)
             else:
                 raise RuntimeError("No file available for flashing")
+
+        self.check_call(cmd + ["write_flash"] + flash_targets)
```
marc-hb: https://docs.python.org/3/library/subprocess.html#subprocess.run

> The recommended approach to invoking subprocesses is to use the run() function for all use cases it can handle.

HalfSweet: done.

soc/sifli/sf32/sf32lb52x/CMakeLists.txt
HalfSweet: done.

gmarull (dismissed):

ck-telecom (approved):

gmarull (approved):

PR #99641, opened 2025-11-19
samples: led_strip: Add Nucleo C071RB overlay

JohnAZoidberg:
Works when connecting LED strip to pins labeled GND, 5V and D11/MOSI


thedjnK (dismissed):

kartben (changes requested):
Thanks for the PR! Unfortunately, we do not add overlay files for stuff that corresponds to "bread boarding" that one has manually done on their board as this simply wouldn't scale 
Overlays in code samples are for things that are physically available on a board but need extra tweaking and configuration to be fully enabled. 
Hope this makes sense :)

samples/drivers/led/led_strip/boards/nucleo_c071rb.overlay
```diff
@@ -0,0 +1,38 @@
...
+#include <zephyr/dt-bindings/led/led.h>
+
+&spi1 {
+	/* Use pin PA7, labeled as D11/MOSI on the C071RB */
+	pinctrl-0 = <&spi1_nss_pa4 &spi1_sck_pa5
+		     &spi1_miso_pa6 &spi1_mosi_pa7>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	led_strip: ws2812@0 {
```
fabiobaltieri, line 20: does this work reliably for you? context: I made a keyboard with exactly this micro and and the string of pixels would eventually glitch out due to the spi stream breaking up for long enough for the ws2812 protocol to reset, the fix was to setup a DMA channel for this, add this here:

```
        dmas = <&dmamux1 0 17 (STM32_DMA_PERIPH_TX | STM32_DMA_PRIORITY_LOW)>,
               <&dmamux1 1 16 (STM32_DMA_PERIPH_RX | STM32_DMA_PRIORITY_LOW)>;
        dma-names = "tx", "rx";
```

then enable

```
&dma1 {
	status = "okay";
};
&dmamux1 {
	status = "okay";
};
```

and in the board Kconfig

```
CONFIG_SPI_STM32_DMA=y
```

for reference: https://github.com/fabiobaltieri/zephyr-input-hid/commit/0663a3f20c6228f0b6188f0a308c6857290e2f42

samples/drivers/led/led_strip/boards/nucleo_c071rb.overlay
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/dt-bindings/led/led.h>
+
+&spi1 {
+	/* Use pin PA7, labeled as D11/MOSI on the C071RB */
+	pinctrl-0 = <&spi1_nss_pa4 &spi1_sck_pa5
+		     &spi1_miso_pa6 &spi1_mosi_pa7>;
```
fabiobaltieri: Drop `spi1_nss_pa4` and `spi1_miso_pa6`, you are not using them and they are free for other functions, `sck` is needed though even though the ws2812 protocol does not use it, but the STM32 SPI controller refuses to do anything if the clock pin is not mapped, since you are doing a sample, can you add a comment about it?

fabiobaltieri (commented):
@kartben I know we normally don't do these but setting this up to work correctly on STM32 has a couple details (the clock pin and the DMA setup) which are otherwise not documented anywhere else, I think we should do an exception and let this go through. We have other cases like this that slipped in anyway but this is actually useful.

samples/drivers/led/led_strip/boards/nucleo_c071rb.overlay
JohnAZoidberg: Aha thanks! I didn't know that, so I added all of them.

samples/drivers/led/led_strip/boards/nucleo_c071rb.overlay
JohnAZoidberg: Ok will try this and see if there's a difference.
The `nucleo_f070rb.overlay` does it this way

fabiobaltieri: you may not see any problem with this sample but you would as the mcu gets busy, if you look at the signal on a scope it should go from a jittery mess to a nice regular pulse stream

simonguinot: Yes, you must use DMA to secure the signal.

JohnAZoidberg: It works well with DMA, updated the devicetree.

fabiobaltieri (approved):

kartben (approved):

soburi (approved):

simonguinot (approved):

PR #74077, opened 2024-06-11
samples: boards: stm32: wkup_pins: no need to configure button GPIO pin

ajarmouni-st:
The gpio pin that will be used as a source for PWR wake-up pins does not actually need to be configured as input, hence `input_gpio_keys` driver does not need to be involved. As far as GPIO controller is concerned, the pin can be in analog state.

Also, it's clearer & safer to call `gpio_pin_configure_dt` directly in the application with the `STM32_GPIO_WKUP` flag, just before calling Poweroff, instead of relying on a driver, like `input_gpio_keys` to do it just after boot behind the scenes, & risk not having the correct wake-up pins config when Poweroff is finally called.

PR #99653, opened 2025-11-19
samples: board: st: wkup_pins: configure the GPIO pin as input

mathieuchopstm:
Certain series (e.g., STM32WBA) require pins to be configured as input to behave properly as wake-up pins.

### DIFF
--- a/samples/boards/st/power_mgmt/wkup_pins/src/main.c
+++ b/samples/boards/st/power_mgmt/wkup_pins/src/main.c
@@ -32,7 +32,7 @@ int main(void)
 	gpio_pin_set(led.port, led.pin, 1);
 
 	/* Setup button GPIO pin as a source for exiting Poweroff */
-	gpio_pin_configure_dt(&button, STM32_GPIO_WKUP);
+	gpio_pin_configure_dt(&button, GPIO_INPUT | STM32_GPIO_WKUP);
 
 	printk("Will wait %ds before powering the system off\n", (WAIT_TIME_US / 1000000));
 	k_busy_wait(WAIT_TIME_US);


etienne-lms (approved):
Tested on wba65: indeed needed together with #99656.

samples/boards/st/power_mgmt/wkup_pins/src/main.c
```diff
@@ -32,7 +32,7 @@ int main(void)
 	gpio_pin_set(led.port, led.pin, 1);
 
 	/* Setup button GPIO pin as a source for exiting Poweroff */
-	gpio_pin_configure_dt(&button, STM32_GPIO_WKUP);
+	gpio_pin_configure_dt(&button, GPIO_INPUT | STM32_GPIO_WKUP);
```
JarmouniA, line 35: In my opinion, since this is a quirk of a specific SoC, it should be hidden in the GPIO driver in
https://github.com/zephyrproject-rtos/zephyr/blob/4758cfadf5aa8664bfe61316aa2311fb2deaad0f/drivers/gpio/gpio_stm32.c#L496
When `STM32_GPIO_WKUP` is specified, we configure the pin as input for WBA SoCs.

mathieuchopstm: `GPIO_OUTPUT | STM32_GPIO_WKUP` also works though. If we had a distinct `STM32_GPIO_LP_RETAIN`, I *could* agree with you, but that's heavier in implementation costs.

In the use case demonstrated by this sample, it just doesn't make sense to *not* have `GPIO_INPUT` anyways - you wouldn't have a button *dedicated* to wake-up, it would also serve as general-purpose input. The sample should really be more something like:
```c
void gpio_isr(void) {
   k_sem_give(&btn_sem);
}

gpio_pin_configure_dt(&button, GPIO_INPUT | STM32_GPIO_WKUP);
gpio_interrupt_configure(&button, &gpio_isr);

while (1) {
  /* schedule timeout in 1s */
  /* wait for btn_sem || timeout */
  if (/* timeout */) {
     break;
  } else {
    /* toggle LED */
  }
}

sys_poweroff();

```

mathieuchopstm:
Tested on most boards with an overlay in the sample (all but `stm32l1_disco`; used `nucleo_c031c6` instead of `nucleo_c092rc`) and confirmed no regression.

I'm leaving `HW Test` blocker anyways since this depends on #99656

samples/boards/st/power_mgmt/wkup_pins/src/main.c
JarmouniA: > In the use case demonstrated by this sample, it just doesn't make sense to not have GPIO_INPUT anyways

No, not quite, at least for STM32 Shutdown mode (which was the mode mapped to Zephyr Poweroff when the wake-up pins feature was added), the GPIO periph settings are unrelated to PWR GPIO settings. You could definitely have a pin working as an output when app is running, and used as a wake-up source from Poweroff. #74077

mathieuchopstm: > > In the use case demonstrated by this sample, it just doesn't make sense to not have GPIO_INPUT anyways
> 
> You could definitely have a pin working as an output when app is running, and used as a wake-up source from Poweroff.

Yes (and I had this in mind) but *the use case demonstrated by this sample* is a button which - by definition - is an input :slightly_smiling_face:

> #74077

To quote you:
> As far as GPIO controller is concerned, the pin can be in analog state.

...is an implementation detail that is true on many series but not WBA.

I'd rather have the sample documenting `GPIO_INPUT | STM32_GPIO_WAKEUP` - (hopefully) guaranteed to work on all SoCs - as the "canonical" method for configuring wake-up pins; it doesn't prevent clever people from using `STM32_GPIO_WAKEUP` as a SoC-specific extension, but ensures no one has "broken wake-up pins for *no apparent reason*" if they blindly copy-paste sample code.

(*"`GPIO_INPUT` is required" --> "`GPIO_INPUT` is required on specific SoCs"* is also not a breaking change, unlike the other way around)

JarmouniA: > Yes (and I had this in mind) but the use case demonstrated by this sample is a button which - by definition - is an input

The usage of the button is just a convenience in the sample, so that we can trigger the PWR wake-up pin without having to attach anything else, we are not using the button in the application for anything, so it isn't input by definition.

> is an implementation detail that is true on many series but not WBA

That's why I recommended it gets handled in the driver as a special case, so that previous out-of-tree apps work on WBA as well without needing any modif (the additional `GPIO_INPUT` flag)

erwango: I prefer we don't put too much assumption on the way the HW will be used. Even if this could seems far fetched, using pin as a wake up while being configured as an OUTPUT is a possible use case (for a good reason or because dev is left with no other option due to HW constraints).
So current proposal is fine to me, as long as it's completed with a patch for WBA to report an error if pin is configured as WKUP w/ no additional INPUT/OUTPUT.

mathieuchopstm: I have opened #100532 as follow-up.

erwango (approved):

PR #99654, opened 2025-11-19
subsys/net: Fix EVENTFD kconfig selection

aescolar:
Since 820cd34dbb5d817bacc6cbdeade99aaa01007a5c (#98765) these net components use ZVFS_EVENTFD directly instead of thru EVENTFD. So, let's select ZVFS_EVENTFD directly instead of indirectly through EVENTFD.


jukkar (approved):
Good point, the code used the zvfs variant already but the Kconfig was missing the proper selection.

awojasinski (approved):

rlubos (approved):

pdgendt (approved):

PR #99656, opened 2025-11-19
soc: st: stm32: wkup_pins: enable retention for STM32WBA wake-up pins

mathieuchopstm:
In order for wake-up pins to behave properly in Standby mode on STM32WBA, the I/O Standby Retention must be enabled. (It only comes in effect when the SoC does enter Standby mode, and has no effect otherwise).

### DIFF
--- a/soc/st/stm32/common/stm32_wkup_pins.c
+++ b/soc/st/stm32/common/stm32_wkup_pins.c
@@ -13,6 +13,7 @@
 #include <zephyr/drivers/gpio.h>
 #include <zephyr/devicetree.h>
 
+#include <stm32_bitops.h>
 #include <stm32_ll_system.h>
 #include <stm32_ll_pwr.h>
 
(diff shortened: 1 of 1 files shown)


erwango (approved):

soc/st/stm32/common/stm32_wkup_pins.c
```diff
@@ -326,6 +330,26 @@ int stm32_pwr_wkup_pin_cfg_gpio(const struct gpio_dt_spec *gpio)
...
+			 * a mere SET_BIT() - this avoids adding another
+			 * look-up table. Since this is series-specific
+			 * code anyways, raw register access tricks like
+			 * this one are fine.
+			 *
+			 * See RefMan/PWR registers layout for details.
+			 */
+			volatile uint32_t *ioretenr_x = (&PWR->IORETENRA) + 2 * i;
+
+			stm32_reg_set_bits(ioretenr_x, 1U << gpio->pin);
```
etienne-lms, line 348:
```suggestion
			stm32_reg_set_bits(ioretenr_x, BIT(gpio->pin));
			break;

mathieuchopstm: I prefer the explicit form because it's more obvious about intent.

etienne-lms: Fine with me.

etienne-lms: Maybe the `break` is worth to be added.

etienne-lms (approved):

soc/st/stm32/common/stm32_wkup_pins.c
mathieuchopstm: good catch! I'll throw a little follow-up...

PR #99680, opened 2025-11-19
soc: nxp: s32: S32K3 use early reset to init ECC ram before using it.

PetervdPerk-NXP:
SOC_RESET_HOOK is already too late because the code touches ram already instead use the newer SOC_EARLY_RESET_HOOK so we can do ECC initialization before using the memory.

@manuargue this fixes startup instability with `CONFIG_DEBUG_THREAD_INFO=y` because now ECC RAM gets initialized correctly before that code touches the ram.


bperseghetti (approved):

manuargue (approved):
Thanks @PetervdPerk-NXP!

PR #99681, opened 2025-11-19
scripts: west_commands: blobs: support blobs cache

thorsten-klein:
This PR adds support for cache directories for blobs during `west blobs fetch` via the west config option `blobs.cache-dirs` or cli argument `--cache-dirs`. The variable may contain a single directory or multiple directories separated by `;`.

Additionally, via `blobs.auto-cache` config option or the `--auto-cache` cli argument an auto-cache directory can be specified. Thi auto-cache is automatically filled when a blob is downloaded, whereby the filename in cache is suffixed by the blob's SHA-256 hash (<filename>.<sha>).

`west blobs fetch` will search in all configured cache directories (including the auto-cache). 
The file in the cache directories can be named either as the blob file (e.g. `some_blob.bin.se`) or suffixed with its hashsum (e.g. `some_blob.bin.se.b94c34fc2a638a1713c79b04992020241aad32bab2be4a85623c8b5bb715e36e`)
If the blob is found, it is copied from the cache to the blobs path; otherwise, it is downloaded from the URL as usual.

### Why?

This feature can help to speed up builds and save network bandwidth in a CI environment, as blobs are cached and not re-downloaded. 
It also helps to become independent from 3rd party servers, as blobs can new be pre-fetched, packaged and uploaded to own server. Later, they can be downloaded and unpacked (to one or more directories) and provided as cache directories to `west blobs fetch`.

### Considered Alternative

Another considered solution was the support for some remapping of urls from blobs, e.g. via config file or manifest file. But I have discarded this idea as the solution from this PR is much simpler.

### DIFF
--- a/doc/develop/west/zephyr-cmds.rst
+++ b/doc/develop/west/zephyr-cmds.rst
@@ -220,6 +220,19 @@ the specific blobs that are fetched, by passing a regular expression::
   # For example, only download esp32 blobs, skip the other variants
   west blobs fetch hal_espressif --allow-regex 'lib/esp32/.*'
 
+An auto-cache directory can be provided via the ``--auto-cache`` cli argument
+or via the ``blobs.auto-cache`` config option. When enabled, the auto-cache
+directory is automatically populated whenever a blob is missing and downloaded.
+
+One or more additional cache directories (separated by ``;``) can be provided
+in ``--cache-dirs`` cli argument or ``blobs.cache-dirs`` config option.
+
+``west blobs fetch`` searches all configured cache directories (including the
+auto-cache) for a matching blob filename. Cached files may be stored either
+under their original filename or with a SHA-256 suffix (``<filename>.<sha>``).
+If found, the blob is copied from the cache to the blob path; otherwise
+it is downloaded from its url to the blob path.
+
 .. _west-twister:
 
 Twister wrapper: ``west twister``
--- a/scripts/west_commands/blobs.py
+++ b/scripts/west_commands/blobs.py
@@ -18,7 +18,6 @@ import zephyr_module
 
 
 class Blobs(WestCommand):
-
     DEFAULT_LIST_FMT = '{module} {status} {path} {type} {abspath}'
 
     def __init__(self):
@@ -27,7 +26,8 @@ class Blobs(WestCommand):
             # Keep this in sync with the string in west-commands.yml.
             'work with binary blobs',
             'Work with binary blobs',
-            accepts_unknown_args=False)
+            accepts_unknown_args=False,
+        )
 
     def do_add_parser(self, parser_adder):
         parser = parser_adder.add_parser(
@@ -61,22 +61,30 @@ class Blobs(WestCommand):
             - uri: URI to the remote location of the blob
             - description: blob text description
             - doc-url: URL to the documentation for this blob
-            '''))
+            '''),
+        )
 
         # Remember to update west-completion.bash if you add or remove
         # flags
-        parser.add_argument('subcmd', nargs=1,
-                            choices=['list', 'fetch', 'clean'],
-                            help='sub-command to execute')
+        parser.add_argument(
+            'subcmd', nargs=1, choices=['list', 'fetch', 'clean'], help='sub-command to execute'
+        )
 
-        parser.add_argument('modules', metavar='MODULE', nargs='*',
-                            help='''zephyr modules to operate on;
-                            all modules will be used if not given''')
+        parser.add_argument(
+            'modules',
+            metavar='MODULE',
+            nargs='*',
+            help='''zephyr modules to operate on;
+                    all modules will be used if not given''',
+        )
 
         group = parser.add_argument_group('west blob list options')
-        group.add_argument('-f', '--format',
-                            help='''format string to use to list each blob;
-                                    see FORMAT STRINGS below''')
+        group.add_argument(
+            '-f',
+            '--format',
+            help='''format string to use to list each blob;
+                    see FORMAT STRINGS below''',
+        )
 
         group = parser.add_argument_group('west blobs fetch options')
         group.add_argument(
@@ -86,8 +94,12 @@ class Blobs(WestCommand):
                     Only local paths matching this regex will be fetched.
                     Note that local paths are relative to the module directory''',
         )
-        group.add_argument('-a', '--auto-accept', action='store_true',
-                            help='''auto accept license if the fetching needs click-through''')
+        group.add_argument(
+            '-a',
+            '--auto-accept',
+            action='store_true',
+            help='''auto accept license if the fetching needs click-through''',
+        )
 
         return parser
 
@@ -125,6 +137,7 @@ class Blobs(WestCommand):
         scheme = urlparse(url).scheme
         self.dbg(f'Fetching {path} with {scheme}')
         import fetchers
+
         fetcher = fetchers.get_fetcher_cls(scheme)
 
         self.dbg(f'Found fetcher: {fetcher}')
@@ -139,8 +152,9 @@ class Blobs(WestCommand):
 
         status = zephyr_module.get_blob_status(blob['abspath'], blob['sha256'])
         if status == zephyr_module.BLOB_OUTDATED:
-            self.err(textwrap.dedent(
-                f'''\
+            self.err(
+                textwrap.dedent(
+                    f'''\
                 The checksum of the downloaded file does not match that
                 in the blob metadata:
                 - if it is not certain that the download was successful,
@@ -153,7 +167,9 @@ class Blobs(WestCommand):
                 Module: {blob['module']}
                 Blob:   {blob['path']}
                 URL:    {blob['url']}
-                Info:   {blob['description']}'''))
+                Info:   {blob['description']}'''
+                )
+            )
             return False
         return True
 
@@ -176,9 +192,11 @@ class Blobs(WestCommand):
 
             if blob['click-through'] and not args.auto_accept:
                 while True:
-                    user_input = input("For this blob, need to read and accept "
-                                       "license to continue. Read it?\n"
-                                       "Please type 'y' or 'n' and press enter to confirm: ")
+                    user_input = input(
+                        "For this blob, need to read and accept "
+                        "license to continue. Read it?\n"
+                        "Please type 'y' or 'n' and press enter to confirm: "
+                    )
                     if user_input.upper() == "Y" or user_input.upper() == "N":
                         break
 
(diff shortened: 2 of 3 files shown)


scripts/west_commands/blobs.py
```diff
@@ -121,16 +122,47 @@ def list(self, args):
...
+        url = blob['url']
+        path = Path(blob['path'])
+
+
+        cached_blob = None
+        sha256 = blob['sha256']
+        cached_filenames = [f'{sha256}-{path.name}', path.name]
+
+        # check if the blob is already cached
+        cache_dirs = os.getenv('ZEPHYR_BLOBS_CACHE', '')
```
pdgendt: Try to avoid using environment variables where CLI arguments or `west config` options are perfectly viable.

thorsten-klein: Yes, you are right. I will update it tomorrow to a west config option.

Any wishes for the option name?
``zephyr.blobs-cache-dir``?

pdgendt: Maybe `blobs.cache-dir` is sufficient?

scripts/west_commands/blobs.py
```diff
@@ -121,16 +123,51 @@ def list(self, args):
...
         scheme = urlparse(url).scheme
         self.dbg(f'Fetching {path} with {scheme}')
         import fetchers
         fetcher = fetchers.get_fetcher_cls(scheme)
 
         self.dbg(f'Found fetcher: {fetcher}')
         inst = fetcher()
+
+        # Fill the cache if only one cache dir is used.
+        if not cached_blob and len(cache_dirs) == 1:
```
pdgendt: This feels like a hidden feature, either we need a separate option, or we need to split into two options (similar to `auto-cache` for west)

thorsten-klein: Any proposal for the west config option? Maybe `blobs.auto-cache-dir` which may always contain only one path?

pdgendt: Maybe. Other question: do we need multiple directories?

thorsten-klein: Users might package the blobs of each module separately (and not one big package containing all blobs), so they will most probably be located in different folders later on. That's why I did it this way.

scripts/west_commands/blobs.py
```diff
@@ -121,16 +131,58 @@ def list(self, args):
...
+        auto_cache_dir = args.auto_cache or config.get('blobs', 'auto-cache', fallback=None)
+
+        # convert all cache dirs to Path
+        cache_paths = [Path(p) for p in cache_dirs.split(';') if p]
+        if auto_cache_dir:
+            cache_paths.append(Path(auto_cache_dir))
+
+        # check if the blob is already cached
+        filenames = [f'{sha256}-{path.name}', path.name]
+        for cache_path in cache_paths:
```
pdgendt: The `sha256` should be a suffix rather than a prefix, which would allow alphabetical sorting to list different versions of the same file next to each other.

I'm not convinced we should have `path.name` too, I understand the convenience but this can easily clash for different files or versions.

scripts/west_commands/blobs.py
```diff
@@ -121,16 +131,58 @@ def list(self, args):
...
+
+        cached_blob = None
+        sha256 = blob['sha256']
+
+        # collect all cache dirs from west config and env variable
+        cache_dirs = args.cache_dirs or config.get('blobs', 'cache-dirs', fallback='')
+        auto_cache_dir = args.auto_cache or config.get('blobs', 'auto-cache', fallback=None)
+
+        # convert all cache dirs to Path
+        cache_paths = [Path(p) for p in cache_dirs.split(';') if p]
```
pdgendt: No need to check if the paths exist?

scripts/west_commands/blobs.py
thorsten-klein: > I'm not convinced we should have path.name too, I understand the convenience but this can easily clash for different files or versions.

This is helpful if you package the blobs into a package. It feels strange if the files MUST be renamed to a specific name scheme.

Note that any clashes are avoided since the hash sum is checked in this function, so a cached file is only used if it is really the correct one.

So I would like to keep this, if it is okay for you.

scripts/west_commands/blobs.py
thorsten-klein: Good catch. It makes sense to only append paths that exist in order to optimize the function 👍

UPDATE: I check in the for loop if the paths exist and skip that path if it does not exist.

scripts/west_commands/blobs.py
thorsten-klein: > The sha256 should be a suffix

I have adapted the PR to suffix the hashsum in the end. 
e.g. the file in the cache can be named `blob.bin.se` or `blob.bin.se.b94c34fc2a638a1713c79b04992020241aad32bab2be4a85623c8b5bb715e36e`
Unfortunately I cannot keep the file extension. But this should not be problematic I guess?

scripts/west_commands/blobs.py
```diff
@@ -5,12 +5,14 @@
...
 import os
 import re
+import shutil
 import sys
 import textwrap
 from pathlib import Path
 from urllib.parse import urlparse
 
 from west.commands import WestCommand
+from west.configuration import config
```
marc-hb:
```suggestion
from west.configuration import wconfig
```

`config` is incredibly vague. It is a variable, a function, a module, an object?

scripts/west_commands/blobs.py
```diff
@@ -121,16 +131,58 @@ def list(self, args):
...
+    def fetch_blob(self, args, blob):
+        url = blob['url']
+        path = Path(blob['path'])
+
+        cached_blob = None
+        sha256 = blob['sha256']
+
+        # collect all cache dirs from west config and env variable
+        cache_dirs = args.cache_dirs or config.get('blobs', 'cache-dirs', fallback='')
+        auto_cache_dir = args.auto_cache or config.get('blobs', 'auto-cache', fallback=None)
```
marc-hb: These new configs need some documentation?

scripts/west_commands/blobs.py
```diff
@@ -121,16 +131,58 @@ def list(self, args):
...
+
+        # convert all cache dirs to Path
```
marc-hb:
```suggestion
```
Comments should add information or structure, not paraphrase obvious one-liners.

scripts/west_commands/blobs.py
```diff
@@ -121,16 +131,58 @@ def list(self, args):
...
         scheme = urlparse(url).scheme
-        self.dbg(f'Fetching {path} with {scheme}')
         import fetchers
         fetcher = fetchers.get_fetcher_cls(scheme)
 
         self.dbg(f'Found fetcher: {fetcher}')
         inst = fetcher()
+
+        # Fill the auto-cache if in use
+        if not cached_blob and auto_cache_dir:
```
marc-hb, line 234: Why keep printing "Found fetcher" in the debug logs when it won't be used at all?

I agree with SonarQubeCloud that this function is getting a bit long; you can probably split it in two: 1. `get_blob()` that searches the cache and invokes `fetch_blob()` when not found. Also, probably reserve the term "fetch" for the network and use a different one for the cache to make the distinction clearer.

scripts/west_commands/blobs.py
pdgendt: No need to use the global configuration, each west command has the config, just test `self.has_config` and use `self.config`.

scripts/west_commands/blobs.py
```diff
@@ -100,6 +101,18 @@ def do_add_parser(self, parser_adder):
             action='store_true',
             help='''auto accept license if the fetching needs click-through''',
         )
+        group.add_argument(
+            '--cache-dirs',
+            help='''Semicolon-separated list of directories to search for cached
+                                    blobs before downloading. Cache files may use the original
+                                    filename or be suffixed with `.<sha256>`.''',
```
pdgendt: `help` indentation looks awkward.

pdgendt: There are more `help` indentation issues left.

scripts/west_commands/blobs.py
```diff
@@ -61,22 +62,30 @@ def do_add_parser(self, parser_adder):
...
 
-        parser.add_argument('modules', metavar='MODULE', nargs='*',
-                            help='''zephyr modules to operate on;
-                            all modules will be used if not given''')
+        parser.add_argument(
+            'modules',
+            metavar='MODULE',
+            nargs='*',
+            help='''zephyr modules to operate on;
+                            all modules will be used if not given''',
```
pdgendt: Fix indent

scripts/west_commands/blobs.py
```diff
@@ -61,22 +62,30 @@ def do_add_parser(self, parser_adder):
...
 
         group = parser.add_argument_group('west blob list options')
-        group.add_argument('-f', '--format',
-                            help='''format string to use to list each blob;
-                                    see FORMAT STRINGS below''')
+        group.add_argument(
+            '-f',
+            '--format',
+            help='''format string to use to list each blob;
+                                    see FORMAT STRINGS below''',
```
pdgendt: Ditto

scripts/west_commands/blobs.py
```diff
@@ -121,26 +146,107 @@ def list(self, args):
...
+        path = Path(blob['abspath'])
+
+        # collect existing cache dirs specified as args, otherwise from west config
+        cache_dirs = args.cache_dirs
+        auto_cache_dir = args.auto_cache
+        if self.has_config:
+            if not cache_dirs:
+                cache_dirs = self.config.get('blobs.cache-dirs')
+            if not auto_cache_dir:
+                auto_cache_dir = self.config.get('blobs.auto-cache')
```
pdgendt: If the user passed an empty string, it should take precedence over the config IMO, so comparing to `None`.

```suggestion
            if cache_dirs is None:
                cache_dirs = self.config.get('blobs.cache-dirs')
            if auto_cache_dir is None:
                auto_cache_dir = self.config.get('blobs.auto-cache')
```

Also would be nice to have expanded user home directory, same as `west` caches.

doc/develop/west/zephyr-cmds.rst
```diff
@@ -220,6 +220,19 @@ the specific blobs that are fetched, by passing a regular expression::
   # For example, only download esp32 blobs, skip the other variants
   west blobs fetch hal_espressif --allow-regex 'lib/esp32/.*'
 
+An auto-cache directory can be provided via the ``--auto-cache`` cli argument
+or via the ``blobs.auto-cache-dir`` config option. When enabled, the auto-cache
```
pdgendt: Different name in the implementation

```suggestion
or via the ``blobs.auto-cache`` config option. When enabled, the auto-cache
```

doc/develop/west/zephyr-cmds.rst
```diff
@@ -220,6 +220,19 @@ the specific blobs that are fetched, by passing a regular expression::
...
+directory is automatically populated whenever a blob is missing and downloaded.
+
+One or more additional cache directories (separated by ``;``) can be provided
+in ``--cache-dirs`` cli argument or ``blobs.cache-dir`` config option.
```
pdgendt: Same issue

```suggestion
in ``--cache-dirs`` cli argument or ``blobs.cache-dirs`` config option.
```

carlescufi (approved):
Thanks for this! Checked the code and looks reasonable to me

pdgendt (approved):

PR #99684, opened 2025-11-19
soc: nordic: uicr: Parse pinctrls for local domain peripherals

jonathannilsen:
Generate PERIPHCONF SPU/CTRLSEL entries based on pinctrl properties
on the local domain peripherals as well as the global ones.
This fixes an issue where the required pin configuration was not
generated for the pinctrls on the radio local domain GPIOTE0 on
nrf54h20.

The secure attribute of nodes without an address and without a bus node
are now interpreted as being secure by default, instead of failing with
an error. This prevents the parsing of certain nodes from triggering
failing the build (in particular the 'ieee802154' node in
cpurad_peripherals).

### DIFF
--- a/soc/nordic/common/uicr/periphconf/builder.py
+++ b/soc/nordic/common/uicr/periphconf/builder.py
@@ -150,7 +150,7 @@ class PeriphconfBuilder:
 
         :param node_or_nodelabel: node or label of the node to process.
         """
-        self._add_peripheral_cfg(node_or_nodelabel, is_global=False, add_gpios=False)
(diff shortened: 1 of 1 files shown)


SebastianBoe (dismissed):

nordic-krch (approved):
It fixes the issue with missing UICR_GPIO_PIN_CNF_CTRLSEL_SET entries for GPIOTE0.

masz-nordic (approved):

hakonfam (approved):

PR #99685, opened 2025-11-19
drivers: usb: udc: skeleton: share thread creation code across instances

mathieuchopstm:
Instead of having one "make_thread" function per instance, called through a function pointer in the instance configuration, save the thread stack information instead and consume it from shared code in the driver pre-init function. This makes the driver more readable (less code in the instance creation macro) and should reduce its footprint impact (since the thread creation code is no longer duplicated).

### DIFF
--- a/drivers/usb/udc/udc_skeleton.c
+++ b/drivers/usb/udc/udc_skeleton.c
@@ -43,7 +43,8 @@ struct udc_skeleton_config {
 	size_t num_of_eps;
 	struct udc_ep_config *ep_cfg_in;
 	struct udc_ep_config *ep_cfg_out;
-	void (*make_thread)(const struct device *dev);
+	k_thread_stack_t *thread_stk;
+	size_t thread_stk_sz;
 	int speed_idx;
 };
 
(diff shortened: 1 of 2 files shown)


drivers/usb/udc/udc_skeleton.c
```diff
@@ -390,26 +404,6 @@ static const struct udc_api udc_skeleton_api = {
 	K_THREAD_STACK_DEFINE(udc_skeleton_stack_##n,				\
 			      CONFIG_UDC_SKELETON_STACK_SIZE);			\
 										\
-	static void udc_skeleton_thread_##n(void *dev, void *arg1, void *arg2)	\
-	{									\
-		skeleton_thread_handler(dev);					\
-	}									\
-										\
-	static void udc_skeleton_make_thread_##n(const struct device *dev)	\
```
jfischer-no, line 398: There was a reason for doing it this way, but I do not remember what it was. :crystal_ball:

jfischer-no:
> `tests/drivers/udc` test `drivers.usb.udc.skeleton` does build the skeleton driver, but I don't know if CI builds it or not.
> 
> I did run the test locally to verify nothing was broken (although, in this instance, the thread is useless anyways...):

I thought it would be picked up for the PR run, but I cannot find it in the logs. Any ideas on how to fix that?

drivers/usb/udc/udc_skeleton.c
mathieuchopstm: Possibly to distinguish between threads for different UDCs? (but then I'd expect a `k_thread_set_name()`, and you can also do so using ELF symbols)

tests/drivers/udc/testcase.yaml
```diff
@@ -26,3 +26,5 @@ tests:
       - CONFIG_UDC_BUF_POOL_SIZE=32768
     platform_allow:
       - native_sim/native/64
+    integration_platforms:
+      - native_sim/native/64
```
jfischer-no, line 30: Does it make sense to add `- native_sim` to platform_allow?

```suggestion
      - native_sim
    integration_platforms:
      - native_sim/native/64
```

mathieuchopstm: There shouldn't be any significant difference between `native_sim` and `native_sim//64` as far as the skeleton driver is concerned, so I don't think it's necessary?

jfischer-no (approved):

josuah (approved):

PR #99686, opened 2025-11-19
scripts: gen_cfb_font_header.py: Improve PNG support

msalau:
Changes:
* Allow matrix arrangement of glyphs
* Remove the output file in case of a failure

The current implementation supports only glyphs arranges as a line thus making an image extremely wide.

The commit also fixes a bug in the script that keeps half-generated output files.

### DIFF
--- a/scripts/build/gen_cfb_font_header.py
+++ b/scripts/build/gen_cfb_font_header.py
@@ -126,9 +126,21 @@ def extract_image_glyphs():
     """Extract font glyphs from an image file"""
     image = Image.open(args.input)
 
+    if (image.width % args.width) != 0:
+        raise ValueError("Incorrect image width")
+    if (image.height % args.height) != 0:
+        raise ValueError("Incorrect image height")
+    if (image.width * image.height) < (args.width * args.height * (args.last - args.first + 1)):
+        raise ValueError("Incorrect image size")
+
     x_offset = 0
+    y_offset = 0
     for i in range(args.first, args.last + 1):
-        glyph = image.crop((x_offset, 0, x_offset + args.width, args.height))
+        if x_offset >= image.width:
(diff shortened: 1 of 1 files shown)


scripts/build/gen_cfb_font_header.py
```diff
@@ -6,6 +6,7 @@
 
 import argparse
 import sys
+import os
```
tejlmand: please reorder:
```suggestion
import os
import sys
```

scripts/build/gen_cfb_font_header.py
```diff
@@ -268,7 +281,12 @@ def parse_args():
 def main():
     """Parse arguments and generate CFB font header file"""
     parse_args()
-    generate_header()
+    try:
+        generate_header()
+    except:
+        args.output.close()
+        os.remove(args.output.name)
```
tejlmand: removing a file when something fails can make debugging much harder.

Having a partially generated file / file with invalid content available can sometimes help in understanding what went wrong.

Please explain why you think removing the file in this case is the best option.

Could we perhaps insert some help comments in the generated file when png content is invalid ?

msalau: Hello @tejlmand 

It turns out removing the file is not necessary, since CMake keeps track whether the file has been generated successfully or not (based on the return code of the generator script).

Thank you!
I'll remove this feature.

Best regards,
Maksim

msalau:
Rebased the branch and fixed the code to use more specific `ValueError` instead of generic `Exception` during image size verification.

tejlmand (approved):

nashif (approved):

JarmouniA (commented):
This script should be part of the display area, @tejlmand do you agree?

PR #99699, opened 2025-11-19
net: l2: ieee802154: add support for IEEE802154_CFI_DATA_REQUEST

spiriou:
802.15.4 specifies that a device may send a DATA_REQUEST command to the coordinator after waiting for "macResponseWaitTime", starting after the acknowledgment to the association request command.

This change adds support to generate DATA_REQUEST frames and sends one for the association usecase only. The two other usecases, for MLME-POLL or for beacon-enabled PAN, are not yet supported.

### DIFF
--- a/subsys/net/l2/ieee802154/ieee802154_mgmt.c
+++ b/subsys/net/l2/ieee802154/ieee802154_mgmt.c
@@ -574,7 +574,30 @@ static int ieee802154_associate(uint64_t mgmt_request, struct net_if *iface,
 	 * TODO: The Association Response command shall be sent to the device
 	 *       requesting association using indirect transmission.
 	 */
-	k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+	ret = k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+
+	if (ret == -EAGAIN) {
+		/* Timeout, send DATA_REQUEST */
+		net_pkt_unref(pkt);
+		pkt = ieee802154_create_mac_cmd_frame(iface, IEEE802154_CFI_DATA_REQUEST, &params);
+		if (pkt == NULL) {
+			ret = -ENOBUFS;
+			k_sem_give(&ctx->scan_ctx_lock);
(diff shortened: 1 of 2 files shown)


subsys/net/l2/ieee802154/ieee802154_mgmt.c
```diff
@@ -574,7 +574,31 @@ static int ieee802154_associate(uint64_t mgmt_request, struct net_if *iface,
...
 	 */
-	k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+	ret = k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+
+	if (ret == -EAGAIN) {
+		/* Timeout, send DATA_REQUEST */
+		net_pkt_unref(pkt);
+		pkt = ieee802154_create_mac_cmd_frame(
+				iface, IEEE802154_CFI_DATA_REQUEST, &params);
+		if (!pkt) {
```
jukkar: Please change this to `if (pkt == NULL) {` as dictated by MISRA so that we do not introduce new code that is using boolean checks for non boolean variables.

subsys/net/l2/ieee802154/ieee802154_mgmt.c
```diff
@@ -574,7 +574,30 @@ static int ieee802154_associate(uint64_t mgmt_request, struct net_if *iface,
...
 	 *       requesting association using indirect transmission.
 	 */
-	k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+	ret = k_sem_take(&ctx->scan_ctx_lock, K_USEC(ieee802154_get_response_wait_time_us(iface)));
+
+	if (ret == -EAGAIN) {
+		/* Timeout, send DATA_REQUEST */
+		net_pkt_unref(pkt);
+		pkt = ieee802154_create_mac_cmd_frame(iface, IEEE802154_CFI_DATA_REQUEST, &params);
+		if (pkt != NULL) {
```
rlubos: That should be `pkt == NULL`, no?

rlubos (approved):

jukkar (approved):

PR #99702, opened 2025-11-19
net: l2: ieee802154: add support for MLME coordinator address

spiriou:
There are two ways in the current implementation to join a network, either with an association request or a via MLME-SET(macShortAddress) to rejoin a previously associated network.

The coordinator address needs to be valid in the rejoin case. It can be lost, for example if the device power cycles. This commit fixes that gap with new MLME GET/SET for:
- macCoordShortAddress
- macCoordExtendedAddress


subsys/net/l2/ieee802154/ieee802154_mgmt.c
```diff
@@ -747,7 +747,8 @@ static int ieee802154_set_parameters(uint64_t mgmt_request,
 		return -EINVAL;
 	}
 
-	if (mgmt_request == NET_REQUEST_IEEE802154_SET_EXT_ADDR) {
+	if ((mgmt_request == NET_REQUEST_IEEE802154_SET_EXT_ADDR) ||
+		(mgmt_request == NET_REQUEST_IEEE802154_SET_COORD_EXT_ADDR)) {
```
rlubos: Please fix the alignment
```suggestion
	if ((mgmt_request == NET_REQUEST_IEEE802154_SET_EXT_ADDR) ||
	    (mgmt_request == NET_REQUEST_IEEE802154_SET_COORD_EXT_ADDR)) {
```

subsys/net/l2/ieee802154/ieee802154_mgmt.c
```diff
@@ -865,7 +876,8 @@ static int ieee802154_get_parameters(uint64_t mgmt_request,
 		return -EINVAL;
 	}
 
-	if (mgmt_request == NET_REQUEST_IEEE802154_GET_EXT_ADDR) {
+	if (mgmt_request == NET_REQUEST_IEEE802154_GET_EXT_ADDR ||
+		mgmt_request == NET_REQUEST_IEEE802154_GET_COORD_EXT_ADDR) {
```
rlubos: Please fix the alignment
```suggestion
	if (mgmt_request == NET_REQUEST_IEEE802154_GET_EXT_ADDR ||
	    mgmt_request == NET_REQUEST_IEEE802154_GET_COORD_EXT_ADDR) {
```

rlubos (approved):

jukkar (approved):

masz-nordic (approved):

PR #99708, opened 2025-11-19
xtensa: move SoC specific out of `arch/` and into `soc/`

dcpleung:
This is mostly about moving SoC specific features and stuff out of the architecture layer and into the SoC layer... mixed in some bug fixes.

See individual commits for more details.


tests/lib/newlib/heap_listener/testcase.yaml
```diff
@@ -1,3 +1,18 @@
+common:
```
keith-packard, line 1: If newlib for this platform is so bad, shouldn't we keep it from being used by applications?

dcpleung: It is possible that newer toolchains from Cadence have already sorted out the newlib issues, so I don't want to simply disable usage of newlib. It would need to be a choice by the application developers. I don't have access to newer toolchains at this point so I cannot test.

Besides, the default is picolibc. So if the developers really need to use newlib, they will have to explicitly enable it and test.

doc/releases/migration-guide-4.4.rst
```diff
@@ -162,3 +167,6 @@ Trusted Firmware-M
 
 Architectures
 *************
+
+* ``CONFIG_ARCH_HAS_COHERENCE`` to :kconfig:option:`CONFIG_CACHE_CAN_SAY_MEM_COHERENCE` as
+  the feature is cache related so move it under cache.
```
aescolar: Minor: It would be nice to phrase it either as what action a user needs to do, or what has changed in a way it is clearly implied what to change.
```suggestion
* ``CONFIG_ARCH_HAS_COHERENCE`` was renamed  :kconfig:option:`CONFIG_CACHE_CAN_SAY_MEM_COHERENCE`.
```

dcpleung: Updated.

nashif (dismissed):

peter-mitsis (dismissed):

peter-mitsis (dismissed):

nashif (dismissed):

nashif (approved):

peter-mitsis (approved):

PR #99721, opened 2025-11-20
soc: st: stm32: add '-align' flag for signing tool v2.21.0+

### LINKED ISSUE #99456: STM32N6 images signed with STM32CubeProgrammer v2.21.0 fail to run
### Describe the bug

When I follow the instructions at https://docs.zephyrproject.org/latest/boards/st/stm32n6570_dk/doc/index.html images signed using the latest `STM32_SigningTool_CLI` version `2.21.0` do not run on either of the STM32N6570-DK or the Nucleo N657X0-Q boards.

Reverting back to version `2.20.0` of the STM32CubeProgrammer release fixes the issue (images run correctly per the instructions in the Zephyr docs).

I'm not exactly sure yet if this is a bug in the latest `STM32_SigningTool_CLI` tool or an issue on the Zephyr side, but I wanted to report it in case this saves somebody else a couple hours of head scratching wondering what's going on.

Also, be aware that the `2.20.0` release has an issue that prevents `west flash` from working on Windows & Ubuntu for the Nucleo N657X0-Q board, so if you revert to that version just be aware that flashing the board might be broken (strangely the STM32N6570-DK board works fine with `2.20.0` on these OS). It looks like [this flashing issue is fixed](https://community.st.com/t5/stm32cubeprogrammer-mcus/nucleo-n657x0-q-unable-to-erase-flash/m-p/854898/highlight/true#M9265) in `2.21.0`.

### Regression

- [ ] This is a regression.

### Steps to reproduce

1. Install STM32CubeProgrammer `2.21.0`
2. Follow the instructions at https://docs.zephyrproject.org/latest/boards/st/stm32n6570_dk/doc/index.html

### Impact

Workaround by reverting to `2.20.0`

### Environment

Same issue on Windows, macOS, Ubuntu hosts

cdwilson:
Starting in `v2.21.0`, the STM32 signing tool (`STM32_SigningTool_CLI`) stopped automatically adding padding bytes at the beginning of the payload to align it to the `0x400` offset. To restore this behavior, the `-align` flag must be passed to the signing tool post-build command.
    
This PR checks for signing tool version `v2.21.0` or higher and appends the `-align` flag to the post-build signing command. It also changes local CMake variable names to lower case and corrects some indentation issues.
    
Fixes zephyrproject-rtos/zephyr#99456

### DIFF
--- a/soc/st/stm32/stm32n6x/CMakeLists.txt
+++ b/soc/st/stm32/stm32n6x/CMakeLists.txt
@@ -3,11 +3,11 @@
 zephyr_include_directories(${ZEPHYR_BASE}/drivers)
 zephyr_sources(
   soc.c
-  )
+)
 
 zephyr_sources_ifdef(CONFIG_STM32N6_NPU
   npu/npu_stm32n6.c
-  )
+)
 
 zephyr_include_directories(.)
 
@@ -22,26 +22,47 @@ zephyr_linker_sources_ifdef(CONFIG_BOOTLOADER_MCUBOOT SECTIONS ram_check.ld)
 
 if(NOT CONFIG_BOOTLOADER_MCUBOOT)
 
-  if(${CMAKE_HOST_SYSTEM_NAME} STREQUAL Windows)
-    set(SIGNING_TOOL STM32_SigningTool_CLI.exe)
-  else()
-    set(SIGNING_TOOL STM32_SigningTool_CLI)
-  endif()
+  set(signing_tool STM32_SigningTool_CLI)
 
-  find_file(SIGNING_TOOL_FIND ${SIGNING_TOOL})
-  if(SIGNING_TOOL_FIND STREQUAL SIGNING_TOOL_FIND-NOTFOUND)
+  # find_program will automatically search for .exe extension on Windows
+  find_program(signing_tool_find ${signing_tool})
+  if(signing_tool_find STREQUAL signing_tool_find-NOTFOUND)
     message(WARNING "
-    Signing Image tool (${SIGNING_TOOL}) is not available.
-    Signed image will not be generated.
-    You won't be able to run application on the board.
-    Refer to board documentation for more information")
+      Signing Image tool (${signing_tool}) is not available.
+      Signed image will not be generated.
+      You won't be able to run application on the board.
+      Refer to board documentation for more information.
+    ")
   else()
-    set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
-      COMMAND ${SIGNING_TOOL}
+    message(STATUS "Found STM32 signing tool: ${signing_tool_find}")
+
+    execute_process(
+      COMMAND ${signing_tool_find} --version
+      OUTPUT_VARIABLE version
+      RESULT_VARIABLE result
+    )
+
+    set(signing_tool_args
       -in ${PROJECT_BINARY_DIR}/${CONFIG_KERNEL_BIN_NAME}.bin
       -nk -t fsbl -hv 2.3 --silent
       -o ${PROJECT_BINARY_DIR}/${CONFIG_KERNEL_BIN_NAME}.signed.bin
       -dump ${PROJECT_BINARY_DIR}/${CONFIG_KERNEL_BIN_NAME}.signed.bin
+    )
+
+    set(signing_tool_version_regex "v([0-9]+[.][0-9]+[.][0-9]+)")
+    if(result EQUAL 0 AND version MATCHES ${signing_tool_version_regex})
+      if(${CMAKE_MATCH_1} VERSION_GREATER_EQUAL 2.21.0)
+        list(APPEND signing_tool_args -align)
+      endif()
+    else()
+      message(FATAL_ERROR "
+        Unable to determine ${signing_tool} version (expected match to
+        ${signing_tool_version_regex} regex)
+      ")
+    endif()
+
+    set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
+      COMMAND ${signing_tool} ${signing_tool_args}
       WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
     )
 


cdwilson:
Set this to draft because I realized this prob needs some updates to the docs. If this looks like the right direction, LMK and I'll update this PR with the doc changes as well.

erwango:
So, you can look at [nrf-regtool](https://github.com/zephyrproject-rtos/zephyr/blob/f1c2ce4f4026bd5b212c6e4cc7d754aba69b189e/modules/hal_nordic/nrf-regtool/nrf-regtoolConfigVersion.cmake#L6) which performs a version check once, then store it for later usage.
We should probably do something similar. Are you interested to look at this ?

etienne-lms (commented):
Worth to add a note in the migration guide to tell flash programming N6 boards require at least STM32_SigningTool_CLI v2.21.0 with a reference to #99456?

soc/st/stm32/stm32n6x/CMakeLists.txt
```diff
@@ -28,20 +28,52 @@ if(NOT CONFIG_BOOTLOADER_MCUBOOT)
...
+        "appending \"-align\" flag to post-build signing command "
+        "(applicable only with header v2.3 for MCUs, starting from "
+        "version v${SIGNING_TOOL_MIN_VERSION})"
+        )
+      endif()
+    else()
+      message(FATAL_ERROR
+      "Unable to determine ${SIGNING_TOOL} version (expected match to "
+      "${SIGNING_TOOL_VERSION_REGEX} regex)"
+      )
```
cdwilson: Not sure if we want to bomb out here or just throw a warning. I figured it's better to have an explicit error than to arbitrarily choose either appending or omitting the `-align` flag in the case where we can't determine the version.

soc/st/stm32/stm32n6x/CMakeLists.txt
```diff
@@ -28,20 +28,53 @@ if(NOT CONFIG_BOOTLOADER_MCUBOOT)
...
-      COMMAND ${SIGNING_TOOL}
+    message(STATUS "Found STM32 signing tool: ${SIGNING_TOOL_FIND}")
+
+    execute_process(
+      COMMAND ${SIGNING_TOOL_FIND} --version
+      OUTPUT_VARIABLE version
+      RESULT_VARIABLE result
+    )
+
+    set(SIGNING_TOOL_ARGS
```
etienne-lms: Minor comment: CMake conventions recommend to use lower case names for varaibles local to a CMake file, and upper case names for variables exported to other CMake file.
That said, before your changes, this **CMakeLists.txt** file already defined  `SIGNING_TOOL` and `SIGNING_TOOL_ARGS` that are local. Maybe worth to change them also, either as a follow-up P-R, or in your P-R if you will to.

soc/st/stm32/stm32n6x/CMakeLists.txt
```diff
@@ -28,20 +28,53 @@ if(NOT CONFIG_BOOTLOADER_MCUBOOT)
...
+      set(SIGNING_TOOL_MIN_VERSION "2.21.0")
+
+      if(SIGNING_TOOL_VERSION VERSION_GREATER_EQUAL SIGNING_TOOL_MIN_VERSION)
+        list(APPEND SIGNING_TOOL_ARGS -align)
+
+        message(STATUS
+        "${SIGNING_TOOL} version is v${SIGNING_TOOL_VERSION}, "
+        "appending \"-align\" flag to post-build signing command "
+        "(applicable only with header v2.3 for MCUs, starting from "
+        "version v${SIGNING_TOOL_MIN_VERSION})"
```
etienne-lms: Could you indent with 2 more space chars, here at below (in the fatal message).

zacck (commented):
So I just tested this on both 2.20 and 2.21 and it works really smooth

soc/st/stm32/stm32n6x/CMakeLists.txt
cdwilson: @etienne-lms thanks for your feedback! I updated the PR to use lower case names for all the local variables (LMK if there is any I missed).

soc/st/stm32/stm32n6x/CMakeLists.txt
```diff
@@ -22,26 +22,57 @@ zephyr_linker_sources_ifdef(CONFIG_BOOTLOADER_MCUBOOT SECTIONS ram_check.ld)
...
+      set(signing_tool_version_min 2.21.0)
+
+      if(signing_tool_version VERSION_GREATER_EQUAL signing_tool_version_min)
+        list(APPEND signing_tool_args -align)
+
+        message(STATUS
+          "${signing_tool} version is v${signing_tool_version}, "
+          "appending \"-align\" flag to post-build signing command "
+          "(applicable only with header v2.3 for MCUs, starting from "
+          "version v${signing_tool_version_min})"
```
erwango: To be honnest, I'm unclear if this add clarity.

erwango (dismissed):

soc/st/stm32/stm32n6x/CMakeLists.txt
etienne-lms: I think too the message can be omitted. An inline comment should be enough.

cdwilson: Removed these lines

erwango:
> Adding DNM tag as I'm not able to to run ext flash loaded application on N6-DK using v2.21.0.

Was a local mess up and I was actually missing current fix.

erwango (dismissed):

soc/st/stm32/stm32n6x/CMakeLists.txt
```diff
@@ -22,26 +22,55 @@ zephyr_linker_sources_ifdef(CONFIG_BOOTLOADER_MCUBOOT SECTIONS ram_check.ld)
...
+      set(signing_tool_version ${CMAKE_MATCH_1})
+      set(signing_tool_version_min 2.21.0)
+
+      if(signing_tool_version VERSION_GREATER_EQUAL signing_tool_version_min)
+        list(APPEND signing_tool_args -align)
+
+        message(STATUS
+          "${signing_tool} version is v${signing_tool_version}, "
+          "appending \"-align\" flag to post-build signing command"
+        )
```
etienne-lms: I don't think the message is really useful. Not a blocking issue.

etienne-lms (commented):
LGTM.
Could you mention in the commit message that you change local CMake variable names to lower case names and correct some indentation issues. For consistency of commit messages to describe changes.

soc/st/stm32/stm32n6x/CMakeLists.txt
cdwilson: Removed

etienne-lms (approved):

erwango (approved):
Thanks for this patch!

PR #99727, opened 2025-11-20
tests: zbus: publish_stats: update thresholds

JordanYates:
`zbus_chan_pub_stats_msg_age` returns milliseconds, while `clock_window` is in clock ticks. Update the thresholds to refer to the same 50ms window as `clock_window`.

Fixes #99500.


rodrigopex (changes requested):
There is repetitive usage of a constant. Could you move that to a define or static constant with a meaningful name? That would reduce maintenance effort.

rodrigopex (approved):

pdgendt (approved):

PR #99729, opened 2025-11-20
stm32: xspi: Allow flash driver usage in RAM_LOAD mode

erwango:
Currently, STM32 XSPI flash driver initialization aborts if application is run from external flash.
Though, constraints are actually different if:
- application runs in XIP and is read from ext flash
- application runs in RAM LOAD mode and is read from RAM, leaving the ext flash available for flash driver operations

Fix XSPI driver to allow flash operations in RAM LOAD.
Make necessary changes on N6 DK to be able to make use of it.

Once approved, it will have to be extended to O/QSPI drivers.
Then, we should also make sure this work using MSPI drivers.

Edit: Depends on https://github.com/mcu-tools/mcuboot/pull/2547

### DIFF
--- a/boards/st/stm32n6570_dk/stm32n6570_dk_common.dtsi
+++ b/boards/st/stm32n6570_dk/stm32n6570_dk_common.dtsi
@@ -23,7 +23,7 @@
 		zephyr,touch = &gt911;
 		spi-flash0 = &mx66uw1g45g;
 		zephyr,flash-controller = &mx66uw1g45g;
-		zephyr,flash = &mx66uw1g45g;
+		zephyr,flash = &ext_flash;
 		zephyr,code-partition = &slot0_partition;
 	};
 
@@ -400,32 +400,40 @@ zephyr_udc0: &usbotg_hs1 {
 		four-byte-opcodes;
 		status = "okay";
 
-		partitions {
-			compatible = "fixed-partitions";
-			#address-cells = <1>;
-			#size-cells = <1>;
+		#address-cells = <1>;
+		#size-cells = <1>;
+		ranges = <0x0 0x70000000 DT_SIZE_M(128)>;
 
-			/*
-			 * Following flash partition is dedicated to the use of bootloader
-			 */
-			boot_partition: partition@0 {
-				label = "mcuboot";
-				reg = <0x00000000 DT_SIZE_K(64)>;
-			};
+		ext_flash: ext-flash@0 {
+			compatible = "soc-nv-flash";
+			reg = <0x0 DT_SIZE_M(128)>;
+			write-block-size = <1>;
+			erase-block-size = <DT_SIZE_K(4)>;
 
-			slot0_partition: partition@10000 {
-				label = "image-0";
-				reg = <0x10000 DT_SIZE_K(1536)>;
-			};
+			partitions {
+				compatible = "fixed-partitions";
+				#address-cells = <1>;
+				#size-cells = <1>;
 
-			slot1_partition: partition@210000 {
-				label = "image-1";
-				reg = <0x210000 DT_SIZE_K(1536)>;
-			};
+				boot_partition: partition@0 {
+					label = "mcuboot";
+					reg = <0x00000000 DT_SIZE_K(64)>;
+				};
 
-			storage_partition: partition@410000 {
-				label = "storage";
-				reg = <0x410000 DT_SIZE_K(64)>;
+				slot0_partition: partition@10000 {
+					label = "image-0";
+					reg = <0x10000 DT_SIZE_K(1536)>;
+				};
+
+				slot1_partition: partition@210000 {
+					label = "image-1";
+					reg = <0x210000 DT_SIZE_K(1536)>;
+				};
+
+				storage_partition: partition@410000 {
+					label = "storage";
+					reg = <0x410000 DT_SIZE_K(64)>;
+				};
 			};
 		};
 	};
--- a/drivers/flash/flash_stm32_xspi.c
+++ b/drivers/flash/flash_stm32_xspi.c
@@ -997,6 +997,7 @@ static int stm32_xspi_set_memorymap(const struct device *dev)
 	LOG_DBG("MemoryMap mode enabled");
 	return 0;
 }
+#endif /* CONFIG_STM32_MEMMAP */
 
 static int stm32_xspi_abort(const struct device *dev)
 {
@@ -1009,10 +1010,7 @@ static int stm32_xspi_abort(const struct device *dev)
 
 	return 0;
 }
-#endif /* CONFIG_STM32_MEMMAP */
 
-
-#if defined(CONFIG_STM32_MEMMAP) || defined(CONFIG_STM32_APP_IN_EXT_FLASH)
 /* Function to return true if the octoflash is in MemoryMapped else false */
 static bool stm32_xspi_is_memorymap(const struct device *dev)
 {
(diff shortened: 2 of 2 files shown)


drivers/flash/flash_stm32_xspi.c
```diff
@@ -1218,7 +1215,7 @@ static int flash_stm32_xspi_read(const struct device *dev, off_t addr,
 		return 0;
 	}
 
-#if defined(CONFIG_STM32_MEMMAP) || defined(CONFIG_STM32_APP_IN_EXT_FLASH)
+#if defined(CONFIG_STM32_MEMMAP) || defined(CONFIG_STM32_APP_IN_EXT_FLASH) && defined(CONFIG_XIP)
```
etienne-lms: Could you add inner parentheses to prevent from confusion?
```suggestion
#if defined(CONFIG_STM32_MEMMAP) || (defined(CONFIG_STM32_APP_IN_EXT_FLASH) && defined(CONFIG_XIP))
```
Could you update the `#endif` counter part?

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2093,7 +2090,7 @@ static int flash_stm32_xspi_init(const struct device *dev)
 		return -ENODEV;
 	}
 
-#ifdef CONFIG_STM32_APP_IN_EXT_FLASH
+#if defined(CONFIG_STM32_APP_IN_EXT_FLASH) && defined(CONFIG_XIP)
```
etienne-lms, line 2093: Could you update the `#endif` counter part?

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2195,13 +2192,29 @@ static int flash_stm32_xspi_init(const struct device *dev)
...
 
 	LOG_DBG("XSPI Init'd");
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory map could have been set by previous application */
+		/* Abort to allow following card oriented transactions */
+		/* If needed by the application, it will be re-enabled just after */
+		ret = stm32_xspi_abort(dev);
+		if (ret != 0) {
+			LOG_ERR("Failed to abort memory-mapped access before init");
```
etienne-lms: Is it safe to continue the function execution? It seems to me this should return with an error code.

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2195,13 +2192,30 @@ static int flash_stm32_xspi_init(const struct device *dev)
...
 		LOG_ERR("XSPI Init failed");
 		return -EIO;
 	}
 
 	LOG_DBG("XSPI Init'd");
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory map could have been set by previous application */
+		/* Abort to allow following card oriented transactions */
+		/* If needed by the application, it will be re-enabled just after */
```
etienne-lms: Nitpicking suggestion as this a single inline comment.
```suggestion
		/* Memory map could have been set by previous application.
		 * If so, abort on-going operations to allow following card oriented transactions.
		 * If memory map mode is needed by the application, it will be re-enabled afterward.
		 */

File: drivers/flash/flash_stm32_xspi.c [hunk 5]
```diff
@@ -2195,13 +2192,30 @@ static int flash_stm32_xspi_init(const struct device *dev)
 		dev_data->hxspi.Init.DelayHoldQuarterCycle = HAL_XSPI_DHQC_ENABLE;
 	}
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory map could have been set by previous application */
```
JarmouniA:
```suggestion
		/* Flash memory-mapping could have been set by previous application */
```

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2195,13 +2192,30 @@ static int flash_stm32_xspi_init(const struct device *dev)
...
+
 	if (HAL_XSPI_Init(&dev_data->hxspi) != HAL_OK) {
 		LOG_ERR("XSPI Init failed");
 		return -EIO;
 	}
 
 	LOG_DBG("XSPI Init'd");
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory map could have been set by previous application */
```
JarmouniA:
```suggestion
		/* Memory-mapping could have been set by previous application */
```

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2195,13 +2192,30 @@ static int flash_stm32_xspi_init(const struct device *dev)
...
+		/* Abort to allow following card oriented transactions */
```
JarmouniA: card!?

boards/st/stm32n6570_dk/stm32n6570_dk_common.dtsi
```diff
@@ -23,7 +23,7 @@
 		zephyr,touch = &gt911;
 		spi-flash0 = &mx66uw1g45g;
 		zephyr,flash-controller = &mx66uw1g45g;
-		zephyr,flash = &mx66uw1g45g;
+		zephyr,flash = &ext_flash;
```
JarmouniA, line 26: More of a nit really, but `ospi-nor-flash@0` node should not have the label `mx66uw1g45g`, rather something like `ext_flash_controller`, and consequently `zephyr,flash-controller` should be changed.

drivers/flash/flash_stm32_xspi.c
erwango: I was about to say copy/paste, but actually I don't even know where it is coming from.
Fixed and actually moved later, closer to Jedec related stuff.

drivers/flash/flash_stm32_xspi.c
erwango: Reworded, but in a different way.

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2195,6 +2192,12 @@ static int flash_stm32_xspi_init(const struct device *dev)
 		dev_data->hxspi.Init.DelayHoldQuarterCycle = HAL_XSPI_DHQC_ENABLE;
 	}
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory-mapping could have been set by previous application */
+		/* Force HAL instance in correct state */
```
etienne-lms: /* Memory-mapping could have been set by previous application.
		 * Force HAL instance in correct state.
		 */

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2268,6 +2271,17 @@ static int flash_stm32_xspi_init(const struct device *dev)
 	/* Run IRQ init */
 	dev_cfg->irq_config(dev);
 
+	if (stm32_xspi_is_memorymap(dev)) {
+		/* Memory-mapping could have been set by previous application */
+		/* Abort to allow following Jedec transactions, it will be */
+		/* re-enabled afterwards if needed by the application. */
```
etienne-lms: /* Memory-mapping could have been set by previous application.
		 * Abort to allow following Jedec transactions, it will be
		 * re-enabled afterwards if needed by the application.
		 */

boards/st/stm32n6570_dk/stm32n6570_dk_common.dtsi
erwango: It seems my comment was lost:
I propose to fix this globally in one shot (at least on stm32 based boards). Would you mind raising an issue, so i doesn't get lost ?

JarmouniA (dismissed):

etienne-lms (dismissed):

JarmouniA (approved):

FRASTM (approved):

JarmouniA (approved):

mathieuchopstm (approved):

etienne-lms (approved):

PR #99742, opened 2025-11-20
Bluetooth: Host: Add legacy pairing test config

HaavardRei:
Adds the `CONFIG_BT_SMP_LEGACY_PAIR_ONLY` Kconfig option to force devices to use legacy pairing. This has a dependency on `CONFIG_BT_TESTING` as it is only intended for testing purposes, and use of legacy pairing is discouraged.


jhedberg (dismissed):

subsys/bluetooth/host/Kconfig
```diff
@@ -1163,6 +1163,15 @@ config BT_CONN_DISABLE_SECURITY
 	  WARNING: This option enables anyone to snoop on-air traffic.
 	  Use of this feature in production is strongly discouraged.
 
+config BT_SMP_ENFORCE_LEGACY_PAIRING
```
Thalley: Should it be


```suggestion
config BT_SMP_LEGACY_PAIR_ONLY
```

to following the naming of 
```
config BT_SMP_OOB_LEGACY_PAIR_ONLY
config BT_SMP_SC_ONLY
config BT_SMP_SC_PAIR_ONLY
```

subsys/bluetooth/host/Kconfig
```diff
@@ -1163,6 +1163,15 @@ config BT_CONN_DISABLE_SECURITY
...
+	bool "Enforce legacy pairing"
+	depends on BT_TESTING
```
Thalley, line 1168: Is it not allowed to only support legacy pairing in recent version of the core spec?

HaavardRei: It's allowed, but since Zephyr 3.0 the use of legacy pairing has been discouraged for being less/not secure, so I didn't want this Kconfig to imply otherwise. Zephyr-devices support legacy pairing with devices that do not support secure connections, but since Zephyr-devices do support it they will always default to this (and thus this config is needed to force it).

subsys/bluetooth/host/Kconfig
HaavardRei: Good point, added!

subsys/bluetooth/host/Kconfig
Thalley: I'm wondering if we should allow setting this without `BT_TESTING`, but treat it as a warning for the above reasons. That would be similar to how we treat e.g. `BT_USE_DEBUG_KEYS`, where we have both Kconfig and CMake warnings if it's enabled. 

Alternatively we should consider adding the same `depends on BT_TESTING` for things like `BT_USE_DEBUG_KEYS`.

HaavardRei: Both suggestions are fine by me, do you have any thoughts @jhedberg?

jhedberg: I think `depends on BT_TESTING` is the right way. `BT_USE_DEBUG_KEYS` should have it as well.

Thalley: @jhedberg Should we apply `depends on BT_TESTING` for all the other similar configs? And then if we do, should we remove the warning for them as well, and add/keep only a warning for `BT_TESTING`?

Thalley (approved):

subsys/bluetooth/host/Kconfig
HaavardRei: I think it makes sense to add the dependency on BT_TESTING to all relevant configs, keeping the warnings in the Kconfig file (but remove the CMakeList warnings and add a common one for BT_TESTING?).

alwa-nordic (approved):

PR #97364, opened 2025-10-10
STM32: add ethernet support to stm32h7rs

TD-JBL:
This PR adds ethernet support for STM32H7RS series microcontrollers.
It's almost a 1:1 copy of h7x ethernet configuration. Only register addresses, clock sources and interrupt numbers changed in device tree and the used RAM for DMA buffer. 

## Changes:
* Add ethernet node with mac and mdio subnodes to dts similar to h7-family.
* Add `SOC_SERIES_STM32H7RSX` where appicable next to `SOC_SERIES_STM32H7X`.
* Add ethernet DMA buffer/descriptor region (sram2) and read only flash region (0x08FFF800) with unique device ID registers to MPU region list.
* Ethernet DMA buffer/descriptor memory section is added to soc linker script.
* Enable support for ethernet on nucleo_h7s3l8 board.

## Used DMA memory
As all h7rs variants have the same memory configuration, there is no need to support multiple memories as DMA buffer. `sram1` and `sram2` (AHB SRAM1 and AHB SRAM2) are identical, so I hardcoded the DMA buffer into `sram2`. 

At first I tried not defining any special memory and let the driver use the NOCACHE memory, but after reading #29915 and #30403 I decided to keep it as near as possible to the h7x implementation.

Edit: Is it possible or useful to add a Kconfig option or dts property to select the used sram node?

## 0x08FFF800 read only flash region with unique device ID registers
The controller memfaults without this region in the MPU list while trying to read the uniqe device ID registers. The PR #94019 introduced a different way of preventing speculative access in entire memory space than the one used for h7x:
`MPU_REGION_ENTRY("UNMAPPED", 0, {REGION_4G | MPU_RASR_XN_Msk | P_NA_U_NA_Msk}), `
Every accessed memory must get a MPU region after blocking access to the complete memory space with this entry.

The unique device ID is used to create a random mac address by the ethernet driver in some conditions.

## Testing Status:
* It responds to ping ;)
* Tested on NUCLEO-H7S3L8 with custom bootloader/app. As I'm not familiar with mcuboot, I didn't run tests with samples like echo_server, they are too big to fit into internal flash and need XiP.

Maybe someone more familiar with running mcuboot+samples on this board  can run some tests?

PR #99745, opened 2025-11-20
soc: st: stm32: h7rsxx: Fix MPU region for read-only id flash region

TD-JBL:
Set correct MPU memory type and size for the uniqe device id, package code and ADC VREF/TS calibration read-only flash region.

~~(DEVICE_NON_SHAREABLE | REGION_512B | P_RO_U_NA_Msk)) is the same as REGION_IO_ATTR (used for PERIPH region) but RO instead of RW.~~
_Edit:_ REGION_IO_ATTR configures the MPU region to device-memory with RW access, which is also used for the PERIPH region. To avoid complicating things unnecessarily, we use this type instead of defining a new device-RO. 

The size is reduced to 512 bytes, because RM0477 table 28 and chapter 5.3.12 state that this read-only flash area has a size of 512 bytes.

I introduced this region with PR #97364, both configurations work, but it is more correct with this one.

Testing: Network still initializes with this changes (reads uniqe device ID for creating the mac address per default). 
Cross-check: Will mem-fault during zephyr init when commenting out this `MPU_REGION_ENTRY`.

Open question: Create a `#define` for this type? `REGION_IO_ATTR_RO` for example

### DIFF
--- a/soc/st/stm32/stm32h7rsx/mpu_regions.c
+++ b/soc/st/stm32/stm32h7rsx/mpu_regions.c
@@ -27,8 +27,8 @@ static const struct arm_mpu_region mpu_regions[] = {
 	/* Region 3 */
 	MPU_REGION_ENTRY("SRAM_0", CONFIG_SRAM_BASE_ADDRESS, REGION_RAM_ATTR(REGION_SRAM_SIZE)),
 
-	/* Region 4 - Ready only flash with unique device id */
-	MPU_REGION_ENTRY("ID", 0x08FFF800, REGION_FLASH_ATTR(REGION_2K)),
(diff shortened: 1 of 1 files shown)


djiatsaf-st (dismissed):
LGTM

mathieuchopstm:
> Open question: Create a `#define` for this type? `REGION_IO_ATTR_RO` for example

I'm not sure if this has a lot of value considering that

> `(DEVICE_NON_SHAREABLE | REGION_512B | P_RO_U_NA_Msk)` is the same as `REGION_IO_ATTR` but RO instead of RW

(i.e., do we really care about the region being read-only?)

mathieuchopstm (approved):

erwango (approved):

PR #99749, opened 2025-11-20
drivers: usb: dwc2: Do cache operations on ownership change

tmon-nordic:
Perform cache operations in thread context when the buffer ownership changes between USB stack and UDC driver. This offers clearly measurable throughput increase on Mass Storage when double buffering is enabled. When endpoint operations are not double buffered the difference is negligible. The positive side effect is reducing number of operations performed in interrupt context.

### DIFF
--- a/drivers/usb/udc/udc_dwc2.c
+++ b/drivers/usb/udc/udc_dwc2.c
@@ -422,6 +422,11 @@ static int dwc2_ctrl_feed_dout(const struct device *dev, const size_t length)
 		return -ENOMEM;
 	}
 
+	if (dwc2_in_buffer_dma_mode(dev)) {
+		/* Get rid of all dirty cache lines */
+		sys_cache_data_invd_range(buf->data, net_buf_tailroom(buf));
+	}
+
 	udc_buf_put(ep_cfg, buf);
 	atomic_set_bit(&priv->xfer_new, 16);
 	k_event_post(&priv->drv_evt, BIT(DWC2_DRV_EVT_XFER));
@@ -592,8 +597,6 @@ static int dwc2_tx_fifo_write(const struct device *dev,
 
 		sys_write32((uint32_t)buf->data,
 			    (mem_addr_t)&base->in_ep[ep_idx].diepdma);
-
-		sys_cache_data_flush_range(buf->data, len);
 	}
 
 	diepctl = sys_read32(diepctl_reg);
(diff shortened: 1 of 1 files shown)


drivers/usb/udc/udc_dwc2.c
josuah, line 2913: Does this one not need to be moved to the main thread as well? Or is it unrelated (control endpoint).

josuah (commented):
Making sure I understood before giving review approval:

There is a first cache invalidation for both IN/OUT when buffers are enqueued.

- For IN buffers it just flushes the data to send
- For OUT buffers coming, invalidation makes sure there is no cache that risks to overwrite the received data
- Then there need another cache ~~flush~~ **invalidation** for the data received.

drivers/usb/udc/udc_dwc2.c
tmon-nordic: It must not be moved to main thread because we have to keep the latest received SETUP in `priv->setup` - there can be multiple setup packets received back-to-back which makes handling complicated (due to how the Buffer DMA works). Control transfers are not really a hot path (unless there is some strange protocol involved) so I think it doesn't make sense to change this.

tmon-nordic:
> * For IN buffers it just flushes the data to send

Yes, the cached data must be in memory so DWC2 controller will read it.

> * For OUT buffers coming, invalidation makes sure there is no cache that risks to overwrite the received data

Yes, this is to ensure there are not any dirty lines that the cache controller could potentially flush (evict) while the DWC2 is really the only of the buffer. If such eviction happens after DWC2 writes data to the affected location, then the data would essentially be corrupted.

> * Then there need another cache flush for the data received.

Note that buffer cache is invalidated (discarded), not flushed (committed/writeback). To make sure that the data in cache is what DWC2 wrote. The cache lines for sure are not dirty at this point (unless there is a software bug somewhere that writes to the buffer after giving the ownership to UDC driver), but if for whatever reason the data was pre-fetched there is no way to tell if it was before or after DWC2 DMA wrote payload to memory.


This reasoning is the same as earlier. The only difference this PR does is when the cache operations take place.

josuah (dismissed):
Thank you for explaining, it is more clear for me now.

drivers/usb/udc/udc_dwc2.c
tmon-nordic, line 430: Just allocated buffer may have dirty cache lines - the buffer should be invalidated here.

jfischer-no (approved):

carlescufi (approved):

PR #89124, opened 2025-04-25
[DNM] RFC: add clock management drivers (duplicate of #72102)

danieldegrasse:
This PR is a continuation of #72102, since I no longer have push access to that branch. This RFC implements the clock management framework. For a more detailed overview, please see the introduction in #72102

PR #99750, opened 2025-11-20
doc: add policy re vendor specific hardware functionality

nashif:
Add docs defining how vendor-specific hardware functionality must be
contributed, structured, and exposed within the Zephyr Project.

Fixes #91373

Signed-off-by: Anas Nashif <anas.nashif@intel.com>

### DIFF
--- a/doc/develop/api/vendor_interfaces_policy.rst
+++ b/doc/develop/api/vendor_interfaces_policy.rst
@@ -1,0 +1,137 @@
+.. _vendor_specific_api_policy:
+
+Vendor-Specific API Placement Policy
+####################################
+
+Overview
+*********
+
+This policy defines how vendor-specific hardware functionality must be contributed,
+structured, and exposed within the Zephyr Project. It ensures that Zephyr’s public API
+surface remains stable, that vendor extensions do not bypass the standard API lifecycle,
+and that potential cross-vendor features have a clear path to evolve into
+generic driver classes.
+
+Vendor-specific APIs must not be exposed as public Zephyr interfaces unless they meet
+the criteria required for inclusion in Zephyr’s formal API lifecycle
+(experimental → stable → deprecated). Features that are inherently tied to a single
+SoC or vendor must remain encapsulated within the vendor’s SoC directories.
+
+This policy provides a consistent decision model for contributors and maintainers.
+
+Guiding Principles
+******************
+
+- **Public Zephyr APIs must not contain vendor-specific interfaces.**
+
+  Interfaces placed under ``include/zephyr/`` are expected to evolve through the standard
+  API lifecycle. Vendor-specific interfaces that cannot satisfy this evolution
+  must not be exposed as public Zephyr APIs.
+
+- **Prefer generic solutions over vendor-specific ones.**
+
+  When a hardware feature applies to multiple vendors or might in the future
+  it should be implemented as part of an existing driver class or as a new cross-vendor
+  driver class.
+
+- **Vendor-specific features must be encapsulated.**
+
+  Vendor-only mechanisms should live within the vendor’s platform namespace (SoCs, boards, etc.).
+
+- **Evolution must remain possible.**
+
+  Placement must not prevent later promotion of a vendor-specific implementation
+  into a portable Zephyr driver class.
+
+Decision Process
+****************
+
+When adding new hardware functionality or vendor-specific behavior,
+contributors must follow the decision tree below, in the given order:
+
+- **Does the feature logically extend an existing driver class?**
+
+  If yes, extend the existing class; do not create a vendor-specific API.
+
+- **Could the feature become a cross-vendor driver class?**
+
+  If the hardware concept is not clearly unique to a single vendor, or
+  resembles patterns likely to appear across multiple architectures,
+  create a new driver class.
+
+- **Is the functionality clearly vendor-specific and unlikely to generalize?**
+
+  In this case, place both headers and implementations under:
+
+  .. code-block:: none
+
+      soc/<vendor>/..
+
+
+  This ensures true encapsulation, long-term maintainability, and clean removal
+  when an SoC is dropped.
+
+  Vendor-specific code should be implemented in headers located under
+  ``soc/<vendor>/include`` and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.
+
+  Such interfaces would be private and/or internal to the vendor to support their
+  specific hardware and will be scrutinized in order to avoid creating feature islands specific
+  to vendors.
+
+  Applications using these interfaces, given that they are internal and/or private to the vendor,
+  shall expect no guarantees as to the lifecycle or backwards compatibility.
+
+Allowed Locations Summary
+*************************
+
+1. When extending existing driver classes
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>.h
+      drivers/<class>/
+
+  - Use when:
+
+    The functionality fits or extends an existing class.
+
+
+2. Vendor-specific extensions to existing driver classes
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>/
+      drivers/<class>/
+
+  - Use when:
+
+    The feature is a vendor-specific extension to an existing class, but the core
+    concept is not unique to a single vendor and could be standardized in the future.
+
+3. Create New Driver Class (Cross-Vendor)
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>.h
+      drivers/<class>/
+
+  - Use when:
+
+    The feature could reasonably apply to multiple vendors or evolve into a generic abstraction.
+
+4. Vendor-Specific, SoC-Local Code (Default for Vendor-Unique Features)
+
+  - Where:
+
+    Vendor-specific code should be implemented in headers located under ``soc/<vendor>/include``
+    and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.
+
+  - Use when:
+
+    The feature is unique to a vendor or SoC and is not intended as a public API.
--- a/doc/develop/api/index.rst
+++ b/doc/develop/api/index.rst
@@ -8,5 +8,6 @@ API Status and Guidelines
 
    overview.rst
    api_lifecycle.rst
+   vendor_interfaces_policy.rst
    design_guidelines.rst
    terminology.rst


teburd (dismissed):
This makes good sense to me and I think pushes people to hopefully work towards cross vendor APIs when possible, which should be ultimately the goal given this is an OS (an abstraction over hardware).

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+This policy provides a consistent decision model for contributors and maintainers.
+
+Guiding Principles
+******************
+
+- **Public Zephyr APIs must not contain vendor-specific interfaces.**
+
+  Interfaces placed under include/zephyr/ are expected to evolve through the standard
+  API lifecycle. Vendor-specific functionality that cannot satisfy this evolution
+  must remain internal.
```
henrikbrixandersen: We likely need to clarify what is meant by "internal" here.

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+  When a hardware feature could apply to multiple vendors or might in the future
+  it should be implemented as part of an existing driver class or as a new cross-vendor
+  driver class.
+
+- **Vendor-specific features must be encapsulated.**
+
+  Vendor-only mechanisms should live within the vendor’s SoC directory.
+  If multiple SoCs from the same vendor share internal logic, a vendor-specific
+  driver directory may be used.
```
henrikbrixandersen: Some vendor-specific APIs are for discrete peripherals (i.e. not part of a SoC, but an external peripheral). We need to take these into account as well.

henrikbrixandersen (changes requested):

doc/develop/api/vendor_specific_api_policy.rst
nashif: meaning "not public", open for suggestions

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+      drivers/<vendor>/
+
+  This directory is allowed only for internal vendor logic and must not expose public Zephyr APIs.
+
+- **Vendor-specific public APIs should be avoided.**
+
+  Only consider exposing a vendor-specific public API if it is expected to evolve
+  into a stable, cross-vendor Zephyr API and can follow the API lifecycle.
+
+Allowed Locations Summary
```
carlescufi, line 84: Perhaps differentiate between implementation (.c) and interface (.h)

mathieuchopstm: To be more explicit, differentiate between where files are located, and how they are included.

For example, the single file `soc/<vendor>/<soc>/header.h` could be included as:
* `#include <VENDOR/SOC/header.h>` if `zephyr_include_directories(soc/)`
* `#include <header.h>` if `zephyr_include_directories(soc/<vendor>/<soc>/)`

Ideally, vendor-specific headers should have an "include path" such that, just from the `#include` directive, it can be identified as vendor-specific.

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+    The logic is vendor-specific but shared across multiple SoCs.
+
+5. Vendor-Specific Public API (Discouraged)
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<vendor>/ (exceptional)
```
erwango: We should explicit the required conditions to have it accepted. I don't want to open a "Binary Blobs" discussion and process, but exceptional cases are more frequent than we think.

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+  - Use when:
+
+    The interface is expected to become a stable Zephyr API.
```
erwango: Then, can't it be an unstable Zephyr API directly ?

nashif: Any API following the API lifecycle is by definition non vendor specific. We do not accept any APIs that for example have vendor prefixes and such or very specific to one vendor. This bullet probably should just be dropped, it was added here for completeness, but seems to be confusing based on the discussion. We should just say No.

doc/develop/api/vendor_specific_api_policy.rst
teburd: Do you have an example of something like this in mind?

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+Decision Process
+****************
+
+When adding new hardware functionality or vendor-specific behavior,
+contributors must follow the decision tree below, in the given order:
+
+- **Does the feature logically extend an existing driver class?**
+
+  If yes, extend the existing class; do not create a vendor-specific API.
+
```
erwango, line 55: Should we mention the api extensions, at this step, as an existing solution which might be used at some point in the decision process ?

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
+.. _vendor_specific_api_policy:
+
+Vendor-Specific API Placement Policy
+####################################
+
+Overview
+*********
+
+This policy defines how vendor-specific hardware functionality must be contributed,
+structured, and exposed within the Zephyr Project. It ensures that Zephyr’s public API
```
carlescufi, line 10:
```suggestion
structured, and exposed within the Zephyr Project. It ensures that Zephyr’s :term:`public API <public API>`
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+surface remains stable, that vendor extensions do not bypass the standard API lifecycle,
```
carlescufi, line 11:
```suggestion
surface remains stable, that vendor extensions do not bypass the standard :ref:`API lifecycle <api lifecycle>`,
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+and that potential cross-vendor features have a clear path to evolve into
+generic driver classes.
+
+Vendor-specific APIs must not be exposed as public Zephyr interfaces unless they meet
```
carlescufi, line 15:
```suggestion
Vendor-specific APIs must not be exposed as :term:`public APIs <public API>` unless they meet
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+the criteria required for inclusion in Zephyr’s formal API lifecycle
+(experimental → stable → deprecated). Features that are inherently tied to a single
+SoC or vendor must remain encapsulated within the vendor’s SoC directories.
```
carlescufi, line 18:
```suggestion
SoC or vendor must remain encapsulated within the vendor’s SoC (file:`soc/`) directories.
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+This policy provides a consistent decision model for contributors and maintainers.
+
+Guiding Principles
+******************
+
+- **Public Zephyr APIs must not contain vendor-specific interfaces.**
```
carlescufi, line 25:
```suggestion
- **Public APIs must not contain vendor-specific interfaces.**
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+  Interfaces placed under include/zephyr/ are expected to evolve through the standard
+  API lifecycle. Vendor-specific functionality that cannot satisfy this evolution
+  must remain not be exposed as public Zephyr APIs.
```
carlescufi:
```suggestion
  must not be exposed as :term:`public APIs <public API>`
```

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+  When a hardware feature could apply to multiple vendors or might in the future
+  it should be implemented as part of an existing driver class or as a new cross-vendor
+  driver class.
+
+- **Vendor-specific features must be encapsulated.**
+
+  Vendor-only mechanisms should live within the vendor’s platform namespace (SoCs, boards, etc.).
+  If multiple platform from the same vendor share internal logic, a vendor-specific
+  driver directory may be used.
```
carlescufi: I would remove these two lines. The directories are described later.

doc/develop/api/vendor_interfaces_policy.rst
carlescufi: Agreed, here we have to be clear about the case where extending the class is not possible, but vendors need a vendor-specific extension to the API. We must document where those additional APIs are supposed to live, and what prefixes they should use, because for example today:
https://github.com/zephyrproject-rtos/zephyr/tree/main/include/zephyr/drivers/clock_control
we have tons of additional APIs under that folder, most using non-well-defined namespaces

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+  resembles patterns likely to appear across multiple architectures,
+  create a new driver class.
+
+- **Is the functionality clearly vendor-specific and unlikely to generalize?**
+
+  In this case, place both headers and implementations under:
+
+  .. code-block:: none
+
+      soc/<vendor>/<soc>/
```
carlescufi: You have to allow `soc/<vendor>/common` as well for hardware features that are common to multiple series, but unique to the vendor.
Also, we should have a `soc/<vendor>/include` probably? see my other comments related to this

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+  This ensures true encapsulation, long-term maintainability, and clean removal
+  when an SoC is dropped.
+
+- **Is the functionality shared internally across multiple SoCs from the same vendor?**
+
+  If so, place the shared implementation under:
+
+  .. code-block:: none
+
+      drivers/<vendor>/
```
carlescufi: I would remove this. We already have `soc/<vendor>/` for those drivers, and having vendor-specific drivers in multiple places is confusing

doc/develop/api/vendor_specific_api_policy.rst
carlescufi: Also, please note that these will not be public APIs, but will be accessible to other drivers and applications using that particular vendor and SoC

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>/
+      drivers/<class>/
+
+  - Use when:
+
+    The functionality fits or extends an existing class.
+
```
carlescufi, line 99: Here you are missing vendor-specific extensions to existing driver classes, because the one you just described seems to be for API extensions common to all driver class implementations, and not for vendor-specific ones. Also, I assume vendor-specific extensions should not be "Preferred"

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+    .. code-block:: none
+
+      soc/<vendor>/<soc>/
+
+  - Use when:
+
+    The feature is unique to a vendor or SoC and is not intended as a public API.
+
+4. Vendor-Specific Shared Driver Directory
```
carlescufi: I would remove this one

doc/develop/api/vendor_specific_api_policy.rst
```diff
@@ -0,0 +1,154 @@
...
+
+    .. code-block:: none
+
+      drivers/<vendor>/
+
+  - Use when:
+
+    The logic is vendor-specific but shared across multiple SoCs.
+
+5. Vendor-Specific APIs exposed as public Zephyr APIs (Discouraged)
```
carlescufi: If this is about vendor-specific extensions to existing driver classes, see my comment above. Like I said before, I think we should add `soc/<vendor>/include` or something to the default set of include folders, instead of mixing locations for new driver classes that are vendor-specific

doc/develop/api/vendor_interfaces_policy.rst
carlescufi: Yes, I think we should specify where the `.c` files live, and regarding the include mechanism, I already advocated for a `soc/<vendor>/include` folder that would be automatically added to the include path

doc/develop/api/vendor_specific_api_policy.rst
carlescufi: I first need to understand, as I mentioned above, whether those are vendor-specific extensions to existing classes or actual new driver clases

carlescufi (changes requested):

doc/develop/api/vendor_interfaces_policy.rst
teburd: clock_control as a whole is vendor defined per the API so this isn't a great example.

We should really work on fixing that, clocks are not things that are unique, the elements of a clock tree are all very common. Actually this is a great thing point at where I think we failed to find consensus early on something and we are paying the price with the added baggage of years and years of doing it with vendor specifics.

A better example is sensors where there's literally hundreds of devices with often unique features, most of which we've managed to encapsulate in a pretty good API at this point.

There have been a few cases though that warranted some headers for things like loading firmware images on them or fun little sensor fusion algorithms.

So in a way, sensors are already doing this today and doing it (mostly) the right way. As time goes on we have found more things to push into common code. At times some stuff never makes it there. That's ok!

mathieuchopstm:
> I think we should accept 3 types of driver APIs:
> 
> 1. Common, vendor-agnostic driver classes: `.h` in `include/zephyr/driver/<class.h>` and `.c` in `drivers/<class>/<vendor_class.c>`. These are public APIs

I believe this is what we are doing now so LGTM.

> 2. Extensions to existing driver classes: `.h.` in `include/zephyr/driver/<class>/<vendor_class.h>` and `.c` in `drivers/<class>/<vendor_class.c>`. **These are public APIs**

Seems fair to me. This shouldn't be a deterrent either as our `API Lifecycle` allows an API to remain `Unstable` indefinitely, so vendors which don't want to provide stability guarantees can do just that. (to quote: `An API *can* be declared stable after [...]` - not `shall be declared`).

However, I will point out that the [`API Lifecycle` policy](https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html) wording is not currently appropriate for hardware-specific APIs (which these `Extensions to existing driver classes` would all be, I believe?). For example: `[An experimental API for peripherals] shall be promoted to unstable when it has at least two implementations on different hardware platforms` doesn't make sense for APIs which target one specific HW platform. This is a problem since `Public APIs cannot be modified without following the provisions described in the API lifecycle section`. 

Either the `API Lifecycle` documentation needs to be updated, or a specific API lifecycle needs to be documented and applied for these `Extensions to existing driver classes`.

Also, we would need to define a schedule for when this policy becomes effective: there are most likely things currently in `include/zephyr/driver/<class>/<vendor_class.h>` that are **not** intended to be `public API`s - so we can't just say "*everything there is public*" - but we also don't want to have headers rotting forever there.

> 3. Vendor-specific driver classes: `.h` in `soc/<vendor>/include` and `.c` in `soc/<vendor>/common` or `soc/<vendor>/<series>`. These do not fit the current definition of public, private or internal, we would need to add a new category ("vendor-specific API") in the glossary.

<details>
<summary>I don't think this is <i>necessary</i>...</summary>

To quote the glossary, an `internal API` is `any internal function, structure or macro defined anywhere in the Zephyr source tree [...] intended to “extend” Zephyr [...] to be used only between certain software components`, where a `software component` is `a self-contained, modular, and replaceable part of the Zephyr source code` - for example `a driver, a subsystem or an applications[sic]`.

Under this definition, I think vendor-specific driver classes (/APIs) would fit as `internal API`s: they are intended to "extend" Zephyr and to only be used between certain software components: the application (*or vendor-specific drivers*) and the SoC-specific code.

</details>

...but having a dedicated category may still be better (e.g., allows having specific policies for that particular category).

One thing needs clarification for me: *what exactly would go in this class?*

From a look at `soc/st/stm32/common/*.h`, we have *mostly* `internal API`s(?) - helper functions for SoC-specific code and STM32-specific drivers - but some of them *might* be useful to applications as well? For example, `stm32_backup_domain_{dis/en}able_access()` *might* make sense to call from the application if it accesses some BKP domain resources without going through a driver. (This is probably a very common situation: SoC-specific code in Zephyr manages a shared resource, and the application may want to access that resource directly but doing so without cooperating with Zephyr could result in the system entering an inconsistent state).

So would this class contain only `internal`-ish APIs intended for use only by SoC-specific/vendor-specific code? Or only `public`-ish APIs intended for use by application code? Or both, and if so, how would one distinguish them?

Finally, there is one point I disagree with: I don't think we should standardize *where the headers are located in tree* (except that it must be `soc/<vendor>/...`) but *how they are exposed*, i.e. what their include path is. (c.f. my previous comment: https://github.com/zephyrproject-rtos/zephyr/pull/99750#discussion_r2560734964)

<details>
  <summary>Minor remarks on this topic</summary>

This gives more freedom to SoC vendors on how they want to structure their SoC-specific code and APIs.

For example, building for `STM32H7` could add `soc/st/stm32/common/include` ***and*** `soc/st/stm32/stm32h7x/include` to include path, whereas building for `STM32F4` would add `soc/st/stm32/common/include` and `soc/st/stm32/stm32f4x/include` instead. Another vendor could make a `soc/<vendor>/include/` directory with SoC-specific subdirectories instead...

but maybe we **don't** want freedom here, and instead we should enforce a specific format? :man_shrugging:
</details>

> The alternative, if we don't want to add a new category in the glossary (i.e. "Vendor-specific API") that can be placed outside `include/` but at the same time can be used by all components in and out of tree, would be to place vendor-specific driver classes like so: `.h` in `include/zephyr/<vendor>/<class.h>` and `.c` in `drivers/<vendor>/<class.c>` or `soc/<vendor>/<class.c>` or similar.

I think this would break the "encapsulation" of SoC-specific headers/code. @henrikbrixandersen IIRC had a point about HW-specific-but-not-tied-to-SoC things though, maybe we do need `drivers/<vendor>` for that, but it is orthogonal to the SoC-specific question IMO.

carlescufi:
| API | Location of header | Location of implementation | API Type |
| --- | --- | --- | --- | 
| Common, vendor-agnostic driver classes | `include/zephyr/driver/<class.h>` | `drivers/<class>/<vendor_class.c>` | Public API |
| Vendor-specific extensions to existing driver classes | `include/zephyr/driver/<class>/<vendor_class.h>` | `drivers/<class>/<vendor_class.c>` |

doc/develop/api/vendor_specific_api_policy.rst
teburd: `soc/<vendor>` is in my opinion the wrong place for vendor specific drivers.

`drivers/<vendor>` is the right place for these, with added scrutiny (arch wg/requiring +4 approvals, whatever we can figure this out) to ensure we aren't putting things there that could be a driver class.

Unless you are talking about some hardware that you do not want to expose to Zephyr applications? Then perhaps `soc/<vendor>` makes sense but that's not what you are proposing it seems like?

doc/develop/api/vendor_interfaces_policy.rst
teburd: I disagree with this, I would much prefer having drivers in a `drivers/<vendor>` location, which makes sense... these are drivers for hardware blocks after all. I'm perfectly fine if what we need to do to get there is some added scrutiny.

doc/develop/api/vendor_interfaces_policy.rst
nashif: most of those files under clock_control are not really vendor APIs or extensions to the class, those are mostly defines and macros used in the driver implementation. I think nrf is the only one with a true API extension and additional interfaces on top of the common clock control API...

doc/develop/api/vendor_interfaces_policy.rst
nashif: hmmm, why would we support API extensions that are generic this way instead of just extending the API? AFAIK all of this is needed only and only makes sense when the extension is vendor specific, otherwise it is just another interface added to an existing API...

doc/develop/api/vendor_interfaces_policy.rst
teburd: The extension to the API aren't the issue in the case of clock control, the issues I have with it are already well described in #89124

carlescufi:
> > Why do we need vendor specific API in the main tree ?
> 
> looks like the motivation of this PR and policy is not clear enough, it is exactly that, we do NOT want vendor specific functionality used at a low level to bring up hardware to be exposed as zephyr APIs. Application should NOT be using such interfaces directly.
> 
> If we allow this and start doing that, we will have vendors creating SILOs of application facing inerfaces applicable only to those vendors and covered under the zephyr project and exposed as Zephyr project interfaces. Such interface should really be only allowed in very rare cases to support things like IPC, low level memory management, dealing with shared resources etc to support special SoC configurations and setups, they should not be exposed as application facing interfaces or "API"s.
> 
> Extending existing APIs with vendor specific functionality is a completely different area which requires to be revisited and we need to do some cleanup and tweaking to our existing guidelines, but it is not the same as introducing a completely new interface for a new functionality that is vendor specific and exposing it as a zephyr API...

In order for us to understand why we need vendor-specific APIs (something that has come up multiple times before) I propose that we look at real-world examples today in the Zephyr tree. I know there are quite a few of those, but it'd be great to get some representative examples of them.

@mathieuchopstm @henrikbrixandersen @bjarki-andreasen could you perhaps help us list some of the most important ones for the chips you know of?

doc/develop/api/vendor_specific_api_policy.rst
nashif: I am removing this as we all agree we do not want this :)

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,126 @@
...
+This policy provides a consistent decision model for contributors and maintainers.
+
+Guiding Principles
+******************
+
+- **Public Zephyr APIs must not contain vendor-specific interfaces.**
+
+  Interfaces placed under include/zephyr/ are expected to evolve through the standard
+  API lifecycle. Vendor-specific functionality that cannot satisfy this evolution
+  must not be exposed as public Zephyr APIs.
```
mathieuchopstm, line 29: I think this was already commented on, but there is a contradiction between the title (*no vendor-specific interface*) and the description (*vendor-specific functionality is allowed if it follows the Zephyr API lifecycle*).

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,126 @@
...
+SoC or vendor must remain encapsulated within the vendor’s SoC directories.
+
+This policy provides a consistent decision model for contributors and maintainers.
+
+Guiding Principles
+******************
+
+- **Public Zephyr APIs must not contain vendor-specific interfaces.**
+
+  Interfaces placed under include/zephyr/ are expected to evolve through the standard
```
mathieuchopstm:
```suggestion
  Interfaces placed under ``include/zephyr/`` are expected to evolve through the standard
```

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,126 @@
...
+
+  Vendor-specific code should be implemented in headers located under
+  ``soc/<vendor>/include`` and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.
+
+  Such interfaces would be private and/or internal to the vendor to support their
+  specific hardware will be scrutinized in order to avoid creating feature islands specific
+  to vendors.
+
+  Applications using these interfaces, given that they are internal and/or private to the vendor,
+  shall expect no guarantees as to the lifecycle or backwards compatibility.
```
mathieuchopstm, line 82: I feel like these two bullet points could be merged together under the single `Is functionality clearly vendor-specific`

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,126 @@
...
+
+
+- **Exposing Vendor-specific Interfaces as public Zephyr APIs is not allowed**
+
+  Vendor-specific code should be implemented in headers located under
+  ``soc/<vendor>/include`` and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.
+
+  Such interfaces would be private and/or internal to the vendor to support their
+  specific hardware will be scrutinized in order to avoid creating feature islands specific
+  to vendors.
```
mathieuchopstm, line 79: typo?
```suggestion
  Such interfaces would be private and/or internal to the vendor to support their
  specific hardware and will be scrutinized in order to avoid creating feature islands specific
  to vendors.
```

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,126 @@
...
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>/
+      drivers/<class>/
+
+  - Use when:
+
+    The functionality fits or extends an existing class.
```
mathieuchopstm, line 98: From the wording above, `extending a driver class` = add new things 
to the generic API. But then the path should be `include/zephyr/drivers/<class>.h`. Is this what was intended?

doc/develop/api/vendor_interfaces_policy.rst
mathieuchopstm: Does this mean that things like the `SoC-specific flags` in GPIO/PWM, or `HW specific` fields in DMA configurations that exist today will be deprecated and removed, and will have to be replaced with e.g. vendor-specific `vnd_gpio_set_xxx_custom_flag()` for example? This seems *very* heavyweight.

Wouldn't it be much more reasonable to allow vendor-specific fields in public APIs (***when justified, with the usual "if it's actually cross-vendor it should be generic"***/etc provisions) and move these flags currently defined in `include/zephyr/drivers/<class>/<class>_<vnd>.h` to `soc/<vnd>/include/` instead?

doc/develop/api/vendor_interfaces_policy.rst
nashif: changed to use "interfaces", otherwise I do not see a contradiction...

doc/develop/api/vendor_interfaces_policy.rst
nashif: sure

doc/develop/api/vendor_interfaces_policy.rst
carlescufi: I would leave the SoC-specific flags and fields in common driver APIs out of the discussion for now. It's hard enough to get consensus on a base to work on.

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,138 @@
...
+
+
+2. Vendor-specific extensions to existing driver classes
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>
```
carlescufi: Maybe a trailing `/` to show this is intended to be in the folder?

`include/zephyr/drivers/<class>/`

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,138 @@
...
+    The feature is a vendor-specific extension to an existing class, but the core
+    concept is not unique to a single vendor and could be standardized in the future.
+
+3. Create New Driver Class (Cross-Vendor)
+
+  - Where:
+
+    .. code-block:: none
+
+      include/zephyr/drivers/<class>/
```
carlescufi:
```suggestion
      include/zephyr/drivers/<class>.h
```

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,138 @@
...
+      drivers/<class>/
+
+  - Use when:
+
+    The feature could reasonably apply to multiple vendors or evolve into a generic abstraction.
+
+4. Vendor-Specific, SoC-Local Code (Default for Vendor-Unique Features)
```
carlescufi, line 128: This is my only blocking comment:

This should reflect the paragraph above:

>Vendor-specific code should be implemented in headers located under ``soc/<vendor>/include`` and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.

carlescufi (changes requested):
Thanks very much @nashif. I only have one blocking comment that I think needs addressing.

carlescufi (changes requested):
Thanks very much @nashif. I only have one blocking comment.

doc/develop/api/vendor_interfaces_policy.rst
carlescufi: i.e. it should cover both SoC-specific and family/series specific code with the locations for both

henrikbrixandersen (approved):

carlescufi (approved):
LGTM, thanks @nashif

mathieuchopstm (approved):

doc/develop/api/vendor_interfaces_policy.rst
```diff
@@ -0,0 +1,137 @@
...
+
+  - Where:
+
+    Vendor-specific code should be implemented in headers located under ``soc/<vendor>/include``
+    and code in ``soc/<vendor>/common`` or ``soc/<vendor>/<series>``.
```
mathieuchopstm, line 133: For consistency with above (L74-75), it should probably be:

_headers located under `soc/<vendor>/include` **or `soc/<vendor>/<series>/include`**_

mathieuchopstm (commented):
Actually, one nit:

jfischer-no (approved):

PR #99754, opened 2025-11-20
net: config: sntp: Asynchronous resynchronization

rlubos:
Make SNTP resynchronization asynchronous, so that it doesn't block the system work queue while waiting for response.

Fixes #99634

### DIFF
--- a/subsys/net/lib/config/init_clock_sntp.c
+++ b/subsys/net/lib/config/init_clock_sntp.c
@@ -124,10 +124,232 @@ end:
 }
 
 #ifdef CONFIG_NET_CONFIG_SNTP_INIT_RESYNC
-static void sntp_resync_handler(struct k_work *work)
+#define SNTP_SERVER_PORT 123
+
+static void sntp_async_service_handler(struct net_socket_service_event *pev);
+NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(sntp_service_async, sntp_async_service_handler, 1);
+
+static void sntp_async_timeout(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(sntp_async_timeout_work, sntp_async_timeout);
+
+static struct sntp_ctx sntp_async_ctx;
+static struct net_sockaddr_storage sntp_addr;
+static net_socklen_t sntp_addrlen;
+
+static void sntp_async_timeout(struct k_work *work)
 {
 	ARG_UNUSED(work);
-	(void)net_init_clock_via_sntp();
+
+	LOG_WRN("SNTP query timed out");
+
+	sntp_close_async(&sntp_service_async);
+
+	/* No response, reschedule */
+	k_work_reschedule(&sntp_resync_work_handle, RESYNC_FAILED_INTERVAL);
+}
+
+static void sntp_async_service_handler(struct net_socket_service_event *pev)
+{
+	struct sntp_time ts;
+	int ret;
+
+	ret = sntp_read_async(pev, &ts);
+	if (ret < 0) {
+		LOG_ERR("Failed to read SNTP response (%d)", ret);
+		goto out;
+	}
+
+	ret = sntp_set_clocks(&ts);
+
+out:
+	/* Close the service */
+	sntp_close_async(&sntp_service_async);
+
+	k_work_cancel_delayable(&sntp_async_timeout_work);
+
+	if (ret < 0) {
+		k_work_reschedule(&sntp_resync_work_handle, RESYNC_FAILED_INTERVAL);
+	}
+}
+
+static int sntp_query_async(struct net_sockaddr *addr, net_socklen_t addrlen)
+{
+	int ret;
+
+	ret = sntp_init_async(&sntp_async_ctx, addr, addrlen, &sntp_service_async);
+	if (ret < 0) {
+		LOG_ERR("Failed to initialize SNTP context (%d)", ret);
+		goto end;
+	}
+
+	ret = sntp_send_async(&sntp_async_ctx);
+	if (ret < 0) {
+		LOG_ERR("Failed to send SNTP query (%d)", ret);
+		sntp_close_async(&sntp_service_async);
+		goto end;
+	}
+
+	k_work_reschedule(&sntp_async_timeout_work,
+			  K_MSEC(CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT));
+
+end:
+	return ret;
+}
+
+static void dns_result_cb(enum dns_resolve_status status,
+			  struct dns_addrinfo *info,
+			  void *user_data)
+{
+	int ret;
+
+	if (status == DNS_EAI_CANCELED || status == DNS_EAI_FAIL) {
+		/* If IPv4 query failed, try IPv6. Otherwise, just schedule next retry. */
+		if (sntp_addr.ss_family == NET_AF_INET && IS_ENABLED(CONFIG_NET_IPV6)) {
+			sntp_addr.ss_family = NET_AF_INET6;
+			sntp_addrlen = 0;
+
+			ret = dns_get_addr_info(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+						DNS_QUERY_TYPE_AAAA, NULL, dns_result_cb,
+						NULL, CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT);
+			if (ret == 0) {
+				return;
+			}
+		}
+
+		LOG_WRN("DNS query timed out");
+
+		k_work_reschedule(&sntp_resync_work_handle, RESYNC_FAILED_INTERVAL);
+
+		return;
+	}
+
+	if (status == DNS_EAI_ALLDONE) {
+		/* If address was found, schedule SNTP query, if that fails schedule next retry. */
+		ret = sntp_query_async(net_sad(&sntp_addr), sntp_addrlen);
+		if (ret < 0) {
+			k_work_reschedule(&sntp_resync_work_handle, RESYNC_FAILED_INTERVAL);
+		}
+
+		return;
+	}
+
+	if (status == DNS_EAI_INPROGRESS && info != NULL) {
+		if (sntp_addrlen > 0) {
+			/* Already got address, skip others. */
+			return;
+		}
+
+		if (IS_ENABLED(CONFIG_NET_IPV4) && info->ai_family == NET_AF_INET) {
+			sntp_addrlen = info->ai_addrlen;
+			sntp_addr.ss_family = NET_AF_INET;
+			net_ipv4_addr_copy_raw(net_sin(net_sad(&sntp_addr))->sin_addr.s4_addr,
+					       net_sin(&info->ai_addr)->sin_addr.s4_addr);
+		} else if (IS_ENABLED(CONFIG_NET_IPV6) && info->ai_family == NET_AF_INET6) {
+			sntp_addrlen = info->ai_addrlen;
+			sntp_addr.ss_family = NET_AF_INET6;
+			net_ipv6_addr_copy_raw(net_sin6(net_sad(&sntp_addr))->sin6_addr.s6_addr,
+					       net_sin6(&info->ai_addr)->sin6_addr.s6_addr);
+		} else {
+			/* Ignore. */
+			return;
+		}
+
+		(void)net_port_set_default(net_sad(&sntp_addr), SNTP_SERVER_PORT);
+	}
+}
+
+static int dns_query_async(void)
+{
+	enum dns_query_type type;
+	int ret;
+
+	memset(&sntp_addr, 0, sizeof(sntp_addr));
+	sntp_addrlen = 0;
+
+	if (IS_ENABLED(CONFIG_DNS_RESOLVER)) {
+		if (IS_ENABLED(CONFIG_NET_IPV4)) {
+			sntp_addr.ss_family = NET_AF_INET;
+			type = DNS_QUERY_TYPE_A;
+		} else {
+			sntp_addr.ss_family = NET_AF_INET6;
+			type = DNS_QUERY_TYPE_AAAA;
+		}
+
+		ret = dns_get_addr_info(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+					type, NULL, dns_result_cb,
+					NULL, CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT);
+		if (ret < 0) {
+			LOG_ERR("Failed to initiate DNS query for SNTP server (%d)", ret);
+		}
+
+		return ret;
+	}
+
+	/* Fallback for IP address string. */
+	if (net_ipaddr_parse(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+			     sizeof(CONFIG_NET_CONFIG_SNTP_INIT_SERVER) - 1,
+			     net_sad(&sntp_addr))) {
+		if (IS_ENABLED(CONFIG_NET_IPV4) && sntp_addr.ss_family == NET_AF_INET) {
+			sntp_addrlen = sizeof(struct net_sockaddr_in);
+		} else if (IS_ENABLED(CONFIG_NET_IPV6) && sntp_addr.ss_family == NET_AF_INET6) {
+			sntp_addrlen = sizeof(struct net_sockaddr_in6);
+		} else {
+			/* Unsupported family */
+			return -EINVAL;
+		}
+
+		ret = net_port_set_default(net_sad(&sntp_addr), SNTP_SERVER_PORT);
+		if (ret < 0) {
+			return ret;
+		}
+
+		ret = sntp_query_async(net_sad(&sntp_addr), sntp_addrlen);
+	} else {
+		LOG_ERR("Failed to parse SNTP server address, enable CONFIG_DNS_RESOLVER");
+		ret = -EINVAL;
+	}
+
+	return ret;
+}
+
+static void sntp_resync_handler(struct k_work *work)
+{
+	int ret;
+
+	ARG_UNUSED(work);
+
+#ifdef CONFIG_NET_CONFIG_SNTP_INIT_SERVER_USE_DHCPV4_OPTION
+	struct net_if *iface = net_if_get_default();
+
+	if (!net_ipv4_is_addr_unspecified(&iface->config.dhcpv4.ntp_addr)) {
+		sntp_addrlen = sizeof(struct net_sockaddr_in);
+		sntp_addr.ss_family = NET_AF_INET;
+		net_ipv4_addr_copy_raw(net_sin(net_sad(&sntp_addr))->sin_addr.s4_addr,
+					       iface->config.dhcpv4.ntp_addr.s4_addr);
+		net_sin(net_sad(&sntp_addr))->sin_port = net_htons(SNTP_SERVER_PORT);
+
+		ret = sntp_query_async(net_sad(&sntp_addr), sntp_addrlen);
+		if (ret < 0) {
+			LOG_ERR("Cannot set time using SNTP: %d", ret);
+		}
+
+		goto out;
+	}
+
+	if (sizeof(CONFIG_NET_CONFIG_SNTP_INIT_SERVER) == 1) {
+		/* Empty address, skip using SNTP via Kconfig defaults */
+		ret = -EINVAL;
+		goto out;
+	}
+#endif
+
+	ret = dns_query_async();
+
+#ifdef CONFIG_NET_CONFIG_SNTP_INIT_SERVER_USE_DHCPV4_OPTION
+out:
+#endif
+	k_work_reschedule(&sntp_resync_work_handle,
+			  (ret < 0) ? RESYNC_FAILED_INTERVAL : RESYNC_INTERVAL);
 }
 #endif /* CONFIG_NET_CONFIG_SNTP_INIT_RESYNC */
 
--- a/subsys/net/lib/config/init_clock_sntp.c
+++ b/subsys/net/lib/config/init_clock_sntp.c
@@ -82,10 +82,28 @@ static void sntp_set_rtc(__maybe_unused const struct timespec *tspec)
 #endif
 }
 
+static int sntp_set_clocks(struct sntp_time *ts)
+{
+	struct timespec tspec;
+	int ret;
+
+	tspec.tv_sec = ts->seconds;
+	tspec.tv_nsec = ((uint64_t)ts->fraction * NSEC_PER_SEC) >> 32;
+	ret = sys_clock_settime(SYS_CLOCK_REALTIME, &tspec);
+	if (ret < 0) {
+		LOG_ERR("Setting sys clock failed (%d)", ret);
+	}
+
+	sntp_set_rtc(&tspec);
+
+	LOG_DBG("Time synced using SNTP, SNTP Time: %" PRIu64, ts->seconds);
+
+	return ret;
+}
+
 int net_init_clock_via_sntp(void)
 {
 	struct sntp_time ts;
-	struct timespec tspec;
 	int res = sntp_init_helper(&ts);
 
 	if (res < 0) {
@@ -93,13 +111,7 @@ int net_init_clock_via_sntp(void)
 		goto end;
 	}
 
-	tspec.tv_sec = ts.seconds;
-	tspec.tv_nsec = ((uint64_t)ts.fraction * (1000 * 1000 * 1000)) >> 32;
-	res = sys_clock_settime(SYS_CLOCK_REALTIME, &tspec);
-
-	sntp_set_rtc(&tspec);
-
-	LOG_DBG("Time synced using SNTP");
+	res = sntp_set_clocks(&ts);
 
 end:
 #ifdef CONFIG_NET_CONFIG_SNTP_INIT_RESYNC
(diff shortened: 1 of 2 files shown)


subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
 #ifdef CONFIG_NET_CONFIG_SNTP_INIT_RESYNC
+static void sntp_async_service_handler(struct net_socket_service_event *pev);
+NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(service_sntp_async, sntp_async_service_handler, 1);
+
+static void sntp_async_timeout(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(sntp_async_timeout_work, sntp_async_timeout);
+
+struct sntp_ctx sntp_async_ctx;
+struct net_sockaddr_storage sntp_addr;
+net_socklen_t sntp_addrlen;
```
pdgendt: Why aren't these `static`?

rlubos: They should be, fixed now.

jukkar (dismissed):

pdgendt (dismissed):

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+	ret = dns_query_async();
+	if (ret < 0) {
+		LOG_ERR("Failed to initiate DNS query for SNTP server (%d)", ret);
+		goto out;
+	}
+
+out:
+	k_work_reschedule(
+		&sntp_resync_work_handle,
+		(ret < 0) ? K_SECONDS(CONFIG_NET_CONFIG_SNTP_INIT_RESYNC_ON_FAILURE_INTERVAL)
```
jeppenodgaard:
```suggestion
		(ret < 0) ? RESYNC_FAILED_INTERVAL
```

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+static void sntp_async_timeout(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(sntp_async_timeout_work, sntp_async_timeout);
+
+static struct sntp_ctx sntp_async_ctx;
+static struct net_sockaddr_storage sntp_addr;
+static net_socklen_t sntp_addrlen;
+
+#define RESYNC_FAILED_INTERVAL K_SECONDS(CONFIG_NET_CONFIG_SNTP_INIT_RESYNC_ON_FAILURE_INTERVAL)
+
+static void sntp_async_timeout(struct k_work *work)
```
jeppenodgaard, line 139: Nit: AFAIK you can use `__unused` instead of `ARG_UNUSED()`
```suggestion
static void sntp_async_timeout(struct k_work __unused *work)
```

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+{
+	ARG_UNUSED(work);
+
+	LOG_DBG("SNTP query timed out");
```
jeppenodgaard: Also nit: I would elevate this log message to INF/WRN.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+	int ret;
+
+	ret = sntp_read_async(pev, &ts);
+	if (ret < 0) {
+		LOG_ERR("Failed to read SNTP response (%d)", ret);
+		goto out;
+	}
+
+	tspec.tv_sec = ts.seconds;
+	tspec.tv_nsec = ((uint64_t)ts.fraction * (1000 * 1000 * 1000)) >> 32;
```
jeppenodgaard:
```suggestion
	tspec.tv_nsec = ((uint64_t)ts.fraction * NSEC_TO_SEC) >> 32;
```

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+	ret = sys_clock_settime(SYS_CLOCK_REALTIME, &tspec);
+	if (ret < 0) {
+		LOG_ERR("Cannot set time using SNTP (%d)", ret);
+		goto out;
+	}
+
+	sntp_set_rtc(&tspec);
+
+	LOG_DBG("Time synced using SNTP, SNTP Time: %llu", ts.seconds);
```
jeppenodgaard: Nit:
```suggestion
	LOG_DBG("Time synced using SNTP, SNTP Time: %" PRIu64, ts.seconds);
```

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+
+	ret = sntp_send_async(&sntp_async_ctx);
+	if (ret < 0) {
+		LOG_ERR("Failed to send SNTP query (%d)", ret);
+		sntp_close_async(&service_sntp_async);
+		goto end;
+	}
+
+	k_work_reschedule(&sntp_async_timeout_work,
+			  K_MSEC(CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT));
```
jeppenodgaard, line 193: SNTP requests must be send with at least 15 s interval but I guess `sntp_send_async`/`zsock_send` cannot have send anything if it returns an errno?

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+		type = DNS_QUERY_TYPE_A;
+	} else {
+		sntp_addr.ss_family = NET_AF_INET6;
+		type = DNS_QUERY_TYPE_AAAA;
+	}
+
+	ret = dns_get_addr_info(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+				type, NULL, dns_result_cb,
+				NULL, CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT);
+	if (ret < 0) {
```
jeppenodgaard:
```suggestion
	return ret;
```
and remove the lines below.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+		LOG_ERR("Failed to initiate DNS query for SNTP server (%d)", ret);
+		goto out;
+	}
+
+out:
+	k_work_reschedule(
+		&sntp_resync_work_handle,
+		(ret < 0) ? K_SECONDS(CONFIG_NET_CONFIG_SNTP_INIT_RESYNC_ON_FAILURE_INTERVAL)
+			  : K_SECONDS(CONFIG_NET_CONFIG_SNTP_INIT_RESYNC_INTERVAL));
```
maass-hamburg: The regular `net_init_clock_via_sntp` is no longer used, when CONFIG_NET_CONFIG_SNTP_INIT_USE_CONNECTION_MANAGER is enabled, as it uses the resync variant always.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+
+	if (IS_ENABLED(CONFIG_NET_IPV4)) {
+		sntp_addr.ss_family = NET_AF_INET;
+		type = DNS_QUERY_TYPE_A;
+	} else {
+		sntp_addr.ss_family = NET_AF_INET6;
+		type = DNS_QUERY_TYPE_AAAA;
+	}
+
+	ret = dns_get_addr_info(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
```
maass-hamburg: This also works when DNS is enabled, but the Kconfig is a IP address right?
Just like `getaddrinfo()`

jukkar: DNS resolving works fine if given an IP address, it just does not need to resolve anything in that case.

subsys/net/lib/config/init_clock_sntp.c
rlubos: It needs to be visible though as `net_config_init_app()` uses it in a regular if with `IS_ENABLED()` macro:
https://github.com/zephyrproject-rtos/zephyr/blob/b5744c3158c2c0b12ff0e78d680fa653e0cfe3e1/subsys/net/lib/config/init.c#L539-L542

subsys/net/lib/config/init_clock_sntp.c
rlubos: `dns_get_addr_info()` handles IP address strings.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+	struct sntp_time ts;
+	int ret;
+
+	ret = sntp_read_async(pev, &ts);
+	if (ret < 0) {
+		LOG_ERR("Failed to read SNTP response (%d)", ret);
+		goto out;
+	}
+
+	tspec.tv_sec = ts.seconds;
```
jeppenodgaard: These lines are very identical to `net_init_clock_via_sntp`. Can you wrap them in a function?

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
...
+	tspec.tv_nsec = ((uint64_t)ts.fraction * (1000 * 1000 * 1000)) >> 32;
+	ret = sys_clock_settime(SYS_CLOCK_REALTIME, &tspec);
+	if (ret < 0) {
+		LOG_ERR("Cannot set time using SNTP (%d)", ret);
+		goto out;
```
jeppenodgaard: Why not set rtc if set sys clock fails?

jeppenodgaard (commented):
LGTM. Non-blocking comments.
I will try it out now.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -112,10 +112,230 @@ int net_init_clock_via_sntp(void)
 }
 
 #ifdef CONFIG_NET_CONFIG_SNTP_INIT_RESYNC
+static void sntp_async_service_handler(struct net_socket_service_event *pev);
+NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(service_sntp_async, sntp_async_service_handler, 1);
```
jeppenodgaard:
```suggestion
NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(sntp_service_async, sntp_async_service_handler, 1);
```
to attempt prevent name clashing (`samples/net/sockets/sntp_client/src/main.c` uses the same name.)
This is also the prefer used for most other names in this file.

jeppenodgaard (dismissed):
To compile `DNS_RESOLVER` must be enabled. This was not required before.

Oh - and thank you for opening this PR so quickly. I like this solution :)

subsys/net/lib/config/init_clock_sntp.c
rlubos: I've just changed that globally, also for the regular resync interval.

subsys/net/lib/config/init_clock_sntp.c
rlubos: I didn't see a single instance in our codebase where `__unused` was used with function argument, I'll stick to `ARG_UNUSED()` which is a common pattern.

subsys/net/lib/config/init_clock_sntp.c
rlubos: Done

subsys/net/lib/config/init_clock_sntp.c
rlubos: Fixed in helper function

subsys/net/lib/config/init_clock_sntp.c
rlubos: Fixed in helper function

subsys/net/lib/config/init_clock_sntp.c
rlubos: `sntp_service_async` is reinitialized on each iteration so a fresh socket should be opened. The timeout takes care of shutting down the service if no response is received.

subsys/net/lib/config/init_clock_sntp.c
rlubos: This code was refactored anyway due to no DNS_RESOLVER fallback.

subsys/net/lib/config/init_clock_sntp.c
rlubos: Added helper function for setting clocs.

subsys/net/lib/config/init_clock_sntp.c
rlubos: Not happening anymore.

rlubos:
Beside some smaller changes addressing review comments I've added a fallback for a scenario where DNS_RESOLVED is not enabled in the system, please re-review.

subsys/net/lib/config/init_clock_sntp.c
rlubos: Renamed

subsys/net/lib/config/init_clock_sntp.c
maass-hamburg: __maybe_unused is used

maass-hamburg: __maybe_unused is used

pdgendt (dismissed):

subsys/net/lib/config/init_clock_sntp.c
maass-hamburg: Oh I misswrote, i meant when DNS is disabled

subsys/net/lib/config/init_clock_sntp.c
jeppenodgaard: Is there any risk that an SNTP query will be sent if `sntp_send_async` fails? It looks like it boils down to the implementation of `zsock_send`. I do not know that subsystem well enough to tell if it can return an -errno if it sent a valid packet on the wire..

jeppenodgaard (commented):
LGTM!

L4 connect -> disconnect -> connect will always result in a SNTP query:
```bash
[00:01:00.715,000] <dbg> net_config: sntp_set_clocks: Time synced using SNTP, SNTP Time: 1763728695
[00:01:02.364,000] <inf> phy_mii: PHY (0) is down
[00:01:03.965,000] <inf> phy_mii: PHY (0) Link speed 100 Mb, full duplex
...
[00:01:04.812,000] <dbg> net_config: sntp_set_clocks: Time synced using SNTP, SNTP Time: 1763728699
```
I am pretty sure this has always been an issue due to:
```c
	if (mgmt_event == NET_EVENT_L4_CONNECTED) {
		k_work_reschedule(&sntp_resync_work_handle, K_NO_WAIT);
	} else if (mgmt_event == NET_EVENT_L4_DISCONNECTED) {
		k_work_cancel_delayable(&sntp_resync_work_handle);
	}
```

It is probably too corner case to report/fix.

subsys/net/lib/config/init_clock_sntp.c
rlubos: That shouldn't ever be the case, if `zsock_send()` fails the packet should not have been queued for transmission.

Not sure I follow the exact issue here, even if for whatever reason the packet was sent, what's the problem?

rlubos:
> I am pretty sure this has always been an issue due to:
> 
> ```c
> 	if (mgmt_event == NET_EVENT_L4_CONNECTED) {
> 		k_work_reschedule(&sntp_resync_work_handle, K_NO_WAIT);
> 	} else if (mgmt_event == NET_EVENT_L4_DISCONNECTED) {
> 		k_work_cancel_delayable(&sntp_resync_work_handle);
> 	}
> ```
> 
> It is probably too corner case to report/fix.

Is that really an issue though? What's the problem with the query being sent when the interface goes up?

Anyway, it has nothing to do with this PR, please note there's still a NACK from your side.

jukkar (dismissed):

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -82,40 +84,273 @@ static void sntp_set_rtc(__maybe_unused const struct timespec *tspec)
...
+
+#ifdef CONFIG_NET_CONFIG_SNTP_INIT_SERVER_USE_DHCPV4_OPTION
+	struct net_if *iface = net_if_get_default();
+
+	if (!net_ipv4_is_addr_unspecified(&iface->config.dhcpv4.ntp_addr)) {
+		sntp_addrlen = sizeof(struct net_sockaddr_in);
+		sntp_addr.ss_family = NET_AF_INET;
+		net_ipv4_addr_copy_raw(net_sin(net_sad(&sntp_addr))->sin_addr.s4_addr,
+					       iface->config.dhcpv4.ntp_addr.s4_addr);
+		net_sin(net_sad(&sntp_addr))->sin_port = net_htons(123);
```
maass-hamburg: '123' should be a define and it should maybe be said where this value comes from

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -82,40 +84,273 @@ static void sntp_set_rtc(__maybe_unused const struct timespec *tspec)
...
+			sntp_addr.ss_family = NET_AF_INET6;
+			type = DNS_QUERY_TYPE_AAAA;
+		}
+
+		ret = dns_get_addr_info(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+					type, NULL, dns_result_cb,
+					NULL, CONFIG_NET_CONFIG_SNTP_INIT_TIMEOUT);
+		if (ret < 0) {
+			LOG_ERR("Failed to initiate DNS query for SNTP server (%d)", ret);
+		}
```
maass-hamburg, line 283: We could return ret here, so the else case doesn't need to be in a enclosed.

maass-hamburg (changes requested):

subsys/net/lib/config/init_clock_sntp.c
rlubos: Added a symbol for the server port number

subsys/net/lib/config/init_clock_sntp.c
rlubos: Done

subsys/net/lib/config/init_clock_sntp.c
jeppenodgaard: > Not sure I follow the exact issue here, even if for whatever reason the packet was sent, what's the problem?

The packet would default be sent again after 3 s which is not allowed:
https://github.com/zephyrproject-rtos/zephyr/pull/99754#issuecomment-3569583001

Resolving as it should never be the case.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -82,40 +84,276 @@ static void sntp_set_rtc(__maybe_unused const struct timespec *tspec)
...
+	}
+
+	/* Fallback for IP address string. */
+	if (net_ipaddr_parse(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+			     sizeof(CONFIG_NET_CONFIG_SNTP_INIT_SERVER) - 1,
+			     net_sad(&sntp_addr))) {
+		if (sntp_addr.ss_family == NET_AF_INET) {
+			sntp_addrlen = sizeof(struct net_sockaddr_in);
+			net_sin(net_sad(&sntp_addr))->sin_port = net_htons(SNTP_SERVER_PORT);
+		} else {
```
jeppenodgaard: Sorry about the late comment.

Building without IPv6 generates a warning:

```bash
west build -p always -b nucleo_h563zi samples/net/sockets/sntp_client/ -- -DCONFIG_NET_CONFIG_CLOCK_SNTP_INIT=y -DCONFIG_NET_CONFIG_SNTP_INIT_SERVER=\"1.2.3.4\" -DCONFIG_NET_CONFIG_SNTP_INIT_RESYNC=y -DCONFIG_NET_CONFIG_NEED_IPV6=n -DCONFIG_NET_IPV6=n
```

```bash
In function 'dns_query_async',
    inlined from 'sntp_resync_handler' at subsys/net/lib/config/init_clock_sntp.c:347:8:
subsys/net/lib/config/init_clock_sntp.c:303:74: warning: array subscript 'struct net_sockaddr_in6[0]' is partly outside array bounds of 'struct net_sockaddr_storage[1]' [-Warray-bounds]
  303 |                                 net_sin6(net_sad(&sntp_addr))->sin6_port = net_htons(123);
subsys/net/lib/config/init_clock_sntp.c: In function 'sntp_resync_handler':
subsys/net/lib/config/init_clock_sntp.c:134:36: note: object 'sntp_addr' of size 8
  134 | static struct net_sockaddr_storage sntp_addr;
```

This fixes the warning but I have no idea if that's how to do it.
```suggestion
		} else if (IS_ENABLED(CONFIG_NET_IPV6)) {
```

jeppenodgaard: Not sure why the build command does not trigger another warning.

I need to change this as well with my custom setup:
```diff
diff --git a/subsys/net/lib/config/init_clock_sntp.c b/subsys/net/lib/config/init_clock_sntp.c
index d8e259718a6..895b4fdebab 100644
--- a/subsys/net/lib/config/init_clock_sntp.c
+++ b/subsys/net/lib/config/init_clock_sntp.c
@@ -232,36 +232,23 @@ static void dns_result_cb(enum dns_resolve_status status,
        }

        if (status == DNS_EAI_INPROGRESS && info != NULL) {
-               switch (info->ai_family) {
-               case NET_AF_INET:
-                       if (sntp_addrlen > 0) {
-                               /* Already got address, skip others. */
-                               break;
-                       }
+               if (sntp_addrlen > 0) {
+                       /* Already got address, skip others. */
+                       return;
+               }

+               if (IS_ENABLED(CONFIG_NET_IPV4) && info->ai_family == NET_AF_INET) {
                        sntp_addrlen = info->ai_addrlen;
                        sntp_addr.ss_family = NET_AF_INET;
                        net_ipv4_addr_copy_raw(net_sin(net_sad(&sntp_addr))->sin_addr.s4_addr,
                                               net_sin(&info->ai_addr)->sin_addr.s4_addr);
                        net_sin(net_sad(&sntp_addr))->sin_port = net_htons(123);
-                       break;
-
-               case NET_AF_INET6:
-                       if (sntp_addrlen > 0) {
-                               /* Already got address, skip others. */
-                               break;
-                       }
-
-                       sntp_addrlen = info->ai_addrlen;
+               } else if (IS_ENABLED(CONFIG_NET_IPV6) && info->ai_family == NET_AF_INET6) {
+                                               sntp_addrlen = info->ai_addrlen;
                        sntp_addr.ss_family = NET_AF_INET6;
                        net_ipv6_addr_copy_raw(net_sin6(net_sad(&sntp_addr))->sin6_addr.s6_addr,
                                               net_sin6(&info->ai_addr)->sin6_addr.s6_addr);
                        net_sin6(net_sad(&sntp_addr))->sin6_port = net_htons(123);
-                       break;
-
-               default:
-                       /* Ignore other records. */
-                       break;
                }
        }
 }
```

rlubos: Ugh, thanks for noticing, that async DNS part is getting more and more cumbersome...

I've changed that code to use `if()` with extra config checks as suggested.

jeppenodgaard: https://github.com/zephyrproject-rtos/zephyr/pull/99754#discussion_r2555511680 still needs to be addressed.

rlubos: > [#99754 (comment)](https://github.com/zephyrproject-rtos/zephyr/pull/99754#discussion_r2555511680) still needs to be addressed.

Fixed now, sorry I've missed that somehow.

subsys/net/lib/config/init_clock_sntp.c
```diff
@@ -82,40 +84,263 @@ static void sntp_set_rtc(__maybe_unused const struct timespec *tspec)
...
+	if (net_ipaddr_parse(CONFIG_NET_CONFIG_SNTP_INIT_SERVER,
+			     sizeof(CONFIG_NET_CONFIG_SNTP_INIT_SERVER) - 1,
+			     net_sad(&sntp_addr))) {
+		if (sntp_addr.ss_family == NET_AF_INET) {
+			sntp_addrlen = sizeof(struct net_sockaddr_in);
+			net_sin(net_sad(&sntp_addr))->sin_port = net_htons(SNTP_SERVER_PORT);
+		} else if (IS_ENABLED(CONFIG_NET_IPV6)) {
+			sntp_addrlen = sizeof(struct net_sockaddr_in6);
+			net_sin6(net_sad(&sntp_addr))->sin6_port = net_htons(SNTP_SERVER_PORT);
+		}
```
maass-hamburg: why not using ``net_port_set_default()`` like in ``sntp_simple_addr()``, also we don't want to overwrite the port if it is set in CONFIG_NET_CONFIG_SNTP_INIT_SERVER

maass-hamburg (changes requested):

subsys/net/lib/config/init_clock_sntp.c
rlubos: Totally forgot we had function like this. Good point about not overwriting custom port number, I've updated both cases to use `net_port_set_default()` instead of setting port manually.

jukkar (approved):

maass-hamburg (approved):

PR #99756, opened 2025-11-20
drivers: usb: udc: fix OTGHS on STM32F7 SoCs with USBPHYC

mathieuchopstm:
From testing on STM32F723E-DISCO, it seems necessary to enable `OTGHSULPI` clock in addition to `USBPHYC` when the internal `USBPHYC` HS PHY is used. (`USBPHYC` is found on STM32F723xx, STM32F730Z8 and STM32F730I8 SoCs)


jfischer-no (approved):

jfischer-no:
>Tested on HW using samples/subsys/usb/cdc_acm with the following applied:

On stm32f723e_disco, why is the high-speed controller not the default one?

erwango (approved):

PR #99760, opened 2025-11-20
subsys: tracing: sysview: add config option to name selected ISRs.

sidprice:
This commit allows a user to select specific ISRs to be named in the SystemView output. The format for the option value is as described in the SystemView documentation, e.g. "I#33=RTC1, I#19=SPI0/TWI0".

### DIFF
--- a/subsys/tracing/sysview/Kconfig
+++ b/subsys/tracing/sysview/Kconfig
@@ -32,4 +32,11 @@ config SEGGER_SYSVIEW_SECTION_DTCM
 
 endchoice
 
+config SEGGER_SYSVIEW_CUSTOM_ISR_NAMES
+	string "SystemView ISR name descriptions"
+	help
+	  Custom ISR name descriptions for SystemView tracing.
+	  Format: "I#<num>=<name>, I#<num>=<name>, ..."
+	  Example: "I#33=RTC1, I#19=SPI0/TWI0"
+
 endif
--- a/subsys/tracing/sysview/sysview_config.c
+++ b/subsys/tracing/sysview/sysview_config.c
@@ -62,7 +62,9 @@ static void cbSendSystemDesc(void)
 	SEGGER_SYSVIEW_SendSysDesc("C=" CONFIG_BOARD_QUALIFIERS);
 #endif
 
-#ifdef CONFIG_SYMTAB
+#ifdef CONFIG_SEGGER_SYSVIEW_CUSTOM_ISR_NAMES
+	SEGGER_SYSVIEW_SendSysDesc(CONFIG_SEGGER_SYSVIEW_CUSTOM_ISR_NAMES);
+#elif defined(CONFIG_SYMTAB)
 	char isr_desc[SEGGER_SYSVIEW_MAX_STRING_LEN];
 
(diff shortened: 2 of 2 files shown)


carlescufi (dismissed):
LGTM, perhaps I'd add a link to the relevant SystemView documentation that defines the format of the ISR names

carlescufi (dismissed):

subsys/tracing/sysview/Kconfig
```diff
@@ -32,4 +32,11 @@ config SEGGER_SYSVIEW_SECTION_DTCM
 
 endchoice
 
+config SEGGER_SYSVIEW_ISR_NAMES
```
carlescufi: Looking at this again, perhaps this should be named `SEGGER_SYSVIEW_CUSTOM_ISR_NAMES` because in the source there's already support for ISR names (although not the ones you may be interested in and with names that you don't want)

sidprice: Thanks, I made the suggested changes.

carlescufi (approved):

subsys/tracing/sysview/Kconfig
```diff
@@ -32,4 +32,11 @@ config SEGGER_SYSVIEW_SECTION_DTCM
 
 endchoice
 
+config SEGGER_SYSVIEW_CUSTOM_ISR_NAMES
```
kartben, line 35: please add to new kconfig option to 4.4. release notes

sidprice: Is this comment addressed to me?

If so, how/where do I do that?

kartben: In doc/releases/release-notes-4.4.rst please add a Tracing section under "New APIs and options" where you mention the new option similar to how it's been done for e.g. Settings. Thanks!

nashif:
Using kconfigs for this type of information, which depending on how many ISRs you have an that they are named can get very "busy" is not the right thing IMO. This also just adds that to sysview and other tracing tools will not benefit from, we need this implemented at the tracing subsystem level to support tracealyzer and ctf as well.

I will be looking into this and how we can provide a generic solution that does not involve kconfigs.

sidprice:
@nashif Thanks for your comment about this PR and the approach I proposed. If there is a "generic", non-KConfig solution I would be thrilled. However, I do not have the domain knowledge or the required trace tools to investigate and/or implement that approach.

Given that, my main area of disagreement with your comment is with things getting busy. The current implementation using SYMTAB outputs every ISR on the MCU. On an nRF MCU for example, this adds 40+ traces to the SystemView timeline. I really don't think it could get busier than that.

Also, SystemView is a debugging tool and generally one focuses on just a few of the ISRs while chasing problems and/or doing performance measurement on specific features of the target system. Using "plain English" names, filtered to a specific area of interest brings a level of focus debugging often requires.

So, again, I hear your comment about a generic tracing solution, however, I am not the person who could pursue that.

Finally, it is a pity that discussion was not raised earlier on Discord, where I raised the subject and mentioned the approach I had in mind. That post got no comments as I recall.

PR #99767, opened 2025-11-20
drivers: riscv: Add AIA (Advanced Interrupt Architecture) support

AFOliveira:
This PR implements support for RISC-V Advanced Interrupt Architecture (AIA) as proposed in RFC #82487, with some changes from the RFC scope.

## Overview

Implements the APLIC (Advanced Platform-Level Interrupt Controller) and IMSIC (Incoming Message-Signaled Interrupt Controller) drivers, enabling AIA support for RISC-V SoCs. The implementation provides a PLIC-compatible interface while supporting modern MSI-based interrupt delivery. Also enables QEMU support to ensure correctness.

## Implementation

Following the RFC guidance, this implementation:

- **Presents a PLIC-like interface**: Applications use standard `irq_enable()` / `irq_disable()` APIs
- **Hides AIA complexity**: Unified `riscv_aia.h` API abstracts APLIC+IMSIC hierarchy and MSI configuration
- **Integrates with privileged RISC-V common code**: Uses same pattern as PLIC in `soc_common_irq.c`
- **Supports wired interrupts**: Full peripheral interrupt routing via APLIC → IMSIC → MEXT

## Beyond RFC Scope

The RFC proposed a single driver that "could be subdivided internally" and would later be "reworked into two separate drivers." This implementation goes directly to that split architecture:

- **Separate drivers from start**: `intc_riscv_imsic.c`, `intc_riscv_aplic_msi.c`, and `intc_riscv_aia.c` coordinator
- **SMP support**: Per-CPU IMSIC initialization with interrupt migration between harts
- **Devicetree separation**: Separate nodes for APLIC and per-hart IMSIC instances

## Testing

- Two working samples: `uart_echo_aia` (single-hart) and `uart_echo_aia_smp` (multi-hart)
- Test suite
- Tested on QEMU RISC-V32 AIA platform
- Demonstrates interrupt delivery, routing, and hart migration

I am keeping this as draft for now as I still need to catch the _infinite_ loose ends that continuously appear. Reviews are very welcome.


dts/bindings/interrupt-controller/riscv,imsic.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+properties:
+  reg:
+    required: true
+    description: Base address and size of the IMSIC file
+
+  riscv,num-ids:
+    type: int
+    description: Number of interrupt identities (EIIDs) supported
+    default: 256
```
fkokosinski: default value must be justified/explained in the descritpion

dts/bindings/interrupt-controller/riscv,imsic.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  riscv,hart-id:
+    type: int
+    description: Hart ID associated with this IMSIC instance
+    default: 0
```
fkokosinski: ditto

drivers/interrupt_controller/intc_riscv_imsic.c
```diff
@@ -0,0 +1,244 @@
...
+	uint32_t p1 = read_imsic_csr(ICSR_EIP1);
+
+	return p0 | (p1 ? 0x80000000U : 0U); /* signal >32 pending via MSB */
+}
+
+/* Separate IRQ registration for hart 0 vs other harts to avoid duplicate registration */
+static void imsic_irq_config_func_0(void)
+{
+	/* Only hart 0 (instance 0) registers the global MEXT IRQ handler */
+	IRQ_CONNECT(11, 0, imsic_mext_isr, DEVICE_DT_INST_GET(0), 0);
```
fkokosinski: Hardcoded IRQ number?

include/zephyr/drivers/interrupt_controller/riscv_imsic.h
```diff
@@ -89,10 +94,12 @@ void riscv_imsic_complete(uint32_t eiid);
...
  * be called from that hart (e.g., using k_thread_cpu_mask_enable).
  *
+ * Following PLIC pattern: no parameter validation at API level.
+ * Invalid EIIDs are caught in the ISR if they fire.
+ *
  * @param eiid External Interrupt ID to enable (0-2047)
- * @return 0 on success, negative errno on failure
  */
-int riscv_imsic_enable_eiid(uint32_t eiid);
+void riscv_imsic_enable_eiid(uint32_t eiid);
```
fkokosinski, line 64: shouldn't these symbols be declared as voids in the previous commit?

include/zephyr/drivers/interrupt_controller/riscv_imsic.h
```diff
@@ -101,10 +108,12 @@ int riscv_imsic_enable_eiid(uint32_t eiid);
...
  * be called from that hart.
  *
+ * Following PLIC pattern: no parameter validation at API level.
+ * Invalid EIIDs are caught in the ISR if they fire.
+ *
  * @param eiid External Interrupt ID to disable (0-2047)
- * @return 0 on success, negative errno on failure
  */
-int riscv_imsic_disable_eiid(uint32_t eiid);
+void riscv_imsic_disable_eiid(uint32_t eiid);
```
fkokosinski, line 78: ditto

include/zephyr/drivers/interrupt_controller/riscv_aia.h
```diff
@@ -13,11 +13,14 @@
...
-int riscv_aia_irq_is_enabled(uint32_t irq);
-void riscv_aia_set_priority(uint32_t irq, uint32_t prio);
+/* Check if source is an APLIC source vs local interrupt */
+bool riscv_aia_is_aplic_source(uint32_t src);
+
+/* APLIC source control (1:1 EIID mapping) */
+void riscv_aia_irq_enable(uint32_t src);
+void riscv_aia_irq_disable(uint32_t src);
+int riscv_aia_irq_is_enabled(uint32_t src);
+void riscv_aia_set_priority(uint32_t src, uint32_t prio);
```
fkokosinski, line 69: why are we reformatting this header in this commit? should this be formatted this way in the commit that originally adds this header?

fkokosinski (commented):
Thanks a lot for this PR! Please note that since the changes you introduce here are significant (2,5k+ LOC added), it will take some time for me to dig through this :)

Here's some remarks/questions for starters

dts/bindings/interrupt-controller/riscv,imsic.yaml
AFOliveira: Thanks for pointing that out, indeed I followed a common-value that isn't really a required default, just a value that I looked at as a default. I'll replace it with the spec-mandated min and max values.

drivers/interrupt_controller/intc_riscv_imsic.c
AFOliveira: It's spec-mandated that interrupt 11 is Machine External Interrupt, however I think that magic wasn't very smart on my side, do you agree with substituting it with RISCV_IRQ_MEXT macro?

include/zephyr/drivers/interrupt_controller/riscv_imsic.h
AFOliveira: of course, I just realized I messed commit history, I'll fix it.

include/zephyr/drivers/interrupt_controller/riscv_aia.h
AFOliveira: I just realized I messed commit history with fix-ups, I'll fix it.

drivers/interrupt_controller/intc_riscv_imsic.c
```diff
@@ -0,0 +1,291 @@
...
+
+	return 0;
+}
+
+/* Runtime API: claim interrupt (read and acknowledge) */
+uint32_t riscv_imsic_claim(void)
+{
+	uint32_t topei;
+
+	__asm__ volatile("csrrw %0, " STRINGIFY(CSR_MTOPEI) ", x0" : "=r"(topei));
```
christian-herber-nxp: add read/write function to https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/arch/riscv/csr.h and use this instead.

include/zephyr/drivers/interrupt_controller/riscv_imsic.h
```diff
@@ -0,0 +1,149 @@
...
+#define EIDELIVERY_MODE_DMSI (1U << 29) /* DMSI only: 01 = 0x20000000 */
+#define EIDELIVERY_MODE_DDI  (2U << 29) /* DDI (not supported): 10 = 0x40000000 */
+#define EIDELIVERY_MODE_BOTH (3U << 29) /* Both: 11 = 0x60000000 */
+
+/* Helper functions for IMSIC indirect CSR access */
+static inline uint32_t read_imsic_csr(uint32_t icsr_addr)
+{
+	uint32_t value;
+
+	__asm__ volatile("csrw 0x350, %0" : : "r"(icsr_addr)); /* miselect */
```
christian-herber-nxp: https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/arch/riscv/csr.h already defines icsr read/write function. For sure, those should be used. I do not believe an additional imsic csr access function adds readability over that.

drivers/interrupt_controller/CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ zephyr_library_sources_ifdef(CONFIG_NXP_SIUL2_EIRQ intc_nxp_siul2_eirq.c)
 zephyr_library_sources_ifdef(CONFIG_PLIC intc_plic.c)
 zephyr_library_sources_ifdef(CONFIG_RENESAS_RX_ICU intc_renesas_rx_icu.c)
 zephyr_library_sources_ifdef(CONFIG_RENESAS_RZ_EXT_IRQ intc_renesas_rz_ext_irq.c)
+zephyr_library_sources_ifdef(CONFIG_RISCV_AIA intc_riscv_aia.c)
+zephyr_library_sources_ifdef(CONFIG_RISCV_APLIC_MSI intc_riscv_aplic_msi.c)
+zephyr_library_sources_ifdef(CONFIG_RISCV_IMSIC intc_riscv_imsic.c)
```
fkokosinski, line 66: non-IMSIC lines are included here

drivers/interrupt_controller/intc_riscv_imsic.c
```diff
@@ -0,0 +1,247 @@
+/*
```
fkokosinski, line 1: No KConfig for this driver?

drivers/interrupt_controller/Kconfig.riscv_aia
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2025 Synopsys, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+config RISCV_IMSIC
```
fkokosinski, line 4: IMSIC has been added in the previous commit

drivers/interrupt_controller/intc_riscv_aia.c
```diff
@@ -0,0 +1,170 @@
+/*
```
fkokosinski, line 1: Missing Kconfig symbol for this driver in this commit (has been added in the previous commit)

drivers/interrupt_controller/Kconfig.riscv_aia
```diff
@@ -0,0 +1,28 @@
...
+	select RISCV_IMSIC
+	help
+	  RISC-V Advanced Platform-Level Interrupt Controller (APLIC) in MSI mode.
+	  Routes wire interrupts to IMSIC as Message-Signaled Interrupts.
+	  Automatically enabled when an APLIC-MSI node exists in devicetree.
+
+config RISCV_AIA
+	bool "RISC-V AIA unified coordinator"
+	default y
+	depends on RISCV_IMSIC || RISCV_APLIC_MSI
```
fkokosinski, line 26: should't this depend on `RISCV_HAS_AIA` as well?

fkokosinski (commented):
Some problems w/r to the commit history still unfortunately persist.

Could we split this PR into two: one which adds the base AIA support and one of the irq controller drivers, and then a subsequent one adding the other controller?

Judging by Kconfig, these two are fairly independent of each other. I reckon that would reduce the review time quite significantly.

drivers/interrupt_controller/Kconfig.riscv_aia
AFOliveira: sorry for the continuation of the error.

drivers/interrupt_controller/intc_riscv_imsic.c
AFOliveira: I tried to put them both (APLIC and IMSIC) inside the AIA Kconfig as AIA to keep them unified, i.e. I don't expect anyone to use both drivers separately.

drivers/interrupt_controller/Kconfig.riscv_aia
AFOliveira: PLIC also doesn't depend on any RISCV_HAS_PLIC - it just depends on DT_HAS_SIFIVE_PLIC_1_0_0_ENABLED, I just tried to follow the same approach. but with  

 - RISCV_IMSIC: default y if DT_HAS_RISCV_IMSIC_ENABLED
 - RISCV_APLIC_MSI: default y if DT_HAS_RISCV_APLIC_MSI_ENABLED
 
 I think this makes sense as AIA is kind of the abstraction of both, but I may add that dependency as well!

drivers/interrupt_controller/intc_riscv_aplic_msi.c
```diff
@@ -0,0 +1,282 @@
...
+{
+	const struct aplic_cfg *cfg = dev->config;
+	uint32_t v = rd32(cfg->base, APLIC_DOMAINCFG);
+
+	if (enable) {
+		/* Enable domain; keep delivery mode as configured by platform.
+		 * Do not modify DM bit - respect hardware/props configuration.
+		 */
+		v |= APLIC_DOMAINCFG_IE;
+		v &= ~APLIC_DOMAINCFG_BE; /* LE */
```
christian-herber-nxp: It seems strange that clearing of BE would be hardcoded here. 
99.9% of SoCs will be LE (with a 0.1% uncertainty).
As this is a read-modify-write sequence, and I would expect this LE systems to be configured for LE out of reset, this line seems unnecessary.

If you truly want to set it, you should rather make the setting based on __BYTE_ORDER__ macro.

AFOliveira: Yeah, I think it's very much redudant to have BE there, you're right. I'll just remove it and hope Linus never ever sees this :)

PR #99777, opened 2025-11-21
drivers: espi: npcx: workaround for missed IRQ1/12 on NPCX9M7FB

TomChang19:
When the firmware writes data to the KBC/AUX output buffer, the hardware automatically generates IRQ1/12 to notify the host. After the host read the data, the EC received an output buffer empty (OBE) interrupt to indicate that the next byte could be written. However, on the NPCX9M7FB, this mechanism may cause IRQ1/12 to be missed if the firmware writes new data immediately after detecting that the output buffer is empty, resulting in incorrect behavior.

To avoid missing IRQ1/12, the flow is modified so that it first checks VWIRE_AVAIL to ensure there are no pending VW events. If no VW event is pending, the firmware writes data to the output buffer and explicitly asserts IRQ1/12 to notify the host. After the host reads the data and the OBE interrupt triggers, the firmware clears IRQ1/12.

This updated sequence prevents IRQ1/12 from being lost and ensures correct notification behavior.

### DIFF
--- a/drivers/espi/host_subs_npcx.c
+++ b/drivers/espi/host_subs_npcx.c
@@ -293,6 +293,9 @@ static void host_kbc_obe_isr(const void *arg)
 
 	LOG_DBG("%s: kbc status 0x%02x", __func__, inst_kbc->HIKMST);
 
+	if (IS_ENABLED(CONFIG_ESPI_NPCX_i8042_KBC_AUX_VWIRE_IRQ_WORKAROUND)) {
+		inst_kbc->HIIRQC &= ~(BIT(NPCX_HIIRQC_IRQ1B) | BIT(NPCX_HIIRQC_IRQ12B));
+	}
 	/*
 	 * Notify application that host already read out data. The application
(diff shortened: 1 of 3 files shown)


fabiobaltieri (dismissed):

drivers/espi/host_subs_npcx.c
```diff
@@ -293,6 +293,10 @@ static void host_kbc_obe_isr(const void *arg)
 
 	LOG_DBG("%s: kbc status 0x%02x", __func__, inst_kbc->HIKMST);
 
+#if defined(CONFIG_ESPI_NPCX_i8042_KBC_AUX_VWIRE_IRQ_WORKAROUND)
```
keith-zephyr: Non-blocking nit.  It is preferred to use `if IS_ENABLED(CONFIG...))` when possible.  I think you can convert all #ifdefs in this file.

TomChang19: Done.

fabiobaltieri (approved):

albertofloyd (approved):

PR #99784, opened 2025-11-21
drivers: counter: lptmr: Fix get_pending_int wrong implementation

ZhaoxiangJin:
Originally, function 'mcux_lptmr_get_pending_int' checks if both
TCF and TIE flags are set to determine if an interrupt is pending.
But function 'LPTMR_GetStatusFlags' only returns the status flags,
that is we will never get TIE flag from it. then the function
always returns false.

The correct approach is to read the CSR register directly and check
if both TIE and TCF bits are 1. If yes, there's a pending interrupt;
if not, there isn't.

### DIFF
--- a/drivers/counter/counter_mcux_lptmr.c
+++ b/drivers/counter/counter_mcux_lptmr.c
@@ -202,11 +202,8 @@ static uint32_t mcux_lptmr_get_pending_int(const struct device *dev)
 {
 	const struct mcux_lptmr_config *config = dev->config;
 	uint32_t mask = LPTMR_CSR_TCF_MASK | LPTMR_CSR_TIE_MASK;
-	uint32_t flags;
 
-	flags = LPTMR_GetStatusFlags(config->base);
-
-	return ((flags & mask) == mask);
+	return (uint32_t)!!((config->base->CSR & mask) == mask);
 }
 
 static uint32_t mcux_lptmr_get_top_value(const struct device *dev)


ZhaoxiangJin:
CC @FelixWang47831

FelixWang47831 (approved):

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -198,6 +200,7 @@ static DEVICE_API(counter, mcux_lptmr_driver_api) = {
 		.mode = DT_INST_PROP(n, timer_mode_sel),			\
 		.pin = DT_INST_PROP_OR(n, input_pin, 0),			\
 		.polarity = DT_INST_PROP(n, active_low),			\
+		.free_running = DT_INST_PROP(n, free_running),			\
```
Holt-Sun: You used DT_INST_PROP(n, free_running) which requires the property to exist, but binding does not mark free-running as required, builds will fail if older DTS files omit it.

```suggestion
		.free_running = DT_INST_PROP_OR(n, free_running, 0),			\
```

ZhaoxiangJin: boolean property default false, so there is no build issue if this property is not set.

Holt-Sun (dismissed):

FelixWang47831 (approved):

zejiang0jason:
Hi, I am not sure about the purpose, from the counter driver model, my understanding is the peripheral counter should be reset when  reached the top value. What is the purpose of free-running mode?

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -91,12 +93,8 @@ static int mcux_lptmr_set_top_value(const struct device *dev,
 static uint32_t mcux_lptmr_get_pending_int(const struct device *dev)
 {
 	const struct mcux_lptmr_config *config = dev->config;
-	uint32_t mask = LPTMR_CSR_TCF_MASK | LPTMR_CSR_TIE_MASK;
-	uint32_t flags;
-
-	flags = LPTMR_GetStatusFlags(config->base);
 
-	return ((flags & mask) == mask);
+	return return (uint32_t)!!LPTMR_GetStatusFlags(config->base);
```
FelixWang47831: TCF(Timer Compare Flag) is set when counter equals compared value even if TIE(Timer Interrupt Enable) is not set. 
Since this API is to get pending interrupt, so I think it's necessary to check TIE:

`return (config->base->CSR & mask) == mask;`

ZhaoxiangJin: you are correct, i have updated the patch

ZhaoxiangJin:
Hello @Holt-Sun, @FelixWang47831, any comments regarding this [topic](https://github.com/zephyrproject-rtos/zephyr/pull/99784#issuecomment-3573936096)? Thanks.

Holt-Sun (approved):

ZhaoxiangJin:
> > I rhink the current top callback can meet your free-running alarm needs.
> 
> Yes, top feature allows to set lower value at which counter will reset. In implementations that i did i actually use one of the compare channels to set that. If not set then counter resets after reaching it's HW top value.

So in reality, whether to reset after reaching top is decided by the driver itself rather than being mandated by the API.

ZhaoxiangJin:
Add Holt as the PR assignee since he is the NXP counter expert and this PR only changes the NXP driver part.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -94,11 +94,8 @@ static uint32_t mcux_lptmr_get_pending_int(const struct device *dev)
 {
 	const struct mcux_lptmr_config *config = dev->config;
 	uint32_t mask = LPTMR_CSR_TCF_MASK | LPTMR_CSR_TIE_MASK;
-	uint32_t flags;
-
-	flags = LPTMR_GetStatusFlags(config->base);
 
-	return ((flags & mask) == mask);
+	return (uint32_t)!!((config->base->CSR & mask) == mask);
```
zejiang0jason, line 206: this doesn't match the commit message description:
> The correct behavior is to check only the TCF
(Timer Compare Flag) to see if an interrupt is pending,
regardless of the TIE (Timer Interrupt Enable) status.
This commit updates the function to return the TCF flag
directly.

ZhaoxiangJin: commit message updated.

Holt-Sun (dismissed):

nordic-krch:
> I think it's strange to assume that software should restrict counter hardware from performing a hardware wrap when reaching top.

I don't recall such assumption. Counter API assumes that counter always wraps after reaching top. There is the flag `COUNTER_TOP_CFG_DONT_RESET` but it only indicates what to do just after updating the top (reset or continue to run as is).

Holt-Sun (approved):

nordic-krch (approved):

PR #99786, opened 2025-11-21
mgmt: mcumgr: grp: img_mgmt: Fix detecting where a slot resides

### LINKED ISSUE #99762: mcumgr: Image management incorrectly identifies active slot when slots are on different flash devices
### Describe the bug

When MCUboot is configured with `slot0_partition` and `slot1_partition` on different flash devices (e.g., slot 1 on external flash), mcumgr image management cannot correctly identify the active application partition. This results in uploading new firmware to the wrong slot, corrupting the currently running image and making the device unbootable.

### Regression

- [x] This is a regression.

### Impact

### Additional Context

The `FIXED_PARTITION_IS_RUNNING_APP_PARTITION` macro used in `img_mgmt_active_slot()` only compares the partition offset against `CONFIG_FLASH_LOAD_OFFSET` without considering which flash device the partition resides on.

https://github.com/zephyrproject-rtos/zephyr/blob/32615695ad1b96670be8f2a38a8a9fd27e2d8ae5/subsys/dfu/img_util/flash_img.c#L26-L28

When two partitions on different flash devices have overlapping offset ranges that include `CONFIG_FLASH_LOAD_OFFSET`, the macro can incorrectly identify a partition on external flash as the running partition.

In Zephyr versions prior to v4.3.0, the implementation used exact equality, so this could accidentally work in multi-device configurations when the slot offsets differed between flash devices.

nordicjm:
Fixes an issue introduced in commit
32615695ad1b96670be8f2a38a8a9fd27e2d8ae5 which wrongly did not check what the residing device was on before determining if a slot was part of a partition area

Fixes #99762

### DIFF
--- a/subsys/dfu/img_util/flash_img.c
+++ b/subsys/dfu/img_util/flash_img.c
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/devicetree.h>
 #include <zephyr/dfu/mcuboot.h>
 #include <zephyr/dfu/flash_img.h>
 #include <zephyr/dfu/mcuboot.h>
@@ -23,11 +24,16 @@ LOG_MODULE_REGISTER(flash_img, CONFIG_IMG_MANAGER_LOG_LEVEL);
 #include <bootutil/bootutil_public.h>
 #endif
 
+#define FIXED_PARTITION_GET_FLASH_NODE(node_id)                                    \
+	COND_CODE_1(DT_NODE_HAS_COMPAT(DT_PARENT(node_id), fixed_subpartitions),   \
+		    (DT_PARENT(DT_GPARENT(node_id))), (DT_GPARENT(node_id)))
+
 #define FIXED_PARTITION_IS_RUNNING_APP_PARTITION(label)                                            \
+	DT_SAME_NODE(FIXED_PARTITION_GET_FLASH_NODE(DT_CHOSEN(zephyr_code_partition)),             \
+		     FIXED_PARTITION_GET_FLASH_NODE(DT_NODELABEL(label))) &&                       \
 	(FIXED_PARTITION_OFFSET(label) <= CONFIG_FLASH_LOAD_OFFSET &&                              \
 	 FIXED_PARTITION_OFFSET(label) + FIXED_PARTITION_SIZE(label) > CONFIG_FLASH_LOAD_OFFSET)
 
-#include <zephyr/devicetree.h>
 #if defined(CONFIG_TRUSTED_EXECUTION_NONSECURE) && (CONFIG_TFM_MCUBOOT_IMAGE_NUMBER == 2)
 #define UPLOAD_FLASH_AREA_LABEL slot1_ns_partition
 #else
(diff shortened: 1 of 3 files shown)


pdgendt (dismissed):

pdgendt (dismissed):

tomchy (changes requested):
This change introduces a regression - it is no longer possible to use a different partition for the code than `slot0/1_partition`.
In our setup - we define subpartitions within the `slot0/1_partition` for the `zephyr,code-partition`.

subsys/dfu/img_util/flash_img.c
```diff
@@ -23,11 +24,16 @@ LOG_MODULE_REGISTER(flash_img, CONFIG_IMG_MANAGER_LOG_LEVEL);
 #include <bootutil/bootutil_public.h>
 #endif
 
+#define FIXED_PARTITION_GET_FLASH_NODE(node_id)                                    \
```
JarmouniA, line 27: This is basically a combination of https://github.com/zephyrproject-rtos/zephyr/blob/8a72d776c5af6522478155adc91a5f4edb2794e3/include/zephyr/devicetree/fixed-partitions.h#L151
and
https://github.com/zephyrproject-rtos/zephyr/blob/8a72d776c5af6522478155adc91a5f4edb2794e3/include/zephyr/devicetree/fixed-partitions.h#L86
but without checking if Flash node has `soc_nv_flash` compatible.

nordicjm: it does not get the device either which is the important part so that it works in macros

de-nordic (approved):

pdgendt (approved):

tomchy (approved):

jakubtopic (approved):

subsys/dfu/img_util/flash_img.c
```diff
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/devicetree.h>
 #include <zephyr/dfu/mcuboot.h>
```
kartben, line 15: random (unrelated) nit: this header is included twice

subsys/dfu/img_util/flash_img.c
```diff
@@ -23,11 +24,16 @@ LOG_MODULE_REGISTER(flash_img, CONFIG_IMG_MANAGER_LOG_LEVEL);
 #include <bootutil/bootutil_public.h>
 #endif
 
+#define FIXED_PARTITION_GET_FLASH_NODE(node_id)                                    \
+	COND_CODE_1(DT_NODE_HAS_COMPAT(DT_PARENT(node_id), fixed_subpartitions),   \
+		    (DT_PARENT(DT_GPARENT(node_id))), (DT_GPARENT(node_id)))
```
JarmouniA, line 29: The 2nd part should not really be an unconditional else, should be checking for fixed_partitions compat.

The justification for the extra check would be that the description of the `zephyr,code-partition` chosen property does not say it only takes fixed-partitions or fixed-subpartitions compatible nodes.
https://docs.zephyrproject.org/latest/build/dts/api/api.html#zephyr-specific-chosen-nodes

subsys/dfu/img_util/flash_img.c
JarmouniA: > it does not get the device either

Not really sure what you mean.

PR #99796, opened 2025-11-21
tests: drivers: can: api: Validate CAN statistics

nordic-segl:
Extend CAN stats test. Add validation of returned data. Assume that current value of error counters is smaller than arbitrary threshold.

This is to detect bug in CAN driver implementation where CAN stats are not correctly initialized.

### DIFF
--- a/tests/drivers/can/api/src/stats.c
+++ b/tests/drivers/can/api/src/stats.c
@@ -9,6 +9,9 @@
 
 #include "common.h"
 
+/* Stats are cleared on every can_start/stop. */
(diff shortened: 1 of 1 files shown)


nordic-segl:
Extend CAN tests with scenario that could detect following issue:
https://github.com/zephyrproject-rtos/zephyr/pull/99259

henrikbrixandersen (changes requested):
These tests were left out for a reason as they depend on external events. There is no guarantee that these events are zero at this point in time.

PR #99798, opened 2025-11-21
drivers: pwm: mcux: enable when CPU is inactive

### LINKED ISSUE #95096: FlexPWM not working as intended
### Describe the bug

Hello, I'm trying to use the FlexPWM peripheral of the MCXA156 SoC in a FRDM_MCXA156 devboard but I can't get it to fully work.
Whatever I try, the pulse and period seen in the associated output pins do not correspond with the values I am assigning it.
I have found a way to reproduce the issue using the blinky_pwm example. Using the reproduction steps below I only get a small pulse of less than 1ms in pin P3_6 (where channel 0 of flexpwm0_pwm0 is linked) whenever the PWM is reconfigured, but nothing in between. AFAIK the example should be setting a 50% duty cycle signal and that is not what I am seeing.

### Regression

- [ ] This is a regression.
It is not a regression; I've tried the same thing on commit 9f6fdae4306473eeb0da4de2218c197ae8e574b4 (Where PWM was implemented for the FRDM_MCXA156 board) and the issue still happens.

### Steps to reproduce

1. `west init repro`
2. `cd repro && west update`
3. Copy the following overlay file to `samples/basic/blinky_pwm/boards/frdm_mcxa156.overlay`:
```
#include <zephyr/dt-bindings/pwm/pwm.h>

/ {
	pwmleds {
		compatible = "pwm-leds";
		pwm_led_0: pwm_led_0 {
			pwms = <&flexpwm0_pwm0 0 PWM_MSEC(1) PWM_POLARITY_NORMAL>;
			label = "PWM LED 0";
		};
	};

	aliases {
		pwm-led0 = &pwm_led_0;
		pwm-0 = &flexpwm0_pwm0;
	};
};
```
3. `cd zephyr && west build -b frdm_mcxa156 samples/basic/blinky_pwm`
4. `west flash`
5. See incorrect behavior when probing P3_6.

### Relevant log output

```shell
[00:00:00.000,069] <dbg> pwm_mcux: pwm_mcux_init: Set prescaler 128, reload mode 2
*** Booting Zephyr OS build v4.2.0-2444-g0d46a934171a ***
PWM-based blinky
Calibrating for channel 0...
[00:00:00.017,179] <err> pwm_mcux: Too long period (750000), adjust pwm prescaler!
[00:00:00.025,412] <err> pwm_mcux: Too long period (375000), adjust pwm prescaler!
[00:00:00.033,645] <err> pwm_mcux: Too long period (187500), adjust pwm prescaler!
[00:00:00.041,879] <err> pwm_mcux: Too long period (93750), adjust pwm prescaler!
Done calibrating; maximum/minimum periods 62500000/7812500 nsec
Using period 62500000
Using period 31250000
Using period 15625000
Using period 7812500
Using period 15625000
Using period 31250000
...
```

### Impact

### Environment

- Windows 11
- Zephyr SDK 0.17.2
- Zephyr commit 0d46a934171a11d0fe08ea2a615f79a92ac0ab9b

### Additional Context

_No response_

aperea-cc:
Set the undocumented WAITEN bit so that the PWM remains on even if the CPU goes into a sleep mode, which AFAIK is the default behavior for other PWM drivers in Zephyr.

Fixes [#95096](https://github.com/zephyrproject-rtos/zephyr/issues/95096).

The comment includes the GitHub issue because the WAITEN bit is not mentioned in the reference manual.

### DIFF
--- a/drivers/pwm/pwm_mcux.c
+++ b/drivers/pwm/pwm_mcux.c
@@ -242,6 +242,10 @@ static int pwm_mcux_init(const struct device *dev)
 		return -EIO;
 	}
 
+	/* Enable WAITEN bit to not disable FlexPWM when CPU is inactive */
+	/* see: https://github.com/zephyrproject-rtos/zephyr/issues/95096 */
+	config->base->SM[config->index].CTRL2 |= 1 << 14;
+
 	/* Disable fault sources */
 	for (i = 0; i < FSL_FEATURE_PWM_FAULT_CH_COUNT; i++) {
 		config->base->SM[config->index].DISMAP[i] = 0x0000;


drivers/pwm/pwm_mcux.c
```diff
@@ -242,6 +242,10 @@ static int pwm_mcux_init(const struct device *dev)
 		return -EIO;
 	}
 
+	/* Enable WAITEN bit to not disable FlexPWM when CPU is inactive */
+	/* see: https://github.com/zephyrproject-rtos/zephyr/issues/95096 */
+	config->base->SM[config->index].CTRL2 |= 1 << 14;
```
FelixWang47831, line 247: Actually Line 236 `pwm_config.enableWait = config->run_wait;` has the same effect.

Just because the macro  FSL_FEATURE_PWM_HAS_NO_WAITEN has been set as 1 in https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/devices/MCX/MCXA/MCXA156/MCXA156_features.h#L462 makes it can not be set.

I don't  think it's a good idea to hardcode set WAITEN here becuase we can't guarantee it's valid for all devices. The best way should be updating the macro in MCXA156_features.h.

aperea-cc: Wow, I didn't see line 236. It seems that was added 3 years ago.

My bad! I'll wait for a HAL fix then.

PR #99800, opened 2025-11-21
Verify memory controller node while determining a running application partition.

### LINKED ISSUE #99762: mcumgr: Image management incorrectly identifies active slot when slots are on different flash devices

tomchy:
Verify memory controller node while determining a running application partition.
Currently the check uses only offsets, which may give false results if there is a partition on the external memory with a matching offset range.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/99762

### DIFF
--- a/include/zephyr/storage/flash_map.h
+++ b/include/zephyr/storage/flash_map.h
@@ -19,7 +19,7 @@
  *
  * @defgroup flash_area_api flash area Interface
  * @since 1.11
- * @version 1.0.0
+ * @version 1.1.0
  * @ingroup storage_apis
  * @{
  */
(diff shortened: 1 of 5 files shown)


include/zephyr/storage/flash_map.h
```diff
@@ -465,6 +465,31 @@ uint8_t flash_area_erased_val(const struct flash_area *fa);
...
+ *
+ * @param label DTS node label of a partition
+ *
+ * @return Pointer to a device.
+ */
+#define FIXED_PARTITION_MTD(label) \
+	COND_CODE_1( \
+		DT_FIXED_SUBPARTITION_EXISTS(DT_NODELABEL(label)), \
+			(DT_MTD_FROM_FIXED_SUBPARTITION(DT_NODELABEL(label))), \
+			(DT_MTD_FROM_FIXED_PARTITION(DT_NODELABEL(label))))
```
JarmouniA, line 479:
```suggestion
			(DT_MTD_FROM_FIXED_PARTITION(DT_NODELABEL(label))))

```

JarmouniA (commented):
API minor version should be incremented.

nvlsianpu (approved):

masz-nordic (approved):

nordicjm (approved):

PR #99801, opened 2025-11-21
scripts: traceconfig: Improve file path display

yvesll:
For Windows, if the build dir is under same drive with zephyr base, the generated link shall be relative url like posix style. 
If they are in different drive, url is meaningless. So just use raw link format so that some editors can directly jump to.

I have found no way to render windows absolute path in markdown and let it work properly.

### DIFF
--- a/scripts/kconfig/traceconfig.py
+++ b/scripts/kconfig/traceconfig.py
@@ -6,6 +6,7 @@
 import argparse
 import os
 import pickle
+from pathlib import Path
 
 from tabulate import tabulate
 
@@ -18,12 +19,33 @@ KIND_TO_COL = {
 }
 
 
-def source_link(refpath, fn, ln):
-    if refpath:
-        fn = os.path.relpath(fn, refpath)
+def is_cross_drive(path1: str | os.PathLike, path2: str | os.PathLike) -> bool:
+    """Check if two paths are on different drives (Windows only)"""
 
-    link_fn = fn
-    disp_fn = os.path.normpath(fn).replace("../", "").replace("_", "\\_")
+    if os.name != 'nt':
+        return False
+
+    # handles "C:" and UNC like "\\SERVER\SHARE"
+    d1 = Path(path1).drive.upper()
+    d2 = Path(path2).drive.upper()
+
+    return bool(d1 and d2 and d1 != d2)
+
+
+def source_link(refpath, fn, ln, use_absolute):
+    full_path = Path(fn).absolute().as_posix().replace("_", "\\_")
+
+    if use_absolute:
+        return f"{full_path}:{ln}"
+
+    try:
+        fn = os.path.relpath(fn, refpath)
+    except ValueError:
+        # fallback to absolute path
+        return f"{full_path}:{ln}"
+
+    link_fn = Path(fn).as_posix()
+    disp_fn = link_fn.replace("../", "").replace("_", "\\_")
 
     return f"[{disp_fn}:{ln}](<{link_fn}#L{ln}>)"
 
@@ -31,12 +53,12 @@ def source_link(refpath, fn, ln):
 def write_markdown(trace_data, output):
     sections = {"user": [], "hidden": [], "unset": []}
 
-    if os.name == "nt":
-        # relative paths on Windows can't span drives, so don't use them
-        # generated links will be absolute
-        refpath = None
-    else:
-        refpath = os.path.dirname(os.path.abspath(output))
+    refpath = os.path.dirname(os.path.abspath(output))
+
+    # Check if we need to use absolute paths (cross-drive on Windows)
+    use_absolute = False
+    if os.name == 'nt' and is_cross_drive(refpath, __file__):
+        use_absolute = True
 
     for sym in trace_data:
         sym_name, sym_vis, sym_type, sym_value, sym_src, sym_loc = sym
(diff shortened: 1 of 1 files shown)


scripts/kconfig/traceconfig.py
```diff
@@ -18,25 +19,50 @@
...
-        # relative paths on Windows can't span drives, so don't use them
-        # generated links will be absolute
-        refpath = None
-    else:
-        refpath = os.path.dirname(os.path.abspath(output))
+    refpath = os.path.dirname(os.path.abspath(output))
+
+    # Check if we need to use absolute paths (cross-drive on Windows)
+    use_absolute = False
+    if os.name == 'nt' and is_cross_drive(refpath, __file__):
```
pillo79, line 60: This test will only detect if _this_ file (ie. Zephyr tree) is on a different drive compared to the output file (ie. build tree), but wouldn't catch app and module files, for example. This would need to be done in `source_link` for every single link.

scripts/kconfig/traceconfig.py
```diff
@@ -18,25 +19,50 @@
...
+    except Exception:
+        return False
+
+
+def source_link(refpath, fn, ln, use_absolute):
+    full_path = Path(fn).resolve().as_posix().replace("_", "\\_")
+
+    if use_absolute:
+        return f"{full_path}:{ln}"
+
```
pillo79, line 40: This can safely be removed, because you added a `try` that catches invalid relative path requests 😉. 
Removing this also removes the need for the `is_cross_drive` function and the extra argument.

pillo79 (commented):
Thank you, I do not usually use Windows so the patch is very much appreciated. 

However, it looks to me you tested the condition several times. I believe you can achieve the same result just wrapping the `relpath` call in the `try ... catch` and using the full_path in the exceptional case. Can you please test this approach as well?

scripts/kconfig/traceconfig.py
yvesll: After onetime calling `is_cross_drive`, the `use_absolute`'s value is certain. Move the logic into try catch can work but may cause performance issue under Windows as everytime it may perform the exception process flow.
IMO, `use_absolute` here is more readable.

scripts/kconfig/traceconfig.py
yvesll: Yes, you are correct. I did not consider out of tree apps/modules. Can we let the try-catch deal with such cases? Most definitions are in zephyr tree.

scripts/kconfig/traceconfig.py
```diff
@@ -18,25 +19,50 @@
...
+    try:
         fn = os.path.relpath(fn, refpath)
+    except ValueError:
+        # fallback to absolute path
+        return f"{full_path}:{ln}"
 
-    link_fn = fn
-    disp_fn = os.path.normpath(fn).replace("../", "").replace("_", "\\_")
+    link_fn = fn.replace("\\", "/")
+    disp_fn = os.path.normpath(fn).replace("\\", "/").replace("../", "").replace("_", "\\_")
```
yvesll: BTW, do you remember why escape the `_` here? If the `disp_fn` is `hello\_world/Kconfig`, it will not be shown in the preview.

pillo79: Not sure why it doesn't show on your parser (FWIW I tested with VSCode's preview), but MD can be tricky. The [Markdown Guide](https://www.markdownguide.org/basic-syntax/#characters-you-can-escape) says underscores can be escaped, and several places on the web mention escaping is needed to prevent the underscore from being treated as a special char (VSC required it).

scripts/kconfig/traceconfig.py
pillo79: I understand, and performance may be a valid concern on Windows. My real goal here was to remove the need for the `is_cross_drive` function.

scripts/kconfig/traceconfig.py
```diff
@@ -18,25 +19,50 @@
...
+    """Check if two paths are on different drives (Windows only)"""
+
+    if os.name != 'nt':
+        return False
+
+    try:
+        p1 = Path(path1).resolve()
+        p2 = Path(path2).resolve()
+        # On Windows, drive is the first part before ':'
+        drive1 = str(p1).split(':')[0] if ':' in str(p1) else None
```
pillo79: Is there no API for this? Could the path be a network folder in `\\server\share\path` notation, for example?

scripts/kconfig/traceconfig.py
```diff
@@ -18,25 +19,50 @@
...
+        drive2 = str(p2).split(':')[0] if ':' in str(p2) else None
+        return drive1 != drive2 and drive1 is not None and drive2 is not None
+    except Exception:
+        return False
```
pillo79: Wouldn't True be more appropriate here? In case the above test broke for some reason, I would assume it's safer to use absolute paths?

pillo79 (commented):
Sorry for the delay, lost track of this useful PR. 
I don't have a strong opinion here, so I was also looking forward to @tejlmand 's comments 🙂

scripts/kconfig/traceconfig.py
yvesll: As we only need to check the relationship between `ZEPHYR_BASE` and `output_dir`, I think one-time call for `is_cross_drive` may be valid.

scripts/kconfig/traceconfig.py
yvesll: Removed try catch as Path.drive will never raise an exception.

scripts/kconfig/traceconfig.py
yvesll: Yes, `Path.drive` is better

tejlmand (approved):

pillo79 (approved):

PR #99811, opened 2025-11-21
drivers: dai: add ability to use dai.h from user threads

kv2019i:
Add user-space support to the dai.h interface. No functional impact to builds when CONFIG_USERSPACE is not set.

### DIFF
--- a/drivers/dai/dai_handlers.c
+++ b/drivers/dai/dai_handlers.c
@@ -59,6 +59,18 @@ static inline int z_vrfy_dai_config_get(const struct device *dev,
 }
 #include <zephyr/syscalls/dai_config_get_mrsh.c>
 
+static inline int z_vrfy_dai_get_properties_copy(const struct device *dev,
+						enum dai_dir dir,
+						int stream_id,
+						struct dai_properties *dst)
+{
+	K_OOPS(K_SYSCALL_DRIVER_DAI(dev, get_properties_copy));
+	K_OOPS(K_SYSCALL_MEMORY_WRITE(dst, sizeof(*dst)));
+
+	return z_impl_dai_get_properties_copy(dev, dir, stream_id, dst);
+}
+#include <zephyr/syscalls/dai_get_properties_copy_mrsh.c>
+
 static inline int z_vrfy_dai_trigger(const struct device *dev,
 				     enum dai_dir dir,
 				     enum dai_trigger_cmd cmd)
--- a/drivers/dai/dai_handlers.c
+++ b/drivers/dai/dai_handlers.c
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2025 Intel Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/drivers/dai.h>
+#include <zephyr/internal/syscall_handler.h>
+
+/**
+ * Maximum size of bespoke objects passed to DAI driver.
+ * The objects get allocated temporarily on stack for validation,
+ * so size needs to be limited.
+ */
+#define DAI_MAX_BESPOKE_CFG_SIZE 256
+
+static inline int z_vrfy_dai_probe(const struct device *dev)
+{
+	K_OOPS(K_SYSCALL_DRIVER_DAI(dev, probe));
+
+	return z_impl_dai_probe(dev);
+}
+#include <zephyr/syscalls/dai_probe_mrsh.c>
+
+static inline int z_vrfy_dai_remove(const struct device *dev)
+{
+	K_OOPS(K_SYSCALL_DRIVER_DAI(dev, remove));
+
+	return z_impl_dai_remove(dev);
+}
+#include <zephyr/syscalls/dai_remove_mrsh.c>
+
+static inline int z_vrfy_dai_config_set(const struct device *dev,
+					const struct dai_config *cfg,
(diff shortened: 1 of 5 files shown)


drivers/dai/dai_handlers.c
```diff
@@ -0,0 +1,111 @@
...
+					const void *bespoke_cfg)
+{
+	K_OOPS(K_SYSCALL_DRIVER_DAI(dev, config_set));
+	K_OOPS(K_SYSCALL_MEMORY_READ(cfg, sizeof(*cfg)));
+
+	/* TODO: using CONFIG_MMU_PAGE_SIZE as a placeholder, need to
+	 * merge with Adrian's PR to add bespoke_cfg size */
+	K_OOPS(K_SYSCALL_MEMORY_READ(bespoke_cfg, CONFIG_MMU_PAGE_SIZE));
+
+	/* TODO: shall a kernel copy made of cfg */
```
lyakh: sorry, not sure what this comment means?

kv2019i: @lyakh So basicly, can we trust to pass an object that sits in user-space accessible memory to a kernel function. In Linux this is a no-no as another user-space thread could be active (on another CPU) and modify the data while kernel is accessing it (on another). 

On a resource constrained device, this is  a bit harder pill to swallow as some of these objects are rather big. But I'm thinking it we want to merge this in Zephyr, we need to do a copy of all input data to kernel, validate and run the function. In SOF side (the application using Zephyr), we could optimize this as we know there will not be other user-space threads (enforced by system design).

lyakh: @kv2019i ah, so the typo is "a kernel copy be made" - that's what I couldn't guess, sorry :-)

teburd: You need to deeply copy this (nothing referring to user mode accessible memory), then validate it. This will be rather troublesome with a void * at an API level.

kv2019i: Ack @teburd , will at deep copies to next version. @abonislawski  is added a size field in a separate PR, so that allows to handle the bespoke data objects in a sane fashion. I'll be adding a new variant of get_properties() that can be used from user-space.

kv2019i (commented):
Thanks all to reviewers so far!

kv2019i:
@lyakh @softwarecki @lgirdwood @nashif can you take a look?

lgirdwood (approved):

softwarecki (approved):
Good job wrapping `zephyr_syscall_header` in `if(CONFIG_DAI_USERSPACE)`.  
This was missing in https://github.com/thesofproject/sof/pull/10341, so it's great to see it addressed here.

fabiobaltieri (approved):

drivers/dai/intel/ssp/ssp.c
```diff
@@ -2548,6 +2548,20 @@ static const struct dai_properties *dai_ssp_get_properties(const struct device *
...
 
+static int dai_ssp_get_properties_copy(const struct device *dev,
+				enum dai_dir dir, int stream_id, struct dai_properties *prop)
+{
+	const struct dai_properties *kernel_prop = dai_ssp_get_properties(dev, dir, stream_id);
+
+	if (!prop) {
+		return -EINVAL;
+	}
+
```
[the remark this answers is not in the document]
kv2019i: Will solve in a follow-up PR.

include/zephyr/drivers/dai.h
```diff
@@ -442,6 +460,31 @@ static inline const struct dai_properties *dai_get_properties(const struct devic
...
 
+/**
+ * @brief Fetch properties of a DAI driver
+ *
+ * @param dev Pointer to the device structure for the driver instance
+ * @param dir Stream direction: RX or TX as defined by DAI_DIR_*
+ * @param stream_id Stream id: some drivers may have stream specific
+ *        properties, this id specifies the stream.
+ * @param dst address where to write properties to
+ * @retval Zero on success
```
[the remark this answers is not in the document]
kv2019i: Will solve in a follow-up PR.

include/zephyr/drivers/dai.h
```diff
@@ -442,6 +460,31 @@ static inline const struct dai_properties *dai_get_properties(const struct devic
...
+				      int stream_id,
+				      struct dai_properties *dst);
+
+static inline int z_impl_dai_get_properties_copy(const struct device *dev,
+						 enum dai_dir dir,
+						 int stream_id,
+						 struct dai_properties *dst)
+{
+	const struct dai_driver_api *api = (const struct dai_driver_api *)dev->api;
+
```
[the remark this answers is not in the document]
kv2019i: I'll implement this in a follow-up.

drivers/dai/dai_handlers.c
```diff
@@ -0,0 +1,145 @@
...
+					const void *bespoke_cfg,
+					size_t size)
+{
+	uint8_t bespoke_cfg_kernel[DAI_MAX_BESPOKE_CFG_SIZE];
+
+	if (size > DAI_MAX_BESPOKE_CFG_SIZE) {
+		return -EINVAL;
+	}
+
+	K_OOPS(K_SYSCALL_DRIVER_DAI(dev, config_set));
```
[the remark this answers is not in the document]
kv2019i: I'll fix this in a follow-up PR.

PR #99814, opened 2025-11-21
Fix POSIX APIs usage in network samples

jukkar:
Convert coap_download and coap_upload to use POSIX APIs just to be consistent with other samples in sockets directory.

Replace non-POSIX network symbols by POSIX ones in some of the network samples. Also replace POSIX symbols by Zephyr network APIs where applicable i.e., when POSIX is disabled.

We do not want to have network samples that mix both Zephyr network API with the relevant POSIX API like zsock_socket() and socket() etc.

Add also more tests to `tests/net/all` to catch more compile issues.


samples/net/sockets/coap_download/src/main.c
```diff
@@ -64,9 +64,9 @@ static void do_coap_download(struct sockaddr *sa)
 					      .user_data = NULL};
 
 	LOG_INF("Starting CoAP download using %s",
-		(NET_AF_INET == sa->sa_family) ? "IPv4" : "IPv6");
+		(AF_INET == sa->sa_family) ? "IPv4" : "IPv6");
 
-	sockfd = zsock_socket(sa->sa_family, SOCK_DGRAM, 0);
+	sockfd = socket(sa->sa_family, SOCK_DGRAM, 0);
```
rlubos, line 72: I think we're missing here POSIX includes for `socket()`/`close()`?

jukkar: yep, fixed now

pdgendt (approved):

maass-hamburg (approved):

rlubos (approved):

PR #49908, opened 2022-09-05
use dts for mmu mapping

JiafeiPan:
devicetree: add mmu mapping support

   Currently for ARM64, mmu mapping entries are hardcoded in soc's
   mmu_regions.c, and such mapping will be used for all board and
   applications although some mapping are not useful for some
   applications.

   So this patch provides another method to add mmu mapping entry
   by devicetree, so such each application can specify mmu mapping
   by devicetree overlay.

   Devicetree need to do the following update:
    1. Add property "zephyr,memory-region" to specify mmu entry name.
    2. Add property "zephyr,memory-region-mmu" to specify mmu entry
       attribute with string array which is valid c token from  architecture
        speficied MMU header file.
    3. The address and size will from "reg" node.

    For example:

            ccm: ccm@30380000 {
                    compatible = "nxp,imx-ccm";
                    reg = <0x30380000 DT_SIZE_K(64)>;
                    zephyr,memory-region = "CCM";
                    zephyr,memory-region-mmu = "MT_DEVICE_nGnRnE", "MT_P_RW_U_NA", "MT_NS";
                    #clock-cells = <3>;
           };

    And then in C code to include mmu entries specified in devicetree:
    For example in "mmu_regions.c" for ARM64 platform
    ...
    ...

    static const struct arm_mmu_region mmu_regions[] = {
            ...
            LINKER_DT_REGION_MMU(MMU_REGION_FLAT_ENTRY)
            ...
    };

With this dts example, the code generated will be:
```
   static const struct arm_mpu_region mpu_regions[] = {
       ...
       { .name = "CCM",
         .base = 0x30380000,
         .size = DT_SIZE_K(64),
         .attr = MT_DEVICE_nGnRnE | MT_P_RW_U_NA | MT_NS, },
       ...
   };
```

Issue #46477, opened 2022-06-13
About MMU mapping on ARM64


carlocaione:
### Facts
On ARM64 we can MMU-map a memory region in two different ways:
- Directly interfacing with the MMU code
- Going through the Zephyr MMU / device MMIO APIs.

### Direct interface with MMU code for direct mapping
This is done by the ARM64 MMU code to setup the basic Zephyr regions (text, data, etc..) in: https://github.com/zephyrproject-rtos/zephyr/blob/bfec3b2ab4cc19a204a6f390c2c3b176cb32a695/arch/arm64/core/mmu.c#L649-L679

but it also used by the soc-specific code to map regions for peripherals that do not support the device MMIO APIs, for example in: https://github.com/zephyrproject-rtos/zephyr/blob/bfec3b2ab4cc19a204a6f390c2c3b176cb32a695/soc/arm64/qemu_cortex_a53/mmu_regions.c#L11-L22

This mapping is done directly in the MMU driver code and it is usually a direct (1:1) mapping.

### Using the device MMIO (or MMU) APIs
There has been lately a certain effort to make the drivers using the device MMIO APIs. These API are leveraging the Zephyr MMU code to map the physical MMIO region of a peripheral to a virtual memory region automatically at init time (see `include/zephyr/sys/device_mmio.h`)

In general the mapping is not a direct mapping, but instead the virtual region is carved out from a memory pool of virtual addresses configured using `CONFIG_KERNEL_VM_BASE` and `CONFIG_KERNEL_VM_SIZE`.

### Problems
There are several.
1. The two methods are orthogonal, the only point of contact is the MMU driver that is actually doing the mapping.
2. The Zephyr MMU code is using a simple mechanism to keep tracking of the allocated pages that is being bypassed by the direct interface with the MMU code, so in theory could be conflicts.
3. Especially on ARM64 we have (theoretically) plenty of virtual memory so we really would like to do direct mapping for MMIO driver regions but this is not currently possible with the Zephyr MMU code.

### Solution?
The easiest one is to give up the direct interface and instead relying exclusively on the Zephyr MMU code. This would force us to give up the 1:1 mapping or adding support for that.

Tagging the main actors involved @dcpleung @npitre @povergoing

povergoing:
[...]

PR #99818, opened 2025-11-21
Add devicetree-based MMU region configuration

kedareswararao:
This pull request does the below
1) Enable automatic MMU configuration for ARM64 peripherals using devicetree
memory attributes. This eliminates the need to manually maintain region
lists in code when adding new peripherals.
2) Add test case for testing device-tree based memory attributes for ARM64 platform 

Tested changes on qemu(A53 and Versal Net) and Versalnet APU hardware


JiafeiPan (dismissed):
I once submitted one PR to achieve similar function, according to review comments, we still depend on device driver to take care of memory mapping.
https://github.com/zephyrproject-rtos/zephyr/pull/49908

Add it seems this PR need to add separate dts nodes but not reuse device's dts nodes' "reg" property, it is not convenient.

mbolivar (commented):
No objections to the DT changes.

kedareswararao:
> No objections to the DT changes.

Thanks @mbolivar if you are okay with the changes could you please approve

rruuaanng (dismissed):

kedareswararao:
@JiafeiPan @mbolivar @nashif : Please review when you have sometime

tests/arch/arm64/arm64_mmu_mem_attr/README.rst
```diff
@@ -0,0 +1,366 @@
...
+  preventing automatic scanning of all device nodes.
+* **Supports HAL use cases**: Allows MMU configuration for vendor HAL drivers that
+  cannot be modified to use Zephyr's Device MMIO API.
+
+Example DTS pattern:
+
+.. code-block:: devicetree
+
+   /* Separate memory-region node for MMU configuration */
+   uart0_mmu: memory-region@ff010000 {
```
pdgendt: Using the same unit address will trigger `dtc` warnings with `-Wunique_unit_address_if_enabled`. Slightly annoying, I think.

kedareswararao: @pdgendt : Thanks for the review. In the initial implementation, the memory attributes were defined directly within the peripheral nodes. Based on @JiafeiPan's feedback, the attribute handling has been separated into a dedicated node.

One alternative approach is to introduce a unique mmu-regions node that references peripherals via phandles, for example:

```dts
/ {
    mmu-regions {
        compatible = "zephyr,mmu-regions";
        #address-cells = <2>;
        #size-cells = <2>;
        
        uart0-mmu {
            device = <&uart0>;  /* phandle to get address/size */
            zephyr,memory-attr = <DT_MEM_ARM64_MMU_DEVICE>;
        };
    };
};
```

if this approach is okay will update the code changes accordingly, please let me know your thoughts

pdgendt: I don't know what the best approach would be right now, but the warnings about duplicate unit addresses are a bit worrying. On top, I recently added [a PR](https://github.com/zephyrproject-rtos/zephyr/pull/99120) to get a child node from the unit address, but this does not work if child nodes don't have unique addresses.

kedareswararao: @mbolivar @JiafeiPan @rruuaanng Could you please let me know if above suggested approach is fine or not to avoid the unique unit address issue

carlocaione (changes requested):
We have `MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY` already that it does almost this but not entirely.

What about something like this instead?

```c
#define MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(compat) \
      DT_FOREACH_STATUS_OKAY_VARGS(compat, MMU_REGION_DT_FLAT_ENTRY_FROM_DT)

  #define MMU_REGION_DT_FLAT_ENTRY_FROM_DT(node_id) \
      MMU_REGION_FLAT_ENTRY(DT_NODE_FULL_NAME(node_id), \
                            DT_REG_ADDR(node_id), \
                            DT_REG_SIZE(node_id), \
                            DT_MEM_ATTR_TO_MT(DT_PROP(node_id, zephyr_memory_attr))),

  static const struct arm_mmu_region mmu_regions[] = {
      MMU_REGION_FLAT_ENTRY("GIC", ...),
      MMU_REGION_FLAT_ENTRY("GIC", ...),
      /* Memory regions, auto from DT with per-node attributes */
      MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region)
  };
```
So we would:
- have the generation at compile time instead of run-time
- reuse what is already there
- one single line in each `mmu_regions.c`

This is all untested so I'm possibly making this easier than it actually is.

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -191,6 +192,80 @@ struct arm_mmu_ptables {
...
+ *       with multiple reg entries are not fully covered.
+ */
+#define MMU_REGION_DT_FLAT_ENTRY_FROM_DT(node_id)			\
+	COND_CODE_1(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),	\
+		(MMU_REGION_FLAT_ENTRY(DT_NODE_FULL_NAME(node_id),	\
+			DT_REG_ADDR(node_id),				\
+			DT_REG_SIZE(node_id),				\
+			DT_MEM_ATTR_TO_MT(				\
+				DT_PROP(node_id, zephyr_memory_attr))),)\
+		, ())
```
pdgendt: Nit: Use `IF_ENABLED` instead

```suggestion
#define MMU_REGION_DT_FLAT_ENTRY_FROM_DT(node_id)			\
	IF_ENABLED(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),	\
		(MMU_REGION_FLAT_ENTRY(DT_NODE_FULL_NAME(node_id),	\
			DT_REG_ADDR(node_id),				\
			DT_REG_SIZE(node_id),				\
			DT_MEM_ATTR_TO_MT(				\
				DT_PROP(node_id, zephyr_memory_attr))),))
```

soc/arm/qemu_virt_arm64/mmu_regions.c
```diff
@@ -19,6 +19,8 @@ static const struct arm_mmu_region mmu_regions[] = {
 			      DT_REG_ADDR_BY_IDX(DT_INST(0, arm_gic), 1),
 			      DT_REG_SIZE_BY_IDX(DT_INST(0, arm_gic), 1),
 			      MT_DEVICE_nGnRnE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE),
+
+	MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region)
```
carlocaione: at this point I see no reason for this to be in a SoC specific `mmu_regions.c`. Why not moving it directly into `arch/arm64/core/mmu.c` (after the `mmu_config.mmu_regions` loop)?

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -191,6 +192,80 @@ struct arm_mmu_ptables {
...
+		(MT_DEVICE_nGnRE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_DEVICE_GRE) ?		\
+		(MT_DEVICE_GRE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_NORMAL_NC) ?		\
+		(MT_NORMAL_NC | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_NORMAL) ?		\
+		(MT_NORMAL | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :		\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_NORMAL_WT) ?		\
+		(MT_NORMAL_WT | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+		0)
```
carlocaione: can we have a build error here instead of falling into `0`?

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -10,6 +10,7 @@
 #ifndef _ASMLANGUAGE
 #include <stdint.h>
 #include <stdlib.h>
+#include <zephyr/dt-bindings/memory-attr/memory-attr-arm64.h>
```
carlocaione: move this out of `_ASMLANGUAGE`

carlocaione (changes requested):

include/zephyr/arch/arm64/arm_mmu.h
kedareswararao: Fixed

soc/arm/qemu_virt_arm64/mmu_regions.c
kedareswararao: yes thanks for the suggestion fixed now

include/zephyr/arch/arm64/arm_mmu.h
kedareswararao: updated as suggested

include/zephyr/arch/arm64/arm_mmu.h
kedareswararao: Fixed

arch/arm64/core/mmu.c
```diff
@@ -891,6 +891,23 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
...
+	IF_ENABLED(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),	\
+		(BUILD_ASSERT(DT_MEM_ATTR_TO_MT(			\
+			DT_PROP(node_id, zephyr_memory_attr)) != 0,	\
+			"Unsupported zephyr,memory-attr on "		\
+			DT_NODE_PATH(node_id));))
+
+DT_FOREACH_STATUS_OKAY(zephyr_memory_region, _MMU_REGION_DT_VALIDATE_ATTR)
+
+static const struct arm_mmu_region mmu_dt_regions[] = {
+	MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region)
```
JiafeiPan, line 895: If all zephyr,memory-region nodes lack the property, this can collapse to { }, depending on toolchain/C standard interpretation, a deduced 0-element array can be problematic.

Suggest to add an empty at the end of the array:

```
static const struct arm_mmu_region mmu_dt_regions[] = {
	MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region),
    { 0 }
};
```
And then at [929](https://github.com/zephyrproject-rtos/zephyr/pull/99818/changes#diff-1330026b372d0de336b29d3e65a2f99d07d8859af1cd7dea54e21a3f331414aeR929)
`for (index = 0U; index < ARRAY_SIZE(mmu_dt_regions) - 1; index++) {`

JiafeiPan (commented):
Would you please update the document for how to use this feature?

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -191,6 +193,73 @@ struct arm_mmu_ptables {
...
+ *
+ * @param dt_attr The DT memory attribute value from zephyr,memory-attr.
+ */
+#define DT_MEM_ATTR_TO_MT(dt_attr)						\
+	((_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_DEVICE) ?		\
+		(MT_DEVICE_nGnRnE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_DEVICE_nGnRE) ?		\
+		(MT_DEVICE_nGnRE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+	 (_DT_MEM_ARM64_ARCH_BITS(dt_attr) == ATTR_MMU_DEVICE_GRE) ?		\
+		(MT_DEVICE_GRE | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
```
carlocaione: we really, REALLY, REALLY have to discourage people to use this to map device memory that should go through (hopefully) MMIO device API.

So we should remove entirely all the `MT_DEVICE_*` options or at very least error out at build time explaining to people that they are using the wrong tool for the job (mapping MMIO device memory).

arch/arm64/core/mmu.c
```diff
@@ -909,6 +926,14 @@ static void setup_page_tables(struct arm_mmu_ptables *ptables)
...
 	}
 
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
+	for (index = 0U; index < ARRAY_SIZE(mmu_dt_regions); index++) {
+		region = &mmu_dt_regions[index];
+		max_va = MAX(max_va, region->base_va + region->size);
+		max_pa = MAX(max_pa, region->base_pa + region->size);
+	}
+#endif
+
```
carlocaione: this and the next block are repeated in this function, factor them out.

carlocaione (changes requested):

include/zephyr/arch/arm64/arm_mmu.h
kedareswararao: @carlocaione   
I understand the concern about discouraging device memory mapping through this mechanism long-term.  However, today many  drivers in Zephyr access MMIO registers using static addresses (DT_REG_ADDR() / DT_INST_REG_ADDR()) without going through  DEVICE_MMIO_MAP().  
   
  On ARM64 with MMU enabled, those accesses will fault if the peripheral address range is not present in the page tables. Porting every driver to the dynamic MMIO API is a significant effort that cannot happen in a single patch series.  The DT  auto-mapping for device memory is intentionally provided as a pragmatic bridge: board DTS authors can declare their peripheral  regions with zephyr,memory-attr = <DT_MEM_ARM64_MMU_DEVICE> so that the static mappings are created at boot, and drivers 
    continue to work without modification.
 
  As individual drivers are ported to the MMIO device API, the corresponding DT entries can be removed.  Would it be acceptable  to keep the device memory support with a warning/comment in the header that it is intended as a transitional mechanism, rather than blocking it entirely at build time?

arch/arm64/core/mmu.c
kedareswararao: will factor the DT region iteration into the existing mmu_config.mmu_regions loops to avoid the duplication

arch/arm64/core/mmu.c
kedareswararao: sure will fix

include/zephyr/dt-bindings/memory-attr/memory-attr-arm64.h
```diff
@@ -0,0 +1,64 @@
...
+/** @brief DT value for Device GRE memory. */
+#define DT_MEM_ARM64_MMU_DEVICE_GRE	DT_MEM_ARM64(ATTR_MMU_DEVICE_GRE)
+/** @brief DT value for Normal non-cacheable memory. */
+#define DT_MEM_ARM64_MMU_NORMAL_NC	DT_MEM_ARM64(ATTR_MMU_NORMAL_NC)
+/** @brief DT value for Normal write-back cacheable memory. */
+#define DT_MEM_ARM64_MMU_NORMAL		DT_MEM_ARM64(ATTR_MMU_NORMAL)
+/** @brief DT value for Normal write-through cacheable memory. */
+#define DT_MEM_ARM64_MMU_NORMAL_WT	DT_MEM_ARM64(ATTR_MMU_NORMAL_WT)
+/** @brief DT value for unknown or unsupported memory type. */
+#define DT_MEM_ARM64_MMU_UNKNOWN	DT_MEM_ARCH_ATTR_UNKNOWN
```
mathieuchopstm, line 60: Nit: I find the double-shifting (1) very silly and (2) very hard to read.

Can't we simply do the following, even if it differs from other implementations?
```suggestion
/** @brief Device memory: nGnRnE (non-Gathering, non-Reordering, no Early write acknowledgment). */
#define DT_MEM_ARM64_MMU_DEVICE			BIT(20)
/** @brief Device memory: nGnRE (non-Gathering, non-Reordering, Early write acknowledgment). */
#define DT_MEM_ARM64_MMU_DEVICE_nGnRE		BIT(21)
/** @brief Device memory: GRE (Gathering, Reordering, Early write acknowledgment). */
#define DT_MEM_ARM64_MMU_DEVICE_GRE		BIT(22)
/** @brief Normal memory: non-cacheable. */
#define DT_MEM_ARM64_MMU_NORMAL_NC		BIT(23)
/** @brief Normal memory: write-back cacheable. */
#define DT_MEM_ARM64_MMU_NORMAL			BIT(24)
/** @brief Normal memory: write-through cacheable. */
#define DT_MEM_ARM64_MMU_NORMAL_WT		BIT(25)
```
(I removed the naked `ATTR_...` values since they should never be used)

include/zephyr/dt-bindings/memory-attr/memory-attr-arm64.h
```diff
@@ -0,0 +1,64 @@
...
+/** @brief Device memory: nGnRE (non-Gathering, non-Reordering, Early write acknowledgment). */
+#define ATTR_MMU_DEVICE_nGnRE		BIT(1)
+/** @brief Device memory: GRE (Gathering, Reordering, Early write acknowledgment). */
+#define ATTR_MMU_DEVICE_GRE		BIT(2)
+/** @brief Normal memory: non-cacheable. */
+#define ATTR_MMU_NORMAL_NC		BIT(3)
+/** @brief Normal memory: write-back cacheable. */
+#define ATTR_MMU_NORMAL			BIT(4)
+/** @brief Normal memory: write-through cacheable. */
+#define ATTR_MMU_NORMAL_WT		BIT(5)
```
mathieuchopstm: That's basically an enumeration like `memory-attr-arm.h` which is exactly what we do **NOT** want as memory attributes.

Take inspiration from `memory-attr-riscv.h` or `memory-attr-xtensa.h` instead, or even from https://github.com/zephyrproject-rtos/zephyr/issues/105467 if you dare.

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -191,6 +193,73 @@ struct arm_mmu_ptables {
 	DT_FOREACH_STATUS_OKAY_VARGS(compat, \
 	MMU_REGION_DT_FLAT_ENTRY, attr)
 
+/**
+ * @brief Extract ARM64 architecture bits from a DT memory attribute value.
+ *
+ * @param dt_attr The DT memory attribute value.
+ */
+#define _DT_MEM_ARM64_ARCH_BITS(dt_attr) \
+	(DT_MEM_ARM64_GET(dt_attr) >> DT_MEM_ARCH_ATTR_SHIFT)
```
mathieuchopstm:
```suggestion
#define _DT_MEM_ARM64_ARCH_BITS(dt_attr) \
	(DT_MEM_ARCH_ATTR_GET(dt_attr) >> DT_MEM_ARCH_ATTR_SHIFT)
```
(Really, this should be a generic macro...)

dts/bindings/base/zephyr,memory-common.yaml
```diff
@@ -32,6 +32,10 @@ properties:
       Attribute or set of attributes (bitmask) for the memory region. See
       'include/zephyr/dt-bindings/memory-attr/memory-attr.h' for a
       comprehensive list with description of possible values.
+      On ARM64, the architecture-specific bits are used by the
+      MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT() macro to
+      automatically generate static MMU region entries at compile time
+      for nodes with compatible = "zephyr,memory-region".
```
mathieuchopstm: The whole `zephyr,memory-region` feature is already underdocumented today, I would really appreciate if we stopped the bleeding and actually added stuff to https://docs.zephyrproject.org/latest/services/mem_mgmt/index.html (or dedicated arch-specific pages) instead of bindings no one will see.

It's also just not sustainable to add a section here for all archs, else the binding will end up with a 500 lines long property description.

mathieuchopstm (dismissed):
-1 for the enum style attributes - ARM32 is already enough of a mess due to that

include/zephyr/arch/arm64/arm_mmu.h
carlocaione: I very strongly disagree with this. I'm ok with this work being a tool to map generic memory regions leveraging `zephyr,memory-region` but using this work also for MMIO mapping is a big NO-NO for me.

I have several concerns about this when considering MMIO regions:
1. If we merged this we would have 3 different and overlapping mechanisms for memory mapping: (a) `MMU_REGION_FLAT_ENTRY` in `mmu_regions.c`, (b) the `MMIO` device API and (c) this new based on `DT`
2. It doesn't interact with the device `MMIO` API at all. It's a third mapping path, alongside `mmu_regions` and `device_map()`. Now there are three independent mechanisms.
3. The creation of a separate `zephyr,memory-region` node for each peripheral that needs mapping (with the same address / size nonetheless) is really cumbersome and not the right abstraction.

I understand that this is a problem, but we should find a way (and I'm trying to understand how to do this) for a definitive solution. I raised this problem back in 2022 (see https://github.com/zephyrproject-rtos/zephyr/issues/46477 ) and it is time to find a way to definitely fix this instead of patching here and there and introducing even more fragmentation.

Sorry, but I'm personally going to NACK this if we are going to support this also for MMIO regions.

kedareswararao: Got it updated to  keep the changes only to memory and for devices it's throughing build error as suggested

arch/arm64/core/mmu.c
```diff
@@ -891,23 +891,62 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
...
 
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
+
+#define _MMU_REGION_DT_VALIDATE_ATTR(node_id)				\
+	IF_ENABLED(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),	\
+		(BUILD_ASSERT(!_DT_MEM_ATTR_IS_DEVICE(			\
+			DT_PROP(node_id, zephyr_memory_attr)),		\
+			"Device memory on " DT_NODE_PATH(node_id)	\
+			" must not use DT auto-mapping"			\
+			" - use the MMIO device API instead");))
```
mathieuchopstm: If we don't want to allow device memory, then we shouldn't provide any DEVICE attribute in the first place.

arch/arm64/core/mmu.c
```diff
@@ -924,10 +963,13 @@ static void setup_page_tables(struct arm_mmu_ptables *ptables)
...
-		region = &mmu_config.mmu_regions[index];
-		add_arm_mmu_region(ptables, region, MT_NO_OVERWRITE);
-	}
+	map_mmu_regions(ptables, mmu_config.mmu_regions,
+			mmu_config.num_regions, MT_NO_OVERWRITE);
+
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
+	map_mmu_regions(ptables, mmu_dt_regions,
+			ARRAY_SIZE(mmu_dt_regions) - 1, MT_NO_OVERWRITE);
+#endif
```
mathieuchopstm: How about gating the `mmu_dt_regions` array behind `#if DT_HAS_COMPAT_STATUS_OKAY()` like done here, which avoids the dummy entry and `ARRAY_SIZE() - 1`?

tests/lib/shared_multi_heap/src/main.c
```diff
@@ -105,12 +106,12 @@ static void fill_multi_heap(void)
 		reg_map = &map[idx];
 
 		/* zephyr,memory-attr property not found. Skip it. */
-		if (reg_map->region.attr == DT_MEM_ARM_MPU_UNKNOWN) {
+		if (reg_map->region.attr == SMH_DT_MEM_UNKNOWN_DEFAULT) {
```
mathieuchopstm, line 89: how about just using `DT_MEM_ARCH_ATTR_UNKNOWN`?

mathieuchopstm (commented):
much better!

arch/arm64/core/mmu.c
kedareswararao: Good point — agreed. If we only support Normal memory through this mechanism, there's no reason to define an ATTR_ARM64_DEVICE 
bit just to BUILD_ASSERT against it. I'll remove ATTR_ARM64_DEVICE, the _DT_MEM_ATTR_IS_DEVICE() helper, and the device-memory 
BUILD_ASSERT entirely. The DT_MEM_ATTR_TO_MT() conversion only handles Normal memory types (via DT_MEM_CACHEABLE + 
ATTR_ARM64_CACHE_WB), so no guard is needed.

Updated code now

arch/arm64/core/mmu.c
kedareswararao: Updated code

tests/lib/shared_multi_heap/src/main.c
kedareswararao: Fixed

arch/arm64/core/mmu.c
```diff
@@ -891,23 +891,51 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
 	}
 }
 
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
```
carlocaione: Is this needed? I would expect `MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT` to simply return nothing is no such node exists.

arch/arm64/core/mmu.c
```diff
@@ -891,23 +891,51 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
...
+
+static const struct arm_mmu_region mmu_dt_regions[] = {
+	MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region)
+};
+
+#endif /* DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region) */
+
+static inline void max_region_bounds(const struct arm_mmu_region *regions,
+				     unsigned int count,
```
carlocaione: `size_t`? here and everywhere

arch/arm64/core/mmu.c
```diff
@@ -891,23 +891,51 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
...
 
-	for (index = 0U; index < mmu_config.num_regions; index++) {
-		region = &mmu_config.mmu_regions[index];
-		max_va = MAX(max_va, region->base_va + region->size);
-		max_pa = MAX(max_pa, region->base_pa + region->size);
-	}
+	max_region_bounds(mmu_config.mmu_regions, mmu_config.num_regions,
+			  &max_va, &max_pa);
+
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
```
carlocaione: is this needed?

arch/arm64/core/mmu.c
```diff
@@ -924,10 +952,13 @@ static void setup_page_tables(struct arm_mmu_ptables *ptables)
...
 	 * Those must not conflict with our default mapping.
 	 */
-	for (index = 0U; index < mmu_config.num_regions; index++) {
-		region = &mmu_config.mmu_regions[index];
-		add_arm_mmu_region(ptables, region, MT_NO_OVERWRITE);
-	}
+	map_mmu_regions(ptables, mmu_config.mmu_regions,
+			mmu_config.num_regions, MT_NO_OVERWRITE);
+
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_memory_region)
```
carlocaione: ditto

doc/hardware/arch/arm64.rst
```diff
@@ -0,0 +1,90 @@
+.. _arm64_developer_guide:
+
+Arm64 Developer Guide
```
carlocaione: ARM64

tests/lib/shared_multi_heap/src/main.c
```diff
@@ -67,34 +73,25 @@ static struct region_map *get_region_map(void *v_addr)
...
-	 * `RAM_NOCACHE` can be accessed and memory can be retrieved from using
-	 * the attribute `SMH_REG_ATTR_NON_CACHEABLE`.
-	 *
-	 * [MPU attr]   -> [SMH attr]
-	 *
-	 * RAM          -> SMH_REG_ATTR_CACHEABLE
-	 * RAM_NOCACHE  -> SMH_REG_ATTR_NON_CACHEABLE
-	 */
+#if defined(CONFIG_ARM64)
+	if (DT_MEM_ATTR_GET(dt_attr) & DT_MEM_CACHEABLE) {
```
carlocaione, line 63: Can we maybe introduce a generic `dt_mem_attr_is_cacheable()` function? likewise for all the generic attributes.

carlocaione (changes requested):

arch/arm64/core/mmu.c
kedareswararao: You're right — MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT simply expands to nothing when no such node exists, resulting in  an empty array. Removed all three #if guards.

arch/arm64/core/mmu.c
kedareswararao: Done — changed to size_t in both max_region_bounds() and map_mmu_regions(), including the loop variables.

arch/arm64/core/mmu.c
kedareswararao: Removed

arch/arm64/core/mmu.c
kedareswararao: Fixed

doc/hardware/arch/arm64.rst
kedareswararao: Fixed — changed to "ARM64 Developer Guide".

tests/lib/shared_multi_heap/src/main.c
kedareswararao: Good idea. I think this is better suited as a follow-up patch since it would be a subsystem-level addition to memory-attr.h and deserves its own review. I'll submit that separately.

tests/lib/shared_multi_heap/src/main.c
carlocaione, line 1: In this file now the block with `smh_reg_map` should now be removed, otherwise we are going do dynamically map the regions in parallel with the static allocation.

include/zephyr/arch/arm64/arm_mmu.h
```diff
@@ -191,6 +193,67 @@ struct arm_mmu_ptables {
...
+ * the MMIO device API instead.
+ *
+ * @param dt_attr The DT memory attribute value from zephyr,memory-attr.
+ */
+#define DT_MEM_ATTR_TO_MT(dt_attr)						\
+	((DT_MEM_ATTR_GET(dt_attr) & DT_MEM_CACHEABLE) ?			\
+		((_DT_MEM_ARM64_ARCH_BITS(dt_attr) & ATTR_ARM64_CACHE_WB) ?	\
+			(MT_NORMAL | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE) :	\
+			(MT_NORMAL_WT | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE)) : \
+		(MT_NORMAL_NC | MT_P_RW_U_NA | MT_DEFAULT_SECURE_STATE))
```
carlocaione, line 219: The only problem with this is that now the unsupported `zephyr,memory-attr` values are silently mapped as normal memory instead of being rejected.

Can we have something like:
```c
#define ARM64_MMU_VALIDATE_DT_REGION(node_id)                                  \
       IF_ENABLED(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),               \
               (BUILD_ASSERT(DT_MEM_ARM64_MMU_IS_VALID(                        \
                       DT_PROP(node_id, zephyr_memory_attr)),                  \
                       "Unsupported zephyr,memory-attr for ARM64 MMU region");))

#define DT_MEM_ARM64_MMU_IS_VALID(dt_attr) ...

DT_FOREACH_STATUS_OKAY(zephyr_memory_region, ARM64_MMU_VALIDATE_DT_REGION)
```

carlocaione (changes requested):
A couple of more comments but this is looking almost fine already.

tests/lib/shared_multi_heap/src/main.c
kedareswararao: Thanks for the review updated now

include/zephyr/arch/arm64/arm_mmu.h
kedareswararao: > The only problem with this is that now the unsupported `zephyr,memory-attr` values are silently mapped as normal memory instead of being rejected.
> 
> Can we have something like:
> 
> ```c
> #define ARM64_MMU_VALIDATE_DT_REGION(node_id)                                  \
>        IF_ENABLED(DT_NODE_HAS_PROP(node_id, zephyr_memory_attr),               \
>                (BUILD_ASSERT(DT_MEM_ARM64_MMU_IS_VALID(                        \
>                        DT_PROP(node_id, zephyr_memory_attr)),                  \
>                        "Unsupported zephyr,memory-attr for ARM64 MMU region");))
> 
> #define DT_MEM_ARM64_MMU_IS_VALID(dt_attr) ...
> 
> DT_FOREACH_STATUS_OKAY(zephyr_memory_region, ARM64_MMU_VALIDATE_DT_REGION)
> ```

Thanks for the review updated now

arch/arm64/core/mmu.c
```diff
@@ -891,23 +891,46 @@ static inline void add_arm_mmu_region(struct arm_mmu_ptables *ptables,
 	}
 }
 
+static const struct arm_mmu_region mmu_dt_regions[] = {
+	MMU_REGION_DT_COMPAT_FOREACH_FLAT_ENTRY_FROM_DT(zephyr_memory_region)
+};
+
+DT_FOREACH_STATUS_OKAY(zephyr_memory_region, ARM64_MMU_VALIDATE_DT_REGION)
```
carlocaione, line 898: nit: this should be in include/zephyr/arch/arm64/arm_mmu.h

carlocaione (approved):
This is nice, thanks for the effort.

wearyzen (approved):

rruuaanng (approved):

mbolivar (changes requested):
Putting a -1 since someone was asking for a merge on this. I want to review when the merge window opens again. Thanks.

mbolivar (approved):
Please don't define new `DT_` macros like this outside of `include/zephyr/devicetree*`. I see the namespace has gotten pretty messy and I'm working on cleaning it up, but in the meantime please don't make it worse.

I'm writing up some docs on how I usually review PRs like this, but the TL;DR for now is:

- the `DT_` namespace belongs to `<zephyr/devicetree.h>`

PR #99827, opened 2025-11-21
samples|tests: replace usage of legacy Mbed TLS crypto with PSA Crypto API

valeriosetti:
This PR replaces usage of legacy Mbed TLS crypto support in samples and tests. It continues the transition to PSA API only usage.


samples/net/lwm2m_client/overlay-dtls-cert.conf
```diff
@@ -1,38 +0,0 @@
-CONFIG_LWM2M_DTLS_SUPPORT=y
```
rlubos, line 1: It was probably an overlook this overlay was not mentioned in the readme or added to sample.yaml, the configuration still seem to cover an actual use-case, please don't remove it. CC @SeppoTakalo 

I can send a follow up to have this config covered in the readme and CI.

samples/net/sockets/big_http_download/src/big_http_download.c
```diff
@@ -11,7 +11,7 @@
 #include <ctype.h>
 #include <errno.h>
 
-#include "mbedtls/md.h"
+#include "psa/crypto.h"
```
tomi-font: Angle brackets rather?

tomi-font (dismissed):

rlubos (approved):

jukkar (approved):

Issue #99779, opened 2025-11-21
USBD: CDC ACM performance regression


JordanYates:
### Describe the bug

The CDC ACM class in the new USB next stack has significantly degraded performance compared to the legacy stack.
Measured write throughput on a nRF52840DK is approximately half, 1.8 Mbps vs 3.4 Mbps previously.

### Regression

- [x] This is a regression.

### Steps to reproduce

1. Cherry-pick the following commit: https://github.com/Embeint/zephyr/commit/ed075abad29c0c5ec0abe3cfc9294523f4f71009
2. Compile `zephyr/samples/usb/cdc_acm` and `zephyr/samples/usb/legacy/cdc_acm`
3. Run the `serial_test.py` script against both applications

### Relevant log output

```shell
# New stack
Read 131072 in 0.58 seconds (1793288 kbps)
Read 354304 in 1.58 seconds (1792186 kbps)
Read 578560 in 2.58 seconds (1792059 kbps)
Read 802816 in 3.58 seconds (1792067 kbps)
Read 1027072 in 4.58 seconds (1792077 kbps)
Read 1250304 in 5.58 seconds (1792059 kbps)
Read 1474560 in 6.58 seconds (1792048 kbps)

# Legacy stack
Read 153600 in 0.50 seconds (2465198 kbps)
Read 590848 in 1.50 seconds (3153578 kbps)
Read 1027072 in 2.50 seconds (3288777 kbps)
Read 1464320 in 3.50 seconds (3348318 kbps)
Read 1901568 in 4.50 seconds (3383012 kbps)
Read 2339840 in 5.50 seconds (3404404 kbps)
Read 2776064 in 6.50 seconds (3418291 kbps)
Read 3213312 in 7.50 seconds (3428161 kbps)
```

### Impact

Functional Limitation – Some features not working as expected, but system usable.

### Environment

- Zephyr v4.3.0

### Additional Context

Another issue with performance, but for MSC not CDC ACM: #93704

Possibly related to the double memcpy in the new driver, which doesn't appear to exist in the legacy driver.

https://github.com/zephyrproject-rtos/zephyr/blob/da9aa8fcad1a015afa7c07e19f0216a14eb44ae7/subsys/usb/device_next/class/usbd_cdc_acm.c#L784-L786

[...]

PR #99838, opened 2025-11-22
usb: device_next: class: cdc_acm: throughput improvements (low hanging fruit)

JordanYates:
Do not artificially limit the size of the USB request to the endpoint MPS, as this forces multiple round trips through the CDC ACM class implementation when the TX FIFO is larger than the endpoint MPS. The lower USB layers handles packetizing to the maximum endpoint size. The maximum allocation size is limited to prevent starving other USB users.

When a custom pool is used for CDC ACM, allow the buffer sizes to be configured in Kconfig.

Throughput testing on a nRF52840DK with `CONFIG_UDC_BUF_POOL_SIZE=1024` shows a throughput increase from ~1.7 Mbps to ~2.7 Mbps.

Steps towards improving the issue of #99779, but this is only the absolute lowest hanging fruit).
Throughput tested on the reproducer steps from the linked issue.
The limit of `CONFIG_UDC_BUF_POOL_SIZE / 4` was picked out of thin air as a reasonable default, happy to hear arguments against.

### DIFF
--- a/subsys/usb/device_next/class/Kconfig.cdc_acm
+++ b/subsys/usb/device_next/class/Kconfig.cdc_acm
@@ -37,6 +37,16 @@ config USBD_CDC_ACM_BUF_POOL
 	  ACM instances and the size of the bulk endpoints. When disabled, the
 	  implementation uses the UDC driver's pool.
 
+config USBD_CDC_ACM_BUF_POOL_SIZE
+	int "Size of the buffers in the dedicated buffer pool"
+	depends on USBD_CDC_ACM_BUF_POOL
+	default 512 if USBD_MAX_SPEED_HIGH
+	default 64 if USBD_MAX_SPEED_FULL
+	help
+	  Increasing the size of buffers past the endpoint MPS can reduce the
+	  number of round trips through the CDC ACM class when the TX FIFO is larger,
+	  than the endpoint MPS, increasing throughput.
+
 module = USBD_CDC_ACM
 module-str = usbd cdc_acm
 default-count = 1
--- a/subsys/usb/device_next/class/usbd_cdc_acm.c
+++ b/subsys/usb/device_next/class/usbd_cdc_acm.c
@@ -132,7 +132,8 @@ static void cdc_acm_irq_rx_enable(const struct device *dev);
 #if CONFIG_USBD_CDC_ACM_BUF_POOL
 UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
 		    DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) * 2,
-		    USBD_MAX_BULK_MPS, sizeof(struct udc_buf_info), NULL);
+		    CONFIG_USBD_CDC_ACM_BUF_POOL_SIZE, sizeof(struct udc_buf_info),
+		    NULL);
 
 static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data, const uint8_t ep,
 					 const size_t bytes_pending)
--- a/subsys/usb/device_next/class/usbd_cdc_acm.c
+++ b/subsys/usb/device_next/class/usbd_cdc_acm.c
@@ -134,10 +134,11 @@ UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
 		    DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) * 2,
 		    USBD_MAX_BULK_MPS, sizeof(struct udc_buf_info), NULL);
 
-static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
+static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data, const uint8_t ep,
+					 const size_t bytes_pending)
 {
 	ARG_UNUSED(c_data);
+	ARG_UNUSED(bytes_pending);
 	struct net_buf *buf = NULL;
 	struct udc_buf_info *bi;
 
@@ -156,10 +157,30 @@ static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
  * The required buffer is 128 bytes per instance on a full-speed device. Use
  * common (UDC) buffer, as this results in a smaller footprint.
  */
-static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
+static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data, const uint8_t ep,
+					 const size_t bytes_pending)
 {
-	return usbd_ep_buf_alloc(c_data, ep, USBD_MAX_BULK_MPS);
+	size_t alloc_size;
+
+	/* The lower layers handle packetizing to the endpoint MPS, therefore
+	 * allocating buffers largers than the endpoint MPS reduces the number
+	 * of round trips through this class driver when the TX FIFO size is
+	 * larger than the endpoint MPS, improving throughput.
+	 */
+	if (bytes_pending <= USBD_MAX_BULK_MPS) {
+		alloc_size = bytes_pending;
+	} else {
+		/* Limit the maximum allocation size to 1/4 of the total pool size
+		 * to prevent starving other USB users.
+		 */
+		alloc_size = min(bytes_pending, CONFIG_UDC_BUF_POOL_SIZE / 4);
+		/* Round down to a multiple of USBD_MAX_BULK_MPS so only the very
+		 * last packet is not maximally sized.
+		 */
+		alloc_size = ROUND_DOWN(alloc_size, USBD_MAX_BULK_MPS);
+	}
+
+	return usbd_ep_buf_alloc(c_data, ep, alloc_size);
 }
 #endif /* CONFIG_USBD_CDC_ACM_BUF_POOL */
 
@@ -651,7 +672,8 @@ static void cdc_acm_tx_fifo_handler(struct k_work *work)
 		return;
 	}
 
-	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_in(c_data));
+	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_in(c_data),
+				ring_buf_size_get(data->tx_fifo.rb));
 	if (buf == NULL) {
 		atomic_clear_bit(&data->state, CDC_ACM_TX_FIFO_BUSY);
 		cdc_acm_work_schedule(&data->tx_fifo_work, K_MSEC(1));
@@ -707,7 +729,7 @@ static void cdc_acm_rx_fifo_handler(struct k_work *work)
 		return;
 	}
 
-	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data));
+	buf = cdc_acm_buf_alloc(c_data, cdc_acm_get_bulk_out(c_data), USBD_MAX_BULK_MPS);
 	if (buf == NULL) {
 		return;
 	}


subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -135,9 +135,10 @@ UDC_BUF_POOL_DEFINE(cdc_acm_ep_pool,
 		    USBD_MAX_BULK_MPS, sizeof(struct udc_buf_info), NULL);
 
 static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
+					 const uint8_t ep, size_t size_hint)
```
jfischer-no: Please add const and name it just size.
```c
static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
					 const uint8_t ep, const size_t size)
```

subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -157,9 +158,21 @@ static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
  * common (UDC) buffer, as this results in a smaller footprint.
  */
 static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
-					 const uint8_t ep)
+					 const uint8_t ep, size_t size_hint)
```
jfischer-no:
```c
static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
					 const uint8_t ep, const size_t size)
```

subsys/usb/device_next/class/usbd_cdc_acm.c
```diff
@@ -157,9 +158,21 @@ static struct net_buf *cdc_acm_buf_alloc(struct usbd_class_data *const c_data,
...
-	return usbd_ep_buf_alloc(c_data, ep, USBD_MAX_BULK_MPS);
+	/* The USB stack handles packetizing to the endpoint size, therefore
+	 * allocating buffers largers than the endpoint size reduces the number
+	 * of round trips through this class driver when the TX FIFO size is
+	 * larger than the endpoint, improving throughput.
+	 *
+	 * Limit the maximum allocation size to 1/4 of the total pool size
+	 * to prevent starving other USB users. Ensure the maximum allocation
+	 * never drops below USBD_MAX_BULK_MPS.
+	 */
```
jfischer-no: >	/* The USB stack handles packetizing to the endpoint size, therefore

"USB stack" does nothing like that. That is the UDC driver (and sometimes controller) who handles a transfer, transaction by transaction. Please change this comment to be just:

```c
	/*
	 * Limit the maximum allocation size to 1/4 of the UDC pool size
	 * to prevent starving other pool users. Ensure the maximum allocation
	 * never drops below USBD_MAX_BULK_MPS.
	 */
```
Please fix your commit message as well.

subsys/usb/device_next/class/Kconfig.cdc_acm
```diff
@@ -37,6 +37,16 @@ config USBD_CDC_ACM_BUF_POOL
 	  ACM instances and the size of the bulk endpoints. When disabled, the
 	  implementation uses the UDC driver's pool.
 
+config USBD_CDC_ACM_BUF_POOL_SIZE
+	int "Size of the buffers in the dedicated buffer pool"
+	depends on USBD_CDC_ACM_BUF_POOL
+	default 512 if USBD_MAX_SPEED_HIGH
+	default 64 if USBD_MAX_SPEED_FULL
```
jfischer-no, line 44: For the full speed, it defaults to the a size that is worse than using UDC pool. Perhaps the default should always be 512. And there should be a `range 512 $(UINT16_MAX)` below `default 512` and
```c
BUILD_ASSERT((CONFIG_USBD_CDC_ACM_BUF_POOL_SIZE % USBD_MAX_BULK_MPS) == 0, \
	     "USBD_CDC_ACM_BUF_POOL_SIZE is not multiple of bulk endpoint MPS");
```
in usbd_cdc_acm.c.

subsys/usb/device_next/class/Kconfig.cdc_acm
```diff
@@ -37,6 +37,16 @@ config USBD_CDC_ACM_BUF_POOL
...
+	help
+	  Increasing the size of buffers past the endpoint size can reduce the
```
jfischer-no: Please use endpoint MPS (maximum packet size) and change it to
```
	  Increasing the size of buffers to a multiple of endpoint MPS can reduce
	  the number of round trips through the CDC ACM class when the TX FIFO is
	  larger than the endpoint MPS, increasing throughput.
```

jfischer-no (changes requested):

subsys/usb/device_next/class/Kconfig.cdc_acm
JordanYates: My intent here was not to change any behaviors, just make the values configurable from Kconfig. The pool was previously created with a size of `USBD_MAX_BULK_MPS`, defined as:
```
/* 1 if USB device stack is compiled with High-Speed support */
#define USBD_SUPPORTS_HIGH_SPEED IS_EQ(CONFIG_USBD_MAX_SPEED, 1)

/* Maximum bulk max packet size the stack supports */
#define USBD_MAX_BULK_MPS COND_CODE_1(USBD_SUPPORTS_HIGH_SPEED, (512), (64))
```
Unless I am misunderstanding something.

If the current values are not optimal, that can be a future PR IMO.

JordanYates: I missed this originally, but why is there any requirement for the pool to be a multiple of the MPS size? The driver makes no attempt to ensure that only buffers with a multiple of MPS bytes are submitted.

subsys/usb/device_next/class/usbd_cdc_acm.c
JordanYates: Added `const`, left the name as `size_hint` as that is what it is. A hint of how much data is pending to be sent, allowing this function to allocate an appropriately sized buffer (through the additional modifications in the function).

subsys/usb/device_next/class/usbd_cdc_acm.c
JordanYates: Removed "USB stack" references. Left the higher reasoning, since it was not obvious to me as someone looking at the driver for the first time that the UDC request sizes don't need to match the MPS.

subsys/usb/device_next/class/usbd_cdc_acm.c
jfischer-no: Please name it just size.

subsys/usb/device_next/class/usbd_cdc_acm.c
jfischer-no: Your higher reasoning is redundant. There is not need to explain how the UDC driver works. Remove it as suggested.

subsys/usb/device_next/class/Kconfig.cdc_acm
jfischer-no: With your changes, buffers may be of arbitrary size. Using a buffer smaller than the MPS size does not make sense. Using a buffer that is not a multiple of MPS may result in all transfers ending with a short transaction. Depending on the host software, this may or may not cause issues, but it is still less efficient.

Please fix it as suggested in https://github.com/zephyrproject-rtos/zephyr/pull/99838#discussion_r2561838264.

subsys/usb/device_next/class/usbd_cdc_acm.c
JordanYates: `size` is wrong. The parameter is not the buffer size to allocate. The parameter is how much data is pending to transmit, which can be used to determine an appropriate size to allocate. If you want it called something like `pending_data`, I'm happy to oblige, I'm just not going to rename it to something misleading.

subsys/usb/device_next/class/usbd_cdc_acm.c
JordanYates: Why is it redundant? It's explaining why the function is allocating buffer sizes other than MPS. That seems like an important point for a function called `cdc_acm_buf_alloc`.

subsys/usb/device_next/class/Kconfig.cdc_acm
JordanYates: > With your changes, buffers may be of arbitrary size. Using a buffer smaller than the MPS size does not make sense. Using a buffer that is not a multiple of MPS may result in all transfers ending with a short transaction. Depending on the host software, this may or may not cause issues

Right now in-tree the code already sends packets that are less than MPS size, if there are fewer than MPS bytes pending. There is no restriction that can be placed on the buffer size that will result in the host software not needing to handle short transfers.

> but it is still less efficient.

Less efficient is not a reason to throw a compile time error. The defaults are set up for the maximum efficiency, if someone explicitly wants to change the value to `537`, I don't see why it should be blocked.

tmon-nordic: For maximum throughput, following heuristic generally works fine:
- if there's less than wMaxPacketSize byte - send that number of bytes
- if there's more than wMaxPacketSize byte, send ROUND_DOWN(available, wMaxPacketSize).
- for precise wMaxPacketSize multiples the easiest option is to set ZLP flag. A more complex scheme would involve enqueuing first without ZLP flag (anticipating that before the transfer is completed there are already new bytes available to transmit), but then the device has to schedule ZLP if no new bytes are available.

Reasoning is that overhead is roughly the same for every packet (regardless of how many bytes are within the packet), and sending as many wMaxPacketSize packets as possible is the key to good throughput. ZLPs are necessary because hosts may be issuing URBs for more than wMaxPacketSize. For example Linux uses 2 * wMaxPacketSize unless SINGLE_RX_URB quirk is set for device. The SINGLE_RX_URB quirk is precisely for devices that do not send ZLPs after the last transmitted IN packet is wMaxPacketSize multiple and there is no more data avilable - I believe Zephyr should be nice citizen so it should send ZLPs.

JordanYates: > * if there's less than wMaxPacketSize byte - send that number of bytes
> * if there's more than wMaxPacketSize byte, send ROUND_DOWN(available, wMaxPacketSize).

That's the current behavior, which is partly to blame for the performance regression in the first place. I don't believe the backend sends any types/lengths of packets that it wasn't before as a result of this PR. Are you suggesting to leave the behavior as is, or am I misunderstanding your comment?

I have no comment about ZLPs, this PR is the extent of my knowledge about the USB stack. Sounds like an issue for another PR.

tmon-nordic: This is not the current behavior. Current behavior is to send up to `USBD_MAX_BULK_MPS` which is clearly different from what `ROUND_DOWN(available, wMaxPacketSize)` is.

For example if there are `5 * wMaxPacketSize + 13` bytes available, then current behavior (when operating at maximum compiled in operating speed) is to enqueue `USBD_MAX_BULK_MPS` at least 5 times. With the `ROUND_DOWN` it would be to queue `5 * wMaxPacketSize` and then enqueue the remainder. The delay in enqueuing the remainder is the anticipation that while the `5 * wMaxPacketSize` is being transmitted that there may be more data that will become available.

JordanYates: Ok, I read `wMaxPacketSize` as being equivalent to `BULK_MPS`, it wasn't mentioned anywhere what it was supposed to be. Especially since you mentioned `2 * wMaxPacketSize` for Linux.

JordanYates: Updated to align with this logic

tmon-nordic: > Ok, I read `wMaxPacketSize` as being equivalent to `BULK_MPS`, it wasn't mentioned anywhere what it was supposed to be. Especially since you mentioned `2 * wMaxPacketSize` for Linux.

`wMaxPacketSize` refers to the active speed bulk `wMaxPacketSize`. `USBD_MAX_BULK_MPS` refers to the maximum compiled in bulk `wMaxPacketSize`, but High-Speed capable device can be connected to Full-Speed only host (or otherwise forced to operate at Full-Speed only).

The thing is that `USBD_MAX_BULK_MPS` may but not necessarily be equal to `wMaxPacketSize`. For example if there 192 bytes available, then for High-Speed capable device this may either be a short packet or a multiple of `wMaxPacketSize` - the only way to know is for device to check its active speed.

subsys/usb/device_next/class/usbd_cdc_acm.c
jfischer-no: I want to call it size because it is size of the buffer to allocate and the size of the pending data. Allocated size should not be less than requested by the caller. You need to fix it or I will not accept this changes.

subsys/usb/device_next/class/usbd_cdc_acm.c
jfischer-no: You comment is redundant because it is common behavior of the driver/controller in Zephyr.

subsys/usb/device_next/class/usbd_cdc_acm.c
JordanYates: > I want to call it size because it is size of the buffer to allocate and the size of the pending data. 

That doesn't make any sense, the two are not the same thing. The trivial case that proves the point is how many bytes should be allocated if there are 1024 bytes pending and `CONFIG_UDC_BUF_POOL_SIZE=512`?

> Allocated size should not be less than requested by the caller

Right now, in tree, this function allocates less data than is pending. It was even written by you (8457f502). Why is that suddenly incorrect? This is a completely different request than the rest of the comment chain.

> You need to fix it or I will not accept this changes.

The variable name or this new comment?

subsys/usb/device_next/class/Kconfig.cdc_acm
JordanYates: > `wMaxPacketSize` refers to the active speed bulk `wMaxPacketSize`.

This doesn't make anything clearer to me, but it doesn't matter.

> The thing is that USBD_MAX_BULK_MPS may but not necessarily be equal to wMaxPacketSize. For example if there 192 bytes available, then for High-Speed capable device this may either be a short packet or a multiple of wMaxPacketSize - the only way to know is for device to check its active speed.

Ok, I don't know how to turn that information into code though.

If the current implementation is wrong or suboptimal, can you please suggest logic that is better?

tmon-nordic: > Ok, I don't know how to turn that information into code though.

Call `cdc_acm_get_bulk_mps()` instead of using `USBD_MAX_BULK_MPS`.

But then the `data->zlp_needed` assignment would have to consider not only the current transfer size, but also whether there is more data pending (if there is more data pending, then it should set `zlp_needed` to false even if sending `wMaxPacketSize` multiple).

Issue #40378, opened 2021-11-16
Bluetooth: Controller: Use defines for access_addr and crc_init array sizes


cvinayak:
**Is your enhancement proposal related to a problem? Please describe.**
`access_addr` always seems to be of size 4; that should be #defined.
`crc_init` always seem to be of size 3; that should be #defined.

https://github.com/zephyrproject-rtos/zephyr/pull/31009/files#r736229677
https://github.com/zephyrproject-rtos/zephyr/pull/31009/files#r736229386

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context or graphics (drag-and-drop an image) about the feature request here.

wolterhv:
Aren't these already defined in `subsys/bluetooth/controller/ll_sw/pdu.h` as `PDU_ACCESS_ADDR_SIZE` and `PDU_CRC_SIZE`? 

Excerpt of `pdu.h`:

```c
/*
 * PDU fields sizes
 */

#define PDU_PREAMBLE_SIZE(phy) (phy&0x3)
#define PDU_ACCESS_ADDR_SIZE   4
#define PDU_HEADER_SIZE        2
#define PDU_MIC_SIZE           4
#define PDU_CRC_SIZE           3
#define PDU_OVERHEAD_SIZE(phy) (PDU_PREAMBLE_SIZE(phy) + \
                                PDU_ACCESS_ADDR_SIZE + \
                                PDU_HEADER_SIZE + \
                                PDU_CRC_SIZE)
```

If so, I'm happy to take care of this one as my first issue.

bradfordarroyo:
Great

PR #99839, opened 2025-11-22
Bluetooth: Controller: Use macros for access address and CRC init sizes

ssam18:
## Description

This PR addresses issue #40378 by replacing hardcoded array size literals with symbolic constants for  and  arrays in the Bluetooth controller code.

## Changes

1. **New define added** ():
   -  (3 bytes) - Specifically for CRC initialization values

2. **Updated files**:
   -  - 6 instances
   -  - 6 instances


cvinayak (commented):
LGTM. Nitpick, reorder variable definitions in the inverse order of their line length (reverse christmas tree).

cvinayak (changes requested):
Seems you have randomly shuffled the lines.  See how LHS has the declarations ordered by line length before your change.

cvinayak (dismissed):
Comments need to be addressed.

PR #99840, opened 2025-11-22
scripts: runners: ecpprog: add -k/--divider argument

josuah:
Some minor improvement to the `ecpprog` runner, used to program Lattice ECP5 and CrossLink FPGAs devkit (with FTDI).

Add a `-k`/`--divider` slowing down the clock.


pdgendt (commented):
It might be more useful to add the `tool_opt` capability so users are capable of passing any arguments they need?

PR #99855, opened 2025-11-23
drivers: can: sja1000: consolidate SFF/EFF payload loops

ndrs-pst:
Utilize `data_reg` to select the appropriate SFF/EFF data register base and use a single read/write loop for frame data access.
This replaces the separate SFF/EFF loops, streamlining the code and reducing its footprint.

### DIFF
--- a/drivers/can/can_sja1000.c
+++ b/drivers/can/can_sja1000.c
@@ -275,7 +275,7 @@ static void can_sja1000_read_frame(const struct device *dev, struct can_frame *f
 {
 	uint32_t id;
 	uint8_t info;
-	int i;
+	uint8_t data_reg;
 
 	memset(frame, 0, sizeof(*frame));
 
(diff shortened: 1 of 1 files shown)


drivers/can/can_sja1000.c
```diff
@@ -275,6 +275,7 @@ static void can_sja1000_read_frame(const struct device *dev, struct can_frame *f
 {
 	uint32_t id;
 	uint8_t info;
+	uint8_t base_reg;
```
henrikbrixandersen: Could we name this `data_reg` or similar? The meaning of `base` is a bit overloaded. Same for the other instance.

drivers/can/can_sja1000.c
```diff
@@ -303,25 +304,19 @@ static void can_sja1000_read_frame(const struct device *dev, struct can_frame *f
...
+		base_reg = CAN_SJA1000_SFF_DATA;
+	}
 
-		if ((frame->flags & CAN_FRAME_RTR) == 0U) {
-			for (i = 0; i < frame->dlc; i++) {
-				frame->data[i] = can_sja1000_read_reg(dev, CAN_SJA1000_SFF_DATA +
-								      i);
-			}
+	if ((frame->flags & CAN_FRAME_RTR) == 0U) {
+		for (i = 0; i < frame->dlc; i++) {
```
henrikbrixandersen: Please follow the recommendation from SonarCloud and move the `int i` declaration into the loop since it is now only used there. Same for the other instance.

henrikbrixandersen (changes requested):
Nice optimization, thanks. A few minor comments below:

drivers/can/can_sja1000.c
ndrs-pst: OK, address it. Since there are other places where `int i` is declared outside, I will move it inside the loop as well.

drivers/can/can_sja1000.c
ndrs-pst: OK, I’ll rename it to `data_reg`.

henrikbrixandersen (approved):

martinjaeger (approved):

PR #99860, opened 2025-11-23
drivers: spi: siwx91x: SPI driver fixes to enable higher burst rates 

smalae:
1. Setting INTF PLL to 160 MHz so GSPI has a stable, high-rate base clock, and QSPI gets the expected 80 MHz. 
2. Rounding the GSPI frequency down to a valid step derived from the INTF PLL (f = PLL/(2*div)), and logging the requested vs programmed rate. 
3. Enabling higher GPDMA burst sizes (1/2/4) when safe (SPI ≥ 10 MHz and both channels use GPDMA). The burst size is selected from the largest value that satisfies TX/RX buffer alignment, and FIFO thresholds are set accordingly. A temporary limitation notes DMA flow-control issues at low SPI rates, so bursts are gated behind the 10 MHz threshold. 
4. Align mosi_overrun to 32 bytes
5. Fixing 16-bit transfers by swapping write/read data paths in the GSPI controller (GSPI_CONFIG2_b), ensuring the correct signal presentation on the SPI lines. 
6. Guarding async callback registration to avoid build errors when using blocking DMA transfers without CONFIG_SPI_ASYNC. 
7. Updating tests: increase CONFIG_SPI_IDEAL_TRANSFER_DURATION_SCALING from 24 → 26 to account for the (slightly) increased transfer time due to burst-size calculation.
8.  Enable the spi_loopback test for the PSRAM board siwx917_rb4342a.


drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -48,7 +49,7 @@ struct gspi_siwx91x_config {
 	const struct device *clock_dev;
 	clock_control_subsys_t clock_subsys;
 	const struct pinctrl_dev_config *pcfg;
-	uint8_t mosi_overrun __aligned(4);
+	uint8_t mosi_overrun __aligned(32);
```
tbursztyka, line 52: Why this change?

drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -76,13 +76,35 @@ static bool spi_siwx91x_is_dma_enabled_instance(const struct device *dev)
 #endif
 }
 
+void gspi_siwx91x_pick_lower_freq(uint32_t clock_hz, uint32_t requested_hz, uint32_t *actual_hz_out,
+				  uint32_t *div_out)
+{
+	/* Calculate divider that ensures freq <= requested */
+	uint32_t divider = (clock_hz + (2U * requested_hz) - 1U) / (2U * requested_hz);
```
jerome-pouiller: ```uint32_t divider = DIV_ROUND_UP(clock_hz, 2 * requested_hz);```

drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -76,13 +76,35 @@ static bool spi_siwx91x_is_dma_enabled_instance(const struct device *dev)
...
+	/* Compute the actual achievable frequency */
+	uint32_t actual_hz = clock_hz / (2U * divider);
+
+	if (divider == 0U) {
```
jerome-pouiller, line 87: If `divider == 0`, `actual_hz = clock_hz / (2U * divider)` has already raised a fault.

drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -423,6 +428,52 @@ static size_t gspi_siwx91x_longest_transfer_size(struct spi_context *instance_ct
...
+	if (gspi_all_aligned(ctx, 4)) {
+		return 4;
+	}
+
+	if (gspi_all_aligned(ctx, 2)) {
+		return 2;
+	}
+
+	return 1;
+}
```
jerome-pouiller, line 463: I believe this function is more complex than required. I believe we can get the same result with:
```
static int gspi_siwx91x_burst_size_buf(const struct spi_buf *spi_buf)
{
	int burst_len = 4;

	if (!ctx->current_tx[i].len || !ctx->current_tx[i].buf) {
		return burst_len;
	}
	burst_len = MIN(burst_len, BIT(find_lsb_set(spi_buf->buf) - 1));
	burst_len = MIN(burst_len, BIT(find_lsb_set(spi_buf->len) - 1));
	return burst_len;
}

static int gspi_siwx91x_burst_size(struct spi_context *ctx)
{
	int burst_len = 4;

	for (int i = 0; i < ctx->tx_count; i++) {
		burst_len = MIN(burst_len, gspi_siwx91x_burst_size_buf(ctx->current_tx + i));
	}
	for (int i = 0; i < ctx->rx_count; i++) {
		burst_len = MIN(burst_len, gspi_siwx91x_burst_size_buf(ctx->current_rx + i));
	}
	return burst_len;
}
```

tests/drivers/spi/spi_loopback/testcase.yaml
```diff
@@ -360,6 +360,12 @@ tests:
     filter: CONFIG_SOC_FAMILY_SILABS_SIWX91X
     platform_allow:
       - siwx917_rb4338a
+      - siwx917_rb4342a
+    extra_configs:
+      - CONFIG_SPI_SILABS_SIWX91X_GSPI_DMA=y
+      - CONFIG_SPI_IDEAL_TRANSFER_DURATION_SCALING=26
+    extra_args:
+      - DTC_OVERLAY_FILE="boards/siwx917_rb4338a.overlay"
```
jerome-pouiller: I suggest to introduce `boards/siwx917_rb4342a.overlay` rather than using `extra_args:`

tests/drivers/spi/spi_loopback/testcase.yaml
```diff
@@ -360,6 +360,12 @@ tests:
     filter: CONFIG_SOC_FAMILY_SILABS_SIWX91X
     platform_allow:
       - siwx917_rb4338a
+      - siwx917_rb4342a
+    extra_configs:
+      - CONFIG_SPI_SILABS_SIWX91X_GSPI_DMA=y
+      - CONFIG_SPI_IDEAL_TRANSFER_DURATION_SCALING=26
```
jerome-pouiller: Redundant with the `.conf` file, no?

drivers/spi/spi_silabs_siwx91x_gspi.c
smalae: This update ensures proper buffer alignment when using the SI91x GPDMA driver. A separate commit has been created to address this issue explicitly.

drivers/spi/spi_silabs_siwx91x_gspi.c
smalae: Updated

drivers/spi/spi_silabs_siwx91x_gspi.c
smalae: Done

drivers/spi/spi_silabs_siwx91x_gspi.c
smalae: Updated as per the suggestion with minor adjustments.

tests/drivers/spi/spi_loopback/testcase.yaml
smalae: Done

tests/drivers/spi/spi_loopback/testcase.yaml
smalae: Removed

jerome-pouiller (dismissed):

Martinhoff-maker (dismissed):

drivers/spi/spi_silabs_siwx91x_gspi.c
jhedberg: @tbursztyka are you ok with this explanation?

jerome-pouiller (approved):

Martinhoff-maker (approved):

tbursztyka (approved):

PR #93302, opened 2025-07-18
NXP driver: lptmr: add alarm and guard period API functions.

Holt-Sun:
Add alarm APIs: set_alarm, cancel_alarm
Add guard period APIs: set_guard_period

PR #99862, opened 2025-11-23
Enable alarm feature for lptmr

ZhaoxiangJin:
LPTMR is a commonly used wake-up source for many NXP MCUs and may be used as the companion low-power timer for the Cortex-M systick. This commit enabled LPTMR set and cacnel ALARM functions.

Enabling single-shot alarm support requires reprogramming the hardware compare register (COMPARE/CMR), which is the same resource used for the periodic TOP callback. Because of this hardware limitation, we added Kconfig 'CONFIG_COUNTER_MCUX_LPTMR_ALARM' to select whether alarm support is enabled. When alarm support is enabled the driver will not allow support set TOP operation at runtime (attempts to set TOP will return -ENOTSUP); When alarm support is
disabled the driver will not allow set and cancel ALARM operation at runtime (attempts to set and cancel ALARM will return -ENOTSUP). Choose the configuration that matches your application's requirements.

### DIFF
--- a/drivers/counter/counter_mcux_lptmr.c
+++ b/drivers/counter/counter_mcux_lptmr.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020 Vestas Wind Systems A/S
+ * Copyright 2025 NXP
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -11,6 +12,7 @@
 #include <zephyr/drivers/counter.h>
 #include <zephyr/irq.h>
 #include <fsl_lptmr.h>
+#include <zephyr/spinlock.h>
 
 struct mcux_lptmr_config {
 	struct counter_config_info info;
@@ -25,8 +27,15 @@ struct mcux_lptmr_config {
 };
 
 struct mcux_lptmr_data {
+#if defined(CONFIG_COUNTER_MCUX_LPTMR_ALARM)
+	counter_alarm_callback_t alarm_callback;
+	void *alarm_user_data;
+	bool alarm_active;
+	struct k_spinlock lock;
+#else
 	counter_top_callback_t top_callback;
 	void *top_user_data;
+#endif
 };
 
 static int mcux_lptmr_start(const struct device *dev)
@@ -60,6 +69,106 @@ static int mcux_lptmr_get_value(const struct device *dev, uint32_t *ticks)
 	return 0;
 }
 
+#if defined(CONFIG_COUNTER_MCUX_LPTMR_ALARM)
+static int mcux_lptmr_set_alarm(const struct device *dev, uint8_t chan_id,
+				const struct counter_alarm_cfg *alarm_cfg)
+{
+	ARG_UNUSED(chan_id);
+
+	const struct mcux_lptmr_config *config = dev->config;
+	struct mcux_lptmr_data *data = dev->data;
+
+	/* Counter API: Alarm callback cannot be NULL. */
+	if ((alarm_cfg == NULL) || (alarm_cfg->callback == NULL) || (alarm_cfg->ticks == 0U) ||
+		(alarm_cfg->ticks > config->info.max_top_value)) {
+		return -EINVAL;
+	}
+
+	k_spinlock_key_t key = k_spin_lock(&data->lock);
+
+	if (data->alarm_active) {
+		k_spin_unlock(&data->lock, key);
+		return -EBUSY;
+	}
+
+	data->alarm_callback = alarm_cfg->callback;
+	data->alarm_user_data = alarm_cfg->user_data;
+	data->alarm_active = true;
+
+	k_spin_unlock(&data->lock, key);
+
+	if (config->base->CSR & LPTMR_CSR_TEN_MASK) {
+		/* Already enabled, first stop then set period (HW constraint). */
+		LPTMR_StopTimer(config->base);
+		LPTMR_SetTimerPeriod(config->base, alarm_cfg->ticks);
+	} else {
+		LPTMR_SetTimerPeriod(config->base, alarm_cfg->ticks);
+		/* RM recommendation: clear status flag after setting period
+		 * when timer is disabled.
+		 */
+		LPTMR_ClearStatusFlags(config->base, kLPTMR_TimerCompareFlag);
+	}
+
+	LPTMR_EnableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+	LPTMR_StartTimer(config->base);
+
+	return 0;
+}
+
+static int mcux_lptmr_cancel_alarm(const struct device *dev, uint8_t chan_id)
+{
+	ARG_UNUSED(chan_id);
+
+	const struct mcux_lptmr_config *config = dev->config;
+	struct mcux_lptmr_data *data = dev->data;
+	k_spinlock_key_t key;
+
+	key = k_spin_lock(&data->lock);
+	if (!data->alarm_active) {
+		k_spin_unlock(&data->lock, key);
+		return 0;
+	}
+
+	LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+
+	data->alarm_callback = NULL;
+	data->alarm_user_data = NULL;
+	data->alarm_active = false;
+
+	k_spin_unlock(&data->lock, key);
+
+	LPTMR_StopTimer(config->base);
+
+	return 0;
+}
+
+static int mcux_lptmr_set_top_value(const struct device *dev,
+				    const struct counter_top_cfg *cfg)
+{
+	ARG_UNUSED(dev);
+	ARG_UNUSED(cfg);
+
+	return -ENOTSUP;
+}
+#else
+static int mcux_lptmr_set_alarm(const struct device *dev, uint8_t chan_id,
+				const struct counter_alarm_cfg *alarm_cfg)
+{
+	ARG_UNUSED(dev);
+	ARG_UNUSED(chan_id);
+	ARG_UNUSED(alarm_cfg);
+
+	return -ENOTSUP;
+}
+
+static int mcux_lptmr_cancel_alarm(const struct device *dev, uint8_t chan_id)
+{
+	ARG_UNUSED(dev);
+	ARG_UNUSED(chan_id);
+
+	return -ENOTSUP;
+}
+
 static int mcux_lptmr_set_top_value(const struct device *dev,
 				    const struct counter_top_cfg *cfg)
 {
@@ -87,6 +196,7 @@ static int mcux_lptmr_set_top_value(const struct device *dev,
 
 	return 0;
 }
+#endif /* CONFIG_COUNTER_MCUX_LPTMR_ALARM */
 
 static uint32_t mcux_lptmr_get_pending_int(const struct device *dev)
 {
@@ -122,9 +232,36 @@ static void mcux_lptmr_isr(const struct device *dev)
 	flags = LPTMR_GetStatusFlags(config->base);
 	LPTMR_ClearStatusFlags(config->base, flags);
 
+#if defined(CONFIG_COUNTER_MCUX_LPTMR_ALARM)
+	k_spinlock_key_t key = k_spin_lock(&data->lock);
+	counter_alarm_callback_t callback = data->alarm_callback;
+
+	if ((callback != NULL) && (data->alarm_active)) {
+		void *user_data = data->alarm_user_data;
+
+		LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+
+		data->alarm_callback = NULL;
+		data->alarm_user_data = NULL;
+		data->alarm_active = false;
+
+		k_spin_unlock(&data->lock, key);
+
+		uint32_t current_count = LPTMR_GetCurrentTimerCount(config->base);
+
+		LPTMR_StopTimer(config->base);
+
+		callback(dev, 0, current_count, user_data);
+
+		return;
+	}
+
+	k_spin_unlock(&data->lock, key);
+#else
 	if (data->top_callback) {
 		data->top_callback(dev, data->top_user_data);
 	}
+#endif
 }
 
 static int mcux_lptmr_init(const struct device *dev)
@@ -156,6 +293,8 @@ static int mcux_lptmr_init(const struct device *dev)
 static DEVICE_API(counter, mcux_lptmr_driver_api) = {
 	.start = mcux_lptmr_start,
 	.stop = mcux_lptmr_stop,
+	.set_alarm = mcux_lptmr_set_alarm,
+	.cancel_alarm = mcux_lptmr_cancel_alarm,
 	.get_value = mcux_lptmr_get_value,
 	.set_top_value = mcux_lptmr_set_top_value,
 	.get_pending_int = mcux_lptmr_get_pending_int,
@@ -189,7 +328,7 @@ static DEVICE_API(counter, mcux_lptmr_driver_api) = {
 			.freq = DT_INST_PROP(n, clock_frequency) /		\
 				BIT(DT_INST_PROP(n, prescale_glitch_filter)),	\
 			.flags = COUNTER_CONFIG_INFO_COUNT_UP,			\
-			.channels = 0,						\
+			.channels = 1,						\
 		},								\
 		.base = (LPTMR_Type *)DT_INST_REG_ADDR(n),			\
 		.clk_source = DT_INST_PROP(n, clk_source),			\
--- a/drivers/counter/Kconfig.mcux_lptmr
+++ b/drivers/counter/Kconfig.mcux_lptmr
@@ -9,3 +9,24 @@ config COUNTER_MCUX_LPTMR
 	depends on DT_HAS_NXP_LPTMR_ENABLED
 	help
 	  Enable support for the MCUX Low Power Timer (LPTMR).
+
+config COUNTER_MCUX_LPTMR_ALARM
+	bool "LPTMR alarm support"
+	depends on COUNTER_MCUX_LPTMR
+	help
+	  LPTMR is a commonly used wake-up source for many NXP MCUs and
+	  may be used as the companion low-power timer for the Cortex-M
+	  systick. Enabling single-shot alarm support requires reprogramming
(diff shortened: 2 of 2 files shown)


drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,141 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+		data->alarm_target_ticks = 0U;
+		data->alarm_active = false;
+
+		/* Re-enable LPTMR and interrupt if periodic top callback is not NULL. */
+		if (data->top_callback != NULL) {
+			LPTMR_EnableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+			LPTMR_StartTimer(config->base);
+		}
+
+		callback(dev, 0, data->alarm_target_ticks, data->alarm_user_data);
```
Holt-Sun: This logic is wrong: you set data->alarm_target_ticks = 0, set data->alarm_user_data = NULL, then pass them to callback, no meaning. You should firstly store data->alarm_target_ticks and data->alarm_user_data, then clear them, and pass the stored into callback.

Holt-Sun (changes requested):

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,141 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+		/* Set back the current top value */
+		LPTMR_SetTimerPeriod(config->base, data->current_top);
+
+		data->alarm_callback = NULL;
+		data->alarm_user_data = NULL;
+		data->alarm_target_ticks = 0U;
+		data->alarm_active = false;
+
+		/* Re-enable LPTMR and interrupt if periodic top callback is not NULL. */
+		if (data->top_callback != NULL) {
```
Holt-Sun: Driver should restore the timer run state after a one-shot alarm or cancel. Any counter can be used with https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/rtc/rtc_counter.c to function as RTC. 
It is not good to restart on “top_callback != NULL”. 
Suggest to track whether the timer was running when the alarm was armed.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,141 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+
+		target = alarm_cfg->ticks;
+		delta = target - now;
+	} else { /* Relative */
+		delta = alarm_cfg->ticks;
+
+		if (delta > data->current_top) { /* Too large delta */
+			return -EINVAL;
+		}
+
```
Holt-Sun:
```suggestion
        /* Disallow zero period; coerce to one tick for immediate-ish firing */
        if (delta == 0U) {
            delta = 1U;
        }
```

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,141 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+static int mcux_lptmr_cancel_alarm(const struct device *dev, uint8_t chan_id)
+{
+	ARG_UNUSED(chan_id);
+
+	const struct mcux_lptmr_config *config = dev->config;
+	struct mcux_lptmr_data *data = dev->data;
+
+	/* Stop LPTMR, CNR[COUNTER] is reset when LPTMR is disabled.
+	 * CSR[TCF] is cleared by disabling LPTMR.
+	 */
```
Holt-Sun: Suggest to firstly check if the alarm exists.
Don't stop counter if it is running before.

ZhaoxiangJin: I'm not currently planning to check for the alarm's existence at the beginning, because the normal logic is set-callback-cancel or cancel-set-callback. The flag used for detection should be set in the set function, but then cleared by the callback. After looking at several other counter drivers, it seems that none of them check for alarms.

removed stop function.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: I prefer check `alarm_cfg->ticks==0U` at the beginning

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: yes, that's my bad, should first store.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: In the patch, current_top is used to record/track the current counter top, in the alarm isr, the current_top is set back. Please check.

I will update the logic here: 
1. do not disable interrupt and stop timer
2. set current_top back
3. check if the top_callbackis null. If it is, then disable interrupt

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -65,22 +70,54 @@ static int mcux_lptmr_set_top_value(const struct device *dev,
...
+				 */
+				if (cfg->flags & COUNTER_TOP_CFG_RESET_WHEN_LATE) {
+					LPTMR_StopTimer(config->base);
+					LPTMR_SetTimerPeriod(config->base, cfg->ticks);
+					LPTMR_StartTimer(config->base);
+				} else {
+					return -ETIME;
+				}
+			} else {
+				/* Current counter value is samller than the new top value.
```
Holt-Sun: Typo, shall be smaller.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,135 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+		LPTMR_SetTimerPeriod(config->base, data->current_top);
+
+		/* If no periodic top callback, mask further interrupts */
+		if (!data->top_callback) {
+			LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+		}
+
+		uint32_t current_count = LPTMR_GetCurrentTimerCount(config->base);
+
+		callback(dev, 0, current_count, user_data);
```
Holt-Sun, line 254: Here is the defintion of 
```c
/** @brief Alarm callback
 *
 * @param dev       Pointer to the device structure for the driver instance.
 * @param chan_id   Channel ID.
 * @param ticks     Counter value that triggered the alarm.
 * @param user_data User data.
 */
typedef void (*counter_alarm_callback_t)(const struct device *dev,
					 uint8_t chan_id, uint32_t ticks,
					 void *user_data);
```
The "ticks" should be alarm_target_ticks instead of current_count.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,135 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+	} else {
+		/* No consumer registered, just ack the interrupt. */
+		LPTMR_ClearStatusFlags(config->base, LPTMR_GetStatusFlags(config->base));
 	}
 }
 
+static int mcux_lptmr_set_alarm(const struct device *dev, uint8_t chan_id,
+				const struct counter_alarm_cfg *alarm_cfg)
+{
+	ARG_UNUSED(chan_id);
```
Holt-Sun: Should validate chan_id, only 1 chan_id is supported.

Holt-Sun (changes requested):

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: I don't think there's a clear specification here that ticks should be the target. I've looked at implementations of other counter drivers - some use current, some use target. In my view, current is more meaningful because callbacks typically indicate "an event has occurred", and developers often want to know "what was the count at the moment it occurred". This directly reflects the system state (timer wrap, delays, external influences, etc.).

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: counter common API already [validated ](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/counter.h#L505)this

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -117,16 +154,139 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+		target = (now + delta) % data->current_top;
+	}
+
+	/* Ensure timer is stopped and status flags cleared before reprogramming
+	 * so that a smaller new period does not immediately assert TCF due to
+	 * an already advanced CNR value.
+	 * CNR[COUNTER] is reset when LPTMR is disabled.
+	 */
+	LPTMR_StopTimer(config->base);
+	LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
```
zejiang0jason: Actually, I think support alarm feature by start/stop LPTMR is not a good idea, because it will results in drift, especially when the document has mentioned it:
image

ZhaoxiangJin: It seems that as long as it's in counter mode, regardless of what method is used to implement the alarm, the issue mentioned in the RM NOTE will exist.

ZhaoxiangJin: Implementing lptmr alarm through stop/start is the only feasible approach, because lptmr only has one compare register, and the scenario we must consider is: setting an alarm (modifying the compare register) when a top already exists. However, according to the lptmr manual, we can only rewrite the compare register when lptmr is disabled. When lptmr is enabled, it can only be rewritten when CDR[TCF] is 1, but implementing this in software would be more complex.

zejiang0jason: right, I doubt whether LPTMR is suitable for alarm.

ZhaoxiangJin: I don't think there should be any major problems, unless the counter testcase design is flawed😄

Both hardware and software serve the requirements. Although the current hardware features are insufficient, if we can achieve the requirements through software means, then at least from the current perspective, it should be reasonable.

ZhaoxiangJin:
> I suddenly remember that I have created one PR to add alarm and guard period API functions before: #93302 The PR is closed because "using the same counter register for both top callback and alarms? both APIs can be used simultaneously".

Did you get counter_basic_api testcase passed? testcases `test_set_top_value_with_alarm `and `single_shot_alarm_top ` seems test top and alarm simultaneously.

Holt-Sun:
> > I suddenly remember that I have created one PR to add alarm and guard period API functions before: #93302 The PR is closed because "using the same counter register for both top callback and alarms? both APIs can be used simultaneously".
> 
> Did you get counter_basic_api testcase passed? testcases `test_set_top_value_with_alarm `and `single_shot_alarm_top ` seems test top and alarm simultaneously.

I checked the test_set_top_value_with_alarm and single_shot_alarm_top, they test alarm and top in a function but sequential. The counter_top_cfg should be with flag COUNTER_TOP_CFG_DONT_RESET, then the COUNTER_TOP_CFG_DONT_RESET shall be true, otherwise it will reset the alarm counter.

I understand the top and alarm are designed for different counters with the same top.

ZhaoxiangJin:
Update:
Considering that enabling both alarm and top (especially top callback) functionality simultaneously in complex usage scenarios can indeed lead to race conditions, I have added a Kconfig option `CONFIG_COUNTER_MCUX_LPTMR_ALARM` to isolate the two functionalities, allowing users to enable the required functionality based on their needs

Hello @Holt-Sun, @zejiang0jason, please help to review again, thanks.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -60,6 +69,131 @@ static int mcux_lptmr_get_value(const struct device *dev, uint32_t *ticks)
...
+	} else { /* Relative */
+		delta = alarm_cfg->ticks;
+
+		if (delta > config->info.max_top_value) { /* Too large delta */
+			return -EINVAL;
+		}
+
+		now = LPTMR_GetCurrentTimerCount(config->base);
+		/* Absolute target relative to now, modulo current top value */
+		target = (now + delta) % config->info.max_top_value;
```
Holt-Sun: The "target" is calculated but never used. Should the hardware be programmed with target instead of delta?

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -60,6 +69,131 @@ static int mcux_lptmr_get_value(const struct device *dev, uint32_t *ticks)
...
+
+	if (config->base->CSR & LPTMR_CSR_TEN_MASK) {
+		/* Already enabled, first stop then set period (HW constraint). */
+		LPTMR_StopTimer(config->base);
+		LPTMR_SetTimerPeriod(config->base, delta);
+	} else {
+		LPTMR_SetTimerPeriod(config->base, delta);
+	}
+
+	LPTMR_EnableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
```
Holt-Sun, line 112: Timer is started before alarm metadata is set. If IRQ fires immediately, the ISR might see inconsistent state. These assignments should occur before starting the timer.

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -122,9 +257,34 @@ static void mcux_lptmr_isr(const struct device *dev)
...
+
+		LPTMR_StopTimer(config->base);
+		LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);
+
+		k_spin_unlock(&data->lock, key);
+		data->alarm_callback = NULL;
+		data->alarm_user_data = NULL;
+		data->alarm_active = false;
+
+		uint32_t current_count = LPTMR_GetCurrentTimerCount(config->base);
```
Holt-Sun: Timer is stopped before reading.  If TEN is 0, it resets the LPTMR internal logic, including CNR[COUNTER].

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -122,9 +257,34 @@ static void mcux_lptmr_isr(const struct device *dev)
 	flags = LPTMR_GetStatusFlags(config->base);
 	LPTMR_ClearStatusFlags(config->base, flags);
 
+#if defined(CONFIG_COUNTER_MCUX_LPTMR_ALARM)
+	counter_alarm_callback_t callback = data->alarm_callback;
```
Holt-Sun: The spinlock is acquired, then immediately unlocked, but the callback state is read without proper protection after unlock.

```suggestion
k_spinlock_key_t key = k_spin_lock(&data->lock);

if ((data->alarm_callback != NULL) && (data->alarm_active)) {
	counter_alarm_callback_t callback = data->alarm_callback;
	void *user_data = data->alarm_user_data;

	LPTMR_StopTimer(config->base);
	LPTMR_DisableInterrupts(config->base, kLPTMR_TimerInterruptEnable);

	// Clear state BEFORE unlocking
	data->alarm_callback = NULL;
	data->alarm_user_data = NULL;
	data->alarm_active = false;

	k_spin_unlock(&data->lock, key);

	uint32_t current_count = LPTMR_GetCurrentTimerCount(config->base);
	callback(dev, 0, current_count, user_data);
	return;
}

k_spin_unlock(&data->lock, key);
```

Holt-Sun (changes requested):

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: in current solution, `set_top `and `set_alarm `will not work simultaneously, so ticks should be written directly to the CMR register; there is no need to calculate `target` and `delta`. patch updated.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: yes, should get current counter first, then stop, updated.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: Yes, due to a flaw in the timing of lock acquisition/release in ISR, race conditions can indeed occur. The code has been updated.

Holt-Sun (dismissed):

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -122,9 +227,34 @@ static void mcux_lptmr_isr(const struct device *dev)
 	flags = LPTMR_GetStatusFlags(config->base);
 	LPTMR_ClearStatusFlags(config->base, flags);
 
+#if defined(CONFIG_COUNTER_MCUX_LPTMR_ALARM)
+	k_spinlock_key_t key = k_spin_lock(&data->lock);
+	counter_alarm_callback_t callback = data->alarm_callback;
+
+	if ((callback != NULL) && (data->alarm_active)) {
```
zejiang0jason, line 239: if callback is NULL, but `data->alarm_active` is true, is it not necessary to stop the alarm?

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -60,6 +69,101 @@ static int mcux_lptmr_get_value(const struct device *dev, uint32_t *ticks)
...
+	const struct mcux_lptmr_config *config = dev->config;
+	struct mcux_lptmr_data *data = dev->data;
+
+	/* Counter API: Alarm callback cannot be NULL. */
+	if ((alarm_cfg == NULL) || (alarm_cfg->callback == NULL) || (alarm_cfg->ticks == 0U) ||
+		(alarm_cfg->ticks > config->info.max_top_value)) {
+		return -EINVAL;
+	}
+
+	k_spinlock_key_t key = k_spin_lock(&data->lock);
```
zejiang0jason, line 87: The period between spinlock should be as short as possible.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: alarm callback can not be NULL, this is API constraint, if it is NULL, then will set_alarm failure.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: correct, the current code can be improved. It only needs to protect the shared data.

zejiang0jason (approved):

Holt-Sun (approved):

PR #99870, opened 2025-11-23
net: follow up convert IEEE 802.15.4 drivers to use renamed network APIs

kartben:
Fixed a couple instances missed in earlier commit
985d1c0351291a9d6976601ab25ef62ad70c91da where `AF_UNSPEC` should be renamed to `NET_AF_UNSPEC`.


jukkar:
Duplicate of #99859

PR #86920, opened 2025-03-11
drivers: usb: rename UDC buffer to USB buffer 

jfischer-no:
Rename it to USB buffers so we can use it in host support without
confusion. Keep the UDC_* macros for now, we can deprecate and remove
them later.

Resolves #99472

PR #99879, opened 2025-11-24
Feature/fix non cache buffer of uhc ehci

MarkWangChinese:
- Differentiating between IN and OUT transfers for proper buffer setup.
    For OUT transfers: use existing data from buf->data with buf->len.
    For IN transfers: use available space from net_buf_tail() with tailroom.
- Improves buffer management in MCUX USB host controller driver by:
    Using net_buf_tailroom() to prevent buffer overflows in transfer callback.
    Add proper error handling for nocache memory allocation failures.
    Only allocate setup packet for control endpoint (EP0).
    Copy data to nocache buffer for OUT transfers.

### DIFF
--- a/drivers/usb/uhc/uhc_mcux_ehci.c
+++ b/drivers/usb/uhc/uhc_mcux_ehci.c
@@ -168,13 +168,12 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 		uhc_mcux_nocache_free(transfer->setupPacket);
 	}
 #endif
-	if ((xfer->buf != NULL) && (transfer->transferBuffer != NULL)) {
-		if (transfer->transferSofar > 0) {
+	if ((xfer->buf != NULL) && (transfer->transferBuffer != NULL) &&
+	    USB_EP_DIR_IS_IN(xfer->ep) && (transfer->transferSofar > 0)) {
 #if defined(CONFIG_NOCACHE_MEMORY)
-			memcpy(xfer->buf->__buf, transfer->transferBuffer, transfer->transferSofar);
+		memcpy(xfer->buf->__buf, transfer->transferBuffer, transfer->transferSofar);
 #endif
-			net_buf_add(xfer->buf, transfer->transferSofar);
-		}
+		net_buf_add(xfer->buf, transfer->transferSofar);
 #if defined(CONFIG_NOCACHE_MEMORY)
 		uhc_mcux_nocache_free(transfer->transferBuffer);
 #endif
--- a/drivers/usb/uhc/uhc_mcux_ehci.c
+++ b/drivers/usb/uhc/uhc_mcux_ehci.c
@@ -171,7 +171,7 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 	if ((xfer->buf != NULL) && (transfer->transferBuffer != NULL) &&
 	    USB_EP_DIR_IS_IN(xfer->ep) && (transfer->transferSofar > 0)) {
 #if defined(CONFIG_NOCACHE_MEMORY)
-		memcpy(xfer->buf->__buf, transfer->transferBuffer, transfer->transferSofar);
+		memcpy(net_buf_tail(xfer->buf), transfer->transferBuffer, transfer->transferSofar);
 #endif
 		net_buf_add(xfer->buf, transfer->transferSofar);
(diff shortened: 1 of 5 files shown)


drivers/usb/uhc/uhc_mcux_ehci.c
```diff
@@ -198,10 +197,33 @@ static usb_host_transfer_t *uhc_mcux_hal_init_transfer(const struct device *dev,
...
-		mcux_xfer->transferBuffer = uhc_mcux_nocache_alloc(mcux_xfer->transferLength);
+	if (USB_EP_GET_IDX(xfer->ep) == 0 && mcux_xfer->setupPacket != NULL) {
+		uint8_t *nocache_buf = uhc_mcux_nocache_alloc(8u);
+
+		if (nocache_buf == NULL) {
+			k_mem_slab_free(&mcux_uhc_transfer_pool, mcux_xfer);
+			return NULL;
+		}
+
+		memcpy(nocache_buf, mcux_xfer->setupPacket, 8u);
```
jfischer-no: >Only allocate setup packet for control endpoint (EP0).

This should go into a separate commit.

Also, it is not obvious that after uhc_mcux_hal_init_transfer_common(), mcux_xfer->setupPacket has the same value as xfer->setup_pkt. I think you should keep original code `memcpy(nocache_buf, xfer->setup_pkt, 8u);`.

jfischer-no (changes requested):

drivers/usb/uhc/uhc_mcux_ehci.c
MarkWangChinese: Updated. I split as three commits. Please help to review again.

jfischer-no (approved):
LGTM

drivers/usb/uhc/uhc_mcux_ehci.c
```diff
@@ -168,13 +168,12 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 		uhc_mcux_nocache_free(transfer->setupPacket);
 	}
 #endif
-	if ((xfer->buf != NULL) && (transfer->transferBuffer != NULL)) {
-		if (transfer->transferSofar > 0) {
+	if ((xfer->buf != NULL) && (transfer->transferBuffer != NULL) &&
+	    USB_EP_DIR_IS_IN(xfer->ep) && (transfer->transferSofar > 0)) {
 #if defined(CONFIG_NOCACHE_MEMORY)
-			memcpy(xfer->buf->__buf, transfer->transferBuffer, transfer->transferSofar);
+		memcpy(net_buf_tail(xfer->buf), transfer->transferBuffer, transfer->transferSofar);
```
josuah, line 174: Is this what [`net_buf_add_mem(xfer->buf, transfer->transferBuffer, transfer->transferSoFar)`](https://docs.zephyrproject.org/latest/services/net_buf/index.html#common-operations) does?

Then it would also require to put `net_buf_add()` in an `#else`.

`if (IS_ENABLED(CONFIG_NOCACHE_MEMORY))` can replace the `#if defined(CONFIG_NOCACHE_MEMORY)` if wanting a bit more clarity, but not always possible to use it (i.e. undefined struct fields).

josuah (commented):
LGTM but here is an optional suggestion, so I am waiting your confirmation and then approve (to avoid accidental merge while making changes).

drivers/usb/uhc/uhc_mcux_ehci.c
MarkWangChinese: Hi @josuah Thanks for the review. All the codes that are wrapped by `CONFIG_NOCACHE_MEMORY` in this file will be removed if https://github.com/zephyrproject-rtos/zephyr/pull/86920 is merged and usb host use the same way to define the transfer buffer as usb device stack.
The reason is: the codes just use internal non-cached buffer to replace the original buffer because usb controller needs the buffer to be non-cached or cache-line aligned.
After the codes are removed, then only the `net_buf_add` is kept. it is more easy to remove the codes in this way. Is it OK from your side?

Note: I agree that your suggestion is better if the codes need to be kept forever.

josuah (approved):

drivers/usb/uhc/uhc_mcux_ehci.c
josuah: Good to be reminded, thanks. No problem on my side of course.

PR #99884, opened 2025-11-24
lora: preamble and header detection callbacks

JordanYates:
Add the option to receive callbacks when the LoRa modem detects a valid preamble or packet header. This can be used to detect whether the modem is in the middle of receiving a packet, allowing the user to possibly defer a mode switch (to TX or IDLE) if it is. This is useful due to the generally long packet durations on LoRa.

Only implemented and tested for the `lora-basics-modem` backend. Example interrupt timings when receiving a 15 byte packet at 125kHz bandwidth, SF 10: 
```
[00:00:30.675,264] <err> lbm_driver: Preamble detected IRQ
[00:00:30.821,120] <err> lbm_driver: Header valid IRQ
[00:00:30.985,216] <inf> phy_lora: LoRa RX RSSI: -15 dBm, SNR: 9 dB
```

### DIFF
--- a/drivers/lora/lora_basics_modem/lbm_common.c
+++ b/drivers/lora/lora_basics_modem/lbm_common.c
@@ -289,7 +289,7 @@ release:
 	return ret;
 }
 
-int lbm_lora_recv_async(const struct device *dev, lora_recv_cb cb, void *user_data)
+int lbm_lora_recv_async(const struct device *dev, const struct lora_recv_async_callbacks *cb)
 {
 	const struct lbm_lora_config_common *config = dev->config;
 	struct lbm_lora_data_common *data = dev->data;
@@ -304,6 +304,10 @@ int lbm_lora_recv_async(const struct device *dev, lora_recv_cb cb, void *user_da
 		return 0;
 	}
 
+	if (cb->recv == NULL) {
+		return -EINVAL;
+	}
+
 	/* Ensure available */
 	if (!modem_acquire(dev)) {
 		return -EBUSY;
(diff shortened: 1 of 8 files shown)


drivers/lora/lora_basics_modem/lbm_common.h
```diff
@@ -54,11 +54,8 @@ struct lbm_lora_data_common {
 			int16_t rssi_dbm;
 			int8_t snr_db;
 		} sync;
-		struct {
-			/* Async RX params */
-			lora_recv_cb rx_cb;
-			void *user_data;
-		} async;
+		/* Async RX params */
+		struct lora_recv_async_callbacks async;
```
mniestroj, line 58: This should be part of "lora: single asynchronous RX parameter" (previous commit), right?

JordanYates: It could be, but it doesn't have to be. After the first commit, the LBM backend is extracting the fields from the struct into local variables like the other backends.

The second commit is the one that adds new callbacks, where it is simpler to just assign the struct directly rather than add more fields.

mniestroj (dismissed):

VynDragon:
Okay that was actually a bug in the PR, here are 2 ways to fix it (with my test code included) @JordanYates 
[fix1.patch](https://github.com/user-attachments/files/24156489/fix1.patch)
[fix2.patch](https://github.com/user-attachments/files/24156490/fix2.patch)

drivers/lora/lora_basics_modem/lbm_common.c
```diff
@@ -289,33 +289,57 @@ int lbm_lora_recv(const struct device *dev, uint8_t *msg, uint8_t msg_len, k_tim
...
 	const struct lbm_lora_config_common *config = dev->config;
 	struct lbm_lora_data_common *data = dev->data;
+	ral_irq_t irqs = RAL_IRQ_LORA;
 	ral_status_t status;
 
 	/* Cancel ongoing reception */
 	if (cb == NULL) {
+		if (data->rx_state.async.preamble_detected || data->rx_state.async.header_valid) {
+			/* Reset IRQ state */
+			(void)ral_set_dio_irq_params(&config->ralf.ral, irqs);
```
VynDragon, line 303: IRQs should also be reset when starting synchronous RX so as to not end RX at healder_valid or preamble detected in that mode.

VynDragon (changes requested):

PR #99888, opened 2025-11-24
drivers: uart_nrfx_uarte: Fix runtime device PM for interrupt driven API

anangl:
Add missing getting/putting of the device when the RX interrupt is enabled/disabled. Also fix enabling and disabling of the TX interrupt so that the device is got/put only if the interrupt wasn't already enabled/disabled and device PM reference counting is done correctly.

This fixes `console_getchar()` that would hang when used with `CONFIG_PM_DEVICE_RUNTIME=y`.

### DIFF
--- a/drivers/serial/uart_nrfx_uarte.c
+++ b/drivers/serial/uart_nrfx_uarte.c
@@ -2832,27 +2832,35 @@ static void uarte_nrfx_irq_tx_enable(const struct device *dev)
 {
 	NRF_UARTE_Type *uarte = get_uarte_instance(dev);
 	struct uarte_nrfx_data *data = dev->data;
+	bool already_enabled;
 
-	if (IS_ENABLED(CONFIG_PM_DEVICE_RUNTIME)) {
-		pm_device_runtime_get(dev);
-	}
+	pm_device_runtime_get(dev);
 
 	unsigned int key = irq_lock();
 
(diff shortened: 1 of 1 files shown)


drivers/serial/uart_nrfx_uarte.c
```diff
@@ -2888,15 +2896,23 @@ static void uarte_nrfx_irq_rx_enable(const struct device *dev)
...
 }
 
 /** Interrupt driven receiver disabling function */
 static void uarte_nrfx_irq_rx_disable(const struct device *dev)
 {
 	NRF_UARTE_Type *uarte = get_uarte_instance(dev);
 
-	nrf_uarte_int_disable(uarte, NRF_UARTE_INT_ENDRX_MASK);
+	if (nrf_uarte_int_enable_check(uarte, NRF_UARTE_INT_ENDRX_MASK)) {
+		pm_device_runtime_put(dev);
```
nordic-krch: i think that it should be `pm_device_runtime_put_async(dev, K_NO_WAIT);` as it may happen from the interrupt context. For ISR_SAFE case that will be the same thing.

anangl: Applied.

nordic-krch (approved):

carlescufi (approved):

ppryga-nordic (approved):

PR #99144, opened 2025-11-10
drivers: rtc: microchip: Add rtc driver

Farsin-Nasar-Microchip:
Add rtc driver for Microchip RTC G1.
Add the device tree node and the binding file for microchip RTC G1.
Add RTC node in sam_e54_xpro.dts
Update sam_e54_xpro.yaml to reflect RTC G1 support on the board.
Added sam_e54_xpro.conf file for test

PR #93440, opened 2025-07-21
kernel: clarify and assert interrupts aren't disabled upon context switch

mathieuchopstm:
The `do_swap()` routine used when `CONFIG_USE_SWITCH=y` asserts that caller thread does not hold any spinlock when `CONFIG_SPIN_VALIDATE` is enabled. However, there is no similar check in place when `CONFIG_USE_SWITCH=n`.

Copy this assertion in the `USE_SWITCH=n` implementation of `z_swap_irqlock()`.

PR #99889, opened 2025-11-24
driver: rtc: microchip: Add RTC-G2

TonyHan11:
Changes included in this PR:
 - update `drivers/rtc/rtc_sam.c` to support sama7d65
 - add `rtc` to the support list in `sama7d65_curiosity.yaml`
 - enable `tests/drivers/rtc/rtc_api` for sama7d65_curiosity
 
Tested with cases under `tests/drivers/rtc` on sama7d65_curiosity board.


drivers/rtc/rtc_sam.c
```diff
@@ -27,6 +27,68 @@
 #define RTC_SAM_CALIBRATE_PPB_QUANTA (1500)
 #define RTC_SAM_CALIBRATE_PPB_LOW_SCALE (30500)
 
+/* The SAMA7D6 HALs change the name of the fields, so we need to
+ * define it this way to match how the other SoC variants name it
+ */
+#if defined(CONFIG_SOC_SERIES_SAMA7D6)
+#undef RTC_CR_UPDTIM
+#define RTC_CR_UPDTIM RTC_CR_UPDTIM_Msk
```
nandojve: NAK, this must be in the hal. No fixups in Zephyr!

No new defines in HAL, it must go direct in the headers so maybe you need to update the ATDF files.

nandojve (changes requested):

drivers/rtc/rtc_sam.c
TonyHan11: OK, updated these definitions to HAL in https://github.com/zephyrproject-rtos/hal_microchip/pull/46, will update after the PR is merged, thanks.

nandojve: You need to bring the change to here to see if everything goes well.
Then when this PR is ready we merge in the HAL and you update the hash here.

TonyHan11: Thanks, added a commit to update `west.yml`.

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_RTC_ALARM=y
+CONFIG_RTC_UPDATE=y
+CONFIG_RTC_CALIBRATION=y
+# CONFIG_SPIN_VALIDATE is not set
```
bjarki-andreasen: Remove this line

drivers/rtc/rtc_mchp_g1.c
```diff
@@ -0,0 +1,691 @@
+/*
+ * Copyright (c) 2023 Bjarki Arge Andreasen
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT microchip_rtc_g1
```
bjarki-andreasen: Is this file unmodified other than the compat? If it is, the existing sam driver can be extended to support both compats, if not, append your copyright to the header

bjarki-andreasen (changes requested):
Could you update the PR title to describe what is changing and why? Its hard to follow why sam0 is going away and mchp_g1 seems to be duplicating a lot from the atmel_sam driver

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.conf
TonyHan11: The line `# CONFIG_SPIN_VALIDATE is not set` could not be removed due to it will cause the following failure:
```
*** Booting Zephyr OS build v4.3.0-3992-g9ea10229d959 ***
Running TESTSUITE rtc_api
START - test_alarm
ASSERTION FAIL [arch_irq_unlocked(key) || _kernel.cpus[0].current->base.thread_state & (((1UL << (0))) | ((1UL << (3))))] @ WEST_TOPDIR/zephyr/kernel/include/kswap.h:210
        Context switching while holding lock!
E: r0/a1:  0x00000004  r1/a2:  0x000000d2  r2/a3:  0x00800002
E: r3/a4:  0xe2020200 r12/ip:  0x00000000 r14/lr:  0x6000a9d0
E:  xpsr:  0x6000011f
E: Faulting instruction address (r15/pc): 0x60002e68
E: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
E: Current thread: 0x6002c010 (test_alarm)
E: Halting system
```

drivers/rtc/rtc_mchp_g1.c
TonyHan11: As the `G1 RTC` driver is already merged to the `main` branch by #99144 for SAME54, the driver in this PR is renamed to `G2 RTC` to avoid the conflicts.

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.conf
bjarki-andreasen: That failure is a fatal flaw in interrupt locking, it has to be addressed or it will destroy the kernel. The reason `CONFIG_SPIN_VALIDATE` can be deselected is for speed optimization in production code which has already been tested and validated to not cause this error, its not so that it can be ignored...

TonyHan11: Reproduced `ASSERTION FAIL` on sama7g54-ek board with `tests/drivers/rtc/rtc_api`. The assertion was added with the commit ID 5183fc56938 from [PR #93440](https://github.com/zephyrproject-rtos/zephyr/pull/93440/commits/124fb897b4908d7bcd0612bc8765445e01e81cce), to be debugged. Thanks.

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
```diff
@@ -65,11 +65,19 @@
...
 			interrupts = <GIC_SPI 6 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
 			clocks = <&clk32k 1>;
+			protection = <&syscwp>;
+			status = "disabled";
+		};
+
+		syscwp: syscwp@e001d900 {
+			compatible = "syscon";
+			reg = <0xe001d900 0x8>;
+			status = "disabled";
```
nandojve: if and only if there is nothing that user needs to configure in this particular node, you can remove the `status = "disabled";`

boards/microchip/sam/sama7d65_curiosity/sama7d65_curiosity.dts
```diff
@@ -91,3 +91,11 @@
...
 	clock-frequency = <DT_FREQ_M(10)>;
 };
+
+&rtc {
+	status = "okay";
+};
+
+&syscwp {
+	status = "okay";
+};
```
nandojve: if and only if you could remove that status I believe you can drop this too.

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.overlay
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) 2025 Microchip Technology Inc. and its subsidiaries
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/ {
+	aliases {
+		rtc = &rtc;
+	};
```
nandojve, line 10: recommendation is add this inside soc devicetree.
With that, we decrease maintenance and all boards inherit this node.

nandojve (changes requested):

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.overlay
TonyHan11: Updated with the following, thanks.
- For `syscwp`: `status = "disabled"` removed from the dtsi file and `&syscwp` removed from the board dts file.
- `aliases` for `rtc` added to soc devicetree and the file for adding the aliases deleted.

tests/drivers/rtc/rtc_api/boards/sama7d65_curiosity.conf
TonyHan11: `# CONFIG_SPIN_VALIDATE is not set` removed and `ASSERTION FAIL` fixed, thanks.

nandojve (dismissed):

dts/bindings/rtc/microchip,rtc-g2.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright(c) 2026 Microchip Technology Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+title: Microchip G2 RTC Controller
+
+description: |
+  Microchip G2 RTC driver.
+
+  Group G2 RTC driver supports following hardware peripherals:
```
JarmouniA:
```suggestion
  Group G2 RTC supports following hardware peripherals:
```
#103087

dts/bindings/rtc/microchip,rtc-g2.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright(c) 2026 Microchip Technology Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+title: Microchip G2 RTC Controller
+
+description: |
+  Microchip G2 RTC driver.
```
JarmouniA:
```suggestion
  Microchip G2 RTC peripheral.
```

drivers/rtc/Kconfig.mchp
```diff
@@ -7,3 +7,11 @@ config RTC_MCHP_G1
...
 	help
 	  Enable RTC driver for Microchip G1.
+
+config RTC_MCHP_G2
+	bool "Microchip RTC G2 driver"
+	default y
+	depends on DT_HAS_MICROCHIP_RTC_G2_ENABLED
+	select SYSCON
+	help
+	  Microchip Real-Time Clock (RTC) G2 driver used on SAM SoC series.
```
JarmouniA, line 17: That's what `drivers/rtc/rtc_sam.c` is for, no? Is this for a different peripheral?

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
```diff
@@ -65,11 +69,18 @@
 		};
 
 		rtc: rtc@e001d800 {
-			compatible = "atmel,sam-rtc";
+			compatible = "microchip,rtc-g2";
```
JarmouniA, line 72: Why!?

JarmouniA (changes requested):

dts/bindings/rtc/microchip,rtc-g2.yaml
TonyHan11: Thanks, updated the title and description in the dts bindings file.

drivers/rtc/Kconfig.mchp
TonyHan11: This driver is very similar with `drivers/rtc/rtc_sam.c`. The biggest difference is the definitions of the RTC registers. 
`drivers/rtc/rtc_sam.c` is used for SAMA7G54 with a patch in `hal` ([update register definitions for RTC](https://github.com/zephyrproject-rtos/hal_microchip/pull/30/changes/139ecf982b4cdb13e5c664798a56aecab1236ab4)). 
Here a new driver is used for sama7d65 which needs no changes on `hal`.

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
TonyHan11: Here the new driver `drivers/rtc/rtc_mchp_g2.c` is used instead of using `drivers/rtc/rtc_sam.c`.

JarmouniA: Use the same compatible with the new driver. The legacy driver can be disabled by default, and the new one enabled by default in Kconfig.
See Additional Context in #103087

TonyHan11: > Use the same compatible with the new driver. The legacy driver can be disabled by default, and the new one enabled by default in Kconfig. See Additional Context in #103087

The old compatible `atmel,sam-rtc` is still used by sam3x, sam4e, sam4s, samx7x and sama7g54 SoCs. Do you mean to use it for sama7d65 too and distinguish which .c file to be used in Kconfig or CMakefile.txt? Thanks.

JarmouniA: > Do you mean to use it for sama7d65 too

Yes if it's the same hw device as in the other SoCs using it.

sunil-abraham (dismissed):
Looks good to me.

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
TonyHan11: > > Do you mean to use it for sama7d65 too
> 
> Yes if it's the same hw device as in the other SoCs using it.

Actually the hw is not the same for MCUs (sam3x, sam4, samx7x) and MPUs (sama7g54, sama7d65). The hw for MPUs is more powerful as it supports UTC mode and tamper timestamping while the hw for MCUs doesn't. It's planned to let sama7g54 use the new driver too.

nandojve (changes requested):
@JarmouniA , @bjarki-andreasen ,

Is there anything missing?

nandojve (dismissed):

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
JarmouniA: Please clarify that in the binding description.

Also, just for your information, a device DT node can have multiple compatibles, from the most specific to the least.
https://docs.zephyrproject.org/latest/build/dts/bindings-syntax.html#compatible

bjarki-andreasen (dismissed):

Holt-Sun (dismissed):

TonyHan11:
Rebased by solving conflicts in `boards/microchip/sam/sama7d65_curiosity/sama7d65_curiosity.yaml` introduced by bf9007f3819

@nandojve , @bjarki-andreasen , @Holt-Sun , @sunil-abraham ,
Would you like to review and approve it again? Thank you very much.

bjarki-andreasen (approved):

Holt-Sun (approved):

nandojve (approved):

AzharMCHP (approved):

NhMchp (approved):

JarmouniA (commented):
Still not addressed https://github.com/zephyrproject-rtos/zephyr/pull/99889#discussion_r2807595636

dts/arm/microchip/sam/sama7/sama7d6/sama7d6.dtsi
nandojve: @JarmouniA ,

Microchip have many IPs and those could have evolution over the time. A IP that seems to be the same sometimes are not. As far I know, we don't need to explain what are the differences between the IPs in the devicetree. The information which is relevant to all the new bindings is the `module name="rtc" id="6056" version="0"` which define the IP. That information is relevant to the vendor and Microchip provide correct instructions to their customers when they need to customize a board.

JarmouniA (approved):

PR #99894, opened 2025-11-24
net: Using const char * in net_hostname_set()

### LINKED ISSUE #99873: net: net_hostname_set() does not work from C++ with a const char *
### Describe the bug

When attempting to use net_hostname_set() with a const char * from C++, the build fails.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Add the following to samples/cpp/hello_world/prj.conf

- CONFIG_NETWORKING
- CONFIG_NET_HOSTNAME_ENABLE
- CONFIG_NET_HOSTNAME_DYNAMIC

Add the following to the top of samples/cpp/hello_world/src/main.cpp
```
- #include <zephyr/net/hostname.h>
```

Then in main()
```
- const char *hostname = "zephyr";
- net_hostname_set(hostname, strlen(hostname));
```

### Relevant log output

```shell
src/main.cpp:13:26: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
   13 |         net_hostname_set(hostname, strlen(hostname));
      |                          ^~~~~~~~
      |                          |
      |                          const char*
In file included from src/main.cpp:8:
zephyr/include/zephyr/net/hostname.h:79:28: note:   initializing argument 1 of 'int net_hostname_set(char*, size_t)'
   79 | int net_hostname_set(char *host, size_t len);
      |                      ~~~~~~^~~~
```

### Impact

### Environment

- OS: macOS Tahoe
- SDK 0.17.4
- Zephyr v4.3.0

### Additional Context

This can be resolved by adding `const` to the function arguments
```
diff --git a/include/zephyr/net/hostname.h b/include/zephyr/net/hostname.h
index 2278121aa15..242f9af3ced 100644
--- a/include/zephyr/net/hostname.h
+++ b/include/zephyr/net/hostname.h
@@ -76,7 +76,7 @@ static inline const char *net_hostname_get(void)
  * @return 0 if ok, <0 on error
  */
 #if defined(CONFIG_NET_HOSTNAME_DYNAMIC)
-int net_hostname_set(char *host, size_t len);
+int net_hostname_set(const char *host, size_t len);
 #else
 static inline int net_hostname_set(char *host, size_t len)
 {
diff --git a/subsys/net/hostname.c b/subsys/net/hostname.c
index 7b85b4782a8..bf7a4329add 100644
--- a/subsys/net/hostname.c
+++ b/subsys/net/hostname.c
@@ -43,7 +43,7 @@ const char *net_hostname_get(void)
 }
 
 #if defined(CONFIG_NET_HOSTNAME_DYNAMIC)
-int net_hostname_set(char *host, size_t len)
+int net_hostname_set(const char *host, size_t len)
 {
        if (len > NET_HOSTNAME_MAX_LEN) {
                return -ENOMEM;
```

ajkontra:
When calling net_hostname_set() from C++, you will hit compile errors if you attempt to use a const char *. Since the internals of net_hostname_set() just uses memcpy(), we should pass in the new hostname as a const char * to better support C++.

Tested using samples/cpp/hello_world, with an added call to net_hostname_set().

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/99873

### DIFF
--- a/include/zephyr/net/hostname.h
+++ b/include/zephyr/net/hostname.h
@@ -76,9 +76,9 @@ static inline const char *net_hostname_get(void)
  * @return 0 if ok, <0 on error
  */
 #if defined(CONFIG_NET_HOSTNAME_DYNAMIC)
-int net_hostname_set(char *host, size_t len);
+int net_hostname_set(const char *host, size_t len);
 #else
-static inline int net_hostname_set(char *host, size_t len)
+static inline int net_hostname_set(const char *host, size_t len)
 {
 	ARG_UNUSED(host);
 	ARG_UNUSED(len);
--- a/subsys/net/hostname.c
+++ b/subsys/net/hostname.c
@@ -43,7 +43,7 @@ const char *net_hostname_get(void)
 }
 
 #if defined(CONFIG_NET_HOSTNAME_DYNAMIC)
-int net_hostname_set(char *host, size_t len)
+int net_hostname_set(const char *host, size_t len)
 {
 	if (len > NET_HOSTNAME_MAX_LEN) {
 		return -ENOMEM;


include/zephyr/net/hostname.h
```diff
@@ -76,7 +76,7 @@ static inline const char *net_hostname_get(void)
  * @return 0 if ok, <0 on error
  */
 #if defined(CONFIG_NET_HOSTNAME_DYNAMIC)
-int net_hostname_set(char *host, size_t len);
+int net_hostname_set(const char *host, size_t len);
 #else
 static inline int net_hostname_set(char *host, size_t len)
```
pdgendt: This line needs to be updated as well then.

rlubos (approved):

pdgendt (approved):

jukkar (approved):

PR #73042, opened 2024-05-20
twister: fix serial connection for flash_before

szczys:
The --flash-before flag allows devices to be flashed before the serial connection is established. However, the implementation was incomplete and only worked if the port address already existed at the start of the run. This is incompatible with devices that implement the USB in software (eg: USB-CDC).

This commit fixes the implementation to delay setting up the connection until after the device is flashed, and to retry the connection for two seconds after flashing to give the device time to enumerate as a USB device.

I considered making a separate twister flag for this behavior, to be used in conjunction with --flash-before. But on further consideration I think this is exactly the behavior originally envisioned, where there is no attempt whatsoever to connect to the serial port until after the firmware being tested has been flashed to the device.

This builds on the work of @LucasTambor in #66534 and addresses some of the issues I saw @JPHutchins discussing in https://github.com/zephyrproject-rtos/zephyr/pull/57660#issuecomment-1545167109.

PR #99912, opened 2025-11-24
pytest: Fix shell tests on IOTDK and HSDK

### LINKED ISSUE #99911: pytest: Shell tests fail on IOTDK and HSDK boards
### Describe the bug

Pytest shell tests consistently fail on IOTDK and HSDK boards:
- `sample.shell.shell_module`
- `sample.sensor.shell.pytest`

**HSDK board:**
- **Occurs:** Immediately after flashing, when test tries to read from serial port
- **Root Cause:** After flashing, the USB-to-serial device resets, invalidating the existing serial connection file descriptor

**IOTDK board:**
- **Occurs:** After several seconds of operation, board stops responding to UART input
- **Root Cause:** Hardware limitation - UART peripheral is in a power domain that gets gated when CPU enters sleep mode

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell
DEBUG: Get device type "hardware"
DEBUG: Opening serial connection for /dev/serial-board-...
DEBUG: rff: resetting device #5 (SN ...)... OK
DEBUG: Flashing command: /opt/python/venv/bin/west flash --skip-rebuild --build-dir /verification_workspace/twister-out/iotdk_arc_iot/zephyr/samples/subsys/shell/shell_module/sample.shell.shell_module -- --serial=...
DEBUG: Flashing finished
DEBUG: Found matching key: CONFIG_SHELL_PROMPT_UART="uart:~$ "
INFO: Wait for prompt
ERROR
------
```

### Impact

### Environment

- OS: Linux
- Toolchain: Zephyr SDK
- Zephyr version: main c9b01e8257dbe54e0f3b765f89fb69ffbafa201f

### Additional Context

#36254 - Original IOTDK UART issue report

M-Moawad:
This change adds a board-level `testing.flash_before flag` in Twister so the pytest/shell harness flashes the board before opening the serial port (enabled on snps/hsdk, snps/hsdk4xd, and snps/iotdk). It also introduces a configurable `CONFIG_ARC_IOT_CUSTOM_CPU_IDLE` option for the arc_iot SoC that skips sleep to keep peripherals powered when needed.

Why:

Avoid USB CDC disconnects during flashing on ARC development boards
Prevent UART power loss on iotdk when entering sleep

Fixes: #99911

### DIFF
--- a/boards/snps/hsdk/hsdk.yaml
+++ b/boards/snps/hsdk/hsdk.yaml
@@ -9,6 +9,7 @@ toolchain:
 supported:
   - smp
 testing:
+  flash_before: true
   ignore_tags:
     - net
     - bluetooth
--- a/boards/snps/hsdk/hsdk_arc_hsdk_2cores.yaml
+++ b/boards/snps/hsdk/hsdk_arc_hsdk_2cores.yaml
@@ -9,6 +9,7 @@ toolchain:
 supported:
   - smp
 testing:
+  flash_before: true
   ignore_tags:
     - net
     - bluetooth
--- a/boards/snps/hsdk4xd/hsdk4xd.yaml
+++ b/boards/snps/hsdk4xd/hsdk4xd.yaml
@@ -9,6 +9,7 @@ toolchain:
 supported:
   - smp
 testing:
+  flash_before: true
   ignore_tags:
     - net
     - bluetooth
--- a/boards/snps/iotdk/iotdk.yaml
+++ b/boards/snps/iotdk/iotdk.yaml
@@ -7,6 +7,7 @@ toolchain:
   - cross-compile
 ram: 128
 testing:
+  flash_before: true
   ignore_tags:
     - net
     - bluetooth
--- a/doc/develop/test/twister.rst
+++ b/doc/develop/test/twister.rst
@@ -223,6 +223,11 @@ testing:
     power-efficient but slow CPU or simulation platform which can perform instruction accurate
     simulation but does it slowly.
 
+  flash_before: [True|False] (default False)
+    For pytest/shell harness hardware testing, flash the device before opening the serial port.
+    This prevents serial port disconnection issues during flashing on some boards (e.g., those
+    with USB CDC that reset during flash operations).
+
 env:
   A list of environment variables. Twister will check if all these environment variables are set,
   and otherwise skip this platform. This allows the user to define a platform which should be
--- a/samples/sensor/sensor_shell/sample.yaml
+++ b/samples/sensor/sensor_shell/sample.yaml
@@ -23,6 +23,7 @@ tests:
     timeout: 180
     extra_configs:
       - arch:posix:CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT=y
+      - platform:iotdk/arc_iot:CONFIG_ARC_IOT_CUSTOM_CPU_IDLE=y
     extra_args: DTC_OVERLAY_FILE=fake_sensor.overlay
     integration_platforms:
       - native_sim
--- a/samples/subsys/shell/shell_module/sample.yaml
+++ b/samples/subsys/shell/shell_module/sample.yaml
@@ -13,6 +13,7 @@ tests:
       - native_sim
     extra_configs:
       - arch:posix:CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT=y
+      - platform:iotdk/arc_iot:CONFIG_ARC_IOT_CUSTOM_CPU_IDLE=y
   sample.shell.shell_module.usb:
     depends_on: usbd
     tags:
--- a/samples/subsys/smf/hsm_psicc2/sample.yaml
+++ b/samples/subsys/smf/hsm_psicc2/sample.yaml
@@ -15,5 +15,7 @@ common:
       - ".*<inf> hsm_psicc2_thread: s2_entry.*"
       - ".*<inf> hsm_psicc2_thread: s21_entry.*"
       - ".*<inf> hsm_psicc2_thread: s211_entry.*"
+  extra_configs:
+    - platform:iotdk/arc_iot:CONFIG_ARC_IOT_CUSTOM_CPU_IDLE=y
 tests:
   sample.smf.hsm_psicc2: {}
--- a/samples/subsys/testsuite/pytest/shell/testcase.yaml
+++ b/samples/subsys/testsuite/pytest/shell/testcase.yaml
@@ -7,6 +7,7 @@ common:
   filter: CONFIG_SERIAL and not CONFIG_SMP and dt_chosen_enabled("zephyr,shell-uart")
   extra_configs:
     - arch:posix:CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT=y
+    - platform:iotdk/arc_iot:CONFIG_ARC_IOT_CUSTOM_CPU_IDLE=y
   min_ram: 40
   integration_platforms:
     - native_sim
--- a/scripts/pylib/twister/twisterlib/harness.py
+++ b/scripts/pylib/twister/twisterlib/harness.py
@@ -500,8 +500,11 @@ class Pytest(Harness):
         if hardware.post_script:
             command.append(f'--post-script={hardware.post_script}')
 
-        if hardware.flash_before:
-            command.append(f'--flash-before={hardware.flash_before}')
+        # Check flash_before from both hardware map and platform (board YAML)
+        # Platform flash_before is intended for boards with USB reset issues during flashing
+        flash_before = hardware.flash_before or self.instance.platform.flash_before
+        if flash_before:
+            command.append(f'--flash-before={flash_before}')
 
         for fixture in hardware.fixtures:
             command.append(f'--twister-fixture={fixture}')
--- a/scripts/pylib/twister/twisterlib/platform.py
+++ b/scripts/pylib/twister/twisterlib/platform.py
@@ -72,6 +72,7 @@ class Platform:
         self.ignore_tags = []
         self.only_tags = []
         self.default = False
+        self.flash_before = False
         # if no flash size is specified by the board, take a default of 512K
         self.flash = 512
         self.supported = set()
@@ -116,6 +117,7 @@ class Platform:
         self.default = testing.get("default", self.default)
         self.binaries = testing.get("binaries", [])
         self.timeout_multiplier = testing.get("timeout_multiplier", self.timeout_multiplier)
+        self.flash_before = testing.get("flash_before", self.flash_before)
 
         # testing data for variant
         testing_var = variant_data.get("testing", data.get("testing", {}))
@@ -124,6 +126,7 @@ class Platform:
         self.only_tags = testing_var.get("only_tags", self.only_tags)
         self.default = testing_var.get("default", self.default)
         self.binaries = testing_var.get("binaries", self.binaries)
+        self.flash_before = testing_var.get("flash_before", self.flash_before)
         renode = testing.get("renode", {})
         self.uart = renode.get("uart", "")
         self.resc = renode.get("resc", "")
--- a/scripts/schemas/twister/platform-schema.yaml
+++ b/scripts/schemas/twister/platform-schema.yaml
@@ -104,6 +104,9 @@ schema;platform-schema:
           required: false
         "default":
           type: bool
+        "flash_before":
+          type: bool
+          required: false
         "binaries":
           type: seq
           seq:
--- a/soc/snps/arc_iot/CMakeLists.txt
+++ b/soc/snps/arc_iot/CMakeLists.txt
@@ -7,4 +7,6 @@ zephyr_sources(
   sysconf.c
   )
 
+zephyr_sources_ifdef(CONFIG_ARC_IOT_CUSTOM_CPU_IDLE soc_idle.c)
+
 set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/linker.ld CACHE INTERNAL "")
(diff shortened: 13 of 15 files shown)


scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
```diff
@@ -148,6 +148,11 @@ def _flash_and_run(self) -> None:
                 self._run_custom_script(self.device_config.post_flash_script, self.base_timeout)
             if process is not None and process.returncode == 0:
                 logger.debug('Flashing finished')
+                # Reconnect serial after flash (USB device was reset)
+                if self._serial_connection and self._serial_connection.is_open:
+                    self._serial_connection.close()
+                    time.sleep(0.5)
+                    self._connect_device()
```
gchwier: have you tried `--flash-before` argument? See https://github.com/zephyrproject-rtos/zephyr/pull/73042

arch/arc/core/cpu_idle.S
```diff
@@ -43,12 +43,23 @@ SECTION_FUNC(TEXT, arch_cpu_idle)
...
 #endif
 
+/*
+ * Workaround for iotdk board. When iotdk board enter sleep mode,
+ * some peripherals (eg: uart) will power off and board can't
+ * wake up through peripherals interrupt.
+ */
+#ifdef CONFIG_BOARD_IOTDK
+	j_s.d [blink]
+	seti 0
```
ruuddw: What's the purpose of this seti here?

M-Moawad: The `arch_cpu_idle` function is expected to always return with interrupts enabled`This function always exits with interrupts unlocked.`.

On normal boards, the sleep instruction (with bit 4 set) takes care of this by enabling interrupts during sleep and keeping them enabled after wake-up.

For IOTDK, we skip sleep to keep UART powered, but we still need to make sure interrupts are turned back on before returning. That’s why `seti 0` is required.

**Note:** Removing `seti 0` caused tests to **fail** because interrupts stayed disabled.

scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
M-Moawad: Thank you for your feedback!

The `--flash-before` flag fixes the two shell tests on boards but causes multiple other tests to fail because it skips reading early boot messages that many tests depend on.
 
 Should we run these two tests separately from the full regression suite, or is there a better approach?
 
My reconnection patch addresses the stale serial connection issue without changing test semantics . Do you think this approach is preferable to using `--flash-before` globally?

gchwier: IMO it’s not a good idea to apply the patch to the common code, as it’s only needed for a couple of tests. It’s better to apply the changes only to the tests that require them.

You can add a separate pytest scenario for those two tests and in that scenario you can modify flag, before flashing,
e.g.:
```python
import pytest
from twister_harness.device.device_adapter import DeviceAdapter
from twister_harness.fixtures import determine_scope

@pytest.fixture(scope=determine_scope)
def flash_before(device_object: DeviceAdapter):
    backup_flash_before = device_object.device_config.flash_before
    device_object.device_config.flash_before = True
    yield
    device_object.device_config.flash_before = backup_flash_before


def test_with_flash_before(flash_before, dut: DeviceAdapter):
    pass
```

you can also try to override `dut` fixture or use `unlaunched_dut` to change flag before launching.

M-Moawad: Thank you for suggesting the use of pytest fixtures. I tested your approach and confirmed that it works well for tests using `harness: pytest`. However, it doesn’t apply to tests with `harness: shell`, as the shell harness bypasses pytest fixtures entirely.

I found that adding `pytest_args: [--flash-before=true]` in the YAML configuration works for both harness types. For example:

*   `sample.sensor.shell.pytest` uses `harness: pytest`
*   `sample.shell.shell_module` uses `harness: shell`
*   `samples/subsys/testsuite/pytest/shell` (which I initially missed in the issue) mixes both harness types

This makes the YAML approach more universal since it addresses both pytest and shell harness tests with a single configuration change. I also tested on boards that previously passed without the `flash-before` flag, and they still pass after adding it.

Do you think adding `--flash-before` will have any unintended impact on those tests or boards?

M-Moawad: I see that some tests are starting to fail on the qemu_riscv64/qemu_virt_riscv64 target, so I will create separate test scenarios for ARC to address this.

M-Moawad: @gchwier Could you review this again?

ruuddw (dismissed):

scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
gchwier: Creating separate tests / configurations for ARC with `--flash-before` is ok .. However I'm not sure if you want to create a separate test for every possible configuration (I mean, that you probably do not need a separate cases for `vt100_colors_off`).

M-Moawad: @gchwier 
Yes, `--flash-before` is required for all specified test scenarios for ARC targets (hsdk, hsdk/arc_hsdk/2cores, hsdk4xd, and iotdk) because they fail on CI without it. 

For `vt100_colors_off`, I confirmed it’s specifically needed for `iotdk`; removing the flag causes the prompt wait to fail after flashing:
```bash

============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-8.4.2, pluggy-1.6.0 -- /opt/python/venv/bin/python3
cachedir: .pytest_cache
rootdir: /verification_workspace
collecting ... collected 1 item
zephyr/scripts/pylib/shell-twister-harness/test_shell.py::test_shell_harness
-------------------------------- live log setup --------------------------------
DEBUG: Get device type "hardware"
DEBUG: Opening serial connection for /dev/serial-...
DEBUG: rff: resetting device #5 (SN ...)... OK
DEBUG: Flashing command: /opt/python/venv/bin/west flash --no-rebuild --build-dir /verification_workspace/twister-out/iotdk_arc_iot/zephyr/samples/subsys/testsuite/pytest/shell/sample.harness.shell.vt100_colors_off -- --serial=...
DEBUG: Flashing finished
DEBUG: Found matching key: CONFIG_SHELL_PROMPT_UART="uart:~$ "
INFO: Wait for prompt
ERROR
------------------------------ live log teardown -------------------------------
DEBUG: Closed serial connection for /dev/serial-...
```

gchwier: You fixed it only for those couple of pytest scenarios. What about other tests where ARC target can be used?
I see that `flash_before` is also used for non-pytest harnesses, when `serial_pty` is used (see https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/pylib/twister/twisterlib/handlers.py#L783).
I have now also concerns to use `pytest_args: [--flash-before=true]`. If you run it with `--flash-before` in args, or with `flash_before` in hardware map, then you will send it to pytest twice.

You can try to update method `get_hardware`:
https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/pylib/twister/twisterlib/handlers.py#L703
and add befeore return:
```
        if self.instance.platform.arch == "arc":
            hardware.flash_before = True
```
I do not know if it is the best place to update `flash_before` parameter, but there you can read `platform.arch`. And with that change you do not need to update any test ... flash_before will be set for every ARC target, not needed to add any args.

M-Moawad: I tested your suggestion of updating method `get_hardware` to set flash_before=True on ARC, but it causes most ARC tests to fail because the serial connection is established only after flashing, so early boot and application output are missed.

Pytest and shell harnesses handle flash_before correctly (they wait for the prompt after flashing), so my YAML changes scope the flag to ARC scenarios using those harnesses—the ones affected by the USB reset—without breaking other tests. Only the ARC pytest/shell scenarios affected by the USB reset are targeted; other tests are already filtered out for ARC.

If we want a more generic solution that doesn’t require per-test YAML, we could centralize the logic to set flash_before only when arch == "arc" and the harness is pytest or shell.

I’m not sure how to apply this condition `(arch == "arc" and harness in ["pytest", "shell"])`. Do you have a recommendation on where and how to implement it?

ruuddw: > I tested your suggestion of updating method get_hardware to set flash_before=True on ARC

I don't think this should be done for all ARC architecture based devices. The issue sounds to me to be about the way Synopsys ARC development systems (boards) handle jtag via USB reset. So please don't make "flash_before" true for all arch/ARC. Can it be made a board yaml option picked up by the test harness?

gchwier: @M-Moawad it would be better to have it more generic, but if it works only with pytest (and pytest related like shell), you can apply filters in `_generate_parameters_for_hardware` method:
https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/pylib/twister/twisterlib/harness.py#L452

@ruuddw only ARC and only Synopsys ... maybe can be filtered with a vendor name.
You can try with:
```
        if self.instance.platform.arch == "arc" and self.instance.platform.vendor == "snps":
            hardware.flash_before = True
```

ruuddw: Synopsys also has simulator boards (nSIM), I'd still prefer some board-level solution. Future boards may handle jtag reset differently as well.

M-Moawad: I've implemented the **board-level `flash_before` YAML solution** as suggested by @ruuddw. Here's what was done:

**Changes:**
1. **`platform-schema.yaml`**: Added `flash_before` field to the platform schema (testing section)
2. **`platform.py`**: Load `flash_before` from board YAML files
3. **handlers.py**: Added `_get_flash_before()` helper method that:
   - Checks hardware map setting (Highest priority) (To address @gchwier concern "If you run it with --flash-before in args, or with flash_before in hardware map, then you will send it to pytest twice.")
   - Checks board YAML setting **only for pytest/shell harnesses**
   - Returns effective `flash_before` value without modifying the shared DUT object
4. **harness.py**: Updated `_generate_parameters_for_hardware()` to pass `--flash-before` flag to pytest when either hardware map or board YAML has it set
5. **Board YAML files**: Added `flash_before: true` for `hsdk`, `hsdk/arc_hsdk/2cores`, `hsdk4xd`, and `iotdk` boards

arch/arc/core/cpu_idle.S
```diff
@@ -43,12 +43,23 @@ SECTION_FUNC(TEXT, arch_cpu_idle)
 	POPR blink
 #endif
 
+/*
+ * Workaround for iotdk board. When iotdk board enter sleep mode,
+ * some peripherals (eg: uart) will power off and board can't
+ * wake up through peripherals interrupt.
+ */
+#ifdef CONFIG_BOARD_IOTDK
```
ruuddw: Looking at this again, I don't think this is the proper place for this workaround.
Please use "CONFIG_ARCH_HAS_CUSTOM_CPU_(ATOMIC)_IDLE" and override the function at SoC (or board) level.

ruuddw (changes requested):
See comments, please move the custom sleep from arch/arc to soc.

arch/arc/core/cpu_idle.S
M-Moawad: Added the SoC level implementation, please review

M-Moawad: I see some failed checks here on the CI, and I am checking them

scripts/pylib/twister/twisterlib/handlers.py
gchwier, line 1: All changes in this file are not needed, because for pytest (and shell) harness you do not use this part of code, flashing and connecting to UART is done in pytest-harness.

gchwier (commented):
> To address @gchwier concern "If you run it with --flash-before in args, or with flash_before in hardware map, then you will send it to pytest twice.

My concern was to use `--pytest-args` (but you removed those changes).
Please clean up commits (just remove instead of revert).

scripts/pylib/twister/twisterlib/handlers.py
M-Moawad: Cleaned up the commits and removed changes of handlers.py

gchwier: consider also updating docs with `flash_before` parameter:
https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/develop/test/twister.rst#L201
https://docs.zephyrproject.org/latest/develop/test/twister.html#board-configuration -> testing

M-Moawad: Docs updated with `flash_before` parameter

gchwier (dismissed):
The Twister part looks OK

NIT: however, it would be better to review and test the logic here:
https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/pylib/twister/twisterlib/handlers.py#L741
to ensure that `flash_before` is also applied to non‑pytest tests (this could be handled by someone else in a separate PR if there is a need for it)

M-Moawad:
Hi @gchwier,

The flash_before commit is unchanged from your previous approval. I've updated the iotdk sleep/idle commit, which removed your approval. Could you please take another look?

M-Moawad:
Hi @ruuddw,

 I've updated the iotdk sleep/idle commit. Could you please take another look?

ruuddw (approved):
Thanks, all looks good to me now. CI seems to be some infrastructure issue ("No space left on device").

gchwier (approved):

PR #99917, opened 2025-11-24
boards: stm32h7s78_dk: add FDCAN1 support

### LINKED ISSUE #99752: boards: st: stm32h7s78_dk: add optional FDCAN support
### Summary

The STM32H7S78-DK board under Zephyr currently doesn't support FDCAN or better said doesn't have it enabled by default.

This is due to FDCAN1 and FDCAN2 pins already being used by something else.

In the table bellow I will list all the pins that can be used for FDCAN1 and FDCAN2 and what already uses them.
Resources for information this are [stm32h7s78_dk-common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi), [stm32h7s7l8hxh-pinctrl.dtsi](https://github.com/zephyrproject-rtos/hal_stm32/blob/main/dts/st/h7/stm32h7s7l8hxh-pinctrl.dtsi) and [board's datasheet](https://www.st.com/resource/en/datasheet/stm32h7s3a8.pdf) (Table 22, Pages 132-137).

**FDCAN1:**
| Pin            | Functionality   | State                  |
| :---------- | :--------------- | :----------------- |
| PA11         | RX                   | Taken by `ltdc`   |
| PB8           | RX                   | *free*                 |
| PD0          | RX                   | Taken by `uart4` |
| PA12        | TX                    | Taken by `ltdc`   |
| PB9          | TX                    | Taken by `i2c1`   |
| PB1          | TX                    | Taken by `uart4` |

**FDCAN2:**
| Pin            | Functionality   | State                           |
| :---------- | :--------------- | :----------------------- |
| PB5          | RX                   | Taken by `timer3`        |
| PB12        | RX                   | Taken by `ltdc`             |
| PB1          | TX                    | *free*<sup>*</sup> |
| PB6          | TX                    | Taken by `i2c1`           |
| PB13        | TX                    | Taken by `ltdc`           |

**Taken by `mac` in #99296*

I propose an addition of something like this (of course adapted to this board):
https://github.com/zephyrproject-rtos/zephyr/blob/db0c34d836aeb79ddbdff0b219499009eebc59d9/boards/waveshare/open103z/waveshare_open103z.dts#L160-L169
to the [stm32h7s78_dk-common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi).

This way the user can enable FDCAN without needing to first find the pins in datasheet or pinctrl file, which makes it easier and less complex. The user can add FDCAN support with an overlay like this:
https://github.com/zephyrproject-rtos/zephyr/blob/f1c2ce4f4026bd5b212c6e4cc7d754aba69b189e/samples/drivers/can/counter/boards/waveshare_open103z.overlay#L7-L15

The main question is what to disable, aka which pins to select for FDCAN1 or FDCAN2.

- `ltdc` pins are used for display.
- `i2c1` pins are used for touch, for the display.
- `uart4` is used for console and shell.

Note that I already successfully tested the [can/counter](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/drivers/can/counter) sample with FDCAN1 using `pa11` and `pa12` pins, with display not in use.

Would an addition like this be good? Or does it indirectly introduce more complexity into [stm32h7s78_dk-common.dtsi](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi)?

If this addition is welcome, I would be happy to open a PR for this, with specified pins.

bklaric1:
## Changes
- Added FDCAN1 configuration to `stm32h7s78_dk-common.dtsi` and disabled it by default. Also explained that in order to use FDCAN1, the user needs to disable `i2c1`, since they share the `pb9` pin.
- ~Updated `stm32h7s78_dk.yaml` and `stm32h7s78_dk_stm32h7s7xx_ext_flash_app.yaml` to include the CAN feature.~
- Updated `index.rst` to include the CAN feature and documented the steps needed in order to use it.

## Tests
I tested these changes successfully with the following example: [can/counter](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/drivers/can/counter)

## Notes
Closes #99752

### DIFF
--- a/boards/st/stm32h7s78_dk/doc/index.rst
+++ b/boards/st/stm32h7s78_dk/doc/index.rst
@@ -180,6 +180,7 @@ Default Zephyr Peripheral Mapping:
 - USB OTG FS DM/DP : PM12/PM11
 - XSPI1 NCS/DQS0/DQS1/CLK/IO: PO0/PO2/PO3/PO4/PP0..15
 - I2C1 SCL/SDA: PB6/PB9
+- FDCAN1 RX/TX : PB8/PB9
 
 System Clock
 ------------
@@ -199,6 +200,13 @@ USB
 
 STM32H7S78-DK Discovery board has 2 USB Type-C connectors. Currently, only
 USB port2 (FS) is supported.
+
+FDCAN
+-----
+
+STM32H7S78-DK Discovery board has 2 FDCAN bus interfaces.
+FDCAN1 is configured but not enabled by default. To enable it, make sure
+that ``i2c1`` is disabled, since they share the PB9 pin.
 
 Programming and Debugging
 *************************
--- a/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
+++ b/boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
@@ -18,6 +18,7 @@
 		zephyr,sram = &sram0;
 		zephyr,display = &ltdc;
 		zephyr,touch = &display_ctp;
+		zephyr,canbus = &fdcan1;
 	};
 
 	psram: memory@90000000 {
@@ -202,6 +203,11 @@
 &i2c1 {
 	pinctrl-0 = <&i2c1_scl_pb6 &i2c1_sda_pb9>;
 	pinctrl-names = "default";
+	/*
+	 * Make sure FDCAN and i2c1 are not enabled at the same time
+	 * because they share the pb9 pin
+	 * reference: DS14359 Rev 4 page 134
+	 */
 	status = "okay";
 
 	display_ctp: gt911@5d {
@@ -210,6 +216,16 @@
 		irq-gpios = <&gpioe 3 (GPIO_ACTIVE_HIGH | GPIO_PULL_DOWN)>;
 		status = "okay";
 	};
+};
+
+&fdcan1 {
+	pinctrl-0 = <&fdcan1_rx_pb8 &fdcan1_tx_pb9>;
+	pinctrl-names = "default";
+	/*
+	 * fdcan1 is disabled by default in favor of i2c1 due to pb9 pin conflict
+	 * reference: DS14359 Rev 4 page 134
+	 */
+	status = "disabled";
 };
 
 &xspi1 {


boards/st/stm32h7s78_dk/stm32h7s78_dk.yaml
```diff
@@ -16,4 +16,5 @@ supported:
   - octospi
   - usbd
   - memc
+  - can
```
erwango: If so, then target will be automatically picked up by twister for CAN related tests and will fail building since the node will be missing.
Either you need to add systematically the overlay or remove this mention.
I would opt for the overlay.

Though, you probably don't have to to it both for this and _ext_app_flash variant. Current one would be enough

boards/st/stm32h7s78_dk/doc/index.rst
```diff
@@ -200,6 +201,13 @@ USB
 STM32H7S78-DK Discovery board has 2 USB Type-C connectors. Currently, only
 USB port2 (FS) is supported.
 
+FDCAN
+-----
+
+STM32H7S78-DK Discovery board has 2 FDCAN bus interfaces. Currently, only
+FDCAN1 is supported. It is not enabled by default. To enable it, make sure
```
erwango:
```suggestion
STM32H7S78-DK Discovery board has 2 FDCAN bus interfaces.
FDCAN1 configured but not enabled by default. To enable it, make sure
```
I think you should not get the one instance support sound as a support limitation, but rather as a configuration limitation.
Probably we need to review the USB mention above. I guess this was a real support limitation initially which has been solved meanwhile.

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
```diff
@@ -202,6 +203,11 @@
 &i2c1 {
 	pinctrl-0 = <&i2c1_scl_pb6 &i2c1_sda_pb9>;
 	pinctrl-names = "default";
+	/*
+	 * make sure FDCAN and i2c1 are not enabled at the same time
```
erwango: Nit pick
```suggestion
	 * Make sure fdcan1 and i2c1 are not enabled at the same time
```

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
```diff
@@ -212,6 +218,17 @@
 	};
 };
 
+&fdcan1 {
+	pinctrl-0 = <&fdcan1_rx_pb8 &fdcan1_tx_pb9>;
+	pinctrl-names = "default";
+	/*
+	 * make sure FDCAN and i2c1 are not enabled at the same time
+	 * because they share the pb9 pin
```
erwango:
```suggestion
	 * fdcan1 is disabled by default in favor of i2c1 due to pb9 pin conflict
```

erwango (changes requested):

boards/st/stm32h7s78_dk/doc/index.rst
bklaric1: Done. Had trouble with the wording for this one. Thanks.

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
bklaric1: Done.

boards/st/stm32h7s78_dk/stm32h7s78_dk-common.dtsi
bklaric1: Done.

djiatsaf-st (commented):
Could you please also test or provide support for the following tests?

- tests/drivers/can/api
- tests/drivers/can/shell
- tests/drivers/can/timing

These are usually the main tests we use to validate CAN peripherals.

boards/st/stm32h7s78_dk/stm32h7s78_dk.yaml
bklaric1: Oh, didn't think this through. You're right.
If I understand it correctly, I should add an overlay such as this: [waveshare_open103z.overlay](https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/can/counter/boards/waveshare_open103z.overlay)
to [CAN tests](https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/drivers/can)?
Or do you mean something else with:
> Either you need to add systematically the overlay or remove this mention.

bklaric1: @djiatsaf-st answered this indirectly with his [comment](https://github.com/zephyrproject-rtos/zephyr/pull/99917#pullrequestreview-3504550573) if I am not mistaken.

henrikbrixandersen: This change will cause this board to be picked up by _all_ samples + tests tagged with `can`, but overlays were only provided for a few of these.

If CAN is not available in the default board configuration, this cannot be added as a feature of the board.

tests/drivers/can/shell/boards/stm32h7s78_dk.overlay
henrikbrixandersen, line 1: This is really not needed. Testing the CAN shell module on this particular board provides no real benefit (and the test does not even use the onboard CAN controller).

henrikbrixandersen (changes requested):
I suggest leaving out `can` as a supported feature (since CAN is not enabled in the default board configuration) and thus also leaving out the test overlays (as they will never be used except in your own testing efforts).

tests/drivers/can/shell/boards/stm32h7s78_dk.overlay
bklaric1: > This is really not needed. Testing the CAN shell module on this particular board provides no real benefit (and the test does not even use the onboard CAN controller).

Yeah, I saw that after I pushed the commit. Sorry.

> I suggest leaving out `can` as a supported feature (since CAN is not enabled in the default board configuration) and thus also leaving out the test overlays (as they will never be used except in your own testing efforts).

Should I then remove the test overlays and the twister tag from the YAML file?

henrikbrixandersen: > Should I then remove the test overlays and the twister tag from the YAML file?

That would be my recommendation, yes.

boards/st/stm32h7s78_dk/stm32h7s78_dk.yaml
djiatsaf-st: > @djiatsaf-st answered this indirectly with his [comment](https://github.com/zephyrproject-rtos/zephyr/pull/99917#pullrequestreview-3504550573) if I am not mistaken.

From my point of view, adding the overlays would be better, and therefore, removing "can" from stm32h7s78_dk.yaml makes sense since there is not enough flash on this variant.

The goal is to run the tests on the H7S_DK board and ensure all tests work well. It should not be just about adding the overlays.

I suspect there is insufficient flash to perform these tests. Rather than adding the overlay for stm32h7s78_dk, I would prefer to add stm32h7s78_dk_stm32h7s7xx_ext_flash_app.overlay for the **can/timing and can/api** tests on this variant.

henrikbrixandersen: > The goal is to run the tests on the H7S_DK board and ensure all tests work well. It should not be just about adding the overlays.

The overlays make little sense if the board is not tagged with `can`, as twister will not pick up these tests without it.

tests/drivers/can/shell/boards/stm32h7s78_dk.overlay
bklaric1: Okay. I will amend my last commit to remove the added tests and tag.

henrikbrixandersen (dismissed):

erwango (dismissed):

etienne-lms (changes requested):
Could you squash the 2 commits?
I can remove my 'Request changes' feedback if @erwango and @henrikbrixandersen request so, but I think it's a bit ugly to have a commit right followed by a fixup change.

bklaric1:
Okay, no problems. I even offered to do it initially.
I will squash 2. commit into 1. commit.

boards/st/stm32h7s78_dk/doc/index.rst
```diff
@@ -180,6 +180,7 @@ Default Zephyr Peripheral Mapping:
 - USB OTG FS DM/DP : PM12/PM11
 - XSPI1 NCS/DQS0/DQS1/CLK/IO: PO0/PO2/PO3/PO4/PP0..15
 - I2C1 SCL/SDA: PB6/PB9
+- FDCAN1 RX/TX : PB8, PB9
```
FRASTM:
```suggestion
- FDCAN1 RX/TX : PB8/PB9
```

FRASTM (commented):
very minor remark
LGTM

boards/st/stm32h7s78_dk/doc/index.rst
bklaric1: Done.

erwango (approved):

henrikbrixandersen (approved):

etienne-lms (approved):

PR #98197, opened 2025-10-24
twister: Allow passing arguments to pre/post scripts via hardware-map

anilozrenk:
This pr extends Twister’s hardware-map functionality to support passing
arguments to user-defined scripts (pre_script, post_flash_script, and post_script).

Previously, Twister allowed defining script paths for DUT setup and teardown
but did not support passing additional arguments. With this change, users can
add argument sequences to their hardware-map entries, providing greater control
and flexibility for device-specific automation.

New optional fields introduced in `hwmap-schema.yaml` under `script_param`:
- pre_script_arg
- post_flash_script_arg
- post_script_arg

Each field accepts a list of string arguments which are appended to the
corresponding script invocation by Twister’s DeviceHandler.

Example usage in `hardware-map.yaml`:
```
- connected: true
  baud: 115200
  id: 000822008434
  flash_timeout: 60
  platform: my_board
  product: J-Link
  runner: jlink
  serial: /dev/ttyACM0
  pre_script: path_to/pre_script.sh
  script_param:
    pre_script_timeout: 100
    pre_script_arg:
      - argument1
      - arg2
      - 42
```

PR #99918, opened 2025-11-24
twister: add hardware map hook support

nashif:
This deprecates old support for custom scripts which was done for each
type of script individually and was duplicating code and adding too many
options for supporting and extending scripting.

This also does what #98197 is trying to do.

pushed as DRAFT to get early feedback while things are being reworked....

### DIFF
--- a/scripts/pylib/twister/twisterlib/hardwaremap.py
+++ b/scripts/pylib/twister/twisterlib/hardwaremap.py
@@ -14,6 +14,7 @@ from pathlib import Path
 import scl
 import yaml
 from natsort import natsorted
+import subprocess
 from twisterlib.environment import ZEPHYR_BASE
 
 try:
@@ -42,10 +43,7 @@ class DUT:
                  serial_pty=None,
                  connected=False,
                  runner_params=None,
-                 pre_script=None,
-                 post_script=None,
-                 post_flash_script=None,
-                 script_param=None,
+                 hooks=None,
                  runner=None,
                  flash_timeout=60,
                  flash_with_test=False,
@@ -59,17 +57,13 @@ class DUT:
         self._available = Value("i", 1)
         self._failures = Value("i", 0)
         self.connected = connected
-        self.pre_script = pre_script
+        self.hooks = hooks or {}
         self.id = id
         self.product = product
         self.runner = runner
         self.runner_params = runner_params
         self.flash_before = flash_before
         self.fixtures = []
-        self.post_flash_script = post_flash_script
-        self.post_script = post_script
-        self.pre_script = pre_script
-        self.script_param = script_param
         self.probe_id = None
         self.notes = None
         self.lock = Lock()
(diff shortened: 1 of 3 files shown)


scripts/pylib/twister/twisterlib/hardwaremap.py
```diff
@@ -125,6 +119,32 @@ def to_dict(self):
...
+    def run_hook(self, hook_type):
+        hook = None
+        for h in self.hooks:
+            if h.get('type', None) == hook_type:
+                hook = h
+        if not hook:
+            return
+
+        script = hook.get('script')
+        args = hook.get('args', [])
```
hakehuang, line 131: can we pass in more dynamic variable here, such as:
1. the current case name
2. the current debugger
3. current os.environments
such as the hook can do based on case/debugger/some other environments?

nashif: > can we pass in more dynamic variable here, such as:
> 
> 1. the current case name
> 2. the current debugger
> 3. current os.environments
>    such as the hook can do based on case/debugger/some other environments?

I guess we can pass some of this as environment variables that can be used by the script.

scripts/pylib/twister/twisterlib/hardwaremap.py
```diff
@@ -271,11 +300,41 @@ def load(self, map_file):
         hwm_schema = scl.yaml_load(self.schema_path)
         duts = scl.yaml_load_verify(map_file, hwm_schema)
         for dut in duts:
+            # deprecated fields are still supported for backward compatibility
             pre_script = dut.get('pre_script')
             script_param = dut.get('script_param')
             post_script = dut.get('post_script')
             post_flash_script = dut.get('post_flash_script')
+            # prepare hooks dictionary for forward compatibility
```
gchwier, line 308: nit: you can extract backward compatibility logic to separate method

something like:
```python
def _prepare_hooks_from_scripts(self, pre_script=None, post_script=None, post_flash_script=None, script_param=None):
    """Convert script parameters to hooks format for backward compatibility."""
    hooks = {}
    if pre_script:
        hooks['pre'] = {
            'script': pre_script,
            'args': [],
            'timeout': script_param.get('pre_script_timeout', 60) if script_param else 60
        }
    if post_script:
        hooks['post'] = {
            'script': post_script,
            'args': [],
            'timeout': script_param.get('post_script_timeout', 60) if script_param else 60
        }
    if post_flash_script:
        hooks['post_flash'] = {
            'script': post_flash_script,
            'args': [],
            'timeout': script_param.get('post_flash_timeout', 60) if script_param else 60
        }
    return hooks
```
you can use it also in `add_device` method and in unit/blackbox tests

gchwier (commented):
It also need to be updated in pytest-twister-harnes plugin, or at least should be passed in old format here:
https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/pylib/twister/twisterlib/harness.py#L499-L506

Would be good to change the way how Twister passes parameters to the pytest plugin. Instead of listing parameters in the command line, it would be better to serialize them into a JSON file .. then will be easier to pass more complex data structures (e.g. multiple DUTs for multi device testing scenarios) - however this is a plan for the future

PR #99924, opened 2025-11-24
Mspi dw format and data only support

Hi-Im-David:
Data only transfer:
The IP doesn't support a data only transfer (no cmd or addr). It is required instead that the first dataframe of the data buffer is sent as an address.

`current_packet` has been added as a variable in the dev data structure to create a copy of the packet so it can be modified and used during a data only transfer. The first dataframe can then be safely copied to the address without modifying the original packet passed through by the user.

Format registers:
Since the API takes in an input buffer of uint8_t, data will appear big-endian in memory. Since the system is little-endian, when for example DFS = 32, the order will come out in reverse order. Therefore, we can use the format registers and set BPP to 8 which results in the data being formatted to big-endian. Setting BPP to always be 8 ensures big-endian output, regardless of the DFS. These format registers are also used for display use-case and can be overwritten to match the BPP of your display in the devicetree

### DIFF
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -162,6 +162,9 @@ DEFINE_MM_REG_WR(xip_write_wrap_inst,	0x144)
 DEFINE_MM_REG_WR(xip_write_ctrl,	0x148)
 #endif
 
+/* Ceiling division by 32 */
+#define CEIL_DIV_32(x) (((x) + 31U) >> 5)
+
 #include "mspi_dw_vendor_specific.h"
 
 static int start_next_packet(const struct device *dev);
@@ -1288,15 +1291,15 @@ static int start_next_packet(const struct device *dev)
 #if defined(CONFIG_MSPI_DMA)
 	if (dev_data->xfer.xfer_mode == MSPI_DMA) {
 		/* For DMA mode, set start level based on transfer length to prevent underflow */
-		uint32_t total_transfer_bytes = packet->num_bytes + dev_data->xfer.addr_length +
-						dev_data->xfer.cmd_length;
-		uint32_t transfer_frames = total_transfer_bytes >> dev_data->bytes_per_frame_exp;
+		uint32_t transfer_frames = (packet->num_bytes >> dev_data->bytes_per_frame_exp) +
+					   CEIL_DIV_32(dev_data->xfer.addr_length)
+					 + CEIL_DIV_32(dev_data->xfer.cmd_length);
 
-		/* Use minimum of transfer length or FIFO depth, but at least 1 */
+		/* Above dma_start_level, the transfer will start.
+		 * Use minimum of transfer length and FIFO depth.
+		 */
 		uint8_t dma_start_level = MIN(transfer_frames - 1,
 					      dev_config->tx_fifo_depth_minus_1);
-
-		dma_start_level = (dma_start_level > 0 ? dma_start_level : 1);
 
 		/* Only TXFTHR needs to be set to the minimum number of frames */
 		write_txftlr(dev, FIELD_PREP(TXFTLR_TXFTHR_MASK, dma_start_level));
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -86,6 +86,8 @@ struct mspi_dw_data {
 #endif
 
 	struct mspi_xfer xfer;
+	struct mspi_xfer_packet current_packet;
+	bool has_cmd_addr;
 
 #if defined(CONFIG_MSPI_DW_HANDLE_FIFOS_IN_SYSTEM_WORKQUEUE)
 	struct k_work fifo_work;
@@ -1086,8 +1088,8 @@ static int start_next_packet(const struct device *dev)
 {
 	const struct mspi_dw_config *dev_config = dev->config;
 	struct mspi_dw_data *dev_data = dev->data;
-	const struct mspi_xfer_packet *packet =
-		&dev_data->xfer.packets[dev_data->packets_done];
+	dev_data->current_packet = dev_data->xfer.packets[dev_data->packets_done];
+	struct mspi_xfer_packet *packet = &dev_data->current_packet;
 	bool xip_enabled = COND_CODE_1(CONFIG_MSPI_XIP,
 				       (dev_data->xip_enabled != 0),
 				       (false));
@@ -1096,11 +1098,12 @@ static int start_next_packet(const struct device *dev)
 	uint32_t imr = 0;
 	int rc = 0;
 
-	if (packet->num_bytes == 0 &&
-	    dev_data->xfer.cmd_length == 0 &&
-	    dev_data->xfer.addr_length == 0) {
+	if (!dev_data->has_cmd_addr && !(packet->num_bytes > 0)) {
 		return 0;
 	}
+
+	const bool tx_data_only = !dev_data->has_cmd_addr && (packet->num_bytes > 0) &&
+				  (packet->dir == MSPI_TX);
 
 	dev_data->dummy_bytes = 0;
 	dev_data->bytes_to_discard = 0;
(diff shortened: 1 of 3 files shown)


swift-tk (dismissed):

drivers/mspi/mspi_dw.c
```diff
@@ -1086,8 +1088,8 @@ static int start_next_packet(const struct device *dev)
 {
 	const struct mspi_dw_config *dev_config = dev->config;
 	struct mspi_dw_data *dev_data = dev->data;
-	const struct mspi_xfer_packet *packet =
-		&dev_data->xfer.packets[dev_data->packets_done];
+	dev_data->current_packet = dev_data->xfer.packets[dev_data->packets_done];
+	struct mspi_xfer_packet *packet = &dev_data->current_packet;
```
anangl, line 1092: Copying and modifying the packets is confusing and error-prone (even this commit lacks the corresponding modification of `tx_data()`), and it is in fact unneeded. See #101689 which also adds handling of RX data-only packets.

anangl (changes requested):

dts/bindings/mspi/nordic,nrf-qspi-v2.yaml
```diff
@@ -6,3 +6,16 @@ description: Nordic QSPI v2 Interface using SSI IP
 compatible: "nordic,nrf-qspi-v2"
 
 include: snps,designware-ssi.yaml
+properties:
+  bits-per-pixel:
```
JarmouniA, line 10: This feels wrong, why would a generic xSPI binding have a display-specific property?

Could you give an example of the DT nodes in this setup?

JarmouniA (changes requested):

drivers/mspi/mspi_dw.c
Hi-Im-David: @anangl The copying of the packet is done because the `vendor_specific_start_dma_xfer()` function also needs to know the modified address value/length and number of data frames in the case of a data only transfer. In the case of your commit would all be stored as local variables in `start_next_packet()` and not in `dev_data`. To avoid duplication of code for the setup of a data-only transfer, I did it this way but if you would prefer I can add separate logic in the vendor_specific_start_dma_xfer function?

dts/bindings/mspi/nordic,nrf-qspi-v2.yaml
Hi-Im-David: Thanks for reviewing @JarmouniA.

For context, the bits-per-pixel setting actually programs a register in the MSPI controller and affects how TX data is reordered. So while the concept originates from the display, the configuration must ultimately be applied at the controller level but, in hind-sight, I agree that my approach goes against the Zephyr driver model.

One alternative could be to move bits-per-pixel to the display child node and have the controller driver read it. However, that would assume there is only one child device providing this value. If there were multiple children with different pixel formats, it would be unclear which one the controller should use.

Another approach could be to introduce an explicit MSPI API that allows a child device (display) to configure the bits-per-pixel setting at init or per transfer. @swift-tk, please could you provide some input here as well.

Thanks.

drivers/mspi/mspi_dw.c
anangl: > @anangl The copying of the packet is done because the `vendor_specific_start_dma_xfer()` function also needs to know the modified address value/length and number of data frames in the case of a data only transfer. In the case of your commit would all be stored as local variables in `start_next_packet()` and not in `dev_data`.

The thing is that there's no need to modify the packet at all and I don't do it in my PR. It is sufficient to just configure the non-zero length of the address. Since the address value and the packet data are written to the controller the same way, through the data register, you don't need to write the address value separately. You can just write the original packet data and the controller will use the first data frame as the address.

dts/bindings/mspi/nordic,nrf-qspi-v2.yaml
swift-tk: I had suggested using the already provided endian attribute. Would that work? I’m not sure how does the bits per pixel affects ordering here, could you provide more detail?

JarmouniA: > Could you give an example of the DT nodes in this setup?

@Hi-Im-David with the display controller node.

Hi-Im-David: @swift-tk Endianness alone cannot discard padding. For 24-bit pixels stored as uint32_t, a standard 32-bit transfer would clock out 8 extra bits of padding (2 additional cycles in Quad mode), immediately desyncing the display. The BPP setting triggers the hardware to strip that padding and treat 24 bits as the logical "unit," ensuring the physical SCLK timing and data packing on the wires match the display's requirements.

e.g
without BPP of 24-bit:  [R1][G1][B1][Pad][R2][G2][B2][Pad]
with BPP of 24-bit:  [R1][G1][B1][R2][G2][B2]

But I agree with your previous statement to add support for endian in this driver but that is separate to display support.

Hi-Im-David: @JarmouniA currently there is no display driver in zephyr that uses the MSPI API so I haven't had a chance to test it yet, instead this is tested by viewing the waveform directly at the moment but in the current state of this PR, it would look something like this:

```
&qspi00 {
    status = "okay";
    op-mode = "MSPI_OP_MODE_CONTROLLER";
    bits-per-pixel = <24>; 

    display-device@0 {
        compatible = "vendor,displayX";
        reg = <0>;
        /* Display parameters... * /
    };
};
```

I agree that the `bits-per-pixel` should probably be moved to be part of a child node or the MSPI API itself.

swift-tk: This is another good example where additional-property would be helpful if bits-per-pixel can not become a generic binding. You get to define it in controller child-binding and then pass it to device level. 

If you want it to be generic, consider something like data-format as the controller should not be concerned with what types of data it is transferring: whether it is a pixel or other stuff.  It should just be data in bytes.

I’d suggest add this support when you have coded a display driver using MSPI API and show usage then. I wonder what would the transfer data size in mspi_xfer_packet become? Is it still 4 bytes or 3 for one pixel?

I also would imagine that the display driver may not be shared with the controllers that are not capable of such data format as they would require repacking the data by software…. In that case, making it generic is hardly beneficial to the community…

drivers/mspi/mspi_dw.c
Hi-Im-David: Ah I had assumed the DMA transfer would require the transfer list to still have 2 entries, 1 for the address and 1 for the data but you are right. Thanks very much for pointing that out and creating a PR. I will remove my commit from this PR.

dts/bindings/mspi/nordic,nrf-qspi-v2.yaml
JarmouniA: > @JarmouniA currently there is no display driver in zephyr that uses the MSPI API so I haven't had a chance to test it yet, instead this is tested by viewing the waveform directly at the moment but in the current state of this PR, it would look something like this:
> 
> ```
> &qspi00 {
>     status = "okay";
>     op-mode = "MSPI_OP_MODE_CONTROLLER";
>     bits-per-pixel = <24>; 
> 
>     display-device@0 {
>         compatible = "vendor,displayX";
>         reg = <0>;
>         /* Display parameters... * /
>     };
> };
> ```
> 
> I agree that the `bits-per-pixel` should probably be moved to be part of a child node or the MSPI API itself.

If MSPI is used to emulate MIPI-DBI mode C, there is a specific way how that's done with normal SPI in Zephyr, see:
- `dts/bindings/mipi-dbi/zephyr,mipi-dbi-spi.yaml`, and the corresponding driver `drivers/mipi_dbi/mipi_dbi_spi.c`
- `dts/bindings/mipi-dbi/mipi-dbi-spi-device.yaml` (that's included in binding of a panel controller, ex. https://github.com/zephyrproject-rtos/zephyr/blob/f8dba1baee3627e55afeb6045f353701b674376e/dts/bindings/display/ilitek%2Cili9xxx-common.yaml#L12)
- 
ex. of usage in DT https://github.com/zephyrproject-rtos/zephyr/blob/f8dba1baee3627e55afeb6045f353701b674376e/boards/shields/x_nucleo_gfx01m2/x_nucleo_gfx01m2.overlay#L55

PR #99925, opened 2025-11-24
lora: remove `EXPERIMENTAL` from already unstable api

sophiekovalevsky:
Based on api lifecycle documentation, the lora driver has been recently moved to unstable. This means that the state of this driver is no longer experimental. Due this new state, the change should be updated on Kconfig.

This commit remove the `EXPERIMENTAL` symbol selection from the driver itself. It also remove the word `experimental` from the lora basic modem backend since the api lifecycle documentation only describes the state of the api based upon the api itself and not any particular backend implementation.

### DIFF
--- a/drivers/lora/Kconfig
+++ b/drivers/lora/Kconfig
@@ -7,9 +7,8 @@
 # Top-level configuration file for LORA drivers.
 
 menuconfig LORA
-	bool "LoRa drivers [EXPERIMENTAL]"
+	bool "LoRa drivers"
 	select POLL
-	select EXPERIMENTAL
 	help
 	  Include LoRa drivers in the system configuration.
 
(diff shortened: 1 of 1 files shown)


henrikbrixandersen (dismissed):

drivers/lora/Kconfig
```diff
@@ -23,12 +22,12 @@ config LORA_MODULE_BACKEND_LORAMAC_NODE
...
 
 config LORA_MODULE_BACKEND_LORA_BASICS_MODEM
-	bool "LoRa Basic modem backend [EXPERIMENTAL]"
+	bool "LoRa Basic modem backend"
 	depends on ZEPHYR_LORA_BASICS_MODEM_MODULE
 	depends on DT_HAS_SEMTECH_SX1262_ENABLED || DT_HAS_SEMTECH_SX1261_ENABLED || DT_HAS_SEMTECH_SX1272_ENABLED || DT_HAS_SEMTECH_SX1276_ENABLED
 	select USE_LORA_BASICS_MODEM_DRIVERS
 	help
-	  Experimental support for implementing the LoRa API using the LoRa Basics Modem module.
+	  Unstable support for implementing the LoRa API using the LoRa Basics Modem module.
```
nashif: "Unstable support" sounds very odd. Why do we need this? The help text of a Kconfig should just focus on what this config does, stability of an API should not be expressed here. The experimental part is about the kconfig being experimental, does not map to the API life cycle. (We have never done it for any other API this way)

nashif (changes requested):

drivers/lora/Kconfig
JordanYates:
```suggestion
	  LoRa API support using the LoRa Basics Modem module.
```

sophiekovalevsky: Thanks for clarifying. Suggestion made by @JordanYates seems to cover what the description should focus on.

nashif (approved):

henrikbrixandersen (approved):

mniestroj (approved):

JordanYates (approved):

PR #99927, opened 2025-11-24
net: mqtt_sn: Improve IPv6 support

aurel32:
This PR fixes various issues I've found in the MQTT-SN code when trying to get it working with IPv6. I can't split in in smaller chunks if needed

### DIFF
--- a/subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
+++ b/subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
@@ -152,8 +152,18 @@ static int tp_udp_init(struct mqtt_sn_transport *transport)
 	}
 
 	optval = CONFIG_MQTT_SN_LIB_BROADCAST_RADIUS;
-	err = zsock_setsockopt(udp->sock, NET_IPPROTO_IP, ZSOCK_IP_MULTICAST_TTL,
-			       &optval, sizeof(optval));
+	if (udp->bcaddr.sa_family == NET_AF_INET && IS_ENABLED(CONFIG_NET_IPV4)) {
+		err = zsock_setsockopt(udp->sock, NET_IPPROTO_IP,
+				       ZSOCK_IP_MULTICAST_TTL, &optval,
+				       sizeof(optval));
+	} else if (udp->bcaddr.sa_family == NET_AF_INET6 && IS_ENABLED(CONFIG_NET_IPV6)) {
+		err = zsock_setsockopt(udp->sock, NET_IPPROTO_IPV6,
+				       ZSOCK_IPV6_MULTICAST_HOPS, &optval,
+				       sizeof(optval));
+	} else {
+		LOG_ERR("Unknown AF");
+		return -EINVAL;
+	}
 	if (err < 0) {
 		return -errno;
 	}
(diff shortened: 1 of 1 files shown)


subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
```diff
@@ -152,8 +152,18 @@ static int tp_udp_init(struct mqtt_sn_transport *transport)
 	}
 
 	optval = CONFIG_MQTT_SN_LIB_BROADCAST_RADIUS;
-	err = zsock_setsockopt(udp->sock, NET_IPPROTO_IP, ZSOCK_IP_MULTICAST_TTL,
-			       &optval, sizeof(optval));
+	if (udp->bcaddr.sa_family == AF_INET && IS_ENABLED(CONFIG_NET_IPV4)) {
```
aurel32: I guess I can reduce the Cognitive Complexity by splitting `tp_udp_init()` into `tp_udp_init_ipv4()` and `tp_udp_init_ipv6()`. On the other hand that will increase the risk that the IPv4 and IPv6 versions diverge.

subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
```diff
@@ -63,7 +63,7 @@ static int tp_udp_init(struct mqtt_sn_transport *transport)
 	}
 
 	if (IS_ENABLED(CONFIG_MQTT_SN_LOG_LEVEL_DBG)) {
-		char ip[30], *out;
+		char ip[40], *out;
```
jukkar: Please replace the magic constant by `NET_INET6_ADDRSTRLEN` which contains the max length of the IPv6 address.

subsys/net/lib/mqtt_sn/Kconfig
```diff
@@ -38,7 +38,7 @@ config MQTT_SN_LIB_MAX_GATEWAYS
 
 config MQTT_SN_LIB_MAX_ADDR_SIZE
 	int "Maximum address size for the transport"
-	default 21
+	default 24
 	range 1 $(UINT8_MAX)
```
jukkar, line 42: This looks unnecessary kconfig option. First of all there is no information what the value actually means. After looking the code, this tells the sizeof the address structure (either `struct sockaddr_in` or `struct sockaddr_in6`). Why do we even ask this from the user as it is very difficult to understand what the value should be in the first place.
Better would be to select the value automatically depending on if the `CONFIG_NET_IPV6` is selected or not.
So I suggest that:
* We remove this Kconfig option
* Add relevant code in `mqtt_sn.c` to select the value automatically, for example like this

```
#define MAX_ADDR_SIZE COND_CODE_1(CONFIG_NET_IPV6, (sizeof(struct sockaddr_in6)), (sizeof(struct sockaddr_in)))
```
* and replace current `CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE` by `MAX_ADDR_SIZE`

jukkar (commented):
I think you can ignore the Sonar warning, the `tp_udp_init()` looks ok to me.

subsys/net/lib/mqtt_sn/Kconfig
aurel32: I think the reason for defining this value is because the MQTT-SN protocol can be used for any kind of message-based transport (e.g. people use it on Lora or serial port), even if Zephyr only provides a a UDP transport. Alternative transport can be provided by the app, in that case I believe a different size might be used.

Therefore I believe this option should be kept. If increasing that value to 24 is an issue, I can make it dependent on CONFIG_NET_IPV6, and maybe also on CONFIG_MQTT_SN_TRANSPORT_UDP.

subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
aurel32: Thanks for the hint, that's even bigger than I thought, it didn't come to my mind that mapped IPv4 addresses are even longer... That is now fixed.

rlubos (dismissed):

subsys/net/lib/mqtt_sn/Kconfig
jukkar: I am questioning here how the user would be able to figure out proper value for this Kconfig option. The help text talks about addresses but that is way too vague to be useful. Some examples would be nice to have here. Now user needs to go to the source code to look how this kconfig option is used to try to understand what to set here.
Also if UDP transport is selected, then depending on whether IPv6/4 is enabled, different size could be set. It would be extra helpful if some compile check could be added, like some `BUILD_ASSERT` that makes sure that the value is not too small.

aurel32: That's a good point indeed, and I started to work on that. That said in the process I realized that 21 bytes is actually fine for IPv6/UDP, that's the size of a `net_sockaddr_in6` struct. I am not sure how I ended-up with the conclusion it required 24 bytes in my previous testing... I have therefore dropped that commit from the PR, and I'll open a different one to clarify that config option and add a `BUILD_ASSERT`.

aurel32: This is now https://github.com/zephyrproject-rtos/zephyr/pull/100124

rlubos (approved):

jukkar (approved):

PR #99928, opened 2025-11-24
cmake: extensions: remove invalid add_dependencies

### LINKED ISSUE #99926: west spdx broken with CMake 4.2.0
### Describe the bug

Not clear yet if it's a CMake 4.2.0 bug, or something in Zephyr, but CMake segfaults when trying to follow https://docs.zephyrproject.org/latest/develop/west/zephyr-cmds.html#software-bill-of-materials-west-spdx

### Regression

- [ ] This is a regression.

### Steps to reproduce

```
west spdx --init -d BUILD_DIR 
west -v build -d BUILD_DIR -b wio_terminal samples/hello_world -- -DCONFIG_BUILD_OUTPUT_META=y
```

### Relevant log output

```shell
FATAL ERROR: command exited with status -11: /opt/homebrew/bin/cmake -DWEST_PYTHON=/Users/kartben/zephyrproject/.venv/bin/python3.13 -B/Users/kartben/zephyrproject/zephyr/BUILD_DIR -GNinja -DBOARD=wio_terminal -DCONFIG_BUILD_OUTPUT_META=y -S/Users/kartben/zephyrproject/zephyr/samples/hello_world
```

### Impact

### Environment

- macOS
- main @ 7a65336edc345a81508303ae97caf8fa176a5a9f
- CMake 4.2.0

### Additional Context

downgrading to cmake 4.1.3 gets everything back on track so there is a "workaround", I guess

kartben:
The `add_dependencies` command in CMake is used to specify dependencies between targets - here it was used to specify a dependency on a folder, which is not valid (and causes failures on CMake 4.2 when using the cmake-file-api, guess it's otherwise just silently ignored in earlier CMake versions).

This was introduced with commit 80a6a206289537fa220e34fe485b11fccef9ac85 - cc @gmarull

### DIFF
--- a/cmake/modules/extensions.cmake
+++ b/cmake/modules/extensions.cmake
@@ -1771,10 +1771,6 @@ function(zephyr_syscall_include_directories)
       syscalls_interface INTERFACE
       ${include_dir}
     )
-    add_dependencies(
-      syscalls_interface
-      ${include_dir}
-    )
 
     unset(include_dir)
   endforeach()
@@ -1793,10 +1789,6 @@ function(zephyr_syscall_header)
 
     target_sources(
       syscalls_interface INTERFACE
-      ${header_file}
-    )
-    add_dependencies(
-      syscalls_interface
       ${header_file}
     )
 


gmarull (commented):
Does change keep original functionality working?

kartben:
> Does change keep original functionality working?

I am afraid I don't know what the original functionality was supposed to be, since `zephyr_syscall_include_directories` is not really exercised in CI or used anywhere in-tree. I would think these add_dependencies are just silently ignored right now (and in any case I really don't think the full paths of header files are valid/declared CMake targets anyway?)

tejlmand (approved):

pdgendt (approved):

henrikbrixandersen (approved):

PR #92854, opened 2025-07-08
doc: twister: inline `twister --help` output into the docs

kartben:
In order to make it easier for folks to discover available twister options when using/searching the documentation, or when making external search in their favorite search engine, inline the output of `twister --help` into the docs so as to have an always up-to-date and searchable snapshot of it.

CI output [here](https://builds.zephyrproject.io/zephyr/pr/92854/docs/develop/test/twister.html#:~:text=The%20list%20of%20command%20line%20options%20supported%20by%20twister%20can%20be%20viewed%20using%3A)

image

PR #79754, opened 2024-10-12
doc: extensions: boards: add ability to filter by hw capability

kartben:
Augment the board catalog with the list of supported hardware features for each board, by building -- cmake only, as we only need to get to the point where we obtain a `zephyr.dts` -- hello world for all boards and the analyzing the contents of the resulting Devicetree.

The board index page search form has been updated to allow to filter by supported HW capability, e.g. what board supported in zephyr has "display + lora + i2s"

A follow-up PR will leverage the metadata to automatically populate the currently very manual, often stale, table of "Supported Features" that boards' READMEs typically include (folks might remember seeing a demo of this -- I decided to push it to a different PR to simplify review process)

image

Doc GH workflow running on pull requests will skip the build of the 800+ hello worlds as it is quite resource-heavy and only the scheduled build (running every 3 hours) will do it (similar to how we skip generation of the fulls docs of the devicetree bindings).

Related: due to this PR updating the doc-build.yml workflow, it's not publishing the CI output to builds.zephyrproject.io ; best way to test the actual user experience is to get the most recent successful build from my fork ([kartben/zephyr/actions/workflows/doc-build.yml?query=event%3Aschedule](https://github.com/kartben/zephyr/actions/workflows/doc-build.yml?query=event%3Aschedule)), download the html-output artifact, unzip it, and open index.html in a browser

PR #99931, opened 2025-11-24
doc: requirements: add missing junitparser, needed by twister

marc-hb:
Fixes commit cae12e6d6a7d5  ("doc: twister: inline `twister --help` output into the docs")

---------

cc:
- #92854

### DIFF
--- a/doc/requirements.in
+++ b/doc/requirements.in
@@ -20,6 +20,9 @@ pykwalify
 pytest
 pyserial
 
+# Since commit cae12e6d6a7d5, building docs runs "twister"
+junitparser
+
 # Doxygen doxmlparser
 doxmlparser
 


kartben (changes requested):
junitparser is already included when setting up a zephyr environment

"In order to install the documentation tools, first install Zephyr as described in Getting Started Guide."
 https://docs.zephyrproject.org/latest/contribute/documentation/generation.html#:~:text=In%20order%20to%20install%20the%20documentation%20tools%2C%20first%20install%20Zephyr%20as%20described%20in%20Getting%20Started%20Guide.

marc-hb:
A long time ago, it was possible to build docs without a compiler and a gazillion of `pip install --break-system-packages ...`. `doc-requirements.txt` used to fill that purpose.

To help move faster through the stages of grief, should the comment at the top of this file be more explicit about the massive dependencies? Very terse and pretty ambiguous right now:

```
# DOC: used to generate docs
```

marc-hb:
> Feel free to propose an update to the requirements file 

Thanks! submitted:
- #99933

> And yeah, a long time ago for sure: building the docs in a working zephyr environment has been a requirement for the past 7 years at least ... https://github.com/zephyrproject-rtos/zephyr/commit/0d811b9aee6a6410f73fdfb4d4072d4d16b05a52

This is only documentation. It's always been simpler and safer to ask people to install everything; no doubt about that. Much less support workload.

But in CI / reality, it looks like you changed this only 1 year ago. Quoting:

- https://github.com/zephyrproject-rtos/zephyr/pull/79754#issuecomment-2654340856 / b6922c83a8a5eace

> One thing that might be worth considering is that you now also need a full install of the SDK to build the docs without error. What I mean by this is that I get a lot of build errors on the helloworld sample that is used to create the devicetree simply because I do not have all toolchains installed (in my case I typically install arm and x86_64 and sometimes riscV but none of the others)....

That subset / ability was nice for doc writers and quick fixes... Never mind, the doc build has become too complex and too slow anyway, moving on...

marc-hb:
> BTW twister --help failures are not fatal: you only get a cryptic warning once. Then the include is silently missing. This seems typical with Sphinx (e.g. https://github.com/zephyrproject-rtos/zephyr/issues/99932)

Just in case you missed this (and someone cares)

PR #99934, opened 2025-11-25
Add sys_getopt_r, sys_getopt_long_r and sys_getopt_long_only_r

keith-packard:
Add getopt APIs inspired by the re-entrant newlib/picolibc versions. Instead of stashing getopt state in global or thread-local variables, have callers allocate space for them instead.

The existing APIs remain available and unchanged, but all in-tree users are migrated to the new APIs and the documentation
refers to the new ones.

As a part of this work, the POSIX getopt implementation has been aligned with the POSIX standard as closely as possible; one minor extension has been adapted from glibc/newlib/picolibc to make re-use of the API possible in testing.

### DIFF
--- a/include/zephyr/sys/sys_getopt.h
+++ b/include/zephyr/sys/sys_getopt.h
@@ -34,6 +34,9 @@ extern int sys_getopt_opterr;
 extern int sys_getopt_optind;
 extern int sys_getopt_optopt;
 
+/* Initialize local sys_getopt_state with this value */
+#define SYS_GETOPT_STATE_INITIALIZER	{0}
+
 #define sys_getopt_no_argument       0
 #define sys_getopt_required_argument 1
 #define sys_getopt_optional_argument 2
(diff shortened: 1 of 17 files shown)


jukkar (approved):
Thanks Keith! LGTM

include/zephyr/sys/sys_getopt.h
```diff
@@ -34,6 +34,9 @@ extern int sys_getopt_opterr;
 extern int sys_getopt_optind;
 extern int sys_getopt_optopt;
 
+/* Initialize local sys_getopt_state with this value */
+#define SYS_GETOPT_STATE_INITIALIZER	{0}
```
aescolar, line 38: Should this initialize the last 2 getopt_long members to -1 ?

aescolar (commented):
Overall this look nice to me.
@keith-packard note the commit title issue CI is complaining about.

include/zephyr/sys/sys_getopt.h
keith-packard: When `optind` is set to zero, each of the getopt functions will call `sys_getopt_init_r` to completely re-initialize
the struct, including setting `nonopt_start` and `nonopt_end` to -1. That means this initializer can just be all zeros.

peter-mitsis (approved):

jakub-uC (approved):

PR #99940, opened 2025-11-25
dma: rpi_pico: fix transfer count calculation

### LINKED ISSUE #99435: rpi_pico: Pico DMA driver expects number of transfers instead of number of bytes in `dma_block_config.block_size` field
### Describe the bug

In the `struct dma_block_config`, the field `block_size` is described as `Number of bytes to be transferred for this block`.

The `dma_rpi_pico` driver simply passes this value on to the pico's `dma_channel_configure` function.
This function explicitly states that this value does not encode the number of bytes:
```c
/* (...)
 * \if rp2040_specific
 * On RP2040 this is just the number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size) from 0 -> 2^32 - 1.
 * \endif
 *
 * \if rp2350_specific
 * On RP2350 the low 28 bits are used to encode a number of transfers (NOT bytes, see \ref channel_config_set_transfer_data_size) and non-zero values of the top 4 bits are used to specify other options.
 * \endif
 *
 * (...)
 */
```

In my example, this has lead to the DMA sending the same buffer of u32 values 4 times until I noticed this.

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Arch Linux (dev container), Fedora OS Silverblue (host)
- Toolchain: zephyr-sdk 0.17.4
- Zephyr version: 4.3.99

### Additional Context

_No response_

Majistor:
The Pico SDK expects a transfer count (items), not a byte count. The driver incorrectly passed the byte size, causing buffer overflows on 16-bit and 32-bit transfers.Fix this by storing the transfer width in the channel configuration and dividing the block size by that width to calculate the correct transfer count.

Fixes #99435

Signed-off-by: Ashirwad Paswan ashi06712@gmail.com

### DIFF
--- a/drivers/dma/dma_rpi_pico.c
+++ b/drivers/dma/dma_rpi_pico.c
@@ -43,6 +43,7 @@ struct dma_rpi_pico_channel {
 	void *source_address;
 	void *dest_address;
 	size_t block_size;
+	uint32_t source_data_size;
 };
 
 struct dma_rpi_pico_data {
@@ -197,6 +198,7 @@ static int dma_rpi_pico_config(const struct device *dev, uint32_t channel,
 
 	data->channels[channel].callback = dma_cfg->dma_callback;
 	data->channels[channel].user_data = dma_cfg->user_data;
+	data->channels[channel].source_data_size = dma_cfg->source_data_size;
 	data->channels[channel].direction = dma_cfg->channel_direction;
 
 	return 0;
@@ -221,7 +223,8 @@ static int dma_rpi_pico_reload(const struct device *dev, uint32_t ch, uint32_t s
 	data->channels[ch].dest_address = (void *)dst;
 	data->channels[ch].block_size = size;
 	dma_channel_configure(ch, &data->channels[ch].config, data->channels[ch].dest_address,
-			      data->channels[ch].source_address, data->channels[ch].block_size,
+			      data->channels[ch].source_address,
+			      data->channels[ch].block_size / data->channels[ch].source_data_size,
 			      true);
 
 	return 0;
@@ -241,7 +244,8 @@ static int dma_rpi_pico_start(const struct device *dev, uint32_t ch)
 	dma_irqn_set_channel_enabled(dma_rpi_pico_channel_irq(dev, ch), ch, true);
 
 	dma_channel_configure(ch, &data->channels[ch].config, data->channels[ch].dest_address,
-			      data->channels[ch].source_address, data->channels[ch].block_size,
+			      data->channels[ch].source_address,
+			      data->channels[ch].block_size / data->channels[ch].source_data_size,
 			      true);
 
 	return 0;


devbhavej (commented):
Hey @Majistor, do you mind running clang-format or git-clang-format for your change to address the warnings raised by the CI? 

This should help (https://docs.zephyrproject.org/latest/contribute/style/index.html).

drivers/dma/dma_rpi_pico.c
```diff
@@ -220,8 +221,9 @@ static int dma_rpi_pico_reload(const struct device *dev, uint32_t ch, uint32_t s
 	data->channels[ch].source_address = (void *)src;
 	data->channels[ch].dest_address = (void *)dst;
 	data->channels[ch].block_size = size;
+	data->channels[ch].transfer_size = dma_cfg->source_data_size;
```
soburi: `transfer_size` is only assigned here, so I think keeping it in `data->channels[ch]` is unnecessary.
It would be simpler to use `dma_cfg->source_data_size` directly to compute the transfer count.

drivers/dma/dma_rpi_pico.c
```diff
@@ -220,8 +221,9 @@ static int dma_rpi_pico_reload(const struct device *dev, uint32_t ch, uint32_t s
...
 	dma_channel_configure(ch, &data->channels[ch].config, data->channels[ch].dest_address,
-			      data->channels[ch].source_address, data->channels[ch].block_size,
+			      data->channels[ch].source_address, data->channels[ch].block_size/data->channels[ch].transfer_size,
```
soburi: It is better to divide by `data->channels[ch]->source_data_size`.

And it is possible that the data size cannot be divided by the transfer size.
Adding a check is recommended.

soburi (commented):
Thank you for your patch.
I just have a few comments to make.

drivers/dma/dma_rpi_pico.c
Majistor: hey! sure. I will try to do it asap

drivers/dma/dma_rpi_pico.c
```diff
@@ -43,6 +43,7 @@ struct dma_rpi_pico_channel {
 	void *source_address;
 	void *dest_address;
 	size_t block_size;
+	uint32_t transfer_size;
```
soburi: No more need this value.

.gitignore
```diff
@@ -123,3 +123,6 @@ ZephyrModuleFile.txt
 
 # Node dependecies
 node_modules
+#personal
```
soburi: Seems unnecessary, if necessary, let's make it a separate PR.

soburi (changes requested):

.gitignore
```diff
@@ -1,125 +0,0 @@
-# SPDX-License-Identifier: Apache-2.0
```
soburi, line 1: Don't delete everything. Make sure there are no differences.

soburi (changes requested):

soburi (changes requested):
Please address CI check.

soburi (dismissed):

soburi:
@Majistor 
Please fix CI errors.

soburi (approved):

ajf58 (approved):

PR #99946, opened 2025-11-25
Fixed sw switch timing issue

jackjinxb:


subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -794,7 +794,9 @@ void sw_switch(uint8_t dir_curr, uint8_t dir_next, uint8_t phy_curr, uint8_t fla
 			delay = HAL_RADIO_NS2US_ROUND(
 			    hal_radio_tx_ready_delay_ns_get(phy_next,
 							    flags_next) +
-			    hal_radio_rx_chain_delay_ns_get(phy_curr, 1));
+			    hal_radio_rx_chain_delay_ns_get(phy_curr, 1)+
+				hal_radio_tx_chain_delay_ns_get(phy_curr,
+								flags_curr));
```
cvinayak, line 799: Tx after Rx: RADIO_END arrives after `rx_chain_delay` on last bit received on air, and TXEN has to be triggered `tx_ready_delay` in advance.

The change requested is incorrect.

cvinayak (changes requested):

PR #99949, opened 2025-11-25
drivers: fpga_ice40: fix CS hack; fix `k_usleep` during spinlock

### LINKED ISSUE #99948: drivers: ice40_fpga: `k_usleep` while holding a spinlock
### Describe the bug

The [ICE40 FPGA driver]() uses spinlocks for synchronisation (most certainly a mistake).

https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/drivers/fpga/fpga_ice40_spi.c#L49

Further, it uses `k_usleep()` for its required delays.

https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/drivers/fpga/fpga_ice40_spi.c#L78

This in turn triggers: https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/kernel/include/kswap.h#L210

While it might be tempting to change that single `k_usleep()` to a `k_busy_wait()`, the issue is deeper: Almost all SPI drivers use `spi_context_wait_for_completion` which contains `k_sem_take`:

https://github.com/zephyrproject-rtos/zephyr/blob/8c264d30ea1b5c218a2efd4288820814ff16f1e2/drivers/spi/spi_context.h#L219

I understand that the spinlock was introduced because of the very strict timing requirements of the ICE40, but this `k_sem_take` makes almost all SPI peripheral drivers incompatible with the ICE40 driver.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Load an FPGA using the `ice40_fpga` driver.

### Relevant log output

```shell
ASSERTION FAIL [arch_irq_unlocked(key) || _kernel.cpus[0].current->base.thread_state & (((1UL << (0))) | ((1UL << (3))))] @ WEST_TOPDIR/zephyr/kernel/include/kswap.h:210
        Context switching while holding lock!
[00:00:00.012,000] <err> os: r0/a1:  0x00000004  r1/a2:  0x000000d2  r2/a3:  0x00000007
[00:00:00.012,000] <err> os: r3/a4:  0x00000004 r12/ip:  0x0000007d r14/lr:  0x3809b257
[00:00:00.012,000] <err> os:  xpsr:  0x01000000
[00:00:00.012,000] <err> os: Faulting instruction address (r15/pc): 0x3802cd2c
[00:00:00.012,000] <err> os: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
[00:00:00.012,000] <err> os: Current thread: 0x203105d8 (main)
[00:00:00.012,000] <err> os: Halting system
```

### Impact

Minor - works when asserts disabled, but there's probably a reason for this assert

### Environment

Zephyr `4.3.0`
Zephyr SDK `0.17.4`
OS: Windows 11

### Additional Context

_No response_

Finomnis:
Fixes #99948.

Two asserts were triggered in this module:
- Holding a spinlock while `k_usleep` triggers an assert in `k_swap`
- All gpio_set must be on valid pin configs, which the CS hack in this module breaks.

### DIFF
--- a/drivers/fpga/fpga_ice40_bitbang.c
+++ b/drivers/fpga/fpga_ice40_bitbang.c
@@ -115,13 +115,20 @@ static void fpga_ice40_spi_send_data(size_t delay, volatile gpio_port_pins_t *se
  *
  * https://www.latticesemi.com/~/media/LatticeSemi/Documents/Handbooks/iCE40FamilyHandbook.pdf
  */
-static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32_t img_size)
+static void fpga_ice40_load_handler(struct k_work *item)
 {
+	struct fpga_ice40_work_item *work_item =
+		CONTAINER_OF(item, struct fpga_ice40_work_item, work);
+	const struct device *dev = work_item->dev;
+	uint32_t *image_ptr = work_item->image_ptr;
+	const uint32_t img_size = work_item->image_size;
+
 	int ret;
 	uint32_t crc;
 	gpio_port_pins_t cs;
 	gpio_port_pins_t clk;
 	k_spinlock_key_t key;
+	struct k_spinlock spinlock;
 	gpio_port_pins_t pico;
 	gpio_port_pins_t creset;
 	struct fpga_ice40_data *data = dev->data;
@@ -130,12 +137,14 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
 
 	if (!device_is_ready(config_bitbang->clk.port)) {
 		LOG_ERR("%s: GPIO for clk is not ready", dev->name);
-		return -ENODEV;
+		ret = -ENODEV;
+		goto exit;
 	}
 
 	if (!device_is_ready(config_bitbang->pico.port)) {
 		LOG_ERR("%s: GPIO for pico is not ready", dev->name);
-		return -ENODEV;
+		ret = -ENODEV;
+		goto exit;
 	}
 
 	/* prepare masks */
@@ -150,7 +159,8 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
 		LOG_WRN("already loaded with image CRC32c: 0x%08x", data->crc);
 	}
 
-	key = k_spin_lock(&data->lock);
+	/* spinlock to prevent interruptions; this entire function is very timing sensitive */
+	key = k_spin_lock(&spinlock);
 
 	/* clear crc */
 	data->crc = 0;
@@ -234,9 +244,13 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
 	}
 
 unlock:
-	k_spin_unlock(&data->lock, key);
+	k_spin_unlock(&spinlock, key);
 
-	return ret;
+exit:
+	work_item->result = ret;
+	k_sem_give(&work_item->finished);
+
+	return;
 }
 
 static DEVICE_API(fpga, fpga_ice40_api) = {
@@ -261,7 +275,8 @@ static DEVICE_API(fpga, fpga_ice40_api) = {
 		.mhz_delay_count = DT_INST_PROP(inst, mhz_delay_count),                            \
 	};                                                                                         \
                                                                                                    \
-	FPGA_ICE40_CONFIG_DEFINE(inst, &fpga_ice40_config_bitbang_##inst);                         \
+	FPGA_ICE40_CONFIG_DEFINE(inst, fpga_ice40_load_handler,                                    \
+				 &fpga_ice40_config_bitbang_##inst);                               \
                                                                                                    \
 	DEVICE_DT_INST_DEFINE(inst, fpga_ice40_init, NULL, &fpga_ice40_data_##inst,                \
 			      &fpga_ice40_config_##inst, POST_KERNEL, CONFIG_FPGA_INIT_PRIORITY,   \
--- a/drivers/fpga/fpga_ice40_common.c
+++ b/drivers/fpga/fpga_ice40_common.c
@@ -31,10 +31,9 @@ void fpga_ice40_crc_to_str(uint32_t crc, char *s)
 enum FPGA_status fpga_ice40_get_status(const struct device *dev)
 {
 	enum FPGA_status st;
-	k_spinlock_key_t key;
 	struct fpga_ice40_data *data = dev->data;
 
-	key = k_spin_lock(&data->lock);
+	k_mutex_lock(&data->lock, K_FOREVER);
 
 	if (data->loaded && data->on) {
 		st = FPGA_STATUS_ACTIVE;
@@ -42,7 +41,7 @@ enum FPGA_status fpga_ice40_get_status(const struct device *dev)
 		st = FPGA_STATUS_INACTIVE;
 	}
 
-	k_spin_unlock(&data->lock, key);
+	k_mutex_unlock(&data->lock);
 
 	return st;
 }
@@ -50,11 +49,10 @@ enum FPGA_status fpga_ice40_get_status(const struct device *dev)
 static int fpga_ice40_on_off(const struct device *dev, bool on)
 {
 	int ret;
-	k_spinlock_key_t key;
 	struct fpga_ice40_data *data = dev->data;
 	const struct fpga_ice40_config *config = dev->config;
 
-	key = k_spin_lock(&data->lock);
+	k_mutex_lock(&data->lock, K_FOREVER);
 
 	ret = gpio_pin_configure_dt(&config->creset, on ? GPIO_OUTPUT_HIGH : GPIO_OUTPUT_LOW);
 	if (ret < 0) {
@@ -65,7 +63,7 @@ static int fpga_ice40_on_off(const struct device *dev, bool on)
 	ret = 0;
 
 unlock:
-	k_spin_unlock(&data->lock, key);
+	k_mutex_unlock(&data->lock);
 
 	return ret;
 }
@@ -96,6 +94,11 @@ int fpga_ice40_init(const struct device *dev)
 {
 	int ret;
 	const struct fpga_ice40_config *config = dev->config;
+	struct fpga_ice40_data *data = dev->data;
+
+	k_mutex_init(&data->lock);
+	k_work_init(&data->work_item.work, config->load_handler);
+	k_sem_init(&data->work_item.finished, 0, 1);
 
 	if (!device_is_ready(config->creset.port)) {
 		LOG_ERR("%s: GPIO for creset is not ready", dev->name);
@@ -120,4 +123,25 @@ int fpga_ice40_init(const struct device *dev)
 	}
 
 	return 0;
+}
+
+int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32_t img_size)
+{
+	struct fpga_ice40_data *data = dev->data;
+
+	struct fpga_ice40_work_item *work_item = &data->work_item;
+
+	k_mutex_lock(&data->lock, K_FOREVER);
+
+	work_item->dev = dev;
+	work_item->image_ptr = image_ptr;
+	work_item->image_size = img_size;
+
+	k_work_submit(&work_item->work);
+	k_sem_take(&work_item->finished, K_FOREVER);
+	int result = work_item->result;
+
+	k_mutex_unlock(&data->lock);
+
+	return result;
 }
--- a/drivers/fpga/fpga_ice40_common.h
+++ b/drivers/fpga/fpga_ice40_common.h
@@ -27,7 +27,7 @@
 #define FPGA_ICE40_LEADING_CLOCKS_MIN  8
 #define FPGA_ICE40_TRAILING_CLOCKS_MIN 49
 
-#define FPGA_ICE40_CONFIG_DEFINE(inst, derived_config_)                                            \
+#define FPGA_ICE40_CONFIG_DEFINE(inst, load_handler_, derived_config_)                             \
 	BUILD_ASSERT(DT_INST_PROP(inst, spi_max_frequency) >= FPGA_ICE40_SPI_HZ_MIN);              \
 	BUILD_ASSERT(DT_INST_PROP(inst, spi_max_frequency) <= FPGA_ICE40_SPI_HZ_MAX);              \
 	BUILD_ASSERT(DT_INST_PROP(inst, config_delay_us) >= FPGA_ICE40_CONFIG_DELAY_US_MIN);       \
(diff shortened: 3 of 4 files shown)


JarmouniA (commented):
I believe the problem here is not in the spinlock usage, which is valid in my opinion since this driver could be used in multi-core setups, rather  in the `k_usleep` introduced in 2f6dff5 without taking into consideration the usage of spinlocks. The mistake went unnoticed because the triggered assert was added just recently.
https://docs.zephyrproject.org/latest/kernel/services/smp/smp.html#spinlocks

Finomnis:
@JarmouniA Would you mind elaborating on spinlock vs mutex in this scenario? Why would one be picked over the other, what are the advantages/disadvantages?

The lock will be held during the entire 'load', which can be 50ms. During this entire time, nothing else on this chip can happen if you are in a spinlock, if I understand spinlocks correctly, because it completely prevents the scheduler from unscheduling the current thread.

I don't quite understand the rationale for this, because the underlying SPI could be DMA+Interrupt based. Is it really wise to hold a spinlock during a DMA transfer that might signal its termination with an interrupt?

drivers/fpga/fpga_ice40_common.h
```diff
@@ -38,11 +38,11 @@
...
 	BUILD_ASSERT(DT_INST_PROP(inst, trailing_clocks) <= UINT8_MAX);                            \
-	                                                                                           \
+                                                                                                   \
 	static const struct fpga_ice40_config fpga_ice40_config_##inst = {                         \
-		.bus = SPI_DT_SPEC_INST_GET(inst,                                                  \
-					    SPI_OP_MODE_MASTER | SPI_MODE_CPOL | SPI_MODE_CPHA |   \
-						    SPI_WORD_SET(8) | SPI_TRANSFER_MSB),           \
+		.bus = SPI_DT_SPEC_INST_GET(inst, SPI_OP_MODE_MASTER | SPI_MODE_CPOL |             \
+							  SPI_MODE_CPHA | SPI_WORD_SET(8) |        \
+							  SPI_TRANSFER_MSB),                       \
```
cfriedt, line 45: It looks like there are some unrelated formatting changes. Can those be undone?

cfriedt:
@Finomnis - I think the `k_usleep()` should be converted to a `k_busy_wait()`. That would definitely be an issue for spinlocks.

The reason that spinlocks were used originally was to avoid any scheduling latencies that a mutex might introduce that could affect the bit-bang interface, which is quite sensitive to timing issues.

We want to avoid the thread being rescheduled while loading the ice40 bitstream.

drivers/fpga/fpga_ice40_common.h
Finomnis: I thought we have an official `.clang-format` file, should we not update old code?

Finomnis:
@cfriedt Better like this?

drivers/fpga/fpga_ice40_spi.c
```diff
@@ -33,21 +33,31 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
...
 	 * as the configuration sequence requires this signal to be inactive
 	 * during the leading and trailing clock phase.
 	 */
-	bus.config.cs.gpio.port = NULL;
+	__ASSERT(bus.config.cs.cs_is_gpio, "driver only compatible with gpio based cs");
+	struct spi_cs_control previous_cs = bus.config.cs;
+
+	if (previous_cs.cs_is_gpio) {
+		bus.config.cs = (struct spi_cs_control){
+			.cs_is_gpio = false,
```
benediktibk, line 51: Wouldn't it be sufficient to just set cs_is_gpio instead of the copy assignment? Or am I missing something here?

benediktibk (changes requested):

drivers/fpga/fpga_ice40_spi.c
Finomnis: This is a `union`, setting `cs_is_gpio` changes which union type this struct gets interpreted as, so I think this is dangerous. The gpio id might get interpreted as a delay

Finomnis: @benediktibk Here: https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/spi.h#L244

benediktibk: Now I get it, delay is also part of the "other" union. But besides that: delay is specified as us and the others as ns?

Finomnis: Oops

Finomnis: @benediktibk Should we even copy this? Or how is the required timing for the FPGA? Should we simply zero it?

Finomnis: If it has any effect in the first place ... I'm unsure. Probably SPI driver dependent. I'll copy it properly to keep the behaviour consistent to how it was before.

Finomnis: Fixed

Finomnis: But resetting it afterwards is pointless as the entire config is memcpy'd

Finomnis:
Still getting this with the current state of the PR:

```
ASSERTION FAIL [arch_irq_unlocked(key) || _kernel.cpus[0].current->base.thread_state & (((1UL << (0))) | ((1UL << (3))))] @ WEST_TOPDIR/zephyr/kernel/include/kswap.h:210
        Context switching while holding lock!
[00:00:00.014,000] <err> os: r0/a1:  0x00000004  r1/a2:  0x000000d2  r2/a3:  0x00000003
[00:00:00.014,000] <err> os: r3/a4:  0x00000004 r12/ip:  0x00000087 r14/lr:  0x3809b583
[00:00:00.014,000] <err> os:  xpsr:  0x01000000
[00:00:00.014,000] <err> os: Faulting instruction address (r15/pc): 0x3802ccb0
[00:00:00.014,000] <err> os: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
[00:00:00.014,000] <err> os: Current thread: 0x20301f00 (main)
[00:00:00.037,000] <err> os: Halting system
```

cfriedt:
Kconfig to select mutex or spinlock seems like maybe overkill.

I kind of wonder if there is a way to perform bitstream loading in a high priority or perhaps cooperative workqueue thread.

cfriedt:
@Finomnis - you could theoretically use the system work queue, and I believe it's cooperative already.

drivers/fpga/fpga_ice40_bitbang.c
```diff
@@ -115,27 +115,36 @@ static void fpga_ice40_spi_send_data(size_t delay, volatile gpio_port_pins_t *se
  *
  * https://www.latticesemi.com/~/media/LatticeSemi/Documents/Handbooks/iCE40FamilyHandbook.pdf
  */
-static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32_t img_size)
+static void fpga_ice40_load_handler(struct k_work *item)
 {
+	struct fpga_ice40_work_item *work_item =
+		CONTAINER_OF(item, struct fpga_ice40_work_item, work);
+	const struct device *dev = work_item->dev;
+	const uint32_t *image_ptr = work_item->image_ptr;
```
benediktibk: If you remove the const here you get rid of the sonar cube complaint later. It also isn't const in the work item, so no real need to make it const.

benediktibk (approved):
I like the idea, but unfortunately I don't have the STM32 based hardware at the moment available to test it out. I hope to get my hands on it within the next few months, but right now I can only hope it won't break the fickle ICE40 loading procedure.

Finomnis:
> I like the idea, but unfortunately I don't have the STM32 based hardware at the moment available to test it out. I hope to get my hands on it within the next few months, but right now I can only hope it won't break the fickle ICE40 loading procedure.

It's pretty broken already 🤷

drivers/fpga/fpga_ice40_bitbang.c
Finomnis: Fixed

benediktibk (approved):

drivers/fpga/fpga_ice40_common.c
```diff
@@ -96,6 +94,15 @@ int fpga_ice40_init(const struct device *dev)
 {
 	int ret;
 	const struct fpga_ice40_config *config = dev->config;
+	struct fpga_ice40_data *data = dev->data;
+
+	data->loaded = false;
```
cfriedt: The `data` object should already be zero initialized, so you can drop these 4 lines.

drivers/fpga/fpga_ice40_spi.c
```diff
@@ -138,32 +150,33 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
...
+		goto cleanup;
 	}
 
 	ret = 0;
 	data->loaded = true;
 	fpga_ice40_crc_to_str(crc, data->info);
 	LOG_INF("Loaded image with CRC32 0x%08x", crc);
 
-unlock:
+cleanup:
```
cfriedt: Changing this label isn't really needed and adds a number of unnecessary lines in this diff.

Can the label rename be undone?

I also see `img_size` being changed to `imag_size` in a number of places where it does not look necessary.

If the renames must stay in this change, can it be moved to a separate commit.

cfriedt (commented):
Suggestions to move gratuitous renames to separate commit.

drivers/fpga/fpga_ice40_spi.c
Finomnis: Sure :) can do. Renamed it because it now does more than unlock

drivers/fpga/fpga_ice40_common.c
Finomnis: I'm not sure if I follow; would you mind explaining where this initialization happens?

drivers/fpga/fpga_ice40_spi.c
Finomnis: Reverted renames.

benediktibk (approved):

drivers/fpga/fpga_ice40_common.c
Finomnis: @cfriedt Is it being initialized somewhere? Or should I keep the lines?

JarmouniA (dismissed):

drivers/fpga/fpga_ice40_common.c
mathieuchopstm: I think @cfriedt refers to the fact that data is zero-initialized by default.

mathieuchopstm (commented):
Agree with @cfriedt - some cosmetic changes are mixed with functional changes, which is not ideal. LGTM otherwise.

drivers/fpga/fpga_ice40_common.c
Finomnis: @mathieuchopstm are static structs in C zero initialized if they don't have a `= {}` initializer? I thought they require that, otherwise they are just random data

Finomnis: I guess you are right. Will update the code.

Finomnis: Fixed.

drivers/fpga/fpga_ice40_bitbang.c
```diff
@@ -115,27 +115,36 @@ static void fpga_ice40_spi_send_data(size_t delay, volatile gpio_port_pins_t *se
...
+
 	int ret;
 	uint32_t crc;
 	gpio_port_pins_t cs;
 	gpio_port_pins_t clk;
-	k_spinlock_key_t key;
 	gpio_port_pins_t pico;
 	gpio_port_pins_t creset;
+	struct k_spinlock spinlock;
+	k_spinlock_key_t spinlock_key;
```
mathieuchopstm: `key` --> `spinlock_key`

drivers/fpga/fpga_ice40_bitbang.c
```diff
@@ -145,12 +154,13 @@ static int fpga_ice40_load(const struct device *dev, uint32_t *image_ptr, uint32
 	creset = BIT(config->creset.pin);
 
 	/* crc check */
-	crc = crc32_ieee((uint8_t *)image_ptr, img_size);
+	crc = crc32_ieee((uint8_t *)image_ptr, image_size);
```
mathieuchopstm: `img_size` --> `image_size`

mathieuchopstm (commented):
> > Agree with @cfriedt - some cosmetic changes are mixed with functional changes, which is not ideal. LGTM otherwise.
> 
> @mathieuchopstm Would you mind pointing out which ones? I already reverted the changes @cfriedt was talking about.

These two renames caught my eyes (they generate noise in the diff):

drivers/fpga/fpga_ice40_bitbang.c
Finomnis: As I moved the spin lock in here I thought it's a good chance to name the "key" a little bit more descriptive. There's already a diff for all relevant lines in this change, so I thought why not. Was that wrong?

drivers/fpga/fpga_ice40_bitbang.c
Finomnis: That's technically a different variable. Should I name the new variable the same as the previous one to prevent the diff? I thought if I created a completely new variable I should maybe adjust the name to be consistent with `image_ptr`

drivers/fpga/fpga_ice40_bitbang.c
mathieuchopstm: It's not *wrong* but it does generate noise in the diff. Also, `key` is a very common name for such variables so I don't think the `spinlock_` prefix adds much value.

drivers/fpga/fpga_ice40_bitbang.c
mathieuchopstm: To be consistent with `image_ptr`, it should (arguably) be named `image_sz` rather than `image_size` 🙂 

I'd keep `image_sz`/`image_size` in the structure and `img_size` here to avoid noise in diff - again, not much value brought by the rename.

Finomnis: Better?

drivers/fpga/fpga_ice40_bitbang.c
Finomnis: Better?

JarmouniA (approved):

mathieuchopstm (approved):

drivers/fpga/fpga_ice40_common.c
mathieuchopstm: > @mathieuchopstm are static structs in C zero initialized if they don't have a `= {}` initializer? I thought they require that, otherwise they are just random data

This is only true for local variables, not globals.

cfriedt (approved):

PR #99966, opened 2025-11-25
drivers: adc: microchip: Introduce G1 ADC Driver

ArunMCHP:
This pull request adds ADC driver support for the Microchip ADC peripheral. It introduces the ADC driver implementation and updates the devicetree for the SAM D5x/E5x series to include the required node and bindings.


dts/bindings/adc/microchip,adc-g1.yaml
```diff
@@ -0,0 +1,74 @@
...
+      Clock phandles used to drive the ADC.
+
+  clock-names:
+    required: true
+    description: |
+      Clock name corresponding to the `clocks` entry.
+
+  vref-mv:
+    type: int
+    description: |
```
nandojve: User `must` select this in the board.

```diff
  vref-mv:
    type: int
+    required: true
    description: |
```

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
```diff
@@ -254,6 +254,38 @@
...
 			clock-names = "mclk", "gclk";
 		};
+
+		adc0: adc@43001c00 {
+			compatible = "microchip,adc-g1";
+			reg = <0x43001c00 0x4a>;
+			interrupts = <118 0>, <119 0>;
+			interrupt-names = "overrun", "resrdy";
+			/* 48 MHz GCLK / 4 prescaler = 12 MHz ADC clock. */
+			prescaler = <4>;
```
nandojve: prescaler is application dependent.
here we add only values that should never be changed from the user in the board.
drop `prescaler` and `vref-mv`.

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,911 @@
...
+		/* In single-ended mode, connect negative input to GND */
+		adc_reg->ADC_INPUTCTRL |= ADC_INPUTCTRL_MUXNEG(ADC_INPUTCTRL_MUXNEG_GND_Val);
+	}
+
+	return 0;
+}
+
+static int8_t adc_set_oversampling(adc_registers_t *adc_reg, uint8_t oversampling)
+{
+
```
nandojve: useless space

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,911 @@
...
+}
+
+static void adc_mchp_isr(const struct device *dev)
+{
+	struct adc_mchp_dev_data *dev_data = dev->data;
+	uint16_t result;
+
+	adc_interrupt_clear(ADC_REGS);
+	result = adc_get_conversion_result(ADC_REGS);
+	*dev_data->buffer++ = result;
```
nandojve: This is hard to read. Let compiler optimize the code.
`	*dev_data->buffer++ = result;`

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,911 @@
...
+	return ret;
+}
+
+#ifdef CONFIG_ADC_ASYNC
+
+static int adc_mchp_read_async(const struct device *dev, const struct adc_sequence *sequence,
+			       struct k_poll_signal *async)
+{
+	struct adc_mchp_dev_data *data = dev->data;
+
```
nandojve: drop empty line

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,911 @@
...
+	case 1:
+		ADC_CALIB_WRITE(sw0_word, 1, BIASCOMP);
+		ADC_CALIB_WRITE(sw0_word, 1, BIASR2R);
+		ADC_CALIB_WRITE(sw0_word, 1, BIASREFBUF);
+		break;
+	default:
+		/* Invalid ADC instance */
+		LOG_ERR("Invalid ADC index for calibration");
+		break;
+	}
```
nandojve: Should we consider the fail ASAP here ?
BTW, how this **n** will be invalid if it came from devicetree `DT_INST_FOREACH_STATUS_OKAY`?

```c
if (not valid index) {
    LOG_ERR("Invalid ADC index for calibration");
    return;
}

ADC_CALIB_WRITE(sw0_word, n, BIASCOMP);
ADC_CALIB_WRITE(sw0_word, n, BIASR2R);
ADC_CALIB_WRITE(sw0_word, n, BIASREFBUF);
```

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.yaml
```diff
@@ -10,6 +10,7 @@ toolchain:
 flash: 1024
 ram: 256
 supported:
+  - adc
```
nandojve, line 13: samples/tests are filtering so you need to add the board there.

nandojve (changes requested):

drivers/adc/adc_mchp_g1.c
ArunMCHP: removed

drivers/adc/adc_mchp_g1.c
ArunMCHP: removed

drivers/adc/adc_mchp_g1.c
ArunMCHP: Moved this logic into the init function. Since there are only two ADC instances (ADC0 and ADC1), we should avoid relying on driver instantiation and its order. Instead, the instance ID is now obtained from the dts property during initialization.

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.yaml
ArunMCHP: Added.

dts/bindings/adc/microchip,adc-g1.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - 1: ADC1
+
+  vref-mv:
+    type: int
+    required: true
+    description: |
+      ADC reference voltage in millivolts.
+
+      This should match the reference selected in hardware or software config.
+
```
nandojve: This 2 properties are very questionable.

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,917 @@
...
+	adc_controller_enable(ADC_REGS);
+	adc_context_unlock_unconditionally(&dev_data->ctx);
+
+	return ret;
+}
+
+static DEVICE_API(adc, adc_mchp_api) = {
+	.channel_setup = adc_mchp_channel_setup,
+	.read = adc_mchp_read,
+	.ref_internal = ADC_MCHP_VREF_MV,
```
nandojve: I have some problems to understand the logic behind this because this is a multi instance ADC with REFCTRL.

* Why select the ref_internal only from instance 0 ? Is only possible to have 1 reference for both instances?
* Is the reference fixed at compile time and can not be changed anymore ? So, the user can never switch the reference ?

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,917 @@
...
+
+static void adc_factory_calib_value(const struct device *dev)
+{
+	const struct adc_mchp_dev_config *const dev_cfg = dev->config;
+
+	ADC_REGS->ADC_CALIB = 0;
+/* Set ADC Calib register */
+#if defined(ADC_CALIB_BIASCOMP_Pos)
+	if (dev_cfg->adc_instance_id == 0) {
+		ADC_CALIB_WRITE_FIELD(0, BIASCOMP);
```
nandojve: In the devicetree we define the address of the ADC instance which should be used to operate direct using the offsets. Why are you not using `adc_instance_id` here ? You are going in the opposite direction with this.

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,917 @@
...
+	((((uint64_t)(sampling_time_ns) * (gclk_adc_hz)) / ((prescaler_val) * 1000000000ULL)) - 1)
+
+/* Macro to Write the calibration values */
+#define ADC_CALIB_WRITE_FIELD(adc_id, field)                                                       \
+	do {                                                                                       \
+		volatile uint32_t *sw0_word = (volatile uint32_t *)ADC_CALIBRATION_AREA;           \
+		uint8_t val = (((*sw0_word) & ADC##adc_id##_##field##_Msk) >>                      \
+			       ADC##adc_id##_##field##_Pos);                                       \
+		ADC_REGS->ADC_CALIB |= ADC_CALIB_##field(val);                                     \
+	} while (0)
```
nandojve: Is the difference between one instance to the another just an offset and the positions are the same? If yes, you should add the offset in devicetree instead to use adc_id.

nandojve (changes requested):

drivers/adc/adc_mchp_g1.c
ArunMCHP: * The calibration values are stored in the NVM Software Calibration Area, which is not related to the actual ADC peripheral address.
* The bit positions for these calibration fields differ for each ADC instance.
* Therefore, we cannot use the peripheral offset from the devicetree to determine them.

image

image

drivers/adc/adc_mchp_g1.c
ArunMCHP: No. The bit positions for these calibration fields differ for each ADC instance. We can't use offset.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Why are you not using adc_instance_id here?

Since this macro uses token pasting, I cannot directly pass a variable as the adc_id argument. If I pass a variable, the preprocessor will treat the variable name as literal text and paste it into the token, which leads to invalid symbol names.
Because of this, I need to check the ADC instance number at runtime and then pass the correct instance ID as a literal (e.g., 0 or 1) to the macro.

nandojve: If you add a property in the devicetree that informs the number of bits to shift right you do the job. Then the indexes will be always define to parse in the bit 0 position.

SAM D5X/E5X and PIC32CX 41/6/61
So ADC0 will shift 2 and ADC1 will shift 16.

In the second example, which should be another device with a different driver BTW

ADC0 will shift 0 and ADC1 will shift 6.

The same principle can be applied to any calibration value and it is instance defined so we can add this info in the devicetree and user never will touch because it is fixed.

BTW, the NVM node should define the place of the Calibration Area and then we pick inside the drivers that need.

ArunMCHP: We are trying to use same G1 driver for SAM D5X/E5X, PIC32CX 41/61 and PIC32CM JH.

nandojve: > We are trying to use same G1 driver for SAM D5X/E5X, PIC32CX 41/61 and PIC32CM JH.

You need only define 2 bindings and then based on which one you select the parser to pick the bits from the correct place. The solution still the same in my eyes if the position are the only thing that changes.

https://github.com/zephyrproject-rtos/zephyr/blob/3be4c4c5a81949b771c7d59b1852ef3e61836df9/drivers/ethernet/eth_sam_gmac.c#L23-L27

dts/bindings/adc/microchip,adc-g1.yaml
ArunMCHP: Removed these props. Thanks!

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added `calib-mapping` property to the DTS and programmed the ADC calibration register using values from the NVM software calibration area, which is defined under the NVM node.

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+		break;
+	case ADC_RESOLUTION_10BIT:
+		if (oversampling != 0) {
+			LOG_ERR("Invalid oversampling: %d\n", oversampling);
+			return -EINVAL;
+		}
+		resolution_val = ADC_CTRLB_RESSEL_10BIT;
+		break;
+	case ADC_RESOLUTION_12BIT:
+		if (oversampling) {
```
NhMchp: add explicit checking

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+	ret = adc_set_resolution(ADC_REGS, sequence->resolution, sequence->oversampling);
+	if (ret != 0) {
+		LOG_ERR("Invalid resolution : %d\n", sequence->resolution);
+		return ret;
+	}
+
+	/* Verify all requested channels are initialized and store resolution */
+	channels = sequence->channels;
+	channel_count = 0;
+	while (channels) {
```
NhMchp: explicit check

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+}
+
+static void adc_mchp_isr(const struct device *dev)
+{
+	struct adc_mchp_dev_data *dev_data = dev->data;
+	uint16_t result;
+
+	adc_interrupt_clear(ADC_REGS);
+	result = adc_get_conversion_result(ADC_REGS);
+	*dev_data->buffer++ = result;
```
NhMchp: split to multiple line for better readability

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+{
+	/* Validate gain */
+	if (gain != ADC_GAIN_1_2 && gain != ADC_GAIN_2_3 && gain != ADC_GAIN_4_5 &&
+	    gain != ADC_GAIN_1) {
+		LOG_ERR("Invalid gain : %d\n", gain);
+		return -EINVAL;
+	}
+
+	/* Validate reference */
+	if (reference < ADC_REF_VDD_1 || reference > ADC_REF_EXTERNAL1) {
```
NhMchp: add brackets () for each expression

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+		LOG_ERR("Invalid Sample Length : %d\n", sample_length);
+		return -EINVAL;
+	}
+
+	/*
+	 * Validate input_positive:
+	 * - Must be <= MCHP_ADC_INPUT_POS_MAX
+	 * - Range 0x10–0x17 is reserved
+	 */
+	if (input_positive > MCHP_ADC_INPUT_POS_MAX ||
```
NhMchp: add brackets () for each expression

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+	return 0;
+}
+
+static int8_t adc_set_input_positive(adc_registers_t *adc_reg, uint8_t input_positive)
+{
+	/* Validate input_positive:
+	 * - Must be <= 0x1E (highest valid value is DAC0)
+	 * - Values 0x10 to 0x17 are reserved and must not be used
+	 */
+	if ((input_positive > MCHP_ADC_INPUT_POS_MAX) ||
```
NhMchp, line 304: add brackets () for each expression

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+#endif /* MCHP_SUPC_API_SUPPORT_AVAILABLE */
+
+	return 0;
+}
+
+static int8_t adc_set_input_negative(adc_registers_t *adc_reg, uint8_t input_negative,
+				     bool differential)
+{
+	/* Validate input_negative range (only valid for differential mode) */
+	if (input_negative > 7) {
```
NhMchp: avoid magic number '7'

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+	 * 0x5 = 32 samples
+	 * 0x6 = 64 samples
+	 * 0x7 = 128 samples
+	 * 0x8 = 256 samples
+	 * 0x9 = 512 samples
+	 * 0xA = 1024 samples
+	 *
+	 * Valid range: 0 to 10 (inclusive)
+	 */
+	if (oversampling > 10) {
```
NhMchp: avoid magic number

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+	/* Start the ADC conversion */
+	adc_trigger_conversion(ADC_REGS);
+}
+
+static int adc_check_buffer_size(const struct adc_sequence *sequence, uint8_t active_channels)
+{
+	size_t needed_buffer_size;
+
+	needed_buffer_size = active_channels * sizeof(uint16_t);
+	if (sequence->options) {
```
NhMchp: explicit checking

drivers/adc/adc_mchp_g1.c
```diff
@@ -0,0 +1,900 @@
...
+	dev_data->channels = ctx->sequence.channels;
+
+	adc_start_channel(dev_data->dev);
+}
+
+static void adc_context_update_buffer_pointer(struct adc_context *ctx, bool repeat_sampling)
+{
+	struct adc_mchp_dev_data *data = CONTAINER_OF(ctx, struct adc_mchp_dev_data, ctx);
+
+	if (repeat_sampling) {
```
NhMchp: add explicit checking

NhMchp (changes requested):

drivers/adc/adc_mchp_g1.c
ArunMCHP: added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Fixed.

drivers/adc/adc_mchp_g1.c
ArunMCHP: added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Fixed.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added.

drivers/adc/adc_mchp_g1.c
ArunMCHP: Added.

NhMchp (approved):

nandojve (approved):

PR #99968, opened 2025-11-25
drivers: serial: uart_mcux_iuart: Adjust fifo fill behavior for trans…

Fladdan:
Until now transmission FIFO was not used. With this chnage the FIFO will be filled up, until it is full.

Currently only Transmit Buffer FIFO Empty flag is checked:
image
i.MX 8M Plus Applications Processor Reference Manual, Rev. 1, 06/2021

Now bytes are added to the FIFO until it is full:
image
i.MX 8M Plus Applications Processor Reference Manual, Rev. 1, 06/2021

### DIFF
--- a/drivers/serial/uart_mcux_iuart.c
+++ b/drivers/serial/uart_mcux_iuart.c
@@ -89,7 +89,7 @@ static int mcux_iuart_fifo_fill(const struct device *dev,
 	int num_tx = 0U;
 
 	while ((len - num_tx > 0) &&
-	       (UART_GetStatusFlag(config->base, kUART_TxEmptyFlag))) {
+	       !(config->base->UTS & UART_UTS_TXFULL_MASK)) {
 
 		UART_WriteByte(config->base, tx_data[num_tx++]);
 	}


Fladdan:
@pdgendt This could be a relevant PR for you, since you contributed to other nxp driver wrappers as well.

drivers/serial/uart_mcux_iuart.c
```diff
@@ -89,7 +89,7 @@ static int mcux_iuart_fifo_fill(const struct device *dev,
 	int num_tx = 0U;
 
 	while ((len - num_tx > 0) &&
-	       (UART_GetStatusFlag(config->base, kUART_TxEmptyFlag))) {
+	       !(config->base->UTS & UART_UTS_TXFULL_MASK)) {
```
pdgendt, line 92: This would be the only occurrence in this driver where registers are read directly. Not sure how portable this is.

I'm going to let NXP devs respond here.

Fladdan: There is now higher level, which make the TXFULL field available. 
If we don't want to access directly, there is a chagne needed in the fsl driver @pdgendt.
What is your opinion @mmahadevan108 @dleach02 ?

nandojve: Hi @dleach02 , @mmahadevan108 ,

Daniel found an issue that we don't want to have a patch for this downstream.
Please advice!

dleach02:
Let me get our internal UART dev to engage on this.

mmahadevan108 (approved):

nandojve (approved):

PR #100034, opened 2025-11-25
toolchain: gcc: Add a new LTO_SINGLE_THREADED option for LTO

carlescufi:
As described in this issue:
https://github.com/zephyrproject-rtos/sdk-ng/issues/1038 `-flto=auto` fails on some (slower) Windows machines due to an issue with the Zephyr SDK's GCC toolchain for Windows.
In order to allow users to work around this issue, introduce a new CONFIG_LTO_SINGLE_THREADED option that switches to `-flto=1`, which enforces a single thread when processing LTO.


tejlmand (changes requested):
Thanks for fixing this, but we need corresponding property names in the templates:
https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/compiler/compiler_flags_template.cmake
https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/linker/linker_flags_template.cmake

and it seems the existing `optimization_lto` and `lto_arguments` are also missing :slightly_frowning_face:

carlescufi:
> @RobinKastberg will you take a look at corresponding IAR flags to use ?
> @evgeniy-paltsev will you take a look at corresponding MetaWare flags to use ?

@tejlmand I don't think LTO was supported on these toolchains to start with, so this new single-threaded option is not applicable to those until they actually support LTO itself?

tejlmand (approved):

nordicjm (approved):

PR #100038, opened 2025-11-25
drivers: watchdog: stm32: Add the Early Wakeup support for IWDG

juickar:
This PR adds the early wakeup fonctionnality for the IWDG.

### DIFF
--- a/drivers/watchdog/Kconfig.stm32
+++ b/drivers/watchdog/Kconfig.stm32
@@ -14,9 +14,10 @@ menuconfig IWDG_STM32
 	help
 	  Enable IWDG driver for STM32 line of MCUs
 
+if IWDG_STM32
+
 config IWDG_STM32_INITIAL_TIMEOUT
 	int "Value for IWDG timeout in ms"
-	depends on IWDG_STM32
 	default 100
 	range 1 26214
 	help
@@ -29,6 +30,25 @@ config IWDG_STM32_INITIAL_TIMEOUT
 
 	  Limiting maximum timeout to a safe value of 26214 ms here, which was
 	  calculated for highest LSI frequency among STM32 MCUs of 40 kHz.
+
+config IWDG_STM32_EARLY_WAKEUP
+	bool "Early Wakeup Interrupt support for STM32 IWDG"
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_WATCHDOG),interrupts)
+	help
+	  Enable support for Early Wakeup Interrupt.
+
+config IWDG_STM32_EWI_TIME
+	int "Value for IWDG early wake-up in ticks"
+	depends on IWDG_STM32_EARLY_WAKEUP
+	default 100
+	range 2 4095
+	help
+	  Set the early wake-up interrupt value for IWDG in ticks.
+	  It represents the time before reset when the early wake-up occurs.
+	  This value is ignored if the reload value provided at runtime
+	  is smaller. In such cases, the interrupt triggers one tick after
+	  the watchdog is reloaded.
+endif
 
 config WWDG_STM32
 	bool "System Window Watchdog (WWDG) Driver for STM32 family of MCUs"
--- a/drivers/watchdog/wdt_iwdg_stm32.c
+++ b/drivers/watchdog/wdt_iwdg_stm32.c
@@ -54,6 +54,39 @@
 #define IWDG_SR_UPDATE_TIMEOUT	(6U * IWDG_PRESCALER_MAX * \
 				 MSEC_PER_SEC / LSI_VALUE)
 
+#ifdef CONFIG_IWDG_STM32_EARLY_WAKEUP
+
+void iwdg_stm32_isr(const struct device *dev)
+{
+	struct iwdg_stm32_data *data = dev->data;
+	IWDG_TypeDef *iwdg = ((const struct iwdg_stm32_config *)dev->config)->instance;
+
+	if (LL_IWDG_IsEnabledIT_EWI(iwdg) && LL_IWDG_IsActiveFlag_EWIF(iwdg)) {
+		LL_IWDG_ClearFlag_EWIF(iwdg);
+		if (data->callback != NULL) {
+			data->callback(dev, 0);
+		}
+	}
+}
+
+static void iwdg_stm32_irq_config(const struct device *dev)
+{
+	IWDG_TypeDef *idg = ((const struct iwdg_stm32_config *)dev->config)->instance;
+
+	IRQ_CONNECT(DT_INST_IRQN(0), DT_INST_IRQ(0, priority), iwdg_stm32_isr,
+		    DEVICE_DT_INST_GET(0), 0);
+
+	irq_enable(DT_INST_IRQN(0));
+
+	LL_IWDG_ClearFlag_EWIF(idg);
+
+	while (LL_IWDG_IsActiveFlag_EWU(idg)) {
+	}
+	LL_IWDG_EnableIT_EWI(idg);
+}
+
+#endif /* CONFIG_IWDG_STM32_EARLY_WAKEUP */
+
 /**
  * @brief Calculates prescaler & reload values.
  *
@@ -125,6 +158,27 @@ static int iwdg_stm32_setup(const struct device *dev, uint8_t options)
 
 	tickstart = k_uptime_get_32();
 
+#ifdef CONFIG_IWDG_STM32_EARLY_WAKEUP
+	if (data->reload < 2U) {
+		/* Early wake-up is not possible if counter is
+		 * reloaded with value 1 as the system will be
+		 * reset immediately when the counter decrements.
+		 * (Note: reload=1 is NOT RECOMMENDED per RefMan)
+		 */
+		return -EINVAL;
+	}
+
+	/* If Kconfig is higher than reload, set early wake-up
+	 * as high as possible (= reload - 1); otherwise, set
+	 * to the requested value.
+	 */
+	uint32_t ewi_time = MIN(CONFIG_IWDG_STM32_EWI_TIME, data->reload);
+
+	LL_IWDG_SetEwiTime(cfg->instance, ewi_time);
+
+	iwdg_stm32_irq_config(dev);
+#endif /* CONFIG_IWDG_STM32_EARLY_WAKEUP */
+
 	/* Wait for the update operation completed */
 	while (LL_IWDG_IsReady(cfg->instance) == 0) {
 		if ((k_uptime_get_32() - tickstart) > IWDG_SR_UPDATE_TIMEOUT) {
@@ -155,7 +209,11 @@ static int iwdg_stm32_install_timeout(const struct device *dev,
 	uint32_t reload = 0U;
 
 	if (config->callback != NULL) {
-		return -ENOTSUP;
+		if (IS_ENABLED(CONFIG_IWDG_STM32_EARLY_WAKEUP)) {
+			data->callback = config->callback;
+		} else {
+			return -ENOTSUP;
+		}
 	}
 	if (data->reload) {
 		/* Timeout has already been configured */
@@ -260,7 +318,10 @@ static int iwdg_stm32_init(const struct device *dev)
 static const struct iwdg_stm32_config iwdg_stm32_dev_cfg = {
 	.instance = (IWDG_TypeDef *)DT_INST_REG_ADDR(0),
 };
-static struct iwdg_stm32_data iwdg_stm32_dev_data;
+
+static struct iwdg_stm32_data iwdg_stm32_dev_data = {
+	.callback = NULL,
+};
 
 DEVICE_DT_INST_DEFINE(0, iwdg_stm32_init, NULL,
 		    &iwdg_stm32_dev_data, &iwdg_stm32_dev_cfg,
--- a/drivers/watchdog/wdt_iwdg_stm32.h
+++ b/drivers/watchdog/wdt_iwdg_stm32.h
@@ -28,6 +28,8 @@ struct iwdg_stm32_config {
 struct iwdg_stm32_data {
 	uint32_t prescaler;
 	uint32_t reload;
+	/* IWDG user defined callback on EWI */
+	wdt_callback_t callback;
 };
 
 #endif	/* ZEPHYR_DRIVERS_WATCHDOG_IWDG_STM32_H_ */
--- a/dts/arm/st/h5/stm32h5.dtsi
+++ b/dts/arm/st/h5/stm32h5.dtsi
@@ -289,6 +289,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <35 0>;
 			status = "disabled";
 		};
 
--- a/dts/arm/st/u5/stm32u5.dtsi
+++ b/dts/arm/st/u5/stm32u5.dtsi
@@ -279,6 +279,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <27 0>;
 			status = "disabled";
 		};
 
--- a/dts/arm/st/wba/stm32wba.dtsi
+++ b/dts/arm/st/wba/stm32wba.dtsi
@@ -295,6 +295,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <27 0>;
 			status = "disabled";
 		};
 
--- a/samples/drivers/watchdog/sample.yaml
+++ b/samples/drivers/watchdog/sample.yaml
@@ -101,6 +101,21 @@ tests:
       - stm32h573i_dk
     integration_platforms:
       - nucleo_f103rb
+  sample.drivers.watchdog.stm32_iwdg_ewi:
+    extra_args: DTC_OVERLAY_FILE=boards/stm32_iwdg.overlay
+    filter: dt_compat_enabled("st,stm32-watchdog")
+    extra_configs:
+      - CONFIG_IWDG_STM32_EARLY_WAKEUP=y
+    platform_allow:
+      - nucleo_h563zi
+      - nucleo_h7s3l8
+      - nucleo_u083rc
+      - nucleo_u385rg_q
+      - nucleo_u575zi_q
+      - b_u585i_iot02a
+      - nucleo_wba55cg
+    integration_platforms:
+      - nucleo_wba55cg
   sample.drivers.watchdog.gd32_fwdgt:
     filter: dt_compat_enabled("gd,gd32-fwdgt")
     extra_args: DTC_OVERLAY_FILE=boards/gd32_fwdgt.overlay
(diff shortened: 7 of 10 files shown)


drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +30,19 @@ config IWDG_STM32_INITIAL_TIMEOUT
 	  Limiting maximum timeout to a safe value of 26214 ms here, which was
 	  calculated for highest LSI frequency among STM32 MCUs of 40 kHz.
 
+config IWDG_STM32_HAS_EARLY_WAKEUP
+	bool "Early Wakeup Interrupt support for STM32 IWDG"
```
mathieuchopstm: `HAS_xxx` options should not be visible.

Either make two options or rename this one.

drivers/watchdog/Kconfig.stm32
mathieuchopstm, line 1: Consider enclosing all options inside a `if IWDG_STM32` instead of adding `depends on` to each option.

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +30,19 @@ config IWDG_STM32_INITIAL_TIMEOUT
...
+	default y
+	depends on IWDG_STM32 && (SOC_SERIES_STM32H5X \
+	        || SOC_SERIES_STM32H7RSX \
+		|| SOC_SERIES_STM32MP2X \
+		|| SOC_SERIES_STM32U0X \
+		|| SOC_SERIES_STM32U3X \
+		|| SOC_SERIES_STM32U5X \
+		|| SOC_SERIES_STM32WBAX)
```
mathieuchopstm: I would exclude unsupported series instead since all new ones should have the feature (or introduce a new compatible)

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -174,7 +188,20 @@ static int iwdg_stm32_install_timeout(const struct device *dev,
...
 
+	if (ewi_time >= reload) {
+		if (reload <= 2U) {
+			/* Reload value is too small and EWI time cannot be set */
+			return -EINVAL;
+		}
+		/* EWI time should be less than reload time but should be at least 2 */
+		ewi_time = (reload > 2U) ? (reload - 1U) : 2U;
+	}
+	LL_IWDG_SetEwiTime(cfg->instance, ewi_time);
```
mathieuchopstm: Per RefMan, this should be done as part of `setup` after enabling the watchdog.

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -174,7 +188,20 @@ static int iwdg_stm32_install_timeout(const struct device *dev,
...
+	uint32_t ewi_time = EWI_TIME;
 
+	if (ewi_time >= reload) {
+		if (reload <= 2U) {
+			/* Reload value is too small and EWI time cannot be set */
+			return -EINVAL;
+		}
+		/* EWI time should be less than reload time but should be at least 2 */
+		ewi_time = (reload > 2U) ? (reload - 1U) : 2U;
+	}
```
mathieuchopstm: 1. Create a Kconfig option for the early wake-up threshold instead of hardcoding it
2. Simplify the logic + rewording
```suggestion
        if (reload == 1U) { //i.e., < 2U
            /* Early wake-up is not possible if counter is
             * reloaded with value 1 as the system will be
             * reset immediately when the counter decrements.
             * (Note: reload=1 is NOT RECOMMENDED per RefMan)
             */
             return -EINVAL;       
        }

        /*
         * If Kconfig is higher than reload, set early wake-up
         * as high as possible (= reload - 1); otherwise, set
         * to the requested value.
         */
	uint32_t ewi_time = MIN(CONFIG_EWI_TIME, reload);

	if (ewi_time >= reload) {
		if (reload <= 2U) {
			/* Reload value is too small and EWI time cannot be set */
			return -EINVAL;
		}
		/* EWI time should be less than reload time but should be at least 2 */
		ewi_time = (reload > 2U) ? (reload - 1U) : 2U;
	}
```

Note: as far as I understand, it is valid to have `RLR = EWIT` (since early wake-up occurs when counter is *strictly less than* `EWIT`)?

Note 2: Kconfig should be `range 2 4095`

drivers/watchdog/Kconfig.stm32
mathieuchopstm: Detect availability based on `interrupts` property:
```suggestion
	depends on IWDG_STM32 && $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_IWDG),interrupts)
```
(`DT_COMPAT_ST_STM32_IWDG` is generated automatically by build system)

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +31,22 @@ config IWDG_STM32_INITIAL_TIMEOUT
...
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_WATCHDOG),interrupts)
+	help
+	  Enable support for Early Wakeup Interrupt.
+
+config IWDG_STM32_EWI_TIME
+	int "Value for IWDG early wake-up in ticks"
+	default 100
+	range 2 4095
+	help
+	  Set the early wake-up interrupt value for IWDG in ticks.
```
mathieuchopstm, line 46: Nits:
1. More verbosity, e.g., "the amount of ticks before reset when the interrupt will occur"
2. indicate that this is ignored if value provided at runtime is smaller than this

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -260,9 +307,32 @@ static int iwdg_stm32_init(const struct device *dev)
...
+	irq_enable(DT_INST_IRQN(0));
+	LL_IWDG_EnableWriteAccess(idg);
+	LL_IWDG_ClearFlag_EWIF(idg);
+
+	while (LL_IWDG_IsActiveFlag_EWU(idg)) {
+	}
+
+	LL_IWDG_EnableIT_EWI(idg);
+	LL_IWDG_DisableWriteAccess(idg);
+}
```
mathieuchopstm: Move the definition where you placed the declaration.

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -189,6 +219,23 @@ static int iwdg_stm32_feed(const struct device *dev, int channel_id)
...
+	IWDG_TypeDef *iwdg = IWDG_STM32_STRUCT(dev);
+
+	if (LL_IWDG_IsEnabledIT_EWI(iwdg)) {
+		if (LL_IWDG_IsActiveFlag_EWIF(iwdg)) {
+			LL_IWDG_ClearFlag_EWIF(iwdg);
+			if (data->callback != NULL) {
+				data->callback(dev, 0);
+			}
+		}
+	}
```
mathieuchopstm: Nit: merge both checks to reduce indentation level
```suggestion
	if (LL_IWDG_IsEnabledIT_EWI(iwdg) && LL_IWDG_IsActiveFlag_EWIF(iwdg)) {
		LL_IWDG_ClearFlag_EWIF(iwdg);
		if (data->callback != NULL) {
			data->callback(dev, 0);
		}
	}
```

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -134,7 +138,28 @@ static int iwdg_stm32_setup(const struct device *dev, uint8_t options)
...
+	/* If Kconfig is higher than reload, set early wake-up
+	 * as high as possible (= reload - 1); otherwise, set
+	 * to the requested value.
+	 */
+	uint32_t ewi_time = MIN(CONFIG_IWDG_STM32_EWI_TIME, data->reload);
+
+	LL_IWDG_SetEwiTime(cfg->instance, ewi_time);
+
+	iwdg_stm32_irq_config(dev);
+#endif /* CONFIG_IWDG_STM32_EARLY_WAKEUP */
```
mathieuchopstm, line 180: This should come before the `while (!IsReady())` per RefMan

drivers/watchdog/Kconfig.stm32
juickar: done

drivers/watchdog/wdt_iwdg_stm32.c
juickar: done

drivers/watchdog/wdt_iwdg_stm32.c
juickar: done

drivers/watchdog/wdt_iwdg_stm32.c
juickar: done, also added a LL_IWDG_DisableWriteAccess() call at the end of the function after the final reload

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +31,24 @@ config IWDG_STM32_INITIAL_TIMEOUT
...
+	  It represents the time before reset when the early wake-up occurs.
+	  This value is ignored if the reload value provided
+	  at runtime is smaller.
```
mathieuchopstm:
```suggestion
	  It represents the time before reset when the early wake-up occurs.
	  This value is ignored if the reload value provided at runtime
	  is smaller.
```
+ maybe `(In such cases, the interrupt triggers one tick after the watchdog is reloaded)`

drivers/watchdog/wdt_iwdg_stm32.h
```diff
@@ -28,6 +28,14 @@ struct iwdg_stm32_config {
...
 	uint32_t reload;
+	/* IWDG user defined callback on EWI */
+	wdt_callback_t callback;
 };
 
+#define IWDG_STM32_CFG(dev) ((const struct iwdg_stm32_config *const)(dev)->config)
+
+#define IWDG_STM32_DATA(dev) ((struct iwdg_stm32_data *const)(dev)->data)
+
+#define IWDG_STM32_STRUCT(dev) ((IWDG_TypeDef *)(IWDG_STM32_CFG(dev))->instance)
```
mathieuchopstm: These macros add very little value:
* the first one is only used by the third one
* the second one is used exactly once but is useless: the cast it performs is done implicitly anyways
* the third macro is used twice, and merely saves explicitly loading `dev->config` in a dedicated variable

```suggestion
// I would also get rid of this macro, but if anything...
#define IWDG_STM32_STRUCT(dev) (((const struct iwdg_stm32_config *)dev->config)->instance)
```
\+ corresponding driver code updates

dts/arm/st/h5/stm32h5.dtsi
```diff
@@ -289,6 +289,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <35 7>;
```
mathieuchopstm: Don't set priorities on interrupts at DTSI level (unless something *requires* it).
```suggestion
			interrupts = <35 0>;
```

dts/arm/st/u5/stm32u5.dtsi
```diff
@@ -279,6 +279,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <27 7>;
```
mathieuchopstm: Ditto (prio)

dts/arm/st/wba/stm32wba.dtsi
```diff
@@ -295,6 +295,7 @@
 		iwdg: watchdog@40003000 {
 			compatible = "st,stm32-watchdog";
 			reg = <0x40003000 0x400>;
+			interrupts = <27 7>;
```
mathieuchopstm: Ditto (prio)

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -134,7 +188,7 @@ static int iwdg_stm32_setup(const struct device *dev, uint8_t options)
 
 	/* Reload counter just before leaving */
 	LL_IWDG_ReloadCounter(cfg->instance);
-
+	LL_IWDG_DisableWriteAccess(cfg->instance);
```
mathieuchopstm: Not necessary: `0xAAAA: reloads the [watchdog] and write-protects registers.`
```suggestion
```

drivers/watchdog/Kconfig.stm32
juickar: Done

drivers/watchdog/wdt_iwdg_stm32.h
juickar: Done

dts/arm/st/h5/stm32h5.dtsi
juickar: Done

dts/arm/st/u5/stm32u5.dtsi
juickar: Done

dts/arm/st/wba/stm32wba.dtsi
juickar: Done

drivers/watchdog/wdt_iwdg_stm32.c
juickar: Done

mathieuchopstm (dismissed):

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -150,12 +204,17 @@ static int iwdg_stm32_install_timeout(const struct device *dev,
...
 	uint32_t timeout = config->window.max * USEC_PER_MSEC;
 	uint32_t prescaler = 0U;
 	uint32_t reload = 0U;
 
 	if (config->callback != NULL) {
+#ifdef CONFIG_IWDG_STM32_EARLY_WAKEUP
+		data->callback = config->callback;
+#else
 		return -ENOTSUP;
+#endif /* CONFIG_IWDG_STM32_EARLY_WAKEUP */
```
erwango: We can probably use `IS_ENABLED()` here

drivers/watchdog/wdt_iwdg_stm32.h
```diff
@@ -28,6 +28,10 @@ struct iwdg_stm32_config {
 struct iwdg_stm32_data {
 	uint32_t prescaler;
 	uint32_t reload;
+	/* IWDG user defined callback on EWI */
+	wdt_callback_t callback;
 };
 
+#define IWDG_STM32_STRUCT(dev) (((const struct iwdg_stm32_config *)dev->config)->instance)
```
erwango: Those (`WWDG_STM32_STRUCT` and friends ) should actually be removed rather than added.
They were all over but were cleaned up, but wwdg felt through the cracks, as it seems.

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +31,25 @@ config IWDG_STM32_INITIAL_TIMEOUT
 	  Limiting maximum timeout to a safe value of 26214 ms here, which was
 	  calculated for highest LSI frequency among STM32 MCUs of 40 kHz.
 
+config IWDG_STM32_EARLY_WAKEUP
+	bool "Early Wakeup Interrupt support for STM32 IWDG"
+	default y
```
erwango: As most users didn't need it until now, keep default disabled unless there is a good reason to do otherwise.
Drawback is that we need to enable at least once in wwdg tests using a board overlay to be sure it is compiled in CI.

erwango (changes requested):
Otherwise LGTM

drivers/watchdog/Kconfig.stm32
mathieuchopstm: (Note that leaving it enabled by default should not be a change for users since `callback=NULL` is valid)

erwango: Yeah, but has footprint impact.

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -150,12 +204,17 @@ static int iwdg_stm32_install_timeout(const struct device *dev,
 				      const struct wdt_timeout_cfg *config)
 {
 	struct iwdg_stm32_data *data = dev->data;
+	__maybe_unused const struct iwdg_stm32_config *cfg = dev->config;
```
GeorgeCGV: Doesn't appear to be used.

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +31,25 @@ config IWDG_STM32_INITIAL_TIMEOUT
 	  Limiting maximum timeout to a safe value of 26214 ms here, which was
 	  calculated for highest LSI frequency among STM32 MCUs of 40 kHz.
 
+config IWDG_STM32_EARLY_WAKEUP
+	bool "Early Wakeup Interrupt support for STM32 IWDG"
+	default n
```
mathieuchopstm: `n` is the default
```suggestion
```

drivers/watchdog/wdt_iwdg_stm32.c
```diff
@@ -54,6 +54,39 @@
 #define IWDG_SR_UPDATE_TIMEOUT	(6U * IWDG_PRESCALER_MAX * \
 				 MSEC_PER_SEC / LSI_VALUE)
 
+#ifdef CONFIG_IWDG_STM32_EARLY_WAKEUP
+
+void iwdg_stm32_isr(const struct device *dev)
+{
+	struct iwdg_stm32_data *data = ((struct iwdg_stm32_data *const)(dev)->data);
```
mathieuchopstm: No cast needed. (Also, `const` in `xxx *const` is useless)
```suggestion
	struct iwdg_stm32_data *data = dev->data;
```

drivers/watchdog/Kconfig.stm32
```diff
@@ -30,6 +31,24 @@ config IWDG_STM32_INITIAL_TIMEOUT
...
 	  calculated for highest LSI frequency among STM32 MCUs of 40 kHz.
 
+config IWDG_STM32_EARLY_WAKEUP
+	bool "Early Wakeup Interrupt support for STM32 IWDG"
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_WATCHDOG),interrupts)
+	help
+	  Enable support for Early Wakeup Interrupt.
+
+config IWDG_STM32_EWI_TIME
+	int "Value for IWDG early wake-up in ticks"
```
mathieuchopstm, line 41: I guess this should `depends on IWDG_STM32_EARLY_WAKEUP_TIME`

samples/drivers/watchdog/sample.yaml
```diff
@@ -101,6 +101,22 @@ tests:
       - stm32h573i_dk
     integration_platforms:
       - nucleo_f103rb
+  sample.drivers.watchdog.stm32_iwdg_ewi:
+    extra_args: DTC_OVERLAY_FILE=boards/stm32_iwdg.overlay
+    build_only: true
```
mathieuchopstm: Any reason for `build_only`?

drivers/watchdog/Kconfig.stm32
juickar: done

samples/drivers/watchdog/sample.yaml
juickar: removed

mathieuchopstm (approved):

erwango (approved):

PR #99863, opened 2025-11-23
ESP32-S3: Restructure LCD_CAM DT layout and add parallel LCD support

xingrz:
The ESP32 LCD_CAM peripheral contains two functional modules: an LCD output block and a DVP camera input block, sharing the same clock and pinctrl resources.

In the current tree, only the camera input path is implemented as a video driver.

This PR restructures the LCD_CAM Devicetree layout by introducing a common `&lcd_cam` node and splitting its functionality into `&lcd_cam_dvp` and `&lcd_cam_disp` child nodes. It also adds support for the LCD output module through a new MIPI-DBI driver.

The new MIPI-DBI driver is enabled on the DNESP32S3B board to drive its I8080-interface LCD.

Related HAL update: https://github.com/zephyrproject-rtos/hal_espressif/pull/501

https://github.com/user-attachments/assets/c5c43adb-9a0c-45d6-b47b-e6b6d0c89fc7

Closes #88385

PR #99114, opened 2025-11-08
Add board Alientek ATK-DNESP32S3B

xingrz:
[DNESP32S3B](http://www.openedv.com/docs/boards/esp32/ATK-DNESP32S3BVXX.html) is a development board from Alientek, featuring a compatible ESP32-S3-WROOM-1-N16R8 module and rich set of on-board peripherals, providing an out-of-the-box development experience for AIoT applications.

image

PR #100039, opened 2025-11-25
boards: alientek: dnesp32s3b: Enable additional peripherals

xingrz:
This PR builds on #99114 and enables additional peripherals on the DNESP32S3B board. DT configurations are added for the following on-board peripherals with existing in-tree support:

* AT24x EEPROM
* Speaker and DMIC
* microSD card slot
* USB device controller
* 2 PH-2.0 UART connectors (pinctrl only, remain disabled)

Support for the I8080-driven LCD is submitted separately in #99863.

Remaining peripherals:

* QMI8658A 6-axis IMU - not yet supported; may be implemented in the future
* AP3216C ALS&PS - not planned; manufacturer is gone and documentation is no longer accessible

### DIFF
--- a/samples/drivers/i2s/echo/boards/dnesp32s3b_procpu.overlay
+++ b/samples/drivers/i2s/echo/boards/dnesp32s3b_procpu.overlay
@@ -1,0 +1,6 @@
+/*
+ * Copyright (c) 2025 Chen Xingyu <hi@xingrz.me>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+i2s_rxtx: &i2s0 {};
--- a/boards/alientek/dnesp32s3b/dnesp32s3b-pinctrl.dtsi
+++ b/boards/alientek/dnesp32s3b/dnesp32s3b-pinctrl.dtsi
@@ -20,6 +20,30 @@
 		};
 	};
 
+	uart1_default: uart1_default {
+		group1 {
+			pinmux = <UART1_TX_GPIO18>;
+			output-high;
+		};
+
+		group2 {
+			pinmux = <UART1_RX_GPIO8>;
+			bias-pull-up;
+		};
+	};
+
+	uart2_default: uart2_default {
+		group1 {
+			pinmux = <UART2_TX_GPIO5>;
+			output-high;
+		};
+
+		group2 {
+			pinmux = <UART2_RX_GPIO6>;
+			bias-pull-up;
+		};
+	};
+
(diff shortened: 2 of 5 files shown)


sylvioalves (dismissed):

samples/drivers/i2s/echo/boards/dnesp32s3b_procpu.overlay
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) 2025 Chen Xingyu <hi@xingrz.me>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+i2s_rxtx: &i2s0 {};
```
sylvioalves, line 6: Any reason not to enable dma here?
```c
i2s_rxtx: &i2s0 {
	status = "okay";

	dmas = <&dma 2>, <&dma 3>;
	dma-names = "rx", "tx";
};
```

xingrz:
> I think the I2S audio support commit should be placed prior the I2S echo sample overlay addition, right?

Do you mean placing the board I2S audio support commit before the sample overlay commit? If so, that ordering is already in place. Please let me know if I misunderstood.

samples/drivers/i2s/echo/boards/dnesp32s3b_procpu.overlay
xingrz: It's already enabled at SoC level

https://github.com/zephyrproject-rtos/zephyr/blob/15716542c20dea8d32a4e8e3072a90c6635ee589/dts/xtensa/espressif/esp32s3/esp32s3_common.dtsi#L319-L320

sylvioalves: what about `status = "okay";`?

xingrz: It is already enabled at the board level in `boards/alientek/dnesp32s3b/dnesp32s3b_procpu.dts`. Unlike `esp32s3_devkitc_procpu.overlay`, which is a general-purpose development board kept minimal for flexibility, the DNESP32S3B is a production board with a fixed audio pinout, so I think enabling I2S at the board level is appropriate here.

sylvioalves: I missed that, thanks!

sylvioalves (dismissed):

PR #100042, opened 2025-11-25
drivers: haptics: Add support for CS40L5x haptic driver

Liam-Ogletree:
This PR includes three commits:

1. Adds haptics driver for Cirrus Logic CS40L5x haptic parts
2. Adds a sample application for CS40L5x haptic parts, which includes a custom shell interface for demonstration
3. Adds CS40L5x to `zephyr/tests/drivers/build_all/haptics/`

The driver and sample application were both tested with a NUCLEO-F401RE development platform and a stand-alone demo platform. All features -- including I2C, ROM effect playback, custom PCM/PWLE effect playback, "buzz" playback, edge-triggered playback, gain configuration, and runtime haptics logging -- were tested with each of the hardware variants (50/51/52/53).


include/zephyr/drivers/haptics/cs40l5x.h
```diff
@@ -0,0 +1,473 @@
+/*
+ * Copyright (c) 2025 Cirrus Logic, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DRIVERS_HAPTICS_CS40L5X_H_
+#define ZEPHYR_INCLUDE_DRIVERS_HAPTICS_CS40L5X_H_
+
+#include "zephyr/drivers/gpio.h"
+#include "zephyr/drivers/i2c.h"
```
rriveramcrus: Should be < > instead of " "

include/zephyr/drivers/haptics/cs40l5x.h
```diff
@@ -0,0 +1,473 @@
...
+	/**< Optional GPIO for hardware and DSP interrupts */
+	struct gpio_dt_spec interrupt_gpio;
+	/**< Optional GPIOs for edge-triggered haptic effects */
+	struct cs40l5x_trigger_gpios trigger_gpios;
+	/**< Optional flash storage, configurable via devicetree property 'flash-storage' */
+	const struct device *flash;
+	/**< Optional flash register offset, configurable via devicetree property 'flash-offset' */
+	const off_t flash_offset;
+	/**< Boost setting, configurable via devicetree property 'external-boost' */
+	const bool external_boost;
```
rriveramcrus: Nit: A pointer to a `struct device` that represents the external boost converter would be advantageous to have in the event it could be switched off when unused, but I am okay to leave this for a future PR.

drivers/haptics/Kconfig.cs40l5x
```diff
@@ -0,0 +1,87 @@
...
+		Enable upload and playback of custom PCM and PWLE waveforms.
+
+config HAPTICS_CS40L5X_DYNAMIC_F0
+	bool "Dynamic F0"
+	default y
+	help
+		Enable dynamic F0 compensation for haptic effects.
+
+config HAPTICS_CS40L5X_LOGGER
+	bool "Haptics Logging"
```
rriveramcrus: This is a bit confusing and could be conflated with the existing kernel logging. Maybe "DSP Logging" is more appropriate.

include/zephyr/drivers/haptics/cs40l5x.h
Liam-Ogletree: Changed `external_boost` to be a `struct device` and updated `cirrus,cs40l5x-base.yaml` and core driver files accordingly. In the driver:

- checks for `config->external_boost` now check whether `config->external_boost != NULL` 
- updated devicetree macro
- updated `CS40L5X_ANY_DEV_USE_EXTERNAL_BOOST` and `CS40L5X_ANY_DEV_USE_INTERNAL_BOOST` accordingly (from `DT_ANY_INST_HAS_BOOL_STATUS_OKAY` -> `DT_ANY_INST_HAS_PROP_STATUS_OKAY`)
- added `regulator_enable()` and `regulator_disable()` to PM functions `cs40l5x_pm_resume()` and `cs40l5x_pm_suspend()`, respectively

drivers/haptics/Kconfig.cs40l5x
Liam-Ogletree: done

include/zephyr/drivers/haptics/cs40l5x.h
Liam-Ogletree: done

rriveramcrus (dismissed):
I'm pretty happy with this, @kartben would you be able to take a look at this or should I look for some volunteers?

drivers/haptics/Kconfig.cs40l5x
```diff
@@ -0,0 +1,87 @@
+# Copyright 2025 Cirrus Logic, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig HAPTICS_CS40L5X
+	bool "CS40L5x Haptics Driver"
+	default y
+	depends on DT_HAS_CIRRUS_CS40L5X_ENABLED
+	help
+		Enable CS40L5x haptics driver.
```
fabiobaltieri: indentation is off, look at other kconfig files

drivers/haptics/cs40l5x-i2c.c
```diff
@@ -0,0 +1,79 @@
...
+#include <zephyr/drivers/haptics/cs40l5x.h>
+
+#if DT_ANY_INST_ON_BUS_STATUS_OKAY(i2c)
+#include <zephyr/drivers/i2c.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/sys/byteorder.h>
+
+LOG_MODULE_REGISTER(CS40L5X_I2C, CONFIG_HAPTICS_LOG_LEVEL);
+
+#define CS40L5X_REG_WIDTH  (4)
```
fabiobaltieri: nit: drop the parenthesis, only neeed when you have arguments, no point adding noise

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+#define CS40L5X_ANY_DEV_USE_INTERRUPTS     DT_ANY_INST_HAS_PROP_STATUS_OKAY(int_gpios)
+#define CS40L5X_ANY_DEV_USE_TRIGGER_GPIOS  DT_ANY_INST_HAS_PROP_STATUS_OKAY(trigger_gpios)
+#define CS40L5X_ANY_DEV_USE_EXTERNAL_BOOST DT_ANY_INST_HAS_PROP_STATUS_OKAY(external_boost)
+#define CS40L5X_ANY_DEV_USE_INTERNAL_BOOST !DT_ALL_INST_HAS_PROP_STATUS_OKAY(external_boost)
+#define CS40L5X_ANY_DEV_USE_FLASH_STORAGE  DT_ANY_INST_HAS_PROP_STATUS_OKAY(flash_storage)
+
+#define CS40L5X_ANY_DEV_USE_HIBERNATION                                                            \
+	(IS_ENABLED(CONFIG_PM_DEVICE) && IS_ENABLED(CONFIG_PM_DEVICE_RUNTIME))
+
+#define CS40L5X_REG_DEVID              (0x00000000)
```
fabiobaltieri: same as above, bit of a nitpick but you can drop the parenthesis from literals, they add noise and nothing else,  just keep them for the ones that have ops (like CS40L5X_REG_IRQ1_MASK2)

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+	const struct cs40l5x_config *const config = dev->config;
+
+	return config->bus_io->write(dev, tx, len);
+}
+
+static int cs40l5x_regmap_multi_write(const struct device *const dev,
+				      const struct cs40l5x_multi_write *const multi_write,
+				      const uint32_t len)
+{
+	int error = 0;
```
fabiobaltieri: drop the intializer, I'd also use `ret` everywhere, seems to be the more common one in the code base, `return ret` in line 444, `return 0` at the end of the function

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+
+		if (reg_val == val) {
+			return 0;
+		}
+
+		(void)k_sleep(CS40L5X_T_DEFAULT_DELAY);
+	} while (!sys_timepoint_expired(end));
+
+	LOG_INST_ERR(config->log, "timed out polling 0x%08x, expected 0x%08x but received 0x%08x",
+		     addr, val, reg_val);
```
fabiobaltieri, line 473: heyd you reallly want the LOG_INST stuff? Just do `LOG_ERR` and drop the rest, can't see the benefit of it and it adds lot of noise in the code

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+}
+
+static inline bool cs40l5x_valid_wavetable_source(const struct device *const dev,
+						  const enum cs40l5x_bank bank, const uint8_t index)
+{
+	__maybe_unused struct cs40l5x_data *const data = dev->data;
+
+	switch (bank) {
+	case CS40L5X_ROM_BANK:
+		return (index < CS40L5X_NUM_ROM_EFFECTS);
```
fabiobaltieri: drop the parenthesis

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+	case CS40L5X_CUSTOM_BANK:
+		return (index > CS40L5X_NUM_CUSTOM_EFFECTS ||
+			!IS_ENABLED(CONFIG_HAPTICS_CS40L5X_CUSTOM_EFFECTS))
+			       ? false
+			       : data->custom_effects[index];
+	case CS40L5X_BUZ_BANK:
+		return (index < CS40L5X_NUM_BUZ_EFFECTS);
```
fabiobaltieri: same...

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+			return cs40l5x_regmap_poll(dev, CS40L5X_REG_DSP_V1MBOX,
+						   CS40L5X_EXP_MBOX_CLEAR, CS40L5X_T_MBOX_CLEAR);
+		}
+
+		(void)k_sleep(CS40L5X_T_DEFAULT_DELAY);
+	} while (!sys_timepoint_expired(end));
+
+	LOG_INST_ERR(config->log, "failed write to mailbox (%d)", error);
+
+	return error;
```
fabiobaltieri: `return 0`

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+}
+
+static int cs40l5x_increment_mailbox(const struct device *const dev, uint32_t *const mbox_ptr)
+{
+	*mbox_ptr = (*mbox_ptr + CS40L5X_REG_WIDTH < CS40L5X_REG_DSP_MBOX_END)
+			    ? *mbox_ptr + CS40L5X_REG_WIDTH
+			    : CS40L5X_REG_DSP_MBOX_BASE;
```
fabiobaltieri: use a normal `if`, the terniary here adds nothing in code compactness and just makes it hard to read

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+		return;
+	}
+
+	trigger_config = (item.edge == CS40L5X_GPIO_LOGIC_HIGH)
+				 ? config->trigger_gpios.rising_edge
+				 : config->trigger_gpios.falling_edge;
+
+	LOG_INST_INF(config->log, "trigger   | %s %u (%d dB)",
+		     cs40l5x_print_bank(trigger_config[item.gpio].bank),
+		     trigger_config[item.gpio].index, trigger_config[item.gpio].attenuation);
```
fabiobaltieri: should probably be `LOG_DBG`

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+		error = cs40l5x_regmap_read(dev, mbox_rd_ptr, &mbox_val);
+		if (error < 0) {
+			LOG_INST_DBG(config->log, "failed to read mailbox (%d)", error);
+			return error;
+		}
+
+		switch (mbox_val) {
+		case CS40L5X_MBOX_PLAYBACK_COMPLETE_MBOX:
+			data->effects_in_flight -= 1;
+			LOG_INST_INF(config->log, "complete  | mailbox playback");
```
fabiobaltieri: should also probably be LOG_DBG, same for the rest of the file

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+}
+
+static int cs40l5x_process_interrupts(const struct device *const dev,
+				      const uint32_t *const irq_ints)
+{
+	__maybe_unused const struct cs40l5x_config *const config = dev->config;
+	int error = 0;
+
+	if (FIELD_GET(CS40L5X_MASK_IRQ1_AMP, irq_ints[CS40L5X_INT1]) != 0) {
+		LOG_INST_ERR(config->log, "amplifier short failure");
```
fabiobaltieri: the should probably be warning? the whole model here is weird, they print a message and nothing else, plus what's the point of retunring -EIO? It's not an IO error, the function worked correctly, it processes interrupts, maybe you want to have a callback system and let the application do something with this stuff?

drivers/haptics/cs40l5x-i2c.c
```diff
@@ -0,0 +1,79 @@
...
+static struct device *cs40l5x_regmap_get_device_i2c(const struct device *const dev)
+{
+	const struct cs40l5x_config *const config = dev->config;
+
+	return (struct device *)&config->bus.i2c.bus;
+}
+
+static int cs40l5x_regmap_read_i2c(const struct device *const dev, const uint32_t addr,
+				   uint32_t *const rx, const uint32_t len)
+{
```
fabiobaltieri: why is this called regmap? that has a very specifc meaning, these seems to be just normal read and write functions

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+	return (calibration->f0 == CS40L5X_FLASH_MEMORY_ERASED &&
+		calibration->redc == CS40L5X_FLASH_MEMORY_ERASED);
+}
+
+static int cs40l5x_load_calibration(const struct device *const dev)
+{
+	const struct cs40l5x_config *const config = dev->config;
+	struct cs40l5x_data *const data = dev->data;
+	struct cs40l5x_calibration calibration = {};
+	int error, __maybe_unused warning;
```
fabiobaltieri: doesn't look maybe unused does it?

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+		LOG_INST_DBG(config->log, "unable to reset DSP mailbox (%d)", error);
+	}
+
+	return error;
+}
+
+static int cs40l5x_bringup(const struct device *const dev)
+{
+	const struct cs40l5x_config *const config = dev->config;
+	int error, warning;
```
fabiobaltieri: does not look like these two interact with each other? just `int ret`, simple and short

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+{
+	uint32_t addr, offset, sample = 0;
+	uint8_t current_word = 0;
+	uint16_t i = 0;
+	int error;
+
+	addr = (index == CS40L5X_CUSTOM_0) ? CS40L5X_REG_CUSTOM_DATA_0 : CS40L5X_REG_CUSTOM_DATA_1;
+
+	while (i < num_samples) {
+		sample |= FIELD_PREP(GENMASK(23, 16), (uint8_t)samples[i++]);
```
fabiobaltieri: this is a 0 |= return value, just direct assign and drop the initializers

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+
+		if (i < num_samples) {
+			sample |= FIELD_PREP(GENMASK(15, 8), (uint8_t)samples[i++]);
```
fabiobaltieri: I suggest moving `i++` in its own line, having it hidden in there makes this hard to follow

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+
+	for (int i = 0; i < gpios->num_gpio; i++) {
+		warning = gpio_pin_configure_dt(&gpios->gpio[i], GPIO_OUTPUT);
+		if (warning < 0) {
+			LOG_INST_DBG(config->log, "skipped %s (%d)", gpios->gpio[i].port->name,
+				     warning);
+			continue;
+		}
+
+		gpios->ready[i] = true;
```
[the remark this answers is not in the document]
Liam-Ogletree: `ready` is `bool *ready` (not something like `bool ready[3]`), so I'm not violating const correctness in this case. This is similar to modifying `foo` in the following example:

`void fn(uint32_t *const foo) { *foo = 0; /* this is okay */ }`

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+		LOG_INST_WRN(config->log, "calibration data not found (%d)", -EINVAL);
+	} else {
+		data->calibration = calibration;
+
+		LOG_INST_INF(config->log, "Loaded    | ReDC: 0x%08X, F0: 0x%08X",
+			     data->calibration.redc, data->calibration.f0);
+	}
+
+	warning = pm_device_runtime_put(config->flash);
+	if (error < 0) {
```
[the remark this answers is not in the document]
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+
+	error = flash_write(config->flash, config->flash_offset, &data->calibration,
+			    sizeof(data->calibration));
+	if (error < 0) {
+		LOG_INST_ERR(config->log, "failed write to flash storage (%d)", error);
+	}
+
+exit_pm:
+	warning = pm_device_runtime_put(config->flash);
+	if (error < 0) {
```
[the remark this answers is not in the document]
Liam-Ogletree: done

samples/drivers/haptics/cs40l5x/src/main.c
```diff
@@ -0,0 +1,372 @@
...
+#include <zephyr/drivers/gpio.h>
+#endif /* CS40L5X_DEMO_TRIGGER */
+
+#if CONFIG_SHELL
+#include <stdlib.h>
+#endif /* CONFIG_SHELL */
+
+#define LOG_LEVEL LOG_LEVEL_DBG
+LOG_MODULE_REGISTER(main, LOG_LEVEL_DBG);
+
```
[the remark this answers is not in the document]
Liam-Ogletree: done

drivers/haptics/cs40l5x-i2c.c
```diff
@@ -0,0 +1,79 @@
...
+	const struct cs40l5x_config *const config = dev->config;
+	int error;
+
+	(void)sys_put_be32(addr, (uint8_t *)&addr);
+
+	error = i2c_write_read_dt(&config->bus.i2c, (uint8_t *)&addr, CS40L5X_ADDR_WIDTH,
```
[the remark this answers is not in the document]
Liam-Ogletree: done

drivers/haptics/cs40l5x-i2c.c
```diff
@@ -0,0 +1,79 @@
...
+static int cs40l5x_regmap_write_i2c(const struct device *const dev, const uint32_t *const tx,
+				    const uint32_t len)
+{
+	const struct cs40l5x_config *const config = dev->config;
+
+	for (int i = 0; i < len; i++) {
+		(void)sys_put_be32(tx[i], (uint8_t *)&tx[i]);
+	}
+
+	return i2c_write_dt(&config->bus.i2c, (uint8_t *)tx, len * CS40L5X_REG_WIDTH);
```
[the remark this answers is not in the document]
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2713 @@
...
+	__maybe_unused struct cs40l5x_data *const data = dev->data;
+
+	switch (bank) {
+	case CS40L5X_ROM_BANK:
+		return (index < CS40L5X_NUM_ROM_EFFECTS);
+	case CS40L5X_CUSTOM_BANK:
+		return (index > CS40L5X_NUM_CUSTOM_EFFECTS ||
+			!IS_ENABLED(CONFIG_HAPTICS_CS40L5X_CUSTOM_EFFECTS))
+			       ? false
+			       : data->custom_effects[index];
```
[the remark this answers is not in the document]
Liam-Ogletree: This is not a logic error -- it intentionally evaluates to false for valid indices if `CONFIG_HAPTICS_CS40L5X_CUSTOM_EFFECTS=n`, indicating the user does not allow for uploading/playback of custom PCM/PWLE effects.

Nevertheless, I think the proposed change is cleaner and have updated the code accordingly.

drivers/haptics/cs40l5x-i2c.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done -- applied everywhere

drivers/haptics/cs40l5x.c
Liam-Ogletree: done (and updated all `int error` to be `int ret`)

drivers/haptics/cs40l5x.c
Liam-Ogletree: A common use case (especially for automotive applications) for this part will have multiple CS40L5x's on a shared bus, so we expect instance-specific debug/information/warning/error logging to be helpful. Unless `LOG_INST_*` is discouraged for other reasons, I think we want to keep these.

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done -- applied everywhere

drivers/haptics/cs40l5x.c
Liam-Ogletree: These are all fatal hardware errors, so the thought (for the time being) was to just log the error and return an error code to force exit the rest of the interrupt handler.

However, I think having a registrable error callback to let the application attempt recovery as you suggested is much better. I've added `void cs40l5x_register_error_callback()` (taking the `register_error_callback()` API from the audio codec API as inspiration) and updated the driver accordingly. I've also added a dummy callback in the sample application to demonstrate what the application callback needs to look like. Thanks for this feedback!

**Separately, I think the haptics API could probably use an error callback system comparable to the audio codec API's, which I can look into after this PR.**

drivers/haptics/cs40l5x-i2c.c
Liam-Ogletree: updated in `cs40l5x-i2c.c`, `cs40l5x-spi.c`, and `cs40l5x.h`

drivers/haptics/cs40l5x.c
Liam-Ogletree: I would agree, but the sonarqubecloud bot flagged this for both` cs40l5x_load_calibration()` and `cs40l5x_store_calibration()`. I've removed the `__maybe_unused`, and we can decide whether to accept the sonarqubecloud "issue" for this as erroneous?

drivers/haptics/cs40l5x.c
Liam-Ogletree: done -- simplified in other cases where appropriate

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/Kconfig.cs40l5x
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: Weird -- sonarqubecloud bot didn't complain this time, so I guess we're good. Thanks for the feedback on this!

drivers/haptics/cs40l5x.c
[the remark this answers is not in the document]
Liam-Ogletree: I suppose this is a better example of what is going on:

image

drivers/haptics/cs40l5x-i2c.c
fabiobaltieri: still relevant

drivers/haptics/cs40l5x-i2c.c
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @file
+ * @brief I2C Driver for Cirrus Logic CS40L5x Haptic Devices
+ */
+
+#define DT_DRV_COMPAT cirrus_cs40l5x
+
+#include <zephyr/drivers/haptics/cs40l5x.h>
+
+#if DT_ANY_INST_ON_BUS_STATUS_OKAY(i2c)
```
fabiobaltieri, line 16: the condition for this to be referenced is `DT_INST_ON_BUS(inst, i2c),` how about matching it in the code? seems like a more defensive pattern, same (inverted) for spi

drivers/haptics/Kconfig.cs40l5x
```diff
@@ -0,0 +1,87 @@
...
+
+if HAPTICS_CS40L5X
+
+config HAPTICS_CS40L5X_I2C
+	def_bool $(dt_compat_on_bus,$(DT_COMPAT_CIRRUS_CS40L5X),i2c)
+	select I2C
+
+config HAPTICS_CS40L5X_SPI
+	def_bool $(dt_compat_on_bus,$(DT_COMPAT_CIRRUS_CS40L5X),spi)
+	select SPI
```
fabiobaltieri, line 19: these are only for the `select I2C` and `select SPI` right? just checking, there's no reference to the symbol itself

drivers/haptics/cs40l5x-i2c.c
fabiobaltieri: I still see a lot of references in drivers/haptics/cs40l5x.c

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2757 @@
...
+
+LOG_MODULE_REGISTER(CS40L5X, CONFIG_HAPTICS_LOG_LEVEL);
+
+#define CS40L5X_ANY_DEV_USE_INTERRUPTS     DT_ANY_INST_HAS_PROP_STATUS_OKAY(int_gpios)
+#define CS40L5X_ANY_DEV_USE_TRIGGER_GPIOS  DT_ANY_INST_HAS_PROP_STATUS_OKAY(trigger_gpios)
+#define CS40L5X_ANY_DEV_USE_EXTERNAL_BOOST DT_ANY_INST_HAS_PROP_STATUS_OKAY(external_boost)
+#define CS40L5X_ANY_DEV_USE_INTERNAL_BOOST !DT_ALL_INST_HAS_PROP_STATUS_OKAY(external_boost)
+#define CS40L5X_ANY_DEV_USE_FLASH_STORAGE  DT_ANY_INST_HAS_PROP_STATUS_OKAY(flash_storage)
+
+#define CS40L5X_ANY_DEV_USE_HIBERNATION                                                            \
```
fabiobaltieri: nit, just `#define CS40L5X_ANY_DEV_USE_HIBERNATION \`, no need to throw the `\` on the other side of the screen if there's only one line

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2757 @@
...
+		return "BUZ";
+	default:
+		return NULL;
+	}
+}
+
+static inline bool cs40l5x_valid_wavetable_source(const struct device *const dev,
+						  const enum cs40l5x_bank bank, const uint8_t index)
+{
+	__maybe_unused struct cs40l5x_data *const data = dev->data;
```
fabiobaltieri: don't think you need the `__maybe_unused` here, can you try to drop it?

drivers/haptics/cs40l5x.c
fabiobaltieri: k it's just that all those `__maybe_unused` are a bit annoying, I presume you get a warning if you have no logging and drop those right?

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2757 @@
...
+	if (*index == gpios->num_gpio) {
+		return -EINVAL;
+	}
+
+	return 0;
+}
+
+static void cs40l5x_trigger_log(const struct device *const dev)
+{
+	__maybe_unused struct cs40l5x_trigger_config *trigger_config;
```
fabiobaltieri: same `__maybe_unused` should not cause problems here since you always assign it

drivers/haptics/cs40l5x.c
```diff
@@ -0,0 +1,2757 @@
...
+	warning = k_mutex_unlock(&data->lock);
+	if (warning < 0) {
+		LOG_INST_DBG(config->log, "failed mutex unlock (%d)", warning);
+	}
+
+error_pm:
+	warning = pm_device_runtime_put(dev);
+	if (warning < 0) {
+		LOG_INST_DBG(config->log, "failed PM put for device (%d)", warning);
+	}
```
fabiobaltieri: wouldn't bother checking the return value for these, if these start failing the system is likely really broken really bad, and you probably don't have the debugging strings on anyway, better value in having simpler code imo

samples/drivers/haptics/cs40l5x/src/main.c
```diff
@@ -0,0 +1,378 @@
...
+	},
+	{
+		.duration = 0x0003,
+		.level = 0x200,
+		.frequency = 0x320,
+		.flags = 0x3,
+	},
+};
+
+void cs40l5x_dummy_callback(const struct device *const dev, const uint32_t errors)
```
fabiobaltieri: `static`

samples/drivers/haptics/cs40l5x/src/main.c
```diff
@@ -0,0 +1,378 @@
...
+const int8_t pcm_samples[78] = {
+	0x00, 0x07, 0x0D, 0x12, 0x18, 0x1E, 0x22, 0x27, 0x2B, 0x2D, 0x30, 0x31, 0x32,
+	0x32, 0x31, 0x30, 0x2D, 0x2B, 0x27, 0x22, 0x1E, 0x18, 0x12, 0x0D, 0x07, 0x00,
+	0xF9, 0xF3, 0xED, 0xE8, 0xE2, 0xDD, 0xD9, 0xD5, 0xD2, 0xD0, 0xCE, 0xCD, 0xCD,
+	0xCE, 0xD0, 0xD2, 0xD5, 0xD9, 0xDD, 0xE2, 0xE8, 0xED, 0xF3, 0xF9, 0x00, 0x07,
+	0x0D, 0x12, 0x18, 0x1E, 0x22, 0x27, 0x2B, 0x2D, 0x30, 0x31, 0x32, 0x32, 0x31,
+	0x30, 0x2D, 0x2B, 0x27, 0x22, 0x1E, 0x18, 0x12, 0x0D, 0x07, 0x00, 0xF9, 0xF3,
+};
+
+const struct cs40l5x_pwle_section pwle_sections[] = {
```
fabiobaltieri: `static`

samples/drivers/haptics/cs40l5x/src/main.c
```diff
@@ -0,0 +1,378 @@
...
+	SHELL_CMD_ARG(stop, NULL, CS40L5X_STOP_HELP, cmd_stop, 1, 0),
+#if CS40L5X_DEMO_TRIGGER
+	SHELL_CMD_ARG(trigger, &dsub_gpio_name, CS40L5X_TRIGGER_HELP, cmd_trigger, 3, 0),
+#endif /* CS40L5X_DEMO_TRIGGER */
+	SHELL_SUBCMD_SET_END);
+
+SHELL_CMD_REGISTER(cs40l5x, &cs40l5x_cmds, "CS40L5x shell commands", NULL);
+#endif /* CONFIG_SHELL */
+
+const int8_t pcm_samples[78] = {
```
fabiobaltieri: `static`

drivers/haptics/cs40l5x-i2c.c
Liam-Ogletree: fixed

drivers/haptics/cs40l5x.c
Liam-Ogletree: Yes -- I've tried to drop as many `__maybe_unused` as possible, see my reply on a related comment

> I've dropped the `__maybe_unused` where possible (still have the prefix before `cs40l5x_deinit()` and in specific instances where twister tests will fail otherwise), which were added after getting sonarqubecloud bot warnings on my first submission. I would be happy to accept the rest of these warnings, because I agree they are not needed

drivers/haptics/cs40l5x-i2c.c
Liam-Ogletree: fixed

drivers/haptics/cs40l5x-i2c.c
Liam-Ogletree: I might be misunderstanding, but given that a typical use case for this device will have multiple CS40L5Xs on a shared bus, the current implementation seems simpler than iterating over all DT_DRV_COMPAT instances. My preference would be to keep this as-is using [`DT_ANY_INST_ON_BUS_STATUS_OKAY()`](https://docs.zephyrproject.org/latest/doxygen/html/group__devicetree-inst.html#gaa4ff1fe4242399fbd667fbee7e98040a)

drivers/haptics/Kconfig.cs40l5x
Liam-Ogletree: yes, these are just used for the `select I2C` and `select SPI`

drivers/haptics/cs40l5x.c
Liam-Ogletree: done

drivers/haptics/cs40l5x.c
Liam-Ogletree: I've dropped the `__maybe_unused` where possible (still have the prefix before `cs40l5x_deinit()` and in specific instances where twister tests will fail otherwise), which were added after getting sonarqubecloud bot warnings on my first submission. I would be happy to accept the rest of these warnings, because I agree they are not needed

drivers/haptics/cs40l5x.c
Liam-Ogletree: see comment above

drivers/haptics/cs40l5x.c
Liam-Ogletree: Removed checks for all instances of `pm_device_runtime_put()` and `k_mutex_unlock()` and replaced with `(void)pm_device_runtime_get()` and `(void)k_mutex_unlock()`

samples/drivers/haptics/cs40l5x/src/main.c
Liam-Ogletree: done

samples/drivers/haptics/cs40l5x/src/main.c
Liam-Ogletree: done

samples/drivers/haptics/cs40l5x/src/main.c
Liam-Ogletree: done

fabiobaltieri (approved):

rriveramcrus (approved):

PR #100047, opened 2025-11-25
maintainer: hal_openisa: Add zejiang0jason as collaborator

dleach02:
Adding Zejiang0jason as collaborator to the openisa HAL. He was the developer of the original SDK.

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -5768,6 +5768,7 @@ West:
   status: odd fixes
   collaborators:
     - dleach02
+    - zejiang0jason
   files:
     - modules/Kconfig.vega
   labels:


dleach02:
Just a side note, there has not been an issue or PR towards this platform in years (excepting the HWMv2).

MAINTAINERS.yml
```diff
@@ -5768,6 +5768,7 @@ West:
   status: odd fixes
```
nashif, line 5768: we need a maintainer here, otherwise any changes submitted here will not be able to be merged

nashif (approved):

MAINTAINERS.yml
dleach02: I think we should actually remove this ... but lets consider moving Zejiang to maintainer as the next step ;-)

zejiang0jason (approved):

PR #100048, opened 2025-11-25
random: ctr-drbg: Use PSA crypto instead of MBED TLS

ceolin:
Change ctr_drb random implementation to use PSA Crypto API. Since name convention is very different and PSA abstracts the algorithm used to generate CSPRNG, file and Kconfig options were changed. Current symbols were deprecated and just select the new one.


doc/services/crypto/random/index.rst
yvesll: As `CTR_DRBG_CSPRNG_GENERATOR` was deprecated, the default value shall be updated. Also for Kconfig source like https://github.com/zephyrproject-rtos/zephyr/blob/dbdd5e91e716c6db1c1f81f804ed8fe3d0961977/drivers/entropy/Kconfig.mcux#L56

subsys/random/random_psa.c
```diff
@@ -0,0 +1,38 @@
...
+#include <psa/crypto.h>
+
+static bool _initialised;
+static K_MUTEX_DEFINE(_lock);
+
+int z_impl_sys_csrand_get(void *dst, size_t outlen)
+{
+	int ret = 0;
+
+	k_mutex_lock(&_lock, K_FOREVER);
```
yvesll: IMO, the lock only need be applied to `psa_crypto_init`

subsys/random/random_psa.c
```diff
@@ -0,0 +1,38 @@
...
+		if (psa_crypto_init() != PSA_SUCCESS) {
+			ret = -EIO;
+			goto end;
+		}
+
+		_initialised = true;
+	}
+
+	if (psa_generate_random((uint8_t *)dst, outlen) != PSA_SUCCESS) {
+		ret = -EIO;
```
yvesll: Can you use more errnos, maybe a map to `psa_status_t`

subsys/random/random_psa.c
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2019, NXP
```
yvesll, line 2: Need update.

doc/services/crypto/random/index.rst
yvesll, line 62: BTW, can you help wrap `CSPRNG_GENERATOR_CHOICE` and `RNG_GENERATOR_CHOICE` in line 38 with backtick. It would be nice to have it 😄 .

ceolin: yep, good catch !

subsys/random/random_psa.c
ceolin: I thought about it too, but was worried about having it duplicated all over the code. I think we should probably have a common code doing that since `psa api` is used in multiple places. Are you okay with me addressing it in a separate pr ?

subsys/random/random_psa.c
ceolin: sure

subsys/random/random_psa.c
ceolin: yep, you're right.

subsys/random/random_psa.c
yvesll: Sure, thank you.

valeriosetti (changes requested):
Question: isn't the result of this PR a sort of duplication to the combination of [entropy_psa_crypto.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/entropy/entropy_psa_crypto.c) + [random_entropy_device.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/random/random_entropy_device.c)?

I'm not against it, but since PSA Crypto API provide random values and not just entropy, we might want to remove `entropy_psa_crypto.c` and just keep this implementation. Wdyt?

valeriosetti (approved):

frkv (approved):
This PR looks good for me, assuming we verify the kernel memory management/protection in follow up work (ref discussions in Zephyr Security WG)...

yvesll (approved):

PR #100051, opened 2025-11-25
Remove self-test from initialization probe.

markrages:
Self test requires the accelerometer to be in a stable orientation. Therefore it should not be a part of initialization, or init will fail when the device is moving around.


ubieda (dismissed):

jeppenodgaard (dismissed):

ttmut:
What is the intended use case for `adxl367_self_test`? It is kind of an orphan function now as there are no consumers of it and it is not declared in any header file for external usage, even though it is non-static.

PR #100057, opened 2025-11-25
boards: shields: m5stack_cardputer: Correct the display parameter

vlone836:
Refer to the M5Stack official code, it need to correct the value of x-offset of st7789v display driver from 53 to 52, or the leftmost column of the LCD shows an anomaly.

### DIFF
--- a/boards/shields/m5stack_cardputer/m5stack_cardputer.overlay
+++ b/boards/shields/m5stack_cardputer/m5stack_cardputer.overlay
@@ -33,7 +33,7 @@
 
 			width = <135>;
 			height = <240>;
-			x-offset = <53>;
+			x-offset = <52>;
 			y-offset = <40>;
 
 			vcom = <0x28>;


boards/shields/m5stack_cardputer/m5stack_cardputer.overlay
vlone836, line 1: Refer to the M5Stack official code: https://github.com/m5stack/M5GFX/blob/7614fa3961b8d6e797ccbfa2ef73359698236c03/src/M5GFX.cpp#L340C7-L340C30

vlone836: before：
before
after:
after

vlone836: ST7789V datasheet:
https://m5stack.oss-cn-shenzhen.aliyuncs.com/resource/docs/datasheet/unit/lcd/ST7789V2_SPEC_V1.0.pdf

kartben (approved):

soburi (approved):

PR #100066, opened 2025-11-26
scripts: guiconfig: Add dark mode

yvesll:
TK under Windows/Linux did not have a native support for dark mode.

Kconfiglib: https://github.com/zephyrproject-rtos/Kconfiglib/pull/35

### DIFF
--- a/scripts/kconfig/guiconfig.py
+++ b/scripts/kconfig/guiconfig.py
@@ -277,6 +277,47 @@ def _needs_save():
     # No need to prompt for save
     return False
 
+def _detect_system_dark_mode():
+    if sys.platform == "win32":
+        try:
+            import winreg
+            registry = winreg.ConnectRegistry(None, winreg.HKEY_CURRENT_USER)
+            key = winreg.OpenKey(registry,
+                r"Software\Microsoft\Windows\CurrentVersion\Themes\Personalize")
+            value, _ = winreg.QueryValueEx(key, "AppsUseLightTheme")
+            winreg.CloseKey(key)
+            return value == 0
+        except (ImportError, OSError, WindowsError) as _:
+            return False
+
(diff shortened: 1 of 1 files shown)


tejlmand (changes requested):
PRs for kconfiglib and related tools should go here https://github.com/zephyrproject-rtos/Kconfiglib and from there be cherry-picked into Zephyr when approved.

Blocked until reviewed in kconfiglib repo.

scripts/kconfig/guiconfig.py
```diff
@@ -67,14 +69,14 @@
 
 if _PY2:
     # Python 2
-    from Tkinter import *
+    from Tkinter import * # novermin
```
tejlmand: why `# novermin` at the end ?

This is not part of code in: https://github.com/zephyrproject-rtos/Kconfiglib/blob/main/guiconfig.py

Comment also applies to other occurrences. in this file.

yvesll: to suppress python lint error

yvesll: As kconfiglib did not clean up python2 code, it may have some lint issues during zephyr compliance check.

tejlmand: there was a Python2 cleanup done here: https://github.com/zephyrproject-rtos/Kconfiglib/pull/22

If anything was missed in that PR then it would be better to make a followup PR.
We should try to avoid, or at least minimize, differences between Zephyr and its upstream counterpart.

yvesll: Make sense, thanks.

tejlmand (approved):

yvesll:
@nashif please take a review, thank you.

carlescufi (approved):

PR #100067, opened 2025-11-26
Bluetooth: Classic: HFP: Fix SCO conn cannot be released issue

lylezhu2012:
There is an issue found that the sco conn cannot be released when the SCO has been disconnected. The sco conn count is referred in the sco connected callback due to the `sco_conn` is NULL while `sco_conn` should not be NULL when the sco connected callback is triggered.

It is a race condition issue where the sco connected callback is triggered before `sco_conn` is updated. The sco conn reference count has been referred incorrectly, actually it should not be referred.

Replace direct pointer access to `sco_conn` with atomic operations to prevent race conditions in concurrent SCO connection scenarios.

If the value of `sco_conn` is not NULL when trying to set the return value of `bt_conn_create_sco()` to `sco_conn`, it means the race condition occurs. In this situation, discount the sco conn reference count to fix the issue.

### DIFF
--- a/subsys/bluetooth/host/classic/hfp_hf.c
+++ b/subsys/bluetooth/host/classic/hfp_hf.c
@@ -3446,8 +3446,8 @@ static void hfp_hf_sco_connected(struct bt_sco_chan *chan)
 {
 	struct bt_hfp_hf *hf = CONTAINER_OF(chan, struct bt_hfp_hf, chan);
 
-	if (hf->sco_conn == NULL) {
-		hf->sco_conn = bt_conn_ref(chan->sco);
+	if (atomic_ptr_cas(&hf->sco_conn, NULL, chan->sco)) {
+		bt_conn_ref(chan->sco);
 	}
 
 	if ((bt_hf != NULL) && (bt_hf->sco_connected != NULL)) {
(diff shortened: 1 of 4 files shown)


subsys/bluetooth/host/classic/hfp_hf.c
```diff
@@ -3524,8 +3527,22 @@ static int hfp_hf_create_sco(struct bt_hfp_hf *hf)
...
-	if (hf->sco_conn == NULL) {
+	sco = bt_conn_create_sco(&hf->acl->br.dst, &hf->chan);
+	updated = atomic_ptr_cas(&hf->sco_conn, NULL, sco);
+	if (!updated) {
+		LOG_WRN("SCO is not NULL (%p), target (%p)", atomic_ptr_get(&hf->sco_conn), sco);
+		__ASSERT(atomic_ptr_get(&hf->sco_conn) == sco,
+				"Concurrent SCO connection creation detected");
+		/* The `hf->sco_conn` has been udpated in callback `hfp_hf_sco_connected()`.
+		 * The refernce count has been updated in callback `hfp_hf_sco_connected()`.
+		 * The refernce count should be unreferred in this case.
```
jhedberg, line 3538: Minor typos here: s/refernce/reference/

lylezhu2012: Yes. It has been fixed by this commit [a1f7481](https://github.com/zephyrproject-rtos/zephyr/commit/a1f74818d8fde06469624d6f1facd16f20de37a1)

makeshi (approved):

gzh-terry (approved):

PR #100073, opened 2025-11-26
drivers: mipi_dsi: Add support Renesas RA MIPI DSI for RA8P1

khoa-nguyen-18:
- Update `dsi_renesas_ra.c` driver and `renesas,ra-mipi-dsi.yaml` binding to align with `RA8P1` support
- Add dts node for MIPI support for `ek_ra8p1/r7ka8p1kflcac/cm85`
- Add `rtkmipilcdb00000be` shield support for `ek_ra8p1/r7ka8p1kflcac/cm85`
- Add support `samples/modules/lvgl/demos` for `ek_ra8p1/r7ka8p1kflcac/cm85` running with the `rtkmipilcdb00000be` shield

### DIFF
--- a/boards/renesas/ek_ra8p1/Kconfig.defconfig
+++ b/boards/renesas/ek_ra8p1/Kconfig.defconfig
@@ -10,4 +10,22 @@ config SD_CMD_TIMEOUT
 
 endif # DISK_DRIVER_SDMMC
 
+if DISPLAY
+
+config MEMC
+	default y
+
+config RENESAS_RA_GLCDC_FRAME_BUFFER_SECTION
+	default ".sdram"
+	depends on RENESAS_RA_GLCDC
+
+if LVGL
+
+config LV_Z_VDB_CUSTOM_SECTION
+	default y
+
+endif # LVGL
+
+endif # DISPLAY
+
 endif # BOARD_EK_RA8P1
--- a/boards/renesas/ek_ra8p1/ek_ra8p1.dtsi
+++ b/boards/renesas/ek_ra8p1/ek_ra8p1.dtsi
@@ -143,6 +143,12 @@
 
 &ioport1 {
 	status = "okay";
+
+	mipi_dphy_gpio: mipi-dphy-enable {
+		gpio-hog;
+		gpios = <8 GPIO_ACTIVE_LOW>;
+		output-low;
+	};
 };
 
 &ioport2 {
--- a/boards/shields/rtkmipilcdb00000be/boards/ek_ra8p1_r7ka8p1kflcac_cm85.overlay
+++ b/boards/shields/rtkmipilcdb00000be/boards/ek_ra8p1_r7ka8p1kflcac_cm85.overlay
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Renesas Electronics Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+&mipi_dphy_gpio {
+	/delete-property/ output-low;
+	output-high;
+};
+
+&renesas_mipi_i2c {
+	clock-frequency = <DT_FREQ_K(100)>;
+};
+
+&zephyr_lcdif {
+	input-pixel-format = <PANEL_PIXEL_FORMAT_RGB_565>;
+};
+
+&zephyr_mipi_dsi {
+	pll-div = <3>;
+	pll-mul-int = <125>;
+	pll-mul-frac = "0.00";
+	pll-out-div = <1>;
+	lp-divisor = <5>;
+	ulps-wakeup-period = <97>;
+	video-mode-delay = <186>;
+	timing = <1183 11 26 40>;
+
+	phys-timing {
+		t-init = <74999>;
+		t-clk-prep = <9>;
+		t-hs-prep = <6>;
+		t-lp-exit = <7>;
+		dphytim4 = <28 1 20 7>;
+		dphytim5 = <19 8 12>;
+	};
+};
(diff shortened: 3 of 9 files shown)


boards/renesas/ek_ra8p1/board.c
```diff
@@ -0,0 +1,34 @@
...
+
+	if (gpio_pin_configure_dt(&mipi_dphy_en_gpios, GPIO_OUTPUT_ACTIVE)) {
+		LOG_ERR("Failed to configure MIPI D-PHY enable gpio");
+		return -EIO;
+	}
+
+	return 0;
+}
+
+SYS_INIT(mipi_dphy_init, POST_KERNEL, CONFIG_MIPI_DSI_INIT_PRIORITY);
```
faxe1008: Better to use `board_late_init_hook` I think.

boards/renesas/ek_ra8p1/Kconfig.defconfig
```diff
@@ -10,4 +10,22 @@ config SD_CMD_TIMEOUT
...
+if DISPLAY
+
+config MEMC
+	default y
+
+config RENESAS_RA_GLCDC_FRAME_BUFFER_SECTION
+	default ".sdram"
+	depends on RENESAS_RA_GLCDC
+
+endif # DISPLAY
```
JarmouniA:
```suggestion
```

boards/renesas/ek_ra8p1/Kconfig.defconfig
```diff
@@ -10,4 +10,22 @@ config SD_CMD_TIMEOUT
...
+
+if LVGL
+
+config LV_Z_VDB_CUSTOM_SECTION
+	default y
+
+endif # LVGL
+
```
JarmouniA, line 28:
```suggestion

endif # DISPLAY

```

boards/renesas/ek_ra8p1/board.c
JarmouniA: > Better to use `board_late_init_hook` I think.

That may be problematic because of the initialization order requirement.

boards/renesas/ek_ra8p1/Kconfig.defconfig
```diff
@@ -10,4 +10,22 @@ config SD_CMD_TIMEOUT
 
 endif # DISK_DRIVER_SDMMC
 
+if DISPLAY
+
+config MEMC
+	default y
+
+config RENESAS_RA_GLCDC_FRAME_BUFFER_SECTION
```
JarmouniA, line 18: non-blocking (may be done in a follow-up PR).
Could you consider converting this to use generic linker sections, similar to
https://github.com/zephyrproject-rtos/zephyr/blob/f908d0b5f22adaaf912e729317c323864f1e1641/drivers/display/display_stm32_ltdc.c#L549

boards/renesas/ek_ra8p1/board.c
uLipe: Is it not possible to achieve the same effect using the gpio-hog pins and removing this file? I did that before on a board that needed to turn-on some CAN phy at early stage.

JarmouniA: > Is it not possible to achieve the same effect using the gpio-hog pins and removing this file?

I think so.

khoa-nguyen-18: Thank you all for the suggestions. I’ve applied the `gpio-hog`, and it works well. As a result, I removed the `board.c`

boards/renesas/ek_ra8p1/Kconfig.defconfig
khoa-nguyen-18: Thank you, I have updated this scope config

boards/renesas/ek_ra8p1/Kconfig.defconfig
khoa-nguyen-18: Thanks for your suggestion. I agree with it, but it seems to be out of scope for this PR. I will create a separate PR to update this soon

JarmouniA (approved):

uLipe (approved):

boards/renesas/ek_ra8p1/ek_ra8p1.dtsi
```diff
@@ -143,6 +143,12 @@
 
 &ioport1 {
 	status = "okay";
+
+	mipi_dphy_gpio: mipi-dphy-enable {
+		gpio-hog;
+		gpios = <8 GPIO_ACTIVE_LOW>;
+		output-low;
```
thenguyenyf, line 150: non-blocking: I think it should be enabled by default by the {board}.dts
```suggestion
		output-high;
```

boards/shields/rtkmipilcdb00000be/boards/ek_ra8p1_r7ka8p1kflcac_cm85.overlay
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Renesas Electronics Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+&mipi_dphy_gpio {
+	/delete-property/ output-low;
+	output-high;
+};
```
thenguyenyf, line 10:
```suggestion
```

thenguyenyf (approved):

boards/renesas/ek_ra8p1/ek_ra8p1.dtsi
khoa-nguyen-18: Since the default display kit that comes with the board is parallel, not MIPI, it is reasonable to disable the MIPI gpio-hog by default

KhiemNguyenT (approved):
LGTM. Thanks.

boards/renesas/ek_ra8p1/Kconfig.defconfig
khoa-nguyen-18: Hi @JarmouniA, I’ve migrated to use the generic linker sections in this PR (#101489). You can take a look when you have time. Thanks for your suggestion

PR #100083, opened 2025-11-26
samples: bluetooth: peripheral_past: Fix uninitialized parameter

aaronyegx:
The uninitialized past_param.options is uncertain and causes bt_le_per_adv_sync_transfer_subscribe() returning error.

> PAST subscribe failed (err -22)


Thalley (approved):

cvinayak (approved):

jhedberg (changes requested):
@aaronyegx the change itself is fine, but could you please make the commit message a bit cleaerer. You can use the same subject as I did for this PR, and don't call it "uncertain", rather "uninitialized" is more accurate.

JarmouniA (approved):

jhedberg (approved):

PR #96086, opened 2025-09-16
ZBus: Multi-domain support

Trond-F-Christiansen:
This PR introduces multi-domain support for ZBus, enabling message forwarding between domains.

- **Shadow Channels**:
A reserved channel type used for observing channels originating in other domains.
- **Proxy Agents**:
Background processes that forward messages between domains. Handles message forwarding, receiving, acknowledgments, and publishing to shadow channels.
   - **IPC backend**: For multi domain between cores.
   - **UART backend**: For multi domain between devices.

Fixes #94857

PR #100085, opened 2025-11-26
Zbus: channel from name

Trond-F-Christiansen:
Add a new API function zbus_chan_from_name() that allows retrieving a
zbus channel by its name string. This complements the existing
zbus_chan_from_id() function and provides more flexibility for channel
lookup operations.

Split from #96086

### DIFF
--- a/subsys/zbus/zbus.c
+++ b/subsys/zbus/zbus.c
@@ -156,6 +156,26 @@ const struct zbus_channel *zbus_chan_from_id(uint32_t channel_id)
 
 #endif /* CONFIG_ZBUS_CHANNEL_ID */
 
+#if defined(CONFIG_ZBUS_CHANNEL_NAME)
+
+const struct zbus_channel *zbus_chan_from_name(const char *name)
+{
+	CHECKIF(name == NULL) {
+		return NULL;
+	}
+
+	STRUCT_SECTION_FOREACH(zbus_channel, chan) {
+		if (strcmp(chan->name, name) == 0) {
+			/* Found matching channel */
+			return chan;
+		}
+	}
+	/* No matching channel exists */
+	return NULL;
+}
+
+#endif /* CONFIG_ZBUS_CHANNEL_NAME */
+
 static inline int _zbus_notify_observer(const struct zbus_channel *chan,
 					const struct zbus_observer *obs, k_timepoint_t end_time,
 					struct net_buf *buf)
--- a/tests/subsys/zbus/channel_name/prj.conf
+++ b/tests/subsys/zbus/channel_name/prj.conf
@@ -1,0 +1,11 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+CONFIG_ZTEST=y
+CONFIG_ASSERT=y
+CONFIG_LOG=y
+CONFIG_ZBUS=y
+CONFIG_ZBUS_CHANNEL_NAME=y
--- a/tests/subsys/zbus/channel_name/src/main.c
+++ b/tests/subsys/zbus/channel_name/src/main.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/zbus/zbus.h>
+#include <zephyr/ztest.h>
+#include <zephyr/ztest_assert.h>
+#include <zephyr/sys/util_macro.h>
+
+struct msg {
+	int x;
+};
+
+#define CHANNEL_COUNT 10
+
+#define DEFINE_TEST_CHANNELS(i, _)                                                                 \
+	ZBUS_CHAN_DEFINE(test_chan_##i, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY,              \
+			 ZBUS_MSG_INIT(0));
+
+LISTIFY(CHANNEL_COUNT, DEFINE_TEST_CHANNELS, (;))
+
+ZTEST(channel_name, test_channel_retrieval)
+{
+	/* invalid cases */
+	zexpect_is_null(zbus_chan_from_name("unknown"));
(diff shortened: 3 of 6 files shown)


subsys/zbus/zbus.c
```diff
@@ -156,6 +156,26 @@ const struct zbus_channel *zbus_chan_from_id(uint32_t channel_id)
 
 #endif /* CONFIG_ZBUS_CHANNEL_ID */
 
+#if defined(CONFIG_ZBUS_CHANNEL_NAME)
+
+const struct zbus_channel *zbus_chan_from_name(const char *name)
+{
+	if (!name) {
```
pdgendt:
```suggestion
	if (name == NULL) {
```

tests/subsys/zbus/channel_name/src/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+	/* Standard retrieval */
+	zassert_equal(&chan_a, zbus_chan_from_name("chan_a"));
+	zassert_equal(&chan_b, zbus_chan_from_name("chan_b"));
+	zassert_equal(&chan_c, zbus_chan_from_name("chan_c"));
+
+	/* Ensure no cross-talk between names */
+	zassert_not_equal(&chan_d, zbus_chan_from_name("chan_e"));
+	zassert_not_equal(&chan_e, zbus_chan_from_name("chan_f"));
+	zassert_not_equal(&chan_f, zbus_chan_from_name("chan_d"));
```
pdgendt: Nit: prefer `zexpect_` over `zassert_` if the test can continue after the statement. This way users will get to see all checks that failed, rather than only the first.

Also pointers need to be checked using `_equal_ptr`. I don't see the benefit of checking the non-equal ones though.

```suggestion
	zexpect_is_null(zbus_chan_from_name("unknown"));
	zexpect_is_null(zbus_chan_from_name(""));

	/* Standard retrieval */
	zexpect_equal_ptr(&chan_a, zbus_chan_from_name("chan_a"));
	zexpect_equal_ptr(&chan_b, zbus_chan_from_name("chan_b"));
	zexpect_equal_ptr(&chan_c, zbus_chan_from_name("chan_c"));
```

tests/subsys/zbus/channel_name/src/main.c
```diff
@@ -0,0 +1,33 @@
...
+struct msg {
+	int x;
+};
+
+ZBUS_CHAN_DEFINE(chan_a, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+ZBUS_CHAN_DEFINE(chan_b, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+ZBUS_CHAN_DEFINE(chan_c, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+ZBUS_CHAN_DEFINE(chan_d, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+ZBUS_CHAN_DEFINE(chan_e, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+ZBUS_CHAN_DEFINE(chan_f, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
```
pdgendt: These are no longer used.

pdgendt (dismissed):

tests/subsys/zbus/channel_name/src/main.c
```diff
@@ -0,0 +1,27 @@
...
+ZBUS_CHAN_DEFINE(test_chan, struct msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY, ZBUS_MSG_INIT(0));
+
+ZTEST(channel_name, test_channel_retrieval)
+{
+	zexpect_is_null(zbus_chan_from_name("unknown"));
+	zexpect_is_null(zbus_chan_from_name(NULL));
+	zexpect_is_null(zbus_chan_from_name(""));
+
+	/* Standard retrieval */
+	zexpect_equal_ptr(&test_chan, zbus_chan_from_name("test_chan"));
```
rodrigopex: Trying to find a channel by name in a single channel list is under testing. I would suggest using `FOR_EACH` or `LISTIFY` to create multiple channels. Retrieve the first channel, one or two from the middle, and the latest channel in the list. In this way, you would test the boundaries.

subsys/zbus/zbus.c
```diff
@@ -156,6 +156,26 @@ const struct zbus_channel *zbus_chan_from_id(uint32_t channel_id)
 
 #endif /* CONFIG_ZBUS_CHANNEL_ID */
 
+#if defined(CONFIG_ZBUS_CHANNEL_NAME)
+
+const struct zbus_channel *zbus_chan_from_name(const char *name)
+{
+	if (name == NULL) {
```
rodrigopex:
```suggestion
	CHECKIF (name == NULL) {
```

rodrigopex: We can use [CONFIG_ASSERT_ON_ERRORS](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_ASSERT_ON_ERRORS) to transform this `if` into an ASSERT. I am going to submit a PR doing that for all the zbus functions.

tests/subsys/zbus/channel_name/prj.conf
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2025 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+CONFIG_ZTEST=y
+CONFIG_ASSERT=n
```
rodrigopex: Why are you disabling asserts?

rodrigopex (approved):
Thanks for the PR! Small changes only.

rodrigopex (changes requested):
Sorry, accidentally hit approve.

pdgendt (approved):

rodrigopex (approved):

PR #100104, opened 2025-11-26
Add sensor: Sensirion SEN6x

M1cha:
I got my hands on a Sensirion SEN66 ([datasheet](https://sensirion.com/media/documents/FAFC548D/68C12881/Sensirion_Datasheet_SEN6x.pdf)) and decided to write a Zephyr driver for it. The driver quite big, because SEN6x are multi-sensor devices, so there are many sensor channels and many I2C commands for controlling the behavior of the sensor.

A few remarks:
* The sensor is available with an I2C interface only, which simplified the code a lot even though I'm using RTIO.
* The sensor does not have an interrupt pin. The implications of that are documented in the Kconfig.
* I'm not sure if `SENSOR_CHAN_NOX_INDEX` and `SENSOR_CHAN_VOC_INDEX` belong in the common space or if I should put them after `SENSOR_CHAN_PRIV_START`, because an index is always vendor or even device specific.
* Please pay special attention to `sen6x_one_shot_decode`. This is the first time I worked with Q-numbers, but after learning about them, I think that what most drivers do is unnecessarily complex. The best implementation I found (I didn't look very long), is from [bme280](https://github.com/laurenmurphyx64/zephyr/blob/1ba9504811d8340345ea5b8f578dff3dd5824f8c/drivers/sensor/bosch/bme280/bme280_decoder.c#L71), but even that goes through extra pow/shift's and makes the division more complicated by converting it to a Q-number and fixing up the result instead of just dividing by the original integer. If my simplified math in the SEN6x driver is correct, we could simplify a lot of code in the other drivers as well.
* Is what I do in `sen6x_save_status_buffer` safe and/or fast enough? As far as I understand RTIO, it can be called in an interrupt context.
* I did not implement fetch/get, because it's about to be deprecated.
* @jfaeh @petrosyan-van You seem to be working for Sensirion. I own SEN66 only and would appreciate if you could test this driver with SEN60, SEN63C, SEN65 and SEN68. Ideally also with firmware versions which require the legacy handling in `sen6x_activate_sht_heater`.

Sensirion_Datasheet_SEN6x.pdf V0.91 from Mouser: [GitHub](https://github.com/user-attachments/files/24205968/Sensirion_Datasheet_SEN6x.pdf), [Archive](https://web.archive.org/web/20251217064804/https://eu.mouser.com/datasheet/3/1278/1/Sensirion_Datasheet_SEN6x.pdf)


include/zephyr/drivers/sensor/sen6x.h
```diff
@@ -0,0 +1,399 @@
...
+ * @param[in] dev   The device to operate on.
+ * @param[in] value Ambient pressure in meters.
+ *
+ * @retval 0        On success
+ * @retval -EINVAL  \p value is out of range.
+ * @retval -ENOTSUP Command is not supported on this device.
+ * @retval -EPERM   Command is not allowed in the current mode.
+ * @retval -errno   For other failures.
+ */
+int sen6x_set_sensor_altitude(const struct device *dev, uint16_t value);
```
teburd, line 364: This should be an attribute

include/zephyr/drivers/sensor/sen6x.h
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * @param[in] dev   The device to operate on.
+ * @param[in] value Ambient pressure in hPa.
+ *
+ * @retval 0        On success
+ * @retval -EINVAL  \p value is out of range.
+ * @retval -ENOTSUP Command is not supported on this device.
+ * @retval -errno   For other failures.
+ */
+int sen6x_set_ambient_pressure(const struct device *dev, uint16_t value);
```
teburd, line 343: This should be an attribute

include/zephyr/drivers/sensor/sen6x.h
teburd, line 1: We should be trying very hard to not have sensor specific APIs

The status changes through a callback could be done as events provided by the stream API instead (avoiding callbacks which have issues in Zephyr once you enable userspace).

The various parameters that are being exposed here should try be done either as DT parameters for statically set values or attributes for dynamic set/get functionality.

steffengoertz:
We’d really appreciate a review from someone familiar with the RTIO (@ubieda , @teburd maybe?) subsystem.

This driver is intentionally written as a small but complete RTIO-based sensor driver, and our main goal with this PR is to validate that we’re using RTIO as intended. We’re hoping it can serve as a reference-quality example for future RTIO/sensor subsystem drivers.

Thanks for your time, and no rush, feedback is much more important to us than quick approval.

drivers/sensor/sensirion/sen6x/sen6x.c
```diff
@@ -0,0 +1,1265 @@
...
+		return ret;
+	}
+
+	do {
+		cqe = rtio_cqe_consume(ctx);
+		if (cqe != NULL) {
+			ret = cqe->result;
+			rtio_cqe_release(ctx, cqe);
+		}
+	} while (cqe != NULL);
```
ubieda: Consider using `rtio_flush_completion_queue()`

(only pointing once, although I see other instances)

drivers/sensor/sensirion/sen6x/sen6x.c
```diff
@@ -0,0 +1,1265 @@
...
+	sys_put_be16(reg, reg_be);
+
+	rtio_sqe_prep_tiny_write(write_reg_sqe, iodev, RTIO_PRIO_NORM, reg_be, sizeof(reg_be),
+				 NULL);
+	write_reg_sqe->flags |= RTIO_SQE_TRANSACTION;
+	rtio_sqe_prep_write(write_buf_sqe, iodev, RTIO_PRIO_NORM, buf, size, NULL);
+	write_buf_sqe->iodev_flags |= RTIO_IODEV_I2C_STOP;
+	write_buf_sqe->flags |= RTIO_SQE_CHAINED;
+
+	rtio_sqe_prep_delay(delay_sqe, delay, NULL);
```
ubieda, line 123: What's this delay for?

drivers/sensor/sensirion/sen6x/sen6x.c
```diff
@@ -0,0 +1,1265 @@
...
+	uint8_t *data = data_;
+	size_t position = 0;
+
+	for (size_t i = 0; i < num_checksums * 3; i += 3) {
+		data[position++] = data[i + 0];
+		data[position++] = data[i + 1];
+	}
+}
+
+static void status_work_handler(struct k_work *work)
```
ubieda, line 388: Mind explaining the need for this work handler?

ubieda (commented):
First-pass review. Haven't fully digested the whole PR, but I focused on the usage of RTIO. In the mean-time, please address all errors/warnings pointer out by SonarQube

drivers/sensor/sensirion/sen6x/sen6x.h
```diff
@@ -0,0 +1,81 @@
...
+#define MAX_NUMBER_CONCENTRATION_COUNT 5
+
+#define MAX_MEASURED_VALUES_SIZE      (MAX_MEASURED_VALUES_COUNT * 3)
+#define MAX_NUMBER_CONCENTRATION_SIZE (MAX_NUMBER_CONCENTRATION_COUNT * 3)
+
+struct sen6x_encoded_data {
+	struct sen6x_encoded_header header;
+
+	uint8_t status[6];
+	uint8_t data_ready[3];
```
dsseng, line 49: What are these constants? Could they maybe be extracted and explained (e.g. channel count)?

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
```diff
@@ -0,0 +1,461 @@
...
+	case SENSOR_CHAN_PM_0_5_NUM:
+		return MAX_MEASURED_VALUES_COUNT + 0;
+	case SENSOR_CHAN_PM_1_0_NUM:
+		return MAX_MEASURED_VALUES_COUNT + 1;
+	case SENSOR_CHAN_PM_2_5_NUM:
+		return MAX_MEASURED_VALUES_COUNT + 2;
+	case SENSOR_CHAN_PM_4_0_NUM:
+		return MAX_MEASURED_VALUES_COUNT + 3;
+	case SENSOR_CHAN_PM_10_NUM:
+		return MAX_MEASURED_VALUES_COUNT + 3;
```
dsseng, line 29: I am just assuming here, sorry if I'm wrong

```suggestion
		return MAX_MEASURED_VALUES_COUNT + 4;
```

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
```diff
@@ -0,0 +1,461 @@
...
+	if (ret >= MAX_MEASURED_VALUES_COUNT && !edata->header.has_number_concentration) {
+		return -ENODATA;
+	}
+
+	ret = sen6x_get_channel_format(chan_spec.chan_type, &is_signed, NULL);
+	if (ret) {
+		return ret;
+	}
+	channel_data = &edata->channels[ret * 3];
+
```
dsseng: seems to be a more logical grouping. also, what is 3 here?

```suggestion

	channel_data = &edata->channels[ret * 3];
```

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
```diff
@@ -0,0 +1,461 @@
...
+		return -ENODATA;
+	}
+	if (ret >= MAX_MEASURED_VALUES_COUNT && !edata->header.has_number_concentration) {
+		return -ENODATA;
+	}
+
+	channel_data = &edata->channels[ret * 3];
+
+	ret = sen6x_get_channel_format(chan_spec.chan_type, &is_signed, &divisor);
+	if (ret) {
```
dsseng: Please avoid implicit boolean conversions, [rule](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#misrac-rule-14-4)

```suggestion
	if (ret < 0) {
```

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
```diff
@@ -0,0 +1,461 @@
...
+	}
+	if (ret < MAX_MEASURED_VALUES_COUNT && !edata->header.has_measured_values) {
+		return -ENODATA;
+	}
+	if (ret >= MAX_MEASURED_VALUES_COUNT && !edata->header.has_number_concentration) {
+		return -ENODATA;
+	}
+
+	ret = sen6x_get_channel_format(chan_spec.chan_type, &is_signed, NULL);
+	if (ret) {
```
dsseng: https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#misrac-rule-14-4 (see my other comment)

drivers/sensor/sensirion/sen6x/shell.c
```diff
@@ -0,0 +1,579 @@
...
+	return value;
+}
+
+static unsigned long long sen6x_get_ull(const struct shell *sh, const char *arg, int *out_ret,
+					const char *name)
+{
+	int ret = 0;
+	unsigned long long value = shell_strtoull(arg, 0, &ret);
+
+	if (ret) {
```
dsseng: https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#misrac-rule-14-4

There're more entries I've preferred no to mark to avoid cluttering

drivers/sensor/sensirion/sen6x/sen6x.c
```diff
@@ -0,0 +1,1265 @@
...
+
+	rtio_sqe_prep_read(read_buf_sqe, iodev, RTIO_PRIO_NORM, buf, size, NULL);
+	read_buf_sqe->iodev_flags |= RTIO_IODEV_I2C_STOP | RTIO_IODEV_I2C_RESTART;
+
+	/** Send back last SQE so it can be concatenated later. */
+	if (out) {
+		*out = read_buf_sqe;
+	}
+
+	return 3;
```
dsseng, line 62: I mostly see this return being checked as an error. What is 3 supposed to mean? Given that some of the paths calling this function check for `if (ret)` it might yield unexpected results

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: Thanks. It looks like a lot of drivers need to be updated with this, because I didn't come across this function while reading other drivers code.

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: As you can see, all i2c transaction functions in this driver do this. That's because sen6x needs processing time for every transaction and can't respond immediately. See `4.5 Power-Up and Communication Start`:
> After the stop condition, the sensor validates and processes the received 
data. During this time, the sensor does not acknowledge any I2C requests (address header will be NACK’d). As 
soon as the command is fully processed, the I2C interface becomes ready again. The controller can then either 
read the result with a read operation or send the next command with a write operation.

In `4.8 I2C Commands` You can see the execution time for each command. Some of them are wrong in the table though. The correct one is always in the section that describes the command in detail.

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: It checks if the status register has changed and runs user provided callbacks if it has. This allows e.g. getting notified if the fan is broken. Since the sensor doesn't have an interrupt pin, the driver polls it with every data read, if callbacks are registered.

drivers/sensor/sensirion/sen6x/sen6x.h
M1cha: data_ready: That's just the size of the data_ready register. I thought this wouldn't need a define, since the field name itself is the description.
status: technically the same reason as data_ready, but since sen60's status register is only 3 bytes, this is more of a max buffer size.

Shall I create "max-buffer-size" defines for both?

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
M1cha: oops. good catch. thanks.

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
M1cha: 3 is the size of every channels data. (uint16 or int16 data + crc8).

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: It's the number of sqes that were submitted. I do this because other drivers do this as well, but I could change it, since I'm not using that information.

I don't see any code which checks the return value of these functions with `if (ret)`, can you point me to one?

M1cha:
Sensirion has uploaded a new datasheet and made the old URL invalid. In there they e.g. removed documentation for SEN60. I didn't have the version I developed against saved, but I found one which should be the same on mouser. I've uploaded it to both GitHub and the archive.

About the remaining sonarqube issues:
- `use of GNU statement expression extension from macro expansion`: Can't be fixed because it's triggered by using common zephyr macros.
- `cast from 'const struct device *' to 'void *' drops const qualifier`: Can't be fixed, because `rtio_sqe_prep_callback_no_cqe` takes a non-const argument. It also shouldn't be changed to const, because that would cause issues for users who want to pass non-const data.
- `Replace this "switch" statement by "if" statements to increase readability.`: I disagree, because we are checking an enum. I don't see how an if is more readable than a switch. If we wanted to add more cases in future we'd have to convert the if back into a switch again anyway.

btw, sonarqube needs to work on their UI. I can choose between tiny text or tiny usable screen space due to giant toolbars which increase in size with zoom.

drivers/sensor/sensirion/sen6x/sen6x.c
dsseng: You fixed all of those, thanks! I am not an RTIO person, let's keep it as is for them to review

drivers/sensor/sensirion/sen6x/sen6x.h
dsseng: Maybe just add a comment explaining what they hold, should be enough to understand the code then

teburd (commented):
Mostly the flows with RTIO look fine to me. I'm mostly concerned about the public sensor specific header I'm seeing, which I think much of it could be done with attributes or with a stream setup where status changes and periodic sampling could be done in place of the start/stop (stream and cancel), status change callbacks (events in the stream with/without data), and various attributes.

include/zephyr/drivers/sensor/sen6x.h
M1cha: I know it's not ideal, but there are two reasons why I did not use attributes:
- They act on channels of a device rather than just the device itself, which doesn't make sense for most of the functionality of sen6x.
- They use the wrong datatype. They can only pass a signed number (integer + fraction), but I need `char*`, unsigned numbers (It would be sad to loose the type safety here) and even complex structs (e.g. `sen6x_algorithm_tuning_parameters`). 

Additionally, some of the functions start processes that are run on the sensors MCU until completion. I don't think that any of the get/set functions I implemented should be DT parameters because it makes sense to set them dynamically at runtime(e.g. the calibration with values loaded from flash).

How would you like me to proceed? Is this maybe a limitation of the sensor API? In this case we should probably discuss extending it.

M1cha:
I consider this blocked by https://github.com/zephyrproject-rtos/zephyr/issues/103433 , because the recent addition of `SENSOR_CHAN_PM_*_COUNT` is in conflict with this PR and simply adding more new channels with weird new names feels a bit weird.

teburd:
Lets talk about this in the Sensor WG meeting next week if you are available (4/20/2025)

The calendar and mailing list are open.

M1cha:
Yes, that works for me. See you there.

include/zephyr/drivers/sensor.h
lindblandro, line 102: Is this different than `SENSOR_CHAN_HUMIDITY`? 

The sensor channel enum is missing multiple water related channels (absolute humidity, relative saturation), but wouldn't this in almost all cases mean the same thing as `SENSOR_CHAN_HUMIDITY`. If the sensor has multiple humidity channels then indexing those in a `struct sensor_chan_spec` would be less confusing IMO.

lindblandro, line 132: This seems to be a Sensirion specific unit, related to their raw measurement values. Probably belongs to sensor private channels.

lindblandro, line 147: This seems to be a Sensirion specific unit, similarly VOC index. They should probably go to the private header.

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
lindblandro: Isn't `ret` always 0 at this point?

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
lindblandro, line 360: Please use a different variable name for this.

lindblandro, line 397: You could also get the shift value from `sen6x_get_channel_format()`. Choosing a fixed shift value for each channel that covers the specified range might make sense, depending on the specifications.

drivers/sensor/sensirion/sen6x/sen6x.h
lindblandro: The number 3 pops in other places as well, mentioned in this comment: https://github.com/zephyrproject-rtos/zephyr/pull/100104/changes#r2625769026

It's present in `MAX_MEASURED_VALUES_SIZE`, `MAX_NUMBER_CONCENTRATION_SIZE`, here and a few places in the decoder implementation. Is the 3 in all cases the size of the data ready register?

drivers/sensor/sensirion/sen6x/sen6x.c
lindblandro: There's a lot of duplicates and the device seems to require a delay between each register write and the next command. Have you considered extracting the write + delay into a separate function?

drivers/sensor/sensirion/sen6x/sen6x.c
lindblandro, line 251: There's also bit specific version if you need this in boolean contexts only. For example something like `atomic_set_bit(&data->device_status, SEN6X_MEASUREMENT_ENABLED_BIT)`. What kind of async usage patterns are you expecting and does the atomic variable guard against them?

lindblandro, line 363: The constant 3 is here again, related to the data ready size?

lindblandro, line 380: 3 here as well.

lindblandro, line 402: Is this a stale development time comment?

lindblandro, line 395: This 3 is probably not related to the other 3's? Could you add macros or other named constants, it would help reading the code and might help later.

lindblandro, line 649: 3 here as well.

lindblandro (commented):
Minor code quality issues. Seems like an interesting little device. I have similar obstacles with complex sensor calibration data as is here solved with a device specific interface, so I have some interest to see how this works out.

include/zephyr/drivers/sensor.h
M1cha: I basically did the same as SENSOR_CHAN_AMBIENT_LIGHT/SENSOR_CHAN_LIGHT. Intuitively I think you're right, but the question if one should create a channel type or an index seems to have been answered arbitrarily with the other channels:
- For example, why are accelerometer axes like X/Y/Z channel types?
- There's also SENSOR_CHAN_DIE_TEMP/SENSOR_CHAN_AMBIENT_TEMP
- SENSOR_CHAN_GAUGE_* seems the weirdest to me, they have temps, voltage, ... even though they already exist as non-gauge channels.

M1cha: I think you're right. conceptionally they are the same as the existing ones with a different unit for the volume, but the semantics seem to be different:
- exact count rather than `>=`
- raw instead of calibrated.

I'll name them properly and move them.

M1cha: You're right, all indices are sensor specific by their definition.

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
M1cha: oops, good catch.

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
M1cha: `sen6x_get_channel_format` is supposed to return fundamental channel information, not precalculated values to simplify usage. If there would be multiple users which need to calculate the shift value I might agree with you, but the way it is right now I don't think that's necessary.

drivers/sensor/sensirion/sen6x/sen6x.h
M1cha: No, for data_ready it's a register size. for MAX_MEASURED_VALUES_SIZE and MAX_NUMBER_CONCENTRATION_SIZE it's the size of a channel in an array of data. each channel consists of a u16be value and a u8 checksum. Keep in mind that this is just a (well designed) coincidence: https://github.com/zephyrproject-rtos/zephyr/pull/100104/changes/BASE..94bbba95e4e8be12774fad636c4fb05e1e54131e#r3108639771

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: Creating the delay requires an allocation via `rtio_sqe_acquire`. These should be done by the code which creates the sqe chain, because:
- All of them need to be done first, before creating any of the sqes.
- The users need to know the length of the chain to send it out.

IMO, putting write+delay into a separate function would just make things more complicated and error prone by trying to hide a detail that the user must know and handle anyway.

drivers/sensor/sensirion/sen6x/sen6x.c
M1cha: IIRC, I use an atomic, to make sure the reader doesn't see corrupted data if read and write happen at the same time.

M1cha: For most transfers, the sensor adds a 1 byte checksum after every 2 bytes, no matter what the data means. Since most meaningful values are 2 bytes long, this results in one value per 3 bytes. I don't think that has to be the case though.

from the datasheet:
> The 8-bit CRC checksum transmitted after each data word is generated by a CRC algorithm. Its properties are displayed in Table 70. The CRC covers the contents of the two previously transmitted data bytes. To calculate the checksum, only these two previously transmitted data bytes are used. Note that command words are not followed by CRC.

M1cha: https://github.com/zephyrproject-rtos/zephyr/pull/100104/changes/BASE..94bbba95e4e8be12774fad636c4fb05e1e54131e#r3108639771

M1cha: No, it describes what this code is doing. it converts the format of sen60s status register to the format that all the other variants of this sensor use, to simplify the public api. This way, the user doesn't have to know that there are multiple formats.

M1cha: No, it's the size of the status register(which just happens to be 2 bytes as well) + the 1 byte checksum. That's why the other sensor variants use 6. (5 bytes status + 1 byte checksum).

Yes, I can add a macro for the channel size, but It would not be used in all the places which use `3` because some have a different meaning.

M1cha: https://github.com/zephyrproject-rtos/zephyr/pull/100104/changes/BASE..94bbba95e4e8be12774fad636c4fb05e1e54131e#r3108639771

ubieda:
> I consider this blocked by #103433 , because the recent addition of `SENSOR_CHAN_PM_*_COUNT` is in conflict with this PR and simply adding more new channels with weird new names feels a bit weird.

Sensor WG:
- Suggestion to either adhere to existing `SENSOR_CHAN_PM_*_COUNT` channels or go with sensor-specific private channels.

dts/bindings/sensor/sensirion,sen6x.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+    Example:
+
+    &i2c0 {
+      ...
+
+      sen6x: sen6x@6b {
+        ...
+
+        model = SEN6X_MODEL_SEN66;
```
ajwahab:
```suggestion
        model = "SEN66";
```

drivers/sensor/sensor_shell.c
```diff
@@ -88,6 +90,7 @@ static const char *const sensor_channel_name[SENSOR_CHAN_COMMON_COUNT] = {
 	[SENSOR_CHAN_PM_10_COUNT] = "pm_10_count",
 	[SENSOR_CHAN_DISTANCE] = "distance",
 	[SENSOR_CHAN_CO2] = "co2",
+	[SENSOR_CHAN_HCHO] = "hch0",
```
ajwahab:
```suggestion
	[SENSOR_CHAN_HCHO] = "hcho",
```

dts/bindings/sensor/sensirion,sen6x.yaml
M1cha: This is a C enum and not using the enum name as a prefix:
- would bloat the global namespace with possibly conflicting symbols
- is uncommon in zephyr.

ajwahab: Thanks for clarifying.  Build failed when using the enum, hence the suggested change.

M1cha: Sorry for my previous comment, you're completely right. This used to be an enum but I later changed it to a string. A string has none of the issues I mentioned. Thanks for the fix.

include/zephyr/drivers/sensor/sen6x.h
M1cha: > The status changes through a callback could be done as events provided by the stream API instead (avoiding callbacks which have issues in Zephyr once you enable userspace).

Are you sure that's possible and/or the right way? As far as I understand the streaming API is meant to be used for sensors which can notify the CPU via an interrupt pin that new data is available, so the sensor framework can then fetch it and notify users. sen6x doesn't have an interrupt pin, so you always have to poll.

Implementing the streaming API would mean that sen6x would have to poll internally in a thread or work handler.

M1cha:
@MaureenHelm Can you take a look at this PR?

steffengoertz:
@teburd @ubieda just bumping this one. Whenever you get a chance, could you take another look? If you don't see any further issues, feel free to give it the green light. Thank you!

MaureenHelm (commented):
Please add this sensor to `tests/drivers/build_all/sensor`

M1cha:
@MaureenHelm Thanks for the hint, I didn't know we had these build_all tests, they already found a missing include :)

The first push contains the change you requested, the second push was just a rebase to solve the merge conflicts.

include/zephyr/drivers/sensor/sen6x.h
```diff
@@ -0,0 +1,440 @@
...
+ * @retval 0        On success
+ * @retval -EINVAL  \p value is out of range.
+ * @retval -ENOTSUP Command is not supported on this device.
+ * @retval -EPERM   Command is not allowed in the current mode.
+ * @retval -errno   For other failures.
+ */
+int sen6x_set_sensor_altitude(const struct device *dev, uint16_t value);
+
+/**
+ * @brief Set Sensor Altitude.
```
ubieda: Duplicate brief. Please revisit all API doc.

ubieda:
Let's fix failing CI checks. I also think the set of device-specific should be revisited (as mentioned here https://github.com/zephyrproject-rtos/zephyr/pull/100104#discussion_r2606601918) or at least respond to those threads.

M1cha:
I'm having trouble debugging the CI failure and it's currently blocked by being unable to use gdb: https://discord.com/channels/720317445772017664/1424288135743012924/1521801219478388788

M1cha:
about the device specifics: I responded in this thread: https://github.com/zephyrproject-rtos/zephyr/pull/100104#discussion_r2742785921

ubieda:
Please rebase on top of `upstream/main`. That should make CI pass.

M1cha:
@ubieda 

It doesn't (I can reproduce this locally). There seem to be some pretty bad memory corruption issues in the sensor_shell sample for `qemu_cortex_r5/zynqmp_rpu`. Even without any of my changes, running it and spamming `sensor get sensor@0`, leads to random crashes or infinite loops. `west build  zephyr/samples/sensor/sensor_shell -b qemu_cortex_r5/zynqmp_rpu -t run -- -DDTC_OVERLAY_FILE=fake_sensor.overlay`

ajwahab:
Please remove the stale label.

M1cha:
I don't know why, but the latest rebase fixed the issue(both locally and in CI). Please take another look.

include/zephyr/drivers/sensor/sen6x.h
M1cha: @teburd Could you take a look at my responses in this thread?

M1cha:
@MaureenHelm Please take another look.

ubieda (approved):

include/zephyr/drivers/sensor/sen6x.h
```diff
@@ -0,0 +1,440 @@
...
+ * @defgroup sen6x_interface SEN6X
+ * @ingroup sensor_interface_ext
+ * @brief Sensirion SEN6X sensor
+ * @{
+ */
+
+/**
+ * @brief Custom sensor channels for SEN6X.
+ */
+enum sensor_channel_sen6x {
```
yperess: This is exactly the sensor-specific API surface @teburd flagged as a concern back in January and it's still here unchanged — private channels, custom structs, driver-specific getters/setters. That design objection from a maintainer hasn't been resolved by any of the commits since.

include/zephyr/drivers/sensor/sen6x.h
yperess: Still relevant — the driver still exposes a dedicated public header with private channels and custom get/set structs rather than attributes or stream events. This should be resolved before merge.

drivers/sensor/sensirion/sen6x/sen6x_decoder.c
yperess: This and the other MISRA 14.4 spots below still look unaddressed.

PR #100109, opened 2025-11-26
drivers/sensor: lsm6dsv16x: add device self test

avisconti:
Add device Self Test procedure. It is required to enable the per device self-test DT property as well as the LSM6DSV16X_SELF_TEST configuration.

### DIFF
--- a/drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
+++ b/drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
@@ -18,6 +18,7 @@
 #include <zephyr/sys/__assert.h>
 #include <zephyr/logging/log.h>
 
+#include <zephyr/drivers/sensor/lsm6dsvxxx.h>
 #include <zephyr/dt-bindings/sensor/lsm6dsv16x.h>
 #include "lsm6dsv16x.h"
 #include "lsm6dsv16x_decoder.h"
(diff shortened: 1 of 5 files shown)


drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
```diff
@@ -1120,6 +1356,14 @@ static int lsm6dsv16x_init_chip(const struct device *dev)
 		k_sleep(K_MSEC(30));
 	}
 
+#ifdef CONFIG_LSM6DSV16X_SELF_TEST
+	lsm6dsv16x->xl_st_result = 0;
```
ndreys, line 1362: Don't we want to zero out `gy_st_result` here too?

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
```diff
@@ -501,6 +725,12 @@ static int lsm6dsv16x_accel_get_config(const struct device *dev,
 	struct lsm6dsv16x_data *data = dev->data;
 
 	switch (attr) {
+#ifdef CONFIG_LSM6DSV16X_SELF_TEST
+	case SENSOR_ATTR_GET_SELF_TEST_RESULT:
+		val->val1 = (data->xl_st_result == 0) ? 0 : 1;
```
ndreys: So `lsm6dsv16x_gyro_self_test` and `lsm6dsv16x_accel_self_test` will return `-1` for failure and `0` for success, right? So then we convert it to `0` -> success, `1` -> failure? That seems a bit confusing, TBH. Also maybe we can define "pass" / "fail" constants in `include/zephyr/drivers/sensor/lsm6dsvxxx.h` to clarify the API?

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
```diff
@@ -282,6 +283,229 @@ static int lsm6dsv16x_accel_wake_duration_set(const struct device *dev,
...
+			break;
+		}
+	}
+
+	/* Disable Self Test */
+	lsm6dsv16x_xl_self_test_set(ctx, LSM6DSV16X_XL_ST_DISABLE);
+
+	/* Disable sensor. */
+	lsm6dsv16x_xl_data_rate_set(ctx, LSM6DSV16X_ODR_OFF);
+
```
ndreys, line 401: As a suggestion, I'd add a `LOG_INF` print here and in `lsm6dsv16x_gyro_self_test` to report the self-test status. We already print stuff like `chip ID` when a device is being probed and it's really convenient to do a quick check in e.g. RTT logs

include/zephyr/drivers/sensor/lsm6dsvxxx.h
```diff
@@ -0,0 +1,39 @@
...
+#include <zephyr/drivers/sensor.h>
+
+/**
+ * @brief Custom sensor attributes for LSM6DSVXXX
+ */
+enum sensor_attribute_lsm6dsvxxx {
+	/**
+	 * Gets the self-test mode result in sensor_value.val1 field.
+	 */
+	SENSOR_ATTR_GET_SELF_TEST_RESULT = SENSOR_ATTR_PRIV_START,
```
ndreys, line 32: ST IMUs are not the only IMUs that have self-test feature, so maybe it'd make sense to make this a new common sensor attribute instead?

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
avisconti: right, thanks

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
avisconti: Yes, I can define something like

```
enum lsm6dsvxxx_self_test_result {
        LSM6DSVXXX_ST_OK = 0,
        LSM6DSVXXX_ST_FAIL = 1,
};
```

Wha do you think?

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
avisconti: it makes sense

include/zephyr/drivers/sensor/lsm6dsvxxx.h
avisconti: I think it might be a good idea. Maybe @MaureenHelm or @teburd may comment on that.

avisconti:
> I don't know if Zephyr uses Tested-by tag, but FWIW, I backported this patch onto Zephyr 3.7 and used it on a device with LSM6DSV16x IMU. I don't have a known bad IMU to test "self-test" failure case, but pass seems to work as expected.

OK Andrey, thanks for your review.
I actually try to test the failure injecting a vibration during self-test of the IMU, and the procedure is failing as expected. Not sure if I could find a failing part somewhere here...

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
ndreys: Yeah, seems reasonable

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
```diff
@@ -282,6 +283,231 @@ static int lsm6dsv16x_accel_wake_duration_set(const struct device *dev,
...
+	LOG_INF("Accel self-test result: %d", st_result);
```
ndreys: Hmm, `st_result` isn't going to match the constants we now have in `enum lsm6dsvxxx_self_test_result`, so maybe we should change this to 

```
LOG_INF("Accel self-test result: %s", st_result ? "FAIL" : "PASS");
```
for clarity?

MaureenHelm (dismissed):

drivers/sensor/st/lsm6dsv16x/lsm6dsv16x.c
avisconti: seems reasonable. Thanks for your review

ndreys (approved):

MaureenHelm (approved):

ubieda (approved):

PR #100116, opened 2025-11-26
soc: silabs: siwx91x: link pm state to power domain and adjust driver

Martinhoff-maker:
The goal of this PR is to have correlation between pm state of the cpu and the soc power domain.
It also correct the driver behavior that are on the power domain.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/103339

### DIFF
--- a/drivers/dma/dma_silabs_siwx91x.c
+++ b/drivers/dma/dma_silabs_siwx91x.c
@@ -669,6 +669,7 @@ static void siwx91x_dma_isr(const struct device *dev)
 	}
 
 	if (data->chan_info[channel].Cnt == data->chan_info[channel].Size) {
+		sys_write32(BIT(channel), (mem_addr_t)&cfg->reg->UDMA_DONE_STATUS_REG);
 		if (data->zephyr_channel_info[channel].channel_active) {
 			pm_device_runtime_put_async(dev, K_NO_WAIT);
 			data->zephyr_channel_info[channel].channel_active = false;
@@ -678,7 +679,6 @@ static void siwx91x_dma_isr(const struct device *dev)
 			data->zephyr_channel_info[channel].dma_callback(
 				dev, data->zephyr_channel_info[channel].cb_data, channel, 0);
 		}
-		sys_write32(BIT(channel), (mem_addr_t)&cfg->reg->UDMA_DONE_STATUS_REG);
 	} else {
 		/* Call UDMA ROM IRQ handler. */
 		ROMAPI_UDMA_WRAPPER_API->uDMAx_IRQHandler(&udma_resources, udma_resources.desc,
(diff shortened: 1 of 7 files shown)


drivers/dma/dma_silabs_siwx91x.c
```diff
@@ -495,17 +484,7 @@ static int siwx91x_dma_start(const struct device *dev, uint32_t channel)
 		return -EINVAL;
 	}
 
-	/* Get the power management policy state lock */
-	if (!data->zephyr_channel_info[channel].channel_active) {
```
smalae, line 499: Should we also call pm_device_runtime_put() and pm_device_runtime_get() in DMA drivers to ensure that the user doesn’t have to explicitly manage this for memory-to-memory transfers?

drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -566,6 +567,9 @@ static int gspi_siwx91x_transceive(const struct device *dev, const struct spi_co
 	if (spi_siwx91x_is_dma_enabled_instance(dev)) {
 		/* Perform DMA transceive */
 		ret = gspi_siwx91x_transceive_dma(dev, config);
+		if (ret < 0) {
+			pm_device_runtime_put(dev);
```
jerome-pouiller, line 571: What happens if `asynchronous==false && dma=true && ret==0` ?

drivers/spi/spi_silabs_siwx91x_gspi.c
```diff
@@ -566,6 +567,9 @@ static int gspi_siwx91x_transceive(const struct device *dev, const struct spi_co
...
+		}
 		spi_context_release(&data->ctx, ret);
```
jerome-pouiller, line 573: BTW, is the context released when `asynchronous==true && dma=true && ret<0` ?

drivers/dma/dma_silabs_siwx91x.c
Martinhoff-maker: I wonder to add that like intel has done here: dma_intel_adsp_hda.c
I will try but I fear that we miss some cases that introduce some bugs.

Martinhoff-maker: that's done, seems to works

drivers/spi/spi_silabs_siwx91x_gspi.c
Martinhoff-maker: Since this PR is mainly about PM, would do it in another PR.

drivers/spi/spi_silabs_siwx91x_gspi.c
Martinhoff-maker: same as above comment.

jerome-pouiller (approved):

smalae (approved):

asmellby (approved):

PR #100118, opened 2025-11-26
RFC: doc: glossary: Consolidate "software component" as a term

carlescufi:
In order to define different parts of the Zephyr codebase, we first need a clear term that defines what a "part of the software" is, so generalize the existing term "software component" to be used from now on as the standardized term.

Given that the kernel is currently defined, as per: https://docs.zephyrproject.org/latest/kernel/services/index.html as:
"The Zephyr kernel lies at the heart of every Zephyr application. It provides a low footprint, high performance, multi-threaded execution environment with a rich set of available features. The rest of the Zephyr ecosystem, including device drivers, networking stack, and application-specific code, uses the kernel’s features to create a complete application."

label it as a software component as well.

### DIFF
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -127,8 +127,9 @@ Glossary of Terms
       executing on the processor.
 
    kernel
-      The set of Zephyr-supplied files that implement the Zephyr kernel,
-      including its core services, device drivers, network stack, and so on.
+      The :term:`software component` in Zephyr that provides the
+      hardware-agnostic core of the RTOS, including booting, threading,
+      scheduling, synchronization, user mode, memory management and more.
 
    power domain
       A power domain is a collection of devices for which power is
@@ -169,9 +170,9 @@ Glossary of Terms
       features, and that the vendor typically names and markets together.
 
    software component
-      A software component is a self-contained, modular, and replaceable part of
-      the Zephyr source code. A driver, a subsystem or an applications are all
-      examples of software components present in Zephyr.
+      A software component is a well-defined and modular part of
+      the Zephyr source code. The kernel, a driver, a library, a subsystem or a
+      sample are all examples of software components present in Zephyr.
 
    subsystem
        A subsystem refers to a logically distinct part of the operating system
--- a/doc/introduction/index.rst
+++ b/doc/introduction/index.rst
@@ -23,11 +23,11 @@ The Zephyr kernel supports multiple architectures, including:
 
 The full list of supported boards based on these architectures can be found :ref:`here <boards>`.
 
-In the context of the Zephyr OS, a :term:`subsystem` refers to a logically distinct
+In the context of the Zephyr OS, a :term:`software component` refers to a logically distinct
 part of the operating system that handles specific functionality or provides
-certain services. Subsystems can include components such as networking,
+certain services. Zephyr contains software components such as networking,
 file systems, device driver classes, power management, and communication protocols,
-among others. Each subsystem is designed to be modular and can be configured,
+among others. Each software component is designed to be modular and can be configured,
 customized, and extended to meet the requirements of different embedded
 applications.
 
@@ -182,7 +182,7 @@ Zephyr offers a large and ever growing number of features including:
 
 **Native port**
   :zephyr:board:`Native sim <native_sim>` allows running Zephyr as a Linux application with support
-  for various subsystems and networking.
+  for most components, including networking.
 
 
 .. include:: ../../README.rst


doc/introduction/index.rst
```diff
@@ -182,7 +182,7 @@ Zephyr offers a large and ever growing number of features including:
 
 **Native port**
   :zephyr:board:`Native sim <native_sim>` allows running Zephyr as a Linux application with support
-  for various subsystems and networking.
+  for various components and networking.
```
aescolar: I would either leave this line as it was, or change it to `for most components including networking`. Otherwise this gives the impression of a quite smaller scope of what is possible.

doc/glossary.rst
```diff
@@ -127,8 +127,9 @@ Glossary of Terms
       executing on the processor.
 
    kernel
-      The set of Zephyr-supplied files that implement the Zephyr kernel,
-      including its core services, device drivers, network stack, and so on.
+      The :term:`software component` in Zephyr that provides the
+      hardware-agnostic core of the RTOS, including booting, threading,
+      scheduling, synchronization, user mode, memory management and more.
```
mathieuchopstm, line 132: > hardware-agnostic core

would mean that `arch` is not part of the kernel, which sounds wrong?

mathieuchopstm (commented):
Seems fine to me otherwise

doc/glossary.rst
carlescufi: Thanks! That is exactly the kind of discussion I wanted to have here. I am very much OK with arch being part of the kernel component, so I am completely fine with removing the hardware-agnostic part and even specifically mentioning the arch part as part of the kernel.

mathieuchopstm: To be clear, I have no issues with `hardware-agnostic core` by itself; however, if we keep it (*and if `arch` is indeed considered part of the kernel!*), then `arch` should be explicitly mentioned as also being part of the "kernel" software component *in addition to the HW-agnostic code*, e.g. roughly:

```suggestion
      The :term:`software component` in Zephyr that provides the
      hardware-agnostic core of the RTOS, including booting, threading,
      scheduling, synchronization, user mode, memory management and more,
      as well as architecture-specific implementation of certain routines
      used by the hardware-agnostic code.
```

Another wording (without the `hardware-agnostic` part) is possible too, but IMO your proposal is fine except for this tiny detail.

carlescufi: Thanks @mathieuchopstm! I will wait for more feedback from others before I make any changes. Your proposed change is fine with me.

doc/introduction/index.rst
```diff
@@ -23,11 +23,11 @@ The Zephyr kernel supports multiple architectures, including:
 
 The full list of supported boards based on these architectures can be found :ref:`here <boards>`.
 
-In the context of the Zephyr OS, a :term:`subsystem` refers to a logically distinct
+In the context of the Zephyr OS, a :term:`software component` refers to a logically distinct
```
nashif, line 26: thank you for taking care of this, I think a review of terminology used and making things less confusing in our docs and codebase is long overdue, however, we should try to have a deep review and fix this once and for all and take any major actions needed (renaming stuff) in one single step. few comments.

- I guess the main driver for this is the fact that we have drivers being classified as subsystem? Well, I guess it was necessary at some time to call them something and we might have picked the wrong name and if it is a problem, we probably need to address this.
- if we are going this path, I would remove this section completely and just rely on this being defined in the gloassry, no need to keep it duplicatated here.
- Also, I would look across the documentation and the code base and address existing usage of subsystem
  - example, take https://docs.zephyrproject.org/latest/kernel/usermode/kernelobjects.html where usage of 'driver subsystems' is common. What changes do we need to make there and in the code to make this compatible with the new definition, or we do not need to touch it at all?
  - also look at other items in the gloassary and define how those relate to SC
  - visualizing all of this would be very helpful, maybe using some of our existing (probably outdated) architecture diagrams
- Based on this definition, everything is a a component, a device driver class is a component, a driver is a component, a sample using those drivers is a component, so I am not sure how useful this is, looking at our docs, there are 4 mentions, 2 of which go bak to ETSI, see below:
- Many related standards/project define software component already, need to make sure we do not conflict or overload the definition, or it will be confusing to users:
- Autosar: *AUTOSAR software components are reusable building blocks of AUTOSAR software. An AUTOSAR software component encapsulates one or more algorithms and communicates with its environment through well-defined ports. For example, a throttle application might include AUTOSAR software components that represent sensors for throttle and acceleration pedal sensors, a throttle position monitor, a controller, and an actuator.*
- [IETSI](https://cse.msu.edu/~cse435/Handouts/Standards/IEEE24765.pdf):
   *a general term used to refer to a software system or an element, such as module, unit, data, or document. IEEE Std 1061-1998 (R2004) IEEE Standard for Software Quality Metrics Methodology.2.2. 2. a functionally or logically distinct part of a software configuration item, distinguished for the purpose of convenience in designing and specifying a complex SCI as an assembly of subordinate elements"*

carlescufi: >I guess the main driver for this is the fact that we have drivers being classified as subsystem? Well, I guess it was necessary at some time to call them something and we might have picked the wrong name and if it is a problem, we probably need to address this.

No, the main driver is that we are using today the word "subsystem" in two different ways:
1. A subsystem as in a generic part of the code  (software component)
2. A subsystem as in each of the sub-folders of the `subsys/` folder, which we defined [here](https://github.com/zephyrproject-rtos/zephyr/issues/99250#issuecomment-3559049013)

>if we are going this path, I would remove this section completely and just rely on this being defined in the gloassry, no need to keep it duplicatated here.

Sounds good, I can rework that section instead of removing it entirely because maybe not all the text is fitting for the glossary

>example, take https://docs.zephyrproject.org/latest/kernel/usermode/kernelobjects.html where usage of 'driver subsystems' is common. What changes do we need to make there and in the code to make this compatible with the new definition, or we do not need to touch it at all?

Yep, I was thinking about tackling that next, as part of this PR or in the next one. In my opinion this should be changed to something  like `__driver_class` or similar, alongside with all the uses of "subsystem" in the doc page you linked. But this is exactly what this RFC is for, to gather input before I make those changes.

>Based on this definition, everything is a a component, a device driver class is a component, a driver is a component, a sample using those drivers is a component, so I am not sure how useful this is, looking at our docs, there are 4 mentions, 2 of which go bak to ETSI, see below:

Well, not quite. An internal header file is not a component. A driver class is, but we can argue that the component is the driver class headers, the common code and even the implementations themselves. But I don't see that as a bad thing myself.

>Many related standards/project define software component already, need to make sure we do not conflict or overload the definition, or it will be confusing to users:

Absolutely, feedback welcome on the glossary entry for this, this was just a first pass

PR #100120, opened 2025-11-26
doc: connectivity: networking: Add MAC address configuration page

pdgendt:
Add a page with information on how ethernet drivers can be assigned a MAC address, using the MAC address configuration mechanism.

Add `net_eth_mac_load` to the public header.

https://builds.zephyrproject.io/zephyr/pr/100120/docs/connectivity/networking/api/mac_config.html

### DIFF
--- a/doc/connectivity/networking/api/mac_config.rst
+++ b/doc/connectivity/networking/api/mac_config.rst
@@ -1,0 +1,147 @@
+.. _mac_address_config:
+
+MAC Address Configuration
+*************************
+
+Ethernet drivers can delegate most MAC address handling during initialization to
+:c:struct:`net_eth_mac_config` and :c:func:`net_eth_mac_load`. The structure
+is typically stored in the driver configuration and initialized with
+:c:macro:`NET_ETH_MAC_DT_CONFIG_INIT` or :c:macro:`NET_ETH_MAC_DT_INST_CONFIG_INIT`,
+which translate the devicetree properties into one of the following behaviors:
+
+* :c:enumerator:`NET_ETH_MAC_STATIC` – use the full ``local-mac-address`` property.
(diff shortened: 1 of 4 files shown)


doc/connectivity/networking/api/mac_config.rst
```diff
@@ -0,0 +1,123 @@
+.. _mac_address_config:
+
+MAC Address Configuration
+*************************
+
+Ethernet drivers can delegate most MAC address handling to
+:c:struct:`net_eth_mac_config` and :c:func:`net_eth_mac_load`. The structure
```
maass-hamburg: We should make it more clear that this is only for init and not for during runtime.

doc/connectivity/networking/api/mac_config.rst
```diff
@@ -0,0 +1,123 @@
...
+           macaddr_cell: cell@0 {
+               reg = <0x0 0x6>;
+               #nvmem-cell-cells = <0>;
+           };
+       };
+   };
+
+When no MAC-related properties are present, :c:func:`net_eth_mac_load` returns ``-ENODATA`` and
+the driver is expected to use its existing mechanism (for example, reading the hardware registers or
+using a build-time constant).
```
maass-hamburg, line 123: A reference to the way how to change it during runtime would be nice, if there isn't any a short explanation how to do it during runtime would be nice too.

pdgendt: Would you expect this to be documented using the `set_config` ethernet API function or the `net_mgmt` function? I don't think these are documented?

maass-hamburg: the `net_mgmt` one.

maass-hamburg (approved):

rlubos (approved):

jukkar (approved):

PR #100123, opened 2025-11-26
drivers: disk: stm32_sdmmc: add PM runtime support

Praneethsvch:
Added PM device runtime actions so driver now honors device runtime actions when PM_DEVICE and/or PM_DEVICE_RUNTIME is used. This allows dynamic suspend and resume actions. 

- `stm32_sdmmc_pm_action()` handling for PM_DEVICE_ACTION_RESUME/SUSPEND
- Runtime power management via pm_device_runtime_enable()
- Automatic clock gating and power GPIO control during suspend/resume
- Safe synchronization with DMA and IRQ state transitions
- Optional system PM policy locks around active transfers

This ensures the SDMMC peripheral enters low-power states when idle and
restores correctly when accessed again via disk or FS layers.

Tested on: STM32U575 (nucleo_u575zi_q)

### DIFF
--- a/drivers/disk/sdmmc_stm32.c
+++ b/drivers/disk/sdmmc_stm32.c
@@ -14,10 +14,12 @@
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/drivers/gpio.h>
 #include <zephyr/drivers/reset.h>
-#include <zephyr/pm/policy.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/irq.h>
 #include <zephyr/cache.h>
+#include <zephyr/pm/device.h>
+#include <zephyr/pm/device_runtime.h>
+#include <zephyr/pm/policy.h>
 #include <soc.h>
 #include <stm32_ll_rcc.h>
 
@@ -149,7 +151,7 @@ static void stm32_sdmmc_isr(const struct device *dev)
 }
 
 #define DEFINE_HAL_CALLBACK(name)                                                                  \
-	void name(HandleTypeDef *hsd)                                                           \
+	void name(HandleTypeDef *hsd)                                                              \
 	{                                                                                          \
 		struct stm32_sdmmc_priv *priv = CONTAINER_OF(hsd, struct stm32_sdmmc_priv, hsd);   \
                                                                                                    \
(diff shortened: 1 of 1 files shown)


drivers/disk/sdmmc_stm32.c
```diff
@@ -372,6 +374,9 @@ static int stm32_sdmmc_access_init(struct disk_info *disk)
 	HAL_StatusTypeDef hal_ret;
 	int err;
 
+#ifdef CONFIG_PM_DEVICE_RUNTIME
+	pm_device_runtime_get(dev);
+#endif
```
etienne-lms: `pm_device_runtime_*()` are already stubbed when `CONFIG_PM_DEVICE_RUNTIME` is disabled.
```suggestion
	pm_device_runtime_get(dev);
```
Ditto below.


That said, device the device for runtime PM would make `stm32_sdmmc_activate()` to be called, so all the clock/pinctrl/SD/MMC initialization already proceeded, no?

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
...
+		}
+	}
+#endif
+
+	k_sem_give(&priv->thread_lock);
+	pm_policy_state_lock_put(PM_STATE_SUSPEND_TO_IDLE, PM_ALL_SUBSTATES);
+	return 0;
+}
+
+#ifdef CONFIG_PM_DEVICE
```
etienne-lms, line 1110: Condition can be remove.

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
...
+int stm32_sdmmc_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	int err;
+
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+		err = stm32_sdmmc_activate(dev);
```
etienne-lms: Nitpicking: can remove `err` and use `return stm32_sdmmc_xxx(dev);` straight.

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
...
+
+	/* Gate SDMMC clocks and reset peripheral to kill any latent state */
+	(void)stm32_sdmmc_clock_disable(priv);
+	__HAL_RCC_SDMMC1_FORCE_RESET();
+	__HAL_RCC_SDMMC1_RELEASE_RESET();
+
+	/* Apply sleep/low-leakage pin state before removing power */
+	int ret = pinctrl_apply_state(priv->pcfg, PINCTRL_STATE_SLEEP);
+	if (ret && ret != -ENOTSUP) {
+		LOG_WRN("sleep pinctrl failed: %d", ret);
```
etienne-lms, line 1103: `LOG_WARN_ONCE()` should be enough.
That said `stm32_sdmmc_pwr_off()`, called right above, already does the job.

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
...
+
+	/* DMA fully idle (and detach from HAL) */
+#if STM32_SDMMC_USE_DMA
+	(void)stm32_sdmmc_dma_deinit(priv);
+#endif
+
+	/* Gate SDMMC clocks and reset peripheral to kill any latent state */
+	(void)stm32_sdmmc_clock_disable(priv);
+	__HAL_RCC_SDMMC1_FORCE_RESET();
+	__HAL_RCC_SDMMC1_RELEASE_RESET();
```
etienne-lms: Prefer `reset_line_toggle_dt(&priv->reset);` over use of HAL functions.
I wonder if reset should be done at deinit, rather at init, which is already done.

Note: reset operation needs clock to be enabled.

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
 	memcpy(csd, &priv->hsd.CSD, sizeof(priv->hsd.CSD));
 }
 
+static int stm32_sdmmc_activate(const struct device *dev)
+{
+	pm_policy_state_lock_get(PM_STATE_SUSPEND_TO_IDLE, PM_ALL_SUBSTATES);
```
etienne-lms: Is this needed?

drivers/disk/sdmmc_stm32.c
bearsh: if `CONFIG_PM_DEVICE_RUNTIME` is defined, the following code is being called twice, once in `stm32_sdmmc_activate()` and then here

drivers/disk/sdmmc_stm32.c
```diff
@@ -491,6 +497,10 @@ static int stm32_sdmmc_access_deinit(struct stm32_sdmmc_priv *priv)
 	stm32_sdmmc_pwr_off(priv);
 
 	priv->status = DISK_STATUS_UNINIT;
+
+#ifdef CONFIG_PM_DEVICE_RUNTIME
+	pm_device_runtime_put(dev);
+#endif
```
bearsh, line 477: here the same, `stm32_sdmmc_suspend()` repeats code already executed in `stm32_sdmmc_access_deinit()`

bearsh (changes requested):
`pm_device_runtime_get()` / `pm_device_runtime_put()` is not called in `stm32_sdmmc_access_read()` `stm32_sdmmc_access_write()`

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
...
+	int ret = pinctrl_apply_state(priv->pcfg, PINCTRL_STATE_SLEEP);
+	if (ret && ret != -ENOTSUP) {
+		LOG_WRN("sleep pinctrl failed: %d", ret);
+	}
+
+	/* Optionally cut card power (if you wired pwr_gpios) */
+	stm32_sdmmc_pwr_off(priv);
+
+	/* If HSI48 was used only for SDMMC, system PM may turn it off later.
+	   If you want to force it off right here (safe if USB not used): */
```
etienne-lms: As you said, this may be unsafe. I would leave it running for now until there are some refcount on clock gating requests.

drivers/disk/sdmmc_stm32.c
```diff
@@ -923,6 +933,163 @@ void stm32_sdmmc_get_card_csd(const struct device *dev, uint32_t csd[4])
 	memcpy(csd, &priv->hsd.CSD, sizeof(priv->hsd.CSD));
 }
 
+static int stm32_sdmmc_activate(const struct device *dev)
```
etienne-lms: Most of the sequence is already done in in `stm32_sdmmc_access_init()`. Could you factorize their common code?

drivers/disk/sdmmc_stm32.c
bearsh: I think so, SDMMC, at least on a U5 is only working in run and sleep but not stop. I don't thinks this is different in other families...

'
bearsh, line 1: committing this file looks like a mistake?

drivers/disk/sdmmc_stm32.c
Praneethsvch: refactored into new helper functions:  stm32_sdmmc_hw_activate/deactivate

drivers/disk/sdmmc_stm32.c
Praneethsvch: Done.  stm32_sdmmc_hw_activate/deactivate now contains the common code

bearsh:
Hi @Praneethsvch 
it seems there a still issues with this PR... 

Personally, I would like to have as much PM related stuff merge as possible (and feasible) but I'm not a subsystem maintainer...

it would be nice to have statement from the st guys (maybe @erwango) how to proceed

erwango:
Heads up, we plan to migrate SDMMC STM32 to SDHC driver (https://github.com/zephyrproject-rtos/zephyr/pull/101617). I hope it could be merged in coming weeks.
Target is to deprecate existing SDMMC driver as part of this release, so I'd advise to check this new driver and see if PM implementation could be adapted to this driver.

PR #100124, opened 2025-11-26
net: mqtt_sn: document address size and add checks for UDP

aurel32:
Add guidelines in Kconfig explaining the minimum required MQTT_SN_LIB_MAX_ADDR_SIZE for UDP transport and IPv4 and or IPv6. Also explain when this value might need to be changed.

In addition, introduce BUILD_ASSERT checks in the UDP transport implementation to ensure the configured maximum address size is large enough for the enabled IP families.


subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
```diff
@@ -19,6 +19,12 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_DECLARE(net_mqtt_sn, CONFIG_MQTT_SN_LOG_LEVEL);
 
+BUILD_ASSERT(CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE >= sizeof(struct net_sockaddr_in) ||
+	     !IS_ENABLED(CONFIG_NET_IPV4));
+
+BUILD_ASSERT(CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE >= sizeof(struct net_sockaddr_in6) ||
```
rlubos, line 25: This now asserts in our CI, perhaps the default needs to be updated after all? Or need to override this value in `tests/net/all`

aurel32: You are correct, there is a real issue to be fixed. Yesterday when adding BUILD_ASSERT, I decided to restart from a clean sample, and chose the `mqtt_sn_publisher` one. It happens that it recently got `CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE` bumped to 24, so it is not exhibiting the issue. I should have better checked the output configuration.

The size of the structure for IPv6 is indeed 21 bytes if you compute its size, but alignment makes it 24 bytes. I have fixed the value in the `Kconfig` comment, provided correct default values for both IPv4 and IPv6

rlubos (approved):

jukkar (approved):

subsys/net/lib/mqtt_sn/mqtt_sn_transport_udp.c
M1cha, line 25: Is this assert a good idea? You might have ipv6 enabled for other uses, but might only need to use ipv4 for mqtt-sn.

PR #100126, opened 2025-11-26
doc: services: implement grouping of services docs

kartben:
https://builds.zephyrproject.io/zephyr/pr/100126/docs/services/index.html

Currently we have 39 (and not every "service" even has proper documentation at the moment...) top-level entries listed at the top level of [OS Services](https://docs.zephyrproject.org/latest/services/index.html) page. This list is virtually useless for anyone who does not already know whether the "service" or class of service they need even exists.

This PR overhauls the (OS) Services landing page by grouping these services into logical categories/buckets using a grid layout. This improves navigation and provides a better high-level overview of what Zephyr has to offer.

image

I have settled on 12 categories at this point, which one could say is still a lot, but it's important that the grouping doesn't become too vague/meaningless. 

Things that this PR does not necessarily address / A note on terminology:

- "Services" is a very broad term, and things that are (currently, as well as after this PR) listed under Services are a mixed bag of "libraries"/"utilities"/"subsystems"/"...". We certainly need to be better at terminology (see e.g. https://github.com/zephyrproject-rtos/zephyr/pull/100118) but in the meantime I don't think it is fundamentally wrong to call these things "services".


doc/services/index.rst
```diff
@@ -3,42 +3,84 @@
...
+
+        Networking, Bluetooth, USB, and other connectivity services.
+
+    .. grid-item-card:: :material-twotone:`bug_report;36px` Debugging & Observability
+        :link: debugging_observability_services
+        :link-type: ref
+
+        Logging, tracing, shell, and other debugging and observability tools.
+
+    .. grid-item-card:: :material-twotone:`import_export;36px` Input / Output
```
kartben: maybe "User Interface"?

ljd42:
Super Idea @kartben 

I have created something similar (albeit not that detailed) for presentation purposes. 

Find below some possible tinkering, in the case we want to reduce the number of categories to 8 or 9. Just throwing some ideas, it's not a change request :wink: 

**Device Management**
Everything we need to manage the device (boot/reboot, authentication, update)
- Cryptography
- Device Management
- Linkable Loadable Extensions (LLEXT)
- Task Watchdog
- "Security & Identity" could be included here too

**IPC & Messaging**
add Asynchronous notification?

**I/O and Data Processing**
Everything related to I/O processing
In addition to list proposed, move these services here:
- Real Time I/O (RTIO)
- Sensing Subsystem
- DSP

**System and Utilities**
- Binary Descriptors
- Resource Management
- Memory Attributes
- Network Buffers
- State Machine Framework (could also be listed in Device Management) 

**Remove**
- Algorithms 
- Application Services

fabiobaltieri (dismissed):
looking good

doc/services/index.rst
```diff
@@ -1,44 +1,93 @@
...
+      :link: io_services
+      :link-type: ref
+
+      Console, input devices abstraction, RTIO, and other I/O utilities.
+
+   .. grid-item-card:: :material-twotone:`save;36px` Storage & Configuration
+      :link: storage_services
+      :link-type: ref
+
+      Flash access, file systems, key-value stores, and persistent data.
```
nordicjm: the problem is retention fits none of this, it's not flash, it's not a file system, it's not key-value store and it's not persistent, from the retention doc page:

> The stored data should not persist in the event of a power failure (or during some low-power modes on some devices)

kartben: went for "Flash access, file systems, key-value stores, and data retention."

doc/services/messaging.rst
```diff
@@ -0,0 +1,15 @@
...
+This section provides an overview of some of the inter-process and inter-thread communication
+mechanisms available in Zephyr.
+
+.. toctree::
+   :maxdepth: 1
+
+   zbus/index.rst
+   ipc/index.rst
+   notify.rst
+   virtualization/index.rst
```
mathieuchopstm, line 15: This renders into `Virtualization` which is *very much* unexpected for "IPC", although it starts making sense when you do click on it and see `Inter-VM Shared Memory`; consider linking to the latter directly instead.

doc/services/messaging.rst
```diff
@@ -0,0 +1,15 @@
...
+***************
+
+This section provides an overview of some of the inter-process and inter-thread communication
+mechanisms available in Zephyr.
+
+.. toctree::
+   :maxdepth: 1
+
+   zbus/index.rst
+   ipc/index.rst
```
mathieuchopstm, line 12: This goes to an almost empty page containing only a link to the `IPC service`: consider making this page what is currently `IPC service` (this is arguably outside PR scope)

doc/services/index.rst
```diff
@@ -1,44 +1,93 @@
...
+enabling application portability across different platforms.
+
+.. grid:: 1 1 2 2
+   :gutter: 3
+
+   .. grid-item-card:: :material-twotone:`memory;36px` System
+      :link: system_services
+      :link-type: ref
+
+      Memory management, low-level infrastructure, and other utility APIs.
```
mathieuchopstm: * `memory management` feels like it should be in `Kernel`, *especially* since it only applies to the `Memory Attributes` page here
* `low-level infrastructure` feels like `Kernel` territory: in my mind, low-level=things that have """no dependency""" (vs. high-level which builds *on top* of low-level primitives)
* `utility APIs` sounds to me like `util_macro.h`, i.e. little stateless helper functions, rather than full-blown libraries as (I think?) you intended

It doesn't help that half of the the (four!) pages currently grouped here don't fit in any of the categories:
* `Network Buffers` is high-level (built on top of `net_buf_simple`, which is however arguably a `utility API`)
* `Task Watchdog` is high-level (built on top of many Kernel primitives (+WDT API))

I would sort of expect them to be in `Application Services`, but `netbuf` is for drivers too and "**Application** Services" carries a "if you're a driver, don't come here" meaning, so that wouldn't be a good fit either (but it feels like the correct group for `Task Watchdog`)

doc/services/index.rst
```diff
@@ -1,44 +1,93 @@
...
+
+   .. grid-item-card:: :material-twotone:`device_hub;36px` Connectivity
+      :link: connectivity_services
+      :link-type: ref
+
+      Networking, Bluetooth, USB, and other connectivity services.
```
mathieuchopstm: There are three categories for "I/O":
* `Connectivity`
* `Input / Output`
* Arguably `IPC & Messaging`

doc/services/index.rst
```diff
@@ -1,44 +1,93 @@
...
+      :link: observability_services
+      :link-type: ref
+
+      Logging, tracing, shell, and other debugging and observability tools.
+
+   .. grid-item-card:: :material-twotone:`import_export;36px` Input / Output
+      :link: io_services
+      :link-type: ref
+
+      Console, input devices abstraction, RTIO, and other I/O utilities.
```
mathieuchopstm: These two are also conflicting: `Logging`/`tracing`/`shell` very much feel like `Input / Output`

kartben:
Thanks for the feedback @mathieuchopstm and @henrikbrixandersen.

> > I must admit, my eyes _really_ prefer just a bulleted list like what we currently have here: [docs.zephyrproject.org/latest/services/index.html](https://docs.zephyrproject.org/latest/services/index.html)
> > While the squares, symbols, bold text, greyed text etc. may appear more "cool", it certainly doesn't make it easier to navigate (same goes for the recent redesign of the front, [docs.zephyrproject.org/latest](https://docs.zephyrproject.org/latest/)).
> 
> +1 on this: as proposed, if I know what I'm looking for (say, `LLEXT`) but not in which category it falls, I have to half-guess, click on a big square, _load a new page_ and hope my guess is correct.

Replying to the "[...] if I know what I'm looking for [...]" bit specifically: I would _really_ hope that people use built-in search (pro-tip: hit `/` key and start searching straight away, no mouse involved :)) as I think this will always be the fastest way (by far!) to navigate the docs for someone who know what they're looking for (kind of like how I hope one doesn't try to manually navigate the doxygen docs hierarchy to get to the docs of that function that they already know how it's called!).

I will definitely look into fixing the missing "+"s, as I had noticed that too. Having the buttons _and_ the bullet list is definitely a way to fix this but this comes with other Sphinx issues.

mathieuchopstm (commented):
Would it be possible to sort all entries within a group in alphabetical order?

e.g.
```
Connectivity
Networking
Bluetooth
USB
Controller Area Network (CAN) Bus Protocols
LoRa and LoRaWAN
Modbus
Modem modules
```

-->

```
Bluetooth
Connectivity
Controller Area Network (CAN) Bus Protocols
LoRa and LoRaWAN
Modbus
Modem modules
Networking
USB
```

Otherwise, this looks pretty nice.

kartben:
> Would it be possible to sort all entries within a group in alphabetical order?
> 
> [...]

@mathieuchopstm Yeah it's kind of an intentional "editorial" choice on this one, which I think might be worth keeping. Similarly I don't think we would necessarily want the "Logging, Tracing & Debugging" section to start with "CPU Load" ... but I appreciate this is a double-edged sword and may be hard to maintain and arbitrate going forward.

mathieuchopstm:
> > Would it be possible to sort all entries within a group in alphabetical order?
> > [...]
> 
> Yeah it's kind of an intentional "editorial" choice on this one, but I am happy with making this alphabetically sorted

For some sections, it *might* make sense to have a different order (e.g., `Device Management` being a the top of `Device Management`), but I can't really see one for `Connectivity` for example.

> (mind you it's impossible to enforce in CI with `zephyr-keep-sorted` tags though, because the file name of the rst document, used for the sorting, may be different from the actual title of the document)

True, but I think it would still be beneficial overall (e.g., `State Machine Framework` is `smf` - can go high a few ranks due to `'m' < 't'`, but still remains in the overall `S` group)

henrikbrixandersen:
Love the categories, looks good. I would prefer the UI not to "move" on mouse-over, though.

kartben:
> Love the categories, looks good.

Thanks!

> I would prefer the UI not to "move" on mouse-over, though.

Right, I think I _could_ be convinced to drop this UI choice but on the other hand that's the exact same styling we've had on the home page for a while now, so it's consistent :)

henrikbrixandersen:
> Right, I think I _could_ be convinced to drop this UI choice but on the other hand that's the exact same styling we've had on the home page for a while now, so it's consistent :)

Ugh, right. You can tell how seldom I actually use the links on the page as opposed to the sidebar links.

doc/services/frameworks.rst
```diff
@@ -0,0 +1,14 @@
...
+
+This section covers high-level frameworks and libraries that help structure
+application logic, load code at runtime, and model sensors.
+
+.. toctree::
+   :maxdepth: 1
+
+   llext/index.rst
+   sensing/index.rst
+   smf/index.rst
```
henrikbrixandersen, line 14: SMF is not only for application-level code, perhaps it ought to go somewhere else - not sure where, though.

kartben: Yeah LLEXT and SMF I was really tempted to put them under "Cool stuff, check it out" category. I'll think about it, and maybe other will chime in with ideas

henrikbrixandersen: They are more like "frameworks"?

henrikbrixandersen (approved):

ljd42:
Look great @kartben, Thanks!!

ljd42 (approved):

PR #100139, opened 2025-11-27
RFC: drivers: biometrics: Add driver subsystem, GT5x, and ZFM-X0 fingerprint sensor drivers

heronet:
This PR proposes a new `biometrics` driver subsystem to standardise interactions with biometric sensors (Fingerprint, Iris, Face, etc.).

Currently, Zephyr lacks a dedicated API for these devices. Users are forced to use raw UART/I2C, which is not portable. This RFC introduces a generic API and a reference implementation.

This subsystem provides a unified API for:
- Enrollment (Start, Capture, Finalize)
- Matching (Verify 1:1, Identify 1:N)
- Template Management (Store, Delete, List)
- Feedback (LED control)

On the naming of the (Biometrics) API:

> A biometric sensor is a device that measures and analyzes biological characteristics or behavioral patterns of individuals for the purpose of identifying or verifying their identity. These sensors capture unique physical or behavioral traits that are distinct to each person, allowing for accurate authentication.

(source: https://ieeexplore.ieee.org/document/10808205) (thanks to @rodrigopex)

More references:
1. [ISO/IEC 2382-37](https://www.iso.org/standard/73514.html): International standard for "Biometric recognition"
2. Android SDK uses [android.hardware.biometrics](https://developer.android.com/reference/android/hardware/biometrics/package-summary) for its authentication system


drivers/biometrics/biometrics_as608.c
```diff
@@ -0,0 +1,918 @@
...
+		ret = 0;
+	} else {
+		ret = as608_err_to_errno(ret);
+	}
+
+	k_mutex_unlock(&data->lock);
+	return ret;
+}
+
+static const struct biometric_driver_api as608_api = {
```
pdgendt: Use the `DEVICE_API` macro to place it in a dedicated linker region (generated automatically). This is useful for filtering devices, for example in shell commands.

```suggestion
static DEVICE_API(biometric, as608_api) = {
```

pdgendt (commented):
Cool stuff! This will need to go through the architecture working group as it introduces a new driver API class.

Some nice additions would be:
- Documentation!
- Maintainer area
- A simulated/emulated test
- Shell commands for testing

drivers/biometrics/Kconfig.as608
```diff
@@ -0,0 +1,29 @@
...
+	default y
+	depends on DT_HAS_SYNOCHIP_AS608_ENABLED
+	select SERIAL
+	select UART_INTERRUPT_DRIVEN
+	help
+	  Enable driver for AS608 optical fingerprint sensor.
+
+if BIOMETRICS_AS608
+
+config AS608_PASSWORD
```
kartben: This should probably be in Devicetree since/if the password is unique to each sensor?

samples/drivers/biometrics/fingerprint/README.rst
```diff
@@ -0,0 +1,90 @@
+.. _fingerprint_sensor_sample:
+
+Fingerprint Sensor Sample
+##########################
+
+Overview
+********
```
kartben: this needs to use `.. zephyr:code-sample::` (template here https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/templates/sample.tmpl or see other code samples for reference)

drivers/biometrics/Kconfig.as608
heronet: All of these devices come with the default password 0x00000000. Most users don't even know about it. But it is changable.

I had to choose between Kconfig or Devicetree here. Putting it in Kconfig will allow the api be similar to other APIs that usually take strings from Kconfig. But having multiple devices with different passwords will be much more straightforward using Devicetree. 

I'll move this to dts

samples/drivers/biometrics/fingerprint/README.rst
heronet: fixed

include/zephyr/drivers/biometrics.h
```diff
@@ -0,0 +1,726 @@
...
+					 struct biometric_capabilities *caps);
+
+static inline int z_impl_biometric_get_capabilities(const struct device *dev,
+						    struct biometric_capabilities *caps)
+{
+	const struct biometric_driver_api *api;
+
+	if (dev == NULL || caps == NULL) {
+		return -EINVAL;
+	}
```
fabiobaltieri: these should probably be `__ASSERT_NO_MSG`, bit of a waste to have runtime validation in all the time for every single API

include/zephyr/drivers/biometrics.h
```diff
@@ -0,0 +1,726 @@
...
+
+	api = (const struct biometric_driver_api *)dev->api;
+
+	if (api->get_capabilities == NULL) {
+		return -ENOSYS;
+	}
```
fabiobaltieri: does it really make sense to check this for every single api call? I'd imagine some of them are mandatory and you could assume that the handler is there

include/zephyr/drivers/biometrics.h
heronet: I've replaced all runtime NULL checks with `__ASSERT_NO_MSG`. Thanks for the suggestion

include/zephyr/drivers/biometrics.h
heronet: Agreed. I've categorized the APIs and removed the NULL checks for mandatory operations

samples/drivers/biometrics/fingerprint/app.overlay
```diff
@@ -0,0 +1,24 @@
...
+&uart1 {
+	status = "okay";
+	current-speed = <57600>;
+
+	as608: as608 {
+		compatible = "synochip,as608";
+		status = "okay";
+		comm-addr = <0xffffffff>;
+		label = "AS608_FINGERPRINT";
+	};
```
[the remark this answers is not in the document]
heronet: No

drivers/biometrics/biometrics_zfm_x0.c
```diff
@@ -0,0 +1,928 @@
...
+	return data->rx_pkt.buf[AS608_HDR_SIZE];
+}
+
+/* Timeout conversion with overflow check */
+static int as608_poll_finger(const struct device *dev, uint32_t timeout_ms)
+{
+	const int64_t start = k_uptime_get();
+	int ret;
+
+	while ((k_uptime_get() - start) < timeout_ms) {
```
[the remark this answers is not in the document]
heronet: timeout_ms is never added to start

drivers/biometrics/biometrics_zfm_x0.c
```diff
@@ -0,0 +1,928 @@
...
+	int ret = 0;
+
+	k_mutex_lock(&data->lock, K_FOREVER);
+
+	if (data->enroll_state != AS608_ENROLL_IDLE) {
+		ret = -EBUSY;
+		goto unlock;
+	}
+
+	if (template_id == 0 || template_id > data->max_templates) {
```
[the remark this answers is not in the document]
heronet: The API requires 1 based indexing. Added documentation for this

include/zephyr/drivers/biometrics.h
```diff
@@ -0,0 +1,680 @@
...
+
+/**
+ * @brief Biometric LED states
+ */
+enum biometric_led_state {
+	BIOMETRIC_LED_OFF,    /**< LED off */
+	BIOMETRIC_LED_ON,     /**< LED on continuously */
+	BIOMETRIC_LED_BLINK,  /**< LED blinking */
+	BIOMETRIC_LED_BREATHE /**< LED breathing effect */
+};
```
bjarki-andreasen, line 67: This seems very specific to one fingerprint sensor, do all fingerprint sensors have 1, and only 1 led, and can it commonly breath?

include/zephyr/drivers/biometrics.h
```diff
@@ -0,0 +1,680 @@
...
+	/** Minimum confidence score for match acceptance (sensor-specific range) */
+	BIOMETRIC_ATTR_MATCH_THRESHOLD,
+	/** Minimum quality score for enrollment acceptance (sensor-specific range) */
+	BIOMETRIC_ATTR_ENROLLMENT_QUALITY,
+	/** Security level (0-10000), higher values reduce false accepts */
+	BIOMETRIC_ATTR_SECURITY_LEVEL,
+	/** Default operation timeout in milliseconds */
+	BIOMETRIC_ATTR_TIMEOUT_MS,
+	/** Anti-spoofing detection sensitivity (0-10000) */
+	BIOMETRIC_ATTR_ANTI_SPOOF_LEVEL,
```
bjarki-andreasen: why 0 to 10000? probably 1 to 10 or something with less resolution would be sufficient?

include/zephyr/drivers/biometrics.h
```diff
@@ -0,0 +1,680 @@
...
+
+/**
+ * @brief Delete all biometric templates
+ *
+ * @param dev Pointer to the biometric device
+ *
+ * @retval 0 Success
+ * @retval -errno Negative errno code on failure
+ */
+__syscall int biometric_template_delete_all(const struct device *dev);
```
bjarki-andreasen, line 586: biometric is not the appropriate name for this class of device, the device does not produce a biometric output for the application to read. This class of device uses some biometric sensor (though not necessarily I would argue), a low res camera likely for a finger print sensor, reads it, does a bunch of processing to compare it to a database it has stored, and returns an id which matches. The device from the perspective of the application is a "unique identifier reader".

I would class this device as a uid_reader or something like that, it could read anything, iris, fingerprint, rfid card, gate (walking pattern), and it returns an id (hopefully unique) to which it matches.

bjarki-andreasen (dismissed):

rodrigopex:
@heronet 

Could you try adding another sensor driver implementation to this API to ensure the APIs are sufficiently generic? It would be ideal to have at least one of each device available in the enum. 

The name seems to be okay: 

> A biometric sensor is a device that measures and analyzes biological characteristics or behavioral patterns of individuals for the purpose of identifying or verifying their identity. These sensors capture unique physical or behavioral traits that are distinct to each person, allowing for accurate authentication.

(source: https://ieeexplore.ieee.org/document/10808205)

include/zephyr/drivers/biometrics.h
heronet: I just wanted to do higher resolution. But I think treating this as a LEVEL,  1-10 sounds better. I'll change it

include/zephyr/drivers/biometrics.h
bjarki-andreasen: Ignore this, biometric seems to be used for bio identification commonly, though I find that as poor choice of term as AI for LLMs, but whatever I don't define the words

heronet: The naming seems to be correct as @rodrigopex pointed out in a comment. Industry precedent supports 'biometrics'. So I think we can keep it. 

1. [ISO/IEC 2382-37](https://www.iso.org/standard/73514.html): International standard for "Biometrics"
2. Android SDK uses [android.hardware.biometrics](https://developer.android.com/reference/android/hardware/biometrics/package-summary) for its authentication system

And `uid_reader` sounds too generic and may conflich with nfc/rfid which read external tokens rather than biological characteristics. The enrollment workflow (`enroll_start` → `enroll_capture` → `enroll_finalize`) is specific to biometric systems and doesn't apply to RFID readers

include/zephyr/drivers/biometrics.h
heronet: The LED control is an optional feature. Drivers that don't support it simply return `-ENOSYS` as the API callback is nullable. Many fingerprint sensors (AS608, R502A, GT521F52) do include LEDs for user feedback, but sensors without LEDs can omit this API function entirely.

heronet:
> > @nordicjm After doing some deep diving I found out that the (almost) exact same fingerpring firmware and AS608 chip is used in JM101, Grow R307, Adafruit Optical Fingerprint(DY-50, FPM10A), ZhianTec ZFM 20/60/70, DFrobot Optical Fingerprint. But the original seems to be ZFM as both the Adafruit and DFRobot datasheets links to this one.
> > I think `synochip,as608` is not the most accurate compatible afterall. `zhiantec,zfm-x0` sounds the most accurate to me. So I changed it to this one.
> > Adafruit: https://cdn-shop.adafruit.com/datasheets/ZFM%20user%20manualV15.pdf DFRobot: https://dfimg.dfrobot.com/nobody/wiki/f4fd9de3830b858d4c661a6f12acf305.pdf ZFM-70: https://www.electrokit.com/upload/quick/01/6c/77db_41022692_TDS.pdf Grow R307: https://www.openhacks.com/uploadsproductos/r307_fingerprint_module_user_manual.pdf
> 
> `zhiantec,zfm-70` would be fine for dts compatible (assuming it is the one you in your device), as that string should refer to one specific device, e.g. lis2dh does the same then checks if the device has other compatibles to see if it should enabled quriks so https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/seeed/wio_wm1110_dev_kit/wio_wm1110_dev_kit.dts#L97 has both compatibles - the lis2dh is required for the driver, the lis3dh enables various extra parts in the driver https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/sensor/st/lis2dh/lis2dh.c#L591

Well I have DY-50/FPM10A, JM-101, and Grow R307 and all of these are compatible with this driver. The ZFMs are the same according to the datasheets. I beleive the relevant features are all the same so using seperate compatible wouldn't be necessary.

Should I go with `grow,r307` or `zhiantec,zfm-70`?

heronet:
Hi @nashif @henrikbrixandersen @carlescufi @pdgendt @rodrigopex @fabiobaltieri @bjarki-andreasen @JarmouniA,

I've addressed all the review feedback. It's been a couple weeks without movement. 

Is there anything else I need to do to move this forward? Happy to make any changes or additions needed.

Thanks!

pdgendt:
> I've addressed all the review feedback. It's been a couple weeks without movement.

I think its due to being so close to the holidays, it will probably get more reviews from next week onwards.

heronet:
Hi @pdgendt @rodrigopex @fabiobaltieri @nashif @henrikbrixandersen,

Hope the new year is going well! It's been a while since the holidays..
Would appreciate any further feedback when you have the bandwidth.

Thanks!

tests/drivers/biometrics/biometrics_emul/src/main.c
```diff
@@ -0,0 +1,340 @@
...
+#include <zephyr/device.h>
+#include <zephyr/drivers/biometrics.h>
+
+/* Defined in biometrics_emul.c */
+void biometrics_emul_set_match_score(const struct device *dev, int score);
+void biometrics_emul_set_match_id(const struct device *dev, int id);
+void biometrics_emul_set_match_fail(const struct device *dev, bool fail);
+void biometrics_emul_set_capture_timeout(const struct device *dev, bool timeout);
+void biometrics_emul_set_image_quality(const struct device *dev, int quality);
+enum biometric_led_state biometrics_emul_get_led_state(const struct device *dev);
```
bjarki-andreasen: expose these in device specific header `include/zephyr/drivers/biometrics/emul.h`

bjarki-andreasen (dismissed):
Still think the LED pattern part should not be part of this API, but overall it looks good

tests/drivers/biometrics/biometrics_emul/src/main.c
heronet: Thanks for you review. I'll move these headers.

But the LED situation is a little complicated. Many of the times these LEDs are not gpio exposed and need command manipulation (i.e through UART) to toggle. Most Arduino drivers include these controls in driver level. If not through this subsystem, can you suggest any other way to solve this problem that I should try?

My current implementation just makes it optional, throwing -ENOSYS if not needed.

heronet: @bjarki-andreasen I moved the headers to `include/zephyr/drivers/biometrics/emul.h`.

bjarki-andreasen: We have an LED subsystem for leds, so devices which include multiple devices (called MFDs) expose the devices individually like:
```
/ {
        aliases {
                fingerprint = &fingerprint;
                led = &led;
        };
};

&uart {
        /* parent device (MFD) */
        mfd {
                /* fingerprint reader child device exposed with biometrics API */
                fingerprint: fingerprint {
                };

                /* LED child exposed with using LED API */
                led: led {
                };
        };
};

```

bjarki-andreasen (dismissed):

tests/drivers/biometrics/biometrics_emul/src/main.c
heronet: Thanks for your approval!

I just decided to include this samll functionality instead of including another subsystem(`CONFIG_LED=y`), since it is very common for these sensors to have some form of LED indicators. There could also exist some sensors with automatic, firmware LED control, with no user configurablity.

My goal was to make it easy and intuitive to control the LED as a feature of the sensor itself, not a separate device and as seamless as it is in other ecosystems like Arduino. I think having 2 devices to control breaks the illusion of uniformity.

bjarki-andreasen: I get that, its just not flexible. For the RTC subsystem, many external RTCs have both a clock calendar, and some form of retained memory, and for most external modules, be that modems, charger ics or whatever, they typically expose a few gpios, besides their primary use. The flexible way to handle this is to define the individual devices within the external modules, just like we define individual devices within the soc.

heronet: I totally get the flexibility point and I agree with you on that. I am a firm believer of the unix philosophy of doing one thing and doing it well. But if I take a step back and look at it from a practical standpoint, though semantically correct, the response LED of a fingerprint sensor individually is not as useful as say the temperature sensor of a DS3231 RTC module which actually is a proper MFD. It's more similar to the backlight of displays where it's not useful individually. The display drivers toggle the BL without using the LED subsystem too.

But it's the initial version and I will be maintaining it. So we can always move to the MFD solution if we need to in the future. Thanks for thinking this through and taking your time on reviewing this!

dts/bindings/biometrics/zhiantec,zfm-x0.yaml
```diff
@@ -0,0 +1,29 @@
...
+      Device communication address used in ZFM protocol packets.
+      Default is 0xFFFFFFFF. Can be changed via system parameter commands.
+
+  password:
+    type: int
+    default: 0x00000000
+    description: |
+      Sensor password for authentication. All sensors ship with default
+      password 0x00000000. Only change if you've modified the sensor's
+      password using the password change command.
```
mbolivar: Missing explanation for **why** the default is chosen; please fix this after reading https://docs.zephyrproject.org/latest/build/dts/bindings-upstream.html#rules-for-default-values

mbolivar (changes requested):
Nice work! Please add a rationale for your default values or mark them as `required`, then I'm happy to approve this PR.

dts/bindings/biometrics/zhiantec,zfm-x0.yaml
heronet: Thanks for the review. I have updated the descriptions  with better explanation. These values are factory programmed so I chose to keep them as defaults. But let me know if `required` is the better approach.

mbolivar (approved):

pdgendt (approved):

PR #100145, opened 2025-11-27
drivers/pwm/it8xxx2: add pwm init output level property

RuibinChang:
When EC reboot, pwm pins go back to default GPI mode, and after we set pin mode to pwm mode, it would output low, this may cause PWM-LED flicker, so add the property to set PWM channel init output level.

### DIFF
--- a/drivers/pwm/pwm_ite_it8xxx2.c
+++ b/drivers/pwm/pwm_ite_it8xxx2.c
@@ -35,6 +35,8 @@ struct pwm_it8xxx2_cfg {
 	uintptr_t reg_pwmpol;
 	/* PWM channel */
 	int channel;
+	/* Init PWM channel output high */
+	bool pwm_init_high;
 	/* PWM prescaler control register base */
 	struct pwm_it8xxx2_regs *base;
 	/* Select PWM prescaler that output to PWM channel */
@@ -232,7 +234,9 @@ static int pwm_it8xxx2_init(const struct device *dev)
 {
 	const struct pwm_it8xxx2_cfg *config = dev->config;
 	struct pwm_it8xxx2_regs *const inst = config->base;
+	volatile uint8_t *reg_dcr = (uint8_t *)config->reg_dcr;
 	volatile uint8_t *reg_pcssg = (uint8_t *)config->reg_pcssg;
+	volatile uint8_t *reg_pwmpol = (uint8_t *)config->reg_pwmpol;
 	int ch = config->channel;
 	int prs_sel = config->prs_sel;
 	int pcssg_shift;
@@ -262,6 +266,17 @@ static int pwm_it8xxx2_init(const struct device *dev)
 	 */
 	inst->CTR1M = 0;
 
+	/* Set PWM init output level */
+	*reg_dcr = 0;
+	if (config->pwm_init_high) {
+		*reg_pwmpol |= BIT(ch);
+	} else {
+		*reg_pwmpol &= ~BIT(ch);
+	}
+
+	/* PWM channel clock source not gating */
+	pwm_enable(dev, 1);
+
 	/* Enable PWMs clock counter */
 	inst->ZTIER |= IT8XXX2_PWM_PCCE;
 
@@ -290,6 +305,7 @@ static DEVICE_API(pwm, pwm_it8xxx2_api) = {
 		.reg_pcsgr = DT_INST_REG_ADDR_BY_IDX(inst, 2),				\
 		.reg_pwmpol = DT_INST_REG_ADDR_BY_IDX(inst, 3),				\
 		.channel = DT_PROP(DT_INST(inst, ite_it8xxx2_pwm), channel),		\
+		.pwm_init_high = DT_INST_PROP(inst, pwm_init_high),			\
 		.base = (struct pwm_it8xxx2_regs *) DT_REG_ADDR(DT_NODELABEL(prs)),	\
 		.prs_sel = DT_PROP(DT_INST(inst, ite_it8xxx2_pwm), prescaler_cx),	\
 		.pcfg = PINCTRL_DT_INST_DEV_CONFIG_GET(inst),				\
--- a/dts/bindings/pwm/ite,it8xxx2-pwm.yaml
+++ b/dts/bindings/pwm/ite,it8xxx2-pwm.yaml
@@ -28,6 +28,12 @@ properties:
       3 = PWM_CHANNEL_3, 4 = PWM_CHANNEL_4, 5 = PWM_CHANNEL_5,
       6 = PWM_CHANNEL_6, 7 = PWM_CHANNEL_7
 
+  pwm-init-high:
+    type: boolean
+    description: |
+      After setting PWM pin to alternate mode, initialize PWM pin to output high,
+      otherwise output low.
+
   pwmctrl:
     type: phandle
     required: true


dts/bindings/pwm/ite,it8xxx2-pwm.yaml
```diff
@@ -28,6 +28,16 @@ properties:
...
 
+  init-output-level:
+    type: int
+    enum:
+      - 0
+      - 1
+      - 255
+    description: |
+      0 = PWM_IT8XXX2_INIT_OUTPUT_LOW, 1 = PWM_IT8XXX2_INIT_OUTPUT_HIGH,
+      255 = PWM_IT8XXX2_INIT_OUTPUT_NONE
```
fabiobaltieri: hey, something similar came up in the past in the context of display backlight control, I'm thinking it would make sense to have this property in `pwm-controller.yaml` and the definitions in `include/zephyr/dt-bindings/pwm/pwm.h`, what do you guys think? @anangl @henrikbrixandersen

include/zephyr/dt-bindings/pwm/it8xxx2_pwm.h
```diff
@@ -23,6 +23,11 @@
 #define PWM_CHANNEL_6		6
 #define PWM_CHANNEL_7		7
 
+/* Init PWM channel output level */
+#define PWM_IT8XXX2_INIT_OUTPUT_LOW	0
+#define PWM_IT8XXX2_INIT_OUTPUT_HIGH	1
+#define PWM_IT8XXX2_INIT_OUTPUT_NONE	0xff
```
fabiobaltieri: what does NONE do in this context? I'd imagine once you set the pin as PWM you'd either want it full off or full on.

RuibinChang: Because `init-output-level` property is not must required, so far only Kaladin needs this, the other boards don't declare this property, then  go through the original flow.

fabiobaltieri: ok but then what happens in practice when you don't specify the init level? sure it'll either stay high or low, then you may as well only have those two

RuibinChang: I remove `PWM_IT8XXX2_INIT_OUTPUT_NONE` in `.yaml`, now only HIGH/LOW can be set to the `init-output-level` property. However, for the boards that don't declare `init-output-level` property, a `PWM_INIT_OUTPUT_PROP_UNDEF` flag (defined in `driver.c`) is needed to configure them and the compatible original flow.

fabiobaltieri: Sorry I still don't understand, what happens on the PWM line if `init-output-level` is unset and that block of code is skipped? I'd imagine the line is going to stay low, then why not just default the property to low? Or have a `init-high` boolean property. I mean if the pin is initialized as output it has to be either high or low.

RuibinChang: If PWM pin doesn't connect to anything, and init-output-level is unset, after setting PWM pin to PWM mode, then the pin will output low. I updated it to init low if the property isn't declared.

fabiobaltieri: ok brilliant, getting a bit simpler now, so one more question, how about either:

1. adding a `default: PWM_IT8XXX2_INIT_OUTPUT_HIGH` to the binding, or, probably better
2. making this a boolean property, something like `ite,pwm-init-high` or something

With either of these you could remove the `COND_CODE_1`, just set it all the time

fabiobaltieri: Then the remaining question is whether this should be a generic property and go into pwm-controller.yaml but that probably needs a bit more research, like if it should be per channel or something like that, probably fine doing it just for ite now and then work on a general approach if other driver needs it, unless you are happy to put the work upfront and propose a generic API, maybe a `pwm-init-high-mask` with a bitmask of channels that should initialize high

RuibinChang: Done, add boolean property `pwm-init-high`.

dts/bindings/pwm/ite,it8xxx2-pwm.yaml
```diff
@@ -28,6 +28,12 @@ properties:
       3 = PWM_CHANNEL_3, 4 = PWM_CHANNEL_4, 5 = PWM_CHANNEL_5,
       6 = PWM_CHANNEL_6, 7 = PWM_CHANNEL_7
 
+  pwm-init-high:
+    type: boolean
+    description: |
+      After setting PWM pin to alternate mode, initialize PWM pin to output high,
+      otherwise output low.
```
henrikbrixandersen, line 35: This feels a bit like a hack to me, I wonder if we can come up with a more general solution to this.

The issue, I take it, happens in the window from the PWM driver is initialised until the PWM consumer (here the PWM LED driver) initialises the PWM channel?

dts/bindings/pwm/ite,it8xxx2-pwm.yaml
henrikbrixandersen: Yeah, maybe. Need to think about this.

dts/bindings/pwm/ite,it8xxx2-pwm.yaml
fabiobaltieri: Think so, the problem is that for gpio the pin is initialized by the comsumer so there's a chance to specify a init-high or init-low there but for pwm it's always an output and initialized at pwm driver level. I think we should just have this property at pwm-controller level but I'm happy with this on the driver in the meantime and have it moved separately

fabiobaltieri (dismissed):

dts/bindings/pwm/ite,it8xxx2-pwm.yaml
RuibinChang: The symptom is:
After gpio pin is set to PWM mode at init(), the pin will output low at default, so LED will be light (LED is low-activated). And until set_cycles() is called to set output high, then LED will be turn off the light.

So add property to set init output level for pin in pwm mode.

fabiobaltieri: actually now that I think about it, wouldn't `PWM_POLARITY_INVERTED` invert the behavior and fix the problem? As in: you'd expect the pwm to initialize in the inactive state by default, if the LED is active low that should have PWM_POLARITY_INVERTED set and the inactive state means high, would that make sense?

fabiobaltieri: Right so `PWM_POLARITY_INVERTED` is set on `pwm_it8xxx2_set_cycles`, but that's in devicetree, it's available in init, could you move the polarity handling bit here instead? both this feature and that are setting `**reg_pwmpol` so that seems to make a lot more sense to me.

henrikbrixandersen: `PWM_POLARITY_INVERTED` and related flags does not need to come from the devicetree (they can come directly from code) and even if they are present in devicetree, they're not directly available from the PWM controller driver code (as they are present on PWM consumers, not PWM providers - similar to GPIOs).

fabiobaltieri: ok sorry I was indeed messing things up, even when it's in dt the flag is in the consumer property so it's not available in init --- apologies for the noise scrape my comment

fabiobaltieri: so Brix how would you feel about this property living in `pwm-controller.yaml`?

RuibinChang: @henrikbrixandersen It would be appreciated if there are any comments.

fabiobaltieri (dismissed):

drivers/pwm/pwm_ite_it8xxx2.c
```diff
@@ -262,6 +266,18 @@ static int pwm_it8xxx2_init(const struct device *dev)
 	 */
 	inst->CTR1M = 0;
 
+	/* Set PWM init output level */
+	if (config->pwm_init_high) {
+		*reg_dcr = 0;
```
GTLin08: Since both branches set` *reg_dcr = 0;`, could this assignment be moved outside the conditional to avoid duplication?

RuibinChang: Done

GTLin08 (approved):

Chenhongren (approved):

PR #100153, opened 2025-11-27
Bluetooth: Classic: HFP_HF: Refactor at parser

lylezhu2012:
In current implementation, the additional buffer is required by AT to store the parsing AT response data temporarily. To store the parsed AT response data completely, the buffer size is as high as the MTU of the RFCOMM connection. And the buffer is dedicated for each HFP HF connection. That means the RAM usage depends on the MAX HFP HF connection count. Actually, the RFCOMM receiving buffer is valid when processing the received AT response. The changes aim to remove the additional buffer and leverage the receiving buffer to process the AT response.

And there is an issue found that the type of the additional buffer length is `uint8_t`, while the configured RFCOMM receiving data length is 65535. When the additional buffer length exceeds 255, the AT cannot process the received data normally.

Replace manual buffer management with net_buf_simple in the AT command parser. This change eliminates the need for manual position tracking and buffer length management by leveraging Zephyr's existing buffer utilities. And leverage RFCOMM receiving buffer instead of allocating dedicated buffer.

### DIFF
--- a/subsys/bluetooth/host/classic/at.c
+++ b/subsys/bluetooth/host/classic/at.c
@@ -17,11 +17,41 @@
 
 #include "at.h"
 
+typedef bool (*pull_char_t)(uint8_t *c, void *data);
+
+static void pull_chars(struct net_buf_simple *buf, pull_char_t cb, void *data)
+{
+	while (buf->len > 0) {
+		if (!cb(buf->data, data)) {
+			break;
+		}
+		net_buf_simple_pull_u8(buf);
+	}
+}
+
+struct at_comma_data {
+	bool found;
+};
+
+static bool pull_comma(uint8_t *c, void *data)
+{
+	struct at_comma_data *comma_data = data;
+
+	if (comma_data->found) {
+		return false;
+	}
+
+	comma_data->found = *c == ',';
+	return comma_data->found;
+}
+
 static void next_list(struct at_client *at)
 {
-	if (at->buf[at->pos] == ',') {
-		at->pos++;
-	}
+	struct at_comma_data data;
+
+	data.found = false;
+
+	pull_chars(&at->rsp_buf, pull_comma, &data);
 }
 
 int at_check_byte(struct net_buf *buf, char check_byte)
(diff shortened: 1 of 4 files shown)


MarkWangChinese (dismissed):
LGTM

makeshi (dismissed):

subsys/bluetooth/host/classic/at.c
```diff
@@ -38,82 +44,105 @@ int at_check_byte(struct net_buf *buf, char check_byte)
...
+		init_rsp_buffer(at, buf);
+	}
+
+	while (buf->len > 0) {
+		if (buf->data[0] == stop_byte) {
 			at->cmd_state = cmd_state;
+			net_buf_pull_u8(buf);
 			break;
 		}
+		net_buf_simple_add(&at->rsp_buf, sizeof(uint8_t));
```
jhedberg, line 183: Aren't you missing initialization of the added byte here? Maybe you'd want this instead?
```c
	net_buf_simple_add_u8(&at->rsp_buf, '\0');
```

subsys/bluetooth/host/classic/at.c
```diff
@@ -19,8 +19,14 @@
...
-	if (at->buf[at->pos] == ',') {
-		at->pos++;
+	char c = ',';
+
+	if (at->rsp_buf.len < sizeof(c)) {
+		return;
+	}
+
+	if (at->rsp_buf.data[0] == (uint8_t)c) {
+		net_buf_simple_pull(&at->rsp_buf, sizeof(c));
```
jhedberg: It seems to me like you could simplify quite a lot of code by adding a helper something like the following:
```c

typedef int (*char_cmp)(int c);

static bool pull_char(struct net_buf_simple *buf, char_cmp cmp, char *pulled)
{
	if (buf->len < 1) {
		return false;
	}

	if (!cmp((int)buf->data[0])) {
		return false;
	}

	if (pulled) {
		*pulled = (char)buf->data[0];
	}

	net_buf_simple_pull(buf, 1);
	
	return true;	
}
```

subsys/bluetooth/host/classic/at.c
```diff
@@ -38,82 +44,105 @@ int at_check_byte(struct net_buf *buf, char check_byte)
...
-	while (at->buf[at->pos] == ' ') {
-		at->pos++;
+	char space = ' ';
+
+	while (at->rsp_buf.len > 0) {
+		if (at->rsp_buf.data[0] != (uint8_t)space) {
+			break;
+		}
+		net_buf_simple_pull(&at->rsp_buf, sizeof(space));
 	}
```
jhedberg: With my helper proposal the above would become:
```c
static int is_space_char(int c)
{
	return (c == (int)' ');
}

static void skip_space(struct at_client *at)
{
	while (pull_char(&at->rsp_buf, is_space_char, NULL));
}
```

subsys/bluetooth/host/classic/at.c
```diff
@@ -38,82 +44,105 @@ int at_check_byte(struct net_buf *buf, char check_byte)
...
+	*val = 0;
+
+	while (at->rsp_buf.len > 0) {
+		if (isdigit((unsigned char)at->rsp_buf.data[0]) == 0) {
+			break;
+		}
+
+		*val = *val * 10 + at->rsp_buf.data[0] - (uint8_t)'0';
+		net_buf_simple_pull_u8(&at->rsp_buf);
 	}
```
jhedberg: This would become:
```c
	char digit;
	
	while (pull_char(at->rsp_buf, isdigit, &digit)) {
		*val = *val * 10 + digit - (uint8_t)'0';
	}
```

jhedberg (changes requested):
"Requesting changes" until my comments are answered (otherwise this is at risk at getting merged)

jhedberg (changes requested):
"Requesting changes" until my comments are answered (otherwise this is at risk at getting merged)

subsys/bluetooth/host/classic/at.c
lylezhu2012: The buffer of the data is came from the L2CAP RECV. The `at->rsp_buf` is initialised by using receiving buffer `buf`. And the valid data length is 0 by default. Please refer to,
```
	if (at->rsp_buf.len == 0) {
		init_rsp_buffer(at, buf);
	}
```
The actual data length needs to be calculated by AT driver.
 
The change here is only used to calculate the length of the data needed to be parsed. So just update the `at->rsp_buf->len` here.

So the data of `at->rsp_buf` will not be changed here.

subsys/bluetooth/host/classic/at.c
lylezhu2012: Yes. It is a good idea. It can help make the code more readable.

subsys/bluetooth/host/classic/at.c
jhedberg: Ok. I see. Normally the data between `buf->data` and `buf->data + buf->len` is what's considered the "valid" part of the buffer, and anything outside the "invalid" (or at least uninitialized) part. However, if you're explicitly initializing the entire buffer  elsewhere then I guess it's fine (just a bit confusing for the reader - I wonder if this could be clarified somehow, e.g. with code comments).

subsys/bluetooth/host/classic/at.c
jhedberg: Since my other comment turned out not to be an actual bug, do you want to do this refactoring as part of this PR or in a later one? This proposal wasn't a blocking review from my side.

lylezhu2012: Updated.

subsys/bluetooth/host/classic/at.c
lylezhu2012: Updated.

subsys/bluetooth/host/classic/at.c
lylezhu2012: Updated.

subsys/bluetooth/host/classic/at.c
lylezhu2012: > Since my other comment turned out not to be an actual bug, do you want to do this refactoring as part of this PR or in a later one? This proposal wasn't a blocking review from my side.

I have update it.

subsys/bluetooth/host/classic/at.c
lylezhu2012: Sure. Let me add some comments for it.

jhedberg (dismissed):

subsys/bluetooth/host/classic/at.c
lylezhu2012: Added comments for it.

MarkWangChinese (approved):

makeshi (approved):

PR #100157, opened 2025-11-27
tests: drivers: dma: change power-state checks in dma loop test

Martinhoff-maker:
This patch allows a DMA device that is part of a power domain to start the test in a suspended power-management state. If another device makes the power domain active (for example, the console UART), the DMA device within the same power domain may enter a suspended state. The test will now verify that the device is suspended/off before the test begins, active during the test, and restored to its initial state at the end.


teburd (approved):

hakehuang (approved):
regression pass on NXP platforms

PR #100164, opened 2025-11-27
maintainers: Update to reflect current status for my work

carlescufi:
I unfortunately do not have the time to maintain documentation, so mark myself as collaborator instead of maintainer. Since most of my work now revolves aournd the nRF platforms, add myself as a maintainer there.

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -1205,8 +1205,8 @@ Documentation:
   status: maintained
   maintainers:
     - kartben
-    - carlescufi
   collaborators:
+    - carlescufi
     - nashif
   files:
     - CODE_OF_CONDUCT.md
@@ -6424,6 +6424,7 @@ nRF Platforms:
     - anangl
     - masz-nordic
     - bjarki-andreasen
+    - carlescufi
   collaborators:
     - jaz1-nordic
     - kl-cruz


MAINTAINERS.yml
```diff
@@ -6424,6 +6424,7 @@ nRF Platforms:
     - anangl
     - masz-nordic
     - bjarki-andreasen
+    - carlescufi
```
kartben, line 6427: Mmmh, do you really need _four_ maintainers? This is really too many IMO, and something we've typically been trying to avoid.

carlescufi: Ideally yes, because we do have quite a lot of PRs coming in from multiple teams at Nordic, and different PRs typically need different reviewers. If you don't mind, this would make our lives easier.

kartben: > ... we do have quite a lot of PRs coming in from multiple teams at Nordic, and different PRs typically need different reviewers ...

But that doesn't sound right, you are saying that there are PRs that end up being assigned to the wrong people / people not in a position to properly approve/review based on some internal organization criteria that shouldn't really be the project's concern. Can't the "Nordic Platform" area be split according to whatever this split is?

carlescufi: Right, that's what I am saying. The current "nRF Platform" entry covers a lot of ground, and depending on the area touched one or another maintainer is required. So if the solution is to split it up I wonder if there are any precedents of splitting up a platform entry in maintainers so I can take some inspiration from.

kartben: > Right, that's what I am saying. The current "nRF Platform" entry covers a lot of ground, and depending on the area touched one or another maintainer is required. So if the solution is to split it up I wonder if there are any precedents of splitting up a platform entry in maintainers so I can take some inspiration from.

To be completely honest, based on your recent "approval" activity as an nRF _collaborator_ it really doesn't feel like  adding a fourth _maintainer_ is needed at all

PR #100078, opened 2025-11-26
silabs: siwx91x: Fix to enable PSRAM half-sleep support to reduce sleep current

fimohame:
Previously, the PSRAM boards exhibited high sleep current due to lack of half-sleep support. This PR implements half-sleep mode for PSRAM, reducing power consumption during device sleep.

Changes include:
- DeviceTree and bindings updated to support 'half-sleep-supported' and 'ldo' for PSRAM
- Build system updates for necessary sources and driver flags

With these fixes, PSRAM now enters half-sleep mode correctly, resulting in lower sleep current on siwx91x-PSRAM based boards which supports half sleep mode.

Signed-off-by: S Mohamed Fiaz <Fiaz.Mohamed@silabs.com>

PR #100168, opened 2025-11-27
soc: silabs: siwx91x: removed sscanf for nwp firmware version check

Martinhoff-maker:
This patch removes the use of sscanf to maintain compatibility with tests that use the minimal cpp library. The expected version is now defined using multiple individual values rather than a single formatted string.

Fixes CI failure here:
https://github.com/zephyrproject-rtos/zephyr/actions/runs/19623866468/job/56189276469?pr=99860
and here:
https://github.com/zephyrproject-rtos/zephyr/pull/99400
and here:
https://github.com/zephyrproject-rtos/zephyr/pull/100078

### DIFF
--- a/modules/hal_silabs/wiseconnect/nwp_fw_version.c
+++ b/modules/hal_silabs/wiseconnect/nwp_fw_version.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2025 Silicon Laboratories Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* This value needs to be updated when the Wiseconnect SDK (hal_silabs/wiseconnect) is updated
+ * Currently mapped to Wiseconnect SDK 3.5.2
+ */
+
+#include <nwp_fw_version.h>
+
+const sl_wifi_firmware_version_t siwx91x_nwp_fw_expected_version = {
+	.rom_id = 0x0B,
+	.major = 2,
+	.minor = 14,
+	.security_version = 5,
+	.patch_num = 2,
+	.customer_id = 0,
+	.build_num = 7,
+};
--- a/modules/hal_silabs/wiseconnect/nwp_fw_version.h
+++ b/modules/hal_silabs/wiseconnect/nwp_fw_version.h
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/* This value needs to be updated when the Wiseconnect SDK (hal_silabs/wiseconnect) is updated
- * Actually mapped to Wiseconnect SDK 3.5.0
- */
-#define SIWX91X_NWP_FW_EXPECTED_VERSION "B.2.14.5.2.0.7"
+#include <sl_wifi_types.h>
+
+extern const sl_wifi_firmware_version_t siwx91x_nwp_fw_expected_version;
--- a/soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
+++ b/soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
@@ -261,7 +261,6 @@ static void siwx91x_configure_network_stack(sl_si91x_boot_configuration_t *boot_
 
 static int siwx91x_check_nwp_version(void)
 {
-	sl_wifi_firmware_version_t expected_version;
 	sl_wifi_firmware_version_t version;
 	int ret;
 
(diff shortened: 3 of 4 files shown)


soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
```diff
@@ -261,24 +261,23 @@ static void siwx91x_configure_network_stack(sl_si91x_boot_configuration_t *boot_
...
 	sl_wifi_firmware_version_t version;
+	sl_wifi_firmware_version_t expected_version = {
+		.rom_id = SIWX91X_NWP_FW_EXPECTED_ROM_ID,
+		.major = SIWX91X_NWP_FW_EXPECTED_MAJOR,
+		.minor = SIWX91X_NWP_FW_EXPECTED_MINOR,
+		.security_version = SIWX91X_NWP_FW_EXPECTED_SECURITY_VERSION,
+		.patch_num = SIWX91X_NWP_FW_EXPECTED_PATCH_NUM,
+		.customer_id = SIWX91X_NWP_FW_EXPECTED_CUSTOMER_ID,
+		.build_num = SIWX91X_NWP_FW_EXPECTED_BUILD_NUM,
+	};
```
jerome-pouiller: I suggest to avoid these intermediate symbols. You can directly declare a `static const sl_wifi_firmware_version_t siwx91x_nwp_fw_expected_version ...` in `modules/hal_silabs/wiseconnect/nwp_fw_version.h`.

modules/hal_silabs/wiseconnect/nwp_fw_version.h
```diff
@@ -5,6 +5,12 @@
  */
 
 /* This value needs to be updated when the Wiseconnect SDK (hal_silabs/wiseconnect) is updated
- * Actually mapped to Wiseconnect SDK 3.5.0
+ * Actually mapped to Wiseconnect SDK 3.5.2
  */
-#define SIWX91X_NWP_FW_EXPECTED_VERSION "B.2.14.5.2.0.7"
+#define SIWX91X_NWP_FW_EXPECTED_ROM_ID           11
```
jerome-pouiller: `0x0B` ?

soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
Martinhoff-maker: I've done that to avoid `#include "sl_wifi_types.h` in the file and to limit propagation of dependency

modules/hal_silabs/wiseconnect/nwp_fw_version.h
Martinhoff-maker: agree, it will be more user friendly to update the version if he clearly see the pattern

soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
Martinhoff-maker: changed

jerome-pouiller (dismissed):

jhedberg (dismissed):

modules/hal_silabs/wiseconnect/nwp_fw_version.h
```diff
@@ -5,6 +5,17 @@
...
+
+static const sl_wifi_firmware_version_t siwx91x_nwp_fw_expected_version = {
+	.rom_id = 0x0B,
+	.major = 2,
+	.minor = 14,
+	.security_version = 5,
+	.patch_num = 2,
+	.customer_id = 0,
+	.build_num = 7,
+};
```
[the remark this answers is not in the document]
Martinhoff-maker: it will be fine since it's only a one time reference so there will only be 1 copy.

soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
```diff
@@ -270,38 +269,29 @@ static int siwx91x_check_nwp_version(void)
...
-	       &expected_version.major,
-	       &expected_version.minor,
-	       &expected_version.security_version,
-	       &expected_version.patch_num,
-	       &expected_version.customer_id,
-	       &expected_version.build_num);
-
 	/* Ignore rom_id:
 	 * B is parsed as an hex value and we get 11 in expected_version.rom_id
 	 * We received rom_id=17 in version.rom_id, we suspect a double hex->decimal conversion
```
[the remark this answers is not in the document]
Martinhoff-maker: updated

modules/hal_silabs/wiseconnect/nwp_fw_version.h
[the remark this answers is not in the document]
jhedberg: > it will be fine since it's only a one time reference so there will only be 1 copy.

Header files usually come with the implication that they may be included in multiple places (otherwise there's little benefit in a header file compared to just putting the content in the single c-file that needs it). Would it make sense to just move this into the c-file then? Alternatively, put the struct into the c-file, but leave defines in the header file for each of these values, which the c-file then uses to initialize the struct.

Another thought (although admittedly out of scope for this PR): even though hal_silabs contains an import of actual wiseconnect files, would it perhaps make sense to have our import script generate a header file in the hal_tree with the imported version, which is then incuded from the Zephyr main tree? That way we'd get stronger linkage with what's actually in the HAL tree and less risk of mistakes (forgetting to update the main tree).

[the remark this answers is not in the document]
Martinhoff-maker: but adressed

modules/hal_silabs/wiseconnect/nwp_fw_version.h
```diff
@@ -5,6 +5,17 @@
  */
 
 /* This value needs to be updated when the Wiseconnect SDK (hal_silabs/wiseconnect) is updated
- * Actually mapped to Wiseconnect SDK 3.5.0
+ * Actually mapped to Wiseconnect SDK 3.5.2
```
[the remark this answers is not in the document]
Martinhoff-maker: changed

modules/hal_silabs/wiseconnect/nwp_fw_version.h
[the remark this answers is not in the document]
Martinhoff-maker: To answer your comment, I don't want to only have it in the c-file because I think we lost the link that this is associated to the Wiseconnect module. For the alternative, this what I've done at the start (see the diff).

Yeah that could be a solution. I thought that we disallow to have out of hal content (files) in the hal_silabs repo in order to have a better maintainability of the repo. That said, if it's part of the import script, that will not create maintenance issue.

Martinhoff-maker:
V2: doing a compromise between information location and declaration. 

Plan to take a action to create a file in the hal with the import script for each new hal update.

modules/hal_silabs/wiseconnect/nwp_fw_version.c
```diff
@@ -0,0 +1,21 @@
...
+
+const sl_wifi_firmware_version_t siwx91x_nwp_fw_expected_version = {
+	.rom_id = 0x0B,
+	.major = 2,
+	.minor = 14,
+	.security_version = 5,
+	.patch_num = 2,
+	.customer_id = 0,
+	.build_num = 7,
+};
```
jhedberg, line 21: Not blocking this because of it, but I'm still a bit confused by why you need this struct variable at all. Can't you just have stuff like:
```c
#define SIWX91X_NWP_EXPECTED_ROM_ID 0x0B
#define SIWX91X_NWP_EXPECTED_MAJOR  2
#define SIWX91X_NWP_EXPECTED_MINOR  14
...
```

And then in the version checking code you'd have stuff like
```c
	if (version.major != SIWX91X_NWP_EXPECTED_MAJOR) {
		return -EINVAL;
	}

	...
```

Martinhoff-maker: That's what I've done first. I don't have strong opinion between the 2 ways of doing this. I suggest to merge it like this and I will follow up with the hal change, or you can discuss about it with @jerome-pouiller.

jerome-pouiller: My main concern in the original PR was about the [number of indirection layers to the real result][1] (I wanted to drop either `expected_version`, either the defines).

In addition, I believe that exposing a structured data would make sense here (while the defines are only linked by the common prefix in their names).

[1]: https://github.com/zephyrproject-rtos/zephyr/commit/2455fb344465ac9ce829e6801083ee6a5fc97c21#diff-11a824d4dafe70e6d9ce1fb166acf120f1c766101ac5cbdc171614dae0881307R265

jhedberg: @jerome-pouiller if we did a direct comparison of two structs, that would indeed make a lot of sense. However, the code is doing component-by-component comparisons. Also:
 - We don't care about all struct fields (unless I misunderstood something), rather only a subset, so there's some wasted space with defining the entire struct
 - With the plan to move this to be auto-generated in the HAL (by the import script) having a struct would mean needing both a header file and a c-file, which seems a bit overkill for just a version check.

Anyway, I don't want to get stuck on this issue, since it's more important to get the CI problems sorted out. We can sort out the rest in a follow-up PR like @Martinhoff-maker proposes.

jerome-pouiller: My main concern was redundancy between defines and `expected_version` variable.

Outside of that, I don't really mind.

jerome-pouiller (approved):

jhedberg (approved):

asmellby (approved):

PR #100174, opened 2025-11-27
boards: nxp: Fix rt700 display sample build cmd in doc

yvesll:
- Use correct board with id.
- Move display support chapter before board footer.


JimmyChen-NXP (changes requested):
In commit head, there is typo for the baords, it shall be boards

DerekSnell (approved):

JimmyChen-NXP (approved):

PR #100176, opened 2025-11-27
drivers: usb: udc: stm32: driver rework for instance-aware support

mathieuchopstm:
Rework STM32 UDC driver to be instance-aware-ish: only one instance is still supported, but the code is organized to regroup all instance-specific data together, such that transition towards a multi-instance driver can be done cleanly.

Multi-instance is not implemented as it requires more heavy changes for PHY management - these will be provided by another PR.

### DIFF
--- a/drivers/usb/udc/udc_stm32.c
+++ b/drivers/usb/udc/udc_stm32.c
@@ -188,6 +188,9 @@ struct udc_stm32_config {
 	/* EP configurations */
 	struct udc_ep_config *in_eps;
 	struct udc_ep_config *out_eps;
+	/* Worker thread stack info */
+	k_thread_stack_t *thread_stack;
+	size_t thread_stack_size;
 	/* Maximal packet size allowed for endpoints */
 	uint16_t ep_mps;
 	/* Number of entries in `pclken` */
@@ -1234,6 +1237,8 @@ static const struct udc_api udc_stm32_api = {
  * WARNING: Don't mix USB defined in STM32Cube HAL and CONFIG_USB_* from Zephyr
  * Kconfig system.
  */
+K_THREAD_STACK_DEFINE(udc0_thr_stk, CONFIG_UDC_STM32_STACK_SIZE);
+
 static struct udc_ep_config udc0_in_ep_cfg[DT_INST_PROP(0, num_bidir_endpoints)];
 static struct udc_ep_config udc0_out_ep_cfg[DT_INST_PROP(0, num_bidir_endpoints)];
 
@@ -1269,6 +1274,8 @@ static const struct udc_stm32_config udc0_cfg  = {
 	.ep_mps = UDC_STM32_NODE_EP_MPS(DT_DRV_INST(0)),
 	.selected_phy = UDC_STM32_NODE_PHY_ITFACE(DT_DRV_INST(0)),
 	.selected_speed = UDC_STM32_NODE_SPEED(DT_DRV_INST(0)),
+	.thread_stack = udc0_thr_stk,
+	.thread_stack_size = K_THREAD_STACK_SIZEOF(udc0_thr_stk),
 };
 
 static int udc_stm32_clock_enable(const struct device *dev)
@@ -1505,8 +1512,6 @@ static const struct gpio_dt_spec ulpi_reset =
 
 static char udc_msgq_buf_0[CONFIG_UDC_STM32_MAX_QMESSAGES * sizeof(struct udc_stm32_msg)];
 
-K_THREAD_STACK_DEFINE(udc_stm32_stack_0, CONFIG_UDC_STM32_STACK_SIZE);
-
 static int udc_stm32_driver_init0(const struct device *dev)
 {
 	struct udc_stm32_data *priv = udc_get_private(dev);
@@ -1567,8 +1572,8 @@ static int udc_stm32_driver_init0(const struct device *dev)
 	k_msgq_init(&priv->msgq_data, udc_msgq_buf_0, sizeof(struct udc_stm32_msg),
 		    CONFIG_UDC_STM32_MAX_QMESSAGES);
 
-	k_thread_create(&priv->thread_data, udc_stm32_stack_0,
-			K_THREAD_STACK_SIZEOF(udc_stm32_stack_0), udc_stm32_thread_handler,
+	k_thread_create(&priv->thread_data, cfg->thread_stack,
+			cfg->thread_stack_size, udc_stm32_thread_handler,
 			(void *)dev, NULL, NULL, K_PRIO_COOP(CONFIG_UDC_STM32_THREAD_PRIORITY),
 			K_ESSENTIAL, K_NO_WAIT);
 	k_thread_name_set(&priv->thread_data, dev->name);
--- a/drivers/usb/udc/udc_stm32.c
+++ b/drivers/usb/udc/udc_stm32.c
@@ -49,9 +49,6 @@ LOG_MODULE_REGISTER(udc_stm32, CONFIG_UDC_DRIVER_LOG_LEVEL);
 #define UDC_STM32_IRQ_NAME     usb
 #endif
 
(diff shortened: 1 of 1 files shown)


drivers/usb/udc/udc_stm32.c
```diff
@@ -1239,13 +1238,22 @@ static struct udc_data udc0_data = {
 	.priv = &udc0_priv,
 };
 
+static void udc0_irq_connect(void)
+{
+	IRQ_CONNECT(
+		DT_INST_IRQ_BY_NAME(0, UDC_STM32_IRQ_NAME, irq),
+		DT_INST_IRQ_BY_NAME(0, UDC_STM32_IRQ_NAME, priority),
+		HAL_PCD_IRQHandler, &udc0_priv.pcd, 0);
```
jfischer-no: Formating is a bit weird, can you fix it please?

```suggestion
	IRQ_CONNECT(DT_INST_IRQ_BY_NAME(0, UDC_STM32_IRQ_NAME, irq),
		    DT_INST_IRQ_BY_NAME(0, UDC_STM32_IRQ_NAME, priority),
		    HAL_PCD_IRQHandler, &udc0_priv.pcd, 0);
```

drivers/usb/udc/udc_stm32.c
```diff
@@ -1570,7 +1569,7 @@ static int udc_stm32_driver_init0(const struct device *dev)
 
 	priv->dev = dev;
 
-	k_msgq_init(&priv->msgq_data, udc_msgq_buf_0, sizeof(struct udc_stm32_msg),
+	k_msgq_init(&priv->msgq_data, priv->msgq_buf, sizeof(struct udc_stm32_msg),
```
jfischer-no, line 1573: Before you rework it further, I recommend switching to k_event first.

mathieuchopstm: I already have a full patchset that adds multi-instance support ready: https://github.com/mathieuchopstm/zephyr/commits/topic/STM32-USB-CleanUp-Full/

I am just submitting it in smaller pieces for easier review (what remains should be integrated in tree through 2~3 PRs).

As such, I would like to avoid functional changes until multi-instance is fully integrated in tree. (One might argue parts of these PRs *are* functional changes, but they should be ultimately equivalent to the current behavior)

drivers/usb/udc/udc_stm32.c
mathieuchopstm: Done.

drivers/usb/udc/udc_stm32.c
mathieuchopstm: Though on this topic, I assume your proposal is something like this?

<details>

```diff
#define EVENT_SETUP BIT(0)
#define EVENT_IN BIT(1)
#define EVENT_OUT BIT(2)

void HAL_PCD_DataOutStageCallback(PCD_HandleTypeDef *hpcd, uint8_t epnum)
{
	uint32_t rx_count = HAL_PCD_EP_GetRxCount(hpcd, epnum);
	struct udc_stm32_data *priv = hpcd2data(hpcd);
-	struct udc_stm32_msg msg = {
-		.type = UDC_STM32_MSG_DATA_OUT,
-		.ep = epnum,
-		.rx_count = rx_count,
-	};
-	int err;
-
-	err = k_msgq_put(&priv->msgq_data, &msg, K_NO_WAIT);
-	if (err != 0) {
-		LOG_ERR("UDC Message queue overrun");
-	}

+	//immediately add `rx_count` to pending buffer of `epnum`
+	/* ... */
+
+	priv->pending_out_ep |= BIT(epnum);
+	k_event_set(priv->evf, EVENT_OUT);
}

void HAL_PCD_DataInStageCallback(PCD_HandleTypeDef *hpcd, uint8_t epnum)
{
	struct udc_stm32_data *priv = hpcd2data(hpcd);
-	struct udc_stm32_msg msg = {
-		.type = UDC_STM32_MSG_DATA_IN,
-		.ep = epnum,
-	};
-	int err;
-
-	err = k_msgq_put(&priv->msgq_data, &msg, K_NO_WAIT);
-	if (err != 0) {
-		LOG_ERR("UDC Message queue overrun");
-	}
+ 	priv->pending_in_ep |= BIT(epnum);
+	k_event_set(priv->evf, EVENT_IN);
}

static void udc_stm32_thread_handler(void *arg1, void *arg2, void *arg3)
{
	const struct device *dev = arg1;
	struct udc_stm32_data *priv = udc_get_private(dev);
	struct udc_stm32_msg msg;

	while (true) {
-		k_msgq_get(&priv->msgq_data, &msg, K_FOREVER);
-		switch (msg.type) {
-		case UDC_STM32_MSG_SETUP:
-			handle_msg_setup(priv);
-			break;
-		case UDC_STM32_MSG_DATA_IN:
-			handle_msg_data_in(priv, msg.ep);
-			break;
-		case UDC_STM32_MSG_DATA_OUT:
-			handle_msg_data_out(priv, msg.ep, msg.rx_count);
-			break;
-		}
+		uint32_t events;
+		k_event_get(priv->evf, &events, K_FOREVER);
+
+		if (events & EVENT_SETUP) {
+			handle_msg_setup(priv);
+		}
+		if (events & EVENT_IN) {
+			/* foreach bit in pending_in_ep: */
+				handle_msg_data_out(priv, LOG2(bit));
+		}
+		if (events & EVENT_OUT) {
+			/* foreach bit in pending_out_ep: */
+				handle_msg_data_in(priv, LOG2(bit));
+		}
	}
}
```

(maybe using `k_event` for the `pending_{in,out}_ep` bitfields too)

</details>

jfischer-no: > As such, I would like to avoid functional changes until multi-instance is fully integrated in tree.

It is up to you.

> Though on this topic, I assume your proposal is something like this?

You can use udc_sam0.c or udc_dwc2.c drivers as a reference, in DWC2 driver it was introduced in commits c2f2d8ce5d9f7485892265d24dec8fb48de84e61, 2aa26ad2615297c3c69649cfd89b345838413c75, 264e9feca09e531ee8bbb34c2b437e5f8ccd3367.

jfischer-no: btw, msgq allows you to have single thread for all driver instances, whether it makes sense here or not, but with k_events you will need a thread per instance.

drivers/usb/udc/udc_stm32.c
```diff
@@ -1277,6 +1284,10 @@ static const struct udc_stm32_config udc0_cfg  = {
 	.selected_speed = UDC_STM32_NODE_SPEED(DT_DRV_INST(0)),
 	.thread_stack = udc0_thr_stk,
 	.thread_stack_size = K_THREAD_STACK_SIZEOF(udc0_thr_stk),
+
+#if DT_INST_NODE_HAS_PROP(0, disconnect_gpios)
+	.disconnect_gpio = &udc0_disconnect_gpio,
```
jfischer-no: Please remove the two `#if DT_INST_NODE_HAS_PROP(0, disconnect_gpios)`. Add `const struct gpio_dt_spec disconnect_gpio;` to driver's config and use `.disconnect_gpio = GPIO_DT_SPEC_INST_GET_OR(0, disconnect_gpios, {0}),` here, and in the code above replace the check with `if (config->disconnect_gpio.port) {}`.

mathieuchopstm: I initially thought it'd be cheaper to have the `gpio_dt_spec` out of band (saves 4 bytes when not present, which is the common use case if tree is to be trusted), but after more thinking your proposal should be about as expensive.

Changing to `GPIO_DT_SPEC_INST_GET_OR()`.

(Either way, when multi-instance is fully integrated, it should be easy to get rid of this overhead when no instance has `disconnect_gpios`)

jfischer-no (approved):

drivers/usb/udc/udc_stm32.c
jfischer-no: But the code looks much better now, and I think it is even worth sacrificing ten times that amount of flash.

erwango (approved):

PR #100181, opened 2025-11-27
samples: usb: common: make vid configurable

fabiobaltieri:
Add a SAMPLE_USBD_VID Kconfig to allow changing the USB VID on samples using the common code. Can be handy for testing things against code that expects a specific vid.

### DIFF
--- a/samples/subsys/usb/common/Kconfig.sample_usbd
+++ b/samples/subsys/usb/common/Kconfig.sample_usbd
@@ -21,6 +21,12 @@ config SAMPLE_USBD_PRODUCT
 	help
 	  USB device sample product stringa.
 
+config SAMPLE_USBD_VID
+	hex "USB device sample Vendor ID"
+	default 0x2fe3
+	help
+	  USB device sample Vendor ID.
+
 config SAMPLE_USBD_PID
 	hex "USB device sample Product ID"
 	default 0x0001
--- a/samples/subsys/usb/common/Kconfig.sample_usbd
+++ b/samples/subsys/usb/common/Kconfig.sample_usbd
@@ -25,13 +25,16 @@ config SAMPLE_USBD_VID
 	hex "USB device sample Vendor ID"
 	default 0x2fe3
 	help
-	  USB device sample Vendor ID.
+	  USB device sample Vendor ID. The default id (0x2fe3) is associated to
+	  Zephyr Project, you must use your own VID for samples and applications
+	  outside of Zephyr Project.
 
 config SAMPLE_USBD_PID
 	hex "USB device sample Product ID"
 	default 0x0001
 	help
-	  USB device sample Product ID.
+	  USB device sample Product ID. You must use your own PID for samples
+	  and applications outside of Zephyr Project.
 
 config SAMPLE_USBD_SELF_POWERED
 	bool "USB device sample Self-powered attribute"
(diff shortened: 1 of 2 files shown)


jfischer-no (dismissed):

samples/subsys/usb/common/Kconfig.sample_usbd
```diff
@@ -21,6 +21,12 @@ config SAMPLE_USBD_PRODUCT
 	help
 	  USB device sample product stringa.
 
+config SAMPLE_USBD_VID
+	hex "USB device sample Vendor ID"
+	default 0x2fe3
+	help
+	  USB device sample Vendor ID.
```
josuah: Some hint about what 0x2fe3 is and how it can be used?

If taking the CDC ACM sample:

https://github.com/zephyrproject-rtos/zephyr/blob/8e687fc0dd6e20dd75f9e6e8c29287cc81cfefe9/subsys/usb/device_next/app/Kconfig.cdc_acm_serial#L41-L43

josuah (commented):
LGTM but giving a chance to answer before the +1...

P.S.: For curiosity: https://github.com/search?q=repo%3Azephyrproject-rtos%2Fzephyr%200x2fe3&type=code

samples/subsys/usb/common/Kconfig.sample_usbd
fabiobaltieri: Sure why not.

josuah (dismissed):
Thanks!

samples/subsys/usb/common/Kconfig.sample_usbd
```diff
@@ -25,13 +25,16 @@ config SAMPLE_USBD_VID
 	hex "USB device sample Vendor ID"
 	default 0x2fe3
 	help
-	  USB device sample Vendor ID.
+	  USB device sample Vendor ID. The default id (0x2fe3) is associated to
+	  NordicSemiconductor, you must use your own VID for samples and
```
jfischer-no: `to Zephyr Project`, see also https://www.usb.org/sites/default/files/vendor_ids10282025b_1.pdf `Decimal Format 12259`.

jfischer-no (changes requested):

samples/subsys/usb/common/Kconfig.sample_usbd
fabiobaltieri: Interesting.

```
/usr/share/misc$ grep 2fe3 usb.ids 
2fe3  NordicSemiconductor
```

how do we fix this one?

tmon-nordic: The IDs are taken from http://www.linux-usb.org/usb-ids.html database. Fix would be in this database.

I reported this in 2022 https://usb-ids.gowdy.us/read/UD/2fe3 but looks like I have chosen wrong report channel... (because no action was taken)

jfischer-no (approved):

samples/subsys/usb/common/Kconfig.sample_usbd
fabiobaltieri: Yeah @jfischer-no and I were looking into it, seems like we can't even make an account with the website to add a comment, though it has been updated thre months ago... we'll figure something.

jfischer-no: Opened https://github.com/zephyrproject-rtos/zephyr/discussions/100305
@aurel32 FYI ^

tmon-nordic (approved):

PR #100182, opened 2025-11-27
arch/posix: Support cross-compiling

### LINKED ISSUE #70967: How build cross-compile native_posix app
Hi,

I what run Zephyr samples on Raspberry Pi 4B(Which is arm linux os),

How i can build some zephyr sample run directly run Raspberry Pi?

i'm test native_posix, but this board only generate host exe not arm linux.

aescolar:
Support cross-compiling native/posix_arch boards from one Linux host for another Linux target host.
This is done by allowing users to set NATIVE_TARGET_HOST for cmake which will override the automatic detection.

New doc: https://builds.zephyrproject.io/zephyr/pr/100182/docs/boards/native/native_sim/doc/cross_compile.html 

As pre-work: Fix building native/posix_arch boards on 32bit arm hosts.

Fixes #70967
Fixes: #99434

### DIFF
--- a/arch/posix/CMakeLists.txt
+++ b/arch/posix/CMakeLists.txt
@@ -39,32 +39,27 @@ set_property(TARGET native_simulator PROPERTY LOCALIZE_EXTRA_OPTIONS "")
 
 set(NSI_DIR ${ZEPHYR_BASE}/scripts/native_simulator CACHE PATH "Path to the native simulator")
 
-if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${CMAKE_HOST_SYSTEM_NAME}.${CMAKE_HOST_SYSTEM_PROCESSOR}.cmake)
-  # @Intent: Set necessary compiler & linker options for this specific host architecture & OS
-  include(${CMAKE_HOST_SYSTEM_NAME}.${CMAKE_HOST_SYSTEM_PROCESSOR}.cmake)
-else() # Linux.x86_64
-  if(CONFIG_64BIT)
-    # some gcc versions fail to build without -fPIC
-    zephyr_compile_options(-m64 -fPIC)
-    zephyr_link_libraries(-m64)
-
-    target_link_options(native_simulator INTERFACE "-m64")
-    target_compile_options(native_simulator INTERFACE "-m64")
+if(NATIVE_TARGET_HOST) # Allow users to manually select the target for cross-compiling use cases
+  set(TARGET_HOST ${NATIVE_TARGET_HOST})
+else()
+  if("${CMAKE_HOST_SYSTEM_PROCESSOR}" MATCHES "arm.*")
+    # All 32bit arm variants
+    set(TARGET_HOST "arm")
+  elseif("${CMAKE_HOST_SYSTEM_PROCESSOR}" MATCHES ".*86.*")
+    # x86_64/i*86
+    set(TARGET_HOST "x86_64")
   else()
-    zephyr_compile_options(-m32)
-    zephyr_link_libraries(-m32)
-
-    target_link_options(native_simulator INTERFACE "-m32")
-    target_compile_options(native_simulator INTERFACE "-m32")
-
-    # When building for 32bits x86, gcc defaults to using the old 8087 float arithmetic
-    # which causes some issues with literal float comparisons. So we set it
-    # to use the SSE2 float path instead
-    # (clang defaults to use SSE, but, setting this option for it is safe)
-    check_set_compiler_property(APPEND PROPERTY fpsse2 "SHELL:-msse2 -mfpmath=sse")
-    zephyr_compile_options($<TARGET_PROPERTY:compiler,fpsse2>)
-    target_compile_options(native_simulator INTERFACE "$<TARGET_PROPERTY:compiler,fpsse2>")
+    set(TARGET_HOST ${CMAKE_HOST_SYSTEM_PROCESSOR})
   endif()
+endif()
+
+if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${TARGET_HOST}.cmake)
+  # Set necessary compiler & linker options for this specific host architecture
+  include(${TARGET_HOST}.cmake)
+elseif(DEFINED NATIVE_TARGET_HOST)
+  message(WARNING "NATIVE_TARGET_HOST set to ${NATIVE_TARGET_HOST}, but ${CMAKE_CURRENT_LIST_DIR}/"
+          "${TARGET_HOST}.cmake not found. No custom target options will be applied."
+  )
 endif()
 
 zephyr_compile_options(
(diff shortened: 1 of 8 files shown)


boards/native/native_sim/doc/cross_compile.rst
```diff
@@ -0,0 +1,100 @@
...
+.. note::
+
+   This instructions guide you on how to set an Ubuntu 24.04 x86_64 development machine for
+   cross-compiling for 32bit (``arm`` or ``armhf``) and 64bit (``aarch64``) ARM targets.
+   And with that setup how to cross-compile for such a target.
+   But there is different ways to setup your cross-compiling environment, and you may have a
+   different host or target architecture. Therefore you can treat this guide as inspiration.
+
+The first thing you need, is to have a cross-compiling toolchain in your development computer.
+On an Ubuntu 24.04 x86_64 you can install the ``gcc-aarch64-linux-gnu`` package for 64bit ARM
```
carlescufi, line 29: Should we mention the Zephyr SDK?

aescolar: Not really, we need a Linux ABI toolchain (i.e. that will produce a binary that runs in Linux, not in a raw embedded target)

cmake/modules/FindHostTools.cmake
```diff
@@ -73,7 +73,8 @@ endif()
 # Default to the host system's toolchain if we are targeting a host based target
 if((${BOARD_DIR} MATCHES "boards\/native") OR ("${ARCH}" STREQUAL "posix")
    OR ("${BOARD}" STREQUAL "unit_testing"))
-  if(NOT "${ZEPHYR_TOOLCHAIN_VARIANT}" STREQUAL "llvm")
+  if((NOT "${ZEPHYR_TOOLCHAIN_VARIANT}" STREQUAL "llvm") AND
+     (NOT "${ZEPHYR_TOOLCHAIN_VARIANT}" STREQUAL "cross-compile"))
```
tejlmand: this could come as a surprise to user using `cross-compile` for target but relies on current behavior for native sim or unit test builds.

LLVM was originally excluded because same toolchain can be used for host builds as well as cross-compiling, and I see the docs now describe using `cross-compile` to not only cross-compile for a target board board but also native board, so I see the reasons, but still concerned for users who might rely on current cross compile behavior.

arch/posix/CMakeLists.txt
```diff
@@ -39,34 +39,27 @@ set_property(TARGET native_simulator PROPERTY LOCALIZE_EXTRA_OPTIONS "")
...
-    target_compile_options(native_simulator INTERFACE "$<TARGET_PROPERTY:compiler,fpsse2>")
+    set(TARGET_HOST ${CMAKE_HOST_SYSTEM_PROCESSOR})
   endif()
 endif()
 
+if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${TARGET_HOST}.cmake)
+  # Set necessary compiler & linker options for this specific host architecture
+  include(${TARGET_HOST}.cmake)
+else()
+  message(WARNING "${CMAKE_CURRENT_LIST_DIR}/${TARGET_HOST}.cmake not found")
```
tejlmand: when printing a warning then it should serve some form of purpose for which the user can take action.

Like changing a setting, fixing code, etc. or informing something is not supported (perhaps an eroor is better in such case) but in this case it's unclear what the user can actually do.

Is the user expected to create the file, and which what content ?
Or is this really something which is not supported --> is an error better ?

tejlmand (commented):
Generally looks good, but some small observations.

cmake/modules/FindHostTools.cmake
aescolar: Would you prefer me to change it in some way?
If you want i can make the check for cross-compile conditional to NATIVE_TARGET_HOST having been set. That should prevent any change for current users.

arch/posix/CMakeLists.txt
aescolar: There are two cases I expected:
* Somebody wrote something like `NATIVE_TARGET_HOST=arch64` (instead of `aarch64`) for which an error is indeed more appropriate.
* Somebody wants to build for another target which does not require any special options and they don't want to bother adding a file that is just empty (Like for `arm` which did not really need any special treatment: You can see that `arm.cmake` is just handholding the user a bit)
So I would not want to prevent a user from going ahead if they know what they are doing, but at the same time, it is not the most normal case so I think it is worthwhile displaying a warning.

Anyhow, your input is welcome. You tell if you want it otherwise :)

cmake/modules/FindHostTools.cmake
aescolar: I changed, so it will stop overriding `cross-compile` only when also `NATIVE_TARGET_HOST` has been set. 
So if some user relies on having `ZEPHYR_TOOLCHAIN_VARIANT=cross-compile` set all the time for their real target, and having this logic correct it for native_sim/unit_tests, it will continue to do so.

arch/posix/CMakeLists.txt
tejlmand: I was just trying to understand what I could use this warning for myself.
How it would help me as developer.

Your reply helped me notice that you actually make it possible for users to manually specify `NATIVE_HOST`.
Did related that at first to this code :+1: 

Let me try to see if I can give a proposal.

cmake/modules/FindHostTools.cmake
tejlmand: I think whoever relies on current behavior will like that.

arch/posix/CMakeLists.txt
tejlmand: how about:
```
if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${TARGET_HOST}.cmake)
  # Set necessary compiler & linker options for this specific host architecture
  include(${TARGET_HOST}.cmake)
elseif(DEFINED NATIVE_TARGET_HOST)
  message(WARNING "NATIVE_TARGET_HOST set to ${NATIVE_TARGET_HOST}, but ${CMAKE_CURRENT_LIST_DIR}/${TARGET_HOST}.cmake not found."
    "No custom target options will be applied."
  )
endif()
```

Tried to avoid indicating a spelling mistake as that is just annoying to be told if spelling is correct.
Also I wanted to avoid indicating that there is no support, so that's how I ended with the option of just informing file is not found and no custom target options are applied :laughing: 

The last case, of a `CMAKE_HOST_SYSTEM_PROCESSOR` not matching know target is just taken as a platform not needing special target adjustments (until we're informed otherwise), hence no warning in this case.

Note, remember to check line length.

aescolar: Applied as is (with line wrap :) )

tejlmand (approved):

carlescufi (approved):

tpambor (approved):

PR #100187, opened 2025-11-27
drivers: uart: save individual uart instances interrupts in PM Action

### LINKED ISSUE #101151: drivers: serial: NXP uart_mcux_flexcomm: instance interrupt config not saved during PM_DEVICE_ACTION_TURN_OFF
### Describe the bug

The PM action saves the interrupts enabled using a global variable called usart_intenset. The problem is when you have multiple uart instances that have different configurations, when entering/exiting from a power level, only one configuration is saved and will be applied to all uart instances when exiting from a power level. We need to keep this setting individual for each instance. 

### Regression

- [ ] This is a regression.

### Steps to reproduce

1. Enable multiple UART instances
2. Select `CONFIG_PM=y` and enabled the `standby` node for PM3 mode
3. Force the app to enter the Idle state for at least 351ms so the PM policy will enter Standby (PM3) mode

The INTEN register for all UART instances will be saved and then restored using the single `usart_intenset` variable.

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

davidmissael:
The PM action saves the interrupts enabled using a global variable called usart_intenset. The problem is when you have multiple uart instances that have different configurations, when entering/exiting from a power level, only one configuration is saved and will be applied to all uart instances when exiting from a power level. We need to keep this setting individual for each instance. To do this, usart_intenset was added as a new element to mcux_flexcomm_data structure. In this way, each uart instance will keep/restore its own setting.

### DIFF
--- a/drivers/serial/uart_mcux_flexcomm.c
+++ b/drivers/serial/uart_mcux_flexcomm.c
@@ -89,6 +89,7 @@ struct mcux_flexcomm_data {
 #ifdef CONFIG_UART_INTERRUPT_DRIVEN
 	uart_irq_callback_user_data_t irq_callback;
 	void *irq_cb_data;
+	uint32_t usart_intenset;
 #endif
 #ifdef CONFIG_UART_ASYNC_API
 	uart_callback_t async_callback;
@@ -1205,10 +1206,12 @@ static void mcux_flexcomm_pm_restore_wake(const struct device *dev,
 }
 #endif /* FC_UART_IS_WAKEUP */
 
-static uint32_t usart_intenset;
 static int mcux_flexcomm_pm_action(const struct device *dev, enum pm_device_action action)
 {
+#ifdef CONFIG_UART_INTERRUPT_DRIVEN
 	const struct mcux_flexcomm_config *config = dev->config;
+	struct mcux_flexcomm_data *data = dev->data;
+#endif
 	int ret;
 
 	switch (action) {
@@ -1217,14 +1220,18 @@ static int mcux_flexcomm_pm_action(const struct device *dev, enum pm_device_acti
 	case PM_DEVICE_ACTION_SUSPEND:
 		break;
 	case PM_DEVICE_ACTION_TURN_OFF:
-		usart_intenset = USART_GetEnabledInterrupts(config->base);
+#ifdef CONFIG_UART_INTERRUPT_DRIVEN
+		data->usart_intenset = USART_GetEnabledInterrupts(config->base);
+#endif
 		break;
 	case PM_DEVICE_ACTION_TURN_ON:
 		ret = mcux_flexcomm_init_common(dev);
 		if (ret) {
 			return ret;
 		}
-		USART_EnableInterrupts(config->base, usart_intenset);
+#ifdef CONFIG_UART_INTERRUPT_DRIVEN
+		USART_EnableInterrupts(config->base, data->usart_intenset);
+#endif
 		break;
 	default:
 		return -ENOTSUP;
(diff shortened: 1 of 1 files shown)


drivers/serial/uart_mcux_flexcomm.c
```diff
@@ -1217,14 +1220,18 @@ static int mcux_flexcomm_pm_action(const struct device *dev, enum pm_device_acti
 	case PM_DEVICE_ACTION_SUSPEND:
 		break;
 	case PM_DEVICE_ACTION_TURN_OFF:
-		usart_intenset = USART_GetEnabledInterrupts(config->base);
+#ifdef CONFIG_UART_INTERRUPT_DRIVEN
+		data->usart_intenset = USART_GetEnabledInterrupts(config->base);
```
Albort12138, line 1221: Could you share your use case? In what scenario would PM_DEVICE_ACTION_TURN_OFF be executed? Why, during TURN_OFF, do you only save the enabled interrupts instead of performing any operations related to turning off?

davidmissael: Hi @Albort12138 PM_DEVICE_ACTION_TURN_OFF is being executed when the device enters in a lowpower mode. The specific case for the RW612, when entering in PM3, the microcontoller does not have registers retention (only SRAM), so any configuration at any peripheral will be lost, that's why this configuration has to be saved during PM_DEVICE_ACTION_TURN_OFF and restored at PM_DEVICE_ACTION_TURN_ON. This is the only operation we need to save in this action to recover the uart while exiting from low power mode. **If you take a look on the driver that is currently  implemented** this is the only operation performed, actually I'm not modifying that logic, this PR just tries to save this context individually per each uart instance defined in the application instead of having one single global variable that retains that configuration and restores all the other ones. We need one context per instance.
@dleach02 @decsny @mmahadevan108 since you are the driver's maintainers, is there anything else we can add to the @Albort12138's question?

Albort12138: Hi @davidmissael, thanks your feedback! If I understand correctly, you're calling pm_device_action_run() to turn off the USART device in your use case?

zejiang0jason: Hi @Albort12138 , RW610 has a `peripheral_domain`, for the peripherals assigned to this domain, PM_DEVICE_ACTION_TURN_OFF will be called when the system enters standby mode:
https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/nxp/nxp_rw6xx_common.dtsi#L87-L91
https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/power_domain/power_domain_soc_state_change.c#L68-L76.

flexcom is added in this domain.

zejiang0jason (dismissed):

drivers/serial/uart_mcux_flexcomm.c
decsny: turn off action happens when the power domain the device is on gets suspended. read doc here https://docs.zephyrproject.org/latest/services/pm/power_domain.html

drivers/serial/uart_mcux_flexcomm.c
```diff
@@ -1205,10 +1206,12 @@ static void mcux_flexcomm_pm_restore_wake(const struct device *dev,
 }
 #endif /* FC_UART_IS_WAKEUP */
 
-static uint32_t usart_intenset;
 static int mcux_flexcomm_pm_action(const struct device *dev, enum pm_device_action action)
 {
+#ifdef CONFIG_UART_INTERRUPT_DRIVEN
```
decsny: I don't understand why you need to chop up the code conditionally based on if this flag is set in the build. Wouldn't you want to restore interrupts in any case?

decsny (dismissed):

drivers/serial/uart_mcux_flexcomm.c
davidmissael: Hi @decsny yes I added this change within the conditions to comply on how the structure "mcux_flexcomm_data" is defined since all its elements depends on compile conditions. I can just add the element in the structure without depending an any conditions as you mention. Thanks for the feedback

zejiang0jason (approved):

mmahadevan108 (approved):

PR #100190, opened 2025-11-28
soc: xlnx: zynqmp_apu: Add basic SoC support for Zynq MPSoC APU

manataji:
Summary

Add initial SoC support for 
the Zynq UltraScale+ MPSoC APU (Cortex-A53), including:
Kconfig, Kconfig.soc, Kconfig.defconfig
CMakeLists.txt
mmu_regions.c
soc.yml

Rationale
ARM Generic Timer base frequency is not described in DT; 
it is provided by firmware via CNTFRQ_EL0.

Testing
twister build-only passes for kv260_a53 hello_world (board PR pending).
docs build unaffected (board docs will be added in the board PR).

Notes
This PR is a prerequisite for the kv260_a53 board support.
No DTS or driver changes here; board PR will add DTS and docs.

Signed-off-by: Manabu Tajima mana.taj@gmail.com

PR #100192, opened 2025-11-28
boards: amd: kv260_a53: Add KV260 A53 board definition

manataji:
Summary

Add KV260 A53 board definition under boards/amd/kv260_a53:
DTS: PS UART1 console
Kconfig.board, board.cmake
kv260_a53.yaml (includes documentation: doc/index.rst)
Documentation with build/run instructions

Testing

hello_world builds and runs via U-Boot (console at 115200).
twister build-only for hello_world succeeds.
docs build succeeds (htmldocs).
Depends-on

Depends-on: zephyrproject-rtos/zephyr#100190

Notes

SMP disabled by default.
ARM Generic Timer used; base frequency matches CNTFRQ_EL0.

Signed-off-by: Manabu Tajima mana.taj@gmail.com

### DIFF
--- a/boards/amd/kv260_a53/Kconfig.kv260_a53
+++ b/boards/amd/kv260_a53/Kconfig.kv260_a53
@@ -1,0 +1,7 @@
+# Copyright (c) 2019 Carlo Caione <ccaione@baylibre.com>
+# SPDX-License-Identifier: Apache-2.0
+
+config BOARD_KV260_A53
+        bool "AMD KV260 Cortex-A53" 
+	select SOC_XILINX_ZYNQMP_APU
+	default y
--- a/boards/amd/kv260_a53/board.cmake
+++ b/boards/amd/kv260_a53/board.cmake
@@ -1,0 +1,3 @@
+# Copyright (c) 2022 Linaro.
+#
+# SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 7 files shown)


michalsimek (changes requested):
There is no reason to introduce ZynqMP APU configuration based on kv260 because there should be generic APU support for ZynqMP where kv260 can be one specific board under it. 
Even current kv260 RPU should be IMHO redesign to fall to the same category as we are adding support for other SOCs (look at versal2_apu, versalnet_apu). It means this one should be zynqmp_apu. And kv260_r5 should be renamed to zynqmp_rpu (with kv260_r5 in it - likely with changed name).

manataji:
Thank you very much for your humble opinion. You mentioned considering a redesign of board/amd/kv260_r5, so does that mean that a design change beyond just changing the KV260_R5 name defined in the folder and file might be necessary?

boards/amd/kv260_a53/Kconfig.kv260_a53
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2019 Carlo Caione <ccaione@baylibre.com>
+# SPDX-License-Identifier: Apache-2.0
+
+config BOARD_KV260_A53
+        bool "AMD KV260 Cortex-A53"
```
nordicjm, line 5: hwmv1 is not supported

nordicjm (changes requested):

boards/amd/kv260_a53/Kconfig.kv260_a53
manataji: Thank you very much for checking. I understand what you pointed out.

PR #100193, opened 2025-11-28
doc: build: dts-vs-kconfig: Add information about generated symbols

### LINKED ISSUE #88460: No documentation for DT_HAS_[compatible]_ENABLED
In some Kconfig files (for example: "zephyr/drivers/lora/Kconfig.sx12xx") there are line that look like this:
```
depends on DT_HAS_SEMTECH_SX1272_ENABLED || DT_HAS_SEMTECH_SX1276_ENABLED
```
It isn't clear how the symbol DT_HAS_SEMTECH_SX1262_ENABLED was constructed.
I assume it depends on a line like 'compatible = "semtech, sx1262";' existing.

This isn't documented anywhere and should be.

pdgendt:
Devicetree bindings generate Kconfig variables, and enabled nodes additionally generate symbols. Add a section to the docs that describes this.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/88460

https://builds.zephyrproject.io/zephyr/pr/100193/docs/build/dts/dt-vs-kconfig.html#automatic-kconfig-symbols-from-devicetree

### DIFF
--- a/doc/build/dts/dt-vs-kconfig.rst
+++ b/doc/build/dts/dt-vs-kconfig.rst
@@ -69,3 +69,45 @@ There are **exceptions** to these rules:
 * Devicetree's ``chosen`` keyword, which allows the user to select a specific
   instance of a hardware device to be used for a particular purpose. An example
   of this is selecting a particular UART for use as the system's console.
+
+Automatic Kconfig symbols from devicetree
+*****************************************
+
+During the devicetree processing step, CMake runs
+:zephyr_file:`scripts/dts/gen_driver_kconfig_dts.py`, which scans all
+:ref:`DTS root <dts_root>` directories, including :zephyr_file:`dts/bindings` and
+writes ``Kconfig.dts`` into the build's ``KCONFIG_BINARY_DIR`` (for example
+``<build>/zephyr`` or ``<build>/<image>/zephyr`` in case of :ref:`sysbuild`).
+For each ``compatible = "vendor,chip"`` that appears in a binding, the generated
+file contains:
+
+.. code-block:: kconfig
+
+   DT_COMPAT_VENDOR_CHIP := vendor,chip
+
+   config DT_HAS_VENDOR_CHIP_ENABLED
+           def_bool $(dt_compat_enabled,$(DT_COMPAT_VENDOR_CHIP))
+
+The assignment keeps the literal ``compatible`` string available to the
+preprocessor functions described in :ref:`kconfig-functions`, so that Kconfig
+files can call helpers such as ``$(dt_compat_on_bus,$(DT_COMPAT_<compatible>),i2c)``.
+Characters like ``-``, ``,`` and ``@`` are converted to underscores when the
+symbol names are created.
+
+The hidden boolean symbol becomes ``CONFIG_DT_HAS_<compatible>_ENABLED`` after
+Kconfig runs. Its value tracks whether the current devicetree contains at least
+one node with that ``compatible`` whose :ref:`status <dt-important-props>` is
+``okay``. Symbols for enabling drivers should almost certainly be ``default y``
+and have ``depends on CONFIG_DT_HAS_<compatible>_ENABLED``, for example:
+
+.. code-block:: kconfig
+
+   config SENSOR_VENDOR_CHIP
+           bool "Vendor Chip sensor"
+           default y
+           depends on DT_HAS_VENDOR_CHIP_ENABLED
+
+Because these symbols are generated automatically, adding a new binding with
+a ``compatible`` property is all that is required to make the corresponding
+``DT_HAS_<compatible>_ENABLED`` and ``DT_COMPAT_<compatible>`` constructs
+available to Kconfig.


doc/build/dts/dt-vs-kconfig.rst
```diff
@@ -69,3 +69,42 @@ There are **exceptions** to these rules:
 * Devicetree's ``chosen`` keyword, which allows the user to select a specific
   instance of a hardware device to be used for a particular purpose. An example
   of this is selecting a particular UART for use as the system's console.
+
+Automatic Kconfig symbols from devicetree
+*****************************************
+
+During the devicetree processing step CMake runs
```
JarmouniA:
```suggestion
During the devicetree processing step, CMake runs
```

JarmouniA (dismissed):

doc/build/dts/dt-vs-kconfig.rst
rruuaanng, line 1: Would it be more appropriate to use `DT_COMPAT_<compatible>` instead of `DT_COMPAT_VENDOR_CHIP`? I think using pseudocode to describe it might make things cleare. For the other generated `CONFIG_DT_HAS_*_ENABLED` definitions, I think it's also possible to use some conventional wildcards to describe them. WDYT :)

pdgendt: Which occurrence would you change, could you add suggestions?

doc/build/dts/dt-vs-kconfig.rst
```diff
@@ -69,3 +69,42 @@ There are **exceptions** to these rules:
...
+
+.. code-block:: kconfig
+
+   config SENSOR_VENDOR_CHIP
+           bool "Vendor Chip sensor"
+           depends on DT_HAS_VENDOR_CHIP_ENABLED
+
+Because these symbols are generated automatically, adding a new binding is all
+that is required to make the corresponding ``DT_HAS_*_ENABLED`` and ``DT_COMPAT_*``
+constructs available to Kconfig.
```
rruuaanng:
```suggestion
The assignment keeps the literal ``compatible`` string available to the
preprocessor functions described in :ref:`kconfig-functions`, so that Kconfig
files can call helpers such as ``$(dt_compat_on_bus,$(DT_COMPAT_<compatible>),i2c)``.
Characters like ``-``, ``,`` and ``@`` are converted to underscores when the
symbol names are created, which is why a Kconfig option would reference the
generated identifier as ``DT_HAS_<compatible>_ENABLED``.

The hidden boolean symbol becomes ``CONFIG_DT_HAS_<compatible>_ENABLED`` after
Kconfig runs. Its value tracks whether the current devicetree contains at least
one node with that ``compatible`` whose :ref:`status <dt-important-props>` is
``okay``. Driver menus commonly use the symbol to guard options that only make
sense when the hardware is present, for example:

.. code-block:: kconfig

   config SENSOR_VENDOR_CHIP
           bool "Vendor Chip sensor"
           depends on DT_HAS_VENDOR_CHIP_ENABLED

Because these symbols are generated automatically, adding a new binding is all
that is required to make the corresponding ``DT_HAS_<compatible>_ENABLED`` and ``DT_COMPAT_<compatible>``
constructs available to Kconfig.
```

JarmouniA (dismissed):

rruuaanng (dismissed):

doc/build/dts/dt-vs-kconfig.rst
```diff
@@ -69,3 +69,41 @@ There are **exceptions** to these rules:
 * Devicetree's ``chosen`` keyword, which allows the user to select a specific
   instance of a hardware device to be used for a particular purpose. An example
   of this is selecting a particular UART for use as the system's console.
+
+Automatic Kconfig symbols from devicetree
+*****************************************
+
+During the devicetree processing step CMake runs
+:zephyr_file:`scripts/dts/gen_driver_kconfig_dts.py`, which scans every binding
+under :file:`dts/bindings` and writes ``Kconfig.dts`` into the build's
```
nordicjm: should probably mention it's in dts roots, including the zephyr one, since modules can add them and users on the command line

doc/build/dts/dt-vs-kconfig.rst
```diff
@@ -69,3 +69,41 @@ There are **exceptions** to these rules:
...
+``KCONFIG_BINARY_DIR`` (usually ``build/zephyr``). For each ``compatible =
```
nordicjm: `<build>/zephyr` or `<build>/<image>/zephyr` in case of sysbuild, should not say that the folder is called build because that only applies when using west, and when using west without options

mbolivar (changes requested):
This is good overview, but I think that it obscures the take-home message, which is:

"your driver should almost certainly be `default y` and have `depends on DT_COMPAT_<YOURCOMPATIBLE>_ENABLED`"

The description of what DT_COMPAT_..._ENABLED is and how it works are nice to know but they're ultimately implementation details that aren't as important as the take-home message.

pdgendt:
> This is good overview, but I think that it obscures the take-home message, which is:
> 
> "your driver should almost certainly be `default y` and have `depends on DT_COMPAT_<YOURCOMPATIBLE>_ENABLED`"
> 
> The description of what DT_COMPAT_..._ENABLED is and how it works are nice to know but they're ultimately implementation details that aren't as important as the take-home message.

Updated the example with your suggestion, thanks!

keith-zephyr (dismissed):

fabiobaltieri (dismissed):

pdgendt:
Ping @mbolivar PTAL

doc/build/dts/dt-vs-kconfig.rst
```diff
@@ -69,3 +69,44 @@ There are **exceptions** to these rules:
...
+and have ``depends on CONFIG_DT_HAS_<compatible>_ENABLED``, for example:
+
+.. code-block:: kconfig
+
+   config SENSOR_VENDOR_CHIP
+           bool "Vendor Chip sensor"
+           default y
+           depends on DT_HAS_VENDOR_CHIP_ENABLED
+
+Because these symbols are generated automatically, adding a new binding is all
```
JarmouniA:
```suggestion
Because these symbols are generated automatically, adding a new binding with ``compatible`` property is all
```

JarmouniA (commented):
Note that the script appears to be ignoring child-bindings
(like https://github.com/zephyrproject-rtos/zephyr/blob/5a58989daddda2a8a3c87396acaf4275b6c72596/dts/bindings/i2c/ti%2Ctca9544a.yaml#L10),
so `DT_COMPAT_{CHILD-COMPAT}` or `DT_HAS_{CHILD-COMPAT}_ENABLED` aren't generated.

mbolivar (approved):

rruuaanng (approved):

JarmouniA (approved):

doc/build/dts/dt-vs-kconfig.rst
mathieuchopstm, line 1: You may want to update the following places to refer to this new section/replace them:

https://github.com/zephyrproject-rtos/zephyr/blob/4e62f0510c93b72de3a62f1795facf6cf048b63b/doc/contribute/style/kconfig.rst?plain=1#L69-L71

https://github.com/zephyrproject-rtos/zephyr/blob/4e62f0510c93b72de3a62f1795facf6cf048b63b/doc/build/kconfig/tips.rst?plain=1#L631-L635

pdgendt: Opened https://github.com/zephyrproject-rtos/zephyr/pull/102985

PR #100194, opened 2025-11-28
drivers: led: pca9533: fix ms_to_psc wrap-around

jeppenodgaard:
If `period_ms` is 3 or less `tmp` is 0. Since `tmp` is type `uint32` subtracting 1 from 0 will wrap around to `UINT32_MAX` and is then clamped to `UINT8_MAX`.

Fix the issue by changing `tmp` and `1` to signed types.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/100020

### DIFF
--- a/drivers/led/pca9533.c
+++ b/drivers/led/pca9533.c
@@ -65,9 +65,9 @@ struct pca9533_data {
  */
 static uint8_t ms_to_psc(uint32_t period_ms)
 {
-	uint32_t tmp = (period_ms * 152U + 500U) / 1000U;
+	int32_t tmp = (period_ms * 152U + 500U) / 1000U;
 
-	return CLAMP(tmp - 1U, 0U, UINT8_MAX);
+	return CLAMP(tmp - 1, 0U, UINT8_MAX);
 }
 
 /**


fabiobaltieri (approved):

kartben (approved):

simonguinot (approved):
Thanks :)

drivers/led/pca9533.c
```diff
@@ -65,9 +65,9 @@ struct pca9533_data {
  */
 static uint8_t ms_to_psc(uint32_t period_ms)
 {
-	uint32_t tmp = (period_ms * 152U + 500U) / 1000U;
+	int32_t tmp = (period_ms * 152U + 500U) / 1000U;
```
thedjnK, line 68: should probably have
```
	int32_t tmp = (int32_t)((period_ms * 152U + 500U) / 1000U);
```
or will get a future warning on type assignment

thedjnK (approved):

drivers/led/pca9533.c
jeppenodgaard: I am not familiar with the warning.
Can you share a link please?

jeppenodgaard: I am not familiar with the warning.
Can you share a link please?

PR #100197, opened 2025-11-28
llext: adding missing parentheses on macro INSTR_FETCHABLE

andreagilardoni:
C macro INSTR_FETCHABLE is missing parentheses in its definition, this could lead to incorrect expansion when combined with logical operators. For instance we can have incorrect behaviors at this line: https://github.com/zephyrproject-rtos/zephyr/blob/f908d0b5f22adaaf912e729317c323864f1e1641/subsys/llext/llext_mem.c#L147


pillo79 (approved):
Thanks for spotting this!

vlad-kulikov (approved):

laurenmurphyx64 (approved):
Thank you!

teburd (approved):

PR #100203, opened 2025-11-28
gpio: mcux_rgpio: Support IRQ output select and secure access checks for i.MX95 M7

PetervdPerk-NXP:
Support for selecting the IRQ output via the `irq-output-select` property in devicetree. Updates interrupt configuration and ISR logic to use the selected IRQ index. For i.MX95 M7, ensures pins and IRQs are configured only when secure access is allowed by checking PCNS and ICNS registers.

### DIFF
--- a/drivers/gpio/gpio_mcux_rgpio.c
+++ b/drivers/gpio/gpio_mcux_rgpio.c
@@ -35,6 +35,7 @@ struct mcux_rgpio_config {
 
 	const struct pinctrl_soc_pinmux *pin_muxes;
 	uint8_t mux_count;
+	uint8_t irq_sel;
 };
 
 struct mcux_rgpio_data {
@@ -144,6 +145,14 @@ static int mcux_rgpio_configure(const struct device *dev,
 	}
 #endif
 
+#if !defined(__ARM_FEATURE_CMSE)
+	base->PCNS &= ~BIT(pin);
+	if (base->PCNS & BIT(pin)) {
+		/* We don't have access to this pin */
+		return -ENOTSUP;
+	}
+#endif
+
 	memcpy(&pin_cfg.pinmux, &config->pin_muxes[cfg_idx], sizeof(pin_cfg));
 	/* cfg register will be set by pinctrl_configure_pins */
 	pin_cfg.pin_ctrl_flags = reg;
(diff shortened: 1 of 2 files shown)


bperseghetti (dismissed):

drivers/gpio/gpio_mcux_rgpio.c
```diff
@@ -144,6 +145,14 @@ static int mcux_rgpio_configure(const struct device *dev,
 	}
 #endif
 
+#if defined(CONFIG_SOC_MIMX9596_M7)
```
zejiang0jason: Let me test whether RT1180 CM7 also need this, if so, suggest add this for all ARM cores which doesn't support secure/non-secure mode.

zejiang0jason: I have tried RT1180 CM7, the behavior is the similar. So for the architecture without secure/non-secure feature, the bit should be cleared. We can use:
```suggestion
#if !defined(__ARM_FEATURE_CMSE)
```
https://arm-software.github.io/acle/cmse/cmse.html#feature-macro

bperseghetti (approved):

Holt-Sun (approved):

zejiang0jason (approved):

mmahadevan108 (approved):

SimonBrix:
Yes, I think that a revert to `defined(CONFIG_SOC_MIMX9596_M7)` as the original PR proposed, is needed, to not cause problems with controlling GPIO from a non-secure context.

Or is it possible to check if it is running in non-secure mode and `PCNS[pin] == 0` and also make that return `-ENOTSUP`, but otherwise continue in case it is in non-secure mode and `PCNS[pin] == 1`, which means that it is able to control the GPIO from non-secure context?

zejiang0jason:
from https://github.com/zephyrproject-rtos/zephyr/issues/117830, seems this code doesn't work for IMX95 CM7 either, which means `defined(CONFIG_SOC_MIMX9596_M7)` is not right condition.

What I can confirm is, on RT1180 CM7, the pin bit in `PCNS` should be cleared.

update:
I test IMX95 with MCUX SDK baremetal code, the `PCNS`  should be cleared too, otherwise CM7 can't change RGPIO pin registers.

PR #100209, opened 2025-11-28
drivers: entropy: gecko_trng: Fully initialize on read

asmellby:
If the TRNG has been unretained, it's not sufficient to enable the clock, it must be fully reinitialized.

This is a minimal fix to ensure entropy is accumulated after sleep. Long term, this driver should be made to use device power management with power domain awareness.

### DIFF
--- a/drivers/entropy/entropy_gecko_trng.c
+++ b/drivers/entropy/entropy_gecko_trng.c
@@ -44,6 +44,25 @@
 #endif	/* _SILICON_LABS_32B_SERIES_2_CONFIG_* */
 #endif	/* CONFIG_CRYPTO_ACC_GECKO_TRNG */
 
+static int entropy_gecko_trng_init(const struct device *dev)
+{
+	/* Enable the TRNG0 clock. */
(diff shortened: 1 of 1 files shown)


drivers/entropy/entropy_gecko_trng.c
```diff
@@ -44,6 +44,8 @@
 #endif	/* _SILICON_LABS_32B_SERIES_2_CONFIG_* */
 #endif	/* CONFIG_CRYPTO_ACC_GECKO_TRNG */
 
+static int entropy_gecko_trng_init(const struct device *dev);
```
jhedberg: Did you consider just moving the init function higher up in the c-file to avoid having to do a forward declaration?

asmellby: I considered it and did this instead to minimize the diff. I could move it if that's preferred.

asmellby: Done.

jhedberg (approved):

Martinhoff-maker (approved):

jerome-pouiller (approved):

Issue #84303, opened 2025-01-21
Bluetooth: Controller: ISO: ISO receive lost when using framed


Thalley:
**Describe the bug**
In a handful of cases we see that connected ISO SDUs are not received when using framing. 

This is evident by some of the LE Audio tests using `441_1_1` and `441_2_1`:

![Image](https://github.com/user-attachments/assets/692ce71b-55f0-47c3-b4b9-d7c1ddfddce3)

**To Reproduce**
1. Apply https://github.com/zephyrproject-rtos/zephyr/pull/83539
2. Run https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/bsim/bluetooth/audio/compile.sh
3. Run `Execute_AC_1 441_1_1`

**Expected behavior**
Expect the controller to properly handle TX and RX when using framing. 

**Impact**
A subset of ISO/LE Audio won't work

**Logs and console output**
```
d_01: @00:00:03.945466  Waiting for data
d_01: @00:00:03.961964  [00:00:03.961,944] <inf> bap_stream_rx: [0|0]: Incoming audio on stream 0x8502100 len 0, flags 0x0C, seq_num 0 and ts 3967732
d_01: @00:00:03.961964  [00:00:03.961,944] <inf> bap_stream_rx: [0|1]: Incoming audio on stream 0x8502100 len 0, flags 0x0C, seq_num 0 and ts 3967732
d_00: @00:00:03.974931  [00:00:03.974,914] <inf> bap_stream_tx: Stream 0x8502220 sent 0 SDUs
d_01: @00:00:03.976963  [00:00:03.976,959] <inf> bap_stream_rx: [1|3]: Incoming audio on stream 0x8502100 len 0, flags 0x0C, seq_num 2 and ts 3975109
d_01: @00:00:03.976963 ERROR: (WEST_TOPDIR/zephyr/tests/bsim/bluetooth/audio/src/bap_stream_rx.c:66): ISO receive lost
```

**Environment (please complete the following information):**
 - Commit SHA or Version used: https://github.com/zephyrproject-rtos/zephyr/pull/83539

**Additional context**
May relate to https://github.com/zephyrproject-rtos/zephyr/issues/83585

Thalley:
The following are still failing with `main` from today: 

cap_ac_2_441_1_1
cap_ac_2_441_2_1
cap_ac_5_441_2_1
cap_ac_10_441_1_1
cap_ac_10_441_2_1

[...]

PR #100210, opened 2025-11-28
tests: Bluetooth: Audio: Reenable some diabled tests that now pass

Thalley:
Enabled some tests that previously failed, but were not reenable when the fix was applied.

### DIFF
--- a/tests/bsim/bluetooth/audio/test_scripts/audio_config_tests/cap_ac_1_441_1_1.sh
+++ b/tests/bsim/bluetooth/audio/test_scripts/audio_config_tests/cap_ac_1_441_1_1.sh
@@ -4,6 +4,4 @@
 
 source $(dirname "$0")/_ac_common.sh
 
-# bap_stream_rx.c:66): ISO receive lost
-# https://github.com/zephyrproject-rtos/zephyr/issues/84303
-# ac_config=1 ac_tx_preset=441_1_1 ac_acc_cnt=1 Execute_cap_unicast_ac $@
+ac_config=1 ac_tx_preset=441_1_1 ac_acc_cnt=1 Execute_cap_unicast_ac $@
--- a/tests/bsim/bluetooth/audio/test_scripts/audio_config_tests/cap_ac_12_441_1_2.sh
+++ b/tests/bsim/bluetooth/audio/test_scripts/audio_config_tests/cap_ac_12_441_1_2.sh
@@ -4,5 +4,4 @@
 
 source $(dirname "$0")/_ac_common.sh
 
-# BT_ISO_FLAGS_LOST
-# ac_config=12 ac_tx_preset=441_1_2 ac_acc_cnt=1 Execute_cap_broadcast_ac $@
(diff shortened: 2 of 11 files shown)


cvinayak (approved):
Just thinking aloud, as I did not fix anything related to these tests passing. I suspect they may fail in the future. If I can help then, it may enlighten me of any cases the Controller has not taken care. Till then, so be it. Approved.

Thalley:
> Just thinking aloud, as I did not fix anything related to these tests passing. I suspect they may fail in the future. If I can help then, it may enlighten me of any cases the Controller has not taken care. Till then, so be it. Approved.

Good point to raise. Some of the previous failures may have been related to other changes and bug fixes in the controller, or simply fixed due to timing changes.

tests/bsim/bluetooth/audio/test_scripts/audio_config_tests/cap_ac_1_441_1_1.sh
```diff
@@ -4,6 +4,4 @@
 
 source $(dirname "$0")/_ac_common.sh
 
-# bap_stream_rx.c:66): ISO receive lost
-# https://github.com/zephyrproject-rtos/zephyr/issues/84303
```
aescolar, line 8: You mentioned in these tickets that this is still an issue..(?)
You saw that thru some other tests? or testing in some other way?
(I'm just wondering if we are going to see these tests fail again in a couple of weeks when the timing changes a bit)

Thalley: That is a fair concern. The issue is still open as some tests are still failing due to the issue, and we may indeed run into an issue in the future where the tests in this PR (or any of the other CAP_AC tests) will fail due to timing changes. 

On the other hand, it's misleading to keep it as is, i.e. if someone tries to reproduce https://github.com/zephyrproject-rtos/zephyr/issues/84303 by searching for the failing tests, and those tests are passing, then it's additional steps for the person working on the issue to find the ones that are actually failing. 

@cvinayak If you have additional thoughts or concerns here as well, please let me know

cvinayak: ISO receive lost error continues to be an issue, overlapping (initial overlap or drifting over a period of the test execution) roles are scheduled with some fairness and hence at such cases ISO SDUs will be missed.

Keeping the passing tests enabled will definitely expose timing changes (random number sequence changes, scheduling changes etc..) and possibly expose new symptoms that can aid is identifying latent bugs in current implementation of the Controller.

We can always decide to disable tests in the future after identifying the root cause (could be new symptoms).

kartben (approved):

larsgk (approved):

aescolar (changes requested):
Just to avoid confusion I will leave a clear nack.

Thalley:
Closing until known issues are fixed

PR #100214, opened 2025-11-28
Bluetooth: Controller: nrf54lx: Fix address resolution wait

### LINKED ISSUE #100101: BLE device bricked after disconnect with CONFIG_BT_APP_PASSKEY=y on nRF54L15 (Zephyr 4.3.0)
### Describe the bug

**Summary**
After enabling CONFIG_BT_APP_PASSKEY=y, the device becomes bricked after each BLE disconnect on nRF54L15. Even after reset and BLE initialization, the core enters a bricked state—only erasing firmware helps.
Environment

Zephyr version: 4.3.0
Hardware: Nordic nRF54L15
Configuration: BLE with passkey enabled

**Steps to Reproduce**

Enable `CONFIG_BT_APP_PASSKEY=y `in prj.conf
Build and flash the application
Connect via BLE and disconnect
Observe device becomes unresponsive

**Expected Behavior**
Device should handle BLE disconnections gracefully and remain responsive.
**Actual Behavior**
Device enters a bricked state after disconnect. Debugger shows the core stuck in an infinite loop.
Troubleshooting Attempted

Disabled `CONFIG_BT_SETTINGS=n` → No effect
Disabled `CONFIG_BT_CTLR_PRIVACY=n `→ Resolves the issue (but not ideal)

**Debug Output**
The core is stuck in an infinite loop in 
`zephyr/subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c`
in the radio_ar_has_match() function:

```
uint32_t radio_ar_has_match(void)
{
	if (!radio_bc_has_match()) {
		return 0U;
	}

	nrf_aar_int_enable(NRF_AAR, AAR_INTENSET_END_Msk);

	// Infinite loop - never returns:
	while (NRF_AAR->EVENTS_END == 0U) {
		cpu_sleep();
	}

	nrf_aar_int_disable(NRF_AAR, AAR_INTENCLR_END_Msk);
	NVIC_ClearPendingIRQ(nrfx_get_irq_number(NRF_AAR));

	if (NRF_AAR->EVENTS_RESOLVED && !NRF_AAR->EVENTS_NOTRESOLVED) {
		return 1U;
	}

	return 0U;
}
```

The `NRF_AAR->EVENTS_END` event never gets set, causing the `cpu_sleep()` loop to never exit.
**Current Workaround**
Disable privacy:
properties `CONFIG_BT_CTLR_PRIVACY=n`
This resolves the issue but likely disables an important security feature.

cvinayak:
Fix address resolution wait to not be infinite loop and instead exit with a reasonable count down value.

Fixes #100101

### DIFF
--- a/subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
+++ b/subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
@@ -2693,6 +2693,28 @@ void radio_ar_status_reset(void)
 	hal_radio_nrf_ppi_channels_disable(BIT(HAL_TRIGGER_AAR_PPI));
 }
 
+static void ar_end_wait(void)
+{
+	/* Blocking Address Resolution (AR) wait */
+
+	/* An arbitrary iteration which seems sufficient testing on target with one IRK to resolve.
+	 * We should ideally not be waiting anyway, and in theory be resolved before we are here.
+	 */
+	uint32_t countdown = 200U;
+
+	while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
+		countdown--;
+
+		/* We do not CPU sleep on select target SoCs but spin wait.
+		 * For all other platforms, interrupt on END is enabled by the caller of this
+		 * function, we will be woken up when the EVENTS_END occurs.
+		 */
+		if (!IS_ENABLED(CONFIG_SOC_SERIES_NRF54L)) {
+			cpu_sleep();
+		}
+	}
+}
+
 uint32_t radio_ar_has_match(void)
 {
 	if (!radio_bc_has_match()) {
@@ -2701,9 +2723,7 @@ uint32_t radio_ar_has_match(void)
 
 	nrf_aar_int_enable(NRF_AAR, AAR_INTENSET_END_Msk);
 
-	while (NRF_AAR->EVENTS_END == 0U) {
-		cpu_sleep();
-	}
+	ar_end_wait();
 
 	nrf_aar_int_disable(NRF_AAR, AAR_INTENCLR_END_Msk);
 
@@ -2753,9 +2773,7 @@ uint8_t radio_ar_resolve(const uint8_t *addr)
 
 	nrf_aar_task_trigger(NRF_AAR, NRF_AAR_TASK_START);
 
-	while (NRF_AAR->EVENTS_END == 0) {
-		cpu_sleep();
-	}
+	ar_end_wait();
 
 	nrf_aar_int_disable(NRF_AAR, AAR_INTENCLR_END_Msk);
 


carlescufi (dismissed):

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,23 @@ void radio_ar_status_reset(void)
...
+	 */
+	uint32_t countdown = 200U;
+
+	while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
+		countdown--;
+
+#if !defined(CONFIG_SOC_SERIES_NRF54LX)
+		/* We do not sleep on target SoC. */
+		cpu_sleep();
+#endif /* !CONFIG_SOC_SERIES_NRF54LX */
```
Thalley: IS_ENABLED?

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,23 @@ void radio_ar_status_reset(void)
 	hal_radio_nrf_ppi_channels_disable(BIT(HAL_TRIGGER_AAR_PPI));
 }
 
+static void ar_end_wait(void)
```
Thalley, line 2696: What is `ar`? :)

cvinayak: Address Resolution.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2693,6 +2723,25 @@ void radio_ar_status_reset(void)
...
+	/* An arbitrary iteration which seems sufficient testing on target with one IRK to resolve.
+	 * We should ideally be not be waiting anyway, and in theory be resolved before we are here.
+	 */
+	uint32_t countdown = 200U;
+
+	while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
+		countdown--;
+
+		/* We do not sleep on target SoC. */
+		if (!IS_ENABLED(CONFIG_SOC_SERIES_NRF54LX)) {
```
Thalley: Should this be


```suggestion
	/* We do not sleep on target SoC. */
	if (!IS_ENABLED(CONFIG_SOC_SERIES_NRF54LX)) {
		while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
			countdown--;

```

The `while` loop doesn't really do anything when `CONFIG_SOC_SERIES_NRF54LX=y` except iterating 200 without a `cpu_sleep()` (which the compiler may actually remove anyways?)

cvinayak: You are right... I will follow up on this.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,23 @@ void radio_ar_status_reset(void)
 	hal_radio_nrf_ppi_channels_disable(BIT(HAL_TRIGGER_AAR_PPI));
 }
 
+static void ar_end_wait(void)
+{
+	/* An arbitrary iteration which seems sufficient testing on target with one IRK to resolve.
+	 * We should ideally be not be waiting anyway, and in theory be resolved before we are here.
```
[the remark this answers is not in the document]
cvinayak: done

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,23 @@ void radio_ar_status_reset(void)
...
+	 */
+	uint32_t countdown = 200U;
```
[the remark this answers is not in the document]
cvinayak: Keeping as-is for now.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,25 @@ void radio_ar_status_reset(void)
...
+
+	while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
+		countdown--;
+
+		/* We do not sleep on target SoC. */
+		if (!IS_ENABLED(CONFIG_SOC_SERIES_NRF54LX)) {
```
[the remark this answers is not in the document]
cvinayak: Correct use of not to sleep on target. we will sleep in simulation platforms.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
cvinayak: ok .`NRF_AAR->EVENTS_END` has `volatile` scope and compiler will not optimize the loop.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
```diff
@@ -2723,6 +2723,23 @@ void radio_ar_status_reset(void)
...
+
+	while ((NRF_AAR->EVENTS_END == 0U) && (countdown != 0U)) {
+		countdown--;
+
+#if !defined(CONFIG_SOC_SERIES_NRF54LX)
+		/* We do not sleep on target SoC. */
+		cpu_sleep();
+#endif /* !CONFIG_SOC_SERIES_NRF54LX */
+	}
+}
```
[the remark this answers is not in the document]
cvinayak: This is ok... `radio_ar_has_match()` will be used by caller to check the results of the address resolution.

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
Thalley: Not sure I understand your comment - Can you elaborate?

cvinayak: defined as `__IOM uint32_t EVENTS_END;` in the `NRF_AAR_Type` structure, where `__IOM` is `#define __IOM    volatile`

Thalley: I need more information :)

What I suggested was to put the entire loop inside `IS_ENABLED(CONFIG_SOC_SERIES_NRF54LX)`, instead of having the entire loop content inside `IS_ENABLED(CONFIG_SOC_SERIES_NRF54LX)`. It's cleaner and more readable

cvinayak: The loop is needed for all target SoCs (and all bsim targets including nrf54l15bsim) except that nRF54L15 SoC silicon shall not CPU sleep.

Thalley: Aaah, I see that 

```c
	while (NRF_AAR->EVENTS_END == 0) {
		cpu_sleep();
	}
```

Was moved now. Thought it made no sense to have a while loop just to sleep, and that it was an error ⁠

In that case it makes sense, but we really should have a comment explaining why we have a loop just doing cpu_sleep() for the other platforms

Thalley: Update: I do see that we actually have that already, so never mind :D

Thalley: I do still think it makes sense to put the `IS_ENABLED` as an outer check; even if `compiler will not optimize the loop.` is true, then it's an improvement in readability IMO (but not a blocker)

Thalley (dismissed):

subsys/bluetooth/controller/ll_sw/nordic/hal/nrf5/radio/radio.c
Thalley: Actually even if `NRF_AAR->EVENTS_END has volatile scope and compiler will not optimize the loop.` is true, wouldn't it still be optimized in terms of CPU size since it's effectively behind a `if (0)` even if the code size doesn't change due to `volatile`?

cvinayak: Add some explanation.

Thalley (approved):

aescolar (approved):

PR #100216, opened 2025-11-28
treewide: use auto-generated Kconfig compatible macro variables

mathieuchopstm:
Replace some manually-defined `DT_COMPAT_<>` Kconfig macro variables with their automatically generated counterparts. In most cases, this is straightforward as the manually defined macro is named identically to the one generated by the build system.

See also #100215

### DIFF
--- a/drivers/usb/uhc/Kconfig.max3421e
+++ b/drivers/usb/uhc/Kconfig.max3421e
@@ -1,8 +1,6 @@
 # Copyright (c) 2022 Nordic Semiconductor ASA
 # SPDX-License-Identifier: Apache-2.0
 
-DT_COMPAT_UHC_MAX3421E := maxim,max3421e-spi
-
 config UHC_MAX3421E
 	bool "MAX3421E driver"
 	default y
--- a/modules/hal_nordic/nrfx/Kconfig
+++ b/modules/hal_nordic/nrfx/Kconfig
@@ -277,13 +277,9 @@ config NRFX_SPI2
 	depends on $(dt_nodelabel_exists,spi2) && SOC_SERIES_NRF52X
 	select NRFX_SPI
 
-DT_COMPAT_NORDIC_NRF_SPIM := nordic,nrf-spim
-
 config NRFX_SPIM
 	bool "SPIM driver"
 	depends on $(dt_has_compat,$(DT_COMPAT_NORDIC_NRF_SPIM))
-
-DT_COMPAT_NORDIC_NRF_SPIS := nordic,nrf-spis
 
 config NRFX_SPIS
 	bool "SPIS driver"
--- a/drivers/debug/Kconfig.nrf
+++ b/drivers/debug/Kconfig.nrf
@@ -1,8 +1,6 @@
 # Copyright (c) 2024 Nordic Semiconductor ASA
 # SPDX-License-Identifier: Apache-2.0
 
-DT_COMPAT_NORDIC_NRF_TBM := nordic,nrf-tbm
-
 config DEBUG_NRF_ETR
 	bool "Coresight ETR handler (with Nordic TBM)"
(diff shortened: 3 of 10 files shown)


drivers/usb/uhc/Kconfig.max3421e
```diff
@@ -1,8 +1,6 @@
 # Copyright (c) 2022 Nordic Semiconductor ASA
 # SPDX-License-Identifier: Apache-2.0
 
-DT_COMPAT_UHC_MAX3421E := maxim,max3421e_spi
```
JarmouniA: The compatible name should probably be fixed as well, to use `-` instead of `_`. Even though I am unable to find a rule in the guidelines regarding compatible naming.
https://github.com/zephyrproject-rtos/zephyr/blob/1a49b418f53aff370a88dfaf78de1930c456d0d1/dts/bindings/usb/maxim%2Cmax3421e_spi.yaml#L7
https://docs.zephyrproject.org/latest/contribute/style/devicetree.html
https://docs.kernel.org/devicetree/bindings/dts-coding-style.html

mathieuchopstm: Outside PR scope.

@MaureenHelm @ozersa @ttmut @yasinustunerg @galak @microbuilder ^^ FYI

MaureenHelm: #100326

maass-hamburg (approved):

MaureenHelm (approved):

tmon-nordic (approved):

butok (approved):

modules/hal_nordic/nrfx/Kconfig
```diff
@@ -277,14 +277,10 @@ config NRFX_SPI2
...
 	select NRFX_SPI
 
-DT_COMPAT_NORDIC_NRF_SPIM := nordic,nrf-spim
-
 config NRFX_SPIM
 	bool "SPIM driver"
 	depends on $(dt_has_compat,$(DT_COMPAT_NORDIC_NRF_SPIM))
 
-DT_COMPAT_NORDIC_NRF_SPIS := nordic,nrf-spis
-
```
jfischer-no, line 287: I do not think you can remove them. They are not the same as generated `DT_HAS_"compatiple"_ENABLED` macros and actually used below as arguments in dt_has_compat().

jfischer-no (changes requested):

modules/hal_nordic/nrfx/Kconfig
mathieuchopstm: These macros are generated automatically by `gen_driver_kconfig_dts.py` - see `<build dir>/Kconfig/Kconfig.dts`.

This is a pattern already present in tree - see for example:
https://github.com/zephyrproject-rtos/zephyr/blob/3b23c76a246973fe7b1570447398f3bb1ef434e1/drivers/audio/Kconfig.pcm1681#L4-L9

jfischer-no (approved):

JarmouniA (commented):
Note that the `scripts/dts/gen_driver_kconfig_dts.py` script appears to be ignoring child-bindings
(like https://github.com/zephyrproject-rtos/zephyr/blob/5a58989daddda2a8a3c87396acaf4275b6c72596/dts/bindings/i2c/ti%2Ctca9544a.yaml#L10),
so `DT_COMPAT_{CHILD-COMPAT}` or `DT_HAS_{CHILD-COMPAT}_ENABLED` aren't generated. Any in-tree Kconfig relying on them is silently not working at the moment.
I verified that the compatibles being replaced in this PR are not of child-bindings.

JarmouniA (approved):

henrikbrixandersen:
Retriggered CI.

maass-hamburg (approved):

PR #100218, opened 2025-11-28
soc: st: stm32: reorganize family-wide common Kconfig

mathieuchopstm:
Split STM32 family-wide common Kconfig in two separate files: a new Kconfig file `soc/st/stm32/common/Kconfig` to hold options that affect the common code found in the same directory, and the existing top-level Kconfig file `soc/st/stm32/Kconfig` which now only holds options used by multiple series but not consumed by the common code - for example, options that are used by SoC-specific code or consumed by the STM32Cube HAL module go in this file.

### DIFF
--- a/soc/st/stm32/Kconfig
+++ b/soc/st/stm32/Kconfig
@@ -10,57 +10,28 @@ if SOC_FAMILY_STM32
 
 rsource "*/Kconfig"
 
-# STM32 wide symbols definitions
+# Kconfig options used to configure the family-wide code
+# found in `common/` are declared in `common/Kconfig`.
+# Some options are also shared by multiple series, but not
+# used by this family-wide common code; instead, they are
+# consumed by certain drivers, SoC-specific code or the
+# STM32Cube HAL module.
+#
+# (Please indicate as a comment where each option is consumed)
 
-# Workaround for not being able to have commas in macro arguments
-DT_CHOSEN_Z_CCM := zephyr,ccm
-
-config STM32_CCM
-	def_bool $(dt_chosen_enabled,$(DT_CHOSEN_Z_CCM))
-
+# This option is consumed by the HAL module
 config USE_STM32_ASSERT
 	depends on ASSERT
 	bool "STM32Cube HAL and LL drivers asserts"
 	help
 	  Enable asserts in STM32Cube HAL and LL drivers.
 
-config STM32_BACKUP_SRAM
-	bool "STM32 Backup SRAM"
-	default y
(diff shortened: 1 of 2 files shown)


soc/st/stm32/common/Kconfig
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) 2025 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+#
+# Kconfig options for STM32 family-wide common code
```
erwango: Disambiguate 'common'
```suggestion
# Kconfig options for STM32 family-wide common/ code
```

soc/st/stm32/Kconfig
```diff
@@ -10,57 +10,29 @@ if SOC_FAMILY_STM32
...
-# STM32 wide symbols definitions
-
-# Workaround for not being able to have commas in macro arguments
-DT_CHOSEN_Z_CCM := zephyr,ccm
-
-config STM32_CCM
-	def_bool $(dt_chosen_enabled,$(DT_CHOSEN_Z_CCM))
-
+# `common/Kconfig` declares all Kconfig options used to
+# configure the family-wide code found in `common/`.
```
erwango: A bit unclear. what about:
```
Kconfig options used to configure the family-wide code 
found in `common/` are made available in common/Konfig
```

soc/st/stm32/Kconfig
```diff
@@ -10,57 +10,29 @@ if SOC_FAMILY_STM32
...
+#
+# However, some Kconfig options are shared by all series but
+# not consumed by common code; instead, SoC-specific code,
+# certain drivers or the STM32Cube HAL module will make use
+# of them. Such Kconfig options are declared here.
```
erwango:
```suggestion
# Kconfig options shared by all series, but not consumed by common code,
# such as SoC-specific code, certain drivers or the STM32Cube HAL module
# are declared below.
```

erwango (changes requested):

soc/st/stm32/common/Kconfig
mathieuchopstm: Removed `common` and added more verbose message instead

soc/st/stm32/Kconfig
mathieuchopstm: Slightly reworded: `made available` ==> `declared`.

soc/st/stm32/Kconfig
mathieuchopstm: Done with slight reword.

erwango (approved):

etienne-lms (approved):

PR #100221, opened 2025-11-28
scripts: gen_relocate_app.py: Replace quoted section name with wildcard.

### LINKED ISSUE #49155: Ability to relocate noinit section variables
It's currently not possible to relocate a zephyr heap (located in `noinit` section). 
Additionally, the `__noinit` and `__noinit_named` macros extend to a section containing `__FILE__` (for gcc) which contains characters that are incompatible for the `linker.cmd` to work.

Adding an additional target to the cmake `zephyr_code_relocate` extension for the `noinit` section.

I've submitted a proposal [here](https://github.com/zephyrproject-rtos/zephyr/pull/39615), however this affects the build mechanism greatly, and doesn't work for generator expressions.

**Additional context**
A simple example to demonstrate relocating a heap to `DTCM`, which currently doesn't build/work.

```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.20.0)
find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
project(relocate)

# relocating everything to DTCM would also relocate instructions
zephyr_code_relocate(src/main.c DTCM_NOINIT)

target_sources(app PRIVATE src/main.c)
```

```c
/* src/main.c */
#include <zephyr/kernel.h>

K_HEAP_DEFINE(custom_heap, 1024);

void main(void)
{
	void *ptr = k_heap_alloc(&custom_heap, 64, K_FOREVER);
	k_heap_free(&custom_heap, ptr);
}
```

```ini
# prj.conf
CONFIG_CODE_DATA_RELOCATION=y
```

iabdalkader:
ld seems to treat quotes in section names as literal characters, which makes it fail to match section names outputted by this script. This fix replaces quoted section names with a wildcard, which matches the section correctly.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/49155

Context: I'm trying to relocate the kernel heap to another SRAM, using `CONFIG_CODE_DATA_RELOCATION`. In `CMakeLists.txt` I have the following:

```
zephyr_code_relocate(FILES ${ZEPHYR_BASE}/kernel/mempool.c LOCATION SRAM2_BSS_NOINIT)
```

The generated linker scripts seems to have the right section in SRAM2. Note that the section name is quoted:
```
 .sram2_noinit_reloc (NOLOAD) :
        {
                . = ALIGN(4);
                KEEP(*mempool.c.obj(.noinit."WEST_TOPDIR/zephyr/kernel/mempool.c".kheap_buf__system_heap))
                . = ALIGN(4);
 } > SRAM2 AT > SRAM2
```

In the final elf, the section is empty because ld fails to match it:

```
  .sram2_noinit_reloc
                  0x0000000030020000        0x0
                  0x0000000030020000                . = ALIGN (0x4)
   *mempool.c.obj(SORT_BY_ALIGNMENT(.noinit.) SORT_BY_ALIGNMENT(WEST_TOPDIR/zephyr/kernel/mempool.c)
```

With this patch:
```
.sram2_noinit_reloc
                0x0000000030020000    0x20000
                0x0000000030020000                . = ALIGN (0x4)
 *mempool.c.obj(SORT_BY_ALIGNMENT(.noinit.*.kheap_buf__system_heap))
 .noinit."WEST_TOPDIR/zephyr/kernel/mempool.c".kheap_buf__system_heap
```

### DIFF
--- a/scripts/build/gen_relocate_app.py
+++ b/scripts/build/gen_relocate_app.py
@@ -332,9 +332,9 @@ def print_linker_sections(list_sections: 'list[OutputSection]'):
     out = ''
     for section in sorted(list_sections):
         template = PRINT_TEMPLATE if section.keep else PRINT_TEMPLATE_NOKEEP
-        out += template.format(
-            obj_file_name=section.obj_file_name, section_name=section.section_name
-        )
+        # Replace quoted section name with wildcard.
+        section_name = re.sub(r'\.\"[^\"]+\"\.', '.*.', section.section_name)
+        out += template.format(obj_file_name=section.obj_file_name, section_name=section_name)
     return out
 
 


scripts/build/gen_relocate_app.py
```diff
@@ -332,9 +332,9 @@ def print_linker_sections(list_sections: 'list[OutputSection]'):
     out = ''
     for section in sorted(list_sections):
         template = PRINT_TEMPLATE if section.keep else PRINT_TEMPLATE_NOKEEP
-        out += template.format(
-            obj_file_name=section.obj_file_name, section_name=section.section_name
-        )
+        # Replace quoted section name with wildcard.
+        section_name = re.sub(r'\.\"[^\"]+\"\.', '.*.', section.section_name)
+        out += template.format(obj_file_name=section.obj_file_name, section_name=section_name)
```
tejlmand, line 337: I would like to have this fixed before entering the script.

I wonder how / when this issue first started to appear.

tejlmand (changes requested):
thanks for the fix, but I would like to get a better understanding of when this started to fail, cause it looks like a regression.

iabdalkader:
FYI, since I opened this, there's been two other fixes for this exact same issue, that address the quote issue at a higher level, so they're probably better than this one.

https://github.com/zephyrproject-rtos/zephyr/pull/100260
https://github.com/zephyrproject-rtos/zephyr/pull/100917

I haven't actually tested either one of them, but happy to confirm if they at least fix our issue or not.

PR #100223, opened 2025-11-28
drivers: remove on-stack copying of dma cfg on stm32 drivers

mariopaja:
This change removes the on-stack copying and instead uses a pointer to the configuration to  avoid unnecessary stack usage

### DIFF
--- a/drivers/video/video_stm32_dcmi.c
+++ b/drivers/video/video_stm32_dcmi.c
@@ -45,6 +45,7 @@ struct video_stm32_dcmi_data {
 	struct k_fifo fifo_in;
 	struct k_fifo fifo_out;
 	struct video_buffer *vbuf;
+	struct stream dma;
 };
 
 struct video_stm32_dcmi_config {
@@ -52,7 +53,6 @@ struct video_stm32_dcmi_config {
 	irq_config_func_t irq_config;
 	const struct pinctrl_dev_config *pctrl;
 	const struct device *sensor_dev;
-	const struct stream dma;
 };
 
 static void stm32_dcmi_process_dma_error(DCMI_HandleTypeDef *hdcmi)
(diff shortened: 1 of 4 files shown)


drivers/video/video_stm32_dcmi.c
```diff
@@ -45,14 +45,14 @@ struct video_stm32_dcmi_data {
 	struct k_fifo fifo_in;
 	struct k_fifo fifo_out;
 	struct video_buffer *vbuf;
+	struct stream dma;
```
gautierg-st, line 48: This change (moving `struct stream dma` from config to data) should at least be explained in the commit message (maybe even be its own commit since it's more than changing `.` to `->`).
Otherwise LGTM.

mathieuchopstm (commented):
LGTM except for the remark of @gautierg-st

mathieuchopstm (dismissed):

drivers/video/video_stm32_dcmi.c
```diff
@@ -147,16 +147,16 @@ static int stm32_dma_init(const struct device *dev)
 	 * the minimum information to inform the DMA slot will be in used and
 	 * how to route callbacks.
 	 */
-	struct dma_config dma_cfg = config->dma.cfg;
+	struct dma_config *dma_cfg = &data->dma.cfg;
```
avolmat-st:
```suggestion
struct dma_config *dma_cfg = &dma->cfg;
```

avolmat-st (changes requested):
Very minor point so I am also ok to approve it as is. Since dma pointer in introduced in the stm32_dma_init function, one of dma_cfg pointer can directly be retrieved via the dma pointer as done in all other places of that function.

JarmouniA (commented):
This kind of change should be examined carefully, not saying it is wrong.
It's not just a matter of stack-usage, it's also a question of: Do we want the subsequent values assigned to the members of the stack-allocated struct to be setted globally, or just be passed to the functions called locally with the struct as an arg and then be forgotten?
Since the drivers in qst seem to be working fine without those values being setted globally (unless there are undiscovered bugs, or the values are setted elsewhere), I kind of question the need to set them in the global config/data struct.

gautierg-st (approved):

avolmat-st (approved):

erwango (approved):

mariopaja:
@gautierg-st I will have a look into it

PR #100231, opened 2025-11-28
drivers: display: SDL: Add SDL_DISPLAY_ROUNDED_MASK option

faxe1008:
While designing some GUI for rounded displays, I encountered the convenience issue that I like to prototype my UIs in the native sim and only do a final check on actual hardware, since its much fast to iterate. Anyways for rounded displays its annoying to have to guess which part of the screen is not visible. It might not be completely pixel perfect but it allows for first good estimate:

Screenshot 2025-11-28 192851

On real hardware:
Unbenannt

Note: the second commit may be considered optional, but at the point where the function takes 18 parameters it felt sensibly to add this refactoring.

### DIFF
--- a/drivers/display/display_sdl_bottom.c
+++ b/drivers/display/display_sdl_bottom.c
@@ -66,63 +66,63 @@ static int sdl_create_rounded_display_mask(uint16_t width, uint16_t height, uint
 	return 0;
 }
 
-int sdl_display_init_bottom(uint16_t height, uint16_t width, uint16_t zoom_pct,
-			    bool use_accelerator, void **window, const void *window_user_data,
-			    const char *title, void **renderer, void **mutex, void **texture,
-			    void **read_texture, void **background_texture,
-			    uint32_t transparency_grid_color1, uint32_t transparency_grid_color2,
-			    uint16_t transparency_grid_cell_size, void **round_disp_mask,
-			    uint32_t mask_color)
+int sdl_display_init_bottom(struct sdl_display_init_params *params)
 {
 	/* clang-format off */
-	*window = SDL_CreateWindow(title, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
-				   width * zoom_pct / 100,
-				   height * zoom_pct / 100, SDL_WINDOW_SHOWN);
+	*params->window = SDL_CreateWindow(params->title, SDL_WINDOWPOS_UNDEFINED,
+				   SDL_WINDOWPOS_UNDEFINED,
+				   params->width * params->zoom_pct / 100,
+				   params->height * params->zoom_pct / 100, SDL_WINDOW_SHOWN);
 	/* clang-format on */
-	if (*window == NULL) {
-		nsi_print_warning("Failed to create SDL window %s: %s", title, SDL_GetError());
+	if (*params->window == NULL) {
+		nsi_print_warning("Failed to create SDL window %s: %s", params->title,
+				  SDL_GetError());
 		return -1;
 	}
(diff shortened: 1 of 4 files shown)


JarmouniA:
Cc @aescolar regarding the 2nd commit.

drivers/display/display_sdl_bottom.c
```diff
@@ -98,6 +99,56 @@ int sdl_display_init_bottom(uint16_t height, uint16_t width, uint16_t zoom_pct,
 	}
 	SDL_UnlockTexture(*background_texture);
 
+	/* Create ellipse mask texture if rounded mask is enabled */
+	if (rounded_mask) {
+		*mask_texture = SDL_CreateTexture(*renderer, SDL_PIXELFORMAT_ARGB8888,
```
aescolar: Given that SonarCloud is complaining about the complexity of this function and this is quite an isolated calculation, maybe this new block should be refactored into its own function?

drivers/display/display_sdl_bottom.c
```diff
@@ -136,6 +188,14 @@ void sdl_display_write_bottom(const uint16_t height, const uint16_t width, const
 				       color_tint & 0xff);
 		SDL_RenderCopy(renderer, texture, NULL, NULL);
 		SDL_SetTextureColorMod(texture, 255, 255, 255);
+
+		/* Apply ellipse mask if enabled */
+		if (rounded_mask && mask_texture != NULL) {
```
aescolar: nit: it would be possible to just do (mask_texture != NULL) and save passing rounded_mask and the first part of the check.
(same in sdl_display_blanking_off_bottom)

drivers/display/display_sdl_bottom.c
```diff
@@ -18,7 +18,8 @@ int sdl_display_init_bottom(uint16_t height, uint16_t width, uint16_t zoom_pct,
 			    const char *title, void **renderer, void **mutex, void **texture,
 			    void **read_texture, void **background_texture,
 			    uint32_t transparency_grid_color1, uint32_t transparency_grid_color2,
-			    uint16_t transparency_grid_cell_size)
+			    uint16_t transparency_grid_cell_size, bool rounded_mask,
+			    void **mask_texture, uint32_t mask_color)
```
aescolar: nit: maybe mask_texture could be called round_disp_mask to be more descriptive?

aescolar (commented):
Quite neat feature :)
Just a few nits.
(I don't lean either way regarding the 2nd commit :) )

drivers/display/display_sdl_bottom.c
faxe1008: Great point

drivers/display/display_sdl_bottom.c
```diff
@@ -66,63 +66,63 @@ static int sdl_create_rounded_display_mask(uint16_t width, uint16_t height, uint
...
 }
 
-int sdl_display_init_bottom(uint16_t height, uint16_t width, uint16_t zoom_pct,
-			    bool use_accelerator, void **window, const void *window_user_data,
-			    const char *title, void **renderer, void **mutex, void **texture,
-			    void **read_texture, void **background_texture,
-			    uint32_t transparency_grid_color1, uint32_t transparency_grid_color2,
-			    uint16_t transparency_grid_cell_size, void **round_disp_mask,
-			    uint32_t mask_color)
+int sdl_display_init_bottom(const struct sdl_display_init_params *params)
```
danieldegrasse: Can we drop the `const` qualifier here? Since we are editing the contents of the structure it shouldn't be const anyways, should it?

JarmouniA (approved):

aescolar (approved):

PR #100233, opened 2025-11-28
tests: settings: add a persistence test

dsseng:
Validate that after some time the settings are commited to the
persistent storage and are not lost on reboot.

Signed-off-by: Dmitrii Sharshakov <d3dx12.xx@gmail.com>


tests/subsys/settings/functional/src/settings_basic_test.c
```diff
@@ -1307,3 +1315,65 @@ ZTEST(settings_functional, test_single_save)
 	settings_deregister(&first_settings);
 #endif
 }
+
+#ifdef CONFIG_REBOOT
+#define TEST_VALUE 0xfeedda7a
+ZTEST(settings_functional, test_persistence)
+{
+	// TODO: Z_TEST_SKIP_IFNDEF(CONFIG_TEST_SETTINGS_PERSISTENCE);
```
dsseng, line 1323: I am not sure when do we want to enable or disable this test. It requires the device to be able to programmatically reboot, as well as persist the `__noinit` variable this test uses to keep track of its reboots. This certainly works with QEMU for TF-M tests, but I have not tested on any physical hardware

PR #94235, opened 2025-08-07
bt: services: gap_default: Extract GAP service from gatt implementation

rkel:
Remove Generic Access Service implementation from gatt files and move it into separate file in services directory.
The proposed name is just gap_default as it is default implementation that may be switched off by the user in purpose of providing the GAP implementation directly in the application.

In the end it serves the same purpose that #93946 but gives the user total freedom of implementation when needed in the same time being 100% compatible with the current solution.

I was testing it manually and now I can enable and disable default GAP implementation.

The main reason of this PR is to discuss if we wish to go into this solution instead of #93946.

To be done:
[v] - GAP service validation during bt_enable.
[v] - review all configurations that need to be moved to Kconfig.gap_default
[v] - test of the functionality (BabbleSim)
[v] - sample with GAP reimplementation on App side
[v] - update documentation

PR #100235, opened 2025-11-28
Bluetooth: GATT: Use check offset against 0 when GAP name or apperance is written

rkel:
Update write_name function to check the offset value against 0 as long writes are not used here.
Simplify write_appearance function also to remove long writes support.
The appearance property is always 16 bit and is should not even support long writes.

### DIFF
--- a/subsys/bluetooth/host/gatt.c
+++ b/subsys/bluetooth/host/gatt.c
@@ -121,7 +121,7 @@ static ssize_t write_name(struct bt_conn *conn, const struct bt_gatt_attr *attr,
 	/* adding one to fit the terminating null character */
 	char value[CONFIG_BT_DEVICE_NAME_MAX + 1] = {};
 
-	if (offset >= CONFIG_BT_DEVICE_NAME_MAX) {
+	if (offset > 0) {
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
 	}
 
@@ -155,21 +155,18 @@ static ssize_t write_appearance(struct bt_conn *conn, const struct bt_gatt_attr 
 			 const void *buf, uint16_t len, uint16_t offset,
 			 uint8_t flags)
 {
-	uint16_t appearance_le = sys_cpu_to_le16(bt_get_appearance());
-	char * const appearance_le_bytes = (char *)&appearance_le;
 	uint16_t appearance;
 	int err;
 
-	if (offset >= sizeof(appearance_le)) {
+	if (offset > 0) {
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
 	}
 
-	if ((offset + len) > sizeof(appearance_le)) {
+	if (len != sizeof(appearance)) {
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_ATTRIBUTE_LEN);
 	}
 
-	memcpy(&appearance_le_bytes[offset], buf, len);
-	appearance = sys_le16_to_cpu(appearance_le);
+	appearance = sys_get_le16(buf);
 
 	err = bt_set_appearance(appearance);
 


subsys/bluetooth/host/gatt.c
```diff
@@ -120,18 +120,21 @@ static ssize_t write_name(struct bt_conn *conn, const struct bt_gatt_attr *attr,
...
 	if (offset + len > CONFIG_BT_DEVICE_NAME_MAX) {
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_ATTRIBUTE_LEN);
 	}
 
-	memcpy(value, buf, len);
+	memcpy(value, current_name, current_name_len + 1);
+	memcpy(value + offset, buf, len);
 
-	value[len] = '\0';
+	value[offset + len] = '\0';
```
Thalley: I understand the `+ offset` changes that seem valid, but why do we care about `current_name` and `current_name_len`?

subsys/bluetooth/host/gatt.c
```diff
@@ -120,18 +120,21 @@ static ssize_t write_name(struct bt_conn *conn, const struct bt_gatt_attr *attr,
 {
 	/* adding one to fit the terminating null character */
 	char value[CONFIG_BT_DEVICE_NAME_MAX + 1] = {};
+	const char *current_name = bt_get_name();
+	size_t current_name_len = strlen(current_name);
 
-	if (offset >= CONFIG_BT_DEVICE_NAME_MAX) {
+	if (offset > current_name_len) {
```
Thalley: Correct me if I'm wrong, but isn't `offset` only used in long writes? Which require the characteristic to have `BT_GATT_PERM_PREPARE_WRITE`, but `BT_GATT_CHARACTERISTIC(BT_UUID_GAP_DEVICE_NAME` does not have that? 

So in the current implementation, `offset` cannot be anything but 0

Thalley (changes requested):

subsys/bluetooth/host/gatt.c
rkel: @Thalley Thank you for that comment - as I had already stated in the PR description, we should allow to use offset or check it against 0. Current implementation is misleading. Then the whole change in this pr should be simply if (offset > 0) -> return error.

subsys/bluetooth/host/gatt.c
rkel: The reason for that is that w are changing current name - only a part between offset and offset + len. And if current name contains "ABC" we cannot write it with offset +5 because the result would still be "ABC". I was trying to move the changes from write_appearance - but maybe it does not make sense for both of them to be so complicated and they both should just check offset against 0?

subsys/bluetooth/host/gatt.c
Thalley: If we only support "regular" writes to the name, then `offset` will always be 0 and we should not allow the use of offset. `offset` is only ever non-0 in the case of long writes, which require `BT_GATT_PERM_PREPARE_WRITE`. It's probably allowed to support long writes for the name, but it will require additional work

rkel: So to summarize - the write_apperance properly supports long writes, but long writes are never really enabled here. So we could possibly spare a few bytes of code by simplifying it.

In other hand - write_name in current form is total mess - it check the offset against name max length, but then totally ignores the offset values. Also - the long writes are never enabled. Can I just simplify both this functions in this PR instead of complicating write_name?

I am pointing this out, because right now I have some small window to work on this: #94235. And there would be conflict between this changes and the ones done in #94235.

Thalley: Neither `write_appearance` nor `write_name` support long writes. 

`write_appearance` should never support it (given that it's a 16-bit value). `write_name` could in theory support it in the future, but now it doesn't. 

We can remove any `offset` checks in both of them, as they are basically invalid checks given that regular writes will never have a 0-zero offset

Thalley: Oh, nvm, the `write_appearance` may actually contain a string too, so we could modify that to support long write as well in the future

Thalley: > Correct me if I'm wrong, but isn't `offset` only used in long writes?

Correcting myself here. Apparently `BT_GATT_PERM_PREPARE_WRITE` does not indicate permissions for prepare write, and all GATT characteristics that support write support prepare and execute writes implicitly, and must handle cases where `offset != 0`

rkel: Now you surprised me - I was reading this document: https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-access-profile.html#UUID-f3f0ba03-b592-12ce-5534-f0be9b18ae9d, I can see here: ``The Appearance characteristic value shall be the enumerated value as defined in [Assigned Numbers](https://www.bluetooth.com/specifications/assigned-numbers). The Appearance characteristic value shall be 2 octets in length.``. Is there any other additional specification that allows it to be a string?

Thalley: > Now you surprised me - I was reading this document: https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-access-profile.html#UUID-f3f0ba03-b592-12ce-5534-f0be9b18ae9d, I can see here: `The Appearance characteristic value shall be the enumerated value as defined in [Assigned Numbers](https://www.bluetooth.com/specifications/assigned-numbers). The Appearance characteristic value shall be 2 octets in length.`. Is there any other additional specification that allows it to be a string?

I misread our code. I expected it to just by 16-bits as well, but then saw `char * const appearance_le_bytes` that I somehow (probably due to `char`) misinterpreted as a string. 

However `write_appearance` should reject any write request that isn't `len == 2` and `offset = 0` in any case :)

subsys/bluetooth/host/gatt.c
```diff
@@ -155,21 +155,19 @@ static ssize_t write_appearance(struct bt_conn *conn, const struct bt_gatt_attr
...
-	if ((offset + len) > sizeof(appearance_le)) {
+	if (len != sizeof(appearance)) {
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_ATTRIBUTE_LEN);
-	}
+	}`
 
-	memcpy(&appearance_le_bytes[offset], buf, len);
-	appearance = sys_le16_to_cpu(appearance_le);
+	memcpy(&appearance, buf, len);
+	appearance = sys_le16_to_cpu(appearance);
```
Thalley: This is fine, but suggest to just do

```suggestion
	appearance = sys_get_le16(buf);
```

subsys/bluetooth/host/gatt.c
```diff
@@ -155,21 +155,19 @@ static ssize_t write_appearance(struct bt_conn *conn, const struct bt_gatt_attr
 			 const void *buf, uint16_t len, uint16_t offset,
 			 uint8_t flags)
 {
-	uint16_t appearance_le = sys_cpu_to_le16(bt_get_appearance());
-	char * const appearance_le_bytes = (char *)&appearance_le;
 	uint16_t appearance;
 	int err;
 
-	if (offset >= sizeof(appearance_le)) {
+	if (offset >= 0) {
```
Thalley:
```suggestion
	if (offset > 0) {
```

Thalley (approved):
I think these are good changes. It disables support for GATT long writes, but they were not properly implement for the name before, and never needed for appearance.

alwa-nordic (approved):

jhedberg (approved):

PR #100240, opened 2025-11-29
Drivers: interrupt: Add NXP GINT driver

### LINKED ISSUE #41979: Add more GPIO interrupt sources with GINT driver support for LPC5500 family
The current GPIO driver used for the LPC5500 family and the superset board LPCXpresso55S69 configure pin interrupts using the on-chip peripheral PINT.  The PINT allows up to 8 GPIO pins to trigger interrupts, and each have a different IRQ.  But the PINT is limited to just 8 pins in this family.  The LPC5500 devices also have another GPIO interrupt controller called GINT that enables many more pins to generate interrupts using shared IRQs.  

This enhancement should add support in the GPIO driver for the GINT, for applications that desire more than 8 GPIO interrupt pins.  It is acceptable for the enhanced driver to require the application to choose just one option at build time, to simplify for everyone: use either PINT or GINT.

The driver source file currently using the PINT is \drivers\gpio\gpio_mcux_lpc.c, and is included when CONFIG_GPIO_MCUX_LPC is selected.  Changing this driver would also impact the supported devices LPC54114, i.MX RT685 and the open [PR for RT595](https://github.com/zephyrproject-rtos/zephyr/pull/41669).

Thank you

zejiang0jason:
Add NXP GINT driver
Add the test
Enable driver and test for NXP boards

Depends on:
https://github.com/zephyrproject-rtos/zephyr/pull/103328


zejiang0jason:
The DTS compliance fail needs this fix: https://github.com/zephyrproject-rtos/zephyr/pull/100133

drivers/interrupt_controller/intc_nxp_gint.c
```diff
@@ -0,0 +1,308 @@
...
+
+static void nxp_gint_isr(const struct device *dev)
+{
+	struct nxp_gint_data *data = dev->data;
+	const struct nxp_gint_config *config = dev->config;
+	uint32_t ctrl;
+
+	/* Clear interrupt flag */
+	ctrl = gint_read_reg(config, GINT_CTRL_OFFSET);
+	gint_write_reg(config, GINT_CTRL_OFFSET, ctrl);
```
Holt-Sun: Should we firstly check `if (ctrl & GINT_CTRL_INT_BIT)`, then `gint_write_reg(config, GINT_CTRL_OFFSET, ctrl);` to clear the flag?

tests/drivers/interrupt_controller/intc_nxp_gint/src/main.c
```diff
@@ -0,0 +1,391 @@
...
+static void test_gint_edge_callback(const struct device *dev, void *user_data)
+{
+	atomic_inc(&gint_callback_count);
+}
+
+static void test_gint_level_callback(const struct device *dev, void *user_data)
+{
+	atomic_inc(&gint_callback_count);
+
+	nxp_gint_disable_pin(gint_dev, TEST_PORT_0, TEST_PIN_0);
```
Holt-Sun, line 63: Disable the same port pin twice.

drivers/interrupt_controller/intc_nxp_gint.c
```diff
@@ -0,0 +1,313 @@
...
+static inline uint32_t gint_read_reg(const struct nxp_gint_config *config, uint32_t offset)
+{
+	return sys_read32(config->base + offset);
+}
+
+static inline void gint_write_reg(const struct nxp_gint_config *config,
+				  uint32_t offset,
+				  uint32_t value)
+{
+	sys_write32(value, config->base + offset);
```
hakehuang: write shall use a mutex to protect multiply write

hakehuang:
> Hi @hakehuang , could you please test for the supported boards? thanks.
> - frdm_mcxw23
> - mcxw23_evk
> - lpcxpresso55s36
> - lpcxpresso51u68
> - lpcxpresso55s06
> - lpcxpresso55s16
> - lpcxpresso55s28
> - lpcxpresso55s69/lpc55s69/cpu0

board testing pass on v4.3.0-1313-g93b81ec66550.

drivers/interrupt_controller/intc_nxp_gint.c
```diff
@@ -0,0 +1,313 @@
...
+}
+
+#define NXP_GINT_IRQ_CONFIG(n)							\
+	static void nxp_gint_irq_init_##n(void)					\
+	{									\
+		IRQ_CONNECT(DT_INST_IRQN(n),					\
+			    DT_INST_IRQ(n, priority),				\
+			    nxp_gint_isr,					\
+			    DEVICE_DT_INST_GET(n),				\
+			    0);							\
```
hakehuang, line 293: your indent is wrong

drivers/interrupt_controller/intc_nxp_gint.c
```diff
@@ -0,0 +1,313 @@
...
+static void nxp_gint_isr(const struct device *dev)
+{
+	struct nxp_gint_data *data = dev->data;
+	const struct nxp_gint_config *config = dev->config;
+	uint32_t ctrl;
+
+	ctrl = gint_read_reg(config, GINT_CTRL_OFFSET);
+
+	/* Clear interrupt flag */
+	if (0U != (ctrl & GINT_CTRL_INT_BIT)) {
```
hakehuang, line 94: clear before your callback, will cause interrupt lost during callback execution, so this is better move after callback call

drivers/interrupt_controller/intc_nxp_gint.c
zejiang0jason: I guess this is because github shows tab size as 4 characters, I set tab size as 8 in my local env, the result is right.

drivers/interrupt_controller/intc_nxp_gint.c
zejiang0jason: why will it cause interrupt lost? if the interrupt is pending again after cleared, then the interrupt will happen again

hakehuang: ⚠️ CAN LOSE (Edge-triggered or pulse):
   - External interrupts (EXTI) configured as edge
   - Some DMA completion interrupts
   - Software-triggered interrupts (counting events)
   - If you clear before reading the cause, events can merge
   
  sorry, to clarify , this is about data processing issue, for GINT, it shall be used combine with other peripheral, and callback usually handler the data processing, so to ensure the data process is clean, later clear interrupt is more reliable.

zejiang0jason:
Hello @mmahadevan108 , could you please help review? thanks.

soc/nxp/lpc/lpc51u68/soc.c
mmahadevan108, line 1: Commit subject should be changed to `soc` instead of `drivers`

dts/arm/nxp/lpc/nxp_lpc51u68.dtsi
mmahadevan108, line 1: Commit subject should change from `drivers` to `dts`

drivers/interrupt_controller/Kconfig.nxp_gint
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 NXP
```
mmahadevan108: Please update Copyright. Same comment across this PR.

tests/drivers/interrupt_controller/intc_nxp_gint/CMakeLists.txt
mmahadevan108, line 1: Please squash the 2 test commits

Holt-Sun (approved):

mmahadevan108 (approved):

PR #100245, opened 2025-11-29
doc: use dnf5 for Fedora documentation tools

SanjayrajN-in:
In newer Fedora versions the package manager CLI is dnf5 instead of dnf, so commands using dnf may fail. This change updates the Fedora documentation tools instructions to use dnf5 and mentions that some systems may still use dnf.


str4t0m:
Thanks for the PR.
To the best of my knowledge, on Fedora installations >= 41 `usr/bin/dnf` links to `usr/bin/dnf5`, and there is no plan to change that. Could you clarify if you have faced any issue or why you would prefer to document dnf5 explicitly?

PR #96902, opened 2025-10-02
boards: nxp: Convert zephyr_compile_definitions

dnltz:
Either convert zephyr_compile_definitions directly in the board directory to zephyr_library_compile_definitions or move those definitions into the soc directory and convert there.

PR #95705, opened 2025-09-09
boards: phytec: Add support for phyCORE-RT1170

dnltz:
This is another attempt to add support for the phyBOARD-Atlas development kit featuring the phyCORE-RT1170 SOM.

While #86756 had both Cortex-M CPUs included, this PR only adds the M7 and partly prepares this board for adding the M4 later.  We first need to figure out how to properly integrate the M4 in our development kit as well as testing infrastructure.

Depends on the following PRs being merged first:
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/95629

PR #91316, opened 2025-06-09
Support usdhc feature on rt700 cm33_cpu0 core

lucien-nxp:
enable tests/subsys/fs/fat_fs_api case with mmc feature

PR #100252, opened 2025-11-30
boards: nxp: adjusting the scope of boot-related build flags

ZhaoxiangJin:
* For the RT10xx and RT11xx families, the compile flag ``BOARD_FLASH_SIZE``, originally defined in
    ``boards/nxp/mimxrt10xx_evk/CMakeLists.txt`` and ``boards/nxp/mimxrt11xx_evk/CMakeLists.txt``, is
    used only by the HAL header ``fsl_flexspi_nor_boot.h``, which is included by
    :zephyr_file:`soc/nxp/imxrt/imxrt10xx/soc.c` and :zephyr_file:`soc/nxp/imxrt/imxrt11xx/soc.c`.
    To avoid potential collisions with other global flags, the macro is now defined at the SoC layer
    using ``zephyr_library_compile_definitions()`` in :zephyr_file:`soc/nxp/imxrt/imxrt10xx/CMakeLists.txt`
    and :zephyr_file:`soc/nxp/imxrt/imxrt11xx/CMakeLists.txt`. This change has been applied to all
    RTxxxx boards.

  * For the RTxxx family, the compile flag ``BOARD_FLASH_SIZE``, originally defined in
    ``boards/nxp/mimxrtxxx_evk/CMakeLists.txt``, is not used in the Zephyr tree and has
    therefore been removed from all RTxxx board CMakeLists.txt files.

  * For the RTxxx family, the compile flag ``BOOT_HEADER_ENABLE``, previously defined in
    ``boards/nxp/mimxrtxxx_evk/CMakeLists.txt`` and used in ``boards/nxp/rtxxx/<boot_header>.c``,
    has been replaced by a Kconfig option. Consequently, the line
    ``zephyr_compile_definitions(BOOT_HEADER_ENABLE=1)`` has been removed from the RTxxx board
    CMakeLists.txt files.

  * Removed compile flag ``BOOT_HEADER_ENABLE`` definition from :zephyr_file:`boards/nxp/rd_rw612_bga/CMakeLists.txt`,
    as it is not used in the Zephyr tree.

  * Originally, the compile flags ``XIP_BOOT_HEADER_ENABLE`` and ``XIP_BOOT_HEADER_DCD_ENABLE`` were
    used in ``boards/nxp/rt1xxx/<boot_header>.c``. These flags have been converted to Kconfig options
    across NXP RTxxxx evaluation boards, allowing boot-header configuration via the Kconfig build system
    instead of compile-time defines. Consequently, we removed ``zephyr_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)``
    and ``zephyr_compile_definitions(XIP_BOOT_HEADER_DCD_ENABLE=1)`` from the RTxxxx board-level CMakeLists.txt files.
    Because these macros are also required by ``hal_nxp/rt10xx/fsl_flexspi_nor_boot.h`` and
    ``hal_nxp/rt11xx/fsl_flexspi_nor_boot.h``, they were added to the corresponding SoC-layer CMakeLists.txt files
    using ``zephyr_library_compile_definitions()`` to limit their scope.

rework https://github.com/zephyrproject-rtos/zephyr/pull/96902

### DIFF
--- a/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
@@ -4,7 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-zephyr_sources(soc.c)
+zephyr_library()
+zephyr_library_sources(soc.c)
 
 if(CONFIG_PM)
   zephyr_sources(power.c)
--- a/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# Copyright 2024 NXP
+# Copyright 2024-2025 NXP
 #
 # SPDX-License-Identifier: Apache-2.0
 #
@@ -20,6 +20,12 @@ if(CONFIG_MEMC_MCUX_FLEXSPI)
   zephyr_sources(flexspi.c)
   if(CONFIG_FLASH_MCUX_FLEXSPI_XIP)
     zephyr_code_relocate(FILES flexspi.c LOCATION ${CONFIG_FLASH_MCUX_FLEXSPI_XIP_MEM}_TEXT)
+  endif()
+endif()
+
+if(CONFIG_NXP_IMXRT_BOOT_HEADER)
+  if(CONFIG_BOOT_FLEXSPI_NOR)
+    zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
   endif()
 endif()
 
--- a/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
@@ -27,6 +27,10 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
   if(CONFIG_BOOT_FLEXSPI_NOR)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
   endif()
+
+  if(CONFIG_DEVICE_CONFIGURATION_DATA)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_DCD_ENABLE=1)
+  endif()
 endif()
 
 zephyr_include_directories(.)
--- a/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
@@ -4,7 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-zephyr_sources(soc.c)
+zephyr_library()
+zephyr_library_sources(soc.c)
 zephyr_sources_ifdef(CONFIG_PM power.c)
 
 zephyr_include_directories(.)
--- a/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
@@ -27,6 +27,10 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
     zephyr_library_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
   endif()
+
+  if(CONFIG_DEVICE_CONFIGURATION_DATA)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_DCD_ENABLE=1)
+  endif()
 endif()
 
 set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/linker.ld CACHE INTERNAL "")
--- a/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
@@ -22,5 +22,10 @@ if(CONFIG_SECOND_CORE_MCUX_REMOTE_DIR)
   zephyr_include_directories(${CONFIG_SECOND_CORE_MCUX_REMOTE_DIR})
 endif()
 
+if(CONFIG_NXP_IMXRT_BOOT_HEADER)
+  if(CONFIG_BOOT_FLEXSPI_NOR)
+    zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
+  endif()
+endif()
 
 set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/linker.ld CACHE INTERNAL "")
--- a/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
+++ b/soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
@@ -25,6 +25,7 @@ endif()
 if(CONFIG_NXP_IMXRT_BOOT_HEADER)
   if(CONFIG_BOOT_FLEXSPI_NOR)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
   endif()
 endif()
 
--- a/boards/nxp/mimxrt1010_evk/CMakeLists.txt
+++ b/boards/nxp/mimxrt1010_evk/CMakeLists.txt
@@ -17,7 +17,6 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
     # This flash configuration block may need modification if another
     # flash chip is used on your custom board. See NXP AN12238 for more
     # information.
-    zephyr_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
     zephyr_library_sources(xip/evkmimxrt1010_flexspi_nor_config.c)
     zephyr_library_include_directories(xip)
   endif()
--- a/boards/nxp/mimxrt1010_evk/CMakeLists.txt
+++ b/boards/nxp/mimxrt1010_evk/CMakeLists.txt
@@ -18,7 +18,6 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
     # flash chip is used on your custom board. See NXP AN12238 for more
     # information.
     zephyr_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
-    zephyr_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
     zephyr_library_sources(xip/evkmimxrt1010_flexspi_nor_config.c)
     zephyr_library_include_directories(xip)
   endif()
--- a/boards/nxp/mimxrt1010_evk/xip/evkmimxrt1010_flexspi_nor_config.c
+++ b/boards/nxp/mimxrt1010_evk/xip/evkmimxrt1010_flexspi_nor_config.c
@@ -7,7 +7,7 @@
 
 #include "evkmimxrt1010_flexspi_nor_config.h"
 
-#if defined(XIP_BOOT_HEADER_ENABLE) && (XIP_BOOT_HEADER_ENABLE == 1)
+#if defined(CONFIG_NXP_IMXRT_BOOT_HEADER) && defined(CONFIG_BOOT_FLEXSPI_NOR)
 __attribute__((section(".boot_hdr.conf"), used))
 
 const flexspi_nor_config_t qspi_flash_config = {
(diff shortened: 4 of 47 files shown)


ZhaoxiangJin:
hello @hakehuang, could you please help to run regression test base on this PR? thanks.

hakehuang:
> hello @hakehuang, could you please help to run regression test base on this PR? thanks.

will test and feedback

boards/nxp/frdm_rw612/CMakeLists.txt
```diff
@@ -23,12 +22,3 @@ if(CONFIG_DT_HAS_NXP_ENET_MAC_ENABLED AND ${xtal32_status})
     "mutually exclusive on FRDM_RW612 due to shared PCB nets "
     "between the ethernet PHY and the external oscillator")
 endif()
-
-# Set TX power limit file to override the default one
```
MaochenWang1, line 26: could we keep this part unchanged?
It's an example showed to customer to add their own WIFI_BT_TX_PWR_LIMITS_OVERRIDE on different boards.

ZhaoxiangJin: Hello @MaochenWang1, we just moved this code to the glue layer, since this definition is only used in hal_nxp.

ZhaoxiangJin:
cc @lucien-nxp

nordicjm (dismissed):
great, thanks

nordicjm:
Although having looked into the macro function, isn't that just doing the exact same thing as zephyr_compile_definitions is and adding things globally?

ZhaoxiangJin:
> Although having looked into the macro function, isn't that just doing the exact same thing as zephyr_compile_definitions is and adding things globally?

yes, this function also add flags to global targets, changed to use `zephyr_library_compile_definitions`.

ZhaoxiangJin:
> > 1. some definitions are not used in zephyr and hal_nxp, so remove them, for examples: [commit1](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/629e76398f17693281e1f02e1be29dad6cbc89e3).
> 
> mcxn947 do have flash boot option, why remove it?

Actually through a global search, I found that this flag is not used by mcxn947. I don't know why it was added, perhaps @mmahadevan108 and @DerekSnell know something.

> > 2. some definitions are not used in hal_nxp only, so move them to hal_nxp cmake, for examples: [commit2](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/757b8e9b1afe4fbe1c158005dd4170e6543716c1), [commit3](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/24a60d462080031dec289f34dd0a3eddbc4b6ebe) and [commit4](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/cc16ec98bef9beb43ea02ac754479c493e4b8d07).
> 
> I am ok with [commit2](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/757b8e9b1afe4fbe1c158005dd4170e6543716c1)
> 
> but see below, I do not believe the BOARD_FLASH_SIZE bind with SOC is correct. they should always go with the platform.
> 
> ```
> if(CONFIG_SOC_SERIES_IMXRT10XX OR CONFIG_SOC_SERIES_IMXRT11XX)
>   if(CONFIG_NXP_IMXRT_BOOT_HEADER AND CONFIG_BOOT_FLEXSPI_NOR)
>     mcux_add_macro(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
>   endif()
> ```

Yes, you are correct, but after analysis, I found that this flag should be added in the soc level cmake (reasons can be found in commit message), the PR has been updated ([commit3](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/91e40fa60280c8cad8be2fbbeb9d8689a410586d)), please review, thanks.

> and for
> 
> ```
> if(CONFIG_SOC_SERIES_IMXRT7XX AND CONFIG_IMX_USDHC)
>   mcux_add_macro(FSL_SDK_ENABLE_DRIVER_CACHE_CONTROL=1)
> endif()
> ```
> 
> the Driver cache control is required for RT7xxx, it is not only for hal, but for shim driver as well. My concerns are thinking about if we are moving to riscv/rust solution those need a quick integration with Zephyr, and we need skip our current arm based SDK. then it is the necessary to move Marcos from hal to Zephyr native, not reverse.

This flag is added [here](https://github.com/zephyrproject-rtos/zephyr/pull/91316/files#diff-e85c6c4b9eddde2584dbea9a2dcbaacbfa81151e314c4d405a1b8e1ab48e8d17R31), after through a global search, I found that this flag is only used in hal_nxp fsl_usdhc.c and fsl_enet.c, then I think it is better add it in glue cmake. 

> > 3. some definitions are used in zephyr only, so convert to use `zephyr_library_compile_definitions`, for examples: [commit8](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/438aef826e7e8e03f3194354bc2b38d56b8c2295).
> 
> I would encourage to reuse the zephyr defines as possible, instead of defining another one, and translate them in hal, remember the policy is `Zephyr first`.

`zephyr_library_compile_definitions` is a zephyr function, it is better than `zephyr_compile_definitions` because `zephyr_library_compile_definitions` passes the flag only to the target that requires it.

> > 4. some definitions are used in zephyr and hal_nxp, but according [Jmie](https://github.com/zephyrproject-rtos/zephyr/pull/95705#discussion_r2343840196)'s suggestion, uses  `zephyr_library_compile_definitions` is a better way, so in zephyr side we use `zephyr_library_compile_definitions`, in hal_nxp side we use `mcux_add_macro `, for examples: [commit5](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/4225b7b43c601846d5fb9e273a9b8c4ce569884f), [commit6](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/e35595f98ecf1791e83db2c209cd976962c5ff42) and [commit7](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/235870428e98b9d4b53d591f687497657e2cc59c).
> 
> I would suggest that we clarify with [Jmie] on this, as those are required for NXP soc and its hal, so a global Marco is good in this case. Adding a translation layer is very prune to error. we can just add a comment in those cmakefile to clarify this.

Currently most flags are passed in the soc level cmake, could you please help to review again? then we can continue discuss, thanks.

DerekSnell:
> > > 1. some definitions are not used in zephyr and hal_nxp, so remove them, for examples: [commit1](https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/629e76398f17693281e1f02e1be29dad6cbc89e3).
> > 
> > 
> > mcxn947 do have flash boot option, why remove it?
> 
> Actually through a global search, I found that this flag is not used by mcxn947. I don't know why it was added, perhaps @mmahadevan108 and @DerekSnell know something.
> 
Hi @ZhaoxiangJin and @hakehuang ,
I see you are right, and the flag `XIP_BOOT_HEADER_ENABLE` does not need to be set for the MCX devices.  Like you, I do not see this flag used anywhere in the HAL for those devices.  And I just tested on the board variant `frdm_mcxn947//cpu0/qspi` and it boots without that flag.

I see that flag is used on the RT1xxx devices, so I suspect it was copied from them when QSPI boot was enabled on this platform.  But we can remove it.  Thanks

soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ if(CONFIG_MEMC_MCUX_FLEXSPI)
   endif()
 endif()
 
+if(CONFIG_NXP_IMXRT_BOOT_HEADER)
+  if(CONFIG_BOOT_FLEXSPI_NOR)
+    zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
```
nordicjm, line 28: where is the zephyr library? This isn't doing anything, all this is doing is the exact same thing of adding a global define to every file, this is not fixed. Open the `build.ninja` file and search, this is applied to:
```
/tmp/aa/zephyr/lib/libc/validate_libc.c
/tmp/aa/zephyr/lib/heap/heap.c
/tmp/aa/zephyr/lib/os/cbprintf_packaged.c
/tmp/aa/zephyr/lib/os/clock.c
/tmp/aa/zephyr/lib/os/printk.c
/tmp/aa/zephyr/lib/os/sem.c
/tmp/aa/zephyr/lib/os/thread_entry.c
/tmp/aa/zephyr/lib/os/cbprintf_complete.c
/tmp/aa/zephyr/lib/os/assert.c
/tmp/aa/zephyr/lib/utils/dec.c
/tmp/aa/zephyr/lib/utils/hex.c
/tmp/aa/zephyr/lib/utils/rb.c
/tmp/aa/zephyr/lib/utils/timeutil.c
/tmp/aa/zephyr/lib/utils/bitarray.c
/tmp/aa/zephyr/lib/utils/bitmask.c
/tmp/aa/zephyr/lib/utils/getopt/getopt.c
/tmp/aa/zephyr/lib/utils/getopt/getopt_common.c
/tmp/aa/zephyr/lib/utils/ring_buffer.c
/tmp/aa/zephyr/lib/utils/last_section_id.c
/tmp/aa/zephyr/subsys/mem_mgmt/mem_attr.c
/tmp/aa/zephyr/subsys/tracing/tracing_none.c

/tmp/aa/zephyr/soc/nxp/imxrt/imxrt10xx/soc.c
/tmp/aa/zephyr/soc/nxp/imxrt/mpu_regions.c
```
the bottom 2 might be valid for it, the others are not

If you want an example of how to do this properly then you can look at https://github.com/zephyrproject-rtos/zephyr/pull/101350 build hello_world for e.g. nrf52840dk/nrf52840 and look in the build.ninja file, then search for `-DNRF` and see how the defines needed by HAL are applied directly to files in nrfx only, the defines applied to main.c are clean in that they are Zephyr-applied defines only `-DKERNEL -DK_HEAP_MEM_POOL_SIZE=0 -DPICOLIBC_LONG_LONG_PRINTF_SCANF -DZVFS_OPEN_SIZE=0 -D_POSIX_THREAD_SAFE_FUNCTIONS=200809L -D__LINUX_ERRNO_EXTENSIONS__ -D__PROGRAM_START -D__ZEPHYR__=1`

soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ endif()
 if(CONFIG_NXP_IMXRT_BOOT_HEADER)
   if(CONFIG_BOOT_FLEXSPI_NOR)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
```
nordicjm, line 28: same with this

soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
   if(CONFIG_BOOT_FLEXSPI_NOR)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
   endif()
+
+  if(CONFIG_DEVICE_CONFIGURATION_DATA)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_DCD_ENABLE=1)
```
nordicjm, line 32: same with this

soc/nxp/imxrt/imxrt11xx/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ if(CONFIG_NXP_IMXRT_BOOT_HEADER)
     zephyr_library_compile_definitions(BOARD_FLASH_SIZE=CONFIG_FLASH_SIZE*1024)
     zephyr_library_compile_definitions(XIP_BOOT_HEADER_ENABLE=1)
   endif()
+
+  if(CONFIG_DEVICE_CONFIGURATION_DATA)
+    zephyr_library_compile_definitions(XIP_BOOT_HEADER_DCD_ENABLE=1)
```
nordicjm, line 32: and this

nordicjm (changes requested):

nordicjm (dismissed):

soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
nordicjm: Not addressed

nordicjm (changes requested):

soc/nxp/imxrt/imxrt10xx/CMakeLists.txt
ZhaoxiangJin: fixed actually, see https://github.com/zephyrproject-rtos/zephyr/pull/100252/changes/c484e026bacbed2ff72590abb209e514f0ac4693.
image

nordicjm: Not fixed as per https://github.com/zephyrproject-rtos/zephyr/pull/100252/commits/f77ee7f4ca93e7e623cb8aa47932a71db3f87b7a fix it properly without a fixup commit later on as per contribution guidelines

ZhaoxiangJin: moved https://github.com/zephyrproject-rtos/zephyr/pull/100252/changes/8861860d3628ba2b711d61d85c25a662c89ec3aa to the first one
ping @nordicjm

nordicjm (approved):

dnltz (approved):

dnltz:
@mmahadevan108 can you please take a look at this PR?

iuliana-prodan (approved):

PR #95637, opened 2025-09-08
Add support for Versal Gen 2 APU (Cortex a78 processor)

kedareswararao:
This pull request adds support for Versal Gen 2 APU which based on the ARM A78 processor.

PR #100254, opened 2025-11-30
Add support for Versal APU (Cortex a72 processor)

kedareswararao:
This pull request adds support for Versal APU which based on the ARM A72 processor.

### DIFF
--- a/boards/amd/versal_apu/board.cmake
+++ b/boards/amd/versal_apu/board.cmake
@@ -1,0 +1,18 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+if(CONFIG_BUILD_WITH_TFA)
+  set(TFA_PLAT "versal")
+  # Add Versal specific TF-A build parameters
+  set(TFA_EXTRA_ARGS "TFA_NO_PM=1;PRELOADED_BL33_BASE=${CONFIG_SRAM_BASE_ADDRESS}")
+  if(CONFIG_TFA_MAKE_BUILD_TYPE_DEBUG)
+    set(BUILD_FOLDER "debug")
+  else()
+    set(BUILD_FOLDER "release")
+  endif()
+  set(XSDB_BL31_PATH ${PROJECT_BINARY_DIR}/../tfa/versal/${BUILD_FOLDER}/bl31/bl31.elf)
+  board_runner_args(xsdb "--bl31=${XSDB_BL31_PATH}")
+endif()
+
+include(${ZEPHYR_BASE}/boards/common/xsdb.board.cmake)
--- a/boards/amd/versal_apu/doc/index.rst
+++ b/boards/amd/versal_apu/doc/index.rst
@@ -1,0 +1,83 @@
+.. zephyr:board:: versal_apu
+
+Overview
+********
+This configuration provides support for the APU (Application Processing Unit) on AMD
+Versal devices. The APU can operate as follows:
+
+* Two independent Cortex-A72 cores
+
+This processing unit is based on an ARM Cortex-A72 CPU, and it enables the following devices:
+
+* ARM GIC v3 Interrupt Controller
+* ARMv8 Generic Timer
+* SBSA UART
+
+Hardware
+********
+Supported Features
+==================
+
+.. zephyr:board-supported-hw::
+
+Devices
+=======
+System Timer
+------------
+
+This board configuration uses a system timer tick frequency of 100 MHz.
+
+Serial Port
+-----------
+
+This board configuration uses a single serial communication channel with the
+on-chip UART0.
+
+Memories
+--------
+
+Although Flash, DDR and OCM memory regions are defined in the DTS file,
+all the code plus data of the application will be loaded in the sram0 region,
+which points to the DDR memory. The OCM memory area is currently available
+for usage, although nothing is placed there by default.
+
+Known Problems or Limitations
+=============================
+
+The following platform features are unsupported:
+
+* Only the first CPU in the APU subsystem is supported.
+
+Programming and Debugging
+*************************
+
+.. zephyr:board-supported-runners::
+
+For deployment on real Versal hardware, XSDB and a PDI file are required. The PDI file contains the hardware initialization and boot configuration needed for the physical device.
+
+Build the application:
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/hello_world
+   :board: versal_apu
+   :goals: build
+
+Flash to real hardware with PDI file:
+
+.. code-block:: console
+
+   west flash --runner xsdb --pdi /path/to/your.pdi --bl31 /path/to/your_bl31.elf
+
+You should see the following message on the console:
+
+.. code-block:: console
+
+   Hello World! versal_apu/versal_apu
+
+
+References
+**********
+
+1. ARMv8-A Architecture Reference Manual (ARM DDI 0487)
+2. Arm Cortex-A72 Core Technical Reference Manual (Doc ID 100095)
+3. AMD Versal Technical Reference Manual
--- a/boards/amd/versal_apu/versal_apu.dts
+++ b/boards/amd/versal_apu/versal_apu.dts
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/dts-v1/;
+#include <arm64/xilinx/versal_a72.dtsi>
+
+/ {
+	compatible = "xlnx,versal_apu";
+
+	chosen {
+		zephyr,sram = &sram0;
+		zephyr,console = &uart0;
+		zephyr,shell-uart = &uart0;
+	};
+};
+
+&cpu0 {
+	clock-frequency = <100000000>;
+};
+
+&soc {
+	sram0: memory@0 {
+		compatible = "mmio-sram";
+		reg = <0x0 0x0 0x0 DT_SIZE_G(2)>;
+	};
+};
+
+&uart0 {
+	status = "okay";
+	current-speed = <115200>;
+	clock-frequency = <100000000>;
+};
+
+&uart1 {
+	status = "okay";
+	current-speed = <115200>;
+	clock-frequency = <100000000>;
+};
--- a/dts/arm64/xilinx/versal_a72.dtsi
+++ b/dts/arm64/xilinx/versal_a72.dtsi
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/dts-v1/;
+
+#include <mem.h>
+#include <arm64/armv8-a.dtsi>
+#include <zephyr/dt-bindings/interrupt-controller/arm-gic.h>
+#include <amd/versal_apu.dtsi>
+
+/ {
+	model = "AMD Versal APU";
+	#address-cells = <2>;
+	#size-cells = <2>;
+
+	cpus {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		cpu0: cpu@0 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a72";
+			reg = <0x0>;
+			enable-method = "psci";
+		};
+	};
+
+	psci {
+		compatible = "arm,psci-0.2", "arm,psci";
+		method = "smc";
+	};
+
+	timer {
+		compatible = "arm,armv8-timer";
+		interrupt-parent = <&gic>;
+		interrupts = <GIC_PPI 13 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>,
+			     <GIC_PPI 14 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>,
+			     <GIC_PPI 11 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>,
+			     <GIC_PPI 10 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
+	};
+
+	soc {
+		interrupt-parent = <&gic>;
+
+		gic: interrupt-controller@f9000000 {
+			compatible = "arm,gic-v3", "arm,gic";
+			reg = <0x0 0xf9000000 0x0 0x80000>,
+			      <0x0 0xf9080000 0x0 0x80000>;
+			interrupt-controller;
+			#interrupt-cells = <4>;
+		};
+	};
+};
--- a/soc/xlnx/versal/CMakeLists.txt
+++ b/soc/xlnx/versal/CMakeLists.txt
@@ -4,10 +4,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-zephyr_sources(soc.c)
+zephyr_sources_ifdef(CONFIG_SOC_VERSAL_RPU soc.c)
+zephyr_sources_ifdef(CONFIG_ARM_MMU arm_mmu_regions.c)
 
 zephyr_include_directories(.)
 
 if(CONFIG_SOC_VERSAL_RPU)
   set(SOC_LINKER_SCRIPT ${ZEPHYR_BASE}/include/zephyr/arch/arm/cortex_a_r/scripts/linker.ld CACHE INTERNAL "")
+endif()
+
+if(CONFIG_SOC_VERSAL_APU)
+  # Use generic ARM64 linker script
+  set(SOC_LINKER_SCRIPT ${ZEPHYR_BASE}/include/zephyr/arch/arm64/scripts/linker.ld CACHE INTERNAL "")
 endif()
--- a/soc/xlnx/versal/Kconfig.soc
+++ b/soc/xlnx/versal/Kconfig.soc
@@ -4,10 +4,24 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+config SOC_AMD_VERSAL
+	bool
+
 config SOC_VERSAL_RPU
 	bool
+	select SOC_AMD_VERSAL
 	help
 	  AMD Versal RPU
 
+config SOC_VERSAL_APU
+	bool
+	select SOC_AMD_VERSAL
+	help
+	  AMD Versal APU
+
+config SOC_FAMILY
+	default "amd_versal" if SOC_AMD_VERSAL
+
 config SOC
 	default "versal_rpu" if SOC_VERSAL_RPU
+	default "versal_apu" if SOC_VERSAL_APU
--- a/boards/amd/versal_apu/Kconfig.defconfig
+++ b/boards/amd/versal_apu/Kconfig.defconfig
@@ -1,0 +1,9 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+if BOARD_VERSAL_APU
+
+config BUILD_OUTPUT_BIN
+	default y
+
+endif # BOARD_VERSAL_APU
--- a/boards/amd/versal_apu/Kconfig.versal_apu
+++ b/boards/amd/versal_apu/Kconfig.versal_apu
@@ -1,0 +1,5 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+config BOARD_VERSAL_APU
+	select SOC_VERSAL_APU
--- a/boards/amd/versal_apu/board.yml
+++ b/boards/amd/versal_apu/board.yml
@@ -1,0 +1,9 @@
+# Copyright (c) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+board:
+  name: versal_apu
+  full_name: AMD Versal APU Development Board
+  vendor: amd
+  socs:
+  - name: versal_apu
--- a/boards/amd/versal_apu/support/xsdb.cfg
+++ b/boards/amd/versal_apu/support/xsdb.cfg
@@ -1,0 +1,28 @@
+# Copyright (c) 2025 Advanced Micro Devices, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+proc load_image args  {
+	set elf_file [lindex $args 0]
+	set pdi_file [lindex $args 1]
+	set bl31_file [lindex $args 2]
+
+	if { [info exists ::env(HW_SERVER_URL)] } {
+		connect -url $::env(HW_SERVER_URL)
+	} else {
+		connect
+	}
+
+	device program $pdi_file
+	targets -set -nocase -filter {name =~ "Versal*"}
+	after 100
+	targets -set -nocase -filter {name =~ "*Cortex-A72 #0"}
+	rst -proc
+	after 100
+	dow -force $elf_file
+	dow -force $bl31_file
+	con
+	exit
+}
+
+load_image {*}$argv
--- a/boards/amd/versal_apu/versal_apu.yaml
+++ b/boards/amd/versal_apu/versal_apu.yaml
@@ -1,0 +1,10 @@
+identifier: versal_apu
+name: AMD Development board for Versal APU
+arch: arm64
+toolchain:
+  - zephyr
+testing:
+  ignore_tags:
+    - net
+    - bluetooth
+vendor: amd
--- a/boards/amd/versal_apu/versal_apu_defconfig
+++ b/boards/amd/versal_apu/versal_apu_defconfig
@@ -1,0 +1,22 @@
+# The Zephyr build from this defconfig is expected to boot from
+# Xilinx Arm Trusted Firmware (ATF).
+# Boot Flow is: Boot PDI -> TF-A -> Zephyr
+
+CONFIG_BUILD_WITH_TFA=y
+
+CONFIG_ARM64_VA_BITS_40=y
+CONFIG_ARM64_PA_BITS_40=y
+CONFIG_TIMER_READS_ITS_FREQUENCY_AT_RUNTIME=y
+
+# Enable UART driver
+CONFIG_SERIAL=y
+
+# Enable console
+CONFIG_CONSOLE=y
+CONFIG_UART_CONSOLE=y
+
+# Enable serial port
+CONFIG_UART_PL011=y
+
+# This should be commented in order to test at EL1 S (EL1 Secure)
+CONFIG_ARMV8_A_NS=y
--- a/dts/vendor/amd/versal_apu.dtsi
+++ b/dts/vendor/amd/versal_apu.dtsi
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/ {
+	soc: soc {
+		#address-cells = <2>;
+		#size-cells = <2>;
+
+		/*
+		 * OCM (On-Chip Memory) is used by TF-A (Trusted Firmware-A) by default.
+		 * Enable this node only if TF-A is not using this memory region or if
+		 * explicit coordination is established between TF-A and Zephyr.
+		 */
+		ocm: memory@fffc0000 {
+			compatible = "zephyr,memory-region";
+			reg = <0x0 0xfffc0000 0x0 DT_SIZE_K(256)>;
+			status = "disabled";
+			zephyr,memory-region = "OCM";
+		};
+
+		uart0: uart@ff000000 {
+			compatible = "arm,sbsa-uart";
+			reg = <0x0 0xff000000 0x0 DT_SIZE_K(64)>;
+			status = "disabled";
+			interrupt-names = "irq_0";
+			interrupts = <GIC_SPI 18 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
+		};
+
+		uart1: uart@ff010000 {
+			compatible = "arm,sbsa-uart";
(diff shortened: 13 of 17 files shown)


soc/xlnx/versal/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+#
+
+zephyr_sources_ifdef(
+  CONFIG_ARM_MMU
+  arm_mmu_regions.c
+)
+
+zephyr_include_directories(.)
+
+if(CONFIG_SOC_AMD_VERSAL_APU)
```
dsseng: This does not seem to be absolutely necessary - this directory is only included when enabling this SoC. Or is this port supposed to support other cores, like a secondary AArch32 core?

dts/arm64/xilinx/versal_a72.dtsi
```diff
@@ -0,0 +1,63 @@
...
+			compatible = "arm,cortex-a72";
+			reg = <0>;
+			enable-method = "psci";
+		};
+
+		cpu1: cpu@1 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a72";
+			enable-method = "psci";
+			reg = <0x1>;
```
dsseng: It would be better to make core declarations look uniform:

```suggestion
		cpu0: cpu@0 {
			device_type = "cpu";
			compatible = "arm,cortex-a72";
			reg = <0x0>;
			enable-method = "psci";
		};

		cpu1: cpu@1 {
			device_type = "cpu";
			compatible = "arm,cortex-a72";
			reg = <0x1>;
			enable-method = "psci";
```

boards/amd/versal_apu/versal_apu.dts
```diff
@@ -0,0 +1,39 @@
...
+	chosen {
+		zephyr,sram = &sram0;
+		zephyr,console = &uart0;
+		zephyr,shell-uart = &uart0;
+	};
+};
+
+&cpu0 {
+	clock-frequency = <100000000>;
+};
```
dsseng, line 22: Should CPU1 frequency be also specified? (it's my assumption the cores are symmetrical, not sure if they have the same frequency)

```suggestion
&cpu0 {
	clock-frequency = <100000000>;
};

&cpu1 {
	clock-frequency = <100000000>;
};
```

boards/amd/versal_apu/versal_apu.dts
```diff
@@ -0,0 +1,39 @@
...
+
+&soc {
+	sram0: memory@0 {
+		compatible = "mmio-sram";
+		reg = <0x0 0x0 0x0 0x7ffc0000>;
```
dsseng: Could this maybe be expressed in DT_SIZE_[G,M,K] terms? I am not sure whether that would be easier to read, so feel free to disregard this suggestion

boards/amd/versal_apu/doc/index.rst
```diff
@@ -0,0 +1,107 @@
...
+all the code plus data of the application will be loaded in the sram0 region,
+which points to the DDR memory. The OCM memory area is currently available
+for usage, although nothing is placed there by default.
+
+Known Problems or Limitations
+==============================
+
+The following platform features are unsupported:
+
+* Only the first CPU in the APU subsystem is supported.
```
dsseng, line 49: Should cpu1 have `status = "disabled"` in the DT then?

dsseng (commented):
Thanks for this PR, please find some suggestions, feel free to disagree with the proposed changes.

soc/xlnx/versal/CMakeLists.txt
kedareswararao: This soc has RPU as well will add that support on top of this that's why kept this if check

dts/arm64/xilinx/versal_a72.dtsi
kedareswararao: Sure will update

boards/amd/versal_apu/versal_apu.dts
kedareswararao: will remove cpu1 all together as we are supporting only cpu0 as of now

boards/amd/versal_apu/versal_apu.dts
kedareswararao: sure will update

boards/amd/versal_apu/doc/index.rst
kedareswararao: will remove cpu1 all together

soc/xlnx/versal/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2025 Advanced Micro Devices, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+zephyr_sources_ifdef(
+  CONFIG_ARM_MMU
+  arm_mmu_regions.c
+)
```
nordicjm:
```suggestion
zephyr_sources_ifdef(CONFIG_ARM_MMU arm_mmu_regions.c)
```

boards/amd/versal_apu/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+# Check if the board-specific qemu.dts file exists
+if(EXISTS "${BOARD_DIR}/${BOARD}-qemu.dts")
+  # Ensure DTC executable is available
+  if(DTC_FOUND)
+    set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
+      COMMAND ${DTC} -I dts -O dtb -q "${BOARD_DIR}/${BOARD}-qemu.dts" -o "${PROJECT_BINARY_DIR}/${BOARD}-qemu.dtb"
+    )
+  else()
+    message(FATAL_ERROR "DTC not found, but required for compiling ${BOARD}-qemu.dts")
```
nordicjm: dts? Because dtc is above with required flag

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4423 @@
...
+	};
+
+	memory@8_0000_0000 {
+		compatible = "qemu:memory-region";
+		device_type = "memory";
+		container = <0x09>;
+		phandle = <0x8c>;
+	};
+
+	memory@0x50000000000ULL {
```
nordicjm: this looks to be 3 completely different ways of putting in addresses, plus is using weird compatibles? We have other qemu boards which do not use a "qemu:" tag in the compatible, and I see no in tree compatible for this

nordicjm (changes requested):

soc/xlnx/versal/CMakeLists.txt
kedareswararao: Sure will fix

boards/amd/versal_apu/CMakeLists.txt
kedareswararao: Thanks for the review. The DTS file is already present in the board folder, but it still requires compilation using the dtc tool. The error being reported is simply indicating that this dtc compilation step is needed. Let me know if you'd like me to add anything further on this.

boards/amd/versal_apu/versal_apu-qemu.dts
kedareswararao: These DTS definitions, including the memory nodes, are consumed by the QEMU model. The compatible string is also used by QEMU to identify and configure the platform accordingly. please let me know if you need any additional clarification

nordicjm: @mbolivar please review

figlesia-xilinx: Hi,

Above are AMD/Xilinx QEMU specific nodes instantiating **QEMU memory region container** nodes that server similar functionality as **simple-bus** nodes in the Device tree specification and act as resource containers (but unlike simple-bus nodes these memory container nodes are not automatically mapped into any CPU address space). That is, this is part of the **Versal virtual machine** instantiated by the AMD/Xilinx QEMU.

Best regards,
Francisco

soc/xlnx/versal/CMakeLists.txt
dsseng: Thank you for explaining, this makes sense now

boards/amd/versal_apu/doc/index.rst
```diff
@@ -0,0 +1,107 @@
...
+Memories
+--------
+
+Although Flash, DDR and OCM memory regions are defined in the DTS file,
+all the code plus data of the application will be loaded in the sram0 region,
+which points to the DDR memory. The OCM memory area is currently available
+for usage, although nothing is placed there by default.
+
+Known Problems or Limitations
+==============================
```
michalsimek: remove one = from line above

soc/xlnx/versal/Kconfig.soc
```diff
@@ -4,10 +4,24 @@
...
+	select SOC_AMD_VERSAL
+	help
+	  AMD Versal APU
+
+config SOC_FAMILY
+	default "amd_versal" if SOC_AMD_VERSALNET
+
 config SOC
 	default "versal_rpu" if SOC_VERSAL_RPU
+	default "versal_apu" if SOC_AMD_VERSAL_APU
```
michalsimek: Can you please create one more patch to do s/SOC_VERSAL_RPU/SOC_AMD_VERSAL_RPU/ ?

michalsimek (commented):
I have run it on vek280 and also on qemu target. Both targets works fine. 
I think it would be worth to also wire TF-A build for this target because based on your configuration BL33 has to start at 0 which is not current default setup. 
Anyway with custom TF-A build not a problem to run it on the target.
That's why please fix that doc fix. At best one more patch (sed for RPU) and TF-A wiring. 
Only doc fix is must. After that I will approve.

dts/arm64/xilinx/versal_a72.dtsi
```diff
@@ -0,0 +1,56 @@
...
+
+		gic: interrupt-controller@f9000000 {
+			compatible = "arm,gic-v3", "arm,gic";
+			reg = <0x0 0xf9000000 0x0 0x80000>,
+			      <0x0 0xf9080000 0x0 0x80000>;
+			interrupt-controller;
+			#interrupt-cells = <4>;
+		};
+
+		timer {
```
michalsimek: Actually one more change request.  This should get out of SOC bus. 
/tmp/versal_apu/zephyr/zephyr.dts:70.9-77.5: Warning (simple_bus_reg): /soc/timer: missing or empty reg/ranges property

Move it to root to avoid this warning.

boards/amd/versal_apu/doc/index.rst
kedareswararao: Sure will fix

soc/xlnx/versal/Kconfig.soc
kedareswararao: Sure will create

dts/arm64/xilinx/versal_a72.dtsi
kedareswararao: will fix

kedareswararao:
> I have run it on vek280 and also on qemu target. Both targets works fine. I think it would be worth to also wire TF-A build for this target because based on your configuration BL33 has to start at 0 which is not current default setup. Anyway with custom TF-A build not a problem to run it on the target. That's why please fix that doc fix. At best one more patch (sed for RPU) and TF-A wiring. Only doc fix is must. After that I will approve.

Sure will integrate the TF-A and will send updated pull request

soc/xlnx/versal/Kconfig.soc
nordicjm: can you not suggest making invalid changes? Name and value must match as is listed on https://docs.zephyrproject.org/latest/hardware/porting/soc_porting.html#write-kconfig-files

kedareswararao: Thanks for the review okay will make it uniform

boards/amd/versal_apu/board.cmake
```diff
@@ -15,6 +15,18 @@ set(QEMU_FLAGS_${ARCH}
...
+  set(TFA_PLAT "versal")
+  # Add Versal specific TF-A build parameters
+  set(TFA_EXTRA_ARGS "TFA_NO_PM=1;PRELOADED_BL33_BASE=${CONFIG_SRAM_BASE_ADDRESS}")
+  if(CONFIG_TFA_MAKE_BUILD_TYPE_DEBUG)
+    set(BUILD_FOLDER "debug")
+  else()
+    set(BUILD_FOLDER "release")
+  endif()
+endif()
+
```
michalsimek, line 17: I think we should also do this to help everybody to just pass pdi for the board but reuse and automatically wire TF-A which has been built that people don't need to think about it.

diff --git a/boards/amd/versal_apu/board.cmake b/boards/amd/versal_apu/board.cmake
index 301e251d50a6..aea2bf738fd6 100644
--- a/boards/amd/versal_apu/board.cmake
+++ b/boards/amd/versal_apu/board.cmake
@@ -2,7 +2,6 @@
 #
 
-include(${ZEPHYR_BASE}/boards/common/xsdb.board.cmake)
 set(SUPPORTED_EMU_PLATFORMS qemu)
 set(QEMU_ARCH xilinx-aarch64)
 set(QEMU_CPU_TYPE_${ARCH} cortexa72)
@@ -25,8 +24,13 @@ if(CONFIG_BUILD_WITH_TFA)
   else()
     set(BUILD_FOLDER "release")
   endif()
+
+  set(XSDB_BL31_PATH ${CMAKE_BINARY_DIR}/tfa/${TFA_PLAT}/${BUILD_FOLDER}/bl31/bl31.elf)
+  board_runner_args(xsdb "--bl31=${XSDB_BL31_PATH}")
 endif()
 
+include(${ZEPHYR_BASE}/boards/common/xsdb.board.cmake)
+
 set(QEMU_KERNEL_OPTION
     "-device;loader,cpu-num=0,file=\$<TARGET_FILE:\${logical_target_for_zephyr_elf}>"
 )

kedareswararao: Thanks for the review updated

michalsimek (dismissed):
Tested on QEMU and also on VEK280 and both are working fine.

kedareswararao:
@nordicjm: Addressed review comments please review
@henrikbrixandersen @ibirnbaum please review

boards/amd/versal_apu/versal_apu-qemu.dts
henrikbrixandersen, line 1: Same comment as https://github.com/zephyrproject-rtos/zephyr/pull/95637#pullrequestreview-3599583077

kedareswararao: @henrikbrixandersen: As suggested in the other pull request Added detailed comment about the usage of this file please review when you have sometime.

boards/amd/versal_apu/versal_apu.dts
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2025, Advanced Micro Devices, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/dts-v1/;
+#include <arm64/xilinx/versal_a72.dtsi>
+
+/ {
+	chosen {
```
kylebonnici, line 13: Please add `compatible` as this required as per DTS Spec table 3.1

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+
+	ddr_4@0x60000000000ULL {
+		compatible = "qemu:memory-region";
+		container = <0x8d>;
+		qemu,ram = <0x01>;
+		reg = <0x600 0x00 0x01 0x00 0x00>;
+		phandle = <0x134>;
+	};
+
+	__symbols__ {
```
kylebonnici: As per Spec 2.2.1 "The node-name shall start with a lower or upper case character..."

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+ * - Keep this file synchronized with QEMU's AMD Versal machine model
+ * - Do not add Zephyr-specific properties here
+ * - For Zephyr driver configuration, modify versal_apu.dts instead
+ *
+ */
+
+/dts-v1/;
+
+/ {
+	#address-cells = <0x02>;
```
kylebonnici: `compatible` and `model` properties are required see DTS Spec Table 3.1

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+	usb0_ma {
+		doc-ignore = <0x01>;
+		compatible = "qemu:memory-transaction-attr";
+		secure = <0x00>;
+		requester-id = <0x230>;
+		phandle = <0x23>;
+	};
+
+	amba_root@0 {
+		#address-cells = <0x02>;
```
kylebonnici: Given that node has address `...@0` then the node must have a reg property. See DTS Spec section 2.2.1 
"...If the node has no reg property , the @unit-address must be omitted..."

This applies to a number of nodes in this file but I will only comment here

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+			#size-cells = <0x02>;
+			#priority-cells = <0x01>;
+			compatible = "simple-bus";
+			ranges;
+			phandle = <0x09>;
+
+			downstream_amba_lpd {
+				compatible = "qemu:memory-region";
+				alias = <0x06>;
+				reg = <0x00 0x00 0xffffffff 0xffffffff 0xffffffff>;
```
kylebonnici: Given that node does not have address  i.e. no ...@... then reg property must be omitted. See DTS Spec section 2.2.1
"...If the node has no reg property , the @unit-address must be omitted..."

This applies to a number of nodes in this file but I will only comment here

kylebonnici: DTS Spec 2.3.6

"Property name: reg
Property value: encoded as an arbitrary number of(address, length) pairs." Where address and size are defined by parent #address-cells and #size-cells respectively.

With the above therefore the reg can only have (2, 2) hence IMO `reg = <0x00 0x00 0xffffffff 0xffffffff 0xffffffff>;` is in violation of the spec.

This applies to a number of nodes in this file but I will only comment here

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+			};
+
+			xppu_lpd@0xff990000 {
+				compatible = "xlnx,versal-xppu";
+				reg-extended = <0x06 0x00 0xff990000 0x00 0x10000 0x00 0x16 0x00
+						0xff000000 0x00 0xe00000 0x02 0x16 0x00 0xfe000000
+						0x00 0x1000000 0x02 0x09 0x00 0xe0000000 0x00
+						0x10000000 0x02>;
+				mr = <0x06>;
+				interrupts = <0x33>;
```
kylebonnici: AFAIK interrupt controller is found by defining `interrupt-parent`. when this is not defined AFAIK in zephyr the compiler will recursive walking up the tree until an ancestor with `interrupt controller ` is found.

I cannot see any ancestor with such a property hence IMO no interrupt-parent will be found. IMO either `interrupts = <0x33>;` is unused as this need parent to determine the `#interrupt-cells` or parent is missing.

this applies to multiple cases but In this file but I will only comment on this one.

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+				xlnx,intc-positive = <0xffff>;
+				xlnx,intc-use-ext-intr = <0x01>;
+				phandle = <0x7d>;
+			};
+
+			ddrmc0_gpo@10 {
+				#gpio-cells = <0x01>;
+				gpio-controller;
+				compatible = "xlnx,io-gpo-1.02.a\0xlnx,io_gpo";
+				reg = <0x00 0x1b010 0x04>;
```
kylebonnici: `reg` address must match node name as per DTS Spec section. 2.2.1 " The unit-address must match the first address specified in the reg property
of the node"

there are a number of these issue in the this file I will only be commenting on this one

boards/amd/versal_apu/versal_apu-qemu.dts
```diff
@@ -0,0 +1,4453 @@
...
+		compatible = "qemu:memory-region";
+		device_type = "memory";
+		container = <0x09>;
+		phandle = <0x8c>;
+	};
+
+	memory@0x50000000000ULL {
+		compatible = "qemu:memory-region";
+		device_type = "memory";
+		container = <0x09>;
```
kylebonnici: I am not sure if this property is allowed at all as it is not in the part of the standard properties + it is not part of the memory node as defined in the DTS spec section 3.4

kylebonnici (changes requested):

boards/amd/versal_apu/versal_apu-qemu.dts
kylebonnici: `0x`  has no meaning as all address are all hex even If spec is not clear on this. 
 `ULL` have no meaning IMO as the address is of what ever size `#address_cell` is set to. 

IMO we should stick to one format `@FE00`.

boards/amd/versal_apu/versal_apu-qemu.dts
michalsimek: This DT is for QEMU emulation as it is described in note which has been requested by @henrikbrixandersen .
yes this file has a lot of issues but it's only purpose is to provide a way for zephyr to be tried and tested on qemu without need to have HW available. And source format is IMHO better then dtb blob. If you insist to fix all these issue you have reported likely we will just remove the whole QEMU support because fixes have to happen in QEMU AMD repo.

kylebonnici: I am not requesting this. 'Zephyr' could consider to make an official exception and state that `compatible ` in root not is not mandatory as required by the DTS spec .AFAIK most (if not all) of the root node `compatible ` are never linked to any bindings yml.

But if we are to abide by the spec then we must follow it IMO

michalsimek: It is not about Zephyr consideration. Zephyr can decide not to have these files in the tree. DT spec/style in this file is something what it is related to AMD Qemu wiring only. They should tell how that DT/configuration should look like. And as of today that's current reality how that files look like for the whole SOC emulation.

kylebonnici: Does this mean that AMD Qemu dose not follow in full https://devicetree-specification.readthedocs.io/en/latest/?

nordicjm: yeah I'm very surprised to read this, wasn't DTS initially created *by* xilinux (AMD) and now you're not even following the specifications you yourselves created?

figlesia-xilinx: Hi,

The AMD/Xilinx QEMU Hardware Devicetrees (HWDTBs) extend the Devicetree Specification to describe heterogeneous SoCs with multiple CPU clusters. The target use case is to describe a virtual platform model of an SoC that can be dynamically (runtime) instantiated by AMD/Xilinx QEMU. From Zephyr perspective this is QEMU configuration.

Best regards

kylebonnici: Where can one find the specs on how `AMD/Xilinx QEMU Hardware Devicetrees (HWDTBs) extend the Devicetree Specification`

figlesia-xilinx: Hi Kyle,

There is currently no public specification for the QEMU HWDTB. For the most accurate understanding, I recommend reviewing the [qemu-devicetrees](https://github.com/Xilinx/qemu-devicetrees) repository along with the [AMD/Xilinx QEMU](https://github.com/Xilinx/qemu) source, which has supported DTB-based configuration for over a decade. Please note, however, that from Zephyr’s perspective (as mentioned in the note), this DTS is simply QEMU configuration.

Best regards

kylebonnici: @figlesia-xilinx can you share if  QEMU HWDTB extends (adds more rules on top of) the DTS Spec 0.4 version or does it go back on some of the expectations set by DTS 0.4 spec + add more rules on top of DTS Spec 0.4? I.e making  QEMU HWDTB spec not compatible with the DTS 0.4 spec?

Can you give examples from my above comments of Spec violations in these files  as per the DTS Spec 0.4 that are not violation in the  QEMU HWDTB spec?

michalsimek: Is this related to https://github.com/kylebonnici/dts-linter.git ? I finally understand connection to another patch you sent me some days/weeks ago to Linux.
One thing are rules which you are talking about and another is really binding for QEMU. 
As was said there is no public specification of QEMU HWDTBs. If this is about formatting which your tool is good at we can do formatting. Pretty much QEMU is using different properties for achieving run time selfconfiguration. Pretty much the same case as is with Zephyr/Linux differences in bindings.

boards/amd/versal_apu/versal_apu-qemu.dts
michalsimek: I think we all understand the purpose of this file. It is also clear that a lot of clean up have to happen to be easy going (and not really talking about QEMU HW spec). There is no reason for us to clean all the issues. If should be enough to run your dts-linter to clean it up, likely we don't have a problem.
If you are requesting something more than that I will ask author to remove qemu support and we will keep it only in downstream port. 
Please let me know your preference.

boards/amd/versal_apu/versal_apu-qemu.dts
kylebonnici: The PR I set to linux was about formatting. The above is not about formatting as this PR already passed CI [Compliance](https://github.com/zephyrproject-rtos/zephyr/actions/runs/20705537069/job/59435432020?pr=100254) check that does the formatting. 

The issues I have raised here are different. The are about this dts file not meeting (as per my interpretation) the DTS spec v0.4 as defined here https://devicetree-specification.readthedocs.io/en/latest/

Lets discus the three comments below and why the cannot/should not be address.

https://github.com/zephyrproject-rtos/zephyr/pull/100254#discussion_r2681511512
https://github.com/zephyrproject-rtos/zephyr/pull/100254#discussion_r2681522937
https://github.com/zephyrproject-rtos/zephyr/pull/100254#discussion_r2681552315

figlesia-xilinx: Hi Kyle,

This dts is configuration for the AMD/Xilinx QEMU not for Zephyr. That is, from QEMU perspective the comments are not issues and addressing them for meeting the v0.4 DTS spec will break the configured QEMU virtual machine. The comments are a few of the differences between AMD/Xilinx QEMU DTS and the v0.4 DTS spec required for runtime configuration of the QEMU virtual machine.

Comment 1) QEMU will try to instantiate what is placed as compatible, if chosen wrongly it will break the QEMU model
Comment 2) The node is a QEMU specific container node which requires the reg property
Comment 3) There is an interrupt-map pointing QEMU how to wire up the interrupt and to which interrupt controller

Hope this clarifies.

Best regards

kylebonnici: Yes it does help me understand the situation better.

My concern is that if we take with as an this [comment](https://github.com/zephyrproject-rtos/zephyr/pull/100254#discussion_r2681552315) as example, I am not sure if the APIs in Zephyr such as [DT_NUM_REGS](https://docs.zephyrproject.org/latest/doxygen/html/group__devicetree-reg-prop.html#ga6cdd22b6a2881b09ed63d9d262566a0a) would work as expected. 

I do not know how these QEMU Boards are used and if the [Devicetree API](https://docs.zephyrproject.org/latest/doxygen/html/group__devicetree.html) is intended to be used with these boards, and if so what the side effect could occur given these discrepancies in specs.

@nordicjm not sure who the best person to tag to pitch in on these concerns and also in a general if the Zephyr Project needs to make sure that all board in Zephyr comply with the DTS Spec or not. 

The feedback we get is also relevant for https://github.com/zephyrproject-rtos/zephyr/issues/101907

michalsimek: I think that you are still mixing two things together. You have Zephyr DT which is used by Zephyr itself (and that's follow all rules) and then you have this QEMU DT which is just configuring QEMU model itself. Zephyr never works with it, never reads it. It is just information for qemu itself to configure itself. And format for it is DT but pretty much it is just qemu configuration file. 

It means DT_NUM_REGS, etc never going to read this DT.

henrikbrixandersen: Should this QEMU configuration not live in the qemu repo, then?

michalsimek: it should be possible to push it to Zephyr SDK or be as external module. That's definitely possible.

JarmouniA (commented):
The fact that QEMU devicetree files have the same file extension as the regular devicetree files, even though they do not follow the devicetree spec, is wrong.
To keep these files in the main Zephyr repo, the file extension has to be changed. Otherwise they will have to be moved elsewhere (to a Zephyr module repo for ex. ...).

kylebonnici:
> It means it doesn't follow all the binding rules or generic rules but DTC has no problem to work with it.

IMO even if DTC works, is just a coincidence, the DTC compile has to follow some spec and hence this assumption can break any point in time if the DTC opts or will need to be more strict on the spec at some point in time in the future.

nordicjm (dismissed):

boards/amd/versal_apu/versal_apu.dts
michalsimek: Agree it should be here.

kedareswararao: Thanks for the review added compatible property now

michalsimek (approved):

nordicjm (approved):

PR #100255, opened 2025-11-30
dts: arm: st: h7: add ITCM and DTCM memory regions for STM32H7

iabdalkader:
Add ITCM and DTCM memory region definitions for the STM32H7.

ITCM: 64 KB @ 0x00000000
DTCM: 128 KB @ 0x20000000

### DIFF
--- a/dts/arm/st/h7/stm32h7.dtsi
+++ b/dts/arm/st/h7/stm32h7.dtsi
@@ -49,6 +49,18 @@
 		};
 	};
 
+	itcm: memory@0 {
+		compatible = "zephyr,memory-region", "arm,itcm";
+		reg = <0x00000000 DT_SIZE_K(64)>;
+		zephyr,memory-region = "ITCM";
+	};
+
+	dtcm: memory@20000000 {
+		compatible = "zephyr,memory-region", "arm,dtcm";
+		reg = <0x20000000 DT_SIZE_K(128)>;
+		zephyr,memory-region = "DTCM";
+	};
+
 	ext_memory: memory@90000000 {
 		compatible = "zephyr,memory-region";
 		reg = <0x90000000 DT_SIZE_M(256)>; /* max addressable area */
--- a/dts/arm/st/h7/stm32h723.dtsi
+++ b/dts/arm/st/h7/stm32h723.dtsi
@@ -232,20 +232,6 @@
 		zephyr,memory-region = "SRAM4";
 	};
 
-	/* Data TCM RAM */
-	dtcm: memory@20000000 {
-		compatible = "zephyr,memory-region", "arm,dtcm";
-		reg = <0x20000000 DT_SIZE_K(128)>;
-		zephyr,memory-region = "DTCM";
-	};
-
-	/* Instruction TCM RAM (64KB as `TCM_AXI_SHARED` is `000`) */
-	itcm: memory@0 {
-		compatible = "zephyr,memory-region", "arm,itcm";
-		reg = <0x00000000 DT_SIZE_K(64)>;
-		zephyr,memory-region = "ITCM";
-	};
-
 	otghs_fs_phy: otghs_fs_phy {
 		compatible = "usb-nop-xceiv";
 		#phy-cells = <0>;
--- a/dts/arm/st/h7/stm32h742.dtsi
+++ b/dts/arm/st/h7/stm32h742.dtsi
@@ -90,18 +90,6 @@
 		zephyr,memory-region = "SRAM4";
 	};
 
-	dtcm: memory@20000000 {
-		compatible = "zephyr,memory-region", "arm,dtcm";
-		reg = <0x20000000 DT_SIZE_K(128)>;
-		zephyr,memory-region = "DTCM";
-	};
-
-	itcm: memory@0 {
-		compatible = "zephyr,memory-region", "arm,itcm";
-		reg = <0x00000000 DT_SIZE_K(64)>;
-		zephyr,memory-region = "ITCM";
-	};
-
 	otghs_fs_phy: otghs_fs_phy {
 		compatible = "usb-nop-xceiv";
 		#phy-cells = <0>;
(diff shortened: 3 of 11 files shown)


dts/arm/st/h7/stm32h747Xi_m7.dtsi
```diff
@@ -13,6 +13,18 @@
 		/delete-node/ cpu@1;
 	};
 
+	itcm: memory@0 {
+		compatible = "zephyr,memory-region", "arm,itcm";
```
JarmouniA: nit: compatibles go from, left to right, most to least specific, `arm,i/dtcm` is more specific than `zephyr,memory-region` (but one could say they are not comparable to begin with, hence the "nit").

dts/arm/st/h7/stm32h747Xi_m7.dtsi
```diff
@@ -13,6 +13,18 @@
...
+		reg = <0x00000000 DT_SIZE_K(64)>;
+		zephyr,memory-region = "ITCM";
+	};
+
+	dtcm: memory@20000000 {
+		compatible = "zephyr,memory-region", "arm,dtcm";
```
JarmouniA: ditto

dts/arm/st/h7/stm32h747Xi_m7.dtsi
iabdalkader: That makes sense, and I personally don't care a lot about this. However, this is exactly how it's defined for every other chip, should we break this convention?
```
grep -r dtcm *
f7/stm32f745.dtsi:      dtcm: memory@20000000 {
f7/stm32f745.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
f7/stm32f765.dtsi:      dtcm: memory@20000000 {
f7/stm32f765.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
f7/stm32f722.dtsi:      dtcm: memory@20000000 {
f7/stm32f722.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h747Xi_m7.dtsi: dtcm: memory@20000000 {
h7/stm32h747Xi_m7.dtsi:         compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h742.dtsi:      dtcm: memory@20000000 {
h7/stm32h742.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h723.dtsi:      dtcm: memory@20000000 {
h7/stm32h723.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h745Xi_m7.dtsi: dtcm: memory@20000000 {
h7/stm32h745Xi_m7.dtsi:         compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h7a3.dtsi:      dtcm: memory@20000000 {
h7/stm32h7a3.dtsi:              compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h757Xi_m7.dtsi: dtcm: memory@20000000 {
h7/stm32h757Xi_m7.dtsi:         compatible = "zephyr,memory-region", "arm,dtcm";
h7/stm32h755Xi_m7.dtsi: dtcm: memory@20000000 {
h7/stm32h755Xi_m7.dtsi:         compatible = "zephyr,memory-region", "arm,dtcm";
h7rs/stm32h7rs.dtsi:    dtcm: memory@20000000 {
h7rs/stm32h7rs.dtsi:            compatible = "zephyr,memory-region", "arm,dtcm";
```

Same for itcm:
```
grep -r itcm *
f7/stm32f722.dtsi:      itcm: memory@0 {
f7/stm32f722.dtsi:              compatible = "zephyr,memory-region", "arm,itcm";
h7/stm32h747Xi_m7.dtsi: itcm: memory@0 {
h7/stm32h747Xi_m7.dtsi:         compatible = "zephyr,memory-region", "arm,itcm";
h7/stm32h742.dtsi:      itcm: memory@0 {
h7/stm32h742.dtsi:              compatible = "zephyr,memory-region", "arm,itcm";
h7/stm32h723.dtsi:      itcm: memory@0 {
h7/stm32h723.dtsi:              compatible = "zephyr,memory-region", "arm,itcm";
h7/stm32h7a3.dtsi:      itcm: memory@0 {
h7/stm32h7a3.dtsi:              compatible = "zephyr,memory-region", "arm,itcm";
h7rs/stm32h7rs.dtsi:    itcm: memory@0 {
h7rs/stm32h7rs.dtsi:            compatible = "zephyr,memory-region", "arm,itcm";
```

JarmouniA: > should we break this convention?

No need.

erwango (commented):
From the search I've just made, all STM32H7 based soc use 64/128K ITCM/DTCM.
Unless I'm wrong, it would be better to put it directly in h7.dtsi (and delete node from m4 cores).

iabdalkader:
> it would be better to put it directly in h7.dtsi (and delete node from m4 cores).

We'll also need to delete any itcm/dtcm defined by _m7 files. This will be a much bigger change, 10+ files, is that okay?

erwango (changes requested):
Thanks!

But please merge all your commits (at the minimum #1 and #3) as having them split will break bisectability (it won't compile due to duplicated nodes)

dts/arm/st/h7/stm32h747Xi_m7.dtsi
iabdalkader: I put "arm,dtcm" first, but then realized it's not consistent with other series (f7, h74s) so I reverted to `"zephyr,memory-region", "arm,dtcm"`. Maybe someone else can clean up all of those if needed.

erwango (approved):
I could not find a H7 that is not compatible with this setting from datasheet.

mathieuchopstm (approved):
Seems correct per [ST Product Selector](https://www.st.com/en/microcontrollers-microprocessors/stm32h7-series/products.html) (which seems to have a bug making no TCM appear despite some products having it... glanced at a few datasheets too)

PR #100256, opened 2025-11-30
Interrupt controller API

bjarki-andreasen:
Continuation of #66505

Goal is to cover all interrupt controllers in https://github.com/zephyrproject-rtos/zephyr/tree/main/include/zephyr/drivers/interrupt_controller under one API and subsystem for configuring interrupts, and registering IRQ handlers.

Currently working on:
  * gpio interrupt wrapper
  * interrupt nexus handling


include/zephyr/sys/irq.h
```diff
@@ -0,0 +1,485 @@
...
+ *     foo: foo {
+ *             interrupts = <1 1>, <2 2>;
+ *     };
+ * @endcode
+ *
+ * Example usage:
+ *
+ *     static int foo_irq_0_handler(const void *data)
+ *     {
+ *             return 1;
```
mathieuchopstm, line 125: Shouldn't this be `return SYS_IRQ_HANDLED;` then?

Secondary remark: until now, "basic" ISRs returned `void` instead - having to manually return 1 feels like it'll be a pain point (and possibly bloat the image too).

include/zephyr/sys/irq.h
```diff
@@ -0,0 +1,485 @@
...
+ */
+#define SYS_DT_INTL_ID_BY_NAME(node_id, name)					\
+	_CONCAT_3(								\
+		DT_DEP_ORD(DT_IRQ_INTC_BY_NAME(node_id, name)),			\
+		_,								\
+		DT_IRQ_BY_NAME(node_id, name, irq)				\
+	)
+
+/**
+ * @brief Get IRQ identifier from devicetree IRQ by index
```
mathieuchopstm, line 47: By `IRQ identifier`, do you mean IRQn (IRQ number) or is it something separate? What's the difference if so?

include/zephyr/sys/irq.h
```diff
@@ -0,0 +1,485 @@
...
+ * @code{.dts}
+ *     foo: foo {
+ *             interrupts = <1 1 3>, <2 2 4>;
+ *     };
+ * @endcode
+ *
+ * Example usage:
+ *
+ *     SYS_DT_IRQ_FLAGS_BY_IDX(DT_NODELABEL(foo), 0) // irq flags of <1 1>
+ *     SYS_DT_IRQ_FLAGS_BY_IDX(DT_NODELABEL(foo), 1) // irq flags of <2 2>
```
mathieuchopstm, line 298: I guess this should be `irq flags of <1 1 2>` and `2 2 4`

include/zephyr/sys/irq.h
```diff
@@ -0,0 +1,485 @@
...
+#define SYS_DT_INST_IRQ_FLAGS(inst) \
+	SYS_DT_IRQ_FLAGS(DT_DRV_INST(inst))
+
+/** @} */
+
+/**
+ * @brief Configure IRQ
+ *
+ * @param irqn System IRQ number obtained using SYS_DT_IRQN()
+ * @param flags The IRQ configuration flags obtained using SYS_DT_IRQ_FLAGS()
```
mathieuchopstm, line 380: `IRQ_CONNECT` accepted flags from user code too (i.e., `ZLI`) - are those forbidden now?

include/zephyr/sys/irq.h
```diff
@@ -0,0 +1,485 @@
...
+ * @param handler IRQ handler
+ * @param data Data passed to IRQ handler
+ *
+ * @note The dynamic IRQ structure must remain valid until released
+ * @note Invoking from ISR is not permitted
+ *
+ * @retval 0 if successful
+ * @retval -errno code if failure
+ */
+int sys_irq_request(uint16_t irqn, struct sys_irq *irq, sys_irq_handler_t handler,
```
mathieuchopstm, line 468: Prefer `install`/`remove`? Maybe with a `_dynamic_` in the name too - since this is gated behind `CONFIG_SYS_IRQ_DYNAMIC` if I understand the intent correctly?

From the name, `sys_irq_request` feels like "request IRQ trigger by software" (i.e., what `sys_irq_trigger` is supposed to do)

include/zephyr/sys/internal/irq.h
```diff
@@ -0,0 +1,24 @@
...
+#ifndef ZEPHYR_INCLUDE_SYS_INTERNAL_IRQ_H_
+#define ZEPHYR_INCLUDE_SYS_INTERNAL_IRQ_H_
+
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+
+/** IRQ interrupt line structure */
+struct sys_irq_intl {
+	/** Interrupt controller */
+	const struct device *intc;
```
mathieuchopstm, line 16: Looking at how you're using this, it might be better footprint-wise to have a table of interrupt controller devices and save a `uint16_t intc_index` here instead (save 4 bytes per entry on 32-bit, 12 on 64-bit).

scripts/build/gen_sys_irq.py
```diff
@@ -0,0 +1,507 @@
...
+        self.irqs.append(irq)
+
+    def add_identifier(self, identifier: int):
+        assert identifier > -1, 'Idenfifier must be positive'
+        self.identifier = identifier
+
+class INTTree():
+    def __init__(self, intcs: list[INTC], intls: list[INTL], irqs: list[IRQ]):
+        self.tree = []
+        self.tree_by_intc = []
```
mathieuchopstm, line 91: Wouldn't it make more sense for something called `tree_by_intc` to be a `dict`? (`dict[INTC, list[INTBranch]]` in this case)

include/zephyr/drivers/intc.h
```diff
@@ -0,0 +1,130 @@
...
+	int (*clear_irq)(const struct device *dev, uint16_t irq);
+};
+
+/** @endcond */
+
+/**
+ * @brief Configure interrupt line
+ *
+ * @param dev Interrupt controller instance
+ * @param irq Interrupt line of interrupt controller
```
mathieuchopstm, line 43: Name should match description: shouldn't this be `intl`?

include/zephyr/drivers/intc/irq_vector.h
```diff
@@ -0,0 +1,121 @@
...
+ * @note See ISR_DIRECT_DECLARE()
+ *
+ * @param node_id Interrupt generating device's devicetree node
+ * @param name Name of IRQ within interrupt generating device's devicetree node
+ */
+#define INTC_DT_DEFINE_IRQ_VECTOR_BY_NAME(node_id, name)			\
+	ARCH_ISR_DIRECT_DECLARE(						\
+		INTC_DT_VECTOR_SYMBOL(						\
+			DT_IRQ_INTC_BY_NAME(node_id, name),			\
+			DT_IRQ_BY_NAME(node_id, name, irq)			\
```
mathieuchopstm, line 78: As far as I understand, this returns the raw `interrupts[0].irq` cell, which seems to be called `intl` so far?

drivers/intc/intc_arm_v8m_nvic/arm_v8m_nvic.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#define ARM_V8M_NVIC_VECTOR_DEFINE(inst, intln)							\
+	__weak void INTC_DT_INST_VECTOR_SYMBOL(inst, intln)(void)				\
+	{											\
+		arm_v8m_nvic_vector_wrapper();							\
+	}
+
+#define ARM_V8M_NVIC_VECTORS_DEFINE(inst)							\
+	_Pragma("GCC diagnostic push")								\
```
mathieuchopstm, line 126: Nit (at this stage): don't we have something portable for that?

drivers/intc/intc_arm_v8m_nvic/arm_v8m_nvic.c
```diff
@@ -0,0 +1,141 @@
...
+	_Pragma("GCC diagnostic ignored \"-Wattributes\"")					\
+	INTC_DT_INST_FOREACH_INTL(inst, ARM_V8M_NVIC_VECTOR_DEFINE)				\
+	_Pragma("GCC diagnostic pop")
+
+#define ARM_V8M_NVIC_VECTOR_TABLE_ENTRY_DEFINE(inst, intln) \
+	INTC_DT_INST_VECTOR_SYMBOL(inst, intln)
```
mathieuchopstm, line 132: Not sure I see the point of the wrapper.

drivers/intc/intc_arm_v8m_nvic/arm_v8m_nvic.c
```diff
@@ -0,0 +1,141 @@
...
+	.max_irq_priority = BIT_MASK(DT_INST_PROP(0, arm_num_irq_priority_bits)),
+};
+
+DEVICE_DT_INST_DEFINE(0, NULL, NULL, NULL, &config, PRE_KERNEL_1, 0, &api);
+
+const sys_irq_intl_handler_t _intl_handlers[] = {
+	INTC_DT_INST_FOREACH_INTL_SEP(0, INTC_DT_INST_INTL_HANDLER_SYMBOL, (,))
+};
+
+static void arm_v8m_nvic_vector_wrapper(void)
```
mathieuchopstm, line 100: Isn't this basically `z_isr_wrapper`?

include/zephyr/drivers/intc/irq_flags/arm_v8m_nvic.h
mathieuchopstm, line 1: I really like this mechanism of "vendor provides its serialization mechanism in a generic fashion".

Tiny nit: for larger platforms, maybe 32 bits won't be enough: a `irq_flags_t` type that could typedef to 64-bit (depending on a Kconfig) might be a sensible choice.

dts/arm/nordic/nrf5340_cpuapp_peripherals.dtsi
mathieuchopstm, line 1: Not against `status = "reserved"` - in fact I think it's a very good idea - but this warrants some docs update.

drivers/clock_control/clock_control_nrf.c
```diff
@@ -646,7 +646,7 @@ static void lfclk_spinwait(enum nrf_lfclk_start_mode mode)
 			/* Clear pending interrupt, otherwise new clock event
 			 * would not wake up from idle.
 			 */
-			NVIC_ClearPendingIRQ(DT_INST_IRQN(0));
+			sys_irq_clear(SYS_DT_IRQN(DT_NODELABEL(clock)));
```
mathieuchopstm, line 649: Seeing this: have you considered the scenario where this wouldn't be an IRQ line directly wired to NVIC? (IIRC, NVIC is level-sensitive so this *should* work, but this is an arch-specific detail)

Should we require clear-at-leaf-clears-entire-interrupt-tree behavior? or is clearing the interrupt at all intermediary levels handled?

drivers/gpio/gpio_nrfx.c
```diff
@@ -627,8 +632,11 @@ static int gpio_nrfx_init(const struct device *port)
 	}
 
 #ifdef CONFIG_GPIO_NRFX_INTERRUPT
+	if (sys_irq_configure(cfg->gpiote_irqn, cfg->gpiote_irq_flags)) {
```
mathieuchopstm, line 635: I really like that the runtime configure call is now (1) explicit in driver control flow and (2) can be made from anywhere (vs. now where it's hidden as part of `IRQ_CONNECT`)

doc/kernel/services/sys_irq.rst
mathieuchopstm, line 1: This doesn't document the two concepts that are IMO key to this new system:
* interrupt lines
* IRQ numbers

My understanding is that:
* interrupt lines are values 0~N ***specific to a given `INTC`*** corresponding to, well, an interrupt line (physical or virtual) on that controller
* IRQ numbers (`IRQn`) are ***global values*** that uniquely identify a specific *interrupt line* on a specific *`INTC`*
* in the Devicetree, ***interrupt lines*** are specified in the `interrupts` property
* in C code, ***`IRQn`s*** are manipulated by the application

\+ this indicates it replaces multi-level interrupts, but what about Shared Interrupts? (arguably, they *ought* to be replaceable by MLIs)

drivers/intc/intc_nrf_clic/nrf_clic.c
```diff
@@ -0,0 +1,86 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/intc.h>
+#include <zephyr/drivers/intc/intl.h>
+#include <hal/nrf_vpr_clic.h>
+
+#define DT_DRV_COMPAT nordic_nrf_clic
+
+static int driver_configure_irq(const struct device *dev, uint16_t irq, uint32_t flags)
+{
+	if (flags > CLIC_CLIC_CLICINT_PRIORITY_PRIOLEVEL3) {
```
mathieuchopstm, line 16: quite a mouthful :smile:

mathieuchopstm (commented):
The overall idea feels very sound and very sensible. I haven't fully wrapped my mind around the implementation yet - lacks of docs doesn't help, but I guess you haven't gotten around to documenting thoroughly  since this is still WIP; nevertheless, I'll still provide some early feedback (mixed high and low level remarks).

One general remark: unless I missed it, there doesn't seem to be a replacement for a MLI platform? That sounds like a must have to "prove" the new framework meets design goals (you may be working on it now though - if so, ignore remark)

drivers/clock_control/clock_control_nrf.c
bjarki-andreasen: The design treats every interrupt controller as an isolated, level agnostic device, like any other hardware block would be.

Take for example this system with to levels of interrupt:
```
intc0: intc0 {
        interrupt-controller;
};

intc1: intc1 {
        interrupt-parent = <&intc0>;
        interrupts = <0>;
};

uart: uart@0 {
        interrupt-parent = <&intc1>;
        interrupts = <3 0>;
};
```
say uart driver wants to enable its interrupt. It will
```
sys_irq_enable(SYS_DT_INST_IRQN(inst));
```
this calls
```
intc_irq_enable(DEVICE_DT_GET(DT_NODELABEL(intc1)), 3);
```
now, intc1 is not a root interrupt controller, it has an `interrupts` property, so it will enable its own IRQ, just like the UART did
```
sys_irq_enable(SYS_DT_INST_IRQN(inst)); /* enables interrupt line 0 of intc0 */
```
then enable its interrupt line 3.

So the chaining is built in to the design, there is no encoding of levels needed.

drivers/intc/intc_nrf_clic/nrf_clic.c
bjarki-andreasen: ha, ups

doc/kernel/services/sys_irq.rst
bjarki-andreasen: The design is intended to replace the way we do MLIs, as the "levels" are effectively designed out.

Shared interrupts are also designed in, there is an example in the original PRs description https://github.com/zephyrproject-rtos/zephyr/pull/66505, I need to actually add it to the docs in this POR with nice diagrams.

In short, shared interrupts are detected and added to the system interrupt line handler, this PR introduces SYS_IRQ_HANDLED which is returned from irq handlers, so we can track which if any interrupt generating device which shares an interrupt actually generated and handled it, to catch spurious interrupts.

cfriedt:
Just wanted to land a comment here because it was mentioned today in the Arch WG meeting.

This is an interesting concept. In maybe 2022, when @ycsin was working on it, there was quite a bit of talk about switching to a chained IRQ handler architecture, which I think might have been proposed originally by @dcpleung .

IIRC, Linux does the same thing, and the reason is to support cascaded interrupt controllers. Are cascaded interrupt controllers covered here? There are a few architectures that would benefit from that, as a significantly more mature architecture than MLI.

bjarki-andreasen:
> Just wanted to land a comment here because it was mentioned today in the Arch WG meeting.
> 
> This is an interesting concept. In maybe 2022, when @ycsin was working on it, there was quite a bit of talk about switching to a chained IRQ handler architecture, which I think might have been proposed originally by @dcpleung .
> 
> IIRC, Linux does the same thing, and the reason is to support cascaded interrupt controllers. Are cascaded interrupt controllers covered here? There are a few architectures that would benefit from that, as a significantly more mature architecture than MLI.

Yes, any "level" of interrupt controller, of any kind is supported, given it models interrupt controllers as individual devices, with a device driver API like any other. This maintains a parent/child relationship allowing "bubbling" of interrupts up the chain of interrupt controllers, while every device and interrupt controller only "sees" its parent, and only cares about its own interrupt handler.

include/zephyr/drivers/intc.h
bjarki-andreasen: Yes, should be `intl`

PR #100270, opened 2025-12-01
soc: nxp: mcxw2xx: Improve OS tick timer selection

zejiang0jason:
When os_timer is enabled in dts, then os_timer will be used as OS tick timer.

To make systick as the default OS tick timer, currently os_timer is not enabled in dts. When users want to use os_timer as OS tick timer, they need to override the dts.

Improve the method, enable the os_timer is dts, but not enable in Kconfig by default. If need to use os_timer as OS tick, just pass CONFIG_MCUX_OS_TIMER=y

### DIFF
--- a/boards/nxp/frdm_mcxw23/frdm_mcxw23_common.dtsi
+++ b/boards/nxp/frdm_mcxw23/frdm_mcxw23_common.dtsi
@@ -176,12 +176,11 @@
 	pinctrl-names = "default";
 };
 
-/*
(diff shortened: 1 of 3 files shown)


ZhaoxiangJin (approved):

Albort12138 (approved):

zejiang0jason:
add and assign to @Holt-Sun , who is the NXP timer owner.

Holt-Sun (approved):

mmahadevan108 (approved):

PR #100273, opened 2025-12-01
tests: arch: arm: custom_interrupt: fix build issue

biwenli:
Fix build issue when multi level interrupts is enabled, define new macro TEST_1ST_LEVEL_INTERRUPTS_MAX to support multi level interrupts case.
  - core_cm33.h:2559:47: error: iteration 496 invokes undefined behavior [-Werror=aggressive-loop-optimizations]

### DIFF
--- a/tests/arch/arm/arm_custom_interrupt/src/arm_custom_interrupt.c
+++ b/tests/arch/arm/arm_custom_interrupt/src/arm_custom_interrupt.c
@@ -17,6 +17,12 @@ static volatile bool custom_set_priority_called;
 static volatile bool custom_eoi_called;
 static volatile bool irq_handler_called;
 
+#if CONFIG_2ND_LVL_ISR_TBL_OFFSET > 0
+#define TEST_1ST_LEVEL_INTERRUPTS_MAX (CONFIG_2ND_LVL_ISR_TBL_OFFSET - 1)
+#else
+#define TEST_1ST_LEVEL_INTERRUPTS_MAX (CONFIG_NUM_IRQS - 1)
+#endif
+
 /* Define out custom SoC interrupt controller interface methods.
  * These closely match the normal Cortex-M implementations.
  */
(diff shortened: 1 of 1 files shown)


tests/arch/arm/arm_custom_interrupt/src/arm_custom_interrupt.c
```diff
@@ -17,6 +17,12 @@ static volatile bool custom_set_priority_called;
 static volatile bool custom_eoi_called;
 static volatile bool irq_handler_called;
 
+#ifdef CONFIG_MULTI_LEVEL_INTERRUPTS
```
wearyzen:
```suggestion
#ifdef CONFIG_2ND_LVL_ISR_TBL_OFFSET
```

biwenli: > 

The option CONFIG_2ND_LVL_ISR_TBL_OFFSET always be defined when multi level interrupts are enabled.

wearyzen: https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#workflow-suggestions-that-help-reviewers 
> Unless they applied the reviewer’s recommendation exactly, authors must not resolve and hide comments, they must let the initial reviewer do it.

wearyzen: I was referring to a similar change that went in earlier but I see that the default value of CONFIG_2ND_LVL_ISR_TBL_OFFSET is 0 (https://github.com/zephyrproject-rtos/zephyr/blob/ab09c94377c562deb7f1b1feca1fce55e1819e02/drivers/interrupt_controller/Kconfig.multilevel#L44) so a check like below seems better instead:
```suggestion
#if CONFIG_2ND_LVL_ISR_TBL_OFFSET > 0
```

biwenli: > Unless they applied the reviewer’s recommendation exactly, authors must not resolve and hide comments, they must let the initial reviewer do it.

Yes, correct. When multi level interrupts are not enabled, then the default value of option CONFIG_2ND_LVL_ISR_TBL_OFFSET is 0. Let's correct it.

wearyzen (approved):

wearyzen (approved):

nashif (approved):

PR #100278, opened 2025-12-01
boards: nxp: imx943_evk: m33: use common dtsi

biwenli:
Use common dtsi imx943_evk_mimx94398_cm.dtsi

### DIFF
--- a/boards/nxp/imx943_evk/imx943_evk_mimx94398_cm.dtsi
+++ b/boards/nxp/imx943_evk/imx943_evk_mimx94398_cm.dtsi
@@ -4,8 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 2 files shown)


dbaluta (dismissed):

JiafeiPan (approved):

ZhaoxiangJin:
Confused, what does `cm ` in `imx943_evk_mimx94398_cm.dtsi` refer to? Is it an abbreviation for `common`? If yes, why not just use `common `directly? That way the information would be clearer. We/the community have such [precedents ](https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/nxp/frdm_rw612/frdm_rw612_common.dtsi)

ZhaoxiangJin (approved):

PR #100284, opened 2025-12-01
scripts: west_commands: zspdx: add default values from zehpyr meta from purl and version

SchlegNi:
With this PR west spdx will include the information about the modules from zephyr.meta into the SBOM as default values, and override this values if a module has defined a security entry in its module.yml file.
With this change, the modules-deps.spdx would represent which versions are actually used. And tools like dependency-track are able to track used packages by the purl entry.

Signed-off-by: Nils Schlegelmilch [n.schlegelmilch@vega.com](mailto:n.schlegelmilch@vega.com)


pdgendt:
Also no fixup commits 😆 amend or squash and force push.

PR #100288, opened 2025-12-01
drivers: i2s: nrf_tdm: Add guards for buffers size

adamkondraciuk:
TDM peripheral requires TXD.MAXCNT and RXD.MAXCNT registers to be:
- multiple of 4 bytes
- larger than 8 bytes

### DIFF
--- a/drivers/i2s/i2s_nrf_tdm.c
+++ b/drivers/i2s/i2s_nrf_tdm.c
@@ -32,6 +32,11 @@ LOG_MODULE_REGISTER(tdm_nrf, CONFIG_I2S_LOG_LEVEL);
  */
 #define NRFX_TDM_STATUS_TRANSFER_STOPPED BIT(1)
 
+/* Due to hardware limitations, the TDM peripheral requires the rx/tx size
+ * to be greater than 8 bytes.
+ */
+#define NRFX_TDM_MIN_TRANSFER_SIZE_ALLOWED 8
+
 /* Maximum clock divider value. Corresponds to CKDIV2. */
 #define NRFX_TDM_MAX_SCK_DIV_VALUE TDM_CONFIG_SCK_DIV_SCKDIV_Max
 #define NRFX_TDM_MAX_MCK_DIV_VALUE TDM_CONFIG_MCK_DIV_DIV_Max
@@ -475,8 +480,10 @@ static int tdm_nrf_configure(const struct device *dev, enum i2s_dir dir,
 
 	__ASSERT_NO_MSG(tdm_cfg->mem_slab != NULL && tdm_cfg->block_size != 0);
 
-	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0) {
-		LOG_ERR("This device can transfer only full 32-bit words");
+	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0 ||
+		tdm_cfg->block_size <= NRFX_TDM_MIN_TRANSFER_SIZE_ALLOWED) {
+		LOG_ERR("This device can only transmit full 32-bit words greater than %u bytes.",
+			NRFX_TDM_MIN_TRANSFER_SIZE_ALLOWED);
 		return -EINVAL;
 	}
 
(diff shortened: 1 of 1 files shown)


drivers/i2s/i2s_nrf_tdm.c
```diff
@@ -475,9 +478,11 @@ static int tdm_nrf_configure(const struct device *dev, enum i2s_dir dir,
 
 	__ASSERT_NO_MSG(tdm_cfg->mem_slab != NULL && tdm_cfg->block_size != 0);
 
-	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0) {
-		LOG_ERR("This device can transfer only full 32-bit words");
-		return -EINVAL;
+	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0 ||
+		tdm_cfg->block_size < NRFX_TDM_MAXCNT_MIN_VALUE) {
```
nordic-segl: You wrote in the commit body that MAXCNT must be larger than 8 bytes.
This condition is NOT meet when buffer size is equal to 8 bytes. -> No error reported.
Does buffer size of 8 bytes is supported or not?

drivers/i2s/i2s_nrf_tdm.c
```diff
@@ -32,6 +32,9 @@ LOG_MODULE_REGISTER(tdm_nrf, CONFIG_I2S_LOG_LEVEL);
  */
 #define NRFX_TDM_STATUS_TRANSFER_STOPPED BIT(1)
 
+/* Minimum value of the MAXCNT register. */
+#define NRFX_TDM_MAXCNT_MIN_VALUE 8
```
masz-nordic: This should be
`TDM_RXD_MAXCNT_MAXCNT_Min * BITS_PER_BYTE`

adamkondraciuk: Why? This is a hardware limitation and is not related to the `TDM_RXD_MAXCNT_MAXCNT_Min ` value

masz-nordic: Then the define description is wrong.

nordic-segl (approved):

bjarki-andreasen (dismissed):

nordic-krch (dismissed):

kl-cruz (dismissed):

drivers/i2s/i2s_nrf_tdm.c
```diff
@@ -475,9 +480,11 @@ static int tdm_nrf_configure(const struct device *dev, enum i2s_dir dir,
...
 	__ASSERT_NO_MSG(tdm_cfg->mem_slab != NULL && tdm_cfg->block_size != 0);
 
-	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0) {
-		LOG_ERR("This device can transfer only full 32-bit words");
-		return -EINVAL;
+	if ((tdm_cfg->block_size % sizeof(uint32_t)) != 0 ||
+		tdm_cfg->block_size <= NRFX_TDM_MIN_TRANSFER_SIZE_ALLOWED) {
+		LOG_ERR("This device can only transmit full 32-bit words greater than %u bytes.",
+			NRFX_TDM_MIN_TRANSFER_SIZE_ALLOWED);
+		return -EIO;
```
nordic-segl: i2s_configure() can not return -EIO. It must be -EINVAL.
See API documentation https://docs.zephyrproject.org/latest/doxygen/html/group__i2s__interface.html#ga299003d72146c127f88d7c12c08889cc

Same in line 692 and ?686?.

nordic-segl (changes requested):

nordic-piks (changes requested):
Please align test if needed:
https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/drivers/i2s/i2s_additional/src/main.c#L763

nordic-segl (approved):

drivers/i2s/i2s_nrf_tdm.c
nordic-segl: Argh.
I didn't noticed that lines 692 and 686 are in i2s_write() function which instead shall return -EIO while it can't return -EINVAL.
https://docs.zephyrproject.org/latest/doxygen/html/group__i2s__interface.html#ga01edf23acc6c16bbaf718dab8061a7a0

Thus, line 487 -> shall be -EINVAL
lines 692 and 686 -> shall be -EIO

nordic-segl (approved):

nordic-piks (approved):

bjarki-andreasen (approved):

nordic-krch (approved):

PR #100291, opened 2025-12-01
Add TI RTI watchdog driver support for am243x-evm

Sunil-Hegde:
- Change the dtsi to include only r5f0_0
- Add main_rti0 and main_rti1 for a53 cores
- Add main_rti8 through main_rti11 for respective r5 cores
- Add mcu_rti0 for the m4 core
- Adds watchdog support for TI am243x-evm on r5f0_0 and m4 core
- Add callback support in the existing driver

These changes are tested on am243x-evm on both r5f0_0 and m4 cores.

### DIFF
--- a/boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
+++ b/boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
@@ -6,7 +6,7 @@
 /dts-v1/;
 
 #include <zephyr/dt-bindings/gpio/gpio.h>
-#include <ti/am64x_r5.dtsi>
+#include <ti/am64x_r5f0_0.dtsi>
 #include "am243x_evm_am2434_r5f0_0-pinctrl.dtsi"
 
 / {
@@ -25,6 +25,7 @@
 	aliases {
 		led0 = &ld26;
 		adc0 = &main_adc0;
+		watchdog0 = &main_rti8;
 	};
 
 	leds: leds {
@@ -181,6 +182,10 @@
 		zephyr,resolution = <12>;
 		zephyr,oversampling = <4>;
 	};
+};
+
+&main_rti8 {
+	status = "okay";
 };
 
 &main_mbox6 {
--- a/drivers/watchdog/wdt_ti_rti.c
+++ b/drivers/watchdog/wdt_ti_rti.c
@@ -7,6 +7,7 @@
 #define DT_DRV_COMPAT ti_j7_rti_wdt
 
 #include <zephyr/drivers/watchdog.h>
+#include <zephyr/irq.h>
 #include <zephyr/kernel.h>
 #include <stdint.h>
 
@@ -19,6 +20,9 @@
 #define WDT_PRELOAD_SHIFT 13
 
 #define WDT_PRELOAD_MAX 0xfff
+
+/* Only the last 6 bits are used */
+#define CLEAR_WDSTATUS 0x3f
 
 #define RTIWWDRX_NMI   0xa
 #define RTIWWDRX_RESET 0x5
@@ -59,12 +63,15 @@ struct wdt_ti_rti_regs {
 
 struct wdt_ti_rti_data {
 	DEVICE_MMIO_RAM;
+
+	wdt_callback_t callback;
 };
 
 struct wdt_ti_rti_config {
 	DEVICE_MMIO_ROM;
 
 	uint64_t freq;
+	void (*irq_config_func)(void);
 };
 
 static int wdt_ti_rti_setup(const struct device *dev, uint8_t options)
@@ -79,6 +86,8 @@ static int wdt_ti_rti_setup(const struct device *dev, uint8_t options)
 		regs->GCTRL = RTIGCTRL_RUN_BY_DBG;
 	}
 
+	/* Clear the Watchdog status register before enabling */
+	regs->WDSTATUS = CLEAR_WDSTATUS;
 	regs->DWDCTRL = WDENABLE_KEY;
 
 	return 0;
@@ -119,9 +128,12 @@ static int wdt_ti_rti_window_size(const struct wdt_window window)
 static int wdt_ti_rti_timeout(const struct device *dev, const struct wdt_timeout_cfg *cfg)
 {
 	const struct wdt_ti_rti_config *config = DEV_CFG(dev);
+	struct wdt_ti_rti_data *data = DEV_DATA(dev);
 	struct wdt_ti_rti_regs *regs = DEV_REGS(dev);
 	uint32_t timer_margin;
 	int window_size;
+
+	data->callback = cfg->callback;
 
 	window_size = wdt_ti_rti_window_size(cfg->window);
 	if (window_size < 0) {
@@ -134,6 +146,7 @@ static int wdt_ti_rti_timeout(const struct device *dev, const struct wdt_timeout
 		return -EINVAL;
 	}
 
+	/* Only NMI is supported */
 	if (cfg->flags == WDT_FLAG_RESET_SOC) {
 		regs->WWDRXNCTRL = RTIWWDRX_NMI;
 	}
@@ -158,8 +171,23 @@ static int wdt_ti_rti_feed(const struct device *dev, int channel_id)
 	return 0;
 }
 
+static void wdt_ti_rti_isr(const struct device *dev)
+{
+	struct wdt_ti_rti_data *data = DEV_DATA(dev);
+	struct wdt_ti_rti_regs *regs = DEV_REGS(dev);
+	volatile uint32_t status = regs->WDSTATUS;
+	/* Clear status register before servicing the watchdog */
+	regs->WDSTATUS = CLEAR_WDSTATUS;
+	if (status & BIT(5) && data->callback) {
+		data->callback(dev, 0);
+	}
+	regs->WDKEY = WDKEY_SEQ0;
+	regs->WDKEY = WDKEY_SEQ1;
+}
+
 static int wdt_ti_rti_init(const struct device *dev)
 {
+	const struct wdt_ti_rti_config *config = DEV_CFG(dev);
 	struct wdt_ti_rti_regs *regs;
 
 	DEVICE_MMIO_MAP(dev, K_MEM_CACHE_NONE);
@@ -169,6 +197,7 @@ static int wdt_ti_rti_init(const struct device *dev)
 		return -EINVAL;
 	}
 
+	config->irq_config_func();
 	return 0;
 }
 
@@ -180,11 +209,18 @@ static DEVICE_API(wdt, wdt_ti_rti_api) = {
 };
 
 #define WDT_TI_RTI_INIT(i)                                                                         \
+	static void wdt_ti_rti_irq_config_##i(void)                                                \
+	{                                                                                          \
+		IRQ_CONNECT(DT_INST_IRQN(i), DT_INST_IRQ(i, priority), wdt_ti_rti_isr,             \
+			    DEVICE_DT_INST_GET(i), 0);                                             \
+		irq_enable(DT_INST_IRQN(i));                                                       \
+	};                                                                                         \
 	static struct wdt_ti_rti_data wdt_ti_rti_data_##i = {};                                    \
                                                                                                    \
 	static struct wdt_ti_rti_config wdt_ti_rti_config_##i = {                                  \
 		DEVICE_MMIO_ROM_INIT(DT_DRV_INST(i)),                                              \
 		.freq = DT_INST_PROP(i, clock_frequency),                                          \
+		.irq_config_func = wdt_ti_rti_irq_config_##i,                                      \
 	};                                                                                         \
                                                                                                    \
 	DEVICE_DT_INST_DEFINE(i, wdt_ti_rti_init, NULL, &wdt_ti_rti_data_##i,                      \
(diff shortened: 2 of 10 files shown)


boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/dts-v1/;
+/ dts - v1 / ;
```
natto1784: this is not necessary

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
...
-		zephyr,ipc-shm = &ipc_shm0;
+	chosen
+	{
+		zephyr, sram = &sram;
+		zephyr, sram1 = &atcm;
+		zephyr, console = &main_uart0;
+		zephyr, shell - uart = &main_uart0;
+		zephyr, ipc = &ipc0;
+		zephyr, ipc - shm = &ipc_shm0;
 	};
```
natto1784, line 23: same here

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
...
 
-	aliases {
+	aliases
+	{
```
natto1784: same here

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
...
 		led0 = &ld26;
 		adc0 = &main_adc0;
+		watchdog0 = &rti8;
```
natto1784: we ideally should not have any aliases in the main DT. I also have a PR that removes the `adc0` alias as it should be used in an overlay for ADC applications. Same thing applies to watchdog

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
...
 	};
 
-	leds: leds {
+leds:
+	leds
+	{
```
natto1784: dont change

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
```diff
@@ -3,187 +3,229 @@
...
 
-	ipc0: ipc {
+ipc0:
+	ipc
+	{
 		compatible = "zephyr,mbox-ipm";
 		mboxes = <&main_mbox6 0>, <&main_mbox6 1>;
-		mbox-names = "rx", "tx";
+		mbox - names = "rx", "tx";
+	};
```
natto1784: do not edit existing unrelated code

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -59,28 +60,28 @@ struct wdt_ti_rti_regs {
...
 	uint64_t freq;
+	void (*irq_config_func)(void);
 };
 
 static int wdt_ti_rti_setup(const struct device *dev, uint8_t options)
 {
 	ARG_UNUSED(options);
 
 	struct wdt_ti_rti_regs *regs = DEV_REGS(dev);
-
```
natto1784, line 75: nitpick: can use a space here

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -180,11 +198,18 @@ static DEVICE_API(wdt, wdt_ti_rti_api) = {
...
+		IRQ_CONNECT(DT_INST_IRQN(i), DT_INST_IRQ(i, priority), wdt_ti_rti_isr,             \
+			    DEVICE_DT_INST_GET(i), 0);                                             \
+		irq_enable(DT_INST_IRQN(i));                                                       \
+	};                                                                                         \
 	static struct wdt_ti_rti_data wdt_ti_rti_data_##i = {};                                    \
                                                                                                    \
 	static struct wdt_ti_rti_config wdt_ti_rti_config_##i = {                                  \
 		DEVICE_MMIO_ROM_INIT(DT_DRV_INST(i)),                                              \
 		.freq = DT_INST_PROP(i, clock_frequency),                                          \
+		.irq_config_func = wdt_ti_rti_irq_config_##i,                                      \
```
natto1784, line 223: not correctly formatted line length

natto1784 (changes requested):

drivers/watchdog/wdt_ti_rti.c
natto1784: github messed with the tab width, looks mostly fine bar some whitespaces otherwise

boards/ti/am243x_evm/am243x_evm_am2434_r5f0_0.dts
Sunil-Hegde: done

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -59,28 +60,28 @@ struct wdt_ti_rti_regs {
 
 struct wdt_ti_rti_data {
 	DEVICE_MMIO_RAM;
+	wdt_callback_t callback;
 };
 
 struct wdt_ti_rti_config {
 	DEVICE_MMIO_ROM;
-
```
glneo, line 66: Leave these newlines as-is, same below.

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -59,28 +60,28 @@ struct wdt_ti_rti_regs {
...
 	struct wdt_ti_rti_regs *regs = DEV_REGS(dev);
 
 	if (options & WDT_OPT_PAUSE_HALTED_BY_DBG) {
 		regs->GCTRL = RTIGCTRL_HALT_BY_DBG;
 	} else {
 		regs->GCTRL = RTIGCTRL_RUN_BY_DBG;
 	}
-
+	/* Clear the Watchdog status register before enabling */
+	regs->WDSTATUS = 0x3f;
```
glneo: `0x3f` is a magic number here.

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -112,17 +113,21 @@ static int wdt_ti_rti_window_size(const struct wdt_window window)
...
 
 static int wdt_ti_rti_timeout(const struct device *dev, const struct wdt_timeout_cfg *cfg)
 {
 	const struct wdt_ti_rti_config *config = DEV_CFG(dev);
+	struct wdt_ti_rti_data *data = DEV_DATA(dev);
 	struct wdt_ti_rti_regs *regs = DEV_REGS(dev);
 	uint32_t timer_margin;
 	int window_size;
 
+	if (cfg->callback != NULL) {
```
glneo: No need to check here, assigning NULL to a var that is already NULL doesn't hurt. And if it wasn't NULL before, then maybe it should be after this call. Just check for NULL before you call it (which is already done below).

drivers/watchdog/wdt_ti_rti.c
Sunil-Hegde: done.

drivers/watchdog/wdt_ti_rti.c
Sunil-Hegde: changed it to a macro

drivers/watchdog/wdt_ti_rti.c
Sunil-Hegde: done.

drivers/watchdog/wdt_ti_rti.c
natto1784: i can still see newline deletions at other places for no reason.

Sunil-Hegde: oh sorry i thought that was only for rti_data and rti_config
i will fix everything

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -168,7 +196,8 @@ static int wdt_ti_rti_init(const struct device *dev)
 	if (!regs) {
 		return -EINVAL;
 	}
-
+
```
natto1784: whitespace

dts/arm/ti/am64x_r5.dtsi
```diff
@@ -188,6 +188,11 @@
 	interrupt-parent = <&vim>;
 };
 
+&main_rti8 {
+	interrupts = <0 0 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
```
glneo: IRQ number of 0 doesn't seem right..

dts/vendor/ti/am64x_main.dtsi
```diff
@@ -272,6 +272,14 @@
 		#size-cells = <0>;
 	};
 
+	main_rti8: watchdog@e080000 {
```
glneo, line 291: Why only add instance 8? If you want to only add the instances that connect to the R5 cores at least add those all (8 through 11), but I would prefer even the ones not routed to the R5s added also for completeness.

natto1784: Agreed, `am64x_main.dtsi` should have `main_rti{0-1,8-11}` since we use this file for both A53 and R5 cores. For now, enabling only `main_rti8` for R5F0_0 is fine in `boards` directory.

Additionally `am64x_mcu.dtsi` should have `mcu_rti0` which would be similarly have its status set as `okay` in the board DTS for M4F0_0

Interrupts should be defined for all of them

dts/arm/ti/am64x_r5.dtsi
natto1784: RTI 8 through 11 have the interrupt ID of 0 for their respective cores.

dts/vendor/ti/am64x_main.dtsi
Sunil-Hegde: I have made the changes and tested the same on both r5f0_0 and m4 cores

natto1784 (dismissed):

samples/drivers/watchdog/boards/am243x_evm_am2434_m4.overlay
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2025 Texas Instruments Incorporated
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/ {
+	aliases {
+		watchdog0 = &mcu_rti0;
```
glneo: You should be able to add these aliases to the base board DTS, no need to have an overlay for every application that uses them.

Sunil-Hegde: I was suggested that aliases are not supposed to be in the main DTS but should be application specific.

natto1784: Hi @glneo, I suggested to avoid use of aliases in the board DT; particularly since aliases, I believe should be left to user's discretion. However, if the need to keep at least one alias for each peripheral type in the board DT is felt, I understand and feel that it would also be fine.

glneo: To begin, Zephyr often uses the `aliases` node incorrectly currently. In many cases it is used to choose a node from a set of valid options (e.g. `sw0 = &button3;` is saying I am choosing my main application button to be `button3`), but these are uses for `chosen` nodes. And vice versa (e.g. `zephyr,flash` isn't chosen, it is the only flash available, it is simply being aliased so the OS/build can find it in a uniform way.

I fully agree with you that choices should not be in the board DT, but this isn't really a choice. This is a case when `aliases` is used correctly. This is the only timer that will function for the M4 core. Every example would have select the same, so no need to have this out in each and every example overlay.

Sunil-Hegde: I removed the overlays and added the alias in r5f0_0.dts and m4.dts

glneo (dismissed):

drivers/watchdog/wdt_ti_rti.c
```diff
@@ -9,6 +9,7 @@
 #include <zephyr/drivers/watchdog.h>
 #include <zephyr/kernel.h>
 #include <stdint.h>
+#include <zephyr/irq.h>
```
glneo: Couple more minor items, could you move this up with the rest of the `zephyr/x.h` includes.

glneo:
For the commits themselves, you should flip the order. The patch adding support for IRQs should come before adding them in DT. Also you don't need the ".h" in the driver name in the commit subject.

drivers/watchdog/wdt_ti_rti.c
Sunil-Hegde: I have changed this, also, "Also you don't need the ".h" in the driver name in the commit subject." this means I have change the commit to only include wdt_ti_rti right?

natto1784: that's right

glneo (approved):
Your commit message state what you did, but we can see that in the commit changes themselves. Your commit messages should say "why" you did these things. Not super critical here, just something to keep in mind.

Ayush1325 (approved):
LGTM. Thankfully, I decided to do a PR search before trying to implement it myself.

Sunil-Hegde:
Can this PR be merged since it has the required approvals?

natto1784 (approved):

PR #100293, opened 2025-12-01
Revert "doc: storage: retention: move retention docs under storage"

nordicjm:
This reverts commit 8c6c3501388eb9890191694123ec1570ca64cc8c.

This change was not approved by the maintainer (i.e. me) and wrongly moved retention (which is for RAM or RAM-like register peripherals) under storage, which is meant for things like flash


nordicjm:
> Thanks, that helps. Will be maybe grouping all things settings/configuration, then...
Or it might also be that the categories end up being more like "tags", with things potentially belonging to more than one (e.g. Retention has to do with Storage and Configuration)

I think that's probably a good idea to have a dedicated page for "settings" and settings-like systems

kartben:
@nordicjm the Services rework PR I was referring to is up for review https://github.com/zephyrproject-rtos/zephyr/pull/100126

PR #98117, opened 2025-10-23
Bluetooth: Fix double set random address for every advertising enable

LingaoM:
For Bluetooth Mesh, all Mesh messages will use private address by default, and then will call `bt_le_ext_adv_update_param`
for Mesh message, this function will call `bt_id_set_adv_own_addr` to set controller random address

But when call `bt_le_ext_adv_start` to start advertising, will also call `bt_id_set_adv_private_addr` to set new random address, that is duplicated, BTW, also increase latency to send mesh message, because every hci command/event complete/status need time.

This PR also fix two bbsim test problem:
1. Mesh relay, look like timing changed after this PR, not real problem.
2. PAST, look like address in `ADV_EXT_IND`, `AUX_ADV_IND` shall not same with `ADV_SYNC_IND`;

PR #100295, opened 2025-12-01
bluetooth: host: Fix stale RPA usage after invalidation

PavelVPV:
Add !BT_ADV_RPA_VALID check to force RPA regeneration when re-enabling an advertising set after RPA rotation occurred while disabled.

The BT_ADV_RANDOM_ADDR_UPDATED flag was added to prevent unnecessary address regeneration (RPA/NRPA) between bt_le_ext_adv_param_set() and bt_le_ext_adv_start() calls. However, this revealed an issue:

When RPA rotation (le_force_rpa_timeout) occurs while an advertiser is disabled, BT_ADV_RPA_VALID is cleared but the RPA is not regenerated. On subsequent bt_le_ext_adv_start() without a new param_set() call:
- BT_ADV_RANDOM_ADDR_UPDATED is already cleared (from previous start)
- Without BT_PER_ADV_ENABLED, no regeneration occurs
- Stale RPA is used, violating privacy requirements

Add !BT_ADV_RPA_VALID check for both connectable and non-connectable advertisers to ensure fresh RPA generation when the previous RPA was invalidated while the advertiser was disabled.

Fixes regression introduced in #98117.

### DIFF
--- a/subsys/bluetooth/host/adv.c
+++ b/subsys/bluetooth/host/adv.c
@@ -1514,13 +1514,16 @@ int bt_le_ext_adv_start(struct bt_le_ext_adv *adv,
 		if (IS_ENABLED(CONFIG_BT_PRIVACY) &&
 		    !atomic_test_bit(adv->flags, BT_ADV_USE_IDENTITY) &&
 		    (!atomic_test_and_clear_bit(adv->flags, BT_ADV_RANDOM_ADDR_UPDATED) ||
-		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED))) {
+		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED) ||
+		     !atomic_test_bit(adv->flags, BT_ADV_RPA_VALID))) {
 			bt_id_set_adv_private_addr(adv);
 		}
 	} else {
 		if (!atomic_test_bit(adv->flags, BT_ADV_USE_IDENTITY) &&
 		    (!atomic_test_and_clear_bit(adv->flags, BT_ADV_RANDOM_ADDR_UPDATED) ||
-		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED))) {
+		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED) ||
+		     (IS_ENABLED(CONFIG_BT_PRIVACY) &&
+		      !atomic_test_bit(adv->flags, BT_ADV_RPA_VALID)))) {
 			bt_id_set_adv_private_addr(adv);
 		}
 	}


KyraLengfeld (approved):

subsys/bluetooth/host/adv.c
```diff
@@ -1514,13 +1514,16 @@ int bt_le_ext_adv_start(struct bt_le_ext_adv *adv,
 		if (IS_ENABLED(CONFIG_BT_PRIVACY) &&
 		    !atomic_test_bit(adv->flags, BT_ADV_USE_IDENTITY) &&
 		    (!atomic_test_and_clear_bit(adv->flags, BT_ADV_RANDOM_ADDR_UPDATED) ||
-		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED))) {
+		     atomic_test_bit(adv->flags, BT_PER_ADV_ENABLED) ||
+		     !atomic_test_bit(adv->flags, BT_ADV_RPA_VALID))) {
```
Thalley, line 1518: Is `bt_le_adv_start` covered as well?

PavelVPV: The `BT_ADV_RANDOM_ADDR_UPDATED` flag should not affect `bt_le_adv_start` as far as I see:
https://github.com/zephyrproject-rtos/zephyr/blob/d02cdc734e6e4ad83960a7b97787729509c848b2/subsys/bluetooth/host/adv.c#L1279-L1288

But, `bt_le_adv_start` is implemented differently from `bt_le_ext_adv_start`:
https://github.com/zephyrproject-rtos/zephyr/blob/d02cdc734e6e4ad83960a7b97787729509c848b2/subsys/bluetooth/host/adv.c#L1513-L1528
https://github.com/zephyrproject-rtos/zephyr/blob/d02cdc734e6e4ad83960a7b97787729509c848b2/subsys/bluetooth/host/adv.c#L1214-L1242

and can be vulnerable to this as well if e.g. `bt_le_oob_get_local` is called from another thread that gets execution time while `bt_le_adv_start` sending commands to Controller.

Though I don't have means to test this. If it is true, it should be fixed separately.

Thalley: > Though I don't have means to test this. If it is true, it should be fixed separately.

That's fair. 

Since `bt_le_adv_start` calls `bt_le_adv_start_ext` I think any fixes for `bt_le_adv_start_ext` should be applied automatically to `bt_le_adv_start`

Thalley (approved):

alwa-nordic:
Wow the logic was already a mess and it's getting more complex. It should work as a quick fix though.

I think there's a small error in the commit message, correct? @PavelVPV Too late to fix it now, but I just want to note it here for future git archeology.

```diff
- - BT_ADV_RANDOM_ADDR_UPDATED is already cleared (from previous start)
+ - BT_ADV_RANDOM_ADDR_UPDATED is already set (from previous start)
```

Do you have any thoughts about how to simplify this in future work, @PavelVPV? I think it's not sustainable to layer more and more logic for something that feels like it should be fairly straight forward.

I'm thinking an alternative would be to update the documentation of `BT_ADV_RANDOM_ADDR_UPDATED`...

```diff
-Advertising random address has been updated in the controller before enabling advertising
+Advertising random address is up to date in the controller
```

...and clear `BT_ADV_RANDOM_ADDR_UPDATED` when the RPA expires.

PR #100301, opened 2025-12-01
drivers: input: CST816S: Add power management / motion interrupt support

Kampi:
- Add power management support
- Add Gesture / Motion interrupt

### DIFF
--- a/drivers/input/input_cst816s.c
+++ b/drivers/input/input_cst816s.c
@@ -1,5 +1,8 @@
 /*
  * Copyright (c) 2021 Qingsong Gou <gouqs@hotmail.com>
+ * Copyright (c) 2022 Jakob Krantz <mail@jakobkrantz.se>
+ * Copyright (c) 2023 Daniel Kampert <kontakt@daniel-kampert.de>
+ * Copyright (c) 2025 ZSWatch Project
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -11,6 +14,7 @@
 #include <zephyr/drivers/i2c.h>
 #include <zephyr/drivers/gpio.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/pm/device.h>
 #include <zephyr/dt-bindings/input/cst816s-gesture-codes.h>
 
 LOG_MODULE_REGISTER(cst816s, CONFIG_INPUT_LOG_LEVEL);
@@ -31,6 +35,7 @@ LOG_MODULE_REGISTER(cst816s, CONFIG_INPUT_LOG_LEVEL);
 #define CST816S_REG_BPC1H               0xB2
 #define CST816S_REG_BPC1L               0xB3
 #define CST816S_REG_POWER_MODE          0xA5
+#define CST816S_REG_SLEEP_MODE          0xE5
 #define CST816S_REG_CHIP_ID             0xA7
 #define CST816S_REG_PROJ_ID             0xA8
 #define CST816S_REG_FW_VERSION          0xA9
@@ -88,6 +93,18 @@ struct cst816s_config {
 #ifdef CONFIG_INPUT_CST816S_INTERRUPT
 	const struct gpio_dt_spec int_gpio;
 #endif
+
+#ifdef CONFIG_PM_DEVICE
+	/** cst816s power management profile. */
+	struct __packed {
+		uint8_t auto_wake_time_min; /**< Auto-recalibration period during low-power mode */
+		uint8_t scan_th;            /**< Low-power scan wake-up threshold */
+		uint8_t scan_win;           /**< Measurement range for low-power scan */
+		uint8_t scan_freq;          /**< Frequency for low-power scan */
+		uint8_t scan_i_dac;         /**< Current for low-power scan */
+		uint8_t auto_sleep_time_s; /**< Time of inactivity before entering low-power mode */
+	} lp_profile;
+#endif
 };
 
 /** cst816s data. */
@@ -106,11 +123,15 @@ struct cst816s_data {
 #endif
 };
 
+/* NOTE: This results in reliable low-power operation with good
+ * wake sensitivity. Consumes about 80uA in suspend mode. Can probably be tuned more to reduce power
+ * further while keeping good wake sensitivity.
+ */
+
 static int cst816s_process(const struct device *dev)
 {
 	const struct cst816s_config *cfg = dev->config;
-
-	int r;
+	int ret;
 	uint8_t event;
 	uint16_t row, col;
 	bool pressed;
@@ -120,23 +141,23 @@ static int cst816s_process(const struct device *dev)
 #ifdef CONFIG_INPUT_CST816S_EV_DEVICE
 	uint8_t gesture;
 
-	r = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_GESTURE_ID, &gesture, sizeof(gesture));
-	if (r < 0) {
+	ret = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_GESTURE_ID, &gesture, sizeof(gesture));
+	if (ret < 0) {
 		LOG_ERR("Could not read gesture-ID data");
-		return r;
+		return ret;
 	}
 #endif
 
-	r = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_XPOS_H, (uint8_t *)&x, sizeof(x));
-	if (r < 0) {
+	ret = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_XPOS_H, (uint8_t *)&x, sizeof(x));
+	if (ret < 0) {
 		LOG_ERR("Could not read x data");
-		return r;
+		return ret;
 	}
 
-	r = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_YPOS_H, (uint8_t *)&y, sizeof(y));
-	if (r < 0) {
+	ret = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_YPOS_H, (uint8_t *)&y, sizeof(y));
+	if (ret < 0) {
 		LOG_ERR("Could not read y data");
-		return r;
+		return ret;
 	}
 	col = sys_be16_to_cpu(x) & 0x0fff;
 	row = sys_be16_to_cpu(y) & 0x0fff;
@@ -166,7 +187,7 @@ static int cst816s_process(const struct device *dev)
 	}
 #endif
 
-	return r;
+	return ret;
 }
 
 static void cst816s_work_handler(struct k_work *work)
@@ -212,8 +233,8 @@ static void cst816s_chip_reset(const struct device *dev)
 static int cst816s_chip_init(const struct device *dev)
 {
 	const struct cst816s_config *cfg = dev->config;
-	int ret = 0;
-	uint8_t chip_id = 0;
+	int ret;
+	uint8_t chip_id;
 
 	cst816s_chip_reset(dev);
 
@@ -233,6 +254,13 @@ static int cst816s_chip_init(const struct device *dev)
 		return -ENODEV;
 	}
 
+	ret = i2c_reg_update_byte_dt(&cfg->i2c, CST816S_REG_MOTION_MASK, CST816S_MOTION_EN_DCLICK,
+				     CST816S_MOTION_EN_DCLICK);
+	if (ret < 0) {
+		LOG_ERR("Could not enable double-click motion mask");
+		return ret;
+	}
+
 	ret = i2c_reg_update_byte_dt(&cfg->i2c, CST816S_REG_IRQ_CTL,
 				     CST816S_IRQ_EN_TOUCH | CST816S_IRQ_EN_CHANGE,
 				     CST816S_IRQ_EN_TOUCH | CST816S_IRQ_EN_CHANGE);
@@ -240,7 +268,7 @@ static int cst816s_chip_init(const struct device *dev)
 		LOG_ERR("Could not enable irq");
 		return ret;
 	}
-	return ret;
+	return 0;
 }
 
 static int cst816s_init(const struct device *dev)
@@ -292,16 +320,108 @@ static int cst816s_init(const struct device *dev)
 	return ret;
 }
 
-#define CST816S_DEFINE(index)                                                                      \
-	static const struct cst816s_config cst816s_config_##index = {                              \
-		.i2c = I2C_DT_SPEC_INST_GET(index),                                                \
-		COND_CODE_1(CONFIG_INPUT_CST816S_INTERRUPT,                                        \
-			    (.int_gpio = GPIO_DT_SPEC_INST_GET(index, irq_gpios),), ())           \
-			.rst_gpio = GPIO_DT_SPEC_INST_GET_OR(index, rst_gpios, {}),                \
-	};                                                                                         \
-	static struct cst816s_data cst816s_data_##index;                                           \
-	DEVICE_DT_INST_DEFINE(index, cst816s_init, NULL, &cst816s_data_##index,                    \
-			      &cst816s_config_##index, POST_KERNEL, CONFIG_INPUT_INIT_PRIORITY,    \
-			      NULL);
+#ifdef CONFIG_PM_DEVICE
+static int cst816s_apply_profile(const struct cst816s_config *cfg)
+{
+	int ret;
+
+	ret = i2c_burst_write_dt(&cfg->i2c, CST816S_REG_LP_AUTO_WAKEUP_TIME,
+				 (const uint8_t *)&cfg->lp_profile, sizeof(cfg->lp_profile));
+	if (ret) {
+		LOG_WRN("Write power profile failed");
+		return ret;
+	}
+
+	return 0;
+}
+
+static int cst816s_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	int ret;
+	const struct cst816s_config *cfg = dev->config;
+
+	/* For some reason the CST816S does not respond to I2C commands after we use the standby
+	 * profile. Workaround for now is to just always reset it before we change power modes.
+	 */
+	cst816s_chip_reset(dev);
+	ret = cst816s_chip_init(dev);
+	if (ret < 0) {
+		LOG_ERR("Chip init failed during PM action");
+		return ret;
+	}
+
+	switch (action) {
+	case PM_DEVICE_ACTION_SUSPEND:
+	/* For PM TURN_ON means device is in suspend mode, hence apply suspend profile. */
+	case PM_DEVICE_ACTION_TURN_ON:
+		ret = cst816s_apply_profile(cfg);
+		if (ret < 0) {
+			LOG_WRN("Could not apply suspend profile");
+			return ret;
+		}
+
+		ret = i2c_reg_write_byte_dt(&cfg->i2c, CST816S_REG_DIS_AUTO_SLEEP, 0x00);
+		if (ret < 0) {
+			LOG_WRN("Could not enable auto sleep");
+			return ret;
+		}
+		break;
+	case PM_DEVICE_ACTION_TURN_OFF:
+		/* Put into Deep Sleep mode. */
+		ret = i2c_reg_write_byte_dt(&cfg->i2c, CST816S_REG_SLEEP_MODE,
+					    CST816S_POWER_MODE_SLEEP);
+		if (ret < 0) {
+			LOG_WRN("Could not enter deep sleep mode");
+			return ret;
+		}
+		break;
+	case PM_DEVICE_ACTION_RESUME:
+		/* Nothing to do, device is already reset above. */
+		break;
+	default:
+		return -ENOTSUP;
+	}
+
+	return 0;
+}
+#endif
+
+/* clang-format off */
+#define CST816S_DEFINE(index)             \
+	IF_ENABLED(CONFIG_PM_DEVICE, (                                    \
+		BUILD_ASSERT(DT_INST_PROP(index, scan_th) >= 1 &&              \
+				DT_INST_PROP(index, scan_th) <= 255,           \
+				"scan_th must be >= 1 and <= 255");            \
+		BUILD_ASSERT(DT_INST_PROP(index, scan_freq) >= 1 &&            \
+				DT_INST_PROP(index, scan_freq) <= 255,         \
+				"scan_freq must be >= 1 and <= 255");          \
+		BUILD_ASSERT(DT_INST_PROP(index, scan_win) <= 255,           \
+				"scan_win must be <= 255");             \
+		BUILD_ASSERT(DT_INST_PROP(index, scan_i_dac) >= 1 &&           \
+				DT_INST_PROP(index, scan_i_dac) <= 255,        \
+				"scan_i_dac must be >= 1 and <= 255");         \
+	))                                                                         \
+	static struct cst816s_data cst816s_data_##index; \
+	static const struct cst816s_config cst816s_config_##index = { \
+		.i2c = I2C_DT_SPEC_INST_GET(index), \
+		.rst_gpio = GPIO_DT_SPEC_INST_GET_OR(index, rst_gpios, {}), \
+		IF_ENABLED(CONFIG_INPUT_CST816S_INTERRUPT, \
+				(.int_gpio = GPIO_DT_SPEC_INST_GET(index, irq_gpios),))  \
+				 IF_ENABLED(CONFIG_PM_DEVICE, \
+				(.lp_profile = {                                             \
+					.auto_wake_time_min = DT_INST_PROP(index, auto_wake_time), \
+					.scan_th = DT_INST_PROP(index, scan_th),     \
+					.scan_win = DT_INST_PROP(index, scan_win),   \
+					.scan_freq = DT_INST_PROP(index, scan_freq), \
+					.scan_i_dac = DT_INST_PROP(index, scan_i_dac), \
+					.auto_sleep_time_s = DT_INST_PROP(index, auto_sleep_time), \
+			},)) }; \
+                                                                                         \
+	PM_DEVICE_DT_INST_DEFINE(index, cst816s_pm_action);          \
+                                                                                           \
+	DEVICE_DT_INST_DEFINE(index, cst816s_init, PM_DEVICE_DT_INST_GET(index), \
+			      &cst816s_data_##index, &cst816s_config_##index, POST_KERNEL, \
+			      CONFIG_INPUT_INIT_PRIORITY, NULL);
 
 DT_INST_FOREACH_STATUS_OKAY(CST816S_DEFINE)
+/* clang-format on */
--- a/drivers/input/input_cst816s.c
+++ b/drivers/input/input_cst816s.c
@@ -232,6 +232,7 @@ static void cst816s_chip_reset(const struct device *dev)
 
 static int cst816s_chip_init(const struct device *dev)
 {
+	uint8_t irq_mask;
 	const struct cst816s_config *cfg = dev->config;
 	int ret;
 	uint8_t chip_id;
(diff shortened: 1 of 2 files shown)


JarmouniA (commented):
Although the driver is used by ZSWatch, that info is not relevant and should not be mentioned in the commit message.

drivers/input/input_cst816s.c
```diff
@@ -106,45 +118,62 @@ struct cst816s_data {
...
+	uint8_t lp_scan_i_dac;
+};
+
+/* NOTE: I can not make a correlation between datasheet descriptions and actual behavior of those.
+ * Those values comes from experimentation. This results in reliable low-power operation with good
+ * wake sensitivity. Consumes about 80uA in suspend mode. Can probably be tuned more to reduce power
+ * further while keeping good wake sensitivity.
+ */
+static const struct cst816s_lp_profile cst816s_suspend_profile = {
+	.autosleep_time_s = 2,
```
faxe1008: would see those as dt binding properties with those default values so they are tunable.

drivers/input/input_cst816s.c
```diff
@@ -106,45 +118,62 @@ struct cst816s_data {
...
+	row = sys_be16_to_cpu(output.y) & 0x0FFF;
 
-	event = (x & 0xff) >> CST816S_EVENT_BITS_POS;
+	event = (output.x & 0xFF) >> CST816S_EVENT_BITS_POS;
 	pressed = (event == EVENT_PRESS_DOWN) || (event == EVENT_CONTACT);
 
-	LOG_DBG("event: %d, row: %d, col: %d", event, row, col);
+	LOG_DBG("Event: %u", event);
+	LOG_DBG("Pressed: %u", pressed);
+	LOG_DBG("Gesture: %u", output.gesture);
```
faxe1008: One log should be enough or?

drivers/input/input_cst816s.c
Kampi: That´s a good idea. I will prepare it

drivers/input/input_cst816s.c
```diff
@@ -76,12 +81,28 @@ LOG_MODULE_REGISTER(cst816s, CONFIG_INPUT_LOG_LEVEL);
...
+#define CST816S_GESTURE_DOUBLE_CLICK 0x0B
+#define CST816S_GESTURE_LONG_PRESS   0x0C
+
 #define EVENT_PRESS_DOWN 0x00U
 #define EVENT_LIFT_UP    0x01U
 #define EVENT_CONTACT    0x02U
 #define EVENT_NONE       0x03U
 
-/** cst816s configuration (DT). */
+#define RST_GPIO_SPEC(idx) GPIO_DT_SPEC_INST_GET_OR(idx, rst_gpios, {})
```
fabiobaltieri: just use `GPIO_DT_SPEC_INST_GET_OR(idx, rst_gpios, {})` directly, no point wrapping it to use it once

previous code was fine, also not power management related

drivers/input/input_cst816s.c
```diff
@@ -76,12 +81,28 @@ LOG_MODULE_REGISTER(cst816s, CONFIG_INPUT_LOG_LEVEL);
...
+#define CST816S_INT_GPIO(idx) \
+	COND_CODE_1( \
+		CONFIG_INPUT_CST816S_INTERRUPT, \
+		(.int_gpio = GPIO_DT_SPEC_INST_GET(idx, irq_gpios),), \
+		() \
+	)
```
fabiobaltieri: use `IF_ENABLED`, leave it as part of `CST816S_DEFINE`, also not powre management related

drivers/input/input_cst816s.c
```diff
@@ -90,7 +111,6 @@ struct cst816s_config {
 #endif
 };
 
-/** cst816s data. */
```
fabiobaltieri, line 93: cosmetic, should not be in this commit

drivers/input/input_cst816s.c
```diff
@@ -106,42 +126,57 @@ struct cst816s_data {
 #endif
 };
 
+struct cst816s_output {
+	uint8_t gesture;
+	uint8_t points;
+	uint16_t x;
+	uint16_t y;
+};
```
fabiobaltieri: not pm

drivers/input/input_cst816s.c
```diff
@@ -106,42 +126,57 @@ struct cst816s_data {
...
+	uint8_t lp_scan_win;
+	uint8_t lp_scan_freq;
+	uint8_t lp_scan_i_dac;
+};
+
+/* NOTE: I can not make a correlation between datasheet descriptions and actual behavior of those.
+ * Those values comes from experimentation. This results in reliable low-power operation with good
+ * wake sensitivity. Consumes about 80uA in suspend mode. Can probably be tuned more to reduce power
+ * further while keeping good wake sensitivity.
+ */
```
fabiobaltieri: who's "I"? this could use some rewording

drivers/input/input_cst816s.c
```diff
@@ -106,42 +126,57 @@ struct cst816s_data {
...
-	if (r < 0) {
-		LOG_ERR("Could not read gesture-ID data");
-		return r;
-	}
-#endif
+	struct cst816s_output output;
+	const struct cst816s_config *cfg = dev->config;
 
-	r = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_XPOS_H, (uint8_t *)&x, sizeof(x));
-	if (r < 0) {
```
fabiobaltieri, line 131: why change the format to drop the intermediate return variable? if anything rename it to `ret` to be coherent with the rest of the file and use it to produce more meaningful error messages rather than throwing it away

drivers/input/input_cst816s.c
```diff
@@ -106,42 +126,57 @@ struct cst816s_data {
...
 
-	r = i2c_burst_read_dt(&cfg->i2c, CST816S_REG_YPOS_H, (uint8_t *)&y, sizeof(y));
-	if (r < 0) {
-		LOG_ERR("Could not read y data");
-		return r;
-	}
-	col = sys_be16_to_cpu(x) & 0x0fff;
-	row = sys_be16_to_cpu(y) & 0x0fff;
+	col = sys_be16_to_cpu(output.x) & 0x0fff;
+	row = sys_be16_to_cpu(output.y) & 0x0fff;
```
fabiobaltieri: not pm

drivers/input/input_cst816s.c
```diff
@@ -155,18 +190,27 @@ static int cst816s_process(const struct device *dev)
...
+#endif
 
-	LOG_DBG("gesture: %d", gesture);
+	return 0;
+}
+
+static int cst816s_write_reg_checked(const struct i2c_dt_spec *i2c, uint8_t reg, uint8_t val,
+				     const char *label)
+{
+	int ret = i2c_reg_write_byte_dt(i2c, reg, val);
```
fabiobaltieri: again not pm? really struggling to understand what changed and why

drivers/input/input_cst816s.c
```diff
@@ -203,6 +247,7 @@ static void cst816s_chip_reset(const struct device *dev)
 			LOG_ERR("Could not configure reset GPIO pin");
 			return;
 		}
+
```
fabiobaltieri: you got the idea...

drivers/input/input_cst816s.c
```diff
@@ -233,6 +278,12 @@ static int cst816s_chip_init(const struct device *dev)
 		return -ENODEV;
 	}
 
+	if (i2c_reg_update_byte_dt(&cfg->i2c, CST816S_REG_MOTION_MASK, CST816S_MOTION_EN_DCLICK,
+				   CST816S_MOTION_EN_DCLICK) < 0) {
+		LOG_ERR("Could not enable double-click motion mask!");
+		return -ENODATA;
```
fabiobaltieri: the rest of the code uses `ret` and return it, do the same, would be nice if the error logs it too but at least keep the styling coherent

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
 		      K_MSEC(CONFIG_INPUT_CST816S_PERIOD));
 #endif
 
+	ret = cst816s_chip_init(dev);
+
+	if (ret < 0) {
+		return ret;
+	}
```
fabiobaltieri: drop the blank line between assignment and check, but also why did this move?

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
...
+	ret |= cst816s_write_reg_checked(&cfg->i2c, CST816S_REG_LP_AUTO_WAKEUP_TIME,
+					 p->lp_auto_wake_time_min, "lp_auto_wake_time");
+	ret |= cst816s_write_reg_checked(&cfg->i2c, CST816S_REG_LP_SCAN_TH, p->lp_scan_th,
+					 "lp_scan_th");
+	ret |= cst816s_write_reg_checked(&cfg->i2c, CST816S_REG_LP_SCAN_WIN, p->lp_scan_win,
+					 "lp_scan_win");
+	ret |= cst816s_write_reg_checked(&cfg->i2c, CST816S_REG_LP_SCAN_FREQ, p->lp_scan_freq,
+					 "lp_scan_freq");
+	ret |= cst816s_write_reg_checked(&cfg->i2c, CST816S_REG_LP_SCAN_I_DAC, p->lp_scan_i_dac,
+					 "lp_scan_i_dac");
```
fabiobaltieri: no please, this pattern of "fail a call, try the other one" is totally pointless, if any of these fails the whole system is broken, there's zero reasons to try and go ahead and return a random error, just  check and return after every call, no need to get fancy and print custom error messages just `if (ret < 0) {return ret;}`

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
...
+	if (ret) {
+		return -EIO;
+	}
+
+	return 0;
+}
+
+static int cst816s_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	int status = 0;
```
fabiobaltieri: drop the initializer

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
...
+	const struct cst816s_config *cfg = dev->config;
+
+	/* For some reason the CST816S does not respond to I2C commands after we use the standby
+	 * profile. Workaround for now is to just always reset it before we change power modes.
+	 */
+	cst816s_chip_reset(dev);
+	status = cst816s_chip_init(dev);
+	LOG_WRN("PM action: %u, reset status: %d", action, status);
```
fabiobaltieri: debugging leftover? drop

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
...
+
+	switch (action) {
+	case PM_DEVICE_ACTION_SUSPEND:
+	/* For PM TURN_ON means device is in suspend mode, hence apply suspend profile. */
+	case PM_DEVICE_ACTION_TURN_ON: {
+		status = cst816s_apply_profile(cfg, &cst816s_suspend_profile);
+
+		if (i2c_reg_write_byte_dt(&cfg->i2c, CST816S_REG_DIS_AUTO_SLEEP, 0x00) < 0) {
+			LOG_WRN("Could not enable auto sleep");
+		}
```
fabiobaltieri: use `ret` like the rest of the file, but also what's the story here? if `cst816s_apply_profile` fails you still want do the register write, or if `cst816s_apply_profile` succeed you are fine with the reg write failing? I'm confused

drivers/input/input_cst816s.c
```diff
@@ -289,19 +337,98 @@ static int cst816s_init(const struct device *dev)
...
+		break;
+	}
+	case PM_DEVICE_ACTION_RESUME: {
+		/* Nothing to do, device is already reset above. */
+		break;
+	}
+	default: {
+		return -ENOTSUP;
+	}
```
fabiobaltieri: drop the inner `{`

fabiobaltieri (commented):
hey, sorry don't want to sound pedantic but the PM comment has a lot of stuff and not a lot of PM and is really hard to follow, this needs some more splits, did a pass anyway

drivers/input/input_cst816s.c
```diff
@@ -292,14 +327,104 @@ static int cst816s_init(const struct device *dev)
...
+	if (ret) {
+		LOG_WRN("Write lp_scan_freq 0x%02x failed (%d)", p->lp_scan_freq, ret);
+		return ret;
+	}
+
+	ret = i2c_reg_write_byte_dt(&cfg->i2c, CST816S_REG_LP_SCAN_I_DAC, p->lp_scan_i_dac);
+	if (ret) {
+		LOG_WRN("Write lp_scan_i_dac 0x%02x failed (%d)", p->lp_scan_i_dac, ret);
+		return ret;
+	}
```
faxe1008, line 338: #define CST816S_REG_LP_AUTO_WAKEUP_TIME 0xF4
#define CST816S_REG_LP_SCAN_TH          0xF5
#define CST816S_REG_LP_SCAN_WIN         0xF6
#define CST816S_REG_LP_SCAN_FREQ        0xF7
#define CST816S_REG_LP_SCAN_I_DAC       0xF8
#define CST816S_REG_AUTOSLEEP_TIME      0xF9

Since they are continous can you maybe reorder the struct fields and write the entire struct at once?

drivers/input/input_cst816s.c
```diff
@@ -292,16 +320,87 @@ static int cst816s_init(const struct device *dev)
...
+			LOG_WRN("Could not enable auto sleep! (%d)", ret);
+		}
+		break;
+	case PM_DEVICE_ACTION_TURN_OFF:
+		/* Put into Deep Sleep mode */
+		ret = i2c_reg_write_byte_dt(&cfg->i2c, CST816S_REG_SLEEP_MODE,
+					    CST816S_POWER_MODE_SLEEP);
+		if (ret < 0) {
+			LOG_WRN("Could not enter deep sleep mode! (%d)", ret);
+		}
```
fabiobaltieri, line 381: `return ret` here and in 361, `return 0` at the end of the function, make it a bit easier to follow, one does not have to track ret

drivers/input/input_cst816s.c
```diff
@@ -292,16 +320,87 @@ static int cst816s_init(const struct device *dev)
...
+			   (.int_gpio = GPIO_DT_SPEC_INST_GET(index, irq_gpios),))                                                 \
+				 IF_ENABLED(CONFIG_PM_DEVICE,                                        \
+               (.lp_profile = {                                             \
+                   .autosleep_time_s = DT_INST_PROP(index, auto_sleep_time), \
+                   .auto_wake_time_min = DT_INST_PROP_OR(index, auto_wake_time, 0), \
+                   .scan_th = DT_INST_PROP_OR(index, scan_th, 0),     \
+                   .scan_win = DT_INST_PROP_OR(index, scan_win, 0),   \
+                   .scan_freq = DT_INST_PROP_OR(index, scan_freq, 0), \
+                   .scan_i_dac = DT_INST_PROP_OR(index, scan_i_dac, 0), \
+               },)) }; \
```
fabiobaltieri: check the intentation, it's a bit all over the place (are you on 8 spaces per tab?)

dts/bindings/input/hynitron,cst816s.yaml
```diff
@@ -21,3 +21,44 @@ properties:
       The reset signal defaults to active low to the
       sensor. The property value should ensure the flags properly
       describe the signal that is presented to the driver.
+
+  auto-wake-time:
+    type: int
+    default: 5
+    description:
+      Auto-recalibration period during low-power mode in minutes. Range 1–5.
+      Default 5
```
JarmouniA: Default values should be justified in the property description.

faxe1008:
You can use BUILD_ASSERT in the device dt macro to check them for valid range with IN_RANGE

JarmouniA (dismissed):

drivers/input/input_cst816s.c
Kampi: No 4 spaces but there were tabs mixed in. Copilot marked it already.

JarmouniA:
> So if I have implemented it, I can close it.

Only the reviewer can determine if it was actually implemented correctly, unless the requested change is fairly trivial and unambiguous.

If the author resolves my comments, as a reviewer I have to click on "show resolved" for every one to see what was the comment (I could have reviewed the PR 1 month ago) & to see the author's response if any, which is a lot of work.

Also, from my experience some authors mark comments they do not want to implement or discuss as resolved to hide them, which is unfortunate (not saying it's the case here).

fabiobaltieri:
Anyway we are steering this issue, @JarmouniA if you want to change the guideline I'd say just open a PR to change it and we can discuss that on the PR. I don't think it's that important, at the end of the day you can't rely on people following the guideline and have to double check anyway.

JarmouniA:
> Anyway we are steering this issue, @JarmouniA if you want to change the guideline

I'm fine with it as is.

drivers/input/input_cst816s.c
```diff
@@ -233,6 +254,13 @@ static int cst816s_chip_init(const struct device *dev)
 		return -ENODEV;
 	}
 
+	ret = i2c_reg_update_byte_dt(&cfg->i2c, CST816S_REG_MOTION_MASK, CST816S_MOTION_EN_DCLICK,
+				     CST816S_MOTION_EN_DCLICK);
+	if (ret < 0) {
+		LOG_ERR("Could not enable double-click motion mask!");
+		return -ENODATA;
```
fabiobaltieri:
```suggestion
		LOG_ERR("Could not enable double-click motion mask: %d", ret);
		return ret;
```

fabiobaltieri: Or whatever error ret string format is in the rest of the file, sorry losing track of it :-)

drivers/input/input_cst816s.c
```diff
@@ -292,16 +320,123 @@ static int cst816s_init(const struct device *dev)
...
+		return ret;
+	}
+
+	switch (action) {
+	case PM_DEVICE_ACTION_SUSPEND:
+	/* For PM TURN_ON means device is in suspend mode, hence apply suspend profile. */
+	case PM_DEVICE_ACTION_TURN_ON:
+		ret = cst816s_apply_profile(cfg);
+		if (ret < 0) {
+			LOG_WRN("Could not apply suspend profile! (%d)", ret);
```
fabiobaltieri:
```suggestion
		if (ret < 0) {
			LOG_WRN("Could not apply suspend profile (%d)", ret);
```

Sorry I know this may be the most nitpicky comment ever, it's odd to have the extra emphasis on this, does not strike me as good use of flash memory :-)

drivers/input/input_cst816s.c
```diff
@@ -292,16 +320,123 @@ static int cst816s_init(const struct device *dev)
...
+
+	return 0;
+}
+#endif
+
+/* clang-format off */
+#define CST816S_DEFINE(index)             \
+	IF_ENABLED(CONFIG_PM_DEVICE, (                                         \
+		BUILD_ASSERT(DT_INST_PROP(index, auto_wake_time) >= 1 &&       \
+				DT_INST_PROP(index, auto_wake_time) <= 5,       \
```
fabiobaltieri: `\` are out again also can you put them all under a single `IF_ENABLED`?

Kampi: Haha nice. I've asked Claude how I can combine these macros and he suggested this and your combined approach. My approach was mentioned as the most usual one :D

But sure, I can combine them :)

fabiobaltieri: well the trivial one would be to bring the config in the condition like `BUILD_ASSERT(CONFIG_PM_DEVICE && DT_INST_PROP(...`, sounds spammy though, the other one would be to have a dedicated macro with the check and define it to a noop `ifndef CONFIG_PM_DEVICE`, or group them under the `IF_ENABLED` here, you are basically using `IF_ENABLED` as a `#idef` that you can use within another macro. I'd be fine with both last two

dts/bindings/input/hynitron,cst816s.yaml
```diff
@@ -21,3 +21,49 @@ properties:
       The reset signal defaults to active low to the
       sensor. The property value should ensure the flags properly
       describe the signal that is presented to the driver.
+
+  auto-wake-time:
+    type: int
+    default: 5
+    description:
```
JarmouniA, line 34:
```suggestion
    enum:
      - 1
      - 2
      - 3
      - 4
      - 5
    description:
```

Kampi: I´ve thought about this solution yesterday, but I don´t like it. The reason is that a valid time is in range 1 to 5. I can use an enum from 1 to 5, for sure, but then I have to increase the index by 1 (or is there a better solution to let the enum start with index 1?)
That´s the reason why I decided to use a BUILD_ASSERT and check for a valid time there.

drivers/input/input_cst816s.c
Kampi: I´ve applied the same format to all of the error messages

dts/bindings/input/hynitron,cst816s.yaml
joelguittet: My opinion on this is that a duration in general is not an enum so I'm fine with the original proposal

drivers/input/input_cst816s.c
```diff
@@ -233,14 +254,21 @@ static int cst816s_chip_init(const struct device *dev)
 		return -ENODEV;
 	}
 
+	ret = i2c_reg_update_byte_dt(&cfg->i2c, CST816S_REG_MOTION_MASK, CST816S_MOTION_EN_DCLICK,
```
joelguittet, line 258: I guess this not linked directly to PM (it's not conditioned on CONFIG_PM) so I would expect a separated commit for this modification, to explain the purpose

dts/bindings/input/hynitron,cst816s.yaml
JarmouniA: > I can use an enum from 1 to 5, for sure, but then I have to increase the index by 1

I don't understand what's the problem here.

JarmouniA: > My opinion on this is that a duration in general is not an enum

!?? The enum is just a way for DT compiler to validate the values, has nothing to do with the type of the property.
https://docs.zephyrproject.org/latest/build/dts/bindings-syntax.html#enum

JarmouniA: > That´s the reason why I decided to use a BUILD_ASSERT and check for a valid time there.

The enum here will catch invalid values at the DT parsing stage, before the C code is even preprocessed, that's its advantage.

joelguittet: Agree with you. But it's not only a technical DT point of view. We speak of documentation and user friendly presentation.
If i see an enum then i would expect an enum in the datasheet, but it's not.

joelguittet: I don't know what is the common way for all drivers but I would suggest to do as usual in zephyr. If we do enum in general, I'm fine with an enum. Consistency is a key :-)

Kampi: > > I can use an enum from 1 to 5, for sure, but then I have to increase the index by 1
> 
> I don't understand what's the problem here.

There is no problem with your solution. If there is a general way to handle duration I will use it for sure. From my point of view it´s a bit more complicated to use this selection based approach and increase the index by one instead to use an assert to catch invalid values.

But as @joelguittet mentioned: if there is a general way of handling limited values for durations I will use this way to go.

drivers/input/input_cst816s.c
Kampi: Yes / no. There was one suggestion to change messages related to the PM change and I´ve changed the rest to make it consistent. But I can also move them into a separate commit.

dts/bindings/input/hynitron,cst816s.yaml
JarmouniA: > If we do enum in general, I'm fine with an enum

We do, search for `enum` in `dts/bindings`.

Kampi: @JarmouniA I´ve changed it to an enum. Please take a look at it.

drivers/input/input_cst816s.c
```diff
@@ -88,6 +93,18 @@ struct cst816s_config {
 #ifdef CONFIG_INPUT_CST816S_INTERRUPT
 	const struct gpio_dt_spec int_gpio;
 #endif
+
+#ifdef CONFIG_PM_DEVICE
+	/** cst816s power management profile. */
```
uLipe, line 98: Nit: it would be good to document also the struct fields.

uLipe (commented):
Looks good, I just have a nit.

drivers/input/input_cst816s.c
Kampi: Will do

Kampi: Will do

Kampi: Will do

drivers/input/input_cst816s.c
```diff
@@ -106,11 +123,15 @@ struct cst816s_data {
 #endif
 };
 
+/* NOTE: This results in reliable low-power operation with good
```
faxe1008, line 126: stray comment

faxe1008 (dismissed):

joelguittet (approved):

uLipe (dismissed):

faxe1008 (dismissed):

drivers/input/input_cst816s.c
```diff
@@ -292,16 +315,108 @@ static int cst816s_init(const struct device *dev)
...
+					.scan_freq = DT_INST_PROP(index, scan_freq), \
+					.scan_i_dac = DT_INST_PROP(index, scan_i_dac), \
+					.auto_sleep_time_s = DT_INST_PROP(index, auto_sleep_time), \
+			},)) }; \
+                                                                                         \
+	PM_DEVICE_DT_INST_DEFINE(index, cst816s_pm_action);          \
+                                                                                           \
+	DEVICE_DT_INST_DEFINE(index, cst816s_init, PM_DEVICE_DT_INST_GET(index), \
+			      &cst816s_data_##index, &cst816s_config_##index, POST_KERNEL, \
+			      CONFIG_INPUT_INIT_PRIORITY, NULL);
```
fabiobaltieri: hey the `\` are all over the place again

Kampi: Hi @fabiobaltieri 
sorry I don´t get your point. What do you mean? The CI format checker doesn´t complain about it.

fabiobaltieri: x

that's how it looks like, the `\` should be aligned at some point (with 8 spaced tab), the CI does not check for that

drivers/input/input_cst816s.c
```diff
@@ -292,16 +320,108 @@ static int cst816s_init(const struct device *dev)
...
-	DEVICE_DT_INST_DEFINE(index, cst816s_init, NULL, &cst816s_data_##index,                    \
-			      &cst816s_config_##index, POST_KERNEL, CONFIG_INPUT_INIT_PRIORITY,    \
-			      NULL);
+#ifdef CONFIG_PM_DEVICE
+static int cst816s_apply_profile(const struct cst816s_config *cfg)
+{
+	int ret;
+
+	ret = i2c_burst_write_dt(&cfg->i2c, CST816S_REG_LP_AUTO_WAKEUP_TIME,
+				 (uint8_t *)&cfg->lp_profile, sizeof(cfg->lp_profile));
```
fabiobaltieri: there's a SonarQubeCloud warning for this, can you use `const uint8_t *` here?

` cast from 'const struct cst816s_config::(unnamed) *' to 'unsigned char *' drops const qualifier `

drivers/input/input_cst816s.c
Kampi: Okay got it. But clang-format messed it up every time so I must change it to an unchecked section for clang-format.

fabiobaltieri: yeah that's good but then you have to align them by hand

Kampi: Yeah I had this before but I removed it because I was unsure if this is good. I will add it again.

fabiobaltieri (approved):

uLipe (approved):

PR #100306, opened 2025-12-01
Add DAC emulator driver

lindblandro:
This driver allows for reading back the latest written value for a channel in tests. The design is mimicked from the "adc-emul" driver.


dts/bindings/dac/zephyr,dac-emul.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2025 Vaisala Oyj
+# SPDX-License-Identifier: Apache-2.0
+
+description: Emulated DAC node
```
martinjaeger:
```suggestion
description: Emulated DAC
```

tests/drivers/dac/dac_emul/src/main.c
```diff
@@ -0,0 +1,589 @@
...
+#define DAC_EMUL_UNCONFIGURED_NODE DT_NODELABEL(dac_emul_unconfigured)
+#define DAC_EMUL_MANY_NODE         DT_NODELABEL(dac_emul_many)
+
+/* Test devices */
+static const struct device *dac_emul0 = DEVICE_DT_GET(DAC_EMUL0_NODE);
+static const struct device *dac_emul_unconfigured = DEVICE_DT_GET(DAC_EMUL_UNCONFIGURED_NODE);
+static const struct device *dac_emul_many = DEVICE_DT_GET(DAC_EMUL_MANY_NODE);
+
+/* External API for testing */
+extern int dac_emul_value_get(const struct device *dev, uint8_t channel, uint32_t *value);
```
martinjaeger: Why not include `zephyr/drivers/dac/dac_emul.h` instead?

tests/drivers/dac/dac_emul/src/main.c
```diff
@@ -0,0 +1,589 @@
...
+
+/*** Setup and Teardown ***/
+
+static void *dac_emul_setup(void)
+{
+	zassert_not_null(dac_emul0, "DAC emulator 0 device not found");
+	zassert_not_null(dac_emul_unconfigured, "DAC emulator 1 device not found");
+	zassert_not_null(dac_emul_many, "DAC emulator many device not found");
```
martinjaeger: These pointers cannot be `NULL`, as you used `DEVICE_DT_GET` (which will fail at linker time if a device is not found) and not `DEVICE_DT_GET_OR_NULL`.

tests/drivers/dac/dac_emul/src/main.c
```diff
@@ -0,0 +1,589 @@
...
+	zassert_true(device_is_ready(dac_emul_many), "DAC emulator many not ready");
+
+	return NULL;
+}
+
+static void dac_emul_before(void *fixture)
+{
+	ARG_UNUSED(fixture);
+	/* Reset state before each test if needed */
+}
```
martinjaeger: Just pass `NULL` in ZTEST_SUITE() call instead of an empty function.

tests/drivers/dac/dac_emul/src/main.c
```diff
@@ -0,0 +1,589 @@
...
+
+/*** Basic Initialization Tests ***/
+
+ZTEST(dac_emul_tests, test_device_ready)
+{
+	zassert_true(device_is_ready(dac_emul0), "Device should be ready");
+	zassert_true(device_is_ready(dac_emul_unconfigured), "Device should be ready");
+	zassert_true(device_is_ready(dac_emul_many), "Device should be ready");
+}
```
martinjaeger: This test can never fail, as you have already asserted that the device is ready in `_setup`. I suggest to remove it.

tests/drivers/dac/dac_emul/src/main.c
```diff
@@ -0,0 +1,589 @@
...
+	ret = dac_write_value(dac_emul0, 0, 512);
+	zassert_equal(ret, 0, "Write should succeed");
+
+	/* Read back */
+	uint32_t value;
+
+	ret = dac_emul_value_get(dac_emul0, 0, &value);
+	zassert_equal(ret, 0, "Read should succeed");
+	zassert_equal(value, 512, "Read value should match");
+}
```
martinjaeger: As far as I can see, this does exactly the same as `test_write_value_basic`. I suggest to keep only one of the two.

martinjaeger (changes requested):
The test cases are very extensive and I'm not sure if that's really required. Can you please explain the rationale behind all these tests? Are we not just checking each and every corner of a driver which is only for testing and will not even be used in any product at any time?

lindblandro:
> The test cases are very extensive and I'm not sure if that's really required. Can you please explain the rationale behind all these tests? Are we not just checking each and every corner of a driver which is only for testing and will not even be used in any product at any time?

To be honest, I mainly added these as I saw that there were tests for `adc_emul`.  Those probably make more sense as it's the beginning of the signal processing chain and the ADC interface has more features. In any case I'll prune and simplify these tests.

tests/drivers/dac/dac_emul/src/main.c
lindblandro: I removed the redundant test.

tests/drivers/dac/dac_emul/src/main.c
lindblandro: Removed this.

tests/drivers/dac/dac_emul/src/main.c
lindblandro: Fixed.

tests/drivers/dac/dac_emul/src/main.c
lindblandro: Removed these checks.

tests/drivers/dac/dac_emul/src/main.c
lindblandro: No idea why I did it like that in the first place. Changed to use the correct header.

dts/bindings/dac/zephyr,dac-emul.yaml
lindblandro: Removed "node".

martinjaeger (dismissed):
I was wondering if we should move some of the tests to the existing `dac_api` test, but I realized that most tests for the emul read the values back (which would not work with actual HW). So let's keep like this for now.

Thanks for your patience.

drivers/dac/dac_emul.c
```diff
@@ -0,0 +1,196 @@
+/*
+ * Copyright (c) 2025 Vaisala Oyj
```
petejohanson-adi:
```suggestion
 * Copyright (c) 2026 Vaisala Oyj
```

And elsewhere in the new files.

drivers/dac/dac_emul.c
```diff
@@ -0,0 +1,196 @@
...
+		k_mutex_unlock(&data->channel_mutex);
+
+		LOG_DBG("Channel %u configured: %u-bit resolution", channel, resolution);
+
+		return 0;
+	}
+
+	LOG_ERR("Could not acquire channel lock (%d)", rc);
+	return (rc == -EAGAIN) ? -EBUSY : rc;
+}
```
petejohanson-adi: This is an odd inversion, IMHO, should return with an error code in the conditional, and otherwise continue:


```suggestion
	if (rc) {
		LOG_ERR("Could not acquire channel lock (%d)", rc);
		return (rc == -EAGAIN) ? -EBUSY : rc;
	}

	struct dac_emul_channel *chan = &data->channels[channel];

	chan->resolution = resolution;
	k_mutex_unlock(&data->channel_mutex);

	LOG_DBG("Channel %u configured: %u-bit resolution", channel, resolution);

	return 0;
}
```

drivers/dac/dac_emul.c
```diff
@@ -0,0 +1,196 @@
...
+
+	if (rc == 0) {
+		rc = write_value_locked(chan, channel, value);
+
+		k_mutex_unlock(&data->channel_mutex);
+		return rc;
+	}
+
+	LOG_ERR("Could not acquire channel lock (%d)", rc);
+	return (rc == -EAGAIN) ? -EBUSY : rc;
```
petejohanson-adi: Ditto.

drivers/dac/dac_emul.c
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if (value == NULL) {
+		LOG_ERR("Null pointer provided");
+		return -EINVAL;
+	}
+
+	int rc = k_mutex_lock(&data->channel_mutex, DAC_EMUL_TIMEOUT);
+
+	if (rc == 0) {
```
petejohanson-adi: Same.

include/zephyr/drivers/dac/dac_emul.h
petejohanson-adi, line 34: Does it make sense at all to also support a callback API, to get notified whenever a value is written to a channel you're interested in? Would be the "inverse" of the same functionality to provide different emulated ADC values via callback.

petejohanson-adi (changes requested):
A few thoughts. Thanks!

drivers/dac/dac_emul.c
lindblandro: They were accurate when the PR was opened, but sure :)

drivers/dac/dac_emul.c
lindblandro: I honestly can't remember anymore why I did it that way. I usually use just `if (k_mutex_lock() != 0) {}`. I'll remove the inversion.

include/zephyr/drivers/dac/dac_emul.h
lindblandro: That sounds like a good idea, you could maybe link the DAC output to an ADC emulator input through a conversion function (e.g. DAC output -> temperature -> ADC input). Maybe there are some audio related things that could also be verified this way.

My goal for this driver was to test a heater implementation and the current implementation allows me to verify the basic functionality. The callback API would probably allow me to e.g. simulate a PID control loop in software, but I'd like to think about the details some more and get this basic implementation merged first.

petejohanson-adi: Fair. Something that can get added as a follow up, not a blocker, IMHO.

drivers/dac/dac_emul.c
petejohanson-adi: Probably should be returning the `rc` value from two lines up?

petejohanson-adi (changes requested):
Just one minor thing after the refactor for the checks, otherwise LGTM.

drivers/dac/dac_emul.c
lindblandro: Fixed.

petejohanson-adi (approved):

martinjaeger (approved):

PR #98458, opened 2025-10-29
host: Implement bt_gatt_cb_unregister method for sys_slist_find_and_remove.

zhongzhijie1:
Introduce bt_gatt_cb_unregister() to unregister  callbacks.

Replace callback_list traversal with SYS_SLIST_FOR_EACH_CONTAINER_SAFE()

PR #100308, opened 2025-12-01
bluetooth: buf: Fix callback protection for ISR context

PavelVPV:
Fix assertion failure when buf_rx_freed_notify() is called from ISR by replacing k_sched_lock with atomic_ptr operations for callback pointer access.

The scheduler lock is retained during callback execution in thread context to maintain backward compatibility, but is skipped in ISR context where it's not available.

Updated documentation to clarify the behavior difference between thread and ISR contexts.

### DIFF
--- a/include/zephyr/bluetooth/buf.h
+++ b/include/zephyr/bluetooth/buf.h
@@ -196,15 +196,22 @@ struct net_buf *bt_buf_get_rx(enum bt_buf_type type, k_timeout_t timeout);
  * @ref bt_buf_get_rx function. However, this callback is called from the context of the buffer
  * freeing operation and must not attempt to allocate a new buffer from the same pool.
  *
- * @warning When this callback is called, the scheduler is locked and the callee must not perform
- * any action that makes the current thread unready. This callback must only be used for very
- * short non-blocking operation (e.g. submitting a work item).
+ * This callback must only be used for very short non-blocking operations (e.g. submitting a work
+ * item). When called from thread context, the scheduler is locked during execution and the
+ * callee must not perform any action that makes the current thread unready. When called from ISR
+ * context, the callback runs without scheduler lock.
+ *
+ * @funcprops \isr_ok
  *
  * @param type_mask A bit mask of buffer types that have been freed.
  */
 typedef void (*bt_buf_rx_freed_cb_t)(enum bt_buf_type type_mask);
 
 /** Set the callback to notify about freed buffer in the incoming data pool.
+ *
+ * It's safe to call this inside the callback itself.
+ *
+ * @funcprops \isr_ok
  *
  * @param cb Callback to notify about freed buffer in the incoming data pool. If NULL, the callback
  *           is disabled.
--- a/subsys/bluetooth/host/buf.c
+++ b/subsys/bluetooth/host/buf.c
@@ -18,6 +18,7 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/net_buf.h>
 #include <zephyr/sys/__assert.h>
+#include <zephyr/sys/atomic.h>
 #include <zephyr/sys/util_macro.h>
 #include <zephyr/sys_clock.h>
 
@@ -36,17 +37,26 @@ LOG_MODULE_REGISTER(bt_buf, CONFIG_BT_LOG_LEVEL);
  */
 #define SYNC_EVT_SIZE (BT_BUF_RESERVE + BT_HCI_EVT_HDR_SIZE + 255)
 
-static bt_buf_rx_freed_cb_t buf_rx_freed_cb;
+static atomic_ptr_t buf_rx_freed_cb;
 
 static void buf_rx_freed_notify(enum bt_buf_type mask)
 {
-	k_sched_lock();
+	bt_buf_rx_freed_cb_t cb;
+	bool in_isr = k_is_in_isr();
 
-	if (buf_rx_freed_cb) {
-		buf_rx_freed_cb(mask);
+	if (!in_isr) {
+		k_sched_lock();
 	}
 
-	k_sched_unlock();
+	cb = (bt_buf_rx_freed_cb_t)atomic_ptr_get(&buf_rx_freed_cb);
+
+	if (cb != NULL) {
+		cb(mask);
+	}
+
+	if (!in_isr) {
+		k_sched_unlock();
+	}
 }
 
 #if defined(CONFIG_BT_ISO_RX)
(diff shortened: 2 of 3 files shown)


subsys/bluetooth/host/buf.c
```diff
@@ -36,17 +36,26 @@ LOG_MODULE_REGISTER(bt_buf, CONFIG_BT_LOG_LEVEL);
...
-	if (buf_rx_freed_cb) {
-		buf_rx_freed_cb(mask);
+	if (!in_isr) {
+		k_sched_lock();
 	}
 
-	k_sched_unlock();
+	cb = (bt_buf_rx_freed_cb_t)atomic_ptr_get(&buf_rx_freed_cb);
+
+	if (cb) {
```
Thalley:
```suggestion
	if (cb != NULL) {
```

include/zephyr/bluetooth/buf.h
```diff
@@ -196,9 +196,11 @@ struct net_buf *bt_buf_get_rx(enum bt_buf_type type, k_timeout_t timeout);
  * @ref bt_buf_get_rx function. However, this callback is called from the context of the buffer
  * freeing operation and must not attempt to allocate a new buffer from the same pool.
  *
- * @warning When this callback is called, the scheduler is locked and the callee must not perform
- * any action that makes the current thread unready. This callback must only be used for very
- * short non-blocking operation (e.g. submitting a work item).
+ * @warning This callback must only be used for very short non-blocking operations  (e.g. submitting
+ * a work item). When called from thread context, the scheduler is locked during execution and the
+ * callee must not perform any action that makes the current thread unready. When called from ISR
```
HaavardRei, line 201: Nit: By callee are you referring to the callback being executed or whatever calls the callback?

HaavardRei (dismissed):

subsys/bluetooth/host/buf.c
```diff
@@ -36,17 +36,26 @@ LOG_MODULE_REGISTER(bt_buf, CONFIG_BT_LOG_LEVEL);
...
+		cb(mask);
+	}
```
Thalley, line 55: I guess this "suffers" from the same "issue" as pointed out by https://github.com/zephyrproject-rtos/zephyr/pull/98458#issuecomment-3474097193 that `cb` can be called after `bt_buf_rx_freed_cb_set(NULL)`, right?

subsys/bluetooth/host/iso.c
```diff
@@ -52,14 +52,18 @@ LOG_MODULE_REGISTER(bt_iso, CONFIG_BT_ISO_LOG_LEVEL);
...
 {
+	bt_iso_buf_rx_freed_cb_t cb;
+
+	cb = (bt_iso_buf_rx_freed_cb_t)atomic_ptr_get(&buf_rx_freed_cb);
+
 	net_buf_destroy(buf);
 
-	if (buf_rx_freed_cb) {
-		buf_rx_freed_cb();
+	if (cb) {
```
Thalley:
```suggestion
	if (cb != NULL) {
```

Thalley (changes requested):
Minor comments w.r.t. complying with https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html rule 85, but otherwise I think this is good. Not sure if the out-of-order change is something that needs to be considered?

include/zephyr/bluetooth/buf.h
PavelVPV: The callback.

KyraLengfeld (dismissed):

subsys/bluetooth/host/buf.c
PavelVPV: True. I can add this the API documentation: "Don't unset a callback while it might be executing. Don't free callback resources until after unregistering."

include/zephyr/bluetooth/buf.h
```diff
@@ -196,15 +196,19 @@ struct net_buf *bt_buf_get_rx(enum bt_buf_type type, k_timeout_t timeout);
...
+ * context, the callback runs without scheduler lock.
  *
  * @param type_mask A bit mask of buffer types that have been freed.
  */
 typedef void (*bt_buf_rx_freed_cb_t)(enum bt_buf_type type_mask);
 
 /** Set the callback to notify about freed buffer in the incoming data pool.
+ *
+ * Don't unset a callback while it might be executing. Don't free callback resources until after
+ * unregistering.
```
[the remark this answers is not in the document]
PavelVPV: This is too verbose IMHO.

subsys/bluetooth/host/buf.c
```diff
@@ -36,17 +37,26 @@ LOG_MODULE_REGISTER(bt_buf, CONFIG_BT_LOG_LEVEL);
...
+	bool in_isr = k_is_in_isr();
 
-	if (buf_rx_freed_cb) {
-		buf_rx_freed_cb(mask);
+	if (!in_isr) {
+		k_sched_lock();
 	}
 
-	k_sched_unlock();
+	cb = (bt_buf_rx_freed_cb_t)atomic_ptr_get(&buf_rx_freed_cb);
```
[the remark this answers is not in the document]
PavelVPV: Calling `atomic_ptr_get` before `k_sched_lock` will increase the window where the thread could be preempted before `k_sched_lock` is called, thus will brake backward compatibility. It is perfectly fine if callback has been changed before `k_sched_lock` is called.

Thalley (dismissed):

HaavardRei (dismissed):

KyraLengfeld (dismissed):

include/zephyr/bluetooth/buf.h
```diff
@@ -196,15 +196,19 @@ struct net_buf *bt_buf_get_rx(enum bt_buf_type type, k_timeout_t timeout);
...
  * freeing operation and must not attempt to allocate a new buffer from the same pool.
  *
- * @warning When this callback is called, the scheduler is locked and the callee must not perform
- * any action that makes the current thread unready. This callback must only be used for very
- * short non-blocking operation (e.g. submitting a work item).
+ * @warning This callback must only be used for very short non-blocking operations (e.g. submitting
+ * a work item). When called from thread context, the scheduler is locked during execution and the
+ * callee must not perform any action that makes the current thread unready. When called from ISR
+ * context, the callback runs without scheduler lock.
  *
```
alwa-nordic, line 203: Let's make it clear that functions of this type must be ISR-safe with `@funcprops \isr_ok`. I think this can replace the warning above.

include/zephyr/bluetooth/buf.h
alwa-nordic: "Don't unset a callback while it might be executing." is not meaningful advice. (How do I know when it might be executing?) And it's not actually a problem. It's one of the moments where there is no race condition, assuming the invoking function is not reentrant.

"Don't free callback resources until after unregistering." - What is meant by this? If it means invoking the callback must be safe while it's registered, I think that's obvious and including it will only cause confusion because of misinterpretation.

I suggest we remove both.

I think we mean instead want to say "This function does not synchronize with the execution of the callback. An execution of the callback may start after it has been unregistered, depending on thread interleaving."

And I suggest we mark this function `@funcprops \isr_ok` and add "It's safe to call this inside the callback itself.".

alwa-nordic (changes requested):

Thalley (approved):

HaavardRei (approved):

alwa-nordic (approved):

KyraLengfeld (approved):

PR #100309, opened 2025-12-01
drivers: pwm: nrf: fix nrfx instance being zeroed on suspend

nika-nordic:
PWM driver clears its runtime state on each suspend event. However, since nrfx 4.0 integration nrfx driver instance is part of the runtime state structure, so clear action must be limited.


MarekPieta (approved):
Confirmed locally that the reported issue is fixed. Thanks!

bjarki-andreasen (approved):

anangl (approved):

PR #100313, opened 2025-12-01
drivers: gpio: gpio_pca_series: fix unresponsive gpio_pca_series

allemanm:
The gpio_pca_series driver may stop responding if the gpio_int is asserted between reading the interrupt status and clearing it. If the interrupt gpio_int is not deasserted by the time the handler finishes, the interrupt work should be rescheduled.


CkovMk:
This is indeed an issue that I did not think for. 

For extended devices, this driver only clears the bits that is asserted in the read. Should another bit is set during this period, it won't get cleared, causing int_gpio to remain asserted. As int_gpio is configured as edge trigger, it won't trigger interrupt ever again.

pdgendt (approved):

fabiobaltieri (approved):

PR #100319, opened 2025-12-01
stm32: u5: allow selection of VOS via DT

### LINKED ISSUE #98641: clock_stm32_u5: Voltage regulator high ranges cannot be selected when running at low clock frequency
### Summary

On STM32U585 (custom board), migrating to Zephyr v4.3.0-rc1 and the USB device stack “next”, as the current stack is flagged as deprecated. My application runs at HCLK = 40 MHz for power savings. The STM32U5 clock driver currently sets VOS3 whenever HCLK is below ~55 MHz. The new STM32 UDC driver checks the regulator voltage scaling and refuses to initialize the USB device peripheral unless VOS ≥ 2. As a result, USBD init fails, even though the same hardware/clock configuration worked with the old USB device stack.

### Current behavior

- With HCLK = 40 MHz, clock driver selects VOS3.
- During USBD bring-up, UDC checks VOS and fails with an error similar to:

```
/* Check that power range is 1 or 2 */
if (LL_PWR_GetRegulVoltageScaling() < LL_PWR_REGU_VOLTAGE_SCALE2) {
	LOG_ERR("Wrong Power range to use USB OTG HS");
	return -EIO;
}
```

Provide a supported way to run STM32U5 USB device (FS) at low HCLK (e.g., 40 MHz) while selecting VOS2, at least when USB is active, without patching Zephyr source locally and without having to raise HCLK above 55 MHz.

### Why this matters

- We have devices in the field that are already configured for 40 MHz HCLK for power consumption reasons, and I would be also scared a bit to do an OTA that affects the clock in the app, considering that bootloader will still be with old clock
- The legacy USB stack is being deprecated; we need to adopt the “next” stack 

A potential solution can be introducing a small runtime hook so peripherals that require higher VOS (e.g., USB device) can request VOS2 during activation and release it on suspend/disconnect (reference-counted, like the pm runtime).

### Alternatives

_No response_

### Additional Context

_No response_

mathieuchopstm:
Allow selecting a voltage scale using Device Tree on STM32U5 series. This is notably useful when a high system clock frequency is not required, but the application still wants to use the USB IPs which are only available in certain ranges.

Fixes #98641.

### DIFF
--- a/drivers/clock_control/clock_stm32_ll_u5.c
+++ b/drivers/clock_control/clock_stm32_ll_u5.c
@@ -37,6 +37,17 @@
 #define PLL2_ID		2
 #define PLL3_ID		3
 
+/* Shorthand for Power Controller node */
+#define PWR_NODE DT_NODELABEL(pwr)
+
+/* Dummy value to use automatic voltage scale selection */
+#define VOLTAGE_SCALE_AUTOMATIC 0xFFFFFFFFu
+
+#define SELECTED_VOLTAGE_SCALE								\
+	COND_CODE_1(DT_NODE_HAS_PROP(PWR_NODE, voltage_scale),				\
+		(CONCAT(LL_PWR_REGU_VOLTAGE_SCALE, DT_PROP(PWR_NODE, voltage_scale))),	\
+		(VOLTAGE_SCALE_AUTOMATIC))
+
 static uint32_t get_bus_clock(uint32_t clock, uint32_t prescaler)
 {
 	return clock / prescaler;
@@ -430,17 +441,35 @@ static int get_vco_input_range(uint32_t m_div, uint32_t *range, size_t pll_id)
 	return 0;
 }
 
-static void set_regu_voltage(uint32_t hclk_freq)
+static void set_regu_voltage(uint32_t hclk_freq, uint32_t wanted_scale)
 {
+	uint32_t minimal_scale, scale_to_apply;
+
 	if (hclk_freq < MHZ(25)) {
-		LL_PWR_SetRegulVoltageScaling(LL_PWR_REGU_VOLTAGE_SCALE4);
+		minimal_scale = LL_PWR_REGU_VOLTAGE_SCALE4;
 	} else if (hclk_freq < MHZ(55)) {
-		LL_PWR_SetRegulVoltageScaling(LL_PWR_REGU_VOLTAGE_SCALE3);
+		minimal_scale = LL_PWR_REGU_VOLTAGE_SCALE3;
 	} else if (hclk_freq < MHZ(110)) {
-		LL_PWR_SetRegulVoltageScaling(LL_PWR_REGU_VOLTAGE_SCALE2);
+		minimal_scale = LL_PWR_REGU_VOLTAGE_SCALE2;
 	} else {
-		LL_PWR_SetRegulVoltageScaling(LL_PWR_REGU_VOLTAGE_SCALE1);
+		minimal_scale = LL_PWR_REGU_VOLTAGE_SCALE1;
 	}
+
+	if (wanted_scale == VOLTAGE_SCALE_AUTOMATIC) {
+		scale_to_apply = minimal_scale;
+	} else if (wanted_scale < minimal_scale) {
+		/*
+		 * This ought to never happen thanks to the
+		 * compile-time checks, but better safe than
+		 * sorry. Ideally, an error message should be
+		 * logged if this ever occurs...
+		 */
+		scale_to_apply = minimal_scale;
+	} else {
+		scale_to_apply = wanted_scale;
+	}
+
+	LL_PWR_SetRegulVoltageScaling(scale_to_apply);
 	while (LL_PWR_IsActiveFlag_VOS() == 0) {
 	}
 }
@@ -855,7 +884,7 @@ int stm32_clock_control_init(const struct device *dev)
 	old_hclk_freq = __LL_RCC_CALC_HCLK_FREQ(get_startup_frequency(), LL_RCC_GetAHBPrescaler());
 
 	/* Set voltage regulator to comply with targeted system frequency */
-	set_regu_voltage(CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC);
+	set_regu_voltage(CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC, SELECTED_VOLTAGE_SCALE);
 
 	/* Set flash latency */
 	/* If freq increases, set flash latency before any clock setting */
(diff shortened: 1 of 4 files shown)


petrosyan-van:
> @petrosyan-van could you please give some feedback on this? 🙂

Thanks for this PR! I will try it out in upcoming days and will let you know :)

dts/bindings/power/st,stm32u5-pwr.yaml
```diff
@@ -0,0 +1,58 @@
...
+title: STM32U5 power controller
+
+description: |
+  Power controller of STM32U5 series
+
+  Controls two voltage regulators which can output one voltage
+  among four different voltage scales, depending on use case.
+
+include: "st,stm32-dualreg-pwr.yaml"
+
```
petrosyan-van, line 15: I think it's missing `compatible: "st,stm32u5-pwr"` here. Otherwise, it didn't compile for me

dts/bindings/power/st,stm32u5-pwr.yaml
```diff
@@ -0,0 +1,58 @@
...
+      +---------+----------------+-----------------------------------+
+      | Scale # | Max. frequency | Notes                             |
+      +---------+----------------+-----------------------------------+
+      |    1    |     160 MHz    | Highest power consumption         |
+      +---------+----------------+-----------------------------------+
+      |    2    |     110 MHz    |                                   |
+      +---------+----------------+-----------------------------------+
+      |    3    |     55  MHz    | [1]                               |
+      +---------+----------------+-----------------------------------+
+      |    4    |     25  MHz    | [1][2]; lowest power consumption  |
```
petrosyan-van: in `udc_stm32.c` we have:
```
	/* Check that power range is 1 or 2 */
	if (LL_PWR_GetRegulVoltageScaling() < LL_PWR_REGU_VOLTAGE_SCALE2) {
		LOG_ERR("Wrong Power range to use USB OTG HS");
		return -EIO;
	}
```
there is no check if it's OTG_HS or OTG_FS. Seems the requirement is to have at least scale 2 in the code

mathieuchopstm: The code in UDC driver is overzealous - the table is correct and based on STM32U5 series Reference Manual:
image

The UDC driver should definitely be fixed, but this is a separate issue outside PR scope.

dts/bindings/power/st,stm32u5-pwr.yaml
mathieuchopstm: Good catch, I know about this but forgot to do it... thanks!

mathieuchopstm: Fixed with latest push.

dts/bindings/power/st,stm32u5-pwr.yaml
```diff
@@ -0,0 +1,60 @@
...
```
erwango: Nit:
```suggestion
| 4 | 25 MHz | Lowest power consumption [1][2] |
```

dts/bindings/power/st,stm32u5-pwr.yaml
```diff
@@ -7,7 +7,7 @@ description: |
   Power controller of STM32U5 series
 
   Controls two voltage regulators which can output one voltage
-  among four different ranges depending on use case.
+  among four different voltage scales, depending on use case.
```
erwango, line 10: To be fixed in previous commit.

erwango (changes requested):
Otherwise LTGM

petrosyan-van:
I tried with `voltage-scale = <2>;` and the USB stack now works, thanks! :)
Is there a possibility to change it dynamically? I mean, ideally, I would set the `voltage-scale` to `2` (or `3` after UDC driver fix), when USB is connected, but would go to `4` when it is disconnected to save some power. Cause my plan is to enable/disable USB stack on plug/unplug

dts/bindings/power/st,stm32u5-pwr.yaml
mathieuchopstm: Done.

dts/bindings/power/st,stm32u5-pwr.yaml
mathieuchopstm: Done + saved a few characters at the right of table.

erwango (approved):

petrosyan-van (approved):

erwango (dismissed):

drivers/clock_control/clock_stm32_ll_u5.c
```diff
@@ -928,6 +957,18 @@ int stm32_clock_control_init(const struct device *dev)
 	return 0;
 }
 
+/* Asserts fSYSCLK < `freq_mhz` if `vos` is selected on PWR node */
+#define ASSERT_VALID_VOS(vos, freq_mhz)					\
+BUILD_ASSERT(DT_PROP_OR(PWR_NODE, voltage_scale, 0) != (vos)		\
+	     || CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC <= MHZ(freq_mhz),	\
+	"Maximal system clock frequency in voltage scale " #vos	\
+	" is " #freq_mhz " MHz.");
```
etienne-lms:
```suggestion
/* Asserts fSYSCLK <= `freq_mhz` if `vos` is selected on PWR node */
#define ASSERT_VALID_VOS(vos, freq_mhz)					\
	BUILD_ASSERT(DT_PROP_OR(PWR_NODE, voltage_scale, 0) != (vos) ||		\
		     CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC <= MHZ(freq_mhz),	\
		     "Maximal system clock frequency in voltage scale " #vos	\
		     " is " #freq_mhz " MHz.");

File: dts/bindings/power/st,stm32u5-pwr.yaml [hunk 6]
```diff
@@ -0,0 +1,60 @@
...
+
+      When this property is not present, the lowest voltage scale compatible
+      with the system's clock configuration is selected automatically: for
+      example, voltage scale 3 would be used if the system was configured
+      to operate with a system clock frequency of 40 MHz.
+
+      Note that the same behavior is observed if the property is provided but
+      the selected voltage scale is not compatible with the system's clock
+      configuration; in this case, clock configuration takes precedence and
+      the value of this property is ignored.
```
etienne-lms, line 55: This is not true since the implemented build time assertion.

mathieuchopstm: The assertions are a nicety but not contractual behavior.

drivers/clock_control/clock_stm32_ll_u5.c
mathieuchopstm: Done.

erwango (approved):

gautierg-st (approved):

PR #100326, opened 2025-12-01
dts: bindings: usb: Fix max3421e compatible to use hyphen word separator

MaureenHelm:
Fixes the max3421e devicetree compatible to follow the convention of using a hyphen rather than an underscore as the word separator.


kartben (changes requested):
can you please add an entry in the migration guide? Thanks!

JarmouniA (approved):

josuah (approved):
Nothing missing according to grep. 👍

kartben (approved):

ttmut (approved):

mathieuchopstm (approved):

jfischer-no (approved):

kartben:
Removing the Trivial label given this is "just" a rename, yes, but for something that's kind of in API territory, and impacts downstream users - a few extra hours of review exposures shouldn't be an issue. cc @JarmouniA

PR #100333, opened 2025-12-02
bluetooth: host: name `AUTO_PHY` choices

JordanYates:
Add names to the automatic PHY update choices so the defaults can be updated.

### DIFF
--- a/subsys/bluetooth/host/Kconfig
+++ b/subsys/bluetooth/host/Kconfig
@@ -426,7 +426,7 @@ config BT_AUTO_PHY_UPDATE
 	  behavior can be accomplished by enabling BT_AUTO_PHY_PERIPHERAL_2M and
 	  BT_AUTO_PHY_CENTRAL_2M.
 
-choice
+choice BT_AUTO_PHY_PERIPHERAL
 	prompt "Auto PHY update for peripheral role"
 	depends on BT_PERIPHERAL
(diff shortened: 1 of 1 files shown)


cvinayak (approved):

HaavardRei (approved):

Thalley (approved):

jhedberg (approved):
Seems fine to me, but I'd be curious to hear how you prefer to change the defaults and why? The current values are simply to retain as close behavior as possible to what we had before, with the exception that as peripheral you have to explicitly enable the `AUTO_PHY_UPDATE` option to get this.

JordanYates:
> Seems fine to me, but I'd be curious to hear how you prefer to change the defaults and why? The current values are simply to retain as close behavior as possible to what we had before, with the exception that as peripheral you have to explicitly enable the `AUTO_PHY_UPDATE` option to get this.

The defaults might be same (since `AUTO_PHY_UPDATE` previously defaulted to `y`), but right now there is no Kconfig default way to stop any PHY updates from happening (you can't set defaults on choice options, just the choice itself).

https://github.com/zephyrproject-rtos/zephyr/blob/8aaa5031f7e481912ba234de9bf96e1f4cee56be/subsys/bluetooth/host/Kconfig#L449-L452

kartben:
It is worth adding the "new" symbols to the release notes please

PR #100346, opened 2025-12-02
drivers: adc: npcx: add adc and v2t driver support for npck3

alvsun:
Add ADC and ADC-V2T support for npck3.

### DIFF
--- a/drivers/adc/adc_npcx.c
+++ b/drivers/adc/adc_npcx.c
@@ -108,6 +108,13 @@ struct adc_npcx_data {
 	 * of this sequence.
 	 */
 	uint32_t channels;
+
+	/* Bit-mask indicating the V2T channels
+	 * 0: adc raw mode
+	 * 1: adc v2t mode
+	 */
+	uint32_t v2t_channels_msk;
+
 	/* ADC Device pointer used in api functions */
 	const struct device *adc_dev;
 	uint16_t *buffer;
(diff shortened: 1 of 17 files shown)


drivers/adc/adc_npcx.c
```diff
@@ -164,6 +171,83 @@ static inline void adc_npcx_config_channels(const struct device *dev, uint32_t c
...
+	 */
+	inst->ADCCNF &= ~BIT(NPCX_ADCCNF_INTECCEN);
+
+	/* Enable ADC temperature end of cyclic conversion event interrupt */
+	inst->V2T_CTRL |= BIT(NPCX_V2T_CTRL_TINTCC_EN);
+
+	/* Enable the voltage to temperature function */
+	inst->V2T_CTRL |= BIT(NPCX_V2T_CTRL_TEMP_EN);
+
+	LOG_DBG("ADC termperature channel TEMP_CH is %#x\n", inst->TEMP_CH);
```
fabiobaltieri: drop the \n, LOG macros add a newline automatically

drivers/adc/adc_npcx.c
```diff
@@ -323,26 +451,37 @@ static void adc_npcx_start_scan(const struct device *dev)
...
-		LOG_DBG("Start ADC scan conversion and ADCCNF,ADCCS are (%04X,%04X)\n",
-			inst->ADCCNF, inst->ADCCS);
+		LOG_DBG("ADC scan channels ADCCS is (%04X)", inst->ADCCS);
 	}
+
+#ifdef CONFIG_ADC_V2T_NPCX
+	uint32_t v2t_ch_en_msk;
+
+	/* Apply V2T setting if V2T channel is configured */
+	v2t_ch_en_msk = (data->channels & data->v2t_channels_msk);
```
fabiobaltieri: nit: the parenthesis are redundant here, drop

drivers/sensor/nuvoton/nuvoton_adc_v2t_npcx/adc_v2t_npcx.c
```diff
@@ -0,0 +1,167 @@
...
+	int ret = 0;
+
+	if (chan != SENSOR_CHAN_AMBIENT_TEMP) {
+		return -ENOTSUP;
+	}
+
+	switch ((uint16_t)attr) {
+	case SENSOR_ATTR_NPCX_V2T_CHANNEL_CFG:
+		/* Set V2T channel bit-mask */
+		ret = adc_npcx_v2t_set_channels(config->adc_dev, (uint32_t)val->val1);
```
fabiobaltieri, line 57: the error handling here is a bit odd, I'd suggest:
- drop the initializer from `int ret`
- here check the return value and return immediately if there's an error, can't see why you'd want to continue with the rest of the function if this fails
- change the default case to return directly `return -ENOTSUP`
- change the normal return to `return 0` so ret is only used in call and checks

same for attr_get

drivers/adc/adc_npcx.c
alvsun: Done

drivers/adc/adc_npcx.c
alvsun: Done

drivers/sensor/nuvoton/nuvoton_adc_v2t_npcx/adc_v2t_npcx.c
alvsun: Done

fabiobaltieri (approved):

MulinChao (approved):

PR #100349, opened 2025-12-02
arch: arm: mpu: v8: add REGION_IO_ATTR for Cortex-M

bjarki-andreasen:
Add REGION_IO_ATTR for ARM MPU v8 Cortex-M

### DIFF
--- a/include/zephyr/arch/arm/mpu/arm_mpu_v8.h
+++ b/include/zephyr/arch/arm/mpu/arm_mpu_v8.h
@@ -259,6 +259,16 @@
 		IF_ENABLED(CONFIG_ARM_MPU_PXN, (.pxn = !PRIV_EXEC_NEVER,))			   \
 	}
 
+/** @cond INTERNAL_HIDDEN */
+#define REGION_IO_ATTR(base, size)                                                                 \
+	{                                                                                          \
+		.rbar = NOT_EXEC | P_RW_U_NA_Msk,                       /* AP, XN (SH ignored) */  \
+		.mair_idx = MPU_MAIR_INDEX_DEVICE,                      /* Cache-ability */        \
+		.r_limit = REGION_LIMIT_ADDR(base, size),               /* Region Limit */         \
+		IF_ENABLED(CONFIG_ARM_MPU_PXN, (.pxn = PRIV_EXEC_NEVER,))                          \
+	}
+/** @endcond */
+
 #if defined(CONFIG_ARM_MPU_PXN)
 /* Use this attr to define an MPU region in RAM that has code intended to be executed in
  * un-privileged mode but not in privileged mode.


wearyzen (dismissed):

include/zephyr/arch/arm/mpu/arm_mpu_v8.h
```diff
@@ -250,6 +250,14 @@
 		IF_ENABLED(CONFIG_ARM_MPU_PXN, (.pxn = !PRIV_EXEC_NEVER,))			   \
 	}
 
+#define REGION_IO_ATTR(base, size)                                                                 \
+	{                                                                                          \
+		.rbar = NOT_EXEC | P_RW_U_RW_Msk | OUTER_SHAREABLE_Msk, /* AP, XN, SH */           \
```
mathieuchopstm: On ARMv7-M, `REGION_IO_ATTR` is not accessible to user, which I think is more sensible as arbitrary MMIO operations could be used to compromise the kernel (e.g., use a DMA to overwrite your thread's credentials):
https://github.com/zephyrproject-rtos/zephyr/blob/4e62f0510c93b72de3a62f1795facf6cf048b63b/include/zephyr/arch/arm/mpu/arm_mpu_v7m.h#L134

Also, `OUTER_SHAREABLE_Msk` is useless as these bits are ignored for Device memory (DDI0553B):
>**D1.2.171 MPU_RBAR, MPU Region Base Address Register**
>
>SH, bits [4:3]
>Shareability. Defines the Shareability domain of this region for Normal memory.
>...
>For any type of Device memory, the value of this field is ignored.

>**B7.11 Device memory**
>
>*RLDDN* Device memory is always treated as shareable.

```suggestion
		.rbar = NOT_EXEC | P_RW_U_NA_Msk, /* AP, XN (SH ignored) */           \
```

bjarki-andreasen: I see, commited and rebased

mathieuchopstm (dismissed):

include/zephyr/arch/arm/mpu/arm_mpu_v8.h
mathieuchopstm, line 1: CI is complaining due to lack of documentation on the new symbol.

Since this is a private header(?), I suggest wrapping it between a `/** @cond INTERNAL_HIDDEN */` + `/** @endcond */` pair which should silence the warning.

bjarki-andreasen: Let's try

mathieuchopstm (approved):

include/zephyr/arch/arm/mpu/arm_mpu_v8.h
mathieuchopstm: I was thinking about having it right/above the `#ifndef _ASMLANGUAGE` instead such that the issue doesn't appear next time header is modified, but fine by me like that.

wearyzen (approved):

PR #98082, opened 2025-10-22
doc: update to Doxygen 1.15.0

kartben:
Doxygen 1.15.0 is now available and a bug previously impacting `STRIP_FROM_INC_PATH` functionality has been fixed. Update the list of paths accordingly to have headers from lib/libc/minimal/include/ covered.

~~Also fix warnings that were not showing in earlier version but were legit formatting issues.~~

~~DNM until Docker image with new Doxygen version is available (https://github.com/zephyrproject-rtos/docker-image/pull/275) -- CI build will fail in the meantime anyway~~

PR #100351, opened 2025-12-02
doc: generation: bump up known-good Doxygen and Latexmk versions

mathieuchopstm:
The `Documentation Generation` page listed a severely outdated "known-good" version for Doxygen (1.8.13) which is no longer correct: recent attempts to build the documentation using 1.9.1 failed miserably.

Update the minimum version to 1.15.0 which matches the version used by the Zephyr-build Docker image, also used as part of CI, to ensure that what we document as "known-good" is indeed a configuration known to work.

Following the same logic, bump up the known-good version of Latexmk to 4.80 which should match the version found in the Docker image. Even though some older versions (such as 4.76) are also working, this documentation page is informative rather than authoritative, so it's fine to not document the exact minimum version required but merely a version that is known to work.

The Graphviz version is untouched as it seems to match the Docker image.

(see also: #98082)

### DIFF
--- a/doc/contribute/documentation/generation.rst
+++ b/doc/contribute/documentation/generation.rst
@@ -76,9 +76,9 @@ Installing the documentation processors
 
 Our documentation processing has been tested to run with:
 
-* Doxygen version 1.8.13
+* Doxygen version 1.15.0
 * Graphviz 2.43
-* Latexmk version 4.56
+* Latexmk version 4.83
 * All Python dependencies listed in the repository file
   ``doc/requirements.txt``
 


doc/contribute/documentation/generation.rst
```diff
@@ -76,9 +76,9 @@ Installing the documentation processors
 
 Our documentation processing has been tested to run with:
 
-* Doxygen version 1.8.13
+* Doxygen version 1.15.0
 * Graphviz 2.43
-* Latexmk version 4.56
+* Latexmk version 4.80
```
kartben: nit - based on what CI is running as of today (latexmk is _not_ part of the docker image AFAIK so not sure what you meant by version matching the one found in the the docker build image :) )

```suggestion
* Latexmk version 4.83
```

kartben (dismissed):
Thanks a lot for this! ❤️

doc/contribute/documentation/generation.rst
mathieuchopstm: Done.

(The [Ubuntu packages page](https://launchpad.net/ubuntu/noble/+package/latexmk) listed both 4.80 and 4.83 so I went for the conservative approach and wrote 4.80, because I couldn't find an easy way to list the Latexmk version in the Docker image)

kartben: > because I couldn't find an easy way to list the Latexmk version in the Docker image

It's actually apt-get installed in CI  :)
https://github.com/zephyrproject-rtos/zephyr/blob/ad205339ef3cffcd1362ffec24bc30ac3c0f7ec2/.github/workflows/doc-build.yml#L219-L221

LaTeX distros are huge so clutterring the already large Docker image with it seems unnecessary :)

kartben (approved):

pdgendt (approved):
Can't these be generated instead of hard-coded? _dummy question_

pdgendt:
> > Can't these be generated instead of hard-coded? _dummy question_
> 
> yeah probably, but then we prob have many other places where we'd want to look at doing this too :) Starting with [docs.zephyrproject.org/latest/develop/getting_started/index.html#install-dependencies](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#install-dependencies)

Those are "minimum supported" versions, not "tested with".

PR #100354, opened 2025-12-02
tests: flash_simulator: Use ztest_test_skip

### LINKED ISSUE #100352: flash_simulator: Flash Simulator has been reported as always requiring erase.
### Describe the bug

Flash Simulator can be configured as a device with explicit erase, as it has originally been programmed, and without explicit erase, which is more natural way, considering the fact that it is emulated in RAM.

With addition of devices that do not require erase, although still can be erased with Flash API, the Flash Simulator configuration has been added to select it as a non-erase device, leaving the original behaviour, with erase, as default.

Unfortunately, due to bug in driver code imposing the Kconfig settings onto run-time configuration, implemented with commit ff81b5244833, the device would be always reporting itself as requiring erase prior to write.

This happens due to following code:
https://github.com/zephyrproject-rtos/zephyr/blob/ad205339ef3cffcd1362ffec24bc30ac3c0f7ec2/drivers/flash/flash_simulator.c#L163-L171
where `no_explicit_erase` flag will be set `false` regardless of the Kconfig options set, as in all other cases it also evaluates to 0 (false).

This bug affects only configurations that check run-time capabilities of devices as they will see Flash Simulator device as requiring erase, even if device has been set to RAMlike device.

This has not been caught in tests due to tests being written for single Flash Simulator instance and so deciding on paths by the Kconfig options. not by the run-time capabilities.

Affects versions:  3.7.0, 3.7.1, 4.0.0, 4.1.0, 4.2.0, 4.2.1, 4.3.0

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Debian 6.1.147-1 (2025-08-02) x86_64 GNU/Linux
- SDK: zephyr-sdk-0.16.3
- SHA 6da559f56544aa2ed1c93071ffad203a523d432c

### Additional Context

Found by @rkel 

**The impact is set to Annoyance:** the bug does not hamper software operation it just may result in non-optimal code paths being taken, like redundant erase cycles on RAM, due to device reporting itself as requiring explicit erase.

de-nordic:
PR renamed from: flash_simulator: Flash Simulator has been reported as always requiring erase,
due to main issue it was fixing has already been fixed with introduction of multi-instance support to Flash SImulator.

Two commits:
 - ~Fixes #100352, where Flash Simulator would always report itself as requiring erase~
 - Flash Simulator test improvements

### DIFF
--- a/tests/drivers/flash_simulator/flash_sim_impl/src/main.c
+++ b/tests/drivers/flash_simulator/flash_sim_impl/src/main.c
@@ -301,10 +301,10 @@ ZTEST(flash_sim_api, test_align)
 	zassert_equal(-EINVAL, rc, "Unexpected error code (%d)", rc);
 }
 
-#if defined(CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES) &&	\
-	defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 ZTEST(flash_sim_api, test_double_write)
 {
+#if defined(CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES) &&	\
+	defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 	int rc;
 	/* Test checks behaviour of write when attempting to double write
 	 * selected offset. Simulator, prior to write, checks if selected
@@ -325,12 +325,14 @@ ZTEST(flash_sim_api, test_double_write)
 	rc = flash_write(flash_dev, FLASH_SIMULATOR_BASE_OFFSET,
 				 &data, sizeof(data));
 	zassert_equal(-EIO, rc, "Unexpected error code (%d)", rc);
-}
+#else
+	ztest_test_skip();
 #endif
+}
 
-#if !defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 ZTEST(flash_sim_api, test_ramlike)
 {
+#if defined(CONFIG_FLASH_SIMULATOR_RAMLIKE)
 	/* Within code below there is assumption that the src size is
 	 * equal or greater than the FLASH_SIMULATOR_PROG_UNIT
 	 * (write-block-size) of device.
@@ -440,8 +442,10 @@ ZTEST(flash_sim_api, test_ramlike)
 		} while (i & (sizeof(buf) - 1));
 		zassert_equal(0, rc, "Unexpected value read");
 	}
-}
+#else
+	ztest_test_skip();
 #endif
+}
 
 ZTEST(flash_sim_api, test_get_erase_value)
 {


drivers/flash/flash_simulator.c
```diff
@@ -164,8 +164,12 @@ static const struct flash_parameters flash_sim_parameters = {
 	.write_block_size = FLASH_SIMULATOR_PROG_UNIT,
 	.erase_value = FLASH_SIMULATOR_ERASE_VALUE,
 	.caps = {
-#if !defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
+#if defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 		.no_explicit_erase = false,
+#elif defined(CONFIG_FLASH_SIMULATOR_RAMLIKE)
+		.no_explicit_erase = true,
+#else
+#error "Flash Simulator erase capability not configured correctly"
```
rkel: Nitpick:

```suggestion
#error "Flash Simulator erase capability configured incorrectly"
```

tests/drivers/flash_simulator/flash_sim_impl/src/main.c
```diff
@@ -301,10 +301,10 @@ ZTEST(flash_sim_api, test_align)
 	zassert_equal(-EINVAL, rc, "Unexpected error code (%d)", rc);
 }
 
-#if defined(CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES) &&	\
-	defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 ZTEST(flash_sim_api, test_double_write)
 {
+#if defined(CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES) &&	\
+	defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
```
rkel, line 307:
```suggestion
	if (!IS_ENABLED(CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES) ||
	    !IS_ENABLED(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)) {
		ztest_test_skip();
	}
```

tests/drivers/flash_simulator/flash_sim_impl/src/main.c
```diff
@@ -325,12 +325,14 @@ ZTEST(flash_sim_api, test_double_write)
 	rc = flash_write(flash_dev, FLASH_SIMULATOR_BASE_OFFSET,
 				 &data, sizeof(data));
 	zassert_equal(-EIO, rc, "Unexpected error code (%d)", rc);
-}
+#else
+	ztest_test_skip();
```
rkel, line 329: Replaced by compilation level conditional.

tests/drivers/flash_simulator/flash_sim_impl/src/main.c
```diff
@@ -325,12 +325,14 @@ ZTEST(flash_sim_api, test_double_write)
...
 #endif
+}
 
-#if !defined(CONFIG_FLASH_SIMULATOR_EXPLICIT_ERASE)
 ZTEST(flash_sim_api, test_ramlike)
 {
+#if defined(CONFIG_FLASH_SIMULATOR_RAMLIKE)
```
rkel, line 335: Similarly like above - as long as it would compile (for example option does not remove some structure fields from existence) i would opt for compilation level conditionals.

rkel (approved):
Comments - more like style and nitpicking. Not a blocker.

tests/drivers/flash_simulator/flash_sim_impl/src/main.c
de-nordic: All the other paths in test are covered with exactly the same style. I can do the change, as suggested but I would like to do it in separate PR to make the test look uniformly at once.

rkel: Sure. As I had mentioned already - not a blocker.

rghaddab (dismissed):

moonlight83340 (approved):

nashif (approved):

tpambor (approved):

PR #100368, opened 2025-12-02
Bluetooth: Mesh: Document Static OOB entropy requirement

akredalen:
Updates documentation for clarifying that for secure provisioning
with the BTM_ECDH_P256_HMAC_SHA256_AES_CCM algorithm,
the Static OOB value should contain more than 128 bits of
entropy to provide adequate security against attacks.

### DIFF
--- a/doc/connectivity/bluetooth/api/mesh/provisioning.rst
+++ b/doc/connectivity/bluetooth/api/mesh/provisioning.rst
@@ -124,7 +124,9 @@ provisionee:
 
 * **Static OOB:** An authentication value is assigned to the device in
   production, which the provisioner can query in some application specific
-  way.
+  way. For secure provisioning with the BTM_ECDH_P256_HMAC_SHA256_AES_CCM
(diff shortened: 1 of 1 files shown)


include/zephyr/bluetooth/mesh/main.h
```diff
@@ -143,7 +143,12 @@ struct bt_mesh_prov {
 	 */
 	const uint8_t *private_key_be;
 
-	/** Static OOB value */
+	/** Static OOB value
+	 *
+	 *  For secure provisioning with the BTM_ECDH_P256_HMAC_SHA256_AES_CCM algorithm,
```
alxelax: I guess it is better to add to the real documentation rather to API description.
Here https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/connectivity/bluetooth/api/mesh/provisioning.rst?plain=1#L125-L127

alxelax (changes requested):

alxelax (approved):

PavelVPV (approved):

PR #100370, opened 2025-12-02
doc: usb: add VID/PID documentation for the new USB device stack

jfischer-no:
The content is almost identical to that of the legacy stack because we either ported or rewrote all the samples.


josuah (approved):
The links all work:
https://builds.zephyrproject.io/zephyr/pr/100370/docs/connectivity/usb/device_next/vid_pid.html

The legacy stack for comparison:
https://builds.zephyrproject.io/zephyr/pr/100370/docs/connectivity/usb/device/usb_device.html#usb-vendor-and-product-identifiers

fabiobaltieri (approved):

PR #100375, opened 2025-12-02
net: context: release context mutex before clearing IN_USE flag

marcowidmer:
A race condition was observed with concurrent net_context_get() and net_context_unref() calls: If net_context_unref() is interrupted just after clearing the NET_CONTEXT_IN_USE flag, net_context_get() may try to reuse the context and reinitialize its mutex through k_mutex_init(). Once net_context_unref() resumes, it calls k_mutex_unlock() on the newly initialized mutex, which may lead to the following assertion:

ASSERTION FAIL [mutex->lock_count > 0U] @ zephyr/kernel/mutex.c

Fix this by unlocking the context mutex before clearing the NET_CONTEXT_IN_USE flag.

### DIFF
--- a/subsys/net/ip/net_context.c
+++ b/subsys/net/ip/net_context.c
@@ -705,11 +705,11 @@ int net_context_unref(struct net_context *context)
 
 	net_context_set_state(context, NET_CONTEXT_UNCONNECTED);
 
+	k_mutex_unlock(&context->lock);
+
 	context->flags &= ~NET_CONTEXT_IN_USE;
 
 	NET_DBG("Context %p released", context);
-
-	k_mutex_unlock(&context->lock);
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


rlubos (commented):
Nice finding! The fix looks reasonable, however I think we still hit the issue if `net_context_put()` is used? Specifically:
https://github.com/zephyrproject-rtos/zephyr/blob/f45b55404fcaf7b979aa51625fff2e058763962f/subsys/net/ip/net_context.c#L743-L747

Specifically, mutex is unlocked after `net_context_unref()`, so in case of such a race we have the same problem here? We should probably only call `net_context_unref()` after releasing the mutex.

marcowidmer:
@rlubos Good point, here is one more attempt to fix the issue.

rlubos (approved):

jukkar (approved):

PR #100378, opened 2025-12-02
net: ocpp: readonly is a bool

markushx:
GetConfiguration.conf read_only is a bool according to https://openchargealliance.org/wp-content/uploads/2025/04/OCPP_1.6_documentation.zip.

The Python OCPP implementation (used in HomeAssistant) reports an error if read_only is reported as an integer: https://github.com/mobilityhouse/ocpp/blob/ae716a1507d708d83743d1132587cbb03522cda7/ocpp/v16/datatypes.py#L73

Should this be differentiated JSON_TOK_TRUE/JSON_TOK_FALSE based on msg.configuration_key[0].readonly?

### DIFF
--- a/subsys/net/lib/ocpp/ocpp_j.c
+++ b/subsys/net/lib/ocpp/ocpp_j.c
@@ -410,7 +410,7 @@ static int frame_getconfig_msg(char *buf, int len, char *key, char *val,
 		JSON_OBJ_DESCR_PRIM_NAMED(struct json_ocpp_key_val, "key",
 					  key, JSON_TOK_STRING),
 		JSON_OBJ_DESCR_PRIM_NAMED(struct json_ocpp_key_val, "readonly",
-					  readonly, JSON_TOK_NUMBER),
+					  readonly, JSON_TOK_TRUE),
 		JSON_OBJ_DESCR_PRIM_NAMED(struct json_ocpp_key_val, "value",
 					  value, JSON_TOK_STRING),
 	};
--- a/subsys/net/lib/ocpp/ocpp_j.h
+++ b/subsys/net/lib/ocpp/ocpp_j.h
@@ -84,7 +84,7 @@ struct json_ocpp_start_txn_msg {
 struct json_ocpp_getconfig_msg {
 	struct json_ocpp_key_val {
 		char *key;
-		int readonly;
+		bool readonly;
 		char *value;
 	} configuration_key[1];
 


subsys/net/lib/ocpp/ocpp_j.c
```diff
@@ -410,7 +410,7 @@ static int frame_getconfig_msg(char *buf, int len, char *key, char *val,
 		JSON_OBJ_DESCR_PRIM_NAMED(struct json_ocpp_key_val, "key",
 					  key, JSON_TOK_STRING),
 		JSON_OBJ_DESCR_PRIM_NAMED(struct json_ocpp_key_val, "readonly",
-					  readonly, JSON_TOK_NUMBER),
+					  readonly, JSON_TOK_TRUE),
```
SanjayyyV, line 413: The readonly member inside the json_ocpp_getconfig_msg struct in ocpp_j.h has to be also changed to bool instead of int.

ssekar15 (dismissed):
Looks good to me, thanks for contribution.

SanjayyyV (changes requested):

subsys/net/lib/ocpp/ocpp_j.c
markushx: I was wondering whether this needs to be dynamically changed between `JSON_TOK_TRUE `and `JSON_TOK_FALSE`? I am not very familiar with the Zephyr JSON library and OCPP yet.

SanjayyyV: IMO, switching between `JSON_TOK_TRUE` and `JSON_TOK_FALSE` isn’t necessary.
Because during the actual encoding, both `JSON_TOK_TRUE` and `JSON_TOK_FALSE` would fallback to call the same bool_encode function as shown below from `lib/utils/json.c`:
```
        case JSON_TOK_FALSE:
        case JSON_TOK_TRUE:
                return bool_encode(ptr, append_bytes, data);
```
Also the `tests/lib/json/src/main.c` would demonstrate the same.

So the actual boolean value in the readonly variable is what determines the output, not the token type.

jukkar (dismissed):

subsys/net/lib/ocpp/ocpp_j.c
markushx: OK, changed readonly member inside the json_ocpp_getconfig_msg struct to bool.

ssekar15 (approved):

SanjayyyV (approved):

jukkar (approved):

PR #100380, opened 2025-12-02
soc: microchip: mec: Remove MECH172x the HAL based version of MEC172x

scottwcpg:
The goal is to slowly convert all Microchip MEC to in-tree drivers and support code (no HAL). We had partial HAL support for MEC172x as the mech172x folder. We are removing HAL based mech172x since HAL will be deprecated in the future and it has no board support.


jvasanth1 (dismissed):

albertofloyd (dismissed):

jvasanth1 (approved):

albertofloyd (approved):

VenkatKotakonda (approved):
LGTM

PR #100384, opened 2025-12-02
drivers: tmc2209: use m0/m1-gpios instead of msx-gpios

### LINKED ISSUE #88066: dts: stepper: unify msx-gpios

At present bindings for micro-step-resolution-gpios is done in a different manner between different stepper drivers. For instance,
- https://github.com/jilaypandya/zephyr/blob/main/dts/bindings/stepper/adi/adi%2Ctmc2209.yaml
- https://github.com/jilaypandya/zephyr/blob/main/dts/bindings/stepper/ti/ti%2Cdrv8424.yaml

Place msx-gpios in stepper-controller.yaml as a common binding and adjust the two stepper drivers that use m0-gpios and m1-gpios

It is either or between msx-gpios and m0/m1/m2-gpios. 
msx-gpios is extensible and can be used for all the drivers that have 1-3 pins for setting micro-step-resolution.

**Additional context**

jilaypandya:
1. Unify msx-gpios for step-dir drivers. tmc2209 is an exception for now and hence migrating to m0/m1-gpios instead of msx-gpios.
2. Generalize drv84xx tests for all step-dir drivers.
3. Added entry in migration-guide


drivers/stepper/adi_tmc/tmc22xx/tmc22xx.c
```diff
@@ -92,18 +97,28 @@ static int tmc22xx_stepper_configure_msx_pins(const struct device *dev)
 	const struct tmc22xx_config *config = dev->config;
 	int ret;
 
-	for (uint8_t i = 0; i < MSX_PIN_COUNT; i++) {
-		if (!gpio_is_ready_dt(&config->msx_pins[i])) {
-			LOG_ERR("MSX pin %u are not ready", i);
-			return -ENODEV;
-		}
+	if (!gpio_is_ready_dt(&config->m0_pin)) {
+		LOG_ERR("MS1 pin not ready");
```
jbehrensnx, line 100: I would use MS0 and MS1 pins, to keep it consistent with the variable names and dt entries.

drivers/stepper/adi_tmc/tmc22xx/tmc22xx.c
```diff
@@ -124,18 +139,16 @@ static int tmc22xx_stepper_init(const struct device *dev)
 		return ret;
 	}
 
-	if (config->msx_pins) {
-		ret = tmc22xx_stepper_configure_msx_pins(dev);
-		if (ret < 0) {
-			LOG_ERR("Failed to configure MSX pins: %d", ret);
-			return ret;
-		}
+	ret = tmc22xx_stepper_configure_msx_pins(dev);
```
jbehrensnx: If I understand this correctly, with these changes the init functions fails when both microstep pins are undefined, which is different from the current state.

dts/bindings/stepper/adi/adi,tmc2209.yaml
```diff
@@ -25,13 +25,13 @@ properties:
...
 
-  msx-gpios:
+  m0-gpios:
     type: phandle-array
-    description: |
-      An array of GPIO pins for configuring the microstep resolution of the driver.
-      The pins should be listed in the following order:
-      - MS1
-      - MS2
+    description: Microstep configuration pin MS1.
```
jbehrensnx, line 30: See `tmc22xx.c` concerning the pin names.

tests/drivers/stepper/step_dir/boards/native_sim_ti_drv84xx.overlay
```diff
@@ -0,0 +1,30 @@
...
+/ {
+	aliases {
+		step-dir-driver = &ti_drv84xx;
+	};
+
+	ti_drv84xx: ti_drv84xx {
+		status = "okay";
+		compatible = "ti,drv84xx";
+
+		dir-gpios = <&gpio1 0 0>; /* D3 */
```
jbehrensnx: NIT: The `DX` style comments should be removed. They are remnants from when these tests were run on real hardware and I forgot to remove them when transitioning to native_sim.

tests/drivers/stepper/step_dir/boards/native_sim_ti_drv84xx.overlay
```diff
@@ -0,0 +1,30 @@
...
+		step-gpios = <&gpio1 1 0>; /* D4 */
+		sleep-gpios = <&gpio2 0 GPIO_ACTIVE_LOW>; /* D2 */
+		en-gpios = <&gpio2 1 0>; /* 5 */
+		m0-gpios = <&gpio3 0 0>;
+		m1-gpios = <&gpio4 1 0>;
+		counter = <&counter0>;
+
+		#address-cells = <1>;
```
jbehrensnx: NIT: The following three lines should also be no longer necessary if I am not mistaken.

tests/drivers/stepper/step_dir/CMakeLists.txt
```diff
@@ -4,7 +4,8 @@
 cmake_minimum_required(VERSION 3.20.0)
 
 find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
-project(stepper_drv84xx_emul)
+project(step_dir_driver_test)
```
faxe1008, line 7: Just a nit: seems like this was carried over from the rearchitecturing PR.

tests/drivers/stepper/step_dir/src/main.c
```diff
@@ -13,26 +13,67 @@
...
+struct step_dir_fixture {
 	const struct device *dev;
 };
 
+struct stepper_drv_micro_step_res {
+	uint8_t m0: 2;
+	uint8_t m1: 2;
+};
+
+#if CONFIG_DRV84XX
```
faxe1008: I get the idea but this feels to specific inside of the now generalized test.

tests/drivers/stepper/step_dir/src/main.c
```diff
@@ -64,26 +105,29 @@ ZTEST_F(drv84xx_emul, test_enable_gpio_pins)
...
 }
 
-ZTEST_F(drv84xx_emul, test_micro_step_res_set)
+ZTEST_F(step_dir, test_micro_step_res_set)
 {
 	enum stepper_micro_step_resolution res;
 	int value = 0;
 
-	zassert_ok(stepper_set_micro_step_res(fixture->dev, 4));
+	zassert_ok(stepper_set_micro_step_res(fixture->dev, STEPPER_MICRO_STEP_16));
```
jbehrensnx: We cant guarantee that in the future every stepper driver supports this microstep resolution. This also ties into every stepper driver having its own struct for pin configuration - even if just one value is used. This results in this file having to be changed with every new stepper driver.

I have two possible recommendations/alternatives:

The first would be to define 3 new Kconfig symbols:
- The microstep resolution to test
- The m0 value for this resolution
- The m1 value for this resolution

And then have `.conf` files for each stepper driver where they are set. Then simply use the Kconfig symbols in the test.

The second one (that might go beyond the original scope of this PR) is to keep the structs and instead in the test iterate over every entry in the struct instead in a loop, thus testing that every microstep resolution is set correctly instead of just one (depending on the limitations of the gpio emulator the resolutions for the drv84xx driver might need to be limited to just the ones with only `0` and `1`).

jbehrensnx (changes requested):
First of all I approve standardization, through I am surprised that it goes in this direction, as I faintly remember a PR where drv84xx and the allegro driver were intended to be modified in the other direction.

The other thing is the test generalization which I feel should also be described in the PR description as an equal (if not possibly more important) part.

dts/bindings/stepper/adi/adi,tmc2209.yaml
jilaypandya: https://www.analog.com/media/en/technical-documentation/data-sheets/tmc2209_datasheet_rev1.09.pdf

tests/drivers/stepper/step_dir/CMakeLists.txt
jilaypandya: Eventually this would be the test-suite for stepper_drv drivers, so this is actually an intermediate name. Could think of a better name as of now :-)

tests/drivers/stepper/step_dir/src/main.c
jilaypandya: Addressing another comment in the PR along, i will move this to Kconfig. However, we would be then limited to testing only one particular micro-step resolution.

dts/bindings/stepper/adi/adi,tmc2209.yaml
jbehrensnx: Ah, the good old case of different vendors using different names for the same thing. With other words, there is no good solution. The pin names are M1 and M2, but the properties are m0 and m1 for consistency with the other drivers. Might possibly worth it to make a short note about this in the description of one of the pins to clear up the confusion, but this would be on the level of NIT for me.

jbehrensnx (dismissed):

jbehrensnx (approved):

MaureenHelm (approved):

PR #100389, opened 2025-12-02
drivers: ieee802154: SSED attachment management

xavraz:
Few functions have been corrected to the SSED attachment for the MCXW72 as leader :
- mcxw_tx(): tx_frame.tx_delay corrected. Adjusts the scheduled transmission start time to account for the 192µs PHY preamble/header overhead before the MAC payload, converting to symbol units, or defaults to immediate transmission if the time is too soon.

- mcxw_rx_thread(): Stores the frame counter and key ID from a security-enabled enhanced ACK into the received packet's metadata for cryptographic verification.

- mcxw_tx_started() added: Notifies the registered event handler that a frame transmission has started, passing the frame buffer as context.

- pd_mac_sap_handler(): timestamp corrected

- mcxw_configure(): IEEE802154_CONFIG_EVENT_HANDLER case updated


samples/net/openthread/shell/prj-ot-host.conf
```diff
@@ -55,3 +51,8 @@ CONFIG_OPENTHREAD_SRP_CLIENT=y
 CONFIG_OPENTHREAD_SRP_SERVER=y
 CONFIG_OPENTHREAD_MAC_FILTER=y
 CONFIG_OPENTHREAD_IP6_FRAGM=y
+
+CONFIG_OPENTHREAD_PLATFORM_CSL_UNCERT=20
+CONFIG_IEEE802154_MCXW_CSL_ACCURACY=50
```
rlubos: Driver-specific configs should not be placed in a generic config file,  please put it in board specific config in `boards` subdirectory.

rlubos:
I don't think assignees were set correctly given it's a change in NXP driver.

samples/net/openthread/shell/prj-ot-host.conf
xavraz: Hello Robert,
Thank you for your review. I will modified this.

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -540,10 +552,22 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 #if defined(CONFIG_NET_PKT_TXTIME)
 	case IEEE802154_TX_MODE_TXTIME:
 	case IEEE802154_TX_MODE_TXTIME_CCA:
-		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt);
+		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt) / NSEC_PER_USEC;
 		msg->msgData.dataReq.startTime =
 			rf_adjust_tstamp_from_app(mcxw_radio->tx_frame.tx_delay);
-		msg->msgData.dataReq.startTime /= IEEE802154_SYMBOL_TIME_US;
+
+		if ((msg->msgData.dataReq.startTime > 192) &&
```
George-Stefan: Why do you need to subtract 192 and not 160 as it's done in sub_mac.cpp?
https://github.com/zephyrproject-rtos/openthread/blob/main/src/core/mac/sub_mac.cpp#L440

Shouldn't this be in radio.c since it's Thread specific?
^^ @rlubos

samples/net/openthread/shell/prj-ot-host.conf
axelnxp: @xavraz I suggest to update the default value of `CONFIG_IEEE802154_MCXW_CSL_ACCURACY` directly in the original kconfig since it's MCXW specific already, this way it's more global to the platform.
You might also consider setting `CONFIG_OPENTHREAD_PLATFORM_CSL_UNCERT` in a defconfig (likely at soc level for this one), unless it's very specific to this application. If it's a mcxw platform requirement, put it in a defconfig.

samples/net/openthread/shell/prj-ot-host.conf
```diff
@@ -55,3 +51,8 @@ CONFIG_OPENTHREAD_SRP_CLIENT=y
...
+CONFIG_NET_PKT_TXTIME=y
+CONFIG_NET_PKT_TIMESTAMP=y
```
axelnxp: My understanding is that those are required by the driver, am I right ?
If so, make it explicit in the driver kconfig.
Either use `select` (or `imply` if it's optional, meaning the driver can build without it, which I doubt anyway) or `depends on` (if you want to put the reponsibility to the application to enable it).
I'd go for `select` in this case.

drivers/ieee802154/ieee802154_mcxw.c
JarEmb: it is thread specific, I agree to put it in radio.c
In Thread 1.4 spec the CSL Phase indicates the first symbol of the MAC header in consequence, the transmission shall start 192us earlier. 192us is the radio header duration.

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -834,13 +860,15 @@ static void mcxw_receive_at(uint8_t channel, uint32_t start, uint32_t duration)
 	__ASSERT_NO_MSG(mcxw_ctx.state == RADIO_STATE_SLEEP);
 	mcxw_ctx.state = RADIO_STATE_RECEIVE;
 
+	/* For the Endpoint (SSED) purpose */
+	start_csl_receiver();
```
JarEmb: Why this is added? The code after set the PHY to Rx so it is redundant

George-Stefan: You need to prevent NBU to go into low power after a successful csl reception so that timestamp conversion is valid at data indication event.
You need to call stop_csl_receiver() after timestamp conversion.

drivers/ieee802154/ieee802154_mcxw.c
rlubos: That should be documented better probably in the generic ieee802154 radio driver API, but the silent assumption was that the timestamp points to the start of the frame's PHR (I think it was the same as OT, at least at a time). 
Whatever is expected by the individual vendor driver, it's up to the driver to align the timestamp, in nRF we also subtract the SHR duration (160 us) as that's what our driver expects.

samples/net/openthread/shell/prj-ot-host.conf
```diff
@@ -5,8 +5,8 @@ CONFIG_NETWORKING=y
 CONFIG_NET_L2_OPENTHREAD=y
 
 # Logging
-CONFIG_LOG=y
-CONFIG_NET_LOG=y
+CONFIG_LOG=n
+CONFIG_NET_LOG=n
```
axelnxp: If you really want to disable logs by default in the app, do it at board level (create a <board>.conf in the `boards` subfolder) and not in the common configuration file

axelnxp (dismissed):

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -540,10 +549,21 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
...
-		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt);
+		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt) / NSEC_PER_USEC;
 		msg->msgData.dataReq.startTime =
 			rf_adjust_tstamp_from_app(mcxw_radio->tx_frame.tx_delay);
-		msg->msgData.dataReq.startTime /= IEEE802154_SYMBOL_TIME_US;
+
+		/* Compensates the 192µs SHR duration by adjusting the MAC payload start
+		 * time backwards and converting it to symbol periods.
+		 */
+		if ((msg->msgData.dataReq.startTime > IEEE802154_SHR_DURATION_US) &&
```
George-Stefan: start time expiration is done on the NBU side by the PHY.
There is no need to do any validation here

drivers/ieee802154/ieee802154_mcxw.c
George-Stefan: https://github.com/zephyrproject-rtos/zephyr/pull/100389#discussion_r2589135471

I'd like to conclude on this topic.
160 vs 192us.

Thread 1.4 spec uses 192us but sub_mac, nRF driver use 160.

Frame timetamp is at SFD reception before LEN field, but the MAC header is after LEN field.
Thread spec defines the start of MAC header as CSL phase.

The OT code uses the frame timestamp for scheduling => CSL phase - 32us.

That's why subtracting 160us is correct and not 192

@rlubos please confirm

rlubos: Sorry I'm not up to date with Thread 1.4 specs, but looking at the OpenThread documentation:
https://github.com/zephyrproject-rtos/openthread/blob/2bc7712f57af22058770d1ef131ad3da79a0c764/include/openthread/platform/radio.h#L282-L284
They claim to set the timestamp to the end of SFD field (so start of PHR as I wrote above), and in OpenThread radio layer:
https://github.com/zephyrproject-rtos/zephyr/blob/ab09c94377c562deb7f1b1feca1fce55e1819e02/modules/openthread/platform/radio.c#L419-L423
We just pass the value as-is to the driver.

So the timestamp the driver receives should be the start of PHR.

drivers/ieee802154/ieee802154_mcxw.c
George-Stefan: start time expiration check is done on the NBU side by the PHY.
There is no need to do any validation here.

Keep this line only
`msg->msgData.dataReq.startTime -= IEEE802154_SHR_DURATION_US;`

drivers/ieee802154/ieee802154_mcxw.h
```diff
@@ -15,57 +15,66 @@
...
+#define NMAX_RXRING_BUFFERS (8)
+#define RX_ON_IDLE_START    (1)
+#define RX_ON_IDLE_STOP     (0)
+
+#define PHY_TMR_MAX_VALUE (0x00FFFFFF)
+
+/* Duration in microseconds of the IEEE 802.15.4 Synchronization Header
+ * (preamble + SFD) transmitted before the MAC payload.
+ */
+#define IEEE802154_SHR_DURATION_US 192
```
George-Stefan: SHR is preamble (4 bytes) + SFD (1 byte) => 160us

See https://github.com/zephyrproject-rtos/zephyr/pull/100389#discussion_r2584705376

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -600,16 +620,13 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 	if (phy_status == gPhySuccess_c) {
 		mcxw_radio->tx_status = 0;
 		mcxw_radio->state = RADIO_STATE_TRANSMIT;
```
George-Stefan: move `mcxw_radio->state = RADIO_STATE_TRANSMI` before `MAC_PD_SapHandler()` to avoid a race condition with tx aborted event. Don't update the state if tx request (`MAC_PD_SapHandler()`) fails.
For rx, it's correct https://github.com/nxp-upstream/zephyr/blob/ecfda026b2b74885c68e1b279badce7b80fdc7b1/drivers/ieee802154/ieee802154_mcxw.c#L308

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -1005,8 +1026,8 @@ static int mcxw_set_channel(const struct device *dev, uint16_t channel)
 
 static net_time_t mcxw_get_time(const struct device *dev)
 {
-	static uint64_t sw_timestamp;
-	static uint64_t hw_timestamp;
+	static uint64_t sw_timestamp; /* µs, on the Host side, monotonous */
```
George-Stefan: `Host side` is confusing since this is a full SoC build.
`last timestamp` is more appropriate

George-Stefan (changes requested):

drivers/ieee802154/ieee802154_mcxw.c
xavraz: Hello George,
I have removed the test.

JA-NXP (approved):

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -540,10 +549,10 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 #if defined(CONFIG_NET_PKT_TXTIME)
 	case IEEE802154_TX_MODE_TXTIME:
 	case IEEE802154_TX_MODE_TXTIME_CCA:
-		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt);
+		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt) / NSEC_PER_USEC;
```
George-Stefan, line 552: - Add back the SHR duration of 160us and subtract it from tx_frame.tx_delay before converting it.
`mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt) / NSEC_PER_USEC - SHR_duration;`

- Add a comment about why this is needed: it's because of the OT radio API
https://github.com/zephyrproject-rtos/openthread/blob/2bc7712f57af22058770d1ef131ad3da79a0c764/include/openthread/platform/radio.h#L282-L284

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -596,20 +605,17 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 
 	k_sem_reset(&mcxw_radio->tx_wait);
 
+	mcxw_radio->state = RADIO_STATE_TRANSMIT;
```
George-Stefan, line 611: Save current state 
```
tmp_state = mcxw_radio->state;
mcxw_radio->state = RADIO_STATE_TRANSMIT;
```
and restore it on the error case
```
mcxw_radio->state = tmp_state;
return -EIO;
```

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -1005,8 +1015,8 @@ static int mcxw_set_channel(const struct device *dev, uint16_t channel)
 
 static net_time_t mcxw_get_time(const struct device *dev)
 {
-	static uint64_t sw_timestamp;
-	static uint64_t hw_timestamp;
+	static uint64_t sw_timestamp; /* µs, on the last timestamp, monotonous */
```
George-Stefan: typo: `on the last timestamp` should be `the last timestamp`

George-Stefan (changes requested):
Please update the 3rd commit message:
```
[dts: arm: nxp: MCXW72 corrected on the lptmr clock frequency]
The lptmr clock frequency was wrong and equal to 32000.
The clock frequency is 32MHz (32768).
```
`32MHz` should be `32.768KHz`

Please update the last commit message:
```
tx_frame.tx_delay corrected.
  Adjusts the scheduled transmission start time to account for the
  192µs PHY preamble/header overhead before the MAC payload,
  converting to symbol units, or defaults to immediate transmission
  if the time is too soon.
```

should be
```
Adjusts the scheduled transmission start time to account for the SHR
duration (160us) as requested by the OpenThread radio API
```

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -540,10 +549,11 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 #if defined(CONFIG_NET_PKT_TXTIME)
 	case IEEE802154_TX_MODE_TXTIME:
 	case IEEE802154_TX_MODE_TXTIME_CCA:
-		mcxw_radio->tx_frame.tx_delay = net_pkt_timestamp_ns(pkt);
+		mcxw_radio->tx_frame.tx_delay -= IEEE802154_SHR_DURATION_US;
```
George-Stefan: This is not correct,
`tx_delay` is in nanoseconds while SHR duration is in us.

Please fix as explained in https://github.com/zephyrproject-rtos/zephyr/pull/100389/files#r2601444923

George-Stefan (changes requested):

drivers/ieee802154/ieee802154_mcxw.c
```diff
@@ -596,20 +606,20 @@ static int mcxw_tx(const struct device *dev, enum ieee802154_tx_mode mode, struc
 
 	k_sem_reset(&mcxw_radio->tx_wait);
 
+	mcxw_radio_state tmp_state = mcxw_radio->state;
+	mcxw_radio->state = RADIO_STATE_TRANSMIT;
+
 	phy_status = MAC_PD_SapHandler(msg, ot_phy_ctx);
 	if (phy_status == gPhySuccess_c) {
 		mcxw_radio->tx_status = 0;
```
George-Stefan: Remove `mcxw_radio->tx_status = 0;`
It's set by the PD/PLME callbacks

George-Stefan (changes requested):

drivers/ieee802154/ieee802154_mcxw.c
xavraz: Corrected

drivers/ieee802154/ieee802154_mcxw.c
George-Stefan: Please fix

drivers/ieee802154/ieee802154_mcxw.c
George-Stefan: Please fix

drivers/ieee802154/ieee802154_mcxw.c
George-Stefan: Please fix

George-Stefan (changes requested):

George-Stefan (dismissed):

George-Stefan (approved):

axelnxp (approved):

rlubos (approved):

PR #100392, opened 2025-12-02
drivers: pwm: Fix prescaler 256 overflow and get cycles

PetervdPerk-NXP:
Setting a prescaler of 256 would overflow the uint8_t. Also get cycles calculation was wrong.

### DIFF
--- a/drivers/pwm/pwm_nxp_flexio.c
+++ b/drivers/pwm/pwm_nxp_flexio.c
@@ -67,7 +67,7 @@ struct pwm_nxp_flexio_channel_config {
 	/** Counter decrement clock prescaler */
 	enum pwm_nxp_flexio_prescaler prescaler;
 	/** Actual Prescaler divisor */
-	uint8_t prescaler_div;
+	uint16_t prescaler_div;
 };
 
 struct pwm_nxp_flexio_pulse_info {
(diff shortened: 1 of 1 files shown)


drivers/pwm/pwm_nxp_flexio.c
```diff
@@ -191,15 +191,8 @@ static int pwm_nxp_flexio_get_cycles_per_sec(const struct device *dev,
 	struct pwm_nxp_flexio_data *data = dev->data;
 	struct pwm_nxp_flexio_channel_config *pwm_info;
 
-	/* If get_cycles is called directly after init */
-	if (data->period_cycles[channel] == 0) {
```
FelixWang47831, line 195: I'm OK for this patch.
It seems period_cycles can be removed from pwm_nxp_flexio_data structure.

FelixWang47831 (approved):

Holt-Sun (dismissed):

drivers/pwm/pwm_nxp_flexio.c
```diff
@@ -191,15 +187,8 @@ static int pwm_nxp_flexio_get_cycles_per_sec(const struct device *dev,
 	struct pwm_nxp_flexio_data *data = dev->data;
 	struct pwm_nxp_flexio_channel_config *pwm_info;
 
-	/* If get_cycles is called directly after init */
```
zejiang0jason, line 194: Is this not needed?

PetervdPerk-NXP: No because get_cycles is simply the clock divided by prescaler which always works.

zejiang0jason (approved):

Holt-Sun (approved):

mmahadevan108 (approved):

PR #100405, opened 2025-12-02
sensor: rm3100: Add missing cycle_count in edata header.

anthony289478:
Fixes missing cycle_count configuration in `edata->header.cycle_count` which would result in decoding of stream data to always use the 600hz scale factor and values off by a factor of 2.

This was configured when performing one-shot submissions but missed for stream submissions.

Also corrects the units in the comments I found while searching for this problem.

### DIFF
--- a/drivers/sensor/pni/rm3100/rm3100_stream.c
+++ b/drivers/sensor/pni/rm3100/rm3100_stream.c
@@ -7,6 +7,7 @@
 
 #include <zephyr/drivers/sensor.h>
 #include <zephyr/drivers/sensor_clock.h>
+#include <zephyr/dt-bindings/sensor/rm3100.h>
 #include <zephyr/rtio/rtio.h>
 #include <zephyr/sys/check.h>
 #include "rm3100_stream.h"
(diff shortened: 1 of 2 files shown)


drivers/sensor/pni/rm3100/rm3100_stream.c
```diff
@@ -29,6 +30,12 @@ static void rm3100_complete_result(struct rtio *ctx, const struct rtio_sqe *sqe,
 				     data->stream.settings.enabled.drdy);
 	edata->header.channels = 0;
 
+	if (data->settings.odr == RM3100_DT_ODR_600) {
+		edata->header.cycle_count = RM3100_CYCLE_COUNT_HIGH_ODR;
+	} else {
+		edata->header.cycle_count = RM3100_CYCLE_COUNT_DEFAULT;
+	}
```
ubieda: I see this was present in one-shot mode through `rm3100_encode`. I suggest to use in streaming as well.

anthony289478: Updated

ubieda (approved):

MaureenHelm (approved):

PR #98578, opened 2025-10-30
DCache: Add __dcacheline_aligned & __dcacheline_exclusive attributes

JarmouniA:
New version of #95888

Add `__dcacheline_aligned` attribute which ensures data cache line alignment for variables.

Add `__dcacheline_exclusive_noinit` & `__dcacheline_exclusive_data` attributes which make variables aligned to data cache line size and having exclusive data cache line(s).

Usage:
`static uint8_t var1 __dcacheline_aligned;`
`static uint8_t var2 __dcacheline_exclusive_noinit;`
`__dcacheline_exclusive_data static uint8_t var3 = 5;`

Note: Following #110984, #112811 and #112845, `CONFIG_DCACHE_LINE_SIZE` no longer depends on `CONFIG_CACHE_MANAGEMENT=y` and on `CONFIG_DCACHE_LINE_SIZE_DETECT=n`, and its value can never be `<2` when `CONFIG_DCACHE=y`.

PR #100409, opened 2025-12-02
drivers: sdmmc: stm32: introduce cache flush & invalidation for DMA

### LINKED ISSUE #100377: drivers: disk: sdmmc_stm32: Mount fails for H7 with D-Cache enabled
### Describe the bug

Attempting to use an STM32H753 (custom board), with a MicroSD card connected to SDMMC1 (4-bit, IDMA enabled, one partition, FAT).

When trying to mount the filesystem, an error occurs. However querying the card geometry succeeds.

Disabling D-Cache (`CONFIG_DCACHE=n`), or applying the patch below resolves the issue.

### Regression

- [ ] This is a regression.

### Steps to reproduce

```dts
&sdmmc1 {
    clocks = <&rcc STM32_CLOCK(AHB3, 16U)>,
        <&rcc STM32_SRC_PLL1_Q SDMMC_SEL(0)>; /* 48 MHz */
    clk-bypass;
    idma;
    pinctrl-0 = <
        &sdmmc1_d0_pc8
        &sdmmc1_d1_pc9
        &sdmmc1_d2_pc10
        &sdmmc1_d3_pc11
        &sdmmc1_ck_pc12
        &sdmmc1_cmd_pd2
    >;
    pinctrl-names = "default";
    bus-width = <4>;
    pwr-gpios = <&gpioi 1 GPIO_ACTIVE_HIGH>;
    disk-name = "SD";
    status = "okay";
};
```

```c
static FATFS fatfs_work;
static struct fs_mount_t fatfs_mnt = {
    .type = FS_FATFS,
    .fs_data = &fatfs_work,
    .mnt_point = "/SD:",
    .flags = FS_MOUNT_FLAG_NO_FORMAT,
    .storage_dev = DEVICE_DT_GET(DT_NODELABEL(sdmmc1)),
};

void sdmmc_setup(void) {
    int ret;

	/* copied from samples/subsys/fs/fs_sample/src/main.c */ 
    do {
        static const char *disk_pdrv = "SD";
        uint64_t memory_size_mb;
        uint32_t block_count;
        uint32_t block_size;

        if (disk_access_ioctl(disk_pdrv,
                DISK_IOCTL_CTRL_INIT, NULL) != 0) {
            LOG_ERR("Storage init ERROR!");
            break;
        }

        if (disk_access_ioctl(disk_pdrv,
                DISK_IOCTL_GET_SECTOR_COUNT, &block_count)) {
            LOG_ERR("Unable to get sector count");
            break;
        }
        LOG_INF("Block count %u", block_count);

        if (disk_access_ioctl(disk_pdrv,
                DISK_IOCTL_GET_SECTOR_SIZE, &block_size)) {
            LOG_ERR("Unable to get sector size");
            break;
        }
        printk("Sector size %u\n", block_size);

        memory_size_mb = (uint64_t)block_count * block_size;
        printk("Memory Size(MB) %u\n", (uint32_t)(memory_size_mb >> 20));

        if (disk_access_ioctl(disk_pdrv,
                DISK_IOCTL_CTRL_DEINIT, NULL) != 0) {
            LOG_ERR("Storage deinit ERROR!");
            break;
        }
    } while (0);

    ret = fs_mount(&fatfs_mnt); /* returns -19 / ENODEV! */
    if (ret != 0) {
        fatal();
    }
}
```

### Relevant log output

```shell
*** Booting Zephyr OS build v4.3.0-1476-gd31c6e95033f ***
[00:00:01.128,000] <inf> bringup: Block count 62357504
Sector size 512
Memory Size(MB) 30448
[00:00:01.129,000] <err> fs: fs mount error (-19)
```

### Impact

### Environment

- Linux
- arm-none-eabi-gcc (Arm GNU Toolchain 14.3.Rel1 (Build arm-14.174)) 14.3.1 20250623
- Zephyr: d31c6e95033fd6b3763389edba6a655245ae1328

### Additional Context

I've attached a patch - applying this with D-Cache enabled seems to do the trick, but I don't think this is necessarily the correct approach / location...

[0001-WIP-Fix-D-Cache-with-STM32H7-SDMMC-IDMA.patch](https://github.com/user-attachments/files/23885925/0001-WIP-Fix-D-Cache-with-STM32H7-SDMMC-IDMA.patch)

attie-argentum:
I encountered issues while mounting a FAT filesystem with D-Cache enabled on an STM32H753.

This patch introduces a call to `sys_cache_data_flush_and_invd_range()` before `HAL_*_ReadBlocks_DMA()`, and `sys_cache_data_flush_range()` before `HAL_*_WriteBlocks_DMA()`.

Fixes #100377

### DIFF
--- a/drivers/disk/sdmmc_stm32.c
+++ b/drivers/disk/sdmmc_stm32.c
@@ -17,6 +17,7 @@
 #include <zephyr/pm/policy.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/irq.h>
+#include <zephyr/cache.h>
 #include <soc.h>
 #include <stm32_ll_rcc.h>
 
@@ -561,12 +562,31 @@ static int stm32_sdmmc_access_read(struct disk_info *disk, uint8_t *data_buf,
 	}
 #endif
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
+	/* A flush is performed before the DMA operation, to prevent accidental data
+	 * loss when the buffer is not properly aligned to the cache-line (e.g:
+	 * 32-bytes for STM32H7).
+	 */
+	sys_cache_data_flush_and_invd_range((void *)data_buf, BLOCKSIZE * num_sector);
+#endif
+
 	err = stm32_sdmmc_read_blocks(&priv->hsd, data_buf, start_sector, num_sector);
 	if (err != 0) {
 		goto end;
 	}
 
 	k_sem_take(&priv->sync, K_FOREVER);
+
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
+	/* Invalidate again after the operation is complete, to protect against
+	 * speculative / spurious reads. Note that this is slightly unsafe when
+	 * `data_buf` is not aligned to the cache line, and shares the cache line
+	 * with other data... Hopefully the previous (otherwise unnecessary) flush &
+	 * invalidate reduces this risk enough.
+	 * This is a balance between forcing callers to have aligned buffers.
+	 */
+	sys_cache_data_invd_range((void *)data_buf, BLOCKSIZE * num_sector);
+#endif
 
 #if STM32_SDMMC_USE_DMA_SHARED
 	if (HAL_DMA_DeInit(&priv->dma_txrx_handle) != HAL_OK) {
@@ -640,6 +660,10 @@ static int stm32_sdmmc_access_write(struct disk_info *disk,
 		err = -EIO;
 		goto end;
 	}
+#endif
+
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
+	sys_cache_data_flush_range((void *)data_buf, BLOCKSIZE * num_sector);
 #endif
 
 	err = stm32_sdmmc_write_blocks(&priv->hsd, (uint8_t *)data_buf, start_sector, num_sector);


drivers/disk/sdmmc_stm32.c
```diff
@@ -518,6 +519,8 @@ static int stm32_sdmmc_read_blocks(HandleTypeDef *hsd, uint8_t *data_buf,
 
 #if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
 
+	sys_cache_data_flush_and_invd_range(data_buf, BLOCKSIZE * num_sector);
```
etienne-lms: Only invalidating cache lines should be enough:
```suggestion
	sys_cache_data_invd_range(data_buf, BLOCKSIZE * num_sector);

**Review by etienne-lms — COMMENTED** [review]
LGTM with a nitpicking comment.

File: drivers/disk/sdmmc_stm32.c (hunk 1 above)
Comment (attie-argentum) [reply drivers/disk/sdmmc_stm32.c]: I left this as "flush+invalidate" because the underling buffer may not be aligned to the 32-byte cache line... There appear to be ways to handle alignment of that buffer for FAT (`FS_FATFS_WINDOW_ALIGNMENT`), but this imposes subtle intricacies that the caller is responsible for and are not immediately apparent... thoughts welcome.

Comment (mathieuchopstm) [reply drivers/disk/sdmmc_stm32.c]: Fine, but please add a comment reflecting this above the call to `flush_and_invd`.

File: drivers/disk/sdmmc_stm32.c
Comment (mathieuchopstm) [inline drivers/disk/sdmmc_stm32.c:1]: I think a D$ invalidate should also be performed on line 572 after the `k_sem_take(sync, K_FOREVER)` in case a speculative/spurious read is performed while DMA is operating.

(really, only one D$ invalidate should be done at this very place, but since we are """""forced""""" to use Flush+Invalidate, then it must be done twice)

Comment (attie-argentum) [reply drivers/disk/sdmmc_stm32.c:1]: Thanks, done!

File: drivers/disk/sdmmc_stm32.c [hunk 2]
```diff
@@ -568,6 +575,13 @@ static int stm32_sdmmc_access_read(struct disk_info *disk, uint8_t *data_buf,
 
 	k_sem_take(&priv->sync, K_FOREVER);
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
+	/* Invalidate after the operation is complete, to protect against
+	 * speculative / spurious reads.
+	 */
+	sys_cache_data_invd_range(data_buf, BLOCKSIZE * num_sector);
```
mathieuchopstm: Really this is still "unsafe" since a write could occur after the Flush+Invalidate that would be lost here. But there's no real solution besides *ensuring the buffer is properly aligned*, so...

Nevertheless, I'd add a comment for the poor soul that may be hit by this:
```suggestion
	/* Invalidate after the operation is complete to ensure CPU sees data
	 * written by the DMA. Note that this is slightly unsafe if `data_buf`
	 * is not aligned to cache line, but hopefully the (otherwise unneeded)
	 * Flush+Invalidate done earlier should reduce risks (only writes done
	 * between the F+I and this Invalidate will be lost, rather than any
	 * pending write on the dirty cache line).
	 */
	sys_cache_data_invd_range(data_buf, BLOCKSIZE * num_sector);
```

attie-argentum: I'd love to propose instead returning early with an error like `-EINVAL` if the buffer isn't aligned... I've had a deeper dig:

It looks like within the context of FAT alone, this will require both `FATFS.win[]` to be aligned (easy, via `FS_FATFS_WINDOW_ALIGNMENT`) and all read buffers to be aligned (i.e: those given to user-called `fs_read()`, but also the internal calls from `ff.c` to `disk_read()` on portions of the buffer (which seems to always advance in multiples of the sector size)... what about other filesystems?

`sys_cache_data_line_size_get()` appears to be reliable only if `CONFIG_DCACHE_LINE_SIZE_DETECT` or `CONFIG_DCACHE_LINE_SIZE` are set - is this enough to cover all STM32 cases? Zero is returned otherwise, so perhaps fall back on `-ENOTSUP`, just in case.

If we're "_content_" enough with a more robust comment, then that may be the best route forward for the moment?

attie-argentum: PS: This has wide-reaching consequences, including `/subsys/fs/shell.c` for example.

JarmouniA: > sys_cache_data_line_size_get() appears to be reliable only if CONFIG_DCACHE_LINE_SIZE_DETECT or CONFIG_DCACHE_LINE_SIZE are set

`CONFIG_DCACHE_LINE_SIZE` is set for all STM32 SoCs.

JarmouniA: Btw cache line alignment is not sufficient here, the buffer would have to occupy 1 or multiple cache lines exclusively, which can only be ensured with padding, or a special linker section, see #98578

attie-argentum: > the buffer would have to occupy 1 or multiple cache lines exclusively

Operating in multiples of `BLOCKSIZE` (seems to be always `512`) should make this true?

JarmouniA: > > the buffer would have to occupy 1 or multiple cache lines exclusively
> 
> Operating in multiples of `BLOCKSIZE` (seems to be always `512`) should make this true?

Indeed.

attie-argentum:
In light of the discussion, and my brief investigation, I've opted to maintain the `F+I` before, and `I` after. The risk stands, comments revised. I've also relocated the `F+I` to sit alongside the `I` for better visibility.

I have confirmed that things "_work_" when aligning `FATFS.win[]` and the buffer passed to `fs_read()`, but the knock-on impact may be difficult to deal with in reality...

drivers/disk/sdmmc_stm32.c
```diff
@@ -561,13 +562,32 @@ static int stm32_sdmmc_access_read(struct disk_info *disk, uint8_t *data_buf,
 	}
 #endif
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```
JarmouniA, line 565:
```suggestion
#if defined(STM32_SDMMC_USE_DMA) || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```

drivers/disk/sdmmc_stm32.c
```diff
@@ -561,13 +562,32 @@ static int stm32_sdmmc_access_read(struct disk_info *disk, uint8_t *data_buf,
...
 	k_sem_take(&priv->sync, K_FOREVER);
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
+	/* Invalidate again after the operation is complete, to protect against
+	 * speculative / spurious reads. Note that this is slightly unsafe when
+	 * `data_buf` is not aligned to the cache line, and shares the cache line
+	 * with other data... Hopefully the previous (otherwise unnecessary) flush &
+	 * invalidate reduces this risk enough.
+	 * This is a balance between forcing callers to have aligned buffers.
+	 */
```
JarmouniA, line 587: Maybe add an assert here that checks if the buffer is aligned to and its size is a multiple of the dcache line size.

drivers/disk/sdmmc_stm32.c
```diff
@@ -561,13 +562,32 @@ static int stm32_sdmmc_access_read(struct disk_info *disk, uint8_t *data_buf,
...
+#endif
+
 	err = stm32_sdmmc_read_blocks(&priv->hsd, data_buf, start_sector, num_sector);
 	if (err != 0) {
 		goto end;
 	}
 
 	k_sem_take(&priv->sync, K_FOREVER);
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```
JarmouniA, line 580:
```suggestion
#if defined(STM32_SDMMC_USE_DMA) || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```

drivers/disk/sdmmc_stm32.c
```diff
@@ -642,6 +662,10 @@ static int stm32_sdmmc_access_write(struct disk_info *disk,
 	}
 #endif
 
+#if STM32_SDMMC_USE_DMA || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```
JarmouniA, line 665:
```suggestion
#if defined(STM32_SDMMC_USE_DMA) || IS_ENABLED(DT_PROP(DT_DRV_INST(0), idma))
```

drivers/disk/sdmmc_stm32.c
attie-argentum: No, it's always defined...

https://github.com/zephyrproject-rtos/zephyr/blob/2aa1ec16b8ac5ed8fa989fbbd315dcd8146ddb70/drivers/disk/sdmmc_stm32.c#L26

drivers/disk/sdmmc_stm32.c
attie-argentum: I considered it and even tested an implementation, but I'm concerned it'll be too unwieldy... (see discussion above)

It might be worth reviewing what other drivers do in this area?

mathieuchopstm: The issue here is only buffer alignment, not size (which is guaranteed multiple of `BLOCKSIZE`=512).

mathieuchopstm (approved):
Better than nothing. Hopefully large buffers are naturally aligned enough that the limitation isn't an issue anyways...

erwango (approved):

PR #100411, opened 2025-12-03
utils: timeutil: option to disable clock skew

JordanYates:
Accurately applying clock skew correction in `timeutil_sync_ref_from_local` and `timeutil_sync_local_from_ref` requires double-precision floating point operations. If this is the only usage of double precision logic in the build, but clocks skews are not used, this can be a significant ROM overhead. Disabling this option can save ~2.4 kB of ROM.

Defaulting the symbol to `y` means there is no application behavior change unless the user takes action.

### DIFF
--- a/tests/unit/timeutil/testcase.yaml
+++ b/tests/unit/timeutil/testcase.yaml
@@ -8,3 +8,7 @@ tests:
 
   utilities.time.64bit:
     extra_args: M64_MODE=1
+
+  utilities.time.no_skew:
+    extra_configs:
+      - CONFIG_TIMEUTIL_APPLY_SKEW=n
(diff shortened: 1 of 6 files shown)


peter-mitsis (dismissed):

tests/unit/timeutil/testcase.yaml
```diff
@@ -8,3 +8,7 @@ tests:
 
   utilities.time.64bit:
     extra_args: M64_MODE=1
+
+  utilities.time.no_skew:
```
nashif, line 12: what does this test? does not seem necessary

JordanYates: Tests that the library compiles when the option is disabled and doesn't affect any other behaviour.
If desired I could either remove the testcase, or add extra tests that validate that the skew is ignored.

nashif: my point here and if I read the code correctly, this test will just not run part of the testsuite, it does not verify any other behavior.

JordanYates: Updated

nashif (approved):

kartben (approved):

Issue #100415, opened 2025-12-03
mimxrt595_evk flexspi2(memc) cannot work properly


KATE-WANG-NXP:
### Describe the bug

Display sample does not work correctly on RT500 for shield rk055hdmipi4ma0 when the frame buffer is placed on FLEXSPI2.
Build and run the example:
west build -p always -b mimxrt595_evk/mimxrt595s/cm33 samples/drivers/display -d test_rt500 --shield rk055hdmipi4ma0
The program never run to main and the panel is blank
However if changing the CONFIG_MCUX_DCNANO_LCDIF_EXTERNAL_FB_MEM=n in [boards\shields\rk055hdmipi4ma0\boards\mimxrt595_evk_mimxrt595s_cm33.conf ](https://github.com/zephyrproject-rtos/zephyr/blob/d0cda31dc4810336f901700588e183e22cc7bfc2/boards/shields/rk055hdmipi4ma0/boards/mimxrt595_evk_mimxrt595s_cm33.conf#L8) which means the frame buffers placed on RAM and CONFIG_MCUX_DCNANO_LCDIF_EXTERNAL_FB_ADDR=0x38400000 will not be used with CONFIG_MEMC=n, the demo works as expected.

### Regression

- [x] This is a regression.

### Steps to reproduce

1. Build and run the example:
west build -p always -b mimxrt595_evk/mimxrt595s/cm33 samples/drivers/display -d test_rt500 --shield rk055hdmipi4ma0
The program never run to main and the panel is blank
2. Change the CONFIG_MCUX_DCNANO_LCDIF_EXTERNAL_FB_MEM=n in [boards\shields\rk055hdmipi4ma0\boards\mimxrt595_evk_mimxrt595s_cm33.conf ](https://github.com/zephyrproject-rtos/zephyr/blob/d0cda31dc4810336f901700588e183e22cc7bfc2/boards/shields/rk055hdmipi4ma0/boards/mimxrt595_evk_mimxrt595s_cm33.conf#L8), the demo works as expected.

### Relevant log output

```shell

```

### Impact

Functional Limitation – Some features not working as expected, but system usable.

Albort12138:
Hi @KATE-WANG-NXP, fail to reproduce the issue your described locally, with same build command.
```
*** Booting Zephyr OS build v4.3.0-2292-g006a95132386 ***
[00:00:12.637,829] <inf> sample: Display sample for lcdif@210000
[...]

PR #100418, opened 2025-12-03
Update dcnano lcdif display controller and rt500 board setting

KATE-WANG-NXP:
1. R500: Move the memory region definition to board dts. It is better to let the FLEXSPI2 memory region in the final linker file be generated from dts, in this way user/developer can place the data in the region in code more easily.
2. Update dcnano_lcdif driver. 
2.1 For the situation when the RAM space is limited and the driver's frame buffer cannot be place in RAM, add a new property 'ext-ram' in the binding, so if in device tree this property is assigned to a secondary RAM space, place the frame buffer there. The old way is to define the frame bufeer address in Kconfig, which may cause inconvenience if other data also needs to be placed in the same RAM space.
2.2 Update the driver to support new requirement on RT700. The IP requires a 64-byte alignment for the frame buffer stride.
2.3 Update the calculation of frame buffer size. If the pixel format is updated, the frame buffer size shall be updated too.
3. RT500/RT700:  Remove the external frame buffer address configuration for shield rk055hdmipi4ma0. In which memory region it shall be placed is obtained from device tree instead.
4. Enable the 720p panel RK055HDMIPI4MA0 for RT700 and add supported panels in board doc.


JarmouniA (commented):
You probably forgot to add `ext-ram = &psram;` to LCDIF DT node.

KATE-WANG-NXP:
Hi @JarmouniA , the frame buffer can be placed both on RAM and PSRAM, after test on PSRAM I found the demo cannot work properly, which can be reproduced on main, I reported an issue https://github.com/zephyrproject-rtos/zephyr/issues/100415
so in this PR I did not add the secondary ram configuration

JarmouniA (dismissed):

JarmouniA (commented):
typo in the 1st commit title, `dedinition` -> `definition`.

JarmouniA (dismissed):

JarmouniA (approved):

zejiang0jason (approved):

PR #100420, opened 2025-12-03
drivers: gpio: tca6424a: fix race condition in port_get_raw()

wosong:
Avoid overwriting cached input state while ISR-deferred work is pending. Previously, port_get_raw() updated pins_state.input via update_input_regs(), which could corrupt previous_state used for transition detection. Fix by reading input registers directly without updating the global cache.

### DIFF
--- a/drivers/gpio/gpio_tca6424a.c
+++ b/drivers/gpio/gpio_tca6424a.c
@@ -318,7 +318,6 @@ static int tca6424a_pin_config(const struct device *dev, gpio_pin_t pin, gpio_fl
 
 static int tca6424a_port_get_raw(const struct device *dev, gpio_port_value_t *value)
 {
-	struct tca6424a_drv_data *const drv_data = dev->data;
 	uint32_t buf;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


henrikbrixandersen (changes requested):
> Avoid overwriting cached input state while ISR-deferred work is pending. Previously, port_get_raw() updated pins_state.input via update_input_regs(), which could corrupt previous_state used for transition detection. Fix by reading input registers directly without updating the global cache.

Isn't this needed? Reading the raw port needs to update the state regardless if any interrupt is enabled.

drivers/gpio/gpio_tca6424a.c
```diff
@@ -329,7 +329,7 @@ static int tca6424a_port_get_raw(const struct device *dev, gpio_port_value_t *va
 
 	k_sem_take(&drv_data->lock, K_FOREVER);
 
-	ret = update_input_regs(dev, &buf);
+	ret = read_port_regs(dev, TCA6424A_REG_INPUT, &buf);
```
fabiobaltieri, line 329: ok then the lock is not really needed anymore right? can you drop it?

wosong: I agree, the lock can be dropped.

fabiobaltieri (approved):

henrikbrixandersen (approved):

PR #100422, opened 2025-12-03
sensor: ti: ina2xx: fix fault on unsupported channels

JordanYates:
Fix calling `sensor_channel_get` on a channel that is not supported by the hardware from faulting with a NULL de-reference.

This fixes the following pattern (used by `fuel_guage/composite`):
```
sensor_sample_fetch(dev);
if (sensor_channel_get(dev, chan, val) == -EINVAL) {
    printk("Not supported\n");
}
````

### DIFF
--- a/drivers/sensor/ti/ina2xx/ina2xx_common.h
+++ b/drivers/sensor/ti/ina2xx/ina2xx_common.h
@@ -74,13 +74,27 @@ struct ina2xx_channel {
  * are available on all INA2XX variants.
  */
 struct ina2xx_channels {
+#ifdef CONFIG_INA2XX_HAS_CHANNEL_BUS_VOLTAGE
 	const struct ina2xx_channel *voltage;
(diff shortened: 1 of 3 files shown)


MaureenHelm (dismissed):

MaureenHelm (dismissed):

MaureenHelm:
@teburd @ubieda can you take a look?

ubieda (dismissed):

JordanYates:
Updated to revert #101961, as the only consequence of removing `NULL` checks is enabling `NULL` dereference exceptions. The sensor API is a generic API, a generic library built on top of that API does not always know which channels a device supports.

ubieda (approved):

JordanYates:
> > Updated to revert #101961, as the only consequence of removing `NULL` checks is enabling `NULL` dereference exceptions. The sensor API is a generic API, a generic library built on top of that API does not always know which channels a device supports.
> 
> Could you please amend the revert commit messages to mention this

Done

MaureenHelm (approved):

PR #96670, opened 2025-09-28
drivers: mspi: Support MSPI driver for STM32. 

ExaltZephyr:
This PR introduces support for MSPI driver on STM32, enabling functionality APIs for MSPI host controllers.

1. **mspi_stm32_xspi driver**

      -  support xspi
      - tested using stm32h573i_dk  and stm32h7s78_dk boards
      - functionalities:
          - [x]             Indirect Mode (read, write and erase).
          - [x]             DTR support.
          - [x]             Memory map mode (read and write). 
          - [x]             xip configuration. 
          - [x]             DMA Mode 
          - [x]              Zephyr Power Management Support 
    
3. **ospi_stm32 driver**
 
     - support octospi
     - tested using stm32h735g_disco and b_u585i_iot02a
     - functionalities:
        - [x]  Indirect mode (read, write and erase)
         - [x]   DTR support 
         - [x]    Memory map mode  
        - [x]   xip configuration
        - [x]  DMA transfer (b_u585i_iot02a board only, hdma isn't supported yet in zephyr)
        - [x]  zephyr power management support

3. **mspi_stm32_qspi driver**
 
     - support quadspi
     - tested using `arduino_giga_r1/stm32h747xx/m7` and `stm32l496g_disco`
     - functionalities:
        - [x]  Indirect mode (read, write and erase) with MSPI_IO_MODE_SINGLE, MSPI_IO_MODE_QUAD_1_1_4, and MSPI_IO_MODE_QUAD_1_4_4
        - [x]  Memory map mode (read)
        - [x]   xip configuration
        - [x]  DMA transfer (stm32l496g_disco)
        - [x]  Zephyr Power Management Support 
 
Notes: 
- This PR was tested with JEDEC flash drivers and has not been tested with PSRAM.
- mspi_stm32_xspi driver - DMA mode here was tested using stm32h573i_dk board only since no zephyr dma support on stm32h7s78_dk

PR #100429, opened 2025-12-03
include: flash: move spi_nor.h to public headers

ExaltZephyr:
This PR moves spi_nor.h from drivers/flash to include/drivers/flash directory and updates affected flash drivers with new include path

### DIFF
--- a/drivers/flash/flash_andes_qspi.c
+++ b/drivers/flash/flash_andes_qspi.c
@@ -10,12 +10,12 @@
 #include <zephyr/kernel.h>
 #include <zephyr/device.h>
 #include <zephyr/drivers/flash.h>
+#include <zephyr/drivers/flash/spi_nor.h>
 #include <zephyr/init.h>
 #include <string.h>
 #include <zephyr/logging/log.h>
 
 #include "flash_andes_qspi.h"
-#include "spi_nor.h"
 #include "jesd216.h"
 #include "flash_priv.h"
 
--- a/drivers/flash/flash_andes_qspi_xip.c
+++ b/drivers/flash/flash_andes_qspi_xip.c
@@ -15,13 +15,13 @@
 #include <zephyr/kernel.h>
 #include <zephyr/device.h>
 #include <zephyr/drivers/cache.h>
+#include <zephyr/drivers/flash/spi_nor.h>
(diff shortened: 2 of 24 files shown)


de-nordic (changes requested):
NACK. This header is not Zephyr API header for Flash Drivers. This header is provided for SPI drivers to reduce redefinitions and should remain where it was.

erwango:
> NACK. This header is not Zephyr API header for Flash Drivers. This header is provided for SPI drivers to reduce redefinitions and should remain where it was.

@de-nordic This is coming from https://github.com/zephyrproject-rtos/zephyr/pull/96670 review where we see this header duplicated between SPI drivers and MSPI drivers. If this proposal is not acceptable, do you see an acceptable alternative ?

fabiobaltieri:
Wouldn't this be useful for out of three drivers though?

de-nordic:
> Wouldn't this be useful for out of three drivers though?

It is not a standard. Just look at this table taken from JEDEC251:
image
even READ fast has two opcodes given directly and you have ambiguous names, in spi_nor.h, for them like: SPI_NOR_CMD_READ_FAST and SPI_NOR_CMD_2READ; and the SFDP means that you have to read a code from parameter table. So you do not even know what is the third one till run-time (if you read SFDP at run-time).

To use it you literary have to check dev spec, find what opcode you want to use, find the opcode by code in spi_nor.h and then you can use, for example, SPI_NOR_CMD_2READ, in your code. Somebody comes months later and tires to figure out, they have to go reversal on that anyway, because they have no way to match driver code directly do documentation.
We could just as well have command SPI_NOR_0x03, which is easier to match that what we have now.

At this point, with MSPI, it is not even SPI_NOR codes only but rather xSPI_NOR or just xSPI_SOMETHING, if you account for MRAM, RRAM, FRAM, EEPROM and SRAM SPI devices, as they often share the same base (de-facto standard) set of commands, although have no SFDP.

I have my own out of tree drivers and they follow the dev specs, for the basic operations, and I directly reference command names from spec in comments like
```
/* READ b0000_0011 (0x03); 3.2 Read Sequence */
```
otherwise it is just annoying to remind myself every time what id does what (and I see immediately what to expect on logic probe).

PR #100430, opened 2025-12-03
boards: nordic: Add default Kconfig for (Q)SPI NOR page size

nordicjm:
Changes the default page size to 4096 (4KiB) for SPI/QSPI NOR flash devices if MCUboot is enabled to allow for swap updates to work by default


de-nordic (dismissed):

JordanYates:
> This sets defaults for when MCUboot is enabled in the application and a default for MCUboot itself

This is equivalent (for the nrf52840dk specifically) to setting `CONFIG_NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE=4096` in https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/boards/nrf52840dk_nrf52840.conf correct?

Why does `CONFIG_QSPI_NOR=n` belong in the mcuboot repo, while `CONFIG_NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE=4096` belong in the Zephyr repo?

JordanYates:
> Not exactly, no, this sets the default value, it does not force the value

Great, we agree on that.
So why is https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/boards/nrf52840dk_nrf52840.conf "forcing" `CONFIG_NORDIC_QSPI_NOR=n` as a default? The comment even explicitly says `# Ensure that the qspi driver is disabled by default`.

If you want it disabled by default, shouldn't that be set by a `default X if Y` in a `.defconfig`, as this PR does?
That would let someone change the default if they wished by forking Zephyr.

michalek-no (approved):

nordicjm:
> > Not exactly, no, this sets the default value, it does not force the value
> 
> Great, we agree on that. So why is https://github.com/mcu-tools/mcuboot/blob/main/boot/zephyr/boards/nrf52840dk_nrf52840.conf "forcing" `CONFIG_NORDIC_QSPI_NOR=n` as a default? The comment even explicitly says `# Ensure that the qspi driver is disabled by default`.
> 
> If you want it disabled by default, shouldn't that be set by a `default X if Y` in a `.defconfig`, as this PR does? That would let someone change the default if they wished by forking Zephyr.

They can change that, in multiple ways:

- when building MCUboot directly `cmake -GNinja -DBOARD=<blah> -DCONFIG_NORDIC_QSPI_NOR=y path/to/mcubootboot/zephyr`
- when building an application with sysbuild `cmake -GNinja -DBOARD=<blah> -Dmcuboot_CONFIG_NORDIC_QSPI_NOR=y -DAPP_DIR=<blah> path/to/share/sysbuild`
- like the above but using a Kconfig fragment with `-DEXTRA_CONF_FILE=<blah>` or `-Dmcuboot_EXTRA_CONF_FILE=<blah>`
- in a project with a sysbuild Kconfig fragment for MCUboot by placing it in `sysbuild/mcuboot.conf` or `sysbuild/mcuboot/prj.conf` (latter can then have `sysbuld/mcuboot/boards/<board>/conf` files also)
- by using menuconfig

There's also other ways to change Kconfigs on images from sysbuild e.g. in a `sysbuild.cmake` file you can use `set_config_bool(mcuboot CONFIG_NORDIC_QSPI_NOR y)` likewise in a module, or it can be done using an image configuration script

The defaults here, are just board defaults. The MCUboot board defaults are just MCUboot defaults, it doesn't mean they are the only settings you can have, applications might chose to have completely different settings based on what device is being used or what type of build they are using

nordicjm:
> Maybe I'll come from a different angle and you can tell me which step I'm wrong at:
> 
> For MCUBoot:
> 
>     1. `nRF52840dk` wants `QSPI_NOR` disabled by default
> 
>     2. `nRF52840dk` wants  `LAYOUT_PAGE_SIZE` to be 4096 by default, if `QSPI_NOR` does end up being enabled
> 
>     3. Both 1. and 2. are default values for the MCUboot application
> 
>     4. Both 1. and 2. may want to be changed by the user
> 
>     5. Setting the defaults in a `nrf52840dk_defconfig` is inherently less flexible than setting defaults in `nrf52840dk.defconfig`
> 
>     6. Since both are application defaults for the same application, for the same board, they should be set in the same place

3 is wrong, defconfig is not default, defconfig is what is set, defaults are not what is set, they are default values
5 is wrong because they're for different purposes, they are not the same thing although they can be used interchangeably for that (and in applications, generally they are if e.g. one Kconfig's value depends on another one rather that is absolute)
6 is wrong

See https://docs.zephyrproject.org/latest/build/kconfig/setting.html#motivation-for-kconfig-defconfig-files and the bit below

> The Kconfig.defconfig method should be used when the dependencies of the choice might not be satisfied. In that case, you’re setting the default selection whenever the user makes the choice visible.

which is exactly what I've described here

JordanYates:
>> 3. Both 1. and 2. are default values for the MCUboot application
>
> 3 is wrong, defconfig is not default, defconfig is what is set, defaults are not what is set, they are default values

By "are default values for the MCUboot application" I meant conceptually, not the current implementation.

If 3 is truly false, that would mean that either `QSPI_NOR` should not be disabled by default for MCUboot (in which case the current MCUboot default it incorrect), or `LAYOUT_PAGE_SIZE` should not be 4096 by default for MCUboot (in which case this PR is incorrect).

> 5 is wrong because they're for different purposes

I agree they are for different purposes. I don't think that makes `nrf52840dk.defconfig` less flexible.

> 6 is wrong, See https://docs.zephyrproject.org/latest/build/kconfig/setting.html#motivation-for-kconfig-defconfig-files and the bit below

That is not an argument for point 6 being wrong at the highest level. It doesn't explain why the default configuration for the MCUboot application should be split between the MCUboot and Zephyr repositories.

Don't get me wrong, I am in favor of this PR. I would vastly prefer the bootloader defaults for a board defined in repo X to also be specified in repo X, mostly because that scales for arbitrary downstream repos and forks of Zephyr, while specifying it in MCUboot does not.

tejlmand:
> By "are default values for the MCUboot application" I meant conceptually, not the current implementation.

I do see @JordanYates point of view.
MCUboot gets special treatment which is not the luxury of other samples.

Someone cannot do `default XYZ if BLE_PERIPHERAL_HR_SAMPLE` in a board Kconfig, but they can do it for the MCUboot sample because of this: https://github.com/zephyrproject-rtos/zephyr/blob/d0cda31dc4810336f901700588e183e22cc7bfc2/modules/Kconfig.mcuboot#L8-L11

If other samples suddenly starts to create such `config MY_SAMPLE` and add `default X if MY_SAMPLE` in `<board>/Kconfig.defconfig` then our board specific Kconfig's would turn into a mess in no time.

I agree MCUboot is special in Zephyr but what we do / allow with MCUboot also sets a precedence, so perhaps we should consider introducing a helper symbol which MCUboot and other samples can then select.

We have `IS_BOOTLOADER` already, could that be used instead ?

Or we could have `config LOW_FOOTPRINT` or whatever, but I agree, using `MCUBOOT` for testing is a luxury we don't give `MY_SAMPLE` and others.

JordanYates:
> No they are not the same, QSPI_NOR should be disabled in the normal configuration/build and LAYOUT_PAGE_SIZE should be undefined because the required Kconfigs are not enabled

I thought it was self-explanatory that I meant that `LAYOUT_PAGE_SIZE` would be set to 4096 assuming the dependencies are met. Obviously it wouldn't end up as 4096 if the driver itself was not enabled.

> It's an exclusion to exclude other use cases that might expect a faster throughput

Again, what does that have to do with why one Kconfig option is defaulted in MCUboot, while the other is in Zephyr? They're both for the same application, the same board, the same driver, but defaulted in different repos.

> They do, partitions are sized up for 4KiB sector size, so you might have a partition starting at offset 72KiB that is 16KiB, a 64KiB sector size means erasing from 64KiB - 128KiB, which isn't going to work because it's outside the scope of the partition

There are extra constraints sure, but its not a requirement AFAIK. The bootloader could have a view of `4kB` partitions from 0 to 256kB, while the application sees 64kB partitions all the way through. That would enable both compatibility with MCUboot, while still supporting the "high throughput" use case.

de-nordic (approved):

anangl (approved):

PR #100431, opened 2025-12-03
drivers: i2c: mchp: Add DMA support for I2C g1 driver

fabin-mchp:
This pull request introduces DMA support for I2C transfers in the Microchip I2C g1 driver and updates the device tree nodes for the sam_e54_xpro board to enable DMA functionality for I2C peripherals.


drivers/i2c/i2c_mchp_sercom_g1.c
```diff
@@ -10,6 +10,10 @@
 #include <zephyr/irq.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/drivers/i2c.h>
+#ifdef CONFIG_I2C_MCHP_DMA_DRIVEN
```
teburd: ifdef really needed? Any build issues if removed?

drivers/i2c/i2c_mchp_sercom_g1.c
```diff
@@ -517,11 +542,42 @@ static bool i2c_is_terminate_on_error(const struct device *dev)
...
+
+	return retval;
+}
+#endif /*CONFIG_I2C_MCHP_DMA_DRIVEN*/
+
 static void i2c_restart(const struct device *dev)
 {
 	struct i2c_mchp_dev_data *data = dev->data;
 
-	/* left-shift address by 1 for R/W bit. */
```
teburd, line 524: Dropping the comment here seems unrelated

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.dts
```diff
@@ -190,6 +190,8 @@
 	clock-frequency = <I2C_BITRATE_FAST>;
 	pinctrl-0 = <&sercom7_i2c_default>;
 	pinctrl-names = "default";
+	dmas = <&dmac 10 0x12>, <&dmac 11 0x13>;
+	dma-names = "rx", "tx";
```
nandojve, line 194: Are this values fixed ?
If yes, they must to go in the soc devicetree.
Remember, user should touch only things that are not fixed!

drivers/i2c/Kconfig.mchp
```diff
@@ -19,11 +19,22 @@ if I2C_MCHP_SERCOM
...
+	depends on !I2C_MCHP_DMA_DRIVEN
 	help
 	  This enables INTERRUPT driven transactions for the I2C peripheral.
 	  interrupt driven mode can be used in blocking & non blocking mode.
 
+config I2C_MCHP_DMA_DRIVEN
+	bool "DMA support for Microchip I2C devices"
+	default y
+	depends on I2C_CALLBACK
+	select DMA
```
nandojve:
```diff
	depends on I2C_CALLBACK
-	select DMA
+	depends on DMA
```

drivers/i2c/i2c_mchp_sercom_g1.c
```diff
@@ -1515,6 +1856,17 @@ static DEVICE_API(i2c, i2c_mchp_api) = {
 	.i2c_clock.mclk_sys = (void *)(DT_INST_CLOCKS_CELL_BY_NAME(n, mclk, subsystem)),           \
 	.i2c_clock.gclk_sys = (void *)(DT_INST_CLOCKS_CELL_BY_NAME(n, gclk, subsystem)),
 
+#if CONFIG_I2C_MCHP_DMA_DRIVEN
+#define I2C_MCHP_DMA_CHANNELS(n)                                                                   \
+	.i2c_dma.dma_dev = DEVICE_DT_GET(MCHP_DT_INST_DMA_CTLR(n, tx)),                            \
+	.i2c_dma.tx_dma_request = MCHP_DT_INST_DMA_TRIGSRC(n, tx),                                 \
+	.i2c_dma.tx_dma_channel = MCHP_DT_INST_DMA_CHANNEL(n, tx),                                 \
+	.i2c_dma.rx_dma_request = MCHP_DT_INST_DMA_TRIGSRC(n, rx),                                 \
+	.i2c_dma.rx_dma_channel = MCHP_DT_INST_DMA_CHANNEL(n, rx),
```
nandojve, line 1900: This assumes that DMA definition are always available.
What happen if that information is not ?
How the driver should behave ? It should fall back to interrupt mode ?
I assume that it is a valid scenario using one I2C with DMA and another with Interrupt. Is the driver ready for that situation ?

nandojve (changes requested):
DMA require samples and tests working in mainline.

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.dts
fabin-mchp: The channel values `10` and `11` are not fixed. The trigger sources `0x12` and `0x13` are specific to the `sercom7` node. Each SERCOM node will have its own unique trigger source values for `rx` and `tx`.

Given this, it makes sense to define these values within this file, rather than in the SoC device tree.

drivers/i2c/Kconfig.mchp
fabin-mchp: This was added so that users get maximum performance in their I2C driver by default (i.e., using DMA). If we make it depend on DMA, users will need to set CONFIG_DMA=y in their proj.conf. Is that the recommended approach, or is it acceptable to keep the current method?

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.dts
nandojve: Ok, but you say _own unique trigger source values for rx and tx_ I understand this is fixed, right ?

drivers/i2c/Kconfig.mchp
nandojve: Yes, user should select DMA in the application.

drivers/i2c/i2c_mchp_sercom_g1.c
fabin-mchp: Yes. If the DMA definitions are not available, the driver will fall back to interrupt mode.
Using one I2C instance with DMA and another with interrupts is a valid scenario. However, the current driver does not support this configuration. We can consider adding support for mixed DMA/interrupt configurations as a future feature enhancement.

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.dts
fabin-mchp: `dmas = <&dmac 10 0x12>;`
Here, 10 is the DMA channel number and 0x12 is the trigger source.

Since one of these values is fixed and the other is not, we cannot place this property in the SoC DTSI file where only SoC-level fixed properties should reside.

Additionally, SERCOM supports multiple functionalities (I²C, SPI, UART), and the specific function used is determined at the board level. Because of this, the DMA configuration depends on how the SERCOM instance is used on a particular board.

Therefore, placing the DMA-related properties in the board DTS is the most appropriate and flexible approach.

drivers/i2c/Kconfig.mchp
fabin-mchp: Updated

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.dts
nandojve: Correct

drivers/i2c/i2c_mchp_sercom_g1.c
nandojve: > However, the current driver does not support this configuration. 

Ok, but then user can configure the system wrongly because he need to add DMA lines for all instances with `status=OKAY`, right ? If the driver do not support mixed configuration maybe we need to add some assert to pass the message:

_you are selection DMA to be used in I2C but you do not configure the DMA lines for the instance X. You need configure all I2C DMA lines to enable I2C DMA support._

nandojve (changes requested):

drivers/i2c/i2c_mchp_sercom_g1.c
fabin-mchp: I’ve added a compile-time build assert as suggested. This way, if DMA is enabled for I2C, all instances need to have their DMA lines correctly configured, helping to catch any misconfigurations during the build.

teburd (changes requested):

drivers/i2c/i2c_mchp_sercom_g1.c
fabin-mchp: The build works fine without the #ifdef. I added it just to include these headers only when DMA support is enabled.
Should I remove it?

fabin-mchp: Removed #ifdef as per suggestion.

tests/drivers/i2c/i2c_async/src/main.c
```diff
@@ -0,0 +1,80 @@
...
+		zassert_equal(read_data[i], write_data[i],
+			      "Data mismatch at index %d: expected 0x%02X, got 0x%02X", i,
+			      write_data[i], read_data[i]);
+	}
+}
+
+/* Test Setup */
+void *i2c_test_setup(void)
+{
+
```
nandojve: drop empty line

nandojve (changes requested):

tests/drivers/i2c/i2c_async/src/main.c
fabin-mchp: Removed empty line

nandojve (approved):

ArunMCHP (approved):

teburd (approved):

PR #100433, opened 2025-12-03
drivers: gpio: pcf857x: pcf857x_port_get_raw() output parameter check invalid

### LINKED ISSUE #100365: drivers: gpio: pcf877x: driver checks uninitialized value
### Describe the bug

In function pcf857x_port_get_raw() there is a check on the dereferenced value of function parameter 'gpio_port_value_t *value'.

As this is supposed to be an output parameter of this function, it makes no sense to check its value - the pointer should be valid, yes, but the value will be overwritten some lines later...

In the function description it is clearly stated that 'value' must be a "Pointer to a variable where pin values will be stored" ... so pure output parameter.

In my opinion this whole check shall be removed.

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

MikeDK:
The dereferenced value of function parameter 'value' is checked against output pin configuration, which absolutely makes no sense because this parameter is intended as pure output buffer. Nothing is mentioned in the interface that the buffer should have any specific content on call of this function. Thus the nonsensical check is removed.

Fixes: #100365

### DIFF
--- a/drivers/gpio/gpio_pcf857x.c
+++ b/drivers/gpio/gpio_pcf857x.c
@@ -135,11 +135,6 @@ static int pcf857x_port_get_raw(const struct device *dev, gpio_port_value_t *val
 		return -EWOULDBLOCK;
 	}
 
-	if ((~drv_data->pins_cfg.configured_as_outputs & (uint16_t)*value) != (uint16_t)*value) {
-		LOG_ERR("Pin(s) is/are configured as output which should be input.");
-		return -EOPNOTSUPP;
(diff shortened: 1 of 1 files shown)


bbilas (approved):
Tested on `kincony kc868 a32` board.

henrikbrixandersen (approved):

tpambor (approved):

PR #100434, opened 2025-12-03
tests: scripts: add west debug support in twister

hakehuang:
in some multi-core platform, `west flash` can not support all core. `west debug` is supported to test secondary core, add this to twister such as with
`--west-flash-cmd=debug` in twister command
or with
`--west-flash-cmd: debug` in hwmap.yaml

and same with pytest

twister command line setting will override the one in hwamp

### DIFF
--- a/scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
+++ b/scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
@@ -51,13 +51,15 @@ class HardwareAdapter(DeviceAdapter):
 
         command = [
             self.west,
-            'flash',
+            self.device_config.west_flash_cmd,
             '--no-rebuild',
             '--build-dir',
             str(self.device_config.build_dir),
         ]
 
         command_extra_args = []
+        if self.device_config.west_flash_cmd == 'debug':
+            command_extra_args.append('--batch')
         if self.device_config.west_flash_extra_args:
             command_extra_args.extend(self.device_config.west_flash_extra_args)
 
--- a/scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
+++ b/scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
@@ -139,6 +139,11 @@ def pytest_addoption(parser: pytest.Parser):
         '--extra-test-args',
         help='Additional args passed to the test binary'
     )
+    twister_harness_group.addoption(
+        '--west-flash-cmd',
+        choices=('flash', 'debug'),
+        help='west command to use, can be flash or debug'
+    )
 
 
 def pytest_configure(config: pytest.Config):
--- a/scripts/pylib/twister/twisterlib/environment.py
+++ b/scripts/pylib/twister/twisterlib/environment.py
@@ -272,6 +272,16 @@ Artificially long but functional example:
     )
 
     parser.add_argument(
+        "--west-flash-cmd", choices=['flash', 'debug'],
+        help="""Uses the specified west command. twister will use flash if not set
+
+        E.g "twister --device-testing --device-serial /dev/ttyACM0
+                         --west-flash-cmd="flash"
+        will translate to "west flash ..."
+        """
+    )
+
+    parser.add_argument(
         "-a", "--arch", action="append",
         help="Arch filter for testing. Takes precedence over --platform. "
              "If unspecified, test all arches. Multiple invocations "
--- a/scripts/pylib/twister/twisterlib/handlers.py
+++ b/scripts/pylib/twister/twisterlib/handlers.py
@@ -587,10 +587,14 @@ class DeviceHandler(Handler):
             return self._create_flash_command(hardware)
 
         command = ["west"]
+        command_extra_args = []
         if self.options.verbose > 2:
             command.append(f"-{'v' * (self.options.verbose - 2)}")
-        command += ["flash", "--no-rebuild", "-d", self.build_dir]
-        command_extra_args = []
+
+        west_flash_cmd = self.options.west_flash_cmd or hardware.west_flash_cmd or "flash"
+        if west_flash_cmd == "debug":
+            command_extra_args.append('--batch')
+        command.extend([west_flash_cmd, "--no-rebuild", "-d", self.build_dir])
 
         # There are three ways this option is used.
         # 1) bare: default flags
--- a/scripts/pylib/twister/twisterlib/harness.py
+++ b/scripts/pylib/twister/twisterlib/harness.py
@@ -475,6 +475,9 @@ class Pytest(Harness):
         if runner := hardware.runner or options.west_runner:
             command.append(f'--runner={runner}')
 
+        if west_flash_cmd := options.west_flash_cmd or hardware.west_flash_cmd:
+            command.append(f'--west-flash-cmd={west_flash_cmd}')
+
         if hardware.runner_params:
             for param in hardware.runner_params:
                 command.append(f'--runner-params={param}')
--- a/scripts/pylib/pytest-twister-harness/src/twister_harness/twister_harness_config.py
+++ b/scripts/pylib/pytest-twister-harness/src/twister_harness/twister_harness_config.py
@@ -44,6 +44,7 @@ class DeviceConfig:
     fixtures: list[str] = None
     app_build_dir: Path | None = None
     extra_test_args: str = ''
+    west_flash_cmd: str = 'flash'
 
     def __post_init__(self):
         domains = self.build_dir / 'domains.yaml'
@@ -101,7 +102,8 @@ class TwisterHarnessConfig:
             post_script=_cast_to_path(config.option.post_script),
             post_flash_script=_cast_to_path(config.option.post_flash_script),
             fixtures=config.option.fixtures,
-            extra_test_args=config.option.extra_test_args
+            extra_test_args=config.option.extra_test_args,
+            west_flash_cmd=config.option.west_flash_cmd or 'flash'
         )
 
         devices.append(device_from_cli)
--- a/scripts/pylib/twister/twisterlib/hardwaremap.py
+++ b/scripts/pylib/twister/twisterlib/hardwaremap.py
@@ -58,6 +58,7 @@ class DUT:
     probe_id: str | None = None
     notes: str | None = None
     match: bool = False
+    west_flash_cmd: str | None = None
 
     def __post_init__(self):
         """Initialize non-serializable objects after dataclass initialization."""
@@ -66,8 +67,10 @@ class DUT:
         self._available = Value("i", 1)
         self._failures = Value("i", 0)
         self.lock = Lock()
+
         # Ensure serial_baud has a default value
         self.serial_baud = self.serial_baud or 115200
+
 
     @property
     def available(self):
@@ -288,6 +291,7 @@ class HardwareMap:
             product = dut.get('product')
             fixtures = dut.get('fixtures', [])
             connected = dut.get('connected') and ((serial or serial_pty) is not None)
+            west_flash_cmd = dut.get('west_flash_cmd', "")
             if not connected:
                 continue
             for plat in platforms:
@@ -306,7 +310,8 @@ class HardwareMap:
                               post_flash_script=post_flash_script,
                               script_param=script_param,
                               flash_timeout=flash_timeout,
-                              flash_with_test=flash_with_test)
+                              flash_with_test=flash_with_test,
+                              west_flash_cmd=west_flash_cmd)
                 new_dut.fixtures = fixtures
                 new_dut.counter = 0
                 self.duts.append(new_dut)
--- a/scripts/schemas/twister/hwmap-schema.yaml
+++ b/scripts/schemas/twister/hwmap-schema.yaml
@@ -56,6 +56,8 @@ items:
         post_flash_timeout:
           type: integer
       additionalProperties: false
+    west_flash_cmd:
+      type: string
   required:
     - connected
     - id
--- a/scripts/tests/twister/test_handlers.py
+++ b/scripts/tests/twister/test_handlers.py
@@ -1185,6 +1185,7 @@ def test_devicehandler_create_command(
 ):
     handler = DeviceHandler(mocked_instance, 'build', mock.Mock())
     handler.options = mock.Mock(west_flash=self_west_flash,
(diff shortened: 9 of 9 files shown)


hakehuang:
"No space left on device" CI failure

scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -589,7 +589,11 @@ def _create_command(self, runner, hardware):
         command = ["west"]
         if self.options.verbose > 2:
             command.append(f"-{'v' * (self.options.verbose - 2)}")
-        command += ["flash", "--no-rebuild", "-d", self.build_dir]
+        os.environ["ZEPHYR_TWISTER_RUN_IN_DEBUG"] = 'y'
```
gchwier: 1. The environment variable is only set here in Twister; it is not set in pytest-harness.
2. Is it safe to set the environment variable here? What happens if multiple tests are executed by Twister? In that case, you would force the `west` command to include a special option even if it was not explicitly required by the test or board.

scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
```diff
@@ -139,6 +139,10 @@ def pytest_addoption(parser: pytest.Parser):
         '--extra-test-args',
         help='Additional args passed to the test binary'
     )
+    twister_harness_group.addoption(
+        '--west-cmd',
+        help='west command flash(default) | debug'
+    )
```
gchwier, line 146: default is not set, and will be passed as None to DeviceConfig

gchwier (commented):
It might be better to move the changes related to `west` into a separate PR.

scripts/pylib/twister/twisterlib/handlers.py
hakehuang: the environment variable is used in `west flash`, as this is in process context, so the variable only exist in this test process, and its child process.

scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
hakehuang: I will fix it

hakehuang:
need wait https://github.com/zephyrproject-rtos/zephyr/pull/100924 to merge

scripts/pylib/twister/twisterlib/handlers.py
gchwier: When you run Twister with two boards connected, board A uses west debug, board B does not use it. When you set this ENV for board A, then board B also gets that ENV (set in the same Twister process so west process gets it). Better would be to pass extra option via specific parameter to west / Jlink (as proposed in second PR)

scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -587,10 +587,20 @@ def _create_command(self, runner, hardware):
...
+        if hardware.west_cmd:
+            # set environ in this process and notify west to use debug batch mode
+            if hardware.west_cmd == "debug":
+                command_extra_args.append('--batch')
+            command += [hardware.west_cmd, "--no-rebuild", "-d", self.build_dir]
+        else:
+            # set environ in this process and notify west to use debug batch mode
+            if self.options.west_cmd == "debug":
+                command_extra_args.append('--batch')
+            command += [self.options.west_cmd, "--no-rebuild", "-d", self.build_dir]
```
pdgendt: De-duplicate some things

```python
        command += [hardware.west_cmd if hardware.west_cmd or self.options.west_cmd]
        # set environ in this process and notify west to use debug batch mode
        if command[-1] == "debug":
            command_extra_args.append('--batch')
        command += ["--no-rebuild", "-d", self.build_dir]
```

scripts/pylib/twister/twisterlib/handlers.py
hakehuang: @gchwier now it changes to use west parameter. Please take a look

scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -587,10 +587,15 @@ def _create_command(self, runner, hardware):
...
         if self.options.verbose > 2:
             command.append(f"-{'v' * (self.options.verbose - 2)}")
-        command += ["flash", "--no-rebuild", "-d", self.build_dir]
-        command_extra_args = []
+
+        command += [hardware.west_cmd if hardware.west_cmd else self.options.west_cmd]
+        # notify west to use debug batch mode
+        if command[-1] == "debug":
+            command_extra_args.append('--batch')
+        command += ["--no-rebuild", "-d", self.build_dir]
```
pdgendt: Some minor inconsistencies now

```suggestion
        command.append(hardware.west_cmd if hardware.west_cmd else self.options.west_cmd)
        # notify west to use debug batch mode
        if command[-1] == "debug":
            command_extra_args.append('--batch')
        command.extend(["--no-rebuild", "-d", self.build_dir])
```

pdgendt (dismissed):

scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -587,10 +587,15 @@ def _create_command(self, runner, hardware):
...
         if self.options.verbose > 2:
             command.append(f"-{'v' * (self.options.verbose - 2)}")
-        command += ["flash", "--no-rebuild", "-d", self.build_dir]
-        command_extra_args = []
+
+        command.append(hardware.west_cmd if hardware.west_cmd else self.options.west_cmd)
+        # notify west to use debug batch mode
+        if command[-1] == "debug":
+            command_extra_args.append('--batch')
+        command.append("--no-rebuild", "-d", self.build_dir)
```
pdgendt: This does not work

```shell
TypeError: list.append() takes exactly one argument (3 given)
```

```suggestion
        command.extend(["--no-rebuild", "-d", self.build_dir])
```

hakehuang: good catch

pdgendt (dismissed):

scripts/pylib/twister/twisterlib/handlers.py
```diff
@@ -587,10 +587,15 @@ def _create_command(self, runner, hardware):
...
         if self.options.verbose > 2:
             command.append(f"-{'v' * (self.options.verbose - 2)}")
-        command += ["flash", "--no-rebuild", "-d", self.build_dir]
-        command_extra_args = []
+
+        command.append(hardware.west_cmd if hardware.west_cmd else self.options.west_cmd)
+        # notify west to use debug batch mode
+        if command[-1] == "debug":
+            command_extra_args.append('--batch')
+        command.extend(["--no-rebuild", "-d", self.build_dir])
```
gchwier:
```python
        west_cmd = hardware.west_cmd or self.options.west_cmd or "flash"
        if west_cmd == "debug":
            command_extra_args.append('--batch')
        command.extend([west_cmd, "--no-rebuild", "-d", self.build_dir])
```
I thought that command-line parameters should take precedence over values from files or code ... However, I found that `--west-runner`  prioritizes the hardware map first also. Is this behavior intentional? or should we change it?

gchwier (commented):
`--west-cmd` is not passed to Pytest, 
check method `_generate_parameters_for_hardware` in harness.py

scripts/pylib/twister/twisterlib/environment.py
```diff
@@ -270,6 +270,17 @@ def add_parse_arguments(parser = None) -> argparse.ArgumentParser:
         """
     )
 
+    parser.add_argument(
+        "--west-cmd", choices=['flash', 'debug'],
+        default='flash',
```
gchwier: you can remove the default here,
in handlers.py:594:
`west_cmd = self.options.west_cmd or hardware.west_cmd or "flash"`
in harness.py:483:
```
        if west_cmd := hardware.west_cmd or options.west_cmd:
            command.append(f'--west-cmd={west_cmd}')
```

scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
gchwier, line 60: add here:
```
        if self.device_config.west_cmd == 'debug':
            command_extra_args.append('--batch')
```

tested with command:
`./scripts/twister -vv -ll debug -T samples/subsys/testsuite/pytest/shell -s sample.pytest.shell --device-testing --hardware-map ~/tmp/hardware_map.yml --west-flash --no-clean --west-cmd debug`

scripts/pylib/twister/twisterlib/handlers.py
hakehuang: you are right. I set this because I used to set a default value in options parser. and your recommendation above is actually set default to "flash", right? this is kind of ambiguous to user. let me update '--west-cmd' as you recommend and leave `west-runner` as is.

scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
```diff
@@ -139,6 +139,10 @@ def pytest_addoption(parser: pytest.Parser):
         '--extra-test-args',
         help='Additional args passed to the test binary'
     )
+    twister_harness_group.addoption(
+        '--west-cmd',
```
gchwier: for pytest-harness the default value of west-cmd need to be set,
or in twister_harness_config.py should be set:
`west_cmd=config.option.west_cmd  or "flash"`,

you can check with:
`./scripts/twister -vv -ll debug -T samples/subsys/testsuite/pytest/shell -s sample.pytest.shell --device-testing --hardware-map ~/tmp/hardware_map.yml --west-flash --no-clean`  (--west-cmd is not given)

scripts/pylib/twister/twisterlib/environment.py
```diff
@@ -270,6 +270,17 @@ def add_parse_arguments(parser = None) -> argparse.ArgumentParser:
         """
     )
 
+    parser.add_argument(
+        "--west-cmd", choices=['flash', 'debug'],
+        default=None,
```
gchwier: this line can be removed

hakehuang: ok

gchwier (dismissed):

scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
```diff
@@ -139,6 +139,10 @@ def pytest_addoption(parser: pytest.Parser):
         '--extra-test-args',
         help='Additional args passed to the test binary'
     )
+    twister_harness_group.addoption(
```
pdgendt, line 142: Add explicit `choices=('flash', 'debug')` instead of only describing them?

pdgendt (commented):
> @pdgendt I update due to some missing in the pytest harness support, and refine the code as @gchwier recommends, Please take a look. now all test with twister/west/pytest have pass. Thanks in advance.

It does need assignee approval anyway..

gchwier (dismissed):

pdgendt (dismissed):

hakehuang:
@nashif please

scripts/pylib/twister/twisterlib/harness.py
```diff
@@ -475,6 +475,9 @@ def _generate_parameters_for_hardware(self, handler: Handler):
         if runner := hardware.runner or options.west_runner:
             command.append(f'--runner={runner}')
 
+        if west_cmd := hardware.west_cmd or options.west_cmd:
```
gchwier:
```suggestion
        if west_cmd := options.west_cmd or hardware.west_cmd:
```
command-line parameters should take precedence over values from files, please change the order also here

pdgendt (dismissed):

gchwier (dismissed):

fundakol (dismissed):

scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
```diff
@@ -51,13 +51,15 @@ def generate_command(self) -> None:
 
         command = [
             self.west,
-            'flash',
+            self.device_config.west_cmd,
```
nashif: west_cmd is ambigous and can mean many anything, here the context is the flashing command, so we should not be using west_cmd and instead be specific, i.e. west_flash_cmd. So this can be flash or debug which will do something similar to "flash"

nashif (changes requested):

scripts/pylib/pytest-twister-harness/src/twister_harness/device/hardware_adapter.py
hakehuang: @nashif I change to `west_flash_cmd` as you recommend, and the copilot issue is also fixed according to former review, the command line option shall override the hwmap settings. Thanks

scripts/pylib/pytest-twister-harness/src/twister_harness/plugin.py
```diff
@@ -139,6 +139,11 @@ def pytest_addoption(parser: pytest.Parser):
...
+        '--west-flash-cmd',
+        choices=('flash', 'debug'),
+        help='west flash command flash | debug'
```
pdgendt: The help here isn't really helpful?

hakehuang: @pdgendt is it OK, I change to 
```
west command to use, can be flash or debug
```

scripts/pylib/twister/twisterlib/harness.py
gchwier:
```suggestion
            command.append(f'--west-flash-cmd={west_flash_cmd}')
```
you forgot to change the command here, When you try to run with pytest it fails

gchwier:
```suggestion
        if west_flash_cmd := options.west_flash_cmd or hardware.west_flash_cmd:
```
don't need to set default here and extend command, as it is set in pytest-harness

scripts/pylib/twister/twisterlib/handlers.py
gchwier, line 586: I found that we also have a `--flash-command` option, which probably no one uses. It is starting to be really confusing. It would be better to refactor / clean it up and have only one option for custom usage - but I know it needs more work, and there is a risk of introducing regressions and breaking backward compatibility.

gchwier (changes requested):

gchwier (approved):

nashif (approved):

PR #100435, opened 2025-12-03
MAINTAINERS.yml: drivers: ethernet: add usb ethernet drivers

maass-hamburg:
add usb ethernet/networking drivers to
ethernet drivers area, as they use the
ethernet api. Generally I would like them
to move, buts that's for later.


jfischer-no (changes requested):
I do not think it makes sense to move them, as all the emulated drivers are inside the USB folder, nor to add you as a maintainer to these files.

PR #100439, opened 2025-12-03
drivers: mspi_dw: Fix NULL pointer exception when XIP enabled on reset

ltrager:
While testing a device which has XIP enabled on reset I ran into a NULL pointer exception. This was due to an TXEIS and RXUIS being raised on initialization. Since data has not been requested dev_data->xfer.packets has not been allocated causing an exception.

dev_init() has been hardened a bit ensuring the controller is disabled and IRQs have been masked before any other initialization occurs. IRQs are connected last.

### DIFF
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -1866,10 +1866,11 @@ static int dev_init(const struct device *dev)
 
 	vendor_specific_init(dev);
 
+	/* Make sure controller is disabled. */
+	write_ssienr(dev, 0);
+
 	dev_data->ctrlr0 |= FIELD_PREP(CTRLR0_SSI_IS_MST_BIT,
 				       dev_config->op_mode == MSPI_OP_MODE_CONTROLLER);
-
-	dev_config->irq_config();
 
 #if defined(CONFIG_MULTITHREADING)
 	dev_data->dev = dev;
@@ -1916,8 +1917,7 @@ static int dev_init(const struct device *dev)
 		return rc;
 	}
 
-	/* Make sure controller is disabled. */
-	write_ssienr(dev, 0);
+	dev_config->irq_config();
 
 	return 0;
 }
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -451,10 +451,17 @@ static void handle_end_of_packet(struct mspi_dw_data *dev_data)
 static void handle_fifos(const struct device *dev)
 {
 	struct mspi_dw_data *dev_data = dev->data;
-	const struct mspi_xfer_packet *packet =
-		&dev_data->xfer.packets[dev_data->packets_done];
+	const struct mspi_xfer_packet *packet;
 	bool finished = false;
 
+	/* On initialization TXEIS/RXUIS may be raised before any data is
+	 * requested. If packet is NULL we have nothing to process.
+	 */
+	if (!dev_data->xfer.packets) {
+		return;
+	}
+
+	packet = &dev_data->xfer.packets[dev_data->packets_done];
 	if (packet->dir == MSPI_TX) {
 		if (dev_data->buf_pos < dev_data->buf_end) {
 			tx_data(dev);


swift-tk:
ping @anangl

drivers/mspi/mspi_dw.c
```diff
@@ -450,10 +450,17 @@ static void handle_end_of_packet(struct mspi_dw_data *dev_data)
...
 	struct mspi_dw_data *dev_data = dev->data;
-	const struct mspi_xfer_packet *packet =
-		&dev_data->xfer.packets[dev_data->packets_done];
+	const struct mspi_xfer_packet *packet;
 	bool finished = false;
 
+	/* On initialization TXEIS/RXUIS may be raised before any data is
+	 * requested. If packet is NULL we have nothing to process.
+	 */
+	if (!dev_data->xfer.packets) {
```
anangl, line 460: This is masking the symptoms, not solving the root cause. It looks like the initialization routine should be corrected, as the interrupt is not supposed to occur at that point.

> If XIP is enabled on reset when initialization TXEIS/RXUIS may be raised before any data is requested.

You mean XIP mode in SSI? I don't quite understand what it has to do with interrupts.

drivers/mspi/mspi_dw.c
```diff
@@ -1834,17 +1834,22 @@ static int dev_init(const struct device *dev)
 	struct mspi_dw_data *dev_data = dev->data;
 	const struct mspi_dw_config *dev_config = dev->config;
 	const struct gpio_dt_spec *ce_gpio;
+	unsigned int key = irq_lock();
 	int rc;
 
+	/* Make sure controller is disabled. */
+	write_ssienr(dev, 0);
```
anangl, line 1870: This should not be done before `vendor_specific_init(dev);` as the SSI core registers may be inaccessible before some specific initialization is done.

drivers/mspi/mspi_dw.c
```diff
@@ -1834,17 +1834,22 @@ static int dev_init(const struct device *dev)
...
+
+	/* Make sure IRQs are disabled. */
+	write_imr(dev, 0);
```
anangl: This is redundant. Since the controller is disabled, it won't generate any interrupts.

drivers/mspi/mspi_dw.c
```diff
@@ -1834,17 +1834,22 @@ static int dev_init(const struct device *dev)
 	struct mspi_dw_data *dev_data = dev->data;
 	const struct mspi_dw_config *dev_config = dev->config;
 	const struct gpio_dt_spec *ce_gpio;
+	unsigned int key = irq_lock();
```
anangl: Is it really needed? Since `dev_config->irq_config();` is moved to the end of the initialization, what's the point in additionally locking interrupts?

anangl (changes requested):

drivers/mspi/mspi_dw.c
ltrager: When the device I am working on starts the CRM loads the CMRT off QSPI flash which in turn loads the CCP off QSPI flash. The CRM does not have QSPI support, instead QSPI is configured to enable XIP by default on reset. This allows the CRM to load the CMRT via XIP using a memcpy. The CMRT does the same to load CCP firmware which is running Zephyr.

CCP firmware can use XIP reads as well but writes require us to use this driver. I use `zephyr,deferred-init;` to delay loading the driver until we need to write to avoid contention. Typically this is for a firmware update. As soon as I enable the controller I receive a TXEIS and RXUIS interrupt. This happens even if I explicitly set IMR to 0 before enabling the controller and setting up IRQs in dev_init(). When I spoke to the hardware engineer team I was told this is expected and I should ignore the interrupt.

drivers/mspi/mspi_dw.c
ltrager: Done

drivers/mspi/mspi_dw.c
ltrager: Due to XIP being enabled on reset the hardware team also sets IMR to 0xFF on reset. The consensus on my team is its safest to set IMR to 0 on driver init to ensure no IRQ is raised which the driver doesn't expect. To be clear this does not resolve the issue above, it's just an additional safety measure.

drivers/mspi/mspi_dw.c
ltrager: It was an attempt to fix the IRQs from being raised and having to check if dev_data->xfer.packets is NULL. You're right this isn't needed.

ltrager:
@anangl thanks for the review. Sorry it took so long for the response I lost access to my emulation environment for a bit.

drivers/mspi/mspi_dw.c
ltrager: I just double checked this and without this code I get the following even with `write_imr(dev, 0);` below.

```
mcause:  5 (Load access fault)
mepc:    a0018140
mtval:   0
mstatus: a00001800
mdcause: 3 (Bus error)

*** HALTED ***
```

anangl: > As soon as I enable the controller I receive a TXEIS and RXUIS interrupt. This happens even if I explicitly set IMR to 0 before enabling the controller and setting up IRQs in dev_init(). When I spoke to the hardware engineer team I was told this is expected and I should ignore the interrupt.

You meant "as I enable the interrupt", not controller, right? As I understand, that interrupt occurs when `dev_init()` is called and the controller is still disabled?
Anyway, it is a spurious interrupt, so the interrupt handler should not be executed for it. Isn't it possible to clear the interrupt in some other way than by executing the handler?

drivers/mspi/mspi_dw.c
anangl: The IMR is cleared in `start_next_packet()`, before the controller is enabled: https://github.com/zephyrproject-rtos/zephyr/blob/799ab6e4fd486867625662e200c0415487730dbe/drivers/mspi/mspi_dw.c#L1353-L1356 so I don't see a value in doing it also in `dev_init()`. And as you mentioned, this does not prevent that spurious interrupt from occurring.

drivers/mspi/mspi_dw.c
ltrager: The issue we have with our device is the one DW MSPI controller is connected to three separate cores. So one core could do *something* which triggers an interrupt in another. Even if I clear the spurious interrupt on init it can easily happen again which will result in an exception when `dev_data->xfer.packets = NULL`. Thus we need the code here which checks if data is expected, if it isn't just ignore it.

ltrager:
Thanks for the review. Dropped the unneeded `write_imr(dev, 0)` and retested to verify.

drivers/mspi/mspi_dw.c
anangl: I disagree. You rather need to make sure that only one of those cores uses the controller at a given time, so that once a core initializes the controller for its use, it won't encounter any spurious interrupt or other strange behavior of the controller because another core just did _something_. Checking if this pointer is NULL only hides the actual problem.

ltrager: I am working on control firmware for a NIC. During the boot process multiple cores may end up reading from QSPI at once. This primarily happens when the UEFI driver is read via PCIE but may also happen as CRMT and control cores load and need to read some metadata off the flash during boot. The silicon design team enabled XIP reads by default as they have assured me that multiple cores can read via XIP without any contention. XIP reads occur without using the mspi_dw driver at all, in fact I defer initialization. Writes cannot use XIP due to flash limitations but occur rarely. Typically writes are needed for firmware updates and core dumps.

Reads and writes occur via an intermediary driver I wrote. This driver acquires a hardware semaphore to manage contention so only one core should be able to read or write at a time. Reads access flash via XIP but writes initialize the QSPI and flash. This mostly ensures only one core uses flash at a time but comes with a couple of caveats

1. Boot firmware does not have access to the semaphore however it should not read after CMRT is loaded
2. The mechanism which loads the UEFI driver uses a memory map and has no way to utilize the semaphore. Technically this can happen at any time but realistically it only occurs on host boot while firmware updates can only be triggered by the host itself.
3. Even if the semaphore works perfectly Zephyr doesn't currently have the ability to reinitilize a driver. So another core could leave an interrupt unhandled either due to a bug or a timing issue.

So I am trying to ensure only one core is uses flash at a time but there are edge cases which may occur where multiple cores do try to access QSPI at the same time. The thought is this patch ensures if one of these edge cases occur Zephyr does not crash.

Would it help to add a LOG_ERR() message here?

PR #100440, opened 2025-12-03
NXP: add CACHE64 devicetree binding, init driver, and enable on RW6xx/RT5xx

Holt-Sun:
Add a devicetree binding and an init-only memc driver for the NXP
CACHE64 controller used with FlexSPI/XSPI external memory. The driver
programs CACHE64 POLSEL via the MCUX SDK, optionally enables the write
buffer and cache at boot, and is used to configure external flash/PSRAM
policies on RW6xx and RT5xx boards.


Holt-Sun:
@Albort12138 @SuperHeroAbner , please review.

boards/nxp/frdm_rw612/frdm_rw612_common.dtsi
```diff
@@ -353,3 +353,11 @@ zephyr_mipi_dbi_spi: &lcdic {
...
 	status = "okay";
 };
+
+&cache64_ctrl0 {
+	status = "okay";
+};
+
+&cache64_ctrl1 {
+	status = "okay";
+};
```
Albort12138, line 371: https://github.com/zephyrproject-rtos/zephyr/blob/d0cda31dc4810336f901700588e183e22cc7bfc2/soc/nxp/imxrt/imxrt5xx/cm33/soc.c#L357
These code can be removed if we enable cache64 device driver

Albort12138 (changes requested):

dts/arm/nxp/nxp_rt5xx_common.dtsi
```diff
@@ -126,6 +126,26 @@
...
+		/* 0x0000_0000-0x03FF_FFFF: WB (16MB NOR XIP) */
+		regions = <0x4000000 1>;
+		enable-write-buffer;
+	};
+
+	cache64_ctrl1: cache@34000 {
+		compatible = "nxp,cache64";
+		reg = <0x34000 0x1000>;
+		/* 64MB at 0x38000000 */
+		windows = <0x38000000 0x08000000 0>;
```
Albort12138: In https://github.com/zephyrproject-rtos/zephyr/pull/100440/files#diff-6c4dc228deeb08ee6df91add00515bae13ecded47da4ed988e1529c288df54d7
I do not find usage of *windows*

Holt-Sun: You are right, `windows` is not used. I kept it for conceptual completeness, as it will likely be useful in future cache model extensions.

dts/bindings/cache/nxp,cache64.yaml
```diff
@@ -0,0 +1,62 @@
...
+  regions:
+    type: array
+    required: true
+    description: >
+      Array of region configurations. Each region is encoded as:
+      <boundary_addr policy>
+      - boundary_addr: Top address in bytes (will be converted to device units)
+      - policy: 0=Non-cacheable, 1=Write-through, 2=Write-back
+
+      Maximum 8 regions (CACHE64 hardware limit).
```
zejiang0jason: the region count here is 3, https://github.com/nxp-mcuxpresso/mcuxsdk-core/blob/main/drivers/cache/cache64/fsl_cache.h#L33. are they the same thing?

dts/bindings/cache/nxp,cache64.yaml
```diff
@@ -0,0 +1,62 @@
...
+      Boundaries are specified in the FlexSPI physical address domain.
+      The number of regions is determined by the array length (N boundaries = N+1 regions).
+
+      Example: regions = <0x01000000 2>, <0x08000000 1>;
+      This defines 3 regions:
+        Region 0: 0x00000000-0x00FFFFFF -> Write-back (16MB)
+        Region 1: 0x01000000-0x07FFFFFF -> Write-through (112MB)
+        Region 2: 0x08000000-end        -> Non-cacheable
```
zejiang0jason: does it mean that only the region 0 and region 1 policy can be configured? from reference manual, all the three regions can be configured.
image

dts/arm/nxp/nxp_rt5xx_common.dtsi
```diff
@@ -126,6 +126,26 @@
 		reg = <0x13c000 0x1000>, <0x38000000 DT_SIZE_M(128)>;
 	};
 
+	cache64_ctrl0: cache@33000 {
+		compatible = "nxp,cache64";
+		reg = <0x33000 0x1000>;
+		/* 128MB at 0x18000000, no remap */
+		windows = <0x18000000 0x08000000 0>;
+		/* 0x0000_0000-0x03FF_FFFF: WB (16MB NOR XIP) */
+		regions = <0x4000000 1>;
```
zejiang0jason: seems the region should be defined in board level, according to board real memory?

dts/arm/nxp/nxp_rw6xx_common.dtsi
```diff
@@ -151,6 +151,29 @@
...
 	};
 
+	cache64_ctrl0: cache@33000 {
+		compatible = "nxp,cache64";
+		reg = <0x33000 0x1000>;
+		windows = <0x18000000 0x08000000 0>;
+		/* 0x0000_0000-0x03FF_FFFF: WB (16MB NOR XIP) */
+		/* 0x0400_0000-0x047F_FFFF: WT (8MB PSRAM) */
+		regions = <0x4000000 2>,
+			  <0x4800000 1>;
```
zejiang0jason: If one FlexSPI is connected to flash, the other FlexSPI connected to PSRAM, should both memory regions be configured in every cache64_ctrl?

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,139 @@
...
+		.regions_raw = cache64_regions_raw_##inst,                  \
+		.enable_write_buffer =                                      \
+			DT_INST_PROP_OR(inst, enable_write_buffer, false),      \
+	};                                                              \
+	DEVICE_DT_INST_DEFINE(inst,                                     \
+				cache64polsel_init,                                 \
+				NULL,                                               \
+				NULL,                                               \
+				&cache64_config_##inst,                             \
+				PRE_KERNEL_1,                                       \
```
zejiang0jason, line 187: should be in early stage?

dts/arm/nxp/nxp_rt5xx_common.dtsi
zejiang0jason: how to set it when porting to a new SOC?

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,139 @@
...
+			LOG_WRN("Region %u boundary 0x%x not 1KB aligned; aligning down", i,
+				boundary);
+			boundary &= ~(CACHE64_REGION_ALIGNMENT - 1U);
+		}
+		hw_cfg.boundaryAddr[i] = boundary;
+		hw_cfg.policy[i] = (cache64_policy_t)policy;
+		raw_len += CACHE64_REGION_TUPLE_SIZE;
+	}
+
+	/* Set last region policy if provided; else default NC */
```
zejiang0jason: Explain in dts binding, tell user how to configure the policy for the last region.

zejiang0jason (changes requested):

dts/bindings/cache/nxp,cache64.yaml
Holt-Sun: Yes, they are the same. The region and its cache policy should be configurable by users from device tree data.

dts/bindings/cache/nxp,cache64.yaml
Holt-Sun: No, it doesn't mean. The example in the device tree shows that we can configure a region to Write-back, Write-through, and Non-cacheable. The 3rd part is Non-cacheable is most cases but it doesn't mean it s not configured. 

Update the doc to illustrate it.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: It makes sense to init it in early stage: CACHE64 policies and enabling the cache must precede any FlexSPI/XIP access and drivers that may touch external NOR/PSRAM

dts/arm/nxp/nxp_rt5xx_common.dtsi
Holt-Sun: Move the board dts as you suggest. Board owner shall update the windows based on the board cache64 memory range.

dts/arm/nxp/nxp_rw6xx_common.dtsi
Holt-Sun: Yes, the cache64 memory regions and flexspi memory regions are not always 1:1. cache64 memory range shall be set based on real case per instance.

drivers/memc/memc_nxp_cache64.c
zejiang0jason: then should use `EARLY`?

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,164 @@
...
+	uint8_t last_idx =
+		(uint8_t)((raw_len / CACHE64_REGION_TUPLE_SIZE) < (CACHE64_REGION_NUM - 1U)
+				  ? (raw_len / CACHE64_REGION_TUPLE_SIZE)
+				  : (CACHE64_REGION_NUM - 1U));
+	if (last_idx < CACHE64_REGION_NUM) {
```
zejiang0jason: from the calculation of `last_idx`, it is always `<= (CACHE64_REGION_NUM - 1U)` which means this line is always true.

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,164 @@
...
+		uint32_t last_pol = cfg->regions_raw[last_idx * CACHE64_REGION_TUPLE_SIZE + 1];
```
zejiang0jason: for this case:
```
      Example 2: Two boundaries (3 regions), last region uses default policy
        regions = <0x01000000 2>, <0x08000000 1>;
        - Region 0: 0x00000000-0x00FFFFFF -> Write-back (16MB)
        - Region 1: 0x01000000-0x07FFFFFF -> Write-through (112MB)
        - Region 2: 0x08000000-end        -> Non-cacheable (default)
```
Seems the size of `cache64_regions_raw_` is 4 words, this is out-of-boundary accessing.

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,164 @@
...
+		return ret;
+	}
+
+	/* Enable write buffer if configured */
+#if !(defined(FSL_FEATURE_CACHE64_CTRL_HAS_NO_WRITE_BUF) && \
+		FSL_FEATURE_CACHE64_CTRL_HAS_NO_WRITE_BUF)
+	if (cfg->enable_write_buffer) {
+		CACHE64_EnableWriteBuffer(cfg->ctrl, true);
+		LOG_DBG("Write buffer enabled");
+	}
```
zejiang0jason, line 157: is it possible to throw build error, when `enable-write-buffer;` is enabled for SOC which does support such feature?

drivers/memc/memc_nxp_cache64.c
```diff
@@ -0,0 +1,164 @@
...
+		}
+		return 0;
+	}
+
+	/* regions_raw is tuples of <boundary, policy>; count tuples */
+	size_t raw_len = 0U;
+
+	for (uint8_t i = 0U; i < (CACHE64_REGION_NUM - 1U); i++) {
+		uint32_t boundary = cfg->regions_raw[i * CACHE64_REGION_TUPLE_SIZE + 0];
+		uint32_t policy = cfg->regions_raw[i * CACHE64_REGION_TUPLE_SIZE + 1];
```
zejiang0jason: is `cfg->regions_raw`'s size always larger than `(CACHE64_REGION_NUM - 1U) * CACHE64_REGION_TUPLE_SIZE * sizeof(uint32_t)`? This loop is using `(CACHE64_REGION_NUM - 1U)` as boundary.

boards/nxp/rd_rw612_bga/rd_rw612_bga.dtsi
```diff
@@ -336,3 +336,22 @@ nxp_8080_touch_panel_i2c: &arduino_i2c {
 	status = "okay";
 	wakeup-source;
 };
+
+&cache64_ctrl0 {
+	windows = <0x18000000 0x08000000 0>;
+	/* 0x0000_0000-0x03FF_FFFF: WB (16MB NOR XIP) */
+	/* 0x0400_0000-0x047F_FFFF: WT (8MB PSRAM) */
+	regions = <0x4000000 2>,
+		  <0x4800000 1>;
```
zejiang0jason, line 344: just confirm, one cache64 instance need to manage both flash and PSRAM?

zejiang0jason (changes requested):

dts/bindings/cache/nxp,cache64.yaml
```diff
@@ -0,0 +1,77 @@
...
+        Single window, no remap:
+          windows = <0x18000000 0x08000000 0>;
+
+        Two windows, second one remapped:
+          windows = <0x18000000 0x08000000 0>,
+                    <0x38000000 0x04000000 0x10000000>;
+
+        Three windows with continuous remap:
+          windows = <0x60000000 0x02000000 0x00000000>,
+                    <0x70000000 0x02000000 0x02000000>,
```
Albort12138, line 37: What the meaning of "remap_base"?

dts/arm/nxp/nxp_rt5xx_common.dtsi
Albort12138: How to use this in future feature? The window related information already exists in device's header file

drivers/memc/memc_nxp_cache64.c
Holt-Sun: There is already a feature guard FSL_FEATURE_CACHE64_CTRL_HAS_NO_WRITE_BUF used to protect the usage of it.

boards/nxp/rd_rw612_bga/rd_rw612_bga.dtsi
Holt-Sun: In some cases, yes, one cache64 could be use for both flash and PSRAM. It is more responsible for memory range instead of memory type.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: Reported issue: since Inside FLEXSPI_Init function, a call to RESET_PeirpheralReset, this will set the PRSTCTL0 bit for FlexSPI1 and resets the FlexSPI module and the cache controller overwriting the previous Cache configuration and effectively disabling it. 

So cache init shall be put later than flexspi flash, updated the code.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: Code updated.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: Code updated.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: Code updated.

Holt-Sun:
> The existing implementations under memc (Memory Controller) are all for controlling RAM

The memc folder contains drivers for external memory controllers including both RAM and Flash (e.g., memc_mcux_flexspi_s27ks0641.c for NOR Flash). CACHE64 is specifically designed for FlexSPI/XSPI memory windows, so placing it alongside the FlexSPI/XSPI drivers in memc maintains better architectural cohesion. The cache folder is more appropriate for system-level caches (L1/L2 data/instruction caches) which shall be enabled with "CACHE_MANAGEMENT".

dts/bindings/cache/nxp,cache64.yaml
Holt-Sun: In some platforms, there are multiple memory(flexspi) regions in the system memory map, but these are remapped to create a single contiguous region, where accesses to some system regions are aliased into a single region for the cache.

Albort12138: This situation already covered in cache64 driver, why still need to exposed in devicetree?

drivers/memc/memc_nxp_cache64.c
Albort12138: In Zephyr, I think this issue do not exist. The reported issue is caused by invoking reset API in flexspi driver. But in Zephyr, reset control is disabled in flexspi driver. 
And the init level of flexspi is **POST_KERNEL** already later than cache's init level

dts/arm/nxp/nxp_rt5xx_common.dtsi
Holt-Sun: The windows related information could be used to provide memory address range for the cache so that the cache can be added into whole cache system management.

drivers/memc/memc_nxp_cache64.c
Holt-Sun: Use EARLY.

dts/bindings/cache/nxp,cache64.yaml
Holt-Sun: I think it might be useful for later update.
I will keep it in device tree binding, but remove from board dts data.

PR #100441, opened 2025-12-03
drivers: udc_mcux_ip3511: mark ep0 as not busy after setup.

te-johan:
if a setup token is received before the previous out data is handled the endpoint needs to be marked as not busy. this is done similar in dwc2 and nrf udc drivers.

i ran into this issue on a imxrt685 board when connected to a macbook pro m5.
when i was able to reproduce the issue the host sends out a zlp to end the previous transaction, but sends a setup packet around 3 uS later.

### DIFF
--- a/drivers/usb/udc/udc_mcux_ip3511.c
+++ b/drivers/usb/udc/udc_mcux_ip3511.c
@@ -186,10 +186,26 @@ static int udc_mcux_ctrl_feed_dout(const struct device *dev,
 
 static int udc_mcux_handler_setup(const struct device *dev, struct usb_setup_packet *setup)
 {
+	struct udc_ep_config *cfg_out = udc_get_ep_cfg(dev, USB_CONTROL_EP_OUT);
+	struct udc_ep_config *cfg_in = udc_get_ep_cfg(dev, USB_CONTROL_EP_IN);
 	int err;
 	struct net_buf *buf;
 
 	LOG_DBG("setup packet");
+
+	buf = udc_buf_get_all(cfg_out);
+	if (buf) {
+		net_buf_unref(buf);
+	}
+
+	buf = udc_buf_get_all(cfg_in);
+	if (buf) {
+		net_buf_unref(buf);
+	}
+
+	udc_ep_set_busy(cfg_out, false);
+	udc_ep_set_busy(cfg_in, false);
+
 	buf = udc_ctrl_alloc(dev, USB_CONTROL_EP_OUT,
 			sizeof(struct usb_setup_packet));
 	if (buf == NULL) {


drivers/usb/udc/udc_mcux_ip3511.c
```diff
@@ -186,10 +186,16 @@ static int udc_mcux_ctrl_feed_dout(const struct device *dev,
...
 {
+	struct udc_ep_config *cfg_out = udc_get_ep_cfg(dev, USB_CONTROL_EP_OUT);
+	struct udc_ep_config *cfg_in = udc_get_ep_cfg(dev, USB_CONTROL_EP_IN);
 	int err;
 	struct net_buf *buf;
 
 	LOG_DBG("setup packet");
+
+	udc_ep_set_busy(cfg_out, false);
+	udc_ep_set_busy(cfg_in, false);
```
jfischer-no, line 207: Perhaps it is not enough. There could be buffers enqueued that need to be unreferenced. See for example dwc2_handle_evt_setup()
https://github.com/zephyrproject-rtos/zephyr/blob/7e4235ff21266f43b1212ca1fcebdbadfed0a7d0/drivers/usb/udc/udc_dwc2.c#L890-L901

te-johan: thanks for the feedback. it worked good with this change as well.
the hal driver seems to disable ep0 in and out so it seems correct to cleanup the old buffers.

MarkWangChinese (approved):
Thanks @te-johan 
The codes changes are ok to me. I am curious about the case. Is the case as follow:The previous control transfer is setup + "data in" + "status out", when the "data in" is finished the host send one ZLP out (status out) to finish the control transfer, and send the next setup quickly, then the software process the next setup before processing the status out ACK?

jfischer-no:
> index 1418 and 1421 describes what triggers the issue seen. however the issue is first seen in the trace at index 1433, where the host cannot write to the output endpoint since the udc ip3511 driver failed to feed it.
> 
> the specification also mentions how setup transactions should be handled.
> 
> > USB 2.0 5.5.5:
> > If a Setup transaction is received by an endpoint before a previously initiated control transfer is completed,
> > the device must abort the current transfer/operation and handle the new control Setup transaction.

In your trace log, where is the control transfer not completed? At the end of the screenshot, the host abandons the status stage after the OUT-NAK and 975295 PING-NAK handshakes. It looks like a bug in the driver that your commit does not fix. Your commit only fixes the case where the host can send an arbitrary setup packet, and likely this fix keeps the device functional.

tmon-nordic:
> At the end of the screenshot, the host abandons the status stage after the OUT-NAK and 975295 PING-NAK handshakes. It looks like a bug in the driver that your commit does not fix.

I would consider that the driver did not arm the OUT endpoint for status stage, or armed it too late (i.e. after host timeout). The only other explanation is that the data stage ACK from host was not correctly received by the device, which is most likely not the case (and highly unlikely if the problem is easily reproducible).

> Your commit only fixes the case where the host can send an arbitrary setup packet, and likely this fix keeps the device functional.

This is actually required behavior, so on that merit the patch is OK. However the root cause behind not arming OUT endpoint for status stage has to be investigated.

jfischer-no:
> > Your commit only fixes the case where the host can send an arbitrary setup packet, and likely this fix keeps the device functional.
> 
> This is actually required behavior, so on that merit the patch is OK. However the root cause behind not arming OUT endpoint for status stage has to be investigated.

Yes, the commit is fine. @te-johan and @MarkWangChinese just need to agree that there is still an issue with the driver.

tmon-nordic:
> It (1418) is completed, the controller accepted the OUT data and respondent with NYET handshake. Another question is why NYET and not ACK?

NYET is actually correct response to OUT status stage at High-Speed. This basically means that controller is not able to accept next packet, and PING protocol has to be used. This is proper and compliant implementation.

jfischer-no:
> > Is the problem simply that the hardware does not have a method to determine whether status stage was complete if SETUP is received in quick succession (here ~4 us)?
> 
> yes that would be the case. but my understanding is that it is a common limitation, hence similar fixes have been applied to other udc drivers.

Where did you get it? It is not a common limitation. There should be INTSTAT_EP0OUT, INTSTAT_EP0IN and SETUP token received status bits.

MarkWangChinese:
Hi @jfischer-no @tmon-nordic The controller can distinguish the setup and endpoint0 data/status transactions. From my see, the flow is: The previous control transfer's status transaction done and the next setup transaction done is processed by the NXP USB controller driver (hal_nxp) at the same time. The NXP USB controller driver check the setup transaction done firstly in the ISR. In processing the setup transaction done, the driver cancel any data/status transactions in endpoint 0, the cancelling need to callback the status transaction successfully because the HW label it as success, **but the NXP usb controller driver seems to have bug that the status transaction is not call backed in the cancelling**. I will fix it in the NXP USB controller driver. Thanks.

drivers/usb/udc/udc_mcux_ip3511.c
```diff
@@ -186,10 +186,26 @@ static int udc_mcux_ctrl_feed_dout(const struct device *dev,
...
+		net_buf_unref(buf);
+	}
+
+	buf = udc_buf_get_all(cfg_in);
+	if (buf) {
+		net_buf_unref(buf);
+	}
+
+	udc_ep_set_busy(cfg_out, false);
+	udc_ep_set_busy(cfg_in, false);
```
MarkWangChinese, line 207: Hi @te-johan I see that the calls of `udc_ep_set_busy` are protected by `udc_mcux_lock` in other places. Please add `udc_mcux_lock` too.

MarkWangChinese (changes requested):

PR #100449, opened 2025-12-03
Bluetooth: shell: Fix bsim xterm-native-shell

alwa-nordic:
The `xterm-native-shell` snippet was broken since `CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT` was added, which redirected the native PTY to stdin/stdout instead of the PTY xterm connected to. This change disables it and does so explicitly with `=n`, even though that's the default, in case the default changes. This restores output to the xterm.

Its DT overlay also created two native PTYs instead of one, resulting two xterms, only one of which is used, the other is blank. This change removes the part of the overlay causing the second PTY.

This patch also changes the overlay so it creates a new DT node for the PTY and assign it as chosen for `zephyr,console` and `zephyr,shell-uart`, instead of patching the UART original nodes, in case they are used for something else.


tests/bluetooth/shell/snippets/xterm-native-shell/xterm-native-shell.conf
```diff
@@ -1,4 +1,4 @@
-CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT=y
+CONFIG_UART_NATIVE_PTY_0_ON_STDINOUT=n
```
aescolar, line 1: (sorry for that)
minor: maybe you want to juts set CONFIG_UART_NATIVE_PTY_0_ON_OWN_PTY=y if you want to ensure that default?

PR #100451, opened 2025-12-03
drivers: input: input_chsc5x: fix sda and scl being pulled low

allemanm:
The CHSC5X chip requires reads to be aligned to multiples of 4 bytes. If the read length is not a multiple of 4, the device pulls SDA/SCL low for about 140 ms.


pdgendt (approved):

fabiobaltieri (approved):
...interesting feature...

PR #92263, opened 2025-06-26
bluetooth: hci: userchan: support connecting over UNIX socket

dsseng:
Can be used with Bluez btvirt

Signed-off-by: Dmitrii Sharshakov <d3dx12.xx@gmail.com>

PR #100455, opened 2025-12-03
Bluetooth: userchan: Validate UNIX socket path fits in sockaddr_un

### LINKED ISSUE #100009: [Coverity CID: 548288] Copy into fixed size buffer in drivers/bluetooth/hci/userchan_bottom.c

https://github.com/zephyrproject-rtos/zephyr/tree/main/drivers/bluetooth/hci/userchan_bottom.c

Category: Security best practices violations
Function: `user_chan_unix_connect`
Component: Drivers
CID: [548288](https://scan9.scan.coverity.com/#/project-view/29271/12996?selectedIssue=548288)

Details:

https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/bluetooth/hci/userchan_bottom.c#L113

https://scan9.scan.coverity.com/#/project-view/29271/12996?selectedIssue=548288

For more information about the violation, check the [Coverity Reference](https://scan9.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_STRING_OVERFLOW). ([CWE-120](http://cwe.mitre.org/data/definitions/120.html))

Also see the related [Coverity documentation](https://docs.zephyrproject.org/latest/contributing/coverity.html) for more information.

aescolar:
sockaddr_un.sun_path is limited to 108 in Linux
(and to similarly small numbers in other systems).

Let's check that it fits before copying it, and error out otherwise (the user won't be able to connect to that socket if the path is truncated, and if we overflow we will just corrupt memory).

Introduced in #92263
Fixes #100009

### DIFF
--- a/drivers/bluetooth/hci/userchan_bottom.c
+++ b/drivers/bluetooth/hci/userchan_bottom.c
@@ -20,6 +20,7 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <nsi_errno.h>
+#include <nsi_tracing.h>
 
 #define BTPROTO_HCI      1
 #define HCI_CHANNEL_USER 1
@@ -103,6 +104,12 @@ int user_chan_unix_connect(char socket_path[])
 {
 	int fd;
 	struct sockaddr_un addr;
+	size_t socket_path_size = strlen(socket_path);
+
+	if (socket_path_size >= sizeof(addr.sun_path)) {
+		nsi_print_error_and_exit("Unix socket path too long (%zu>=%zu)\n",
+					 socket_path_size, sizeof(addr.sun_path));
+	}
 
 	fd = socket(AF_UNIX, SOCK_STREAM, 0);
 	if (fd < 0) {


drivers/bluetooth/hci/userchan_bottom.c
```diff
@@ -103,6 +104,12 @@ int user_chan_unix_connect(char socket_path[])
 {
 	int fd;
 	struct sockaddr_un addr;
+	int socket_path_size = strlen(socket_path);
+
+	if (socket_path_size >= sizeof(addr.sun_path) - 1) {
+		nsi_print_error_and_exit("Unix socket path too long (%i>%i)\n",
+					 socket_path_size, sizeof(addr.sun_path));
```
Thalley:
```suggestion
	size_t socket_path_size = strlen(socket_path);

	if (socket_path_size >= sizeof(addr.sun_path) - 1) {
		nsi_print_error_and_exit("Unix socket path too long (%zu>%zu)\n",
					 socket_path_size, sizeof(addr.sun_path));
```

Thalley (changes requested):
Minor suggestion to use correct type and format

drivers/bluetooth/hci/userchan_bottom.c
```diff
@@ -103,6 +104,12 @@ int user_chan_unix_connect(char socket_path[])
 {
 	int fd;
 	struct sockaddr_un addr;
+	int socket_path_size = strlen(socket_path);
+
+	if (socket_path_size >= sizeof(addr.sun_path) - 1) {
```
dsseng: Perhaps either -1 or = in >= has to go: we shouldn't decline strlen(...) == size-1, because we still have a byte for null terminator.

I'm not fully confident, as I'm on the go and reviewing without an ide

dsseng (approved):
Thanks for taking care of this, I left a comment, but that doesn't really matter a lot if we allow one character less than possible

drivers/bluetooth/hci/userchan_bottom.c
Thalley: Can `sizeof` even be used here? 

From `man`:
```
   UNIX domain sockets
       #include <sys/un.h>

       struct sockaddr_un {
           sa_family_t     sun_family;     /* Address family */
           char            sun_path[];     /* Socket pathname */
       };
```

Don't we need to use `strlen` on `sun_path`?

dsseng: I doubt strlen gives us a valid value, it will also likely be a read of uninitialized memory. It would make sense for this to be defined in some of the libc files

aescolar: > Can sizeof even be used here?

I think so.
That man page seems to try to not define the size, but is misleadingly showing a flexible array at the end (which should not be the case).
See https://man7.org/linux/man-pages/man7/unix.7.html which matches the actual [definition in glibc](https://elixir.bootlin.com/glibc/glibc-2.39.9000/source/socket/sys/un.h).
If it was a flexible array like that, we'd need to allocate something in the stack after sockaddr_un where sun_path would be kept. (and yes, in that case we would not be able to use sizeof() on it, but that code would not work at all either)

Thalley (approved):

mniestroj (approved):

drivers/bluetooth/hci/userchan_bottom.c
dsseng: grepping through a large enough C corpus (systemd) I can find quite some usage of `sizeof(...sun_path)`, so this should be all good

jhedberg (approved):

PR #100463, opened 2025-12-03
drivers: entropy: stm32: Keep RNG clock running during PKA operation

gfroissard:
When using STM32 PKA hardware accelerator the RNG's clock is needed. This PR makes entropy (RNG module) not turning off its clock if PKA is enabled. Before this PR, it was done only for `STM32WBAX` family.

### DIFF
--- a/drivers/entropy/entropy_stm32.c
+++ b/drivers/entropy/entropy_stm32.c
@@ -140,12 +140,20 @@ static int entropy_stm32_suspend(void)
 	LL_RNG_SetAesReset(rng, 1);
 #endif /* CONFIG_SOC_STM32WB09XX */
 
-#ifdef CONFIG_SOC_SERIES_STM32WBAX
-	uint32_t wait_cycles, rng_rate;
+/* PKA module isn't currently supported in Zephyr, but it could be used outside Zephyr */
+#if defined(PKA)
+	if (__HAL_RCC_PKA_IS_CLK_ENABLED() && LL_PKA_IsEnabled(PKA)) {
+#if defined(CONFIG_SOC_SERIES_STM32WBX) || defined(CONFIG_STM32H7_DUAL_CORE)
+		z_stm32_hsem_unlock(CFG_HW_RNG_SEMID);
+#endif /* CONFIG_SOC_SERIES_STM32WBX || CONFIG_STM32H7_DUAL_CORE */
 
-	if (LL_PKA_IsEnabled(PKA)) {
+		/* PKA needs RNG clock, so exit here if in use */
 		return 0;
 	}
+#endif /* PKA */
+
+#ifdef CONFIG_SOC_SERIES_STM32WBAX
+	uint32_t wait_cycles, rng_rate;
 
 	if (clock_control_get_rate(dev_data->clock,
 			(clock_control_subsys_t) &dev_cfg->pclken[0],


drivers/entropy/entropy_stm32.c
```diff
@@ -140,12 +140,15 @@ static int entropy_stm32_suspend(void)
 	LL_RNG_SetAesReset(rng, 1);
 #endif /* CONFIG_SOC_STM32WB09XX */
 
-#ifdef CONFIG_SOC_SERIES_STM32WBAX
-	uint32_t wait_cycles, rng_rate;
-
+#if defined(PKA)
```
mathieuchopstm, line 144: This will hang on N6 if PKA peripheral is not clocked.

Please add a `if (__HAL_RCC_PKA_IS_CLK_ENABLED())` check before `LL_PKA_IsEnabled()`

mathieuchopstm (changes requested):

drivers/entropy/entropy_stm32.c
gfroissard: I've added this check, Thanks !

mathieuchopstm (dismissed):
Needs some testing (notably N6) but LGTM otherwise.

drivers/entropy/entropy_stm32.c
```diff
@@ -140,12 +140,15 @@ static int entropy_stm32_suspend(void)
 	LL_RNG_SetAesReset(rng, 1);
 #endif /* CONFIG_SOC_STM32WB09XX */
 
-#ifdef CONFIG_SOC_SERIES_STM32WBAX
-	uint32_t wait_cycles, rng_rate;
-
-	if (LL_PKA_IsEnabled(PKA)) {
+#if defined(PKA)
```
erwango, line 144: Please add a comment to explicit that we're talking about the use of PKA "outside of zephyr" as not supported yet.
(otherwise this wouldn't be implemented this way).

drivers/entropy/entropy_stm32.c
```diff
@@ -140,12 +140,15 @@ static int entropy_stm32_suspend(void)
...
+	/* PKA module needs RNG clock */
+	if (__HAL_RCC_PKA_IS_CLK_ENABLED() && LL_PKA_IsEnabled(PKA)) {
 		return 0;
```
erwango:
```
	if (__HAL_RCC_PKA_IS_CLK_ENABLED() && LL_PKA_IsEnabled(PKA)) {
		/* PKA needs RNG clock, so exit here if in use */
```

erwango (changes requested):
Thanks for sharing this fix, some tweaks required.

drivers/entropy/entropy_stm32.c
gfroissard: Thanks for the review, done

drivers/entropy/entropy_stm32.c
gfroissard: Thanks for the suggestion, done

erwango (dismissed):

mathieuchopstm (dismissed):

drivers/entropy/entropy_stm32.c
```diff
@@ -140,12 +140,16 @@ static int entropy_stm32_suspend(void)
...
 
-#ifdef CONFIG_SOC_SERIES_STM32WBAX
-	uint32_t wait_cycles, rng_rate;
-
-	if (LL_PKA_IsEnabled(PKA)) {
+/* PKA module isn't currently supported in Zephyr, but it could be used outside Zephyr */
+#if defined(PKA)
+	if (__HAL_RCC_PKA_IS_CLK_ENABLED() && LL_PKA_IsEnabled(PKA)) {
+		/* PKA needs RNG clock, so exit here if in use */
 		return 0;
```
etienne-lms, line 151: Unlock semaphore for WBX and H7 dual core devices?
```suggestion
#if defined(CONFIG_SOC_SERIES_STM32WBX) || defined(CONFIG_STM32H7_DUAL_CORE)
		z_stm32_hsem_unlock(CFG_HW_RNG_SEMID);
#endif /* CONFIG_SOC_SERIES_STM32WBX || CONFIG_STM32H7_DUAL_CORE */
		return 0;
```
**(edited) actually only needed for WBXX.**

etienne-lms (changes requested):

drivers/entropy/entropy_stm32.c
gfroissard: You're right thanks for catching this. I've added the unlock for both to make it more coherent with the rest of this file

mathieuchopstm (approved):

etienne-lms (approved):

drivers/entropy/entropy_stm32.c
```diff
@@ -140,12 +140,20 @@ static int entropy_stm32_suspend(void)
 	LL_RNG_SetAesReset(rng, 1);
 #endif /* CONFIG_SOC_STM32WB09XX */
 
-#ifdef CONFIG_SOC_SERIES_STM32WBAX
-	uint32_t wait_cycles, rng_rate;
+/* PKA module isn't currently supported in Zephyr, but it could be used outside Zephyr */
+#if defined(PKA)
+	if (__HAL_RCC_PKA_IS_CLK_ENABLED() && LL_PKA_IsEnabled(PKA)) {
```
FRASTM, line 145: instead of using HAL,   could we  continue with LL drivers  
 and `LL_AHB2_GRP1_IsEnabledClock(LL_AHB2_GRP1_PERIPH_PKA)`

FRASTM (approved):

drivers/entropy/entropy_stm32.c
mathieuchopstm: Not possible - the LL function to use varies between series. This HAL function is defined inside `stm32XXxx_hal_rcc.h` so it's basically like an LL call anyways.

erwango (approved):

drivers/entropy/entropy_stm32.c
msmttchr, line 145: Unfortunately, this change broke compilation in STM32WB06/07 who do not have `LL_PKA_IsEnabled` function.

mathieuchopstm: Fixup PR open https://github.com/zephyrproject-rtos/zephyr/pull/101291

PR #98507, opened 2025-10-29
Settings documentation fixes

lindblandro:
Harmonize the settings docstring style within the public header. Try to bring the style closer to what is used elsewhere in-tree.

PR #100465, opened 2025-12-03
Settings doc follow up fixes

lindblandro:
Some minor documentation fixes that were noticed in #98507.

### DIFF
--- a/include/zephyr/settings/settings.h
+++ b/include/zephyr/settings/settings.h
@@ -33,18 +33,24 @@ extern "C" {
  * @{
  */
 
-#define SETTINGS_MAX_DIR_DEPTH  8 /* max depth of settings tree */
+/** Max depth of settings tree */
+#define SETTINGS_MAX_DIR_DEPTH  8
+/** Max setting name length */
 #define SETTINGS_MAX_NAME_LEN   (8 * SETTINGS_MAX_DIR_DEPTH)
+/** Max setting value length */
 #define SETTINGS_MAX_VAL_LEN    256
+/** Setting name field separator */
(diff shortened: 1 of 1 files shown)


de-nordic (changes requested):
Please squash both commits into "docs: settings: add missing API documentation" commit; it is not needed to mention clang formatting.

de-nordic (approved):
Looks OK.

include/zephyr/settings/settings.h
```diff
@@ -33,18 +33,24 @@ extern "C" {
...
-/* place for settings additions:
+/**
+ * Place for settings additions:
  * up to 7 separators, '=', '\0'
  */
 #define SETTINGS_EXTRA_LEN ((SETTINGS_MAX_DIR_DEPTH - 1) + 2)
 
-/* Maximum Settings name length including separators */
+/** Maximum Settings name length including separators */
 #define SETTINGS_FULL_NAME_LEN ((SETTINGS_MAX_NAME_LEN) + (SETTINGS_EXTRA_LEN) + 1)
```
kartben, line 54: Are these really meant to be public API (stuff that end users would care about and need to use) or is it in fact internal?

lindblandro: Frankly, it's hard to tell. The literal value of `SETTINGS_NAME_SEPARATOR` is used in the examples ("path/to/value") so these might document the API contract to some extent. These could also possible be used when writing set/get handlers where users might want to construct setting paths dynamically. For example, the limit values (`SETTINGS_MAX_*`) are used by the settings backend drivers as well as secure storage subsys, settings shell, and bluetooth.

mbolivar (approved):

PR #100467, opened 2025-12-03
drivers: spi: stm32: add support for baudrate prescaler bypass

gautierg-st:
Some STM32 SPI have the ability to bypass the baudrate prescaler, which is equivalent to having a prescaler of 1.
Adds this new value in the prescaler calculation if available.

### DIFF
--- a/drivers/spi/spi_stm32.c
+++ b/drivers/spi/spi_stm32.c
@@ -811,7 +811,10 @@ static int spi_stm32_configure(const struct device *dev,
 {
 	const struct spi_stm32_config *cfg = dev->config;
 	struct spi_stm32_data *data = dev->data;
-	const uint32_t scaler[] = {
+	static const uint32_t scaler[] = {
+#ifdef LL_SPI_BAUDRATEPRESCALER_BYPASS
+		LL_SPI_BAUDRATEPRESCALER_BYPASS,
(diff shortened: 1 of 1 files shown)


drivers/spi/spi_stm32.c
mathieuchopstm, line 1: While at it, please make `const uint32_t scaler[]` static.

drivers/spi/spi_ll_stm32.c
```diff
@@ -824,6 +827,7 @@ static int spi_stm32_configure(const struct device *dev,
 	SPI_TypeDef *spi = cfg->spi;
 	uint32_t clock;
 	int br;
+	int shift;
```
mathieuchopstm: Nit: could be `const`

drivers/spi/spi_stm32.c
gautierg-st: Done

drivers/spi/spi_ll_stm32.c
gautierg-st: Done

mathieuchopstm (commented):
LGTM; a rebase is needed though

gautierg-st:
Rebased

djiatsaf-st (approved):

erwango (approved):

PR #96822, opened 2025-09-30
Add RTT support to Twister and Pytest

MarkoSagadin:
## Summary

This PR ~~fixes shell RTT interactions via `west rtt` command and~~ adds RTT support in Twister and pytest-twister-harness.
This enables device testing via RTT in addition to existing UART and serial pty transports.

## Details

I have split this work into 3 separate commits, where each one makes changes to the specific part of the code. 

I have tried to capture all the details in the commit messages, so check them out, but an overall view of the PR looks like this:

* ~~First commit fixes `west rtt` so that is suitable for interacting with the shell subsys. Previously, it was only suitable for presenting logs captured via the RTT, meaning that arrow and navigation keys weren't correctly interpreted. Separate Unix and Windows implementations were needed to resolve this.~~
* Second commit adds RTT support to the pytest-twister-harness.
* Third commit add RTT support to the Twister itself.

I have been mostly testing this with the nRF7002dk board. 

Below are some commands that I used to test this.

**IMPORTANT**: by PR feedback I have moved the `west rtt` fix commit into a separate PR, #100473. Those wishing to test below commands need to first apply that single commit over this branch so that `west rtt` will properly interact with Zephyr's shell.

<details>
  <summary><h3>Testing instructions - click me</h3></summary>
  
### Run single ZBus test

Uses Uart for communication (see this as a control test group, I want to show that RTT support didn't break anything else).
```bash
west twister \
    -T tests/subsys/zbus/channel_id \
    -p nrf7002dk/nrf5340/cpuapp \
    --device-testing \
    --device-serial /dev/ttyACM1 \
    --log-level DEBUG \
    -vvv
```

Uses RTT for communication (`-x` flags add additional flags to the builds, we use them here to enable RTT as logging backend).

```bash
west twister \
    -T tests/subsys/zbus/channel_id \
    -p nrf7002dk/nrf5340/cpuapp \
    --device-testing \
    --device-rtt \
    --log-level DEBUG \
    -x=CONFIG_USE_SEGGER_RTT=y \
    -x=CONFIG_LOG_BACKEND_RTT=y \
    -x=CONFIG_LOG_BACKEND_UART=n \
    -x=CONFIG_LOG_MODE_DEFERRED=y \
    -vvv
```

### Run several ZBus tests

They use `console` or `ztest` harnesses.

Uses UART:

```bash
west twister \
    -T tests/subsys/zbus \
    -p nrf7002dk/nrf5340/cpuapp \
    --device-testing \
[...]

PR #100473, opened 2025-12-03
Fix west rtt for shell interaction

MarkoSagadin:
## Description

This PR fixes shell RTT interactions via west rtt command. Previously, it was only suitable for presenting logs captured via the RTT, meaning that arrow and navigation keys weren't correctly interpreted. Separate Unix and Windows implementations were needed to resolve this. More details can be found in the commit message.

**Note**: The commit in this PR was moved from the #96822 as per request from this [comment](https://github.com/zephyrproject-rtos/zephyr/pull/96822#issuecomment-3572412537).

### DIFF
--- a/scripts/west_commands/runners/core.py
+++ b/scripts/west_commands/runners/core.py
@@ -13,6 +13,7 @@ as well as some other helpers for concrete runner classes.
 
 import abc
 import argparse
+import contextlib
 import errno
 import logging
 import os
@@ -25,6 +26,7 @@ import signal
 import socket
 import subprocess
 import sys
+import threading
 from dataclasses import dataclass, field
 from enum import Enum
 from functools import partial
@@ -1032,23 +1034,137 @@ class ZephyrBinaryRunner(abc.ABC):
             # Send the given string before entering the interactive mode
             sock.sendall(send_on_connect.encode('ascii'))
 
-        # Otherwise, use a pure python implementation. This will work well for logging,
-        # but input is line based only.
-        sel = selectors.DefaultSelector()
-        sel.register(sys.stdin, selectors.EVENT_READ)
-        sel.register(sock, selectors.EVENT_READ)
-        while True:
-            events = sel.select()
-            for key, _ in events:
-                if key.fileobj == sys.stdin:
-                    # The regression https://github.com/python/mypy/issues/21199 in mypy v1.20
-                    # causes the type of sys.stdin to wrongly become the type of key.fileobj,
-                    # which is a "plain" file without .readline()
-                    text = sys.stdin.readline()  # type: ignore[union-attr]
-                    if text:
-                        sock.send(text.encode())
+        if platform.system() == 'Windows':
+            with self._windows_vt_console_mode():
+                self._threaded_telnet_client(sock)
+        else:
+            with self._unix_stdin_cbreak_mode():
+                self._threaded_telnet_client(sock)
 
-                elif key.fileobj == sock:
-                    resp = sock.recv(2048)
-                    if resp:
-                        print(resp.decode(), end='')
+    @contextlib.contextmanager
+    def _unix_stdin_cbreak_mode(self):
+        import termios
+        import tty
+
+        fd = sys.stdin.fileno()
+        orig_attr = termios.tcgetattr(fd)  # type: ignore
+        try:
+            # Enable cbreak mode on the keyboard input, that way all shell editing
+            # commands (arrow keys, backspace, etc.) work as expected.
+            tty.setcbreak(fd)  # type: ignore
+            yield
+        finally:
+            termios.tcsetattr(fd, termios.TCSADRAIN, orig_attr)  # type: ignore
+
+    @contextlib.contextmanager
+    def _windows_vt_console_mode(self):
+        import ctypes
+        from ctypes import byref, c_ulong
+
+        kernel32 = ctypes.windll.kernel32  # type: ignore[attr-defined]
+
+        # Console mode flag constants
+        ENABLE_PROCESSED_INPUT             = 0x0001
+        ENABLE_LINE_INPUT                  = 0x0002
+        ENABLE_ECHO_INPUT                  = 0x0004
+        ENABLE_VIRTUAL_TERMINAL_INPUT      = 0x0200
+        ENABLE_PROCESSED_OUTPUT            = 0x0001
+        ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+
+        STD_INPUT_HANDLE  = -10
+        STD_OUTPUT_HANDLE = -11
+
+        stdin = kernel32.GetStdHandle(STD_INPUT_HANDLE)
+        stdout = kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
+
+        old_in_mode = c_ulong(0)
+        old_out_mode = c_ulong(0)
+        kernel32.GetConsoleMode(stdin, byref(old_in_mode))
+        kernel32.GetConsoleMode(stdout, byref(old_out_mode))
+
+        # Enable raw VT input: turn on VT translation and PROCESSED_INPUT,
+        # turn off LINE and ECHO so os.read() returns each character immediately.
+        in_mode = (
+            (old_in_mode.value | ENABLE_VIRTUAL_TERMINAL_INPUT | ENABLE_PROCESSED_INPUT)
+            & ~ENABLE_LINE_INPUT
+            & ~ENABLE_ECHO_INPUT
+        )
+        kernel32.SetConsoleMode(stdin, in_mode)
+
+        # Enable VT output processing so ANSI sequences from the target render
+        # correctly.
+        out_mode = (
+            old_out_mode.value
+            | ENABLE_PROCESSED_OUTPUT
+            | ENABLE_VIRTUAL_TERMINAL_PROCESSING
+        )
+        kernel32.SetConsoleMode(stdout, out_mode)
+
+        try:
+            yield
+        finally:
+            kernel32.SetConsoleMode(stdin, old_in_mode.value)
+            kernel32.SetConsoleMode(stdout, old_out_mode.value)
+
+    def _threaded_telnet_client(self, sock: socket.socket) -> None:
+
+        stop_received = threading.Event()
+
+        def send_keyboard_input(stop_received: threading.Event):
+            stdin_fd = sys.stdin.fileno()
+
+            try:
+                while not stop_received.is_set():
+                    data = os.read(stdin_fd, 2048)
+                    if data and not stop_received.is_set():
+                        sock.sendall(data)
+            except OSError as e:
+                print('\n\nCommunication was closed, reason:', e)
+                stop_received.set()
+
+        def print_rtt_output(sock: socket.socket, mask):
+            _ = mask
+
+            resp = sock.recv(2048)
+            if resp:
+                print(resp.decode(errors='replace'), end='', flush=True)
+                return
+
+            print('\n\nConnection closed by the remote host, exiting!')
+            stop_received.set()
+
+        sel = selectors.DefaultSelector()
+
+        sel.register(sock, selectors.EVENT_READ, print_rtt_output)
+
+        # Mark the keyboard thread as a daemon so it does not prevent the
+        # process from exiting if it is still blocked in os.read() after the
+        # socket has been closed.
+        kb_thread = threading.Thread(target=send_keyboard_input,
+                                     args=(stop_received,), daemon=True)
+        kb_thread.start()
+
+        try:
+            while not stop_received.is_set():
+                # Below select() waits for the sock to become ready. Once it is, the
+                # callback (print_rtt_output) is called. The timeout is needed only on
+                # Windows, where signals are processed only _after_ the syscall
+                # finishes (on Linux signals interrupt the syscall). Here this is done
+                # specifically, so that the Ctrl+C combination is quickly processed,
+                # otherwise user would need to press another key (any one) for the
+                # KeyboardInterrupt to be raised.
+                events = sel.select(timeout=0.5)
+                for key, mask in events:
+                    callback = key.data
+                    callback(key.fileobj, mask)
+        except KeyboardInterrupt:
+            stop_received.set()
+        except OSError as e:
+            print('\n\nCommunication was closed, reason:', e)
+            stop_received.set()
+        finally:
+            with contextlib.suppress(Exception):
+                sel.unregister(sock)
+            with contextlib.suppress(OSError):
+                sock.close()
+            sel.close()
--- a/scripts/west_commands/runners/jlink.py
+++ b/scripts/west_commands/runners/jlink.py
@@ -418,6 +418,12 @@ class JLinkBinaryRunner(ZephyrBinaryRunner):
 
                 self.run_telnet_client('localhost', self.rtt_port, sock,
                                        send_on_connect=rtt_config)
+            except ValueError as e:
+                # ValueError exception is thrown, if PC can't connect to the J-Link and
+                # we still try to open a socket to it.
+                # Silently capture the exception, the prompt from JLinkExe already
+                # contains good message about what happened.
+                self.logger.debug(f'ValueError during RTT connection: {e}')
             except OSError as e:
                 self.logger.error(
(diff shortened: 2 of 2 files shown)


MarkoSagadin:
Remove stale label, I still care about getting this PR merged.

marc-hb (commented):
> Since we don't have tty package on the Windows, ...

Maybe I'm missing something but I think this has been available for a long time on Windows, it's just not on by default:

https://stackoverflow.com/questions/36760127/how-to-use-the-new-support-for-ansi-escape-sequences-in-the-windows-10-console

From https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences

```
Sequence | Code | Description | Behavior
-- | -- | -- | --
ESC [ <n> A | CUU | Cursor Up | Cursor up by <n>
ESC [ <n> B | CUD | Cursor Down | Cursor down by <n>
...
```

etc.

mbolivar (dismissed):

MarkoSagadin:
Hi @marc-hb, I have converted the windows-specific function to use the `ENABLE_VIRTUAL_TERMINAL_INPUT` and `ENABLE_VIRTUAL_TERMINAL_PROCESSING` from the `windll` instead of the `KEY_MAP` dict.

Please take another look, this should be ok now.

scripts/west_commands/runners/core.py
```diff
@@ -1028,20 +1029,165 @@ def run_telnet_client(
...
-            for key, _ in events:
-                if key.fileobj == sys.stdin:
-                    text = sys.stdin.readline()
-                    if text:
-                        sock.send(text.encode())
-
-                elif key.fileobj == sock:
+
+
+        # We can't type check functions from tty and termios on Windows operating
```
marc-hb: Windows comment in the unix function?

marc-hb (commented):
Would Linux and macOS lose anything if they were using the same two threads instead of `select()` ? Without the kernel32.setconsolemode() and friends of course. This would considerably reduce the divergence and OS-dependent code, allowing anyone to troubleshoot and fix most issues for any OS.

scripts/west_commands/runners/core.py
MarkoSagadin: Shouldn't be here, I have removed it.

MarkoSagadin:
Hi @marc-hb, I addressed both of your suggestions. The platform specific code is now solved with context managers, the rest of the code uses the the same read and write loops for the all platforms. Threads were replaced with the async coroutines. 

Please have another look when you find time.

marc-hb (commented):
Thanks for unifying the logic across Windows and Unix! This looks great, I think reducing that duplication will pay off from a maintenance perspective.

Wow, I didn't expect you to replace your Threads-based implementation with asyncio so quickly, I was just throwing an idea there. I admit I'm a bit disappointed by the complexity that seems required by `asyncio`; I naively expected the implementation would be somewhat simpler than with Threads... Are you sure you need a custom event loop? Can't "Streams" do the job without one? Also, are so many `try: expect OSerror` blocks required? Again, I didn't mind the threads; please correct me but performance shouldn't be an issue here so the less complex solution is probably the better one.

PS: your commit message is a bit out of date. No big deal.

MarkoSagadin:
So, the event loop is needed due to the `loop.sock_sendall` and `loop.sock_recv`, their api expects it. Keep in mind that `asyncio.get_running_loop()` just returns currently running loop, it doesn't create one.

I took a brief look at the `asyncio.open_connection` (Streams), but I missed the fact that I could also use those, I see now that I can provide an existing socket to the function.

The various `except OSError as e` are remnant from the Windows implementation: if rtt process exists due to some error or due to the Ctrl+C some graceful shut down is needed.

The performance is definitely not paramount here: both threads and asyncio seem to do just fine, there are no visible artifacts / glitches in the terminal.

You are right, the commit message needs an update, I have added few words about the current development.

---

Honestly, I am okay with either solution, I guess that both have some level of complexity which can be seen differently by people with different backgrounds. So please pick what you think would be lower maintenance burden in the long run. If needed I can still revert back to the thread implementation.

This PR was created some 4 months ago, the original code is from a PR that was opened on 30th of September. At this point I just want to get it merged.

marc-hb:
> Honestly, I am okay with either solution, I guess that both have some level of complexity which can be seen differently by people with different backgrounds. So please pick what you think would be lower maintenance burden in the long run. If needed I can still revert back to the thread implementation.

I think people (including myself) will be less familiar with the `async` model. It has big performance advantages in some contexts but this is not one of them. It can also require less code in some contexts but I'm surprised this does not seem to be one of them.

Looking back at your earlier Thread implementation as of https://github.com/zephyrproject-rtos/zephyr/commit/585e67b31cc5c443c741a2abccc11b0cf5d45af9, I just realized you probably don't require to start two new threads, starting a single Thread should be enough, no? So if you fork a (single) thread for `stdin`, then the main program can read on the socket and it can even use `select()` on both Windows and Unix! In case `select()` is "less blocking" than `recv()` and helps with KeyboardInterrupt and what not.

> This PR was created some 4 months ago, the original code is from a PR that was opened on 30th of September. At this point I just want to get it merged.

Sorry about that, as usual there is a maintainer shortage (here's a new one! #107507) and I'm not familiar with RTT. I haven't had the time to review line by line but to be honest I'm glad I didn't yet considering the high level design is still being decided and the code has been changing a lot. This being said, I never saw anything major and as soon as we get one or two "second opinions" to help settle on the design then things should accelerate.

EDIT: no promises but I suspect a single-Thread implementation could look the least review effort and help decide another reviewer to take a look.

BrandonDao:
Hi— I ran into the same bug with `rtt` independently ([issue #107637](https://github.com/zephyrproject-rtos/zephyr/issues/107637)) and had the same conclusion that the root cause of failure on Windows was the use of `select()` on a non-socket.

For what it's worth, I agree with @marc-hb that a simpler threading implementation (main thread + daemon for `stdin`) is likely better thanks to it being cleaner & more maintainable compared to the `asyncio` solution. It's actually the solution I arrived at when trying to roll my own workaround.
* I also agree with @marc-hb's point about using `select()` on the socket in the main thread.

Regardless of the concurrency model, the platform-appropriate stuff (`_windows_vt_console_mode` and `_unix_stdin_cbreak_mode`) feels self-contained + correct.

MarkoSagadin:
Hi @BrandonDao, thank you for interest in your PR. Yes, a while ago I have ran into the same issue as you did, but from a different angle: I wanted to get RTT over shell working :)

Hi @marc-hb! I have attempted to change the implementation to use a single thread + `select()` on the sock. While I got that working (RTT shell works fine), I noticed a strange side-effect: the Ctrl+C combination wasn't correctly working on the Windows, after I pressed it, nothing happened, only after I have pressed a random key the exception was raised. I have googled a bit, this seems to be connected with how `os.read()` behaves on Windows (it appears to ignore such key combination?!).

The two thread approach doesn't experience this issue. For some reason here the `stop.wait(timeout=0.1)` does the trick.

So I have added two additional commits to the PR (see the commit messages), that will also make it a bit easier for me to revert, once we decide on one:

1. First one uses single thread + `select()`
2. Second one uses two threads.

I have tested both implementations extensively, only the second one seems to be the "right" one, in the sense that it doesn't have any weird side effects.

(tbh, one more thread doesn't complicate the implementation that much, since the rtt loop is just moved to a different location, although I am not happy about those `not err.is_set() and not stop.is_set()` statements. They seem unnecessary, although they aren't.)

Please take a look when you find time.

---

> Sorry about that, as usual there is a maintainer shortage (here's a new one! https://github.com/zephyrproject-rtos/zephyr/pull/107507) and I'm not familiar with RTT. I haven't had the time to review line by line but to be honest I'm glad I didn't yet considering the high level design is still being decided and the code has been changing a lot. This being said, I never saw anything major and as soon as we get one or two "second opinions" to help settle on the design then things should accelerate.

No slight was meant against you, I am happy that I got a proper feedback from you :)

marc-hb:
> I have attempted to change the implementation to use a single thread + select() on the sock. While I got that working (RTT shell works fine), I noticed a strange side-effect: the Ctrl+C combination wasn't correctly working on the Windows, after I pressed it, nothing happened, only after I have pressed a random key the exception was raised. I have googled a bit, this seems to be connected with how os.read() behaves on Windows (it appears to ignore such key combination?!).

Interesting, it looks like SIGINT does not interrupt Windows system calls - unlike Unix system calls that get interrupted. AFAICT, SIGINT is not "ignored" when in Windows syscalls but you have to wait until the syscall completes. I found (too much) useful information on these pages: https://vorpus.org/blog/control-c-handling-in-python-and-trio/ , https://github.com/python-trio/trio/issues/42. I bet there are shorter pages but these ones seem extremely well informed (and too detailed).

So, instead of having the main thread doing nothing but waiting for the SIGINT, I think you could have one stdin thread + the main thread doing something like this:

```python

   while True:
      if select.select([sock_read], timeout=1-2 seconds,...)
         break
      # Ctrl-C / SIGINT gets received here with 1-2 seconds latency maximum
      # That comment is important!
```

It may look weird and counter-productive to manually "poll" select() in a loop, but I think it's actually a pretty good trade-off:
- non-zero but very low CPU usage when nothing happens: runs only once every 1-2 seconds and extremely briefly
- zero added network latency cause select() is still there doing its job
- Very decent Ctrl-C latency.
- Barely any additional code.

 
> (tbh, one more thread doesn't complicate the implementation that much, since the rtt loop is just moved to a different location, 

It's not just the complexity of the code, it's also the fact that the more threads there are and the more potential races and situations there are and the more painful it is to use a debugger or to debug generally speaking (another reason why I suggested async a while back but let's park that)

> although I am not happy about those not err.is_set() and not stop.is_set() statements. They seem unnecessary, although they aren't.)

I haven't looked at those in detail yet (sorry again), but I don't see anything shocking yet. Ideally, threads should always ask each other politely to stop. Interupting/killing at random points in time is always bad and the crazy length of https://vorpus.org/blog/control-c-handling-in-python-and-trio/ is a great demonstration of how bad random interruptions are :-)

20 years old throwback:
https://stackoverflow.com/questions/16504140/thread-stop-deprecated

MarkoSagadin:
> AFAICT, SIGINT is not "ignored" when in Windows syscalls but you have to wait until the syscall completes. 

Ah, the more you know. I have briefly looked at the blog post and the issue, but I will admit that I haven't read them completely. I do however understand the issues that can happen by terminating some process/thread unexpectedly. 


> So, instead of having the main thread doing nothing but waiting for the SIGINT, I think you could have one stdin thread + the main thread doing something like this:

This is actually the solution!

If I only add timeout parameter to the `sel.select()`, the Ctrl+C is correctly captured on Windows. I have did that in the [36a6df8](https://github.com/zephyrproject-rtos/zephyr/pull/100473/commits/36a6df8e8443c1e9d914006759cb503ece314b04). Only difference is that I decreased the timeout to 0.5, it feels a bit more snappier than 1 second.

I would be ok with going forward with the single thread plus select implementation. What do you think? I have rebased to the latest main and pushed mentioned timeout change.


> 20 years old throwback:
https://stackoverflow.com/questions/16504140/thread-stop-deprecated

I feel like that there should be (probably is, but I don't know it) some kind of "bible" book on the topic of threads and issues that can arise from them. I know of this: https://greenteapress.com/semaphores/LittleBookOfSemaphores.pdf

MarkoSagadin:
Naturally, a comment above `sel.select(timeout=0.5)`, that explains what is going on, should be added.

scripts/west_commands/runners/core.py
```diff
@@ -1032,22 +1034,129 @@ def run_telnet_client(
...
+                        break
+                    print(resp.decode(errors="replace"), end='', flush=True)
+            except OSError as e:
+                if not err.is_set() and not stop.is_set():
+                    err.set()
+                    print("\n\nCommunication was closed, reason:", e)
+                stop.set()
+
+        stop = threading.Event()
+        err  = threading.Event()
```
marc-hb: Could you make these event names a tiny bit longer? Right now they are so short and generic that they look like things built in either Python or the threading module which makes the code a bit harder to read.

I mean things like `stop_requested` and `recv_error` - or anything better but just as long.

MarkoSagadin: I know what you mean, they are a bit short.

I changed them to `stop_requested` and `error_received`.

MarkoSagadin:
@marc-hb 

I have added a comment and renamed the two variables.

marc-hb:
I just had a look at the 3 approaches and they all look OK! However, since you got the 1-thread approach working now, I think you can drop the 2 threads approach because why fork 2 threads when a single one does the job? Also, it's a bit surprising to have 3 different alternatives as commits in the same PR. Spoiled with choice! But that choice could make some reviewers wait until there are fewer alternatives left to review?

The commit with the 2-threads version https://github.com/zephyrproject-rtos/zephyr/commit/d43ea8b2574d7fe39f96faa696419233c2a42f23 will stay available for a long time (forever?) anyway even after you drop it from the PR.

I had another look at the `async` alternative  and it felt simpler than as last time... Did you simplify it a bit? Or maybe my brain got used to it somehow. I don't have a strong preference but I suspect most people would be more familiar with a thread. The async code is about the same size. I would love someone new to express a preference.

Impressive work anyway, thank you for all the efforts!

MarkoSagadin:
I went ahead and dropped the commits with async and two thread implementations, now the single thread + select remains.

> I had another look at the async alternative and it felt simpler than as last time... Did you simplify it a bit? Or maybe my brain got used to it somehow.

Nope, stayed the same :) 

> Impressive work anyway, thank you for all the efforts!

Thank you for kind words :)

I have also rebased against the latest main and force-pushed. 

>  I would love someone new to express a preference.

Maybe @mbolivar?

scripts/west_commands/runners/jlink.py
```diff
@@ -418,13 +418,22 @@ def do_run(self, command, **kwargs):
 
                 self.run_telnet_client('localhost', self.rtt_port, sock,
                                        send_on_connect=rtt_config)
+            except ValueError:
+                # ValueError expection is thrown, if PC can't connect to the J-Link and
```
tpambor:
```suggestion
                # ValueError exception is thrown, if PC can't connect to the J-Link and
```
nit

scripts/west_commands/runners/core.py
```diff
@@ -1032,23 +1034,131 @@ def run_telnet_client(
...
+        # Console mode flag constants
+        ENABLE_PROCESSED_INPUT             = 0x0001
+        ENABLE_LINE_INPUT                  = 0x0002
+        ENABLE_ECHO_INPUT                  = 0x0004
+        ENABLE_VIRTUAL_TERMINAL_INPUT      = 0x0200
+        ENABLE_PROCESSED_OUTPUT            = 0x0001
+        ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+
+        stdin = kernel32.GetStdHandle(-10)
+        stdout = kernel32.GetStdHandle(-11)
```
tpambor:
```suggestion
        STD_INPUT_HANDLE = -10
        STD_OUTPUT_HANDLE = -11

        stdin = kernel32.GetStdHandle(STD_INPUT_HANDLE)
        stdout = kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
```

tpambor (dismissed):
LGTM

scripts/west_commands/runners/jlink.py
```diff
@@ -418,13 +418,22 @@ def do_run(self, command, **kwargs):
...
+                # we still try to open a socket to it.
+                # Silently capture the exception, the prompt from JLinkExe already
+                # contains good message about what happened.
+                pass
```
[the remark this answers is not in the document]
MarkoSagadin: > and it's not clear what in the try-block would raise ValueError during a failed socket connect (socket.connect/send typically raise OSError).

Well, that is explained in the first sentence of my comment.

I replaced a `pass` with a debug log statement, so if any developer is debugging this part of the code here, the raised exception will be visible.

MarkoSagadin:
Thank you for approval @tpambor! I have addressed your comments and most of the Copilot's comments.

I didn't address two comments related to redirection of the stdin/stdout, since this isn't the usecase that is expected to be covered by the `west rtt` command.

tpambor (approved):

pdgendt (approved):

PR #100477, opened 2025-12-03
dts: nordic: Fix ngpios for nRF54L15 (and variants)

carlescufi:
The correct number is 7, since the QFN52 (QGAA) variant has P0.00 to P0.06 present. See:
https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/chapters/pin.html#ariaid-title6

### DIFF
--- a/dts/vendor/nordic/nrf54l_05_10_15.dtsi
+++ b/dts/vendor/nordic/nrf54l_05_10_15.dtsi
@@ -695,7 +695,7 @@
 				gpio-controller;
 				reg = <0x10a000 0x300>;
 				#gpio-cells = <2>;
-				ngpios = <5>;
+				ngpios = <7>;
 				status = "disabled";
 				port = <0>;
 				gpiote-instance = <&gpiote30>;


bjarki-andreasen (approved):
Looks correct, datasheet says 7 pins, though dk only shows 5

tmon-nordic (dismissed):

dts/vendor/nordic/nrf54l_05_10_15.dtsi
```diff
@@ -695,7 +695,7 @@
 				gpio-controller;
 				reg = <0x10a000 0x300>;
 				#gpio-cells = <2>;
-				ngpios = <5>;
```
tmon-nordic, line 698: Only L15 has 7 pins on gpio0. How to handle the fact that 05 and 10 have only 5? Is separate dtsi needed?

carlescufi: We don't make a distinction for these and here it will only mean that we reflect the max config.

tmon-nordic: We as in who? Zephyr? Isn't devicetree supposed to describe the hardware? Why should there be devicetree definition with wrong pin count for 05 and 10 series?

nika-nordic: @tmon-nordic regarding this statement: `Only L15 has 7 pins on gpio0. How to handle the fact that 05 and 10 have only 5?` - per MDK definition L05, L10 and L15 have 7 pins on P0 available:
- https://github.com/NordicSemiconductor/nrfx/blob/83b3a6b657f656c8feb2984586d369d3c9246080/bsp/stable/mdk/nrf54l05_application_peripherals.h#L778
- https://github.com/NordicSemiconductor/nrfx/blob/83b3a6b657f656c8feb2984586d369d3c9246080/bsp/stable/mdk/nrf54l15_application_peripherals.h#L778

Is reduced number of pins by your statement because of different package? If yes, then I believe we never differentiated between packages, at least on MDK level.

kl-cruz: @carlescufi imo the best could be to remove gpio0 description from this general file and move it to:
- https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/vendor/nordic/nrf54l05.dtsi
- https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/vendor/nordic/nrf54l10.dtsi
- https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/vendor/nordic/nrf54l15.dtsi

tmon-nordic: > Is reduced number of pins by your statement because of different package? If yes, then I believe we never differentiated between packages, at least on MDK level.

The 5 pins comes from current version of datasheet. I don't know why there is discrepancy between datasheet and MDK and which one should be fixed (if any).

kl-cruz (approved):

tmon-nordic (approved):
Turns out the letters in product name describe different variant, i.e. nRF54L10 is different from nRF54LV10. This is correct for L-series without any additional letters.

PR #100478, opened 2025-12-03
cmake: toolchain/xt-clang: allow to skip FLIX generation

### LINKED ISSUE #100892: intel_adsp/ace40/nvl/sim: kernel.memory_protection.syscalls.kyield and kernel.memory_protection.syscalls.timeslicing tests failures
### Describe the bug

Tests kernel.memory_protection.syscalls.timeslicing and kernel.memory_protection.syscalls.kyield fail on intel_adsp/ace40/nvl/sim with unknown error message.

### Regression

- [ ] This is a regression.

### Steps to reproduce

west twister -p intel_adsp/ace40/nvl/sim -s kernel.memory_protection.syscalls.kyield
or
west build -p -b intel_adsp/ace40/nvl/sim tests/kernel/mem_protect/syscalls -T kernel.memory_protection.syscalls.kyield

west twister -p intel_adsp/ace40/nvl/sim -s kernel.memory_protection.syscalls.timeslicing
or
west build -p -b intel_adsp/ace40/nvl/sim tests/kernel/mem_protect/syscalls -T kernel.memory_protection.syscalls.timeslicing

### Relevant log output

```shell
NOTE:        0.0/000: 0[ms]: Simulation initialization done, took: 1.3s, starting simulation...
[TOOL]: Initializing
[TOOL]: Initializing IOB Manager
[TOOL]: Executing step: LOAD_ROM_EXT (file_id=rom_ext_bin)
[TOOL]: Loading ROM_EXT: '/__w/os.rtos.zephyr.zephyr-intel/intel/std_sim/intel_ace40_nvl/dsp_lib_rom_ext_nvl_release.bin'
[TOOL]: Executing step: LOAD_FW (commands: #1, file_id=fw_bin)
[TOOL]: Loading FW: '/tmp/acesim.tmpf_3s9zda/acesim.tmp9ulowpmk'
[TOOL]: Streamer: creating stream: streamer_type=INPUT file_type=FW node_id=0x0 file=/tmp/acesim.tmpf_3s9zda/acesim.tmp9ulowpmk
[TOOL]: Streamer: node_id=0x0 sent: 331776 bytes (331776/331776)
[TOOL]: Streamer: node_id=0x0 sent: 0 bytes (331776/331776)
[TOOL]: Streamer: node_id=0x0 processed data drained
NOTE: 616456645.3/002: 1004[ms]: Simulation finished. Executed 1004ms on target in 00h 03min 32.715s (212715ms), Slowdown: 211.9x
Simulator exited with code: 0
[TOOL]: Exception caught: Step: LOAD_FW (commands: #1, file_id=fw_bin), from: /tmp/acesim.tmpf_3s9zda/acesim.tmpsyzwtzlw:25 failed. Reason: FwReadyNotification not received
[TOOL]: Exiting with code: 1

INFO    -  91/276 intel_adsp/ace40/nvl/sim  kernel.memory_protection.syscalls.timeslicing      FAILED rc=2 (custom 1440.617s <xt-clang>)
INFO    - /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.timeslicing/handler.log
ERROR   - ninja: Entering directory `/.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.timeslicing'
[0/1] cd /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.timeslicing && /.../acesim.py --soc ace40 --rimage /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.timeslicing/zephyr/zephyr.ri


START - test_syscall_switch_stress
Running syscall switch stress test with 16 threads on 4 cpu(s)

INFO    -  92/276 intel_adsp/ace40/nvl/sim  kernel.memory_protection.syscalls.kyield           FAILED rc=2 (custom 1440.616s <xt-clang>)
INFO    - /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.kyield/handler.log
ERROR   - ninja: Entering directory `/.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.kyield'
[0/1] cd /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.kyield && /.../acesim.py --soc ace40 --rimage /.../tests/kernel/mem_protect/syscalls/kernel.memory_protection.syscalls.kyield/zephyr/zephyr.ri
[TOOL]: DSP SIM TOOL, built Sep  6 2024 at 14:53:25
[TOOL]: No outputsfiles to load
```

### Impact

### Environment

Zephyr version: v4.3.0-2203-g9f0ce75b411a
Toolchain: zephyr-sdk-0.17.4

### Additional Context

_No response_

dcpleung:
This adds compiler flag to instruct xt-clang not to generate FLIX (Flexible Length Instruction Xtensions) instructions, where multiple instructions can be fused into one instruction. FLIX can provide performance and code size advantages.

When userspace is enabled, FLIX may reorder memory access where memory is accessed before code determines whether we should perform that memory access. For example, when guarding memory access via k_is_user_context(), and some kernel variables can only be accessed via kernel mode, FLIX may reorder memory access such that these variables are accessed before k_is_user_context(). This results in access violation.

FLIX generation is enabled by default. This lets the application developers to disable FLIX generation if needed.

### DIFF
--- a/Kconfig.zephyr
+++ b/Kconfig.zephyr
@@ -629,6 +629,37 @@ config TOOLCHAIN_SUPPORTS_VLA_IN_STATEMENTS
 	  Hidden symbol to state if the toolchain can handle vla in
 	  statements.
 
+menu "Code Generation"
+
+choice COMPILER_CODEGEN_VLIW
+	prompt "VLIW (Very Long Instruction Word)"
+	default COMPILER_CODEGEN_VLIW_AUTO
+	help
+	  This option instructs the compiler whether to use VLIW
+	  (Very Long Instruction Word) instructions. VLIW allows
+	  the compiler to fuse instructions for parallel execution.
+
+config COMPILER_CODEGEN_VLIW_AUTO
+	bool "Auto"
+	help
+	  Let the compiler decide whether to generate VLIW instructions.
+
+	  This usually means using the compiler defaults.
+
+config COMPILER_CODEGEN_VLIW_ENABLED
+	bool "Generate VLIW"
+	help
+	  Explicitly instructs the compiler to generate VLIW instructions if possible.
+
+config COMPILER_CODEGEN_VLIW_DISABLED
+	bool "DO NOT generate VLIW"
+	help
+	  Explicitly instructs the compiler to NEVER generate VLIW instructions.
+
+endchoice # COMPILER_CODEGEN_VLIW
+
+endmenu # Code Generation
+
 endmenu
 
 choice
--- a/cmake/toolchain/xt-clang/Kconfig
+++ b/cmake/toolchain/xt-clang/Kconfig
@@ -1,0 +1,11 @@
+# Copyright (c) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+if DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
+
+choice COMPILER_CODEGEN_VLIW
+	default COMPILER_CODEGEN_VLIW_DISABLED if USERSPACE
+
+endchoice
+
+endif # DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -593,6 +593,14 @@ if(CONFIG_INSTRUMENTATION)
   endif()
 endif()
 
+if(CONFIG_COMPILER_CODEGEN_VLIW_ENABLED)
+  zephyr_compile_options($<$<COMPILE_LANGUAGE:C>:$<TARGET_PROPERTY:compiler,generate_vliw>>)
+  zephyr_compile_options($<$<COMPILE_LANGUAGE:CXX>:$<TARGET_PROPERTY:compiler,generate_vliw>>)
+elseif(CONFIG_COMPILER_CODEGEN_VLIW_DISABLED)
+  zephyr_compile_options($<$<COMPILE_LANGUAGE:C>:$<TARGET_PROPERTY:compiler,no_generate_vliw>>)
+  zephyr_compile_options($<$<COMPILE_LANGUAGE:CXX>:$<TARGET_PROPERTY:compiler,no_generate_vliw>>)
+endif()
+
 # Allow the user to inject options when calling cmake, e.g.
 # 'cmake -DEXTRA_CFLAGS="-Werror -Wno-deprecated-declarations" ..'
 include(cmake/extra_flags.cmake)
--- a/cmake/compiler/compiler_flags_template.cmake
+++ b/cmake/compiler/compiler_flags_template.cmake
@@ -177,3 +177,8 @@ set_compiler_property(PROPERTY no_data_sections)
 
 # Compiler flag to enable function instrumentation
 set_compiler_property(PROPERTY func_instrumentation)
+
+# Compiler flags to enable or disable code generation of VLIW
+# (Very Long Instruction Word) instructions.
+set_compiler_property(PROPERTY generate_vliw)
+set_compiler_property(PROPERTY no_generate_vliw)
--- a/cmake/compiler/xt-clang/compiler_flags.cmake
+++ b/cmake/compiler/xt-clang/compiler_flags.cmake
@@ -28,3 +28,6 @@ set_compiler_property(PROPERTY warning_shadow_variables)
 check_set_compiler_property(APPEND PROPERTY warning_extended
                             -Wno-unknown-warning-option
 )
+
+# Do not fuse instructions via FLIX
+set_compiler_property(PROPERTY no_generate_vliw -mno-generate-flix)


arch/xtensa/core/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@
 
 zephyr_cc_option(-mlongcalls)
 
+if(CONFIG_TOOLCHAIN_XTCLANG_DO_NOT_GENERATE_FLIX)
+  zephyr_cc_option(-mno-generate-flix)
```
tejlmand: please add this flag to https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/compiler/xt-clang/compiler_flags.cmake and give it a proper name.

We try to avoid, or at least minimize, the direct use of compiler flags as it makes it harder to support multiple toolchains.
I do notice this resides in `arch/xtensa` but principle easily gets copied elsewhere.

cmake/toolchain/xt-clang/Kconfig
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+config TOOLCHAIN_XTCLANG_DO_NOT_GENERATE_FLIX
+	bool "Skip instruction fusions via FLIX"
+	depends on DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
+	default y if USERSPACE
```
tejlmand: Please turn this setting around so it is disabled per default, and then can be enabled by samples which wants it.
Like:
```suggestion
config TOOLCHAIN_XTCLANG_FLIX
	bool "Flexible Length Instruction Xtensions (FLIX)"
	depends on DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
	default y if !USERSPACE
```

Related comment, but not a required change.
Afaik FLIX is just Xtensa's form of `VLIW`, so a more generic option would be to relate the feature to `VLIW` and then let xt-clang define this to be `-mno-generate-flix`.

tejlmand (changes requested):

dcpleung:
Modified to do the compiler flag route.

Kconfig.zephyr
```diff
@@ -629,6 +629,37 @@ config TOOLCHAIN_SUPPORTS_VLA_IN_STATEMENTS
...
+choice COMPILER_CODEGEN_VLIW
+	prompt "VLIW (Very Long Instruction Word)"
+	default COMPILER_CODEGEN_VLIW_AUTO
+	help
+	  This option instructs the compiler whether to use VLIW
+	  (Very Long Instruction Word) instructions. VLIW allows
+	  the compiler to fuse instructions for parallel execution.
+
+config COMPILER_CODEGEN_VLIW_AUTO
+	bool "Let compiler decide"
```
tejlmand: Perhaps `Auto` would look nicer as prompt text, and let the help text describe what `Auto` means.
```suggestion
	bool "Auto"
```

Ref:
image
image

vs:
image

tejlmand (dismissed):
looks good.

A smaller, non-blocking, suggestion to consider.

Kconfig.zephyr
dcpleung: Changed.

nashif (dismissed):

cmake/toolchain/xt-clang/Kconfig
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+if DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
+
+choice COMPILER_CODEGEN_VLIW
+	default COMPILER_CODEGEN_VLIW_DISABLED if USERSPACE
+
+endchoice
+
+endif # DT_HAS_CDNS_TENSILICA_XTENSA_LX7_ENABLED
```
lyakh: should this also be converted? https://github.com/zephyrproject-rtos/zephyr/blob/c86c748ff037259412edb6f3a6bea52a7efad87c/cmake/compiler/xt-clang/target.cmake#L31

dcpleung: ? what do you mean?

lyakh: @dcpleung I mean - should that also use the new compiler property?

dcpleung: It is. The kconfig here is to dictate which compiler property to use in `CMakeLists.txt`.

softwarecki:
I assume you have already tested different approaches, but I want to make sure. I don't like the workaround that disables optimization for the entire project. In my opinion, this is too intrusive. Wouldn't it be better to apply a smaller change only in the problematic spots? I assume the issue with early access to the kernel-only variable occurs in very few places.

Have you tried moving the declaration of this variable below the if? Another option could be defining a pointer to this variable and using it in the return. As a last resort, you could separate access to this variable into a separate function that would be called during return.

The most important question:
Have you tried other barriers? In `include/zephyr/sys/barrier_builtin.h` there are three types of barriers defined. Only `z_barrier_isync_fence_full` uses the assembler `memory` instruction. The other two, `z_barrier_dmem_fence_full` and `z_barrier_dsync_fence_full`, use gcc built-in function `__atomic_thread_fence`.
The `compiler_barrier` also uses the `memory` instruction, which is the `isync` variant. Maybe if we inform the compiler that we want a barrier at this point, it would generate the code differently? Perhaps the `memory` instruction is not clear enough for it? I think looking at the assembly code of this function could help us understand the problem.

lyakh:
@dcpleung are there any xtensa asm instructions that are never included in FLIX?

wjablon1:
If the issue occurs in multiple places, there is a risk that we miss some of them by applying a manual approach like barriers. If we sprinkle the entire kernel code with those barriers, the performance could be the same as with FLIX disabled (or even worse) while the code readability decreased.

The performance measurements would ideal. Also, it’s worth figuring out why Xtensa XOS has just mno-l32r-flix. Maybe there are some special programming techniques used or a combination of compiler options.

I partly agree with @softwarecki that the solution is too intrusive. Some algorithms, that have zero interactions with the kernel, vastly benefit from FLIX. But as a temporary solution, I guess it’s not the end of the world. Especially because FLIX is now disabled for LLEXT modules for another reason (linking problem).

wjablon1 (approved):

wjablon1 (approved):

Kconfig.zephyr
```diff
@@ -629,6 +629,37 @@ config TOOLCHAIN_SUPPORTS_VLA_IN_STATEMENTS
 	  Hidden symbol to state if the toolchain can handle vla in
 	  statements.
 
+menu "Code Generation"
```
wjablon1, line 632: Wouldn't be better to have more flexible/generic solution like in case of "config COMPILER_OPT" (or reuse that if possible)? 
My concern is that if someone wants to fine-tune other optimizations/instructions of a complier, her or she probably needs to follow the same pattern. You can imagine how the kconfig could bloat in the future.

There are more possibilities related to FLIX in the Xtensa compiler. So the solution isn't comprehensive from the beginning. Also I would like to add a similar option but only for LLEXT modules. Should I copy-paste the entire choice section?

dcpleung: The idea is to provide a option which cannot be accidentally overridden. Doing it like `CONFIG_COMPILER_OPT` means that the application can override it, and it is going to cause confusion.

If you want it only for LLEXT, you will probably need to duplicate the compiler properties so it can be enabled in cmake.

softwarecki:
Could you share a part of the assembler code where the FLIX is generated incorrectly?

Kconfig.zephyr
```diff
@@ -629,6 +629,37 @@ config TOOLCHAIN_SUPPORTS_VLA_IN_STATEMENTS
...
+
+choice COMPILER_CODEGEN_VLIW
+	prompt "VLIW (Very Long Instruction Word)"
+	default COMPILER_CODEGEN_VLIW_AUTO
+	help
```
tejlmand, line 637: Should we have a have some form of `ARCH_HAS_VLIW` / `TOOLCHAIN_HAS_VLIW` dependency here ?
Like: `depends on ARCH_HAS_VLIW`.

I'm a bit unsure which form would be most suitable, because it is a feature of the core (architecture) but still requires proper compiler support.

tejlmand (changes requested):

Kconfig.zephyr
dcpleung: Maybe we can do the toolchain one first? If the toolchain does not support generating VLIW instructions, there is no point in exposing these options.

dcpleung: Let me know which way you prefer so I can add the kconfig(s).

SzymonRichert: @tejlmand any update from your side?

softwarecki:
> Besides, fixing one macro does not fix the fundamental issue where the compiler will generate this kind of pattern

Our responsibility is to tell the compiler where its optimizations must not apply. Using memory barriers in kernel code is not a workaround, it is the correct way to enforce ordering rules. For me the real workaround would be globally disabling the FLIX optimization in a case where the issue does not even occur in the FLIX instruction.

Please note that memory barriers are already used with `k_is_user_context` context. For example, the `ARCH_EXCEPT` macro on RISC-V, and `k_is_pre_kernel` use barriers to enforce ordering. The `gen_syscalls.py` script also inserts a barrier after `z_syscall_trap` (which calls `arch_is_user_context`).

There are projects where using macros is considered a bad practice. Here we clearly went in the opposite direction and crossed the critical mass of macros. At this point they are not a tool anymore, they are a lifestyle choice.

dcpleung:
Note that with FLIX generation disabled, the compiler emits code with correct ordering. And that no other compilers have this issue either. From my perspective, adding `compiler_barrier()` is a workaround for a misbehaving compiler... especially as the incorrect ordering is done even without emitting FLIX instructions.

Just looking at your PR, there are lots of code and complexities added for one compiler. And that's going to add quite some emitted code in binaries... for other compilers which absolutely do not need the extra baggage.

Kconfig.zephyr
SzymonRichert: @tejlmand can you take a look ?

Kconfig.zephyr
tejlmand: I think we can leave like this know, but consider if this should have a follow-up PR for toolchain / vliw support.

tejlmand (approved):

ceolin (approved):

PR #100482, opened 2025-12-03
samples: net: openthread: coprocessor: Add nrf52840dongle config

nullr0ute:
Add a config for the nrf52840dongle to use the CDC_ACM interface for the ot_uart. Without this building the OpenThread coprocessor fails to build when using the nrf52840 dongle because the UART isn't defined so fix it.


rlubos (commented):
Please check the compliance, it seems it expects tabs instead of spaces

nullr0ute:
> Please check the compliance, it seems it expects tabs instead of spaces

Thanks, I had on my list to check that, I thought I just did a copy and then edit of the `nrf52840dk_nrf52840.overlay` in that same directory. Fixed now.

rlubos (approved):

jukkar (approved):

PR #100485, opened 2025-12-04
modules: tfm: don't pad unconfirmed image

JordanYates:
If the image is not confirmed, there is no need to pad the entire flash slot with empty data. This reduces the size of
`tfm_s_zephyr_ns.signed.bin` (the file actually sent for an OTA upgrade) from 100% of the slot size down to the size of the application.


tomi-font (dismissed):

JordanYates:
@nandojve I don't really understand most of the comment, what are these "issues with TF-M FWU state machine"?

This change is about whether there is random trailing data in the generated files for `CONFIG_TFM_MCUBOOT_IMAGE_NUMBER==1`, when there is no need for it. It doesn't affect `CONFIG_TFM_MCUBOOT_IMAGE_NUMBER==2`, because that is a fundamentally different problem, since the two sides are upgraded independently AFAIK.

> Can you confirm that the S image is full with trailer when concatenated with a shrink NS image ?

I don't understand the question. This changes nothing related to the merging of hex files. What does "S image is full" mean?

> Another thing, Are you really using the TF-M FWU with the state machine proposed in https://arm-software.github.io/psa-api/fwu/1.0 ?

Nope, I have never claimed to. I would hazard a guess that practically no-one does, since I see not a single reference to `psa_fwu_start` in either upstream Zephyr or Nordics NCS (the only large fork I am aware of).

I don't see how trimming useless trailing bytes would have any effect on it, but if you can point me to the problem in that document I would be happy to take a look.

JordanYates:
> Do you have the S (secure image) with trailing to fill the full S slot when concatenating with the NS (non-Secure) image ? I mean, the outcome of the S+NS+shrink keeps S with the full slot size ?

The question you are essentially asking is "does a .hex file with holes in it generate a .bin with empty data in those holes when converted with objcopy". This has nothing to do with TF-M, `mergehex.py` or the changes in this PR, but yes, it does.

For signed, confirmed, images, the `.hex` file is something like (for IMAGE_NUMBER 1):
`{TF-M S} padding {Zephyr NS} padding {Confirmation bit}`.
Prior to this PR, signed images look like:
`{TF-M S} padding {Zephyr NS} padding`.
After this PR, signed images look like:
`{TF-M S} padding {Zephyr NS}`.

nandojve (approved):

tomi-font (approved):

PR #100491, opened 2025-12-04
tests: bluetooth: classic: sim: add GAP discovery test

lylezhu2012:
Add Bluetooth Classic GAP (Generic Access Profile) discovery test for the simulation testing infrastructure.

The test validates inquiry-based device discovery functionality, including both general and limited discovery procedures. It uses two simulated Bumble controllers connected via TCP to test central and peripheral roles performing discovery, connection establishment, and disconnection.

It is the first test suite based on the Classic simulation environment.


tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,231 @@
...
+	/* Wait for all discovery results to be processed */
+	timeout = GAP_DISCOVERY_TIMEOUT(ARRAY_SIZE(br_discover_result));
+	LOG_DBG("Will wait for GAP discovery done (timeout %us)", (uint32_t)timeout);
+	err = k_sem_take(&br_discover_sem, K_SECONDS(timeout));
+	zassert_equal(err, 0, "Failed to wait for discovery done (err %d)", err);
+
+	err = bt_br_discovery_stop();
+	if (err != 0) {
+		LOG_ERR("Failed to stop GAP discovery procedure (err %d)", err);
+	}
```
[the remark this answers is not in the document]
lylezhu2012: The error code of GAP discovery stop is ignored. Just add a log to print it.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,820 @@
...
+    if [[ $count -eq 0 ]]; then
+        log_warn "No Virtual HCI devices found"
+        return 1
+    fi
+
+    log_info "Found $count Virtual HCI devices"
+
+    # Check and fix status of all devices
+    local has_error=false
+    local auto_fix=true  # Set to false if you don't want automatic fixing
```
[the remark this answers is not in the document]
lylezhu2012: The comments is used to comment the variable can be changed to `false` if needed.

.github/workflows/bt_classic_sim_tests.yml
```diff
@@ -0,0 +1,215 @@
...
+          pip3 install --user -U west
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Install Zephyr SDK
+        run: |
+          cd ~
+          wget https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.16.5/zephyr-sdk-0.16.5_linux-x86_64.tar.xz
+          wget -O - https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.16.5/sha256.sum | shasum --check --ignore-missing
+          tar xvf zephyr-sdk-0.16.5_linux-x86_64.tar.xz
+          cd zephyr-sdk-0.16.5
```
jhedberg: Why such an old version of the SDK? The latest one is `0.17.4`. You might also want to factor out the version to a single updatable variable, and maybe also consider using the `west sdk install` command.

lylezhu2012: Thanks. Will update it.

makeshi (dismissed):

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+        while IFS= read -r -d '' script; do
+            # Check if file is executable
+            if [[ -x "$script" ]]; then
+                TEST_SCRIPTS+=("$script")
+                log_info "  Found executable: $(basename "$script")"
+            else
+                log_warn "  Found non-executable: $(basename "$script")"
+
+                # Try to make it executable
+                if chmod +x "$script" 2>/dev/null; then
```
aescolar: Please don't do this. Do not have test scripts modify the tree itself.
If a file needs to be executable it should be comited to the tree being executable.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+    # Check if ZEPHYR_BASE directory exists
+    if [[ ! -d "$ZEPHYR_BASE" ]]; then
+        log_info "$test_index/$total_tests $test_id \033[31mFAIL\033[0m (ZEPHYR_BASE dir not found)"
+        FAILED_TESTS+=("$test_id")
+        TEST_MESSAGES["$test_id"]="ZEPHYR_BASE directory not found: $ZEPHYR_BASE"
+        TEST_TIMES["$test_id"]=0
+        TEST_EXIT_CODES["$test_id"]=1
+        HAS_FAILURE=true
+        return 1
+    fi
```
aescolar: This (and 234-238) is a lot of error handling code for the same condition. An error which will result in all tests failing anyhow. Handling it per test, and in such a verbose way seems quite unnecessary.

Note this is actually redundant with code in main()

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+set -euo pipefail
+
+# Create gap_discovery directory if it doesn't exist
+mkdir -p ${ZEPHYR_BASE}/bt_classic_sim/gap_discovery
+
+# Redirect all output to log file while also displaying on terminal
+exec > >(tee -a ${ZEPHYR_BASE}/bt_classic_sim/gap_discovery/script_output.log)
+exec 2>&1
+
+# Color definitions
```
aescolar: This script is replicating a lot of code that is also in ci.bt.classic.sh
It would be nicer if it was in a common source file.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+HCI_PORT_0="9000"
+HCI_PORT_1="9001"
+BD_ADDRESS_0=""
+BD_ADDRESS_1=""
+
+# Check and kill existing controllers.py processes
+check_and_kill_existing_controllers() {
+    log_info "Checking for existing controllers.py processes..."
+
+    local controllers_script="${ZEPHYR_BASE}/tests/bluetooth/classic/sim/common/controllers.py"
```
aescolar: What if they were launched from another ZEPHYR_BASE but still hanging around?

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,254 @@
...
+	atomic_clear_bit(test_flags, TEST_FLAG_DEVICE_FOUND);
+	atomic_clear_bit(test_flags, TEST_FLAG_CONN_CONNECTED);
+	atomic_clear_bit(test_flags, TEST_FLAG_CONN_DISCONNECTED);
+
+	k_sem_reset(&br_discover_sem);
+}
+
+static void after(void *f)
+{
+}
```
aescolar: SonarCloud errors would need to be addressed.
For this, the function can just be removed, and NULL used in its place in ZTEST_SUITE

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+    local zephyr_exe="$build_dir/zephyr/zephyr.exe"
+
+    log_info "Checking for generated executable: $zephyr_exe"
+
+    if [[ ! -f "$zephyr_exe" ]]; then
+        log_error "zephyr.exe not found at $zephyr_exe"
+        return 1
+    fi
+
+    log_info "zephyr.exe found successfully at $zephyr_exe"
```
aescolar: Wouldn't this be redundant? (it would have failed just before otherwise)

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+
+    # Optional: Check if the file is executable
+    if [[ ! -x "$zephyr_exe" ]]; then
+        log_warn "zephyr.exe exists but is not executable"
+        chmod +x "$zephyr_exe"
+        log_info "Made zephyr.exe executable"
+    fi
```
aescolar: this is really unnecessary/redundant...

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+        fi
+
+        PERIPHERAL_PID=""
+    fi
+
+    return 0
+}
+
+# Build central device
+build_central_device() {
```
aescolar: This function is practically the same as `build_peer_device()`. If you want to have your own build wrapper like this, it should be refactored into something reused.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+check_and_kill_existing_controllers() {
+    log_info "Checking for existing controllers.py processes..."
+
+    local controllers_script="${ZEPHYR_BASE}/tests/bluetooth/classic/sim/common/controllers.py"
+
+    # Find all processes running controllers.py
+    local pids=$(pgrep -f "$controllers_script" 2>/dev/null || true)
+
+    if [[ -z "$pids" ]]; then
+        log_info "No existing controllers.py processes found"
```
aescolar: This message and this whole function is way too verbose.
There is no need to print "everything is fine" messages all the time.
It is much better to follow a ["silence is golden"](https://www.linfo.org/rule_of_silence.html) approach.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+# Global variables
+BUMBLE_CONTROLLER_PID=""
+PERIPHERAL_PID=""
+HCI_PORT_0="9000"
+HCI_PORT_1="9001"
+BD_ADDRESS_0=""
+BD_ADDRESS_1=""
+
+# Check and kill existing controllers.py processes
+check_and_kill_existing_controllers() {
```
aescolar: Something like this would better be in a common file to be reused.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+
+# Set cleanup on exit
+trap cleanup EXIT
+
+# Main function
+main() {
+    log_info "Starting GAP Discovery test with Bumble controllers..."
+
+    # 0. Check and kill existing controllers.py processes
+    if ! check_and_kill_existing_controllers; then
```
aescolar: Fundamental: so if there is more of these tests, they cannot be run in parallel? (also note hardcoded port numbers) (and they cannot be run in the same machine by different users at the same time?)

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+    for pid in $pids; do
+        log_info "Attempting to kill process $pid..."
+
+        # Try graceful shutdown first
+        if kill -TERM "$pid" 2>/dev/null; then
+            log_info "Sent SIGTERM to process $pid"
+
+            # Wait for graceful shutdown
+            local count=0
+            while kill -0 "$pid" 2>/dev/null && [[ $count -lt 5 ]]; do
```
aescolar: That's a lot of time..

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
aescolar: A basic design consideration/note: If there is left over processes from previous runs, either the previous run was not cleaned up properly due to some catastrophic failure or hard kill by the user not handled by the wrapping scripts (which can happen), or there is a fundamental flaw in how the whole thing is run which should be fixed.

.github/workflows/bt_classic_sim_tests.yml
```diff
@@ -0,0 +1,151 @@
...
+          pip install -r scripts/requirements-actions.txt --require-hashes
+
+      - name: Install Bluetooth Classic Simulation dependencies
+        run: |
+          pip install bumble==0.0.220
+
+      - name: Make test scripts executable
+        run: |
+          find tests/bluetooth/classic/sim -name "*.sh" -type f \
+            -exec chmod +x {} \;
```
aescolar: repeat:
If a file needs to be executable it should be committed to the tree being executable.

.github/workflows/bt_classic_sim_tests.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Run Bluetooth Classic Sim tests
+        id: run_tests
+        run: |
+          export ZEPHYR_BASE=${PWD}
+          export ZEPHYR_TOOLCHAIN_VARIANT=zephyr
+          tests/bluetooth/classic/sim/ci.bt.classic.sh
+
+      - name: Create Test Results
+        run: |
+          junit2html tests/bluetooth/classic/sim/test_logs/junit.xml tests/bluetooth/classic/sim/test_logs/junit.html
```
aescolar: Creating logs around in the tree is really not nice. Better generate all artifacts in a single folder[^1], preferably in a user configurable place (env var defined), and which defaults to be a folder in the top level (like other similar test setups and workflows do)

[^1]: `bt_classic_sim/` is already created by other scripts in this PR

.github/workflows/bt_classic_sim_tests.yml
```diff
@@ -0,0 +1,151 @@
...
+            if [[ "$failures" -gt 0 ]]; then
+              echo "::error::$failures test(s) failed"
+              exit 1
+            fi
+          fi
+
+      - name: Cleanup
+        if: always()
+        run: |
+          sudo pkill -9 zephyr.exe || true
```
aescolar: This should have happened at a lower level. If this is necessary there is a problem with some other script in this PR leaving zombie processes around.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+        else
+            abs_path="$SCRIPT_DIR/$line"
+        fi
+
+        # Check if directory exists
+        if [[ -d "$abs_path" ]]; then
+            TEST_DIRECTORIES+=("$abs_path")
+            log_info "  Added directory: $abs_path"
+        else
+            log_warn "  Directory not found (skipping): $abs_path"
```
aescolar: A user trying to run something that does not exist, should be treated as an error.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+
+        # Execute the script and redirect all output to log file
+        "$script" > "$log_file" 2>&1
+        exit $?
+    )
+    local exit_code=$?
+    set -e
+
+    # Return to original directory (don't fail if this doesn't work)
+    cd "$original_dir" 2>/dev/null || true
```
aescolar: Wasn't the directory change run in a subshell? This would be a nop..

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+
+    # Record end time
+    local end_time=$(date +%s)
```
aescolar: Nit: This only has second resolution.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+
+# Function to generate JUnit XML report
+generate_junit_xml() {
+    log_info "Generating JUnit XML report..."
+
+    local timestamp=$(date -u +"%Y-%m-%dT%H:%M:%S")
+    local total_tests=$((${#PASSED_TESTS[@]} + ${#FAILED_TESTS[@]}))
+    local total_time=0
+
+    # Calculate total time
```
aescolar: nit: I think it would be more appropriate just keeping the overall start in a separate variable, as any overhead in this script around running the tests also counts.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+    echo "Summary log: $SUMMARY_LOG"
+    echo "JUnit XML report: $JUNIT_XML"
+}
+
+# Cleanup function
+cleanup() {
+    log_info "CI BT Classic cleaning up..."
+
+    # Kill any remaining processes (try with and without sudo)
+    pkill -9 btvirt 2>/dev/null || sudo pkill -9 btvirt 2>/dev/null || true
```
aescolar: Major: A user should not need to have sudo permissions to run this script.

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+    pkill -9 btmon 2>/dev/null || sudo pkill -9 btmon 2>/dev/null || true
+    pkill -9 zephyr.exe 2>/dev/null || sudo pkill -9 zephyr.exe 2>/dev/null || true
```
aescolar: Major: Killing any zephyr.exe is not ok. There are likely to be other zephyr.exe running from other unrelated tests.
In principle the same applies to btmon

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+
+    # Generate JUnit XML report
+    generate_junit_xml
+
+    # Print summary
+    print_summary
+
+    # Return appropriate exit code
+    if [[ "$HAS_FAILURE" == true ]]; then
+        log_error "Some tests failed - exiting with error code 1"
```
aescolar: Minor: A whole summary was just printed above. This is just unnecessarily verbose

tests/bluetooth/classic/sim/ci.bt.classic.sh
```diff
@@ -0,0 +1,507 @@
...
+        return 1
+    fi
+
+    if [[ ! -d "$ZEPHYR_BASE" ]]; then
+        log_error "ZEPHYR_BASE directory does not exist: $ZEPHYR_BASE"
+        HAS_FAILURE=true
+        return 1
+    fi
+
+    log_info "ZEPHYR_BASE: $ZEPHYR_BASE"
```
aescolar: Nit: too verbose

aescolar (dismissed):
I have only done a quick very partial review. But so far:
Questions:
* Is this setup deterministic? will results depend on the machine load?
* Can this setup run tests in parallel? (or are we limited to run tests serially independently of the number of cores in the machine) (EDIT: PR was changed to parallelize tests)

Fundamental issue:
* If we are going to have another tests setup, we should have some documentation for it in Zephyr's rst.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
```diff
@@ -0,0 +1,595 @@
...
+        log_error "Failed to start bumble controllers"
+        TEST_RESULT=1
+        return 1
+    fi
+
+    # 2. Display controller information
+    display_controller_info
+
+    # 3. Build peer device
+    if ! build_peer_device; then
```
aescolar: It should be possible for users to build and run separately tests.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Parallelism should be supported. It just needs to use different ports for different cases.

Let me update the test suite to allocate the TCP ports dynamically.

lylezhu2012: Yes, I agreed. 

Let me clean up it. The created processes should be killed before the script existing. And it has be handled by this script.

.github/workflows/bt_classic_sim_tests.yml
lylezhu2012: Sure. let me remove it.

.github/workflows/bt_classic_sim_tests.yml
lylezhu2012: Yes.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: The directory was switched to "$ZEPHYR_BASE" earlier in the current function. Here it is switched back.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Removed. The executable attribute is added when creating the commit.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Added a common file `test_utils.sh`.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Fixed. Terminate the processor according to the processor ID instead of script path.

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
lylezhu2012: Removed.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Fixed. Changed it to `build_local_device()`.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Unnecessary log messages were cleared.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Added a common file `bumble_utils.sh` to manage the simulations controller related operations.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Added a dedicated function `find_available_port()` to allocate the port.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Optimised. Changed it to 3.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: Optimised.

The simulation controller processor will be created at the beginning of the script. And it will be terminated when existing the script.

.github/workflows/bt_classic_sim_tests.yml
lylezhu2012: Removed.
Added the executable attribute when creating the commit.

.github/workflows/bt_classic_sim_tests.yml
lylezhu2012: Put all log files to the same folder `bt_classic_sim`.

.github/workflows/bt_classic_sim_tests.yml
lylezhu2012: Removed.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Returned an error.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Cleaned.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Change the unit to `ms`.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Added. Calculated the total time separately.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: All of them are removed.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/ci.bt.classic.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/gap_discovery_run.sh
lylezhu2012: The script do not support this case.

The BD address of simulation controller is randomly generated.

And the precondition of building includes the peer BD address in current test case.

Therefore, when creating a simulation controller, the build and run processes should be executed sequentially. This ensures that the addresses are not modified.

doc/develop/test/bluetooth_classic_sim.rst
```diff
@@ -0,0 +1,977 @@
...
+
+When writing new tests:
+
+1. **Use sequential BD addresses**: Assign ``BD_ADDR[0]`` to the first device,
+   ``BD_ADDR[1]`` to the second, etc.
+
+2. **Document role-to-address mapping**: Clearly document which role uses which BD address
+   in test documentation and comments.
+
+3. **Define addresses in Kconfig**: For tests that need to know peer addresses, define
```
aescolar: Why not pass them as command line parameters to not need to rebuild?

doc/develop/test/bluetooth_classic_sim.rst
```diff
@@ -0,0 +1,977 @@
...
+
+:zephyr_file:`tests/bluetooth/classic/sim/tests.native_sim.txt`
+
+Purpose
+=======
+
+The ``tests.native_sim.txt`` file serves as a registry of all Bluetooth Classic simulation
+tests that should be included in the build and test execution process. This allows:
+
+- **Centralized test management**: All tests are listed in one location
```
aescolar: The way this is written, it seems you have used an LLM to generate this..
In any case, this is overly verbose and reads like advertisement with little content. Please be brief. Peoples time is valuable, and their attention limited.

lylezhu2012: It is updated.

lylezhu2012:
> We do it with our scripting. In the case of the twister job, it is called again (up to 3 times in total) told to rerun only last run failed tests. An option would be for this PR parallelizing script to retry up to N times if a test fails on its own.

Thanks. Added an argument MAX_RUN_RETRIES for retry.

tests/bluetooth/classic/sim/common/bt_sim_controller.py
```diff
@@ -0,0 +1,349 @@
...
+        on_inquiry: Callable | None = None,
+        on_complete: Callable | None = None,
+    ):
+        """Start periodic timeout callback"""
+        self.flag = True
+        self._is_running = True
+        self.count = count
+        self.interval = interval
+        if on_inquiry:
+            self.on_complete = on_inquiry
```
chengkai15:
```suggestion
            self.on_inquiry = on_inquiry
```

chengkai15:
@lylezhu2012 any update?

nashif:
given the recent changes in twister to support running babblesim tests (using script harness: https://github.com/zephyrproject-rtos/zephyr/pull/108119), was there any attempt to make this type of tests run and be managed by twister? 
It would be great to have testing all done in one worflow and managed the same way, having multiple workflows with different envionment setups that are managed by different folks is not ideal.
But in general I am not opposed for giving this a try short term with the goal of consolidating our workflows and how we run and manage tests mid to long term.

lylezhu2012:
> given the recent changes in twister to support running babblesim tests (using script harness: #108119), was there any attempt to make this type of tests run and be managed by twister? It would be great to have testing all done in one worflow and managed the same way, having multiple workflows with different envionment setups that are managed by different folks is not ideal. But in general I am not opposed for giving this a try short term with the goal of consolidating our workflows and how we run and manage tests mid to long term.

Thanks for sharing. Let me try it.

lylezhu2012:
Hi @nashif, @aescolar, @jhedberg, @Thalley, @chengkai15 

I have removed the changes related to the GitHub workflow and updated the PR using the `twister + harness + script` testing framework.

Please help review the changes.

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,244 @@
...
+	bool found;
+
+	found = peer_device_discovery(false);
+
+	zassert_true(found, "Peer device not found");
+
+	peer_device_connect();
+}
+
+#define BT_COD_MAJOR_SVC_CLASS_LIMITED_DISCOVER BIT(13)
```
chengkai15, line 165: BT_COD_MAJOR_SVC_CLASS_LIMITED_DISCOVER is defined in
include/zephyr/bluetooth/assigned_numbers.h

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,244 @@
...
+	atomic_set_bit(test_flags, TEST_FLAG_CONN_DISCONNECTED);
+}
+
+BT_CONN_CB_DEFINE(conn_callbacks) = {
+	.connected = br_connected,
+	.disconnected = br_disconnected,
+};
+
+/* Multiplier for each discovery result to estimate timeout (seconds per device) */
+#define GAP_DISCOVERY_TIMEOUT_PER_DEVICE 1.25
```
chengkai15, line 82: it seems that the length was in units of 1.28 s

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,244 @@
...
+/* Base timeout added to the total (seconds) */
+#define GAP_DISCOVERY_TIMEOUT_BASE 5.0
+#define GAP_DISCOVERY_TIMEOUT(count) \
```
chengkai15, line 85: For the Z_TIMEOUT_MS / K_MSEC / SYS_CLOCK series, suggest using integer milliseconds for time calculations is the standard convention in Zephyr.

tests/bluetooth/classic/sim/common/common_utils.sh
```diff
@@ -0,0 +1,593 @@
...
+export -f run_script_parallel
+export TMP_RESULTS_DIR
+export BT_CLASSIC_SIM
+export MAX_RUN_RETRIES
+
+# Function to execute test scripts in parallel (if GNU parallel available) or serially
+# Args: script_list (space-separated or newline-separated)
+#       max_jobs (optional, default: number of CPU cores)
+# Returns: 0 if all scripts succeed, 1 if any fail
+run_parallel() {
```
chengkai15: run_parallel/collect_parallel_results — hand-assembles JUnit <testsuites> XML, duplicating what the Twister script harness already provides.

tests/bluetooth/classic/sim/common/thread_utils.sh
```diff
@@ -0,0 +1,140 @@
...
+    fi
+
+    return 0
+}
+
+# Function to execute command asynchronously
+# Args: command_line
+# Returns: PID of the background process
+execute_async() {
+    local command_line="$*"
```
chengkai15: uses eval "$*", which would re-parses arguments and loses quoting

chengkai15:
I haven't gone through everything yet — it's a fairly large change. 
 
 But based on the discussion so far, it seems we're not fully aligned on the overall approach yet. 
 
 Could we settle the framework direction (extend bsim vs. adopt Bumble) before going further?

tests/bluetooth/classic/sim/gap_discovery/Kconfig
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 NXP
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Only for use with the Bluetooth Classic GAP Discovery simulation test
+# Used to shorten test time by limiting discovery duration
+config BT_LIMITED_DISCOVERABLE_DURATION
+	default 3
+	range 1 60
```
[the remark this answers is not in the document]
lylezhu2012: The purpose is to modify the range. The type is not needed.

tests/bluetooth/classic/sim/common/common_utils.sh
```diff
@@ -0,0 +1,593 @@
...
+    while IFS= read -r test_dir; do
+        if [[ -n "${test_dir}" ]]; then
+            local resolved
+            resolved=$(resolve_test_dir "${base_dir}" "${test_dir}")
+            if [[ $? -ne 0 ]]; then
+                echo "Error: Failed to resolve test directory '${test_dir}'" >&2
+                exit 1
+            fi
+
+            if [[ $? -eq 0 && -n "${resolved}" ]]; then
```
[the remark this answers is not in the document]
lylezhu2012: Cleared the code block

tests/bluetooth/classic/sim/common/common_utils.sh
```diff
@@ -0,0 +1,593 @@
...
+                valid_dirs+=("${resolved}")
+            fi
```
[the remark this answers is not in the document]
lylezhu2012: Cleared the code block

tests/bluetooth/classic/sim/common/thread_utils.sh
```diff
@@ -0,0 +1,140 @@
...
+        echo "Error: Failed to start async process" >&2
+        return 1
+    fi
+
+    # Save PID to process-specific array
+    eval "${array_name}+=('$pid')"
+
+    echo "$pid"
+    return 0
+}
```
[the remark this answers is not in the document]
lylezhu2012: ~Ignored and unchanged.~

Fixed.

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,244 @@
...
+
+	err = bt_br_discovery_start(&br_discover_param, br_discover_result,
+				    ARRAY_SIZE(br_discover_result));
+	zassert_equal(err, 0, "Bluetooth inquiry failed (err %d)", err);
+
+	/* Wait for all discovery results to be processed */
+	timeout = GAP_DISCOVERY_TIMEOUT(ARRAY_SIZE(br_discover_result));
+	LOG_DBG("Will wait for GAP discovery done (timeout %us)", (uint32_t)timeout);
+	err = k_sem_take(&br_discover_sem, K_SECONDS(timeout));
+	zassert_equal(err, 0, "Failed to wait for discovery done (err %d)", err);
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

tests/bluetooth/classic/sim/gap_discovery/README.rst
```diff
@@ -0,0 +1,95 @@
...
+   ./tests/bluetooth/classic/sim/gap_discovery/tests_scripts/general_discovery.sh
+   ./tests/bluetooth/classic/sim/gap_discovery/tests_scripts/limited_discovery.sh
+
+The script:
+
+* allocates two TCP ports,
+* starts two Bumble controllers (one per port, with fixed BD_ADDR values),
+* copies the generated ``zephyr.exe`` into ``${BT_CLASSIC_BIN}`` using
+  ``bt_classic_${BOARD_SANITIZED}_bluetooth.classic.gap.discovery.exe`` naming,
+* launches the test binary twice:
```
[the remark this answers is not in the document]
lylezhu2012: Fixed

doc/develop/test/bluetooth_classic_sim.rst
```diff
@@ -0,0 +1,880 @@
...
+   # Create controllers with port-to-address mapping
+   # BD_ADDR[0] = "00:00:01:00:00:01" (peripheral)
+   # BD_ADDR[1] = "00:00:01:00:00:02" (central)
+   BUMBLE_CONTROLLER_PID=$(start_bumble_controllers \
+       "$BUMBLE_CONTROLLERS_LOG" \
+       "$HCI_PORT_0@${BD_ADDR[0]}" \
+       "$HCI_PORT_1@${BD_ADDR[1]}")
+
+   test_case_name="bluetooth.classic.gap.discovery"
+   extra_twister_out "${ZEPHYR_BASE}/twister-out" "${test_case_name}"
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

doc/develop/test/bluetooth_classic_sim.rst
```diff
@@ -0,0 +1,880 @@
...
+       echo "Error: Failed to start bumble controllers" >&2
+       exit 1
+   fi
+
+   # Run test applications with test suite selection
+   test_case_name="bluetooth.classic.my_test"
+   test_exe="${BT_CLASSIC_BIN}/bt_classic_${BOARD_SANITIZED}_${test_case_name}.exe"
+
+   test_case_name="bluetooth.classic.gap.discovery"
+   extra_twister_out "${ZEPHYR_BASE}/twister-out" "${test_case_name}"
```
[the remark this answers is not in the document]
lylezhu2012: Fixed

tests/bluetooth/classic/sim/common/bt_sim_controller.py
```diff
@@ -0,0 +1,928 @@
...
+        else:
+            logger.warning("[%s] Unknown inquiry PDU type: %s", self.name, type(packet))
+
+    def on_hci_inquiry_command(self, command: hci.HCI_Inquiry_Command) -> None:
+        '''
+        See Bluetooth spec 7.1.1 Inquiry command
+        '''
+        logger.debug(
+            f'Inqury LAP {command.lap} length {command.inquiry_length} '
+            f'num_responses {command.num_responses}'
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

tests/bluetooth/classic/sim/common/common_utils.sh
```diff
@@ -0,0 +1,593 @@
...
+        fi
+    fi
+}
+
+run_scripts() {
+    local config_file="$1"
+    local base_dir="$2"
+    local test_dirs
+    local valid_dirs=()
+
```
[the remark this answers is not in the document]
lylezhu2012: Fixed

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/limited_discovery.sh
```diff
@@ -0,0 +1,91 @@
...
+    exit 1
+fi
+
+test_case_name="bluetooth.classic.sim.gap.discovery"
+extra_twister_out "${ZEPHYR_BASE}/twister-out" "${test_case_name}"
+
+peripheral_exe="${BT_CLASSIC_BIN}/bt_classic_${BOARD_SANITIZED}_${test_case_name}.exe"
+ARGS=("--peer_bd_address=${BD_ADDR[1]}" \
+      " -test=gap_peripheral::test_02_gap_peripheral_limited_discovery")
+
```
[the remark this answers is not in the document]
lylezhu2012: Fixed

tests/bluetooth/classic/sim/gap_discovery/tests_scripts/limited_discovery.sh
```diff
@@ -0,0 +1,91 @@
...
+execute_async "$peripheral_exe" --bt-dev="127.0.0.1:$HCI_PORT_0" "${ARGS[@]}" || exit 1
+
+central_exe="${BT_CLASSIC_BIN}/bt_classic_${BOARD_SANITIZED}_${test_case_name}.exe"
+ARGS=("--peer_bd_address=${BD_ADDR[0]}" \
+      " -test=gap_central::test_02_gap_central_limited_discovery")
+
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

doc/develop/test/bluetooth_classic_sim.rst
```diff
@@ -0,0 +1,880 @@
...
+   peripheral_exe="${BT_CLASSIC_BIN}/bt_classic_${BOARD_SANITIZED}_${test_case_name}.exe"
+   ARGS="--peer_bd_address=${BD_ADDR[1]} -test=gap_peripheral::test_01_gap_peripheral_general_discovery"
+   execute_async "$peripheral_exe" --bt-dev="127.0.0.1:$HCI_PORT_0" "$ARGS" || exit 1
+
+   # Connect central executable to HCI_PORT_1
+   # This gives central the BD address 00:00:01:00:00:02
+   # Pass peripheral's address as peer_bd_address parameter
+   central_exe="${BT_CLASSIC_BIN}/bt_classic_${BOARD_SANITIZED}_${test_case_name}.exe"
+   ARGS="--peer_bd_address=${BD_ADDR[0]}  -test=gap_central::test_01_gap_central_general_discovery"
+   execute_async "$central_exe" --bt-dev="127.0.0.1:$HCI_PORT_1" "$ARGS" || exit 1
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
lylezhu2012: Corrected.

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
lylezhu2012: Corrected.

tests/bluetooth/classic/sim/common/common_utils.sh
lylezhu2012: Removed.

tests/bluetooth/classic/sim/common/thread_utils.sh
lylezhu2012: Fixed.

lylezhu2012:
> the overall approach yet.

It is unclear what you mean by "overall approach."

> Could we settle the framework direction (extend bsim vs. adopt Bumble) before going further?

Because Bluetooth Classic is unsupported by BubbleSim, the BSIM cannot be used. So, for Bluetooth Classic, the test framework is ZTEST (twister + harness + script) in my plan.

So, the test framework is clear. It is ZTEST (twister + harness + script). Also it is applicable `BSIM`.

tests/bluetooth/classic/sim/common/port_utils.sh
```diff
@@ -0,0 +1,430 @@
+﻿#!/usr/bin/env bash
```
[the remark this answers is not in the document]
lylezhu2012: Fixed,

tests/bluetooth/classic/sim/gap_discovery/Kconfig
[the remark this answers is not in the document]
lylezhu2012: Not accepted.

tests/bluetooth/classic/sim/gap_discovery/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+set(NO_QEMU_SERIAL_BT_SERVER 1)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(bluetooth)
+
+FILE(GLOB app_sources src/*.c)
+target_sources(app PRIVATE ${app_sources})
```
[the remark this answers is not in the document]
lylezhu2012: Ignored.

tests/bluetooth/classic/sim/common/thread_utils.sh
```diff
@@ -0,0 +1,142 @@
...
+    local pids_copy
+
+    # Get copy of PIDs array
+    eval "pids_copy=(\"\${${array_name}[@]}\")"
+
+    # Wait for all processes
+    for pid in "${pids_copy[@]}"; do
+        wait $pid || let "exit_code=$?"
+    done
+
```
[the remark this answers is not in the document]
lylezhu2012: Optimized.

tests/bluetooth/classic/sim/common/common_utils.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+# Ensure this script is only loaded once
+if [[ -n "${COMPILE_UTILS_LOADED:-}" ]]; then
+    return 0
+fi
+COMPILE_UTILS_LOADED=1
```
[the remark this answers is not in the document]
lylezhu2012: Updated.

tests/bluetooth/classic/sim/common/bumble_utils.sh
```diff
@@ -0,0 +1,166 @@
...
+    if ! "$py" -m pip --version >/dev/null 2>&1; then
+        echo "Error: pip not available for '$py'; cannot install Bumble" >&2
+        return 1
+    fi
+
+    # Install the pinned version
+    if ! "$py" -m pip install --upgrade "Bumble==$required_version"; then
+        echo "Error: Failed to install Bumble==$required_version" >&2
+        return 1
+    fi
```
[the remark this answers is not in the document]
lylezhu2012: If the test is executed by a human, this comment is fine; however, for CI, this comment is unacceptable.

tests/bluetooth/classic/sim/common/bt_sim_link.py
```diff
@@ -0,0 +1,255 @@
...
+                elif sender_controller.inquiry_mode is hci.HCI_INQUIRY_WITH_RSSI_MODE:
+                    inquiry_pdu = sim_ll.InquiryRssiInd(
+                        num_responses=1,
+                        bd_address=[c.public_address],
+                        Page_Scan_Repetition_Mode=[0x00],
+                        Reserved=[0x00],
+                        Class_Of_Device=[c.class_of_device],
+                        Clock_Offset=[0x00],
+                        RSSI=[-60],
+                    )
```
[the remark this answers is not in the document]
lylezhu2012: Updated

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,246 @@
...
+	double timeout;
+	uint64_t timeout_ms;
+
+	atomic_clear_bit(test_flags, TEST_FLAG_DEVICE_FOUND);
+
+	LOG_DBG("Starting Bluetooth inquiry");
+
+	br_discover_param.length = BR_DISCOVER_RESULT_COUNT;
+	br_discover_param.limited = limited;
+
```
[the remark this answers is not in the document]
lylezhu2012: Fixed

tests/bluetooth/classic/sim/gap_discovery/src/test_central.c
```diff
@@ -0,0 +1,246 @@
...
+	memset(br_discover_result, 0, sizeof(br_discover_result));
+
+	err = bt_br_discovery_start(&br_discover_param, br_discover_result,
+				    ARRAY_SIZE(br_discover_result));
+	zassert_equal(err, 0, "Bluetooth inquiry failed (err %d)", err);
+
+	/* Wait for all discovery results to be processed */
+	timeout = GAP_DISCOVERY_TIMEOUT(ARRAY_SIZE(br_discover_result));
+	timeout_ms = (uint64_t)(timeout * 1000.0);
+	LOG_DBG("Will wait for GAP discovery done (timeout %llu ms)", timeout_ms);
```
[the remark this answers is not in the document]
lylezhu2012: Optimized.

tests/bluetooth/classic/sim/gap_discovery/src/test_main.c
```diff
@@ -0,0 +1,55 @@
...
+static void cmd_peer_bd_address_found(char *argv, int offset)
+{
+	char *addr_str = &argv[offset];
+	int err;
+
+	err = bt_addr_from_str(addr_str, &peer_addr);
+	if (err != 0) {
+		LOG_ERR("Failed to parse peer Bluetooth address: %s (err %d)", addr_str, err);
+	}
+}
```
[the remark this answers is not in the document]
lylezhu2012: Added.

jhedberg:
@lylezhu2012 I added you back as assignee, since this doesn't touch runner details or anything else outside of the Bluetooth Classic domain.

tests/bluetooth/classic/sim/gap_discovery/src/test_main.c
```diff
@@ -0,0 +1,57 @@
...
+{
+	static struct args_struct_t args[] = {
+		{
+			.manual = false,
+			.is_mandatory = true,
+			.is_switch = false,
+			.option = "peer_bd_address",
+			.name = "XX:XX:XX:XX:XX:XX",
+			.type = 's',
+			.dest = NULL,
```
aescolar: Nit:
```suggestion
			.is_mandatory = true,
			.option = "peer_bd_address",
			.name = "XX:XX:XX:XX:XX:XX",
			.type = 's',
```
No need to set fields that are 0

tests/bluetooth/classic/sim/gap_discovery/testcase.yaml
```diff
@@ -0,0 +1,14 @@
+common:
+  tags:
+    - bluetooth
+  platform_allow:
+    - native_sim
+  harness: script
+
+tests:
+  bluetooth.classic.sim.gap.discovery:
+    timeout: 120
```
aescolar: This test took 50 seconds on this PR latest CI run, and your own example output was 115s. I think it would be better to have quite a bit more margin here, to account for more loaded runners.
```suggestion
    timeout: 300
```

jhedberg:
Heads-up: #116447 proposes a twister sidecar (per the merged #113253 framework) as the host-side orchestration for this simulation framework, and it carries the Python simulation backend from this PR — your commit, trimmed to the Python modules, with authorship and sign-off preserved and a note in the commit message explaining the trim. Please shout if you'd prefer that handled differently.

The intent is that this PR and #111840 remain the home of the actual test content: with the sidecar in place, the shell utilities and per-test scripts here would drop out, and each scenario becomes a `sidecar: bumble` declaration in `tests.yaml` with the ztest harness. See the `acl_conn` reference test in #116447 for the shape.

PR #100493, opened 2025-12-04
boards: st: nucleo_h563zi: fix PLL clocks

mariopaja:
This change aligns max allowed SPI clock with the datasheet.

SPI1 on H5xx is using PLL1Q for their clock selection
https://github.com/zephyrproject-rtos/zephyr/blob/64a1477976b62606e8844a4bc8de880117a1cc47/dts/arm/st/h5/stm32h5.dtsi#L510

Based on [STM32H562/3xx Datasheet](https://www.st.com/resource/ja/datasheet/stm32h563ri.pdf), the  maximum allowed clock freq for SPI is 125MHz.
f_spi

By changing PLL1Q div to 4, SPI clock drops to 120MHz (just below the max allowed frequency)

The new proposed PLL clocks:
1. PLL1 will be used by SPI
2. PLL2 for FDCAN (as discussed) because clock freq should remain as is
3. PLL3 will be used in the _samples/i2s/output_ for the SAI

h563_pll

### DIFF
--- a/boards/st/nucleo_h563zi/nucleo_h563zi-common.dtsi
+++ b/boards/st/nucleo_h563zi/nucleo_h563zi-common.dtsi
@@ -71,6 +71,16 @@
 	div-m = <2>;
 	mul-n = <120>;
 	div-p = <2>;
+	div-q = <4>;
+	div-r = <2>;
+	clocks = <&clk_hse>;
+	status = "okay";
+};
+
+&pll2 {
+	div-m = <2>;
+	mul-n = <120>;
+	div-p = <2>;
 	div-q = <3>;
 	div-r = <2>;
 	clocks = <&clk_hse>;
@@ -171,7 +181,7 @@
 	pinctrl-0 = <&fdcan1_rx_pd0 &fdcan1_tx_pd1>;
 	pinctrl-names = "default";
 	clocks = <&rcc STM32_CLOCK(APB1_2, 9)>,
-		 <&rcc STM32_SRC_PLL1_Q FDCAN_SEL(1)>;
+		 <&rcc STM32_SRC_PLL2_Q FDCAN_SEL(2)>;
 	clk-divider = <2>;
 	status = "okay";
 };
--- a/dts/arm/st/h5/stm32h562.dtsi
+++ b/dts/arm/st/h5/stm32h562.dtsi
@@ -128,7 +128,7 @@
 			#size-cells = <0>;
 			reg = <0x40015404 0x20>;
 			clocks = <&rcc STM32_CLOCK(APB2, 21)>,
-				 <&rcc STM32_SRC_PLL2_P SAI1_SEL(0)>;
+				 <&rcc STM32_SRC_PLL1_P SAI1_SEL(0)>;
 			dmas = <&gpdma1 1 53 (STM32_DMA_MODE_NORMAL | STM32_DMA_PRIORITY_HIGH |
 				STM32_DMA_16BITS)>;
 			status = "disabled";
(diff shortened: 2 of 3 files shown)


mariopaja:
This issue https://github.com/zephyrproject-rtos/zephyr/issues/91103 could have been related to this. (just speculating not tested)

erwango:
Thanks for the fix @mariopaja.
I can see that FDCAN clock is also using PLL1_Q output. Would you mind checking this change won't have negative impact on that front?

boards/st/nucleo_h563zi/nucleo_h563zi-common.dtsi
```diff
@@ -171,7 +181,7 @@
 	pinctrl-0 = <&fdcan1_rx_pd0 &fdcan1_tx_pd1>;
 	pinctrl-names = "default";
 	clocks = <&rcc STM32_CLOCK(APB1_2, 9)>,
-		 <&rcc STM32_SRC_PLL1_Q FDCAN_SEL(1)>;
+		 <&rcc STM32_SRC_PLL2_Q FDCAN_SEL(1)>;
```
gautierg-st: It should be FDCAN_SEL(2) for PLL2_Q

dts/arm/st/h5/stm32h562.dtsi
```diff
@@ -128,7 +128,7 @@
 			#size-cells = <0>;
 			reg = <0x40015404 0x20>;
 			clocks = <&rcc STM32_CLOCK(APB2, 21)>,
-				 <&rcc STM32_SRC_PLL2_P SAI1_SEL(0)>;
+				 <&rcc STM32_SRC_PLL3_P SAI1_SEL(1)>;
```
gautierg-st: For PLL3_P, it should be SAI1_SEL(2).
That said, I'm not sure we should set that in the dtsi file. Maybe set it to the default register value `<&rcc STM32_SRC_PLL1_P SAI1_SEL(0)>` in dtsi, and set it to PLL3_P in the board dts file.

gautierg-st (changes requested):

gautierg-st (changes requested):
Thanks. Though could you also update `stm32h562.dtsi` while you're at it ?
Both sai nodes indicates `<&rcc STM32_SRC_PLL2_P SAI1_SEL(0)>` but it should be `<&rcc STM32_SRC_PLL1_P SAI1_SEL(0)>`

mariopaja:
> Thanks. Though could you also update `stm32h562.dtsi` while you're at it ? Both sai nodes indicates `<&rcc STM32_SRC_PLL2_P SAI1_SEL(0)>` but it should be `<&rcc STM32_SRC_PLL1_P SAI1_SEL(0)>`

I updated it, but maybe the sai nodes of all platforms should be updated to the default value. Atm I dont know if they use the default value, but it would have to be checked and updated accordingly?

gautierg-st (approved):

henrikbrixandersen (approved):

dts/arm/st/h5/stm32h562.dtsi
```diff
@@ -128,7 +128,7 @@
 			#size-cells = <0>;
 			reg = <0x40015404 0x20>;
 			clocks = <&rcc STM32_CLOCK(APB2, 21)>,
-				 <&rcc STM32_SRC_PLL2_P SAI1_SEL(0)>;
```
erwango, line 131: .dtsi change should be done in a dedicated commit

erwango (changes requested):
Otherwise LGTM

dts/arm/st/h5/stm32h562.dtsi
mariopaja: Should I leave it as is and check all SAI nodes in another PR?

erwango: As you wish, but ok for me to keep in current PR.

gautierg-st (approved):

henrikbrixandersen (approved):

erwango (approved):

Issue #44454, opened 2022-04-01
Bluetooth: GATT: The (mis)use of `ssize_t` for GATT callbacks


Thalley:
**Is your enhancement proposal related to a problem? Please describe.**
The return type `ssize_t` is used multiple places in the GATT API where the functions may return a value >= denoting bytes read/written, and a negative value in case of error. 

The definition of the POSIX type `ssize_t` is:
> Used for a count of bytes or an error indication. According to POSIX, it shall be a signed integer type capable of storing values at least in the range [-1, SSIZE_MAX], and the implementation shall support one or more programming environments where the width of ssize_t is no greater than the width of the type long.

The positive value range is not much of a concern, but the GATT API often defines the negative return value range to be
```
BT_GATT_ERR() with a specific ATT error code.
```

The ATT error code range is 0x00 to 0xFF, and is made negative by the use of `BT_GATT_ERR`, so the return value range of the GATT API functions is thus [-255, SSIZE_MAX], which is of course outside the defined value range of the POSIX standard. 

**Describe the solution you'd like**
This issue has been created to create a discussion about this, and to determine if any change is needed. 

**Describe alternatives you've considered**

**Additional context**

Total definition from the man page: 
```
ssize_t
              Include: <sys/types.h>.  Alternatively, <aio.h>,
              <monetary.h>, <mqueue.h>, <stdio.h>, <sys/msg.h>,
              <sys/socket.h>, <sys/uio.h>, or <unistd.h>.

              Used for a count of bytes or an error indication.
              According to POSIX, it shall be a signed integer type
              capable of storing values at least in the range [-1,
              SSIZE_MAX], and the implementation shall support one or
              more programming environments where the width of ssize_t
              is no greater than the width of the type long.

              Glibc and most other implementations provide a length
              modifier for ssize_t for the [printf(3)](https://man7.org/linux/man-pages/man3/printf.3.html) and the [scanf(3)](https://man7.org/linux/man-pages/man3/scanf.3.html)
              families of functions, which is z; resulting commonly in
              %zd or %zi for printing ssize_t values.  Although z works
              for ssize_t on most implementations, portable POSIX
              programs should avoid using it—for example, by converting
              the value to intmax_t and using its length modifier (j).

              Conforming to: POSIX.1-2001 and later.

[...]

PR #100495, opened 2025-12-04
tests: bluetooth: audio: ascs: fix build issues

biwenli:
Fix build issues,
- error: format %x expects argument of type unsigned int, but argument 7 has type ssize_t {aka long int} [-Werror=format=] zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02x)" , BT_GATT_ERR(ret));

### DIFF
--- a/tests/bluetooth/audio/ascs/src/main.c
+++ b/tests/bluetooth/audio/ascs/src/main.c
@@ -164,7 +164,8 @@ ZTEST_F(ascs_test_suite, test_sink_ase_read_state_idle)
 	zexpect_not_null(fixture->ase_snk.attr);
 
 	ret = ase->read(conn, ase, &hdr, sizeof(hdr), 0);
-	zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02x)", BT_GATT_ERR(ret));
+	zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02x)",
+		      (uint8_t)BT_GATT_ERR(ret));
 	zassert_equal(0x00, hdr.ase_state, "unexpected ASE_State 0x%02x", hdr.ase_state);
 }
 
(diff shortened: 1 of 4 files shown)


tests/bluetooth/audio/ascs/src/main.c
```diff
@@ -164,7 +164,8 @@ ZTEST_F(ascs_test_suite, test_sink_ase_read_state_idle)
 	zexpect_not_null(fixture->ase_snk.attr);
 
 	ret = ase->read(conn, ase, &hdr, sizeof(hdr), 0);
-	zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02x)", BT_GATT_ERR(ret));
+	zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02lx)",
+		      BT_GATT_ERR(ret));
```
Thalley: Thanks for fixing this :) I'm wondering if `lx` is the right format though; `ssize_t` is not a unsigned value, so `lx` should give the same warning as `x`, shouldn't it? 

One alternative would have been to do something like

```suggestion
	zassert_false(ret < 0, "attr->read returned unexpected (err 0x%02x)",
		      (uint8_t)BT_GATT_ERR(ret));
```

i.e. cast the value to an uint8_t, as ATT errors are only 8 bit (which is also evident from the `0x%02X`).

biwenli: > Thanks for fixing this :) I'm wondering if `lx` is the right format though; `ssize_t` is not a unsigned value, so `lx` should give the same warning as `x`, shouldn't it?
> 
> One alternative would have been to do something like
> 
> i.e. cast the value to an uint8_t, as ATT errors are only 8 bit (which is also evident from the `0x%02X`).

Correct. Thanks.

Thalley (approved):

Thalley:
Hmm, this is less than trivial. 

> @Thalley @biwenli wouldn't it make more sense to fix the `BT_GATT_ERR()` macro to do the explicit typecast? 

What type should `BT_GATT_ERR` typecast to? The GATT API does indeed define it as `ssize_t`, but we treat it as an `int`. (see https://github.com/zephyrproject-rtos/zephyr/issues/44454)

> From an API perspective I don't think it's ideal that an API that's named like that returns `ssize_t`. 

Not sure I follow this. `BT_GATT_ERR` does not return any specific type, and the code here is not using an API but rather internal function calls as part of a unit test. 

> The way the macro is defined, the caller should already have verified that the passed input is negative, so doing the explicit typecase should IMO be fine as part of the macro definition.

`BT_GATT_ERR` is designed to handle positive error values (`BT_ATT_ERR_*`) and make the positive value a negative value for error handling. 

I'm not opposed to modify `BT_GATT_ERR`, but that could easily end up being an a breaking API change, and I think it's worth fixing this build error without fixing the macro.

kartben (approved):

PR #100500, opened 2025-12-04
Enable DTCM and ITCM of core R8 on Renesas RZ/V2H

nhutnguyenkc:
In the previous implementation, the sram node was treated as a single large RAM block that included ITCM, DTCM, and general SRAM, starting at address 0x0 with a total size of 256 MB.

However, Arm Cortex-R8 MPCore Processor r0p2 Technical Reference Manual, DTCM cannot store executable instructions. If an application grows large enough, the linker may place instructions in the DTCM area, which begins at address 0x20000, leading to invalid execution.

This PR updates the memory layout to use the sram3 node and relocates the vector table into the ITCM region. The entire DTCM region is now left unused by default so that applications may choose how to use it explicitly.

The MPU configuration has also been updated to align with the new memory layout.


binhnguyen2434 (approved):
LGTM!

tiennguyenzg (approved):

henrikbrixandersen (approved):

PR #100501, opened 2025-12-04
soc: it8xxx2: block idle if timer is close to expiration

Chenhongren:
If the event timer or free-run timer is close to expiration when system enters idle with I2C target DMA mode enabled, the memory and CPU clocks may become unsynchronized after wakeup. This causes CPU to fetch incorrect data and eventually trigger SoC watchdog timeout.

Due to this hardware limitation, SoC should skip entering idle mode if the remaining timer value is less than 150µs(safe margin).

### DIFF
--- a/drivers/timer/ite_it8xxx2_timer.c
+++ b/drivers/timer/ite_it8xxx2_timer.c
@@ -73,6 +73,13 @@ const int32_t z_sys_timer_irq_for_test = DT_IRQ_BY_IDX(DT_NODELABEL(timer), 5, i
 #define EVEN_TIMER_MAX_CNT_SYS_TICK	(EVENT_TIMER_MAX_CNT \
 					/ HW_CNT_PER_SYS_TICK)
 
+/* Timer tick threshold to prevent SoC from entering idle mode.
+ * Calculated as 150µs converted to timer ticks using the formula:
+ *   ticks = us * timer_clk_src / 1000000
+ * where (event/free run timers)timer_clk_src is fixed at 32768Hz
+ */
+#define IDLE_BLOCK_TIMER_TICKS DIV_ROUND_UP(150 * CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC, 1000000)
+
 static struct k_spinlock lock;
 /* Last HW count that we called sys_clock_announce() */
 static volatile uint32_t last_announced_hw_cnt;
@@ -388,6 +395,12 @@ static int timer_init(enum ext_timer_idx ext_timer,
 	}
 
 	return 0;
+}
+
+bool ite_ec_timer_block_idle(void)
+{
+	return (IT8XXX2_EXT_CNTOX(EVENT_TIMER) < IDLE_BLOCK_TIMER_TICKS) ||
+	       (IT8XXX2_EXT_CNTOX(FREE_RUN_TIMER) < IDLE_BLOCK_TIMER_TICKS);
 }
 
 static int sys_clock_driver_init(void)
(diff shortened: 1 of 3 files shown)


soc/ite/ec/common/soc_timer.h
```diff
@@ -0,0 +1,31 @@
...
+#else
+#define timer_block_idle
+#endif /* CONFIG_SOC_IT8XXX2 */
+
+/**
+ * @brief check if the SoC should be prevented from entering idle mode
+ *
+ * @return true if idle mode should be skipped, false otherwise
+ */
+bool timer_block_idle(void);
```
fabiobaltieri: hey since this is an ite ec specific function, how about just calling it `ite_ec_timer_block_idle` and then here instead of jumping around with the macro renaming the function you could either do the forward declaration or do an empty inline if you are not on `CONFIG_SOC_IT8XXX2`. Would make things a bit simpler I think.

Chenhongren: The change has been updated. Please help review it again, thanks.

fabiobaltieri (dismissed):

Dino-Li (dismissed):

drivers/timer/ite_it8xxx2_timer.c
```diff
@@ -73,6 +73,13 @@ const int32_t z_sys_timer_irq_for_test = DT_IRQ_BY_IDX(DT_NODELABEL(timer), 5, i
 #define EVEN_TIMER_MAX_CNT_SYS_TICK	(EVENT_TIMER_MAX_CNT \
 					/ HW_CNT_PER_SYS_TICK)
 
+/* Timer tick threshold to prevent SoC from entering idle mode.
+ * Calculated as 150µs converted to timer ticks using the formula:
+ *   ticks = us * timer_clk_src / 1000000
+ * where (event/free run timers)timer_clk_src is fixed at 32768Hz
+ */
+#define IDLE_BLOCK_TIMER_TICKS DIV_ROUND_UP(150 * 32768, 1000000)
```
keith-zephyr: Should this use `CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC`?
```suggestion
#define IDLE_BLOCK_TIMER_TICKS DIV_ROUND_UP(150 * CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC, 1000000)
```

Chenhongren: We prefer to retain this change, as the Kconfig setting may differ between projects, whereas the event and free-run timers use fixed 32768Hz clock source defined in the timer driver.
https://github.com/zephyrproject-rtos/zephyr/blob/8053d7793f13a89bc660381a8eae5e8172e79c7e/drivers/timer/ite_it8xxx2_timer.c#L402
https://github.com/zephyrproject-rtos/zephyr/blob/8053d7793f13a89bc660381a8eae5e8172e79c7e/drivers/timer/ite_it8xxx2_timer.c#L414

Chenhongren: After investigation, we confirmed that CONFIG_SYS_CLOCK_HW_CYCLES_PER_SEC is fixed at 32768Hz with build-time check in place. 
https://github.com/zephyrproject-rtos/zephyr/blob/8053d7793f13a89bc660381a8eae5e8172e79c7e/drivers/timer/ite_it8xxx2_timer.c#L21
Therefore, we can safely use this Kconfig option to replace hardcoded value.

@fabiobaltieri @Dino-Li @keith-zephyr please help review again, thanks.

Dino-Li (approved):

GTLin08 (approved):

PR #100509, opened 2025-12-04
MAINTAINERS: Include native drivers host side in native area

aescolar:
Let's include the "bottom" side of native drivers in the native_sim area for the drivers which were missing. So they are covered from the "platform" side of things by this area.

The regex picks .c and .h files, in the drivers and subsys folders which have "bottom" at the end of their name.

Before we were already covering drivers which had "posix" or "native" anywhere in their name, but that did not include all which are not named following that pattern.
(i.e. userchan, linux_evdev, sdl_touch, display_sdl, gpio_emul_sdl, fuse_fs_access)

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -3775,6 +3775,8 @@ Native_sim and POSIX arch:
     - scripts/valgrind.supp
     - soc/native/
     - tests/boards/native_sim/
+  files-regex:
+    - (drivers/|subsys/).*(bottom)\.([c|h])
   labels:
     - "area: native port"
   description: >-


JarmouniA (approved):

MAINTAINERS.yml
```diff
@@ -3775,6 +3775,8 @@ Native_sim and POSIX arch:
     - scripts/valgrind.supp
     - soc/native/
     - tests/boards/native_sim/
+  files-regex:
+    - (drivers/|subsys/).*(bottom)\.([c|h])
```
nashif, line 3779: something like `drivers/input/linux_evdev_bottom.c` is now part of the input subsystem, also contributed and maintained within this subsystem (by @fabiobaltieri ), adding this here, creates a duplication and ambiguity about who actually maintains that file. Do we really want this? Does this really belong into POSIX arch?

aescolar: The idea with this change is to cover it from both sides (platform and subsystem) like we do with quite a few other drivers. With the understanding that it is better to have more eyes in reviews than less.
I'm not going to be blocking PRs to their drivers classes with nits or comments on the driver API :).

Anyhow I added all the respective subsystem maintainers to this PR in case they disagree.

nashif: see https://github.com/zephyrproject-rtos/zephyr/pull/100435 for something similar being done.

PR #100512, opened 2025-12-04
Flash simulator callbacks

rkel:
Implement a functionality to set write and erase callbacks that can modify the simulation on the fly when the operations are in progress. This allows simulation for modules that takes care about data integrity and should react when anything breaks when accessing FLASH.

I did not implement read operation callback as when reading we can usually modify the memory content before calling a module operation that uses it.


drivers/flash/flash_simulator.c
```diff
@@ -498,6 +551,26 @@ void *z_impl_flash_simulator_get_memory(const struct device *dev,
...
+#endif
+}
+
+#ifdef CONFIG_FLASH_SIMULATOR_CALLBACKS
+void z_impl_flash_simulator_set_callbacks(const struct device *dev,
+					  const struct flash_simulator_cb *cb)
+{
+	ARG_UNUSED(dev);
+	flash_simulator_cbs = cb;
+}
```
de-nordic, line 571: seems like this can return flash parameters for any dev it gets. We should address that in the future.

de-nordic (approved):

drivers/flash/flash_simulator.c
rkel: Yes, but it is exactly the same issue with z_impl_flash_simulator_get_memory. The case is that the driver is not a multi-instance implementation. From the drivers itself it seems really easy to be implemented as a multi-instance driver, but the STATS may be problematic - I am not sure. I had just an idea to quickly reimplement it as multi-instance driver,  but I had stopped looking at all the STATS_SECT_xxx macros and I admit I did not have possibility to go deeper into it.

tpambor (approved):

rghaddab (approved):

Issue #27805, opened 2020-08-26
samples without README


pabigot:
Taking #27130 to the next step: The following samples do not have a README:

* [ ] samples/boards/esp32/ethernet
* [ ] samples/boards/nrf/nrf53_sync_rtc/net
* [ ] samples/drivers/lcd_cyclonev_socdk
* [ ] samples/drivers/led_apa102c_bitbang
* [ ] samples/drivers/spi_fujitsu_fram
* [ ] samples/sensor/bmg160
* [ ] samples/sensor/max44009 (has a README)
* [ ] samples/sensor/sx9500  (has a README.txt)
* [ ] samples/shields/npm6001_ek
* [ ] samples/subsys/ipc/ipc_service/static_vrings/remote
* [ ] samples/subsys/logging/multidomain
* [ ] samples/subsys/mgmt/osdp/control_panel
* [ ] samples/subsys/mgmt/osdp/peripheral_device
* [ ] samples/subsys/pm/device_pm
* [ ] samples/subsys/pm/latency
* [ ] samples/subsys/rtio/sensor_batch_processing
* [ ] samples/subsys/shell/devmem_load
* [x] samples/bluetooth/mtu_update/central
* [x] samples/bluetooth/mtu_update/peripheral
* [x] samples/drivers/ipm/ipm_mcux/remote
* [x] samples/drivers/lora/receive
* [x] samples/drivers/lora/send
* [x] samples/drivers/mbox/remote
* [x] samples/drivers/spi_flash (pr #27805)
* [x] samples/drivers/watchdog
* [x] samples/modules/thrift/hello/client (in hello/)
* [x] samples/modules/thrift/hello/server (in hello/)
* [x] samples/net/mdns_responder (pr #27866)
* [x] samples/posix/eventfd
* [x] samples/sensor/fxos8700-hid
* [x] samples/sensor/grove_light
* [x] samples/sensor/grove_temperature
* [x] samples/sensor/thermometer (has a README.txt)
* [x] samples/subsys/console/echo
* [x] samples/subsys/fs/format
* [x] samples/subsys/shell/shell_module
* [x] samples/subsys/tracing (has a README.txt)

pabigot:
```
find samples/ -name sample.yaml \
| sed -e 's@/[^/]*$@@' \
| while read dir ; do \
  if $(cd $dir && ls | grep -i '^readme' > /dev/null) ; then \
    :  ; \
  else  \
[...]

PR #100515, opened 2025-12-04
samples: drivers: led: add README for APA102C bit-bang sample

Med-El-Amrani:
Add comprehensive README.rst documentation for the APA102C LED bit-bang sample. The README includes:

- Overview of the sample functionality
- Hardware setup and wiring instructions
- Configuration options
- Building and running instructions
- Safety warning about LED brightness

Fixes #27805

### DIFF
--- a/samples/drivers/led/apa102c_bitbang/README.rst
+++ b/samples/drivers/led/apa102c_bitbang/README.rst
@@ -1,0 +1,89 @@
+.. zephyr:code-sample:: apa102c-bitbang
+   :name: APA102C LED (bit-bang)
+   :relevant-api: gpio_interface
+
+   Control APA102C RGB LEDs using GPIO bit-banging.
+
+Overview
+********
+
+This sample demonstrates how to control APA102C RGB LEDs using GPIO bit-banging
+(manual software control of GPIO pins). The sample cycles through four colors:
+red, green, blue, and white.
+
+The APA102C is a smart RGB LED that uses a simple two-wire interface (data and
+clock). This sample uses the :ref:`GPIO API <gpio_api>` to manually control
+these signals.
+
+.. warning::
+
+   APA102C LEDs are extremely bright even at low settings. To protect your eyes,
+   do not look directly at the LEDs. This sample uses very low brightness
+   settings for safety.
+
+Hardware Setup
+**************
+
+The APA102C requires 5V data and clock signals. You will need:
+
+- One or more APA102C LEDs
+- A logic level shifter to convert 3.3V signals to 5V
+
+.. warning::
+
+   Do not connect 3.3V GPIO outputs directly to 5V device inputs unless your
+   board's GPIO pins are explicitly 5V tolerant. Consult your board's datasheet.
+
+
+Wiring
+======
+
+By default, the sample uses the following GPIO pins:
+
+- **Data pin**: GPIO 16
+- **Clock pin**: GPIO 19
+
+Connect these pins to the APA102C LED:
+
+- Data pin → DI (Data Input) on LED
+- Clock pin → CI (Clock Input) on LED
+- Also connect power (5V) and ground
+
+Configuration
+*************
+
+The sample can be configured by modifying ``src/main.c``:
+
+- ``GPIO_DATA_PIN``: Change the data pin number
+- ``GPIO_CLK_PIN``: Change the clock pin number
+- ``NUM_LEDS``: Set the number of LEDs in your chain
+- ``SLEEPTIME``: Adjust the color change interval
+
+Requirements
+************
+
+Your board must:
+
+#. Have GPIO support
+#. Have the ``gpio-0`` devicetree alias defined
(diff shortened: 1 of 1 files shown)


samples/drivers/led/apa102c_bitbang/README.rst
```diff
@@ -0,0 +1,84 @@
...
+- Logic level shifter (preferred) or pull-up resistors to convert 3.3V to 5V
+- Make sure the GPIO pins are 5V tolerant if using pull-up resistors
+
+Wiring
+======
+
+By default, the sample uses the following GPIO pins:
+
+- **Data pin**: GPIO 16
+- **Clock pin**: GPIO 19
```
simonguinot, line 44: Hi @Med-El-Amrani,

Thanks for this patch.

While you are at it, maybe you could introduce some Kconfig options to let users configure the GPIO numbers ?

samples/drivers/led/apa102c_bitbang/README.rst
```diff
@@ -0,0 +1,84 @@
...
+   do not look directly at the LEDs. This sample uses very low brightness
+   settings for safety.
+
+Hardware Setup
+**************
+
+The APA102C requires 5V data and clock signals. You will need:
+
+- One or more APA102C LEDs
+- Logic level shifter (preferred) or pull-up resistors to convert 3.3V to 5V
```
fabiobaltieri: how does a pullup convert 3.3V to 5V?

samples/drivers/led/apa102c_bitbang/README.rst
Med-El-Amrani: Hi @simonguinot,
Thank you for the suggestion! This is my first contribution to Zephyr, and I appreciate your guidance.
I'd be happy to add the Kconfig options. Could you confirm if this approach would be acceptable:

1. Add a `Kconfig` file with `CONFIG_APA102C_DATA_PIN` and `CONFIG_APA102C_CLK_PIN` options
2. Update `src/main.c` to use these config options instead of hard-coded values
3. Update `README.rst` to document the new configuration options

I'll implement this and update the PR once I get confirmation. Thanks for your help!

samples/drivers/led/apa102c_bitbang/README.rst
Med-El-Amrani: Hi @fabiobaltieri,

Thank you for pointing this out! 
I copied that wording from the original source code comments, which mention 
"pull-up resistors" as an option. However, I realize now that this is either 
imprecise or refers to a very specific case (open-drain GPIO configuration 
with 5V-tolerant pins and external pull-ups to 5V).

You're right that this is confusing and not the recommended approach for most users. 
A logic level shifter is indeed the proper and safer solution.

I'll update the README. Thank you for the correction!

fabiobaltieri:
> Eventually, this is of some interest for demonstrating how to use the device on a board where the SPI protocol is not available?

Well but it's demonstrating in the wrong way right? The right one would be with the existing drivers.

samples/drivers/led/apa102c_bitbang/README.rst
thedjnK: > While you are at it, maybe you could introduce some Kconfig options to let users configure the GPIO numbers ?

DTS for pins, not Kconfig - thanks!

samples/drivers/led/apa102c_bitbang/README.rst
```diff
@@ -0,0 +1,89 @@
...
+Connect these pins to the APA102C LED:
+
+- Data pin → DI (Data Input) on LED
+- Clock pin → CI (Clock Input) on LED
+- Also connect power (5V) and ground
+
+Configuration
+*************
+
+The sample can be configured by modifying ``src/main.c``:
```
thedjnK, line 55:
```
The sample can be configured by modifying :file:`src/main.c`:
```

samples/drivers/led/apa102c_bitbang/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+- ``GPIO_DATA_PIN``: Change the data pin number
+- ``GPIO_CLK_PIN``: Change the clock pin number
+- ``NUM_LEDS``: Set the number of LEDs in your chain
+- ``SLEEPTIME``: Adjust the color change interval
```
thedjnK, line 60: actually I'm really confused by this, the configuration should be done in dts, it should not be done in an application .c file, and it shouldn't even be possible... Not that this is caused by you or this PR but the sample or driver looks flat out wrong, where is the driver in this? How is it reusable? It's just calling GPIO functions with a wait period @simonguinot

fabiobaltieri: I think this sample should go.

simonguinot: > I think this sample should go.

Yeah you are right @fabiobaltieri.

@Med-El-Amrani please can you take care of removing this sample ? Thanks in advance.

PR #100518, opened 2025-12-04
soc: nordic: nrf54l: Remove selecting a DTS-type option in Kconfig

grochu:
Remove selecting `CONFIG_HAS_HW_NRF_RADIO_IEEE802154` that should be generated from DTS based on the SoC selection.


soc/nordic/nrf54l/Kconfig
```diff
@@ -16,7 +16,6 @@ config SOC_NRF54L_CPUAPP_COMMON
 	select CPU_CORTEX_M33
 	select CPU_CORTEX_M_HAS_DWT
 	select CPU_HAS_ICACHE
-	select HAS_HW_NRF_RADIO_IEEE802154
```
mstasiaknordic, line 19: Shouldn't specific nRF54L SoCs symbols, e.g. `SOC_NRF54L05_CPUAPP`, select it then? Or is it going to be completely removed from any nRF54L? Or do there targets already have the required nodes in DTS for the symbol to be generated from?

grochu: It shouldn't be selected from Kconfig directly as it depends on a DTS property. The dependency is defined here: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/nordic/common/Kconfig.peripherals#L145

mstasiaknordic (approved):

bjarki-andreasen (approved):

nordic-piks (approved):

vlad-kulikov (approved):

PR #100519, opened 2025-12-04
drivers: usb: udc: stm32: move power configuration to common code

mathieuchopstm:
Create infrastructure for shared USB common code on STM32 family, and move the Power Controller configuration logic to common code.

The power configuration is left unchanged for now, but it can be further cleaned up and improved in the future.

### DIFF
--- a/drivers/usb/common/stm32/Kconfig
+++ b/drivers/usb/common/stm32/Kconfig
@@ -1,0 +1,16 @@
+# Copyright (c) 2025 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+
+if SOC_FAMILY_STM32
+
+config STM32_USB_COMMON
+	bool
+	help
+	  Enable compilation of STM32 USB common code.
+	  This option is selected by the USB drivers when appropriate.
+
+module = STM32_USB_COMMON
+module-str = STM32 USB common code
+source "subsys/logging/Kconfig.template.log_config"
+
+endif  # SOC_FAMILY_STM32
--- a/drivers/usb/common/stm32/stm32_usb_common.h
+++ b/drivers/usb/common/stm32/stm32_usb_common.h
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2025 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_
+#define ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_
+
+/**
+ * @brief Configures the Power Controller as necessary
+ * for proper operation of the USB controllers
+ *
+ * @returns 0 on success, negative error code otherwise.
+ */
+int stm32_usb_pwr_enable(void);
+
+#endif /* ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_ */
--- a/drivers/usb/common/stm32/stm32_usb_common.h
+++ b/drivers/usb/common/stm32/stm32_usb_common.h
@@ -15,4 +15,13 @@
  */
 int stm32_usb_pwr_enable(void);
 
+/**
+ * @brief Configures the Power Controller to disable
+ * USB-related regulators/etc if no controller is
+ * still active (refcounted).
+ *
+ * @returns 0 on success, negative error code otherwise.
+ */
+int stm32_usb_pwr_disable(void);
+
 #endif /* ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_ */
--- a/drivers/usb/common/stm32/stm32_usb_pwr.c
+++ b/drivers/usb/common/stm32/stm32_usb_pwr.c
@@ -131,3 +131,73 @@ fini:
 
 	return err;
 }
+
+int stm32_usb_pwr_disable(void)
+{
+	uint32_t new_count;
+	int err;
+
+	err = k_sem_take(&pwr_refcount_mutex, K_FOREVER);
+	if (err != 0) {
+		return err;
+	}
+
+	new_count = --usb_pwr_refcount;
+	if (new_count > 0) {
+		/* There are other users - don't disable now */
+		err = 0;
+		goto fini;
+	}
+
+#if defined(CONFIG_SOC_SERIES_STM32H7X)
+	LL_PWR_DisableUSBVoltageDetector();
+#elif defined(CONFIG_SOC_SERIES_STM32U5X)
+# if DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs)
+	LL_PWR_DisableUSBEPODBooster();
+	while (LL_PWR_IsActiveFlag_USBBOOST() != 0) {
+		/* Wait for USB EPOD BOOST off */
+	}
+
+	LL_PWR_DisableUSBPowerSupply();
+# endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs) */
+
+	LL_PWR_DisableVddUSB();
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* Disable Vdd33USB voltage monitoring */
+	LL_PWR_DisableVddUSBMonitoring();
+
+	/* Disable VDDUSB */
+	LL_PWR_DisableVddUSB();
+#elif defined(CONFIG_SOC_SERIES_STM32WBAX)
+	/* Disable USB OTG booster */
+	LL_PWR_DisableUSBBooster();
+
+	while (LL_PWR_IsActiveFlag_USBBOOSTRDY()) {
+		/* Wait until USB OTG booster is off */
+	}
+
+	/* Disable USB OTG internal power */
+	LL_PWR_DisableUSBPWR();
+
+	/* Disable VDD11USB */
+	LL_PWR_DisableVdd11USB();
+
+	while (LL_PWR_IsActiveFlag_VDD11USBRDY()) {
+		/* Wait until VDD11USB supply is off */
+	}
+
+	/* Enable VDDUSB power isolation */
+	LL_PWR_DisableVddUSB();
+#elif defined(PWR_USBSCR_USB33SV) || defined(PWR_SVMCR_USV)
+	/* Enable VDDUSB power isolation */
+	LL_PWR_DisableVDDUSB();
+#elif defined(PWR_CR2_USV)
+	/* Enable VDDUSB power isolation */
+	LL_PWR_DisableVddUSB();
+#endif
+
+fini:
+	k_sem_give(&pwr_refcount_mutex);
+
+	return err;
+}
--- a/drivers/usb/common/stm32/stm32_usb_pwr.c
+++ b/drivers/usb/common/stm32/stm32_usb_pwr.c
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2025 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <soc.h>
+#include <stm32_ll_bus.h>
+#include <stm32_ll_pwr.h>
+#include <stm32_ll_rcc.h>
+#include <stm32_ll_system.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/util.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(stm32_usb_pwr, CONFIG_STM32_USB_COMMON_LOG_LEVEL);
+
+/*
+ * Keep track of whether power is already
+ * enabled here to simplify the USB drivers.
+ */
+static K_SEM_DEFINE(pwr_refcount_mutex, 1, 1);
+static uint32_t usb_pwr_refcount;
+
+int stm32_usb_pwr_enable(void)
+{
+	uint32_t old_count;
+	int err;
+
+	err = k_sem_take(&pwr_refcount_mutex, K_FOREVER);
+	if (err != 0) {
+		return err;
+	}
+
+	old_count = usb_pwr_refcount++;
+	if (old_count > 0) {
+		/* Already enabled - nothing to do */
+		err = 0;
+		goto fini;
+	}
+
+#if defined(CONFIG_SOC_SERIES_STM32H7X)
+	LL_PWR_EnableUSBVoltageDetector();
+
+	/* Per AN2606: USBREGEN not supported when running in FS mode. */
+	LL_PWR_DisableUSBReg();
+	while (!LL_PWR_IsActiveFlag_USB()) {
+		LOG_INF("PWR not active yet");
+		k_msleep(100);
+	}
+#elif defined(CONFIG_SOC_SERIES_STM32U5X)
+	__ASSERT_NO_MSG(LL_AHB3_GRP1_IsEnabledClock(LL_AHB3_GRP1_PERIPH_PWR));
+
+	/* Check that power range is 1 or 2 */
+	if (LL_PWR_GetRegulVoltageScaling() < LL_PWR_REGU_VOLTAGE_SCALE2) {
+		LOG_ERR("Wrong Power range to use USB OTG HS");
+		err = -EIO;
+		goto fini;
+	}
+
+	LL_PWR_EnableVddUSB();
+
+# if DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs)
+	/* Enable HS PHY power supply */
+	LL_PWR_EnableUSBPowerSupply();
+	LL_PWR_EnableUSBEPODBooster();
+	while (LL_PWR_IsActiveFlag_USBBOOST() != 1) {
+		/* Wait for USB EPOD BOOST ready */
+	}
+# endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs) */
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* Enable Vdd33USB voltage monitoring */
+	LL_PWR_EnableVddUSBMonitoring();
+	while (!LL_PWR_IsActiveFlag_USB33RDY()) {
+		/* Wait for Vdd33USB ready */
+	}
+
+	/* Enable VDDUSB */
+	LL_PWR_EnableVddUSB();
+#elif defined(CONFIG_SOC_SERIES_STM32WBAX)
+	/* Remove VDDUSB power isolation */
+	LL_PWR_EnableVddUSB();
+
+	/* Make sure that voltage scaling is Range 1 */
+	__ASSERT_NO_MSG(LL_PWR_GetRegulCurrentVOS() == LL_PWR_REGU_VOLTAGE_SCALE1);
+
+	/* Enable VDD11USB */
+	LL_PWR_EnableVdd11USB();
+
+	/* Enable USB OTG internal power */
+	LL_PWR_EnableUSBPWR();
+
+	while (!LL_PWR_IsActiveFlag_VDD11USBRDY()) {
+		/* Wait for VDD11USB supply to be ready */
+	}
+
+	/* Enable USB OTG booster */
+	LL_PWR_EnableUSBBooster();
+
+	while (!LL_PWR_IsActiveFlag_USBBOOSTRDY()) {
+		/* Wait for USB OTG booster to be ready */
+	}
+#elif defined(PWR_USBSCR_USB33SV) || defined(PWR_SVMCR_USV)
+	/*
+	 * VDDUSB independent USB supply (PWR clock is on)
+	 * with LL_PWR_EnableVDDUSB function (higher case)
+	 */
+	LL_PWR_EnableVDDUSB();
+#elif defined(PWR_CR2_USV)
+	/*
+	 * Required for at least STM32L4 devices as they electrically
+	 * isolate USB features from VDDUSB. It must be enabled before
+	 * USB can function. Refer to section 5.1.3 in DM00083560 or
+	 * DM00310109.
+	 */
+	LL_PWR_EnableVddUSB();
+#endif
+	/* Successful control flow in series-specific code above
+	 * will fall here. Set `err` to success value in a single
+	 * place to avoid duplication.
+	 */
+	err = 0;
+
+fini:
+	if (err != 0) {
+		/* An error occurred - drop reference before unlocking */
+		usb_pwr_refcount--;
+	}
+
+	k_sem_give(&pwr_refcount_mutex);
+
+	return err;
+}
--- a/drivers/usb/udc/udc_stm32.c
+++ b/drivers/usb/udc/udc_stm32.c
@@ -22,6 +22,7 @@
 #include <zephyr/sys/util.h>
 
 #include "udc_common.h"
+#include <stm32_usb_common.h>
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(udc_stm32, CONFIG_UDC_DRIVER_LOG_LEVEL);
@@ -720,6 +721,13 @@ int udc_stm32_init(const struct device *dev)
 	struct udc_stm32_data *priv = udc_get_private(dev);
 	const struct udc_stm32_config *cfg = dev->config;
 	HAL_StatusTypeDef status;
+	int err;
+
+	err = stm32_usb_pwr_enable();
+	if (err != 0) {
+		LOG_ERR("Error enabling USB power: %d", err);
+		return err;
+	}
 
 	if (udc_stm32_clock_enable(dev) < 0) {
 		LOG_ERR("Error enabling clock(s)");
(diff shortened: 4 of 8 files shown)


drivers/usb/common/stm32/stm32_usb_common.h
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2025 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
tmon-nordic, line 6: This header file is missing include guards.

mathieuchopstm: Fixed.

tmon-nordic (dismissed):

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no, line 1: >Create infrastructure for shared USB common code on STM32 family, and move
the Power Controller configuration logic to common code.

But it is not shared. You just moved the code from UDC driver. It is not worth it to touch legacy driver at this point. What is the reason to move the code?

If there is stm32_usb_pwr_enable(void), then you need also stm32_usb_pwr_disable(). stm32_usb_pwr.c actually looks like regulator driver. Why not use regulator API and implement a regulator driver? What you would need to implement is just

```c
static DEVICE_API(regulator, api) = {
	.enable = stm32_usbreg_enable,
	.disable = stm32_usbreg_disable,
};
```

jfischer-no (changes requested):

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: > But it is not shared.

It is not shared *for now*, but would become shared when/if an STM32 UHC driver is implemented.

> It is not worth it to touch legacy driver 

What do you mean by `legacy driver`? This is the mainline UDC driver, `usb_dc_stm32.c` is left untouched.

> If there is stm32_usb_pwr_enable(void), then you need also stm32_usb_pwr_disable().

I did not implement `disable()` because it requires refcounting, but...

> stm32_usb_pwr.c actually looks like regulator driver.

...good remark - and this solves refcounting issues. I'll try implementing as a regulator driver.

mathieuchopstm: After a quick and dirty implementation, I don't think a regulator driver will be an acceptable solution. Turning this code into a regulator driver (and thus enabling the regulator subsystem!) adds an unjustified ~1K to the ROM footprint (on our smallest product with USB support, that's 2% of ROM size!)

**Would you be OK if the code was in `soc/st/common` instead of `drivers/usb/common`?**
(note that next PR in patchlist **will** place things in `drivers/usb/common` either way - unless that directory is pure legacy, in which case what's there should be migrated to a Nordic SoC-specific folder :slightly_smiling_face:)

jfischer-no: > It is not shared _for now_, but would become shared when/if an STM32 UHC driver is implemented.

Just FYI, UHC shim driver for DWC2 controller would not be acceptable.

> > It is not worth it to touch legacy driver
> 
> What do you mean by `legacy driver`? This is the mainline UDC driver, `usb_dc_stm32.c` is left untouched.

From the commit message it is not clear where it would be shared, so I have to guess.

> > If there is stm32_usb_pwr_enable(void), then you need also stm32_usb_pwr_disable().
> 
> I did not implement `disable()` because it requires refcounting, but...

But why move still "not yet" shared code and keep it half-backed?

> After a quick and dirty implementation, I don't think a regulator driver will be an acceptable solution. Turning this code into a regulator driver (and thus enabling the regulator subsystem!) adds an unjustified ~1K to the ROM footprint (on our smallest product with USB support, that's 2% of ROM size!)

AFAIK there is no regulator subsystem and the increased footprint sounds acceptable to me, we should stimulate the market, but I do not insist on it. Please add stm32_usb_pwr_disable().

mathieuchopstm: > > It is not shared _for now_, but would become shared when/if an STM32 UHC driver is implemented.
> 
> Just FYI, UHC shim driver for DWC2 controller would not be acceptable.

ST's USB controller also supports Host mode. (not that this matters anyways - regardless of which driver controls the IP, it will have to enable power the same way, hence the move to common code) 

> But why move still "not yet" shared code and keep it half-backed?

I want to avoid functional changes as much as possible while implementing multi-instance support.

tmon-nordic (dismissed):

drivers/usb/common/stm32/Kconfig
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2025 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+
+if SOC_FAMILY_STM32
+
+config STM32_USB_COMMON
+	bool
+	help
+	  Enable compilation of STM32 USB common code.
+
```
jfischer-no: Please remove this empty line.

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,111 @@
...
+	LL_PWR_EnableUSBVoltageDetector();
+
+	/* Per AN2606: USBREGEN not supported when running in FS mode. */
+	LL_PWR_DisableUSBReg();
+	while (!LL_PWR_IsActiveFlag_USB()) {
+		LOG_INF("PWR not active yet");
+		k_msleep(100);
+	}
+#elif defined(CONFIG_SOC_SERIES_STM32U5X)
+	/* Sequence to enable the power of the OTG HS on a stm32U5 serie : Enable VDDUSB */
```
jfischer-no: ` on a STM32U5` ?

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,111 @@
...
+	LL_PWR_EnableVddUSB();
+
+	#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs)
+		/* Configure VOSR register of USB HSTransceiverSupply(); */
+		LL_PWR_EnableUSBPowerSupply();
+		LL_PWR_EnableUSBEPODBooster();
+		while (LL_PWR_IsActiveFlag_USBBOOST() != 1) {
+			/* Wait for USB EPOD BOOST ready */
+		}
+	#endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs) */
```
jfischer-no: We usually do not indent blocks inside #if/#endif.

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,111 @@
...
+		/* Wait for USB OTG booster to be ready */
+	}
+#elif defined(PWR_USBSCR_USB33SV) || defined(PWR_SVMCR_USV)
+	/*
+	 * VDDUSB independent USB supply (PWR clock is on)
+	 * with LL_PWR_EnableVDDUSB function (higher case)
+	 */
+	LL_PWR_EnableVDDUSB();
+#elif defined(PWR_CR2_USV)
+	/*cd
```
jfischer-no: `	/*cd`
----------^ ?

drivers/usb/udc/udc_stm32.c
```diff
@@ -1618,25 +1549,12 @@ static int udc_stm32_driver_init0(const struct device *dev)
...
-		LL_PWR_EnableVddUSB();
-		LL_APB1_GRP1_DisableClock(LL_APB1_GRP1_PERIPH_PWR);
-	}
-	#else
-	LL_PWR_EnableVddUSB();
-#endif /* defined(LL_APB1_GRP1_PERIPH_PWR) */
-#endif /* PWR_CR2_USV */
+	/* Configure Power Controller as necessary for USB */
+	err = stm32_usb_pwr_enable();
+	if (err < 0) {
```
jfischer-no: Please use `	if (err != 0) {`. Btw, you forgot to document return values.

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,111 @@
...
+
+/*
+ * Keep track of whether power is already
+ * enabled here to simplify the USB drivers.
+ */
+static atomic_t usb_pwr_refcount = ATOMIC_INIT(0);
+
+int stm32_usb_pwr_enable(void)
+{
+	atomic_val_t old = atomic_inc(&usb_pwr_refcount);
```
jfischer-no: Should it be incremented if the function exits with `-EIO`?

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,111 @@
...
+
+	if (old > 0) {
```
jfischer-no: Do you need old? Can it be `if (atomic_get(&usb_pwr_refcount)) {` here and `atomic_inc(&usb_pwr_refcount);` at the end of the function?

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: This is just an artifact from verbatim copy. It can be cleaned up at a later time, but I'd like to keep verbatim for now to make bissect easier.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: See https://github.com/zephyrproject-rtos/zephyr/pull/100519#discussion_r2618797627

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: See https://github.com/zephyrproject-rtos/zephyr/pull/100519#discussion_r2618797627

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: No, but in the only situation where it returns, USB is unusable anyways so it doesn't really matter. Can be fixed later when the function is improved, I'd like to keep verbatim for now.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: As far as I understand atomics, a loop + `atomic_cas` would be necessary if I used `atomic_get`. The current implementation seems simpler.

drivers/usb/udc/udc_stm32.c
mathieuchopstm: Done.

drivers/usb/common/stm32/Kconfig
mathieuchopstm: Done.

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: I do not think there is any reason to postpone these fixes.

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: This looks like a bug, please fix it in this commit.

drivers/usb/common/stm32/stm32_usb_common.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_
+#define ZEPHYR_DRIVERS_USB_COMMON_STM32_STM32_USB_COMMON_H_
+
+/**
+ * @brief Configures the Power Controller as necessary
+ * for proper operation of the USB controllers
+ *
+ * @returns 0 on success, negative value on error
```
jfischer-no: What are positive values then?

drivers/usb/udc/udc_stm32.c
jfischer-no: `if (err != 0) {`

drivers/usb/common/stm32/stm32_usb_common.h
mathieuchopstm: Cannot happen.

drivers/usb/udc/udc_stm32.c
mathieuchopstm: It is done - check the diff.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Done

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Done

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Done

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -118,3 +119,62 @@ int stm32_usb_pwr_enable(void)
 
 	return err;
 }
+
+int stm32_usb_pwr_disable(void)
+{
+	atomic_val_t old = atomic_dec(&usb_pwr_refcount);
```
etienne-lms: Are atomic operations sufficient here. Consider 2 threads running
- Thread A runs and has enabled a USB device;
- Thread A starts to shutdown its USB device, calls `stm32_usb_pwr_disable()` that returns with `old == 0`;
- Thread B gets scheduled, powers up another USB device: calls `stm32_usb_pwr_enable()` and start enabling already enabled USB common power supplies;
- Thread A gets scheduled and completes its `stm32_usb_pwr_disable()` sequence.

Don't you need a mutex or like here?

mathieuchopstm: Yes; fixed.

drivers/usb/udc/udc_stm32.c
jfischer-no: No it is not, use `if (err != 0) {`!

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -64,7 +63,7 @@ int stm32_usb_pwr_enable(void)
 	LL_PWR_EnableVddUSB();
 
 	#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs)
-		/* Configure VOSR register of USB HSTransceiverSupply(); */
+		/* Enable HS PHY power supply */
```
jfischer-no: This commit should be squashed with the first commit.

jfischer-no (changes requested):

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: > * Thread A runs and has enabled a USB device;
> 
>     * Thread A starts to shutdown its USB device, calls `stm32_usb_pwr_disable()` that returns with `old == 0`;
> 
>     * Thread B gets scheduled, powers up another USB device: calls `stm32_usb_pwr_enable()` and start enabling already enabled USB common power supplies;

In udc_stm32.c, it is never called from a thread.

drivers/usb/common/stm32/stm32_usb_common.h
jfischer-no: Then change it to `@returns 0 on success, other value on error`.

drivers/usb/udc/udc_stm32.c
mathieuchopstm: ???
image

drivers/usb/common/stm32/stm32_usb_common.h
mathieuchopstm: already done

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: done

drivers/usb/udc/udc_stm32.c
jfischer-no: ~~I still saw the original state. I even reloaded the page.~~

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,130 @@
...
+uint32_t usb_pwr_refcount;
+
+int stm32_usb_pwr_enable(void)
+{
+	uint32_t old_count;
+	int err;
+
+	err = k_sem_take(&pwr_refcount_mutex, K_FOREVER);
+	if (err < 0) {
+		return err;
```
jfischer-no, line 32: `(err < 0)` looks like a spreading disease. All the undefined and unused return values should be threaten as errors. Even https://docs.zephyrproject.org/latest/kernel/services/synchronization/semaphores.html#taking-a-semaphore is doing it right.

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,130 @@
...
+#elif defined(PWR_CR2_USV)
+	/*
+	 * Required for at least STM32L4 devices as they electrically
+	 * isolate USB features from VDDUSB. It must be enabled before
+	 * USB can function. Refer to section 5.1.3 in DM00083560 or
+	 * DM00310109.
+	 */
+	LL_PWR_EnableVddUSB();
+#endif
+	err = 0;
```
jfischer-no, line 122: At this point, it cannot have a value other than 0.

drivers/usb/udc/udc_stm32.c
jfischer-no: It is still there, check L:1561, it will be removed in the second commit, which introduces power disable support. It does not belong there and is confusing.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: If this variable ever gets reused in the future, it might end up with a positive, non-zero value. So I would prefer to keep it this way.

drivers/usb/udc/udc_stm32.c
mathieuchopstm: Ok, I see it now. fixed.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: If anything, quick tree grepping seems to indicate `ret < 0` is more commonly used than `ret != 0` - even `kernel/` uses both patterns :slightly_smiling_face:

This is just a matter of taste; there is zero functional difference between both approaches, unless one considers the kernel API as untrustworthy (i.e., not abiding to the *de facto* rule that errors are signaled by *negative* return values, and thus that *positive or zero* values signal success).

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,130 @@
...
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(stm32_usb_pwr, CONFIG_STM32_USB_COMMON_LOG_LEVEL);
+
+/*
+ * Keep track of whether power is already
+ * enabled here to simplify the USB drivers.
+ */
+K_SEM_DEFINE(pwr_refcount_mutex, 1, 1);
+uint32_t usb_pwr_refcount;
```
etienne-lms: `static`? (for both)

etienne-lms (commented):
LGTM aside 1 comment.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: done

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -11,6 +11,7 @@
 #include <stm32_ll_system.h>
 #include <string.h>
 #include <zephyr/kernel.h>
+#include <zephyr/sys/atomic.h>
```
etienne-lms: **atomic.h** can be removed.
Sorry I missed it in my previous review.

By the way, **string.h** is already included from **zephyr/sys/util.h**.

drivers/usb/common/stm32/stm32_usb_pwr.c
```diff
@@ -0,0 +1,201 @@
...
+	/* Check that power range is 1 or 2 */
+	if (LL_PWR_GetRegulVoltageScaling() < LL_PWR_REGU_VOLTAGE_SCALE2) {
+		LOG_ERR("Wrong Power range to use USB OTG HS");
+		err = -EIO;
+		goto fini;
+	}
+
+	LL_PWR_EnableVddUSB();
+
+	#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32_otghs)
```
etienne-lms: Indentation is a bit weird here. Inherited from original implementation. Worth to be cleaned (as in `stm32_usb_pwr_disable()`)  in a appended commit?

mathieuchopstm: I prefer to keep indented (although - arguably - the `#` should be in column 0), it is more readable (makes it obvious the check is not part of the `#if`/`#elif`/`#endif` chain)

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: > By the way, **string.h** is already included from **zephyr/sys/util.h**.

I'd rather avoid depending on what the Zephyr headers include when it's not obvious.

But good remark, I don't think I'm using anything from that header anymore.

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: No it is not matter of taste. It is matter of consistency and error checking. `if (err)` or `if (err != 0)` is the right way here. `(err < 0)` does not check undefined return values. And until it is clearly documented in the kernel semaphore API that "positive" return value "signal success", `(err < 0)` is just wrong.
You have to fix it.

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: We do not indent `#if` blocks. You have to fix it.

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: Would you remove `#include <zephyr/sys/atomic.h>`?

drivers/usb/udc/udc_stm32.c
```diff
@@ -966,6 +967,12 @@ static int udc_stm32_shutdown(const struct device *dev)
 		/* continue anyway */
 	}
 
+	err = stm32_usb_pwr_disable();
+	if (err != 0) {
+		LOG_ERR("Error disabling USB power: %d", err);
```
etienne-lms, line 972: Since this failure does not fail the sequence:
```suggestion
		LOG_ERR("Error disabling USB power: %d, ignored", err);

mathieuchopstm: But then I would also have to update the other error messages above too...

I don't think it's that important anyways: the messages are indicative, they don't have to be coupled with the code's behavior (as long as the indication isn't misleading, which isn't the case here).

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Done. (Inner `#if`s now use the *space-after-`#`* pattern found in kernel and various other places in tree)

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Done. + removed `#include <string.h>` too

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: > And until it is clearly documented in the kernel semaphore API that "positive" return value "signal success", `(err < 0)` is just wrong.

Well how fortunate for me because this is explicitly documented here: https://github.com/zephyrproject-rtos/zephyr/wiki/Naming-Conventions#return-codes
> Zephyr uses the standard codes in errno.h for all APIs.
>
> As a general rule, 0 indicates success; a **negative** `errno.h` code indicates an error condition. 

And here too: https://docs.zephyrproject.org/latest/develop/languages/c/minimal_libc.html#error-numbers
> Error numbers are used throughout Zephyr APIs to signal error conditions as return values from functions. They are typically returned as the **negative** value of the integer literals defined in this section, and are defined in the errno.h header file.

And here too if that wasn't enough: https://docs.zephyrproject.org/latest/kernel/drivers/index.html#error-handling
> When it is appropriate to return error conditions for the caller to check, 0 should be returned on success and a POSIX errno.h code returned on failure.

And the kernel explicitly lists only three possible return values for `k_sem_take`:
https://github.com/zephyrproject-rtos/zephyr/blob/dd76aa4402d5eaad4e47bcd715f99d39679e9348/include/zephyr/kernel.h#L3669-L3672
Note how the only legal values are `0` (on success) and `-EBUSY`/`-EAGAIN` (**negative**, on error).

The only guideline that *could* be invoked for justifying the change is: https://docs.zephyrproject.org/latest/contribute/style/code.html
> For style aspects where the guidelines don’t offer explicit guidance or permit multiple valid ways to express something, contributors should follow the style of existing code in the tree, with higher importance given to “nearby” code (first look at the function, then the same file, then subsystem, etc).

But that is debatable considering that other things in `drivers/usb/` also use the `err < 0` pattern.

---

I've done the change anyway because I just want to get the code in tree and move on, but these whole arguments are quite frankly tiring.

drivers/usb/common/stm32/stm32_usb_pwr.c
mathieuchopstm: Added a justification comment on top.

mathieuchopstm:
+ rebased

etienne-lms (approved):

erwango (approved):

drivers/usb/common/stm32/stm32_usb_pwr.c
jfischer-no: Nice guidelines! But, as I already wrote, only 0 means success, "And until it is clearly documented in the kernel semaphore API that "positive" return value "signal success", (err < 0) is just wrong." So, please use consistently `(err != 0)` or `(err)` in all the USB contributions for the function calls where only 0 return value is defined as success.

>But that is debatable considering that other things in drivers/usb/ also use the err < 0 pattern.

Unfortunately. One stands out particularly negatively.

jfischer-no (approved):

PR #96039, opened 2025-09-15
drivers: can: tcan4x5x: Add device PM support

rerickson1:
Add support for device runtime power management.
Suspending the device will place it into sleep mode, its lowest
power state.
All config will be lost when suspending the device.
Do not allow suspending the device if it is started or has
RX filters registered.
The application has to reconfigure the CAN device to resume
expected operation.

Testing:

Tested with BL54L15 DVK and CAN FD 6 Click.
https://docs.zephyrproject.org/latest/boards/ezurio/bl54l15_dvk/doc/bl54l15_dvk.html
https://www.mikroe.com/can-fd-6-click

VBAT/VSUP = 12V
Observed 20 uA current draw when suspended.

PR #100520, opened 2025-12-04
drivers: can: mcp251xfd: Add support for sleep mode

tomridl:
Add support for MCP251xFD sleep mode using device power management. When in sleep mode, controller will wake automatically on CAN activity via WAKIF.
Sleep mode is entered by calling 
`pm_device_action_run(can_dev, PM_DEVICE_ACTION_SUSPEND);`
Sleep mode can be exited by calling 
`pm_device_action_run(can_dev, PM_DEVICE_ACTION_RESUME);`
Add config options to enable power management support as well as an option to enable management of transceiver standby automatically when sleep mode is entered.

### DIFF
--- a/drivers/can/Kconfig.mcp251xfd
+++ b/drivers/can/Kconfig.mcp251xfd
@@ -58,31 +58,4 @@ config CAN_MAX_FILTER
 	help
 	  Maximum number of filters supported by the can_add_rx_callback() API call.
 
-config CAN_MCP251XFD_PM_DEVICE
-	bool "Enable power management support"
-	default n
-	depends on PM_DEVICE
-	help
-	  Enable power management support for the MCP251XFD CAN controller.
-	  When enabled, the driver implements suspend and resume callbacks
-	  that transition the device to sleep mode on suspend and restore
-	  the previous operational mode on resume.
-
-config CAN_MCP251XFD_TRANSCEIVER_STANDBY_ON_SLEEP
-	bool "Put CAN transceiver in standby during sleep"
-	default n
-	depends on CAN_MCP251XFD_PM_DEVICE
-	help
-	  When enabled, the CAN transceiver will be put into standby mode
-	  when the CAN controller enters sleep mode. This provides the lowest
-	  power consumption.
-
-	  When disabled, the transceiver remains active during CAN sleep,
-	  which uses more power but may provide faster wake response on
-	  some hardware configurations.
-
-	  If XSTBY is enabled in the device tree and is directly connected to the
-	  standby pin of the transceiver, this option is not requrired, as the controller
-	  will automatically put the transceiver into standby via the XSTBY pin.
-
 endif # CAN_MCP251XFD
--- a/drivers/can/Kconfig.mcp251xfd
+++ b/drivers/can/Kconfig.mcp251xfd
@@ -58,4 +58,31 @@ config CAN_MAX_FILTER
 	help
 	  Maximum number of filters supported by the can_add_rx_callback() API call.
 
+config CAN_MCP251XFD_PM_DEVICE
+	bool "Enable power management support"
+	default n
+	depends on PM_DEVICE
+	help
+	  Enable power management support for the MCP251XFD CAN controller.
+	  When enabled, the driver implements suspend and resume callbacks
+	  that transition the device to sleep mode on suspend and restore
+	  the previous operational mode on resume.
+
+config CAN_MCP251XFD_TRANSCEIVER_STANDBY_ON_SLEEP
+	bool "Put CAN transceiver in standby during sleep"
+	default n
+	depends on CAN_MCP251XFD_PM_DEVICE
+	help
+	  When enabled, the CAN transceiver will be put into standby mode
+	  when the CAN controller enters sleep mode. This provides the lowest
+	  power consumption.
+
+	  When disabled, the transceiver remains active during CAN sleep,
+	  which uses more power but may provide faster wake response on
+	  some hardware configurations.
+
+	  If XSTBY is enabled in the device tree and is directly connected to the
+	  standby pin of the transceiver, this option is not requrired, as the controller
+	  will automatically put the transceiver into standby via the XSTBY pin.
+
 endif # CAN_MCP251XFD
--- a/drivers/can/can_mcp251xfd.c
+++ b/drivers/can/can_mcp251xfd.c
@@ -1823,7 +1823,6 @@ done:
 static int mcp251xfd_pm_action(const struct device *dev, enum pm_device_action action)
 {
 	struct mcp251xfd_data *dev_data = dev->data;
-	const struct mcp251xfd_config *dev_cfg = dev->config;
 	int ret = 0;
 
 	switch (action) {
--- a/drivers/can/can_mcp251xfd.c
+++ b/drivers/can/can_mcp251xfd.c
@@ -331,8 +331,8 @@ static int mcp251xfd_set_mode_internal(const struct device *dev, uint8_t request
 
 #if defined(CONFIG_CAN_FD_MODE)
 	if (dev_data->current_mcp251xfd_mode == MCP251XFD_REG_CON_MODE_CONFIG) {
-		if (requested_mode == MCP251XFD_REG_CON_MODE_CAN2_0 ||
-		    requested_mode == MCP251XFD_REG_CON_MODE_EXT_LOOPBACK ||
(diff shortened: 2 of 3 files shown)


drivers/can/Kconfig.mcp251xfd
```diff
@@ -58,4 +58,31 @@ config CAN_MAX_FILTER
...
 
+config CAN_MCP251XFD_PM_DEVICE
+	bool "Enable power management support"
+	default n
+	depends on PM_DEVICE
+	help
+	  Enable power management support for the MCP251XFD CAN controller.
+	  When enabled, the driver implements suspend and resume callbacks
+	  that transition the device to sleep mode on suspend and restore
+	  the previous operational mode on resume.
```
henrikbrixandersen: This should not be a dedicated, driver-specific Kconfig. Enabling `CONFIG_PM_DEVICE` should be sufficient.

drivers/can/can_mcp251xfd.c
```diff
@@ -330,8 +331,8 @@ static int mcp251xfd_set_mode_internal(const struct device *dev, uint8_t request
 
 #if defined(CONFIG_CAN_FD_MODE)
 	if (dev_data->current_mcp251xfd_mode == MCP251XFD_REG_CON_MODE_CONFIG) {
-		if (requested_mode ==  MCP251XFD_REG_CON_MODE_CAN2_0 ||
-		    requested_mode ==  MCP251XFD_REG_CON_MODE_EXT_LOOPBACK ||
+		if (requested_mode == MCP251XFD_REG_CON_MODE_CAN2_0 ||
+		    requested_mode == MCP251XFD_REG_CON_MODE_EXT_LOOPBACK ||
```
henrikbrixandersen: Please do not mix unrelated whitespace changes with functional changes. Same goes for the rest of the file.

drivers/can/can_mcp251xfd.c
```diff
@@ -1099,6 +1111,24 @@ static void mcp251xfd_handle_interrupts(const struct device *dev)
 			}
 		}
 
+#if defined(CONFIG_CAN_MCP251XFD_PM_DEVICE)
+		if ((reg_int & MCP251XFD_REG_INT_WAKIF) != 0) {
+			LOG_INF("WAKIF Interrupt - CAN bus activity wake");
```
henrikbrixandersen: Please reduce the log level here to DBG.

drivers/can/can_mcp251xfd.c
```diff
@@ -1688,6 +1721,274 @@ static int mcp251xfd_init(const struct device *dev)
...
+	ret = mcp251xfd_get_mode_internal(dev, &current_mode);
+	if (ret < 0) {
+		LOG_ERR("Failed to read current mode [%d]", ret);
+		goto done;
+	}
+	LOG_DBG("Current mode: %d, started: %d", current_mode, dev_data->common.started);
+
+	/* If in operational mode (NORMAL/LISTENONLY), first abort any pending TX
+	 * and transition to CONFIG mode for clean sleep entry.
+	 */
```
henrikbrixandersen: Please invert this to disallow entering sleep mode while in operational mode (see discussion in https://github.com/zephyrproject-rtos/zephyr/pull/96039 for details).

Will the MCP251xFD retain CAN RX filter configuration while in sleep mode?

drivers/can/can_mcp251xfd.c
```diff
@@ -1688,6 +1721,274 @@ static int mcp251xfd_init(const struct device *dev)
 	return ret;
 }
 
+#if defined(CONFIG_CAN_MCP251XFD_PM_DEVICE)
+static int mcp251xfd_enter_sleep_mode(const struct device *dev)
```
henrikbrixandersen, line 1711: This function is very long and complicated. Please consider refactoring it for readability.

drivers/can/can_mcp251xfd.c
```diff
@@ -1688,6 +1721,274 @@ static int mcp251xfd_init(const struct device *dev)
...
+			LOG_ERR("Failed to request CONFIG mode [%d]", ret);
+			goto done;
+		}
+
+		/* Wait for CONFIG mode */
+		ret = mcp251xfd_reg_check_value_wtimeout(
+			dev, MCP251XFD_REG_CON,
+			FIELD_PREP(MCP251XFD_REG_CON_OPMOD_MASK, MCP251XFD_REG_CON_MODE_CONFIG),
+			MCP251XFD_REG_CON_OPMOD_MASK, MCP251XFD_MODE_CHANGE_TIMEOUT_USEC,
+			MCP251XFD_MODE_CHANGE_RETRIES, true);
```
henrikbrixandersen: Please avoid formatting like this (Python-like) where a function call ends with `(` on a line. Put the arguments on the same line up to 100 characters in width.

drivers/can/can_mcp251xfd.c
```diff
@@ -1688,6 +1721,274 @@ static int mcp251xfd_init(const struct device *dev)
...
+
+	case PM_DEVICE_ACTION_RESUME:
+		LOG_DBG("PM resume requested");
+
+#if defined(CONFIG_CAN_MCP251XFD_TRANSCEIVER_STANDBY_ON_SLEEP)
+		if (dev_cfg->common.phy != NULL) {
+			/* Bring transceiver out of standby */
+			(void)can_transceiver_enable(dev_cfg->common.phy, dev_data->common.mode);
+		}
+#endif
```
henrikbrixandersen: Suspending/disabling the CAN transceiver due to PM is the task of the CAN transceiver driver (not the individual CAN controller driver), which needs to implement device PM by itself.

henrikbrixandersen (changes requested):
Thanks you for submitting this. A few initial comments below:

drivers/can/can_mcp251xfd.c
tomridl: Yes, the MCP251xFD retains CAN RX filter configuration while in sleep mode (the same is not true for LP mode, however LP mode has not been implemented here. 
According to the MCP25XXFD family reference manual, sleep mode can be entered from normal mode however in testing I found it more reliable to switch to config mode first. I will change this to match the implementation for tcan4x5x and only allow entry from configuration mode.

drivers/can/Kconfig.mcp251xfd
tomridl: Will do - the reason for the extra config was that entering sleep mode could cause CAN packets to be missed on wakeup, so devices that implement CONFIG_PM for other reasons could optionally prevent the CAN controller going to sleep if it is critical to receive every CAN packet. Could an option perhaps be added to the dts instead?
Edit: Doh, if they don't want to use sleep mode then just don't call pm_device_action_run(can_dev, PM_DEVICE_ACTION_SUSPEND);

drivers/can/can_mcp251xfd.c
tomridl: It appears that the existing mcp251xfd_stop and mcp251xfd_start functions already control the transceiver standby, so switching to configuration mode using can_stop (putting the device in configuration mode) will already put the transceiver to sleep. I will switch the PM_DEVICE_ACTION_RESUME to use mcp251xfd_start which will then switch the transceiver out of standby.

henrikbrixandersen: Since the CAN controller will be in stopped mode when entering sleep mode, resuming should not put it into started mode - that's up to the application.

tomridl: Got it - will restore to config mode after wake from sleep and leave it up to the app to start.

tomridl: Although then I'm not sure how to restart the controller when waking from CAN activity - the state_changed_callback doesn't have a defined state for WAKIF so how should this be handled?

drivers/can/can_mcp251xfd.c
tomridl: With the removal of the mode changes this function should now be an acceptable length.

drivers/can/can_mcp251xfd.c
henrikbrixandersen: I have not forgotten about this. I am currently working on establishing a basic CAN power management test suite based on the current functionality of the TCAN4x5x driver - which we can then build upon, adding further test cases - to ensure a unified behaviour across CAN driver power management implementations.

After adding the initial test suite, the first thing on my list is to expand the CAN controller state to represent "suspended".

PR #100172, opened 2025-11-27
Add support for Synaptics SR100 SoCs

aweissel:
Add support for Synaptics SR100 SoCs, the Astra Machina Micro (RDK) board, Synaptics-specific drivers.
Add clock control support to a few Synopsys Designware drivers.

PR #100527, opened 2025-12-04
drivers: usb: dwc2: Add support for Synaptics SR100 SoCs

aweissel:
Add USB device driver support for Synaptics SR100 devices (clock control and caps) as a vendor quirk for the dwc2 USB-UDC driver.

~~Depends on: #100172~~

### DIFF
--- a/drivers/usb/udc/udc_dwc2.h
+++ b/drivers/usb/udc/udc_dwc2.h
@@ -162,6 +162,9 @@ static inline struct usb_dwc2_reg *dwc2_get_base(const struct device *dev)
 #if DT_HAS_COMPAT_STATUS_OKAY(st_stm32f4_fsotg)
 #include "udc_dwc2_stm32f4_fsotg.h"
 #endif
+#if DT_HAS_COMPAT_STATUS_OKAY(syna_sr100_usb)
+#include "udc_dwc2_syna_sr100_usb.h"
+#endif
 
 #define UDC_DWC2_VENDOR_QUIRK_GET(n)						\
 	COND_CODE_1(DT_NODE_VENDOR_HAS_IDX(DT_DRV_INST(n), 1),			\
--- a/drivers/usb/udc/udc_dwc2_syna_sr100_usb.h
+++ b/drivers/usb/udc/udc_dwc2_syna_sr100_usb.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2025 Synaptics, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_DRIVERS_USB_UDC_DWC2_SYNA_SR100_USB_H
+#define ZEPHYR_DRIVERS_USB_UDC_DWC2_SYNA_SR100_USB_H
+
+#include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/reset.h>
+
+struct syna_sr100_usb_clk {
+	const struct device *const dev;
+	const clock_control_subsys_t id;
+	const struct reset_dt_spec reset;
+};
+
+static inline int syna_sr100_usb_pre_enable(const struct syna_sr100_usb_clk *const clk)
+{
+	if (device_is_ready(clk->reset.dev)) {
+		int ret = reset_line_toggle(clk->reset.dev, clk->reset.id);
+
+		if (ret != 0) {
+			return ret;
+		}
+	}
+
+	if (!device_is_ready(clk->dev)) {
+		return -ENODEV;
+	}
+
+	return clock_control_on(clk->dev, clk->id);
+}
+
+static inline int syna_sr100_usb_init_caps(const struct device *dev)
+{
+	struct udc_data *data = dev->data;
+
+	data->caps.hs = true;
+
+	return 0;
+}
+
+#define QUIRK_SYNA_SR100_USB_DEFINE(n)						\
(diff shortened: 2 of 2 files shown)


josuah:
I see no modification to any `.dts` file, is that for an out of tree device?
If not, which in-tree hardware is this meant for?
Thanks for the addition.

drivers/usb/udc/udc_dwc2.c
```diff
@@ -3516,6 +3541,16 @@ static const struct udc_api udc_dwc2_api = {
 	}
 #endif
 
+#define UDC_DWC2_CLOCK_DEFINE(n)						\
+	COND_CODE_1(DT_INST_NODE_HAS_PROP(n, clocks),				\
```
tmon-nordic: You would need to document the clocks property. Currently the PHY clock enable is effectively handled by vendor quirks.

aweissel: Good point; should I add a vendor quirk similar to "stm32f4" instead? I'm not sure if the generic clock_control support added in my PR would break the clock handling in the vendor quirks (see, e.g., samples/subsys/usb/cdc_acm/nucleo_f413zh_dwc2.overlay).
The "clocks" phandle is already documented in "base.yaml".

tmon-nordic: It really depends on how "typical" a hardware is when it comes to clock control. This generic handling should not break others because others currently do not use `clocks` property.

The question is also about what clock this is all about. Peripheral clock? PHY clock? Some other dependent clock that hardware needs to have running but is neither peripheral clock nor PHY clock?

tmon-nordic: Another important question: is it safe to disable the clock when hibernated?

How should we describe all this? Currently vendor quirks are the method of describing this sort of fine details. Because it can vary quite much between different implementations, it is not easy to determine what should consist of a vendor quirk and what should be achieved via some Zephyr standard mechanism.

aweissel: Given these arguments, I'm now more inclined to re-write this PR and implement a vendor quirk for the SR100 platform.
Furthermore, the "clocks" property is used by the stm32f4 quirk (a build with the overlay "samples/subsys/usb/cdc_acm/nucleo_f413zh_dwc2.overlay" would fail), and, as you stated, we don't know if the clock can be disabled. What do you think?

tmon-nordic: Using quirks would keep up the current status-quo. Keeping status-quo can potentially make future rework harder, but I just don't really see a way to easily answer all these questions with devicetree entries without over-complicating it.

tmon-nordic (dismissed):

dts/bindings/usb/snps,dwc2.yaml
```diff
@@ -5,7 +5,7 @@ description: DesignWare OTG USB 2.0 controller
 
 compatible: "snps,dwc2"
 
-include: [base.yaml]
+include: [base.yaml, reset-device.yaml]
```
tmon-nordic: Putting this here sounds wrong to me. I don't quite know how to supply the necessary information though.

This whole issue is arising from the fact that different vendors differently integrate Synopsys USB controller. I think the integration details should not be part of "common baseline" compatible.

@bjarki-andreasen can you come up with some ideas here?

tmon-nordic (dismissed):

bjarki-andreasen (dismissed):

drivers/usb/udc/udc_dwc2_vendor_quirks.h
```diff
@@ -715,6 +715,64 @@ DT_INST_FOREACH_STATUS_OKAY(QUIRK_ESP32_USB_OTG_DEFINE)
...
+#include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/reset.h>
+
+struct syna_usb_clk {
+	const struct device *const dev;
+	const clock_control_subsys_t id;
+	const struct reset_dt_spec reset;
+};
+
+static inline int syna_usb_enable_clk(const struct syna_usb_clk *const clk)
```
rvosk: the name is too generic, can you plz name it as syna_sr100_usb_enable_clk or at least syna_sr_usb_eanble_clk?

drivers/usb/udc/udc_dwc2_vendor_quirks.h
```diff
@@ -715,6 +715,64 @@ DT_INST_FOREACH_STATUS_OKAY(QUIRK_ESP32_USB_OTG_DEFINE)
...
+	{									\
+		return syna_usb_enable_clk(&syna_usb_clk_##n);			\
+	}									\
+										\
+	struct dwc2_vendor_quirks dwc2_vendor_quirks_##n = {			\
+		.caps = syna_usb_init_caps,					\
+		.pre_enable = syna_usb_enable_clk_##n,				\
+	};
+
+DT_INST_FOREACH_STATUS_OKAY(QUIRK_SYNA_USB_DEFINE)
```
rvosk: ditto

drivers/usb/udc/udc_dwc2_vendor_quirks.h
rvosk: in fact, this function not only enable clk but also toggle reset, so naming it as xxx_enable_clk is a bit misleading

tmon-nordic:
Needs rebase to fix the conflicts after vendor quirks got separated into individual per vendor files.

tmon-nordic (approved):

jfischer-no (approved):

PR #100532, opened 2025-12-04
drivers: gpio: stm32: assert valid configuration for wake-up pins on STM32WBA

mathieuchopstm:
When configuring a GPIO pin as wake-up pin on STM32WBA series, ensure that the provided configuration is correct.

While at it, also ensure that errors when performing the wake-up pin configuration in general are reported to the caller, rather than merely logging an error message (which can be invisible).

### DIFF
--- a/drivers/gpio/gpio_stm32.c
+++ b/drivers/gpio/gpio_stm32.c
@@ -530,9 +530,11 @@ static int gpio_stm32_config(const struct device *dev,
 			.dt_flags = (gpio_dt_flags_t)flags,
 		};
 
-		if (stm32_pwr_wkup_pin_cfg_gpio((const struct gpio_dt_spec *)&gpio_dt_cfg)) {
+		err = stm32_pwr_wkup_pin_cfg_gpio(&gpio_dt_cfg);
+		if (err < 0) {
 			LOG_ERR("Could not configure GPIO %s pin %d as a wake-up source",
 					gpio_dt_cfg.port->name, gpio_dt_cfg.pin);
+			return err;
 		}
 #else
 		LOG_DBG("STM32_GPIO_WKUP flag has no effect when CONFIG_POWEROFF=n");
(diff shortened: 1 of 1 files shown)


drivers/gpio/gpio_stm32.c
```diff
@@ -524,6 +524,24 @@ static int gpio_stm32_config(const struct device *dev,
...
+		if (IS_ENABLED(CONFIG_SOC_SERIES_STM32WBAX) &&
+		    (flags & GPIO_OUTPUT) == 0 &&
+		    ((flags & GPIO_INPUT) == 0 || (flags & (GPIO_PULL_DOWN | GPIO_PULL_UP)) == 0)) {
+			/*
+			 * RM0493 Rev. 7 Table 93 / RM0515 Rev. 3 Table 95:
+			 * Only input pins with PU/PD and output pins are retained in Standby.
+			 * Other pins are placed in High-Z state and can't be used for wake-up.
+			 */
+			LOG_ERR("STM32WBA: wake-up pin must be configured as "
+				"output, or input+pull-up/pull-down");
```
etienne-lms, line 541: Worth to print the faulty pin reference: `"%s%u", gpio_dt_cfg.port->name, gpio_dt_cfg.pin`?

etienne-lms (approved):

drivers/gpio/gpio_stm32.c
mathieuchopstm: I see little value:
1. It'll make the already big message even larger
2. The added info has little value (`port->name` is `gpio@xxxxxxxx`, not the nodelabel - alas!)
3. Caller will get an error return value and probably log by itself, with an hopefully more accurate message

This is not an "expected" error anyways: you're supposed to hit it once, fix your app and never see it again.

JarmouniA (approved):

erwango (approved):

PR #96799, opened 2025-09-30
drivers: video: stm32_dcmipp: Initialize ISP only once

ABESTM:
This fixes bug where stm32_dcmipp_isp_init is called multiple times.

PR #100539, opened 2025-12-04
soc: stm32: Introduce stm32_iocell driver

ABESTM:
### Goal
This PR aims to move HSLV (high-speed low voltage) to dedicated driver, since it is now hardcoded in soc or driver files.
The HSLV option is used on advanced devices if the corresponding power supply (e.g. VDD or VDDXSPI) is powered from 1.8V. So, it depends on board and should be handled by device tree.
### Open points
This is still draft since I'm not sure about several points:
* The driver is called via SYS_INIT before kernel is started, but maybe it should be called directly from SoC specific file like `soc/st/stm32/stm32h7rsx/soc.c`
* The driver allows some error logging, but maybe it is too verbose. Also, it is run later, once the kernel is initialized.
* The PR also removes the current fixed configuration in XSPI driver and SoC for N6, but maybe this should be separate PR.
* Make sure the PR covers all affected board (currently only DK boards are updated).

### DIFF
--- a/boards/st/nucleo_n657x0_q/Kconfig.defconfig
+++ b/boards/st/nucleo_n657x0_q/Kconfig.defconfig
@@ -8,4 +8,7 @@ if BOARD_NUCLEO_N657X0_Q
 configdefault NET_L2_ETHERNET
 	default y
 
+configdefault OTP
+	default y if ETH_STM32_HAL && NVMEM
+
 endif # BOARD_NUCLEO_N657X0_Q
--- a/drivers/flash/flash_stm32_xspi.c
+++ b/drivers/flash/flash_stm32_xspi.c
@@ -2114,18 +2114,6 @@ static int flash_stm32_xspi_init(const struct device *dev)
 		LOG_ERR("XSPI mode SPI|DUAL|QUAD/DTR is not valid");
 		return -ENOTSUP;
 	}
-#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
-	__HAL_RCC_SBS_CLK_ENABLE();
-	if (dev_data->hxspi.Instance == XSPI1) {
-		LL_PWR_EnableXSPIM1();
-		LL_SBS_EnableXSPI1SpeedOptim();
-	} else if (dev_data->hxspi.Instance == XSPI2) {
-		LL_PWR_EnableXSPIM2();
-		LL_SBS_EnableXSPI2SpeedOptim();
-	} else {
-		__ASSERT(0, "Not an XSPI instance?!");
-	}
-#endif /* CONFIG_SOC_SERIES_STM32H7RSX */
 
 	/* Signals configuration */
 	ret = pinctrl_apply_state(dev_cfg->pcfg, PINCTRL_STATE_DEFAULT);
--- a/dts/arm/st/n6/stm32n6.dtsi
+++ b/dts/arm/st/n6/stm32n6.dtsi
@@ -16,6 +16,7 @@
 #include <zephyr/dt-bindings/memory-controller/stm32-fmc-nor-psram.h>
 #include <zephyr/dt-bindings/gpio/gpio.h>
 #include <zephyr/dt-bindings/video/video-interfaces.h>
+#include <zephyr/dt-bindings/power/stm32n6_iocell.h>
 #include <freq.h>
 #include <mem.h>
 
@@ -239,6 +240,37 @@
 	};
 
 	soc {
+		iocell: iocell {
+			compatible = "st,stm32-iocell";
+			status = "disabled";
+
+			/* If iocell is enabled,
+			 * the bsec node needs to be enabled as well
+			 */
+			nvmem-cells = <&otp124_hconf1>;
+			nvmem-cell-names = "safeguard";
+
+			vddio {
+				domain = <STM32_DT_IOCELL_VDD>;
+			};
+
+			vddio2 {
+				domain = <STM32_DT_IOCELL_VDDIO2>;
+			};
+
+			vddio3 {
+				domain = <STM32_DT_IOCELL_VDDIO3>;
+			};
+
+			vddio4 {
+				domain = <STM32_DT_IOCELL_VDDIO4>;
+			};
+
+			vddio5 {
+				domain = <STM32_DT_IOCELL_VDDIO5>;
+			};
+		};
+
 		rcc: rcc@56028000 {
 			compatible = "st,stm32n6-rcc";
 			clocks-controller;
@@ -756,6 +788,11 @@
 				compatible = "fixed-layout";
 				#address-cells = <1>;
 				#size-cells = <1>;
+
+				otp124_hconf1: opt124_hconf1@1f0 {
+					reg = <0x1f0 0x4>;
+					#nvmem-cell-cells = <0>;
+				};
 
 				mac_address0: mac-address@2a4 {
 					reg = <0x2a4 0x6>;
--- a/dts/bindings/power/st,stm32-iocell.yaml
+++ b/dts/bindings/power/st,stm32-iocell.yaml
@@ -1,0 +1,57 @@
+# Copyright (c) 2026 STMicroelectronics
+# SPDX-License-Identifier: Apache-2.0
+
+description: STM32 I/O cell controller
+
+compatible: "st,stm32-iocell"
+
+include:
+  - name: base.yaml
+  - name: nvmem-consumer.yaml
+
+properties:
+  "#address-cells":
+    const: 0
+
+  "#size-cells":
+    const: 0
+
+  nvmem-cells:
+    description: |
+      Single nvmem-cells entry named "safeguard" can be provided to access
+      OTP/option byte value where HSLV enable bit is stored.
+      This is optional depending on the SoC series.
+
+child-binding:
+  description: |
+    IO power domains
+
+  properties:
+    domain:
+      type: int
+      description: |
+        I/O domain managed by the cell.
+        This must be one of the SoC-specific values defined in header
+        `include/zephyr/dt-bindings/power/stm32XXX_iocell.h`.
+
+    hslv-mode:
+      type: string
+      default: "off"
+      enum:
+        - "off"
+        - "on"
+        - "auto"
+      description: |
+        Allows to turn on HSLV (high-speed low voltage) option.
+        This option will work only if the corresponding bit is set in
+        option bytes / OTP fuse.
+        Enabling this option on domain that is powered from 3.3V
+        can permanently damage the device! Please consult datasheet for
+        precise threshold value.
+        Selecting "auto" will follow the setting in option bytes / OTP.
+        This can be useful for board allowing both 1.8V and 3.3V operation.
+
+    compensation-enabled:
+      type: boolean
+      description: |
+        Enables the I/O compensation cell
--- a/include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
+++ b/include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file
+ * @brief I/O cell definitions for STM32H7R/S devices
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+
+/**
+ * @name I/O power domains used to configure STM32 iocell device
+ * @{
+ */
+
+#define STM32_DT_IOCELL_VDDIO (0) /**< I/O domain powered by VDD supply */
+#define STM32_DT_IOCELL_XSPI1 (1) /**< I/O domain powered by VDDXSPI1 supply */
+#define STM32_DT_IOCELL_XSPI2 (2) /**< I/O domain powered by VDDXSPI2 supply */
+
+/** @brief Return true if @a domain is valid I/O domain */
+#define STM32_DT_IOCELL_DOMAIN_VALID(domain) (((domain) >= 0) && ((domain) <= 2))
+
+/** @} */
+
+#endif /* ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_ */
--- a/include/zephyr/dt-bindings/power/stm32n6_iocell.h
+++ b/include/zephyr/dt-bindings/power/stm32n6_iocell.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file
+ * @brief I/O cell definitions for STM32N6 devices
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+
+/**
+ * @name I/O power domains used to configure STM32 iocell device
+ * @{
+ */
+
+#define STM32_DT_IOCELL_VDDIO2 (0) /**< I/O domain powered by VDDIO2 supply */
+#define STM32_DT_IOCELL_VDDIO3 (1) /**< I/O domain powered by VDDIO3 supply */
+#define STM32_DT_IOCELL_VDDIO4 (2) /**< I/O domain powered by VDDIO4 supply */
+#define STM32_DT_IOCELL_VDDIO5 (3) /**< I/O domain powered by VDDIO5 supply */
+#define STM32_DT_IOCELL_VDD    (4) /**< I/O domain powered by VDD supply */
+
+/** @brief Return true if @a domain is valid I/O domain */
+#define STM32_DT_IOCELL_DOMAIN_VALID(domain) (((domain) >= 0) && ((domain) <= 4))
+
+/** @} */
+
+#endif /* ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_ */
--- a/soc/st/stm32/common/Kconfig
+++ b/soc/st/stm32/common/Kconfig
@@ -51,3 +51,26 @@ config STM32_BACKUP_PROTECTION
 	help
 	  Enabled for SoCs for which access protection to backup domain
 	  resources needs to be explicitly handled.
+
+config STM32_IOCELL
+	bool "STM32 I/O cell"
+	default y
+	depends on DT_HAS_ST_STM32_IOCELL_ENABLED
+	help
+	  Enable support for STM32 I/O compensation cell and HSLV configuration
+
+config STM32_IOCELL_PRIORITY
+	int "STM32 I/O cell initialization priority"
+	default 50
+	help
+	  Initialization priority of the routine within the EARLY level.
+
+config STM32_IOCELL_CHECK_ENABLE
+	bool "STM32 I/O cell check"
+	default y
+	depends on LOG
+	depends on DT_HAS_ST_STM32_IOCELL_ENABLED
+	help
+	  Enables logging of I/O cell driver. This can be useful for reporting
+	  discrepancy between device-tree configuration and value stored in
+	  device OTP/option bytes
--- a/soc/st/stm32/common/stm32_iocell.c
+++ b/soc/st/stm32/common/stm32_iocell.c
@@ -1,0 +1,300 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+#include <zephyr/init.h>
+#include <zephyr/types.h>
+#include <soc.h>
+#include <zephyr/devicetree.h>
+
+#include <stm32_ll_system.h>
+#include <stm32_ll_pwr.h>
+#include <stm32_bitops.h>
+
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(iocell, CONFIG_SOC_LOG_LEVEL);
+
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+#include <zephyr/dt-bindings/power/stm32h7rs_iocell.h>
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+#include <zephyr/dt-bindings/power/stm32n6_iocell.h>
+#include <zephyr/nvmem.h>
+#define STM32_IOCELL_USE_NVMEM          1
+#define STM32_IOCELL_OTP_HCONF1_DEFAULT 0x0
+#endif /* CONFIG_SOC_SERIES_... */
+
+#define DT_DRV_COMPAT     st_stm32_iocell
+#define STM32_IOCELL_NODE DT_DRV_INST(0)
+
+#if defined(CONFIG_SOC_SERIES_STM32N6X)
+#define IOCELL_FUSE_NAME "OTP fuse"
+#else
+#define IOCELL_FUSE_NAME "option bit"
+#endif /* CONFIG_SOC_SERIES_... */
+
+#define IOCELL_HSLV_DISCLAIMER                                                                     \
+	"The I/O HSLV configuration must not be enabled "                                          \
+	"if the corresponding power supply is above certain threshold. "                           \
+	"Enabling it while the voltage is higher can damage the device"
+
+struct stm32_iocell_config {
+#if defined(STM32_IOCELL_USE_NVMEM)
+	struct nvmem_cell cell;
+#endif
+};
+
+struct stm32_iocell_data {
+#if defined(CONFIG_SOC_SERIES_STM32N6X)
+	uint32_t otp_hconf1_value;
+	int otp_correct;
+#endif
+};
+
+static const struct stm32_iocell_config iocell_config = {
+#if defined(STM32_IOCELL_USE_NVMEM)
+	.cell = NVMEM_CELL_GET_BY_NAME_OR(STM32_IOCELL_NODE, safeguard, {0}),
+#endif
+};
+
+static struct stm32_iocell_data iocell_data = {
+#if defined(CONFIG_SOC_SERIES_STM32N6X)
+	.otp_hconf1_value = STM32_IOCELL_OTP_HCONF1_DEFAULT,
+#endif
+};
+
+/* Returns 1 if HSLV can be configured
+ * Returns 0 if HSLV can't be configured
+ * Returns < 0 on error
+ */
+static int iocell_is_hslv_allowed(uint16_t domain)
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	switch (domain) {
+	case STM32_DT_IOCELL_VDDIO:
+		return (FLASH->OBW1SR & FLASH_OBW1SR_VDDIO_HSLV) == FLASH_OBW1SR_VDDIO_HSLV;
+	case STM32_DT_IOCELL_XSPI1:
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI1_HSLV) == FLASH_OBW1SR_XSPI1_HSLV;
+	case STM32_DT_IOCELL_XSPI2:
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI2_HSLV) == FLASH_OBW1SR_XSPI2_HSLV;
+	default:
+		return -EINVAL;
+	}
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	struct stm32_iocell_data *data = &iocell_data;
+
+	if (!data->otp_correct) {
+		return -EIO;
+	}
+	switch (domain) {
+	case STM32_DT_IOCELL_VDD:
+		return (data->otp_hconf1_value & BIT(17)) != 0;
+	case STM32_DT_IOCELL_VDDIO2:
+		return (data->otp_hconf1_value & BIT(16)) != 0;
+	case STM32_DT_IOCELL_VDDIO3:
+		return (data->otp_hconf1_value & BIT(15)) != 0;
+	case STM32_DT_IOCELL_VDDIO4:
+		return (data->otp_hconf1_value & BIT(14)) != 0;
+	case STM32_DT_IOCELL_VDDIO5:
+		return (data->otp_hconf1_value & BIT(13)) != 0;
+	default:
+		return -EINVAL;
+	}
+#endif /* CONFIG_SOC_SERIES_... */
+}
+
+/* Returns 0 on success
+ * Returns < 0 on error
+ */
+static int iocell_enable_hslv_runtime(uint16_t domain)
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	switch (domain) {
+	case STM32_DT_IOCELL_VDDIO:
+		LL_SBS_EnableIOSpeedOptim();
+		break;
+	case STM32_DT_IOCELL_XSPI1:
+		LL_SBS_EnableXSPI1SpeedOptim();
+		break;
+	case STM32_DT_IOCELL_XSPI2:
+		LL_SBS_EnableXSPI2SpeedOptim();
+		break;
+	default:
+		return -EINVAL;
+	}
+	return 0;
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	switch (domain) {
+	case STM32_DT_IOCELL_VDD:
+		LL_PWR_SetVddIOVoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+		break;
+	case STM32_DT_IOCELL_VDDIO2:
+		LL_PWR_SetVddIO2VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+		break;
+	case STM32_DT_IOCELL_VDDIO3:
+		LL_PWR_SetVddIO3VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+		break;
+	case STM32_DT_IOCELL_VDDIO4:
+		LL_PWR_SetVddIO4VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+		break;
+	case STM32_DT_IOCELL_VDDIO5:
+		LL_PWR_SetVddIO5VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+		break;
+	default:
+		return -EINVAL;
+	}
+	return 0;
+#endif /* CONFIG_SOC_SERIES_... */
+}
+
+static int iocell_config_domain_auto(uint16_t domain, const char *domain_name)
+{
+	int is_allowed = iocell_is_hslv_allowed(domain);
+
+	if (is_allowed < 0) {
+		LOG_ERR("Failed to read " IOCELL_FUSE_NAME " value of HSLV: \"%s\" (%d)",
+			domain_name, domain);
+		return is_allowed;
+	} else if (is_allowed > 0) {
+		return iocell_enable_hslv_runtime(domain);
+	}
+	return 0;
+}
+
+static int iocell_enable_compensation(uint16_t domain)
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	/* Currently targets MCU revisions Y & B */
+
+	/* This implements workaround described in ES0596:
+	 * 2.2.15. I/O compensation could alter duty-cycle of high-frequency output signal
+	 */
+	uint32_t shift;
+	uint32_t nmos, pmos;
+
+	switch (domain) {
+	case STM32_DT_IOCELL_VDDIO:
+		shift = 0;
+		break;
+	case STM32_DT_IOCELL_XSPI1:
+		shift = 1;
+		break;
+	case STM32_DT_IOCELL_XSPI2:
+		shift = 2;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	uint32_t codesel_bit = BIT((shift * 2) + 1);
+	uint32_t en_bit = BIT(shift * 2);
+
+	if (SBS->CCCSR & en_bit) {
+		/* Compensation cell already enabled */
+		return 0;
+	}
+
+	/* Configure Cell selection for domain and enable.
+	 * Clear COMP_CODESEL bit and set COMP_EN bit.
+	 */
+	stm32_reg_modify_bits(&SBS->CCCSR, codesel_bit, en_bit);
+
+	/* Wait for ready flag */
+	while ((SBS->CCCSR & BIT(shift + 8)) == 0) {
+	}
+
+	/* Read and adjust PMOS and NMOS values*/
+	nmos = (SBS->CCVALR >> (8 * shift)) & 0xF;
+	pmos = (SBS->CCVALR >> (8 * shift + 4)) & 0xF;
+	/* Add/subtract 2 from values, preventing overflow/underflow */
+	nmos = (nmos < 0xD) ? (nmos + 2) : 0xF;
+	pmos = (pmos > 0x2) ? (pmos - 2) : 0x0;
+
+	/* Write modified values back */
+	stm32_reg_modify_bits(&SBS->CCSWVALR, 0xFF << (shift * 8),
+			      ((pmos << 4) | nmos) << (shift * 8));
+
+	/* Configure Cell selection for register
+	 * Set COMP_CODESEL bit.
+	 */
+	stm32_reg_set_bits(&SBS->CCCSR, codesel_bit);
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* There is a global workaround implemnted in SystemInit.
+	 * For N6 IO compensation cell is always active.
+	 */
+	ARG_UNUSED(domain);
+#endif /* CONFIG_SOC_SERIES_... */
+	return 0;
+}
+
+__maybe_unused static void iocell_hslv_check_log(int domain_id, int is_allowed, int hslv,
+						 const char *domain_name)
+{
+	if (is_allowed < 0) {
+		/* This shouldn't happen with proper implementation */
+		LOG_ERR("Failed to read " IOCELL_FUSE_NAME " value of HSLV: \"%s\" (%d)",
+			domain_name, domain_id);
+	} else if (!is_allowed && hslv) {
+		LOG_ERR("HSLV configuration for \"%s\" blocked by " IOCELL_FUSE_NAME, domain_name);
+		LOG_ERR(IOCELL_HSLV_DISCLAIMER);
+	} else if (is_allowed && !hslv) {
+		LOG_WRN("HSLV allowed in " IOCELL_FUSE_NAME " for domain \"%s\", but not enabled",
+			domain_name);
+		LOG_WRN(IOCELL_HSLV_DISCLAIMER);
+	} else {
+		/* Everything ok */
+	}
+}
+
+#define STM32_IOCELL_INIT(node_id)                                                                 \
+	BUILD_ASSERT(STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain)),                       \
+		     "Invalid domain ID for: " DT_NODE_FULL_NAME(node_id));                        \
+	if (DT_ENUM_HAS_VALUE(node_id, hslv_mode, auto)) {                                         \
+		result = iocell_config_domain_auto(DT_PROP(node_id, domain),                       \
+						   DT_NODE_FULL_NAME(node_id));                    \
+	} else if (DT_ENUM_HAS_VALUE(node_id, hslv_mode, on)) {                                    \
+		result = iocell_enable_hslv_runtime(DT_PROP(node_id, domain));                     \
+	}                                                                                          \
+	if (result < 0)                                                                            \
+		ret = result;                                                                      \
+	if (DT_PROP(node_id, compensation_enabled)) {                                              \
+		result = iocell_enable_compensation(DT_PROP(node_id, domain));                     \
+	}                                                                                          \
+	if (IS_ENABLED(CONFIG_STM32_IOCELL_CHECK_ENABLE)) {                                        \
+		iocell_hslv_check_log(DT_PROP(node_id, domain),                                    \
+				      iocell_is_hslv_allowed(DT_PROP(node_id, domain)),            \
+				      DT_ENUM_HAS_VALUE(node_id, hslv_mode, on),                   \
+				      DT_NODE_FULL_NAME(node_id));                                 \
+	}                                                                                          \
+	if (result < 0)                                                                            \
+		ret = result;
+
+static int iocell_init(const struct device *dev)
+{
+	ARG_UNUSED(dev);
+	int ret = 0, result = 0;
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	__HAL_RCC_SBS_CLK_ENABLE();
+	/* CSI is required by IO compensation cell */
+	__HAL_RCC_CSI_ENABLE();
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* PWR & BSEC clocks enabled by default */
+
+	if (nvmem_cell_read(&iocell_config.cell, &iocell_data.otp_hconf1_value, 0, 4) != 0) {
+		iocell_data.otp_hconf1_value = STM32_IOCELL_OTP_HCONF1_DEFAULT;
+		iocell_data.otp_correct = 0;
+	} else {
+		iocell_data.otp_correct = 1;
+	}
+#endif /* CONFIG_SOC_SERIES_... */
+	DT_FOREACH_CHILD_STATUS_OKAY(STM32_IOCELL_NODE, STM32_IOCELL_INIT)
+
+	return ret;
+}
+
+DEVICE_DT_DEFINE(STM32_IOCELL_NODE, iocell_init, NULL, &iocell_data, &iocell_config, PRE_KERNEL_1,
+		 CONFIG_STM32_IOCELL_PRIORITY, NULL);
--- a/soc/st/stm32/stm32n6x/Kconfig.defconfig
+++ b/soc/st/stm32/stm32n6x/Kconfig.defconfig
@@ -30,4 +30,14 @@ config STM32_LTDC_FB_SMH_ALIGN
 config ROM_START_OFFSET
 	default 0x400 if BOOTLOADER_MCUBOOT
 
+if STM32_IOCELL
+
+configdefault OTP
+	default y
+
+configdefault NVMEM
+	default y
+
+endif # STM32_IOCELL
+
 endif # SOC_SERIES_STM32N6X
--- a/boards/ruiside/art_pi2/art_pi2.dts
+++ b/boards/ruiside/art_pi2/art_pi2.dts
@@ -121,3 +121,17 @@
 &vbat {
 	status = "okay";
 };
+
+&iocell {
+	status = "okay";
+
+	vddxspi1 {
+		hslv-mode = "on";
+		compensation-enabled;
+	};
+
+	vddxspi2 {
+		hslv-mode = "on";
+		compensation-enabled;
+	};
+};
--- a/boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
+++ b/boards/st/nucleo_h7s3l8/nucleo_h7s3l8.dts
@@ -32,3 +32,12 @@
 		};
 	};
 };
+
+&iocell {
+	status = "okay";
+
+	vddxspi2 {
+		hslv-mode = "on";
+		compensation-enabled;
+	};
+};
--- a/boards/st/nucleo_n657x0_q/nucleo_n657x0_q_common.dtsi
+++ b/boards/st/nucleo_n657x0_q/nucleo_n657x0_q_common.dtsi
@@ -256,6 +256,8 @@ zephyr_udc0: &usbotg_hs1 {
 	pinctrl-names = "default";
 	phy-connection-type = "rmii";
 	phy-handle = <&eth_phy>;
+	nvmem-cells = <&mac_address0>;
+	nvmem-cell-names = "mac-address";
 };
 
 &mdio {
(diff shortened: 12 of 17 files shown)


soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	if (domain == STM32_DT_IOCELL_VDDIO) {
+		return (FLASH->OBW1SR & FLASH_OBW1SR_VDDIO_HSLV) == FLASH_OBW1SR_VDDIO_HSLV;
+	} else if (domain == STM32_DT_IOCELL_XSPI1) {
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI1_HSLV) == FLASH_OBW1SR_XSPI1_HSLV;
+	} else if (domain == STM32_DT_IOCELL_XSPI2) {
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI2_HSLV) == FLASH_OBW1SR_XSPI2_HSLV;
+	} else {
+		return -1;
```
mathieuchopstm: Maybe?
```suggestion
		return 0;
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+	__HAL_RCC_BSEC_CLK_ENABLE();
+	__HAL_RCC_SYSCFG_CLK_ENABLE();
+	/* This implements workaround described in ES0620
+	 * 2.2.21 I/O compensation may deform output signal rise and fall edges
+	 */
+	SYSCFG->VDDCCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO2CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO3CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO4CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO5CCCR = STM32N6_IOCELL_FIXEDVALUE;
```
mathieuchopstm: If there are `LL` APIs to perform this, please use them.

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,48 @@
...
+        can permanently damage the device! Please consult datasheet for
+        pecise threshold value.
+
+    hslv-auto:
+      type: boolean
+      description: |
+        Similar to hslv-enabled, but th HSLV setting will follow
+        option bytes / OTP fuse configuration.
+        Can be useful for board allowing both 1.8V and 3.3V operation.
+        Otherwise users should avoid using this option.
```
mathieuchopstm: Since `hslv-auto` is:
> Similar to `hslv-enabled`

Maybe it would be more sensible to have:
```yml
hslv-mode:
  default: "off"
  enum:
    - "off"
    - "on"    # = 'hslv-enabled;'
    - "auto"  # = 'hslv-auto;'
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+			DT_PROP(node_id, domain));                                                 \
+	}
+
+static int iocell_check(void){
+	DT_FOREACH_CHILD(STM32_IOCELL_NODE, STM32_IOCELL_CHECK)
+	return 0;
+}
+
+SYS_INIT(iocell_check, POST_KERNEL, CONFIG_STM32_IOCELL_CHECK_PRIORITY);
+#endif
```
mathieuchopstm: Please add EOL comments on `#endif` lines - the logic is hard to follow without them:
```suggestion
#endif /* CONFIG_LOG */
```

drivers/flash/flash_stm32_xspi.c
```diff
@@ -2117,11 +2117,6 @@ static int flash_stm32_xspi_init(const struct device *dev)
 		LOG_ERR("XSPI mode SPI|DUAL|QUAD/DTR is not valid");
 		return -ENOTSUP;
 	}
-#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
-	LL_PWR_EnableXSPIM2();
```
mathieuchopstm: This doesn't seem to have a replacement in new code?

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+	SYSCFG->VDDIO2CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO3CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO4CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO5CCCR = STM32N6_IOCELL_FIXEDVALUE;
+#endif
+	DT_FOREACH_CHILD(STM32_IOCELL_NODE, STM32_IOCELL_INIT)
+	return 0;
+}
+
+SYS_INIT(iocell_init, PRE_KERNEL_1, CONFIG_STM32_IOCELL_PRIORITY);
```
mathieuchopstm: Since you have a proper node in DT, it should be possible to `DEVICE_DT_DEFINE` instead, which is preferred.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+		LOG_ERR(IOCELL_HSLV_DISCLAIMER);
+	} else if (safeguard && !hslv) {
+		LOG_WRN("HSLV allowed in " IOCELL_FUSE_NAME " for domain \"%s\", but not enabled",
+			domain_name);
+		LOG_WRN(IOCELL_HSLV_DISCLAIMER);
+	}
+}
+
+#define STM32_IOCELL_CHECK(node_id)                                                                \
+	if (STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain))) {                              \
```
mathieuchopstm: Consider making this a build-time assertion instead (e.g., `foreach(/* child of iocell controller */, BUILD_ASSERT(STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(<child node>, domain)), "Invalid domain on " <child node>);`)

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+#include <zephyr/dt-bindings/power/stm32h7rs_iocell.h>
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+#include <zephyr/dt-bindings/power/stm32n6_iocell.h>
+#define STM32N6_IOCELL_FIXEDVALUE (0x00000287)
+#endif
+
+#define DT_DRV_COMPAT     st_stm32_iocell
+#define STM32_IOCELL_NODE DT_NODELABEL(iocell)
+
+/* Returns 1 if HSLV can be confgured */
```
mathieuchopstm: The expectation is `, 0 otherwise` but this can return 3 values. Please clarify.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+
+static inline void iocell_enable_compensation(uint16_t domain)
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	/* TODO: check MCU revision */
+
+	/* This implements workaround described in ES0596:
+	 * 2.2.15. I/O compensation could alter duty-cycle of high-frequency output signal
+	 */
+	uint32_t shift = domain;
```
mathieuchopstm: I'm not super keen of such optimizations based on definitions from (1) a different file (2) in `dt-bindings` - they are very fragile. Add a warning in the `dt-bindings` header or find another mechanism.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+			/* Read error => force safe configuration */
+			hconf1 = 0x0FFFFFFF;
+		} else {
+			hconf1 = BSEC->FVRw[124];
+		}
+	}
+
+	if (domain == STM32_DT_IOCELL_VDD) {
+		shift = 17;
+	}
```
mathieuchopstm: Similar remark as below for H7RS wrt optimization.

mathieuchopstm (commented):
First pass, incomplete. But overall the idea seems fine.

drivers/flash/flash_stm32_xspi.c
ABESTM: Yes, this is intentional and it is already done during SoC initialization: [soc/st/stm32/stm32h7rsx/soc.c#L64](https://github.com/zephyrproject-rtos/zephyr/blob/3b23c76a246973fe7b1570447398f3bb1ef434e1/soc/st/stm32/stm32h7rsx/soc.c#L64)
Maybe I should mention it in commit message.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI2_HSLV) == FLASH_OBW1SR_XSPI2_HSLV;
+	} else {
+		return -1;
+	}
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	static uint32_t hconf1 = 0xFFFFFFFF;
+	uint32_t shift = 16 - domain;
+
+	if (hconf1 == 0xFFFFFFFF) {
+		WRITE_REG(BSEC->OTPCR, 124);
```
etienne-lms:
```suggestion
		stm32_reg_write(&BSEC->OTPCR, 124);
```
See 2a52a2b7ff48f0deeadcd3aece7f9064a54b130a.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+
+	/* This implements workaround described in ES0596:
+	 * 2.2.15. I/O compensation could alter duty-cycle of high-frequency output signal
+	 */
+	uint32_t shift = domain;
+	uint32_t nmos, pmos;
+	/* TODO: skip if compensation cell already enabled ? */
+
+	/* Configure Cell selection for domain and enable */
+	MODIFY_REG(SBS->CCCSR, BIT((shift * 2) + 1), BIT(shift * 2));
```
etienne-lms:
```suggestion
	stm32_reg_modify_bits(SBS->CCCSR, BIT((shift * 2) + 1), BIT(shift * 2));
```
See 004c613e25d957c35258132e2d60d1f20dc4b47f.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,218 @@
...
+	if (STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain))) {                              \
+		if (DT_PROP(node_id, hslv_enabled) ||                                              \
+		    (DT_PROP(node_id, hslv_auto) &&                                                \
+		     iocell_get_hslv_safeguard(DT_PROP(node_id, domain)))) {                       \
+			iocell_enable_hslv_runtime(DT_PROP(node_id, domain));                      \
+		}                                                                                  \
+		if (DT_PROP(node_id, compensation_enabled)) {                                      \
+			iocell_enable_compensation(DT_PROP(node_id, domain));                      \
+		}                                                                                  \
+	}
```
etienne-lms: `} else { return -EINVAL; }`?

soc/st/stm32/common/stm32_iocell.c
ABESTM: There is no LL API to use, only `HAL_SYSCFG_ConfigVDDIOCompensationCell`: [stm32cube/stm32n6xx/drivers/src/stm32n6xx_hal.c#L1558](https://github.com/zephyrproject-rtos/hal_stm32/blob/50fcec604b9994f67cace4b4d47ab6c6048c9ac7/stm32cube/stm32n6xx/drivers/src/stm32n6xx_hal.c#L1558)

mathieuchopstm: OK for raw accesses then.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I updated for `iocell_init`, but the `iocell_check` still uses `SYS_INIT`

dts/bindings/power/st,stm32-iocell.yaml
ABESTM: Make sense. I fixed it. Thank you.

soc/st/stm32/common/stm32_iocell.c
ABESTM: Thank you. Updated it.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	} else if (domain == STM32_DT_IOCELL_XSPI2) {
+		return (FLASH->OBW1SR & FLASH_OBW1SR_XSPI2_HSLV) == FLASH_OBW1SR_XSPI2_HSLV;
+	} else {
+		return -EINVAL;
+	}
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	static uint32_t hconf1 = 0xFFFFFFFF;
+
+	if (hconf1 == 0xFFFFFFFF) {
+		WRITE_REG(BSEC->OTPCR, 124);
```
mathieuchopstm: Use the helpers from `#include <stm32_bitops.h>` instead:
```suggestion
		stm32_reg_write(BSEC->OTPCR, 124);
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	if (domain == STM32_DT_IOCELL_VDDIO2) {
+		LL_PWR_SetVddIO2VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+	} else if (domain == STM32_DT_IOCELL_VDDIO3) {
+		LL_PWR_SetVddIO3VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+	} else if (domain == STM32_DT_IOCELL_VDDIO4) {
+		LL_PWR_SetVddIO4VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+	} else if (domain == STM32_DT_IOCELL_VDDIO5) {
+		LL_PWR_SetVddIO5VoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
+	} else if (domain == STM32_DT_IOCELL_VDD) {
+		LL_PWR_SetVddIOVoltageRange(LL_PWR_VDDIO_VOLTAGE_RANGE_1V8);
```
mathieuchopstm: Esthetical nit: `IOCELL_VDD` could come first such that we go from `VddIO` to `VddIO5` in order

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+		shift = 1;
+	} else if (domain == STM32_DT_IOCELL_XSPI2) {
+		shift = 2;
+	} else {
+		return;
+	}
+	/* TODO: skip if compensation cell already enabled ? */
+
+	/* Configure Cell selection for domain and enable */
+	MODIFY_REG(SBS->CCCSR, BIT((shift * 2) + 1), BIT(shift * 2));
```
mathieuchopstm: 1. Use the helpers
```suggestion
	stm32_reg_modify_bits(SBS->CCCSR, BIT((shift * 2) + 1), BIT(shift * 2));
```
2. Maybe the comment could be expanded a bit to indicate that the `mask` and `value` are unrelated (usually, `MODIFY_REG` is used to replace a field's value, whereas here it's used to clear one bit and set another at the same time, which is slightly surprising/confusing)

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+
+	/* Wait for read flag */
+	while ((SBS->CCCSR & BIT(shift + 8)) == 0) {
+	}
+
+	/* Read and adjust PMOS and NMOS values*/
```
mathieuchopstm, line 210: Could you add a reference to the documentation from which this algorithm is sourced?

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	/* Write modified values back */
+	MODIFY_REG(SBS->CCSWVALR, 0xFF << (shift * 8), ((pmos << 4) | nmos) << (shift * 8));
+
+	/* Configure Cell selection for register */
+	SET_BIT(SBS->CCCSR, BIT((shift * 2) + 1));
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* There is a global workaround implemnted in iocell_init.
+	 * For N6 IO compoensation cell is always active.
+	 */
+	(void)domain;
```
mathieuchopstm: In Zephyr, prefer instead:
```suggestion
	ARG_UNUSED(domain);
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	/* Read and adjust PMOS and NMOS values*/
+	nmos = (SBS->CCVALR >> (8 * shift)) & 0xF;
+	pmos = (SBS->CCVALR >> (8 * shift + 4)) & 0xF;
+	nmos = (nmos < 0xD) ? (nmos + 2) : 0xF;
+	pmos = (pmos > 0x2) ? (pmos - 2) : 0x0;
+
+	/* Write modified values back */
+	MODIFY_REG(SBS->CCSWVALR, 0xFF << (shift * 8), ((pmos << 4) | nmos) << (shift * 8));
```
mathieuchopstm:
```suggestion
	stm32_reg_modify_bits(SBS->CCSWVALR, 0xFF << (shift * 8), ((pmos << 4) | nmos) << (shift * 8));
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+
+	/* Configure Cell selection for register */
+	SET_BIT(SBS->CCCSR, BIT((shift * 2) + 1));
```
mathieuchopstm:
```suggestion
	stm32_reg_set_bits(SBS->CCCSR, BIT((shift * 2) + 1));
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	/* This implements workaround described in ES0620
+	 * 2.2.21 I/O compensation may deform output signal rise and fall edges
+	 */
+	SYSCFG->VDDCCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO2CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO3CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO4CCCR = STM32N6_IOCELL_FIXEDVALUE;
+	SYSCFG->VDDIO5CCCR = STM32N6_IOCELL_FIXEDVALUE;
+#endif /* CONFIG_SOC_SERIES_... */
+	(void)dev;
```
mathieuchopstm:
```suggestion
	ARG_UNUSED(dev);
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+}
+
+#define STM32_IOCELL_CHECK(node_id)                                                                \
+	BUILD_ASSERT(STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain)),                       \
+		     "Invalid domain ID for: " DT_NODE_FULL_NAME(node_id));                        \
+	iocell_hslv_check_log(                                                                     \
+		DT_PROP(node_id, domain), iocell_get_hslv_safeguard(DT_PROP(node_id, domain)),     \
+		DT_ENUM_HAS_VALUE(node_id, hslv_mode, on), DT_NODE_FULL_NAME(node_id));
+
+#ifdef CONFIG_LOG
```
mathieuchopstm: I'd create a dedicated Kconfig option (e.g., `CONFIG_STM32_IOCELL_CHECK_ENABLE`) instead

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+child-binding:
+  description: |
+    IO power domains
+
+  properties:
+    domain:
+      type: int
+      description: |
+        Domain e.g., STM32_DT_IOCELL_VDDIO
```
mathieuchopstm: Should be a bit more verbose, e.g.:
```suggestion
      description: |
        I/O domain managed by the cell
        
        This must be one of the SoC-specific values defined in header
        `include/zephyr/dt-bindings/power/stm32XXX_iocell.h`.
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	(void)domain;
+#endif /* CONFIG_SOC_SERIES_... */
+}
+
+#define STM32_IOCELL_INIT(node_id)                                                                 \
+	BUILD_ASSERT(STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain)),                       \
+		     "Invalid domain ID for: " DT_NODE_FULL_NAME(node_id));                        \
+	if (DT_ENUM_HAS_VALUE(node_id, hslv_mode, on) ||                                           \
+	    (DT_ENUM_HAS_VALUE(node_id, hslv_mode, auto) &&                                        \
+	     iocell_get_hslv_safeguard(DT_PROP(node_id, domain)))) {                               \
```
mathieuchopstm: 1. this checks for non-zero value, but I think `<0` should NOT result in HSLV being enabled?
2. considering its usage, maybe `iocell_is_hslv_enable_safe` might be a more fitting name

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,48 @@
...
+      description: |
+        Allows to turn on HSLV (high-speed low voltage) option.
+        This option will work only if the corresponding bit is set in
+        option bytes / OTP fuse.
+        Enabling this option on domain that is powered from 3.3V
+        can permanently damage the device! Please consult datasheet for
+        pecise threshold value.
+        Selecting "auto" will follow the setting in option bytes / OTP.
+        This can be useful for board allowing both 1.8V and 3.3V operation.
+        Otherwise users should avoid using this option.
```
etienne-lms:
```suggestion
        Otherwise users should use default "off" mode.

File: dts/bindings/power/st,stm32-iocell.yaml [hunk 26]
```diff
@@ -0,0 +1,48 @@
...
+  description: |
+    IO power domains
+
+  properties:
+    domain:
+      type: int
+      description: |
+        Domain e.g., STM32_DT_IOCELL_VDDIO
+
+    hslv-mode:
```
etienne-lms, line 37: Prefer with an `st,` prefix since this is an ST specific property.
Ditto for `compensation-enabled`.

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+
+#include <zephyr/dt-bindings/dt-util.h>
+
+#define STM32_DT_IOCELL_VDDIO (0)
+#define STM32_DT_IOCELL_XSPI1 (1)
+#define STM32_DT_IOCELL_XSPI2 (2)
+
+#define STM32_DT_IOCELL_DOMAIN_VALID(x) ((x >= 0) && (x <= 2))
```
etienne-lms:
```suggestion
#define STM32_DT_IOCELL_DOMAIN_VALID(x) (((x) >= 0) && ((x) <= 2))
```
or even:
```suggestion
#define STM32_DT_IOCELL_DOMAIN_VALID(x)	IN_RANGE((x), 0, 2)
```
(`#include <sys/util.h>`)

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+
+#include <zephyr/dt-bindings/dt-util.h>
```
etienne-lms: Not neded.

soc/st/stm32/common/Kconfig
```diff
@@ -57,3 +57,24 @@ config STM32_BACKUP_PROTECTION
 	help
 	  Enabled for SoCs for which access protection to backup domain
 	  resources needs to be explicitly handled.
+
+config STM32_IOCELL
+	bool "STM32 I/O cell"
+	default y
+	depends on  DT_HAS_ST_STM32_IOCELL_ENABLED
```
etienne-lms:
```suggestion
	depends on DT_HAS_ST_STM32_IOCELL_ENABLED

File: soc/st/stm32/common/stm32_iocell.c [hunk 30]
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright (c) 2024 STMicroelectronics
```
etienne-lms: 2025

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(iocell, CONFIG_SOC_LOG_LEVEL);
+
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+#include <zephyr/dt-bindings/power/stm32h7rs_iocell.h>
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+#include <zephyr/dt-bindings/power/stm32n6_iocell.h>
+#define STM32N6_IOCELL_FIXEDVALUE (0x00000287)
```
etienne-lms: Can drop parentheses.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+#endif /* CONFIG_SOC_SERIES_... */
+
+#define DT_DRV_COMPAT     st_stm32_iocell
+#define STM32_IOCELL_NODE DT_NODELABEL(iocell)
+
+/* Returns 1 if HSLV can be configured
+ * Returns 0 if HSLV can't be configured
+ * Returns < 0 on error
+ */
+static inline int iocell_get_hslv_safeguard(uint16_t domain)
```
etienne-lms: Can remove the `inline` keywords.

soc/st/stm32/common/stm32_iocell.c
etienne-lms: With an `&` operator:
```patch
-		stm32_reg_write(BSEC->OTPCR, 124);
+		stm32_reg_write(&BSEC->OTPCR, 124);
```

Ditto for `stm32_reg_modify_bits()` and `stm32_reg_set_bits()` in review comments below.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+{
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	if (domain == STM32_DT_IOCELL_VDDIO) {
```
etienne-lms: I'd suggest to use `switch()`/`case` instructions.
Ditto in other occurrences below.

soc/st/stm32/common/stm32_iocell.c
etienne-lms: Nitpicking: prefer located at function entry.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+		     "Invalid domain ID for: " DT_NODE_FULL_NAME(node_id));                        \
+	iocell_hslv_check_log(                                                                     \
+		DT_PROP(node_id, domain), iocell_get_hslv_safeguard(DT_PROP(node_id, domain)),     \
+		DT_ENUM_HAS_VALUE(node_id, hslv_mode, on), DT_NODE_FULL_NAME(node_id));
+
+#ifdef CONFIG_LOG
+
+static int iocell_check(void)
+{
+	DT_FOREACH_CHILD(STM32_IOCELL_NODE, STM32_IOCELL_CHECK)
```
etienne-lms: `DT_FOREACH_CHILD_STATUS_OKAY()` in case one explicitly disables a IOCELL node?

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+		LOG_WRN("HSLV allowed in " IOCELL_FUSE_NAME " for domain \"%s\", but not enabled",
+			domain_name);
+		LOG_WRN(IOCELL_HSLV_DISCLAIMER);
+	} else {
+		/* Everything ok */
+	}
+}
+
+#define STM32_IOCELL_CHECK(node_id)                                                                \
+	BUILD_ASSERT(STM32_DT_IOCELL_DOMAIN_VALID(DT_PROP(node_id, domain)),                       \
```
etienne-lms, line 254: I don't think it's needed in `STM32_IOCELL_CHECK()` since already addressed by `STM32_IOCELL_INIT()`.

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,248 @@
...
+	"if the corresponding power supply is above certain threshold. "                           \
+	"Enabling it while the voltage is higher can damage the device"
+
+__maybe_unused static void iocell_hslv_check_log(int domain_id, int safeguard, int hslv,
+						 const char *domain_name)
+{
+	if (safeguard < 0) {
+		/* This shouldn't happen with proper implementation */
+		LOG_ERR("Failed to read safeguard value of HSLV: \"%s\" (%d)", domain_name,
+			domain_id);
```
etienne-lms: Failing here should panic the system, no?

etienne-lms (commented):
Nitpicking on 2 commit message header lines: fix `h7s78` typo + add `st: ` prefix.
```patch
-boards: stm32hs78_dk: Enable HSLV and IO compensation
+boards: st: stm32h7s78_dk: Enable HSLV and IO compensation
```
```patch
-boards: stm32n6570_dk: Enable HSLV and IO compensation
+boards: st: stm32n6570_dk: Enable HSLV and IO compensation
```

dts/bindings/power/st,stm32-iocell.yaml
mathieuchopstm: I'm not sure as to whether a vendor prefix makes sense here. This is a vendor-specific binding anyways. (i.e., why `st,hslv-mode` but not `st,domain`?)

soc/st/stm32/common/stm32_iocell.c
ABESTM: This is redundant at the moment, but I think it might be better to keep it there for future modifications.

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
ABESTM: Good point.
Maybe better to avoid `IN_RANGE` since `#include <sys/util.h>` causes error in Device tree context.

soc/st/stm32/common/stm32_iocell.c
ABESTM: All bit operations should be moved to `stm32_reg_*` helpers.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I added a comment and also define temporary variables, so it should be easier to understand.

soc/st/stm32/common/stm32_iocell.c
ABESTM: The specific errata is referenced at the beginning of H7RS-specific code:
```
	/* This implements workaround described in ES0596:
	 * 2.2.15. I/O compensation could alter duty-cycle of high-frequency output signal
	 */
```
Do you think there should be more detailed explanation?

soc/st/stm32/common/stm32_iocell.c
ABESTM: I reworked that part. I think it was because the code for one MCU uses the parameter, but for another doesn't.

soc/st/stm32/common/stm32_iocell.c
ABESTM: Added the new option `CONFIG_STM32_IOCELL_CHECK_ENABLE`. I also moved the `#ifdef` at the end of file (just before `SYS_INIT`), so the functions are always compiled and checked, but not sure if this is good approach.

soc/st/stm32/common/stm32_iocell.c
ABESTM: 1. Yes, you are right, I now check against the `1` directly.
2. Yes, this might be good idea (but still on my TODO).

soc/st/stm32/common/stm32_iocell.c
ABESTM: This was removed since handled by: https://github.com/zephyrproject-rtos/hal_stm32/pull/331

soc/st/stm32/common/stm32_iocell.c
ABESTM: It should be now always switch/case for IO domain

soc/st/stm32/common/stm32_iocell.c
ABESTM: I switched to `DT_FOREACH_CHILD_STATUS_OKAY`. I don't think it is a good idea to disable the domain via `status`, so I thinks we should avoid doing this way in the board DT and just modify the parameters instead.
But I guess this it is expected behavior for user when node is disabled.

soc/st/stm32/common/Kconfig.defconfig
```diff
@@ -0,0 +1,9 @@
+if SOC_SERIES_STM32N6X && DT_HAS_ST_STM32_IOCELL_ENABLED
```
erwango: If SoC specific, then it should be moved to soc/stm32n6xx. Keep it here if not functional otherwise (and then please add header).

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,291 @@
...
+	/* TODO: maybe access stm32_iocell_config/data directly ? */
+	__maybe_unused const struct stm32_iocell_config_type *config = dev->config;
+	__maybe_unused struct stm32_iocell_data_type *data = dev->data;
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+	__HAL_RCC_SBS_CLK_ENABLE();
+	/* CSI is required by IO compensation cell */
+	__HAL_RCC_CSI_ENABLE();
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+	/* PWR clocks enabled by default */
+	__HAL_RCC_BSEC_CLK_ENABLE();
```
erwango: Not already done in bsec driver ?

ABESTM: I removed this, since this should be handled by BSEC driver.
Currently there is no such call in BSEC driver, since the clocks are enabled by default after reset.

dts/arm/st/n6/stm32n6.dtsi
```diff
@@ -750,13 +779,18 @@
 			compatible = "st,stm32-bsec";
 			reg = <0x56009000 0x1000>;
 			st,upper-fuse-limit = <256>;
-			status = "disabled";
+			status = "okay";
```
JarmouniA: Devices are enabled at board level.

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,58 @@
...
+    const: 0
+
+  "#size-cells":
+    const: 0
+
+  nvmem-cells:
+    description: |
+      Single nvmem-cells entry named "safeguard" can be provided to access
+      OTP/option byte value where HSLV enable bit is stored.
+      This is optional depending on the STM32 family.
```
JarmouniA:
```suggestion
      This is optional depending on the SoC series.
```

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,58 @@
...
+        - "off"
+        - "on"
+        - "auto"
+      description: |
+        Allows to turn on HSLV (high-speed low voltage) option.
+        This option will work only if the corresponding bit is set in
+        option bytes / OTP fuse.
+        Enabling this option on domain that is powered from 3.3V
+        can permanently damage the device! Please consult datasheet for
+        pecise threshold value.
```
JarmouniA:
```suggestion
        precise threshold value.
```

dts/bindings/power/st,stm32-iocell.yaml
```diff
@@ -0,0 +1,58 @@
...
+        Selecting "auto" will follow the setting in option bytes / OTP.
+        This can be useful for board allowing both 1.8V and 3.3V operation.
+        Otherwise users should use default "off" mode.
```
JarmouniA:
```suggestion
```
it's the default anyway

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+
+#define STM32_DT_IOCELL_VDDIO (0)
+#define STM32_DT_IOCELL_XSPI1 (1)
+#define STM32_DT_IOCELL_XSPI2 (2)
+
```
JarmouniA, line 26: Please document what these constants are and where they are supposed to be used.

include/zephyr/dt-bindings/power/stm32n6_iocell.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+
+#define STM32_DT_IOCELL_VDDIO2 (0)
+#define STM32_DT_IOCELL_VDDIO3 (1)
+#define STM32_DT_IOCELL_VDDIO4 (2)
+#define STM32_DT_IOCELL_VDDIO5 (3)
+#define STM32_DT_IOCELL_VDD    (4)
+
```
JarmouniA, line 28: ditto

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,290 @@
...
+
+#if defined(CONFIG_SOC_SERIES_STM32H7RSX)
+#include <zephyr/dt-bindings/power/stm32h7rs_iocell.h>
+#elif defined(CONFIG_SOC_SERIES_STM32N6X)
+#include <zephyr/dt-bindings/power/stm32n6_iocell.h>
+#include <zephyr/nvmem.h>
+#endif /* CONFIG_SOC_SERIES_... */
+
+#define DT_DRV_COMPAT     st_stm32_iocell
+#define STM32_IOCELL_NODE DT_NODELABEL(iocell)
```
JarmouniA:
```suggestion
#define STM32_IOCELL_NODE DT_DRV_INST(0)
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,290 @@
...
+
+struct stm32_iocell_config_type {
```
JarmouniA:
```suggestion
struct stm32_iocell_config {
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,290 @@
...
+#if defined(CONFIG_NVMEM)
+	struct nvmem_cell cell;
+#endif
+};
+
+struct stm32_iocell_data_type {
```
JarmouniA:
```suggestion
struct stm32_iocell_data {
```

soc/st/stm32/stm32n6x/Kconfig.defconfig
```diff
@@ -30,4 +30,14 @@ config STM32_LTDC_FB_SMH_ALIGN
 config ROM_START_OFFSET
 	default 0x400 if BOOTLOADER_MCUBOOT
 
+if DT_HAS_ST_STM32_IOCELL_ENABLED
```
JarmouniA:
```suggestion
if STM32_IOCELL
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,290 @@
...
+	if (nvmem_cell_read(&config->cell, &data->otp_hconf1_value, 0, 4) != 0) {
+		data->otp_hconf1_value = 0xFFFFFFFF;
+		data->error_flag = -1;
+	}
+#endif /* CONFIG_SOC_SERIES_... */
+	DT_FOREACH_CHILD_STATUS_OKAY(STM32_IOCELL_NODE, STM32_IOCELL_INIT)
+	return 0;
+}
+
+DEVICE_DT_DEFINE(STM32_IOCELL_NODE, iocell_init, NULL, &stm32_iocell_data, &stm32_iocell_config,
```
JarmouniA: No point in creating a globally visible device object here, as no external code is excepted to get a pointer it.

JarmouniA (dismissed):

soc/st/stm32/common/stm32_iocell.c
ABESTM: @JarmouniA I used `SYS_INIT` before, but @mathieuchopstm recommended to use `DEVICE_DT_DEFINE`. To me it seems the `DEVICE_DT_DEFINE` might be more future proof, when new SoCs or features need to be implemented.

dts/arm/st/n6/stm32n6.dtsi
ABESTM: The idea was to have `iocell` driver enabled by default and on STM32N6 it requires `bsec` to be enabled as well.
So, if we enable `bsec` at board level, we should do the same for `iocell` driver.

JarmouniA: I am talking about DT device nodes, you are mixing them with drivers, I don't see where is the issue.

> it requires bsec to be enabled as well.

That dependency can be expressed in Kconfig.

soc/st/stm32/common/stm32_iocell.c
JarmouniA: > To me it seems the DEVICE_DT_DEFINE might be more future proof

Defining all functions as non-static is also more future proof... that's a non-argument in my opinion.
This code is not offering any API, so what's the point of exposing a global device object.

dts/arm/st/n6/stm32n6.dtsi
GseoC: > That dependency can be expressed in Kconfig.

Using `DT_HAS_ST_STM32_BSEC_ENABLED`

ABESTM: You are right, I was referring to `iocell` node which is currently enabled as SoC level (`status = "okay";`).
At the end, I think it might be better to have the `iocell` DT node enabled at board level (and `bsec` as well).
Unless I miss something, the Kconfig won't help, since the `iocell` DT node references `nvmem-cell` from `bsec` and if that node is disabled, the compilation will fail.

JarmouniA: If iocell cannot actually be disabled in hw, you could just block the status property from its binding, and remove the dependency of driver on `DT_HAS_ST_STM32_IOCELL_ENABLED`.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I modified the function name to be more descriptive.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I set all copyright of new files to 2026

soc/st/stm32/common/stm32_iocell.c
mathieuchopstm: That would be sufficient.

soc/st/stm32/common/stm32_iocell.c
mathieuchopstm: > This code is not offering any API, so what's the point of exposing a global device object.

This driver depends on other drivers (i.e., OTP). As far as I know, `SYS_INIT` performs no dependency checking whereas using a proper device ensures `iocell` is always initialized after its dependencies (or build fails).

soc/st/stm32/stm32n6x/Kconfig.defconfig
```diff
@@ -30,4 +30,14 @@ config STM32_LTDC_FB_SMH_ALIGN
...
+
+config OTP
+	default y
+
+config NVMEM
+	default y
```
mathieuchopstm:
```suggestion
configdefault OTP
	default y

configdefault NVMEM
	default y
```

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,290 @@
...
+	if (data->error_flag < 0) {
+		LOG_ERR("Error during initialization");
+	}
+	DT_FOREACH_CHILD_STATUS_OKAY(STM32_IOCELL_NODE, STM32_IOCELL_CHECK)
+	return 0;
+}
+
+#ifdef CONFIG_STM32_IOCELL_CHECK_ENABLE
+
+SYS_INIT(iocell_check, POST_KERNEL, CONFIG_STM32_IOCELL_CHECK_PRIORITY);
```
mathieuchopstm: Still not really a fan of this, why not add it to the main `init` function? because it runs "too early"?

soc/st/stm32/common/stm32_iocell.c
ABESTM: > > This code is not offering any API, so what's the point of exposing a global device object.
> 
> This driver depends on other drivers (i.e., OTP). As far as I know, `SYS_INIT` performs no dependency checking whereas using a proper device ensures `iocell` is always initialized after its dependencies (or build fails).

Perhaps a bit off topic, but currently the check doesn't work properly for `iocell` and `bsec`, since it doesn't refer directly to `bsec` node, but to its nested child. And since the script doesn't detect any initialization level for the child node, it passes the check.
I think I encountered the same issue some time ago with camera sensor in #96799, but I didn't know there is automatic check that should work.
I did some quick modification of the python script to propagate the initialization priority to child nodes recursively. But maybe this could bring some other issues, and it is different topic/discussion. I will try to create PR for that.

soc/st/stm32/common/stm32_iocell.c
ABESTM: Somehow, I thought that logging can't be used in `PRE_KERNEL_1` initialization level, but it seems to work.
So, I will fix this.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I'm not sure if it makes sense here (although I added it in latest version). This code performs check of the runtime configuration vs. OTP/option byte. The error here means it can't perform the check itself, but it might still be correct.
Maybe it makes sense to panic when "auto" option is selected, which should follow OTP/option byte. But even in such case it can just not configure HSLV and report error in log.

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
ABESTM: I added comment in latest update.

soc/st/stm32/common/stm32_iocell.c
ABESTM: I reworked the driver and removed the separate initialization level for the check. The `CONFIG_STM32_IOCELL_CHECK_ENABLE` Kconfig is still there.

include/zephyr/dt-bindings/power/stm32h7rs_iocell.h
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32H7RS_IOCELL_H_
+
+/** @brief I/O power domains that can be configured in iocell driver */
```
JarmouniA:
```suggestion
/** @brief I/O power domains used to configure STM32 iocell device */
```

include/zephyr/dt-bindings/power/stm32n6_iocell.h
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) 2026 STMicroelectronics
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+#define ZEPHYR_INCLUDE_DT_BINDINGS_POWER_STM32N6_IOCELL_H_
+
+/** @brief I/O power domains that can be configured in iocell driver */
```
JarmouniA: ditto

soc/st/stm32/common/Kconfig
```diff
@@ -51,3 +51,26 @@ config STM32_BACKUP_PROTECTION
 	help
 	  Enabled for SoCs for which access protection to backup domain
 	  resources needs to be explicitly handled.
+
+config STM32_IOCELL
+	bool "STM32 I/O cell"
+	default y
+	depends on DT_HAS_ST_STM32_IOCELL_ENABLED
```
JarmouniA, line 58: Shouldn't the dependency on bsec be expressed here?

soc/st/stm32/common/stm32_iocell.c
ABESTM: Regarding my previous comment, I created RFC for this issue: #103093

soc/st/stm32/common/Kconfig
ABESTM: The iocell depends on bsec only on STM32N6 devices. On STM32H7R/S there is no dependency.
Maybe this can be somehow expressed in the N6-specific Kconfig, but I'm not sure how.

mathieuchopstm (dismissed):

boards/st/nucleo_n657x0_q/Kconfig.defconfig
```diff
@@ -8,4 +8,7 @@ if BOARD_NUCLEO_N657X0_Q
 configdefault NET_L2_ETHERNET
 	default y
 
+configdefault OTP
+	default y if NET_L2_ETHERNET && NVMEM
```
etienne-lms: `NET_L2_ETHERNET` may not be the best symbol. It use in N6-DK board hurts a build_all test and would need to be fixed. @GseoC, do you know if `ETH_DRIVER`, or maybe `ETH_STM32_HAL` would better apply?

etienne-lms (commented):
Question to @GseoC, otherwise LGTM.

boards/st/nucleo_n657x0_q/Kconfig.defconfig
GseoC: Hi, yes, better with `ETH_STM32_HAL`

ABESTM: Since I copied this from stm32n6570_dk board, it would make sense to modify its `Kconfig.defconfig` as well. Or is there already separate PR?
What is the test that fails? I would like to test before I push the change.

juickar: Hello @ABESTM, the opened PR is [here](https://github.com/zephyrproject-rtos/zephyr/pull/103236)

ABESTM: Hello @juickar thank you for the update. So I will modify only the Nucleo code in this PR, so it will not collide with #103236

mathieuchopstm (approved):

dts/arm/st/n6/stm32n6.dtsi
```diff
@@ -757,6 +789,11 @@
 				#address-cells = <1>;
 				#size-cells = <1>;
 
+				otp124_hconf1: opt124_hconf1@1f0 {
```
etienne-lms, line 792: Preef `-` over `_` in node names.
```suggestion
				otp124_hconf1: opt124-hconf1@1f0 {

File: soc/st/stm32/common/stm32_iocell.c [hunk 57]
```diff
@@ -0,0 +1,300 @@
...
+#if defined(CONFIG_SOC_SERIES_STM32N6X)
+#define IOCELL_FUSE_NAME "OTP fuse"
+#else
+#define IOCELL_FUSE_NAME "option bit"
+#endif /* CONFIG_SOC_SERIES_... */
+
+#define IOCELL_HSLV_DISCLAIMER                                                                     \
+	"The I/O HSLV configuration must not be enabled "                                          \
+	"if the corresponding power supply is above certain threshold. "                           \
+	"Enabling it while the voltage is higher can damage the device"
```
etienne-lms, line 43: Nitpicking: add a terminal dot `.`?

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,300 @@
...
+struct stm32_iocell_config {
+#if defined(STM32_IOCELL_USE_NVMEM)
+	struct nvmem_cell cell;
+#endif
+};
+
+struct stm32_iocell_data {
+#if defined(CONFIG_SOC_SERIES_STM32N6X)
+	uint32_t otp_hconf1_value;
+	int otp_correct;
```
etienne-lms, line 54: This seeems a boolean value, use `bool`?

soc/st/stm32/common/stm32_iocell.c
```diff
@@ -0,0 +1,300 @@
...
+				      DT_ENUM_HAS_VALUE(node_id, hslv_mode, on),                   \
+				      DT_NODE_FULL_NAME(node_id));                                 \
+	}                                                                                          \
+	if (result < 0)                                                                            \
+		ret = result;
+
+static int iocell_init(const struct device *dev)
+{
+	ARG_UNUSED(dev);
+	int ret = 0, result = 0;
```
etienne-lms, line 279: Could you add an empty line below (local variables definition)?

etienne-lms (commented):
Some minor comments. Non-blocking.

erwango (approved):

PR #100544, opened 2025-12-04
modules: Update trusted-firmware-m for picolibc

keith-packard:
This adds linker script bits and compiler options so that trusted-firmware-m will build with picolibc.

This has been merged to the Zephyr trusted-firmware-m repository.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -380,7 +380,7 @@ manifest:
       groups:
         - tee
     - name: trusted-firmware-m
-      revision: 9889df41022c6679b71da55d93f9e3693a804976
+      revision: fe2b9109fe5de8c4e09f3ccc502a64429b8d45dd
       path: modules/tee/tf-m/trusted-firmware-m
       groups:
         - tee


tpambor (approved):

tomi-font (approved):
About leaving out https://github.com/zephyrproject-rtos/tf-m-tests/pull/18. Do you foresee any scenario where it might be needed? Even though twister passes without it, maybe it wouldn't hurt to still have it in case (what do you think)?

keith-packard:
> About leaving out [zephyrproject-rtos/tf-m-tests#18](https://github.com/zephyrproject-rtos/tf-m-tests/pull/18). Do you foresee any scenario where it might be needed? Even though twister passes without it, maybe it wouldn't hurt to still have it in case (what do you think)?

Yes, it would be useful if Zephyr stopped placing `-DNDEBUG` on the command line when building tests. As this PR is the only thing blocking SDK 1.0 release, I'm mostly focused on reducing the amount of work necessary to finish it up, so removing that change from this PR seemed like a good plan.

dkalowsk (approved):

PR #100553, opened 2025-12-05
RFC: I3C: Common: Allow I2C devices to occupy reserved I3C addresses

cujomalainey:
In the I3C specification it states that the addressing restrictions are for Dynamic Addressing. I2C devices are statically addressed and this would not be backwards compatible if we apply the same I3C addressing restrictions to attaching I2C addresses.


XenuIsWatching:
Table 5 addresses this in Note 5 of the i3c v1.2 spec (as there was additional clarification on this specific topic added in v1.2)
image

Reserved addresses
image

I assume you have a I2C target that uses a reserved address? You may be able to open *some* of those up as not reserved based on section 4.3.2.2.5

PR #100559, opened 2025-12-05
RFC: Wakeup Controller (WUC) Driver

### LINKED ISSUE #100853: RFC: Wakeup Controller (WUC) Driver and Power Management Integration
### Problem Description

In some MCUs, waking up from low-power modes requires special configuration of a dedicated peripheral called the Wakeup Controller (WUC). Currently, Zephyr supports this functionality in two ways:
1. Implemented in interrupt controllers (e.g., [wuc_ite_it51xxx.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/interrupt_controller/wuc_ite_it51xxx.c)):
   This approach lacks a unified interface for WUC control. For example, if different devices share the same peripheral but use different WUC implementations, supporting a new device often requires modifying its driver (e.g.,[espi_it8xxx2.c](https://github.com/zephyrproject-rtos/zephyr/blob/fd63bf93a5a18bf0dc0a18b67b821bfdd68ab5b9/drivers/espi/espi_it8xxx2.c#L2684C1-L2688C7)).
2. Implemented in SoC-specific power management code(e.g., [power.c](https://github.com/zephyrproject-rtos/zephyr/blob/fd63bf93a5a18bf0dc0a18b67b821bfdd68ab5b9/soc/nxp/mcx/mcxw/mcxw7xx/power.c#L25)):  
   This method only supports a limited set of peripherals and lacks flexibility. If an application needs to use an unsupported peripheral as a wakeup source, additional code changes are required, and developers must be familiar with the MCU internals.

**Both approaches have common issues:**
WUC configuration is typically done during initialization, with no flexible API for user-space control.
The relationship between peripherals and WUC is hidden in code rather than expressed in devicetree, reducing transparency and flexibility.

### Proposed Change (Summary)

- Introduce a generic WUC device driver with APIs that can be reused by other drivers, reducing the need for driver-specific changes when supporting new SoCs.
- Allow user-space applications to control WUC with existing API: `pm_device_wakeup_enable()`.
- Represent the peripheral-to-WUC mapping in devicetree instead of hardcoding it in source code.

### Proposed Change (For details please refer to #100559)
The proposal introduces a new Wakeup Controller (WUC) driver API and integrates it with Zephyr’s power management subsystem.
[New WUC Driver API](https://github.com/nxp-upstream/zephyr/blob/create-wakeup-controller/include/zephyr/drivers/wuc.h)
```c
typedef int (*wuc_api_enable_wakeup_source)(const struct device *dev, uint32_t source);

typedef int (*wuc_api_disable_wakeup_source)(const struct device *dev, uint32_t source);

typedef int (*wuc_api_check_wakeup_triggered)(const struct device *dev, uint32_t source);

__subsystem struct wuc_driver_api {
    wuc_api_enable_wakeup_source enable_wakeup_source;
    wuc_api_disable_wakeup_source disable_wakeup_source;
    wuc_api_check_wakeup_triggered check_wakeup_triggered;
};
```

**Function Parameter Explanation**
`const struct device *dev:` 
- Points to the current WUC (Wakeup Controller) device instance
- This is a pointer to the device structure that represents the specific WUC peripheral being operated on
`uint32_t source:`
- Represents the actual wakeup source identifier used within the WUC peripheral
- This is a hardware-specific value that can have different meanings depending on the WUC peripheral implementation
- Different WUC peripherals can interpret this value differently based on their hardware design
- `source` is a uint32_t mask value whose bit fields are defined by the specific WUC implementation, but it must comply with the following general constraints to ensure consistency across SoCs.

**PM Integration:** #100859 

Introduce `CONFIG_PM_WAKEUP_CONTROLLER` Kconfig option to enable this feature.
```
config PM_WAKEUP_CONTROLLER
    bool "Wakeup Controller Support"
    default y
    depends on WUC
    help
      Enable support for Wakeup Controllers. When enabled, devices can be
      associated with a wakeup controller to manage wakeup events.
```

Add pm_wuc_config structure to store WUC device reference and source ID.
```c
#if defined(CONFIG_PM_WAKEUP_CONTROLLER) || defined(__DOXYGEN__)
struct pm_wuc_config {
    /** Wakeup controller device */
    const struct device *wuc;
    /** Wakeup source argument used in wakeup controller */
    uint32_t arg;
};
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
```

Extend `pm_device_base` structure with wakeup controller count and configuration array.
```c
struct pm_device_base {
    /** Device PM status flags. */
    atomic_t flags;
    /** Device power state */
    enum pm_device_state state;
    /** Device PM action callback */
    pm_device_action_cb_t action_cb;
#if defined(CONFIG_PM_DEVICE_RUNTIME) || defined(__DOXYGEN__)
    /** Device usage count */
    uint32_t usage;
#endif /* CONFIG_PM_DEVICE_RUNTIME */
#ifdef CONFIG_PM_DEVICE_POWER_DOMAIN
    /** Power Domain it belongs */
    const struct device *domain;
#endif /* CONFIG_PM_DEVICE_POWER_DOMAIN */
#if defined(CONFIG_PM_WAKEUP_CONTROLLER) || defined(__DOXYGEN__)
    uint8_t wuc_count;
    const struct pm_wuc_config *wuc_configs;
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
};
```
```c
#ifdef CONFIG_PM_WAKEUP_CONTROLLER
#define Z_PM_DEVICE_WAKEUP_CTRL_INIT(_node_id)          \
    .wuc_count = DT_PROP_LEN_OR(_node_id, wakeup_ctrls, 0), \
    .wuc_configs = (const struct pm_wuc_config [])      \
        COND_CODE_1(DT_NODE_HAS_PROP(_node_id, wakeup_ctrls),       \
            ({DT_FOREACH_PROP_ELEM_SEP(_node_id, wakeup_ctrls,  \
                Z_PM_WUC_CFG, (,))}),           \
            ({})),

#define Z_PM_WUC_CFG(node_id, prop, idx)                \
    {                                   \
        .wuc = DEVICE_DT_GET(DT_PHANDLE_BY_IDX(node_id, prop, idx)),    \
        .arg = DT_PHA_BY_IDX(node_id, prop, idx, source),       \
    }
#else
#define Z_PM_DEVICE_WAKEUP_CTRL_INIT(obj)
#endif  /* CONFIG_PM_WAKEUP_CONTROLLER */

#define Z_PM_DEVICE_BASE_INIT(obj, node_id, pm_action_cb, _flags)        \
    {                                    \
        .flags = ATOMIC_INIT(Z_PM_DEVICE_FLAGS(node_id) | (_flags)), \
        .state = PM_DEVICE_STATE_ACTIVE,                 \
        .action_cb = pm_action_cb,                   \
        Z_PM_DEVICE_POWER_DOMAIN_INIT(node_id)               \
        Z_PM_DEVICE_WAKEUP_CTRL_INIT(node_id)            \
    }
```

Add devicetree binding for wakeup-ctrls property in pm.yaml to associate devices with WUC.
```yaml
wakeup-ctrls:
    type: phandle-array
    description: |
      Phandle to the wakeup controller used by this device.
  
  '#wakeup-ctrl-cells':
    type: int
    const: 1
    description: |
      Number of cells in wakeup-ctrl property.
```

Update pm_device_wakeup_enable() to configure wakeup sources automatically when device wakeup is enabled or disabled.
```c
bool pm_device_wakeup_enable(const struct device *dev, bool enable)
{
    atomic_val_t flags, new_flags;
    struct pm_device_base *pm = dev->pm_base;
#if CONFIG_PM_WAKEUP_CONTROLLER
    const int wuc_count = pm->wuc_count;
    const struct pm_wuc_config *wuc_configs = pm->wuc_configs;
    int ret = 0;
    int i = 0;
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
    if (pm == NULL) {
        return false;
    }

    flags = atomic_get(&pm->flags);

    if ((flags & BIT(PM_DEVICE_FLAG_WS_CAPABLE)) == 0U) {
        return false;
    }

    if (enable) {
#if CONFIG_PM_WAKEUP_CONTROLLER
        if ((wuc_count > 0) && (wuc_configs != NULL)) {
            for (i = 0; i < wuc_count; i++) {
                ret = wuc_enable_wakeup_source(wuc_configs[i].wuc, wuc_configs[i].arg);
                if (ret != 0) {
                    return false;
                }
            }
        }
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
        new_flags = flags |
                BIT(PM_DEVICE_FLAG_WS_ENABLED);
    } else {
#if CONFIG_PM_WAKEUP_CONTROLLER
        if ((wuc_count > 0) && (wuc_configs != NULL)) {
            for (i = 0; i < wuc_count; i++) {
                ret = wuc_disable_wakeup_source(wuc_configs[i].wuc, wuc_configs[i].arg);
                if (ret != 0) {
                    return false;
                }
            }
        }   
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
        new_flags = flags & ~BIT(PM_DEVICE_FLAG_WS_ENABLED);
    }

    return atomic_cas(&pm->flags, flags, new_flags);
}
```

Device Tree Integration #100865 :
Devices specify WUC associations via wakeup-ctrls property.
```
lptmr0: timer@2d000 {
        compatible = "nxp,lptmr";
        reg = <0x2d000 0x10>;
        interrupts = <34 0>;
        clock-frequency = <DT_FREQ_K(32)>;
        clk-source = <2>;
        prescale-glitch-filter = <0>;
        resolution = <32>;
        status = "disabled";
        wakeup-source;
        wakeup-ctrls = <&wuu NXP_WUU_MODULE_0_INT>;
    };
```

PM subsystem uses WUC driver API to manage wakeup sources dynamically.
Device Driver Integration:
The MCUX LPTMR (Low Power Timer) driver is typically configured as the system clock source and should be enabled as the default wakeup source to ensure proper system operation after low-power states.

  Conditionally include WUC headers when `CONFIG_PM_WAKEUP_CONTROLLER` and relevant devicetree properties are present.
  Extract WUC device reference and source ID from devicetree.
```
wuu: system-modules@19000 {
        compatible = "nxp,wuu";
        reg = <0x19000 0x1000>;
        interrupts = <22 0>;
        #wakeup-ctrl-cells = <1>;
    };

lptmr0: timer@2d000 {
        compatible = "nxp,lptmr";
        reg = <0x2d000 0x10>;
        interrupts = <34 0>;
        clock-frequency = <DT_FREQ_K(32)>;
        clk-source = <2>;
        prescale-glitch-filter = <0>;
        resolution = <32>;
        status = "disabled";
        wakeup-source;
        wakeup-ctrls = <&wuu NXP_WUU_MODULE_0_INT>;
    };
```

  Enable wakeup source during initialization; disable when timer stops.
```c
static int sys_clock_driver_init(void)
{
    lptmr_config_t config;

#if CONFIG_PM_WAKEUP_CONTROLLER && DT_INST_NODE_HAS_PROP(0, wakeup_ctrls) && DT_INST_NODE_HAS_PROP(0, wakeup_source)
    const struct device *wuc = LPTMR_WUC_DEV;
    uint32_t sourceId = LPTMR_WUC_SOURCE_ID;
    if (wuc_enable_wakeup_source(wuc, sourceId) != 0) {
        return -EIO;
    };
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */

    LPTMR_GetDefaultConfig(&config);
    config.timerMode = kLPTMR_TimerModeTimeCounter;
    config.enableFreeRunning = false;
    config.prescalerClockSource = LPTMR_CLK_SOURCE;
    config.bypassPrescaler = (LPTMR_PRESCALER == 0);
    config.value = (LPTMR_PRESCALER == 0) ? 0 : (LPTMR_PRESCALER - 1);

    LPTMR_Init(LPTMR_BASE, &config);

    IRQ_CONNECT(LPTMR_IRQN, LPTMR_IRQ_PRIORITY, mcux_lptmr_timer_isr, NULL, 0);
    irq_enable(LPTMR_IRQN);

    LPTMR_EnableInterrupts(LPTMR_BASE, kLPTMR_TimerInterruptEnable);
    LPTMR_SetTimerPeriod(LPTMR_BASE, CYCLES_PER_TICK);
    LPTMR_StartTimer(LPTMR_BASE);

    return 0;
}

void sys_clock_disable(void)
{
    LPTMR_DisableInterrupts(LPTMR_BASE, kLPTMR_TimerInterruptEnable);
    LPTMR_StopTimer(LPTMR_BASE);
#if CONFIG_PM_WAKEUP_CONTROLLER && DT_INST_NODE_HAS_PROP(0, wakeup_ctrls) && DT_INST_NODE_HAS_PROP(0, wakeup_source)
    const struct device *wuc = LPTMR_WUC_DEV;
    uint32_t sourceId = LPTMR_WUC_SOURCE_ID;
    (void)wuc_disable_wakeup_source(wuc, sourceId);
#endif /* CONFIG_PM_WAKEUP_CONTROLLER */
}
```

### Dependencies

- subsys/pm/device.c
- dts/bindings/base/pm.yaml
- 

_No response_

### Alternatives Considered

_No response_

Albort12138:
Fix: #100853

#100866 (c89a262d39a8933f34b22e40ddd6b2865d904075)demonstrates an implementation  of this new device driver.
#100865 (91fa5e4cc694666255320fb3d03714e4f2f8370c) demonstrates an integration with other device driver.

### Overview
This proposal introduces a new device driver for the Wakeup Controller (WUC). It involves:
- Adding new APIs to manage the WUC peripheral.

---

### Motivation
In some MCUs, waking up from low-power modes requires special configuration of a dedicated peripheral called the Wakeup Controller (WUC). Currently, Zephyr supports this functionality in two ways:

1. **Implemented in interrupt controllers** (e.g., [`wuc_ite_it51xxx.c`](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/interrupt_controller/wuc_ite_it51xxx.c)):
   This approach lacks a unified interface for WUC control. For example, if different devices share the same peripheral but use different WUC implementations, supporting a new device often requires modifying its driver (e.g., [`espi_it8xxx2.c`](https://github.com/zephyrproject-rtos/zephyr/blob/fd63bf93a5a18bf0dc0a18b67b821bfdd68ab5b9/drivers/espi/espi_it8xxx2.c#L2684C1-L2688C7)).

2. **Implemented in SoC-specific power management code**(e.g., [`power.c`](https://github.com/zephyrproject-rtos/zephyr/blob/fd63bf93a5a18bf0dc0a18b67b821bfdd68ab5b9/soc/nxp/mcx/mcxw/mcxw7xx/power.c#L25)):  
   This method only supports a limited set of peripherals and lacks flexibility. If an application needs to use an unsupported peripheral as a wakeup source, additional code changes are required, and developers must be familiar with the MCU internals.

Both approaches have common issues:
- WUC configuration is typically done during initialization, with no flexible API for user-space control.
- The relationship between peripherals and WUC is hidden in code rather than expressed in devicetree, reducing transparency and flexibility.

**Problems solved by this proposal:**
- Introduce a generic WUC device driver with APIs that can be reused by other drivers, reducing the need for driver-specific changes when supporting new SoCs.
- Represent the peripheral-to-WUC mapping in devicetree instead of hardcoding it in source code.

---

### Design Details
The proposal introduces a new Wakeup Controller (WUC) driver API and integrates it with Zephyr’s power management subsystem.

**[New WUC Driver API](https://github.com/nxp-upstream/zephyr/blob/create-wakeup-controller/include/zephyr/drivers/wuc.h):**
- `wuc_enable_wakeup_source()` – Enable a specific wakeup source.
- `wuc_disable_wakeup_source()` – Disable a specific wakeup source.
- `wuc_check_wakeup_triggered()` – Check if a wakeup was triggered by a specific source.
- `wuc_record_wakeup_source_triggered` - Recode a specific wakeup source occurs.
- `wuc_clear_wakeup_source_triggered` - Clear triggered wakeup source.

### DIFF
--- a/drivers/wuc/Kconfig
+++ b/drivers/wuc/Kconfig
@@ -20,7 +20,7 @@ config WUC_INIT_PRIORITY
 
 
 # zephyr-keep-sorted-start
-
+rsource "Kconfig.mcux_wuc"
 # zephyr-keep-sorted-stop
 
 endif # WAKEUP_CONTROL
--- a/drivers/wuc/Kconfig
+++ b/drivers/wuc/Kconfig
@@ -1,0 +1,26 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig WUC
+	bool "Wakeup Controller drivers"
+	help
+	  Enable wakeup controller driver configuration
+
+if WUC
+
+module = WUC
+module-str = WUC
+source "subsys/logging/Kconfig.template.log_config"
+
+config WUC_INIT_PRIORITY
+	int "Wakeup controller init priority"
+	default KERNEL_INIT_PRIORITY_DEVICE
+	help
+	  Wakeup controller device driver initialization priority.
+
+
+# zephyr-keep-sorted-start
+
+# zephyr-keep-sorted-stop
+
+endif # WAKEUP_CONTROL
--- a/drivers/wuc/Kconfig.mcux_wuc
+++ b/drivers/wuc/Kconfig.mcux_wuc
@@ -1,0 +1,9 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+config WUC_MCUX_LLWU
+	bool "NXP Low Leakage Wakeup Unit"
+	default y
+	depends on DT_HAS_NXP_LLWU_ENABLED
+	help
+	  Enable support for NXP LLWU(Low Leakage Wakeup Unit) wakeup controller driver.
--- a/drivers/wuc/wuc_nxp_llwu.c
+++ b/drivers/wuc/wuc_nxp_llwu.c
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2026 NXP
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#define DT_DRV_COMPAT nxp_llwu
+
+#include <zephyr/init.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/kernel.h>
+#include <zephyr/spinlock.h>
+#include <zephyr/irq.h>
+#include <zephyr/pm/device.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/drivers/wuc.h>
+
+#include <fsl_llwu.h>
+#include <zephyr/dt-bindings/wuc/wuc_nxp_llwu.h>
+
+LOG_MODULE_REGISTER(mcux_llwu, CONFIG_WUC_LOG_LEVEL);
+
+struct mcux_llwu_config {
+	LLWU_Type *base;
+	uint32_t irqn;
+};
+
+struct mcux_llwu_data {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+	/** Bitmask of pin sources that have triggered a wakeup event */
+	uint32_t triggered_pin_sources;
+	/** Bitmask of currently enabled wakeup pins */
+	uint32_t enabled_pins;
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+	struct k_spinlock lock;
+};
+
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+static void mcux_llwu_isr(const struct device *dev)
+{
+	const struct mcux_llwu_config *config = dev->config;
+	struct mcux_llwu_data *data = dev->data;
+
+	K_SPINLOCK(&data->lock) {
+		for (int i = 0; i < 16; i++) {
+			if (IS_BIT_SET(data->enabled_pins, i)) {
+				if (LLWU_GetExternalWakeupPinFlag(config->base, i + 1) == true) {
+					data->triggered_pin_sources |= BIT(i);
+					LLWU_ClearExternalWakeupPinFlag(config->base, i + 1);
+				}
+			}
+		}
+	}
+}
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+
+static int mcux_llwu_enable_wakeup_source(const struct device *dev, uint32_t id)
+{
+	const struct mcux_llwu_config *config = dev->config;
+	struct mcux_llwu_data *data = dev->data;
+	int ret = -EINVAL;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id)) {
+			ret = 0;
+			if (IS_BIT_SET(data->enabled_pins, (NXP_LLWU_GET_PIN_INDEX(id) - 1)) ==
+			    false) {
+				LLWU_SetExternalWakeupPinMode(
+					config->base, NXP_LLWU_GET_PIN_INDEX(id),
+					(llwu_external_pin_mode_t)NXP_LLWU_GET_PIN_EDGE(id));
+				data->enabled_pins |= BIT((NXP_LLWU_GET_PIN_INDEX(id) - 1));
+			}
+
+			if (data->enabled_pins != 0) {
+				irq_enable(config->irqn);
+			}
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+
+#if FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE
+		if (NXP_LLWU_IS_MODULE_SOURCE(id)) {
+			ret = 0;
+			LLWU_EnableInternalModuleInterruptWakup(
+				config->base, NXP_LLWU_GET_MODULE_INDEX(id), true);
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE */
+	}
+
+	return ret;
+}
+
+static int mcux_llwu_disable_wakeup_source(const struct device *dev, uint32_t id)
+{
+	const struct mcux_llwu_config *config = dev->config;
+	struct mcux_llwu_data *data = dev->data;
+	int ret = -EINVAL;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id)) {
+			ret = 0;
+			if (IS_BIT_SET(data->enabled_pins, (NXP_LLWU_GET_PIN_INDEX(id) - 1)) ==
+			    true) {
+				LLWU_SetExternalWakeupPinMode(config->base,
+							      NXP_LLWU_GET_PIN_INDEX(id),
+							      kLLWU_ExternalPinDisable);
+				data->enabled_pins &= ~BIT((NXP_LLWU_GET_PIN_INDEX(id) - 1));
+			}
+
+			if (data->enabled_pins == 0) {
+				irq_disable(config->irqn);
+			}
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+
+#if FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE
+		if (NXP_LLWU_IS_MODULE_SOURCE(id)) {
+			ret = 0;
+			LLWU_EnableInternalModuleInterruptWakup(
+				config->base, NXP_LLWU_GET_MODULE_INDEX(id), false);
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE */
+	}
+
+	return ret;
+}
+
+static int mcux_llwu_check_wakeup_source_triggered(const struct device *dev, uint32_t id)
+{
+	struct mcux_llwu_data *data = dev->data;
+	int ret = 0;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id) &&
+		    IS_BIT_SET(data->triggered_pin_sources, (NXP_LLWU_GET_PIN_INDEX(id) - 1))) {
+			ret = 1;
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+	}
+
+	return ret;
+}
+
+static int mcux_llwu_clear_wakeup_source_triggered(const struct device *dev, uint32_t id)
+{
+	struct mcux_llwu_data *data = dev->data;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id)) {
+			data->triggered_pin_sources &= ~BIT(NXP_LLWU_GET_PIN_INDEX(id) - 1);
+		}
+#endif
+	}
+
+	return 0;
+}
+
+static int mcux_llwu_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	ARG_UNUSED(dev);
+	ARG_UNUSED(action);
+
+	return 0;
+}
+
+static int mcux_llwu_init(const struct device *dev)
+{
+	struct mcux_llwu_data *data = dev->data;
+
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+	IRQ_CONNECT(DT_IRQN(DT_DRV_INST(0)), DT_IRQ(DT_DRV_INST(0), priority), mcux_llwu_isr,
+		    DEVICE_DT_GET(DT_DRV_INST(0)), 0);
+	data->triggered_pin_sources = 0;
+	data->enabled_pins = 0;
+#endif /* FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN */
+
+	return pm_device_driver_init(dev, mcux_llwu_pm_action);
+}
+
+static const struct mcux_llwu_config llwu_config = {
+	.base = (LLWU_Type *)DT_REG_ADDR(DT_DRV_INST(0)),
+	.irqn = DT_IRQN(DT_DRV_INST(0)),
+};
+
+static struct mcux_llwu_data llwu_data;
+
+static const DEVICE_API(wuc, mcux_llwu_api) = {
+	.enable = mcux_llwu_enable_wakeup_source,
+	.disable = mcux_llwu_disable_wakeup_source,
+	.triggered = mcux_llwu_check_wakeup_source_triggered,
+	.clear = mcux_llwu_clear_wakeup_source_triggered,
+};
+
+DEVICE_DT_INST_DEFINE(0, mcux_llwu_init, NULL, &llwu_data, &llwu_config, POST_KERNEL,
+		      WUC_INIT_PRIORITY, &mcux_llwu_api);
--- a/dts/arm/nxp/mcx/nxp_mcxc_common.dtsi
+++ b/dts/arm/nxp/mcx/nxp_mcxc_common.dtsi
@@ -393,6 +393,14 @@
 			#io-channel-cells = <1>;
 			voltage-reference = <2>;
 		};
+
+		llwu: llwu@4007c000 {
+			compatible = "nxp,llwu";
+			reg = <0x4007c000 0xa>;
+			interrupts = <7 0>;
+			#wakeup-ctrl-cells = <1>;
+			status = "disabled";
+		};
 	};
 };
 
--- a/dts/bindings/wuc/nxp,llwu.yaml
+++ b/dts/bindings/wuc/nxp,llwu.yaml
@@ -1,0 +1,24 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: NXP Low-Leakage Wakeup Unit
+
+compatible: "nxp,llwu"
+
+include: [base.yaml, wuc-controller.yaml]
+
+properties:
+  reg:
+    required: true
+
+  interrupts:
+    required: true
+
+  "#wakeup-ctrl-cells":
+    type: int
+    const: 1
+    description: |
+      Number of cells in wakeup-ctrls property.
+
+wakeup-ctrl-cells:
+  - id
--- a/dts/bindings/wuc/wuc-controller.yaml
+++ b/dts/bindings/wuc/wuc-controller.yaml
@@ -1,0 +1,13 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+    Common file for Wakeup Controllers
+
+properties:
+  "#wakeup-ctrl-cells":
+    type: int
+    required: true
+    description: |
+      Number of cells in wakeup-ctrls property. There must be a cell
+      name "id" to use.
--- a/dts/bindings/wuc/wuc-device.yaml
+++ b/dts/bindings/wuc/wuc-device.yaml
@@ -1,0 +1,11 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+    Common file for wake-up source devices referencing wake up controller(s).
+
+properties:
+  wakeup-ctrls:
+    type: phandle-array
+    description: |
+      Phandle and wakeup source identifier of the corresponding wakeup controller.
--- a/include/zephyr/devicetree/wuc.h
+++ b/include/zephyr/devicetree/wuc.h
@@ -1,0 +1,209 @@
+/**
+ * @file
+ * @brief Wakeup Controller Devicetree macro public API header file.
+ */
+
+/*
+ * Copyright 2026 NXP
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DEVICETREE_WUC_H_
+#define ZEPHYR_INCLUDE_DEVICETREE_WUC_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @defgroup devicetree-wakeup-controller Devicetree WUC Controller API
+ * @ingroup devicetree
+ * @ingroup wuc_interface
+ * @{
+ */
+
+/**
+ * @brief Get the node identifier for the controller phandle from a
+ *        "wakeup-ctrls" phandle-array property at an index
+ *
+ * Example devicetree fragment:
+ *
+ *     wuc1: wakeup-controller@... { ... };
+ *
+ *     wuc2: wakeup-controller@... { ... };
+ *
+ *     n: node {
+ *             wakeup-ctrls = <&wuc1 10>, <&wuc2 20>;
+ *     };
+ *
+ * Example usage:
+ *
+ *     DT_WUC_BY_IDX(DT_NODELABEL(n), 0) // DT_NODELABEL(wuc1)
+ *     DT_WUC_BY_IDX(DT_NODELABEL(n), 1) // DT_NODELABEL(wuc2)
+ *
+ * @param node_id node identifier
+ * @param idx logical index into "wakeup-ctrls"
+ * @return the node identifier for the wakeup controller referenced at
+ *         index "idx"
+ * @see DT_PHANDLE_BY_IDX()
+ */
+#define DT_WUC_BY_IDX(node_id, idx) DT_PHANDLE_BY_IDX(node_id, wakeup_ctrls, idx)
+
+/**
+ * @brief Equivalent to DT_WUC_BY_IDX(node_id, 0)
+ * @param node_id node identifier
+ * @return a node identifier for the wakeup controller at index 0
+ *         in "wakeup-ctrls"
+ * @see DT_WUC_BY_IDX()
+ */
+#define DT_WUC(node_id) DT_WUC_BY_IDX(node_id, 0)
+
+/**
+ * @brief Get a wakeup controller specifier's cell value at an index
+ *
+ * Example devicetree fragment:
+ *
+ *     wuc: wakeup-controller@... {
+ *             compatible = "vnd,wuc";
+ *             #wakeup-ctrl-cells = <1>;
+ *     };
+ *
+ *     n: node {
+ *             wakeup-ctrls = <&wuc 10>;
+ *     };
+ *
+ * Bindings fragment for the vnd,wuc compatible:
+ *
+ *     wakeup-ctrl-cells:
+ *       - id
+ *
+ * Example usage:
+ *
+ *     DT_WUC_CELL_BY_IDX(DT_NODELABEL(n), 0, id) // 10
+ *
+ * @param node_id node identifier for a node with a wakeup-ctrls property
+ * @param idx logical index into wakeup-ctrls property
+ * @param cell lowercase-and-underscores cell name
+ * @return the cell value at index "idx"
+ * @see DT_PHA_BY_IDX()
+ */
+#define DT_WUC_CELL_BY_IDX(node_id, idx, cell) DT_PHA_BY_IDX(node_id, wakeup_ctrls, idx, cell)
+
+/**
+ * @brief Equivalent to DT_WUC_CELL_BY_IDX(node_id, 0, cell)
+ * @param node_id node identifier for a node with a wakeup-ctrls property
+ * @param cell lowercase-and-underscores cell name
+ * @return the cell value at index 0
+ * @see DT_WUC_CELL_BY_IDX()
+ */
+#define DT_WUC_CELL(node_id, cell) DT_WUC_CELL_BY_IDX(node_id, 0, cell)
+
+/**
+ * @brief Get the node identifier for the controller phandle from a
+ *        "wakeup-ctrls" phandle-array property at an index
+ *
+ * @param inst instance number
+ * @param idx logical index into "wakeup-ctrls"
+ * @return the node identifier for the wakeup controller referenced at
+ *         index "idx"
+ * @see DT_WUC_BY_IDX()
+ */
+#define DT_INST_WUC_BY_IDX(inst, idx) DT_WUC_BY_IDX(DT_DRV_INST(inst), idx)
+
+/**
+ * @brief Equivalent to DT_INST_WUC_BY_IDX(inst, 0)
+ * @param inst instance number
+ * @return a node identifier for the wakeup controller at index 0
+ *         in "wakeup-ctrls"
+ * @see DT_WUC_BY_IDX()
+ */
+#define DT_INST_WUC(inst) DT_INST_WUC_BY_IDX(inst, 0)
+
+/**
+ * @brief Get a DT_DRV_COMPAT instance's wakeup controller specifier's cell value
+ *        at an index
+ * @param inst DT_DRV_COMPAT instance number
+ * @param idx logical index into wakeup-ctrls property
+ * @param cell lowercase-and-underscores cell name
+ * @return the cell value at index "idx"
+ * @see DT_WUC_CELL_BY_IDX()
+ */
+#define DT_INST_WUC_CELL_BY_IDX(inst, idx, cell) DT_WUC_CELL_BY_IDX(DT_DRV_INST(inst), idx, cell)
+
+/**
+ * @brief Equivalent to DT_INST_WUC_CELL_BY_IDX(inst, 0, cell)
+ * @param inst DT_DRV_COMPAT instance number
+ * @param cell lowercase-and-underscores cell name
+ * @return the value of the cell inside the specifier at index 0
+ */
+#define DT_INST_WUC_CELL(inst, cell) DT_INST_WUC_CELL_BY_IDX(inst, 0, cell)
+
+/**
+ * @brief Get a Wakeup Controller specifier's id cell at an index
+ *
+ * This macro only works for Wakeup Controller specifiers with cells named "id".
+ * Refer to the node's binding to check if necessary.
+ *
+ * Example devicetree fragment:
+ *
+ *     wuc: wakeup-controller@... {
+ *             compatible = "vnd,wuc";
+ *             #wakeup-ctrl-cells = <1>;
+ *     };
+ *
+ *     n: node {
+ *             wakeup-ctrls = <&wuc 10>;
+ *     };
+ *
+ * Bindings fragment for the vnd,wuc compatible:
+ *
+ *     wakeup-ctrl-cells:
+ *       - id
+ *
+ * Example usage:
+ *
+ *     DT_WUC_ID_BY_IDX(DT_NODELABEL(n), 0) // 10
+ *
+ * @param node_id node identifier
+ * @param idx logical index into "wakeup-ctrls"
+ * @return the id cell value at index "idx"
+ * @see DT_PHA_BY_IDX()
+ */
+#define DT_WUC_ID_BY_IDX(node_id, idx) DT_PHA_BY_IDX(node_id, wakeup_ctrls, idx, id)
+
+/**
+ * @brief Equivalent to DT_WUC_ID_BY_IDX(node_id, 0)
+ * @param node_id node identifier
+ * @return the id cell value at index 0
+ * @see DT_WUC_ID_BY_IDX()
+ */
+#define DT_WUC_ID(node_id) DT_WUC_ID_BY_IDX(node_id, 0)
+
+/**
+ * @brief Get a DT_DRV_COMPAT instance's Wakeup Controller specifier's id cell value
+ *        at an index
+ * @param inst DT_DRV_COMPAT instance number
+ * @param idx logical index into "wakeup-ctrls"
+ * @return the id cell value at index "idx"
+ * @see DT_WUC_ID_BY_IDX()
+ */
+#define DT_INST_WUC_ID_BY_IDX(inst, idx) DT_WUC_ID_BY_IDX(DT_DRV_INST(inst), idx)
+
+/**
+ * @brief Equivalent to DT_INST_WUC_ID_BY_IDX(inst, 0)
+ * @param inst DT_DRV_COMPAT instance number
+ * @return the id cell value at index 0
+ * @see DT_INST_WUC_ID_BY_IDX()
+ */
+#define DT_INST_WUC_ID(inst) DT_INST_WUC_ID_BY_IDX(inst, 0)
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* ZEPHYR_INCLUDE_DEVICETREE_WUC_H_ */
--- a/include/zephyr/drivers/wuc.h
+++ b/include/zephyr/drivers/wuc.h
@@ -1,0 +1,310 @@
+/*
+ * Copyright 2026 NXP
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file
+ * @ingroup wuc_interface
+ * @brief Main header file for WUC (Wakeup Controller) driver API.
+ */
+
+#ifndef ZEPHYR_INCLUDE_DRIVERS_WUC_H_
+#define ZEPHYR_INCLUDE_DRIVERS_WUC_H_
+
+/**
+ * @brief Wakeup Controller (WUC) Driver APIs
+ * @defgroup wuc_interface WUC
+ * @since 4.4
+ * @version 0.1.0
+ * @ingroup io_interfaces
+ * @{
+ */
+
+#include <errno.h>
+#include <stddef.h>
+
+#include <zephyr/device.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Wakeup controller device configuration. */
+struct wuc_dt_spec {
+	/** Wakeup controller device. */
+	const struct device *dev;
+	/** Wakeup source identifier. */
+	uint32_t id;
+};
+
+/**
+ * @brief Static initializer for a @p wuc_dt_spec
+ *
+ * This returns a static initializer for a @p wuc_dt_spec structure given a
+ * devicetree node identifier, a property specifying a Wakeup Controller and an index.
+ *
+ * Example devicetree fragment:
+ *
+ *	n: node {
+ *		wakeup-ctrls = <&wuc 10>;
+ *	}
+ *
+ * Example usage:
+ *
+ *	const struct wuc_dt_spec spec = WUC_DT_SPEC_GET_BY_IDX(DT_NODELABEL(n), 0);
+ *	Initializes 'spec' to:
+ *	 {
+ *	         .dev = DEVICE_DT_GET(DT_NODELABEL(wuc)),
+ *	         .id = 10
+ *	 }
+ *
+ * The 'wuc' field must still be checked for readiness, e.g. using
+ * device_is_ready(). It is an error to use this macro unless the node
+ * exists, has the given property, and that property specifies a wakeup
+ * controller wakeup source id as shown above.
+ *
+ * @param node_id devicetree node identifier
+ * @param idx logical index into "wakeup-ctrls"
+ * @return static initializer for a struct wuc_dt_spec for the property
+ */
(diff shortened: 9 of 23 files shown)


include/zephyr/pm/device.h
```diff
@@ -205,6 +215,10 @@ struct pm_device_base {
 	/** Power Domain it belongs */
 	const struct device *domain;
 #endif /* CONFIG_PM_DEVICE_POWER_DOMAIN */
+#if defined(CONFIG_PM_WAKEUP_CONTROLLER) || defined(__DOXYGEN__)
+	uint8_t wuc_count;
```
ZhaoxiangJin: What does `wuc_count `represent? The number of wakeup unit controllers on an SoC? Why is it placed in pm_device?

drivers/timer/mcux_lptmr_timer.c
```diff
@@ -15,6 +15,10 @@
 #include <fsl_lptmr.h>
 #include <zephyr/irq.h>
 
+#if CONFIG_PM_WAKEUP_CONTROLLER && DT_INST_NODE_HAS_PROP(0, wakeup_ctrls) && DT_INST_NODE_HAS_PROP(0, wakeup_source)
```
zejiang0jason: could this be simplified?

drivers/wuc/nxp_wuu.c
```diff
@@ -0,0 +1,222 @@
...
+	static void mcux_wuu_config_func_##n(const struct device *dev);                            \
+	static struct mcux_wuu_config mcux_wuu_config_##n = {                                      \
+		.base = (WUU_Type *)DT_INST_REG_ADDR(n),                                           \
+		.irqn = DT_INST_IRQN(n),                                                           \
+		.irq_config_func = mcux_wuu_config_func_##n,                                       \
+	};                                                                                         \
+	static struct mcux_wuu_data mcux_wuu_data_##n = {                                          \
+		.triggered_pin_sources = 0,                                                        \
+	};                                                                                         \
+	static struct wuc_driver_api mcux_wuu_api_##n = {                                          \
```
zejiang0jason: this doesn't need to be instance specific.

dts/bindings/wuc/nxp,wuu.yaml
```diff
@@ -0,0 +1,25 @@
...
+properties:
+  reg:
+    required: true
+
+  interrupts:
+    required: true
+  
+  '#wakeup-ctrl-cells':
+    type: int
+    const: 1
```
rruuaanng:
```suggestion
  "#wakeup-ctrl-cells":
    const: 1
```

drivers/wuc/nxp_wuu.c
Albort12138: Will Update, thanks your comment

drivers/timer/mcux_lptmr_timer.c
Albort12138: The original huge PR already separated into serval small PRs,  please move to https://github.com/zephyrproject-rtos/zephyr/pull/100865 for details.

JarmouniA (commented):
This PR shall implement the API for at least one platform to demonstrate that it's usable.
Even better, you (or someone else) should try implementing it for a second platform to demonstrate that it's generic enough to be adopted by other vendors.

Edit: just saw #100865, it should be mentioned in the description, or even better added as a separate commit in this PR.

JarmouniA (commented):
A couple of remarks:
- You have to specify from what Zephyr state the system is waking up: Poweroff or low-power states, or both.
- you have to demonstrate that you have looked sufficiently at similar IPs of other vendors, so that the semantics of the core API can become sufficiently vendor-agnostic. Of course, there can be vendor-specific extensions to cover the features that are hard to generalize.

dts/bindings/wuc/nxp,llwu.yaml
```diff
@@ -0,0 +1,24 @@
...
+include: [base.yaml, wuc-controller.yaml]
+
+properties:
+  reg:
+    required: true
+
+  interrupts:
+    required: true
+
+  '#wakeup-ctrl-cells':
```
rruuaanng:
```suggestion
  "#wakeup-ctrl-cells":
```

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+static const struct mcux_llwu_config llwu_config = {
+	.base = (LLWU_Type *)DT_REG_ADDR(DT_DRV_INST(0)),
+	.irqn = DT_IRQN(DT_DRV_INST(0)),
+};
+
+static struct mcux_llwu_data llwu_data;
+
+static const struct wuc_driver_api mcux_llwu_api = {
```
rruuaanng:
```suggestion
static const DEVICE_API(wuc, mcux_llwu_api) = {
```

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,238 @@
...
+static int mcux_llwu_check_wakeup_source_triggered(const struct device *dev, uint32_t id)
+{
+	struct mcux_llwu_data *data = dev->data;
+	int ret = 0;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id)) {
+			if (IS_BIT_SET(data->triggered_pin_sources,
+				       (NXP_LLWU_GET_PIN_INDEX(id) - 1))) {
```
rruuaanng:
```suggestion
		if (NXP_LLWU_IS_PIN_SOURCE(id) && IS_BIT_SET(data->triggered_pin_sources,
		   (NXP_LLWU_GET_PIN_INDEX(id) - 1))) {
```
Merge them. Or:

```
#if xxx
return NXP_LLWU_IS_PIN_SOURCE(id) &&
       IS_BIT_SET(data->triggered_pin_sources,
	  (NXP_LLWU_GET_PIN_INDEX(id) - 1)) ? 1 : 0;
#elif xxx
return ...
#endif
```

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,238 @@
...
+
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+static void mcux_llwu_isr(const struct device *dev)
+{
+	const struct mcux_llwu_config *config = dev->config;
+	struct mcux_llwu_data *data = dev->data;
+
+	K_SPINLOCK(&data->lock) {
+		for (int i = 0; i < 16; i++) {
+			if (IS_BIT_SET(data->enabled_pins, i) == true) {
```
rruuaanng:
```suggestion
			if (IS_BIT_SET(data->enabled_pins, i)) {
```
The semantics of `IS_` are very readable.

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * @brief Enable a wakeup source
+ *
+ * @param dev Pointer to the WUC device structure
+ * @param id Wakeup source identifier
+ *
+ * @retval 0 If successful
+ * @retval -errno Negative errno code on failure
+ */
+__syscall int wuc_enable_wakeup_source(const struct device *dev, uint32_t id);
```
JarmouniA: no need for syscalls since the API is used by drivers.

include/zephyr/drivers/wuc/wuc_nxp_llwu.h
```diff
@@ -0,0 +1,119 @@
...
+
+/* Encode MODULE wakeup source */
+#define NXP_LLWU_MODULE_WAKEUP(module_index) \
+	(NXP_LLWU_WAKEUP_SOURCE_TYPE_MODULE | \
+	 ((module_index) & 0xFF))
+
+/* PIN wakeup sources (PIN 0-16) */
+#define NXP_LLWU_PIN_0_RISING    NXP_LLWU_PIN_WAKEUP(1, NXP_LLWU_PIN_EDGE_RISING)
+#define NXP_LLWU_PIN_0_FALLING   NXP_LLWU_PIN_WAKEUP(1, NXP_LLWU_PIN_EDGE_FALLING)
+#define NXP_LLWU_PIN_0_ANY       NXP_LLWU_PIN_WAKEUP(1, NXP_LLWU_PIN_EDGE_ANY)
```
JarmouniA: better done in DT as child nodes of WUC.
ex. https://github.com/zephyrproject-rtos/zephyr/blob/cdb7debfb726569440fc22ce6673de2b86c728f7/dts/arm/st/u5/stm32u5.dtsi#L950

include/zephyr/drivers/wuc/wuc_nxp_llwu.h
JarmouniA, line 1: should be in include/zephyr/dt-bindings

include/zephyr/drivers/wuc.h
Albort12138: Removed

drivers/wuc/wuc_nxp_llwu.c
Albort12138: Thanks your comments, updated

drivers/wuc/wuc_nxp_llwu.c
Albort12138: Updated

drivers/wuc/wuc_nxp_llwu.c
Albort12138: Updated

dts/bindings/wuc/nxp,llwu.yaml
Albort12138: Updated

Albort12138:
Hi @bjarki-andreasen @rruuaanng @JarmouniA , The PR has been updated. Please take a look when you have time. Thank you very much

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,331 @@
...
+ * Example devicetree fragment:
+ *
+ *	n: node {
+ *		wakeup-ctrls = <&wuc 10>;
+ *	}
+ *
+ * Example usage:
+ *
+ *	const struct wuc_dt_spec spec = WUC_DT_SPEC_GET_BY_IDX(DT_NODELABEL(n), 0);
+ *	// Initializes 'spec' to:
```
bjarki-andreasen: There are comments in the comment?

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,331 @@
...
+typedef int (*wuc_api_disable_wakeup_source)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_check_wakeup_source_triggered)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_clear_wakeup_source_triggered)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_record_wakeup_source_triggered)(const struct device *dev, uint32_t id);
+
+__subsystem struct wuc_driver_api {
+	wuc_api_enable_wakeup_source enable_wakeup_source;
```
bjarki-andreasen: APIs are already name spaced, so _wakeup_source is redundant, could just be enable, disable, triggered, clear etc.

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * @brief Record a wakeup source as triggered.
+ *
+ * @param dev Pointer to the WUC device structure.
+ * @param id Wakeup source identifier.
+ *
+ * @retval 0 If successful.
+ * @retval -errno Negative errno code on failure.
+ */
+static inline int wuc_record_wakeup_source_triggered(const struct device *dev, uint32_t id)
```
bjarki-andreasen: I believe this means "trigger the wakeup source"? I don't immediately understand what "record" means in this context

tests/drivers/wuc/wuc_api/src/main.c
```diff
@@ -0,0 +1,91 @@
...
+	for (i = 0; i < test_wuc_dt_spec_count; i++) {
+		/* clear any previous state. */
+		ret = wuc_clear_wakeup_source_triggered_dt(&test_wuc_dt_specs[i]);
+		zassert_ok(ret, "Fail to clear wakeup source %d", test_wuc_dt_specs[i].id);
+
+		ret = wuc_check_wakeup_source_triggered_dt(&test_wuc_dt_specs[i]);
+		zassert_equal(ret, 0, "Wakeup source should not be triggered initially.");
+
+		/* Record wakeup source as triggered */
+		ret = wuc_record_wakeup_source_triggered_dt(&test_wuc_dt_specs[i]);
```
bjarki-andreasen: Manually triggering a wakeup source is likely not going to be supported by all wakeup controllers, so allowing and checking for -ENOTSUP and skipping this case willbe needed

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * @brief Clear a wakeup source triggered status.
+ *
+ * @param dev Pointer to the WUC device structure.
+ * @param id Wakeup source identifier.
+ *
+ * @retval 0 If successful.
+ * @retval -errno Negative errno code on failure.
+ */
+static inline int wuc_clear_wakeup_source_triggered(const struct device *dev, uint32_t id)
```
bjarki-andreasen, line 282: This API should be optional, some wakeup controllers can not check, clear or trigger the wakeup source, they are simply "triggered" if some hardware event is pending.

bjarki-andreasen (changes requested):

include/zephyr/drivers/wuc.h
Albort12138: Fixed

include/zephyr/drivers/wuc.h
Albort12138: Good suggestion, will update

include/zephyr/drivers/wuc.h
Albort12138: Updated

include/zephyr/drivers/wuc.h
Albort12138: In some WUC implementations, whether a given wakeup source actually woke the system is not reflected in the WUC peripheral itself. In those cases the wakeup source must be manually recorded into the WUC device driver so it can be queried later. For example, we have a WUC called "wuu" and an internal peripheral interrupt ( take timer interrupt as example) used as a wakeup source. When the LPTMR interrupt wakes the MCU, there is no status bit in wuu indicating the LPTMR caused the wakeup. I expect the record API to be used to update the wuu device driver so that subsequent checks will show the LPTMR wakeup was triggered.

bjarki-andreasen: If the wakeup source is not recorded in the wuc, then that's not where it should be gotten from right? For example, if an RTC alarm caused a wakeup, you would check if the RTC alarm fired, since its a combination of the wakeup event it can produce being enabled, and the specific event actually being active. See for example https://github.com/zephyrproject-rtos/zephyr/blob/1b23efc6121eedac9f30391265f767c1f232724c/include/zephyr/drivers/rtc.h#L335 and https://github.com/zephyrproject-rtos/zephyr/blob/1b23efc6121eedac9f30391265f767c1f232724c/include/zephyr/drivers/gpio.h#L1899

dts/bindings/wuc/wuc-controller.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: Wakeup Controller
```
JarmouniA:
```suggestion
description: Common file for Wakeup Controllers
```

dts/bindings/wuc/wuc-device.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: This file needs to be included by devices that need a wake up controller.
```
JarmouniA:
```suggestion
description: Common file for wake-up source devices referencing wake up controller(s).
```

dts/bindings/wuc/wuc-device.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+properties:
+  wakeup-ctrls:
+    type: phandle-array
+    description: |
+      Phandle to the wakeup controller used by this device.
```
JarmouniA:
```suggestion
      Phandle of corresponding wakeup controller.
```

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,347 @@
...
+
+#ifndef ZEPHYR_INCLUDE_DRIVERS_WUC_H_
+#define ZEPHYR_INCLUDE_DRIVERS_WUC_H_
+
+/**
+ * @brief Wakeup Controller (WUC) Driver APIs
+ * @defgroup wuc_interface WUC Driver APIs
+ * @since 4.4
+ * @version 0.1.0
+ * @ingroup io_interfaces
```
JarmouniA, line 21: Really?

tests/drivers/wuc/wuc_api/testcase.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+common:
+  tags:
+    - drivers
+    - wuc
+tests:
+  drivers.wuc.api.nxp:
```
JarmouniA:
```suggestion
  drivers.wuc.api
```

dts/bindings/wuc/wuc-controller.yaml
Albort12138: Updated

include/zephyr/drivers/wuc.h
Albort12138: Your description is very reasonable. If it isn’t recorded in the WUC, it shouldn’t be handled in the WUC device driver. My earlier understanding was wrong—thanks for the explanation.

tests/drivers/wuc/wuc_api/testcase.yaml
Albort12138: Removed

include/zephyr/drivers/wuc.h
Albort12138: Apologies, I am not cognizant of the specific mistakes. Could you kindly furnish more details? 
Thank you very much.

dts/bindings/wuc/wuc-device.yaml
Albort12138: Updated

dts/bindings/wuc/wuc-device.yaml
Albort12138: Updated

tests/drivers/wuc/wuc_api/src/main.c
Albort12138: Removed

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * @param dev Pointer to the WUC device structure.
+ * @param id Wakeup source identifier.
+ *
+ * @retval 1 If wakeup was triggered by this source.
+ * @retval 0 If wakeup was not triggered by this source.
+ * @retval -errno Negative errno code on failure.
+ * @retval -ENOSYS if the interface is not implemented.
+ */
+static inline int wuc_check_wakeup_source_triggered(const struct device *dev, uint32_t id)
```
Albort12138, line 249: Hi @bjarki-andreasen @JarmouniA @rruuaanng , I’m considering changing the API to:
```c
static inline enum wuc_status wuc_get_wakeup_source_status(const struct device *dev,
                               uint32_t id);
```
It would return the status of a specific wakeup source (disabled, enabled, triggered, or unknown).
Do you think this is more useful and user-friendly?

JarmouniA: I think so.

include/zephyr/drivers/wuc.h
JarmouniA: `@ingroup io_interfaces`, not saying it's definitely incorrect.

include/zephyr/drivers/wuc.h
bjarki-andreasen: I don't think so, when would a driver care to check? if a driver wants the wakeup source to be enabled, it will simply enable it, there is no gain in checking if its enabled or not from that perspective, it will be whatever its set to.

JarmouniA: What about a wake-up source that's shared between many peripherals, that's the usecase where I thought the ability to check the state would be useful.

bjarki-andreasen: Wakeup sources are not shared, if they could be, that would require some sort of reference counted or onoff service style API to take/release.

JarmouniA: by shared, I meant can be used by many peripherals, but not at the same time.
The idea is that the driver would be able to check if the source has been enabled by another peripheral, and if so it tries setting up another that's unused (if available).
Or would that be better handed inside the wake-up controller driver. But in that case, it would have to be able to identify the peripheral requesting the source.

Albort12138: Thanks for the discussion. I agree there are valid cases where multiple instances of the same peripheral share a single wakeup source (for example, multiple UART instances tied to one wakeup source).

In that model, the wakeup source is a shared capability, not an instance‑private resource. As long as any consumer enables it, it must stay active.

Given this, a generic “is_enabled” API has unclear semantics in the shared case. For example:

Instance A enables the wakeup source
Instance B disables the same wakeup source
Any is_enabled() query is ambiguous:

Returning true is misleading for B
Returning false is incorrect for A
Returning a refcount or aggregate state isn’t actionable for drivers
Based on this, I do not plan to change the API at this time.

Thanks again for the discussion; it has been very helpful in clarifying the API’s scope and boundaries.

include/zephyr/drivers/wuc.h
bjarki-andreasen: All device driver APIs are in `io_interfaces`, maybe its not relevant for this one

tests/drivers/wuc/wuc_api/testcase.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2026 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+common:
+  tags:
+    - drivers
+    - wuc
+tests:
+  drivers.wuc.api:
+    depends_on: wuc
```
bjarki-andreasen, line 10: please filter on the existence of `DT_PATH(zephyr_user), wakeup_ctrls`

I believe this should work:
```
drivers.wuc.api:
  filter: dt_node_has_prop("/zephyr,user", "wakeup-ctrls")
```

Usually its enough to test just one board per soc, so its typical to only include an overlay for one of them.

bjarki-andreasen (changes requested):
Looking really good!

tests/drivers/wuc/wuc_api/testcase.yaml
Albort12138: Updated.

bjarki-andreasen (dismissed):

include/zephyr/dt-bindings/wuc/wuc_nxp_llwu.h
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @brief Encode a MODULE wakeup source.
+ *
+ * @param module_index Module index (0-255).
+ */
+#define NXP_LLWU_MODULE_WAKEUP(module_index)                                                       \
+	(NXP_LLWU_WAKEUP_SOURCE_TYPE_MODULE | ((module_index) & 0xFF))
+
+/** @name PIN wakeup sources (PIN 0-16) */
```
ceolin: Shouldn't it be `(PIN 0-15)` ?

include/zephyr/dt-bindings/wuc/wuc_nxp_llwu.h
```diff
@@ -0,0 +1,213 @@
...
+/** @{ */
+/** PIN 0 wakeup source on rising edge. */
+#define NXP_LLWU_PIN_0_RISING  NXP_LLWU_PIN_WAKEUP(1, NXP_LLWU_PIN_EDGE_RISING)
```
ceolin, line 67: This is somehow awkward ... it says `PIN_0` but the value is 1

dts/arm/nxp/mcx/nxp_mcxc_common.dtsi
```diff
@@ -389,6 +389,14 @@
 			#io-channel-cells = <1>;
 			voltage-reference = <2>;
 		};
+
+		llwu: llwu@0x4007c000 {
```
ceolin: Does it works ? I thought you couldn't have `0x` in the address.

ceolin (dismissed):
It looks good to me. Minor comments in the driver only.

include/zephyr/dt-bindings/wuc/wuc_nxp_llwu.h
Albort12138: Updated

include/zephyr/dt-bindings/wuc/wuc_nxp_llwu.h
Albort12138: This is to align with requirement of llwu driver.

dts/arm/nxp/mcx/nxp_mcxc_common.dtsi
Albort12138: Fixed

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * For internal use only, skip these in public documentation.
+ */
+typedef int (*wuc_api_enable_wakeup_source)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_disable_wakeup_source)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_check_wakeup_source_triggered)(const struct device *dev, uint32_t id);
+
+typedef int (*wuc_api_clear_wakeup_source_triggered)(const struct device *dev, uint32_t id);
```
bjarki-andreasen, line 174: This could be squashed into be 1 API, which does both check and clear, like https://github.com/zephyrproject-rtos/zephyr/blob/7c37e958750a66c2f2706a157f24d24fe5e51591/include/zephyr/drivers/rtc.h#L335

Since these two APIs have to be synchronized to avoid a data race (one context checks and clears before another gets a change to check, thus only the first context gets correct data), having the triggered API check and clear is a valid optimization.

bjarki-andreasen (changes requested):
See comment

include/zephyr/drivers/wuc.h
Albort12138: When defining the wuc APIs, I also considered this point. However, I ultimately still believe that having an API focus on a single responsibility is more reasonable.
As for the data race issue you mentioned, would it be acceptable if we add more detailed explanations in the documentation?

Edit:
In addition, some WUC peripherals do not support a clear operation, so separating the check and clear functions into two independent APIs would provide better flexibility.

drivers/wuc/Kconfig
```diff
@@ -0,0 +1,26 @@
...
+	default KERNEL_INIT_PRIORITY_DEVICE
+	help
+	  Wakeup controller device driver initialization priority.
+
+
+# zephyr-keep-sorted-start
+rsource "Kconfig.mcux_wuc"
+# zephyr-keep-sorted-stop
+
+endif # WAKEUP_CONTROL
```
sylvioalves, line 26: `endif # WUC`

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,193 @@
...
+
+#if FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE
+		if (NXP_LLWU_IS_MODULE_SOURCE(id)) {
+			LLWU_EnableInternalModuleInterruptWakup(
+				config->base, NXP_LLWU_GET_MODULE_INDEX(id), true);
+		}
+#endif /* FSL_FEATURE_LLWU_HAS_INTERNAL_MODULE */
+	}
+
+	return 0;
```
sylvioalves: `enable` and `disable` calls always return 0, even if the source id won't match. I suggest returning `-EINVAL` in such cases.

dts/bindings/wuc/wuc-controller.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+description: |
+    Common file for Wakeup Controllers
+
+properties:
+  "#wakeup-ctrl-cells":
+    type: int
+    required: true
+    description: |
+      Number of cells in wakeup-ctrls property. There must be a cell
```
sylvioalves, line 12: I think this could go: `There must be a cell name "id" to use.`. I mean, this is an implementation constraint not a binding description.

include/zephyr/drivers/wuc.h
```diff
@@ -0,0 +1,307 @@
...
+ *
+ * @retval 1 If wakeup was triggered by this source.
+ * @retval 0 If wakeup was not triggered by this source.
+ * @retval -errno Negative errno code on failure.
+ * @retval -ENOSYS if the interface is not implemented.
+ */
+static inline int wuc_check_wakeup_source_triggered(const struct device *dev, uint32_t id)
+{
+	if (DEVICE_API_GET(wuc, dev)->triggered == NULL) {
+		return -ENOSYS;
```
sylvioalves, line 252: `-ENOSYS` is indirectly added be `zephyr/device.h`. I suggest including `<errno.h>` in this file as well.

drivers/wuc/Kconfig.mcux_wuc
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+config WUC_MCUX_LLWU
+	bool "NXP Low Leakage Wakeup Unit"
+	default y
+	depends on DT_HAS_NXP_LLWU_ENABLED
+	help
+		Enable support for NXP LLWU(Low Leakage Wakeup Unit) wakeup controller driver.
```
sylvioalves: Use 2-spaces in `help` section.

tests/drivers/wuc/wuc_api/src/main.c
```diff
@@ -0,0 +1,58 @@
...
+{
+	zassert_true(device_is_ready(test_wuc_dt_specs[0].dev), "WUC device is not ready");
+
+	return NULL;
+}
+
+/**
+ * @brief Test wuc_enable_wakeup_source API
+ */
+ZTEST(wuc_api, test_wuc_enable_wakeup_source)
```
sylvioalves, line 33: Sorry if already discussed but there is no test for triggered and clear APIs. The test doesn't verify that enabling the same source twice is idempotent, or that disabling an already-disabled source is safe.

dts/bindings/wuc/wuc-device.yaml
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+    Common file for wake-up source devices referencing wake up controller(s).
+
+properties:
+  wakeup-ctrls:
+    type: phandle-array
+    description: |
+      Phandle of corresponding wakeup controller.
```
sylvioalves: Suggestion: `Phandle and wakeup source identifier of the corresponding wakeup controller.`

dts/bindings/wuc/nxp,llwu.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2025 NXP
```
sylvioalves: Would you update all new files copyright to 2026? (unless it is not new or copied..)

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,193 @@
...
+	struct mcux_llwu_data *data = dev->data;
+
+	K_SPINLOCK(&data->lock) {
+		for (int i = 0; i < 16; i++) {
+			if (IS_BIT_SET(data->enabled_pins, i)) {
+				if (LLWU_GetExternalWakeupPinFlag(config->base, i + 1) == true) {
+					data->triggered_pin_sources |= BIT(i);
+					LLWU_ClearExternalWakeupPinFlag(config->base, i + 1);
+				} else {
+					data->triggered_pin_sources &= ~BIT(i);
```
sylvioalves: Sounds this `else` branch should go. It should only be cleaned by `mcux_llwu_clear_wakeup_source_triggered`. Would you review this again to make sure it is correct?

drivers/wuc/wuc_nxp_llwu.c
```diff
@@ -0,0 +1,193 @@
...
+
+static int mcux_llwu_enable_wakeup_source(const struct device *dev, uint32_t id)
+{
+	const struct mcux_llwu_config *config = dev->config;
+	struct mcux_llwu_data *data = dev->data;
+
+	K_SPINLOCK(&data->lock) {
+#if FSL_FEATURE_LLWU_HAS_EXTERNAL_PIN
+		if (NXP_LLWU_IS_PIN_SOURCE(id)) {
+			if (IS_BIT_SET(data->enabled_pins, (NXP_LLWU_GET_PIN_INDEX(id) - 1)) ==
```
sylvioalves, line 66: You could eventually check pin index here before proceeding(or similar but not mandatory I guess).
```
uint32_t pin_idx = NXP_LLWU_GET_PIN_INDEX(id);
if (pin_idx == 0 || pin_idx > 16) {
    return -EINVAL;
}
```

include/zephyr/devicetree/wuc.h
```diff
@@ -0,0 +1,209 @@
...
+ *             #wakeup-ctrl-cells = <1>;
+ *     };
+ *
+ *     n: node {
+ *             wakeup-ctrls = <&wuc 10>;
+ *     };
+ *
+ * Bindings fragment for the vnd,wuc compatible:
+ *
+ *     wakeup-cells:
```
sylvioalves: should this be `wakeup-ctrl-cells:`?

include/zephyr/devicetree/wuc.h
```diff
@@ -0,0 +1,209 @@
...
```
sylvioalves: same here: `wakeup-ctrl-cells:`?

Albort12138: Fixed

include/zephyr/devicetree/wuc.h
Albort12138: Fixed

drivers/wuc/wuc_nxp_llwu.c
Albort12138: The number of supported pins may vary across SoCs that use the same LLWU peripheral. The current SoC supports only 16 pins, but future SoCs may support more. In that case, I prefer to add a macro in dt-bindings/ instead of modifying the device-driver implementation.
So I do not check pin index here.

drivers/wuc/wuc_nxp_llwu.c
Albort12138: Fixed

dts/bindings/wuc/nxp,llwu.yaml
Albort12138: All new created files updated to 2026.

Albort12138:
Hi @bjarki-andreasen and @ceolin,
I would appreciate your help reviewing this PR. I’m hoping it can be merged in time for the 4.4 feature freeze.
Hi @sylvioalves and @henrikbrixandersen,
I’ve added a code snippet in #103863, and based on that, I believe the existing WUC APIs should already satisfy your use cases. Your review and comments would be greatly appreciated.

bjarki-andreasen (approved):

ZhaoxiangJin (approved):

drivers/wuc/Kconfig
sylvioalves: Will you address this?

sylvioalves:
LGTM on driver API. Left a comment in Kconfig entry.

Albort12138:
Hi @ceolin, @henrikbrixandersen, please take a look at this PR, thank you very much.

Albort12138:
Hi @henrikbrixandersen, my understanding is @sylvioalves has reviewed the PR and confirmed that the existing APIs meet his requirements. For reference, please see #103863.
From your perspective, are there any additional concerns? If not, may we remove the “Architecture Review” label and proceed with this PR?

sylvioalves (approved):

PR #100561, opened 2025-12-05
Bluetooth: Host: Fix unable to create periodic adv sync

gangheivt:
In option of HCI_LE_Periodic_Advertising_Create_Sync, bit 2 depend on Periodic Advertising ADI Support feature.

### DIFF
--- a/samples/bluetooth/bap_broadcast_sink/sample.yaml
+++ b/samples/bluetooth/bap_broadcast_sink/sample.yaml
@@ -9,6 +9,7 @@ tests:
       - qemu_x86
       - nrf5340dk/nrf5340/cpuapp
       - nrf5340bsim/nrf5340/cpuapp
+      - sf32lb52_devkit_lcd/sf32lb525uc6
     integration_platforms:
       - qemu_x86
       - nrf5340dk/nrf5340/cpuapp
--- a/samples/bluetooth/bap_broadcast_sink/src/main.c
+++ b/samples/bluetooth/bap_broadcast_sink/src/main.c
@@ -52,9 +52,9 @@ BUILD_ASSERT(IS_ENABLED(CONFIG_SCAN_SELF) || IS_ENABLED(CONFIG_SCAN_OFFLOAD),
 #endif /* CONFIG_SCAN_SELF */
 
 #define PA_SYNC_INTERVAL_TO_TIMEOUT_RATIO 5 /* Set the timeout relative to interval */
-#define PA_SYNC_SKIP                5
-#define NAME_LEN                    sizeof(CONFIG_TARGET_BROADCAST_NAME) + 1
-#define BROADCAST_DATA_ELEMENT_SIZE sizeof(int16_t)
+#define PA_SYNC_SKIP                      5
+#define NAME_LEN                          sizeof(CONFIG_TARGET_BROADCAST_NAME) + 1
+#define BROADCAST_DATA_ELEMENT_SIZE       sizeof(int16_t)
 
 static K_SEM_DEFINE(sem_broadcast_sink_stopped, 0U, 1U);
 static K_SEM_DEFINE(sem_connected, 0U, 1U);
@@ -404,8 +404,7 @@ static void pa_timer_handler(struct k_work *work)
 			pa_state = BT_BAP_PA_STATE_FAILED;
 		}
 
-		bt_bap_scan_delegator_set_pa_state(req_recv_state->src_id,
-						   pa_state);
+		bt_bap_scan_delegator_set_pa_state(req_recv_state->src_id, pa_state);
 	}
 
 	printk("PA timeout\n");
@@ -438,7 +437,7 @@ static uint16_t interval_to_sync_timeout(uint16_t pa_interval)
 
 static int pa_sync_past(struct bt_conn *conn, uint16_t pa_interval)
 {
-	struct bt_le_per_adv_sync_transfer_param param = { 0 };
+	struct bt_le_per_adv_sync_transfer_param param = {0};
 	int err;
 
 	param.skip = PA_SYNC_SKIP;
@@ -606,10 +605,10 @@ static int bis_sync_req_cb(struct bt_conn *conn,
 		}
 
 		/* The stream stopped callback will be called as part of this,
-		* and we do not need to wait for any events from the
-		* controller. Thus, when this returns, the `big_synced`
-		* is back to false.
-		*/
+		 * and we do not need to wait for any events from the
+		 * controller. Thus, when this returns, the `big_synced`
+		 * is back to false.
+		 */
 		err = bt_bap_broadcast_sink_stop(broadcast_sink);
 		if (err != 0) {
 			printk("Failed to stop Broadcast Sink: %d\n", err);
@@ -853,7 +852,6 @@ static struct bt_le_per_adv_sync_cb bap_pa_sync_cb = {
 	.term = bap_pa_sync_terminated_cb,
 };
 
-
 static int init(void)
 {
 	const struct bt_pacs_register_param pacs_param = {
@@ -965,8 +963,7 @@ static int start_adv(void)
 {
 	const struct bt_data ad[] = {
 		BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
-		BT_DATA_BYTES(BT_DATA_UUID16_ALL,
-			      BT_UUID_16_ENCODE(BT_UUID_BASS_VAL),
+		BT_DATA_BYTES(BT_DATA_UUID16_ALL, BT_UUID_16_ENCODE(BT_UUID_BASS_VAL),
 			      BT_UUID_16_ENCODE(BT_UUID_PACS_VAL)),
 		BT_DATA_BYTES(BT_DATA_SVC_DATA16, BT_UUID_16_ENCODE(BT_UUID_BASS_VAL)),
 		BT_DATA(BT_DATA_NAME_COMPLETE, CONFIG_BT_DEVICE_NAME,
@@ -1108,7 +1105,7 @@ static uint32_t select_bis_sync_bitfield(struct base_data *base_sg_data,
 				/* Partial match */
 				printk("Channel allocation match, partial %d\n", combine_alloc);
 			} else {
-				 /* No action required */
+				/* No action required */
 			}
 		}
 
@@ -1208,8 +1205,7 @@ int main(void)
 				 * should start scanning, or wait for PAST
 				 */
 				printk("Waiting for PA sync request\n");
-				err = k_sem_take(&sem_pa_request,
-						 BROADCAST_ASSISTANT_TIMEOUT);
+				err = k_sem_take(&sem_pa_request, BROADCAST_ASSISTANT_TIMEOUT);
 				if (err != 0) {
 					printk("sem_pa_request timed out, resetting\n");
 					continue;
@@ -1230,8 +1226,7 @@ int main(void)
 
 		err = bt_le_scan_start(BT_LE_SCAN_ACTIVE, NULL);
 		if (err != 0 && err != -EALREADY) {
-			printk("Unable to start scan for broadcast sources: %d\n",
-			       err);
+			printk("Unable to start scan for broadcast sources: %d\n", err);
 			return 0;
 		}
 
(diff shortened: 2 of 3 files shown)


subsys/bluetooth/host/scan.c
```diff
@@ -1970,8 +1970,10 @@ int bt_le_per_adv_sync_create(const struct bt_le_per_adv_sync_param *param,
 	}
 
 	if (param->options & BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE) {
-		cp->options |=
-			BT_HCI_LE_PER_ADV_CREATE_SYNC_FP_FILTER_DUPLICATE;
+		if (BT_FEAT_LE_PER_ADV_ADI_SUPP(bt_dev.le.features)) {
+			cp->options |=
+				BT_HCI_LE_PER_ADV_CREATE_SYNC_FP_FILTER_DUPLICATE;
+		}
```
Thalley: If this is the case, then I think we should reject the request early. 

At the top of this function we have a similar check

```c
	if (!BT_FEAT_LE_EXT_PER_ADV(bt_dev.le.features)) {
		return -ENOTSUP;
	}
```

We should not silently ignore parameters from the application, and we should perform parameter checks before attempting to allocate buffers and PA sync objects, so I suggest to move the check to above (but keep the assignment here)

Thalley (changes requested):

subsys/bluetooth/host/scan.c
gangheivt: If so, sample bap_broadcast_sink, should check feature before call this function. But in APP level, the feature is not exposed. This will make sample failed on chipset that do not support this feature.

jhedberg: Arguably such apps shouldn't be attempted to build or run on these platforms then? I'd consider doing that through the sample metadata (`sample.yaml`) and/or the documentation for the sample.

Thalley: Not sure I follow the issue - How is it different from the current check 

```c
	if (!BT_FEAT_LE_EXT_PER_ADV(bt_dev.le.features)) {
		return -ENOTSUP;
	}
```
? 

It is possible for the application to perform this check. You can get the locally supported features by calling `bt_le_get_local_features` and then use `BT_FEAT_LE_PER_ADV_ADI_SUPP` on the result (both of these are in the public API)

cvinayak: > We should not silently ignore parameters from the application

Agree to this. Application can check and use the correct parameters.

> This will make sample failed on chipset that do not support this feature.

Fix the sample to use the feature if supported and document it accordingly.

Thalley: Even though the application can check this, I still think it makes sense to have the same check in the stack (what this PR is doing), as that's how we treat any other API function that can require some optional feature from the controller. 

> Arguably such apps shouldn't be attempted to build or run on these platforms then?

That becomes quite tricky if not building a combined build though. E.g. samples may build and run for e.g. `native_sim` or the appcore of `nRF5340` for multiple controllers, some of which may not support the feature. If we want the app to default to using `BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE` when supported, then I think it also makes sense to have such a check in the application (I think that was the purpose of adding `bt_le_get_local_features`). 

So my suggestion here would be:
1) Add the check for the feature in the stack, and reject if the controller does not support it
2) In samples (and shell?) where `BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE` is being set as an option, make this conditional on `BT_FEAT_LE_PER_ADV_ADI_SUPP`, and if not supported by the controller, default to `BT_LE_PER_ADV_SYNC_OPT_NONE` or whatever makes sense

cvinayak: > Add the check for the feature in the stack, and reject if the controller does not support it

I do not see a need to add redundant check in Host if a Controller is going to fail at the HCI Command Complete providing a spec compliant error code.

We are not yet in a place to have hot-plug of Controllers, hence it is sufficient for application to not use features that it can know to not work in the board. For the samples, the samples shall have the check; and for user applications the check is unnecessary saving precious code space and run time (added transport latencies, if any).

Thalley: > I do not see a need to add redundant check in Host if a Controller is going to fail at the HCI Command Complete providing a spec compliant error code.

I somewhat agree with this, but
1) It's something we do a lot of places (including doing it already for this function
2) The same could be said about any parameter check done by the host that could be done by the controller
3) Failing earlier can be beneficial (no need to waste time allocating buffers etc.)
4) The current approach of transforming (almost) all controller HCI error codes to `-EIO` may be less useful than e.g. `ENOTSUP` (this part we may be able to do something about though).

gangheivt: > Not sure I follow the issue - How is it different from the current check
> 
> ```c
> 	if (!BT_FEAT_LE_EXT_PER_ADV(bt_dev.le.features)) {
> 		return -ENOTSUP;
> 	}
> ```
> 
> ?
> 
> It is possible for the application to perform this check. You can get the locally supported features by calling `bt_le_get_local_features` and then use `BT_FEAT_LE_PER_ADV_ADI_SUPP` on the result (both of these are in the public API)

I updated scan.c based on this.

gangheivt: > Not sure I follow the issue - How is it different from the current check
> 
> ```c
> 	if (!BT_FEAT_LE_EXT_PER_ADV(bt_dev.le.features)) {
> 		return -ENOTSUP;
> 	}
> ```
> 
> ?
> 
> It is possible for the application to perform this check. You can get the locally supported features by calling `bt_le_get_local_features` and then use `BT_FEAT_LE_PER_ADV_ADI_SUPP` on the result (both of these are in the public API)

I updated bap_broadcast_sink sample based on this.

samples/bluetooth/bap_broadcast_sink/src/main.c
```diff
@@ -1025,9 +1022,20 @@ static int stop_adv(void)
 static int pa_sync_create(void)
 {
 	struct bt_le_per_adv_sync_param create_params = {0};
+	struct bt_le_local_features feature;
+	int r;
+
+	r = bt_le_get_local_features(&feature);
+	if (r < 0) {
+		return r;
+	}
```
Thalley:
```suggestion
	int err;

	err = bt_le_get_local_features(&feature);
	if (err < 0) {
		return err;
	}
```

To be consistent with the rest of the code

subsys/bluetooth/host/scan.c
```diff
@@ -2397,8 +2339,7 @@ int bt_le_per_adv_list_add(const bt_addr_le_t *addr, uint8_t sid)
 	bt_addr_le_copy(&cp->addr, addr);
```
Thalley, line 2343: @vignesh-aerlync Do you think the complience check here could be related your recent PR (https://github.com/zephyrproject-rtos/zephyr/pull/99125), or is it something different?

subsys/bluetooth/host/scan.c
```diff
@@ -2276,9 +2219,8 @@ static int default_past_param_set(uint8_t mode, uint16_t skip, uint16_t timeout,
 	return bt_hci_cmd_send_sync(BT_HCI_OP_LE_DEFAULT_PAST_PARAM, buf, NULL);
```
Thalley: @vignesh-aerlync Do you think the complience check here could be related your recent PR (https://github.com/zephyrproject-rtos/zephyr/pull/99125), or is it something different?

subsys/bluetooth/host/scan.c
```diff
@@ -1924,13 +1885,17 @@ int bt_le_per_adv_sync_create(const struct bt_le_per_adv_sync_param *param,
...
-		   param->skip > BT_GAP_PER_ADV_MAX_SKIP ||
-		   param->timeout > BT_GAP_PER_ADV_MAX_TIMEOUT ||
-		   param->timeout < BT_GAP_PER_ADV_MIN_TIMEOUT) {
+	if (param->sid > BT_GAP_SID_MAX || param->skip > BT_GAP_PER_ADV_MAX_SKIP ||
+	    param->timeout > BT_GAP_PER_ADV_MAX_TIMEOUT ||
+	    param->timeout < BT_GAP_PER_ADV_MIN_TIMEOUT) {
 		return -EINVAL;
 	}
 
+	if ((param->options & BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE) &&
```
Thalley:
```suggestion
	if ((param->options & BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE) != 0 &&
```

Thalley (changes requested):
The (actual) changes look overall good, with some minor suggestions/requests

subsys/bluetooth/host/scan.c
gangheivt: It is different.

subsys/bluetooth/host/scan.c
gangheivt: It is different.

subsys/bluetooth/host/scan.c
gangheivt: Updated.

cvinayak (dismissed):
Keep all the formatting changes in separate commit.

samples/bluetooth/bap_broadcast_sink/src/main.c
Thalley, line 1: Ideally the formatting changes in this file should also either be removed or moved to a separate commit

samples/bluetooth/bap_broadcast_sink/sample.yaml
Thalley, line 1: This change should probably be in a separate commit

Thalley (changes requested):

samples/bluetooth/bap_broadcast_sink/sample.yaml
gangheivt: I added a separate commit to format main.c

Thalley (dismissed):

samples/bluetooth/bap_broadcast_sink/src/main.c
```diff
@@ -1025,9 +1022,21 @@ static int stop_adv(void)
...
+		printk("Failed to get local le features (err %d)\n", err);
+		return err;
+	}
 
 	bt_addr_le_copy(&create_params.addr, &broadcaster_addr);
-	create_params.options = BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE;
+	if (BT_FEAT_LE_PER_ADV_ADI_SUPP(feature.features)) {
+		create_params.options = BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE;
+	} else {
+		create_params.options = 0;
```
cvinayak:
```suggestion
		create_params.options = 0U;
```

cvinayak (commented):
LGTM. A nitpick, since the most line of files are formatted anyway, may be add `U` to unsigned assignments?

to `for (uint8_t i = 0U;`  etc.

samples/bluetooth/bap_broadcast_sink/src/main.c
Thalley: I would actually suggest to use
```suggestion
		create_params.options = BT_LE_PER_ADV_SYNC_OPT_NONE;
```

`0U` is a bit misleading since it's an `enum` which, IIRC, is always signed?

gangheivt: Fixed.

samples/bluetooth/bap_broadcast_sink/src/main.c
```diff
@@ -1025,9 +1022,21 @@ static int stop_adv(void)
...
+		printk("Failed to get local le features (err %d)\n", err);
+		return err;
+	}
 
 	bt_addr_le_copy(&create_params.addr, &broadcaster_addr);
-	create_params.options = BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE;
+	if (BT_FEAT_LE_PER_ADV_ADI_SUPP(feature.features)) {
+		create_params.options = BT_LE_PER_ADV_SYNC_OPT_FILTER_DUPLICATE;
+	} else {
+		create_params.options = 0U;
```
Thalley:
```suggestion
		create_params.options = BT_LE_PER_ADV_SYNC_OPT_NONE;
```

Thalley (changes requested):

samples/bluetooth/bap_broadcast_sink/src/main.c
gangheivt: Updated.

Thalley (approved):
Thanks for the updates :) LGTM

cvinayak (approved):

jhedberg (approved):

PR #100565, opened 2025-12-05
scripts: west_commands: packages: Execute pip install in new command

### LINKED ISSUE #100296: west packages pip --install fails with permission error
### Describe the bug

_west packages pip --install_ fails when it tries to uninstall a specific package, even during a clean, first time setup. Log attached.

The permission error indicates the file is already in use by the environment, since it's not an OS permission issue. There's also a follow-up exception while handling this error.

If I uninstall the package manually before doing the above command everything works as expected.

### Regression

- [ ] This is a regression.

### Steps to reproduce

Script to replicate the error (Git Bash):

```
run() { echo -e "\e[32m>>> $*\e[0m"; "$@"; if [ $? -ne 0 ]; then echo -e "\e[31m>>> $*\n\e[0mCommand above failed."; return 1; fi }
run python -m venv zephyrproject/.venv && \
run source zephyrproject/.venv/Scripts/activate && \
run pip install west && \
run west init --manifest-rev v4.3.0 zephyrproject && \
run cd zephyrproject && \
run west update && \
run west zephyr-export && \
run west packages pip --install && \
run cd zephyr && \
run west sdk install && \
run west config --local build.board disco_l475_iot1 && \
run west build -p always samples/basic/blinky
```

### Relevant log output

```shell
Attempting uninstall: ruamel.yaml.clib
    Found existing installation: ruamel.yaml.clib 0.2.15
    Uninstalling ruamel.yaml.clib-0.2.15:
   ━━━━━━━━━━╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  38/147 [ruff]ERROR: Could not install packages due to an OSError: [WinError 5] Access is denied: 'd:\\zephyrproject\\.venv\\lib\\site-packages\\_ruamel_yaml.cp313-win_amd64.pyd'
Check the permissions.


[notice] A new release of pip is available: 25.2 -> 25.3
[notice] To update, run: python.exe -m pip install --upgrade pip
Traceback (most recent call last):
  File "D:\zephyrproject\.venv\Lib\site-packages\west\app\main.py", line 584, in run_command
    self.run_extension(args.command, argv)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\app\main.py", line 739, in run_extension
    self.cmd.run(args, unknown, self.topdir, manifest=self.manifest,
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 config=self.config)
                 ^^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\commands.py", line 200, in run
    self.do_run(args, unknown)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "D:\zephyrproject\zephyr\scripts\west_commands\packages.py", line 127, in do_run
    return self.do_run_pip(args, unknown[1:])
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\zephyr\scripts\west_commands\packages.py", line 160, in do_run_pip
    subprocess.check_call(
    ~~~~~~~~~~~~~~~~~~~~~^
        [sys.executable, "-m", "pip", "install"]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        + list(chain.from_iterable([("-r", r) for r in requirements]))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        + manager_args
        ^^^^^^^^^^^^^^
    )
    ^
  File "C:\Users\cau\AppData\Local\Programs\Python\Python313\Lib\subprocess.py", line 419, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['D:\\zephyrproject\\.venv\\Scripts\\python.exe', '-m', 'pip', 'install', '-r', WindowsPath('D:/zephyrproject/zephyr/scripts/requirements.txt'), '-r', WindowsPath('D:/zephyrproject/modules/hal/espressif/zephyr/requirements.txt'), '-r', WindowsPath('D:/zephyrproject/bootloader/mcuboot/zephyr/requirements.txt'), '-r', WindowsPath('D:/zephyrproject/modules/lib/nanopb/extra/requirements.txt')]' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "D:\zephyrproject\.venv\Scripts\west.exe\__main__.py", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\app\main.py", line 1199, in main
    app.run(argv or sys.argv[1:])
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\app\main.py", line 278, in run
    self.run_command(argv, early_args)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\app\main.py", line 622, in run_command
    f'{quote_sh_list(cpe.cmd)}', fatal=True)
       ~~~~~~~~~~~~~^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\util.py", line 44, in quote_sh_list
    return ' '.join(shlex.quote(s) for s in cmd)
           ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\zephyrproject\.venv\Lib\site-packages\west\util.py", line 44, in <genexpr>
    return ' '.join(shlex.quote(s) for s in cmd)
                    ~~~~~~~~~~~^^^
  File "C:\Users\cau\AppData\Local\Programs\Python\Python313\Lib\shlex.py", line 327, in quote
    if _find_unsafe(s) is None:
       ~~~~~~~~~~~~^^^
TypeError: expected string or bytes-like object, got 'WindowsPath'
```

### Impact

Showstopper – Prevents release or major functionality; system unusable.

### Environment

- Windows 11 build 26200.7171
- Python 3.13.9

### Additional Context

_No response_

pdgendt:
On non-Windows systems execute a new program, replacing the west packages call, when trying to install packages using pip.
For Windows, update the documented way of installing python packages, or using 'west packages pip --install' print a warning.

Docs preview:
https://builds.zephyrproject.io/zephyr/pr/100565/docs/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies
https://builds.zephyrproject.io/zephyr/pr/100565/docs/develop/beyond-GSG.html#keeping-zephyr-updated

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/100296

### DIFF
--- a/doc/develop/getting_started/index.rst
+++ b/doc/develop/getting_started/index.rst
@@ -266,6 +266,10 @@ chosen. You'll also install Zephyr's additional Python dependencies in a
 
             west packages pip --install
 
+         .. note::
+
+            This could downgrade or upgrade west itself.
+
    .. group-tab:: macOS
 
       #. Create a new virtual environment:
@@ -316,6 +320,10 @@ chosen. You'll also install Zephyr's additional Python dependencies in a
          .. code-block:: bash
 
             west packages pip --install
+
+         .. note::
+
+            This could downgrade or upgrade west itself.
 
    .. group-tab:: Windows
 
@@ -389,9 +397,19 @@ chosen. You'll also install Zephyr's additional Python dependencies in a
 
       #. Install Python dependencies using ``west packages``.
 
-         .. code-block:: bat
+         .. tabs::
 
-            west packages pip --install
+            .. code-tab:: bat
+
+               cmd /c scripts\utils\west-packages-pip-install.cmd
+
+            .. code-tab:: powershell
+
+               python -m pip install @((west packages pip) -split ' ')
+
+         .. note::
+
+            This could downgrade or upgrade west itself.
 
 Install the Zephyr SDK
 **********************
--- a/scripts/west_commands/packages.py
+++ b/scripts/west_commands/packages.py
@@ -4,13 +4,15 @@
 
 import argparse
 import os
+import platform
 import subprocess
 import sys
 import textwrap
 from itertools import chain
-from pathlib import Path
+from pathlib import Path, PureWindowsPath
 
 from west.commands import WestCommand
+from west.util import quote_sh_list
 from zephyr_ext_common import ZEPHYR_BASE
 
 sys.path.append(os.fspath(Path(__file__).parent.parent))
@@ -157,11 +159,38 @@ class Packages(WestCommand):
                 self.die("Running pip install outside of a virtual environment")
 
             if len(requirements) > 0:
-                subprocess.check_call(
-                    [sys.executable, "-m", "pip", "install"]
-                    + list(chain.from_iterable([("-r", r) for r in requirements]))
-                    + manager_args
+                cmd = [sys.executable, "-m", "pip", "install"]
+                cmd += chain.from_iterable([("-r", str(r)) for r in requirements])
+                cmd += manager_args
+                self.dbg(quote_sh_list(cmd))
+
+                # Use os.execv to execute a new program, replacing the current west process,
+                # this unloads all python modules first and allows for pip to update packages safely
+                if platform.system() != 'Windows':
+                    os.execv(cmd[0], cmd)
+
+                # Only reachable on Windows systems
+                # Windows does not really support os.execv:
+                # https://github.com/python/cpython/issues/63323
+                # https://github.com/python/cpython/issues/101191
+                # Warn the users about permission errors as those reported in:
+                # https://github.com/zephyrproject-rtos/zephyr/issues/100296
+                cmdscript = (
+                    PureWindowsPath(__file__).parents[1] / "utils" / "west-packages-pip-install.cmd"
                 )
+                self.wrn(
+                    "Updating packages on Windows with 'west packages pip --install', that are "
+                    "currently in use by west, results in permission errors. Leaving your "
+                    "environment with conflicting package versions. Recommended is to start with "
+                    "a new environment in that case.\n\n"
+                    "To avoid this using powershell run the following command instead:\n"
+                    f"{sys.executable} -m pip install @((west packages pip) -split ' ')\n\n"
+                    "Using cmd.exe execute the helper script:\n"
+                    f"cmd /c {cmdscript}\n\n"
+                    "Running 'west packages pip --install -- --dry-run' can provide information "
+                    "without actually updating the environment."
+                )
+                subprocess.check_call(cmd)
             else:
                 self.inf("Nothing to install")
             return
--- a/doc/develop/beyond-GSG.rst
+++ b/doc/develop/beyond-GSG.rst
@@ -136,16 +136,39 @@ Keeping Zephyr updated
 To update the Zephyr project source code, you need to get the latest
 changes via ``git``. Afterwards, run ``west update`` as mentioned in
 the previous paragraph.
-Additionally, in the case of updated or added Python dependencies, running
-``west packages pip --install`` will make sure these are up-to-date.
+Additionally, check for updated or added Python dependencies.
 
-.. code-block:: console
+.. tabs::
 
-   # replace zephyrproject with the path you gave west init
-   cd zephyrproject/zephyr
-   git pull
-   west update
-   west packages pip --install
+   .. group-tab:: Linux/macOS
+
+      .. code-block:: console
+
+         # replace zephyrproject with the path you gave west init
+         cd zephyrproject/zephyr
+         git pull
+         west update
+         west packages pip --install
+
+   .. group-tab:: Windows
+
+      .. tabs::
+
+         .. code-tab:: bat
+
+            :: replace zephyrproject with the path you gave west init
+            cd zephyrproject\zephyr
+            git pull
+            west update
+            cmd /c scripts\utils\west-packages-pip-install.cmd
+
+         .. code-tab:: powershell
+
+            # replace zephyrproject with the path you gave west init
+            cd zephyrproject\zephyr
+            git pull
+            west update
+            python -m pip install @((west packages pip) -split ' ')
 
 Export Zephyr CMake package
 ***************************
--- a/scripts/utils/west-packages-pip-install.cmd
+++ b/scripts/utils/west-packages-pip-install.cmd
@@ -1,0 +1,27 @@
+:: SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+:: SPDX-License-Identifier: Apache-2.0
+
+@echo off
+rem Collect packages from west and install them with a single pip call.
+setlocal enabledelayedexpansion
+
+set "PACKAGES="
+
+for /f "usebackq delims=" %%p in (`west packages pip`) do (
+	if defined PACKAGES (
+		set "PACKAGES=!PACKAGES! %%p"
+	) else (
+		set "PACKAGES=%%p"
+	)
+)
+
+if not defined PACKAGES (
+	echo west packages pip returned no packages to install.
+	exit /b 0
+)
+
+echo Installing packages with: python.exe -m pip install %PACKAGES%
+python.exe -m pip install %PACKAGES%
+set "RESULT=%ERRORLEVEL%"
+
+endlocal & exit /b %RESULT%


scripts/west_commands/packages.py
```diff
@@ -157,10 +156,14 @@ def do_run_pip(self, args, manager_args):
                 self.die("Running pip install outside of a virtual environment")
 
             if len(requirements) > 0:
-                subprocess.check_call(
+                # Use os.execv to execute a new program, replacing the current west call, as there
+                # are known issues when pip tries to install packages which can be in use by west
+                # itself
+                os.execv(
+                    sys.executable,
```
mbolivar: Unrelated to this PR, but neither the before nor after code is logging the actual thing that is being run. That's what ZephyrBinaryRunner.check_call() and friends do (see `self._log_cmd(cmd)` calls throughout core.py). The point is to make it easier for both users to debug and for maintainers to ask for detailed logs to see what is going on in the user environment.

I suggest taking the time to similarly log the command at a verbose level as part of this fix to make any future debugging easier, since it seems like a 'hot' area.

scripts/west_commands/packages.py
```diff
@@ -157,11 +159,28 @@ def do_run_pip(self, args, manager_args):
...
-                subprocess.check_call(
-                    [sys.executable, "-m", "pip", "install"]
-                    + list(chain.from_iterable([("-r", r) for r in requirements]))
-                    + manager_args
+                cmd = [sys.executable, "-m", "pip", "install"]
+                cmd += list(chain.from_iterable([("-r", str(r)) for r in requirements]))
+                cmd += manager_args
+                self.dbg(' '.join(shlex.quote(s) for s in cmd))
+
+                # Use os.execv to execute a new program, replacing the current west call, as there
```
marc-hb:
```suggestion
                # Use os.execv to execute a new program, replacing the current west process, as there
```

scripts/west_commands/packages.py
```diff
@@ -157,11 +159,28 @@ def do_run_pip(self, args, manager_args):
...
+                # are known issues when pip tries to install packages which can be in use by west
+                # itself
+                if platform.system() != 'Windows':
+                    os.execv(cmd[0], cmd)
+
+                # Only reachable on Windows systems
+                # Relevant issues:
```
marc-hb:
```suggestion
                # Windows does not really support os.execv:
```

scripts/west_commands/packages.py
```diff
@@ -157,11 +159,28 @@ def do_run_pip(self, args, manager_args):
...
-                    [sys.executable, "-m", "pip", "install"]
-                    + list(chain.from_iterable([("-r", r) for r in requirements]))
-                    + manager_args
+                cmd = [sys.executable, "-m", "pip", "install"]
+                cmd += list(chain.from_iterable([("-r", str(r)) for r in requirements]))
+                cmd += manager_args
+                self.dbg(' '.join(shlex.quote(s) for s in cmd))
+
+                # Use os.execv to execute a new program, replacing the current west call, as there
+                # are known issues when pip tries to install packages which can be in use by west
```
marc-hb: Please add a reference to the issue. A 1 or 2 lines summary would be even better.

scripts/west_commands/packages.py
```diff
@@ -157,11 +159,28 @@ def do_run_pip(self, args, manager_args):
...
+                    os.execv(cmd[0], cmd)
+
+                # Only reachable on Windows systems
+                # Relevant issues:
+                # https://github.com/python/cpython/issues/63323
+                # https://github.com/python/cpython/issues/101191
+                self.wrn(
+                    "On Windows, installing packages from 'west packages' has known limitations; "
+                    "if packages conflict with those currently in use\n"
+                    "Using powershell instead run:\n"
```
marc-hb: Same as above: "known" limitations but no reference. What sort of conflict? A version conflict?

marc-hb: What happens when this fails, doesn't it leave things in a messy state? Is it really worth taking that risk? I mean, compared to just refusing to run and printing the correct Powershell command instead. Re-installing packages in use sounds like a pretty bad idea.

BTW why are these packages being re-installed? Version difference? Messy... If that is the reason, then maybe this extension could compare versions and run only after making sure this will not trigger any re-installation.

marc-hb (commented):
Re-installing packages in use sounds like a pretty bad idea.

More generally speaking, this convenience shortcut looks like a lot of hassle, maintenance and risk to save just a tiny bit of typing.

Maybe this shortcut does not belong to `west`. Maybe there is a pratical limit to the number of blades a swiss army knife can have. Maybe the desire to have every single Zephyr-related command starting with `west ...` is not such a good idea.

marc-hb:
From a quick chat with @pdgendt : I have zero concern with `pip install $(west packages pip)`, that's neat. It's really the "integrated installation" that looks like a bad design.

scripts/west_commands/packages.py
pdgendt: Packages _can_ be re-installed if some Zephyr module has updated dependencies. `west packages pip` collects these and hands them over to `pip`. We don't have control over downstream projects.

Luckily `west` doesn't have many dependencies, but the issue was first noticed with NXP's `spsdk` which uses pinned dependencies.

scripts/west_commands/packages.py
```diff
@@ -157,11 +159,27 @@ def do_run_pip(self, args, manager_args):
                 self.die("Running pip install outside of a virtual environment")
 
             if len(requirements) > 0:
-                subprocess.check_call(
-                    [sys.executable, "-m", "pip", "install"]
-                    + list(chain.from_iterable([("-r", r) for r in requirements]))
-                    + manager_args
+                cmd = [sys.executable, "-m", "pip", "install"]
+                cmd += list(chain.from_iterable([("-r", str(r)) for r in requirements]))
```
yvesll: Convert path to str here is more safe. Otherwise it may fail when subprocess raise an exception:
```shell
  File "C:\Users\cau\AppData\Local\Programs\Python\Python313\Lib\shlex.py", line 327, in quote
    if _find_unsafe(s) is None:
       ~~~~~~~~~~~~^^^
TypeError: expected string or bytes-like object, got 'WindowsPath'
```

pdgendt: I'm not sure I get the statement? Did you get the error with the current version?

yvesll: No, I have not encountered the permission issue before. But if someone run fail in subprocess.check_call, this exception may be raised. You can see the end of the log in https://github.com/zephyrproject-rtos/zephyr/issues/100296

yvesll: Sorry, I did not notice that you have added a `str()`. I think it would be better to cp the fix here: https://github.com/zephyrproject-rtos/west/blob/76890ef68124369c9b3ccd1ef968e47ab5f3183f/src/west/util.py#L44

pdgendt: No, it expects a list of strings. But we can make use of it here.

yvesll: But check_call accept Path like object, if other extensions use check call and raise `CalledProcessError`, it will raise a shlex error here: https://github.com/zephyrproject-rtos/west/blob/76890ef68124369c9b3ccd1ef968e47ab5f3183f/src/west/app/main.py#L671

pdgendt: "Our" `check_call` doesn't, it expects a list of strings

https://github.com/zephyrproject-rtos/zephyr/blob/3ea2951bc01c92466a5fed228f0c70aa96dfec72/scripts/west_commands/runners/core.py#L894

yvesll: I saw it's a check_call api from `ZephyrBinaryRunner`. Type hint may not cause an error and both check_call from subprocess and WestCommand did not have any type hint.

marc-hb:
```suggestion
                cmd += chain.from_iterable([("-r", str(r)) for r in requirements])
```
The `list()` adds mental overhead.

scripts/west_commands/packages.py
marc-hb: So what happens when installation fails half way? Only half the requirements are installed and the user gets to keep all the pieces? That minor install convenience is just not worth this risk.

Even outside Windows, is it OK for `west packages pip --install` to update its own dependencies, potentially modifying its own behavior? I'm not sure this ever was OK.

marc-hb (commented):
I think the only reliable and portable way to implement this feature would be:

- Use `pip --dry-run` or similar to determine whether `west packages pip --install` would affect one of `west`'s own dependencies and behavior. In other words: find any conflicting dependency.
- If no conflict found: go ahead and install
- If any conflict found: print them and ask the user to manually upgrade or downgrade those west dependencies (and only them) first. Then, the user can run `west packages pip --install` again. Or, run `pip` directly.


PS: this brings back memories of west [aeca29ac9fd3](https://github.com/zephyrproject-rtos/west/commit/aeca29ac9fd3)... Windows: always the problem child. Off-topic sorry.

pdgendt:
> > But this behavior is how pip works, I'm not sure we can accurately predict this, maybe different versions would work differently.
> 
> If we can't predict what pip mess this might leave behind, then that feature should be removed on Windows until it can be implemented properly. `--install` is not even a critical feature; it's just a convenience shortcut to save some typing!

For Linux, macOS and Powershell on WIndows users, yes. For Windows users running it from `cmd.exe`, it was supposed to be more than that, and a follow-up is needed.

> Alternatively, here's an "actionable" request to be more direct and hopefully clearer: please replace the wording "can result in permission errors" in the new Windows warning this PR adds to the more technically accurate text: "Sorry, we have no idea what kind of pip mess we just got you into if you got a permission error. It could be minor or major but we don't know. Best of luck figuring it out."

Sarcasm aside, thanks.

> > @marc-hb could you please add more actionable requests instead?
> 
> I'm sorry I'm trying to have a design discussion in a quick fix PR, I agree it's not a good place. But I'm not the one trying to quick fix a potentially broken design.

We're all in agreement that there is a flaw with the [_documented_](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies) convenient way of installing python dependencies for Zephyr. We can blame designs and have discussions all we want, but it's a fix nonetheless. I want to incrementally improve things, until the next bug pops up we might have overlooked.

> > Why not, if Zephyr would specify west>=1.6.0, the current extension with older versions can handle that just fine.
> 
> Afraid you lost me there. Trying to move the design discussion to [#100296 (comment)](https://github.com/zephyrproject-rtos/zephyr/issues/100296#issuecomment-3635103334)

`west` is listed in the requirements already?

```shell
$ west packages pip --install -- --dry-run | grep west
Requirement already satisfied: west>=0.14.0 in /home/pdgendt/zephyrproject/.venv/lib/python3.12/site-packages (from -r /home/pdgendt/zephyrproject/zephyr/scripts/requirements-base.txt (line 37)) (1.5.0)
```

How do you upgrade `pip`? Commonly like this:
```shell
$ pip install --upgrade pip
```

scripts/west_commands/packages.py
pdgendt: It is OK if `os.execv` is used on non-Windows platforms.

scripts/west_commands/packages.py
marc-hb: `quote_sh_list` enhancement submitted in https://github.com/zephyrproject-rtos/west/pull/896/, thanks!

marc-hb:
Trying to summarize my current perspective.

- os.exec() looks good, just add the words "this could downgrade or upgrade west" in the GSG (see  #100296)

- On Windows where os.exec() is not available, the `--install` mere convenience should be disabled _until we have some knowledge of how bad are the things it can leave behind_. It's not user-friendly at all to ask users to experiment with supposedly user-friendly features that actually leave an unknown mess behind. There is barely anything less user-friendly than cleaning up an unknown quantity of garbage. The alternative is the powershell and cmd scripts that you already wrote and tested. They are just a little more typing than the `--install` option and that's all? Then they should be the preferred option. More typing is a bit boring but it's nothing compared to failures.

> How do you upgrade pip? Commonly like this:
> `$ pip install --upgrade pip`

Great observation: how does `pip` solve with this problem on Windows? Does it have to, is this example relevant here?

doc/develop/getting_started/index.rst
```diff
@@ -266,6 +266,10 @@ chosen. You'll also install Zephyr's additional Python dependencies in a
 
             west packages pip --install
 
+         .. note::
+
+            This could downgrade or upgrade west itself.
```
marc-hb, line 271:
```suggestion
            This could downgrade or upgrade west itself and it can fail on Windows.
```

pdgendt: This note was added for the "Ubuntu" tab, don't see the importance of adding it.

pdgendt:
> > If this option on Windows is to be removed, it should be deprecated first.
> 
> So, should `--install` be deprecated on Windows? Or is it fixable somehow? With what approximate magnitude of time and effort? These questions matter. This is exactly why I wrote this earlier:
> 
> > I'm not opposed to quick fixes in general but only as long as there's a very clear understanding of all issues and trade-offs and a relatively clear "vision" of what the longer-term fix(s) will be. Otherwise quick fixes can add fuel to the fire and make longer term solutions even more difficult because they can add more user expectations (depending on the version of the extension), because they make the code more complex, because they may pull in the opposite direction, etc

I kind of questioned if it needs to be removed first 😅. My opinion is that it shouldn't, it will succeed in most cases. The warning should make it clear that _if_ something goes wrong, users know where to look and have an alternative.

I think the important bit is the _getting started_ documentation, where users will blindly copy the commands.

> > It was relevant in that it is familiar for Python users to have tools that upgrade the tool itself.
> 
> "was" - so it's not relevant anymore?
> 
> Yes there are many tools able to upgrade themselves but it's never an easy thing. How does `pip` achieves this on Windows and is there any relevant lesson to learn?

_"Was"_ simply to refer to a comment made earlier, it _is_ to prove a point that self-updating tools isn't bad or uncommon.

marc-hb (approved):

yvesll (approved):

carlescufi (approved):

PR #100505, opened 2025-12-04
boards: nordic: nrf54h20dk: Add cpuapp TCM memory section

nordic-krch:
Add memory section which is placed in cpuapp TCM (Tightly Coupled Memory) RAM0 which is fast, non-cacheable.

~~Contains #100564 which fixes failing CI.~~

PR #100503, opened 2025-12-04
manifest: Update segger with enabling DT sections

nordic-krch:
Update segger with a change in SEGGER_RTT_Conf.h that allows to use DT memory section for RTT data.

Added help section to SEGGER_RTT_SECTION_CUSTOM_DTS_REGION Kconfig explaining what alias need to be added to DTS.

PR #100570, opened 2025-12-05
Add support for RTT on nrf54h20 for standard and STM logging

nordic-krch:
Enabled RTT on nrf54h20 cpuapp. RTT does not work with cacheable RAM so non-cacheable area need to be used. TCM RAM is used for that.

Add RTT backend implementation to STM logging processing (debug_nrf_etr driver).

PR in draft as it is based on few other:
- #100503 which enables DTS section to be used for RTT
- #100505 which allows using TCM RAM memory section
- #100564 fix for failing CI


kl-cruz:
@nordic-krch Related PRs have been merged. Please rebase and prepare for review

asmellby (approved):

nika-nordic (approved):

PR #89248, opened 2025-04-29
net: core: Rework l3 handling

go2sh:
This PR rework the l3 handling from the ethernet l2 driver into the net_core. For that it does:

- Loopback flag in net_pkt
- Dedicated l3 handler for ipv4
- Dedicated l3 handler for ipv6
- L3 handler for IEEE802.15.4
- Allow L3 handlers to not specify an l2 to handle DUMMY and VIRTUAL l2.

The idea is to allow virutal intefaces to go throug the l3 handlers. So for example (a potential macsec l2 inteface), the flow could look like: l2 ethernet -> l2 macsec ->l2 vlan -> l3 handler.

PR #100576, opened 2025-12-05
openthread: platform: infra_if: Add IPV4 packet filtering rule for NAT64 translator

Cristib05:
This commit aims to implement an L3 handler for NAT64 translator. If packets are consumed, those are not processed anymore by network stack.
NAT64 packets are created by OpenThread stack and sent on backbone interface using a raw socket.

This commit attempts to fix an issue where a TCP connection is initiated by an OpenThread node, but discarded by network stack since there is no active known connection.

### DIFF
--- a/modules/openthread/platform/infra_if.c
+++ b/modules/openthread/platform/infra_if.c
@@ -26,6 +26,7 @@
 
 #if defined(CONFIG_OPENTHREAD_NAT64_TRANSLATOR)
 #include <zephyr/net/icmp.h>
+#include <zephyr/net/net_pkt_filter.h>
 #include <openthread/nat64.h>
 #endif /* CONFIG_OPENTHREAD_NAT64_TRANSLATOR */
 
@@ -45,9 +46,19 @@ static void handle_ra_from_ot(const uint8_t *buffer, uint16_t buffer_length);
 static struct zsock_pollfd sockfd_raw[MAX_SERVICES];
 static void raw_receive_handler(struct net_socket_service_event *evt);
 static void remove_checksums_for_eth_offloading(uint8_t *buf, uint16_t len);
+static bool infra_if_nat64_try_consume_packet(struct npf_test *test, struct net_pkt *pkt);
 static int raw_infra_if_sock = -1;
 
 NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(handle_infra_if_raw_recv, raw_receive_handler, MAX_SERVICES);
+
+struct ot_nat64_pkt_filter_test {
+	struct npf_test test;
+};
+/* Packet filtering rules for NAT64 translator section */
+static struct ot_nat64_pkt_filter_test ot_nat64_drop_rule_check = {
+	.test.fn = infra_if_nat64_try_consume_packet};
+/* Drop all traffic destined to and consumed by NAT64 translator */
+static NPF_RULE(ot_nat64_drop_pkt_process, NET_DROP, ot_nat64_drop_rule_check);
 #endif /* CONFIG_OPENTHREAD_NAT64_TRANSLATOR */
 
 otError otPlatInfraIfSendIcmp6Nd(uint32_t aInfraIfIndex, const otIp6Address *aDestAddress,
@@ -318,6 +329,9 @@ otError infra_if_nat64_init(void)
 						 ARRAY_SIZE(sockfd_raw), NULL) == 0,
 		     error = OT_ERROR_FAILED);
 
+	npf_insert_ipv4_recv_rule(&ot_nat64_drop_pkt_process);
+	npf_append_ipv4_recv_rule(&npf_default_ok);
+
 exit:
 	return error;
 }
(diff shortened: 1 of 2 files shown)


modules/openthread/platform/infra_if.c
```diff
@@ -366,4 +366,52 @@ otError infra_if_send_raw_message(uint8_t *buf, uint16_t len)
...
+	VerifyOrExit(message != NULL);
+
+	for (buf = pkt->buffer; buf; buf = buf->frags) {
+		if (otMessageAppend(message, buf->data, buf->len) != OT_ERROR_NONE) {
+			otMessageFree(message);
+			ExitNow();
+		}
+	}
+
+	if (otNat64Send(ot_instance, message) == OT_ERROR_NONE) {
```
rlubos, line 463: For my own understanding - does `otNat64Send()` maintain a NAT table and only succeeds if there's a valid NAT entry for a given address/port? Just wondering if we don't end up in a situation, where we forward all IPv4 packets to OpenThread and won't be able to communicate with BR over IPv4 at all.

modules/openthread/platform/infra_if.c
```diff
@@ -366,4 +366,52 @@ otError infra_if_send_raw_message(uint8_t *buf, uint16_t len)
...
+		openthread_mutex_unlock();
+		return NET_OK;
+	}
+
+exit:
+	openthread_mutex_unlock();
+	return NET_CONTINUE;
+}
+
+ETH_NET_L3_REGISTER(OT_NAT64_CONSUMER, NET_ETH_PTYPE_IP, infra_if_nat64_try_consume_packet);
```
rlubos: `ETH_NET_L3_REGISTER()` wasn't intented to have multiple handlers registered per packet type (@jukkar please correct me if I'm wrong). The respective packet handlers are just put in an iterable section, I'm not sure what's the logic behind the ordering of the iterable section entries, but it's likely not guaranteed that your handler will execute before the actual IPv4 one.

Perhaps a better fit here would be the IP packet filtering feature we have (https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/net/net_pkt_filter.h, I realized recently it's not documented well so sorry about that). You can install filtering rules to decide whether the IP stack should continue to process the packet or not. In the rule you could do the NAT64 processing, and inform the IP stack to drop the packet if it was forwarded to OpenThread.

After enabling `CONFIG_NET_PKT_FILTER` and `CONFIG_NET_PKT_FILTER_IPV4_HOOK` you should be able to define a IP filtering rule (please cosider this a pseudo-code as I didn't compile it):
```c
static bool nat64_cb(struct npf_test *test, struct net_pkt *pkt);
static struct npf_test nat64_rule = { 	.fn = nat64_cb };
static NPF_RULE(nat64_rx, NET_DROP, nat64_rule);
```

And then install it during intialization:
```c
npf_append_ipv4_recv_rule(&nat64_rx);
```

`nat64_cb()` should do the same thing you currently do in `infra_if_nat64_try_consume_packet()`, just return `true` if the packet was forwarded to OT and should be dropped by the stack, `false` otherwise.

IPv4 filtering rules are executed directly from the IPv4 stack so there should be no confusion about order of execution:
https://github.com/nrfconnect/sdk-zephyr/blob/c09c6abd11695d6b9c2ea24c4e88822eee9ff3f9/subsys/net/ip/ipv4.c#L344

modules/openthread/platform/infra_if.c
Cristib05: Yes, it maintains a table, and it returns success if there is a valid NAT entry. Not every IPV4 packet will be consumed

modules/openthread/platform/infra_if.c
Cristib05: Thanks, I will have a look at it and update this PR. I was not aware of ETH_NET_L3_REGISTER limitation, or maybe I have not properly understood this. I thought I can make use of it since there was an iterable section for it

rlubos: Or actually, looking at https://docs.zephyrproject.org/latest/kernel/iterable_sections/index.html, the doc page claim that the section entries are ordered by name. But that puzzles me even more as alphabetically, `OT_NAT64_CONSUMER` should be after `IPv4`, which is the handler name for the regular IPv4 stack:
https://github.com/zephyrproject-rtos/zephyr/blob/476483075c3a6414e506bd7a339e4e76faf52405/subsys/net/l2/ethernet/ethernet.c#L454

So at least in theory, your handler would be executed after the IPv4 stack processing, not sure how come this works.

rlubos: Ah, answering to myslef, that's because that IPv4 handler doesn't actually process the packet, but just sets the family on a packet...
https://github.com/zephyrproject-rtos/zephyr/blob/476483075c3a6414e506bd7a339e4e76faf52405/subsys/net/l2/ethernet/ethernet.c#L435-L447

Although I still think it'd be more "future-proof" and appropriate to use the packet filter for this case, there already were attempts to change how IPv4/6 handlers are registered in general (https://github.com/zephyrproject-rtos/zephyr/pull/89248, but for some reaso n the author went radio silent).

rlubos (dismissed):

samples/net/openthread/border_router/prj.conf
```diff
@@ -24,6 +25,10 @@ CONFIG_NET_CONNECTION_MANAGER_MONITOR_STACK_SIZE=2048
 # if there are more than 1 handlers defined.
 CONFIG_ZVFS_OPEN_ADD_SIZE_NET_SAMPLE=16
 
+#Packet filtering options (Required by NAT64)
+CONFIG_NET_PKT_FILTER=y
+CONFIG_NET_PKT_FILTER_IPV4_HOOK=y
```
pdgendt: Would it make sense to `imply` these in border router configs? For example, `CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER`?

Cristib05: Good idea, I guess it makes more sense to `select` them when NAT64 translator is enabled. I want to force them to be set, otherwise if someone sets them to `n` failures will occur.

pdgendt (dismissed):

rlubos (dismissed):

jukkar (dismissed):

pdgendt (approved):

rlubos (approved):

PR #100578, opened 2025-12-05
drivers: rtc: sf32lb: fix alarm_is_pending error

ck-telecom:
Fix alarm_is_pending error

### DIFF
--- a/drivers/rtc/rtc_sf32lb.c
+++ b/drivers/rtc/rtc_sf32lb.c
@@ -11,6 +11,7 @@
 #include <zephyr/device.h>
 #include <zephyr/drivers/rtc.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/sys/atomic.h>
 
 #include <register.h>
 
@@ -59,6 +60,7 @@ struct rtc_sf32lb_alarm_cb {
 struct rtc_sf32lb_data {
 #ifdef CONFIG_RTC_ALARM
 	struct rtc_sf32lb_alarm_cb alarm_cb;
+	ATOMIC_DEFINE(is_pending, 1);
 #endif
 };
 
@@ -79,7 +81,10 @@ static void rtc_irq_handler(const struct device *dev)
 
 	if (isr & RTC_ISR_ALRMF) {
 		sys_clear_bit(config->base + RTC_ISR, RTC_ISR_ALRMF_Pos);
+
 #ifdef CONFIG_RTC_ALARM
+		atomic_set_bit(&data->is_pending, 0);
+
 		if (data->alarm_cb.cb) {
 			data->alarm_cb.cb(dev, 0, data->alarm_cb.user_data);
 		}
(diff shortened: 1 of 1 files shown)


drivers/rtc/rtc_sf32lb.c
```diff
@@ -28,16 +28,24 @@ LOG_MODULE_REGISTER(rtc_sf32lb, CONFIG_RTC_LOG_LEVEL);
 #define SYS_CFG_RTC_TR offsetof(HPSYS_CFG_TypeDef, RTC_TR)
 #define SYS_CFG_RTC_DR offsetof(HPSYS_CFG_TypeDef, RTC_DR)
 
+#define SF32LB_LXT32_NODE         DT_NODELABEL(lxt32)
```
gmarull: you should obtain frequency using clock control API, then adjust as needed. This creates a hidden DT dependency in C code.

gmarull (changes requested):

fabiobaltieri (dismissed):

drivers/rtc/rtc_sf32lb.c
```diff
@@ -308,18 +312,19 @@ static int rtc_sf32lb_alarm_get_time(const struct device *dev, uint16_t id, uint
...
 	}
 
-	if (sys_test_bit(config->base + RTC_ISR, RTC_ISR_ALRMF_Pos)) {
-		sys_clear_bit(config->base + RTC_ISR, RTC_ISR_ALRMF_Pos);
-		return 1;
-	}
+	__disable_irq();
+	ret = data->is_pending ? 1 : 0;
+	data->is_pending = false;
+	__enable_irq();
```
gmarull: could IRQ be masked instead of taking a global lock?

ck-telecom: this could happened in multi-thread with race conditions

gmarull: could an atomic work instead, like test and clear?

ck-telecom: atomic should be okay, but seems that   it will disable irq if the hardware not support atomic instructions, anyway it should be okay with atomic type

drivers/rtc/rtc_sf32lb.c
```diff
@@ -308,18 +312,17 @@ static int rtc_sf32lb_alarm_get_time(const struct device *dev, uint16_t id, uint
 
 static int rtc_sf32lb_alarm_is_pending(const struct device *dev, uint16_t id)
 {
-	const struct rtc_sf32lb_config *config = dev->config;
+	struct rtc_sf32lb_data *data = dev->data;
+	int ret = 0;
```
gmarull:
```suggestion
```

drivers/rtc/rtc_sf32lb.c
```diff
@@ -308,18 +312,17 @@ static int rtc_sf32lb_alarm_get_time(const struct device *dev, uint16_t id, uint
...
 
-	if (sys_test_bit(config->base + RTC_ISR, RTC_ISR_ALRMF_Pos)) {
-		sys_clear_bit(config->base + RTC_ISR, RTC_ISR_ALRMF_Pos);
-		return 1;
-	}
+	ret = atomic_test_bit(&data->is_pending, 0);
+	atomic_clear_bit(&data->is_pending, 0);
 
-	return 0;
+	return ret;
```
gmarull:
```suggestion
	return (int)atomic_test_and_clear_bit(&data->is_pending, 0);
```

drivers/rtc/rtc_sf32lb.c
```diff
@@ -59,6 +59,7 @@ struct rtc_sf32lb_alarm_cb {
 struct rtc_sf32lb_data {
 #ifdef CONFIG_RTC_ALARM
 	struct rtc_sf32lb_alarm_cb alarm_cb;
+	atomic_t is_pending;
```
gmarull:
```suggestion
	ATOMIC_DEFINE(is_pending, 1);
```

gmarull: include `zephyr/sys/atomic.h`

gmarull (approved):

ck-telecom:
@fabiobaltieri ping

bjarki-andreasen (approved):

Issue #100572, opened 2025-12-05
samples: drivers: crypto: Regression in mbedtls variant - encrypt/decrypt failed


mars-low:
### Describe the bug

After https://github.com/zephyrproject-rtos/zephyr/pull/97572, `samples/drivers/crypto/sample.drivers.crypto.mbedtls` doesn't yield correct result for a few platforms anymore. https://github.com/zephyrproject-rtos/zephyr/blob/cc68eaef2b08d39dcf3962965225d495fe0f3257/samples/drivers/crypto/sample.yaml#L9-L22 I verified it during a simulation with `qemu_riscv64`, `qemu_x86` and with the hardware `beaglev_fire/polarfire/u54`. For all these platforms, before the PR, sample used to return `Match` for all supported modes of operation (`ECB`, `CBC`, `CCM`, `GCM`). After the PR, they return `Failed` during simulation.

There are two interesting things about the problem:
1. I don't observe regression for `native_sim` platform which is an integration platform for `sample.drivers.crypto.mbedtls` variant.
2. Even if the sample returned `Failed` for `native_sim`, it still wouldn't be detected by twister due to the regex pattern, which ignores the result. It only checks if the tests were ran and they are executed unconditionally, so it will never fail unless a critical error occurs.

After running on the real board `beaglev_fire/polarfire/u54`, it fails with the fatal error in Zephyr, what may suggest the source of the problem.

<details>

<summary>Show hardware log after the PR on `beaglev_fire/polarfire/u54` (FAILED)</summary>

```txt
HSS: decompressing from eNVM to L2 Scratch ... Passed
DDR training ... Passed ( 5777 ms)

---------------------------------
--        BeagleV-Fire         --
---------------------------------

[5.840061] PolarFire(R) SoC Hart Software Services (HSS) - version 0.99.36-BVF-0.3.0
MPFS HAL version 2.2.104 / DDR Driver version 0.4.023 / Mi-V IHC version 0.1.1 / BOARD=bvf
(c) Copyright 2017-2022 Microchip FPGA Embedded Systems Solutions.

incorporating OpenSBI - version 1.2
(c) Copyright 2019-2022 Western Digital Corporation.

[5.871049] Build ID: 3f0ed536ff6a0aaef3c9b81896adaee39ec4d8c1
[5.877906] Built with the following tools:
[...]

PR #98660, opened 2025-10-31
boards: mpfs_icicle: add renode simulation support

mars-low:
Add Renode simulation support for `mpfs_icicle`.

Renode 1.16.0 required for the support was activated on Zephyr CI in https://github.com/zephyrproject-rtos/zephyr/commit/8e687fc0dd6e20dd75f9e6e8c29287cc81cfefe9.

PR #100580, opened 2025-12-05
tests: crypto: mbedtls: Exclude mpfs_icicle and hifive_unmatched

mars-low:
The regression causes system halt in `sample.drivers.crypto.mbedtls` for `mpfs_icicle` and `hifive_unmatched` with the error `mcause: 6, Store/AMO address misaligned`.

Exclude platforms that support simulation as these errors impact CI. Non-simulated targets are also impacted, for example `beaglev_fire/polarfire/u54`, but these are not excluded to not hide the underlying problem.

The underlying issue is tracked in https://github.com/zephyrproject-rtos/zephyr/issues/100572.

### DIFF
--- a/samples/drivers/crypto/sample.yaml
+++ b/samples/drivers/crypto/sample.yaml
@@ -12,6 +12,10 @@ tests:
     extra_args: EXTRA_CONF_FILE=prj_mtls_shim.conf
     integration_platforms:
       - native_sim
+    platform_exclude:
+      - hifive_unmatched/fu740/s7
+      - hifive_unmatched/fu740/u74
+      - mpfs_icicle/polarfire/e51
     harness_config:
       type: multi_line
       regex:


valeriosetti (changes requested):
Question: is still problem still valid after https://github.com/zephyrproject-rtos/zephyr/pull/100712?
Locally the following command succeeds:
`twister -c -p hifive_unmatched/fu740/s7 -p hifive_unmatched/fu740/u74 -p mpfs_icicle/polarfire/e51 -T samples/drivers/crypto/ -s sample.drivers.crypto.mbedtls`

mars-low:
I confirm that after your last update, https://github.com/zephyrproject-rtos/zephyr/pull/100712 allows to pass tests for all platforms where I previously observed an error:

Simulated in renode:
* hifive_unmatched/fu740/s7
* hifive_unmatched/fu740/u74
* mpfs_icicle/polarfire/e51 (to be added in https://github.com/zephyrproject-rtos/zephyr/pull/98660)

Real hardware:
* beaglev_fire/polarfire/u54

Thanks!

PR #100583, opened 2025-12-05
drivers: crypto: sf32lb: Add sf32lb crypto drivers

HalfSweet:
Including the AES encryption algorithm and SHA hash algorithms.
AES supports the Cartesian product of [AES128, AES192, AES256] with [ECB, CTR, CBC].
SHA algorithms support SHA224 and SHA256.


drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,699 @@
...
+static DEVICE_API(crypto, crypto_sifli_funcs) = {
+#if defined(CONFIG_CRYPTO_SIFLI_AES)
+	.cipher_begin_session = crypto_sifli_session_setup,
+	.cipher_free_session = crypto_sifli_session_free,
+#endif
+#if defined(CONFIG_CRYPTO_SIFLI_HASH)
+	.hash_begin_session = crypto_sifli_hash_begin_session,
+	.hash_free_session = crypto_sifli_hash_free_session,
+#endif
+	.cipher_async_callback_set = NULL,
```
ck-telecom, line 1105:
```suggestion
```

ck-telecom (changes requested):

ck-telecom (dismissed):

drivers/crypto/crypto_sf32lb_priv.h
```diff
@@ -0,0 +1,139 @@
...
+	uint8_t key_len;
+	uint8_t mode;
+	bool in_use;
+};
+
+/* Session structure for hash operations */
+struct crypto_sifli_hash_session {
+	uint8_t algo;
+	bool in_use;
+};
```
valeriosetti, line 142: Is it worth having 2 different structures for sessions? This cause code duplication like `crypto_sifli_hash_get_unused_session_index` and `crypto_sifli_get_unused_session_index` or `crypto_sifli_session_free` and `crypto_sifli_hash_free_session`.

What about instead having a single structure for sessions and adding a `union` inside it? Something like:
```
struct crypto_sifli_session {
	bool in_use;
	union {
#if defined(CONFIG_CRYPTO_SIFLI_AES)
		struct {
			uint32_t key[SIFLI_AES_MAX_KEY_LEN / sizeof(uint32_t)];
			uint8_t key_len;
			uint8_t mode;
		} cipher;
#endif /* CONFIG_CRYPTO_SIFLI_AES */
#if defined(CONFIG_CRYPTO_SIFLI_HASH)
		struct {
			uint8_t algo;
		} hash;
#endif /* CONFIG_CRYPTO_SIFLI_HASH */
	};
};
```

This way you can share functions for getting/releasing a session

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,702 @@
...
+{
+	sys_write32(AES_ACC_COMMAND_AES_ACC_RESET, base + AES_COMMAND_OFFSET);
+	sys_write32(0, base + AES_COMMAND_OFFSET);
+}
+
+static void crypto_sifli_hash_reset(uint32_t base)
+{
+	sys_write32(AES_ACC_COMMAND_HASH_RESET, base + AES_COMMAND_OFFSET);
+	sys_write32(0, base + AES_COMMAND_OFFSET);
+}
```
valeriosetti: Since you nicely added guards in the code for when cipher and hash are enabled in Kconfig, I think it would be nice to move these functions as well under those guards. So
- `crypto_sifli_aes_busy` and `crypto_sifli_aes_reset` under `#if defined(CONFIG_CRYPTO_SIFLI_AES)`
- `crypto_sifli_hash_busy` and `crypto_sifli_hash_reset` under `#if defined(CONFIG_CRYPTO_SIFLI_HASH)`

Wdyt?

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,702 @@
...
+
+	/* Data memory barrier before starting DMA */
+	__DSB();
+
+	/* Start hash operation */
+	sys_write32(AES_ACC_COMMAND_HASH_START, base + AES_COMMAND_OFFSET);
+
+	/* Wait for completion */
+	if (!WAIT_FOR(crypto_sifli_hash_busy(base) != true, CRYPTO_SIFLI_TIMEOUT_US,
+		      k_busy_wait(1))) {
```
valeriosetti, line 945: Why busy waiting when there seems to be support from interrupt (`AES_ACC_IRQ_HASH_DONE_STAT` on line 572)?

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,702 @@
...
+	}
+	sys_write32(aes_setting, base + AES_AES_SETTING_OFFSET);
+
+	/* Start operation */
+	sys_write32(sys_read32(base + AES_COMMAND_OFFSET) | AES_ACC_COMMAND_START,
+		    base + AES_COMMAND_OFFSET);
+
+	/* Wait for completion */
+	if (!WAIT_FOR(crypto_sifli_aes_busy(base) != true, CRYPTO_SIFLI_TIMEOUT_US,
+		      k_busy_wait(1))) {
```
valeriosetti, line 247: As for the hash comment I added below, in `aes_acc.h` there seems to be `AES_ACC_IRQ_DONE_STAT`. Can't we use it instead of polling?

dts/arm/sifli/sf32lb52x.dtsi
```diff
@@ -147,6 +147,14 @@
 			status = "disabled";
 		};
 
+		crypto: crypto@5000d000 {
+			compatible = "sifli,sf32lb-crypto";
+			reg = <0x5000d000 0x1000>;
+			clocks = <&rcc_clk SF32LB52X_CLOCK_AES>;
+			interrupts = <67 0>;
```
valeriosetti, line 180: Interrupt seem to be required according to the `yaml` file, but it's never used in the driver code. However I think it would be a nice improvement instead of polling.

valeriosetti (changes requested):

drivers/crypto/crypto_sf32lb_priv.h
HalfSweet: On the SF32LB platform, AES encryption and HASH operations can run concurrently, so they must maintain separate sessions

drivers/crypto/crypto_sf32lb.c
HalfSweet: done

drivers/crypto/crypto_sf32lb.c
HalfSweet: ASYNC mode support has now been added. If ASYNC is enabled, there will be no need to wait. Conversely, in non-ASYNC mode, I believe there is no need to use interrupts; maintaining polling is sufficient

valeriosetti (dismissed):
PR looks good to me, but there is a CI failure and a conflict. Once you fix that I will re-approve.

HalfSweet:
All check have been passed

valeriosetti (dismissed):

ck-telecom (dismissed):

ceolin (dismissed):

HalfSweet:
There appeared to be some conflicts, so I performed a rebase. Re-approval is required

valeriosetti (dismissed):

dts/bindings/crypto/sifli,sf32lb-crypto.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2025 SiFli Technologies(Nanjing) Co., Ltd
+# SPDX-License-Identifier: Apache-2.0
+
+description: SiFli SF32LB Series Crypto Accelerator (AES/HASH)
+
+compatible: "sifli,sf32lb-crypto"
+
+include: base.yaml
```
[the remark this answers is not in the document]
HalfSweet: The AES module does not have an independent clock

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,998 @@
...
+		LOG_ERR("HASH operation timeout");
+		/* Disable IRQ mask */
+		sys_write32(sys_read32(base + AES_SETTING_OFFSET) &
+				    ~(AES_ACC_SETTING_HASH_DONE_MASK |
+				      AES_ACC_SETTING_HASH_BUS_ERR_MASK |
+				      AES_ACC_SETTING_HASH_PAD_ERR_MASK),
+			    base + AES_SETTING_OFFSET);
+		k_sem_give(&data->device_sem);
+		return -ETIMEDOUT;
+	}
```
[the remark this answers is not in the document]
HalfSweet: done

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,998 @@
...
+
+	/* Enable IRQ masks */
+	sys_write32(AES_ACC_SETTING_DONE_IRQ_MASK | AES_ACC_SETTING_BUS_ERR_IRQ_MASK |
+			    AES_ACC_SETTING_SETUP_ERR_IRQ_MASK,
+		    base + AES_SETTING_OFFSET);
+
+	/* Set DMA addresses and size */
+	sys_write32((uint32_t)in_data, base + AES_DMA_IN_OFFSET);
+	sys_write32((uint32_t)out_data, base + AES_DMA_OUT_OFFSET);
+	sys_write32((size + 15) >> 4, base + AES_DMA_DATA_OFFSET);
```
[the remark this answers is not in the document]
HalfSweet: done

HalfSweet:
ping @valeriosetti @kartben

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,1144 @@
...
+}
+
+static inline void crypto_sifli_cache_flush_invd(void *addr, size_t len)
+{
+	if (addr == NULL || len == 0U) {
+		return;
+	}
+
+	sys_cache_data_flush_and_invd_range(addr, len);
+}
```
ceolin: Are you adding those wrappers because of memory check ? Those checks are done only when user space is enabled. Is that the reason ?

drivers/crypto/Kconfig.sf32lb
```diff
@@ -0,0 +1,42 @@
...
+
+config CRYPTO_SIFLI_AES
+	bool "SiFli AES cipher support"
+	default y
+	help
+	  Enable AES-128/192/256 ECB/CBC/CTR cipher support.
+
+config CRYPTO_SIFLI_ASYNC
+	bool "SiFli crypto async operations support"
+	default n
```
ceolin: Not needed

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,1144 @@
...
+			status = -EIO;
+		}
+
+		/* Call hash callback or signal sync semaphore */
+		if (data->hash_pkt != NULL) {
+			struct hash_pkt *pkt = data->hash_pkt;
+			uint8_t algo = data->hash_algo;
+
+			data->hash_pkt = NULL;
+			data->hash_status = status;
```
ceolin: Is it possible to this handler be invoked due both interruption sources (AES / HASH) ? If so, this `status` variable may have have the contents from the other operation.

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,1144 @@
...
+	sys_cache_data_flush_and_invd_range(addr, len);
+}
+
+#if defined(CONFIG_CRYPTO_SIFLI_ASYNC)
+
+static void crypto_sifli_isr(const struct device *dev)
+{
+	const struct crypto_sifli_config *config = dev->config;
+	struct crypto_sifli_data *data = dev->data;
+	uint32_t base = config->base;
```
ceolin: To keep consistent, why not using `uintpr_t` ?

drivers/crypto/crypto_sf32lb.c
```diff
@@ -0,0 +1,1144 @@
...
+	return 0;
+}
+
+static int crypto_sifli_hash_free_session(const struct device *dev, struct hash_ctx *ctx)
+{
+	struct crypto_sifli_hash_session *session = CRYPTO_SIFLI_HASH_SESSN(ctx);
+
+	if (session != NULL) {
+		session->in_use = false;
+	}
```
ceolin, line 1020: You should erase key material here.

ceolin (changes requested):
A few minor comments. The `status` variable and key material are possible problems, the others are just nitpicking.

Please take a look and check if they make sense. Thanks for the driver, it is looking great.

drivers/crypto/Kconfig.sf32lb
HalfSweet: done

drivers/crypto/crypto_sf32lb.c
HalfSweet: done. It has now been divided into different states

drivers/crypto/crypto_sf32lb.c
HalfSweet: done

drivers/crypto/crypto_sf32lb.c
HalfSweet: Yes, but it seems that such encapsulation is no longer meaningful now that it has been removed

ceolin (approved):

valeriosetti (approved):

PR #100584, opened 2025-12-05
drivers: mcux_os_timer: Add Kconfig option for MIN_DELAY configuration

yeaissa:
Introduce CONFIG_MCUX_OS_TIMER_MIN_DELAY to allow the configuration of the minimum delay cycles.

### DIFF
--- a/drivers/timer/Kconfig.mcux_os
+++ b/drivers/timer/Kconfig.mcux_os
@@ -21,4 +21,14 @@ config MCUX_OS_TIMER_PM_POWERED_OFF
 	  OS Timer is turned off in certain low power modes. When this option is
 	  picked, OS Timer will take steps to store state and reinitialize on wakeups.
 
+config MCUX_OS_TIMER_MIN_DELAY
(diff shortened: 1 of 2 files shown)


drivers/timer/Kconfig.mcux_os
```diff
@@ -21,4 +21,14 @@ config MCUX_OS_TIMER_PM_POWERED_OFF
 	  OS Timer is turned off in certain low power modes. When this option is
 	  picked, OS Timer will take steps to store state and reinitialize on wakeups.
 
+config MCUX_OS_TIMER_MIN_DELAY
+	int "Minimum delay cycles for OS timer match value"
+	default 1000
```
yvesll, line 26: Is it possible to add a range?

yeaissa: I don't think a meaningful range can be defined here. The value is dependent on the timer's clock frequency, which varies significantly across platforms and configurations.

Holt-Sun (approved):

mmahadevan108 (approved):

PR #100586, opened 2025-12-05
drivers: modem: Include empty string arguments in argument parsing

joerchan:
When parsing AT command responses, empty string arguments were being ignored if they occur it the end of the response.
This is not only iconsistent behavior, but also causes issues when parsing responses where we need to know the number of arguments. For example, the +CEREG response and unsolicited command response which we need to know if it is response or unsolicited command response to correctly parse the it, as the first argument will be different. 
Example:
```
Response:     +CEREG: 5,4,\"0A0B\",\"FFFFFFFF\",9,0,0,,
Notification: +CEREG: 4,\"0A0B\",\"FFFFFFFF\",9,0,0,,
```
Here we want to know that response has 8 arguments, not 7, or we will
confuse it with the notification.


rerickson1 (dismissed):

bjarki-andreasen (dismissed):

joerchan:
Added test cases, for existing functionality as well.
The testcase `test_recv_response_parse_args_empty_arg_end` is testing the change of this PR.

rerickson1 (approved):

bjarki-andreasen (approved):

joerchan:
Can someone rerun the checks now that the DNM label is removed?

PR #98836, opened 2025-11-04
drivers: flash: esp32: flash access supervisor

marekmatej:
This PR brings changes to ESP32 flash driver to suport flash access requests from another CPU cores.

PR #100588, opened 2025-12-05
Add Waveshare ESP32-S3-GEEK board

xingrz:
0a096381c116280012d51e49192a4176

This PR adds support for the Waveshare ESP32-S3-GEEK USB dongle.

The board features an ESP32-S3R2 with 2 MB PSRAM, 16 MB on-board flash, a 1.14" LCD, and a microSD slot.


boards/waveshare/esp32s3_geek/esp32s3_geek_esp32s3_procpu.dts
```diff
@@ -0,0 +1,151 @@
...
+	#size-cells = <0>;
+	pinctrl-0 = <&spim2_default>;
+	pinctrl-names = "default";
+};
+
+&usb_serial {
+	status = "okay";
+};
+
+zephyr_udc0: &usb_otg {};
```
josuah, line 118: Just curious: No need for `status = "okay";` on this board?
It will not be initialized unless Kconfig also enables USB.
It seems `status = "okay";` on i.e. `boards/espressif/esp32s3_devkitc/esp32s3_devkitc_procpu.dts`.

xingrz: Actually, only one of `usb_serial` or `usb_otg` can be enabled at a time because they share the same USB port. On the DevKitC, the USB-to-serial function is provided by the on-board CP2102, so the SoC's internal `usb_serial` is not used, and the USB-C receptacle is intended for `usb_otg`.

On this dongle, however, the USB-A plug is the only way to program or access the console unless an external USB-to-serial adapter is attached. For this reason, the USB port is by default used for `usb_serial`.

josuah: Very insightful, thank you! I am leaving the thread open for other reviewers.

xingrz:
@josuah Yes, it's broken since #98836. It breaks `esp32s3_devkitc/esp32s3/appcpu` too.

boards/waveshare/esp32s3_geek/esp32s3_geek_esp32s3_appcpu.yaml
```diff
@@ -24,4 +24,4 @@ testing:
     - cmsis_rtos
     - jwt
     - zdsp
```
sylvioalves, line 26: Add here `psa.secure_storage` so it ignores such test (which is not supposed to be done for `appcpu`).

nordicjm (dismissed):

samples/subsys/usb/cdc_acm_bridge/boards/esp32s3_geek_esp32s3_procpu.overlay
```diff
@@ -0,0 +1,24 @@
+/*
```
jfischer-no: This file is similar to samples/subsys/usb/cdc_acm_bridge/app.overlay, which is allowing this sample to be used with many boards. I do not think there should be another board specific overlay file, as that would finally create an overwhelming number of overlays with almost identical content.

jfischer-no (dismissed):

samples/subsys/usb/cdc_acm_bridge/boards/esp32s3_geek_esp32s3_procpu.overlay
xingrz: Do you mean adding an `arduino_serial` node label pointing to `uart0` so that the generic `app.overlay` can be reused? That would indeed simplify things.

One board-specific requirement remains: on this dongle the hardware USB-to-serial (`usb_serial`) and the USB device controller (`zephyr_udc0`) share the same USB port. Because of that, we must explicitly disable `usb_serial` and enable `zephyr_udc0` in the board overlay; otherwise the UDC driver cannot be initialized.

tmon-nordic: Where is this "hardware USB-to-serial"? Is it internal part of ESP32-S3? I can't see anything other on the schematic.

jfischer-no: > Do you mean adding an `arduino_serial` node label pointing to `uart0` so that the generic `app.overlay` can be reused? That would indeed simplify things.

No.

> One board-specific requirement remains: on this dongle the hardware USB-to-serial (`usb_serial`) and the USB device controller (`zephyr_udc0`) share the same USB port. Because of that, we must explicitly disable `usb_serial` and enable `zephyr_udc0` in the board overlay; otherwise the UDC driver cannot be initialized.

That is a problem with your board and not the sample. Just because something is possible does not mean it should be supported. I want the samples to be as simple and maintainable as possible. If your board does not configure and use the USB device controller by default, then that is a problem with your board, not the samples. If your board does not have a node label that could be used with a sample, that is a problem with your board.

xingrz: > Where is this "hardware USB-to-serial"? Is it internal part of ESP32-S3? I can't see anything other on the schematic.

@tmon-nordic Yes, it's called "USB Serial/JTAG Controller"

https://documentation.espressif.com/esp32-s3_technical_reference_manual_en.pdf 33.3.1 USB Serial/JTAG Host Connection (page 1244):

> The ESP32-S3 has an internal PHY, which is shared between the USB-OTG and the USB Serial/JTAG hardware.

xingrz: @jfischer-no OK, I'll drop this overlay.

marekmatej: > This file is similar to samples/subsys/usb/cdc_acm_bridge/app.overlay, which is allowing this sample to be used with many boards. I do not think there should be another board specific overlay file, as that would finally create an overwhelming number of overlays with almost identical content.

You are suggesting that every board should define the `arduino_uart`? That doesn't seem right. If you are afraid of many *almost-the-same* overlays, then putting it under the `socs` instead of `boards` will cover every ESP board willing to use the subsystem.

xingrz:
Rebased to include a CI fix from #100625

samples/subsys/usb/cdc_acm_bridge/boards/esp32s3_geek_esp32s3_procpu.overlay
tmon-nordic: > 33.3.1 USB Serial/JTAG Host Connection

Thanks, this clarifies the problem a lot. Let me summarize, please correct if I get anything wrong.

So basically esp32-s3 has two USB controllers:
1. Generic purpose DWC2 controller, that can be used by Zephyr USB stack.
2. USB Serial/JTAG Controller, which cannot be used by Zephyr USB stack. This controller implements USB device stack in hardware.

esp32-s3 has one internal USB PHY and a possibility to use external USB PHY. Both controllers can be configured to use either PHY, but obviously the selection is mutually exclusive. The board in question does not have external USB PHY available, and therefore it is up to firmware developer to choose which USB controller to use with internal USB PHY.

When it comes to Zephyr, the DWC2 controller is available as `usb_otg` while the USB Serial/JTAG Controller is available as `usb_serial`. The mutually exclusive use of USB PHY is currently solved by making sure that only one of these nodes has `status = "okay";`.

A side note: is the multiplexer for external PHY somehow modelled in devicetree and/or otherwise configurable using any standard Zephyr mechanism?

jfischer-no: > You are suggesting that every board should define the `arduino_uart`?

https://github.com/zephyrproject-rtos/zephyr/pull/100588#discussion_r2602246609

> That doesn't seem right. If you are afraid of many _almost-the-same_ overlays, then putting it under the `socs` instead of `boards` will cover every ESP board willing to use the subsystem.

I am not afraid. I am very confident that there should not be any of this kind. And certainly not in the SoC directory. There are other ways to solve that. Apart from it, USB samples still do not have to support any possible hardware configurations by default.

sylvioalves: > esp32-s3 has one internal USB PHY and a possibility to use external USB PHY. Both controllers can be configured to use either PHY, but obviously the selection is mutually exclusive. The board in question does not have external USB PHY available, and therefore it is up to firmware developer to choose which USB controller to use with internal USB PHY.

> When it comes to Zephyr, the DWC2 controller is available as usb_otg while the USB Serial/JTAG Controller is available as usb_serial. The mutually exclusive use of USB PHY is currently solved by making sure that only one of these nodes has status = "okay";.

@tmon-nordic, you are correct. 
To add up, some Espressif's devkit like `esp32c6_devkitc` comes with 2 USB connectors:
1. Used with an onboard usb-to-uart bridge `CP2102N`. That provides flash/console over uart.
2. The one we are talking here: `USB Serial/JTAG Controller`, which we can use to flash/console over usb-serial, debug over USB/JTAG or as USB/DWC2.

> A side note: is the multiplexer for external PHY somehow modelled in devicetree and/or otherwise configurable using any standard Zephyr mechanism?

We haven't yet added such devicetree configurable entry to select between internal/external PHY.

sylvioalves (approved):

samples/subsys/usb/cdc_acm_bridge/boards/esp32s3_geek_esp32s3_procpu.overlay
marekmatej: > I am not afraid. I am very confident that there should not be any of this kind. And certainly not in the SoC directory. There are other ways to solve that. Apart from it, USB samples still do not have to support any possible hardware configurations by default.

@jfischer-no  Are we talking about the same thing? I mean `socs` as in here https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/net/wifi/shell/socs
It is standard Zephyr way to provide overlay on SoCs level. Can you point me to a rule that prohibits such usage?

jfischer-no: You have got something mixed up. First, please show me where a maintainer is forced to accept these overlays to be placed in a sample, and also in sample that is otherwise is made to run on majority of the boards out of the box?

Anyway, thanks for another example, https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/net/wifi/shell/socs, with a lot of trivial overlays like https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/net/wifi/shell/socs/esp32s3_procpu.overlay.

marekmatej: @jfischer-no Please note that the topic of this discussion is not the obsolete overlays you mentioned (those should indeed be fixed). The focus here is on the overlays introduced in this PR, which resolve the conflict between UART nodes.

Given your strong position on this, I would like to clarify an apparent inconsistency. You are comfortable with the use of overlays in the following samples:
- https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/mass
- https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/shell
- https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/uac2_explicit_feedback
- https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/uac2_implicit_feedback

However, in this sample:
- https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/usb/cdc_acm_bridge/

that on `esp32s3_devkitc` fails on `parse error: undefined node label 'arduino_serial'`,
You oppose applying the trivial overlay such as:
```
zephyr_udc0: &usb_otg {
    status = "okay";
};
```

Requiring board vendors to add an `arduino_serial` label solely to build this single sample doesn't seem to be the correct nor practical approach to me.

Therefore, I would appreciate clear guidance on the rules for applying the overlays in the Zephyr `samples`. My current understanding is that any sample code is for testing and educational purposes, enabling Zephyr users to adapt their specific platforms to provide desired functionality.

nordicjm (approved):

wmrsouza (approved):

marekmatej (approved):

samples/subsys/usb/cdc_acm_bridge/boards/esp32s3_geek_esp32s3_procpu.overlay
jfischer-no: > @jfischer-no Please note that the topic of this discussion is not the obsolete overlays you mentioned (those should indeed be fixed). The focus here is on the overlays introduced in this PR, which resolve the conflict between UART nodes.

What? It is you who brought these overlays into the discussion. Please go fix it.

> Given your strong position on this, I would like to clarify an apparent inconsistency. You are comfortable with the use of overlays in the following samples:
> 
>     * https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/mass
> 
>     * https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/shell
> 
>     * https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/uac2_explicit_feedback
> 
>     * https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/usb/uac2_implicit_feedback

Unfortunately, they are there either because of historical reasons or of sample limitation. Their mere existence is no justification for adding new overlays.
 
> However, in this sample:
> 
>     * https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/usb/cdc_acm_bridge/
> 
> 
> that on `esp32s3_devkitc` fails on `parse error: undefined node label 'arduino_serial'`, You oppose applying the trivial overlay such as:
> 
> ```
> zephyr_udc0: &usb_otg {
>     status = "okay";
> };
> ```

No idea why are you mixing arduino_serial and zephyr_udc0 node labels. Apart from the fact that your board does not even enable USB device controller node, a **sample does not have to run on any board**. And I already pointed out in other reviews that you could add a snippet to enable USB device controllers, and provide documentation to the users of these boards with a single connector, on how to apply a snippet to build commands.

> Requiring board vendors to add an `arduino_serial` label solely to build this single sample doesn't seem to be the correct nor practical approach to me.

That is something you made up. I did not requested that you add a node label.
 
> Therefore, I would appreciate clear guidance on the rules for applying the overlays in the Zephyr `samples`.

No trivial overlays in USB samples.

PR #100558, opened 2025-12-05
tests: drivers: counter: Improve alarm capability detection

ZhaoxiangJin:
- Add a self-contained alarm_capable() probe that starts the device, sets and cancels a 1-tick alarm, and stops the counter again. This treats -ENOTSUP as “not capable” and avoids leaving pending alarms.
- Switch single/multiple alarm capability checks to reuse alarm_capable(), preventing unsupported drivers from running alarm cases and reducing false failures.
- Ensure capability probing no longer leaves the driver running or with active alarms, preventing teardown/set_top_value failures in later tests.

fix https://github.com/zephyrproject-rtos/zephyr/issues/100529

Issue #100529, opened 2025-12-04
tests: drivers: frdm_mcxn947: drivers.counter.basic_api fails


majunkier:
### Describe the bug

Test drivers.counter.basic_api fails on frdm_mcxn947 after [commit ](https://github.com/zephyrproject-rtos/zephyr/commit/a3663c8c7fd)

### Regression

- [x] This is a regression.

### Steps to reproduce

twister -p frdm_mcxn947/mcxn947/cpu0 --device-testing --hardware-map map.yaml -s drivers.counter.basic_api

### Relevant log output

```shell
*** Booting Zephyr OS build v4.3.0-1638-gd14a547d4259 (delayed boot 500ms) ***
Running TESTSUITE counter_basic
===================================================================
START - test_all_channels
Testing ctimer@c000
Skipped for mrt0_channel@0
W: MRT channel resets upon stopping
Testing lptmr@4a000
Testing lptmr@4b000
 PASS - test_all_channels in 0.501 seconds
===================================================================
START - test_cancelled_alarm_does_not_expire
Skipped for ctimer@c000
Skipped for mrt0_channel@0
W: MRT channel resets upon stopping
Skipped for lptmr@4a000
Skipped for lptmr@4b000
 SKIP - test_cancelled_alarm_does_not_expire in 0.412 seconds
===================================================================
START - test_late_alarm
Skipped for ctimer@c000
Skipped for mrt0_channel@0
W: MRT channel resets upon stopping
Skipped for lptmr@4a000
Skipped for lptmr@4b000
 SKIP - test_late_alarm in 0.412 seconds
===================================================================
START - test_late_alarm_error
Skipped for ctimer@c000
Skipped for mrt0_channel@0
W: MRT channel resets upon stopping
Skipped for lptmr@4a000
Skipped for lptmr@4b000
 SKIP - test_late_alarm_error in 0.412 seconds
===================================================================
START - test_multiple_alarms
Testing ctimer@c000
Skipped for mrt0_channel@0
W: MRT channel resets upon stopping
Skipped for lptmr@4a000
Skipped for lptmr@4b000
 PASS - test_multiple_alarms in 0.466 seconds
===================================================================
START - test_set_top_value_with_alarm
Testing ctimer@c000
Testing mrt0_channel@0
I: "Started" MRT@0x50013000 channel 0 with default value 16777215
W: MRT@0x50013000 channel 0 received requested top value 3000000 which is smaller than current count 16026450
[...]

PR #100589, opened 2025-12-05
drivers: counter: fix mcux lptmr unit test issue.

Holt-Sun:
Fix the MCUX LPTMR counter driver alarm behavior so that, after an alarm
fires, the match value is reset to the maximum top value. Without this,
restarting the LPTMR may hit the previous match value almost
immediately, causing the counter_basic_api test
test_valid_function_without_alarm to fail intermittently.

Extend the counter_basic_api test suite with a dedicated test
configuration that enables CONFIG_COUNTER_MCUX_LPTMR_ALARM and runs
the same tests on any board that instantiates an nxp,lptmr device.
This verifies the alarm-only LPTMR configuration against the common
counter API.

Depends-On: https://github.com/zephyrproject-rtos/zephyr/pull/100558

### DIFF
--- a/drivers/counter/counter_mcux_lptmr.c
+++ b/drivers/counter/counter_mcux_lptmr.c
@@ -247,6 +247,7 @@ static void mcux_lptmr_isr(const struct device *dev)
 		uint32_t current_count = LPTMR_GetCurrentTimerCount(config->base);
 
 		LPTMR_StopTimer(config->base);
+		LPTMR_SetTimerPeriod(config->base, config->info.max_top_value);
 
 		callback(dev, 0, current_count, user_data);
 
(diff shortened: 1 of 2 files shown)


ZhaoxiangJin:
are you fixing https://github.com/zephyrproject-rtos/zephyr/issues/100529? I just tried this patch, but alarm-related testcases still failed.

Holt-Sun:
> are you fixing #100529? I just tried this patch, but alarm-related testcases still failed.

You need to enable the COUNTER_MCUX_LPTMR_ALARM.

I am not directly aiming at https://github.com/zephyrproject-rtos/zephyr/issues/100529, I just realize the lptmr unittest is not actually run with latest added COUNTER_MCUX_LPTMR_ALARM. I am waiting your https://github.com/zephyrproject-rtos/zephyr/pull/100558 merged, then reenable the unit test for lptmr both alarm+top.

zejiang0jason (approved):

ZhaoxiangJin (approved):

FelixWang47831 (approved):

PR #100595, opened 2025-12-05
net: coap: add support for CoAP over reliable transports

mpenate-ellenbytech:
Add implementation of RFC 8323 (CoAP over TCP). This includes:
- TCP-specific packet parsing and creation functions
- BERT (Block-wise Extension for Reliable Transport) support
- CSM signaling code (7.01) and options
- New Kconfig options for reliable transport configuration


subsys/net/lib/coap/Kconfig
```diff
@@ -101,6 +101,12 @@ config COAP_KEEP_USER_DATA
 	help
 	  This option enables keeping application-specific user data
 
+config COAP_RELIABLE
```
rlubos: Should we call the config `COAP_OVER_RELIABLE_TRANSPORT`? `COAP_RELIABLE` IMHO suggests that the core CoAP is not reliable, which isn't really the case, it just ensures reliability at different level.

subsys/net/lib/coap/Kconfig
```diff
@@ -170,6 +176,39 @@ config COAP_CLIENT_TRUNCATE_MSGS
...
+	default 247000
+	range 60000 600000
+	help
+	  Maximum lifetime of a TCP CoAP exchange from initial request to
+	  final response. Default is 247 seconds as per RFC 8323.
+	  Used to determine when request contexts can be safely released.
+
+config COAP_CLIENT_TCP_BLOCKWISE_REUSE_TOKEN
+	bool "Reuse the request token when requesting further blocks"
+	default n
```
rlubos: No need for `default n`, it's implicit.

include/zephyr/net/coap.h
```diff
@@ -1276,6 +1290,34 @@ struct coap_transmission_parameters coap_get_transmission_parameters(void);
  */
 void coap_set_transmission_parameters(const struct coap_transmission_parameters *params);
 
+#if defined(CONFIG_COAP_RELIABLE)
```
rlubos: It's not needed to hide function declarations behind `#ifdef` (actually I recall it was documented somewhere we should avoid that, just don't have it handy).

include/zephyr/net/coap.h
```diff
@@ -1276,6 +1290,34 @@ struct coap_transmission_parameters coap_get_transmission_parameters(void);
...
+
+uint8_t coap_header_get_token_tcp(const struct coap_packet *cpkt, uint8_t *token);
```
rlubos: I wonder if we could have a common prefix `coap_tcp_` for all the functions instead of having the `tcp` part in the end or in the middle, I think it'd be more readable that way, so `coap_tcp_header_get_token()` etc.

rlubos: Plus we should aim to document all functions added to a public header.

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2158,483 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+
+#if defined(CONFIG_COAP_RELIABLE)
+
+int coap_packet_init_tcp(struct coap_packet *cpkt, uint8_t *data, uint16_t max_len,
+			 uint8_t token_len, const uint8_t *token, uint8_t code)
+{
+	uint8_t hdr;
+	bool res;
+
+	if (!cpkt || !data || !max_len) {
```
rlubos: For new code we should aim to follow Misra rules, so need explicit NULL/0 checks here and in similar cases (`if (cpkt == NULL || data == NULL || max_len == 0)`).

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2158,483 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+
+int coap_packet_tcp_update_len(struct coap_packet *cpkt)
+{
+	uint8_t hdr;
+	uint8_t packet_len = (cpkt->data[0] >> 4) & 0xF;
+	uint8_t token_len = (cpkt->data[0]) & 0xF;
+
+	uint64_t options_and_payload_len = cpkt->offset - 1 - 1 - token_len;
+
+	if (options_and_payload_len <= 12) {
```
rlubos: Please avoid using magic numbers - either define symbols for the numbers used in this function (12, 269 etc.), or reuse existing ones if fit.

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2158,483 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+uint8_t coap_header_get_code_tcp(const struct coap_packet *cpkt)
+{
+	if (!cpkt || !cpkt->data) {
+		return 0;
+	}
+
+	uint8_t tkl = cpkt->data[0] & 0x0f;
+	uint8_t code = cpkt->data[cpkt->hdr_len - tkl - 1];
+
+	switch (code) {
```
rlubos: We could possibly share this code matching between standard CoAP/TCP CoAP?

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+				  int error_code);
+
+/**
+ * @brief Send a CoAP message over TCP
+ */
+static int send_request_tcp(int sock, const void *buf, size_t len, int flags)
+{
+	int ret;
+
+	ret = zsock_send(sock, buf, len, flags);
```
rlubos: So with STREAM sockets you should always check the number of bytes sent, `send()` can actually send less bytes than requested in a single call. This is usually handled by calling send in a loop, for example in MQTT:
https://github.com/zephyrproject-rtos/zephyr/blob/3b23c76a246973fe7b1570447398f3bb1ef434e1/subsys/net/lib/mqtt/mqtt_transport_socket_tcp.c#L84-L101

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+static int recv_response_tcp(struct coap_client_tcp *client, struct coap_packet *response,
+			     bool *truncated)
+{
+	int total_len;
+	int available_len;
+	int ret;
+	int flags = ZSOCK_MSG_DONTWAIT;
+
+	if (IS_ENABLED(CONFIG_COAP_CLIENT_TRUNCATE_MSGS)) {
+		flags |= ZSOCK_MSG_TRUNC;
```
rlubos: `MSG_TRUNC` only works with datagram sockets, no point setting this for stream socket.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+		LOG_ERR("Packet too large: %zd > %zu", packet_length + bytes_remaining, max_len);
+		return -EMSGSIZE;
+	}
+
+	while (bytes_remaining > 0) {
+		err = zsock_recv(sock, &byte_buf[packet_length], bytes_remaining, flags);
+		if (err < 0) {
+			if (errno == EAGAIN || errno == EWOULDBLOCK) {
+				continue;
+			}
```
rlubos: Please don't do that, you make the `zsock_recv()` call non-blocking with `ZSOCK_MSG_DONTWAIT`, repeating the loop in case of `EAGAIN` is a recipe for a busy-loop, potentially making the whole system unusable. 

Either use blocking `zsock_recv()` to prevent busy-looping, or we'd need some more sophisticated logic with some sort of state machine to be able to resume receiving in case of `EAGAIN` at the point where it stopped, but only after `poll()` reports new data.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+		}
+	} else {
+		if (client->socklen != 0 && has_ongoing_request(client)) {
+			LOG_WRN("Can't clear address, request ongoing");
+			ret = -EALREADY;
+			goto release;
+		}
+		memset(&client->address, 0, sizeof(client->address));
+		client->socklen = 0;
+	}
```
rlubos: I don't think there's a point of having a address parameter in the CoAP over stream variant? Doesn't seem to be used, and you can't actually re-connect a stream socket to a different address.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+			continue;
+		}
+
+		if (fds[i].revents & ZSOCK_POLLOUT) {
+			coap_client_tcp_timeout_handler(client);
+		}
+
+		if (fds[i].revents & ZSOCK_POLLIN) {
+			struct coap_packet response;
+			bool response_truncated = false;
```
rlubos: There should be no concept of "truncated response" in case of STREAM transport, all logic related to this should be removed.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
...
+	for (int i = 0; i < nfds; i++) {
+		struct coap_client_tcp *client = get_tcp_client(fds[i].fd);
+
+		if (!client) {
+			continue;
+		}
+
+		if (fds[i].revents & ZSOCK_POLLOUT) {
+			coap_client_tcp_timeout_handler(client);
+		}
```
rlubos: Monitoring `ZSOCK_POLLOUT` made sense in case of regular CoAP as we needed to handle retransmissions, but here we just report an error in case of timeout, so that should be handled separately from `ZSOCK_POLLOUT`. Probably no need to handle `ZSOCK_POLLOUT` at all for this case.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1075 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
```
rlubos, line 2: You should update copyright

include/zephyr/net/coap_client_tcp.h
```diff
@@ -0,0 +1,234 @@
...
+#endif
+
+/** Maximum size of a CoAP TCP message */
+#define MAX_COAP_TCP_MSG_LEN (CONFIG_COAP_CLIENT_MESSAGE_HEADER_SIZE + \
+			      CONFIG_COAP_CLIENT_MESSAGE_SIZE)
+
+/**
+ * @brief Representation for CoAP TCP client response data.
+ */
+struct coap_client_tcp_response_data {
```
rlubos: Having two separate implementations for UDP/TCP CoAP client is IMHO fine given the differences, but I think we should aim to share common concepts like `struct coap_client_response_data`, `struct coap_client_request` or callback formats between the two whenever possible. Looking at this header file, those are identical to the original ones. Maybe just extend the orginal `coap_client` header with TCP-specific stuff (new APIs etc.).

rlubos (commented):
An initial pass on the PR.

For the final version, we should also aim to have some sample covering this functionality (or ideally some tests if possible), and documentation.

subsys/net/lib/coap/Kconfig
mpenate-ellenbytech: I agree CoAP Reliable isn't great but this would ideally help support coap over websockets which is still over TCP. However it isn't quite the same as what I would expect if I heard "CoAP TCP". In-line with other comments I'll switch to `COAP_TCP` for now and try to have all functions and declarations have `coap_tcp`

rlubos: I should've express myself more clearly - I think we should call the config option enabling this feature `COAP_OVER_RELIABLE_TRANSPORT`, not just `COAP_RELIABLE` - the latter implies that this makes the CoAP itself reliable, which is already the case with CoAP over UDP.

I've also considered `COAP_OVER_TCP`, but as you say it could be confusing if we added support for TLS or websocket in the future.

mpenate-ellenbytech: I can agree with `COAP_OVER_RELIABLE_TRANSPORT`, switching to this.

Also FWIW I am testing over both basic TCP and secure TLS/TCP.

subsys/net/lib/coap/Kconfig
mpenate-ellenbytech: Done, reflected in most recent force push

include/zephyr/net/coap.h
mpenate-ellenbytech: Done, reflected in most recent force push

include/zephyr/net/coap.h
mpenate-ellenbytech: Added a starting point for documentation, switched to `coap_tcp_`... for all functions declared in `coap.h`

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: Done, reflected in most recent force push. Also re-wrote the update len function to handle shrinking the packet (I think, this should DEFINITELY have a test case)

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: Done created `static uint8_t __coap_header_code_sanitize(uint8_t code)`, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Done, reflected in most recent force push

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Added `recv_offset` to the tcp client as it is linked to the file descriptor currently. This is now non-blocking and returns when the entire packet has yet to be received.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1068 @@
...
+		return -EMSGSIZE;
+	}
+
+	/* Step 5: Check if we have complete packet */
+	if (offset < total_expected) {
+		return -EAGAIN;
+	}
+
+	/* Step 6: Complete packet received, reset offset for next packet */
+	client->recv_offset = 0;
```
rlubos: Could it be the case that we receive two messages from the server one after another?
Note that now `zsock_recv()` can read up to to the buffer size number of bytes, so theoretically we could've read more than a single message in a single `zsock_recv()` call. If we reset the offset here, we might lose the other message.
If `total_expected` is the size of the currently processed message, we should check if `offset > total_expected` and `memmove()` the remaining bytes to the beginning of the buffer. And instead of clearing the offset, decrease it by `total_expected` value.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1078 @@
...
+			k_mutex_unlock(&client->lock);
+
+			/* Consume processed packet and preserve any remaining bytes */
+			if (client->recv_offset > pkt_len) {
+				memmove(client->recv_buf, client->recv_buf + pkt_len,
+					client->recv_offset - pkt_len);
+				client->recv_offset -= pkt_len;
+			} else {
+				client->recv_offset = 0;
+			}
```
mpenate-ellenbytech: Should this oversized recv handling happen within the mutex critical section? The purpose of that mutex is vague to me besides "generally protect the client during multi-threaded activity"

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: See comment above, removed setting `recv_offset` to 0 here and instead after handling the response in the polling function I adjust the value there.

mpenate-ellenbytech:
I added rudimentary handling of the CSM max message size option but the RFC specifies some, lets say complicated implementation requirements that don't fit in with the original idea of CoAP client so I pushed the basic handling and come here to continue the discussion.

[Section 5.3.2](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3.2) on the enabling of block-wise transfers via CSMs:
> If a Max-Message-Size Option is indicated with a value that is
   greater than 1152 (in the same CSM or a different CSM), the
   Block-Wise-Transfer Option also indicates support for BERT (see
   [Section 6](https://datatracker.ietf.org/doc/html/rfc8323#section-6))
And earlier in [Section 5.3](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3):
> Base values are listed below for CSM options.  These are the values
   for the capability and settings before any CSMs send a modified
   value.

The base value for the [Block-Wise-Transfer Capability Option](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3.2) is unspecified meaning that generic block-wise transfer and BERT block-wise transfer should technically be runtime-disabled until this CSM is negotiated with the server correct?

If we were to do this, then the "context" of the socket is now more entangled with CoAP client than it was before. When using a new socket fd with the coap client, the developer needs to ask themselves "Is this socket context already negotiated with basic CSMs?" and inform the client of these previous negotiations if they exist which IMO isn't the best.

@rlubos what do you think about integrating more of the socket connection context into the management handled by CoAP client? Personally I could see a world where a client is initialized with a basic int-returning callback that should return a `zsock_connect`-ed socket FD. Then if anything ever goes wrong with the socket, the client can close it and request a new one be opened using the callback? Then upon using the callback it can automatically handle the CSMs and the book-keeping needed for it. This would restrict a client to one connected context at a time but that restriction already exists in the form of a single socket fd in the UDP client.

subsys/net/lib/coap/coap_client_tcp.c
rlubos: I don't think there's a need, the RX buffer will only be updated from the CoAP TCP client thread so there's likely no need for extra mutex protection. That client mutex to my understanding protects the internal request contexts, where those could be updated from different threads (like for example when cancelling request).

rlubos:
> The base value for the [Block-Wise-Transfer Capability Option](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3.2) is unspecified meaning that generic block-wise transfer and BERT block-wise transfer should technically be runtime-disabled until this CSM is negotiated with the server correct?

That's my understanding as well - lack of Block-wise-transfer option in CSM indicates no block transfer support by the sender, so block transfers should not be used over such connection. Probably good-enough approach for initial implementation.

> @rlubos what do you think about integrating more of the socket connection context into the management handled by CoAP client? Personally I could see a world where a client is initialized with a basic int-returning callback that should return a zsock_connect-ed socket FD. Then if anything ever goes wrong with the socket, the client can close it and request a new one be opened using the callback? 

I agree, client context in CoAP over TCP is practically bound to the underlying TCP/TLS connection, so it makes sense to integrate the client with the socket more closely. I'd imagine we'd have two additional session-related functions for CoAP over TCP, like:
* `coap_client_tcp_connect()`
* `coap_client_tcp_close()`

`coap_client_tcp_connect()` would make a connection at the socket level, and initialize CoAP session (i.e. send CSM). `coap_client_tcp_close()` would close the session (I see in the RFC that there are also various signals to notify this, instead of just closing the connection at the transport level).

As you say, if any error is returned when making requests or via response callback, the session should be considered invalid and restarted by the app.

Socket creation could be handled via callback as you suggests, or simply a FD to a created (not yet connected) socket could be passed to `coap_client_tcp_connect()`. Either should work, and I agree that this should be easier than having to cover socket creation with all the details within the library.

> Then upon using the callback it can automatically handle the CSMs and the book-keeping needed for it. This would restrict a client to one connected context at a time but that restriction already exists in the form of a single socket fd in the UDP client.

I think it's quite natural that single client context maintains a single connection. If needed, multiple client contexts can be created (at least that's how it works with regular CoAP client).

mpenate-ellenbytech:
> > The base value for the [Block-Wise-Transfer Capability Option](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3.2) is unspecified meaning that generic block-wise transfer and BERT block-wise transfer should technically be runtime-disabled until this CSM is negotiated with the server correct?
> 
> That's my understanding as well - lack of Block-wise-transfer option in CSM indicates no block transfer support by the sender, so block transfers should not be used over such connection. Probably good-enough approach for initial implementation.

Added a bool to the client struct `blockwise_enabled` to track status from CSMs received. Added a connect and close function. Trying to send a request when there is no socket (ie. the client needs to connect)  returns `-ENOTCONN`. Poll errors result in the connection being closed and the socket reset to `-1` (causing above mentioned `-ENOTCONN`). 

> I agree, client context in CoAP over TCP is practically bound to the underlying TCP/TLS connection, so it makes sense to integrate the client with the socket more closely. I'd imagine we'd have two additional session-related functions for CoAP over TCP...

The connect function calls a simple callback currently (callback stored in the client struct)
```
	coap_client_tcp_socket_config_cb_t socket_config_cb;
	void *socket_config_cb_user_data;
```

In this callback the application developer can call `zsock_setsockopt` to configure the socket fd (ie. binding to an interface, specifying TLS credential tags, configuring keep alive, etc.). If this callback returns a negative value that value is returned as the result from `coap_client_tcp_connect`. Otherwise after the callback, zsock_connect is attempted (**currently** within the thread context of an application thread, not CoAP thread, should this change?). Failures handled still (closing/cleaning the socket fd), otherwise on successful connect a CSM is immediately queued as the spec states the client should always send a CSM message.

Some public APIs are missing documentation.

I left the CSM function declaration in the header as the runtime application should technically be able to change the MMS at runtime and throughout an active connection.

> I see in the RFC that there are also various signals to notify this, instead of just closing the connection at the transport level

Looking at targeting some of these next.

rlubos:
> Otherwise after the callback, zsock_connect is attempted (currently within the thread context of an application thread, not CoAP thread, should this change?)

I don't think that's a problem, probably even better as this allows to detect connect issues early and return an error from `coap_client_tcp_connect()`.

include/zephyr/net/coap_client_tcp.h
```diff
@@ -0,0 +1,333 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
```
rlubos: Please update the copyright and date

include/zephyr/net/coap_client_tcp.h
rlubos: Any thoughts to this? It still seems to me that part of the concepts (request/response representation, option etc.) could be shared between implementations and only differentiate at the API level.

subsys/net/lib/coap/Kconfig
```diff
@@ -172,6 +179,39 @@ config COAP_CLIENT_TRUNCATE_MSGS
 
 endif # COAP_CLIENT
 
+if COAP_OVER_RELIABLE_TRANSPORT
+
+config COAP_CLIENT_TCP
+	bool
```
rlubos: I think it should be possible to enable/disable this at the application level, one might only want to use low-level COAP TCP APIs w/o enabling the client. Also I don't think it should be enabled by default, let the application enable it if needed.

subsys/net/lib/coap/coap.c
```diff
@@ -979,40 +1002,8 @@ static uint8_t __coap_header_get_code(const struct coap_packet *cpkt)
 	return cpkt->data[1];
 }
 
-int coap_header_set_code(const struct coap_packet *cpkt, uint8_t code)
+static uint8_t __coap_header_code_sanitize(uint8_t code)
```
rlubos: No need for `__` prefix, just make it `coap_header_code_sanitize()`, not even sure if it's still allowed in Zephyr

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2171,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+	if (!res) {
+		return -EINVAL;
+	}
+
+	res = append_u8(cpkt, code);
+	if (!res) {
+		return -EINVAL;
+	}
+
+	if (token && token_len) {
```
rlubos: Please use explicit comparison with pointers or integers(`token != NULL && token_len > 0` etc.).

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2171,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+
+#if defined(CONFIG_COAP_OVER_RELIABLE_TRANSPORT)
+
+int coap_tcp_packet_init(struct coap_packet *cpkt, uint8_t *data, uint16_t max_len,
+			 uint8_t token_len, const uint8_t *token, uint8_t code)
+{
+	uint8_t hdr;
+	bool res;
+
+	if (NULL == cpkt || NULL == data ||  0 == max_len) {
```
rlubos: We usually do varaible first in comparisons (`cpkt == NULL`), it'd be more consistent with the rest of the codebase.

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2171,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+
+	if (NULL == cpkt || NULL == data) {
+		return -EINVAL;
+	}
+
+	if (len < COAP_TCP_BASIC_HEADER_SIZE) {
+		return -EINVAL;
+	}
+
+	if (options) {
```
rlubos: `options != NULL`, please also fix similar cases elsewhere.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1362 @@
...
+			return true;
+		}
+	}
+	return false;
+}
+
+static int coap_client_tcp_init_request(struct coap_client_tcp *client,
+					struct coap_client_tcp_request *req,
+					struct coap_client_tcp_internal_request *internal_req,
+					bool reconstruct)
```
rlubos: Do we still need the `reconstruct` flag, given each request has a separate buffer. Note this was removed from the regular coap_client.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1362 @@
...
+		}
+		if (blockwise_transfer) {
+			internal_req->offset += payload_len;
+		}
+	}
+
+	/* Request next block if needed */
+	if (blockwise_transfer && !last_block) {
+		ret = coap_client_tcp_init_request(client, &internal_req->coap_request,
+						   internal_req, true);
```
rlubos: Why would we set reconstruct flag to true here, if we're initiating next transaction (i.e. next block)? It seems that the behavior configured with `COAP_CLIENT_TCP_BLOCKWISE_REUSE_TOKEN` is now enforced unconditionally.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1362 @@
...
+{
+	int ret = 0;
+	int block_option;
+	int block_num;
+	bool blockwise_transfer = false;
+	bool last_block = false;
+	struct coap_client_tcp_internal_request *internal_req;
+	struct coap_option csm_option;
+	uint8_t cms_option_num;
+
```
rlubos: nit: unneeded empty line.

rlubos (commented):
> @rlubos I think this is at the point that it could be promoted from draft PR. LMK your thoughts.

Yeah I think it can be moved out of draft already.

Generally the implementation looks pretty good, although I still wanted to discuss a bit whether we could share some common data representations between regular coap_client and TCP variant. But if there are good reasons not to, no strong push from my side.

Also I think we need basic example of the new library usage (separate sample most likely). A doc update would also be appreciated.

samples/net/sockets/coap_client_tcp/boards/native_sim.conf
rlubos, line 1: TBH I'm not a fan of using offloaded sockets by default, IMHO the default choice should be the default native_sim networking (especially that's what documented in the readme file). If anything, we can provide an overlay file to facilitate configuring offloaded sockets for native sim (some samples provide `overlay-nsos.conf`).

rlubos (commented):
Nice, thanks for the sample update, I think it's much easier to follow now w/o the built-in server overhead. One final suggestion, but otherwise LGTM. You could possible pull it out of draft already?

samples/net/sockets/coap_client_tcp/boards/native_sim.conf
mpenate-ellenbytech: Yeah, on my setup I've had issues using TAP networking before because of QEMU VMs I have running for other work. I'll try to get it swapped. The docs build fail is why I went back to draft PR.

mpenate-ellenbytech: I suppose re-writing the sample 3 times helps find bugs haha. Noticed I wasn't matching CSM requests (specifically a ping) to old-but-still-allocated internal requests. Cleaning that up also pointed out another bug in the encoded buffer length that I managed to clean up too. Pushing (hopefully) last version soon and opening PR again.

rlubos (dismissed):
LGTM, Thanks

subsys/net/lib/coap/coap.c
```diff
@@ -2135,3 +2170,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+}
+
+uint8_t coap_tcp_header_get_code(const struct coap_packet *cpkt)
+{
+	if (cpkt == NULL || cpkt->data == NULL) {
+		return 0;
+	}
+
+	uint8_t header_ext_len = (cpkt->data[0] >> 4) & 0xF;
+	size_t offset_to_code = 1;
```
jukkar: The variables should be declared at the start of the block

subsys/net/lib/coap/coap.c
```diff
@@ -2135,3 +2170,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+
+	if (header_ext_len >= COAP_TCP_HEADER_LEN_EXT_1B) {
+		offset_to_code += (1 << (header_ext_len - COAP_TCP_HEADER_LEN_EXT_1B));
+	}
+
+	uint8_t code = cpkt->data[offset_to_code];
```
jukkar: ditto

subsys/net/lib/coap/coap.c
```diff
@@ -2135,3 +2170,433 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+		r = update_control_block2_tcp(ctx, block2, size2);
+		if (r) {
+			return r;
+		}
+
+		return update_descriptive_block(ctx, block1, size1 == -ENOENT ? 0 : size1);
+	}
+
+	r = update_control_block1_tcp(ctx, block1, size1);
+	if (r) {
```
jukkar: This should be written as `if (r != 0) {` for non boolean checks. I am seeing these all over the place, could check all the code for these.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+	k_sleep(K_MSEC(COAP_PERIODIC_TIMEOUT));
+}
+
+static void coap_client_tcp_recv(void *coap_cl, void *a, void *b)
+{
+	int ret;
+
+	k_sem_take(&coap_client_tcp_recv_sem, K_FOREVER);
+	while (true) {
+		ret = handle_poll();
```
jukkar, line 1145: Perhaps not really part of this PR, but it would make sense to investigate if the socket service API could be used here so save some resources.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+		client->event_cb = cb;
+		client->event_cb_user_data = user_data;
+	}
+}
+
+#define COAP_CLIENT_TCP_THREAD_PRIORITY CLAMP(CONFIG_COAP_CLIENT_THREAD_PRIORITY, \
+					      K_HIGHEST_APPLICATION_THREAD_PRIO, \
+					      K_LOWEST_APPLICATION_THREAD_PRIO)
+
+K_THREAD_DEFINE(coap_client_tcp_recv_thread, CONFIG_COAP_CLIENT_STACK_SIZE,
```
jukkar, line 1374: We could avoid this thread if using socket services API.

include/zephyr/net/coap_client_tcp.h
```diff
@@ -0,0 +1,363 @@
...
+#if defined(CONFIG_COAP_EXTENDED_OPTIONS_LEN)
+	/** Option len */
+	uint16_t len;
+	/** Buffer for the length */
+	uint8_t value[CONFIG_COAP_EXTENDED_OPTIONS_LEN_VALUE];
+#else
+	/** Option len */
+	uint8_t len;
+	/** Buffer for the length */
+	uint8_t value[12];
```
jukkar: Could you replace the magic constant with a define

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+		size_t ext_bytes = 1 << (len_field - COAP_TCP_HEADER_LEN_EXT_1B);
+
+		header_size += ext_bytes;
+
+		if (offset < header_size) {
+			return -EAGAIN;
+		}
+
+		/* Parse extended length */
+		uint32_t ext_val = 0;
```
jukkar: Declaration at the start of the block please.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+	for (int i = 0; i < CONFIG_COAP_CLIENT_MAX_REQUESTS; i++) {
+		if (!client->requests[i].request_ongoing &&
+		    exchange_lifetime_exceeded(&client->requests[i])) {
+			return &client->requests[i];
+		}
+	}
+
+	/* Find oldest non-observe request to reuse */
+	struct coap_client_tcp_internal_request *oldest = NULL;
+	int64_t oldest_time = INT64_MAX;
```
jukkar: same here

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+				/* Trigger blockwise - we don't know total size yet */
+				total_len = cb_len + 1;
+			}
+		}
+
+		/* Check if blockwise send is needed.
+		 * Only use blockwise if enabled via CSM negotiation AND payload exceeds limit
+		 * or we're continuing a blockwise transfer.
+		 */
+		bool use_blockwise = client->blockwise_enabled &&
```
jukkar: and here

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+		client->recv_offset = 0;
+	}
+
+	k_mutex_unlock(&client->lock);
+}
+
+void coap_client_tcp_cancel_requests(struct coap_client_tcp *client)
+{
+	cancel_requests_with(client, -ECANCELED);
+	k_sleep(K_MSEC(COAP_PERIODIC_TIMEOUT));
```
jukkar, line 1115: Why do we need these 500 ms sleeps here and also in various other places?

samples/net/sockets/coap_client_tcp/src/main.c
```diff
@@ -0,0 +1,183 @@
...
+				      IPPROTO_TCP);
+	if (ret < 0) {
+		LOG_ERR("Failed to connect: %d", ret);
+		return ret;
+	}
+
+	LOG_INF("Connected");
+
+	/* Wait for CSM exchange */
+	k_sleep(K_SECONDS(1));
```
jukkar: I am not really fan of these hard coded sleeps. Would there be any other way to detect what state we are in the connection? For example could we we wait for an event that tells that "CSM exchange" has happened etc and the event would be sent by the core TCP COAP code?
Hardcoding these sleep (even in sample) is error prone because what if CSM exchange takes longer than 1 second (probably never but still a possibility).

jukkar (commented):
Some style issues comments.

The bigger changes could be done in the subsequent PRs if you are willing to commit to develop the code further. Like
* investigating to get rid of handler thread and replace it by socket service api
* investigating to send events in various stages in the connection so that we can avoid having hard coded sleeps in the application

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: That function is mostly a copy of `coap_update_from_block` but calling TCP funcs, I updated the original to have `r != 0`, also fixed other locations.

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: Addressed in next push

include/zephyr/net/coap_client_tcp.h
mpenate-ellenbytech: Want me to add one to the general coap header and replace the constant 12 in the original coap client header as well?

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: Addressed in next push

samples/net/sockets/coap_client_tcp/src/main.c
mpenate-ellenbytech: Already had events in, this was just a simple sample but it really isn't much for me to add it. Next push will wait on sems instead of having arbitrary sleeps.

rlubos (dismissed):
Compliance failure needs fixing:
>  LINE_SPACING: Missing a blank line after declarations
File:subsys/net/lib/coap/coap.c
Line:2400

rlubos (dismissed):

include/zephyr/net/coap.h
```diff
@@ -1278,6 +1315,119 @@ struct coap_transmission_parameters coap_get_transmission_parameters(void);
...
+ * @brief Returns the data pointer and length of the CoAP TCP packet.
+ *
+ * @param cpkt CoAP TCP packet representation
+ * @param len Total length of CoAP payload
+ *
+ * @return data pointer and length if payload exists
+ *         NULL pointer and length set to 0 in case there is no payload
+ */
+const uint8_t *coap_tcp_packet_get_payload(const struct coap_packet *cpkt,
+				       uint32_t *len);
```
pdgendt: Alignment should be fixed for all `_tcp` variants

```suggestion
const uint8_t *coap_tcp_packet_get_payload(const struct coap_packet *cpkt,
					   uint32_t *len);
```

include/zephyr/net/coap_client_tcp.h
jukkar: What ever makes sense, just trying to avoid magic constants.

subsys/net/lib/coap/coap.c
```diff
@@ -82,6 +83,15 @@ static inline void encode_be16(struct coap_packet *cpkt, uint16_t offset, uint16
 	cpkt->offset += 2;
 }
 
+static inline void encode_be32(struct coap_packet *cpkt, uint16_t offset, uint32_t data)
+{
+	cpkt->data[offset] = data >> 24;
+	cpkt->data[offset + 1] = data >> 16;
+	cpkt->data[offset + 2] = data >> 8;
+	cpkt->data[offset + 3] = (uint8_t)data;
```
mniestroj:
```suggestion
	sys_put_be32(data, &cpkt->data[offset]);
```

subsys/net/lib/coap/coap.c
```diff
@@ -1441,7 +1476,7 @@ int insert_option(struct coap_packet *cpkt, uint16_t code, const uint8_t *value,
 static int update_descriptive_block(struct coap_block_context *ctx,
 				    int block, int size)
 {
-	size_t new_current = GET_NUM(block) << (GET_BLOCK_SIZE(block) + 4);
+	size_t new_current = GET_NUM(block) << (MIN(6, GET_BLOCK_SIZE(block)) + 4);
```
mniestroj:
```suggestion
	size_t new_current = GET_NUM(block) << (MIN(COAP_BLOCK_1024, GET_BLOCK_SIZE(block)) + 4);
```

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+ * @brief Send a CoAP message over TCP
+ */
+static int send_request_tcp(int sock, const void *buf, size_t len, int flags)
+{
+	int ret;
+
+	size_t bytes_written = 0;
+
+	while (len > bytes_written) {
+		ret = zsock_send(sock, buf, len, flags);
```
mniestroj: `buf` and `len` should be updated on partial send. Otherwise the same buffer/data will be sent in a while loop.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+			k_mutex_unlock(&client->lock);
+
+			/* Consume processed packet and preserve any remaining bytes */
+			if (client->recv_offset > pkt_len) {
+				memmove(client->recv_buf, client->recv_buf + pkt_len,
+					client->recv_offset - pkt_len);
+				client->recv_offset -= pkt_len;
+			} else {
+				client->recv_offset = 0;
+			}
```
mniestroj: What happens if we receive 2 full CoAP packets with single `zsock_recv()` call? Second packet will be moved with `memmove()`, but when will it be consumed/parsed? The next poll() iteration might not happen (if there is nothing more than those two packets to be received).

subsys/net/lib/coap/coap.c
```diff
@@ -2136,3 +2171,435 @@ void coap_set_transmission_parameters(const struct coap_transmission_parameters
...
+	default:
+		break;
+	}
+
+	hdr = (header_len_ext_val << 4) & 0xF0;
+	hdr |= token_len & 0xF;
+	cpkt->data[0] = hdr;
+
+	/* Header/TKL + Extended Header Length + Code + Token*/
+	cpkt->hdr_len = COAP_TCP_BASIC_HEADER_SIZE + current_ext_len_bytes + token_len;
```
mniestroj:
```suggestion
	cpkt->hdr_len = COAP_TCP_BASIC_HEADER_SIZE + needed_ext_len_bytes + token_len;
```
should it be `needed_ext_len_bytes` (the new size)?

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+int coap_client_tcp_connect(struct coap_client_tcp *client,
+				const struct net_sockaddr *addr, net_socklen_t addrlen, int proto)
+{
+	int ret;
+
+	if (NULL == addr || 0 == addrlen) {
+		return -EINVAL;
+	}
+
+	if (client->fd > 0) {
```
mniestroj:
```suggestion
	if (client->fd >= 0) {
```

0 is also a valid file descriptor. Maybe we should initialize it to -1.

subsys/net/lib/coap/coap_client_tcp.c
```diff
@@ -0,0 +1,1364 @@
...
+	i = n;
+	while (i--) {
+		option->value[i] = val & 0xff;
+		val >>= 8;
+	}
+
+	return n;
+}
+
+static struct coap_client_tcp_option tcp_csm_options[2];
```
mniestroj: Why is that a static variable? This prevents `coap_client_tcp_csm_req()` to be used by multiple threads.

mniestroj (changes requested):

subsys/net/lib/coap/coap.c
mpenate-ellenbytech: Yes, reflected in most recent push. Thanks.

subsys/net/lib/coap/coap_client_tcp.c
mpenate-ellenbytech: I had to add handling for rx'ing multiple packets. I thought about TCP segmentation giving more packets than coap data but not the other way around.

include/zephyr/net/coap_client_tcp.h
mpenate-ellenbytech: Added `MAX_COAP_CLIENT_OPTION_LEN` to coap_client.h

rlubos (approved):

mniestroj (approved):

PR #100596, opened 2025-12-05
drivers: can: use proper namespace for driver-specific Kconfig symbols

henrikbrixandersen:
Currently, quite a few CAN controller drivers define Kconfig symbols with identical names (e.g. `CONFIG_CAN_MAX_FILTER`) but the ranges and defaults of these symbols differ between drivers, causing clashes and misconfigurations when multiple drivers are enabled concurrently.

Rename these driver-specific Kconfig symbols to use proper namespacing and increase their defaults where needed to support typical software needs and in-tree samples.

This problem became evident with the reshuffling of the Kconfig.* include order in 30950b888d368121e15b44e8348054e612b4ec16.


alexanderwachter (dismissed):

martinjaeger (commented):
Now that the Kconfig is changed anyway, should it not be plural? `_MAX_FILTERS`

alexanderwachter (approved):

martinjaeger (approved):

PR #100598, opened 2025-12-05
[Backport v4.3-branch] sys: getopt: Rename getopt to sys_getopt to avoid conflicts with Posix

mmahadevan108:
Backport d164f8b55d08aba8084fcd5e0d3bce93263f1aa3~2..d164f8b55d08aba8084fcd5e0d3bce93263f1aa3 from https://github.com/zephyrproject-rtos/zephyr/issues/98845

Fixes #100731

### DIFF
--- a/doc/services/shell/index.rst
+++ b/doc/services/shell/index.rst
@@ -678,16 +678,16 @@ The shell module supports the following meta keys:
 
 This feature is activated by :kconfig:option:`CONFIG_SHELL_METAKEYS` set to ``y``.
 
-Getopt Feature
-*****************
+sys_getopt Feature
+******************
 
 Some shell users apart from subcommands might need to use options as well.
 the arguments string, looking for supported options. Typically, this task
 is accomplished by the ``getopt`` family functions.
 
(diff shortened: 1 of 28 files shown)


peter-mitsis (approved):

jukkar:
@mmahadevan108 why the backport, the getopt renaming is not really a bug fix?

jukkar (approved):

nashif:
rebase

nashif (changes requested):
I do not think this is something that should be backported. Whatever issues this is supposed to fix need to be dealt with without changing APIs and such.

mmahadevan108:
There are build failures seen in 4.3 release as stated in the bug report. Any suggestions on an alternate fix?

nashif:
> There are build failures seen in 4.3 release as stated in the bug report. Any suggestions on an alternate fix?

matter is not part of zephyr in 4.3 or zephyr in general as of now,  I do not have an immediate alternate fix, but backporting major cleanup and api changes (to a certain degree) is not something we should be doing, this might break existing users, i.e. all of the sudden getopt does not work and people need to use sys_getopt.

PR #100604, opened 2025-12-05
boards: ai_m61_32s_kit: fix wrong values and supported test

VynDragon:
Input the wrong number in flash size, also should not run display tests.

Realized while updating another board.


josuah (approved):
Seems like I missed that during the review.

Good catch!

nordicjm (approved):

PR #100606, opened 2025-12-05
drivers: i2s: Adding new i2s driver for PSE84

Peterson-Brett:
Driver development is still in progress and should not be expected to work for all use cases.

The I2s interface is implemented using the TDM hardware provided by the PSE84 device.


drivers/i2s/i2s_infineon.c
```diff
@@ -0,0 +1,1063 @@
...
+
+	if (CY_TDM_SUCCESS == Cy_AudioTDM_Init(config->reg_addr, &config->tdm_config)) {
+		if (dir == I2S_DIR_RX) {
+			Cy_AudioTDM_EnableRx(&config->reg_addr->TDM_RX_STRUCT);
+			if (I2S_OPT_LOOPBACK & i2s_cfg->options) {
+				Cy_AudioTDM_EnableRxTestMode(&config->reg_addr->TDM_RX_STRUCT);
+			}
+
+		} else {
+			Cy_AudioTDM_EnableTx(&config->reg_addr->TDM_TX_STRUCT);
```
Alex9360, line 611: After enabling tx it immediately goes to the isr as it has less entries than trigger level . CY_TDM_INTR_TX_FIFO_TRIGGER interrupt gets occured. disable interrupt during configuration.

Alex9360 (commented):
I have interfaced the external audio codec to this i2s, it doesn't  shown any pcm data, clock on the i2s pins. have you tested with any codec device?

drivers/i2s/i2s_infineon.c
Peterson-Brett: I have tested with the tlv320dac codec chip. It is necessary to preload the transmit queue (usually by calling i2s_buf_write) before sending the I2S_TRIGGER_START so that data is available to transmit to the hardware when starting a transmission. Otherwise, the hardware will immediately detect an underflow. This flow can be seen in the i2s test code.

drivers/i2s/i2s_infineon.c
```diff
@@ -0,0 +1,1072 @@
...
+	/* empty the queue in case of re-configure */
+	while (k_msgq_get(&stream->queue, &queue_element, K_NO_WAIT) == 0) {
+		/* free all queued buffers */
+		k_mem_slab_free(stream->cfg.mem_slab, queue_element.buffer);
+	}
+
+	/* Save configuration for i2s_config_get */
+	memcpy(&stream->cfg, i2s_cfg, sizeof(struct i2s_config));
+
+	configiure_i2s_clock(dev, dir);
```
teburd: s/configiure/configure

drivers/serial/uart_infineon_pdl.c
```diff
@@ -1385,6 +1386,7 @@ static int ifx_cat1_uart_init(const struct device *dev)
 		data->async.dma_tx.dma_cfg.head_block = &data->async.dma_tx.blk_cfg;
 		data->async.dma_tx.dma_cfg.user_data = (void *)dev;
 		data->async.dma_tx.dma_cfg.dma_callback = dma_callback_tx_done;
+		data->async.dma_tx.dma_cfg.source_handshake = 1;
```
teburd: This should come from DT rather than hard coding, assuming this is the hardware DMA trigger mux source

teburd (changes requested):
Looks pretty good to me

A small typo here that I spotted, and some bigger questions around this i2s API that aren't answerable here in a PR that implements it for our part.

teburd (dismissed):

teburd (dismissed):

drivers/i2s/Kconfig.infineon
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2025 Infineon Technologies AG,
+# or an affiliate of Infineon Technologies AG.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Infineon I2S configuration options
+
+config I2S_INFINEON
```
mcatee-infineon, line 8: Would it make sense to have a select for DMA included here like for the SPI_DMA configuration for the SPI driver. However, I doubt this driver would ever build without DMA configured. So, I could see there being no need.

drivers/i2s/i2s_infineon.c
```diff
@@ -0,0 +1,1070 @@
...
+	int ret = 0;
+
+	switch (dir) {
+	case I2S_DIR_RX:
+		stream = &data->rx;
+		break;
+	case I2S_DIR_TX:
+		stream = &data->tx;
+		break;
+	case I2S_DIR_BOTH:
```
mcatee-infineon, line 713: Would a clarifying comment or log be desired here, similar to other examples of this check in the driver? (e.g. line 634)

drivers/i2s/i2s_infineon.c
```diff
@@ -0,0 +1,1070 @@
+/*
+ * Copyright (c) 2025 Infineon Technologies AG,
+ * or an affiliate of Infineon Technologies AG.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT infineon_i2s
+
+#include <zephyr/drivers/i2s.h>
```
mcatee-infineon, line 10: You would likely want to include the Infineon "infineon_kconfig.h" file at the top of your includes, if this driver is intended to support other Infineon boards in the future. It would potentially be needed to ensure that header defines are parsed correctly for a given board. An example of said include is present in the uart driver modified in this PR.

mcatee-infineon (changes requested):

drivers/i2s/Kconfig.infineon
Peterson-Brett: Yes, selecting DMA should be done here. I'll make this change.

drivers/i2s/i2s_infineon.c
Peterson-Brett: Sure, I'll add the log message to be consistent.

drivers/i2s/i2s_infineon.c
Peterson-Brett: I'll add the include to be consistent with other drivers.

mcatee-infineon (approved):

teburd (approved):

PR #100607, opened 2025-12-05
style: drivers: use Kconfig glob for sourcing all sub-Kconfig

josuah:
Dependency:

- https://github.com/zephyrproject-rtos/zephyr/pull/100624

It seems to work well for I2S:

https://github.com/zephyrproject-rtos/zephyr/blob/8053d7793f13a89bc660381a8eae5e8172e79c7e/drivers/i2s/Kconfig#L28

I have not unified all drivers to either absolute (`source`) or relative (`rsource`), but can do it easily in the process.

It removes one manual step for each contribution.

Current PRs that add a Kconfig entry might need to remove their `Kconfig` addition.


josuah:
Manual ping to the audio collaborator: the Kconfig in audio were renamed to allow a glob to happen:
https://github.com/zephyrproject-rtos/zephyr/pull/100607/commits/0b42527eab651666221771bba7d2ae0d74b0d4e0

- I can revert that if manual sorting is preferred.
- I can further rename the source files to make them match the Kconfig, here I went with least modification possible (Kconfig only).

henrikbrixandersen (changes requested):
Please don't. This will source intermediate editor files, editor backup files etc, and break the development environment for a lot of end users.

PR #100619, opened 2025-12-06
drivers: ethernet: sensry: Fix phy handle instance and MAC config

pdgendt:
- Fix multi-instance PHY handle
- Add MAC address config support


maass-hamburg (approved):

tswaehn (approved):
Just tested on hardware, static and random addresses are working.
(and thank you for fixing my handle-bug)

PR #100623, opened 2025-12-06
Fix comment for I2C_CLK_LIMIT_XTAL definition

kakmonk:
Updating the comment to signal its limitation to XTAL, not RTC, to avoid confusion.


drivers/i2c/i2c_esp32.c
```diff
@@ -39,7 +39,7 @@ LOG_MODULE_REGISTER(i2c_esp32, CONFIG_I2C_LOG_LEVEL);
 #define I2C_CLK_LIMIT_REF_TICK (1 * 1000 * 1000 / 20)	/* REF_TICK, no more than REF_TICK/20*/
```
sylvioalves, line 39: Hi @kakmonk, I think all those defines (`I2C_CLK_LIMI_xxx`) are not being used at all and is probably leftovers from initial code. Perhaps you could update your PR to remove it all instead.

PR #100628, opened 2025-12-07
video: allow placement of the video buffer pool in a ZEPHYR REGION (and adapt STM32N6 / STM32H7RS config)

avolmat-st:
The video buffer pool is placed in the default memory region. Due to that it wasn't possible to allow video buffer in other memory and multi-heap / shared-multi-heap was used.
Another alternative is simply to be able to indicate in which region the video buffer pool should be placed.
Similar approach is now available for LVGL buffers via dedicated config, making usage under the hood of the Z_GENERIC_SECTION mechanism.

Similar approach is proposed here, via addition two new KConfig for video subsystem VIDEO_BUFFER_POOL_ZEPHYR_REGION and VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME, to indicate where the video buffer pool should be placed.

This gives more flexibility on the whole soc memory allocation strategy, by simply partitioning the memory via DT, giving required caching properties as well.
Tricky part is that, depending on the memory, it might not be available yet from the very beginning, making it impossible to use the normal Z_HEAP_DEFINE_BY_SECT macro since this would lead to configuration the heap too early during the boot at a moment when the memory isn't yet available.
To overcome this, this manually configure the k_heap via the k_heap_init on the very first allocation request.

Two other commits are to move away from multi-heap / shared-multi-heap for LTDC / video buffer on STM32 platforms.
(if this is agreed, it would probably be necessary to either remove the SMH configuration of the MEMC XSPI/OSPI  or at least make it configurable ?)


josuah:
It definitely sounds like an improvement, thank you!

Currently we have `CONFIG_VIDEO_BUFFER_POOL_SZ_MAX * CONFIG_VIDEO_BUFFER_POOL_NUM_MAX` to set the size, but currently this does not work (allocation fails) because there is overhead for the heap header struct, for alignment, and fragmentation can happen on top so `malloc()/free()` can randomly make the allocation fail at runtime (which is helped by reusing buffers).

Maybe a different API would be `CONFIG_VIDEO_BUFFER_POOL_SIZE=` computed by the user.

I have the impression that what users would want is `video_buf_alloc(dev);`, then with `video_buf_alloc()`:

- If it asks the number of buffers via a CID
- It asks the alignment requirements via a CID
- It asks the buffer size via the video API
- It enqueues the video buffers right away
- It uses the memory region from Kconfig, or an extra parameter for an `_smh()` variant maybe.

What you proposed does not seem to require any API change, so it is probably not a problem to change the internals frequently... Just an opportunity to rework things.

drivers/video/video_common.c
```diff
@@ -27,10 +27,38 @@ LOG_MODULE_REGISTER(video_common, CONFIG_VIDEO_LOG_LEVEL);
...
+#endif
+
+#define VIDEO_BUFFER_POOL_SIZE (CONFIG_VIDEO_BUFFER_POOL_SZ_MAX * CONFIG_VIDEO_BUFFER_POOL_NUM_MAX)
+
+/*
+ * The k_heap is manually initialized instead of using directly Z_HEAP_DEFINE_IN_SECT
+ * since the section might not be yet accessible from the beginning, making it impossible
+ * to initialize it if done via Z_HEAP_DEFINE_IN_SECT
+ */
+static char VIDEO_BUFFER_POOL_REGION_NAME
```
ngphibang: `static const char`

drivers/video/Kconfig
```diff
@@ -35,6 +35,19 @@ config VIDEO_BUFFER_POOL_ALIGN
...
+	  whose name is specified in CONFIG_VIDEO_BUFFER_ZEPHYR_REGION_NAME.
+
+config VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME
+	string "Name of the zephyr memory-region for video buffer pool"
+	help
+	  Name of the Zephyr memory-region where video buffer pool is placed.
+
 config VIDEO_BUFFER_USE_SHARED_MULTI_HEAP
 	bool "Use shared multi heap for video buffer"
 	default n
```
ngphibang: no need

ngphibang:
> Currently we have CONFIG_VIDEO_BUFFER_POOL_SZ_MAX * CONFIG_VIDEO_BUFFER_POOL_NUM_MAX to set the size, but currently this does not work (allocation fails) because there is overhead for the heap header struct, for alignment, and fragmentation can happen on top so malloc()/free() can randomly make the allocation fail at runtime (which is helped by reusing buffers).

I think this is because one of the two Kconfig exposed to user, the `CONFIG_VIDEO_BUFFER_POOL_SZ_MAX` does not make sense and cause the confusion, i.e., in fact, if we allocate less than `CONFIG_VIDEO_BUFFER_POOL_NUM_MAX`, the size of one buffer can still be greater than `CONFIG_VIDEO_BUFFER_POOL_SZ_MAX`. A simple fix is to expose

- `CONFIG_VIDEO_BUFFER_POOL_NUM_MAX` : max numbers of buffers (for video_buf[] array)
- `CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE` : size in bytes of the video buffer pool. This way user can specify the total size and does not need to adjust the header struct to their buffer size, etc.

drivers/video/Kconfig
josuah: ~~Do you mean to drop SMH support after this is merged?~~

avolmat-st: I think this refer to the need of default n.

drivers/video/video_common.c
avolmat-st: Actually having 2nd thought about this change in const. This is then given to k_heap_init which anyway will need a (void *) so, losing the const statement.

drivers/video/video_stm32_venc.c
```diff
@@ -16,8 +16,14 @@
 #include <zephyr/irq.h>
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
+#if defined(CONFIG_VIDEO_STM32_VENC_SMH_ALLOC)
 #include <zephyr/multi_heap/shared_multi_heap.h>
 
+#define EWL_HEAP_ALIGNED_ALLOC(size)\
+	shared_multi_heap_aligned_alloc(CONFIG_VIDEO_BUFFER_SMH_ATTRIBUTE, ALIGNMENT_INCR, (size))
+#define EWL_HEAP_ALIGNED_FREE(block) shared_multi_heap_free(block)
+#endif
```
etienne-lms: Is mixing header files inclusion and macro definition really nice? Always including **shared_multi_heap.h** should not be an issue. Suggestion
```patch
 #include <zephyr/multi_heap/shared_multi_heap.h>

 #include <ewl.h>
 
 (...)

 #define ALIGNMENT_INCR 8UL

+#if defined(CONFIG_VIDEO_STM32_VENC_SMH_ALLOC)
 #define EWL_HEAP_ALIGNED_ALLOC(size)\
 	shared_multi_heap_aligned_alloc(CONFIG_VIDEO_BUFFER_SMH_ATTRIBUTE, ALIGNMENT_INCR, (size))
 #define EWL_HEAP_ALIGNED_FREE(block) shared_multi_heap_free(block)
+#endif /* CONFIG_VIDEO_STM32_VENC_SMH_ALLOC */

File: samples/drivers/video/capture/boards/arduino_nicla_vision_stm32h747xx_m7.conf [hunk 4]
```diff
@@ -1,5 +1,5 @@
+CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE=320000
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: It is by default. Not too related to the PR but could you drop this as well ?

samples/drivers/video/capture/boards/ek_ra8d1.conf
```diff
@@ -1,4 +1,4 @@
 CONFIG_VIDEO_FRAME_WIDTH=320
 CONFIG_VIDEO_FRAME_HEIGHT=240
-CONFIG_VIDEO_BUFFER_POOL_SZ_MAX=155000
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: ditto

samples/drivers/video/capture/boards/frdm_mcxn947_cpu0.conf
```diff
@@ -1,2 +1,2 @@
-CONFIG_VIDEO_BUFFER_POOL_SZ_MAX=40000
+CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE=80000
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: ditto

samples/drivers/video/capture/boards/stm32n6570_dk_stm32n657xx_sb.conf
```diff
@@ -1,6 +1,6 @@
 CONFIG_VIDEO_FRAME_WIDTH=800
 CONFIG_VIDEO_FRAME_HEIGHT=480
 CONFIG_VIDEO_PIXEL_FORMAT="RGBP"
-CONFIG_VIDEO_BUFFER_POOL_SZ_MAX=800000
+CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE=1600000
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: ditto

samples/subsys/usb/uvc/boards/frdm_mcxn947_mcxn947_cpu0.conf
```diff
@@ -1,2 +1,2 @@
-CONFIG_VIDEO_BUFFER_POOL_SZ_MAX=40000
+CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE=80000
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: ditto

samples/subsys/usb/uvc/prj.conf
```diff
@@ -11,6 +11,6 @@ CONFIG_USB_DEVICE_STACK_NEXT=y
 CONFIG_CDC_ACM_SERIAL_INITIALIZE_AT_BOOT=n
 CONFIG_VIDEO=y
 CONFIG_VIDEO_BUFFER_POOL_NUM_MAX=2
```
ngphibang: ditto

samples/subsys/usb/uvc/src/main.c
```diff
@@ -100,7 +100,7 @@ static int app_add_format(uint32_t pixfmt, uint32_t width, uint32_t height, bool
 		return ret;
 	}
 
-	if (fmt.size > CONFIG_VIDEO_BUFFER_POOL_SZ_MAX) {
+	if (fmt.size > (CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE / CONFIG_VIDEO_BUFFER_POOL_NUM_MAX)) {
```
ngphibang, line 103: The check is not really correct, I think. fmt.size can be greater than `(CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE / CONFIG_VIDEO_BUFFER_POOL_NUM_MAX)` if we allocate less than `CONFIG_VIDEO_BUFFER_POOL_NUM_MAX` buffers

tests/drivers/video/test_pattern/src/main.c
```diff
@@ -206,7 +206,7 @@ static void *test_pattern_setup(void)
 	/* Alloc video buffers and enqueue for capture */
 	zassert(caps.min_vbuf_count <= CONFIG_VIDEO_BUFFER_POOL_NUM_MAX,
 		"not enough buffers");
-	zassert(fmt.size <= CONFIG_VIDEO_BUFFER_POOL_SZ_MAX,
+	zassert(fmt.size <= (CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE / CONFIG_VIDEO_BUFFER_POOL_NUM_MAX),
```
ngphibang, line 209: ditto

samples/subsys/usb/uvc/src/main.c
josuah: It is also possible to allocate a buffer for just 30 lines, like FRDM-MCXN497 and `vbuf->line_offset` (fragmented buffers).

doc/releases/release-notes-4.4.rst
```diff
@@ -173,6 +156,12 @@ New APIs and options
 
   * :c:macro:`COND_CASE_1`
 
+* Video
+
+  * :kconfig:option:`CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE`
+  * :kconfig:option:`CONFIG_VIDEO_BUFFER_POOL_ZEPHYR_REGION`
+  * :kconfig:option:`CONFIG_VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME`
```
josuah: It is tempting to shorten them as:
- `CONFIG_VIDEO_BUFFER_POOL_SIZE` (or `_BUFFER_HEAP_SIZE`?)
- `CONFIG_VIDEO_BUFFER_REGION` (or `_BUFFER_MEMORY_REGION`?)
- `CONFIG_VIDEO_BUFFER_REGION_NAME`

samples/subsys/usb/uvc/src/main.c
avolmat-st: I made a shortcut here to make it easier but indeed that isn't fully correct. In fact the correct check is:

- in case of there is no encoder:
    2 * fmt.size > CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE
- in case of there an encoder:
    (2 * fmt.size + encoder related buffers) > CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE.

samples/drivers/video/capture/boards/arduino_nicla_vision_stm32h747xx_m7.conf
avolmat-st: Just to be sure, you refer to the NUM_MAX=2 right ? Ok then.

ngphibang:
> Should I keep those 5 commits related to POOL_HEAP_SIZE or squash all into 1 commit (which would make conf update / framework update at once which I do not feel right), or 2 commits (one for framework update and 1 for conf updates) but in this case this will break during 1 commit so not good from a bisect point of view.

Yes, we need to keep bisectability. Keeping 5 commits or squashing into 1 commit are both fine to me (personally, I prefer one commit because it is simpler).

Just still a typo in the commit title  "video: allow placement of video buffer **pool** in a Zephyr region"

samples/subsys/usb/uvc/src/main.c
avolmat-st: @ngphibang, @josuah, actually, looking more into detail how it is currently done within the UVC sample, allocation is done as following:

within the main function:
```c
        uint32_t uvc_buf_count = CONFIG_VIDEO_BUFFER_POOL_NUM_MAX;
```
then, based on this that number is either fully allocated for the camera side, or splitted into 2 and half allocated for the camera and half for the encoder.

So, based on that, I am not sure we can make something more generic than what is currently done / proposed which is:

```c
if (fmt.size > (CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE / CONFIG_VIDEO_BUFFER_POOL_NUM_MAX))
```

Of course that one doesn't take into consideration if there is an encoder or not and the size generated by the encoder, but this might still be a good enough approximation no ?

ngphibang: Ok, that's good enough.

tests/drivers/video/test_pattern/src/main.c
avolmat-st: Actually same point here as well if we consider the whole code from the test.

In this test CONFIG_VIDEO_BUFFER_POOL_NUM_MAX buffers are being allocated.


```c

        /* Alloc video buffers and enqueue for capture */
        zassert(caps.min_vbuf_count <= CONFIG_VIDEO_BUFFER_POOL_NUM_MAX,
                "not enough buffers");
        zassert(fmt.size <= (CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE / CONFIG_VIDEO_BUFFER_POOL_NUM_MAX),
                "buffers too large");

        for (int i = 0; i < CONFIG_VIDEO_BUFFER_POOL_NUM_MAX; i++) {
                vbuf = video_buffer_aligned_alloc(fmt.size, CONFIG_VIDEO_BUFFER_POOL_ALIGN,
                                                  K_NO_WAIT);
                zassert_not_null(vbuf);

                vbuf->type = VIDEO_BUF_TYPE_OUTPUT;

                ret = video_enqueue(video_dev, vbuf);
                zassert_ok(ret);
        }
```

samples/drivers/video/capture/boards/arduino_nicla_vision_stm32h747xx_m7.conf
avolmat-st: Done

erwango (dismissed):
Approved from STM32 side.
A rebase is required, though

ngphibang (approved):

JarmouniA (commented):
Nit: I believe you should combine the following commits into 1:
- video: add CONFIG_VIDEO_BUFFER_POOL_HEAP_SIZE in configs
- video: replace CONFIG_VIDEO_BUFFER_POOL_SZ_MAX with POOL_HEAP_SIZE
- video: remove CONFIG_VIDEO_BUFFER_POOL_SZ_MAX in conf files

JarmouniA (approved):

doc/releases/migration-guide-4.4.rst
```diff
@@ -261,6 +261,9 @@ Video
 =====
 
 * CONFIG_VIDEO_OV7670 is now gone and replaced by CONFIG_VIDEO_OV767X.  This allows supporting both the OV7670 and 0V7675.
+* :kconfig:option:`CONFIG_VIDEO_BUFFER_POOL_SZ_MAX` is replaced by
```
aescolar, line 264: Isn't the video API stable, and therefore should this not be deprecated instead of just removed?

avolmat-st: @aescolar, thanks for the comment, I (and probably @josuah and @ngphibang as well (?)) am surprised to see that the video API is considered Stable. But indeed, this is the way it is described.
I went through the history of this API since its introduction in 2019, key commits, regarding introduction and stability informations are as below:

32169886ca7 ("drivers: Add video API")
39930ca2989 ("doc: reference: Add API Stability page")
  This commit introduced the video_api stability information, with video_api being Stable, introduced in version 2.1 and last modified in version 2.1
28d08ada28ea ("include/zephyr: Add 'version' and 'since' tag to groups") => v1.0.0 (info written into include/zephyr/drivers/video.h based on doc/develop/api/overview.rst in which video_api was written as STABLE since v2.1)
46a262ffe62 ("drivers: video: common: Add utilities to seek frmival/caps structures")    => v1.1.0

Basically, after the introduction commit, there has been 1 commit which update the API itself (I am not talking about treewide coccinelle based updates) and after that the API stability page has been introduced in Zephyr, stating that the video_api is **stable**.
This led to finally getting the version v1.0.0 when API version have been introduced (3rd commit).

In the last 1-2 years, there has been much more video activity and we are working on updating the video API in order to make it fit better with the use-cases for which it is being used. In that context we do update things that break existing APIs. Among them, this buffer pool config is more, previously there has been been the control API and the video structures as well are being updated.

So, overall, I don't think that this API should be considered for the time being as **Stable** (@ngphibang, @josuah, please correct me if I am wrong) and I would like to challenge the way it is currently reported.

ngphibang: Yes, I agree. The video API is under changes (e.g. buffer management rework, switch to RTIO, etc. are on-going). I think it should be considered "unstable".

drivers/video/Kconfig
```diff
@@ -35,6 +35,18 @@ config VIDEO_BUFFER_POOL_ALIGN
...
+	bool "Place video buffer pool in a Zephyr memory-region"
+	help
+	  Place video buffer pool in a region, automatically
+	  created from a 'zephyr,memory-region' compatible node,
+	  whose name is specified in CONFIG_VIDEO_BUFFER_ZEPHYR_REGION_NAME.
+
+config VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME
+	string "Name of the zephyr memory-region for video buffer pool"
+	help
+	  Name of the Zephyr memory-region where video buffer pool is placed.
```
ngphibang, line 48: I think `VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME` needs "depends on" or "select" VIDEO_BUFFER_POOL_ZEPHYR_REGION.

BTW, these two Kconfig are for the same purpose. Can we just use `VIDEO_BUFFER_POOL_ZEPHYR_REGION_NAME` and compare it against `""` to avoid the need of activate both at the same time ?

PR #100641, opened 2025-12-08
boards: amd: versalnet_apu: Add SMP board variant support

kedareswararao:
Add a new board variant versalnet_apu/amd_versalnet_apu/smp to support Symmetric Multi-Processing (SMP) configuration for AMD Versal NET APU.

This commit introduces:
- SMP board variant with 4-core configuration (CONFIG_MP_MAX_NUM_CPUS=4)
- PSCI support for CPU power management operations
- ARM64 40-bit virtual and physical address space support
- Non-secure world execution (EL1 NS) configuration
- Compatible with Xilinx ARM Trusted Firmware boot flow

The SMP variant enables out-of-box multicore support for applications that require parallel processing capabilities on the Versal NET APU platform.

Usage:
  west build -b versalnet_apu/amd_versalnet_apu/smp <application>

Boot Flow: Boot PDI -> TF-A -> Zephyr (SMP)

Files added:
- boards/amd/versalnet_apu/versalnet_apu_amd_versalnet_apu_smp.yaml
- boards/amd/versalnet_apu/versalnet_apu_amd_versalnet_apu_smp.dts
- boards/amd/versalnet_apu/versalnet_apu_amd_versalnet_apu_smp_defconfig

Files modified:
- boards/amd/versalnet_apu/board.yml (added smp variant)

### DIFF
--- a/boards/amd/versalnet_apu/board.cmake
+++ b/boards/amd/versalnet_apu/board.cmake
@@ -18,11 +18,19 @@ set(QEMU_FLAGS_${ARCH}
   -m 2g
 )
 
+# Add SMP support if configured maxcpus parameter value is aligned with QEMU DT
+if(CONFIG_SMP AND CONFIG_MP_MAX_NUM_CPUS GREATER 1)
+  list(APPEND QEMU_SMP_FLAGS -smp maxcpus=20)
+endif()
+
 # Set TF-A platform for ARM Trusted Firmware builds
 if(CONFIG_BUILD_WITH_TFA)
   set(TFA_PLAT "versal_net")
-  # Add Versal NET specific TF-A build parameters
-  set(TFA_EXTRA_ARGS "TFA_NO_PM=1;PRELOADED_BL33_BASE=0x0")
+  # Configure TF-A memory location for Versal NET platform
+  # TF-A runs from DDR at address 0xf000000 (changed from default 0xbbf00000).
+  # This DDR location (VERSAL_NET_ATF_MEM_BASE=0xf000000) works for all possible designs.
+  # Note: If TF-A needs to run from OCM instead of DDR, PDI changes would be required.
+  set(TFA_EXTRA_ARGS "RESET_TO_BL31=1;PRELOADED_BL33_BASE=0x0;TFA_NO_PM=1;VERSAL_NET_ATF_MEM_BASE=0xf000000;VERSAL_NET_ATF_MEM_SIZE=0x50000")
   if(CONFIG_TFA_MAKE_BUILD_TYPE_DEBUG)
     set(BUILD_FOLDER "debug")
   else()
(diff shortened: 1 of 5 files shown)


boards/amd/versalnet_apu/board.cmake
```diff
@@ -18,11 +18,15 @@ set(QEMU_FLAGS_${ARCH}
...
+if(CONFIG_SMP AND CONFIG_MP_MAX_NUM_CPUS GREATER 1)
+  list(APPEND QEMU_SMP_FLAGS -smp cpus=${CONFIG_MP_MAX_NUM_CPUS},maxcpus=20)
+endif()
+
 # Set TF-A platform for ARM Trusted Firmware builds
 if(CONFIG_BUILD_WITH_TFA)
   set(TFA_PLAT "versal_net")
-  # Add Versal NET specific TF-A build parameters
-  set(TFA_EXTRA_ARGS "TFA_NO_PM=1;PRELOADED_BL33_BASE=0x0")
+  set(TFA_EXTRA_ARGS "RESET_TO_BL31=1;PRELOADED_BL33_BASE=0x0;TFA_NO_PM=1;VERSAL_NET_ATF_MEM_BASE=0xf000000;VERSAL_NET_ATF_MEM_SIZE=0x50000")
```
michalsimek, line 33: I can't see any explanation why TF-A location where changed. I can't see any issue with using default one.
RESET_TO_BL31=1 is default too.

boards/amd/versalnet_apu/board.cmake
```diff
@@ -18,11 +18,15 @@ set(QEMU_FLAGS_${ARCH}
   -m 2g
 )
 
+# Add SMP support if configured
+if(CONFIG_SMP AND CONFIG_MP_MAX_NUM_CPUS GREATER 1)
+  list(APPEND QEMU_SMP_FLAGS -smp cpus=${CONFIG_MP_MAX_NUM_CPUS},maxcpus=20)
```
michalsimek: Please update comment above that this parameter is aligned with QEMU DT already in the tree where you have 16 APUs + 4 RPUs. That would explain why number 20 is passed and the reason behind.
It means just append -smp maxcpus=20 which should be enough because core itself is adding -smp cpus= already.

michalsimek (changes requested):

boards/amd/versalnet_apu/board.cmake
kedareswararao: In case of Versal_net TF-A runs from DDR. If it needs to run from OCM, then some PDI changes are needed, To make it work for all possible designs location was changed will add comment for the same.

boards/amd/versalnet_apu/board.cmake
kedareswararao: Sure will add comment

michalsimek (approved):

nordicjm (approved):

PR #100642, opened 2025-12-08
mbox: mbox_ti_secproxy: Driver optimization and bugfixes

Kronosblaster:
Improves the TI secure proxy mailbox driver with better error handling, per-channel interrupt support, and thread cleanup.

1: Fix thread verification
  - Replaced busy-wait loop with immediate error returns (-EINVAL, -EBUSY, -ENODATA)
  - Removed redundant message count checks
  - Simplified ISR error handling

  2: Support per-channel interrupts
  - Changed from single shared interrupt to per-channel interrupt array with individual ISR handlers
  - ISR now processes specific channels directly instead of polling all channels
  - Moved device tree configuration to core-specific files (am64x_r5f0_0.dtsi)
  - Added interrupt-names requirement to device tree binding

  3: Add thread flush on channel enable
  - Added secproxy_mailbox_flush_thread() to drain pending messages from RX thread
  - Automatically flushes thread when enabling channel to ensure clean state
  - Prevents stale messages from previous sessions

  4: Add SPDX License Identifier

### DIFF
--- a/drivers/mbox/mbox_ti_secproxy.c
+++ b/drivers/mbox/mbox_ti_secproxy.c
@@ -79,7 +79,7 @@ struct secproxy_mailbox_config {
 	DEVICE_MMIO_NAMED_ROM(target_data);
 	DEVICE_MMIO_NAMED_ROM(rt);
 	DEVICE_MMIO_NAMED_ROM(scfg);
-	uint32_t irq;
+	int32_t interrupts[MAILBOX_MAX_CHANNELS];
 };
 
 static inline int secproxy_verify_thread(struct secproxy_thread *spt, uint8_t dir)
@@ -113,75 +113,74 @@ static inline int secproxy_verify_thread(struct secproxy_thread *spt, uint8_t di
 	return 0;
 }
 
-static void secproxy_mailbox_isr(const struct device *dev)
+static void secproxy_mailbox_isr(const struct device *dev, uint32_t channel)
 {
 	struct secproxy_mailbox_data *data = DEV_DATA(dev);
 	struct secproxy_thread spt;
 	uint32_t data_word;
 	mem_addr_t data_reg;
 
-	for (int i_channel = 0; i_channel < MAILBOX_MAX_CHANNELS; i_channel++) {
-		if (!data->channel_enable[i_channel]) {
-			continue;
+	if (!data->channel_enable[channel]) {
+		return;
+	}
+
+	spt.target_data = SEC_PROXY_THREAD(DEV_TDATA(dev), channel);
+	spt.rt = SEC_PROXY_THREAD(DEV_RT(dev), channel);
+	spt.scfg = SEC_PROXY_THREAD(DEV_SCFG(dev), channel);
+
+	if (secproxy_verify_thread(&spt, THREAD_IS_RX)) {
+		/* Silent failure */
+		return;
+	}
+
+	data_reg = spt.target_data + SEC_PROXY_DATA_START_OFFS;
+	size_t msg_len = MAILBOX_MBOX_SIZE;
+	size_t num_words = msg_len / sizeof(uint32_t);
+	size_t i;
+	struct rx_msg *rx_data = data->user_data[channel];
+
+	if (!rx_data || !rx_data->buf) {
+		LOG_ERR("No buffer provided for channel %d", channel);
+		return;
+	}
+
+	if (rx_data->size < MAILBOX_MBOX_SIZE) {
+		LOG_ERR("Buffer too small for channel %d", channel);
+		return;
+	}
+
+	uint8_t *buf = (uint8_t *)rx_data->buf;
+
+	/* Copy full words */
+	for (i = 0; i < num_words; i++) {
+		data_word = sys_read32(data_reg);
+		memcpy(&buf[i * 4], &data_word, sizeof(uint32_t));
+		data_reg += sizeof(uint32_t);
+	}
+
+	/* Handle trail bytes */
+	size_t trail_bytes = msg_len % sizeof(uint32_t);
+
+	if (trail_bytes) {
+		uint32_t data_trail = sys_read32(data_reg);
+
+		i = msg_len - trail_bytes;
+
+		while (trail_bytes--) {
+			buf[i++] = data_trail & 0xff;
+			data_trail >>= 8;
 		}
+	}
 
-		spt.target_data = SEC_PROXY_THREAD(DEV_TDATA(dev), i_channel);
-		spt.rt = SEC_PROXY_THREAD(DEV_RT(dev), i_channel);
-		spt.scfg = SEC_PROXY_THREAD(DEV_SCFG(dev), i_channel);
+	/* Ensure we read the last register if we haven't already */
+	if (data_reg <= (spt.target_data + SEC_PROXY_DATA_END_OFFS)) {
+		sys_read32(spt.target_data + SEC_PROXY_DATA_END_OFFS);
+	}
 
-		if (secproxy_verify_thread(&spt, THREAD_IS_RX)) {
-			continue;
-		}
-
-		data_reg = spt.target_data + SEC_PROXY_DATA_START_OFFS;
-		size_t msg_len = MAILBOX_MBOX_SIZE;
-		size_t num_words = msg_len / sizeof(uint32_t);
-		size_t i;
-		struct rx_msg *rx_data = data->user_data[i_channel];
-
-		if (!rx_data || !rx_data->buf) {
-			LOG_ERR("No buffer provided for channel %d", i_channel);
-			continue;
-		}
-
-		if (rx_data->size < MAILBOX_MBOX_SIZE) {
-			LOG_ERR("Buffer too small for channel %d", i_channel);
-			continue;
-		}
-
-		uint8_t *buf = (uint8_t *)rx_data->buf;
-
-		/* Copy full words */
-		for (i = 0; i < num_words; i++) {
-			data_word = sys_read32(data_reg);
-			memcpy(&buf[i * 4], &data_word, sizeof(uint32_t));
-			data_reg += sizeof(uint32_t);
-		}
-
-		/* Handle trail bytes */
-		size_t trail_bytes = msg_len % sizeof(uint32_t);
-
-		if (trail_bytes) {
-			uint32_t data_trail = sys_read32(data_reg);
-
-			i = msg_len - trail_bytes;
-
-			while (trail_bytes--) {
-				buf[i++] = data_trail & 0xff;
-				data_trail >>= 8;
-			}
-		}
-
-		/* Ensure we read the last register if we haven't already */
-		if (data_reg <= (spt.target_data + SEC_PROXY_DATA_END_OFFS)) {
-			sys_read32(spt.target_data + SEC_PROXY_DATA_END_OFFS);
-		}
-
-		rx_data->size = msg_len;
-		rx_data->seq = GET_MSG_SEQ(buf);
-		if (data->cb[i_channel]) {
-			data->cb[i_channel](dev, i_channel, data->user_data[i_channel], NULL);
-		}
+	rx_data->size = msg_len;
+	rx_data->seq = GET_MSG_SEQ(buf);
+	if (data->cb[channel]) {
+		data->cb[channel](dev, channel, data->user_data[channel], NULL);
 	}
 }
 
@@ -310,13 +309,18 @@ static int secproxy_mailbox_set_enabled(const struct device *dev, uint32_t chann
 		return -EALREADY;
 	}
 
+	if (cfg->interrupts[channel] < 0) {
+		LOG_ERR("No interrupt configured for channel %d", channel);
+		return -EINVAL;
+	}
+
 	key = k_spin_lock(&data->lock);
 	data->channel_enable[channel] = enable;
 
 	if (enable) {
-		irq_enable(cfg->irq);
+		irq_enable(cfg->interrupts[channel]);
 	} else {
-		irq_disable(cfg->irq);
+		irq_disable(cfg->interrupts[channel]);
 	}
 
 	k_spin_unlock(&data->lock, key);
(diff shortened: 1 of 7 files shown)


drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -104,18 +107,13 @@ static inline int secproxy_verify_thread(struct secproxy_thread *spt, uint8_t di
...
-		k_busy_wait(poll_interval_ms);
-		waited_ms += poll_interval_ms;
-		if (waited_ms >= timeout_ms) {
-			LOG_ERR("Timeout waiting for thread to %s\n",
-				(dir == THREAD_IS_TX) ? "empty" : "fill");
-			return -ETIMEDOUT;
+	if (dir == THREAD_IS_TX) {
+		if ((status & RT_THREAD_STATUS_CUR_CNT_MASK) == (SECPROXY_MAILBOX_NUM_MSGS)) {
+			return -EBUSY;
+		}
```
natto1784: one query - for an outbound channel, this value means the number of free messages that can be written, is that correct? If so, wouldn't any value above 0 be valid in such a case?

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -233,16 +220,9 @@ static int secproxy_mailbox_send(const struct device *dev, uint32_t channel,
 	spt.target_data = SEC_PROXY_THREAD(DEV_TDATA(dev), channel);
 	spt.rt = SEC_PROXY_THREAD(DEV_RT(dev), channel);
 	spt.scfg = SEC_PROXY_THREAD(DEV_SCFG(dev), channel);
+	spt.channel = channel;
 
 	if (secproxy_verify_thread(&spt, THREAD_IS_TX)) {
```
natto1784: nitpick: can use `< 0` here for verbosity

drivers/mbox/mbox_ti_secproxy.c
natto1784: I understand that is correct, however in case of TX, `0` would mean that no new messages can be transferred, right?

Kronosblaster: updated

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -104,18 +106,13 @@ static inline int secproxy_verify_thread(struct secproxy_thread *spt, uint8_t di
...
-			return -ETIMEDOUT;
+	if (dir == THREAD_IS_TX) {
+		if ((status & RT_THREAD_STATUS_CUR_CNT_MASK) == 0) {
+			return -EBUSY;
+		}
+	} else {
+		if ((status & RT_THREAD_STATUS_CUR_CNT_MASK) == 0) {
+			return -ENODATA;
 		}
 	}
```
natto1784: conditions can be reversed i.e, CUR_CNT check can go outside DIR check

drivers/mbox/mbox_ti_secproxy.c
natto1784: better yet, return the value returned by `secproxy_verify_thread` if not zero instead of `-EBUSY`

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -359,25 +348,52 @@ static DEVICE_API(mbox, secproxy_mailbox_driver_api) = {
...
-		DEVICE_MMIO_NAMED_MAP(dev, scfg, K_MEM_CACHE_NONE);                                \
-		IRQ_CONNECT(DT_INST_IRQN(idx), DT_INST_IRQ(idx, priority), secproxy_mailbox_isr,   \
-			    DEVICE_DT_INST_GET(idx),                                               \
-			    COND_CODE_1(DT_INST_IRQ_HAS_CELL(idx, flags),			\
-				(DT_INST_IRQ(idx, flags)), (0)));       \
-		return 0;                                                                          \
-	}                                                                                          \
+#define SECPROXY_THREAD_ISR(i, idx)                                                                \
+	COND_CODE_1(DT_INST_IRQ_HAS_NAME(idx, DT_CAT(rx_, i)),                                     \
+		(	\
```
natto1784: Move all backslashes in macros to column 99 to avoid formatting errors. clang-format won't fix a lot of these on its own.

Kronosblaster: done

drivers/mbox/mbox_ti_secproxy.c
Kronosblaster: done

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -359,28 +351,54 @@ static DEVICE_API(mbox, secproxy_mailbox_driver_api) = {
...
+				secproxy_mailbox_isr(dev, i);                                     \
+			}                                                                         \
+		),                                                                                \
+		())
+
+/* Generate IRQ number or 0 for array initialization */
+#define SECPROXY_IRQ_OR_ZERO(i, inst)                                                             \
+	COND_CODE_1(DT_INST_IRQ_HAS_NAME(inst, DT_CAT(rx_, i)),                                   \
+		(DT_INST_IRQ_BY_NAME(inst, DT_CAT(rx_, i), irq)),                                 \
+		(0))
```
natto1784: is it possible to have another placeholder - like a negative value? since 0 is a valid IRQ.

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -38,15 +38,12 @@ LOG_MODULE_REGISTER(ti_secure_proxy);
...
-#define SECPROXY_MAILBOX_NUM_MSGS 5
-#define MAILBOX_MAX_CHANNELS      32
-#define MAILBOX_MBOX_SIZE         60
+#define MAILBOX_MAX_CHANNELS 32
+#define MAILBOX_MBOX_SIZE    60
 
 #define SEC_PROXY_DATA_START_OFFS 0x4
 #define SEC_PROXY_DATA_END_OFFS   0x3c
 
-#define SEC_PROXY_TIMEOUT_US 1000000
```
glneo, line 49: Why remove the busy wait? I'm not saying I'm opposed to this change, but you need to explain it.

And also it would be good to split that change into its own patch. You really should try to do that for each of the bullet points in your commit messages, each commit should do one single logical thing, having a list of different changes in a commit message is a sign that you are doing too much in one commit.

natto1784: Hi, the rationale behind this change was that the timer is not initialized until `PRE_KERNEL_2`, hence it can cause issues to use busy_wait here. Alternatively, we can also add a check to see whether we are pre_kernel_* stages or not, but we concluded that this busy_wait is not necessary either way.

Kronosblaster: The commits have been split up.

glneo: > but we concluded that this busy_wait is not necessary either way.

How was that concluded? Looking at other implementations of this TI-SCI driver like the one in OP-TEE and TF-A, I see timeout waits.

You also do not explain it in the new split up commit either, the commit message just has: 

> Update secproxy_verify_thread to remove busy wait.

I can read as much in the code itself, commit messages should tell the "why" something is being changed. I think I've given this feedback before to you, please start working on your commit messages.

natto1784: The idea was that since every core has its own dedicated channel for communication on the mailbox instance, a simple check to see whether its empty (for RX) or full (for TX) should work in theory.

However, I do understand now that this is not ideal and retract my conclusion. The busy wait can be kept here, but should go inside a `!k_is_pre_kernel()` check, because we cannot really rely on timers such as dmtimer since it is not necessarily initialized at this point. This is due to how `k_busy_wait` is implemented - speaking of which, maybe we can do `nop`s to busy wait if dmtimer is not initialized.

Kronosblaster: Also U-boot, Linux and MCU+SDK dont have a busy wait

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -297,8 +296,8 @@ static uint32_t secproxy_mailbox_max_channels_get(const struct device *dev)
 
 static int secproxy_mailbox_set_enabled(const struct device *dev, uint32_t channel, bool enable)
 {
-	const struct secproxy_mailbox_config *cfg = DEV_CFG(dev);
 	struct secproxy_mailbox_data *data = DEV_DATA(dev);
+	const struct secproxy_mailbox_config *config = DEV_CFG(dev);
```
glneo: You are making small un-related changes like this where you move the definition down one line, and change the variable name (cfg -> config). I don't have an issue if you think this variable should be named differently but it should go in its own commit. Your goal should be to make each patch as simple and logically granular as possible.

drivers/mbox/mbox_ti_secproxy.c
glneo: I'm not interested in what other implementations do, some have the wait, some do not, what I care about is what is correct. We had the polling timeout wait before, but you are removing it without an explanation beyond you finding some other implementations without this feature.

Kronosblaster: Updated commit message to include rationale

drivers/mbox/mbox_ti_secproxy.c
Kronosblaster: updated

drivers/mbox/mbox_ti_secproxy.c
glneo: I don't see any update to the commit message with any rationale. Seems you have just repeated that found some implementations of the same without the wait. I can find implementations with the wait (OP-TEE/TF-A) so that does tell us which way is correct and why.

drivers/mbox/mbox_ti_secproxy.c
glneo: Still not right, here you update the variable name but not any users of the variable. This doesn't even compile now. It does get fixed in the next patch, but that means build is broken in-between which breaks bisectability. 

Either change all the users here, or just drop this patch and don't worry about changing the name.

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -295,6 +295,21 @@ static uint32_t secproxy_mailbox_max_channels_get(const struct device *dev)
...
 
+static void secproxy_mailbox_flush_thread(const struct device *dev, uint32_t channel)
+{
+	struct secproxy_thread spt;
+
+	spt.target_data = SEC_PROXY_THREAD(DEV_TDATA(dev), channel);
+	spt.rt = SEC_PROXY_THREAD(DEV_RT(dev), channel);
+	spt.scfg = SEC_PROXY_THREAD(DEV_SCFG(dev), channel);
+
+	/* Drain all pending messages from the thread. Blocking call */
```
glneo: Blocking call where? Feels like this comment was copied from somewhere and doesn't apply here.

Kronosblaster: You're right. Updated the comment to be clearer

drivers/mbox/mbox_ti_secproxy.c
Kronosblaster: b8e33f145e1cfc0d0e598d234e6cf1d5c1f5e2d9 and the other commit messages have been updated for clarity.

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -105,18 +103,11 @@ static inline int secproxy_verify_thread(struct secproxy_thread *spt, uint8_t di
 		return -EINVAL;
 	}
 
-	/* Check the message queue before sending/receiving data */
-	int timeout_ms = SEC_PROXY_TIMEOUT_US;
```
glneo, line 109: If what you want to have no delay before timing out, why not flip the below loop to check if timeout has passed first. Then you can set `SEC_PROXY_TIMEOUT_US` in the case you want this to timeout instantly.

Your commit message states that "retry logic should be handled by the calling layer" but I don't see any retry logic anywhere in the calling path of this function nor in any example using mailbox.

Kronosblaster: This check shouldn't be done in this layer at all. As a peripheral driver it should just return the appropriate error message and be done with it. It is much cleaner to return these error messages. Additionally k_busy_wait shouldn't be called pre-kernel where this driver should be functional. Adding a pre-kernel check could be done but that is adding additional complexity where it is not required.

You're right about the commit message and I have updated the commit message to clarify that retry messages can be optionally implemented if required. This layer is only used by TISCI. The retry logic is an improvement for the TISCI driver and will be done in another patch if required. Not currently pushing that as this PR is a dependency for a few other PRs which have been blocked for a while.

Thanks

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -338,13 +324,19 @@ static int secproxy_mailbox_set_enabled(const struct device *dev, uint32_t chann
 		return -EALREADY;
 	}
 
+	if (cfg->interrupts[channel] < 0) {
```
natto1784, line 327: always false

drivers/mbox/mbox_ti_secproxy.c
```diff
@@ -83,20 +79,21 @@ struct secproxy_mailbox_config {
 	DEVICE_MMIO_NAMED_ROM(target_data);
 	DEVICE_MMIO_NAMED_ROM(rt);
 	DEVICE_MMIO_NAMED_ROM(scfg);
-	uint32_t irq;
+	uint32_t interrupts[MAILBOX_MAX_CHANNELS];
```
natto1784:
```suggestion
	int32_t interrupts[MAILBOX_MAX_CHANNELS];
```

drivers/mbox/mbox_ti_secproxy.c
Kronosblaster: SECPROXY_IRQ_OR_INVALID might set it as -1

drivers/mbox/mbox_ti_secproxy.c
glneo: It's easy to say it shouldn't be done in this layer to make this layer less complex, but all you would be doing is pushing that complexity higher in the stack, and potentiality into every user of this mailbox. Meaning more complexity in more spots..

Since that higher layer will likely only be TI-SCI, I won't hold this series up on this, but with the understanding that should this need to be done you will need to be the one to add the retry logic to TI-SCI layer.

glneo (approved):

natto1784 (approved):

Issue #78724, opened 2024-09-19
net: checksum offloading not working for virtual network interfaces used for Virtual LAN (VLAN)


stomschik-ask:
**Describe the bug**
The checksum offloading capabilities of an Ethernet interface are not respected for virtual network interfaces as the function `need_calc_checksum()` in "net_if.c" always returns true for the virtual interfaces. This increases CPU utilization as checksums are calculated in software even though they would be automatically calculated by the Ethernet MAC. Even worse, some Ethernet MACs require the checksum to be cleared to zero for TX checksum offloading as otherwise an incorrect checksum will be generated by the Ethernet MAC (for example [NXP MIMXRT1060-EVK](https://docs.zephyrproject.org/latest/boards/nxp/mimxrt1060_evk/doc/index.html)).

**To Reproduce**
Use the [Virtual LAN](https://docs.zephyrproject.org/latest/samples/net/vlan/README.html) sample
1. disable IPv6 as IPv6 will disable TX/RX checksum offloading on i.MX RT1060 (`CONFIG_NET_IPV6=n`)
2. west build --pristine -b mimxrt1060_evk samples/net/vlan
3. west flash
4. From PC, ping 192.0.2.1 (untagged, working)
5. From PC, ping 198.51.100.1 (VLAN 100, request timed out)
6. Incorrect checksum (0xffff) for IP header and ICMP message will be shown in Wireshark for the Echo reply

**Expected behavior**
TX/RX checksum offloading should also work on virtual network interfaces used for Virtual LAN (VLAN).

**Impact**
TX/RX checksum offloading can be disabled for i.MX RT1060 by setting `CONFIG_ETH_NXP_ENET_HW_ACCELERATION=n`. Therefore, impact is mainly the increased CPU utilization.

**Environment:**
 - OS: Windows
 - Zephyr SDK 0.16.8
 - Zephyr 3.7.0

jukkar:
@stomschik-ask I have a fix proposal at # #78745. Would you be able to check if it works for you?

stomschik-ask:
Thanks, the fix is working fine for me.

PR #73985, opened 2024-06-10
net: Support partial checksum offloading

WorldofJARcraft:
The Xilinx AXI Ethernet subsystem is capable of RX/Tx checksum offloading. While it supports computing IP and UDP/TCP checksums, it does not support computing ICMP checksums and only computes IP checksums for ICMP messages. Thus, this patch adds an default-disabled option for always computing the ICMP checksum in software while relying on hardware to compute the remaining checksums.

PR #100653, opened 2025-12-08
openthread: platform: infra_if: Remove NAT64 packet checksums

Cristib05:
OpenThread NAT64 translator creates a packet that has header checksum computed and appended.
In case of using ethernet as backbone interface, some NICs may want to have checksum set to 0 for a correct computation.
This commit aims to address this issue by checking if hardware has offload capabilities and setting NAT64 packet checksums to 0, if needed.

Might be related to:
https://github.com/zephyrproject-rtos/zephyr/issues/78724
https://github.com/zephyrproject-rtos/zephyr/pull/73985

### DIFF
--- a/modules/openthread/platform/infra_if.c
+++ b/modules/openthread/platform/infra_if.c
@@ -44,6 +44,7 @@ static void handle_ra_from_ot(const uint8_t *buffer, uint16_t buffer_length);
 
 static struct zsock_pollfd sockfd_raw[MAX_SERVICES];
 static void raw_receive_handler(struct net_socket_service_event *evt);
+static void remove_checksums_for_eth_offloading(uint8_t *buf, uint16_t len);
 static int raw_infra_if_sock = -1;
 
 NET_SOCKET_SERVICE_SYNC_DEFINE_STATIC(handle_infra_if_raw_recv, raw_receive_handler, MAX_SERVICES);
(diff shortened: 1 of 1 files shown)


modules/openthread/platform/infra_if.c
```diff
@@ -360,10 +361,60 @@ otError infra_if_send_raw_message(uint8_t *buf, uint16_t len)
...
+	}
+
+	if (net_eth_get_hw_config(ail_iface_ptr, ETHERNET_CONFIG_TYPE_TX_CHECKSUM_SUPPORT,
+				  &config) != 0) {
+		return; /* No TX checksum capabilities*/
+	}
+
+	pkt_cursor = buf + (ipv4_hdr->vhl & 0x0F) * 4;
+
+	if (config.chksum_support & NET_IF_CHECKSUM_IPV4_HEADER) {
```
pdgendt: No implicit boolean conversions, please

```suggestion
	if ((config.chksum_support & NET_IF_CHECKSUM_IPV4_HEADER) != 0U) {
```

Cristib05: done, thanks!

rlubos (dismissed):

rlubos (approved):

modules/openthread/platform/infra_if.c
```diff
@@ -360,10 +361,60 @@ otError infra_if_send_raw_message(uint8_t *buf, uint16_t len)
...
 exit:
 	return error;
 }
+
+static void remove_checksums_for_eth_offloading(uint8_t *buf, uint16_t len)
+{
+	struct net_ipv4_hdr *ipv4_hdr = (struct net_ipv4_hdr *)buf;
+	uint8_t *pkt_cursor = NULL;
+	struct ethernet_config config;
+
```
pdgendt, line 378: `len` does not appear to be used? Maybe add an assert for a minimum length or checks depending on the proto type?

Cristib05: Most likely, length parameter should not be passed to this function. Given that NAT64 packets are created by OpenThread stack, a check should be performed when NAT64 callback is invoked and an otMessage is received.
I expect that OpenThread's NAT64 module is performing a message check, but I will double check that. If it is performed at OpenThread level, then I would say nothing is to be done at border router application level (net/l2/openthread or platform/infra_if).

pdgendt: But you do inspect values in `buf`, why not validate that `len` is large enough and bail out if that's not the case?

Cristib05: It can be done, but it seems that is also done at OT level. I can add it, just in case OT implementation might change, but I don't think this will ever happen.
https://github.com/zephyrproject-rtos/openthread/blob/main/src/core/net/nat64_translator.cpp#L124

pdgendt (approved):

PR #100655, opened 2025-12-08
drivers: clock_control: stm32: enable fractional divider for PLL2/3

tpambor:
Configure the fractional divider (FRACN) for PLL2 and PLL3 on STM32H5 series when specified in the device tree.


erwango (approved):

mathieuchopstm (approved):

FRASTM (approved):

etienne-lms (commented):
This change only targets only STM32H5xx devices. Could you update  the commit message accordingly, at least the hader line, for example:
```patch
-drivers: clock_control: stm32: enable fractional divider for PLL2/3
+drivers: clock_control: stm32: enable H5 fractional divider for PLL2/3

PR #100657, opened 2025-12-08
doc: bluetooth: mesh: Update supported Bluetooth Mesh Protocol version

PavelVPV:
Update Bluetooth Mesh Protocol supported version to 1.1.1.


alxelax (approved):

omkar3141 (approved):

jhedberg (approved):

Thalley (approved):
I'm wondering what the best approach to denote supported versions of BT specs. Currently we just have a table in the documentation for LE Audio as well, but maybe the supported version should be in Kconfig or the public API? Or maybe the documentation like this is the best place?

PR #100660, opened 2025-12-08
arm: mpu: ArmV8: track regions when MPU gap filling is disabled

### LINKED ISSUE #98606: Potential MPU Region Waste When CONFIG_MPU_GAP_FILLING=n
## Description

I've observed what appears to be an inefficiency in the ARMv8-M MPU region allocation logic when `CONFIG_MPU_GAP_FILLING` is disabled.

## Environment

- **Board**: `rd_rw612_bga/rw612` (NXP RW612)
- **SoC**: ARM Cortex-M33 (ARMv8-M Mainline)
- **Zephyr Version**: 4.2.0
- **Configuration**:
  - `CONFIG_USERSPACE=y`
  - `CONFIG_ARM_MPU=y`
  - `CONFIG_FLASH=y`
  - `CONFIG_MPU_GAP_FILLING=n`

## Observation

When running with the above configuration, the MPU regions are:

```
Reg | Status   | Base Addr  | Limit Addr | Perms     | XN  | Share | Type
-------------------------------------------------------------------------
 0  | ENABLED  | 0x18000000 | 0x1BFFFFFF | P_RO_U_RO | No  | Non   | Norm
 1  | ENABLED  | 0x10040000 | 0x10041C9F | P_RO_U_RO | No  | Non   | Norm
 2  | ENABLED  | 0x41300000 | 0x413FF7FF | P_RW_U_NA | Yes | Non   | Norm
 3  | ENABLED  | 0x44300000 | 0x4432E2EF | P_RW_U_NA | Yes | Non   | Norm
 4  | ENABLED  | 0x10041CA0 | 0x10042C7F | P_RO_U_RO | No  | Non   | Norm
 5  | DISABLED | 0x00000000 | 0x00000000 | -         | -   | -     | -
 6  | ENABLED  | 0x100429C0 | 0x100429FF | P_RW_U_RW | Yes | Non   | Norm
 7  | DISABLED | 0x00000000 | 0x00000000 | -         | -   | -     | -
```

0-4 correspond to static regions: flash, SRAM (code relocation), SMU1, SMU2, SRAM (ramfunc).
Region 6 is the dynamic memory region (libc partition).

**Region 5 is disabled**, while regions 0-4 and 6 are enabled. This seems wasteful - why wouldn't region 5 be used before region 6?

## Code Analysis

Looking at `arch/arm/core/mpu/arm_mpu_v8_internal.h`, lines 777-785:

```c
/* We are going to skip the full partition of the background areas.
 * So we can disable MPU regions inside which dynamic memory regions
 * may be programmed.
 */
for (int i = 0; i < MPU_DYNAMIC_REGION_AREAS_NUM; i++) {
    mpu_clear_region(dyn_reg_info[i].index);  // Clears region 5
}

/* The dynamic regions are now programmed on top of
 * existing SRAM region configuration.
 */
mpu_reg_index = mpu_configure_regions(dynamic_regions,
    regions_num, mpu_reg_index, true);  // Starts at static_regions_num = 6
```

The logic appears to be:

1. Region 5 is initially the background SRAM region (part of static regions)
2. `dyn_reg_info[0].index = 5` marks it as the region for dynamic allocation
3. Region 5 gets **cleared** (disabled) via `mpu_clear_region()`
4. Dynamic regions then allocate starting from `mpu_reg_index = static_regions_num = 6`

**Result**: Region 5 is disabled but never reused, wasting one of the 8 available MPU regions.

Is this intended behavior? If so, what's the reasoning behind:

- Clearing region 5 but not reusing it for dynamic allocation?
- Starting dynamic allocation at region 6 instead of region 5?
- On systems with only 8 MPU regions every region is valuable.

I may be missing some subtle requirement that necessitates this design, so I wanted to confirm before proposing a fix.

## Expected Behavior
Clear region 5 and **reuse it** as the first dynamic region. This would allow dynamic regions to use regions 5-7 instead of just 6-7, saving an MPU region.

## Additional comments

When `CONFIG_MPU_GAP_FILLING=y`:
- Region 5 is reprogrammed (not just cleared) and reused as part of the partitioning logic

When `CONFIG_MPU_GAP_FILLING=n`:
- Region 5 is cleared but the allocation starts at region 6, leaving region 5 unused

I can propose a solution and then prepare a PR if this is confirmed to be a bug.

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

wight-u:
Optimizes MPU region usage on Armv8 when CONFIG_MPU_GAP_FILLING is disabled. Previously, available regions were not efficiently tracked, leading to waste. This change implements bitmask tracking to identify free indices, allowing dynamic regions to reuse slots released from static configuration.

Fixes #98606

### DIFF
--- a/arch/arm/core/mpu/arm_mpu.c
+++ b/arch/arm/core/mpu/arm_mpu.c
@@ -201,45 +201,6 @@ static int mpu_configure_region(const uint8_t index,
 		(const struct arm_mpu_region *)&region_conf);
 }
 
-#if !defined(CONFIG_MPU_REQUIRES_NON_OVERLAPPING_REGIONS) || \
-	!defined(CONFIG_MPU_GAP_FILLING)
-/* This internal function programs a set of given MPU regions
- * over a background memory area, optionally performing a
- * sanity check of the memory regions to be programmed.
- */
-static int mpu_configure_regions(const struct z_arm_mpu_partition
-	regions[], uint8_t regions_num, uint8_t start_reg_index,
-	bool do_sanity_check)
-{
-	int i;
-	int reg_index = start_reg_index;
-
-	for (i = 0; i < regions_num; i++) {
-		if (regions[i].size == 0U) {
-			continue;
-		}
-		/* Non-empty region. */
-
-		if (do_sanity_check &&
-				(!mpu_partition_is_valid(&regions[i]))) {
-			LOG_ERR("Partition %u: sanity check failed.", i);
-			return -EINVAL;
-		}
-
-		reg_index = mpu_configure_region(reg_index, &regions[i]);
-
-		if (reg_index == -EINVAL) {
-			return reg_index;
-		}
-
-		/* Increment number of programmed MPU indices. */
-		reg_index++;
-	}
-
-	return reg_index;
-}
-#endif
-
 /* ARM Core MPU Driver API Implementation for ARM MPU */
 
 
--- a/arch/arm/core/mpu/arm_mpu_v7_internal.h
+++ b/arch/arm/core/mpu/arm_mpu_v7_internal.h
@@ -202,12 +202,41 @@ static inline int mpu_buffer_validate(const void *addr, size_t size, int write)
 
 #endif /* CONFIG_USERSPACE */
 
-static int mpu_configure_region(const uint8_t index,
-	const struct z_arm_mpu_partition *new_region);
+/* This internal function programs a set of given MPU regions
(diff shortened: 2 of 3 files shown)


wearyzen (commented):
Hi @wight-u , thank you for the fixing this. The bitmap approach looks good but I think with this we don't need `static_regions_num` anymore. Could you remove it as well?

PR #100665, opened 2025-12-08
boards: nordic: Add bias-pull-up to UART TX pin sleep state

nordic-krch:
When UART TX pin is in sleep state it should have pull up. So far it was floating and that could lead to garbage output on terminal.


nika-nordic (dismissed):

bjarki-andreasen (dismissed):

nordicjm (dismissed):

nordic-krch:
@nika-nordic @bjarki-andreasen @nordicjm can you look again? Had to fix compliance check.

nordicjm (approved):

bjarki-andreasen (approved):

ubieda (approved):

PR #100671, opened 2025-12-08
samples: add a ec_host_cmd_usb sample

fabiobaltieri:
Add a samples implementing some basic EC HC command using the EC HC over USB subsystem.


fabiobaltieri:
> In terms of adding the sample right now. I would wait till a problem with kernel crash is fixed. I don't think we are in rush to add such sample. Unless you have different opinion.

Yeah let's wait, mainly wanted to do a brain dump before I forget everything.

PR #100672, opened 2025-12-08
drivers: display: st: allow-ltdc-callback-override

markussmST:
Allow overriding the LTDC callback for manual handling or extension. In some cases, the default handling may not be desired, for example if a line interrrupt has been set up to do an operation after a certain line has been passed, or if the user wants to use the HAL IRQ handler.

### DIFF
--- a/drivers/display/display_stm32_ltdc.c
+++ b/drivers/display/display_stm32_ltdc.c
@@ -17,6 +17,7 @@
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/drivers/clock_control/stm32_clock_control.h>
 #include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/display/display_stm32_ltdc.h>
 #include <zephyr/drivers/reset.h>
 #include <zephyr/linker/devicetree_regions.h>
 #include <zephyr/pm/device.h>
@@ -89,6 +90,16 @@ static void stm32_ltdc_global_isr(const struct device *dev)
 {
 	struct display_stm32_ltdc_data *data = dev->data;
 
+	/*
+	 * Allow handling or extending the ISR with custom code in weak function.
+	 * An example of this could be to call the HAL LTDC IRQ handler.
+	 * Return 0 to continue with the default handler or 1 if the interrupt
+	 * has been fully handled by the custom handler.
+	 */
+	if (ltdc_isr_custom_handler(&data->hltdc) == 1) {
+		return;
+	}
+
 	if (__HAL_LTDC_GET_FLAG(&data->hltdc, LTDC_FLAG_LI) &&
 	    __HAL_LTDC_GET_IT_SOURCE(&data->hltdc, LTDC_IT_LI)) {
 		if (data->front_buf != data->pend_buf) {
@@ -102,6 +113,12 @@ static void stm32_ltdc_global_isr(const struct device *dev)
 
 		__HAL_LTDC_CLEAR_FLAG(&data->hltdc, LTDC_FLAG_LI);
 	}
+}
+
+__weak int ltdc_isr_custom_handler(LTDC_HandleTypeDef *hltdc)
+{
+	ARG_UNUSED(hltdc);
+	return 0;
 }
 
 static int stm32_ltdc_set_pixel_format(const struct device *dev,
--- a/include/zephyr/drivers/display/display_stm32_ltdc.h
+++ b/include/zephyr/drivers/display/display_stm32_ltdc.h
@@ -1,0 +1,13 @@
+/*
+ * Copyright (c) 2025 STMicroelectronics.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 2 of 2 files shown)


include/zephyr/drivers/display/display_stm32_ltdc.h
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2025 STMicroelectronics.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DISPLAY_STM32_LTDC_H_
+#define ZEPHYR_INCLUDE_DISPLAY_STM32_LTDC_H_
```
etienne-lms, line 8:
```suggestion
#ifndef ZEPHYR_INCLUDE_DRIVERS_DISPLAY_DISPLAY_STM32_LTDC_H_
#define ZEPHYR_INCLUDE_DRIVERS_DISPLAY_DISPLAY_STM32_LTDC_H_
```
to match the header file path.

include/zephyr/drivers/display/display_stm32_ltdc.h
```diff
@@ -0,0 +1,13 @@
...
+
+/* Prototypes of external ISR handler weak functions */
+int ltdc_isr_custom_handler(LTDC_HandleTypeDef *hltdc);
```
etienne-lms, line 11: If this is STM32 specific, it should be prefixed, e.g. `ltdc_stm32_isr_custom_handler()`.

drivers/display/display_stm32_ltdc.c
```diff
@@ -89,6 +90,16 @@ static void stm32_ltdc_global_isr(const struct device *dev)
 {
 	struct display_stm32_ltdc_data *data = dev->data;
 
+	/*
+	 * Allow handling or extending the ISR with custom code in weak function.
+	 * An example of this could be to call the HAL LTDC IRQ handler.
+	 * Return 0 to continue with the default handler or 1 if the interrupt
+	 * has been fully handled by the custom handler.
```
etienne-lms, line 97: This description should be moved into the header file where the function is declared.

drivers/display/display_stm32_ltdc.c
```diff
@@ -89,6 +90,16 @@ static void stm32_ltdc_global_isr(const struct device *dev)
...
+	 */
+	if (ltdc_isr_custom_handler(&data->hltdc) == 1) {
```
JarmouniA, line 99: We don't really want a driver calling a user function.
Instead, use a semaphore to inform a user thread that it can run whatever it wants when the interrupt occurs.

JarmouniA (changes requested):

drivers/display/display_stm32_ltdc.c
markussmST: There are two problems with that approach, in my opinion;
1) If you want to make sure that whatever you want to do takes place before the buffer change, you need the use thread to block the driver thread, and then you're kind of back to having all the same problems with more complexity.
2) There are cases where a line interrupt could be set by the user, for example to wait for a certain line to be drawn before rendering an element, where it is not desirable to update CFBAR like the current ISR in the driver does.

I'm not necessarily dead set on this solution, and maybe in the end the best option is what erwango describes, so it's at least a conscious choice to have the function, or to make it possible to let the ST HAL IRQ handler handle the interrupt instead of the driver via a compile time switch, so the user can implement everything manually.

JarmouniA: FYI Display API is still unstable (https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html#unstable), so we have room to make big changes.

Also, keep in mind that we do not want the app to bypass the API and change the driver's behavior directly.

erwango: @JarmouniA I think the need we have may apply to others as well.

One reason to be able to handle the callback externally is to optimize applications using a single framebuffer. In this case, you may want to render an element into the buffer at a certain position. Instead of waiting for the entire LTDC refresh to finish, you can set a line interrupt to wait for the LTDC to pass the last line of the element in question, and then render the element into the framebuffer to be displayed at the next refresh.

I think API should allow such optimized behavior, but we don't see much how to do it other than handling the callback externally. Would you have an alternative proposal ?

JarmouniA: @erwango Let me look into it and get back with a proposal.

erwango: @JarmouniA Did you had a chance to look at this issue? We'd need it for a better integration of an alternative GFX subystem (alternative to lvgl)

JarmouniA: @erwango Working on it (in my limited free time) sorry for the delay :) Will be able to push something this week.

JarmouniA: @markussmST Please take a look at #102822.

markussmST: Thanks, that looks like it would work. I'll adapt this PR once I get some Zephyr-time :)

PR #100676, opened 2025-12-08
drivers: counter: mcux_lptmr: use guard for late alarms

Holt-Sun:
Use guard period in set_alarm() to detect late settings and return -ETIME. Allow relative ticks > top and ticks==0; only reject absolute ticks > top.
Pend IRQ on late when requested.
Add irqn from DT for pending IRQ.

### DIFF
--- a/drivers/counter/counter_mcux_lptmr.c
+++ b/drivers/counter/counter_mcux_lptmr.c
@@ -24,6 +24,7 @@ struct mcux_lptmr_config {
 	lptmr_pin_select_t pin;
 	lptmr_pin_polarity_t polarity;
 	void (*irq_config_func)(const struct device *dev);
+	unsigned int irqn;
 };
 
 struct mcux_lptmr_data {
@@ -32,6 +33,7 @@ struct mcux_lptmr_data {
 	void *alarm_user_data;
 	bool alarm_active;
 	struct k_spinlock lock;
+	uint32_t guard_period;
 #else
 	counter_top_callback_t top_callback;
 	void *top_user_data;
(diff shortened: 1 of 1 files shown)


ZhaoxiangJin (dismissed):

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -97,22 +136,42 @@ static int mcux_lptmr_set_alarm(const struct device *dev, uint8_t chan_id,
 
 	k_spin_unlock(&data->lock, key);
 
-	if (config->base->CSR & LPTMR_CSR_TEN_MASK) {
-		/* Already enabled, first stop then set period (HW constraint). */
-		LPTMR_StopTimer(config->base);
-		LPTMR_SetTimerPeriod(config->base, alarm_cfg->ticks);
+	current = LPTMR_GetCurrentTimerCount(config->base);
```
zejiang0jason, line 139: should `current` always be 0 when `CONFIG_COUNTER_MCUX_LPTMR_ALARM` is  used?

drivers/counter/counter_mcux_lptmr.c
```diff
@@ -97,22 +136,42 @@ static int mcux_lptmr_set_alarm(const struct device *dev, uint8_t chan_id,
...
+	max_rel_val = (current + data->guard_period) % config->info.max_top_value;
+	ticks = alarm_cfg->ticks;
+	/* 'absolute' already computed for validation above */
+
+	if (!absolute) {
+		/* Late detection heuristic for relative alarms. */
+		irq_on_late = ticks < (config->info.max_top_value / 2U);
```
zejiang0jason, line 146: why `ticks < (config->info.max_top_value / 2U);` means irq_on_late?

Holt-Sun: Because it’s a wrap-around counter, “short” relative delays are unambiguous only if they’re less than half the counter range. The half-range threshold is the standard way to avoid ambiguity on modulo counters (unique shortest distance) which are popular in other counter drivers too.

drivers/counter/counter_mcux_lptmr.c
ZhaoxiangJin: always 0 I believe. when an alarm fires, the interrupt occurs, then the LPTMR gets disabled and the CNR register gets reset, which means the current value becomes 0 the next time set_alarm is called.

Holt-Sun:
Hi @nordic-krch , please review, thank you.

Holt-Sun:

mmahadevan108:
@Holt-Sun please rebase.

What was the issue you were seeing that required this change.

Holt-Sun:
This PR is totally out of date, close it.
The task will be done in another PR.

PR #100682, opened 2025-12-08
ipc: intel_adsp: rework host IPC service backend and add critical send support

tmleman:
This pull request extends the generic IPC service API with a function to support sending critical messages/notifications and refactors the Intel Audio DSP host IPC service backend.

Summary of changes:

- ipc_service: Add ipc_service_send_critical() API and a matching send_critical() backend hook to send high-priority messages via a dedicated fast path that bypasses normal busy checks.
- ipc: intel_adsp: Simplify the host IPC service backend by replacing the custom intel_adsp_ipc_msg descriptor and enum-based length encoding with a fixed two-word (header + payload) IPC payload used consistently in send and receive paths.
- intel_adsp: Wire the emergency/critical send path through the new ipc_service_send_critical() API while keeping the normal IPC flow unchanged.
- intel_adsp: Move the synchronous IPC wait logic into the common intel_adsp_ipc_send_message_sync() helper used by tests, and remove the now-redundant backend-specific sync helper.
- intel_adsp: Update and clarify backend documentation comments to describe the new data format, the role of ipc_ept_cfg::priv, and the semantics of the TX-buffer readiness and RX-buffer release helpers.

Status / TODO:

- ~~Existing intel_adsp IPC tests have not yet been adapted to the refactored backend and will currently fail. Test updates will be added as part of this PR before it is ready to merge.~~

### DIFF
--- a/include/zephyr/ipc/ipc_service.h
+++ b/include/zephyr/ipc/ipc_service.h
@@ -315,6 +315,36 @@ int ipc_service_deregister_endpoint(struct ipc_ept *ept);
  */
 int ipc_service_send(struct ipc_ept *ept, const void *data, size_t len);
 
+/** @brief Send a high-priority message bypassing normal state checks.
+ *
+ *  This function sends a high-priority message using the backend in a
+ *  special fast path that skips normal state and busy checking. It is
+ *  intended for critical system notifications such as crash reports,
+ *  fatal errors, or emergency shutdown signals that must be delivered
+ *  even when the IPC channel is otherwise considered busy or blocked.
+ *
+ *  WARNING: This function should only be used for critical system notifications.
+ *  Misuse can lead to data corruption or system instability. The backend must
+ *  support this operation.
+ *
+ *  @param[in] ept Registered endpoint by @ref ipc_service_register_endpoint.
+ *  @param[in] data Pointer to the critical message buffer to send.
+ *  @param[in] len Number of bytes to send.
+ *
+ *  @retval -EIO when no backend is registered or send_critical hook is missing
+ *               from backend.
+ *  @retval -EINVAL when instance or endpoint is invalid.
+ *  @retval -ENOENT when the endpoint is not registered with the instance.
+ *  @retval -EBADMSG when the data is invalid (i.e. invalid data format,
+ *                   invalid length exceeds backend limits, ...).
+ *  @retval -ENOTSUP when the operation is not supported by backend.
+ *  @retval -ENOMEM when even critical path buffers are exhausted.
+ *
+ *  @retval bytes number of bytes sent.
+ *  @retval other errno codes depending on the implementation of the backend.
+ */
+int ipc_service_send_critical(struct ipc_ept *ept, const void *data, size_t len);
+
 /** @brief Get the TX buffer size
  *
  *  Get the maximal size of a buffer which can be obtained by @ref
--- a/include/zephyr/ipc/ipc_service_backend.h
+++ b/include/zephyr/ipc/ipc_service_backend.h
@@ -71,6 +71,29 @@ struct ipc_service_backend {
 	int (*send)(const struct device *instance, void *token,
 		    const void *data, size_t len);
 
+	/** @brief Pointer to the function that sends critical messages bypassing flow control.
+	 *
+	 *  This function sends high-priority messages directly, bypassing busy checks
+	 *  and normal queuing mechanisms. It should be used only for critical system
+	 *  notifications like crash reports or fatal errors.
+	 *
+	 *  @param[in] instance Instance pointer.
+	 *  @param[in] token Backend-specific token.
+	 *  @param[in] data Pointer to the buffer to send.
+	 *  @param[in] len Number of bytes to send.
+	 *
+	 *  @retval -EINVAL when instance is invalid.
+	 *  @retval -ENOENT when the endpoint is not registered with the instance.
+	 *  @retval -EBADMSG when the message is invalid.
+	 *  @retval -ENOTSUP when critical send is not supported.
+	 *  @retval -ENOMEM when even critical buffers are unavailable.
+	 *
+	 *  @retval bytes number of bytes sent.
+	 *  @retval other errno codes depending on the implementation of the backend.
+	 */
+	int (*send_critical)(const struct device *instance, void *token,
+			     const void *data, size_t len);
+
 	/** @brief Pointer to the function that will be used to register endpoints.
 	 *
 	 *  @param[in] instance Instance to register the endpoint onto.
--- a/soc/intel/intel_adsp/common/ipc.c
+++ b/soc/intel/intel_adsp/common/ipc.c
@@ -43,94 +43,75 @@ static void intel_adsp_ipc_receive_cb(const void *data, size_t len, void *priv)
 	struct intel_adsp_ipc_data *devdata = dev->data;
 	struct intel_adsp_ipc_ept_priv_data *priv_data =
 		(struct intel_adsp_ipc_ept_priv_data *)priv;
-	bool done = true;
 
-	if (len == INTEL_ADSP_IPC_CB_MSG) {
-		const struct intel_adsp_ipc_msg *msg = (const struct intel_adsp_ipc_msg *)data;
+	__ASSERT(len == sizeof(uint32_t) * 2, "Unexpected IPC payload length: %zu", len);
+	__ASSERT(data != NULL, "IPC payload pointer is NULL");
 
-		if (devdata->handle_message != NULL) {
-			done = devdata->handle_message(dev, devdata->handler_arg, msg->data,
-						       msg->ext_data);
-		}
+	const uint32_t *msg = (const uint32_t *)data;
 
-		if (done) {
-			priv_data->cb_ret = INTEL_ADSP_IPC_CB_RET_OKAY;
-		} else {
-			priv_data->cb_ret = -EBADMSG;
-		}
-	} else if (len == INTEL_ADSP_IPC_CB_DONE) {
-		bool external_completion = false;
-
-		if (devdata->done_notify != NULL) {
-			external_completion = devdata->done_notify(dev, devdata->done_arg);
-		}
-
-		if (external_completion) {
-			priv_data->cb_ret = INTEL_ADSP_IPC_CB_RET_EXT_COMPLETE;
-		} else {
-			priv_data->cb_ret = INTEL_ADSP_IPC_CB_RET_OKAY;
-		}
+	if (devdata->handle_message != NULL) {
+		priv_data->msg_done = devdata->handle_message(dev, devdata->handler_arg, msg[0],
+							      msg[1]);
 	}
 }
 
 void intel_adsp_ipc_complete(const struct device *dev)
 {
-	int ret;
-
-	ret = ipc_service_send(&intel_adsp_ipc_ept, NULL, INTEL_ADSP_IPC_SEND_DONE);
+	ARG_UNUSED(dev);
+	int ret = ipc_service_release_rx_buffer(&intel_adsp_ipc_ept, NULL);
 
 	ARG_UNUSED(ret);
+	__ASSERT(ret == 0, "ipc_service_release_rx_buffer() failed: %d", ret);
 }
 
 bool intel_adsp_ipc_is_complete(const struct device *dev)
 {
-	int ret;
-
-	ret = ipc_service_send(&intel_adsp_ipc_ept, NULL, INTEL_ADSP_IPC_SEND_IS_COMPLETE);
-
-	return ret == 0;
+	ARG_UNUSED(dev);
+	return ipc_service_get_tx_buffer_size(&intel_adsp_ipc_ept) > 0;
 }
 
 int intel_adsp_ipc_send_message(const struct device *dev, uint32_t data, uint32_t ext_data)
 {
-	struct intel_adsp_ipc_msg msg = {.data = data, .ext_data = ext_data};
-	int ret;
+	ARG_UNUSED(dev);
+	uint32_t msg[2] = {data, ext_data};
 
-	ret = ipc_service_send(&intel_adsp_ipc_ept, &msg, INTEL_ADSP_IPC_SEND_MSG);
-
-	if (ret < 0) {
-		return ret;
-	}
-
-	return 0;
+	return ipc_service_send(&intel_adsp_ipc_ept, &msg, sizeof(msg));
 }
 
+/*
+ * This helper sends an IPC message and then waits synchronously for completion using the backend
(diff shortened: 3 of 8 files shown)


soc/intel/intel_adsp/common/ipc.c
```diff
@@ -43,94 +43,56 @@ static void intel_adsp_ipc_receive_cb(const void *data, size_t len, void *priv)
...
-	struct intel_adsp_ipc_msg msg = {.data = data, .ext_data = ext_data};
-	int ret;
-
-	ret = ipc_service_send(&intel_adsp_ipc_ept, &msg, INTEL_ADSP_IPC_SEND_MSG);
-
-	if (ret < 0) {
-		return ret;
-	}
+	ARG_UNUSED(dev);
+	uint32_t msg[2] = {data, ext_data};
```
kv2019i, line 76: I've been looking at how to bolt my https://github.com/zephyrproject-rtos/zephyr/pull/98862 on top of this, but seems this is doable. So if the full payload is sent using this interface, the msg would be 4x32bit, with inband header and an optional larger payload.

I do wonder if the "struct intel_adsp_ipc_msg" should be kept as the payload is somewhat special already with the 2x32bit inband message.

subsys/ipc/ipc_service/backends/ipc_intel_adsp_host_ipc.c
```diff
@@ -301,51 +288,99 @@ static int ipc_send_message_emergency(const struct device *dev, uint32_t data, u
...
+ */
+int intel_adsp_ipc_hold_rx_buffer(const struct device *instance, void *token, void *data)
+{
+	ARG_UNUSED(instance);
+	ARG_UNUSED(token);
+	ARG_UNUSED(data);
+	return -ENOTSUP;
+}
+
+int intel_adsp_ipc_release_rx_buffer(const struct device *instance, void *token, void *data)
```
kv2019i, line 332: I think this is nice, using release_rx_buffer() nicely maps to completing the IPC.

soc/intel/intel_adsp/common/ipc.c
```diff
@@ -43,94 +43,56 @@ static void intel_adsp_ipc_receive_cb(const void *data, size_t len, void *priv)
...
-	ret = ipc_service_send(&intel_adsp_ipc_ept, &msg, INTEL_ADSP_IPC_SEND_MSG_SYNC);
-
-	if (ret < 0) {
-		return ret;
-	}
+	ARG_UNUSED(dev);
+	uint32_t msg[2] = {data, ext_data};
 
-	return 0;
+	return ipc_service_send(&intel_adsp_ipc_ept, &msg, sizeof(msg));
```
kv2019i: Doesn't this break the semantics as this is no longer synchronous?

It's another question can we drop the synchronous send variant altogether. It seems this is only used by the unit tests (e.g. zephyr/tests/boards/intel_adsp/smoke/src/hostipc.c ). 

If needed, this could be replaced by async send, and an implementation of "ipc_service_get_tx_buffer()" that allows to wait until send is complete.

kv2019i (commented):
All in all looks good. I think using the service interface rx/tx buffer ops is cleaner mapping. Some questions on whether we keep a struct for the payload (which is still special) and a question on the synchronous send variant.

soc/intel/intel_adsp/common/ipc.c
tmleman: Yes, but I'm fixing this in a later commit (https://github.com/zephyrproject-rtos/zephyr/pull/100682/commits/521a1065773923b043bfd1807d2aaf76e1004c8e). Ideally this should be one commit, but I wanted to discuss the target solution. One of the considered options is to remove this functionality.

soc/intel/intel_adsp/common/ipc.c
tmleman: I haven't seen the changes you're proposing before. I was considering a similar solution, but instead of a structure with a pointer, I saw it more as a buffer that has space for two dwords of message + payload. When SOF sends a message, the IPC backend would know how many bytes to copy to the mailbox based on the passed size.

kv2019i: @tmleman Given SOF client code is using a scatter gather approach (the IPC payload is a tuple of upto two register words and a variable size payload), I think we need some structure for the payload. We will have variance betweeb different HW implementations (some transfer data inband in the doorbell registers, some don't), and between IPC versions (IPC3 has no concept of "ext_data"). 

I was thinking a four item tuple "first_word, second_word, payload_ptr, payload_len" would be enough to cover all variation we know about. Given we are mixing data and pointers, I think it would be better to use a struct to describe the object that is passed to IPC service send routine. I think this makes the code more readable both in SOF side, as well as the various Zephyr backends.

dcpleung (dismissed):

soc/intel/intel_adsp/common/ipc.c
kv2019i: @tmleman Any takes on this? Do you prefer to merge this first, and then follow-up with the payload extension? I'm ok with this as well, but would be good to have a rough consensus of how this can be extended. If this will be forever limited to "two word" payloads, not sure it's worth the effort tot move this to IPC service as we already know now this will only be useful for one type of hardware. And if that's the case, SOF code could just directly use the driver interface.

tmleman: @kv2019i Right now I don’t have a concrete design for the payload descriptor that would be passed to the backend via the IPC service.

The steps I had in mind are:
- After these changes are merged, update the Zephyr version in SOF and run regression tests.
- Introduce IPC service usage in SOF and replace the direct use of the Intel‑specific IPC driver.
- Clean up the Intel ADSP IPC path by removing the old IPC driver so only the backend remains.
- Once we are in that state, we can extend the backend to handle richer IPC payloads (data in mailboxes/shared memory).

Today, the data we pass through IPC service is intentionally backend‑specific, and each vendor can define its own layout. I’d like to hear from other vendors what they would need here. If we see enough commonality for SOF IPC, we can then agree on a shared struct or helper type on top of that.

tmleman: @kv2019i here are two PR in SOF to test those changes:
- https://github.com/thesofproject/sof/pull/10423
- https://github.com/thesofproject/sof/pull/10422

kv2019i: @tmleman Ack thanks, I think we are aligned then. I'm good to take this in steps. We can use the size parameter to incrementally add features (if message length is more than 2 words, then client is using a newer interface).

W.r.t. this:
>Today, the data we pass through IPC service is intentionally backend‑specific, and each vendor can define its own 
> layout. I’d like to hear from other vendors what they would need here.

I think can aim higher already now. If you look at IPC3 drivers in https://github.com/thesofproject/sof/blob/main/src/drivers/ , we have 4 four vendors using exactly the same layout. So I think it's really already aligned in SOF -- we have two layouts currently, IPC3 and IPC4 and the two are never compiled in same binary, so in practise there's no variation and the layout doesn't vary from hardware to another. This has the great benefit that we don't need add another abstract on client (SOF) side, but can directly use Zephyr IPC interface. 

I mean if any folks from other vendors object, I'm happy to change my mind, but given this already worked once, I'd rather go towards this goal and keep code in SOF clean and take full benefit of the hardware abstraction Zephyr provides.

kv2019i (dismissed):
Test results look good, +1 from me.

@tmleman maybe remove the "RFC" tag and "WIP" from summary to gather more feedback.

soc/intel/intel_adsp/common/ipc.c
```diff
@@ -43,94 +43,57 @@ static void intel_adsp_ipc_receive_cb(const void *data, size_t len, void *priv)
...
-		if (external_completion) {
-			priv_data->cb_ret = INTEL_ADSP_IPC_CB_RET_EXT_COMPLETE;
-		} else {
-			priv_data->cb_ret = INTEL_ADSP_IPC_CB_RET_OKAY;
-		}
+
+	__ASSERT(len == sizeof(uint32_t) * 2, "Unexpected IPC payload length: %zu", len);
+	__ASSERT(data != NULL, "IPC payload pointer is NULL");
+
+	const uint32_t *msg = (const uint32_t *)data;
```
lyakh, line 50: looks like the type-cast shouldn't be needed

soc/intel/intel_adsp/common/ipc.c
lyakh: in SOF IPC sending will remain synchronous though?

subsys/ipc/ipc_service/backends/ipc_intel_adsp_host_ipc.c
```diff
@@ -301,51 +288,94 @@ static int ipc_send_message_emergency(const struct device *dev, uint32_t data, u
...
 
-	switch (len) {
-	case INTEL_ADSP_IPC_SEND_MSG: {
-		ret = ipc_send_message(dev, msg->data, msg->ext_data);
+	if (dev == NULL) {
+		return -EINVAL;
+	}
 
-		break;
+	if ((len != (sizeof(uint32_t) * 2)) || data == NULL) {
```
lyakh: 4 parentheses too many: `if (len != sizeof(uint32_t) * 2 || data == NULL)`

subsys/ipc/ipc_service/backends/ipc_intel_adsp_host_ipc.c
```diff
@@ -301,51 +288,94 @@ static int ipc_send_message_emergency(const struct device *dev, uint32_t data, u
...
+	ARG_UNUSED(token);
+
+	if (dev == NULL) {
+		return -EINVAL;
 	}
-	default:
-		ret = -EBADMSG;
 
-		break;
+	if ((len != (sizeof(uint32_t) * 2)) || data == NULL) {
```
lyakh: ditto parentheses

subsys/ipc/ipc_service/backends/ipc_intel_adsp_host_ipc.c
```diff
@@ -301,51 +288,94 @@ static int ipc_send_message_emergency(const struct device *dev, uint32_t data, u
...
+		return -EBADMSG;
 	}
 
-	return ret;
+	const uint32_t *msg = (const uint32_t *)data;
```
lyakh, line 358: supposedly superfluous type-cast, but needs checking

soc/intel/intel_adsp/common/ipc.c
tmleman: @kv2019i We haven't made a decision yet about what we do with 'intel_adsp_ipc_send_message_sync'. I would probably go in the direction of removing this function, I don't think we ever had a use case for it.

soc/intel/intel_adsp/common/ipc.c
```diff
@@ -78,13 +78,31 @@ int intel_adsp_ipc_send_message(const struct device *dev, uint32_t data, uint32_
...
 {
-	ARG_UNUSED(dev);
 	uint32_t msg[2] = {data, ext_data};
+	struct intel_adsp_ipc_data *devdata = dev->data;
 
-	return ipc_service_send(&intel_adsp_ipc_ept, &msg, sizeof(msg));
+	int ret = ipc_service_send(&intel_adsp_ipc_ept, &msg, sizeof(msg));
+
+	if (ret < 0) {
+		k_sem_take(&devdata->sem, timeout);
```
lyakh, line 102: oh... so the only reason it can fail is because it's delayed? I see a few more ways how `ipc_service_send()` can fail. At least we should check a specific error code, since there's no other way to indicate asynchronous execution. Or maybe we can call the backend `.send()` function directly or maybe a new internal function like
```
int intel_adsp_ipc_send_message_sync(...)
{
    return intel_adsp_ipc_send_message_internal(..., timeout);
}
int intel_adsp_ipc_send_message(...)
{
    return intel_adsp_ipc_send_message_internal(..., 0);
}
```
also previously the check was `ret == 0` which wasn't perfect but made a bit more sense: not an error, not a positive byte count that has already been sent, but 0, meaning asynchronous sending in progress.

dcpleung (dismissed):

soc/intel/intel_adsp/common/ipc.c
tmleman: I don't see a use case for this function and I would prefer to remove it.

include/zephyr/ipc/ipc_service.h
```diff
@@ -315,6 +315,36 @@ int ipc_service_deregister_endpoint(struct ipc_ept *ept);
...
+ *  WARNING: This function should only be used for critical system notifications.
+ *  Misuse can lead to data corruption or system instability. The backend must
+ *  support this operation.
+ *
+ *  @param[in] ept Registered endpoint by @ref ipc_service_register_endpoint.
+ *  @param[in] data Pointer to the critical message buffer to send.
+ *  @param[in] len Number of bytes to send.
+ *
+ *  @retval -EIO when no backend is registered or send_critical hook is missing
+ *               from backend.
```
softwarecki, line 335: It seems that -EIO duplicates the -ENOTSUP.

include/zephyr/ipc/ipc_service_backend.h
```diff
@@ -71,6 +71,29 @@ struct ipc_service_backend {
...
+	 *
+	 *  This function sends high-priority messages directly, bypassing busy checks
+	 *  and normal queuing mechanisms. It should be used only for critical system
+	 *  notifications like crash reports or fatal errors.
+	 *
+	 *  @param[in] instance Instance pointer.
+	 *  @param[in] token Backend-specific token.
+	 *  @param[in] data Pointer to the buffer to send.
+	 *  @param[in] len Number of bytes to send.
+	 *
```
softwarecki, line 84: Missing -EIO

soc/intel/intel_adsp/common/ipc.c
softwarecki: Do we want to propagate the timeout occurrence?

tests/boards/intel_adsp/smoke/src/hostipc.c
softwarecki, line 128: Do we want to propagate the timeout occurrence?

softwarecki (approved):
Moving the semaphore handling to `intel_adsp_ipc_send_message_sync` is a good change. It removes the need for `ipc_send_message_sync` and simplifies the flow. Using `hold/release_rx_buffer` for ipc completion is also a clean approach. I like it! I would prefer to keep the `struct intel_adsp_ipc_msg`. The tuple is less readable and any future extension will require updating `sizeof(uint32_t) * 2` in many places.

serhiy-katsyuba-intel (approved):

kv2019i (approved):

dcpleung (approved):

PR #100683, opened 2025-12-08
MT29F4G08 NAND Flash Driver

LinderPi:
Introduces a NAND flash driver for Micron's MT29F4G08 memory device that is configurable via devicetree. A generic flash controller driver communicates to the flexible memory controller of STM32 MCUs. Any other NAND flash driver can be implemented to work via FMC in the future. The chip-specific driver provides the flash API for upper layers and enables the on-die ECC feature.

The flash controller is strongly inspired by the STM32 HAL NAND driver following the ONFI specification. However, it optionally uses DMA to transfer a read page. In future, an additional function can be implemented for entire page reads. At the moment this is not required by upper layers and the direct memory transfer is somehow slower to a yet unknown problem.

As it is the first NAND flash driver in Zephyr, some API extensions are foreseen for bad block management by a flash translation layer (FTL). As proposed in #99908, we are currently working on a NFTL disk driver module based on Dhara. A first version is implemented and available [here](https://github.com/endresshauser-lp/dhara/tree/zephyr) for testing.

As an optional enhancement, a bad block map could be kept in memory such that the driver does not need to ask the memory device for bad blocks on every request.


drivers/flash/Kconfig.stm32_fmc
```diff
@@ -0,0 +1,17 @@
...
+# Copyright (c) 2025 Endress+Hauser GmbH+Co. KG
+# SPDX-License-Identifier: Apache-2.0
+
+DT_STM32_FMC_NAND_HAS_DMA := $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_FMC_NAND),dmas)
+
+config FLASH_STM32_FMC_NAND
+	bool "STM32 FMC NAND flash controller"
+	default y
+	depends on DT_HAS_ST_STM32_FMC_NAND_ENABLED
+	select MEMC
```
etienne-lms, line 12: Is `CONFIG_MEMC` needed?

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,336 @@
...
+							   spare_area);
+		*(int *)out = ((ret == 0) && (spare_area[0] != 0xFF)) ? 1 : 0;
+		break;
+	}
+
+	case NAND_FLASH_MARK_BAD_BLOCK: {
+		/* Mark bad block in first page */
+		size_t block = *(size_t *)in;
+		struct nand_flash_address address;
+		uint8_t spare_area[config->spare_area_size];
```
etienne-lms: Only 1st byte get initialized, when writing to spare area. Would it be worth to fully zero-initialize this buffer?
Or asserting at build time that spare area size is 1.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,336 @@
...
+		LOG_ERR("NAND flash reset failed with error %d", ret);
+		return -EIO;
+	}
+
+#ifdef CONFIG_FLASH_MT29F4G08_ECC
+	/* Enable on-die ECC feature */
+	struct nand_flash_feature ecc_feature = {
+		.feature_addr = ECC_FEATURE_ADDR,
+		.feature_data = ECC_FEATURE_DATA,
+	};
```
etienne-lms, line 297: Could you add an empty line below local variables definitions?

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,336 @@
...
+		.page_size = DT_INST_PROP(n, page_size),                                           \
+		.spare_area_size = DT_INST_PROP(n, spare_area_size),                               \
+		.block_size = DT_INST_PROP(n, block_size),                                         \
+		.plane_size = DT_INST_PROP(n, plane_size),                                         \
+		.flash_size = DT_INST_PROP(n, flash_size),                                         \
+		.setup_time = DT_INST_PROP(n, setup_time),                                         \
+		.wait_setup_time = DT_INST_PROP(n, wait_setup_time),                               \
+		.hold_setup_time = DT_INST_PROP(n, hold_setup_time),                               \
+		.hiz_setup_time = DT_INST_PROP(n, hiz_setup_time),                                 \
+		LAYOUT_PAGES_PROP(n),                                                              \
```
etienne-lms: The terminal comma should move into `LAYOUT_PAGES_PROP()` macro value, in case `CONFIG_FLASH_PAGE_LAYOUT` is disabled.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,336 @@
...
+		}))
+
+#define FLASH_STM32_FMC_MT29F4G08_INIT(n)                                                          \
+	static const struct flash_mt29f4g08_config flash_stm32_fmc_mt29f4g08_config_##n = {        \
+		.controller = DEVICE_DT_GET(DT_INST_PARENT(n)),                                    \
+		.parameters =                                                                      \
+			{                                                                          \
+				.write_block_size = DT_INST_PROP(n, page_size),                    \
+				.erase_value = 0xff,                                               \
+			},                                                                         \
```
etienne-lms:
```suggestion
		.parameters = {                                                                        \
			.write_block_size = DT_INST_PROP(n, page_size),                            \
			.erase_value = 0xff,                                                       \
		},                                                                                  \

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c [hunk 6]
```diff
@@ -0,0 +1,336 @@
...
+	.ex_op = flash_mt29f4g08_ex_op,
+#endif /* CONFIG_FLASH_EX_OP_ENABLED */
+};
+
+/* A page in this context corresponds to the smallest erasable area which is a block */
+#define LAYOUT_PAGES_PROP(n)                                                                       \
+	IF_ENABLED(CONFIG_FLASH_PAGE_LAYOUT,                                                       \
+		(.layout = {                                                                       \
+			.pages_count = DT_INST_PROP(n, flash_size) /                               \
+				       DT_INST_PROP(n, block_size),                                \
```
etienne-lms: Fits into a signle line (easier to read)
```suggestion
			.pages_count = DT_INST_PROP(n, flash_size) / DT_INST_PROP(n, block_size),  \

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c [hunk 7]
```diff
@@ -0,0 +1,336 @@
...
+
+	switch (code) {
+	case NAND_FLASH_IS_BAD_BLOCK: {
+		/* Check bad block in first page */
+		size_t block = *(size_t *)in;
+		struct nand_flash_address address;
+		uint8_t spare_area[config->spare_area_size];
+
+		if (block * config->block_size >= config->flash_size) {
+			*(int *)out = 1;
```
etienne-lms: Maybe not needed to provide an output value when the function returns with an error status.

drivers/flash/flash_stm32_fmc_nand.c
```diff
@@ -0,0 +1,755 @@
...
+	dev_data->total_pages = init->flash_size / init->page_size;
+
+	(void)FMC_NAND_Init(dev_data->instance, &dev_data->init);
+
+	FMC_NAND_PCC_TimingTypeDef timing = {
+		.SetupTime = init->setup_time,
+		.WaitSetupTime = init->wait_setup_time,
+		.HoldSetupTime = init->hold_setup_time,
+		.HiZSetupTime = init->hiz_setup_time,
+	};
```
etienne-lms, line 538: Could you add an empty line below this local variable definition?

drivers/flash/flash_stm32_fmc_nand.c
```diff
@@ -0,0 +1,755 @@
...
+		.source_burst_length = 64,                                                         \
+		.dest_burst_length = 64,                                                           \
+		.block_count = 1,                                                                  \
+		.dma_callback = fmc_nand_dma_callback,                                             \
+	}
+
+#define FMC_NAND_DMA_CHANNEL(node, dir)                                                            \
+	.dma = {COND_CODE_1(DT_DMAS_HAS_NAME(node, dir),                                           \
+			    (FMC_NAND_DMA_CHANNEL_INIT(node, dir)),                                \
+			    NULL)},
```
etienne-lms:
```suggestion
			   (NULL))},

File: drivers/flash/flash_stm32_fmc_nand.c [hunk 10]
```diff
@@ -0,0 +1,755 @@
...
+#define FLASH_STM32_FMC_NAND_INIT(n)                                                               \
+	static uint8_t __nocache __aligned(PAGE_BUFFER_ALIGNMENT)                                  \
+	flash_stm32_fmc_nand_page_buffer_##n[DT_INST_PROP(n, page_buffer_size)];                   \
+                                                                                                   \
+	static const struct flash_stm32_fmc_nand_config flash_stm32_fmc_nand_config_##n = {        \
+		.page_buffer = flash_stm32_fmc_nand_page_buffer_##n,                               \
+	};                                                                                         \
+                                                                                                   \
+	static struct flash_stm32_fmc_nand_data flash_stm32_fmc_nand_data_##n = {                  \
+		FMC_NAND_DMA_CHANNEL(DT_DRV_INST(n), tx_rx)};                                      \
```
etienne-lms:
```suggestion
		FMC_NAND_DMA_CHANNEL(DT_DRV_INST(n), tx_rx)                                        \
	};                                                                                         \

File: drivers/flash/flash_stm32_fmc_nand.h [hunk 11]
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2025 Endress+Hauser GmbH+Co. KG
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __ZEPHYR_DRIVERS_FLASH_STM32_FMC_NAND_H__
+#define __ZEPHYR_DRIVERS_FLASH_STM32_FMC_NAND_H__
```
etienne-lms:
```suggestion
#ifndef __ZEPHYR_DRIVERS_FLASH_FLASH_STM32_FMC_NAND_H__
#define __ZEPHYR_DRIVERS_FLASH_FLASH_STM32_FMC_NAND_H__
```
to match header file path.

include/zephyr/drivers/flash/nand_flash_api_ex.h
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Checks whether a block is marked as bad. As input it takes the block index (size_t *). As
+	 * output it returns 1 if the block is bad, 0 otherwise (int *).
+	 */
+	NAND_FLASH_IS_BAD_BLOCK = FLASH_EX_OP_VENDOR_BASE,
+
+	/**
+	 * Marks a block as bad. As input it takes the block index (size_t *). There is no output.
+	 */
+	NAND_FLASH_MARK_BAD_BLOCK
```
etienne-lms: Nitpicking suggestion, to ease later addition of other extended operations.
```suggestion
	NAND_FLASH_MARK_BAD_BLOCK,
```

drivers/flash/flash_stm32_fmc_nand.c
```diff
@@ -0,0 +1,755 @@
...
+	ARG_UNUSED(dev);
+	ARG_UNUSED(user_data);
+
+	if (status < 0) {
+		LOG_ERR("DMA callback error %d with channel %d", status, channel);
+	}
+}
+#endif /* STM32_FMC_NAND_USE_DMA */
+
+#if STM32_FMC_NAND_USE_DMA
```
etienne-lms:
```suggestion

**Review by etienne-lms — COMMENTED** [review]
Minor comments only.

File: drivers/flash/Kconfig.stm32_fmc (hunk 1 above)
Comment (LinderPi) [reply drivers/flash/Kconfig.stm32_fmc:12]: Yes, otherwise `memc_stm32_fmc_clock_rate()` is not available to log the FMC clock rate.

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c (hunk 2 above)
Comment (LinderPi) [reply drivers/flash/flash_stm32_fmc_mt29f4g08.c]: It makes sense to initialise the buffer. However, the size is not always 1. In our case it is 64.

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c (hunk 7 above)
Comment (LinderPi) [reply drivers/flash/flash_stm32_fmc_mt29f4g08.c]: True, the upcoming FTL is implemented to not consider the return value on failure. From a defensive programming point of view I would like to leave it anyways.

File: drivers/flash/Kconfig.stm32_fmc (hunk 1 above)
Comment (tpambor) [reply drivers/flash/Kconfig.stm32_fmc:12]: Furthermore, clock setup is handled by https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/memc/memc_stm32.c driver, which is only compiled if `CONFIG_MEMC=y`, so this is required even if FMC clock rate would not be logged.

File: drivers/flash/flash_stm32_fmc_nand.c [hunk 14]
```diff
@@ -0,0 +1,767 @@
...
+		.SetupTime = init->setup_time,
+		.WaitSetupTime = init->wait_setup_time,
+		.HoldSetupTime = init->hold_setup_time,
+		.HiZSetupTime = init->hiz_setup_time,
+	};
+
+	(void)FMC_NAND_CommonSpace_Timing_Init(dev_data->instance, &timing,
+					       dev_data->init.NandBank);
+	(void)FMC_NAND_AttributeSpace_Timing_Init(dev_data->instance, &timing,
+						  dev_data->init.NandBank);
```
erwango, line 543: In theory, we're not supposed to call directly these apis, which are not meant to be public, but API internal to the HAL driver. As such there is no guarantee on these APIs (portability, stability, ..).
Can you detail why you chose not to call HAL_NAND_Init() directly ?

LinderPi: The STM32 HAL NAND driver does not provide the entire functionality we need (DMA support, status read, etc.). Additionally, there is no intend to further develop it. This is the reasoning I got from ST's official support: "The NAND is not so popular since the xSPI memories we see the FMC usage less and less."

erwango: > The STM32 HAL NAND driver does not provide the entire functionality we need (DMA support, status read, etc.).

Ok, that's clear. However, I need to check internally if `FMC_NAND_()` usage is safe.

drivers/flash/flash_stm32_fmc_nand.c
```diff
@@ -0,0 +1,767 @@
...
+			dev_data->state = NAND_STATE_ERROR;
+			k_sem_give(&dev_data->lock);
+			return -EIO;
+		}
+		if (HAL_DMA_PollForTransfer(&dev_data->dma.handle, HAL_DMA_FULL_TRANSFER,
+					    DMA_TIMEOUT_MS) != HAL_OK) {
+			dev_data->state = NAND_STATE_ERROR;
+			k_sem_give(&dev_data->lock);
+			return -EIO;
+		}
```
erwango: If not using HAL for FMC, it would be good to avoid using it for DMA.
Was there a reason you chose not to use Zephyr DMA API ?

LinderPi: Our goal was to have the highest transfer rate possible. We got this by using the HAL DMA driver API for the page transfer and the Zephyr DMA API for the chunk transfer. There are also other STMicroelectonics drivers making use of both APIs. I suppose that there is a restriction in the DMA driver. Unfortunately, I did not yet have the time to further investigate this.

drivers/flash/flash_stm32_fmc_nand.c
```diff
@@ -0,0 +1,767 @@
...
+	dev_data->state = NAND_STATE_READY;
+	k_sem_give(&dev_data->lock);
+
+	return 0;
+}
+
+static int flash_stm32_fmc_nand_init(const struct device *dev)
+{
+	struct flash_stm32_fmc_nand_data *dev_data = dev->data;
+	uint32_t fmc_freq;
```
etienne-lms:
```suggestion
	uint32_t __maybe_unused fmc_freq ;

File: drivers/flash/Kconfig.stm32_fmc (hunk 1 above)
Comment (etienne-lms) [reply drivers/flash/Kconfig.stm32_fmc:12]: Ok, thanks, I missed that :-(

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c [hunk 17]
```diff
@@ -0,0 +1,338 @@
...
+
+	while (len > 0) {
+		off_t page_offset = offset % config->page_size;
+		size_t chunk = ((page_offset + len) < config->page_size)
+				       ? len
+				       : (config->page_size - page_offset);
+		struct nand_flash_address address =
+			flash_mt29f4g08_calculate_address(config, offset);
+		int ret = flash_stm32_fmc_nand_read_page_chunk(controller, &address, page_offset,
+							       chunk, (uint8_t *)data);
```
etienne-lms: Could you add an empty line below these local variable definitions?

Ditto in `flash_mt29f4g08_write()`.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,338 @@
...
+		}
+
+		memset(spare_area, 0x00, config->spare_area_size);
+		address = flash_mt29f4g08_calculate_address(config, block * config->block_size);
+		ret = flash_stm32_fmc_nand_read_spare_area(config->controller, &address,
+							   spare_area);
+		*(int *)out =
+			((ret == 0) && (spare_area[0] != 0xFF)) ? NAND_BLOCK_BAD : NAND_BLOCK_GOOD;
+		break;
+	}
```
etienne-lms, line 198: I don't think these braces (in both `case` sequences)  are useful. There indentation could be confusing. Could you remove them?

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,338 @@
...
+			*(int *)out = NAND_BLOCK_BAD;
+			return -EINVAL;
+		}
+
+		memset(spare_area, 0x00, config->spare_area_size);
+		address = flash_mt29f4g08_calculate_address(config, block * config->block_size);
+		ret = flash_stm32_fmc_nand_read_spare_area(config->controller, &address,
+							   spare_area);
+		*(int *)out =
+			((ret == 0) && (spare_area[0] != 0xFF)) ? NAND_BLOCK_BAD : NAND_BLOCK_GOOD;
```
etienne-lms: For consistency:
```suggestion
		if (ret == 0) {
			*(int *)out = (spare_area[0] != 0xFF) ? NAND_BLOCK_BAD : NAND_BLOCK_GOOD;
		}

File: drivers/flash/flash_stm32_fmc_mt29f4g08.c [hunk 20]
```diff
@@ -0,0 +1,338 @@
...
+		break;
+	}
+
+	case NAND_FLASH_MARK_BAD_BLOCK: {
+		/* Mark bad block in first page */
+		size_t block = *(size_t *)in;
+		struct nand_flash_address address;
+		uint8_t spare_area[config->spare_area_size];
+
+		if (block * config->block_size >= config->flash_size) {
```
etienne-lms:
```suggestion
		off_t offset;

		if (u32_mul_overflow(block, config->block_size, &offset) || (offset >= config->flash_size)) {
```
and below:
```patch
-		address = flash_mt29f4g08_calculate_address(config, block * config->block_size);
+		address = flash_mt29f4g08_calculate_address(config, offset);

File: drivers/flash/flash_stm32_fmc_nand.c [hunk 21]
```diff
@@ -0,0 +1,767 @@
...
+			k_sem_give(&dev_data->lock);
+			return -EIO;
+		}
+	} else {
+#endif /* STM32_FMC_NAND_USE_DMA */
+
+		/* Get page data into buffer */
+		for (size_t index = 0; index < dev_data->page_size; index++) {
+			config->page_buffer[index] = sys_read8(NAND_DEVICE);
+		}
```
etienne-lms, line 207: Using `config->page_buffer` allocated in cache memory is useful when using DMAs, but here, maybe worth the straight load data read into the likely cached `data` buffer.
If using this scheme, you could avoid `flash_stm32_fmc_nand_page_buffer_XXX[]` allocations.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
LinderPi: But then I have to set the output value to bad in the else case.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
LinderPi: Formatter puts it that way. Following change would work:
```
		.parameters = {.write_block_size = DT_INST_PROP(n, page_size),                     \
				       .erase_value = 0xff},                                               \
```

drivers/flash/flash_stm32_fmc_nand.c
LinderPi: Formatter puts it that way.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
LinderPi: The advantage is to define operation specific variables in the block. Also, the indentation remains the same.

drivers/flash/flash_stm32_fmc_nand.c
LinderPi: What could work is something like:
```
/* Read chunk of page data directly into output buffer */
for (size_t index = 0; index < dev_data->page_size; index++) {
	if (index >= page_offset && index < (page_offset + chunk)) {
		*data = sys_read8(NAND_DEVICE);
		data++;
	} else {
		(void)sys_read8(NAND_DEVICE);
	}
}
```
In our case, this speeds up reading a page from 3.6 MiBps to 4.5 MiBps. Is this what you have imagined?

drivers/flash/flash_stm32_fmc_nand.c
erwango: > There are also other STMicroelectonics drivers making use of both APIs.

Actually, the only case is for HAL based drivers that use DMA internally to the HAL, and then we call `dma_configure()` in order to set the DMA channels busy in Zephyr DMA API. But otherwise all drivers use zephyr dma API.
Other than performance, would there be issues to use Zephyr DMA API ?

drivers/flash/flash_stm32_fmc_mt29f4g08.c
LinderPi: Nice point. Thanks :)

drivers/flash/flash_stm32_fmc_nand.c
LinderPi: I changed it to use Zephyr DMA API only. Different implementations led to various performance results for reading an entire page:
- Initial: 9.5 MiBps
- Busy waiting using `dma_get_status()`: 8.7 MiBps
- Semaphore synchronisation for first transfer: 8.4 MiBps
- Semaphore synchronisation for both transfers: 6.9 MiBps

I decided to go for a semaphore synchronisation after getting the page data into the buffer. It is a good trade-off between a clean solution and good performance.

dts/bindings/memory-controllers/st,stm32-fmc-nand.yaml
```diff
@@ -0,0 +1,145 @@
...
+    type: int
+    required: true
+    description: |
+      Page buffer size in bytes. Must be equal to the largest NAND page size among the child nodes.
+
+child-binding:
+  description: NAND bank.
+
+  properties:
+    reg:
```
JarmouniA, line 78:
```suggestion
    bank:
```

dts/bindings/flash_controller/micron,mt29f4g08.yaml
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2025 Endress+Hauser GmbH+Co. KG
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  NAND flash driver for Micron MT29F4G08 flash memory.
```
JarmouniA, line 5:
```suggestion
  Micron MT29F4G08 NAND flash memory.
```

dts/bindings/flash_controller/micron,mt29f4g08.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  The MT29F4G08 is a 4 Gb NAND flash organised in pages of 2048 bytes each.
+
+  Refer to the Micron MT29F4G08 datasheet for more details.
+
+compatible: "micron,mt29f4g08"
+
+properties:
```
JarmouniA, line 13: Please add a description to all the properties.

JarmouniA (changes requested):

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,356 @@
...
+	while (len > 0) {
+		off_t page_offset = offset % config->page_size;
+		size_t chunk = ((page_offset + len) < config->page_size)
+				       ? len
+				       : (config->page_size - page_offset);
+		struct nand_flash_address address =
+			flash_mt29f4g08_calculate_address(config, offset);
+
+		ret = flash_stm32_fmc_nand_read_page_chunk(controller, &address, page_offset, chunk,
+							   (uint8_t *)data);
```
ttmut, line 73: Sorry if this was asked before but should not this driver be controller-independent instead? It seems every vendor will have to maintain their own variant of `flash_vendor_mt29f4g08.c`.

tpambor: First, the split between `flash_stm32_fmc_nand.c` and `flash_stm32_fmc_mt29f4g08.c` is due to the MT29F4G08 driver making use of on‑die ECC, meaning error detection and correction are handled inside the NAND device itself. This is a vendor‑specific feature. As a result, supporting on‑die ECC reliably requires a NAND flash vendor‑specific flash driver rather than a fully generic one.

It would indeed be possible to add a controller‑specific but flash chip vendor‑agnostic driver such as a hypothetical `flash_stm32_fmc_onfi.c`. Such a driver could work with ONFI‑compliant NAND devices across different vendors. However, to stay generic, it would need to rely on host ECC (calculated in the MCU) or software ECC, and therefore could not take advantage of on‑die ECC features provided by some flashes.

The reason these drivers are NAND controller‑dependent is timing. Parallel NAND flashes have fairly strict and sometimes subtle timing requirements (setup/hold times, access times, etc.), and each controller exposes its own mechanism and parameters for generating those timings. In practice, it is very difficult to define a standardized, controller‑independent interface that maps cleanly onto the timing configuration model of every NAND controller. For the same reason, we also have vendor‑specific bindings for e.g. SDRAM, where the timing parameters are inherently tied to the memory controller’s register layout and capabilities. See, e.g.:

https://github.com/zephyrproject-rtos/zephyr/blob/32823f39fbf2663fa1bbb75e60e5743cd6d91579/dts/bindings/memory-controllers/st%2Cstm32-fmc-sdram.yaml#L142-L166

ttmut: > First, the split between `flash_stm32_fmc_nand.c` and `flash_stm32_fmc_mt29f4g08.c` is due to the MT29F4G08 driver making use of on‑die ECC, meaning error detection and correction are handled inside the NAND device itself. This is a vendor‑specific feature. As a result, supporting on‑die ECC reliably requires a NAND flash vendor‑specific flash driver rather than a fully generic one.

> 
> It would indeed be possible to add a controller‑specific but flash chip vendor‑agnostic driver such as a hypothetical `flash_stm32_fmc_onfi.c`. Such a driver could work with ONFI‑compliant NAND devices across different vendors. However, to stay generic, it would need to rely on host ECC (calculated in the MCU) or software ECC, and therefore could not take advantage of on‑die ECC features provided by some flashes.
> 

> The reason these drivers are NAND controller‑dependent is timing. Parallel NAND flashes have fairly strict and sometimes subtle timing requirements (setup/hold times, access times, etc.), and each controller exposes its own mechanism and parameters for generating those timings. In practice, it is very difficult to define a standardized, controller‑independent interface that maps cleanly onto the timing configuration model of every NAND controller. For the same reason, we also have vendor‑specific bindings for e.g. SDRAM, where the timing parameters are inherently tied to the memory controller’s register layout and capabilities. See, e.g.:
> 
> https://github.com/zephyrproject-rtos/zephyr/blob/32823f39fbf2663fa1bbb75e60e5743cd6d91579/dts/bindings/memory-controllers/st%2Cstm32-fmc-sdram.yaml#L142-L166

I did not actually mean that there should be single driver that covers all NAND chips. I was asking if it is possible to move the calls to STM32 FMC inside `flash_stm32_fmc_mt29f4g08.c` behind some common API so that it becomes vendor-agnostic, such as `flash_mt29f4g08.c`. Otherwise, we are going to end up with `flash_vendorA_mt29f4g08.c`, `flash_vendorB_mt29f4g08.c` whenever someone decides to add support for their controller.

tpambor: > I did not actually mean that there should be single driver that covers all NAND chips. I was asking if it is possible to move the calls to STM32 FMC inside `flash_stm32_fmc_mt29f4g08.c` behind some common API so that it becomes vendor-agnostic, such as `flash_mt29f4g08.c`. Otherwise, we are going to end up with `flash_vendorA_mt29f4g08.c`, `flash_vendorB_mt29f4g08.c` whenever someone decides to add support for their controller.

Yes, I wanted to make clear both decisions, why it is flash-chip dependent and why it is flash-controller dependent as others might have the same question. The second part was about the flash-controller dependency.

> > The reason these drivers are NAND controller‑dependent is timing. Parallel NAND flashes have fairly strict and sometimes subtle timing requirements (setup/hold times, access times, etc.), and each controller exposes its own mechanism and parameters for generating those timings. In practice, it is very difficult to define a standardized, controller‑independent interface that maps cleanly onto the timing configuration model of every NAND controller. For the same reason, we also have vendor‑specific bindings for e.g. SDRAM, where the timing parameters are inherently tied to the memory controller’s register layout and capabilities. See, e.g.:
> > https://github.com/zephyrproject-rtos/zephyr/blob/32823f39fbf2663fa1bbb75e60e5743cd6d91579/dts/bindings/memory-controllers/st%2Cstm32-fmc-sdram.yaml#L142-L166

dts/bindings/memory-controllers/st,stm32-fmc-nand.yaml
tpambor: The other fmc bindings in tree use `reg`.
https://github.com/zephyrproject-rtos/zephyr/blob/a0fdebcc908ad484c29092c30be5513b9d346651/dts/bindings/memory-controllers/st%2Cstm32-fmc-nor-psram.yaml#L86-L93

https://github.com/zephyrproject-rtos/zephyr/blob/a0fdebcc908ad484c29092c30be5513b9d346651/dts/bindings/memory-controllers/st%2Cstm32-fmc-sdram.yaml#L115-L121

So, maybe keep it to be consistent with the others?

JarmouniA: Makes sense.

include/zephyr/drivers/flash/nand_flash.h
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @brief Structure representing a NAND flash feature.
+ */
+struct nand_flash_feature {
+	/* Feature address. */
+	uint8_t feature_addr;
+
+	/* Feature data. */
+	uint8_t feature_data[4];
+};
```
SuperHeroAbner, line 40: Is this header common for both serial/parallel nand? This feature structure looks for parallel nand with ONFI definition.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,356 @@
...
+	return 0;
+}
+
+static int flash_mt29f4g08_write(const struct device *dev, off_t offset, const void *data,
+				 size_t len)
+{
+	const struct flash_mt29f4g08_config *config = dev->config;
+	const struct device *controller = config->controller;
+	int ret;
+
```
de-nordic, line 93: Chrck 'len==0' first, if call is not going to do any work do not fail it on other checks.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,356 @@
...
+	return 0;
+}
+
+static int flash_mt29f4g08_erase(const struct device *dev, off_t offset, size_t size)
+{
+	const struct flash_mt29f4g08_config *config = dev->config;
+	const struct device *controller = config->controller;
+	int ret;
+
```
de-nordic, line 126: Check size against 0 first, it is pointless to fail call when monwork is going to happen.

dts/bindings/flash_controller/micron,mt29f4g08.yaml
```diff
@@ -0,0 +1,52 @@
...
+    type: int
+    default: 0
+
+  hold-setup-time:
+    type: int
+    default: 0
+
+  hiz-setup-time:
+    type: int
+    default: 0
```
de-nordic, line 52: Missing descriptions, units are not explained.

include/zephyr/drivers/flash.h
```diff
@@ -697,6 +697,33 @@ enum flash_ex_op_types {
...
+	 * Checks whether a block is marked as bad. As input it takes the address of the block
+	 * (off_t *). As output it returns @ref flash_block_status (enum flash_block_status *).
+	 */
+	FLASH_IS_BAD_BLOCK,
+
+	/**
+	 * Marks a block as bad. As input it takes the address of the block (off_t *). There is no
+	 * output.
+	 */
+	FLASH_MARK_BAD_BLOCK,
```
de-nordic, line 711: Please explicitly number these. It is easier to debug when i do not have to count enums, and future depreciation, if happens, would be easier.

drivers/flash/flash_stm32_fmc_mt29f4g08.c
```diff
@@ -0,0 +1,356 @@
...
+		.plane = plane_index % (config->flash_size / config->plane_size),
+	};
+}
+
+static int flash_mt29f4g08_read(const struct device *dev, off_t offset, void *data, size_t len)
+{
+	const struct flash_mt29f4g08_config *config = dev->config;
+	const struct device *controller = config->controller;
+	int ret;
+
```
de-nordic, line 58: Check len first, and return 0 if no work is going to happen.

de-nordic (changes requested):

tpambor:
Currently, no capacity to work on this.

PR #100686, opened 2025-12-08
drivers: display: ssd1306: remove i2c_burst_write usage

TanguyRaufflet:
As mentioned in the i2c_burst_write documentation, this API may not be supported by all I2C devices. The recommendation from this documentation is to use i2c_write with a buffer containing the combined address and data.

The ssd1306 driver was working well when using I2C but not when using I2C over I3C. In this case, the control command was sent in a separate transaction, which the SSD1306 does not support. To fix this issue, we should apply the recommended approach and use i2c_write instead of i2c_burst_write.

Signed-off-by: Tanguy Raufflet [tanguy.raufflet@savoirfairelinux.com](mailto:tanguy.raufflet@savoirfairelinux.com)

### DIFF
--- a/drivers/display/display_ssd1306.c
+++ b/drivers/display/display_ssd1306.c
@@ -93,6 +93,8 @@ LOG_MODULE_REGISTER(ssd1306, CONFIG_DISPLAY_LOG_LEVEL);
 #define SSD1306_RESET_DELAY			1
 #define SSD1306_SUPPLY_DELAY			20
 
+#define SSD1306_I2C_CHUNK_SIZE 128
+
 #ifndef SSD1306_ADDRESSING_MODE
 #define SSD1306_ADDRESSING_MODE		(SSD1306_MEM_ADDRESSING_HORIZONTAL)
 #endif
@@ -153,6 +155,9 @@ struct ssd1306_data {
 #if (DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(solomon_ssd1306, i2c) || \
 	DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(solomon_ssd1309, i2c) || \
 	DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(sinowealth_sh1106, i2c))
+
+K_MUTEX_DEFINE(ssd1306_i2c_buffer_lock);
+
 static bool ssd1306_bus_ready_i2c(const struct device *dev)
 {
 	const struct ssd1306_config *config = dev->config;
@@ -164,10 +169,35 @@ static int ssd1306_write_bus_i2c(const struct device *dev, uint8_t *buf, size_t 
 {
 	const struct ssd1306_config *config = dev->config;
 
-	return i2c_burst_write_dt(&config->bus.i2c,
-				  command ? SSD1306_I2C_ALL_BYTES_CMD :
-				  SSD1306_I2C_ALL_BYTES_DATA,
-				  buf, len);
+	const uint8_t control =
+		command ? SSD1306_I2C_ALL_BYTES_CMD : SSD1306_I2C_ALL_BYTES_DATA;
+	static uint8_t tmp[SSD1306_I2C_CHUNK_SIZE + 1];
+	int ret = 0;
+	size_t off = 0;
+
+	if (len == 0U) {
+		return 0;
+	}
+
+	k_mutex_lock(&ssd1306_i2c_buffer_lock, K_FOREVER);
+
+	while (off < len) {
+		size_t this_len = MIN(SSD1306_I2C_CHUNK_SIZE, len - off);
+
+		tmp[0] = control;
+		memcpy(&tmp[1], &buf[off], this_len);
+
+		ret = i2c_write_dt(&config->bus.i2c, tmp, this_len + 1);
+		if (ret < 0) {
+			goto out;
+		}
+
+		off += this_len;
+	}
+
+out:
+	k_mutex_unlock(&ssd1306_i2c_buffer_lock);
+	return ret;
 }
 
 static const char *ssd1306_bus_name_i2c(const struct device *dev)


VynDragon (changes requested):
This broke the first ssd1306 display (.96 SSD1315) display I tested (was planning for ~~ssd1306~~ sh1106 and ssd1309 too), what have you used for testing?

TanguyRaufflet:
> Nevermind, it was a new episode of 'GCC cant into compiling' breaking my I2C. After fixing the relevant stupidity, I retested all of them and they all work. however
> 
> ```c
> uint8_t tmp_buf[len + 1];
> ```
> 
> I do not like. Can easily balloon to 1KB.

Thanks for the review !

Yes I agree this is not good.

The simplest solution I can see is to create the table with a fixed size. Does that seem acceptable to you?
If not, I think we will have to make the change on the I3C driver side.

drivers/display/ssd1306.c
```diff
@@ -82,10 +86,37 @@ static int ssd1306_write_bus_i2c(const struct device *dev, uint8_t *buf, size_t
...
+
+	while (off < len) {
+		size_t this_len = MIN(SSD1306_I2C_CHUNK_SIZE, len - off);
+
+		tmp[0] = control;
+		memcpy(&tmp[1], &buf[off], this_len);
+
+		ret = i2c_write_dt(&config->bus.i2c, tmp, this_len + 1);
+		if (ret < 0) {
+			k_mutex_unlock(&ssd1306_i2c_buffer_lock);
```
VynDragon: Could use a goto here and deduplicate that to the end, only suggestion

drivers/display/ssd1306.c
```diff
@@ -82,10 +86,37 @@ static int ssd1306_write_bus_i2c(const struct device *dev, uint8_t *buf, size_t
...
+	static uint8_t tmp[SSD1306_I2C_CHUNK_SIZE + 1];
+	int ret = 0;
+
+	if (len == 0U) {
+		return 0;
+	}
+
+	k_mutex_lock(&ssd1306_i2c_buffer_lock, K_FOREVER);
+
+	size_t off = 0;
```
VynDragon: This should be at the top

VynDragon (changes requested):
Will test, only blocking thing is size_t declaration in the wrong palce.

VynDragon:
Ok tests all good just move the size_t decl.

drivers/display/ssd1306.c
TanguyRaufflet: Done.

drivers/display/ssd1306.c
TanguyRaufflet: Yes indeed, fixed.

VynDragon (dismissed):

VynDragon (approved):

drivers/display/display_ssd1306.c
```diff
@@ -153,6 +155,9 @@ struct ssd1306_data {
 #if (DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(solomon_ssd1306, i2c) || \
 	DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(solomon_ssd1309, i2c) || \
 	DT_HAS_COMPAT_ON_BUS_STATUS_OKAY(sinowealth_sh1106, i2c))
+
+K_MUTEX_DEFINE(ssd1306_i2c_buffer_lock);
```
JarmouniA, line 159: Should be defined per device instance.

JarmouniA (changes requested):

drivers/display/display_ssd1306.c
VynDragon: The buffer is not defined per device instance, it is global...

JarmouniA: Ah I missed that, it shouldn't be static. Won't work in a multi display setup.

VynDragon: Then it shall be a semaphore @TanguyRaufflet

JarmouniA: > Then it shall be a semaphore

Not really, we want to guard access to the I2C bus, no?

But I also think it's only necessary at the level of driver's functions called from the API. Two display controllers using this driver won't share the same I2C bus. What we want is bus access calls for the same device from different threads to wait for each other.

VynDragon: Again, no? Look at the code, it's guarding the static temporary buffer shared between all instances...

JarmouniA: > it's guarding the static temporary buffer

I said above that it shouldn't be shared, why would 2 controllers on different I2C buses share a tmp buffer?

VynDragon: because it saves memory, and yes they can share the same I2C bus.

VynDragon: (at least 2 address provided for every ssd130x model and comaptible on the bus part so they can share a bus)

JarmouniA: > because it saves memory

But it introduces extra latency unnecessarily in case the controllers are on different buses.

VynDragon: What is your suggestion then? 128 bytes on the stack is not great, multiple buffers is better but basically wasted ram in almost all cases, and semaphore / mutex has a cost that would be about one millionth to 1 billionth of the transfer time.

JarmouniA: > What is your suggestion then?

Put the buffer in data struct of devices on I2C bus.

JarmouniA:
> I might take and update this PR if you are not interested in finishing the process, if only because my platform has a i2c size limit of 256 bytes and this change saves me from some shenanigans, while it also is aligned with overall i2c goals, is it okay with you @TanguyRaufflet ?

@VynDragon Please go ahead.

TanguyRaufflet:
> I might take and update this PR if you are not interested in finishing the process, if only because my platform has a i2c size limit of 256 bytes and this change saves me from some shenanigans, while it also is aligned with overall i2c goals, is it okay with you @TanguyRaufflet ?

Hello,

Sorry, I haven't had time to update the pull request. Of course, feel free to do it if I want to.

JarmouniA:
Replaced by #107228

PR #100690, opened 2025-12-08
scripts: build: check_init_priorities: take `zephyr,deferred-init` into account

mathieuchopstm:
A device with `zephyr,deferred-init` is initialized by the application, so we cannot check if it is initialized in "the correct order" (ensuring correct order is the application's responsibility). On the other hand, a non-deferred device *must not* (per my understanding) depend on a deferred-init device: this is guaranteed to be initialized out of order (or is it? Calling `device_init()` from an initializer seems backwards - just don't make the device deferred??)

Fixes #100666

### DIFF
--- a/scripts/build/check_init_priorities.py
+++ b/scripts/build/check_init_priorities.py
@@ -262,6 +262,17 @@ class Validator:
 
         self.errors = 0
 
+    def _flag_error(self, msg):
+        """Remember that a validation error occurred and report to user"""
+        if not self.errors:
+            self.log.error(
+                "Device initialization priority validation failed, "
+                "the sequence of initialization calls does not match "
+                "the devicetree dependencies."
+            )
+        self.errors += 1
+        self.log.error(msg)
+
     def _check_dep(self, dev_ord, dep_ord):
         """Validate the priority between two devices."""
         if dev_ord == dep_ord:
@@ -287,14 +298,7 @@ class Validator:
                 f"{dev_node.path} and {dep_node.path} have the same priority: {dev_prio}"
             )
         elif dev_prio < dep_prio:
-            if not self.errors:
-                self.log.error(
-                    "Device initialization priority validation failed, "
-                    "the sequence of initialization calls does not match "
-                    "the devicetree dependencies."
-                )
-            self.errors += 1
-            self.log.error(
+            self._flag_error(
                 f"{dev_node.path} <{dev_init}> is initialized before its dependency "
                 f"{dep_node.path} <{dep_init}> ({dev_prio} < {dep_prio})"
             )
--- a/scripts/build/check_init_priorities.py
+++ b/scripts/build/check_init_priorities.py
@@ -49,6 +49,11 @@ _IGNORE_COMPATIBLES = frozenset(
     ]
 )
 
+# Deferred initialization property name.
+# When present, the device is not initialized during boot.
+# (it is evaluated as if initializing "after everything else")
+_DEFERRED_INIT_PROP_NAME = "zephyr,deferred-init"
+
 # The offset of the init pointer in "struct device", in number of pointers.
 DEVICE_INIT_OFFSET = 5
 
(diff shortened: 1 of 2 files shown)


scripts/build/check_init_priorities.py
```diff
@@ -287,6 +292,27 @@ def _check_dep(self, dev_ord, dep_ord):
                 self.log.info(f"Ignoring priority: {dev_node._binding.compatible}")
                 return
 
+        def _deferred(node):
+            p = node.props.get(_DEFERRED_INIT_PROP_NAME)
+            if p is None:
+                return False
+            return bool(p.val)
```
fabiobaltieri: this looks a bit odd, it's a boolean property could you just do

```python
if not node.props.get(_DEFERRED_INIT_PROP_NAME):
    return False
return True
```
?

scripts/build/check_init_priorities.py
```diff
@@ -300,7 +326,8 @@ def _check_dep(self, dev_ord, dep_ord):
         elif dev_prio < dep_prio:
             self._flag_error(
                 f"{dev_node.path} <{dev_init}> is initialized before its dependency "
-                f"{dep_node.path} <{dep_init}> ({dev_prio} < {dep_prio})")
+                f"{dep_node.path} <{dep_init}> ({dev_prio} < {dep_prio})"
+            )
```
fabiobaltieri: I think this was meant to go in the first patch

fabiobaltieri (commented):
hey thanks!

scripts/build/check_init_priorities.py
pdgendt: One-liner?

```suggestion
            return node.props.get(_DEFERRED_INIT_PROP_NAME) is not None
```

mathieuchopstm: > this looks a bit odd, it's a boolean property could you just do
> 
> ```python
> if not node.props.get(_DEFERRED_INIT_PROP_NAME):
>     return False
> return True
> ```
> 
> ?

I wasn't sure if bool props at `edtlib` level behaved like in the C API where `DT_NODE_HAS_PROP(boolean_property)` is always true (i.e., `node.props.get()` never returns `None`) or if it is not present when the property is not on the node (i.e., `node.props.get() is not None` == C API `DT_PROP(boolean_property)`)

Reading the docstring ~~which I should have done earlier~~:
https://github.com/zephyrproject-rtos/zephyr/blob/a1970ebd2f428dc472d505d7c7e5eee4774ee7b0/scripts/dts/python-devicetree/src/devicetree/edtlib.py#L1016-L1020

So I guess a mere `node.props.get() is not None` suffices

scripts/build/check_init_priorities.py
mathieuchopstm: Good catch. Fixed.

scripts/build/check_init_priorities.py
mathieuchopstm: Pushed updated version.

pdgendt (dismissed):

scripts/build/check_init_priorities.py
mathieuchopstm: Accidentally hit (I assume) the same issue as CI locally: `node.props.get() is not None` does NOT suffice:

```
ERROR: Device initialization priority validation failed, the sequence of initialization calls does not match the devicetree dependencies.
ERROR: Non-deferred device /clocks/cpusw depends on deferred device /soc/rcc@56028000
ERROR: Non-deferred device /clocks/perck depends on deferred device /soc/rcc@56028000
/soc/rcc@56028000 Property(spec=<PropertySpec zephyr,deferred-init type 'boolean'>, val=False, node=<Node /soc/rcc@56028000 in '/local/zprj/builds/default/zephyr/zephyr.dts.pre', binding /local/zprj/zephyrproject/zephyr/dts/bindings/clock/st,stm32n6-rcc.yaml>)
/soc/rcc@56028000 Property(spec=<PropertySpec zephyr,deferred-init type 'boolean'>, val=False, node=<Node /soc/rcc@56028000 in '/local/zprj/builds/default/zephyr/zephyr.dts.pre', binding /local/zprj/zephyrproject/zephyr/dts/bindings/clock/st,stm32n6-rcc.yaml>)
/clocks/cpusw None
/soc/rcc@56028000 Property(spec=<PropertySpec zephyr,deferred-init type 'boolean'>, val=False, node=<Node /soc/rcc@56028000 in '/local/zprj/builds/default/zephyr/zephyr.dts.pre', binding /local/zprj/zephyrproject/zephyr/dts/bindings/clock/st,stm32n6-rcc.yaml>)
/clocks/perck None
ninja: build stopped: subcommand failed.
```

(obtained by adding `print(f"{node.path} {node.props.get(_DEFERRED_INIT_PROP_NAME)}")` in `_deferred`)

fabiobaltieri (approved):

pdgendt (approved):

PR #100694, opened 2025-12-08
tests: secure_storage: psa: its: patch test scope

bjarki-andreasen:
The current testcase.yaml for the psa its test suite does not filter for targets which are supported, it rather accepts everything then excludes unsupported targets. This currently causes tiny unsupported riscv based targets to be included in the test scope.

Rather than extend the platform_exclude list, which can grow forever, refactor the filtering to whitelist supported targets.

See https://github.com/zephyrproject-rtos/zephyr/pull/100662 for an example of CI trying to run for many nrf54h RV32 cores.

### DIFF
--- a/tests/subsys/secure_storage/psa/its/testcase.yaml
+++ b/tests/subsys/secure_storage/psa/its/testcase.yaml
@@ -2,10 +2,18 @@ common:
   integration_platforms:
     - native_sim
     - nrf54l15dk/nrf54l15/cpuapp
-  platform_exclude:
-    - qemu_cortex_m0 # settings subsystem initialization fails
-    - nrf54h20dk/nrf54h20/cpuapp
-    - nrf54h20dk/nrf54h20/cpurad
+  platform_allow:
+    - native_sim
+    - mps2/an385
+    - qemu_x86/atom
+    - qemu_x86_64/atom
+    - nrf54l15dk/nrf54l15/cpuapp
+    - nrf5340dk/nrf5340/cpuapp
+    - nrf52840dk/nrf52840
+    - nrf9151dk/nrf9151
+    - nrf9160dk/nrf9160
+    - nrf9161dk/nrf9161
+    - ophelia4ev/nrf54l15/cpuapp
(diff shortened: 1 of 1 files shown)


tomi-font (changes requested):
Working with filters and only excluding problematic platforms is the better way IMO. As the documentation says, if you want to reduce testing scope, use `integration_platforms`, not `platform_allow`.  The change you are proposing drastically reduces the board targets the tests could run on, and quite some people wouldn't be happy about this change I bet. (Nordic included because you are forgetting a bunch of board targets that are relevant for these tests.)

I see the failures you are encountering in https://github.com/zephyrproject-rtos/zephyr/pull/100662 are due to 3ea2951bc01c92466a5fed228f0c70aa96dfec72. cc @petejohanson-adi 
Maybe we need a different condition than just `CONFIG_FLASH_PAGE_LAYOUT` for the `#ifdef`?

sylvioalves:
@bjarki-andreasen thanks for bringing this. All `esp32-based` boards with `appcpu` core target fails in this test. So either we add the `ignore-tag` in all those board's `yml` file or it gets filtered in the testcase here.

tomi-font:
FYI related PR: https://github.com/zephyrproject-rtos/zephyr/pull/100783

JarmouniA:
> As the documentation says, if you want to reduce testing scope, use integration_platforms, not platform_allow.

@tomi-font That's not true, you are confusing a full Twister run with a run with `--integration`.

"integration_platforms: This option limits the scope to the listed platforms when twister is invoked **with the --integration**"
https://docs.zephyrproject.org/latest/develop/test/twister.html#tests

`filter`, `depends_on`, `vendor_allow/exclude`, and `min_flash/ram` are what limits the scope in a full run.
(arch_allow, arch_exclude, platform_allow, platform_exclude are all syntactic sugar for `filter` expressions).

tomi-font:
Fix available here: https://github.com/zephyrproject-rtos/zephyr/pull/100835

PR #100695, opened 2025-12-08
stm32: perform stm32n6 RIF configuration in a centralized soc function

avolmat-st:
Currently the RIF configuration of the STM32N6 is performed in each related driver. This requires to add those dedicated call in each driver and moreover this might not always be needed.
Currently Zephyr is running in priviledge mode so is able to perform such configuration and needs to perform configuration this way but this is not always the case especially if Zephyr is not running in secure mode.
For this reason, move all those RIF configuration in a dedicated SOC specific function which is only called if Zephyr is "alone" running on the SOC.
In case of another entity has already run before (such as TF-M), this entity will have already set the RIF prior to reaching Zephyr and Zephyr will most probably not have enough right to set those, in such case the function will not be called.


erwango (dismissed):

soc/st/stm32/stm32n6x/soc.c
```diff
@@ -34,6 +34,46 @@ void soc_reset_hook(void)
...
+		RIMC_MasterConfig_t rimc = {						\
+			.MasterCID = RIF_CID_1,						\
+			.SecPriv = RIF_ATTRIBUTE_SEC | RIF_ATTRIBUTE_PRIV,		\
+		};									\
+		HAL_RIF_RIMC_ConfigMasterAttributes(RIF_MASTER_INDEX_##device, &rimc);	\
+	}
+
+#define RIF_SLAVE_SEC_PRIV(device)	\
+	HAL_RIF_RISC_SetSlaveSecureAttributes(RIF_RISC_PERIPH_INDEX_##device,		\
+					      RIF_ATTRIBUTE_SEC | RIF_ATTRIBUTE_PRIV);
```
mathieuchopstm: Nit: function-like macros are usually written to behave as statements (i.e., you have to place `;` at the end of invocation)
```suggestion
#define RIF_MASTER_CID1_SEC_PRIV(device)	\
	do {										\
		RIMC_MasterConfig_t rimc = {						\
			.MasterCID = RIF_CID_1,						\
			.SecPriv = RIF_ATTRIBUTE_SEC | RIF_ATTRIBUTE_PRIV,		\
		};									\
		HAL_RIF_RIMC_ConfigMasterAttributes(RIF_MASTER_INDEX_##device, &rimc);	\
	} while (0)

#define RIF_SLAVE_SEC_PRIV(device)	\
	HAL_RIF_RISC_SetSlaveSecureAttributes(RIF_RISC_PERIPH_INDEX_##device,		\
					      RIF_ATTRIBUTE_SEC | RIF_ATTRIBUTE_PRIV)
```
(+ update of `soc_rif_config()` to match)

soc/st/stm32/stm32n6x/Kconfig
```diff
@@ -31,9 +32,13 @@ config STM32N6_BOOT_SERIAL
...
-	select USE_STM32_HAL_RIF
 	select RESET
 	default y
 	depends on DT_HAS_ST_STM32_NPU_ENABLED
 
+config STM32N6_RIF_OPEN
+	bool "Configure the RIF with all OPEN access"
+	default y
+	depends on TRUSTED_EXECUTION_SECURE
+
```
mathieuchopstm, line 47: Maybe add a little `help` message?
```
When this option is enabled, RIMC of all masters and RISC of all slaves is configured
during SoC initialization. Masters are configured with CID 1 and Secure Privileged
attributes, and slaves are made accessible to Secure Privileged requests.
```
(not sure if this is a proper explanation, I'm not very familiar with RIF)

mathieuchopstm (dismissed):
LGTM w/ two minor nits

mathieuchopstm (approved):

erwango (approved):

maass-hamburg (approved):

etienne-lms (approved):

PR #100696, opened 2025-12-08
arch: arm: cortex_m: implement usermode thread fault recovery

bjarki-andreasen:
Using CONFIG_USERSPACE, any fault caused by a usermode thread is treated like a supervisor thread fault, halting the system.

Usermode threads should not be able to corrupt the kernel, faults caused by a usermode thread should only terminate the thread, not halt the system.

Implement recovery in fault.c which on fault occuring will check if the current thread is a usermode thread, and if so, whether the fault is one which is caused by the usermode thread and is expected not to be able to corrupt the kernel. If so, simply abort the thread and resume the system.

### DIFF
--- a/arch/arm/core/cortex_m/fault.c
+++ b/arch/arm/core/cortex_m/fault.c
@@ -774,6 +774,46 @@ static void reserved_exception(const struct arch_esf *esf, int fault)
 		      fault < 16 ? "Reserved Exception (" : "Spurious interrupt (IRQ ", fault - 16);
 }
 
+#if defined(CONFIG_USERSPACE)
+static void fault_handle_userspace(uint32_t reason, bool *recoverable)
+{
+	if (*recoverable) {
+		/* Fault is recoverable */
+		return;
+	}
+
+	if ((_current->arch.mode & 1) == 0) {
+		/* Current thread is not usermode thread */
+		return;
+	}
+
+	switch (reason) {
+	/* Faults which are expected to be caused by the usermode thread and which should
+	 * not be able to corrupt the kernel.
+	 */
+	case K_ERR_STACK_CHK_FAIL:
+	case K_ERR_ARM_MEM_STACKING:
+	case K_ERR_ARM_MEM_UNSTACKING:
+	case K_ERR_ARM_MEM_DATA_ACCESS:
+	case K_ERR_ARM_MEM_INSTRUCTION_ACCESS:
+	case K_ERR_ARM_USAGE_DIV_0:
+	case K_ERR_ARM_USAGE_STACK_OVERFLOW:
+		break;
+
+	default:
+		return;
+	}
+
+	PR_FAULT_INFO("  Usermode thread %p aborted", _current);
+
+	/* Usermode thread caused fault. We will abort the usermode thread. */
+	k_thread_abort(_current);
+
+	/* Kernel should not be corrupted, fault was handled and is thus recoverable. */
+	*recoverable = true;
+}
+#endif
+
 /* Handler function for ARM fault conditions. */
 static uint32_t fault_handle(struct arch_esf *esf, int fault, bool *recoverable)
 {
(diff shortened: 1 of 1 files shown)


nashif:
is this cortex_m specific? if this is what we end up doing, we need to generalize this and do the same for all arches supporting userspace. @dcpleung  FYI

bjarki-andreasen:
> Following the reverse flow I see that ...
> 
> 1. fault_handle_userspace() is called by fault_handle()
> 2. fault_handle() is called by z_arm_fault()
> 3. z_arm_fault() is called by any of the following types of faults:  hard, mpu, bus, usage, secure, debug monitor and exc spurious
> 
> My understanding of ARM nuances is admittedly not the best, but these to me appear to be likely unrecoverable. Can the ARM experts provide additional comment?

Not for a usermode thread, since it can't access any critical memory without going through syscalls (by default), so a usage error like stack overflow, or DIV by 0, caused by a usermode thread, should not be an unrecoverable fault to the kernel, just the thread itself, which can be safely aborted without affecting the rest of the system.

bjarki-andreasen:
> > > Following the reverse flow I see that ...
> > > 
> > > 1. fault_handle_userspace() is called by fault_handle()
> > > 2. fault_handle() is called by z_arm_fault()
> > > 3. z_arm_fault() is called by any of the following types of faults:  hard, mpu, bus, usage, secure, debug monitor and exc spurious
> > > 
> > > My understanding of ARM nuances is admittedly not the best, but these to me appear to be likely unrecoverable. Can the ARM experts provide additional comment?
> > 
> > 
> > Not for a usermode thread, since it can't access any critical memory without going through syscalls (by default), so a usage error like stack overflow, or DIV by 0, caused by a usermode thread, should not be an unrecoverable fault to the kernel, just the thread itself, which can be safely aborted without affecting the rest of the system.
> 
> +1, for a usermode thread the faults listed in the PR (stack check, stacking,unstacking etc.) catch any corruption outside the user stack before it happens (assuming MPU/Stack guards are configured) so terminating the thread and continuing with kernel and any other working thread should be ok. However, this still needs care. If the thread in question changes the system behavior in a way that the system can't function without it then the system becomes unpredictable. e.g. if a main thread controlling a state machine or actions of other threads overflows and gets killed without rebooting the system then the resulting behavior is unknown. Existing systems not having considered this as part of their design may end up badly with this change so it would be safer to put this behind a config option for now, so existing systems can opt in when ready.

Definately agree that the change in behavior in this PR is a "breaking" change, since what before forced a system reset now can continue operating. Adding a config like `config RECOVER_USERMODE_FAULT` or something is a good and neccesary option.

To preserve the previous behavior, a user could also declare the usermode thread K_ESSENTIAL, this will result in usermode thread fault safely caught, thread is aborted, system is reset as thread was essential. Though this is not exactly what K_ESSENTIAL means, since manually aborting the usermode thread now would cause the system to reset where it did not before...

henrikbrixandersen:
> Definately agree that the change in behavior in this PR is a "breaking" change, since what before forced a system reset now can continue operating. Adding a config like config RECOVER_USERMODE_FAULT or something is a good and neccesary option.
> 
> To preserve the previous behavior, a user could also declare the usermode thread K_ESSENTIAL, this will result in usermode thread fault safely caught, thread is aborted, system is reset as thread was essential. Though this is not exactly what K_ESSENTIAL means, since manually aborting the usermode thread now would cause the system to reset where it did not before...

Another option could be to keep the current behavior, _unless_ some new flag was set on the user mode thread, essentially making the new behavior opt-in per usermode thread.

bjarki-andreasen:
> The original design is to let the application decide what to do in this situation by implementing `k_sys_fatal_error_handler()`. The default implementation is simply to halt the system, but the application can override to not do that. The thread is being terminated either way.

letting the application control the flow of kernel faults seems quite dangerous and non portable, we are passing the stack frame and an error, its not trivial to check that an exception was caused by a usermode thread, or whether that specific fault can be safely handled, not that I'm against the optional hook, but we should have some portable default handling of failed usermode threads IMO. Otherwise why have K_ESSENTIAL, that could also just be up to the application to handle?

PR #100700, opened 2025-12-08
sys: getopt: drop the gcd definition

fabiobaltieri:
This is available in util.h since 646f2553733 and the existing one creates a collision and build breakage. Just drop it.

--

Tested with `west build -p -b native_sim tests/net/wifi/configs -T wifi.build.crypto_none`


nashif (approved):

kartben (dismissed):
It rather looks like the main issue is that the linked commit was merged despite not following our namespacing/prefixing rules, no? They should be `sys_...`

PR #100703, opened 2025-12-08
USB class: support multiple USB host controllers

josuah:
Dependencies:

- https://github.com/zephyrproject-rtos/zephyr/pull/94590

Do not bind an USB host controller context during init.
Instead only use the `c_data->udev = udev;` assignment that happens in `probe()` to give the class the USB context.

This means that class can be used by a controller or another depending on which side the USB device is connected, and this shares resources between the ports.

### DIFF
--- a/include/zephyr/usb/usbh.h
+++ b/include/zephyr/usb/usbh.h
@@ -98,8 +98,7 @@ struct usbh_class_filter {
  */
 struct usbh_class_api {
 	/** Host initialization handler, before any device is connected */
-	int (*init)(struct usbh_class_data *const c_data,
-		    struct usbh_context *const uhs_ctx);
+	int (*init)(struct usbh_class_data *const c_data);
 	/** Request completion handler */
 	int (*completion_cb)(struct usbh_class_data *const c_data,
 			     struct uhc_transfer *const xfer);
@@ -121,8 +120,6 @@ struct usbh_class_api {
 struct usbh_class_data {
 	/** Name of the USB host class instance */
 	const char *name;
-	/** Pointer to USB host stack context structure */
-	struct usbh_context *uhs_ctx;
 	/** Pointer to USB device this class is used for */
 	struct usb_device *udev;
 	/** First interface number or claimed function */
--- a/subsys/usb/host/usbh_class.c
+++ b/subsys/usb/host/usbh_class.c
@@ -14,11 +14,9 @@
 
 LOG_MODULE_REGISTER(usbh_class, CONFIG_USBH_LOG_LEVEL);
 
-void usbh_class_init_all(struct usbh_context *const uhs_ctx)
+void usbh_class_init_all(void)
 {
 	int ret;
-
-	usbh_host_lock(uhs_ctx);
 
 	STRUCT_SECTION_FOREACH(usbh_class_node, c_node) {
 		struct usbh_class_data *const c_data = c_node->c_data;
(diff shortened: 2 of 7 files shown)


subsys/usb/host/usbh_class.c
```diff
@@ -14,12 +14,10 @@
 
 LOG_MODULE_REGISTER(usbh_class, CONFIG_USBH_LOG_LEVEL);
 
-void usbh_class_init_all(struct usbh_context *const uhs_ctx)
+void usbh_class_init_all(void)
 {
 	int ret;
 
-	usbh_host_lock(uhs_ctx);
-
```
josuah, line 22: I imagine no locking at all is acceptable given this is something called in a loop inside `uhs_pre_init()` put in `SYS_INIT()`, so no concurrency in the way it is called?

include/zephyr/usb/usbh.h
```diff
@@ -98,8 +98,7 @@ struct usbh_class_filter {
  */
 struct usbh_class_api {
 	/** Host initialization handler, before any device is connected */
-	int (*init)(struct usbh_class_data *const c_data,
-		    struct usbh_context *const uhs_ctx);
+	int (*init)(struct usbh_class_data *const c_data);
```
jfischer-no, line 101: > Only bind to a controller after probing.

Something like `c_data->uhs_ctx = udev->ctx;` is missed here
https://github.com/zephyrproject-rtos/zephyr/blob/d6c47f2beb2b9e32db9066104f7538e5dfe6629b/subsys/usb/host/usbh_class.c#L128-L130

Also, c_data->uhs_ctx seems to be redundant.

tests/subsys/usb/host/src/class.c
```diff
@@ -38,8 +38,7 @@ static struct usbh_foo_priv usbh_foo_priv = {
 	.state = FOO_CLASS_PRIV_INACTIVE,
 };
 
-static int usbh_foo_init(struct usbh_class_data *const c_data,
-			 struct usbh_context *const uhs_ctx)
+static int usbh_foo_init(struct usbh_class_data *const c_data)
```
jfischer-no, line 41: usbh_uvc_init() is in the tree, and must be fixed as well. Actually, UVC test build in twister should fail.

josuah: Yes it does now.
```
usb.uvc on native_sim/native error (Build failure - error: initialization of int (*)(struct usbh_class_data * const) from incompatible pointer type int (*)(struct usbh_class_data * const,  struct usbh_context * const) [-Werror=incompatible-pointer-types])
```

include/zephyr/usb/usbh.h
josuah: I tried to remove the redundacy by removing the `uhs_ctx` field, available from `udev->ctx`.

jfischer-no: We may need a helper to get the host support context, like 
```diff
diff --git a/subsys/usb/host/usbh_class.h b/subsys/usb/host/usbh_class.h
index 0b3d673c058..eb005b0922e 100644
--- a/subsys/usb/host/usbh_class.h
+++ b/subsys/usb/host/usbh_class.h
@@ -60,4 +60,13 @@ void usbh_class_probe_device(struct usb_device *const udev);
  */
 void usbh_class_remove_all(struct usb_device *const udev);
 
+static inline struct usbh_context *usbh_class_get_ctx(struct usbh_class_data *const c_data)
+{
+	if (c_data->udev != NULL) {
+		return c_data->udev->ctx;
+	}
+
+	return NULL;
+}
+
 #endif /* ZEPHYR_INCLUDE_USBH_CLASS_H */
```

jfischer-no (dismissed):

include/zephyr/usb/usbh.h
josuah: Added and rebased on top of latest `main`.

Edit: that was not the intent to add it *now*, so removed it in order to keep it for later.

jfischer-no (approved):

tmon-nordic (approved):

PR #100709, opened 2025-12-09
tests: riscv: pmp: Make permission change tests dynamic

fsammoura1980:
The existing PMP permission tests relied on hardcoded region indices (e.g., index 0 or 2). This makes the tests fragile and dependent on the exact memory attribute region setup.

This change refactors the tests to dynamically determine region counts and indices based on the output of `mem_attr_get_regions()`:

-   `test_pmp_change_perm_invalid_permission`: Uses the last valid region index to test invalid permission flags.
-   `test_pmp_change_perm_invalid_region_index`: Uses the total number of regions as the out-of-bounds index.
-   `test_successful_permission_change`: Locates the index of the target test region (`dt_regions[0]`) by matching its base address and size. The test now also inverts the X bit of the current permissions to ensure a change is applied and verified.

These changes make the tests more robust and less dependent on a specific static configuration.

### DIFF
--- a/tests/arch/riscv/pmp/mem-attr-entries/src/main.c
+++ b/tests/arch/riscv/pmp/mem-attr-entries/src/main.c
@@ -127,33 +127,53 @@ ZTEST(riscv_pmp_memattr_entries, test_dt_pmp_perm_conversion)
 
 ZTEST(riscv_pmp_memattr_entries, test_pmp_change_perm_invalid_permission)
 {
+	const struct mem_attr_region_t *region;
+	size_t num_regions = mem_attr_get_regions(&region);
+	size_t idx = num_regions - 1;
+
 	/* Invalid permission: 0x08 (bit 3) is outside of R|W|X (0x7) mask */
-	zassert_equal(z_riscv_pmp_change_permissions(0, 0x08), -EINVAL,
+	zassert_equal(z_riscv_pmp_change_permissions(idx, 0x08), -EINVAL,
 		      "Should return -EINVAL for invalid permission bits.");
 }
 
 ZTEST(riscv_pmp_memattr_entries, test_pmp_change_perm_invalid_region_index)
 {
-	/* region_idx = 2, which is out of bounds */
-	size_t idx = 2;
+	const struct mem_attr_region_t *region;
+	size_t num_regions = mem_attr_get_regions(&region);
+	size_t invalid_idx = num_regions;
 
-	zassert_equal(z_riscv_pmp_change_permissions(idx, PMP_R), -EINVAL,
+	zassert_equal(z_riscv_pmp_change_permissions(invalid_idx, PMP_R), -EINVAL,
 		      "Should return -EINVAL for region index out of bounds.");
 }
 
 ZTEST(riscv_pmp_memattr_entries, test_successful_permission_change)
 {
-	const uint8_t new_perm = 0x03; /* Read (0x1) | Write (0x2) */
-	const size_t idx = 0;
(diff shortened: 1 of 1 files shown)


tests/arch/riscv/pmp/mem-attr-entries/src/main.c
```diff
@@ -127,33 +127,53 @@ ZTEST(riscv_pmp_memattr_entries, test_dt_pmp_perm_conversion)
...
+		if ((region[idx].dt_addr == dt_regions[0].base) &&
+		    (region[idx].dt_size == dt_regions[0].size)) {
+			break;
+		}
+	}
 
-	zassert_not_equal(new_perm, dt_regions[idx].perm,
-			  "Ensure new permission is different from existing permission");
+	zassert_not_equal(idx, num_regions,
+			  "No matching memory attribute region found for dt_regions[1].");
```
jimmyzhe: Should be dt_regions[0] here.

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
```diff
@@ -127,33 +127,53 @@ ZTEST(riscv_pmp_memattr_entries, test_dt_pmp_perm_conversion)
...
 
-	int ret = z_riscv_pmp_change_permissions(idx, new_perm);
+	/* Store original permission to revert later if needed, and calculate new permission */
+	uint8_t original_perm = dt_regions[idx].perm;
```
jimmyzhe: Should be `dt_regions[0].perm`?

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
```diff
@@ -127,33 +127,53 @@ ZTEST(riscv_pmp_memattr_entries, test_dt_pmp_perm_conversion)
...
+	uint8_t new_perm = 0x04 ^ original_perm; /* Toggle X bit */
```
jimmyzhe: We can use `PMP_X` here.

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
fsammoura1980: Done

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
fsammoura1980: Done

tests/arch/riscv/pmp/mem-attr-entries/src/main.c
fsammoura1980: Done

jimmyzhe (commented):
> * `test_successful_permission_change`: Locates the index of the target test region (`dt_regions[1]`) by matching its base address and size. The test now also inverts the R/W bits of the current permissions to ensure a change is applied and verified.

Seems that `test_successful_permission_change` uses test region `dt_regions[0]` and inverts the X bit. Could you update the commit message and the comment?

Other than that, these changes LGTM.

fsammoura1980:
> > * `test_successful_permission_change`: Locates the index of the target test region (`dt_regions[1]`) by matching its base address and size. The test now also inverts the R/W bits of the current permissions to ensure a change is applied and verified.
> 
> Seems that `test_successful_permission_change` uses test region `dt_regions[0]` and inverts the X bit. Could you update the commit message and the comment?
> 
> Other than that, these changes LGTM.

Done

jimmyzhe (approved):

fkokosinski (approved):

PR #100711, opened 2025-12-09
dts: arm: nxp: rt700_cpu0: add uuid support

lucien-nxp:
SYSCON0 IP can be accessed by CPU0, HiFi4, EZH-V, eDMA0, eDMA1. 
So cm33_cpu1 can't be supported this feature.
Test hwinfo/api case passed on cm33_cpu0 core


Holt-Sun (approved):

ZhaoxiangJin (approved):
I am ok with this, but I think it would be better to post the test results in the PR description after enabling a certain feature for the board/SoC and conducting tests.

butok (approved):

mmahadevan108 (approved):

PR #100717, opened 2025-12-09
drivers: input: Add sleep-mode-enable property for Pinnacle

petejohanson:
Add a new sleep-mode-enable property for the Cirque Pinnacle input driver, to enable the sleep-mode for those peripherals, which will go into a lower power state after 5 seconds with no fingers detected.

### DIFF
--- a/dts/bindings/input/cirque,pinnacle-common.yaml
+++ b/dts/bindings/input/cirque,pinnacle-common.yaml
@@ -44,6 +44,12 @@ properties:
       taps. When set to 0, no empty packets are sent. Supported values from 0
       to 255.
 
+  sleep-mode-enable:
+    type: boolean
+    description: |
+      Enable sleep mode, allowing the hardware to enter a lower power mode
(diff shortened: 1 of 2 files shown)


dts/bindings/input/cirque,pinnacle-common.yaml
```diff
@@ -44,6 +44,12 @@ properties:
       taps. When set to 0, no empty packets are sent. Supported values from 0
       to 255.
 
+  sleep-mode-enable:
+    type: boolean
+    description: |
+      Enable sleep mode, allowing the hardware to enter a  lower power mode
```
dsseng:
```suggestion
      Enable sleep mode, allowing the hardware to enter a lower power mode
```

dsseng (commented):
nitpick, non-blocking

fabiobaltieri (approved):

dsseng (approved):

kartben (approved):

PR #100720, opened 2025-12-09
soc: nordic: move IronSide SE interface code to hal_nordic

jonathannilsen:
Move most of the code that is used to interface with IronSide SE on the
nRF54H20/nRF9280 from the soc/nordic directory to the hal_nordic
repository. The interface code is now provided by the new IronSide
support package. Build system code and glue code that makes use of
Zephyr APIs is now located in the modules/hal_nordic directory.

Update the directory path for IronSide SE interface code in
MAINTAINERS.yml to match the move from soc/nordic/ironside to
modules/hal_nordic/ironside.

Also included are some refactoring changes and cleanup to match the new
supporting code.

C code and Kconfigs have been renamed to *ironside_se* / *IRONSIDE_SE*
to match the supporting code changes. Users of these APIs in zephyr
have been updated to match.

Individual configurations for different "IronSide services" have been
removed as the API serialization for all of these is now provided in
a single C file / header file (ironside/se/api.h).

The ironside_boot_report_get() API has been removed. The boot report
structure can be accessed through the IRONSIDE_SE_BOOT_REPORT macro.

Most configs relating to UICR / PERIPHCONF have been moved under the
"IronSide SE" menu to make it clear that these are part of the
IronSide SE interface.

The macros that in uicr.h that were used to add entries to the
PERIPHCONF section have been removed. The supporting code now provides
PERIPHCONF_XYZ() macros that can be used to initialize structures that
go into this section, and the zephyr part now only contains a macro
UICR_PERIPHCONF_ENTRY() that is used to place an arbitrary structure
into the section. The gen_periphconf_entries.py script used to generate
PERIPHCONF entries based on devicetree has been updated to use the new
macro system.

IronSide SE integration code/configs is now guarded by HAS_IRONSIDE_SE.

Note that the UICR build system integration that relies on Sysbuild
remains under the soc/nordic directory for now, but will be moved to
the modules/hal_nordic directory once it is clear how the integration
will look there.


modules/hal_nordic/ironside/se/CMakeLists.txt
```diff
@@ -1,6 +1,26 @@
 # Copyright (c) 2025 Nordic Semiconductor ASA
 # SPDX-License-Identifier: Apache-2.0
 
+# The IronSide source directory can be overridden by setting
+# IRONSIDE_SUPPORT_DIR before invoking the build system.
+zephyr_get(IRONSIDE_SUPPORT_DIR SYSBUILD GLOBAL)
```
hakonfam, line 6: Why does this need to be a SYSBUILD variable? Wouldn't `zephyr_get(IRONSIDE_SUPPORT_DIR)` suffice?

modules/hal_nordic/ironside/se/CMakeLists.txt
```diff
@@ -1,6 +1,26 @@
...
+  set(IRONSIDE_SUPPORT_DIR
+    ${ZEPHYR_CURRENT_MODULE_DIR}/ironside CACHE PATH "IronSide Support Directory"
+  )
+endif()
+
+zephyr_include_directories(include)
+zephyr_include_directories(${IRONSIDE_SUPPORT_DIR}/se/include)
+
+zephyr_library()
+zephyr_library_property(ALLOW_EMPTY TRUE)
```
hakonfam, line 17: why is this needed?

hakonfam (commented):
Need more time to go through in more detail, but some initial nits. I agree on the overall direction.

modules/hal_nordic/ironside/se/CMakeLists.txt
jonathannilsen: Tbh I just copied the setup used by nrfx here:
https://github.com/zephyrproject-rtos/zephyr/blob/923ece1eea7d303d06b617bb85159832c6b9c12e/modules/hal_nordic/nrfx/CMakeLists.txt#L6-L12

I'll look into whether it is needed.

modules/hal_nordic/ironside/se/CMakeLists.txt
jonathannilsen: It was added to avoid a warning whenever the library ends up without source files due to none of the kconfig options being enabled. That is still a valid usage of the lib from what I can tell (e.g. if you are only using the headers)

hakonfam (dismissed):
The changes look good. Pulling this out to hal_nordic is a great improvement.

kyberdin (approved):

karstenkoenig (dismissed):

hakonfam (dismissed):

carlescufi (dismissed):

karstenkoenig (approved):

hakonfam (approved):

anangl (approved):

PR #100721, opened 2025-12-09
scripts: footprint: size_report: Close ELF file

pdgendt:
The ELFFile class keep a file stream open, an explicit close needs to be called before exiting.


kartben (changes requested):
ELFFile defines `__enter__()` and `__exit__()` so you should be able to use "with...." and a context manager

kartben (approved):

nashif (approved):

fabiobaltieri (approved):

PR #100722, opened 2025-12-09
Techdhaba test

shweatank:
sample code and changes for zephyr for stm32f407


kartben:
> Hi @kartben purpose of these commits to create more sample projects so beginners , it will be helpful for them to begin with.

I see. Maybe I would recommend opening an issue first to iterate on what you think should be samples we should be adding, and agree on this first. Here it, to be frank, it looks like you just threw some ideas at your LLM of choice and opened a pull request before it was really ready for review. 
Do you have a sense of what type of basic samples you think we should be adding?

PR #100723, opened 2025-12-09
modules: hal_nxp: Fix spc module selection on MCXW SoC family

yeaissa:
Fix wrong condition for spc selection on MCXW family.


ZhaoxiangJin (approved):
thanks for the fix

Holt-Sun (approved):

axelnxp (approved):

modules/hal_nxp/mcux/mcux-sdk-ng/drivers/drivers.cmake
```diff
@@ -191,7 +191,7 @@ if(CONFIG_SOC_FAMILY_MCXN OR CONFIG_SOC_FAMILY_MCXA)
   set(CONFIG_MCUX_COMPONENT_driver.mcx_spc ON)
 endif()
 
-if(CONFIG_DT_HAS_NXP_SPC_ENABLED AND SOC_FAMILY_MCXW)
+if(CONFIG_DT_HAS_NXP_SPC_ENABLED AND CONFIG_SOC_FAMILY_MCXW)
```
mmahadevan108, line 194: Why do we have a conditional on the family. The Binding should be enabled in the appropriate soc dts file and then we do not need this change. 
@ZhaoxiangJin why was this changed to the current implementation?

ZhaoxiangJin: originally logic is: `set_variable_ifdef(CONFIG_DT_HAS_NXP_SPC_ENABLED CONFIG_MCUX_COMPONENT_driver.spc)`. 
MCXA/MCXN now also need to enable SPC in the devicetree, so `CONFIG_DT_HAS_NXP_SPC_ENABLED` is also true for MCXA/MCXN. Based on the previous logic, `CONFIG_MCUX_COMPONENT_driver.spc` driver will also be pulled for MCXA/MCXN, but in the SDK, MCXA/MCXN require `set(CONFIG_MCUX_COMPONENT_driver.mcx_spc ON)` instead. Therefore, control logic needs to be added to `set(CONFIG_MCUX_COMPONENT_driver.spc ON)`.

PR #100724, opened 2025-12-09
doc: net: Enhance network packet filtering documentation

### LINKED ISSUE #100464: npf_ip_src_addr_match() is using packet address family before it is set
### Describe the bug

I'm trying network packet filtering and `NPF_IP_SRC_ADDR_ALLOWLIST` but struggle with it. After debugging a bit, I don't understand how this is supposed to work.

Ethernet packets enter `net_recv_data()`, which calls `net_pkt_filter_recv_ok()` to see if the packet should be accepted or dropped. If accepted, the packet is passed to `net_queue_rx()`, where it eventually ends up in `ethernet_ip_recv()`, which calls `net_pkt_set_family()` on it. Up until then, it is AF_UNSPEC.

One of the filters that can be called by `net_pkt_filter_recv_ok()` is `npf_ip_src_addr_match()`. It looks at the address family to see which address filter to use. But the address family is not set until later, so the address filter is never applied.

Here's the call graph:

```c
net_recv_data()
 net_pkt_filter_recv_ok()
  lock_evaluate()
   evaluate()
    apply_tests()
     npf_ip_src_addr_match()
     {
        uint8_t pkt_family = net_pkt_family(pkt);
 net_queue_rx()
  net_process_rx_packet()
   net_rx()
    processing_data()
     process_data()
      net_if_recv_data()
       ethernet_recv()
        ethernet_ip_recv()
        {
          net_pkt_set_family(pkt, AF_INET6);

```

We have tests for the filters in tests/net/npf but they test the filter with a fully populated net_pkt struct. It appears we don't test passing the packet through the network stack.

### Regression

- [ ] This is a regression.

### Steps to reproduce

```c
struct in6_addr ipv6_address = {.s6_addr = { 0xaa, 0xbb, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x1 }};
NPF_IP_SRC_ADDR_ALLOWLIST(allowlist_ipv6_src_addr, (void *)&ipv6_address, 1, AF_INET6);
NPF_RULE(ipv6_allowlist, NET_OK, allowlist_ipv6_src_addr);

npf_append_recv_rule(&ipv6_allowlist);
npf_append_recv_rule(&npf_default_drop);
```

### Relevant log output

```shell

```

### Impact

### Environment

Zephyr v4.2.1, but I don't see any relevant changes since in 4.3 or master.

Tested on STM32 hardware.

### Additional Context

_No response_

jukkar:
Add more information to network packet filter documentation to make it more useful to end users.

Fixes #100464

### DIFF
--- a/doc/connectivity/networking/api/net_pkt_filter.rst
+++ b/doc/connectivity/networking/api/net_pkt_filter.rst
@@ -38,8 +38,22 @@ A rule is represented by a :c:struct:`npf_rule` object. It can be inserted to,
 appended to or removed from a rule list contained in a
 :c:struct:`npf_rule_list` object using :c:func:`npf_insert_rule()`,
 :c:func:`npf_append_rule()`, and :c:func:`npf_remove_rule()`.
-Currently, two such rule lists exist: ``npf_send_rules`` for outgoing packets,
-and ``npf_recv_rules`` for incoming packets.
+
+There are different sets of rules for different layers in the network stack.
+Some of the rules are applied for L2 layer like Ethernet, and some for L3 layer
+where IPv4 or IPv6 protocol is handled. The ``local_in`` rules are for matching
+incoming protocol types like UDP or TCP packets that are run on top of IPv4 or IPv6.
+The rule support for different layers can be controlled by relevant Kconfig options
+mentioned below.
+
+* ``npf_send_rules`` is a rule list applied to outgoing packets in L2 layer
+* ``npf_recv_rules`` is a rule list applied to incoming packets in L2 layer
+* ``npf_ipv4_recv_rules`` is a rule list applied for incoming IPv4 packets. Can be
+  enabled or disabled by :kconfig:option:`CONFIG_NET_PKT_FILTER_IPV4_HOOK` option.
+* ``npf_ipv6_recv_rules`` is a rule list applied for incoming IPv6 packets. Can be
+  enabled or disabled by :kconfig:option:`CONFIG_NET_PKT_FILTER_IPV6_HOOK` option.
+* ``npf_local_in_recv_rules`` is a rule list applied for incoming UDP or TCP packets.
+  Can be enabled or disabled by :kconfig:option:`CONFIG_NET_PKT_FILTER_LOCAL_IN_HOOK` option.
 
 If a filter rule list is empty then ``NET_OK`` is assumed. If a non-empty
 rule list runs to the end then ``NET_DROP`` is assumed. However it is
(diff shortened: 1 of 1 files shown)


doc/connectivity/networking/api/net_pkt_filter.rst
```diff
@@ -56,6 +70,11 @@ to statically define condition instances for various conditions, and
 :c:macro:`NPF_RULE()` and :c:macro:`NPF_PRIORITY()` to create a rule instance
 with an immediate outcome or a priority change.
 
+See also :zephyr:code-sample:`net-pkt-filter` sample for an example how to create and
+manage packet filters. The network shell has a ``net filter`` command that can be used
+to see the installed rules at runtime.
+
```
rlubos, line 76: nit: extra empty line

rlubos (dismissed):

doc/connectivity/networking/api/net_pkt_filter.rst
```diff
@@ -38,8 +38,22 @@ A rule is represented by a :c:struct:`npf_rule` object. It can be inserted to,
 appended to or removed from a rule list contained in a
 :c:struct:`npf_rule_list` object using :c:func:`npf_insert_rule()`,
 :c:func:`npf_append_rule()`, and :c:func:`npf_remove_rule()`.
-Currently, two such rule lists exist: ``npf_send_rules`` for outgoing packets,
-and ``npf_recv_rules`` for incoming packets.
+
+There are different set of rules for different layers in the network stack.
```
kartben:
```suggestion
There are different sets of rules for different layers in the network stack.
```

doc/connectivity/networking/api/net_pkt_filter.rst
```diff
@@ -56,6 +70,11 @@ to statically define condition instances for various conditions, and
 :c:macro:`NPF_RULE()` and :c:macro:`NPF_PRIORITY()` to create a rule instance
 with an immediate outcome or a priority change.
 
+See also :zephyr:code-sample:`net-pkt-filter` sample for an example how to create and
```
kartben:
```suggestion
See also :zephyr:code-sample:`net-pkt-filter` sample for an example of how to create and
```

kartben (dismissed):

rlubos (approved):

kartben (approved):

zagor:
> I added a link to the pkt_filter sample which contains examples of L3 rules and handling.

Does it? It only calls `npf_append_recv_rule()`, which adds a rule to the `npf_recv_rules` L2 rule list doesn't it?

I think we need to show that the other rule lists exists, and how to use them. And perhaps especially how to use them in combination, which is what I'm struggling with right now. (Such as: _"Allow IPv6 from address X::1 and IPv4 from interface Y, but drop everything else."_)

jukkar:
> Does it? It only calls `npf_append_recv_rule()`, which adds a rule to the `npf_recv_rules` L2 rule list doesn't it?
> 
> I think we need to show that the other rule lists exists, and how to use them. And perhaps especially how to use them in combination, which is what I'm struggling with right now. (Such as: _"Allow IPv6 from address X::1 and IPv4 from interface Y, but drop everything else."_)

I will check and add something related to this then. But as a general note, would you as a end user be ok with this (pointing to the sample) or would you prefer to see examples in inline documentation?

PR #100725, opened 2025-12-09
sys: util: add sys_ prefix to SIGN, gcd, and lcm utils

kartben:
As per our naming conventions for public symbols, these utility functions and macros should be prefixed with "sys_".

These were introduced with https://github.com/zephyrproject-rtos/zephyr/pull/98875 with a lack of prefix.

### DIFF
--- a/include/zephyr/sys/util.h
+++ b/include/zephyr/sys/util.h
@@ -1036,7 +1036,7 @@ static inline size_t sys_count_bits(const void *value, size_t len)
  * @retval -1 if x is negative
  * @retval 0 if x is zero
  */
-#define SIGN(x) (((x) > 0) - ((x) < 0))
+#define SYS_SIGN(x) (((x) > 0) - ((x) < 0))
 
 /**
  * @brief Compute the Greatest Common Divisor (GCD) of two integers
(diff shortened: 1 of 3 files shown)


aescolar (approved):

aescolar:
By the way, this fix is needed to be able to build https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/usb_c/vbus/usbc_vbus_adc.c#L99

carlescufi (approved):

ngphibang (approved):

fabiobaltieri:
> > However, should we do the same for all other macros in `util.h`, such as MIN/MAX(), min/max, min3/max3(), DIV_ROUND_CLOSEST(), etc. it seems all of them don't have a SYS_/sys_ prefix ?

Yeah I don't think that'd be a good idea. In general, I don't think we should blindly follow these rules without applaying common sense first, some of these macros are well used in Linux and we have a lot of developers working cross-project, having different convention just makes it harder for them carry over and add distractions when doing code reviews.

Probably no big deal for this one, there is a linux counterpart but it's not as commonly used as min/max though honestly I would leave it as is.

henrikbrixandersen (approved):

mathieuchopstm (approved):

nashif (approved):

PR #100728, opened 2025-12-09
net: ip: net_context: Use sin6_scope_id for link-local IPv6 addresses

Cristib05:
This commit aims to make use of sin6_scope_id field of net_sockaddr_in6 structure to correctly choose the network interface when destination address in link-local.

### DIFF
--- a/subsys/net/ip/net_context.c
+++ b/subsys/net/ip/net_context.c
@@ -944,6 +944,12 @@ int net_context_bind(struct net_context *context, const struct net_sockaddr *add
 		} else {
 			struct net_if_addr *ifaddr;
 
+			if (net_ipv6_is_ll_addr(&addr6->sin6_addr)) {
+				if (iface == NULL) {
+					iface = net_if_get_by_index(addr6->sin6_scope_id);
+				}
+			}
+
 			ifaddr = net_if_ipv6_addr_lookup(
 					&addr6->sin6_addr,
 					iface == NULL ? &iface : NULL);
@@ -2470,6 +2476,17 @@ static int context_sendto(struct net_context *context,
 				IF_ENABLED(CONFIG_NET_IPV6,
 					   (iface = net_if_get_by_index(
 						   context->options.ipv6_mcast_ifindex)));
+			}
+
+			if (net_ipv6_is_ll_addr(&addr6->sin6_addr) &&
+			    !net_context_is_bound_to_iface(context) &&
+			    COND_CODE_1(CONFIG_NET_IPV6,
+					(addr6->sin6_scope_id > 0), (false))) {
+				IF_ENABLED(CONFIG_NET_IPV6, (
+					   iface = net_if_get_by_index(addr6->sin6_scope_id)));
+				if (iface != NULL) {
+					net_context_set_iface(context, iface);
+				}
 			}
 		}
 


subsys/net/ip/net_context.c
```diff
@@ -2471,6 +2471,12 @@ static int context_sendto(struct net_context *context,
 					   (iface = net_if_get_by_index(
 						   context->options.ipv6_mcast_ifindex)));
 			}
+			if (net_ipv6_is_ll_addr(&addr6->sin6_addr) &&
```
rlubos, line 2481: Shouldn't `sin6_scope_id` be handled for any context type, why is it limited to UDP only?

Cristib05: It should, I have misplaced my code. I guess now it's ok. One question would be if it is necessary to call `net_context_set_iface(context, iface);` whenever `iface` is set based on a particular case. I've seen that further down in `context_sendto`,  `iface` is set by calling `iface = net_context_get_iface(context);` Do you think there could be cases where `iface` could be set to something other than `context->iface`?

rlubos: > It should, I have misplaced my code. I guess now it's ok.

It still seems under `if (IS_ENABLED(CONFIG_NET_UDP) && net_context_get_type(context) == NET_SOCK_DGRAM)` or am I missing something?

> One question would be if it is necessary to call net_context_set_iface(context, iface); whenever iface is set based on a particular case. I've seen that further down in context_sendto, iface is set by calling iface = net_context_get_iface(context); 

Actually, that's a good point, below we check if the interface assigned to the context is up, so we should update the interface with `net_context_set_iface()`:
```
	iface = net_context_get_iface(context);
	if (iface && !net_if_is_up(iface)) {
		return -ENETDOWN;
	}
```

> Do you think there could be cases where iface could be set to something other than context->iface?

Now that I think, there could be the case that the context is bound to a specific interface with `SO_BINDTODEVICE` socket option. I don't think we should override that. So before we update the interface based on `sin6_scope_id`, I think we should check with `net_context_is_bound_to_iface()` whether the context (socket) was assigned a specific interface by the application.

jukkar: For TCP the checks need to be in `net_context_bind()` because in `context_sendto()` we have already bound the TCP socket to the interface and cannot change it.

rlubos (approved):

subsys/net/ip/net_context.c
```diff
@@ -2473,6 +2473,16 @@ static int context_sendto(struct net_context *context,
 			}
 		}
 
+		if (net_ipv6_is_ll_addr(&addr6->sin6_addr) &&
+		    !net_context_is_bound_to_iface(context) &&
+		    COND_CODE_1(CONFIG_NET_IPV6,
+				(addr6->sin6_scope_id > 0), (false))) {
+			IF_ENABLED(CONFIG_NET_IPV6, ({
+				iface = net_if_get_by_index(addr6->sin6_scope_id);
```
rlubos: Or actually, we should check if `iface != NULL` before setting it on the interface, in case application provided invalid `sin6_scope_id`

rlubos (changes requested):

subsys/net/ip/net_context.c
Cristib05: Based on @jukkar's comment, I've added code in `net_context_bind()` function, and in `context_sendto()` function, I've added the new code under `UDP` case. Please let me know if this is correct or not. 
I am not sure if it's worth investing time on tests for this new addition. If you think it is, please let me know.

subsys/net/ip/net_context.c
```diff
@@ -944,6 +944,12 @@ int net_context_bind(struct net_context *context, const struct net_sockaddr *add
 		} else {
 			struct net_if_addr *ifaddr;
 
+			if (net_ipv6_is_ll_addr(&addr6->sin6_addr)) {
+				if (iface == NULL) {
```
rlubos, line 948: Should we also check if `addr6->sin6_scope_id` is a valid interface ID, so `addr6->sin6_scope_id > 0`?

subsys/net/ip/net_context.c
```diff
@@ -2471,8 +2477,20 @@ static int context_sendto(struct net_context *context,
...
+					(addr6->sin6_scope_id > 0), (false))) {
+				IF_ENABLED(CONFIG_NET_IPV6, (
+					   iface = net_if_get_by_index(addr6->sin6_scope_id)));
+				if (iface != NULL) {
+					net_context_set_iface(context, iface);
+				}
+			}
 		}
 
+
```
rlubos: nit: Extra emtpy line

subsys/net/ip/net_context.c
Cristib05: `z_impl_net_if_get_by_index()` function has this check already, returns `NULL` if `index <=0`

rlubos (approved):

jukkar (approved):

PR #100730, opened 2025-12-09
openthread: platform: udp: Correct scope_id assignment for ll addresses

Cristib05:
This commit aims to fix correct scope_id assignment when link local addresses are in use.
Related to #100728


rlubos (approved):

pdgendt (approved):
Thanks for the explanation, is a backport needed?

PR #100739, opened 2025-12-09
llext: Keep constant data in flash.

iabdalkader:
This patch adds a new Kconfig option that allows read-only data sections without relocations to be mapped directly (remain in flash) instead of being copied to RAM during extension loading. This helps extensions that have large constant data, such as certificates, lookup tables etc.. to save significant RAM, especially with MPU enabled.

This feature requires the extension's linker script to place rodata sections without relocations into separate, non-overlapping section called `.rodata.noreloc`. In our application, we automate this by performing two-pass linking with a simple Python/Go script, that separates rodata with/without relocations. The final linker script looks something like this:

```ld
SECTIONS
{
    /* Read-only data WITH relocations - will be copied to RAM by LLEXT */
    .rodata : {
        *(.rodata)
        *(.rodata._ZTVN7arduino10SerialUSB_E)
        *(.rodata._ZTVN7arduino12ZephyrSerialE)
        *(.rodata._ZTVN7arduino14HardwareSerialE)
        *(.rodata._ZTVN7arduino5PrintE)
        *(.rodata._ZTVN7arduino6StreamE)
    }

    /* Read-only data WITHOUT relocations - kept in flash by LLEXT */
    .rodata.noreloc : {
        *(.rodata._ZL10cacert_pem)
        *(.rodata._ZN7arduino5Print7printlnEv.str1.4)
        *(.rodata.setup.str1.4)
        *(.rodata.str1.4)
    }

    /* Merge all .rel.rodata.* sections into .rel.rodata */
    .rel.rodata : {
        *(.rel.rodata)
        *(.rel.rodata.*)
    }

    /* Merge all .rela.rodata.* sections into .rela.rodata */
    .rela.rodata : {
        *(.rela.rodata)
        *(.rela.rodata.*)
    }
}
```

Note that we chose to disable this optimization when MMU or USERSPACE is enabled because the NO_RELOC section may not be aligned to the MPU/MMU requirements on some architectures. If we leave it in place, llext will attempt to create an MPU region for it using a misaligned address, and if we allow it to be copied to RAM, it defeats the purpose of this feature.

### DIFF
--- a/include/zephyr/llext/llext.h
+++ b/include/zephyr/llext/llext.h
@@ -53,6 +53,9 @@ enum llext_mem {
 	LLEXT_MEM_PREINIT,      /**< Array of early setup functions */
 	LLEXT_MEM_INIT,         /**< Array of setup functions */
 	LLEXT_MEM_FINI,         /**< Array of cleanup functions */
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+	LLEXT_MEM_RODATA_NO_RELOC,  /**< Read-only data without relocations (kept in flash) */
+#endif
 
 	LLEXT_MEM_COUNT,        /**< Number of regions managed by LLEXT */
 };
@@ -61,6 +64,22 @@ enum llext_mem {
 
 /* Number of memory partitions used by LLEXT */
 #define LLEXT_MEM_PARTITIONS (LLEXT_MEM_BSS+1)
+
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+/* Section name for read-only data kept in flash */
+#define LLEXT_SECT_RODATA_NO_RELOC llext.rodata.noreloc
+
+/* Full section name as string for comparisons */
+#define LLEXT_SECTION_RODATA_NO_RELOC ("." STRINGIFY(LLEXT_SECT_RODATA_NO_RELOC))
+
+/**
+ * Use this attribute on read-only data that should remain in flash
+ * instead of being copied to RAM.
+ */
+#define LLEXT_RODATA_NO_RELOC Z_GENERIC_DOT_SECTION(LLEXT_SECT_RODATA_NO_RELOC)
+#else
+#define LLEXT_RODATA_NO_RELOC
+#endif
 
 struct llext_loader;
 /** @endcond */
--- a/subsys/llext/Kconfig
+++ b/subsys/llext/Kconfig
@@ -108,6 +108,20 @@ config LLEXT_STORAGE_WRITABLE
 	  Select if LLEXT storage is writable, i.e. if extensions are stored in
 	  RAM and can be modified in place
 
+config LLEXT_RODATA_NO_RELOC
+	bool "Keep read-only data without relocations in flash"
+	depends on !MMU && !USERSPACE
+	help
+	  When enabled, read-only data sections (.rodata) that do not have
+	  any relocations are kept in their original location (typically flash)
+	  instead of being copied to RAM.
+
+	  This is useful for large constant data like certificates, lookup
+	  tables, or string literals that don't contain any pointers.
+
+	  This feature requires the extension to manually tag that data with
+	  the LLEXT_RODATA_NO_RELOC macro.
+
 config LLEXT_EXPORT_DEVICES
 	bool "Export all DT devices to llexts"
 	select LLEXT_EXPORT_SYMBOL_GROUP_DEVICE
--- a/subsys/llext/llext_load.c
+++ b/subsys/llext/llext_load.c
@@ -262,6 +262,10 @@ static int llext_map_sections(struct llext_loader *ldr, struct llext *ext,
 				mem_idx = LLEXT_MEM_TEXT;
 			} else if (shdr->sh_flags & SHF_WRITE) {
 				mem_idx = LLEXT_MEM_DATA;
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+			} else if (strcmp(name, LLEXT_SECTION_RODATA_NO_RELOC) == 0) {
+				mem_idx = LLEXT_MEM_RODATA_NO_RELOC;
+#endif
 			} else {
 				mem_idx = LLEXT_MEM_RODATA;
 			}
@@ -533,6 +537,13 @@ static int llext_map_sections(struct llext_loader *ldr, struct llext *ext,
 			ldr->sect_map[i].offset = shdr->sh_offset - ldr->sects[mem_idx].sh_offset;
 		}
 	}
+
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+	if (ldr->sects[LLEXT_MEM_RODATA_NO_RELOC].sh_flags & SHF_LLEXT_HAS_RELOCS) {
+		LOG_ERR("%s has relocations", LLEXT_SECTION_RODATA_NO_RELOC);
+		return -ENOEXEC;
+	}
+#endif
 
 	return 0;
 }
--- a/tests/subsys/llext/src/rodata_no_reloc_ext.c
+++ b/tests/subsys/llext/src/rodata_no_reloc_ext.c
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2025 Arduino SA
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Test extension for CONFIG_LLEXT_RODATA_NO_RELOC feature.
+ */
+
+#include <stdint.h>
+#include <zephyr/llext/llext.h>
+#include <zephyr/llext/symbol.h>
+#include <zephyr/sys/printk.h>
+#include <zephyr/ztest_assert.h>
+
+/* Exported by the main test to provide ELF buffer location */
+extern const void *rodata_no_reloc_ext_ptr;
+extern const size_t rodata_no_reloc_ext_size;
+
+/* rodata with a relocation - forces .rodata to be relocated */
+Z_GENERIC_SECTION(.rodata)
+static const void *relocated_data = &printk;
+
+/* rodata with no relocation - should stay in place */
+LLEXT_RODATA_NO_RELOC
+static const uint32_t noreloc_data = 0x12345678;
+
+void test_entry(void)
+{
+	uintptr_t elf_start = (uintptr_t)rodata_no_reloc_ext_ptr;
+	uintptr_t elf_end = elf_start + rodata_no_reloc_ext_size;
+	uintptr_t noreloc_addr = (uintptr_t)&noreloc_data;
+	uintptr_t relocated_addr = (uintptr_t)&relocated_data;
+
+	printk("noreloc at %p, relocated at %p, ELF: %p - %p\n",
+	       (void *)noreloc_addr, (void *)relocated_addr,
+	       (void *)elf_start, (void *)elf_end);
+
+	/* Verify data is correct */
+	zassert_equal(noreloc_data, 0x12345678, "noreloc_data value mismatch");
+
+	/* noreloc_data should stay in original ELF buffer */
+	zassert_true(noreloc_addr >= elf_start && noreloc_addr < elf_end,
+		     "noreloc_data should remain in ELF buffer");
+
+	/* relocated_data should be outside ELF buffer */
+	zassert_true(relocated_addr < elf_start || relocated_addr >= elf_end,
+		     "relocated_data should be outside ELF buffer");
+}
+EXPORT_SYMBOL(test_entry);
--- a/tests/subsys/llext/testcase.yaml
+++ b/tests/subsys/llext/testcase.yaml
@@ -44,6 +44,7 @@ tests:
     extra_conf_files: ['no_mem_protection.conf']
     extra_configs:
       - CONFIG_LLEXT_STORAGE_WRITABLE=n
+      - CONFIG_LLEXT_RODATA_NO_RELOC=y
   llext.readonly_mpu:
     arch_allow:
       - arm # Xtensa needs writable storage, currently not supported on RISC-V
@@ -64,6 +65,7 @@ tests:
     extra_configs:
       - CONFIG_LLEXT_HEAP_SIZE=128 # qemu_cortex_a9 requires larger heap
       - CONFIG_LLEXT_STORAGE_WRITABLE=n
+      - CONFIG_LLEXT_RODATA_NO_RELOC=y
   llext.writable:
     arch_allow:
       - arm
--- a/tests/subsys/llext/CMakeLists.txt
+++ b/tests/subsys/llext/CMakeLists.txt
@@ -47,6 +47,10 @@ if(NOT CONFIG_LLEXT_TYPE_ELF_SHAREDLIB)
     list(APPEND ext_names init_fini)
 endif()
 
+if(CONFIG_LLEXT_RODATA_NO_RELOC)
+    list(APPEND ext_names rodata_no_reloc)
+endif()
+
 # generate extension targets for each extension in 'ext_names'
 foreach(ext_name ${ext_names})
   set(ext_src ${PROJECT_SOURCE_DIR}/src/${ext_name}_ext.c)
--- a/tests/subsys/llext/src/test_llext.c
+++ b/tests/subsys/llext/src/test_llext.c
@@ -332,6 +332,17 @@ static LLEXT_CONST uint8_t inspect_ext[] LLEXT_SECT ELF_ALIGN = {
 	#include "inspect.inc"
 };
 
+#if defined(CONFIG_LLEXT_RODATA_NO_RELOC)
+static LLEXT_CONST uint8_t rodata_no_reloc_ext[] ELF_ALIGN = {
+	#include "rodata_no_reloc.inc"
+};
+const void *rodata_no_reloc_ext_ptr = rodata_no_reloc_ext;
+EXPORT_SYMBOL(rodata_no_reloc_ext_ptr);
+const size_t rodata_no_reloc_ext_size = sizeof(rodata_no_reloc_ext);
+EXPORT_SYMBOL(rodata_no_reloc_ext_size);
+LLEXT_LOAD_UNLOAD(rodata_no_reloc)
+#endif
+
 void do_inspect_checks(struct llext_loader *ldr, struct llext *ext, enum llext_mem reg_idx,
 		       const char *sect_name, const char *sym_name)
 {


include/zephyr/llext/llext.h
```diff
@@ -53,6 +53,9 @@ enum llext_mem {
 	LLEXT_MEM_PREINIT,      /**< Array of early setup functions */
 	LLEXT_MEM_INIT,         /**< Array of setup functions */
 	LLEXT_MEM_FINI,         /**< Array of cleanup functions */
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+	LLEXT_MEM_RODATA_NO_RELOC,  /**< Read-only data without relocations (kept in flash) */
+#endif
```
pillo79, line 58: This must be moved upwards. It must be counted in the `LLEXT_MEM_PARTITIONS` (and thus have an index close to `BSS`), because this area needs a memory partition to cover the `USERSPACE` case, or user threads would not be able to access the new section.

subsys/llext/Kconfig
```diff
@@ -108,6 +108,22 @@ config LLEXT_STORAGE_WRITABLE
 	  Select if LLEXT storage is writable, i.e. if extensions are stored in
 	  RAM and can be modified in place
 
+config LLEXT_RODATA_NO_RELOC
+	bool "Keep read-only data without relocations in flash"
+	depends on !LLEXT_STORAGE_WRITABLE
```
pillo79: Why this limitation? It should be possible to keep the original data in a writable buffer. Also, that option is just a default for the `llext_loader`'s storage attribute, so this is not even effective 🙂

subsys/llext/llext_mem.c
```diff
@@ -132,10 +132,15 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
...
 			if (ext->mem[mem_idx]) {
 				if ((IS_ALIGNED(ext->mem[mem_idx], region_align) ||
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+				    (mem_idx == LLEXT_MEM_RODATA_NO_RELOC) ||
+#endif
 				     ldr_parm->pre_located) &&
 				    ((mem_idx != LLEXT_MEM_TEXT) ||
 				     INSTR_FETCHABLE(ext->mem[mem_idx], region_alloc))) {
 					/* Map this region directly to the ELF buffer */
+					LOG_DBG("Region %d KEPT IN FLASH at %p",
```
pillo79: Minor nit since the source data isn't always Flash:
```suggestion
					LOG_DBG("Region %d reused at %p",
```

subsys/llext/llext_mem.c
```diff
@@ -132,10 +132,15 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
 			ext->mem[mem_idx] = llext_peek(ldr, region->sh_offset);
 			if (ext->mem[mem_idx]) {
 				if ((IS_ALIGNED(ext->mem[mem_idx], region_align) ||
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+				    (mem_idx == LLEXT_MEM_RODATA_NO_RELOC) ||
+#endif
```
pillo79: This will create issues with `USERSPACE`, as user threads will not be able to access the data without a properly aligned memory partition. (Kernel is fine though).

pillo79 (commented):
Thanks - it is a tremendously useful feature. While we at Arduino do benefit from a fine-grained automatic grouping script, we should provide a manual option, with a macro for users to mark big literal chunks that the should be placed in a separate section using `Z_GENERIC_SECTION`. This should also help in crafting a set of tests that cover all LLEXT use cases.

Please see a few notes below.

include/zephyr/llext/llext.h
iabdalkader: Does the memory partition create an MPU entry? If so, then this will not work because the section is not necessarily aligned and padded to MPU requirement, as it's kept in flash. Also, user threads should be able to access flash by default/boot MPU config, no?

subsys/llext/llext_mem.c
iabdalkader: This can't be fixed. If we let it align/pad and copy to memory, it would defeat the purpose of this PR, and we can't align/pad it via linker script because that could waste a lot of flash space. See my comment above, I think user threads can already access this region without a memory part (and MPU region) as the default/boot/background MPU should allow it: 
https://docs.zephyrproject.org/latest/kernel/usermode/memory_domain.html#boot-time-memory-configuration
> A read-only, executable region or regions for program text and ro-data, that is accessible to user mode.

Although I'm not 100% sure because I can't get user mode threads to work with our extension loader, some missing config, or deeper issues with our app.

That all said, we could just use `depends on !USERSPACE`?

include/zephyr/llext/llext.h
pillo79: LLEXT creates a `RODATA` mapping because usually the source is in the LLEXT heap, not in Flash (which would be covered by the boot mapping). 

That logic might need some adaptation to cover this new use case. Still, `RODATA_NO_RELOC` needs to be treated like the "other `RODATA`", since the distinction introduced in this PR is arbitrary and only for optimization purposes.

subsys/llext/llext_mem.c
pillo79: You can run the LLEXT test suite with the following Twister command line:

```
west twister -cv -T tests/subsys/llext
```

That includes emulated targets that cover `USERSPACE`. Enabling this feature in the test suite can definitely be helpful in fine tuning it.

include/zephyr/llext/llext.h
iabdalkader: That doesn't quite answer my question. Would adding this region to `LLEXT_MEM_PARTITIONS` create an MPU entry for it? That would be broken, because this region is neither aligned, nor padded. Plus, it will Not automatically work, so `llext_init_mem_part` would also need to be fixed.

> Still, `RODATA_NO_RELOC` needs to be treated like the "other `RODATA`", since the distinction introduced in this PR is arbitrary and only for optimization purposes.

I don't think so, it doesn't contain relocations, it's not relocated to RAM, so it's more similar to `text` than RODATA really. That said, I don't really care too much about this, we don't use user space mode, so I'm going to move it as requested.

tests/subsys/llext/src/rodata_no_reloc_ext.c
```diff
@@ -0,0 +1,45 @@
...
+#include <zephyr/llext/symbol.h>
+#include <zephyr/sys/printk.h>
+#include <zephyr/ztest_assert.h>
+
+/* rodata with a relocation - forces .rodata to be relocated */
+Z_GENERIC_SECTION(.rodata)
+static const void *relocated_data = &printk;
+
+/* rodata with no relocation - should stay in place */
+Z_GENERIC_SECTION(.rodata.noreloc)
```
pillo79: Please define a public macro in `llext.h` so that using this functionality doesn't depend on implementation details. It should also reuse `LLEXT_SECTION_RODATA_NO_RELOC` (removing quotes and `STRINGIFY`ing it for the string compare).

tests/subsys/llext/testcase.yaml
```diff
@@ -64,6 +64,17 @@ tests:
     extra_configs:
       - CONFIG_LLEXT_HEAP_SIZE=128 # qemu_cortex_a9 requires larger heap
       - CONFIG_LLEXT_STORAGE_WRITABLE=n
+  llext.rodata_no_reloc:
+    arch_allow:
+      - arm
+      - riscv
+      - arc
+    filter: not CONFIG_MPU and not CONFIG_MMU
+    extra_conf_files: ['no_mem_protection.conf']
```
pillo79: `filter` and `arch_allow` are useful tools to improve Twister runtime, but must match already existing limitations on the Kconfig or code side. If these filters are really required, please add the same `depends on` constraints in the `LLEXT_RODATA_NO_RELOC` Kconfig. Otherwise, you can add that config to existing test targets (where applicable).

subsys/llext/Kconfig
```diff
@@ -108,6 +108,21 @@ config LLEXT_STORAGE_WRITABLE
 	  Select if LLEXT storage is writable, i.e. if extensions are stored in
 	  RAM and can be modified in place
 
+config LLEXT_RODATA_NO_RELOC
```
pillo79, line 111: The help text below needs some adjustments after the change to a fixed section name.

pillo79 (commented):
Thanks, looks much better now. Although it's a pity to see more string compares in LLEXT, IMO it makes the logic both robust and simple. 👍 

Not sure about the test filtering - testing is currently restricted to a few cases without enforcing corresponding limits in the code, so users can still use the functionality in untested situations (and AFAICS, excluding MPU/MMU avoids the  Portenta H7 use case).

tests/subsys/llext/testcase.yaml
iabdalkader: > If these filters are really required, please add the same `depends on` constraints in the `LLEXT_RODATA_NO_RELOC` Kconfig.

I removed the filters and added x86, let's hope the tests pass. According to comments in there, xtensa requires writable storage, so yes, that seems to be required.

iabdalkader: @pillo79  I removed the filters as you suggested and now some tests fail on Cortex-R5. The failed test seems unrelated to this feature or this new test case as it still fails with `CONFIG_LLEXT_RODATA_NO_RELOC=n`. Now what?

```
START - test_inter_ext
W: Cannot peek region 0: 0x400e034 not aligned to 0x20
W: Cannot peek region 2: 0x400e054 not aligned to 0x20
I: Loaded extension inter_ext_depen
W: Cannot peek region 2: 0x400f164 not aligned to 0x100
I: Loaded extension export_dependen
E: ***** PREFETCH ABORT *****
E: Permission Fault @ 0x04042882
E: r0/a1:  0x00000001  r1/a2:  0x0401dd56  r2/a3:  0x0401dbf0
E: r3/a4:  0x000001e6 r12/ip:  0x0000f2c0 r14/lr:  0x04000569
E:  xpsr:  0x2000013f
E: s[ 0]:  0x00000000  s[ 1]:  0x00000000  s[ 2]:  0x00000000  s[ 3]:  0x00000000
E: s[ 4]:  0x00000000  s[ 5]:  0x00000000  s[ 6]:  0x00000000  s[ 7]:  0x00000000
E: s[ 8]:  0x00000000  s[ 9]:  0x00000000  s[10]:  0x00000000  s[11]:  0x00000000
E: s[12]:  0x00000000  s[13]:  0x00000000  s[14]:  0x00000000  s[15]:  0x00000000
E: fpscr:  0x00000000
E: Faulting instruction address (r15/pc): 0x04042882
E: >>> ZEPHYR FATAL ERROR 0: CPU exception on CPU 0
E: Current thread: 0x40401b0 (test_inter_ext)
E: Halting system
```

pillo79: Please re-read my message @iabdalkader, the ask was to properly sync Twister coverage with code, not remove the filters and "hope". Also, the [failing test in CI](https://github.com/zephyrproject-rtos/zephyr/actions/runs/20104555680/job/57685043892?pr=100739#step:13:1250) is the one added in this PR, while I proposed to add the new config to the **existing** set of tests where appropriate. 

While developing, I strongly suggest using Twister offline to check your PR result: it's faster for you and reduces reviewer notifications. Thanks!

iabdalkader: Removed the new test config as suggested.

include/zephyr/llext/llext.h
```diff
@@ -62,6 +65,20 @@ enum llext_mem {
...
+
+/* Full section name as string for comparisons */
+#define LLEXT_SECTION_RODATA_NO_RELOC ("." STRINGIFY(LLEXT_SECT_RODATA_NO_RELOC))
+
+/**
+ * Use this attribute on read-only data that should remain in flash
+ * instead of being copied to RAM.
+ */
+#define LLEXT_RODATA_NO_RELOC Z_GENERIC_DOT_SECTION(LLEXT_SECT_RODATA_NO_RELOC)
+#endif
```
pillo79, line 82: Please add an empty define for `LLEXT_RODATA_NO_RELOC` in an `#else` clause here, so that existing code still compiles when you disable the feature.

subsys/llext/Kconfig
```diff
@@ -108,6 +108,20 @@ config LLEXT_STORAGE_WRITABLE
...
+	help
+	  When enabled, read-only data sections (.rodata) that do not have
+	  any relocations are kept in their original location (typically flash)
+	  instead of being copied to RAM.
+
+	  This is useful for large constant data like certificates, lookup
+	  tables, or string literals that don't contain any pointers.
+
+	  This feature requires the extension to place rodata sections without
+	  relocations in LLEXT_RODATA_NO_RELOC section.
```
pillo79:
```suggestion
	  This feature requires the extension to manually tag that data with
	  the LLEXT_RODATA_NO_RELOC macro.
```

include/zephyr/llext/llext.h
```diff
@@ -46,6 +46,9 @@ enum llext_mem {
 	LLEXT_MEM_DATA,         /**< Initialized data */
 	LLEXT_MEM_RODATA,       /**< Read-only data */
 	LLEXT_MEM_BSS,          /**< Uninitialized data */
+#ifdef CONFIG_LLEXT_RODATA_NO_RELOC
+	LLEXT_MEM_RODATA_NO_RELOC,  /**< Read-only data without relocations (kept in flash) */
+#endif
```
pillo79: Any downside in removing the `#if` gates in the definition? That would allow removing them later on as well, for cleaner code.

subsys/llext/llext_mem.c
```diff
@@ -132,10 +132,14 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
...
```
pillo79: (here's where the gates can be removed).

subsys/llext/llext_mem.c
```diff
@@ -132,10 +132,14 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
...
 				     ldr_parm->pre_located) &&
 				    ((mem_idx != LLEXT_MEM_TEXT) ||
 				     INSTR_FETCHABLE(ext->mem[mem_idx], region_alloc))) {
 					/* Map this region directly to the ELF buffer */
+					LOG_DBG("Region %d kept at %p", mem_idx, ext->mem[mem_idx]);
```
teburd: Not sure how much help this LOG_DBG is providing, and we're already quite noisy with logs

tests/subsys/llext/src/rodata_no_reloc_ext.c
```diff
@@ -0,0 +1,51 @@
...
+#include <zephyr/llext/symbol.h>
+#include <zephyr/sys/printk.h>
+#include <zephyr/ztest_assert.h>
+
+/* Exported by the main test to provide ELF buffer location */
+extern const void *rodata_no_reloc_ext_ptr;
+extern const size_t rodata_no_reloc_ext_size;
+
+/* rodata with a relocation - forces .rodata to be relocated */
+Z_GENERIC_SECTION(.rodata)
```
teburd, line 22: Z_ macros are meant to be private, concerning that we are needing this here in what should be application code

teburd (changes requested):
Minor things, test at least covers what I'd expect to see covered.

I'd note that the idea of loadable modules having tls certs/keys in ROM without user mode sounds like a bad idea. But what you do with the tools is none of my business really.

Fixing the two nits here is though.

subsys/llext/llext_load.c
pillo79, line 529: Suggestion: if users tag entities that _need_ relocations with `RODATA_NO_RELOC`, it will currently be blindly applied, causing all sorts of issues. Adding `if (target_region == LLEXT_MEM_RODATA_NO_RELOC) { ... } else` here and exiting with an error message would detect this kind of mistakes.

pillo79 (commented):
Thanks for the updates, this version looks much better to me! 
Only a few additional tiny nits below.

include/zephyr/llext/llext.h
iabdalkader: If removed it changes the regions numbers. I think I saw some test hard-coding region numbers. My intention is to make sure this feature absolutely cannot break any existing code, this aggressive gating does that.

tests/subsys/llext/src/rodata_no_reloc_ext.c
iabdalkader: @teburd @pillo79 Is there an alternative public macro?

laurenmurphyx64: @teburd I noticed we also do this in `detached_fn_ext.c`

subsys/llext/llext_load.c
iabdalkader: I added this check, but once outside the loop instead of checking every region.

include/zephyr/llext/llext.h
pillo79: Hardcoding those indexes would be worrying, and would break anyway where the gate is enabled. But for maximum compatibility you can also add it at the end of the enum 👍

tests/subsys/llext/src/rodata_no_reloc_ext.c
pillo79: IMHO they can generally be useful for LLEXT tests, as we do need to mess with the ELF files a lot. In the same vein, we always allowed tests to access internals of the subsystem, which apps should never attempt.

However, in this case the test only cares how LLEXT treats the variable, so hardcoding a section name should not be required IIUC?

include/zephyr/llext/llext.h
iabdalkader: Moved to the end of the enum. I'm assuming we keeping the gating macros, so resolving this and the next one.

iabdalkader: Actually, I'm not sure if you still want to remove them or not. I don't think they should be removed.

subsys/llext/Kconfig
```diff
@@ -108,6 +108,20 @@ config LLEXT_STORAGE_WRITABLE
 	  Select if LLEXT storage is writable, i.e. if extensions are stored in
 	  RAM and can be modified in place
 
+config LLEXT_RODATA_NO_RELOC
+	bool "Keep read-only data without relocations in flash"
+	depends on !USERSPACE
```
teburd: !MPU is likely more relevant in this case, note that the tests disable MPU not USERSPACE.

pillo79: That is a subtle change which wouldn't really match our use case (https://github.com/zephyrproject-rtos/zephyr/pull/100964#issuecomment-3745072419), where we do have `MPU` but not `USERSPACE`. This currently matches the same test performed by that code.

subsys/llext/llext_mem.c
iabdalkader: FYI, after a rebase, this will no longer be needed, because the region will be aligned.

subsys/llext/Kconfig
iabdalkader: Actually, this check should be `!MMU && !USERSPACE`. As Luca pointed out, we use the MPU without USERSPACE. As for MMU, it also requires some particular alignment (for example, 4096), so it won't work with that either.

FWIW, this feature can still be used with MMU, MPU, USERSPACE but you'd have to manually align and round up the region to whatever the MPU/MMU requires.

teburd: That's fair, you *could* use with MMU and MPU, but then there would need to be a runtime check on MPU or MMU and the alignment and size of the .rodata section.

Fair!

pillo79 (approved):
LGTM, thanks for coming up with this!

teburd (approved):
I'm ok with this

PR #100742, opened 2025-12-09
soc: intel_adsp: tools: cavstool.py: add I2S offload support

kv2019i:
Program I2S link for DSP offload, allowing to run DSP tests using the I2S/SSP interface.

The I2S/SSP hardware provides nice loopback capabilities that are useful for testing. On ACE platforms, this capability needs to be enabled on host side, so add the capabiltiy to cavstool.py used to run Zephyr based tests.

### DIFF
--- a/soc/intel/intel_adsp/tools/cavstool.py
+++ b/soc/intel/intel_adsp/tools/cavstool.py
@@ -48,6 +48,9 @@ CSTALL = 8
 SPA    = 16
 CPA    = 24
 
+# LCTL bits
+OFLEN  = 4
+
 class HDAStream:
     # creates an hda stream with at 2 buffers of buf_len
     def __init__(self, stream_id: int):
@@ -256,12 +259,14 @@ def map_regs(log_only):
     hda.SPBFCTL = 0x0704
     hda.PPCTL   = 0x0804
 
+    if ace20 or ace30:
+        hda.HDAML_I2S_LCTL = 0x0C40 + 0x40 * 3 + 4
+
     # Find the ID of the first output stream
     hda_ostream_id = (hda.GCAP >> 8) & 0x0f # number of input streams
     log.info(f"Selected output stream {hda_ostream_id} (GCAP = 0x{hda.GCAP:x})")
     hda.SD_SPIB = 0x0708 + (8 * hda_ostream_id)
     hda.freeze()
-
 
     # Standard HD Audio Stream Descriptor
     sd = Regs(hdamem + 0x0080 + (hda_ostream_id * 0x20))
@@ -394,8 +399,7 @@ def runx(cmd):
     return subprocess.check_output(cmd, shell=True).decode().rstrip()
 
 def mask(bit):
-    if cavs25:
-        return 0b1 << bit
+    return 0b1 << bit
 
 def load_firmware(fw_file):
     try:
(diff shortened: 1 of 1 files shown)


dcpleung (dismissed):

soc/intel/intel_adsp/tools/cavstool.py
```diff
@@ -538,6 +542,11 @@ def load_firmware_ace(fw_file):
         log.info("Waiting for DSP subsystem power on")
         time.sleep(0.1)
 
+    log.info(f"Enabling offload for I2S link")
+    # needed to allow DSP to access SSP DAI
+    hda.HDAML_I2S_LCTL |= mask(OFLEN)
+    log.debug(f"HDAML.I2S_LCTL 0x{hda.HDAML_I2S_LCTL:x}")
```
ujfalusi: Unconditionally enable SSP offload? What can go wrong?

soc/intel/intel_adsp/tools/cavstool.py
```diff
@@ -394,8 +399,7 @@ def runx(cmd):
     return subprocess.check_output(cmd, shell=True).decode().rstrip()
 
 def mask(bit):
-    if cavs25:
-        return 0b1 << bit
+    return 0b1 << bit
```
ujfalusi, line 402: So this works on ACE as well? :D

soc/intel/intel_adsp/tools/cavstool.py
```diff
@@ -256,13 +259,15 @@ def map_regs(log_only):
     hda.SPBFCTL = 0x0704
     hda.PPCTL   = 0x0804
 
+    if adsp_is_ace():
+        hda.HDAML_I2S_LCTL = 0x0C40 + 0x40 * 3 + 4
```
ujfalusi, line 263: HDAML_I2S_LCTL is set only in case of ACE (cAVS also have SSP I believe...)
and at line 546 you write to this unconditionally

soc/intel/intel_adsp/tools/cavstool.py
kv2019i: @ujfalusi This is a bit contentious, but this script/tool is only used to test the DSP and I can't think of cases where we would not enable DSP to program the SSP. Also this matches the behaviour where SSP was accessible by DSP by default. This is really a hw feature to allow use without a DSP at all, and when/if you are using cavstool.py, you are pretty deep on the "let's use DSP" path. :)

But in summarum, I did consider putting this behind a command-line option, but decided not to (as anyone using this on a ACEx.y Intel platform, would want to have the option set).

soc/intel/intel_adsp/tools/cavstool.py
kv2019i: I did test this on cavs2.5! L546 is in funtion load_firmware_ace() which is only run on ACE hw.

soc/intel/intel_adsp/tools/cavstool.py
kv2019i: I know, know. I was very nervous about this change, but I trust our hw engineers. ;)

soc/intel/intel_adsp/tools/cavstool.py
ujfalusi: OK

soc/intel/intel_adsp/tools/cavstool.py
ujfalusi: oh, indeed it is in load_firmware_ace(), I just checked the function name up to **load_firmware**...

ujfalusi (approved):

kv2019i:
@tmleman @abonislawski @softwarecki @nashif would need one more approval

tmleman (approved):
LGTM

dcpleung (approved):

PR #100745, opened 2025-12-09
 drivers: counter: microchip: Add counter using TCC timer

mchp-asif:
Adds g1 driver for counter using tcc peripheral


drivers/counter/counter_mchp_tcc_g1.c
```diff
@@ -0,0 +1,784 @@
...
+
+struct mchp_counter_clock {
+	const struct device *clock_dev;
+	clock_control_subsys_t host_core_sync_clk;
+	clock_control_subsys_t periph_async_clk;
+};
+
+struct tcc_counter_irq_map {
+	const uint32_t ovf_irq_line;      /* Overflow interrupt number */
+	const uint32_t comp_irq_line[6u]; /* Channel interrupt number */
```
nandojve: 6u ? Magic number

drivers/counter/counter_mchp_tcc_g1.c
```diff
@@ -0,0 +1,784 @@
...
+
+	/* Configure drive control register */
+	p_regs->TCC_DRVCTRL = TCC_DRVCTRL_INVEN(0U);
+
+	/* Set the period register to maximum value */
+	max_counter_value = (uint32_t)(BIT64(max_bit_width) - 1);
+
+	p_regs->TCC_PER = max_counter_value;
+
+	for (uint8_t index = 0u; index < max_channels; index++) {
```
nandojve, line 147: Should we capture a error related to the `max_channels `? I mean what is the maximum channels available.

Is this defined by the

https://github.com/zephyrproject-rtos/zephyr/blob/3be4c4c5a81949b771c7d59b1852ef3e61836df9/dts/bindings/counter/microchip%2Ctcc-g1.yaml#L56

If yes, I think there is a way to define max value to the property then at compile time we pick greater then values.

drivers/counter/counter_mchp_tcc_g1.c
```diff
@@ -0,0 +1,784 @@
...
+	/* Immediate interrupt trigger */
+	if ((data->late_alarm_flag == true) && (data->late_alarm_channel == channel)) {
+		data->late_alarm_flag = false;
+	} else {
+		tcc_counter_alarm_irq_clear(cfg->regs, cfg->max_channels, channel);
+	}
+
+	data->channel_data[channel].callback = NULL;
+	cc_value = data->channel_data[channel].compare_value;
+	if (cb != NULL) {
```
nandojve, line 628: Add an empty line before the if ()

tests/drivers/counter/counter_basic_api/testcase.yaml
```diff
@@ -5,7 +5,9 @@ tests:
       - counter
     depends_on: counter
     min_ram: 16
-    platform_exclude: nucleo_f302r8
+    platform_exclude:
+      - nucleo_f302r8
```
nandojve, line 9: In the future, try to make the changes

```diff
-    platform_exclude: nucleo_f302r8
+    platform_exclude:
+      - nucleo_f302r8
```

In a separete commit. This helps if we need to rollback code. I know this example is very small but the principle can be used for simple things or more elaborated ones.

nandojve (changes requested):

dts/bindings/counter/microchip,tcc-g1-counter.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2025 Microchip Technology Inc.
+# SPDX-License-Identifier: Apache-2.0
+title: Microchip TCC G1 counter peripheral.
```
nandojve, line 3: The commit brief is using the board as reference.

`dts: bindings: microchip: sam_e54_xpro: Add counter yaml for tcc`

This binding is not about the sam_e54_xpro baord.

`dts: binding: counter: microchip: Add TCC G1 compatible`

drivers/counter/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ zephyr_library_sources_ifdef(CONFIG_COUNTER_INFINEON_TCPWM counter_infineon_tcpw
 zephyr_library_sources_ifdef(CONFIG_COUNTER_ITE_IT51XXX counter_ite_it51xxx.c)
 zephyr_library_sources_ifdef(CONFIG_COUNTER_ITE_IT8XXX2 counter_ite_it8xxx2.c)
 zephyr_library_sources_ifdef(CONFIG_COUNTER_MAXIM_DS3231 maxim_ds3231.c)
+zephyr_library_sources_ifdef(CONFIG_COUNTER_MCHP_G1_TCC counter_mchp_tcc_g1.c)
```
nandojve, line 31: Now, you are adding a driver. Then the commit brief should tell us that.

this
`drivers: microchip: sam_e54_xpro: Add counter g1 driver with tcc`
should be something like
`drivers: counter: microchip: Add TCC G1 driver`

a driver of type counter from vendor MCHP ...

tests/drivers/counter/counter_basic_api/boards/microchip/sam_e54_xpro_tcc1.overlay
```diff
@@ -0,0 +1,12 @@
+/*
```
nandojve, line 1: `tests: drivers: counter: added sam_e54 test support files`
to
`tests: drivers: counter: Add sam_e54 test support files`

This PR will Add..

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.yaml
```diff
@@ -12,6 +12,7 @@ ram: 256
 supported:
   - clock_control
   - comparator
+  - counter
```
nandojve, line 15: It is important mention what/how

boards: microchip: sam_e54_xpro: Add counter tag in sam_e54_xpro.yaml
- Add counter tag in the board yaml file section to allow CI run TCC G1 counter tests on this board.

nandojve (changes requested):

drivers/counter/counter_mchp_tcc_g1.c
mchp-asif: The maximum channel for a particular instance changes with the instance. It is set in the common dtsi files. 

 Do you mean, we need a validation check for it during `init`?

dts/bindings/counter/microchip,tcc-g1-counter.yaml
mchp-asif: Thank you @nandojve for your suggestions. I have updated each of the commit messages as mentioned, and will take care of it in future PRs as well.

drivers/counter/CMakeLists.txt
mchp-asif: Updated as suggested

boards/microchip/sam/sam_e54_xpro/sam_e54_xpro.yaml
mchp-asif: Updated as suggested

ArunMCHP (approved):

drivers/counter/counter_mchp_tcc_g1.c
mchp-asif: Updated with MACRO

drivers/counter/counter_mchp_tcc_g1.c
nandojve: If I understood correctly you question -> You only need to validate at run-time when a param came from above layers. User can always set an invalid param. However, indexes from instances **usually** you can validate inside the macros and at driver init. Then you speed up run-time execution because these are defined as "static" and can not be changed at run-time.

This means that if you can validate everything at driver init phase the system will fail at start with an error message.

nandojve (approved):

NhMchp (approved):

PR #100746, opened 2025-12-09
drivers: sensor: adi: adxl345: take selected g-range into account when converting samples

stemschmidt:
When running the samples/sensor/accel_trig sample with a adxl345 sensor the logged output values were doubled compared to the expected values.

The sensor_channel_get() path calls adxl345_accel_convert which did not handle the configured g-range, resulting in invalid sample values except for the +/-16g range. To fix this issue the configured range is taken into account and different conversion factors are applied.

adxl345_accel_convert is also made static as it is not used by another file.

### DIFF
--- a/drivers/sensor/adi/adxl345/adxl345.c
+++ b/drivers/sensor/adi/adxl345/adxl345.c
@@ -312,14 +312,22 @@ int adxl345_read_sample(const struct device *dev,
 	return 0;
 }
 
-void adxl345_accel_convert(struct sensor_value *val, int16_t sample)
+static void adxl345_accel_convert(struct sensor_value *val, int16_t sample,
+				  uint8_t selected_range)
 {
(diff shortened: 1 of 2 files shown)


MaureenHelm (commented):
Thanks for the fix. The code changes look good, but the commit message needs to be updated to fix the compliance check:
```
1: UC5 Commit title exceeds max length (77>75): "drivers: sensor: adi: adxl345: take selected range into account in conversion"
11: UC4 Commit message body line exceeds max length (76>75): "adxl345_accel_convert is also made static as it is not used by another file."
```
https://github.com/zephyrproject-rtos/zephyr/actions/runs/20066238808/job/57555432868?pr=100746#step:13:42

ubieda (approved):

MaureenHelm (approved):

PR #100749, opened 2025-12-09
drivers: display: Change display buffer pitch from pixel to byte

zejiang0jason:
The `pitch` in `struct display_buffer_descriptor` is defined as "Number of pixels between consecutive rows in the data buffer".

Limitation:
Some display controller needs the pitch of byte be dividable by an even value (e.g., 4, 64, ...), if the parameter pitch here is the number of pixels, then the RGB888 (3 bytes per pixel) can't work

Change the unit of `pitch` from pixel to byte.

For easy review:
- The first patch https://github.com/zephyrproject-rtos/zephyr/pull/100749/commits/e621c234bcc41baddd6b004a771b0e148d96b2e8 changes the structure definition, it is the key change,
- The second patch https://github.com/zephyrproject-rtos/zephyr/pull/100749/commits/e5229fed0bd66d5693ff5a2ee64d81b71fd5177f updates the display controller and samples accordingly. Currently it is created by AI tool, I am still check the content to make sure no issues.

### DIFF
--- a/drivers/display/display_co5300.c
+++ b/drivers/display/display_co5300.c
@@ -37,6 +37,7 @@ struct co5300_config {
 	uint16_t panel_height;
 	uint16_t channel;
 	uint16_t num_of_lanes;
+	uint16_t frame_pitch_align; /* Pitch alignment requirement, in bytes. */
 };
 
 struct co5300_data {
@@ -46,7 +47,7 @@ struct co5300_data {
 	struct k_sem tear_effect_sem;
 	/* Pointer to framebuffer */
 	uint8_t *frame_ptr;
-	uint32_t frame_pitch;
+	uint32_t frame_pitch; /* Pitch in bytes. */
 };
 
 /* Organized as MIPI_CMD | SIZE OF MIPI PARAM | MIPI PARAM */
@@ -108,12 +109,12 @@ static void co5300_copy_and_adjust_coordinates(const struct device *dev,
 
 	/* Copy the update area to the framebuffer */
 	src = buf;
-	dst = data->frame_ptr + (y * data->frame_pitch * data->bytes_per_pixel)
+	dst = data->frame_ptr + (y * data->frame_pitch)
 		+ (x * data->bytes_per_pixel);
 	for (uint16_t row = 0; row < desc->height; row++) {
 		memcpy(dst, src, desc->width * data->bytes_per_pixel);
-		src += desc->pitch * data->bytes_per_pixel;
-		dst += data->frame_pitch * data->bytes_per_pixel;
+		src += desc->pitch;
+		dst += data->frame_pitch;
 	}
 
 	/*
@@ -139,7 +140,7 @@ static void co5300_copy_and_adjust_coordinates(const struct device *dev,
 	local_desc->height = ROUND_UP(local_desc->height, 2U);
 	local_desc->pitch = data->frame_pitch;
 	local_desc->frame_incomplete = desc->frame_incomplete;
-	local_desc->buf_size = local_desc->width * local_desc->height * data->bytes_per_pixel;
+	local_desc->buf_size = local_desc->pitch * local_desc->height;
 }
 
 static int co5300_write(const struct device *dev,
@@ -222,7 +223,7 @@ static int co5300_write(const struct device *dev,
 
 	/* Start memory write. */
 	/* The address and the total pixel size of the updated area. */
-	src = data->frame_ptr + (local_y * data->frame_pitch * data->bytes_per_pixel)
+	src = data->frame_ptr + (local_y * data->frame_pitch)
 		+ (local_x * data->bytes_per_pixel);
 	tx_size = local_desc.buf_size;
 
@@ -246,11 +247,11 @@ static int co5300_write(const struct device *dev,
 		}
 
 		/* Advance source pointer and decrement remaining */
-		if (local_desc.pitch > local_desc.width) {
+		if (local_desc.pitch > (local_desc.width * data->bytes_per_pixel)) {
 			total_bytes_sent += bytes_written;
 			src += bytes_written + total_bytes_sent /
 				(local_desc.width * data->bytes_per_pixel) *
-				((local_desc.pitch - local_desc.width) * data->bytes_per_pixel);
+				(local_desc.pitch - local_desc.width * data->bytes_per_pixel);
 		} else {
 			src += bytes_written;
 		}
@@ -342,6 +343,8 @@ static int co5300_set_pixel_format(const struct device *dev,
 	/* Update the format in the device data after DCS command succeeds. */
 	data->bytes_per_pixel = bytes_per_pixel;
 	data->pixel_format = mipi_pixel_format;
+	data->frame_pitch = ROUND_UP(config->panel_width * bytes_per_pixel,
+		 config->frame_pitch_align);
 
 	return 0;
 }
@@ -552,13 +555,12 @@ static DEVICE_API(display, co5300_api) = {
 		.tear_effect_gpios = GPIO_DT_SPEC_INST_GET_OR(node_id, tear_effect_gpios, {0}),	\
 		.panel_width = DT_INST_PROP(node_id, width),					\
 		.panel_height = DT_INST_PROP(node_id, height),					\
+		.frame_pitch_align = DT_INST_PROP(node_id, pitch_align),			\
 	};											\
 	CO5300_FRAMEBUFFER_DECL(node_id);							\
 	static struct co5300_data co5300_data_##node_id = {					\
 		.pixel_format = DT_INST_PROP(node_id, pixel_format),				\
 		.frame_ptr = CO5300_FRAMEBUFFER(node_id),					\
-		.frame_pitch = ROUND_UP(DT_INST_PROP(node_id, width),				\
-				DT_INST_PROP(node_id, pitch_align)),				\
 	};											\
 	DEVICE_DT_INST_DEFINE(node_id,								\
 			    &co5300_init,							\
--- a/drivers/display/display_mcux_elcdif.c
+++ b/drivers/display/display_mcux_elcdif.c
@@ -89,7 +89,8 @@ static int mcux_elcdif_write(const struct device *dev, const uint16_t x, const u
 
 	LOG_DBG("W=%d, H=%d, @%d,%d", desc->width, desc->height, x, y);
 
-	if ((x == 0) && (y == 0) && (desc->width == config->rgb_mode.panelWidth) &&
+	if ((x == 0) && (y == 0) &&
+	    (desc->pitch == config->rgb_mode.panelWidth * dev_data->pixel_bytes) &&
 	    (desc->height == config->rgb_mode.panelHeight) && (desc->pitch == desc->width)) {
 		/* We can use the display buffer directly, no need to copy it */
 		LOG_DBG("Setting FB from %p->%p", (void *)dev_data->active_fb, (void *)buf);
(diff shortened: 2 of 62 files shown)


VynDragon:
Thanks, I support this initiative, will try to do a review and testing.

drivers/display/display_co5300.c
```diff
@@ -552,13 +554,12 @@ static DEVICE_API(display, co5300_api) = {
 		.tear_effect_gpios = GPIO_DT_SPEC_INST_GET_OR(node_id, tear_effect_gpios, {0}),	\
 		.panel_width = DT_INST_PROP(node_id, width),					\
 		.panel_height = DT_INST_PROP(node_id, height),					\
+		.frame_pitch_align = DT_INST_PROP_OR(node_id, pitch_align, 1),			\
 	};											\
 	CO5300_FRAMEBUFFER_DECL(node_id);							\
 	static struct co5300_data co5300_data_##node_id = {					\
 		.pixel_format = DT_INST_PROP(node_id, pixel_format),				\
 		.frame_ptr = CO5300_FRAMEBUFFER(node_id),					\
-		.frame_pitch = ROUND_UP(DT_INST_PROP(node_id, width),				\
```
KATE-WANG-NXP, line 559: Do not delete this, frame pitch needs a initial value after system init

drivers/display/display_co5300.c
```diff
@@ -246,11 +247,11 @@ static int co5300_write(const struct device *dev,
 		}
 
 		/* Advance source pointer and decrement remaining */
-		if (local_desc.pitch > local_desc.width) {
+		if (local_desc.pitch > local_desc.width * data->bytes_per_pixel) {
```
KATE-WANG-NXP: local_desc.pitch > (local_desc.width * data->bytes_per_pixel)?

drivers/mipi_dbi/mipi_dbi_nxp_dcnano_lcdif.c
```diff
@@ -244,14 +244,8 @@ static int mipi_dbi_dcnano_lcdif_write_display(const struct device *dev,
 
 	LCDIF_SetFrameBufferConfig(config->base, 0, &fbConfig);
 
-	if (bytes_per_pixel == 3U) {
-		/* For RGB888 the stride shall be calculated as
```
KATE-WANG-NXP, line 248: For pitch in bytes it shall follow the same law here

modules/lvgl/lvgl_display_mono.c
```diff
@@ -106,7 +106,7 @@ void lvgl_flush_cb_mono(lv_display_t *display, const lv_area_t *area, uint8_t *p
 	struct display_buffer_descriptor desc = {
 		.buf_size = (w * h) / 8U,
 		.width = w,
-		.pitch = w,
+		.pitch = DIV_ROUND_UP(w, 8),
```
KATE-WANG-NXP, line 109: Need to updatelvgl_display_16bit/24bit/32bit as well?

samples/drivers/display/src/main.c
```diff
@@ -376,7 +398,29 @@ int main(void)
 		}
 	}
 
-	buf_desc.pitch = ROUND_UP(rect_w, CONFIG_SAMPLE_PITCH_ALIGN);
+	switch (capabilities.current_pixel_format) {
```
KATE-WANG-NXP: May be separate a function?

matt-wood-ct:
Should it be bits rather then bytes, this allows for better packing on low colour depth screens like mono OLEDs & segment LCDs where each pixel may only be one bit

drivers/display/display_co5300.c
zejiang0jason: it is calculated in function `co5300_set_pixel_format`, which is called in `co5300_init`

drivers/display/display_co5300.c
zejiang0jason: updated.

drivers/mipi_dbi/mipi_dbi_nxp_dcnano_lcdif.c
zejiang0jason: updated.

modules/lvgl/lvgl_display_mono.c
zejiang0jason: updated.

drivers/display/display_st7735r.c
```diff
@@ -183,8 +183,9 @@ static int st7735r_write(const struct device *dev,
 	enum display_pixel_format fmt;
 	struct display_buffer_descriptor mipi_desc;
 
-	__ASSERT(desc->width <= desc->pitch, "Pitch is smaller than width");
-	__ASSERT((desc->pitch * ST7735R_PIXEL_SIZE * desc->height)
+	__ASSERT(desc->width * ST7735R_PIXEL_SIZE <= desc->pitch,
+		 "Pitch is smaller than width in bytes");
+	__ASSERT((desc->pitch * desc->height)
 		 <= desc->buf_size, "Input buffer too small");
```
etienne-lms: Would be simpler to read as: 
```suggestion
	__ASSERT(desc->pitch * desc->height <= desc->buf_size, 
		 "Input buffer too small");

File: drivers/display/display_co5300.c [hunk 7]
```diff
@@ -552,13 +555,12 @@ static DEVICE_API(display, co5300_api) = {
 		.tear_effect_gpios = GPIO_DT_SPEC_INST_GET_OR(node_id, tear_effect_gpios, {0}),	\
 		.panel_width = DT_INST_PROP(node_id, width),					\
 		.panel_height = DT_INST_PROP(node_id, height),					\
+		.frame_pitch_align = DT_INST_PROP_OR(node_id, pitch_align, 1),			\
```
etienne-lms: `pitch-align` is already `default: 1` in **chipone,co5300.yaml**.
```suggestion
		.frame_pitch_align = DT_INST_PROP(node_id, pitch_align),			\

File: drivers/display/display_sdl.c [hunk 8]
```diff
@@ -529,18 +527,18 @@ static void sdl_display_read_mono(const uint8_t *read_buf,
...
 			pix_ptr = (const uint32_t *)read_buf + h_idx * desc->pitch + w_idx;
 			buf8 = buf;
 
 			if (IS_ENABLED(CONFIG_SDL_DISPLAY_MONO_VTILED)) {
-				buf8 += (h_idx / 8U) * DIV_ROUND_UP(desc->pitch, 1U) + (w_idx);
+				buf8 += (h_idx / 8U) * desc->pitch + (w_idx);
 				bits = mono_pixel_order(h_idx % 8U);
 			} else {
-				buf8 += (h_idx)*DIV_ROUND_UP(desc->pitch, 8U) + (w_idx / 8U);
+				buf8 += (h_idx)*desc->pitch + (w_idx / 8U);
```
etienne-lms: Nitpicking:
```suggestion
				buf8 += h_idx * desc->pitch + (w_idx / 8U);

File: drivers/mipi_dsi/dsi_mcux_2l.c [hunk 9]
```diff
@@ -647,7 +647,7 @@ static ssize_t dsi_mcux_transfer(const struct device *dev, uint8_t channel,
 		 */
 		if (desc != NULL) {
 #endif
-			if ((desc->pitch * data->src_bytes_per_pixel) >
+			if (desc->pitch >
 				(msg->tx_len / desc->height)) {
```
etienne-lms: Suggestion for clarity:
```c
			if (desc->pitch > (msg->tx_len / desc->height)) {

**Review by etienne-lms — COMMENTED** [review]
A few minor review comments, FWIW. Looks consistent to me.

Such a change in `struct display_buffer_descriptor` ABI would deserve a note in the migration guide.

File: drivers/display/display_sdl.c [hunk 10]
```diff
@@ -218,7 +218,7 @@ static void sdl_display_write_argb8888(void *disp_buf,
 	__ASSERT((desc->pitch * 4U * desc->height) <= desc->buf_size,
```
avolmat-st: I think the * 4U is useless now that pitch is expressed is byte.

drivers/display/display_sdl.c
```diff
@@ -440,7 +438,7 @@ static void sdl_display_read_argb8888(const uint8_t *read_buf,
 {
 	__ASSERT((desc->pitch * 4U * desc->height) <= desc->buf_size, "Read buffer is too small");
```
avolmat-st: I think the *4U should also be removed here.

avolmat-st (changes requested):
Thanks for this development. I fully agree in this modification, on the video side the pitch is also referring to bytes and not pixel.

I reviewed the LTDC and SDL part, two small correction needed in the SDL part.
Moreover since pitch value is modified now, I think an entry into the migration-guide is needed.

zejiang0jason:
1. Updated based on review comment.
2. Merge to one patch.

drivers/display/display_co5300.c
```diff
@@ -139,7 +140,7 @@ static void co5300_copy_and_adjust_coordinates(const struct device *dev,
 	local_desc->height = ROUND_UP(local_desc->height, 2U);
 	local_desc->pitch = data->frame_pitch;
 	local_desc->frame_incomplete = desc->frame_incomplete;
-	local_desc->buf_size = local_desc->width * local_desc->height * data->bytes_per_pixel;
+	local_desc->buf_size = local_desc->pitch * local_desc->height;
```
KATE-WANG-NXP, line 143: Do not change this line, the local_desc->buf_size is the actual updated size, should not use pitch to calculate

KATE-WANG-NXP (changes requested):

drivers/display/display_mcux_elcdif.c
```diff
@@ -89,14 +89,16 @@ static int mcux_elcdif_write(const struct device *dev, const uint16_t x, const u
 
 	LOG_DBG("W=%d, H=%d, @%d,%d", desc->width, desc->height, x, y);
 
-	if ((x == 0) && (y == 0) && (desc->width == config->rgb_mode.panelWidth) &&
+	if ((x == 0) && (y == 0) &&
+	    (desc->pitch == config->rgb_mode.panelWidth * dev_data->pixel_bytes) &&
 	    (desc->height == config->rgb_mode.panelHeight) && (desc->pitch == desc->width)) {
```
ngphibang, line 94: There are some error here. Should be

```
	if ((x == 0) && (y == 0) && (desc->width == config->rgb_mode.panelWidth) &&
	    (desc->height == config->rgb_mode.panelHeight) && (desc->pitch == desc->width * dev_data->pixel_bytes)) {
```

Maybe checking only the resolution (width and height) is enough to know that the buffer is at full LCD resolution so can use the buffer directly, no need to check pitch here.

Also, line 87 need to be changed too (I cannot comment at the line):

`__ASSERT((desc->pitch * desc->height) <= desc->buf_size, "Input buffer too small");`

drivers/display/display_mcux_elcdif.c
```diff
@@ -126,7 +128,7 @@ static int mcux_elcdif_write(const struct device *dev, const uint16_t x, const u
 
 		for (h_idx = 0; h_idx < desc->height; h_idx++) {
 			memcpy(dst, src, dev_data->pixel_bytes * desc->width);
```
ngphibang, line 130: Can we replace by pitch now ?

`memcpy(dst, src, desc->pitch);`

ngphibang (commented):
Some comments for eLCDIF that should be applied on eLCDIFv3 as well since this one takes eLCDIF as reference.

drivers/display/display_mcux_elcdif.c
zejiang0jason: Hi, we only copy the necessary data, if copy length is pitch, it might ruin the data in dst, this branch is updating a window in the dst.

ngphibang: Yeah, I know in this branch the desc->width is different from config->rgb_mode.panelWidth, but it's another thing, we have `desc->pitch = dev_data->pixel_bytes * desc->width`, no ? In the next line, you already did the same ?

```
			-  src += dev_data->pixel_bytes * desc->pitch;
			+ src += desc->pitch;
```

zejiang0jason: Hi, `desc->pitch = dev_data->pixel_bytes * desc->width`, this is not always true, there might be pad at the end, which should not be copied.

ngphibang: Yes, you are right.

PR #100751, opened 2025-12-09
drivers: ksz8081: fix reset pin polarity

fabiobaltieri:
The reset pin, like most reset pins, is active low. Fix the driver to treat it as active low and simplify the pin control logic while at it. Fix all current boards using the wrong pin definition and add a note for out of tree users.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -60,6 +60,10 @@ Ethernet
   is needed. There you need to specify the fixed link parameters using the ``default-speeds``
   property (:github:`100454`).
 
+* The ``reset-gpios`` property of :dtcompatible:`microchip,ksz8081` has been
+  reworked to be used as active low, you may have to set the pin as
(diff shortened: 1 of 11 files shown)


pdgendt (dismissed):

doc/releases/migration-guide-4.4.rst
```diff
@@ -60,6 +60,10 @@ Ethernet
   is needed. There you need to specify the fixed link parameters using the ``default-speeds``
   property (:github:`100454`).
 
+* The ``reset-gpios`` property of :dtcompatible:`microchip,ksz8081` has been
+  reworked to be used as active low, you may have to set the pin as
+  ``GPIO_ACTIVE_LOW`` in devicetree.
```
maass-hamburg: github pr reference should be added otherwise fine

fabiobaltieri: oh, I don't dislike that idea, when did we start doing this? good stuff

pdgendt (approved):

maass-hamburg (approved):

PR #99759, opened 2025-11-20
doc: hardware: porting: clarify board porting recommendations

kartben:
This commit introduces some clarfications to the "General Recommendations" in the board porting documentation, in particular it makes it clearer that DT nodes for key peripherals should be configured AND enabled. It also makes it easier to see the set of recommendations at a glance by adding a short summary for each item.

https://builds.zephyrproject.io/zephyr/pr/99759/docs/hardware/porting/board_porting.html#general-recommendations

PR #100752, opened 2025-12-09
Add pit feature for mcxe31x

lucien-nxp:
-   add some necessary static parameter for pit instances
-   Enable counter_basic_api case by using pit0 instance
-   Test counter_basic_api passed

### DIFF
--- a/boards/nxp/frdm_mcxe31b/frdm_mcxe31b.dts
+++ b/boards/nxp/frdm_mcxe31b/frdm_mcxe31b.dts
@@ -159,6 +159,10 @@
 	overdrive = <12>;
 };
 
+&pit_0 {
+	status = "okay";
+};
+
 &pll {
 	status = "okay";
 	workmode = "Integer";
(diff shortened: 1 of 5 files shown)


boards/nxp/frdm_mcxe31b/frdm_mcxe31b.dts
```diff
@@ -159,6 +159,10 @@
 	overdrive = <12>;
 };
 
+&pit_0 {
```
yvesll, line 162: As pit is an optional counter, is it necessary to enable it on board level without any channel enabled?

Holt-Sun (approved):

boards/nxp/frdm_mcxe31b/frdm_mcxe31b.dts
lucien-nxp: I think board level only enable instances, and set some specific parameter for this instance in specific application code(except if some common setting for this instances).  Channel settings should be put into application level should be better because this should be determined by application code.

Holt-Sun (dismissed):

boards/nxp/frdm_mcxe31b/frdm_mcxe31b.dts
yvesll: The board dts guide has been updated since I last viewed, need a confirm https://github.com/zephyrproject-rtos/zephyr/pull/99759#discussion_r2622499535
Anyway, it's a tiny improvement and will not block this pr.

yvesll (approved):

butok (approved):

nordic-krch (approved):

PR #100758, opened 2025-12-09
net: latmon: Fix net API use

aescolar:
In d45cd6716bbab3a805e3a5fd461934f0dcdc13e5 the mayority of the Zephyr networking code was changed to use the Zephyr native net_/zsock_ prefixed types, but some symbols were missed. 
Let's change them.

Without these fixes/changes the code still builds in most cases as we are by now setting CONFIG_NET_NAMESPACE_COMPAT_MODE.
But when this is not set, things will fail to build.

And change the respective sample app to use native networking API directly in the one call it wasn't.


rlubos (dismissed):

pdgendt (dismissed):

jukkar (commented):
LGTM. One reason why the library code was not noticed to contain POSIX symbols is that `tests/net/all/prj.conf` is missing enabling LATMON i.e., `CONFIG_NET_LATMON=y`
While at it, could you add a commit to this PR enabling that?

aescolar:
> tests/net/all/prj.conf is missing enabling LATMON i.e., CONFIG_NET_LATMON=y
While at it, could you add a commit to this PR enabling that?

Done :)

pdgendt (dismissed):

kartben (dismissed):

aescolar:
The last push was just a rebase + an extra commit with a fix for the issue which was just introduced in main in the openthread integration which also triggered in this PR.

pdgendt (approved):

jhedberg (approved):

rlubos (approved):

aescolar:
I labelled the PR trivial so we can skip the few last minutes of the 48h if we want to. Any release engineer can merge it now.
The changes since the last batch of approvals was just the hotfix for the openthread module integration issue in main which also triggered in this PR. This last fix is conceptually trivial.
Before the PR had been approved by both networking maintainers.

PR #100759, opened 2025-12-09
drivers: gpio: mcp23xxx: fix drive mode validation for input pins

cbrake:
The drive mode validation added in commit ede19a4337f was rejecting valid configurations for input pins. The validation ensures that GPIO_SINGLE_ENDED and GPIO_LINE_OPEN_DRAIN flags match the hardware capabilities of each MCP23xxx variant (push-pull vs open-drain).

However, this validation was applied to all pin configurations, including input pins. Input pins do not have a drive mode, so applications typically configure them without setting drive mode flags. This caused the validation to incorrectly reject input pin configurations with -ENOTSUP.

Fix by only validating drive mode flags when GPIO_OUTPUT is set. Input pins now configure successfully, while output pins still receive proper validation to ensure the requested drive mode matches the hardware capabilities.

Fixes: ede19a4337f ("drivers: mcp23xxx: add support for open-drain chip variants")

### DIFF
--- a/drivers/gpio/gpio_mcp23xxx.c
+++ b/drivers/gpio/gpio_mcp23xxx.c
@@ -180,10 +180,18 @@ static int mcp23xxx_pin_cfg(const struct device *dev, gpio_pin_t pin, gpio_flags
 
 	k_sem_take(&drv_data->lock, K_FOREVER);
 
-	if ((bool)(flags & GPIO_SINGLE_ENDED) != config->is_open_drain ||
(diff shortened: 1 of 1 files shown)


cbrake:
Without this fix, I get the following error when trying to use the MCP23018 IOs as inputs:

`<err> z_dc: Failed to configure GPIO pin 0: -134`

henrikbrixandersen (approved):

MaureenHelm (approved):

PR #100760, opened 2025-12-09
TLS credentials: generate key

maxd-nordic:
Add support to generate a key on-device.
Also: print usage instructions to make commands less painful to use.

We have a usecase to generate p256 keys for use with credentials shell on-device.
Possibly, the option to read the private key should be behind a Kconfig too.

### DIFF
--- a/subsys/net/lib/tls_credentials/tls_credentials_shell.c
+++ b/subsys/net/lib/tls_credentials/tls_credentials_shell.c
@@ -16,6 +16,15 @@ LOG_MODULE_REGISTER(tls_credentials_shell, CONFIG_TLS_CREDENTIALS_LOG_LEVEL);
 #include <strings.h>
 #include <ctype.h>
 
+#if defined(CONFIG_PSA_CRYPTO)
+#include <psa/crypto.h>
+#include <psa/crypto_extra.h>
+#endif /* defined(CONFIG_PSA_CRYPTO)*/
+
+#if defined(MBEDTLS_PEM_WRITE_C)
+#include <mbedtls/pem.h>
+#endif /* defined(MBEDTLS_PEM_WRITE_C)*/
+
 enum cred_storage_fmt {
 	/* Credential is stored as a string and will be passed between the shell and storage
 	 * unmodified.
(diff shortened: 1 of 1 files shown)


subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
 }
 
+#if defined(CONFIG_PSA_CRYPTO)
+/* Key type parameter for generate_ecdsa_keypair */
+enum ecdsa_key_type {
+	ECDSA_P256 = 0, /* secp256r1 / P-256 */
+};
+
+/* Generate ECDSA keypair with specified key type.
+ * Returns the generated key handle in user_keypair_id.
```
yvesll: The return type is not clear. It may be `psa_status_t` or `errno`

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+	return err;
+}
+#endif /* defined(CONFIG_PSA_CRYPTO)*/
+
+/* Generate a private key on device using ECDSA and store it as a credential */
+static int tls_cred_cmd_generate_key(const struct shell *sh, size_t argc, char *argv[])
+{
+#if defined(CONFIG_PSA_CRYPTO)
+	int err = 0;
+	uint32_t keypair_id = 0;
```
yvesll:
```suggestion
	psa_key_id_t keypair_id = 0;
```

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+
+		psa_reset_key_attributes(&key_attributes);
+
+		err = 0;
+	}
+	return err;
+}
+#endif /* defined(CONFIG_PSA_CRYPTO)*/
+
+/* Generate a private key on device using ECDSA and store it as a credential */
```
yvesll: Same for return type here.

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+		psa_set_key_lifetime(&key_attributes, PSA_KEY_LIFETIME_VOLATILE);
+		psa_set_key_algorithm(&key_attributes, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
+		psa_set_key_type(&key_attributes,
+				 PSA_KEY_TYPE_ECC_KEY_PAIR(ecc_family));
+		psa_set_key_bits(&key_attributes, key_bits);
+
+		/* Generate the keypair */
+		status = psa_generate_key(&key_attributes, user_keypair_id);
+		if (status != PSA_SUCCESS) {
+			LOG_INF("psa_generate_key() failed! (Error: %d)", status);
```
yvesll:
```suggestion
			LOG_ERR("psa_generate_key() failed! (Error: %d)", status);
```

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+	if (user_keypair_id != NULL) {
+		psa_key_attributes_t key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+		psa_status_t status;
+		psa_ecc_family_t ecc_family;
+		size_t key_bits;
+
+		/* Initialize PSA Crypto */
+		status = psa_crypto_init();
+		if (status != PSA_SUCCESS) {
+			LOG_INF("psa_crypto_init() failed! (Error: %d)", status);
```
yvesll:
```suggestion
			LOG_ERR("psa_crypto_init() failed! (Error: %d)", status);
```

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +892,180 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+	int err = -1;
+
+	if (user_keypair_id != NULL) {
+		psa_key_attributes_t key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+		psa_status_t status;
+		psa_ecc_family_t ecc_family;
+		size_t key_bits;
+
+		/* Initialize PSA Crypto */
+		status = psa_crypto_init();
```
yvesll, line 927: It would be better to add a mutex lock.

maxd-nordic: psa_crypto_init can be called any number of times and will just return if already initialized. No mutex necessary.

yvesll: Yes, but it's more safe in multi-thread environment

subsys/net/lib/tls_credentials/tls_credentials_shell.c
```diff
@@ -840,6 +905,178 @@ static int tls_cred_cmd_list(const struct shell *sh, size_t argc, char *argv[])
...
+#if defined(CONFIG_PSA_CRYPTO)
+/* Key type parameter for generate_ecdsa_keypair */
+enum ecdsa_key_type {
+	ECDSA_P256 = 0, /* secp256r1 / P-256 */
+};
+
+/* Generate ECDSA keypair with specified key type.
+ * Returns 0 on success or negative error code.
+ */
+static psa_status_t generate_ecdsa_keypair(psa_key_id_t *user_keypair_id,
```
yvesll: IMO, keep int type. As you have log the psa_status_t value, just return a `ECANCELED` or any other better zephyr errno.
For PSA error code in line 929, shell_fprintf may be more clear.

subsys/net/lib/tls_credentials/tls_credentials_shell.c
maxd-nordic: what do you mean? only one thread can access this function.

yvesll: Other threads may call `psa_crypto_init()`, not only `generate_ecdsa_keypair`.
Anyway, as the doc said it can be called in multiple times, it's just a suggestion.

PR #100625, opened 2025-12-06
drivers: flash: Fix flash host type for CI builds

marekmatej:
This fixes the APPCPU synthetic builds.

PR #100765, opened 2025-12-09
drivers: flash: esp32: only link APIs when compiled in

fabiobaltieri:
Heya, this was broken in the weekly, tested with

`west build -p -b esp32_devkitc/esp32/appcpu tests/subsys/mgmt/mcumgr/fs_mgmt_hash_supported`

--

The flash_esp32_read, flash_esp32_write and flash_esp32_erase are only implemented if CONFIG_ESP_FLASH_HOST is defined, do not try to reference them when they are not compiled in.

### DIFF
--- a/drivers/flash/flash_esp32.c
+++ b/drivers/flash/flash_esp32.c
@@ -760,7 +760,7 @@ static DEVICE_API(flash, flash_esp32_driver_api) = {
 	.read = flash_esp32_read_async,
 	.write = flash_esp32_write_async,
 	.erase = flash_esp32_erase_async,
-#else
+#elif CONFIG_ESP_FLASH_HOST
 	.read = flash_esp32_read,
 	.write = flash_esp32_write,
 	.erase = flash_esp32_erase,


sylvioalves:
@fabiobaltieri, we have already https://github.com/zephyrproject-rtos/zephyr/pull/100625 in place. @marekmatej, please check.

drivers/flash/flash_esp32.c
```diff
@@ -760,7 +760,7 @@ static DEVICE_API(flash, flash_esp32_driver_api) = {
 	.read = flash_esp32_read_async,
 	.write = flash_esp32_write_async,
 	.erase = flash_esp32_erase_async,
-#else
+#elif CONFIG_ESP_FLASH_HOST
```
marekmatej, line 763: Very clever solution, but I think it is hiding the root cause. If you don't mind, I prefer to fix this at Kconfig level.

marekmatej: ~~OR, maybe both fixes should go. Kconfig to define the exact conditions for the host flash, following with `elif` instead of `else` as a safety margin.~~

fabiobaltieri: Yeah I mean I can't see how the current code is correct, https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/flash/flash_esp32.c#L371 and the others are under a `#ifdef CONFIG_ESP_FLASH_HOST`, any reference of it has to be under the same condition or it'll lead to a build failure.

If the configuration is not valid for some other reason then it should either be made impossible with Kconfig (whcih I guess you are doing already) or have a corresponding `BUILD_ASSERT`.

marekmatej: Thanks for pointing this out. The relations between `HOST` and `ASYNC` are not clear. I'll fix it.

marekmatej:
@fabiobaltieri I'll fix this in #100625, you can close this one here please.

PR #100769, opened 2025-12-09
tests: assert_custom_header: only build on some platforms

### LINKED ISSUE #100733: tests/subsys/debug/assert_custom_header fails in multiple platforms
### Describe the bug

tests/subsys/debug/assert_custom_header fails to build in multiple (33) platforms

### Regression

- [x] This is a regression.

### Steps to reproduce

1. $ mkdir build && cd build
2. $ cmake -GNinja -DBOARD=nrf5340dk/nrf5340/cpuapp/ns ../tests/subsys/debug/assert_custom_header/
3. $ ninja
4. see error

### Relevant log output

```shell
In file included from build/tfm/../zephyr/misc/generated/syscalls_links/include/zephyr/irq_multilevel.h:15,
                 from build/tfm/../zephyr/misc/generated/syscalls_links/include/zephyr/devicetree.h:20,
                 from modules/trusted-firmware-m/nordic/include/RTE_Device.h:14,
                 from modules/tee/tf-m/trusted-firmware-m/platform/ext/target/nordic_nrf/common/core/./nrfx_config.h:35,
                 from modules/hal/nordic/nrfx/nrfx.h:37,
                 from modules/tee/tf-m/trusted-firmware-m/platform/ext/target/nordic_nrf/common/core/common/cmsis.h:35,
                 from modules/tee/tf-m/trusted-firmware-m/platform/include/tfm_hal_device_header.h:14,
                 from modules/tee/tf-m/trusted-firmware-m/bl2/ext/mcuboot/bl2_main.c:21:
build/tfm/../zephyr/misc/generated/syscalls_links/include/zephyr/sys/__assert.h:154:10: fatal error: zephyr_custom_assert.h: No such file or directory
  154 | #include <zephyr_custom_assert.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
```
33 failures (33 targets fail to build it) in weekly run
https://github.com/zephyrproject-rtos/zephyr/actions/runs/19997797171 
For ex.:
https://github.com/zephyrproject-rtos/zephyr/actions/runs/19997797171/job/57348087573#step:14:6231

### Impact

### Environment

- OS: Any
- Zephyr SDK 0.17.4
- SHA main since ada3aad811d2f74ff50a48843797858c334dffa3

### Additional Context

Test added in https://github.com/zephyrproject-rtos/zephyr/pull/96514
Fails since it was introduced

fabiobaltieri:
Breaks in weekly, example ` west build -p -b thingy53/nrf5340/cpuapp/ns tests/subsys/debug/assert_custom_header`

---

This test does not build correctly on some modules, namely TFM, just run
it on a few platform, that should be enough to validate that the feature
is working correctly.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/100733

### DIFF
--- a/tests/subsys/debug/assert_custom_header/testcase.yaml
+++ b/tests/subsys/debug/assert_custom_header/testcase.yaml
@@ -1,6 +1,11 @@
 common:
   integration_platforms:
     - native_sim
+  platform_allow:
+    - native_sim
+    - native_sim/native/64
+    - qemu_x86
+    - qemu_cortex_m3
 
 tests:
   debug.assert_custom_header:


JarmouniA (dismissed):

dsseng:
This only affects nRF platforms (`mps2/an521/cpu0/ns` and `max32657evkit/max32657/ns` aren't affected for sure), and this is the culprit:

https://github.com/zephyrproject-rtos/zephyr/blob/d41d915330d0b6ea21b273294db7dc7f1be75423/modules/trusted-firmware-m/nordic/include/RTE_Device.h#L14

I have not found patching viable (will require quite some reorganisation of dt header files), but perhaps we should make `CONFIG_ASSERT_CUSTOM_HEADER` conflict with nRF ns targets, as this might impact users as well.

tests/subsys/debug/assert_custom_header/testcase.yaml
```diff
@@ -1,4 +1,5 @@
 common:
+  filter: not CONFIG_BUILD_WITH_TFM
```
nashif: I do not think this is something we need to build/validate on evey platform just turn the `integration_platforms` below into `platform_allow`

nashif (changes requested):

tests/subsys/debug/assert_custom_header/testcase.yaml
tomi-font: > `platform_allow: <list of platforms>`
>
>    Set of platforms that this test scenario should only be run for. Do not use this option to limit testing or building in CI due to time or resource constraints, this option should only be used if the test or sample can only be run on the allowed platform and nothing else.
>
> `integration_platforms: <YML list of platforms/boards>`
>
>    This option limits the scope to the listed platforms when twister is invoked with the --integration option. Use this instead of platform_allow if the goal is to limit scope due to timing or resource constraints.

https://docs.zephyrproject.org/latest/develop/test/twister.html

I get it I guess you're talking about the CI scope that is run outside of PRs, but shouldn't there be a different mechanism for limiting the combinations there as well?

tomi-font (dismissed):

tests/subsys/debug/assert_custom_header/testcase.yaml
nashif: regardless of when CI is run, this is not something that needs to be evaluated and run on every platforms combo we have in zephyr, this is not something that is hardware specific, this just verifies include infrastructure and can be considered a unit test that is enough to only build for one platform for coverage. The way it is done now:

```
real	1m4.019s
user	16m33.648s
sys	3m31.487s
```   

when limiting to native_sim:

```
real	0m5.892s
user	0m8.836s
sys	0m2.734s
```

so it is ~100 time faster and for a good reason :)

fabiobaltieri: ok I've redone this to limit the test to few platforms

fabiobaltieri:
Reworked to add a platform_allow list instead

nashif (approved):

tomi-font (approved):

PR #100770, opened 2025-12-09
net: l2: Add new stats API

krish2718:
This new API takes a category as input primarily to bypass vendor stats query (to avoid FW interactions) as this is called per-packet by the networking stack.

### DIFF
--- a/include/zephyr/net/ethernet.h
+++ b/include/zephyr/net/ethernet.h
@@ -529,6 +529,16 @@ struct ethernet_config {
 
 /** @endcond */
 
+/** Ethernet statistics type (bitmap) */
+enum ethernet_stats_type {
+	/** Common statistics only (excludes vendor statistics) */
+	ETHERNET_STATS_TYPE_COMMON = BIT(0),
+	/** Vendor statistics only */
+	ETHERNET_STATS_TYPE_VENDOR = BIT(1),
+	/** All statistics */
+	ETHERNET_STATS_TYPE_ALL = 0xFFFFFFFFU,
+};
+
 /** Ethernet L2 API operations. */
 struct ethernet_api {
 	/**
@@ -543,6 +553,14 @@ struct ethernet_api {
 	 */
 #if defined(CONFIG_NET_STATISTICS_ETHERNET)
 	struct net_stats_eth *(*get_stats)(const struct device *dev);
+
+	/** Optional function to collect ethernet specific statistics with
+	 * type filter. If NULL, get_stats() will be called instead, which
+	 * is equivalent to calling this with ETHERNET_STATS_TYPE_ALL.
+	 * @param type Bitmask of ethernet_stats_type values.
+	 */
+	struct net_stats_eth *(*get_stats_type)(const struct device *dev,
+						 uint32_t type);
 #endif
 
 	/** Start the device */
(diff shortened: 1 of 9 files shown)


include/zephyr/net/ethernet.h
```diff
@@ -529,6 +529,16 @@ struct ethernet_config {
 
 /** @endcond */
 
+/** Ethernet statistics type */
+enum ethernet_stats_type {
+	/** Native statistics only (excludes vendor statistics) */
+	ETHERNET_STATS_TYPE_NATIVE = 0,
```
rlubos: I think we should rather call this `ETHERNET_STATS_TYPE_COMMON` (as common statistics for all drivers), native is a bit ambiguous.

include/zephyr/net/ethernet.h
```diff
@@ -543,6 +553,14 @@ struct ethernet_api {
 	 */
 #if defined(CONFIG_NET_STATISTICS_ETHERNET)
 	struct net_stats_eth *(*get_stats)(const struct device *dev);
+
+	/** Collect optional ethernet specific statistics with type filter.
+	 * This pointer should be set by driver if statistics needs to be
+	 * collected for that driver. If NULL, get_stats will be called
+	 * with ETHERNET_STATS_TYPE_ALL.
+	 */
```
rlubos: The description is a bit vague to me

> Collect optional ethernet specific statistics with type filter.

So are the statistics optional or the function? Shouldn't this be something like `Optional function to collect ethernet specific statistics with type filter.`?

> If NULL, get_stats will be called with ETHERNET_STATS_TYPE_ALL.

`get_stats()` will not be called with `ETHERNET_STATS_TYPE_ALL` as it doesn't accept this parameter. Should it rather mean  that `If NULL, get_stats() will be called, which corresponds to get_stats2() call with ETHERNET_STATS_TYPE_ALL filter`.

subsys/net/l2/ethernet/eth_stats.h
```diff
@@ -25,7 +25,12 @@ static inline void eth_stats_update_bytes_rx(struct net_if *iface,
 		return;
```
rlubos: Should we rather check `if (api->get_stats == NULL && api->get_stats2 == NULL)` before returning early? We only need one variant available. Currently, nRF driver won't collect stats at all as it doesn't implement `get_stats()` now. This applies to all other functions as well.

include/zephyr/net/ethernet.h
```diff
@@ -543,6 +553,14 @@ struct ethernet_api {
...
+	struct net_stats_eth *(*get_stats2)(const struct device *dev,
```
sylvioalves: Could that be renamed to `get_stats_vendor` or similar?. Unless there is some kind of naming rule,  I think adding suffix `2` can be improved.

krish2718: How about `get_stats_type` where type can be extended in the feature (I will convert from enum to a bitmap)?

sylvioalves: Even better! Thanks.

subsys/net/l2/ethernet/eth_stats.h
krish2718: oops, it's a bug, fixed it locally, thanks.

jukkar (approved):

rlubos (approved):

sylvioalves (approved):

Issue #98647, opened 2025-10-31
meta: Modules should not contain usage of `zephyr_compile_definitions`


nordicjm:
### Describe the bug

It has been noticed that many modules are using the `zephyr_compile_definitions` CMake function to add definitions to a build, these defines are **global** which means **every** file will get them, so when building hello world, the main.c file will get all these flags, none of which are needed for it to be built.
If flags are needed, they are likely needed by a module which is a **library** and should be using the `zephyr_library_compile_options` function which adds the defines to the current library only and does not pollute other files with defines.

The following modules have been found in breach of this and need to be fixed:

- [ ] zcbor @de-nordic
modules/zcbor/CMakeLists.txt:  zephyr_compile_definitions_ifdef(CONFIG_ZCBOR_CANONICAL ZCBOR_CANONICAL)
modules/zcbor/CMakeLists.txt:  zephyr_compile_definitions_ifdef(CONFIG_ZCBOR_STOP_ON_ERROR ZCBOR_STOP_ON_ERROR)
modules/zcbor/CMakeLists.txt:  zephyr_compile_definitions_ifdef(CONFIG_ZCBOR_VERBOSE ZCBOR_VERBOSE)
modules/zcbor/CMakeLists.txt:  zephyr_compile_definitions_ifdef(CONFIG_ZCBOR_ASSERT ZCBOR_ASSERTS)
modules/zcbor/CMakeLists.txt:  zephyr_compile_definitions_ifdef(CONFIG_ZCBOR_BIG_ENDIAN ZCBOR_BIG_ENDIAN)

- [ ] nrf_wifi @krish2718
modules/nrf_wifi/bus/CMakeLists.txt:    zephyr_compile_definitions(

- [ ] nanopb @pdgendt
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(
modules/nanopb/CMakeLists.txt:  zephyr_compile_definitions_ifdef(

[...]

PR #100779, opened 2025-12-10
modules: hal_silabs: wiseconnect: CMakeLists fix

fimohame:
This fix addresses a PM failure caused by an inconsistency between the device tree and the build system. The device tree was updated with `reg = <0x00000400 DT_SIZE_K(319)>;`, but the corresponding value for `SL_SI91X_SI917_RAM_MEM_CONFIG` in CMakeLists.txt was not updated, leading to the problem. The build system now automatically sets `SL_SI91X_SI917_RAM_MEM_CONFIG` based on the sram0 RAM size from the device tree using `dt_prop()` and `list(GET)`, ensuring the configuration matches the device tree and preventing future mismatches.

### DIFF
--- a/modules/hal_silabs/wiseconnect/CMakeLists.txt
+++ b/modules/hal_silabs/wiseconnect/CMakeLists.txt
@@ -5,6 +5,25 @@ set(SISDK_DIR ${ZEPHYR_HAL_SILABS_MODULE_DIR}/simplicity_sdk)
 set(WISECONNECT_DIR ${ZEPHYR_HAL_SILABS_MODULE_DIR}/wiseconnect)
 set(COMMON_DIR ${ZEPHYR_HAL_SILABS_MODULE_DIR}/simplicity_sdk/platform/common)
 
+function(wiseconnect_get_ram_config node_var out_var)
+  # Get reg property from device tree node
+  dt_nodelabel(node_path NODELABEL "${node_var}")
+  dt_prop(reg_property PATH "${node_path}" PROPERTY "reg")
+  list(GET reg_property 1 size_bytes)
+  math(EXPR size_kb "${size_bytes} / 1024")
+
+  # Set output variable based on size
+  if(size_kb EQUAL 195)
+    set(${out_var} 1 PARENT_SCOPE)
+  elseif(size_kb EQUAL 255)
+    set(${out_var} 2 PARENT_SCOPE)
+  elseif(size_kb EQUAL 319)
+    set(${out_var} 3 PARENT_SCOPE)
+  else()
+    message(FATAL_ERROR "Unsupported RAM size config: ${size_kb}KB")
+  endif()
+endfunction()
+
 # Keep these values sync with
 # components/device/silabs/si91x/mcu/core/chip/component/siwg917*.slcc
 zephyr_compile_definitions(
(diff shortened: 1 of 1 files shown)


modules/hal_silabs/wiseconnect/CMakeLists.txt
```diff
@@ -156,9 +156,30 @@ if(CONFIG_BT_SILABS_SIWX91X)
...
+     set(RAM_MEM_CONFIG 1)
+  elseif(sram0_size_kb EQUAL 255)
+     set(RAM_MEM_CONFIG 2)
+  elseif(sram0_size_kb EQUAL 319)
+     set(RAM_MEM_CONFIG 3)
+  else()
+     message(FATAL_ERROR "Unsupported RAM size config: ${sram0_size_kb}")
+  endif()
+
   zephyr_compile_definitions(
```
nordicjm, line 179: change `zephyr_compile_definitions` in this file to `zephyr_library_compile_definitions`

nordicjm (dismissed):

modules/hal_silabs/wiseconnect/CMakeLists.txt
```diff
@@ -156,9 +175,10 @@ if(CONFIG_BT_SILABS_SIWX91X)
...
 
 if(CONFIG_SILABS_SIWX91X_NWP)
+  wsieconnect_get_ram_config("sram0" RAM_MEM_CONFIG)
   zephyr_compile_definitions(
     SLI_SI91X_ENABLE_OS
-    SL_SI91X_SI917_RAM_MEM_CONFIG=2
+    SL_SI91X_SI917_RAM_MEM_CONFIG=${RAM_MEM_CONFIG}
     SL_WIFI_COMPONENT_INCLUDED # Depite de the name, required for everything
   )
   zephyr_include_directories(
```
nordicjm, line 184: and this to `zephyr_library_include_directories`

modules/hal_silabs/wiseconnect/CMakeLists.txt
jerome-pouiller: I wish we can. Unfortunately, these definitions are used in the headers files that are included in Zephyr drivers. `zephyr_library_compile_definitions()` would cause mismatches difficult to identify.

I would like to clean up the Silabs SDK code to avoid this problem, but it is huge task.

Regarding the current PR, `SL_SI91X_SI917_RAM_MEM_CONFIG` was already declared globally in the original code. I believe a such change is not in the scope of this PR.

modules/hal_silabs/wiseconnect/CMakeLists.txt
jerome-pouiller: Ditto, the headers used by Zephyr depends on these directories. `zephyr_library_include_directories()` would break the build.

jerome-pouiller (dismissed):

modules/hal_silabs/wiseconnect/CMakeLists.txt
nordicjm: they need to be cleaned up and removed as per https://github.com/zephyrproject-rtos/zephyr/issues/98647 will accept that it will be done in a future PR

modules/hal_silabs/wiseconnect/CMakeLists.txt
nordicjm: as in, network drivers? If so, can add them directly to the driver's CMake file that needs them, or possibly using zephyr_library_amend() here if they are enabled (unsure of @tejlmand's thoughts on that)

jhedberg (dismissed):

jerome-pouiller (approved):

jhedberg (approved):

PR #100782, opened 2025-12-10
soc: renesas: rzv: Retrieve itcm, dtcm from dtsi for mpu settings

nhutnguyenkc:
Retrieve itcm and dtcm size and base address from dtsi for mpu settings instead of hardcode.


tiennguyenzg (approved):

binhnguyen2434 (approved):
LGTM

PR #100783, opened 2025-12-10
tests: subsys: secure_storage: psa: its: Exclude any nrf54h20 target

### LINKED ISSUE #100781: tests/subsys/secure_storage/psa/its fails to compile for nrf54h20dk@0.9.0/nrf54h20/cpuflpr/xip
### Describe the bug

Test fails to compile after https://github.com/zephyrproject-rtos/zephyr/commit/3ea2951bc01c92466a5fed228f0c70aa96dfec72 on CI: https://github.com/zephyrproject-rtos/zephyr/actions/runs/20026231653/job/57424541399?pr=100505

Should that target be excluded?  nrf54h20dk@0.9.0/nrf54h20/cpuapp and nrf54h20dk@0.9.0/nrf54h20/cpurad already are.

### Regression

- [x] This is a regression.

### Steps to reproduce

Run `west twister -p nrf54h20dk@0.9.0/nrf54h20/cpuflpr/xip -s secure_storage.psa.its.secure_storage.store.settings.64-bit_uids`

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

nordic-krch:
6cc7bdb58a excluded cpuapp and cpurad targets but any nrf54h20dk target should be included. cpuppr and cpuflpr do not compile after 3ea2951bc01c.

CI is red due to that: https://github.com/zephyrproject-rtos/zephyr/actions/runs/20026231653/job/57424541403?pr=100505

Fixes: #100781

### DIFF
--- a/tests/subsys/secure_storage/psa/its/testcase.yaml
+++ b/tests/subsys/secure_storage/psa/its/testcase.yaml
@@ -6,6 +6,10 @@ common:
     - qemu_cortex_m0 # settings subsystem initialization fails
     - nrf54h20dk/nrf54h20/cpuapp
     - nrf54h20dk/nrf54h20/cpurad
+    - nrf54h20dk/nrf54h20/cpuppr
+    - nrf54h20dk/nrf54h20/cpuppr/xip
+    - nrf54h20dk/nrf54h20/cpuflpr
+    - nrf54h20dk/nrf54h20/cpuflpr/xip
   timeout: 600
   tags:
     - psa.secure_storage


tests/subsys/secure_storage/psa/its/testcase.yaml
```diff
@@ -6,6 +6,10 @@ common:
     - qemu_cortex_m0 # settings subsystem initialization fails
     - nrf54h20dk/nrf54h20/cpuapp
     - nrf54h20dk/nrf54h20/cpurad
+    - nrf54h20dk/nrf54h20/cpuppr
```
nordicjm, line 9: this is not maintainable, there should be a proper check on this to ensure it only runs on devices with a flash controller

nordic-krch: CI is red now. Anyone maintaining this sample could propose something?

dsseng: I believe I know the solution, give me 30 minutes, will look it up once at the computer

dsseng: zms test should probably have the tag `zms`, and `settings.*` already depend on settings' Kconfig options it seems

tomi-font: Agreed that we need a proper fix and not just exclude all the board targets which are now failing since https://github.com/zephyrproject-rtos/zephyr/commit/3ea2951bc01c92466a5fed228f0c70aa96dfec72. FYI I have asked @petejohanson to have a look (and he said he'd do so "first thing tomorrow") but @dsseng if you can come up with a fix it'd be much appreciated.

tomi-font: > zms test should probably have the tag zms, and settings.* already depend on settings' Kconfig options it seems

Yes, the problem we have now is that with what is introduced in 3ea2951bc01c92466a5fed228f0c70aa96dfec72 the check to decide whether to erase the flash (`CONFIG_FLASH_PAGE_LAYOUT`) is manifestly not enough for certain board targets.

Issue #93769, opened 2025-07-28
Bluetooth: Host: BT_DEVICE_NAME should be mutually exclusive with BT_DEVICE_NAME_DYNAMIC


Thalley:
### Summary

The 2 Kconfig options should be mutually exclusive, i.e. if `CONFIG_BT_DEVICE_NAME_DYNAMIC=y`, then `CONFIG_BT_DEVICE_NAME` should not be set-able. 

The reason for this is to avoid cases where e.g. `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` and applications mistakenly use  `CONFIG_BT_DEVICE_NAME` when e.g. advertising instead of `bt_get_name` and may thus advertise the wrong name. 

### Describe the solution you'd like

Modify 

```c
config BT_DEVICE_NAME
	string "Bluetooth device name"
	default "Zephyr"
	help
	  Bluetooth device name. Name can be up to 248 bytes long (excluding
	  NULL termination). Can be empty string.
```

to be 

```c
config BT_DEVICE_NAME
	string "Bluetooth device name" if !BT_DEVICE_NAME_DYNAMIC
	default "Zephyr"
	default "" if BT_DEVICE_NAME_DYNAMIC
	help
	  Bluetooth device name. Name can be up to 248 bytes long (excluding
	  NULL termination). Can be empty string.
```

or

```c
config BT_DEVICE_NAME
	string "Bluetooth device name" 
	depends on !BT_DEVICE_NAME_DYNAMIC
	default "Zephyr"
	help
	  Bluetooth device name. Name can be up to 248 bytes long (excluding
	  NULL termination). Can be empty string.
```

axel-lebourhis:
Trickier than what it seems at first glance. Even host files like hci_core.c assume `CONFIG_BT_DEVICE_NAME` is always set and use it as a default value in bt_enable.
I saw there's already an open PR for this issue. If for any reason it doesn’t go through or needs additional work, I’d be happy to help or pick it up if needed — just let me know

Thalley:
> Trickier than what it seems at first glance. Even host files like hci_core.c assume `CONFIG_BT_DEVICE_NAME` is always set and use it as a default value in bt_enable. I saw there's already an open PR for this issue. If for any reason it doesn’t go through or needs additional work, I’d be happy to help or pick it up if needed — just let me know

Agreed. Removed the Good First Issue label. Requires quite some significant work, and probably even a migration guide entry.

[...]

PR #100785, opened 2025-12-10
Bluetooth: Use macro to replace sizeof(CONFIG_BT_DEVICE_NAME) - 1

LingaoM:
1. Add define single marco `BT_DEVICE_NAME_LEN` to replace `(sizeof(CONFIG_BT_DEVICE_NAME) - 1)`
2. replace everywhere.


nordicjm (dismissed):
OK for MCUmgr

include/zephyr/bluetooth/hci_types.h
sjanc: I'm not sure if this is correct,  local_name parameter is defined as utf8s{248}

LingaoM: Yes, it is. thanks.

A UTF-8 encoded User-Friendly Descriptive Name for the device with type utf8s{248} (see [Vol 1] Part E, Section 2.9.3).

PavelVPV (dismissed):

include/zephyr/bluetooth/bluetooth.h
Thalley: I wonder if macros are supposed to be exempt from https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-1-conditional-compilation

Thalley: This might be a bit misleading when the device name is dynamic. Ideally this would never be the case (see https://github.com/zephyrproject-rtos/zephyr/issues/93769), but it is today. 

So this macro arguably shouldn't be allowed when `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` as it may return a length that is not the device name length. 

One possible solution is to define it as
```c
#define BT_DEVICE_NAME_LEN      (strlen(bt_get_name))
```

when `CONFIG_BT_DEVICE_NAME_DYNAMIC=y`

Thalley (changes requested):

include/zephyr/bluetooth/bluetooth.h
LingaoM: > The 2 Kconfig options should be mutually exclusive, i.e. if CONFIG_BT_DEVICE_NAME_DYNAMIC=y, then CONFIG_BT_DEVICE_NAME should not be set-able.

I agree with your idea that these two macros should be mutually exclusive, but the problem this PR aims to solve is to replace all samples usages and provide a unified naming convention;

LingaoM: btw: 
> #define BT_DEVICE_NAME_LEN      (strlen(bt_get_name))
when CONFIG_BT_DEVICE_NAME_DYNAMIC=y

Maybe be build failed in this PR, because macro:`BT_DEVICE_NAME_LEN` used in preprocessing not only build stage.

Thalley: > Maybe be build failed in this PR, because macro:BT_DEVICE_NAME_LEN used in preprocessing not only build stage.

Indeed, but I think that's a good thing. If `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` then the user should never assume that it's a compile-time constant, and the macro should never be used as such.

Thalley: I'm not convinced an `else` clause here is a good solution. If the host isn't enabled, then the compiled source file should not use `BT_DEVICE_NAME_LEN`. So I'd argue that using `BT_DEVICE_NAME_LEN` when `CONFIG_BT_HCI_HOST=n` should result in a build error

LingaoM: According like this way:https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/bluetooth/mesh/access.h#L124

LingaoM:
``` C
╭─mi@lingao ~/works/github/zephyrproject/zephyr  ‹bt_dev_name› 
╰─➤  grep -rn "CONFIG_BT_DEVICE_NAME_DYNAMIC=y" 
samples/bluetooth/peripheral/prj.conf:19:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
samples/boards/phytec/reel_board/mesh_badge/prj.conf:6:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bluetooth/tester/prj.conf:27:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bluetooth/shell/log.conf:18:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bluetooth/shell/audio.conf:29:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bluetooth/shell/prj.conf:20:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/att/long_read/prj.conf:5:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/att/timeout/prj.conf:5:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/att/retry_on_sec_err/server/prj.conf:6:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/att/open_close/prj.conf:5:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/att/read_fill_buf/server/prj.conf:6:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/misc/conn_stress/peripheral/prj.conf:15:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/misc/conn_stress/peripheral/prj.conf:17:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/host/gatt/device_name/prj.conf:7:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
tests/bsim/bluetooth/ll/edtt/gatt_test_app/prj_llcp.conf:24:CONFIG_BT_DEVICE_NAME_DYNAMIC=y
```

Do you mean that these projects should not be used in this way:
``` C
static const struct bt_data sd[] = {
	BT_DATA(BT_DATA_NAME_COMPLETE, CONFIG_BT_DEVICE_NAME, BT_DEVICE_NAME_LEN),
};
```
and replace in function such like:

``` C
func()
{
    struct bt_data sd[] = {
	    BT_DATA(BT_DATA_NAME_COMPLETE, bt_get_name(), BT_DEVICE_NAME_LEN),
    };
};

Or also define macro like:

#define BT_DEVICE_NAME      IS_ENABLED(CONFIG_BT_DEVICE_NAME_DYNAMIC) ? \
                                bt_get_name(): CONFIG_BT_DEVICE_NAME

func()
{
    struct bt_data sd[] = {
	    BT_DATA(BT_DATA_NAME_COMPLETE, BT_DEVICE_NAME, BT_DEVICE_NAME_LEN),
    };
};

```

Thalley: > Do you mean that these projects should not be used in this way:

Depends; if the name actually changes in those samples/tests then they shouldn't do 
```c
static const struct bt_data sd[] = {
	BT_DATA(BT_DATA_NAME_COMPLETE, CONFIG_BT_DEVICE_NAME, BT_DEVICE_NAME_LEN),
};
```
But if the name never changes, then they probably shouldn't have `CONFIG_BT_DEVICE_NAME_DYNAMIC=y`, as `CONFIG_BT_DEVICE_NAME` will not reflect the actual device name. So there may be multiple samples and tests that wrongly sets `CONFIG_BT_DEVICE_NAME_DYNAMIC=y`, or that they aren't properly implementing the feature. 

That's why I requested the 2 Kconfig options to be mutually exclusive in https://github.com/zephyrproject-rtos/zephyr/issues/93769 so that we can at build time better ensure that it's done correctly

Thalley:
```suggestion
#if defined(CONFIG_BT_DEVICE_NAME)
```

Would it make sense to do this instead? I think that better contains the relevant context

LingaoM: @Thalley Changed, Please take a look, thanks.

LingaoM: Use COND_CODE_1 replace.

Thalley:
```suggestion
						(bt_get_name()),	\
```

I think it makes sense that the user does the casting if necessary. The macro be used in other contexts than setting advertising data

Thalley: I understand that the purpose of this is to help setting advertising data, but the way it is defined makes it appear as a general macro, and that makes it somewhat odd IMO as you'd think you could just do `strlen(BT_DEVICE_NAME)` (but of course that isn't possible for advertising data like it is done most places)

Thalley (commented):
I'd like to get some input from the other contributors/maintainers on this. 

What seemed like a trivial fix (for a trivial "issue"), is now moving into multiple macros that have significantly different behavior depending on whether a Kconfig is enabled or not, while still just being macros that are doing something very simple like `strlen` or `sizeof`. 

While the solution here works, I don't think I really like it, but don't let met be a blocker on this, if it's just me :)

include/zephyr/bluetooth/bluetooth.h
Thalley: It does have the advantage that it will give a build error if you tried to use it with `BT_DATA` when `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` which does help prevent a set of bugs (that we can even see fixed in this PR!)

tests/bluetooth/shell/audio.conf
Thalley, line 29: The shell does support dynamic names, so please revert this line

tests/bluetooth/shell/src/main.c
Thalley, line 33: We can remove this I think, since the shell doesn't use this

tests/bsim/bluetooth/host/misc/conn_stress/peripheral/src/main.c
Thalley:
```suggestion
	sd[0].data_len = BT_DEVICE_NAME_LEN;
	sd[0].data = BT_DEVICE_NAME;
```

Right? `CONFIG_BT_DEVICE_NAME` shouldn't be used with `BT_DEVICE_NAME_LEN` as the name of the two may not be the same

samples/bluetooth/beacon/src/main.c
Thalley: In cases like this, I guess you could remove the original defines and replace their uses with the new macros?

subsys/bluetooth/common/dummy.c
Thalley, line 47: Suggest to add a reference to where this maximum size is defined

subsys/bluetooth/mesh/gatt.h
Thalley, line 19: Interesting that we already had these defined locally in mesh

tests/bsim/bluetooth/ll/advx/src/main.c
Thalley:
```suggestion
	    (strlen(name) == BT_DEVICE_NAME_LEN) &&
```

tests/bsim/bluetooth/host/privacy/central/src/tester.c
Thalley:
```suggestion
	ad.data_len = BT_DEVICE_NAME_LEN;
```

Thalley (changes requested):
Given the built-in compile check in the macros, I may want to reconsider my previous statement. Additionally we already had something like this defined in Mesh, so at least the Mesh people should approve of these :) 

Looked through the remaining part of the PR and noticed a few things that may need to be fixed

include/zephyr/bluetooth/bluetooth.h
kartben, line 99: these could use slightly more informative doxygen comments IMO - consider adding some details re: why/when to use, etc ("Helper macro that can be used to xxx")

Thalley: Agree. How they can/shall be used should be documented, as well as the different behavior depending on whether `CONFIG_BT_DEVICE_NAME_DYNAMIC` is enabled, and how `CONFIG_BT_DEVICE_NAME_DYNAMIC=y` make it impossible to use this with `BT_DATA`

Thalley, line 98:
```suggestion
						(sizeof(CONFIG_BT_DEVICE_NAME) - 1U))),	        \
```

Thalley, line 91:
```suggestion
						(bt_get_name()),	        \
						(CONFIG_BT_DEVICE_NAME))),     \
```

tests/bluetooth/shell/prj.conf
Thalley, line 20: The shell does support dynamic names, so please revert this line

include/zephyr/bluetooth/bluetooth.h
Thalley: https://github.com/zephyrproject-rtos/zephyr/pull/100785#discussion_r2609680184

Thalley (changes requested):

PR #100788, opened 2025-12-10
Added preliminary support for sf32lb58 from Sifli

sprhawk:
Added board definitions to sf32lb58_devkit/sf32lb586/hcpu Added soc definitions to sf32lb58 series
Added related clock control and pinctrl drivers to sf32lb58

The board can run with helloworld with uart_sf32lb

Currently not much documentation, I'll added them later, please review the code first.
In my PR, I have a more sophisticated architecture than SF32LB52 port, because sf32lb58 has two systems: `HPSYS` and `LPSYS` opened to user. And I have a tree-like structures to define the clock control of sf32lb58:
peripheral clocks -> `hpsys_rcc` -> `hxt48`/`hrc48`, etc ... -> `hpsys_aon`
That because I want 1. shared driver between `hxt48` and `lxt32`, `hrc48` and `lrc10` drivers, and have a shared controller which is `hpsys_aon`. and because `hpsys` and `lpsys` shares some same registers layout, so I want to share some drivers between `hpsys` and `lpsys`. 
I know it is more complicated, and don't know whether others like it.

Another problem left is SF32LB58 HCPU is booted from LCPU in flash1, currently I use SiFli SDK built bootloader, flashed by scripts, then flash and debug inside zephyr. I will try to work a `mcuboot` based solution to fix it.


gmarull (changes requested):
Please organize PR accordingly, see how SF32LB52 was introduced.

ck-telecom (commented):
could you rename board name to sf32lb58_devkit_lcd?

sprhawk:
> Please organize PR accordingly, see how SF32LB52 was introduced.

I have re-organized PRs

gmarull (changes requested):
please make sure commits are bisectable, e.g. you're adding a board before anything else...

boards/sifli/sf32lb58_devkit_lcd/sf32lb58_devkit_lcd_sf32lb586_hcpu.yaml
```diff
@@ -11,7 +11,5 @@ toolchain:
   - zephyr
 ram: 1024
 supported:
-  - arduino_gpio
-  - arduino_serial
-  - gpio
+  - uart
```
gmarull, line 14: some commits also do not do what title explains, see eg fdcedd00d5d6de965e94f165b62cd5f396873066

PR #100791, opened 2025-12-10
Enhance network packet filter sample app

jukkar:
Add support for IPv4 / IPv6 address filtering.


include/zephyr/net/net_pkt_filter.h
```diff
@@ -236,19 +236,19 @@ bool npf_remove_all_rules(struct npf_rule_list *rules);
...
-		.result = NET_CONTINUE,                                                            \
-		.priority = (_priority),                                                           \
-		.nb_tests = NUM_VA_ARGS_LESS_1(__VA_ARGS__) + 1,                                   \
-		.tests = {FOR_EACH(Z_NPF_TEST_ADDR, (,), __VA_ARGS__)},                            \
+#define NPF_PRIORITY(_name, _priority, ...)                             \
+	struct npf_rule _name = {                                       \
+		.result = NET_CONTINUE,                                 \
+		.priority = (_priority),                                \
+		.nb_tests = NUM_VA_ARGS_LESS_1(__VA_ARGS__) + 1,        \
+		.tests = {FOR_EACH(Z_NPF_TEST_ADDR, (,), __VA_ARGS__)}, \
```
pdgendt: Nit: one macro uses tabs, the other spaces, try to be consistent.

samples/net/pkt_filter/sample.yaml
```diff
@@ -1,14 +1,20 @@
...
-      - net_pkt_filter
-    depends_on: netif
-    integration_platforms:
-      - native_sim
+  sample.net.pkt_filter: {}
+  sample.net.pkt_filter_priorities:
+    extra_args:
+      - CONFIG_NET_SAMPLE_USE_PACKET_PRIORITIES=y
+      - CONFIG_NET_TC_RX_COUNT=2
+      - CONFIG_NET_TC_RX_SKIP_FOR_HIGH_PRIO=y
```
rlubos: I just wonder if there should be an overlay configuration with those, so users can easily apply required configs when building.

rlubos (dismissed):

samples/net/pkt_filter/sample.yaml
jukkar: Sure, why not. I will create a new version.

rlubos (approved):

pdgendt (approved):

PR #100800, opened 2025-12-10
drivers: clock_control: stm32{h5,u5}: fix PLL rate calculation with fractional PLL

tpambor:
When calculating the PLL output rate, take into account the fractional part of the PLL multiplier N. This ensures that the calculated output frequency is correct.


etienne-lms (approved):
LGTM.
Do you want to extend also to **clock_stm32_ll_h7.c**, to cover H7xx and H7RS SoCs series that embed the same PLL fractional mode (same 2¹³ ratio)?

erwango (approved):

PR #100802, opened 2025-12-10
hal_nordic: nrfs: select MULTITHREADING when NRFS_LOCAL_DOMAIN=y

rghaddab:
NRFS_LOCAL_DOMAIN needs to enable MULTITHREADING for events handling. select MULTITHREADING when this option is selected.

### DIFF
--- a/modules/hal_nordic/nrfs/Kconfig
+++ b/modules/hal_nordic/nrfs/Kconfig
@@ -67,6 +67,7 @@ config NRFS_LOCAL_DOMAIN
 	select IPC_SERVICE
 	select MBOX
 	select EVENTS
+	select MULTITHREADING
 	select REBOOT
 	help
 	  This option enables the nRF Services Local Domain libraries.


lstnl (approved):

modules/hal_nordic/nrfs/Kconfig
```diff
@@ -67,6 +67,7 @@ config NRFS_LOCAL_DOMAIN
 	select IPC_SERVICE
 	select MBOX
 	select EVENTS
+	select MULTITHREADING
```
nordicjm, line 70: `depends on`

nordicjm (changes requested):

modules/hal_nordic/nrfs/Kconfig
rghaddab: @nordicjm actually MULTITHREDING is enabled by default in zephyr and this commit will create some dependency loops.
I will close this PR and replace it with a cleaner solution in mcuboot

nordicjm: shouldn't be any dependency loops, IPC_SERVICE has no dependencies, MBOX has no dependencies, EVENTS has, as expected, multithreading as a *dependency*:
```
config EVENTS
        bool "Event objects"
        depends on MULTITHREADING
```
so there is no dependency loop here

PR #100809, opened 2025-12-10
drivers: ethernet: nxp: use net_if_carrier correct

maass-hamburg:
net_if_carrier is to be used independently of the
administrative state (start and stop of the ethernet_api).

### DIFF
--- a/drivers/ethernet/eth_nxp_s32_gmac.c
+++ b/drivers/ethernet/eth_nxp_s32_gmac.c
@@ -52,7 +52,6 @@ struct eth_nxp_s32_config {
 struct eth_nxp_s32_data {
 	struct net_if *iface;
 	uint8_t mac_addr[ETH_NXP_S32_MAC_ADDR_LEN];
-	uint8_t	if_suspended;
 	struct k_sem rx_sem;
 	struct k_sem tx_sem;
 	struct k_thread rx_thread;
@@ -117,13 +116,6 @@ static void phy_link_state_changed(const struct device *pdev,
 		Gmac_Ip_SetSpeed(cfg->instance, gmac_cfg.Speed);
 
 		cfg->base->MAC_CONFIGURATION |= GMAC_MAC_CONFIGURATION_DM(gmac_cfg.Duplex);
-
-		/* net iface should be down even if PHY link state is up
-		 * till the upper network layers have suspended the iface.
-		 */
-		if (ctx->if_suspended) {
-			return;
-		}
 
 		LOG_DBG("Link up");
 		net_eth_carrier_on(ctx->iface);
@@ -228,32 +220,11 @@ static int eth_nxp_s32_init(const struct device *dev)
 static int eth_nxp_s32_start(const struct device *dev)
 {
 	const struct eth_nxp_s32_config *cfg = dev->config;
-	struct eth_nxp_s32_data *ctx = dev->data;
-	struct phy_link_state state;
 
 	Gmac_Ip_EnableController(cfg->instance);
 
 	irq_enable(cfg->rx_irq);
 	irq_enable(cfg->tx_irq);
-
-	/* If upper layers enable the net iface then mark it as
-	 * not suspended so that PHY Link changes can have the impact
-	 */
-	ctx->if_suspended = false;
-
-	if (cfg->phy_dev) {
-		phy_get_link_state(cfg->phy_dev, &state);
-
-		/* Enable net_iface only when Ethernet PHY link is up or else
-		 * if net_iface is enabled when link is down and tx happens
-		 * in this state then the used tx buffers will never be recovered back.
-		 */
-		if (state.is_up == true) {
-			net_eth_carrier_on(ctx->iface);
-		}
-	} else {
-		net_eth_carrier_on(ctx->iface);
-	}
 
 	LOG_DBG("GMAC%d started", cfg->instance);
 
@@ -263,19 +234,11 @@ static int eth_nxp_s32_start(const struct device *dev)
 static int eth_nxp_s32_stop(const struct device *dev)
 {
 	const struct eth_nxp_s32_config *cfg = dev->config;
-	struct eth_nxp_s32_data *ctx = dev->data;
 	Gmac_Ip_StatusType status;
 	int err = 0;
 
 	irq_disable(cfg->rx_irq);
 	irq_disable(cfg->tx_irq);
-
-	/* If upper layers disable the net iface then mark it as suspended
-	 * in order to save it from the PHY link state changes
-	 */
-	ctx->if_suspended = true;
-
-	net_eth_carrier_off(ctx->iface);
 
 	status = Gmac_Ip_DisableController(cfg->instance);
 	if (status != GMAC_STATUS_SUCCESS) {
@@ -306,16 +269,13 @@ static void eth_nxp_s32_iface_init(struct net_if *iface)
 		ctx->mac_addr[0], ctx->mac_addr[1], ctx->mac_addr[2],
 		ctx->mac_addr[3], ctx->mac_addr[4], ctx->mac_addr[5]);
 
-	/* Make sure that the net iface state is not suspended unless
-	 * upper layers explicitly stop the iface
-	 */
-	ctx->if_suspended = false;
-
 	/* No PHY available, link is always up and MAC speed/duplex settings are fixed */
 	if (cfg->phy_dev == NULL) {
 		net_if_carrier_on(iface);
 		return;
 	}
+
+	net_if_carrier_off(iface);
 
 	/*
 	 * GMAC controls the PHY. If PHY is configured either as fixed


subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -1123,7 +1123,6 @@ static int cdc_ncm_iface_start(const struct device *dev)
 	LOG_DBG("Start interface %d", net_if_get_by_iface(data->iface));
 
 	atomic_set_bit(&data->state, CDC_NCM_IFACE_UP);
-	net_if_carrier_on(data->iface);
```
tmon-nordic, line 1129: > This way it is no longer wrong. To really make use of it, net_if_carrier should be changed at the points where a connection is established or disconnected. But that is for later.

Aren't these points simply `usbd_cdc_ncm_enable()` and `usbd_cdc_ncm_disable()`? If yes, why not just put the calls there instead of commenting "that is for later".

tmon-nordic: ~~Or is it more like `usbd_cdc_ncm_shutdown` and `.update` callback in `struct usbd_class_api` which are bound to functional state (device configured)?~~

maass-hamburg: I now changed it, so that it works with plugging the usb cable out and in. (tested on a renesas `ek_ra8d1` with the zperf sample and the provided confs and overlays)

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -877,6 +882,8 @@ static void usbd_cdc_ncm_suspended(struct usbd_class_data *const c_data)
 	struct cdc_ncm_eth_data *data = dev->data;
 
 	atomic_set_bit(&data->state, CDC_NCM_CLASS_SUSPENDED);
+
+	net_if_carrier_off(data->iface);
```
tmon-nordic: I would argue that suspend is not removing carrier. When suspended the device is still connected to host. It may even be possible for the device to issue Remote Wakeup if device reported that it supports Remote Wakeup and host did allow it (enabled the feature before suspending device).

maass-hamburg: But where are we getting the info, that for example the cable is no longer connected?

tmon-nordic: What cable do you mean? If "USB cable" then suspend is low power state when the cable is physically connected.

When USB cable is disconnected, all USB classes disable callbacks are called - here it is `usbd_cdc_ncm_disable()`. When the cable is connected, and device gets configured (so USB functions can operate) then all USB classes enable callbacks are called - here it is `usbd_cdc_ncm_enable()`.

manuargue:
ping @sumitbatra-nxp for the GMAC changes for S32

subsys/usb/device_next/class/usbd_cdc_ncm.c
maass-hamburg: > When USB cable is disconnected, all USB classes disable callbacks are called - here it is `usbd_cdc_ncm_disable()`. When the cable is connected, and device gets configured (so USB functions can operate) then all USB classes enable callbacks are called - here it is `usbd_cdc_ncm_enable()`.

No it isn't, at least for the `device_next` subsys. The disable callbacks (`usbd_class_disable`) are only called on `UDC_EVT_RESET` by `event_handler_bus_reset`.
`UDC_EVT_VBUS_REMOVED` is not used for that.

maass-hamburg: I removed the the `net_if_carrier_off()` in the suspend therefore.

sumitbatra-nxp:
@maass-hamburg thanks for the fix aligning carrier with link state.

Could you confirm you tested (or can reason about) these scenarios for eth_nxp_s32_gmac:

**TX while link is down**: if an application calls .send() while carrier/link is down, do we recover cleanly once link comes back?
Specifically: does Gmac_Ip_SendFrame() ever complete / do TX interrupts fire when link is down, or can TX descriptors/buffers get stuck consumed (leading to timeouts / permanent -ENOBUFS until reboot)?
If this is possible, we may want a small guard in send() to fail fast when carrier is off (e.g. return -ENETDOWN / -EHOSTDOWN) rather than queueing into HW.

**Initial carrier state with PHY:** the PR sets net_eth_carrier_off(iface) in eth_nxp_s32_iface_init() for PHY-present cases, which makes sense so iface doesn’t look “link up” until PHY reports it.

Please double-check the no-PHY / fixed-link case still sets carrier ON (e.g. if (cfg->phy_dev == NULL) net_if_carrier_on(iface);).

Also, does phy_link_callback_set() guarantee an immediate callback for the PHYs used on S32? If not, carrier could remain OFF until the first link transition/poll. If that’s a concern, we might need a one-time phy_get_link_state() after registering the callback to seed initial carrier state.

sumitbatra-nxp:
**Optional validation matrix** (what I had in mind while reviewing):

**For PHY-based Ethernet (S32 GMAC)**:

Boot with cable unplugged → carrier stays OFF

Plug cable → carrier ON, traffic works

Unplug cable → carrier OFF, traffic stops cleanly

Admin down/up with cable plugged → traffic stops/resumes without needing cable replug

TX attempted while link down → no TX deadlock or permanent ENOBUFS after link comes back

Rapid link flap → no stuck semaphores / TX mutex / IRQ storm

**For USB CDC-NCM:**

USB plug/unplug while running iperf → carrier tracks connection, no stale carrier ON after unplug

maass-hamburg:
> At this point this looks consistent with other Ethernet drivers. The only remaining thing I was curious about is TX behavior if .send() is called while carrier/link is down (e.g. whether TX resources always recover cleanly once the link comes back), but if this matches existing driver behavior then I’m fine with it.

Yes, i didn't changed these parts.

maass-hamburg:
> At this point this looks consistent with other Ethernet drivers. The only remaining thing I was curious about is TX behavior if .send() is called while carrier/link is down (e.g. whether TX resources always recover cleanly once the link comes back), but if this matches existing driver behavior then I’m fine with it.

No it isn't, see:

https://github.com/zephyrproject-rtos/zephyr/blob/1da7503904e63d0699b29b947d8d4c497ff3ce5a/subsys/net/ip/net_if.c#L247-L262

``NET_IF_LOWER_UP`` is the carrier. and it is checked by the networking subsystem before sending.

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -635,6 +635,7 @@ static void ncm_handle_notifications(const struct device *dev, const int err)
 	if (data->if_state == IF_STATE_CONNECTION_STATUS_SUBMITTED) {
 		data->if_state = IF_STATE_CONNECTION_STATUS_SENT;
 		LOG_INF("Connection status sent");
+		net_if_carrier_on(data->iface);
```
jfischer-no: Handling notifications (interrupt endpoint transfer is completed) is the consequence of  cdc_ncm_iface_start() scheduling notification work. Removing it from cdc_ncm_iface_start() and calling it here does not really change what happened from network interfaces view. What is the point of this commit?

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -840,6 +841,7 @@ static void usbd_cdc_ncm_update(struct usbd_class_data *const c_data,
 
 	if (data_iface == iface && alternate == 0) {
 		atomic_clear_bit(&data->state, CDC_NCM_DATA_IFACE_ENABLED);
+		net_if_carrier_off(data->iface);
```
jfischer-no: In the current implementation, the network interface on the Zephyr side determines whether the carrier is switched on or off for both sides. Mixing it like that is not correct.


>net_if_carrier is to be used independently of the
administrative state (start and stop of the ethernet_api).

>This seem to work.

Also it does belong in this commit and looks like a stray change.
Please work on your commit messages.

subsys/usb/device_next/class/usbd_cdc_ncm.c
```diff
@@ -868,6 +870,8 @@ static void usbd_cdc_ncm_disable(struct usbd_class_data *const c_data)
 	atomic_clear_bit(&data->state, CDC_NCM_DATA_IFACE_ENABLED);
 	atomic_clear_bit(&data->state, CDC_NCM_CLASS_SUSPENDED);
 
+	net_if_carrier_off(data->iface);
```
jfischer-no: Same as for changes in usbd_cdc_ncm_update().

jfischer-no (dismissed):

subsys/usb/device_next/class/usbd_cdc_ncm.c
maass-hamburg: cdc_ncm_iface_start() will be called by the networking subsystem only once. Meaning, if i disconnect the cable between the board and the host and connect it again, the networking subsystem won't be informed about the disconnection in between and won't f.e. restart the dhcp client. Doing it this way it will be notified. Also this will only be called when there is a connection, if the board is started without the usb cable plugged in, the iface will only go up when it is plugged in.

manuargue (dismissed):
Thanks @sumitbatra-nxp for your feedback on the NXP GMAC driver. Judging by the discussion, the changes LGTM for NXP GMAC driver.

subsys/usb/device_next/class/usbd_cdc_ncm.c
maass-hamburg: `iface_start` and `iface_stop` should never directly contain `net_if_carrier_on` and `net_if_carrier_off`. The carrier is independent of it.

manuargue (dismissed):
approved based on the comments above, I don't have hw to test this.

manuargue (approved):

yangbolu1991 (approved):
I don't work on s32 platform.
But reviewed the changes, It looks good to me.
Thanks.

PR #100814, opened 2025-12-10
drivers: input: input_chsc5x: add delay after reset

allemanm:
The sensor requires at least 94ms of delay before starting communication after a reset. This is not mentioned in the datasheet but is arbitrarily defined.

### DIFF
--- a/drivers/input/input_chsc5x.c
+++ b/drivers/input/input_chsc5x.c
@@ -252,6 +252,9 @@ static int chsc5x_init(const struct device *dev)
 		return ret;
 	}
 
+	/* It takes about 94ms until chip is ready after reset. */
+	k_msleep(100);
+
 	return chsc5x_chip_init(dev);
 };
 
--- a/drivers/input/input_chsc5x.c
+++ b/drivers/input/input_chsc5x.c
@@ -100,7 +100,8 @@ static void chsc5x_isr_handler(const struct device *dev, struct gpio_callback *c
 	k_work_submit(&data->work);
 }
 
-static int chsc5x_chip_init(const struct device *dev)
+#if defined(CONFIG_INPUT_CHSC5X_VERIFY_IC_TYPE)
+static int chsc5x_verify_ic(const struct device *dev)
 {
 	const struct chsc5x_config *cfg = dev->config;
 	int ret;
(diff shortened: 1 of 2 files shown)


drivers/input/input_chsc5x.c
```diff
@@ -252,10 +254,17 @@ static int chsc5x_init(const struct device *dev)
...
+#if defined(CONFIG_INPUT_CHSC5X_VERIFY_IC_TYPE)
 	/* It takes about 94ms until chip is ready after reset. */
 	k_msleep(100);
 
-	return chsc5x_chip_init(dev);
+	ret =  chsc5x_chip_init(dev);
+	if (ret < 0) {
+		LOG_ERR("Failed to initialize chip: %d", ret);
+		return ret;
+	}
```
fabiobaltieri: doesn't sound like a great idea, one may skip this, then in the future something get added to `chsc5x_chip_init` and it'll be sliently skipped, just leave it in all the time

pdgendt: Would it help to rename the function to `chsc5x_verify_ic`? Needing to wait 100 ms on some platform is a long time. While if you have a custom board, you probably know already what IC is on there?

fabiobaltieri: Sure that'd work but is there an actual use case? There's plenty of drivers with similar behavior and this is probably something with a heavy graphical framework on the back, I'd be surprised if those 100ms make any difference. It may as well be otherwise why would have @allemanm bothered implementing it but I'd still want that confirmed in case it was thought as a hypothetical "nice to have".

allemanm: I indeed made the IC type verification optional to speed up startup behavior, in case an application needs to perform tasks quickly after boot.

fabiobaltieri: ok, unless you actually need it I suggest removing it, I understand the "just in case" sentiment but that's more options and more tech debt and more chances of future bug and regressions or race conditions or whatever for something no one may need, 100ms reset delay is fine plenty of drivers have it

pdgendt: We do need it, and keeping a patch downstream is less convenient than having it upstream. Or do you suggest removing the IC type altogether?

fabiobaltieri: Ok if you are using it downstream then let's go with it, I got the opposite impression.

fabiobaltieri (dismissed):

drivers/input/input_chsc5x.c
fabiobaltieri: I'd rename the function maybe, up to you really not entirely convinced myself

pdgendt (approved):

fabiobaltieri (approved):

PR #100815, opened 2025-12-10
boards: arduino: opta: enable SDHC SDIO support

pennam:
This commit enable SDHC SDIO support for OPTA


iabdalkader (approved):
Compared to Giga and double checked the pins.

facchinm (approved):

pillo79 (approved):
LGTM, was looking to introduce variants but will do it in a follow-up refactor.

PR #100816, opened 2025-12-10
samples: code_relocation_nocopy: Allow flash to not start at address 0

asmellby:
The fallback when a XIP external flash doesn't exist assumed that main Flash starts at address 0. Adjust placement of the EXTFLASH region to take the Flash base address and load offset into account.

Also increase the offset of the fake EXTFLASH region to support larger apps.

### DIFF
--- a/samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
+++ b/samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
@@ -66,10 +66,13 @@
 
 /*
  * Add another fake portion of FLASH to simulate a secondary or external FLASH
- * that we can do XIP from.
+ * that we can do XIP from. The simulated EXTFLASH starts at an offset of 0x9000
+ * from the start of the main application FLASH area, taking the flash base address
+ * and load offset into account, and extends to the end of the physical FLASH.
  */
-#define EXTFLASH_ADDR	0x7000
-#define EXTFLASH_SIZE	(CONFIG_FLASH_SIZE * 1K - EXTFLASH_ADDR)
+#define EXTFLASH_OFFSET	0x9000
+#define EXTFLASH_ADDR	(CONFIG_FLASH_BASE_ADDRESS + CONFIG_FLASH_LOAD_OFFSET + EXTFLASH_OFFSET)
+#define EXTFLASH_SIZE	(CONFIG_FLASH_SIZE * 1K - CONFIG_FLASH_LOAD_OFFSET - EXTFLASH_OFFSET)
 
 #endif
 


jerome-pouiller (dismissed):

samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
```diff
@@ -68,8 +68,9 @@
  * Add another fake portion of FLASH to simulate a secondary or external FLASH
  * that we can do XIP from.
  */
-#define EXTFLASH_ADDR	0x7000
-#define EXTFLASH_SIZE	(CONFIG_FLASH_SIZE * 1K - EXTFLASH_ADDR)
+#define EXTFLASH_OFFSET	0x9000
```
tejlmand, line 73: why 0x9000 instead of 0x7000 ?

I know it's the offset which is added, but if flash base and load offset are both 0, then this commit now has EXTFLASH_ADDR of 0x9000 instead of 0x7000.

If I understand the original code correctly, then we try to use a portion of the existing flash as emulating an external flash. Thus any increase to EXTFLASH_ADDR has a risk of sample crossing the real size of the real flash.

So I would like to know why this value was chosen ?

samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
```diff
@@ -68,8 +68,9 @@
  * Add another fake portion of FLASH to simulate a secondary or external FLASH
  * that we can do XIP from.
```
tejlmand: can you update this comment, describing how / why we are doing this calculation ?

samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
asmellby: The number needs to be high enough to steer clear of the application itself. If I compile this sample for e.g. `nrf52dk/nrf52832`, it fails with the current config because the app itself is actually larger than 0x7000 (28 kB):

```
/work/zephyr/zephyr-sdk-0.17.4/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/12.2.0/../../../../arm-zephyr-eabi/bin/ld.bfd: section .extflash_text_reloc LMA [0000000000007000,000000000000702f] overlaps section rodata LMA [00000000000068f0,00000000000072df]
collect2: error: ld returned 1 exit status
```

I chose to bump it up by 8 kB to make it more likely to work on "most" targets. While this does make it no longer work on targets with <32 kB flash, there shouldn't be many of them.

samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
asmellby: Done!

samples/application_development/code_relocation_nocopy/linker_arm_nocopy.ld
nordicjm: why not delete partitions then add 2 partitions with a chosen node so that it cannot overflow, then you get the XIP region from the second partition without having to hard code anything ?

nordicjm (changes requested):
See https://github.com/zephyrproject-rtos/zephyr/pull/100816/files#r2635169329

PR #100818, opened 2025-12-10
Enable nxp analog comparator (acomp) driver

ZhaoxiangJin:
enable nxp acomp driver
enable acomp for rw612
test with gpio_loopback
test result:

<details>
<summary>Logs</summary>

```
*** Booting Zephyr OS build v4.3.0-2167-gc8e9e81fa49c ***
Running TESTSUITE comparator_gpio_loopback
START - test_get_output
PASS - test_get_output in 0.005 seconds
START - test_no_trigger_no_pending
PASS - test_no_trigger_no_pending in 0.003 seconds
START - test_trigger_both_edges_pending
E: Trigger type not support.
SKIP - test_trigger_both_edges_pending in 0.003 seconds
START - test_trigger_callback
PASS - test_trigger_callback in 0.004 seconds
START - test_trigger_falling_edge_pending
PASS - test_trigger_falling_edge_pending in 0.004 seconds
START - test_trigger_rising_edge_pending
PASS - test_trigger_rising_edge_pending in 0.004 seconds
TESTSUITE comparator_gpio_loopback succeeded
 
------ TESTSUITE SUMMARY START ------
 
SUITE PASS - 100.00% [comparator_gpio_loopback]: pass = 5, fail = 0, skip = 1, total = 6 duration = 0.023 seconds
- PASS - [comparator_gpio_loopback.test_get_output] duration = 0.005 seconds
- PASS - [comparator_gpio_loopback.test_no_trigger_no_pending] duration = 0.003 seconds
- SKIP - [comparator_gpio_loopback.test_trigger_both_edges_pending] duration = 0.003 seconds
- PASS - [comparator_gpio_loopback.test_trigger_callback] duration = 0.004 seconds
- PASS - [comparator_gpio_loopback.test_trigger_falling_edge_pending] duration = 0.004 seconds
- PASS - [comparator_gpio_loopback.test_trigger_rising_edge_pending] duration = 0.004 seconds
 
------ TESTSUITE SUMMARY END ------
 
PROJECT EXECUTION SUCCESSFUL
```

</details>


dts/arm/nxp/nxp_rw6xx_common.dtsi
```diff
@@ -604,6 +604,19 @@
...
+			compatible = "nxp,acomp";
+			reg = <0x38400 0x4c>;
+			interrupts = <110 0>;
+			status = "disabled";
+			positive-input = "ch0";
+			negative-input = "ch0";
+			warmup-time-us = <1>;
+			response-mode = "slow";
+			positive-hysteresis-mv = <0>;
+			negative-hysteresis-mv = <0>;
```
bjarki-andreasen: No need to define defaults for a disabled component, user is expected to configure these anyway to match their usecase . Just
```
		acomp: acomp@38400 {
			compatible = "nxp,acomp";
			reg = <0x38400 0x4c>;
			interrupts = <110 0>;
			status = "disabled";
		};
```
will do

dts/bindings/comparator/nxp,acomp.yaml
```diff
@@ -0,0 +1,100 @@
...
+  - pinctrl-device.yaml
+
+properties:
+  reg:
+    required: true
+
+  interrupts:
+    required: true
+
+  positive-input:
```
bjarki-andreasen, line 20: this and many others (all of the properties it seems) are required, so set `required: true` for them. This produces way nicer build errors than the macro madness that occurs when  `DT_ENUM_IDX(DT_DRV_INST(inst), positive_input)` errors out if `positive-input` was not defined in the devicetree

bjarki-andreasen (changes requested):
Looks good, two comments

dts/bindings/comparator/nxp,acomp.yaml
ZhaoxiangJin: positive and negative channel are required, others are optional (depends on application).
Added `required: true` for positive and negative channel.

dts/arm/nxp/nxp_rw6xx_common.dtsi
ZhaoxiangJin: yes, removed these properties from the common dtsi

dts/arm/nxp/nxp_rw6xx_common.dtsi
```diff
@@ -604,6 +604,15 @@
...
 			power-domains = <&peripheral_domain>;
 		};
+
+		acomp: acomp@38400 {
+			compatible = "nxp,acomp";
+			reg = <0x38400 0x4c>;
+			interrupts = <110 0>;
+			status = "disabled";
+			positive-input = "ch0";
+			negative-input = "ch0";
```
bjarki-andreasen: even required properties are only required if status is "okay"

bjarki-andreasen (dismissed):
one nit, otherwise LGTM

dts/arm/nxp/nxp_rw6xx_common.dtsi
ZhaoxiangJin: We will enable this device node in [board.dtsi](https://github.com/zephyrproject-rtos/zephyr/pull/100818/changes#diff-9df863dd3517c1ba60206231142641c1c9cbb285a8a851e5f905fb9e4982f93bR358), so if we don't set these `required:true` properties in [soc.dtsi](https://github.com/zephyrproject-rtos/zephyr/pull/100818/changes), but instead set them in `app.overlay`, then when compiling other cases for this board (non-acomp apps), there will be a devicetree parsing error, indicating that the `required:true` property has not been set."

bjarki-andreasen: a device should normally only be enabled if it is actually used, enabling it but having not assigned pins for it is pretty contradictory, you would want a user to get an error after they enable the comparator, since they intend to use it, rather than having it set to some arbitrary, and invalid default value. (negative and positive inputs can't both be ch0)

ZhaoxiangJin: The acomp pins on this platform don't require additional configuration and the ch0 for positive and negative terminals are independent. However, the current design isn't truly plug-and-play - at minimum, users still need to consult the manual themselves to determine the wiring. So perhaps it would be most appropriate to enable it in the specific app.overlay instead. The PR has already been updated.

bjarki-andreasen (approved):

Holt-Sun (approved):

PR #100820, opened 2025-12-10
boards: silabs: Enable additional peripherals

asmellby:
Configure PWM, ADC and DAC on boards that support it, and declare support for pwm, adc, dac and watchdog testing in board yaml files.

### DIFF
--- a/boards/silabs/dev_kits/xg24_dk2601b/xg24_dk2601b.dts
+++ b/boards/silabs/dev_kits/xg24_dk2601b/xg24_dk2601b.dts
@@ -32,6 +32,9 @@
 		pwm-led0 = &red_pwm_led;
 		pwm-led1 = &green_pwm_led;
 		pwm-led2 = &blue_pwm_led;
+		red-pwm-led = &red_pwm_led;
+		green-pwm-led = &green_pwm_led;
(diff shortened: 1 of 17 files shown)


jerome-pouiller (approved):

boards/silabs/dev_kits/xg24_dk2601b/xg24_dk2601b.dts
```diff
@@ -32,6 +32,9 @@
 		pwm-led0 = &red_pwm_led;
 		pwm-led1 = &green_pwm_led;
 		pwm-led2 = &blue_pwm_led;
```
jerome-pouiller, line 34: In fact, `pwm-led1` and `pwm-led2` are never referenced.

asmellby: They are used in the `fade_led` sample, but the reference is constructed through a macro so it's not searchable: https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/basic/fade_led/src/main.c#L17

jerome-pouiller: good catch.

jhedberg (approved):

PR #100823, opened 2025-12-10
boards: cy8cproto_041tp: i2c support

manoj-aerlync:
I2C Driver for cy8cproto_041tp.This PR has dependency on https://github.com/zephyrproject-rtos/zephyr/pull/100894

### DIFF
--- a/drivers/i2c/i2c_infineon_pdl.c
+++ b/drivers/i2c/i2c_infineon_pdl.c
@@ -163,6 +163,10 @@ static void ifx_master_event_handler(void *callback_arg, uint32_t event)
 		k_sem_give(&data->transfer_sem);
 	}
 
+	if (data->p_target_config == NULL) {
+		return;
+	}
+
 	if (0 != (CY_SCB_I2C_SLAVE_READ_EVENT & event)) {
 		if (data->p_target_config->callbacks->read_requested) {
 			data->p_target_config->callbacks->read_requested(data->p_target_config,
@@ -318,6 +322,7 @@ static int ifx_cat1_i2c_configure(const struct device *dev, uint32_t dev_config)
 	const struct ifx_cat1_i2c_config *config = dev->config;
 	cy_en_scb_i2c_status_t rslt;
 	int ret;
+	bool is_target_mode = false;
 
 	if (dev_config != 0) {
 		switch (I2C_SPEED_GET(dev_config)) {
@@ -343,9 +348,13 @@ static int ifx_cat1_i2c_configure(const struct device *dev, uint32_t dev_config)
 
 		if (dev_config & I2C_MODE_CONTROLLER) {
 			_i2c_default_config.i2cMode = CY_SCB_I2C_MASTER;
+			is_target_mode = false;
 		} else {
 			_i2c_default_config.i2cMode = CY_SCB_I2C_SLAVE;
+			is_target_mode = true;
 		}
+	} else {
+		is_target_mode = (_i2c_default_config.i2cMode == CY_SCB_I2C_SLAVE);
 	}
 
 	/* Acquire semaphore (block I2C operation for another thread) */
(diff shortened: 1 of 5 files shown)


boards/infineon/cy8cproto_041tp/cy8cproto_041tp.dts
```diff
@@ -21,27 +21,3 @@
 		zephyr,shell-uart = &uart0;
 	};
 };
-
-uart0: &scb0 {
```
sreeramIfx: this change does not seem to relevant to this PR and is probably causing the CI failures

sreeramIfx (commented):
Looks good to me mostly. Compliance checks need to be fixed

JarmouniA:
> This PR has dependency on #100894

Then it should stay as a draft until the dependency is merged and it's rebased on it.

dts/arm/infineon/psoc4/psoc4100tp/psoc4100tp.dtsi
```diff
@@ -23,18 +22,18 @@
 
 	soc {
 		pinctrl: pinctrl@40020000 {
-			compatible = "infineon,cat1-pinctrl";
```
sreeramIfx: please move the compatible renaming into a separate commit with appropriate description

drivers/i2c/i2c_infineon_pdl.c
```diff
@@ -318,6 +325,7 @@ static int ifx_cat1_i2c_configure(const struct device *dev, uint32_t dev_config)
 	const struct ifx_cat1_i2c_config *config = dev->config;
 	cy_en_scb_i2c_status_t rslt;
 	int ret;
+	bool is_slave_mode = false;
```
teburd:
```suggestion
	bool is_target_mode = false;
```

dts/arm/infineon/psoc4/psoc4100tp/psoc4100tp.dtsi
manoj-aerlync: These changes are done in the Clock and GPIO Pull Request.

boards/infineon/cy8cproto_041tp/cy8cproto_041tp.dts
manoj-aerlync: Done

drivers/i2c/i2c_infineon_pdl.c
```diff
@@ -517,7 +541,7 @@ static int ifx_cat1_i2c_transfer(const struct device *dev, struct i2c_msg *msg,
 		}
 
 		/* Acquire semaphore (block I2C async transfer for another thread) */
-		ret = k_sem_take(&data->transfer_sem, K_FOREVER);
+		ret = k_sem_take(&data->transfer_sem, I2C_TRANSFER_TIMEOUT_MSEC);
```
teburd: Adding a timeout here has some implications that aren't being tested/dealt with. Leave the K_FOREVER and remove the ret check (k_sem_take is infallible if K_FOREVER).

samples/sensor/bme280/boards/cy8cproto_041tp.overlay
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) 2025 Infineon Technologies AG,
+ * or an affiliate of Infineon Technologies AG.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+i2c1: &scb1 {
```
teburd, line 8: I appreciate that we are adding the overlays for sensors. For i2c controller testing the i2c_ram test is the preferred test to use. Non-blocking, and something for the future if you submit more i2c drivers which I hope you do.

teburd (changes requested):
Some minor changes I'd like to see, should be quick to do.

drivers/i2c/i2c_infineon_pdl.c
manoj-aerlync: If the timeout is K_FOREVER, the code will block indefinitely on the semaphore when the I²C sensor is missing.As a result, the I²C shell becomes unresponsive because it’s stuck waiting and cannot accept further commands.It also never returns to main, and important error logs cannot be seen.

teburd: I’m aware of what happens. I do not see anything here to recover. If there’s no recovery it’s better to fail hard and fail fast.

teburd (approved):

teburd (approved):

sreeramIfx (approved):

nordic-krch (approved):

PR #100836, opened 2025-12-10
sys: reboot and poweroff: disable ZLIs

bjarki-andreasen:
During `sys_reboot`, we discovered deadlocks as ZLIs where triggered while we where actively powering down the SoC.

When powering down or rebooting the SoC, neither ZLIs nor regular IRQs must interrupt the procedure, as we are actively powering down parts of the SoC, which quickly leads to undefined behavior if a ZLI tries to access some hardware which was just powered off or reset.

For regular IRQs, we have `irq_lock()`, we have nothing for disabling ZLIs, as by design, the zephyr kernel is not supposed to be able to interfere with them. However, the poweroff usecase presents a clear scenario where we must be able to disable them.

This PR adds an arch level API for disabling normal IRQs and ZLIs, `arch_zli_lock()` and complimenting `arch_zli_unlock()`. The API is clearly documented with warnings, expected usecase, and is not added exposed in `irq.h` to discourage "common" use of the API, its meant for arch level operations.

The sys_poweroff() and sys_reboot() functions have been extended to use the new API to ensure ZLIs are masked, which should ensure a safe, atomic poweroff/reboot procedure.

### DIFF
--- a/include/zephyr/arch/arm/asm_inline_gcc.h
+++ b/include/zephyr/arch/arm/asm_inline_gcc.h
@@ -105,6 +105,31 @@ static ALWAYS_INLINE bool arch_irq_unlocked(unsigned int key)
 	return key == 0U;
 }
 
+#ifdef CONFIG_ZERO_LATENCY_IRQS
+
+static ALWAYS_INLINE unsigned int arch_zli_lock(void)
+{
+	unsigned int key;
+
+	key = __get_PRIMASK();
+
+	/*
+	 * The cpsid instruction is self synchronizing within the instruction stream, no need for
+	 * an explicit __ISB().
+	 */
+	__disable_irq();
+
+	return key;
+}
+
+static ALWAYS_INLINE void arch_zli_unlock(unsigned int key)
+{
+	__set_PRIMASK(key);
+	__ISB();
+}
+
+#endif /* CONFIG_ZERO_LATENCY_IRQS */
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/os/reboot.c
+++ b/lib/os/reboot.c
@@ -23,6 +23,10 @@ FUNC_NORETURN void sys_reboot(int type)
 
 	(void)irq_lock();
 
+#if defined(CONFIG_ZERO_LATENCY_IRQS)
+	(void)arch_zli_lock();
+#endif /* CONFIG_ZERO_LATENCY_IRQS */
+
 	/* Disable caches to ensure all data is flushed */
 #if defined(CONFIG_ARCH_CACHE)
 #if defined(CONFIG_DCACHE)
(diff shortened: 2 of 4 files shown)


include/zephyr/arch/arm/asm_inline_gcc.h
```diff
@@ -105,6 +105,25 @@ static ALWAYS_INLINE bool arch_irq_unlocked(unsigned int key)
...
 }
 
+#ifdef CONFIG_ZERO_LATENCY_IRQS
+
+static ALWAYS_INLINE unsigned int arch_zli_lock(void)
+{
+	unsigned int key;
+
+	key = __get_PRIMASK();
+	__disable_irq();
```
peter-mitsis: Does this need a call to __ISRB()? I ask because not only do I not know, but I observe that the arch_irq_lock() implementation for ARMV7_M_ARMV8_M_MAINLINE has it and ZERO_LATENCY_IRQS depends on CPU_CORTEX_M_HAS_BASEPRI which in turn depends on ARMV7_M_ARMV8_M_MAINLINE.

peter-mitsis:
I have a question that applies to the implementations of arch_zli_lock() and arch_zli_unlock(). Do these functions need to call __ISRB() ? I ask because I observe that the ARMV7_M_ARMV8_M_MAINLINE implementations for both arch_irq_lock() and arch_irq_unlock() have it (and ZERO_LATENCY_IRQS depends on CPU_CORTEX_M_HAS_BASEPRI which in turn depends on ARMV7_M_ARMV8_M_MAINLINE).

lib/os/reboot.c
```diff
@@ -23,6 +23,10 @@ FUNC_NORETURN void sys_reboot(int type)
 
 	(void)irq_lock();
 
+#if defined(CONFIG_ZERO_LATENCY_IRQS)
+	(void)arch_zli_lock();
```
ZhaoxiangJin, line 27: For ARMV6-M and ARMV8-M BASELINE, `irq_lock` already masked interrupts except NMI and HardFault. Do we still need to invoke `arch_zli_lock`?

bjarki-andreasen: These don't support `CONFIG_ZERO_LATENCY_IRQS` do they? `arch_zli_lock` should not be invoked I believe

carlescufi: They don't, see https://github.com/zephyrproject-rtos/zephyr/blob/0b52928625b06b6b8de9b56524fbc1ae5f8cbd6a/arch/arm/core/cortex_m/Kconfig#L329
and
https://github.com/zephyrproject-rtos/zephyr/blob/0b52928625b06b6b8de9b56524fbc1ae5f8cbd6a/arch/arm/core/cortex_m/Kconfig#L131-L133

ZhaoxiangJin: Ah, I missed this. Peter already mentioned it above, but I didn't notice.

peter-mitsis (approved):

nashif (approved):

PR #50690, opened 2022-09-27
add spi nand driver

rogeryou:
This PR is updated for #43915.

Add SPI NAND block device driver for using SPI NAND Flash like Macronix Flash MX31LF4GE4BC.
This driver is tested on STM32L562E-DK.

The license of BCH code is updated.

PR #100845, opened 2025-12-11
flash: spi_nand: initial generic driver

JordanYates:
Initial generic driver for SPI-NAND devices that expose the ONFI parameter page. Does not support advanced/optional features such as continuous reads, software ECC, or power down modes.

Configuration comes from devicetree in order to support other modules, but is validated against the loaded ONFI data at boot.

Can provide a base towards future work on #99908.

PR #50690 was used as inspiration and for some logic (copyrights preserved), but this PR attempts to sidestep the significant review challenges around the software ECC support by simply providing a baseline driver that uses hardware ECC.

Implementing the driver for the Micron flash devices also exposed non-generic register defines etc used in the previous PR. Everything in this PR should be generic across Micron (tested in hardware) and Macronix (datasheets and the other PR).

Tested on Micron MT29F4G01ABBFDWB. Example debug logs from a (erase-write-read-erase functionality test)
```
[00:00:00.008,480] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 32 us (Status 00)
[00:00:00.008,832] <dbg> spi_nand: onfi_parameters_load:      Manufacturer: MICRON
[00:00:00.008,832] <dbg> spi_nand: onfi_parameters_load:             Model: MT29F4G01ABBFDWB
[00:00:00.008,864] <dbg> spi_nand: onfi_parameters_load:  Page Size (data): 4096
[00:00:00.008,864] <dbg> spi_nand: onfi_parameters_load: Page Size (spare): 256
[00:00:00.008,864] <dbg> spi_nand: onfi_parameters_load:   Pages per Block: 64
[00:00:00.008,864] <dbg> spi_nand: onfi_parameters_load:   Blocks per Unit: 2048
[00:00:00.008,896] <dbg> spi_nand: onfi_parameters_load:             Units: 1
[00:00:00.049,056] <dbg> spi_nand: spi_nand_erase: Erasing block starting at 011380
[00:00:00.049,920] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 768 us (Status 00)
[00:00:00.051,840] <dbg> spi_nand: spi_nand_write: Write 4096 to 01139d:000
[00:00:00.052,160] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 256 us (Status 00)
[00:00:00.052,192] <dbg> spi_nand: spi_nand_read: Read 4096 from 01139d:000
[00:00:00.052,416] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 128 us (Status 00)
[00:00:00.053,952] <dbg> spi_nand: spi_nand_read: Read 51 from 01139d:031
[00:00:00.054,144] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 128 us (Status 00)
[00:00:00.054,336] <dbg> spi_nand: spi_nand_erase: Erasing block starting at 011380
[00:00:00.055,008] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 640 us (Status 00)
[00:00:00.055,072] <dbg> spi_nand: spi_nand_read: Read 4096 from 01139d:000
[00:00:00.055,232] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 128 us (Status 00)
```


LinderPi:
You could already check if the implemented NFTL disk driver module works for you (#100858).

tpambor:
I also gave this a try with a Winbond W25N01GVZEIG
```
[00:00:00.000,000] <dbg> spi_nand: spi_nand_wait_until_ready: Ready after 0 us (Status 00)
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:      Manufacturer: WINBOND
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:             Model: W25N01GV
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:  Page Size (data): 2048
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load: Page Size (spare): 64
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:   Pages per Block: 64
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:   Blocks per Unit: 1024
[00:00:00.001,000] <dbg> spi_nand: onfi_parameters_load:             Units: 1
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+static int spi_nand_read(const struct device *dev, off_t addr, void *dest, size_t size)
+{
+	const struct spi_nand_config *config = dev->config;
+	uint8_t *dest_u8 = dest;
+	uint32_t page_address;
+	uint16_t page_offset;
+	uint16_t bytes_to_end;
+	uint16_t bytes_to_read;
+	int ret = 0;
+
```
de-nordic, line 369: No work when size == 0.

carlescufi:
> > @JordanYates how does this fit #100683? Are we just going to have the same as NOR, where we have an SPI NOR driver and then flash-specific ones?
> 
> This driver issues the SET_FEATURE/GET_FEATURE command to addresses 0xa0, 0xb0, 0xc0. https://github.com/Embeint/zephyr/blob/97f4c915ad2da5ee63638a7fe377a8a7de185d06/drivers/flash/spi_nand.h#L46-L53
> 
> These are in the vendor specific range (0x80-0xff) according to ONFI spec. Most SPI NANDs have similar content there, e.g. ECC enable bit is the same on Micron MT29F4G01ABBFDWB and Winbond W25N01GVZEIG. But there are also differences, I haven't looked at everything but Micron MT29F4G01ABBFDWB has 3 bits of ECC status in the status register, but Winbond W25N01GVZEIG has only 2 bits of ECC status. Currently, the driver in this PR is not checking the ECC status bits.

Thanks @tpambor for the feedback, but what I meant is how will we organize the code for NAND flashes in general. In the case of NOR, we have:

- Individual drivers for different flashes
- `spi_nor.c` which is a generic flash driver over the SPI API
- `flash_mspi_nor.c` which is a generic flash driver over the MSPI API

My question was whether we will have something similar here, in the lines of:

- NAND drivers that target "raw" NAND and require a translation layer (Dhara)
- An `spi_nand.c` as in this PR that handles NAND flashes that do not require an FTL because they do bad block allocation and wear levelling internally, and that are accessible via the SPI API
- A future `flash_mspi_nand.c` that does the same as `spi_nand.c` but over MSPI

CC @JordanYates @anangl @swift-tk

drivers/flash/spi_nand.h
```diff
@@ -0,0 +1,208 @@
...
+	uint16_t vendor_revision;
+	/** Vendor specific information  */
+	uint8_t vendor_info[88];
+
+	/*
+	 * Validity block
+	 */
+
+	uint16_t integrity_crc;
+} __packed;
```
tpambor, line 228: The parameter page is not specific to SPI NAND chips. I think we can share this with parallel NAND flashes a move this structure to a `onfi_nand.h` header.

drivers/flash/spi_nand.h
```diff
@@ -0,0 +1,208 @@
...
+	SPI_NAND_FEATURE_STATUS_OIP = BIT(0),
+	/** Write enable latch */
+	SPI_NAND_FEATURE_STATUS_WEL = BIT(1),
+	/** Block erase operation failed */
+	SPI_NAND_FEATURE_STATUS_ERASE_FAIL = BIT(2),
+	/** Page program operation failed */
+	SPI_NAND_FEATURE_STATUS_PROGRAM_FAIL = BIT(3),
+	/** Page program operation failed */
+	SPI_NAND_FEATURE_STATUS_CACHE_BUSY = BIT(7),
+};
```
tpambor, line 95:
```suggestion
/* Micron:
 * BIT6 BIT5 BIT4  State
 *    0    0    0  No errors
 *    0    0    1  Errors detected and corrected
 *    0    1    0  Errors detected but not corrected
 *    0    1    1  Errors detected and corrected. Refresh recommended
 *    1    0    1  Errors detected and corrected. Refresh necessary
 * Macronix:
 * BIT5 BIT4  State
 *    0    0  No errors
 *    0    1  Errors detected and corrected
 *    1    0  Errors detected but not corrected
 *    1    1  Errors detected and corrected. Refresh recommended
 */

enum spi_nand_feature_status {
	/** Operation in progress */
	SPI_NAND_FEATURE_STATUS_OIP = BIT(0),
	/** Write enable latch */
	SPI_NAND_FEATURE_STATUS_WEL = BIT(1),
	/** Block erase operation failed */
	SPI_NAND_FEATURE_STATUS_ERASE_FAIL = BIT(2),
	/** Page program operation failed */
	SPI_NAND_FEATURE_STATUS_PROGRAM_FAIL = BIT(3),
	/** No ECC errors */
	SPI_NAND_FEATURE_STATUS_ECC_NO_ERROR = 0,
	/** Uncorrectable ECC error */
	SPI_NAND_FEATURE_STATUS_ECC_UNCORRECTABLE = BIT(5),
	/** Mask for ECC status bits */
	SPI_NAND_FEATURE_STATUS_ECC_MASK = BIT(5) | BIT(4),
	/** Page program operation failed */
	SPI_NAND_FEATURE_STATUS_CACHE_BUSY = BIT(7),
};
```

ECC status is sufficiently similar across vendors to be supported here.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+	/* Trigger the read to cache */
+	ret = spi_nand_access(dev, SPI_NAND_CMD_PAGE_READ,
+			      NAND_ACCESS_ADDRESSED | NAND_ACCESS_24BIT_ADDR, page, NULL, 0);
+	if (ret != 0) {
+		return ret;
+	}
+
+	/* Wait until the read to cache completes (poll with no delays) */
+	return spi_nand_wait_until_ready(dev, config->page_read_us, 0, &status);
```
tpambor:
```suggestion
	ret = spi_nand_wait_until_ready(dev, config->page_read_us, 0, &status);
	if (ret != 0) {
		return ret;
	}

	if ((status & SPI_NAND_FEATURE_STATUS_ECC_MASK) ==
	    SPI_NAND_FEATURE_STATUS_ECC_UNCORRECTABLE) {
		LOG_ERR("Uncorrectable ECC error during page read");
		return -EIO;
	} else if ((status & SPI_NAND_FEATURE_STATUS_ECC_MASK) !=
		   SPI_NAND_FEATURE_STATUS_ECC_NO_ERROR) {
		LOG_DBG("Correctable ECC error during page read");
	}

	return 0;
```

Return an error if there is an uncorrectable bit.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+};
+
+/* Indicates that an access command includes bytes for the address.
+ * If not provided the opcode is not followed by address bytes.
+ */
+#define NAND_ACCESS_ADDRESSED BIT(0)
+
+/* Indicates that addressed access uses a 8-bit address regardless of
+ * spi_nand_data::flag_8bit_addr.
+ */
```
tpambor:
```suggestion
/* Indicates that addressed access uses a 8-bit address. */
```
`flag_8bit_addr` does not exist in `spi_nand_data`. Also applies to below 16bit etc.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+	uint32_t page_address;
+	uint8_t status;
+	int ret = 0;
+
+	/* Write area must be subregion of device */
+	if (!valid_region(dev, addr, size)) {
+		return -EINVAL;
+	}
+
+	/* All erases must be block aligned in both start address and size */
```
tpambor:
```suggestion
	/* All writes must be page aligned in both start address and size */
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+			break;
+		}
+
+		/* Wait for the write to complete (poll every 0.1ms) */
+		ret = spi_nand_wait_until_ready(dev, config->page_program_us, 100, &status);
+		if (ret != 0) {
+			break;
+		}
+		if (status & SPI_NAND_FEATURE_STATUS_PROGRAM_FAIL) {
+			LOG_ERR("Erase operation failed");
```
tpambor:
```suggestion
			LOG_ERR("Program operation failed");
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+	 */
+	computed_crc = 0;
+	onfi.integrity_crc = 1;
+
+	/* Scan through the loaded info until we find a valid CRC.
+	 * Use the assumed page size from devicetree.
+	 */
+	page_size = config->parameters->write_block_size;
+	for (int i = 0; i < page_size; i += sizeof(onfi)) {
+		ret = spi_nand_read_from_cache(dev, 0, &onfi, sizeof(onfi));
```
tpambor:
```suggestion
		ret = spi_nand_read_from_cache(dev, i, &onfi, sizeof(onfi));
```
This was always reading the first parameter page.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+		return -ENODEV;
+	}
+
+	acquire_device(dev);
+
+	/* Soft RESET chip */
+	ret = spi_nand_cmd_write(dev, SPI_NAND_CMD_RESET);
+	if (ret != 0) {
+		goto release;
+	}
```
tpambor: Status should be checked after issuing the RESET command.
> After a RESET command (FFh) is issued, the Status register OIP bit0 will go high. This bit can be polled to determine when the Reset operation is complete, as it will return to the default value (0) after the reset operation is finished.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+
+	/* Validate JEDEC ID */
+	ret = spi_nand_cmd_read_dummy(dev, SPI_NAND_CMD_READ_ID, jedec_id, SPI_NAND_MAX_ID_LEN);
+	if (ret != 0) {
+		goto release;
+	}
+	if (memcmp(jedec_id, config->jedec_id, sizeof(jedec_id)) != 0) {
+		LOG_ERR("Device id %02x %02x does not match config %02x %02x", jedec_id[0],
+			jedec_id[1], config->jedec_id[0], config->jedec_id[1]);
+		return -EINVAL;
```
tpambor: `release_device(dev)` is  missing here

drivers/flash/spi_nand.h
```diff
@@ -0,0 +1,208 @@
...
+	/** Operation in progress */
+	SPI_NAND_FEATURE_STATUS_OIP = BIT(0),
+	/** Write enable latch */
+	SPI_NAND_FEATURE_STATUS_WEL = BIT(1),
+	/** Block erase operation failed */
+	SPI_NAND_FEATURE_STATUS_ERASE_FAIL = BIT(2),
+	/** Page program operation failed */
+	SPI_NAND_FEATURE_STATUS_PROGRAM_FAIL = BIT(3),
+	/** Page program operation failed */
+	SPI_NAND_FEATURE_STATUS_CACHE_BUSY = BIT(7),
```
tpambor: Comment is wrong.You might even consider removing this as it is not used and seems not to be present on all chips.

drivers/flash/spi_nand.h
```diff
@@ -0,0 +1,208 @@
...
+	/** Number of blocks per logical unit (LUN) */
+	uint32_t blocks_per_lun;
+	/** Number of logical units (LUNs) */
+	uint8_t num_lun;
+	/** Number of address cycles (4-7: Column cycles, 0-3 Row cycles) */
+	uint8_t address_cycles;
+	/** Number of bits per cell */
+	uint8_t bits_per_cell;
+	/** Bad blocks maximum per LUN */
+	uint16_t back_blocks_per_lun;
```
tpambor:
```suggestion
	uint16_t bad_blocks_per_lun;
```

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: int
+    required: true
+    description: |
+      Maximum erase duration in microseconds.
+
+  page-program-duration-max:
+    type: int
+    required: true
+    description: |
+      Maximum erase duration in microseconds.
```
tpambor:
```suggestion
      Maximum program duration in microseconds.
```

drivers/flash/spi_nand.h
```diff
@@ -0,0 +1,208 @@
...
+	/** Clear WEL bit in the status register */
+	SPI_NAND_CMD_WRITE_DISABLE = 0x04,
+	/** Set WEL bit in the status register */
+	SPI_NAND_CMD_WRITE_ENABLE = 0x06,
+	/** Read data from main storage to NAND cache */
+	SPI_NAND_CMD_PAGE_READ = 0x13,
+	/** Read data from NAND cache */
+	SPI_NAND_CMD_READ_CACHE = 0x03,
+	/** Write memory contents to NAND cache */
+	SPI_NAND_CMD_PROGRAM_LOAD = 0x2,
```
tpambor:
```suggestion
	SPI_NAND_CMD_PROGRAM_LOAD = 0x02,
```
Nit

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,771 @@
...
+#define DEFINE_PAGE_LAYOUT(idx)                                                                    \
+	IF_ENABLED(CONFIG_FLASH_PAGE_LAYOUT,                                                       \
+		   (.layout = {                                                                    \
+			    .pages_count = DT_INST_PROP(idx, size_bytes) /                         \
+					   DT_INST_PROP(idx, erase_block_size),                    \
+			    .pages_size = DT_INST_PROP(idx, erase_block_size),                     \
+		    },))
+/* clang-format on */
+
+#define SPI_NAND_INST(idx)                                                                         \
```
tpambor, line 809:
```suggestion
#define SPI_NAND_INST(idx)                                                                         \
	BUILD_ASSERT(IS_POWER_OF_TWO(DT_INST_PROP(idx, write_block_size)),                         \
		     "write-block-size must be a power of 2");                                     \
	BUILD_ASSERT(IS_POWER_OF_TWO(DT_INST_PROP(idx, erase_block_size)),                         \
		     "erase-block-size must be a power of 2");                                     \
	BUILD_ASSERT(DT_INST_PROP(idx, erase_block_size) % DT_INST_PROP(idx, write_block_size)     \
		    == 0, "erase-block-size must be a multiple of write-block-size");              \
	BUILD_ASSERT(DT_INST_PROP(idx, size_bytes) % DT_INST_PROP(idx, erase_block_size) == 0,     \
		     "size-bytes must be a multiple of erase-block-size");                         \
```
Validate device tree settings at compile time

tpambor:
> I wasn't aware that these were defined as vendor specific commands, but as I mentioned in the PR description I did my best to ensure that only defines that were common across the two manufacturers were used. Maybe it won't end up being true for all NAND flashes, but I would want to see a counter-example before complicating the driver.

I had a closer look across vendor datasheets and these registers are sufficiently similar to implement a generic solution for the basic functionality.

> Correct, the ECC bits are currently not checked since AFAIK its not required for this basic operation mode. Obviously when combined with something like Dhara they start to have meaning, but what is the flash API supposed to do when these bits start to be set?

I would expect that if ECC detects unrecoverable errors, this is propagated as an error (like -EIO) to the upper layers (FTL, FS, application). Dhara currently doesn't do anything too special: if a page contains unrecoverable errors, it propagates that error for data pages and avoids utilizing them for metadata. While handling this isn't strictly required for basic operation, "bad things" may happen if unrecoverable errors are present but not reported, regardless of whether Dhara is used or not.

> Note that the program and erase fail bits are common, and are checked to confirm operation success/failure.

This is fine and all that Dhara expects. Dhara would then mark the block as a bad but these extensions can be implemented in a follow-up PR.

drivers/flash/spi_nand.h
JordanYates: Are you sure? Even in Macronix I'm looking at a datasheet that marks all these bits as reserved.
image

drivers/flash/spi_nand.h
JordanYates: Updated the comment, can you point me at a datasheet that doesn't define the bit?

drivers/flash/spi_nand.h
JordanYates: that can be done when some other user/driver wants the same defines

drivers/flash/spi_nand.h
tpambor: You are looking at the datasheet of MX35LF4G24AD, which has **no** built-in Hardware ECC support but relies on ECC in the MCU or Software ECC, therefore it has these bits reserved. MX35LF4GE4AD is the variant, which **has** built-in Hardware ECC support. As I understood, we will only support flashes with ECC until there is software ECC support. There would also be a warning for MX35LF4G24AD without hardware ECC:

https://github.com/Embeint/zephyr/blob/97f4c915ad2da5ee63638a7fe377a8a7de185d06/drivers/flash/spi_nand.c#L565-L568

image

JordanYates: Ah, that would explain it, I just  grabbed the first Macronix datasheet I found since I didn't have a specific chip to test.

drivers/flash/spi_nand.h
tpambor: Winbond NAND flashes don't have it.
image

drivers/flash/spi_nand.h
JordanYates: Are you able to link me to where you found that datasheet (or a more recent part)? The part number is obsolete based on my googling and I can't find it.

drivers/flash/spi_nand.h
tpambor: Or also for Gigadevice NAND flashes
image

drivers/flash/spi_nand.h
tpambor: https://www.mxic.com.tw/en-us/products/NAND-Flash/Serial-NAND-Flash/Pages/default.aspx

The ones with hardware ECC are those which have ECC requirement as "ECC-free".

JordanYates:
> Is your proposal the these features be handled elsewhere?

My proposal is that advanced features can be debated in future PRs without bogging down the discussion for baseline support. That's what killed the other PR.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		return -EINVAL;
+	}
+
+	acquire_device(dev);
+
+	while (size > 0) {
+		page_address = addr >> config->addr_page_shift;
+		page_offset = addr & config->addr_offset_mask;
+		bytes_to_end = config->parameters->write_block_size - page_offset;
+		bytes_to_read = min(size, bytes_to_end);
```
tpambor:
```suggestion
		bytes_to_read = MIN(size, bytes_to_end);
```
nit

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	}
+
+	acquire_device(dev);
+
+	/* Soft RESET chip */
+	ret = spi_nand_cmd_write(dev, SPI_NAND_CMD_RESET);
+	if (ret != 0) {
+		goto release;
+	}
+	ret = spi_nand_wait_until_ready(dev, "reset", config->reset_us, 100, &status);
```
tpambor, line 718:
```suggestion
	ret = spi_nand_wait_until_ready(dev, config->reset_us, 100, &status);
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	static const struct flash_parameters spi_nand_##idx##_parameters = {                       \
+		.write_block_size = DT_INST_PROP(idx, write_block_size),                           \
+		.erase_value = 0xff,                                                               \
+	};                                                                                         \
+	static const struct spi_nand_config spi_nand_##idx##_config = {                            \
+		.spi = SPI_DT_SPEC_INST_GET(idx, SPI_WORD_SET(8)),                                 \
+		.parameters = &spi_nand_##idx##_parameters,                                        \
+		.flash_size = DT_INST_PROP(idx, size_bytes),                                       \
+		.block_size = DT_INST_PROP(idx, erase_block_size),                                 \
+		.block_erase_us = DT_INST_PROP(idx, block_erase_duration_max),                     \
```
tpambor, line 830:
```suggestion
		.block_erase_us = DT_INST_PROP(idx, block_erase_duration_max),                     \
		.reset_us = DT_INST_PROP(idx, reset_duration_max),                                 \
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	/* Size of device in bytes */
+	uint32_t flash_size;
+	/* Size of erase unit in bytes */
+	uint32_t block_size;
+	/* Maximum duration to erase a block */
+	uint32_t block_erase_us;
+	/* Maximum duration to program a page */
+	uint32_t page_program_us;
+	/* Maximum duration to read a page to cache */
+	uint32_t page_read_us;
```
tpambor:
```suggestion
	/* Maximum duration to read a page to cache */
	uint32_t page_read_us;
	/* Maximum duration to perform soft-reset */
	uint32_t reset_us;
```

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: int
+    required: true
+    description: |
+      Maximum program duration in microseconds.
+
+  page-read-duration-max:
+    type: int
+    required: true
+    description: |
+      Maximum read duration in microseconds.
```
tpambor:
```suggestion
  page-read-duration-max:
    type: int
    required: true
    description: |
      Maximum read duration in microseconds.
      
  reset-duration-max:
    type: int
    required: true
    description: |
      Maximum soft-reset duration in microseconds.
```

tests/drivers/build_all/flash/spi.dtsi
```diff
@@ -44,3 +44,18 @@ s3axx04@3 {
...
+	status = "okay";
+	spi-max-frequency = <5000000>;
+	/* mt29f4g01 parameters */
+	jedec-id = [2c 35];
+	size-bytes = <536870912>;
+	erase-block-size = <262144>;
+	write-block-size = <4096>;
+	block-erase-duration-max = <10000>;
+	page-program-duration-max = <600>;
+	page-read-duration-max = <200>;
```
tpambor, line 60:
```suggestion
	page-read-duration-max = <200>;
	reset-duration-max = <1250>;
```

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	/* Validate JEDEC ID */
+	ret = spi_nand_cmd_read_dummy(dev, SPI_NAND_CMD_READ_ID, jedec_id, SPI_NAND_MAX_ID_LEN);
+	if (ret != 0) {
+		goto release;
+	}
+	if (memcmp(jedec_id, config->jedec_id, sizeof(jedec_id)) != 0) {
+		LOG_ERR("Device id %02x %02x does not match config %02x %02x", jedec_id[0],
+			jedec_id[1], config->jedec_id[0], config->jedec_id[1]);
+		ret = -EINVAL;
+		goto release;
```
de-nordic: There is comparison of all SPI_NAND_MAX_ID_LEN bytes, but only first two are compared in error, it will be funny to get info that devices does not match yet report first two bytes match perfectly.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		LOG_WRN("Parameters at offset %u corrupt (%04X != %04X)", i, computed_crc,
+			onfi.integrity_crc);
+	}
+	if (computed_crc != onfi.integrity_crc) {
+		return -ENOSPC;
+	}
+
+	/* NULL terminate for display */
+	onfi.device_manufacturer[11] = '\0';
+	onfi.device_model[19] = '\0';
```
de-nordic: Instead of doing that you should use formatting field to cut down number of characters to 12 and 20

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		computed_crc =
+			crc16(CRC16_POLY, 0x4F4E, (void *)&onfi, sizeof(onfi) - sizeof(uint16_t));
+		if (computed_crc == onfi.integrity_crc) {
+			break;
+		}
+		LOG_WRN("Parameters at offset %u corrupt (%04X != %04X)", i, computed_crc,
+			onfi.integrity_crc);
+	}
+	if (computed_crc != onfi.integrity_crc) {
+		return -ENOSPC;
```
de-nordic, line 656: At least LOG_DBG would be nice.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	ret = spi_nand_page_read_to_cache(dev, 1);
+	if (ret != 0) {
+		return ret;
+	}
+
+	/* Explicilty set CRCs to be mismatched, since static analysis doesn't know
+	 * config->parameters->write_block_size is always non-zero.
+	 */
+	computed_crc = 0;
+	onfi.integrity_crc = 1;
```
de-nordic, line 634: The comment is little bit misleading. Just state that these are expected to be equal when loop hits valid onfi structure.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		spi_buf[0].len += address_len;
+	}
+
+	if (access & NAND_ACCESS_DUMMY_BYTE) {
+		spi_buf[0].len += 1;
+	}
+
+	const struct spi_buf_set tx_set = {
+		.buffers = spi_buf,
+		.count = (length != 0) ? 2 : 1,
```
de-nordic, line 188:
```suggestion
       /* Non zero length means that data follows opcode, so there are two buffers to tx */
		.count = (length != 0) ? 2 : 1,
```

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2025 Embeint Pty Ltd
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  Properties supporting Zephyr spi-nand flash driver (over the Zephyr SPI
```
de-nordic:
```suggestion
  Properties supporting Zephyr SPI-NAND flash driver (over the Zephyr SPI
```

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+compatible: "jedec,spi-nand"
+
+include: [spi-device.yaml]
+
+properties:
+  jedec-id:
+    type: uint8-array
+    required: true
+    description: JEDEC ID as manufacturer ID, memory type, memory density
```
de-nordic: Just JEDEC ID. As far as I remember that you can only be sure of manufacturer id and memory model.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	ret = spi_nand_wait_until_ready(dev, "reset", config->reset_us, 100, &status);
+	if (ret != 0) {
+		goto release;
+	}
+
+	/* Validate JEDEC ID */
+	ret = spi_nand_cmd_read_dummy(dev, SPI_NAND_CMD_READ_ID, jedec_id, SPI_NAND_MAX_ID_LEN);
+	if (ret != 0) {
+		goto release;
+	}
```
de-nordic, line 727: Validate me on that, and yes the NOR may have the same problem, but JEDEC ID may have something like 17 characters up to memory model. Seems that we have to read the thing in loop till 0x7f stops appearing.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+
+static int spi_nand_write(const struct device *dev, off_t addr, const void *src, size_t size)
+{
+	const struct spi_nand_config *config = dev->config;
+	uint32_t write_block = config->parameters->write_block_size;
+	uint8_t *src_u8 = (void *)src;
+	uint32_t page_address;
+	uint8_t status;
+	int ret = 0;
+
```
de-nordic, line 421: First check size against 0, if there is nothing to work on then there is no reason to fail.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	release_device(dev);
+	return ret;
+}
+
+static int spi_nand_write(const struct device *dev, off_t addr, const void *src, size_t size)
+{
+	const struct spi_nand_config *config = dev->config;
+	uint32_t write_block = config->parameters->write_block_size;
+	uint8_t *src_u8 = (void *)src;
```
de-nordic, line 417: Why `const` gets dropped here on conversion?

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		ret = spi_nand_page_read_to_cache(dev, page_address);
+		if (ret != 0) {
+			break;
+		}
+
+		/* Read data out of cache */
+		ret = spi_nand_read_from_cache(dev, page_offset, dest_u8, bytes_to_read);
+		if (ret != 0) {
+			break;
+		}
```
de-nordic, line 401: LOG_DBG would be nice, because once you break out of this and return value in 374 you do not really know what has caused the problem,

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+				      NAND_ACCESS_WRITE | NAND_ACCESS_ADDRESSED |
+					      NAND_ACCESS_16BIT_ADDR,
+				      0, src_u8, write_block);
+		if (ret != 0) {
+			break;
+		}
+
+		/* Program the cache to the appropriate page */
+		page_address = addr >> config->addr_page_shift;
+		LOG_DBG("Write %d to %06x:000", write_block, page_address);
```
de-nordic: You can as well move it above 424 and add rc in printout.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+		if ((*status & SPI_NAND_FEATURE_STATUS_OIP) == 0U) {
+			t = k_uptime_ticks() - t;
+			LOG_DBG("Ready after %u us (Status %02X)", k_ticks_to_us_near32(t),
+				*status);
+			return ret;
+		}
+		k_sleep(K_USEC(poll_us));
+	} while (!sys_timepoint_expired(timeout));
+
+	LOG_ERR("Timeout waiting for flash ready (Status %02X)", *status);
```
de-nordic: This probably should be debug. Status does not matter that much in production code and you have to deal with timeout in calling code, so you will probably log the timeout info there too; probably with more accurate info like device you have been using.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,769 @@
...
+	return ret;
+}
+
+static int spi_nand_erase(const struct device *dev, off_t addr, size_t size)
+{
+	const struct spi_nand_config *config = dev->config;
+	uint32_t page_address;
+	uint8_t status;
+	int ret = 0;
+
```
de-nordic, line 500: size == 0, means no work. There is  no reason to check other params.

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
...
+    required: true
+    description: |
+      Address alignment required by flash write operations. Equivalent to
+      the page size for SPI NAND flash storage.
+
+  block-erase-duration-max:
+    type: int
+    required: true
+    description: |
+      Maximum erase duration in microseconds.
```
de-nordic:
```suggestion
      Maximum erase duration of a block in microseconds.
```

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  page-program-duration-max:
+    type: int
+    required: true
+    description: |
+      Maximum program duration in microseconds.
```
de-nordic:
```suggestion
      Maximum page program duration in microseconds.
```

dts/bindings/mtd/jedec,spi-nand.yaml
de-nordic:
```suggestion
      Maximum page read duration in microseconds.
```

de-nordic (changes requested):
Sorry it takes long, overworked and under the weather.

drivers/flash/spi_nand.c
JordanYates: Personally I've found it useful even when not actively developing the driver. It is an error, and does provide context that isn't propagated further up, `LOG_ERR` seems appropriate to me.

drivers/flash/spi_nand.c
JordanYates: Because `spi_nand_access` (which takes the pointer) is common between the read and write functionality, and read buffers are mutable.

drivers/flash/spi_nand.c
JordanYates: I'll reword it, this code is only here because static analysis thinks the two values could be uninitialised at the comparison point below the loop because it doesn't know that `config->parameters->write_block_size` is always non-zero.

drivers/flash/spi_nand.c
JordanYates: You can't get to this exit condition without hitting the `LOG_WRN` 3 lines above many many times.

drivers/flash/spi_nand.c
JordanYates: I wrote a whole comment about how `"%20s"` doesn't do that, but then was reading https://cplusplus.com/reference/cstdio/printf/ and found that `"%.20s"` does. Thanks for the pointer :)

drivers/flash/spi_nand.c
JordanYates: Hmmm, the NAND I was testing against defines this as a 2 byte value, but looking at other datasheets I can see they have different lengths. I'll come up with something better.

drivers/flash/spi_nand.c
JordanYates: This worked fine previously because I was doing `#define SPI_NAND_MAX_ID_LEN 2`. I'm not sure the best way to display these for variable length IDs.

drivers/flash/spi_nand.c
JordanYates: Where did you get 17 characters from? The two datasheets I'm looking at define the ID as 2 and 3 bytes long.
Image
Image

drivers/flash/spi_nand.c
JordanYates: One thing I want for this is a specific error code for corrupt data. `-EIO` implies that maybe if the user tries again it will work, which is not the case. Unfortunately I can't see any errors specifically about corruption in `errno.h`. I'm thinking either `ENODATA`, `ENOMSG`, or `EBADMSG`, even though they don't strictly match.

dts/bindings/mtd/jedec,spi-nand.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+  page-read-duration-max:
+    type: int
+    required: true
+    description: |
+      Maximum read duration in microseconds.
+
+  reset-duration-max:
+    type: int
+    default: 1000
```
tpambor: In the micron datasheet is a note:
> For first RESET condition after power-up, tRST will be 1.25ms maximum

drivers/flash/spi_nand.c
tpambor: There could be 3 corrupt and a 4th valid page. I think logging something like "all parameter pages corrupt" would not hurt here.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,844 @@
...
+			crc16(CRC16_POLY, 0x4F4E, (void *)&onfi, sizeof(onfi) - sizeof(uint16_t));
+		if (computed_crc == onfi.integrity_crc) {
+			LOG_DBG("Valid CRC: %04X", computed_crc);
+			break;
+		}
+		LOG_WRN("Parameters at offset %u corrupt (%04X != %04X)", i, computed_crc,
+			onfi.integrity_crc);
+	}
+	if (computed_crc != onfi.integrity_crc) {
+		return -ENOSPC;
```
tpambor, line 656: I think it would be good to leave parameter page read mode by clearing OTP_EN again before returning.

drivers/flash/spi_nand.c
```diff
@@ -0,0 +1,844 @@
...
+
+	/* Read area must be subregion of device */
+	if (!valid_region(dev, addr, size)) {
+		return -EINVAL;
+	}
+
+	if (size == 0) {
+		/* No work to do */
+		return 0;
+	}
```
tpambor, line 373: To be consistent with the others, check size first.

drivers/flash/spi_nand.c
tpambor: Yes, that is due to the JEDEC continuation scheme. Linux currently also does only read 5. See also https://lore.kernel.org/linux-mtd/20240125200108.24374-2-ezra@easyb.ch/

drivers/flash/spi_nand.c
JordanYates: Is it useful in the context that if this fails the device can't be used further as `device_init` has failed, and there are other error paths that leave the bit enabled?

tpambor: True, I was thinking about reducing chances to accidentally program OTP pages when fiddling around with it (during development). I am also fine with leaving as is.

drivers/flash/spi_nand.c
de-nordic: We are not supposed to do that according to https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html, rule 70

drivers/flash/spi_nand.c
de-nordic: I also wonder whether the LOG_WRN, in 651, should be LOG_DBG instead. If it is a case that we just have to search for one valid page to be able to go, the problem is when we find none.

drivers/flash/spi_nand.c
de-nordic: Jesd 106 "Standard Manufacturer’s Identification Code" , the document only contains id codes.
Leave it as is. Till we figure common way to address it with NOR driver.

de-nordic (changes requested):
I think it looks OK, the only controversial addition is the line that removes const, as it breaks one of our coding guidelines.

drivers/flash/spi_nand.c
JordanYates: `spi_nor` does the same thing hidden behind some macros:
https://github.com/zephyrproject-rtos/zephyr/blob/3c64b79ddb645afc2a06dedd6fbb32ea8b88e585/drivers/flash/spi_nor.c#L487

And even if it didn't, the exact same casting would be required when `struct spi_buf` is constructed anyway, since the pointer there is a `void *`, since it is used for both TX and RX operations.

drivers/flash/spi_nand.c
JordanYates: I've updated the maximum length to 8 and add updated the code to compare to the same number of bytes as provided by devicetree.

tpambor (approved):
LGTM. Retested on hardware:
```
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load: Valid CRC: 3D0F
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:      Manufacturer: WINBOND
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:             Model: W25N01GV
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:  Page Size (data): 2048
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load: Page Size (spare): 64
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:   Pages per Block: 64
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:   Blocks per Unit: 1024
[00:00:00.002,000] <dbg> spi_nand: onfi_parameters_load:             Units: 1
uart:~$ flash test spi-nand@0 0 20000 1
Erase OK.
Write OK.
Verified OK.
Erase-Write-Verify test done.
```

de-nordic (approved):

prashant3285:
This issue is due to NAND with multiple planes.
Odd and even blocks are located in different planes and for Reading, Writing a block, its designated plane needs to be selected.

PR #100858, opened 2025-12-11
NFTL Disk Driver

LinderPi:
Zephyr has currently no flash translation layer implemented. This pull request adds the basic possibility to use Dhara as FTL in Zephyr. It is implemented as disk driver that can be placed between a filesystem and a flash driver. The FTL uses API extensions for bad block management, which are also faked on the flash simulator to facilitate tests.

The [Dhara repository](https://github.com/endresshauser-lp/dhara/tree/zephyr) shall be moved to the Zephyr organisation as soon as the module is ready for that (#99908, #100834).

### DIFF
--- a/drivers/disk/ftl_dhara.c
+++ b/drivers/disk/ftl_dhara.c
@@ -1,0 +1,570 @@
+/*
+ * Copyright (c) 2025 Endress+Hauser GmbH+Co. KG
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT zephyr_ftl_dhara
+
+#include <dhara/map.h>
+#include <dhara/nand.h>
+#include <zephyr/drivers/disk.h>
+#include <zephyr/drivers/flash.h>
+#include <zephyr/storage/flash_map.h>
+#include <zephyr/sys/math_extras.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(disk_ftl, CONFIG_DISK_FTL_LOG_LEVEL);
+
+struct disk_ftl_data {
+#if CONFIG_DISK_FTL_SUPPORT_CONCURRENT_ACCESS
+	struct k_sem lock;
+#endif
+	bool initialised;
+	struct disk_info info;
+	const struct flash_area *area;
+	size_t page_size;
+	size_t block_size;
+	size_t buffer_size;
+	uint8_t *page_buffer;
+	struct dhara_map dhara_map;
+	struct dhara_nand dhara_nand;
+	uint8_t *dhara_buffer;
+};
+
+int dhara_nand_is_bad(const struct dhara_nand *n, dhara_block_t b)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t block_addr = b * ctx->block_size + ctx->area->fa_off;
+	int block_status = FLASH_BLOCK_BAD;
+	int ret;
+
+	ret = flash_ex_op(ctx->info.dev, FLASH_EX_OP_IS_BAD_BLOCK, (uintptr_t)&block_addr,
+			  &block_status);
+	if (ret == -ENOTSUP) {
+		LOG_DBG("checking bad block is not supported");
+		return 0;
+	} else if (ret != 0) {
+		LOG_ERR("checking bad block at 0x%08lx failed with error %d", (long)block_addr,
+			ret);
+		return 1;
+	} else if (block_status == FLASH_BLOCK_BAD) {
+		LOG_DBG("block at 0x%08lx is marked bad", (long)block_addr);
+	} else {
+		/* No more action if block is good */
+		LOG_DBG("block at 0x%08lx is good", (long)block_addr);
+	}
+
+	return (block_status == FLASH_BLOCK_BAD) ? 1 : 0;
+}
+
+void dhara_nand_mark_bad(const struct dhara_nand *n, dhara_block_t b)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t block_addr = b * ctx->block_size + ctx->area->fa_off;
+	int ret;
+
+	ret = flash_ex_op(ctx->info.dev, FLASH_EX_OP_MARK_BAD_BLOCK, (uintptr_t)&block_addr, NULL);
+	if (ret == -ENOTSUP) {
+		LOG_DBG("marking bad block is not supported");
+	} else if (ret != 0) {
+		LOG_ERR("marking bad block at 0x%08lx failed with error %d", (long)block_addr, ret);
+	} else {
+		LOG_DBG("marked block bad at 0x%08lx", (long)block_addr);
+	}
+}
+
+int dhara_nand_erase(const struct dhara_nand *n, dhara_block_t b, dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t block_addr = b * ctx->block_size + ctx->area->fa_off;
+	int ret;
+
+	LOG_DBG("erasing block at 0x%08lx", (long)block_addr);
+
+	ret = flash_erase(ctx->info.dev, block_addr, ctx->block_size);
+	if (ret != 0) {
+		LOG_ERR("erasing block at 0x%08lx failed with error %d", (long)block_addr, ret);
+		*err = DHARA_E_BAD_BLOCK;
+		return -1;
+	}
+
+	return 0;
+}
+
+int dhara_nand_prog(const struct dhara_nand *n, dhara_page_t p, const uint8_t *data,
+		    dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t page_addr = p * ctx->page_size + ctx->area->fa_off;
+	int ret;
+
+	LOG_DBG("writing page at 0x%08lx", (long)page_addr);
+
+	ret = flash_write(ctx->info.dev, page_addr, data, ctx->page_size);
+	if (ret != 0) {
+		LOG_ERR("writing page at 0x%08lx failed with error %d", (long)page_addr, ret);
+		*err = DHARA_E_BAD_BLOCK;
+		return -1;
+	}
+
+	return 0;
+}
+
+int dhara_nand_is_free(const struct dhara_nand *n, dhara_page_t p)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	const struct flash_parameters *flash_params = flash_get_parameters(ctx->info.dev);
+	off_t page_addr = p * ctx->page_size + ctx->area->fa_off;
+	int ret;
+
+	LOG_DBG("checking erase status of page at 0x%08lx", (long)page_addr);
+
+	ret = flash_read(ctx->info.dev, page_addr, ctx->page_buffer, ctx->page_size);
+	if (ret != 0) {
+		LOG_ERR("reading page at 0x%08lx failed with error %d", (long)page_addr, ret);
+		return 0;
+	}
+
+	if (ctx->page_buffer[0] != flash_params->erase_value) {
+		return 0;
+	}
+
+	if (memcmp(ctx->page_buffer, ctx->page_buffer + 1, ctx->page_size - 1) != 0) {
+		return 0;
+	}
+
+	return 1;
+}
+
+int dhara_nand_read(const struct dhara_nand *n, dhara_page_t p, size_t offset, size_t length,
+		    uint8_t *data, dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t page_addr = p * ctx->page_size + ctx->area->fa_off;
+	int ret;
+
+	LOG_DBG("reading page at 0x%08lx, offset 0x%zx, length 0x%zx", (long)page_addr, offset,
+		length);
+
+	ret = flash_read(ctx->info.dev, page_addr + offset, data, length);
+	if (ret != 0) {
+		LOG_ERR("reading data at 0x%08lx failed with error %d", (long)(page_addr + offset),
+			ret);
+		*err = DHARA_E_ECC;
+		return -1;
+	}
+
+	return 0;
+}
+
+int dhara_nand_copy(const struct dhara_nand *n, dhara_page_t src, dhara_page_t dst,
+		    dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t src_page_addr = src * ctx->page_size + ctx->area->fa_off;
+	off_t dst_page_addr = dst * ctx->page_size + ctx->area->fa_off;
+	int ret;
+
+	LOG_DBG("copying page from 0x%08lx to 0x%08lx", (long)src_page_addr, (long)dst_page_addr);
+
+	ret = flash_read(ctx->info.dev, src_page_addr, ctx->page_buffer, ctx->page_size);
+	if (ret != 0) {
+		LOG_ERR("reading page at 0x%08lx failed with error %d", (long)src_page_addr, ret);
+		*err = DHARA_E_ECC;
+		return -1;
+	}
+
+	ret = flash_write(ctx->info.dev, dst_page_addr, ctx->page_buffer, ctx->page_size);
+	if (ret != 0) {
+		LOG_ERR("writing page at 0x%08lx failed with error %d", (long)dst_page_addr, ret);
+		*err = DHARA_E_BAD_BLOCK;
+		return -1;
+	}
+
+	return 0;
+}
+
+static __maybe_unused int flash_area_is_uniform(const struct flash_area *fa)
+{
+	int ret;
+	struct flash_pages_info page;
+	size_t block_size = 0;
+	off_t offset = fa->fa_off;
+
+	while (offset < (fa->fa_off + fa->fa_size)) {
+		ret = flash_get_page_info_by_offs(fa->fa_dev, offset, &page);
+		if (ret != 0) {
+			return ret;
+		}
+
+		if (block_size == 0) {
+			block_size = page.size;
+		}
+
+		if (page.size != block_size) {
+			return -EINVAL;
+		}
+
+		offset += page.size;
+	}
+
+	return 0;
+}
+
+/* Everything necessary to acquire owning access to the disk. */
+static void acquire_disk(__maybe_unused struct disk_info *disk)
+{
+#if CONFIG_DISK_FTL_SUPPORT_CONCURRENT_ACCESS
+	struct disk_ftl_data *ctx = CONTAINER_OF(disk, struct disk_ftl_data, info);
+
+	k_sem_take(&ctx->lock, K_FOREVER);
+#endif
+}
+
+/* Everything necessary to release access to the disk. */
+static void release_disk(__maybe_unused struct disk_info *disk)
+{
+#if CONFIG_DISK_FTL_SUPPORT_CONCURRENT_ACCESS
+	struct disk_ftl_data *ctx = CONTAINER_OF(disk, struct disk_ftl_data, info);
+
+	k_sem_give(&ctx->lock);
+#endif
+}
+
+int disk_ftl_access_init(struct disk_info *disk)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(disk, struct disk_ftl_data, info);
+	const struct flash_parameters *flash_params;
+	struct flash_pages_info page; /* Information of the smallest erasable area */
+	dhara_error_t err;
+	int ret;
+
+	acquire_disk(disk);
+
+	if (ctx->initialised) {
+		release_disk(disk);
+		LOG_ERR("FTL is already initialised");
+		return -EALREADY;
+	}
+
+	if (!flash_area_device_is_ready(ctx->area)) {
+		release_disk(disk);
+		LOG_ERR("Flash device %s is not ready", ctx->area->fa_dev->name);
+		return -ENODEV;
+	}
+
+	ctx->info.dev = flash_area_get_device(ctx->area);
+
+	if (ctx->info.dev == NULL) {
+		release_disk(disk);
+		LOG_ERR("Flash device was not found");
+		return -ENODEV;
+	}
+
+	flash_params = flash_get_parameters(disk->dev);
+	ctx->page_size = flash_params->write_block_size;
+	__ASSERT(ctx->buffer_size >= ctx->page_size,
+		 "Buffer size %u is too small for pages with size %u", ctx->buffer_size,
+		 ctx->page_size);
+
+	ret = flash_get_page_info_by_offs(disk->dev, ctx->area->fa_off, &page);
+	if (ret != 0) {
+		release_disk(disk);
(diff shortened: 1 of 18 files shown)


drivers/flash/flash_simulator.c
```diff
@@ -380,6 +381,35 @@ flash_sim_get_parameters(const struct device *dev)
...
+	case NAND_FLASH_IS_BAD_BLOCK: {
+		/* Block is never bad on flash simulator */
+		*(int *)out = 0;
+		break;
+	}
+
+	case NAND_FLASH_MARK_BAD_BLOCK: {
+		/* Blocks cannot be marked bad on flash simulator */
+		break;
+	}
```
de-nordic: I think we should also return -ENOTSUP here, we should describe return values for the NAND_FLASH_MARK_BAD_BLOCK and code calling the op should properly address the ENTOSUP.

de-nordic: I am also not sure whether Flash Simulator, in the current form, is able to properly pretend NAND devices.
I think we should alter it to be able to use it for testing FTL on the CI, which also means that we may have to add ability to actually support bad blocks, as well as ecc or user bits.
I am not sure whether this is job for this pr, nevertheless we will not be able to reliably test ftl without it.

include/zephyr/drivers/flash/nand_flash_api_ex.h
```diff
@@ -0,0 +1,33 @@
...
+	/**
+	 * Checks whether a block is marked as bad. As input it takes the block index (size_t *). As
+	 * output it returns 1 if the block is bad, 0 otherwise (int *).
+	 */
+	NAND_FLASH_IS_BAD_BLOCK = FLASH_EX_OP_VENDOR_BASE,
+
+	/**
+	 * Marks a block as bad. As input it takes the block index (size_t *). There is no output.
+	 */
+	NAND_FLASH_MARK_BAD_BLOCK,
```
de-nordic: It is not vendor base, that is something we should standardize.
I think that we also should drop NAND_, as not only NAND may have such features.

modules/dhara/Kconfig
```diff
@@ -0,0 +1,25 @@
...
+
+config ZEPHYR_DHARA_MODULE
+	bool
+
+config DISK_DRIVER_FTL
+	bool "Flash translation layer"
+	default y
+	depends on DT_HAS_ZEPHYR_FTL_DHARA_ENABLED
+	depends on DISK_DRIVERS
+	select FLASH_EX_OP_ENABLED
```
de-nordic: if DISK_DRIVER_FLASH

drivers/flash/flash_simulator.c
```diff
@@ -380,6 +381,35 @@ flash_sim_get_parameters(const struct device *dev)
...
+{
+	ARG_UNUSED(dev);
+	ARG_UNUSED(in);
+
+	int ret = 0;
+
+	switch (code) {
+	case NAND_FLASH_IS_BAD_BLOCK: {
+		/* Block is never bad on flash simulator */
+		*(int *)out = 0;
```
de-nordic: GOOD = 0, BAD = !GOOD, name values.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+
+int dhara_nand_is_bad(const struct dhara_nand *n, dhara_block_t b)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	int is_bad = 1;
+	int ret;
+
+	ret = flash_ex_op(ctx->info.dev, NAND_FLASH_IS_BAD_BLOCK, (uintptr_t)&b, &is_bad);
+	if (ret == -ENOTSUP) {
+		LOG_INF("Checking bad block is not supported");
```
de-nordic: LOG_DBG as this is only needed on testing and evaluation of user system setup. Otherwise you will have a lot of these.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+		return 0;
+	} else if (ret != 0) {
+		LOG_ERR("Checking bad block failed with error %d", ret);
```
de-nordic: Worth logging the block number? Can we somehow translate it to device offset?

de-nordic: Also, this is for failed query, right? So we still should probably log , in else, when we actually catch bad block.
This may start bloating log at some point, but users should get some info that storage is wearing out, rigth?

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+int dhara_nand_erase(const struct dhara_nand *n, dhara_block_t b, dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	int ret;
+
+	LOG_DBG("%s: block at 0x%X", __func__, b * ctx->block_size);
+
+	ret = flash_erase(ctx->info.dev, b * ctx->block_size, ctx->block_size);
+	if (ret != 0) {
+		LOG_ERR("flash_erase failed with error %d", ret);
```
de-nordic: Shouldn't we log failed range here?

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+		    dhara_error_t *err)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	int ret;
+
+	LOG_DBG("%s: page at 0x%X", __func__, p * ctx->page_size);
+
+	ret = flash_write(ctx->info.dev, p * ctx->page_size, data, ctx->page_size);
+	if (ret != 0) {
+		LOG_ERR("flash_write failed with error %d", ret);
```
de-nordic: Shouldn't we log range here?

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+	if (ret != 0) {
+		LOG_ERR("flash_read failed with error %d", ret);
+		return 0;
+	}
+
+	for (size_t i = 0; i < ctx->page_size; i++) {
+		if (ctx->page_buffer[i] != flash_params->erase_value) {
+			return 0;
+		}
+	}
```
de-nordic: memset

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+                                                                                                   \
+	static struct disk_ftl_data disk_ftl_data_##n = {                                          \
+		.info =                                                                            \
+			{                                                                          \
+				.name = DT_INST_PROP(n, disk_name),                                \
+				.ops = &disk_ftl_ops,                                              \
+			},                                                                         \
+		.partition_id = DT_FIXED_PARTITION_ID(PARTITION_PHANDLE(n)),                       \
+		.partition_offset = DT_REG_ADDR(PARTITION_PHANDLE(n)),                             \
+		.partition_size = DT_REG_SIZE(PARTITION_PHANDLE(n)),                               \
```
de-nordic: Use FIXED_PARTITION macros here, and instead of ID, use the macro, FIXED_PARTITION, that directly takes pointer to `struct flash_area` object instance serving the selected partition.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+#include <zephyr/drivers/flash/nand_flash_api_ex.h>
+#include <zephyr/storage/flash_map.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(disk_ftl_dhara, CONFIG_DISK_FTL_LOG_LEVEL);
+
+struct disk_ftl_data {
+	struct k_sem lock;
+	struct disk_info info;
+	uint8_t partition_id;
```
de-nordic: Instead of using partition_id, use `struct flash_area` pointer.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+	struct flash_pages_info page; /* Information of the smallest erasable area */
+	dhara_error_t err;
+	off_t offset;
+	int ret;
+
+	ret = flash_area_open(ctx->partition_id, &area);
+	if (ret != 0) {
+		LOG_ERR("Opening flash area %u failed with error %d", ctx->partition_id, ret);
+		return ret;
+	}
```
de-nordic: Do not use `flash_area_open`. Try getting `struct flash_area` object pointer in `DISK_FTL_INIT` and use `flash_area_device_is_ready` on the object instead.
Once you do that, you can drop all the `flash_are_close`

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,424 @@
...
+		ctx->dhara_nand.num_blocks);
+
+	dhara_map_init(&ctx->dhara_map, &ctx->dhara_nand, ctx->dhara_buffer, ctx->dhara_gc_ratio);
+
+	ret = dhara_map_resume(&ctx->dhara_map, &err);
+	if (ret != 0) {
+		LOG_INF("dhara_map_resume failed with error %d", err);
+	}
+
+	k_sem_init(&ctx->lock, 1, 1);
```
de-nordic: Shouldn't this be first thing to happen and the entire init should be guarded by semaphore. There can be other init called on already working ftl. Also the `struct disk_ftl_data` should have "initialized" flag set once init is complete, to prevent double inits.

de-nordic: I am also not sure whether we even need semaphores here, but we can leave them for now.
A reason here is that one instance of Dhara, as attached to a partition, will only be used by one other subsystem directly, as partitions, nearly always have only one thing above it.

de-nordic (changes requested):

include/zephyr/drivers/flash/nand_flash_api_ex.h
LinderPi: That is why I put the enumeration in a generic NAND flash header that can be used by any driver. However, I would leave it for NAND memories as long as FTLs are only used there.

modules/dhara/Kconfig
LinderPi: The flashdisk driver must not be used. The NFTL disk driver replaces it for NAND memory devices with additional functionality for wear leveling and bad block management.

drivers/flash/flash_simulator.c
LinderPi: With the current addition, we can at least test the NFTL. `ENOTSUP` is correctly handled in NFTL code, but you are right that I should return it here as long as there is no other alternative.

drivers/disk/ftl_dhara.c
LinderPi: You are right. I changed this and also wanted to add a test for double initialisation, but the later did somehow not work to call the function twice (maybe due to compiler optimisation).

I would leave the semaphores as you stated. Also flashdisk does implement a mutex and has the same use case.

drivers/disk/ftl_dhara.c
LinderPi: I added the block address and a info log for marked bad blocks. In future, this disk driver shall provide an additional health monitoring API to allow applications checking the available good blocks.

include/zephyr/drivers/flash/nand_flash_api_ex.h
de-nordic: > That is why I put the enumeration in a generic NAND flash header that can be used by any driver. However, I would leave it for NAND memories as long as FTLs are only used there.

What I mean is that when we have defined FLASH_EX_OP_VENDOR_BASE, we have left it for vendor specific drivers, so something we can not agree to be universal across multiple vendors.
Ids below, which w already have a [one](https://github.com/zephyrproject-rtos/zephyr/blob/cb2109e7d9241922f5e534f98e4f083caf4471e2/include/zephyr/drivers/flash.h#L699), are for what we standardize for system. It does not mean that every driver has to implement it, it just means that we have recognized that the feature is common enough, but not crucial, to be accepted as something not vendor specific yet 

NAND_ is not vendor specific. MX25_ would be vendor specific STM has their own ex_op and they use the same range.
Now when you compile STM32 with NAND you have a problem where you will have NAND_FLASH_IS_BAD_BLOCK  being recognized by STM32 flash as something different.

It has been done so that we can, at some point, add direct `flash_` API calls that would invoke the `ex_op`, for example `flash_reset_device`, something we can not do with all varieties of vendor specific extensions.

modules/dhara/Kconfig
de-nordic: Sorry, my bad

include/zephyr/drivers/flash/nand_flash_api_ex.h
LinderPi: Makes sense. I moved it to the generic flash API extended operations.

tpambor:
> I've little confused. Should we need to add ctx->partition_offset back to every flash access functions, just like ctx->offset of flashdisk.c did?

Yes, you are right. I have fixed it.

tpambor:
- Updated to latest disk driver API including erase/trim support
- Added support for disk_access tests using native_sim
- Moved KConfig to disk drivers

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(disk_ftl, CONFIG_DISK_FTL_LOG_LEVEL);
+
+struct disk_ftl_data {
+	struct k_sem lock;
+	bool initialised;
+	struct disk_info info;
+	const struct flash_area *area;
+	off_t partition_offset;
```
de-nordic: You probably do not need `partition_offset` and `partition_size` if you have already the `area`.
`flash_area` object already holds the info as `fa_off` and `fa_size`

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+
+	return 0;
+}
+
+int dhara_nand_is_free(const struct dhara_nand *n, dhara_page_t p)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	const struct flash_parameters *flash_params = flash_get_parameters(ctx->info.dev);
+	off_t page_addr = p * ctx->page_size + ctx->partition_offset;
+	uint8_t empty_page[ctx->page_size];
```
de-nordic: No dynamic allocations on stack.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+		}
+
+		if (page.size != ctx->block_size) {
+			LOG_ERR("Non-uniform block size is not supported");
+			k_sem_give(&ctx->lock);
+			return -EINVAL;
+		}
+
+		offset += page.size;
+	}
```
de-nordic: Make the verification code optional and controller via Kconfig. You want to check that on you CI, probably does not make sense on in-field device. Just waste of cycles and silicon.
And I mean the params check to (I just can not select all of the code, for some reason).

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+#include <zephyr/drivers/disk.h>
+#include <zephyr/drivers/flash.h>
+#include <zephyr/storage/flash_map.h>
+#include <zephyr/sys/math_extras.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(disk_ftl, CONFIG_DISK_FTL_LOG_LEVEL);
+
+struct disk_ftl_data {
+	struct k_sem lock;
```
de-nordic, line 21: Parallel access to this will be rather rare, so should be kept under Kconfig.

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+	uint32_t total_sectors;
+	dhara_error_t err;
+	int ret;
+
+	k_sem_take(&ctx->lock, K_FOREVER);
+
+	total_sectors = dhara_map_capacity(&ctx->dhara_map);
+
+	if (u32_add_overflow(start_sector, num_sector, &end_sector) ||
+	    (end_sector > total_sectors)) {
```
de-nordic: You can do that simpler:

```suggestion
	if (total_sectors < num_sector || (total_sector - num_sectors) < start_sector) {
	// error
	}
	end_sector = start_sector + num_sector;
```

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+		LOG_ERR("Requested sectors are out of range");
+		k_sem_give(&ctx->lock);
+		return -EINVAL;
+	}
+
+	for (uint32_t i = 0; i < num_sector; i++) {
+		uint32_t sector = start_sector + i;
+		const uint8_t *buffer = data_buf + (i * ctx->page_size);
```
de-nordic: You know that you can avoid multiplication by adding `page_size` on each iteration?

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,508 @@
...
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t block_addr = b * ctx->block_size + ctx->partition_offset;
+	int ret;
+
+	LOG_DBG("erasing block at 0x%08lx", block_addr);
+
+	ret = flash_erase(ctx->info.dev, block_addr, ctx->block_size);
+	if (ret != 0) {
+		LOG_ERR("erasing block at 0x%08lx failed with error %d", block_addr, ret);
```
de-nordic: You should probably cast `block_addr` to long, as you format it that way.

de-nordic (changes requested):

drivers/disk/ftl_dhara.c
tpambor: Yes, I have dropped them.

drivers/disk/ftl_dhara.c
tpambor: I have converted them to asserts so this code will only be present if asserts are enabled.

drivers/disk/ftl_dhara.c
tpambor: I have added a option to configure this. It is ok to leave it disabled e.g. for FATFS with a single volume.

SuperHeroAbner:
Hi @tpambor and @de-nordic , do we have chance to merge this PR recently into v4.4.0? I'm doing [some work](https://github.com/zephyrproject-rtos/zephyr/pull/104870) based on this. Will wait these work to merge first.

tpambor:
> Hi @tpambor and @de-nordic , do we have chance to merge this PR recently into v4.4.0? I'm doing [some work](https://github.com/zephyrproject-rtos/zephyr/pull/104870) based on this. Will wait these work to merge first.

I would like to land this as well for v4.4. This PR has been open for a while now. I have addressed all open comments. If there are any remaining issues, I will prioritize fixing them. Downstream, we have been using this for a while without issues.

There are already three drivers with PRs pending that can utilize it:
- https://github.com/zephyrproject-rtos/zephyr/pull/104685 (including in-tree board support for testing)
- https://github.com/zephyrproject-rtos/zephyr/pull/100683 (not planned for v4.4)
- https://github.com/zephyrproject-rtos/zephyr/pull/104870

@de-nordic Please take a look

drivers/disk/ftl_dhara.c
```diff
@@ -0,0 +1,571 @@
...
+
+void dhara_nand_mark_bad(const struct dhara_nand *n, dhara_block_t b)
+{
+	struct disk_ftl_data *ctx = CONTAINER_OF(n, struct disk_ftl_data, dhara_nand);
+	off_t block_addr = b * ctx->block_size + ctx->area->fa_off;
+	int ret;
+
+	ret = flash_ex_op(ctx->info.dev, FLASH_EX_OP_MARK_BAD_BLOCK, (uintptr_t)&block_addr, NULL);
+	if (ret == -ENOTSUP) {
+		LOG_INF("marking bad block is not supported");
```
de-nordic: Wouldn't that pointlessly log this info on each attempt? It is not an error and actually user can not do anything about it, once device is in the field.
Shouldn't such attempt be considered an error?

tpambor: It is mostly useful during development/debugging, so I lowered it to `LOG_DBG` like the one in `dhara_nand_is_bad`. Dhara can work also if mark bad is not supported, e.g. when only performing wear-leveling for NOR flashes, so I think it is fine not to consider it as an error.

de-nordic (dismissed):

modules/fatfs/zephyr_fatfs_config.h
```diff
@@ -109,12 +109,13 @@
...
-	DT_FOREACH_STATUS_OKAY(zephyr_sdmmc_disk, _FF_DISK_NAME) \
-	DT_FOREACH_STATUS_OKAY(zephyr_mmc_disk, _FF_DISK_NAME) \
-	DT_FOREACH_STATUS_OKAY(st_stm32_sdmmc, _FF_DISK_NAME)
+#define FF_VOLUME_STRS                                                                             \
+	DT_FOREACH_STATUS_OKAY(zephyr_flash_disk, _FF_DISK_NAME)                                   \
+	DT_FOREACH_STATUS_OKAY(zephyr_ram_disk, _FF_DISK_NAME)                                     \
+	DT_FOREACH_STATUS_OKAY(zephyr_sdmmc_disk, _FF_DISK_NAME)                                   \
+	DT_FOREACH_STATUS_OKAY(zephyr_mmc_disk, _FF_DISK_NAME)                                     \
+	DT_FOREACH_STATUS_OKAY(st_stm32_sdmmc, _FF_DISK_NAME)                                      \
+	DT_FOREACH_STATUS_OKAY(zephyr_ftl_dhara, _FF_DISK_NAME)
```
jfischer-no, line 118: stray changes, have nothing to do with "modules: dhara: Implement Dhara disk driver as Zephyr module"

dts/bindings/disk/zephyr,ftl-dhara.yaml
```diff
@@ -0,0 +1,35 @@
...
+      Buffer size in bytes for the FTL driver. Must be equal or greater than the page size of the
+      backing flash.
+
+  gc-ratio:
+    type: int
+    default: 15
+    description: |
+      Garbage collection ratio for the FTL driver. This is the ratio of garbage collection
+      operations to real writes when automatic collection is active. Smaller values lead to faster
+      and more predictable input/output at the expense of capacity.
```
jfischer-no: Both properties sound like they are software configurations. Do they need to be configurable per instance, or can they be Kconfig options?

jfischer-no (changes requested):

dts/bindings/disk/zephyr,ftl-dhara.yaml
tpambor: `buffer-size` is somewhat hardware related as it has to be at least the size of a page of the flash. So, if one has multiple flashes with different page sizes and this is a KConfig one has to set it to the largest and would ovetallocate for the smaller flash. This was the motivation to have it in DT. `zephyr,flash-disk` has similar properties with `cache-size` and `sector-size`. I have no strong opinion here, if a KConfig is preferable, I can change it.

jfischer-no: What about gc-ratio does it make sense to be configurable per driver instance?

modules/fatfs/zephyr_fatfs_config.h
tpambor: I have split them off into separate commit.

dts/bindings/disk/zephyr,ftl-dhara.yaml
tpambor: I have switched `gc-ratio` to a KConfig. The GC ratio trades off write amplification for available capacity. The capacity for rewritable blocks is the number of pages times 1/(ratio+1), and the write amplification factor is (ratio+1). I don't think this needs to be configured per instance.

de-nordic (approved):
Looks good. Thanks.

jfischer-no (approved):

PR #100862, opened 2025-12-11
profiling: perf: delegate stack unwinding to arch_stack_walk

AndyLiuQian:
Implement stack trace function for arm64 arch.

support arm64 backtrace

### DIFF
--- a/subsys/profiling/perf/backends/CMakeLists.txt
+++ b/subsys/profiling/perf/backends/CMakeLists.txt
@@ -13,3 +13,7 @@ zephyr_sources_ifdef(CONFIG_PROFILING_PERF_BACKEND_X86
 zephyr_sources_ifdef(CONFIG_PROFILING_PERF_BACKEND_X86_64
   perf_x86_64.c
 )
+
+zephyr_sources_ifdef(CONFIG_PROFILING_PERF_BACKEND_ARM64
+  perf_arm64.c
+)
--- a/subsys/profiling/perf/backends/Kconfig
+++ b/subsys/profiling/perf/backends/Kconfig
@@ -12,8 +12,7 @@ config PROFILING_PERF_BACKEND_RISCV
 	bool
 	default y
 	depends on RISCV
-	depends on THREAD_STACK_INFO
-	depends on FRAME_POINTER
+	depends on ARCH_STACKWALK
 	select PROFILING_PERF_HAS_BACKEND
 
 config PROFILING_PERF_BACKEND_X86
@@ -30,4 +29,11 @@ config PROFILING_PERF_BACKEND_X86_64
 	depends on X86_64
 	depends on THREAD_STACK_INFO
 	depends on FRAME_POINTER
+	select PROFILING_PERF_HAS_BACKEND
+
+config PROFILING_PERF_BACKEND_ARM64
+	bool
+	default y
+	depends on ARM64
+	depends on ARCH_STACKWALK
 	select PROFILING_PERF_HAS_BACKEND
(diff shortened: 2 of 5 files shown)


nashif (dismissed):

AndyLiuQian (commented):
Copilot says CONFIG_FRAME_POINTER will be ignore by kconfig. I have already remove depend in Kconfig file.

nashif (dismissed):

npitre (changes requested):
This duplicates the unwinder that already exists in `arch/arm64/core/fatal.c:282` (`walk_stackframe()` / `arch_stack_walk()`, gated on `CONFIG_ARCH_STACKWALK`). The arch version uses the same `fp[0]=prev_fp, fp[1]=lr` pattern but with strictly better validation:

- alignment check on every fp,
- USERSPACE-aware bounds (privileged stack region for K_USER threads),
- `is_address_mapped()` check for demand-paged / MMU configs.

It also handles the IRQ-stack-to-thread-stack transition correctly when given an ESF, because `esf->fp` points into the interrupted thread's stack frames (saved by `vector_table.S` at exception entry). The walker never needs to traverse the IRQ stack itself — that's precisely what we *don't* want in a perf trace.

So the perf-specific bit reduces to recovering the ESF (the existing `_current_cpu->irq_stack - 16` trick is fine), then delegating. Sketch:

```c
struct ctx { uintptr_t *buf; size_t idx, size; };

static bool fill_cb(void *cookie, unsigned long ra, void *fp)
{
    struct ctx *c = cookie;
    if (c->idx >= c->size) return false;
    c->buf[c->idx++] = ra;
    return true;
}

size_t arch_perf_current_stack_trace(uintptr_t *buf, size_t size)
{
    if (size < 2U) return 0;
    const struct arch_esf *esf =
        (struct arch_esf *)*((uintptr_t *)((uintptr_t)_current_cpu->irq_stack - 16));
    struct ctx c = { buf, 0, size };
    buf[c.idx++] = esf->elr;     /* faulting PC */
    buf[c.idx++] = esf->lr;      /* covers prologue/epilogue case */
    arch_stack_walk(fill_cb, &c, _current, esf);
    return c.idx;
}
```

That's ~25 lines instead of 130, with stricter bounds checking, and any future fix to the unwinder automatically benefits perf. Kconfig would need `select ARCH_STACKWALK`.

Note: the same observation applies to the existing `perf_riscv.c`, which similarly reinvents `arch/riscv/core/stacktrace.c`. That should be reworked the same way, but in a separate PR.

Smaller things, separately:

1. Copyright header on `perf_arm64.c` is `Copyright (c) 2025` with no holder name. Compare to `Copyright (c) 2023 KNS Group LLC (YADRO)` on the siblings.

2. The `if (idx >= size) return 0;` in the loop discards a partial trace already written into `buf`. Should be `return idx;`. (Same defect exists in `perf_riscv.c` and `perf_x86_64.c`, so this is a broader cleanup; not specific to this PR.)

AndyLiuQian:
@npitre Sorry I’ve been busy recently. I totally get your idea now. I’ve figured out why x86 and 32-bit ARM can’t adopt this method, and adjusted the RISC-V & ARM64 code per your suggestions.

npitre (approved):
Non-blocking, for a follow-up PR: the unconditional esf->lr (esf->ra on riscv) emit is only correct for a leaf with no frame established yet. For a non-leaf sampled mid-body the live LR/RA points inside the current function and adds a spurious frame. Could be gated to the leaf/no-frame case later.

AndyLiuQian:
@nashif Hi, can this PR be merged into main?

wearyzen (approved):

PR #100863, opened 2025-12-11
drivers: rtc: rv3028: add calibration support

tpambor:
RV3028 RTC supports calibration to adjust for errors from +243.2 ppm to -244.1 ppm in steps of 0.9537 ppm.

Add support for the RTC calibration API by implementing `set_calibration` and `get_calibration` functions.

### DIFF
--- a/drivers/rtc/rtc_rv3028.c
+++ b/drivers/rtc/rtc_rv3028.c
@@ -83,9 +83,12 @@ LOG_MODULE_REGISTER(rv3028, CONFIG_RTC_LOG_LEVEL);
 
 #define RV3028_CLKOUT_FD_LOW            0x7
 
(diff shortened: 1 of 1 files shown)


bjarki-andreasen (approved):
Looks good, have you tried enabling CONFIG_RTC_CALIBRATION for the rtc api test suite? it dumps an array of set/get just to sanity check the conversion works in both directions and rounds properly.

Actually checking the calibration works could also be done, if you have 12 hours and an accurate clock source to use for reference if you really want to make sure it works (calibrate to lowest or highest offset, wait 12 hours, calculate the actual frequency by reading time using shell, use rtc_calibration_from_frequency to get the offset required to correct the clock, wait 12 hours, compare)

fabiobaltieri (approved):

PR #100872, opened 2025-12-11
boards: doc: remove tfm note for stm32u0 boards

lmapii:
The STM32U0 series MCUs don't support TF-M and therefore the note regarding the `build/tfm` directory needs to be removed from the corresponding documentation pages of the Nucleo boards using the STM32U0.

Sidenote: I couldn't get `west debug` working for neither of the U0 nucleo boards after flashing a Zephyr application (flashing works). It only runs without problems with the stock Nucleo firmware and/or after a chip-erase. Therefore, I couldn't really verify whether `west debug` works as advertised.


erwango (approved):
Thanks for this clean up.

mathieuchopstm (approved):

PR #100874, opened 2025-12-11
net: mqtt_sn: Disable support for GWINFO messages

M1cha:
The MQTT-SN v1.2 [0] specification specified the GwAddr as follows: The GwAdd field has a variable length and contains the address of a GW. Its depends on the network over which MQTT-SN operates and is indicated in the first octet of this field. For example, in a ZigBee network the network address is 2-octet long.

It specifies neither the possible values for the first octet, nor the format of the network-specific address that follows. Thus, the specification is incomplete and this functionality unusable.

I also wasn't able to find any implementation which implements a gwinfo message where the address length is not zero. This includes https://github.com/eclipse-paho/paho.mqtt-sn.embedded-c .

The current implementation in Zephyr simply copies a `struct sockaddr` into GwAddr. This is a bad idea for many reasons, the most important one being that the format is not even specified by POSIX [1]. They only say, which defines must exist, not what their values are. And in fact, these even differ between Zephyr and Linux.

Thus, I think it's best to disable the implementation without fully removing it, to prevent people from using this, which may even lead to memory safety issues, depending on the length of
CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE. If we were to receive an updated specification, all we'd have to do is to convert between `struct sockaddr` and mqtt-sn addresses both ways.

[0] https://groups.oasis-open.org/higherlogic/ws/public/download/66091/MQTT-SN_spec_v1.2.pdf
[1] https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_socket.h.html

### DIFF
--- a/subsys/net/lib/mqtt_sn/mqtt_sn.c
+++ b/subsys/net/lib/mqtt_sn/mqtt_sn.c
@@ -596,32 +596,6 @@ static void mqtt_sn_do_searchgw(struct mqtt_sn_client *client)
 }
 
 /**
- * @brief Internal function to send a GWINFO message.
- *
- * @param client
- */
-static void mqtt_sn_do_gwinfo(struct mqtt_sn_client *client)
-{
-	struct mqtt_sn_param response = {.type = MQTT_SN_MSG_TYPE_GWINFO};
-	struct mqtt_sn_gateway *gw;
-	struct mqtt_sn_data addr;
-
-	gw = SYS_SLIST_PEEK_HEAD_CONTAINER(&client->gateway, gw, next);
-
-	if (gw == NULL || gw->addr_len == 0) {
-		LOG_WRN("No Gateway Address");
-		return;
-	}
-
-	response.params.gwinfo.gw_id = gw->gw_id;
-	addr.data = gw->addr;
-	addr.size = gw->addr_len;
-	response.params.gwinfo.gw_add = addr;
-
-	encode_and_send(client, &response, client->radius_gwinfo);
-}
-
-/**
  * @brief Internal function to send a PINGREQ message.
  *
  * @param client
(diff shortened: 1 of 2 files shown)


subsys/net/lib/mqtt_sn/mqtt_sn.c
```diff
@@ -613,6 +613,9 @@ static void mqtt_sn_do_gwinfo(struct mqtt_sn_client *client)
 		return;
 	}
 
+	LOG_WRN("GwAddr is not specified properly. Ignoring SEARCHGW message");
+	return;
```
pdgendt: Code after this statement is unreachable, this does not appear correct.

pdgendt (dismissed):

subsys/net/lib/mqtt_sn/mqtt_sn.c
M1cha: This is the intention: 
> Thus, I think it's best to disable the implementation without fully removing it, to prevent people from using this, which may even lead to memory safety issues, depending on the length of
CONFIG_MQTT_SN_LIB_MAX_ADDR_SIZE. If we were to receive an updated specification, all we'd have to do is to convert between struct sockaddr and mqtt-sn addresses both ways.

Would you like me to remove all of that code instead?

pdgendt: I'd prefer to wrap it in a `#if 0 .... #endif` block if we want to keep it, but more importantly it should have a comment with the rationale for future readers.

M1cha: Unfortunately, an `#if 0` causes a checkpatch warning, so I opted to remove the code and add a comment with the rationale.

rlubos (approved):

aurel32 (approved):
Thanks, sounds good

jukkar (approved):

PR #100875, opened 2025-12-11
drivers: crc: add NXP CRC driver support

Holt-Sun:
This PR adds driver support for the NXP CRC (Cyclic Redundancy Check) peripheral found on MCX series microcontrollers, along with device tree bindings, board configuration, and updated samples and tests.

### DIFF
--- a/drivers/crc/CMakeLists.txt
+++ b/drivers/crc/CMakeLists.txt
@@ -1,4 +1,5 @@
 # Copyright (c) 2025 Renesas Electronics Corporation
+# Copyright 2025 NXP
 # SPDX-License-Identifier: Apache-2.0
 
 zephyr_syscall_header(${ZEPHYR_BASE}/include/zephyr/drivers/crc.h)
@@ -6,6 +7,7 @@ zephyr_syscall_header(${ZEPHYR_BASE}/include/zephyr/drivers/crc.h)
 zephyr_library()
 
 # zephyr-keep-sorted-start
+zephyr_library_sources_ifdef(CONFIG_CRC_DRIVER_NXP crc_nxp.c)
 zephyr_library_sources_ifdef(CONFIG_CRC_DRIVER_RENESAS_RA crc_renesas_ra.c)
 zephyr_library_sources_ifdef(CONFIG_CRC_DRIVER_SF32LB crc_sf32lb.c)
 # zephyr-keep-sorted-stop
--- a/drivers/crc/crc_nxp.c
+++ b/drivers/crc/crc_nxp.c
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2025 NXP
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT nxp_crc
+
+#include <zephyr/drivers/crc.h>
+#include <zephyr/sys/util.h>
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(nxp_crc, CONFIG_CRC_LOG_LEVEL);
+
+/* Namespace collision process: there are crc_result_t defined in both mcux and Zephyr */
+#define crc_result_t mcux_crc_result_t
+#include "fsl_crc.h"
+#undef crc_result_t
+
+struct crc_nxp_config {
+	CRC_Type *base;
+};
+
+struct crc_nxp_data {
+	struct k_sem lock;
+};
+
+static inline void crc_nxp_lock(const struct device *dev)
+{
+	struct crc_nxp_data *data = dev->data;
+
+	k_sem_take(&data->lock, K_FOREVER);
+}
+
+static inline void crc_nxp_unlock(const struct device *dev)
+{
+	struct crc_nxp_data *data = dev->data;
+
+	k_sem_give(&data->lock);
+}
+
+static int crc_nxp_prepare_config(const struct device *dev, const struct crc_ctx *ctx,
+				  crc_config_t *cfg, bool *use32)
+{
+	const struct crc_nxp_config *config = dev->config;
+
+	ARG_UNUSED(config);
+
+	if ((ctx == NULL) || (cfg == NULL)) {
+		return -EINVAL;
+	}
+
+	cfg->polynomial = ctx->polynomial;
+	cfg->seed = ctx->seed;
+	cfg->reflectIn = (ctx->reversed & CRC_FLAG_REVERSE_INPUT) != 0U;
+	cfg->reflectOut = (ctx->reversed & CRC_FLAG_REVERSE_OUTPUT) != 0U;
+	cfg->complementChecksum = false;
+	cfg->crcResult = kCrcFinalChecksum;
+
+	switch (ctx->type) {
+	case CRC16:
+		if (ctx->polynomial != CRC16_POLY) {
+			return -EINVAL;
+		}
+		cfg->seed &= 0xFFFFU;
+		cfg->crcBits = kCrcBits16;
+		*use32 = false;
+		break;
+	case CRC16_CCITT:
+		if (ctx->polynomial != CRC16_CCITT_POLY) {
+			return -EINVAL;
+		}
+		cfg->seed &= 0xFFFFU;
+		cfg->crcBits = kCrcBits16;
+		*use32 = false;
+		break;
+	case CRC16_ANSI:
+		if (ctx->polynomial != CRC16_REFLECT_POLY) {
+			return -EINVAL;
+		}
+		cfg->seed &= 0xFFFFU;
+		cfg->crcBits = kCrcBits16;
+		*use32 = false;
+		break;
+	case CRC16_ITU_T:
+		if (ctx->polynomial != CRC16_CCITT_POLY) {
+			return -EINVAL;
+		}
+		cfg->seed &= 0xFFFFU;
+		cfg->crcBits = kCrcBits16;
+		*use32 = false;
+		break;
+	case CRC32_C:
+		if (ctx->polynomial != CRC32C_POLY) {
+			return -EINVAL;
+		}
+		cfg->crcBits = kCrcBits32;
+		*use32 = true;
+		break;
+	case CRC32_IEEE:
+		if (ctx->polynomial != CRC32_IEEE_POLY) {
+			return -EINVAL;
+		}
+		cfg->crcBits = kCrcBits32;
+		cfg->complementChecksum = true; /* IEEE requires final XOR */
+		*use32 = true;
+		break;
+	default:
+		return -ENOTSUP;
+	}
+
+	return 0;
+}
+
+static int crc_nxp_begin(const struct device *dev, struct crc_ctx *ctx)
+{
+	const struct crc_nxp_config *config = dev->config;
+	crc_config_t cfg;
+	bool use32 = false;
+	int ret;
+
+	if ((ctx == NULL) || (ctx->state != CRC_STATE_IDLE)) {
+		return -EINVAL;
+	}
+
+	crc_nxp_lock(dev);
+
+	ret = crc_nxp_prepare_config(dev, ctx, &cfg, &use32);
+	if (ret != 0) {
+		crc_nxp_unlock(dev);
+		return ret;
+	}
+
+	/* Initialize hardware with protocol settings and seed */
+	CRC_Init(config->base, &cfg);
+
+	ctx->state = CRC_STATE_IN_PROGRESS;
+
+	return 0;
+}
+
+static int crc_nxp_update(const struct device *dev, struct crc_ctx *ctx, const void *buffer,
+			  size_t bufsize)
+{
+	const struct crc_nxp_config *config = dev->config;
+
+	if (ctx->state != CRC_STATE_IN_PROGRESS) {
+		return -EINVAL;
+	}
+
+	/* Allow zero-length updates */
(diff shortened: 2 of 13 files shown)


butok:
Thank you @Holt-Sun for filling the NXP CRC gap.
 Any plan to add a CRC support for the rest of the NXP platforms?

tests/drivers/crc/src/main.c
```diff
@@ -75,13 +77,16 @@ ZTEST(crc, test_crc_8)
...
  * @brief Test that crc16 works
  */
+#ifdef CONFIG_CRC_DRIVER_HAS_CRC16
+
+/* Define result of CRC computation */
+#define RESULT_CRC_16 0xD543
+
 ZTEST(crc, test_crc_16)
 {
 	static const struct device *dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_crc));
```
thenguyenyf: If driver doesn't support this CRC variant, skip this test instead of excluding it from build
```suggestion
ZTEST(crc, test_crc_16)
{
	static const struct device *dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_crc));
	
#ifndef CONFIG_CRC_DRIVER_HAS_CRC16
	ztest_test_skip();
#endif	
```

thenguyenyf: Please update other test cases

tests/drivers/crc/src/main.c
```diff
@@ -268,22 +288,48 @@ ZTEST(crc, test_discontinuous_buf)
...
+		.reversed = CRC_FLAG_REVERSE_INPUT | CRC_FLAG_REVERSE_OUTPUT,
+	};
+#else
+	ctx = (struct crc_ctx){
+		.type = CRC32_C,
+		.polynomial = CRC32C_POLY,
+		.seed = CRC32_C_INIT_VAL,
+		.reversed = CRC_FLAG_REVERSE_INPUT | CRC_FLAG_REVERSE_OUTPUT,
+	};
+#endif
```
thenguyenyf: The update below could reduce the number of code changes and improve readability.
```c
#ifdef CONFIG_CRC_DRIVER_HAS_CRC8

#define CRC_TEST_VARIANT CRC8
#define CRC_TEST_INIT_VAL ...
#define ...

#elif defined(CONFIG_CRC_DRIVER_HAS_CRC16)

#define CRC_TEST_VARIANT CRC16
#define CRC_TEST_INIT_VAL ...
#define ...

#else

#define CRC_TEST_VARIANT CRC32_C
#define CRC_TEST_INIT_VAL ...
#define ...

#endif
```
```suggestion
	ctx = (struct crc_ctx){
		.type = CRC_TEST_VARIANT,
		.polynomial = CRC_TEST_POLY,
		.seed = CRC_TEST_INIT_VAL,
		.reversed = CRC_TEST_REVERSE_CONFIG,
	};
```

thenguyenyf: Please update this for any other similar cases.

drivers/crc/crc_nxp.c
```diff
@@ -0,0 +1,217 @@
...
+		CRC_WriteData(config->base, (const uint8_t *)buffer, bufsize);
+	}
+
+	/* Keep an updated result for streaming verification */
+	if ((ctx->type == CRC32_C) || (ctx->type == CRC32_IEEE)) {
+		ctx->result = CRC_Get32bitResult(config->base);
+	} else {
+		uint16_t r16 = CRC_Get16bitResult(config->base);
+
+		ctx->result = (uint32_t)r16;
```
thenguyenyf: Is there any special reason to use the secondary variable `r16` instead of assigning directly to `ctx->result`?

samples/drivers/crc/README.rst
```diff
@@ -8,6 +8,17 @@ Overview
...
+
+- If the driver enables ``CONFIG_CRC_DRIVER_HAS_CRC8``, the sample uses CRC8.
+- Else if ``CONFIG_CRC_DRIVER_HAS_CRC16`` is enabled, it uses CRC16-CCITT.
+- Else if ``CONFIG_CRC_DRIVER_HAS_CRC32_IEEE`` is enabled, it uses CRC32-IEEE.
+- Else if ``CONFIG_CRC_DRIVER_HAS_CRC32_C`` is enabled, it uses CRC32-C.
+
+Optional expected values can be provided via macros to verify results:
+``EXPECTED_CRC8``, ``EXPECTED_CRC16_CCITT``, ``EXPECTED_CRC32_IEEE``, ``EXPECTED_CRC32_C``.
+If no expected macro is defined for the selected protocol, the sample logs
+the computed CRC value.
```
thenguyenyf: In my opinion, this information is unnecessary in a sample note

samples/drivers/crc/README.rst
```diff
@@ -16,29 +27,45 @@ Building and Running for Renesas RA8M1
 ======================================
 
 The sample can be built and executed for the
-:zephyr:board:`ek_ra8m1` as follows:
+:zephyr:board:`ek_ra8m1` and :zephyr:board:`frdm_mcxe247` as follows:
```
thenguyenyf: Just 1 platform need to add for example in this section. There is no additional param or steps to build your board, please revert it.

samples/drivers/crc/README.rst
```diff
@@ -16,29 +27,45 @@ Building and Running for Renesas RA8M1
...
    :goals: build flash
    :compact:
+.. zephyr-app-commands::
+   :zephyr-app: samples/drivers/crc
+   :board: frdm_mcxe247
+   :goals: build flash
+   :compact:
 
-To build for another board, change "ek_ra8m1" above to that board's name.
+To build for another board, change "ek_ra8m1" or "frdm_mcxe247" above to that board's name.
```
thenguyenyf: Revert it too

samples/drivers/crc/src/main.c
thenguyenyf, line 1: The purpose of a sample is to demonstrate how to use a driver, so it should be implemented as simply as possible. You don't need to include all the configuration options in these samples (this should be done in a zTest instead). So please:
- Keep the call flow as clear as possible. Move all preprocessing outside the main function.
- Add `Kconfig` to allow flexible selection of the supported CRC variant to run with your platform. Refer to https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/can/babbling/Kconfig for more information.
- Revert the main function to calculate only one CRC variant. The console log could be updated to provide more information about the calculation, such as the CRC variant and a comparison of the result and the expected value.

samples/drivers/crc/README.rst
```diff
@@ -16,29 +27,45 @@ Building and Running for Renesas RA8M1
...
 .. code-block:: console
 
-   crc_example: CRC verification succeed.
+   crc_example: CRC8 verification succeeded
+   # or
+   crc_example: CRC16-CCITT verification succeeded
+   # or
+   crc_example: CRC32-IEEE verification succeeded
+   # or
+   crc_example: CRC result: 0x445c
```
thenguyenyf: Only keep one generic log

```suggestion
   crc_example: CRC result: 0x445c
   crc_example: CRC16-CCITT verification succeeded
```

samples/drivers/crc/README.rst
```diff
@@ -16,29 +27,45 @@ Building and Running for Renesas RA8M1
...
 
-.. note:: If the CRC is not supported, the output will be an error message.
+.. note:: If no supported CRC protocol is available on the selected board,
+   the sample will log an error.
```
thenguyenyf:
```suggestion
```

thenguyenyf (changes requested):

samples/drivers/crc/src/main.c
Holt-Sun: 1. Add Kconfig for sample crc
2. Update sample crc main.c as commentd.

drivers/crc/crc_nxp.c
Holt-Sun: Good catch. I think it is perhaps because when I write code I thought here needs a (uint32_t).
Updated to `ctx->result = (uint32_t)CRC_Get16bitResult(config->base);`

tests/drivers/crc/src/main.c
```diff
@@ -45,17 +45,22 @@ static void wait_thread_entry(void *a, void *b, void *c)
...
  * @brief Test that crc8 works
  */
+/* Define result of CRC computation */
+#define RESULT_CRC_8 0xB2
+
 ZTEST(crc, test_crc_8)
 {
+#ifndef CONFIG_CRC_DRIVER_HAS_CRC8
+	ztest_test_skip();
+	return;
```
thenguyenyf:
```suggestion
	ztest_test_skip();
```

thenguyenyf: Similar cases should be updated

tests/drivers/crc/src/main.c
```diff
@@ -268,22 +337,35 @@ ZTEST(crc, test_discontinuous_buf)
...
-		.polynomial = CRC8_POLY,
-		.seed = CRC8_INIT_VAL,
-		.reversed = CRC_FLAG_REVERSE_INPUT | CRC_FLAG_REVERSE_OUTPUT,
+	struct crc_ctx ctx;
+
+	ctx = (struct crc_ctx){
+		.type = CRC_TEST_VARIANT,
+		.polynomial = CRC_TEST_POLY,
+		.seed = CRC_TEST_INIT_VAL,
+		.reversed = CRC_TEST_REVERSE_CONFIG,
```
thenguyenyf:
```suggestion
	struct crc_ctx ctx = {
		.type = CRC_TEST_VARIANT,
		.polynomial = CRC_TEST_POLY,
		.seed = CRC_TEST_INIT_VAL,
		.reversed = CRC_TEST_REVERSE_CONFIG,
```

thenguyenyf: Similar cases should be updated

tests/drivers/crc/src/main.c
```diff
@@ -268,22 +337,35 @@ ZTEST(crc, test_discontinuous_buf)
...
 
 	zassert_equal(crc_begin(dev, &ctx), 0);
 	zassert_equal(crc_update(dev, &ctx, data1, sizeof(data1)), 0);
 	zassert_equal(crc_update(dev, &ctx, data2, sizeof(data2)), 0);
 	zassert_equal(crc_finish(dev, &ctx), 0);
-	zassert_equal(crc_verify(&ctx, RESULT_DISCONTINUOUS_BUFFER), 0);
+	zassert_equal(crc_verify(&ctx, CRC_TEST_EXPECTED), 0);
 }
 
+#undef CRC_TEST_EXPECTED
```
thenguyenyf: I don't recommend using this method. It's better to define an exclusive test result macro for each checkpoint.

drivers/crc/CMakeLists.txt
```diff
@@ -1,11 +1,13 @@
 # Copyright (c) 2025 Renesas Electronics Corporation
+# Copyright 2025 NXP
```
thenguyenyf, line 2: non-blocking: note that you don't need to add your copyright to each global file when adding vendor inclusion

tests/drivers/crc/src/main.c
Holt-Sun: Updated.

thenguyenyf (approved):

ZhaoxiangJin (approved):

drivers/crc/CMakeLists.txt
```diff
@@ -1,11 +1,13 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
 zephyr_syscall_header(${ZEPHYR_BASE}/include/zephyr/drivers/crc.h)
 
 zephyr_library()
 
 # zephyr-keep-sorted-start
+zephyr_library_sources_ifdef(CONFIG_CRC_DRIVER_NXP crc_nxp.c)
```
zejiang0jason, line 10: there are several types of crc in NXP chips, need consider how to name them.

samples/drivers/crc/src/main.c
```diff
@@ -31,55 +76,42 @@ int main(void)
...
-		}
-	} else { /* Reversed is no reversed output */
-		ret = crc_verify(&ctx, 0xB2);
+	LOG_INF("%s result: 0x%08x", CRC_SAMPLE_NAME, (unsigned int)ctx.result);
 
-		if (ret != 0) {
-			LOG_ERR("CRC verification failed: %d", ret);
-			return ret;
-		}
+#if defined(CRC_SAMPLE_EXPECTED) && (CRC_SAMPLE_EXPECTED != 0)
```
zejiang0jason, line 105: In which case the result doesn't need be validated?

samples/drivers/crc/src/main.c
thenguyenyf, line 1: @Holt-Sun . I detect that a twister test was failed due to the update from this PR. The regex in this file https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/crc/sample.yaml has not been updated.

Could you send a subsequent PR to fix it?

kartben: Also please add at least one integration platform (use what is mentioned in the readme) so that the test runs on PRs :)

Holt-Sun: Please review https://github.com/zephyrproject-rtos/zephyr/pull/101523

PR #100876, opened 2025-12-11
usb: device_next: remove udc_buf_get_all()

tmon-nordic:
Function udc_buf_get_all() was intended to be a helper to remove all requests from endpoint FIFO. While for just freeing all queue the it may be argued that there may be some doubtful simplicity argument, merging multiple submitted transfers into one is just enforcing unnecessary complexity on class implementations.

At general level, every submitted (enqueued) request should get corresponding completion (request callback) call. UDC drivers were violating this sensible behavior when dequeuing (cancelling) requests by merging all submitted requests into one.

Remove udc_buf_get_all() and replace all uses with simple loops. For most classes (that submit just one request for an endpoint at a time) this has no functional difference. For classes that implement double buffering this simplifies completion handling.

---

Related discussion: https://github.com/zephyrproject-rtos/zephyr/discussions/99439

### DIFF
--- a/drivers/usb/udc/udc_ambiq.c
+++ b/drivers/usb/udc/udc_ambiq.c
@@ -258,14 +258,10 @@ static int udc_ambiq_ep_dequeue(const struct device *dev, struct udc_ep_config *
 {
 	unsigned int lock_key;
 	struct udc_ambiq_data *priv = udc_get_private(dev);
-	struct net_buf *buf;
 
 	lock_key = irq_lock();
 
-	buf = udc_buf_get_all(ep_cfg);
-	if (buf) {
-		udc_submit_ep_event(dev, buf, -ECONNABORTED);
-	}
+	udc_ep_cancel_queued(dev, ep_cfg);
 
 	udc_ep_set_busy(ep_cfg, false);
 	am_hal_usb_ep_state_reset(priv->usb_handle, ep_cfg->addr);
--- a/drivers/usb/udc/udc_common.c
+++ b/drivers/usb/udc/udc_common.c
@@ -81,27 +81,6 @@ struct net_buf *udc_buf_get(struct udc_ep_config *const ep_cfg)
 	return k_fifo_get(&ep_cfg->fifo, K_NO_WAIT);
 }
 
-struct net_buf *udc_buf_get_all(struct udc_ep_config *const ep_cfg)
-{
-	struct net_buf *buf;
-
-	buf = k_fifo_get(&ep_cfg->fifo, K_NO_WAIT);
-	if (!buf) {
-		return NULL;
-	}
-
-	LOG_DBG("ep 0x%02x dequeue %p", ep_cfg->addr, buf);
-	for (struct net_buf *n = buf; !k_fifo_is_empty(&ep_cfg->fifo); n = n->frags) {
-		n->frags = k_fifo_get(&ep_cfg->fifo, K_NO_WAIT);
-		LOG_DBG("|-> %p ", n->frags);
-		if (n->frags == NULL) {
-			break;
-		}
-	}
-
-	return buf;
-}
-
 struct net_buf *udc_buf_peek(struct udc_ep_config *const ep_cfg)
 {
 	return k_fifo_peek_head(&ep_cfg->fifo);
@@ -125,6 +104,15 @@ void udc_ep_buf_clear_zlp(const struct net_buf *const buf)
 	struct udc_buf_info *bi = udc_get_buf_info(buf);
 
 	bi->zlp = false;
+}
+
+void udc_ep_cancel_queued(const struct device *dev, struct udc_ep_config *const cfg)
+{
+	struct net_buf *buf;
+
+	for (buf = udc_buf_get(cfg); buf; buf = udc_buf_get(cfg)) {
+		udc_submit_ep_event(dev, buf, -ECONNABORTED);
+	}
 }
 
 void udc_setup_received(const struct device *dev, const void *const setup)
--- a/drivers/usb/udc/udc_bflb_v1.c
+++ b/drivers/usb/udc/udc_bflb_v1.c
@@ -1350,10 +1350,7 @@ static int udc_bflb_v1_ep_dequeue(const struct device *dev,
 		fifo_rx_clear(base, ep_idx);
 	}
 
-	buf = udc_buf_get_all(ep_cfg);
-	if (buf != NULL) {
-		udc_submit_ep_event(dev, buf, -ECONNABORTED);
-	}
+	udc_ep_cancel_queued(dev, ep_cfg);
 
 	/* Clear pending completion signal to prevent stale processing */
 	atomic_and(&priv->xfer_finished, ~BIT(ep_to_bnum(ep_cfg->addr)));
--- a/drivers/usb/udc/udc_common.h
+++ b/drivers/usb/udc/udc_common.h
@@ -82,19 +82,6 @@ void udc_ep_set_busy(struct udc_ep_config *const ep_cfg,
 struct net_buf *udc_buf_get(struct udc_ep_config *const ep_cfg);
 
 /**
- * @brief Get all UDC request from endpoint FIFO.
- *
- * Get all UDC request from endpoint FIFO as single-linked list.
- * This function removes all request from endpoint FIFO and
- * is typically used to dequeue endpoint FIFO.
- *
- * @param[in] ep_cfg Pointer to endpoint configuration
- *
- * @return pointer to UDC request or NULL on error.
- */
-struct net_buf *udc_buf_get_all(struct udc_ep_config *const ep_cfg);
-
-/**
  * @brief Peek request at the head of endpoint FIFO.
  *
  * Return request from the head of endpoint FIFO without removing.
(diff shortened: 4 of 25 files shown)


mathieuchopstm (commented):
Seems fair. +1 for STM32 driver.

As there appears to be two main patterns repeating, it may be sensible to provide helper functions for them though, e.g.:
```c
static void udc_buf_unref_all(struct udc_ep_config *cfg)
{
  struct net_buf *buf;

  while ((buf = udc_buf_get(cfg))) {
    net_buf_unref(buf);
  }
}

/* this name is pretty bad but i'm out of ideas */
static void udc_drain_queue_with_ep_event(const struct device *dev, struct udc_ep_config *cfg, int status) {
  struct net_buf *buf;

  while ((buf = udc_buf_get(cfg))) {
    udc_submit_ep_event(dev, buf, status);
  }
}
```

josuah (commented):
It seems like simplifying the logic at a few places, before buffers are listed as a tree (FIFO list + buf->frag), now buffers are listed in the FIFO list only, so that reduces special cases...

To check that conversion is complete:
- all calls to the function removed (guaranteed by the compiler) ✔️ 
- `buf->frags` occurences... it looks like control transfers were not modified, I assume on purpose as it's an needed use of `->frags`?

tmon-nordic:
>    * `buf->frags` occurences... it looks like control transfers were not modified, I assume on purpose as it's an needed use of `->frags`?

Control transfer manually link data stage to setup data. This frags usage is different from the `udc_buf_get_all()` and is not modified by this PR.

josuah (dismissed):
Thanks for confirming. I do not see anything missing, though I did not test it to say.

MarkWangChinese (dismissed):

josuah (dismissed):

mathieuchopstm (dismissed):
STM32 changes LGTM

drivers/usb/udc/udc_ambiq.c
```diff
@@ -296,8 +296,7 @@ static int udc_ambiq_ep_dequeue(const struct device *dev, struct udc_ep_config *
 
 	lock_key = irq_lock();
 
-	buf = udc_buf_get_all(ep_cfg);
-	if (buf) {
+	while ((buf = udc_buf_get(ep_cfg))) {
```
fabiobaltieri: sorry, late comment, isn't there an extra set of parenthesis? you can just do

```c
	while (buf = udc_buf_get(ep_cfg)) {
```

tmon-nordic: If you do `while (buf = udc_buf_get(ep_cfg)) {` then you get warning from gcc about assignment inside `while ()` which is solved by adding extra parenthesis.

drivers/usb/udc/udc_common.c
```diff
@@ -111,27 +111,6 @@ struct net_buf *udc_buf_get(struct udc_ep_config *const ep_cfg)
...
-		n->frags = k_fifo_get(&ep_cfg->fifo, K_NO_WAIT);
-		LOG_DBG("|-> %p ", n->frags);
-		if (n->frags == NULL) {
-			break;
-		}
-	}
-
-	return buf;
-}
-
```
jfischer-no, line 104: IIRC it should be reworked after 3d306c181f0d75c5765f80fdfbee614e1e3f191d, but yes maybe it is better to remove it.

It looks like there should be helpers for the two common patterns to flush the queue and cancel all transfers.

```diff
-	buf = udc_buf_get_all(cfg);
-	if (buf) {
+	while ((buf = udc_buf_get(cfg))) {
		udc_submit_ep_event(dev, buf, -ECONNABORTED);
	}
```

```diff
-	buf = udc_buf_get_all(cfg);
-	if (buf) {
+	while ((buf = udc_buf_get(cfg_out))) {
		net_buf_unref(buf);
	}
```

`while ((buf = udc_buf_get(cfg_out))) ` is a bit problematic as it is foreseeable that people will get confused. Also, I do not want someone to get triggered, for IMO valid reasons here, by the rule "The result of an assignment operator should not be used". Please add helpers like:


```c
/**
 * @brief Unref all queued UDC request
 *
 * Remove all the request from the endpoint FIFO and unref them.
 *
 * @param[in] cfg Pointer to endpoint configuration
 */
void udc_ep_unref_queued(struct udc_ep_config *const cfg);

void udc_ep_unref_queued(struct udc_ep_config *const cfg)
{
	struct net_buf *buf;

	buf = udc_buf_get(cfg)
	while (buf != NULL) {
		net_buf_unref(buf);
		buf = udc_buf_get(cfg)
	}
}
```

```c
/**
 * @brief Cancel all queued UDC request
 *
 * Remove all the request from the endpoint FIFO and submit
 * them to higher level.
 *
 * @param[in] cfg Pointer to endpoint configuration
 */
void udc_ep_cancel_queued(struct udc_ep_config *const cfg);

void udc_ep_cancel_queued(struct udc_ep_config *const cfg)
{
	struct net_buf *buf;

	buf = udc_buf_get(cfg)
	while (buf != NULL) {
		udc_submit_ep_event(dev, buf, -ECONNABORTED);
		buf = udc_buf_get(cfg)
	}
}
```

drivers/usb/udc/udc_ambiq.c
fabiobaltieri: oh, it thinks you meant to do `==`, right... ok thanks for clarifying

fabiobaltieri (dismissed):

jfischer-no (changes requested):

drivers/usb/udc/udc_common.c
tmon-nordic: > while ((buf = udc_buf_get(cfg_out))) is a bit problematic as it is foreseeable that people will get confused

The `while ((` is rather common in Zephyr codebase, for example:
https://github.com/zephyrproject-rtos/zephyr/blob/37eb5ad0318a55bbb83c712151278de9b6145c73/subsys/bluetooth/host/iso.c#L469
https://github.com/zephyrproject-rtos/zephyr/blob/37eb5ad0318a55bbb83c712151278de9b6145c73/subsys/bluetooth/host/l2cap.c#L2399-L2406
https://github.com/zephyrproject-rtos/zephyr/blob/37eb5ad0318a55bbb83c712151278de9b6145c73/subsys/bluetooth/host/l2cap.c#L281

> void udc_ep_unref_queued(struct udc_ep_config *const cfg)

This sounds like a bad advice to me. UDC should not be freeing queued buffers on its own, but rather it should be completing them. The only reason why the requests are currently freed by UDC driver is the broken design for how control transfers are handled. Hopefully this is resolved by #103493

> void udc_ep_cancel_queued(struct udc_ep_config *const cfg)

I don't see the point. Maybe if the reason was provided as parameter (instead of hardcoding `-ECONNABORTED`) there would be some merit to it.

Would you revise your -1 given above argumentation?

MarkWangChinese (dismissed):

drivers/usb/udc/udc_common.c
jfischer-no: >The while (( is rather common in Zephyr codebase, for example:

It is up to Bluetooth to fix "The result of an assignment operator should not be used".

>Would you revise your -1 given above argumentation?

No, after https://github.com/zephyrproject-rtos/zephyr/pull/103493 is merged, there is still at common pattern in your changes
```diff
-	buf = udc_buf_get_all(cfg);
-	if (buf) {
+	while ((buf = udc_buf_get(cfg))) {
		udc_submit_ep_event(dev, buf, -ECONNABORTED);
	}
```

and that should be moved to something like

```c
/**
 * @brief Cancel all queued UDC request
 *
 * Remove all the request from the endpoint FIFO and submit
 * them to higher level.
 *
 * @param[in] cfg Pointer to endpoint configuration
 */
void udc_ep_cancel_queued(struct udc_ep_config *const cfg);

void udc_ep_cancel_queued(struct udc_ep_config *const cfg)
{
	struct net_buf *buf;

	buf = udc_buf_get(cfg)
	while (buf != NULL) {
		udc_submit_ep_event(dev, buf, -ECONNABORTED);
		buf = udc_buf_get(cfg)
	}
}
```

> This sounds like a bad advice to me. UDC should not be freeing queued buffers on its own, but rather it should be completing them. The only reason why the requests are currently freed by UDC driver is the broken design for how control transfers are handled. Hopefully this is resolved by https://github.com/zephyrproject-rtos/zephyr/pull/103493

One still can find this pattern here:
https://github.com/zephyrproject-rtos/zephyr/pull/100876/commits/3a619d9dcfab95641b48a53c0412ce0d29e4a4dc#diff-9636542b03fbd14ba20898a4f5562fc753535a02a6547cb9ac3e346b9a2b6e2dR2182-R2184
and here
https://github.com/zephyrproject-rtos/zephyr/blob/ee0cc5a620500937481ade44e993b87b17736191/drivers/usb/udc/udc_common.c#L626-L628
https://github.com/zephyrproject-rtos/zephyr/blob/ee0cc5a620500937481ade44e993b87b17736191/drivers/usb/udc/udc_common.c#L633-L635

Looks like it still makes sense to have helper function `void udc_ep_unref_queued()`.

tmon-nordic: > It is up to Bluetooth to fix "The result of an assignment operator should not be used".

Please open bug report so they are aware that this is an issue. That having said, I am fairly certain it is not issue and you are completely unreasonably forcing people to use different constructs.

I added `void udc_ep_cancel_queued(const struct device *dev, struct udc_ep_config *const cfg);`

> Looks like it still makes sense to have helper function void udc_ep_unref_queued().

I prefer not to. These should eventually get removed, but it should be considered out-of-scope here.

josuah (dismissed):
`usb: device_next: Allocate control buffers in stack` (244b015e99d85e56f10ed2c7a7eb349590ea151b) already removed all other uses of `udc_buf_get_all()` as a side effect.

josuah (approved):

jfischer-no (approved):

PR #100880, opened 2025-12-11
TF-M Support for NXP mcxn947

waqar-tahir:
TF-M Support is added for NXP mcxn947 platform.
The changes also include the TF-M logging updates from upstream tf-m repo as well as upstream tf-m-tests repo.
Currently, tf-m regression and tfm-ipc tests are enabled in samples for frdmmcxn947 board.


modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -264,6 +265,18 @@ if(CONFIG_BUILD_WITH_TFM)
     list(APPEND TFM_CMAKE_ARGS -DHAL_ADI_PATH=${ZEPHYR_ADI_MODULE_DIR})
   endif()
 
+ if(CONFIG_BOARD_FRDM_MCNX947 OR CONFIG_BOARD_FRDM_MCXN947)
```
tomi-font: `CONFIG_BOARD_FRDM_MCNX947` doesn't exist, remove?

modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -264,6 +265,18 @@ if(CONFIG_BUILD_WITH_TFM)
...
+    # Supply path to NXP HAL sources used for TF-M build
+    list(APPEND TFM_CMAKE_ARGS -DZEPHYR_HAL_NXP_MODULE_DIR=${ZEPHYR_HAL_NXP_MODULE_DIR})
+    # Supply path to NXP HAL sources used for TF-M build
+    set(TFM_PLATFORM_NXP_HAL_FILE_PATH ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/platform/ext/target/nxp/)
+    list(APPEND TFM_CMAKE_ARGS -DTFM_PLATFORM_NXP_HAL_FILE_PATH=${TFM_PLATFORM_NXP_HAL_FILE_PATH})
+    # Disable watchdog support on this platform for tfm_psa_tests
```
tomi-font: Saying why would be more important

modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -264,6 +265,18 @@ if(CONFIG_BUILD_WITH_TFM)
     list(APPEND TFM_CMAKE_ARGS -DHAL_ADI_PATH=${ZEPHYR_ADI_MODULE_DIR})
   endif()
 
+ if(CONFIG_BOARD_FRDM_MCNX947 OR CONFIG_BOARD_FRDM_MCXN947)
+    # Supply path to NXP HAL sources used for TF-M build
+    list(APPEND TFM_CMAKE_ARGS -DZEPHYR_HAL_NXP_MODULE_DIR=${ZEPHYR_HAL_NXP_MODULE_DIR})
+    # Supply path to NXP HAL sources used for TF-M build
+    set(TFM_PLATFORM_NXP_HAL_FILE_PATH ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/platform/ext/target/nxp/)
+    list(APPEND TFM_CMAKE_ARGS -DTFM_PLATFORM_NXP_HAL_FILE_PATH=${TFM_PLATFORM_NXP_HAL_FILE_PATH})
```
tomi-font: Don't duplicate this, just modify the existing occurrence: https://github.com/zephyrproject-rtos/zephyr/blob/0c12a6601cdcf84a5b5a7ccb6459972f190cefa8/modules/trusted-firmware-m/CMakeLists.txt#L257-L261

modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -264,6 +265,18 @@ if(CONFIG_BUILD_WITH_TFM)
     list(APPEND TFM_CMAKE_ARGS -DHAL_ADI_PATH=${ZEPHYR_ADI_MODULE_DIR})
   endif()
 
+ if(CONFIG_BOARD_FRDM_MCNX947 OR CONFIG_BOARD_FRDM_MCXN947)
+    # Supply path to NXP HAL sources used for TF-M build
+    list(APPEND TFM_CMAKE_ARGS -DZEPHYR_HAL_NXP_MODULE_DIR=${ZEPHYR_HAL_NXP_MODULE_DIR})
```
tomi-font: This doesn't seem to exist in TF-M? Why are you adding this to the TF-M args?

modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -264,6 +265,18 @@ if(CONFIG_BUILD_WITH_TFM)
...
+    # Supply path to NXP HAL sources used for TF-M build
+    set(TFM_PLATFORM_NXP_HAL_FILE_PATH ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/platform/ext/target/nxp/)
+    list(APPEND TFM_CMAKE_ARGS -DTFM_PLATFORM_NXP_HAL_FILE_PATH=${TFM_PLATFORM_NXP_HAL_FILE_PATH})
+    # Disable watchdog support on this platform for tfm_psa_tests
+    if(TFM_PSA_TEST_SUITE)
+      list(APPEND TFM_CMAKE_ARGS -DWATCHDOG_AVAILABLE=OFF)
```
tomi-font: Same thing here, where is this checked?

tomi-font (changes requested):

modules/trusted-firmware-m/CMakeLists.txt
waqar-tahir: I have removed 'CONFIG_BOARD_FRDM_MCNX947' in latest revision. Its not being used.

modules/trusted-firmware-m/CMakeLists.txt
waqar-tahir: The change in cmakeLists.txt is cleaned up. As nxp platforms are using drivers from platform/ext/target/nxp/ directory, no need to configure the path from here.

modules/trusted-firmware-m/CMakeLists.txt
waqar-tahir: NXP platform does not need to pass on any path from zephyr CMakeLists.txt now, hence cleaned the file in latest revision.

modules/trusted-firmware-m/CMakeLists.txt
waqar-tahir: Removed this usage as well. Its not required.

modules/trusted-firmware-m/CMakeLists.txt
waqar-tahir: Cleaned it as well. Its not being used.

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns.dts
```diff
@@ -256,3 +257,8 @@ zephyr_udc0: &usb1 {
 &fmu {
 	status = "okay";
 };
+
+&trng {
+	/* status = "okay"; */
```
stgloorious: I would remove this comment. Maybe add the explanation from the commit message here as well.

modules/trusted-firmware-m/Kconfig.tfm
```diff
@@ -12,6 +13,7 @@ config TFM_BOARD
 	string
 	default "adi/max32657" if BOARD_MAX32657EVKIT_MAX32657_NS || BOARD_MAX32658EVKIT_MAX32658_NS
 	default "arm/mps2/an521" if BOARD_MPS2_AN521_CPU0_NS
+	default "nxp/frdmmcxn947" if BOARD_FRDM_MCXN947
```
stgloorious: This should have the `_NS` suffix for consistency

stgloorious (changes requested):

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns.dts
waqar-tahir: Comment updated

modules/trusted-firmware-m/Kconfig.tfm
waqar-tahir: Updated in latest revision. The board name used is: BOARD_FRDM_MCXN947_MCXN947_CPU0_NS

boards/nxp/frdm_mcxn947/frdm_mcxn947.dtsi
```diff
@@ -268,6 +268,11 @@ arduino_serial: &flexcomm2_lpuart2 {};
 		 * Partition sizes must be aligned
 		 * to the flash memory sector size of 8KB.
 		 */
+		slot0_ns_partition: partition@80000 {
+			label = "image-0-nonsecure";
+			reg = <0x00080000 DT_SIZE_K(512)>;
+		};
+
```
stgloorious: I don't think this should be here, it should go into the *_ns.dts only so only the TF-M boards have it. There I would also delete `slot0_partition` and `slot1_partition` or directly move its definition to a separate include file that is included by every board except the `__ns` variant.

So the non-TFM board should still only have `slot0_partition` and `slot1_partition` and the `_ns` variant should have 4 slot partitions (`slot0_s_partition`, `slot0_ns_partition`, `slot1_s_partition`, `slot1_ns_partition`). That is the most common convention afaik.

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns.dts
```diff
@@ -0,0 +1,264 @@
...
+	compatible = "nxp,mcxn947", "nxp,mcx";
+
+	cpus {
+		/delete-node/ cpu@1;
+	};
+
+	chosen {
+		zephyr,sram = &non_secure_ram;
+		zephyr,code-partition = &slot0_ns_partition;
+		zephyr,flash = &slot0_ns_partition;
```
stgloorious:
```suggestion
		zephyr,flash = &flash;
```

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns_defconfig
```diff
@@ -0,0 +1,18 @@
...
+CONFIG_UART_CONSOLE=y
+CONFIG_SERIAL=y
+CONFIG_UART_INTERRUPT_DRIVEN=y
+CONFIG_GPIO=y
+
+CONFIG_ARM_MPU=y
+CONFIG_HW_STACK_PROTECTION=y
+CONFIG_TRUSTED_EXECUTION_NONSECURE=y
+
+CONFIG_TFM_BL2=n
```
stgloorious, line 17: BL2 support would be really helpful. In the current configuration, it also does not work with a board out-of-the-box (I am assuming because the boot ROM is not configured).

stgloorious (changes requested):

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns_defconfig
butok: Guess, there is no reason to disable BL2.

waqar-tahir: FRDM-MCXN947 is not yet supporting TF-M with Bl2 in tf-m upstream. BL2 enablement will come later on and will be enabled for zephyr as soon as tf-m port will support it.
@stgloorious This limitation does not affect the behavior of the Boot ROM on the FRDM-MCXN947. The Boot ROM will continue to operate as designed and will eventually transfer execution to the tfm_secure image located at flash address 0x00000000.

stgloorious: Okay, I see. Then there must be another reason why I cannot boot this at the moment. I will try again later with the latest changes.

waqar-tahir: Just for your info, I am able to boot and pass frdmmcxn947 with following tests locally with these changes.
 - hello_world 
 - tfm_regression test 
 - tfm_ipc tests

boards/nxp/frdm_mcxn947/frdm_mcxn947.dtsi
waqar-tahir: @stgloorious Thanks for the suggestion. I have made the suggested update in the latest revision and '_ns' partition is now only available during builds with tfm. However, since the platform is not yet supporting BL2, hence there are no slot1_xx_partitions.

boards/nxp/frdm_mcxn947/frdm_mcxn947_mcxn947_cpu0_ns.dts
waqar-tahir: This is now fixed in latest revision.

waqar-tahir:
@stgloorious @tomi-font  I have made a new PR for TF-M, without the logging changes, as we are not yet able to identify the fix required(from upstream tfm) for nordic platforms. Now, the TF-M [pr ](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/164) is marked with prefix: [zep fromtree dirty], the change is without the logging related updates for this platform and nxp common files. This also removes the requirement to update the tf-m-tests repo with the logging changes.

tomi-font (dismissed):

modules/trusted-firmware-m/CMakeLists.txt
tomi-font, line 1: commit title should be `modules: tf-m: [...]`

waqar-tahir: commit title update in latest revision.

stgloorious:
> Could you please share you binaries?

Yes, please find attached.
[tfm_merged.hex.zip](https://github.com/user-attachments/files/24188437/tfm_merged.hex.zip)

> Are you able to run tfm based examples ever on your board?

Yes, with the Zephyr patches that @groncarolonxp initially shared, it worked. Both with his TF-M fork and upstream TF-M. See my original PR https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/157

> CONFIG_TFM_BL2=n' has nothing to do how BootROM works [...]

I am aware. But on some platforms you either need to manually configure some options (like enabling and configuring TF-M) with some vendor tool or have BL2 to configure it for you. Because TF-M expects to be booted on an already configured platform afaik.

> [...] this flag was never enabled for frdmmcxn947 even with the changes @groncarolonxp shared with you initially as far as i can tell.

But you're right, the initial patchset did also not include BL2, I thought it did. So this is not the problem.

waqar-tahir:
> Yes, please find attached.
> [tfm_merged.hex.zip](https://github.com/user-attachments/files/24188437/tfm_merged.hex.zip)

@stgloorious I am able to flash your binary and it works.
image

I think, the issue is related to commands being used for erasing of flash and writing the binary. Once TF-M is running, the chip may disable mass flash erase. 
I used following sequence to erase and re-write the new binaries.
1- Erase flash of the board
2- Write the binary (as you are doing it already)

For step 1, you can use NXP SPSDK commands(if you have already installed it, or else: https://spsdk.readthedocs.io/en/stable/examples/_knowledge_base/installation_guide.html):
command: nxpdebugmbox cmd -f mcxn947 erase

Could you please give it a try and let me know if it works for you? Thanks.

stgloorious (approved):

butok (dismissed):

tomi-font (approved):

butok (approved):

stgloorious (approved):

waqar-tahir:
@tomi-font @d3zd3z Hi. Can you kindly help mering the PR or if there is further feedback?
Thanks.

tomi-font:
We can't merge the PR, it's not our responsibility.
It will be picked up by someone responsible for that shortly, as you can see your PR is showing up as mergeable (it fulfills all the requirements to be merged): https://merge-list.zephyrproject.io/

joelguittet:
@waqar-tahir hello, I have seen this recent addition so first thanks for that!
Question: do you have an example to define partition int the flash, with tfp_its, tfm_ps, secure and non secure slot0/1 ?
Thanks

PR #100882, opened 2025-12-11
drivers: i2c: stm32: Fix hang on read log position

### LINKED ISSUE #93881: driver: stm32-i2c-v2: negative return from i2c_target_read_processed causes lockup.
### Describe the bug

According to the [docs](https://docs.zephyrproject.org/apidoc/latest/group__i2c__interface.html#ga4cb0ae2cf41fc2105d1baa5e496e5dae), a negative return from `i2c_target_read_requested_cb_t` or `i2c_target_read_processed_cb_t` should:

> cause the controller to ignore bus operations until a new start condition is received.

However, what actually happens is that the I2C target stops responding, not quite sure what happens, voltage turns off and break points don't get hit anymore.

Here's a normal test with these functions returning 0:
```
I2C> scan
I2C address search:
0x60 (0xC0 W) (0xC1 R)

I2C> [0xC1 r:1]

I2C START
TX: 0xC1 ACK 
RX: 0x42 NACK 
I2C STOP
I2C> [0xC1 r:1]

I2C START
TX: 0xC1 ACK 
RX: 0x42 NACK 
I2C STOP
I2C> [0xC1 r:1]
```

Here's a test with `sample_target_read_processed_cb` returning `-1`:
```
I2C> [0xC1 r:1]

I2C START
TX: 0xC1 ACK 
RX: 0x42 NACK 
I2C STOP
I2C> [0xC1 r:1]

I2C START
TX: 0xC1 NACK 
RX: 0xFF NACK 
I2C STOP
```

FWIW I'm testing with an Arduino Giga R1.

Also, this might be related: https://github.com/zephyrproject-rtos/zephyr/pull/80994

### Regression

- [ ] This is a regression.

### Steps to reproduce

1. In `samples/drivers/i2c/custom_target/src/main.c` change `sample_target_read_processed_cb` to return `-1`.
2. Read from the I2C target.

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

juickar:
Ensure that data is transmitted before logging as it could cause lockups in some setups. see #93881 

Also update i2c_stm32_v2_rtio.c to match this behavior to always send data in order to avoid the I2C to lockup.

Fixes #93881

### DIFF
--- a/drivers/i2c/i2c_stm32_v2.c
+++ b/drivers/i2c/i2c_stm32_v2.c
@@ -270,12 +270,16 @@ static void i2c_stm32_slave_event(const struct device *dev)
 
 	if (LL_I2C_IsActiveFlag_TXIS(i2c)) {
 		uint8_t val = 0x00;
+		int ret = slave_cb->read_processed(slave_cfg, &val);
 
-		if (slave_cb->read_processed(slave_cfg, &val) < 0) {
+		/* We should transmit the data before logging because logging could
+		 * lead the i2c slave to stretch the clock if the data are not
+		 * transmitted fast enough.
+		 */
+		LL_I2C_TransmitData8(i2c, val);
+		if (ret < 0) {
 			LOG_ERR("Error continuing reading");
 		}
-
-		LL_I2C_TransmitData8(i2c, val);
 
 		return;
 	}
--- a/drivers/i2c/i2c_stm32_v2_rtio.c
+++ b/drivers/i2c/i2c_stm32_v2_rtio.c
@@ -115,13 +115,18 @@ static void i2c_stm32_target_event(const struct device *dev)
 	target_cb = target_cfg->callbacks;
 
 	if (LL_I2C_IsActiveFlag_TXIS(i2c)) {
-		uint8_t val;
+		uint8_t val = 0x00;
+		int ret = target_cb->read_processed(target_cfg, &val);
 
-		if (target_cb->read_processed(target_cfg, &val) < 0) {
+		/* We should transmit the data before logging because logging could
+		 * lead the i2c target to stretch the clock if the data are not
+		 * transmitted fast enough.
+		 */
+		LL_I2C_TransmitData8(i2c, val);
+		if (ret < 0) {
 			LOG_ERR("Error continuing reading");
-		} else {
-			LL_I2C_TransmitData8(i2c, val);
 		}
+
 		return;
 	}
 


drivers/i2c/i2c_stm32_v2.c
```diff
@@ -270,13 +270,13 @@ static void i2c_stm32_slave_event(const struct device *dev)
 
 	if (LL_I2C_IsActiveFlag_TXIS(i2c)) {
 		uint8_t val = 0x00;
+		uint32_t ret = slave_cb->read_processed(slave_cfg, &val);
```
mathieuchopstm:
```suggestion
		int ret = slave_cb->read_processed(slave_cfg, &val);
```

drivers/i2c/i2c_stm32_v2.c
```diff
@@ -270,13 +270,13 @@ static void i2c_stm32_slave_event(const struct device *dev)
...
 
-		if (slave_cb->read_processed(slave_cfg, &val) < 0) {
+		LL_I2C_TransmitData8(i2c, val);
```
mathieuchopstm, line 279: Please add a comment about why the TX is done before log to avoid an incorrect code sequence being re-introduced again.

drivers/i2c/i2c_stm32_v2_rtio.c
```diff
@@ -115,13 +115,14 @@ static void i2c_stm32_target_event(const struct device *dev)
 	target_cb = target_cfg->callbacks;
 
 	if (LL_I2C_IsActiveFlag_TXIS(i2c)) {
-		uint8_t val;
+		uint8_t val = 0x00;
+		uint32_t ret = target_cb->read_processed(target_cfg, &val);
```
mathieuchopstm:
```suggestion
		int ret = target_cb->read_processed(target_cfg, &val);
```

drivers/i2c/i2c_stm32_v2_rtio.c
```diff
@@ -115,13 +115,14 @@ static void i2c_stm32_target_event(const struct device *dev)
...
 
-		if (target_cb->read_processed(target_cfg, &val) < 0) {
+		LL_I2C_TransmitData8(i2c, val);
```
mathieuchopstm, line 125: ditto wrt comment

mathieuchopstm (changes requested):

erwango (approved):

mathieuchopstm (approved):

PR #100884, opened 2025-12-11
net: dhcpv4_server: Allow application to reject an address

icsys-aal:
This is useful if the application may wish to reject an address. This works well alongside the address_provider callback, this way you can enforce static leases based on the client id (MAC address).

### DIFF
--- a/include/zephyr/net/dhcpv4_server.h
+++ b/include/zephyr/net/dhcpv4_server.h
@@ -148,6 +148,34 @@ void net_dhcpv4_server_set_provider_cb(net_dhcpv4_server_provider_cb_t cb,
 				       void *user_data);
 
 /**
+ * @typedef net_dhcpv4_server_address_validator_cb_t
+ * @brief Callback used to let application accept or reject a proposed DHCP
+ * address
+ * @details This function is called before accepting an address to a client,
+ * and lets the application reject the address for a given client. If the
+ * callback returns 0, addr needs to be a valid address and will be assigned
+ * to the client. If the callback returns anything non-zero, the client will
+ * be assigned an address from the pool.
+ *
+ * @param iface Pointer to the network interface
+ * @param client_id Pointer to client requesting the address
+ * @param addr Address to be assigned to client
+ * @param user_data A valid pointer to user data or NULL
+ */
+typedef bool (*net_dhcpv4_server_address_validator_cb_t)(struct net_if *iface,
+							const struct dhcpv4_client_id *client_id,
+							const struct net_in_addr *addr,
+							void *user_data);
+/**
+ * @brief Set the callback used to validate new addresses to the DHCP server.
+ *
+ * @param cb User-supplied callback function to call
+ * @param user_data A valid pointer to user data or NULL
+ */
+void net_dhcpv4_server_set_address_validator_cb(net_dhcpv4_server_address_validator_cb_t cb,
+						void *user_data);
+
+/**
  * @}
  */
 
--- a/subsys/net/lib/dhcpv4/dhcpv4_server.c
+++ b/subsys/net/lib/dhcpv4/dhcpv4_server.c
@@ -64,6 +64,7 @@ struct dhcpv4_server_ctx {
 	int sock;
 	struct k_work_delayable timeout_work;
 	struct dhcpv4_addr_slot addr_pool[CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT];
+	struct net_in_addr base_addr;
 	struct net_in_addr server_addr;
 	struct net_in_addr netmask;
 #if defined(DHCPV4_SERVER_ICMP_PROBE)
@@ -71,11 +72,27 @@ struct dhcpv4_server_ctx {
 #endif
 };
 
+static struct net_in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
+							  struct dhcpv4_client_id *client_id);
+
+static void *address_validator_callback_user_data;
+static net_dhcpv4_server_address_validator_cb_t address_validator_callback;
+
 static void *address_provider_callback_user_data;
 static net_dhcpv4_server_provider_cb_t address_provider_callback;
+
 static struct dhcpv4_server_ctx server_ctx[CONFIG_NET_DHCPV4_SERVER_INSTANCES];
 static struct zsock_pollfd fds[CONFIG_NET_DHCPV4_SERVER_INSTANCES];
 static K_MUTEX_DEFINE(server_lock);
+
+static void reset_slot(struct dhcpv4_addr_slot *slot)
+{
+	slot->state = DHCPV4_SERVER_ADDR_FREE;
+	slot->addr.s_addr = 0;
+	slot->expiry = sys_timepoint_calc(K_FOREVER);
+	memset(slot->client_id.buf, 0, DHCPV4_CLIENT_ID_MAX_SIZE);
+	slot->client_id.len = 0;
+}
 
 static void dhcpv4_server_timeout_recalc(struct dhcpv4_server_ctx *ctx)
 {
@@ -842,7 +859,7 @@ static int echo_reply_handler(struct net_icmp_ctx *icmp_ctx,
 	probe_ctx->slot->state = DHCPV4_SERVER_ADDR_DECLINED;
 	probe_ctx->slot->expiry = sys_timepoint_calc(ADDRESS_DECLINED_TIMEOUT);
 
-	/* Try to find next free address */
+	/* Try to find next free slot */
 	for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
 		struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
 
@@ -858,6 +875,7 @@ static int echo_reply_handler(struct net_icmp_ctx *icmp_ctx,
 		dhcpv4_server_timeout_recalc(ctx);
 		goto out;
 	}
+	new_slot->addr = net_dhcpv4_server_find_free_address(ctx, &probe_ctx->slot->client_id);
 
 	if (dhcpv4_probe_address(ctx, new_slot) < 0) {
 		probe_ctx->slot = NULL;
@@ -930,6 +948,7 @@ static void dhcpv4_server_probe_timeout(struct dhcpv4_server_ctx *ctx,
 	if (dhcpv4_send_offer(ctx, &ctx->probe_ctx.discovery, &slot->addr,
 			      slot->lease_time, &ctx->probe_ctx.params,
 			      &ctx->probe_ctx.client_id) < 0) {
+		reset_slot(slot);
 		slot->state = DHCPV4_SERVER_ADDR_FREE;
 		return;
 	}
@@ -1027,30 +1046,20 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 				}
 			}
 		}
-	}
 
-	/* 4. Allocate new address from pool, if available. */
-	if (selected == NULL) {
-		struct net_in_addr giaddr;
+		/* 4. Allocate new slot from pool, if available. */
+		if (selected == NULL) {
+			for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
+				struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
 
-		memcpy(&giaddr, msg->giaddr, sizeof(giaddr));
-		if (!net_ipv4_is_addr_unspecified(&giaddr)) {
-			/* Only addresses in local subnet supported for now. */
-			return;
-		}
-
-		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
-			struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
-
-			if (slot->state == DHCPV4_SERVER_ADDR_FREE) {
-				/* Requested address is free. */
-				selected = slot;
-				probe = true;
-				break;
+				if (slot->state == DHCPV4_SERVER_ADDR_FREE) {
+					selected = slot;
+					probe = true;
+					break;
+				}
 			}
 		}
 	}
-
 	/* In case no free address slot was found, as a last resort, try to
 	 * reuse the oldest declined entry, if present.
 	 */
@@ -1076,6 +1085,28 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 		LOG_ERR("No free address found in address pool");
 	} else {
 		uint32_t lease_time = dhcpv4_get_lease_time(options, optlen);
+
+		if (selected->state == DHCPV4_SERVER_ADDR_DECLINED) {
+			/* Reusing declined address slot, need to clear it first */
+			reset_slot(selected);
+		}
+		if (selected->state == DHCPV4_SERVER_ADDR_FREE) {
+
+			struct net_in_addr selected_addr;
+
+			selected_addr.s_addr = selected->addr.s_addr;
+			if (address_provider_callback != NULL &&
+				address_provider_callback(ctx->iface, &client_id, &selected_addr,
+							address_provider_callback_user_data) == 0) {
+				selected->addr.s_addr = selected_addr.s_addr;
+				probe = false;
+			} else {
+				/* Try to look for next available address */
+				selected->addr = net_dhcpv4_server_find_free_address(ctx,
+							&client_id);
+				probe = true;
+			}
+		}
 
 		if (IS_ENABLED(DHCPV4_SERVER_ICMP_PROBE) && probe) {
 			if (dhcpv4_server_probe_setup(ctx, selected, msg,
@@ -1119,6 +1150,7 @@ static void dhcpv4_handle_request(struct dhcpv4_server_ctx *ctx,
 	struct dhcpv4_client_id client_id;
 	struct net_in_addr requested_ip, server_id, ciaddr, giaddr;
 	int ret;
+	bool success;
 
 	memcpy(&ciaddr, msg->ciaddr, sizeof(ciaddr));
 	memcpy(&giaddr, msg->giaddr, sizeof(giaddr));
@@ -1154,6 +1186,15 @@ static void dhcpv4_handle_request(struct dhcpv4_server_ctx *ctx,
 		if (!net_ipv4_is_addr_unspecified(&ciaddr)) {
 			/* ciaddr MUST be zero */
 			return;
+		}
+
+		if (address_validator_callback != NULL) {
+			success = address_validator_callback(ctx->iface, &client_id, &requested_ip,
+							 address_validator_callback_user_data);
+			if (!success) {
+				dhcpv4_send_nak(ctx, msg, &client_id);
+				return;
+			}
 		}
 
 		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
@@ -1205,6 +1246,15 @@ static void dhcpv4_handle_request(struct dhcpv4_server_ctx *ctx,
 		if (!net_if_ipv4_addr_mask_cmp(ctx->iface, &requested_ip)) {
 			/* Wrong subnet. */
 			dhcpv4_send_nak(ctx, msg, &client_id);
+		}
+
+		if (address_validator_callback != NULL) {
+			success = address_validator_callback(ctx->iface, &client_id, &requested_ip,
+							 address_validator_callback_user_data);
+			if (!success) {
+				dhcpv4_send_nak(ctx, msg, &client_id);
+				return;
+			}
 		}
 
 		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
@@ -1379,9 +1429,8 @@ static void dhcpv4_handle_release(struct dhcpv4_server_ctx *ctx,
 			LOG_DBG("DHCPv4 processing Release - %s",
 				net_sprint_ipv4_addr(&slot->addr));
 
-			slot->state = DHCPV4_SERVER_ADDR_FREE;
-			slot->expiry = sys_timepoint_calc(K_FOREVER);
 			dhcpv4_server_timeout_recalc(ctx);
+			reset_slot(slot);
 			break;
 		}
 	}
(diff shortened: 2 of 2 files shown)


subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -64,15 +64,22 @@ struct dhcpv4_server_ctx {
 	int sock;
 	struct k_work_delayable timeout_work;
 	struct dhcpv4_addr_slot addr_pool[CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT];
-	struct net_in_addr server_addr;
-	struct net_in_addr netmask;
+	struct in_addr base_addr;
+	struct in_addr server_addr;
+	struct in_addr netmask;
```
jukkar: Please do not change the struct types, we namespaced various network APIs in order to avoid conflicts with POSIX.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1072,6 +1092,19 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 	} else {
 		uint32_t lease_time = dhcpv4_get_lease_time(options, optlen);
 
+		struct in_addr selected_addr;
```
jukkar: same here

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1151,6 +1184,16 @@ static void dhcpv4_handle_request(struct dhcpv4_server_ctx *ctx,
 			return;
 		}
 
+		if(address_validator_callback)
```
jukkar: Please write this as `if (address_validator_callback != NULL) {`

I just wonder would it make sense to have the callback specified for each interface separately. This way it would be possible to have different validator in multi interface system.

rlubos (commented):
I need to better understand the sue case for this new callback type. We already have `address_provider_callback`, where the application can take ownership of the address management. Can't it just return some error in case the given client won't be assigned an address? Why do we need both callbacks?

subsys/net/lib/dhcpv4/dhcpv4_server.c
icsys-aal: Personally I think that should be handled by the application, which should be possible given that we provide the network interface in the callback

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1072,6 +1093,20 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 	} else {
 		uint32_t lease_time = dhcpv4_get_lease_time(options, optlen);
 
+		struct net_in_addr selected_addr;
+
+		selected_addr.s_addr = selected->addr.s_addr;
+		ret = address_provider_callback(ctx->iface, &client_id, &selected_addr,
```
rlubos: Need to check first if `address_provider_callback` was set, no?

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1202,6 +1246,15 @@ static void dhcpv4_handle_request(struct dhcpv4_server_ctx *ctx,
 			dhcpv4_send_nak(ctx, msg, &client_id);
 		}
 
+		if (address_validator_callback) {
```
rlubos: Please make all pointer checks agains NULL, so
```
if (address_validator_callback != NULL) {
```

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1415,13 +1467,14 @@ static void dhcpv4_server_timeout(struct k_work *work)
 			} else {
 				LOG_DBG("Address %s expired",
 					net_sprint_ipv4_addr(&slot->addr));
-				slot->state = DHCPV4_SERVER_ADDR_FREE;
+				printk("Resetting index %d from timeout\n", i);
```
rlubos: Don't use printk, use logger instead.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1640,20 +1693,15 @@ int net_dhcpv4_server_start(struct net_if *iface, struct net_in_addr *base_addr)
 	server_ctx[slot].iface = iface;
 	server_ctx[slot].sock = sock;
 	server_ctx[slot].server_addr = *server_addr;
+	server_ctx[slot].base_addr = *base_addr;
 	server_ctx[slot].netmask = netmask;
 
 	k_work_init_delayable(&server_ctx[slot].timeout_work,
 			      dhcpv4_server_timeout);
 
 	LOG_DBG("Started DHCPv4 server, address pool:");
```
rlubos, line 1703: It does not print the address pool now. We can't we keep the log?

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1842,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
...
+{
+	/* Try to look for next available address */
+	struct net_in_addr address;
+
+	for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
+		address.s_addr = htonl(ntohl(ctx->base_addr.s_addr) + i);
+		bool taken = false;
+
+		/* Skip if address is not assignable by application */
+		if (address_validator_callback &&
```
rlubos: `address_validator_callback != NULL`

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1842,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
...
+			address_validator_callback(ctx->iface, client_id, &address,
```
rlubos: Please fix the alignment, the if condition contents should align with the opening bracket, not the block content, it's currently hard to read.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1842,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
...
+				address_validator_callback_user_data) != 0) {
+			continue;
+		}
+
+		/* Check if address is free */
+		for (unsigned int j = 0; j < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++j) {
+			struct dhcpv4_addr_slot *slot = &ctx->addr_pool[j];
+
+			if (slot->state != DHCPV4_SERVER_ADDR_FREE &&
+				net_ipv4_addr_cmp(&address, &slot->addr)) {
```
rlubos: Same here, please fix the alignement.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1842,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
...
+				/* Address is not free */
+				taken = true;
+				break;
+			}
+		}
+
+		if (taken == false) {
```
rlubos: `!taken`

rlubos (commented):
It'd be nice to have some test coverage for this.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1640,16 +1692,15 @@ int net_dhcpv4_server_start(struct net_if *iface, struct net_in_addr *base_addr)
...
 			      dhcpv4_server_timeout);
 
 	LOG_DBG("Started DHCPv4 server, address pool:");
 	for (int i = 0; i < ARRAY_SIZE(server_ctx[slot].addr_pool); i++) {
-		server_ctx[slot].addr_pool[i].state = DHCPV4_SERVER_ADDR_FREE;
-		server_ctx[slot].addr_pool[i].addr.s_addr =
-					net_htonl(net_ntohl(base_addr->s_addr) + i);
+		reset_slot(&server_ctx[slot].addr_pool[i]);
 
 		LOG_DBG("\t%2d: %s", i,
```
rlubos: Yeah but now we'll just get printed all 0 for each address. Either create a dummy `net_htonl(net_ntohl(base_addr->s_addr) + i` address just for the log, or give up on address pool printing (including the log above). Personally I'd prefer the former I think.

include/zephyr/net/dhcpv4_server.h
```diff
@@ -137,6 +137,34 @@ typedef int (*net_dhcpv4_server_provider_cb_t)(struct net_if *iface,
...
+ * be assigned an address from the pool.
+ *
+ * @param iface Pointer to the network interface
+ * @param client_id Pointer to client requesting the address
+ * @param addr Address to be assigned to client
+ * @param user_data A valid pointer to user data or NULL
+ */
+typedef int (*net_dhcpv4_server_address_validator_cb_t)(struct net_if *iface,
+							const struct dhcpv4_client_id *client_id,
+							const struct in_addr *addr,
```
rlubos: Probably the reason of twister failures.
```suggestion
							const struct net_in_addr *addr,
```

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -64,15 +64,23 @@ struct dhcpv4_server_ctx {
...
 	struct dhcpv4_addr_slot addr_pool[CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT];
+	struct net_in_addr base_addr;
 	struct net_in_addr server_addr;
 	struct net_in_addr netmask;
 #if defined(DHCPV4_SERVER_ICMP_PROBE)
 	struct dhcpv4_server_probe_ctx probe_ctx;
 #endif
 };
 
+static struct in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
```
rlubos:
```suggestion
static struct net_in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
```

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1845,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
 	return ret;
 }
 
+/*
+ *	Finds a free address, taking the validator callback into account
+ */
+static struct in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
```
rlubos:
```suggestion
static struct net_in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
```

rlubos (commented):
Some final comments and I think we're good.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1794,17 +1845,69 @@ int net_dhcpv4_server_foreach_lease(struct net_if *iface,
...
+ *	Finds a free address, taking the validator callback into account
+ */
+static struct net_in_addr net_dhcpv4_server_find_free_address(struct dhcpv4_server_ctx *ctx,
+							  struct dhcpv4_client_id *client_id)
+{
+	/* Try to look for next available address */
+	struct net_in_addr address;
+
+	for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
+		address.s_addr = htonl(ntohl(ctx->base_addr.s_addr) + i);
```
rlubos: `htonl` -> `net_htonl`
`ntohl` -> `net_ntohl`
One of the reasons twister fails.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1640,20 +1692,19 @@ int net_dhcpv4_server_start(struct net_if *iface, struct net_in_addr *base_addr)
...
 	for (int i = 0; i < ARRAY_SIZE(server_ctx[slot].addr_pool); i++) {
-		server_ctx[slot].addr_pool[i].state = DHCPV4_SERVER_ADDR_FREE;
-		server_ctx[slot].addr_pool[i].addr.s_addr =
-					net_htonl(net_ntohl(base_addr->s_addr) + i);
+		reset_slot(&server_ctx[slot].addr_pool[i]);
 
 		LOG_DBG("\t%2d: %s", i,
 			net_sprint_ipv4_addr(
-				&server_ctx[slot].addr_pool[i].addr));
+				net_htonl(net_ntohl(base_addr->s_addr) + i)));
```
rlubos: `net_sprint_ipv4_addr()` expects an address pointer, not the address itself.

rlubos (commented):
A couple of issues causing twister errors.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -852,6 +865,8 @@ static int echo_reply_handler(struct net_icmp_ctx *icmp_ctx,
 		}
 	}
 
+	new_slot->addr = net_dhcpv4_server_find_free_address(ctx, &probe_ctx->slot->client_id);
+
```
jukkar: There is something wrong with the next if check, how come we check new_slot after we have assigned a value to it.

include/zephyr/net/dhcpv4_server.h
```diff
@@ -147,6 +137,34 @@ typedef int (*net_dhcpv4_server_provider_cb_t)(struct net_if *iface,
...
+ * callback returns 0, addr needs to be a valid address and will be assigned
+ * to the client. If the callback returns anything non-zero, the client will
+ * be assigned an address from the pool.
+ *
+ * @param iface Pointer to the network interface
+ * @param client_id Pointer to client requesting the address
+ * @param addr Address to be assigned to client
+ * @param user_data A valid pointer to user data or NULL
+ */
+typedef int (*net_dhcpv4_server_address_validator_cb_t)(struct net_if *iface,
```
jukkar: If there is only two possible return values, why not use boolean type?

icsys-aal: To me it feels wrong to use a boolean type in C.

maass-hamburg: But it is fine, especially if there are just two possible return types, it should be used.

rlubos (commented):
Remaining CI issues:

Compliance:
* `4c1e639bf127df66f9c88fd4f4171d12f674077d: subsys/net/lib/dhcpv4/dhcpv4_server.c:814: trailing whitespace.`

Twister:
* Shadowed local variable: `/__w/zephyr/zephyr/subsys/net/lib/dhcpv4/dhcpv4_server.c:1696:36: error: declaration of 'addr' shadows a previous local [-Werror=shadow]
 1696 |                 struct net_in_addr addr = {`
* `reset_slot()` is hidden behind `#if defined(DHCPV4_SERVER_ICMP_PROBE)`: `/__w/zephyr/zephyr/subsys/net/lib/dhcpv4/dhcpv4_server.c:1422:25: error: implicit declaration of function 'reset_slot' [-Werror=implicit-function-declaration]
 1422 |                         reset_slot(slot);`

Please fix

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1029,25 +1048,12 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
...
-		memcpy(&giaddr, msg->giaddr, sizeof(giaddr));
-		if (!net_ipv4_is_addr_unspecified(&giaddr)) {
-			/* Only addresses in local subnet supported for now. */
-			return;
-		}
-
-		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
-			struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
+	/* 4. Allocate new slot from pool, if available. */
+	for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
```
rlubos: This should still be within:
```
if (selected == NULL) {
   ...
}
```
block, otherwise all previous slot selection rules are ignored.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1029,25 +1048,12 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
...
+		struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
 
-			if (slot->state == DHCPV4_SERVER_ADDR_FREE) {
-				/* Requested address is free. */
-				selected = slot;
-				probe = true;
-				break;
-			}
+		if (slot->state == DHCPV4_SERVER_ADDR_FREE) {
+			selected = slot;
```
rlubos: We still need to set probe variable, I see you do it below but I don't think that's correct - why skipping probe for application provided addresses, from the DHCP perspective it makes no difference how the address was chosen. Also restore the `break;` statement when empty slot is found to avoid unnecessary loop iterations.
```suggestion
			selected = slot;
			probe = true;
			break;
```

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1077,6 +1083,20 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
...
+		if (address_provider_callback != NULL &&
+		    address_provider_callback(ctx->iface, &client_id, &selected_addr,
+					      address_provider_callback_user_data) == 0) {
+			selected->addr.s_addr = selected_addr.s_addr;
+			probe = false;
+		} else {
+			/* Try to look for next available address */
+			selected->addr = net_dhcpv4_server_find_free_address(ctx, &client_id);
+			probe = true;
+		}
```
rlubos: This whole block should be hidden behind `if (selected->state == DHCPV4_SERVER_ADDR_FREE) { }`.
According to slot selection rule 1., we will resend the offer message for a given client ID if the address is already assigned or reserved. We should not allocate a new address in such cases.

This and above broke the `test_discover_repeat` test from the test suite.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1077,6 +1083,20 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 	} else {
 		uint32_t lease_time = dhcpv4_get_lease_time(options, optlen);
```
rlubos, line 1088: In case we're reusing the oldest DECLINED slot, we need to clear it first so that the address can be reused according to address selection rules below.
```
		if (selected->state == DHCPV4_SERVER_ADDR_DECLINED) {
			/* Reusing declined address slot, need to clear it first */
			reset_slot(selected);
		}
```

This broke the `test_declined_reuse` test from the test suite.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1645,20 +1683,21 @@ int net_dhcpv4_server_start(struct net_if *iface, struct net_in_addr *base_addr)
...
 			      dhcpv4_server_timeout);
 
 	LOG_DBG("Started DHCPv4 server, address pool:");
 	for (int i = 0; i < ARRAY_SIZE(server_ctx[slot].addr_pool); i++) {
-		server_ctx[slot].addr_pool[i].state = DHCPV4_SERVER_ADDR_FREE;
-		server_ctx[slot].addr_pool[i].addr.s_addr =
-					net_htonl(net_ntohl(base_addr->s_addr) + i);
+		reset_slot(&server_ctx[slot].addr_pool[i]);
+
+		struct net_in_addr tempAddr = {
```
rlubos: snake case please
```suggestion
		struct net_in_addr temp_addr = {
```

rlubos (commented):
Now that the tests are building the DHCPv4 server testsuite exposed some regressions with this PR. Left some suggestions on how to fix them.

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1027,30 +1046,18 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
...
 
-		memcpy(&giaddr, msg->giaddr, sizeof(giaddr));
-		if (!net_ipv4_is_addr_unspecified(&giaddr)) {
-			/* Only addresses in local subnet supported for now. */
-			return;
-		}
-
-		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
+		/* 4. Allocate new slot from pool, if available. */
+		for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
```
rlubos: That should still be in a separate `if (selected == NULL) { }` block as it was, otherwise, we'll always override the selected slot from rule 3 (if found).

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1029,24 +1048,13 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
...
-
-		for (int i = 0; i < ARRAY_SIZE(ctx->addr_pool); i++) {
+		for (unsigned int i = 0; i < CONFIG_NET_DHCPV4_SERVER_ADDR_COUNT; ++i) {
 			struct dhcpv4_addr_slot *slot = &ctx->addr_pool[i];
 
 			if (slot->state == DHCPV4_SERVER_ADDR_FREE) {
-				/* Requested address is free. */
 				selected = slot;
-				probe = true;
-				break;
```
rlubos, line 1049: My earlier comment https://github.com/zephyrproject-rtos/zephyr/pull/100884#discussion_r2852518406 - this was already fixed and now we're back to the starting point. Please fix.

rlubos (changes requested):

subsys/net/lib/dhcpv4/dhcpv4_server.c
```diff
@@ -1027,30 +1046,18 @@ static void dhcpv4_handle_discover(struct dhcpv4_server_ctx *ctx,
 				}
 			}
 		}
-	}
-
-	/* 4. Allocate new address from pool, if available. */
-	if (selected == NULL) {
```
rlubos, line 1033: Now the change requested in https://github.com/zephyrproject-rtos/zephyr/pull/100884#discussion_r2896462259 was reverted...

rlubos (approved):

maass-hamburg (approved):

PR #19208, opened 2019-09-17
Bluetooth: Host: Remove printk dependency from settings

kimsekkelund:
Some modules use snprintk to format the settings keys. Unfortunately
snprintk is tied with printk which is very large for some embedded
systems.
To be able to have settings enabled without also enabling printk
support, change creation of settings key strings to use bin2hex, strlen
and strcpy instead.
A utility function to make decimal presentation of a byte value is
added.

Signed-off-by: Kim Sekkelund <ksek@oticon.com>

PR #100887, opened 2025-12-11
Bluetooth: Host: settings without snprintk use

cvinayak:
Update settings implementation for not reusing the OS snprintk implementation with CONFIG_PRINTK=n.

With PR ​

```
west build -p -d build/peripheral_hids_nrf54l15dk -b nrf54l15dk/nrf54l15/cpuapp samples/bluetooth/peripheral_hids -DCONFIG_BOOT_BANNER=n -DCONFIG_PRINTK=n -DCONFIG_BT_SETTINGS_USE_PRINTK=n -DCONFIG_LOG=n -DCONFIG_OUTPUT_DISASSEMBLY=y
```

```
           FLASH:      155012 B      1428 KB     10.60%
             RAM:       27420 B       188 KB     14.24%
        IDT_LIST:          0 GB        32 KB      0.00%
```

Without PR, base branch https://github.com/zephyrproject-rtos/zephyr/tree/0b52928625b06b6b8de9b56524fbc1ae5f8cbd6a

```
           FLASH:      157012 B      1428 KB     10.74%
             RAM:       27420 B       188 KB     14.24%
        IDT_LIST:          0 GB        32 KB      0.00%
```
​

[0001-samples-Bluetooth-peripheral_hids-Reduce-code-size-w.patch](https://github.com/user-attachments/files/24770057/0001-samples-Bluetooth-peripheral_hids-Reduce-code-size-w.patch)

20260121:
```
-- Zephyr version: 4.3.99 (/home/vich/workspace/zephyrproject/zephyr), build: v4.3.0-2228-g3aa46ff1adea
[387/387] Linking C executable zephyr/zephyr.elf
           FLASH:      154912 B      1428 KB     10.59%
             RAM:       27420 B       188 KB     14.24%
        IDT_LIST:          0 GB        32 KB      0.00%
```

### DIFF
--- a/subsys/bluetooth/host/att.c
+++ b/subsys/bluetooth/host/att.c
@@ -376,7 +376,6 @@ static struct net_buf *att_create_rsp_pdu(struct bt_att_chan *chan, uint8_t op);
 
 static void att_disconnect(struct bt_att_chan *chan)
 {
-	char addr[BT_ADDR_LE_STR_LEN];
 	int err;
 
 	/* In rare circumstances we are "forced" to disconnect the ATT bearer and the ACL.
@@ -385,8 +384,7 @@ static void att_disconnect(struct bt_att_chan *chan)
 	 * invalid
 	 */
 
-	bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
-	LOG_DBG("ATT disconnecting device %s", addr);
+	LOG_DBG("ATT disconnecting device %s", bt_addr_le_str(bt_conn_get_dst(chan->att->conn)));
 
 	bt_att_disconnected(&chan->chan.chan);
 
@@ -3195,12 +3193,11 @@ static void att_chan_detach(struct bt_att_chan *chan)
 
 static void att_timeout(struct k_work *work)
 {
-	char addr[BT_ADDR_LE_STR_LEN];
 	struct k_work_delayable *dwork = k_work_delayable_from_work(work);
 	struct bt_att_chan *chan = CONTAINER_OF(dwork, struct bt_att_chan, timeout_work);
 
-	bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
-	LOG_ERR("ATT Timeout for device %s. Disconnecting...", addr);
+	LOG_ERR("ATT Timeout for device %s. Disconnecting...",
+		bt_addr_le_str(bt_conn_get_dst(chan->att->conn)));
 
 	/* BLUETOOTH SPECIFICATION Version 4.2 [Vol 3, Part F] page 480:
 	 *
--- a/subsys/bluetooth/host/settings.c
+++ b/subsys/bluetooth/host/settings.c
@@ -33,28 +33,45 @@
 LOG_MODULE_REGISTER(bt_settings);
 
 #if defined(CONFIG_BT_SETTINGS_USE_PRINTK)
-void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
+int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 			    const bt_addr_le_t *addr, const char *key)
 {
+	int err;
+
 	if (key) {
-		snprintk(path, path_size,
-			 "bt/%s/%02x%02x%02x%02x%02x%02x%u/%s", subsys,
-			 addr->a.val[5], addr->a.val[4], addr->a.val[3],
-			 addr->a.val[2], addr->a.val[1], addr->a.val[0],
-			 addr->type, key);
+		err = snprintk(path, path_size, "bt/%s/%02x%02x%02x%02x%02x%02x%u/%s", subsys,
+			       addr->a.val[5], addr->a.val[4], addr->a.val[3], addr->a.val[2],
+			       addr->a.val[1], addr->a.val[0], addr->type, key);
 	} else {
-		snprintk(path, path_size,
-			 "bt/%s/%02x%02x%02x%02x%02x%02x%u", subsys,
-			 addr->a.val[5], addr->a.val[4], addr->a.val[3],
-			 addr->a.val[2], addr->a.val[1], addr->a.val[0],
-			 addr->type);
+		err = snprintk(path, path_size, "bt/%s/%02x%02x%02x%02x%02x%02x%u", subsys,
+			       addr->a.val[5], addr->a.val[4], addr->a.val[3], addr->a.val[2],
+			       addr->a.val[1], addr->a.val[0], addr->type);
+	}
+
+	if (err < 0) {
+		return -EINVAL;
 	}
 
 	LOG_DBG("Encoded path %s", path);
+
+	return 0;
 }
-#else
-void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
-			    const bt_addr_le_t *addr, const char *key)
+
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	int err;
+
+	err = snprintk(path, path_size, "bt/%s", key);
+	if (err < 0) {
+		return -EINVAL;
+	}
+
+	return 0;
+}
+
+#else /* !CONFIG_BT_SETTINGS_USE_PRINTK */
+int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
+			   const bt_addr_le_t *addr, const char *key)
 {
 	size_t len = 3;
 
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 	}
 
 	LOG_DBG("Encoded path %s", path);
+
+	return 0;
 }
-#endif
+
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	size_t len = 3;
+
+	/* Skip if path_size is less than 3; strlen("bt/") */
+	if (len < path_size) {
+		/* Key format:
+		 *  "bt/<key>", "/<key>" is optional
+		 */
+		strcpy(path, "bt/");
+
+		/* Concatenate key as much the free space permits */
+		strncpy(&path[len], key, path_size - len);
+		len = strlen(path);
+
+		/* If path string is full, always null terminate at path_size */
+		if (len >= path_size) {
+			/* Truncate string */
+			path[path_size - 1] = '\0';
+		}
+
+	} else if (path_size > 0) {
+		/* path_size not sufficient for "bt/" */
+		*path = '\0';
+	}
+
+	LOG_DBG("Encoded path %s", path);
+
+	return 0;
+}
+#endif /* !CONFIG_BT_SETTINGS_USE_PRINTK */
 
 int bt_settings_decode_key(const char *key, bt_addr_le_t *addr)
 {
@@ -324,21 +374,23 @@ __weak void bt_testing_settings_delete_hook(const char *key)
 int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, const void *value,
 		      size_t val_len)
 {
-	int err;
-	char id_str[4];
 	char key_str[BT_SETTINGS_KEY_MAX];
+	char id_str[4];
+	int err;
 
 	if (addr) {
 		if (id) {
 			u8_to_dec(id_str, sizeof(id_str), id);
 		}
 
-		bt_settings_encode_key(key_str, sizeof(key_str), key, addr, (id ? id_str : NULL));
+		err = bt_settings_encode_key(key_str, sizeof(key_str), key, addr,
+					     (id ? id_str : NULL));
 	} else {
-		err = snprintk(key_str, sizeof(key_str), "bt/%s", key);
-		if (err < 0) {
-			return -EINVAL;
-		}
+		err = bt_settings_encode_key_no_addr(key_str, sizeof(key_str), key);
+	}
+
+	if (err != 0) {
+		return err;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_TESTING)) {
@@ -350,21 +402,23 @@ int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, con
 
 int bt_settings_delete(const char *key, uint8_t id, const bt_addr_le_t *addr)
 {
-	int err;
-	char id_str[4];
 	char key_str[BT_SETTINGS_KEY_MAX];
+	char id_str[4];
+	int err;
 
 	if (addr) {
 		if (id) {
 			u8_to_dec(id_str, sizeof(id_str), id);
 		}
 
-		bt_settings_encode_key(key_str, sizeof(key_str), key, addr, (id ? id_str : NULL));
+		err = bt_settings_encode_key(key_str, sizeof(key_str), key, addr,
+					     (id ? id_str : NULL));
 	} else {
-		err = snprintk(key_str, sizeof(key_str), "bt/%s", key);
-		if (err < 0) {
-			return -EINVAL;
-		}
+		err = bt_settings_encode_key_no_addr(key_str, sizeof(key_str), key);
+	}
+
+	if (err != 0) {
+		return err;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_TESTING)) {
--- a/subsys/bluetooth/host/settings.c
+++ b/subsys/bluetooth/host/settings.c
@@ -87,9 +87,9 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 
 	/* Concatenate subsys as much the free space permits */
 	strncpy(&path[len], subsys, (path_size - len));
+	len += MIN(strlen(subsys), (path_size - len));
 
 	/* Postfix '/' if there is free space */
-	len = strlen(path);
 	if (len < path_size) {
 		path[len] = '/';
 		len++;
@@ -97,8 +97,7 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 
 	/* Concatenate addr as much the free space permits */
 	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-		len += bin2hex(&addr->a.val[i], 1, &path[len],
-			       path_size - len);
+		len += bin2hex(&addr->a.val[i], 1, &path[len], (path_size - len));
 	}
 
 	/* Postfix addr type if there is free space */
@@ -114,11 +113,11 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 	if ((key != NULL) && (len < path_size)) {
 		path[len] = '/';
 		len++;
-		strncpy(&path[len], key, path_size - len);
-		len += strlen(&path[len]);
+		strncpy(&path[len], key, (path_size - len));
+		len += MIN(strlen(key), (path_size - len));
 	}
 
-	/* Insufficient path_size, include null termination */
+	/* Insufficient path_size, including null termination */
 	if (len >= path_size) {
 		return -EINVAL;
 	}
@@ -143,10 +142,10 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 	strcpy(path, "bt/");
 
 	/* Concatenate key as much the free space permits */
-	strncpy(&path[len], key, path_size - len);
-	len = strlen(path);
+	strncpy(&path[len], key, (path_size - len));
+	len += MIN(strlen(key), (path_size - len));
 
-	/* Insufficient path_size, include null termination */
+	/* Insufficient path_size, including null termination */
 	if (len >= path_size) {
 		return -EINVAL;
 	}
--- a/subsys/bluetooth/host/settings.c
+++ b/subsys/bluetooth/host/settings.c
@@ -24,7 +24,6 @@
 
 #include "common/bt_settings_commit.h"
 #include "common/bt_str.h"
-#include "common/assert.h"
 #include "hci_core.h"
 #include "settings.h"
 #include "sys/types.h"
@@ -76,8 +75,10 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 {
 	size_t len = 3;
 
-	/* path_size is less than 3; strlen("bt/") */
-	BT_ASSERT(path_size >= len);
+	/* path_size is less than or equal 3; strlen("bt/") */
+	if (path_size <= len) {
+		return -EINVAL;
+	}
 
 	/* Key format:
 	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
@@ -117,10 +118,9 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 		len += strlen(&path[len]);
 	}
 
-	/* If path string is full, always null terminate at path_size */
+	/* Insufficient path_size, include null termination */
 	if (len >= path_size) {
-		/* Truncate string */
-		path[path_size - 1] = '\0';
+		return -EINVAL;
 	}
 
 	LOG_DBG("Encoded path %s", path);
@@ -132,8 +132,10 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 {
 	size_t len = 3;
 
-	/* path_size is less than 3; strlen("bt/") */
-	BT_ASSERT(path_size >= len);
+	/* path_size is less than or equal 3; strlen("bt/") */
+	if (path_size <= len) {
+		return -EINVAL;
+	}
 
 	/* Key format:
 	 *  "bt/<key>"
@@ -144,10 +146,9 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 	strncpy(&path[len], key, path_size - len);
 	len = strlen(path);
 
-	/* If path string is full, always null terminate at path_size */
+	/* Insufficient path_size, include null termination */
 	if (len >= path_size) {
-		/* Truncate string */
-		path[path_size - 1] = '\0';
+		return -EINVAL;
 	}
 
 	LOG_DBG("Encoded path %s", path);
--- a/subsys/bluetooth/host/settings.c
+++ b/subsys/bluetooth/host/settings.c
@@ -48,7 +48,8 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 			       addr->a.val[1], addr->a.val[0], addr->type);
 	}
 
-	if (err < 0) {
+	if ((err < 0) || (err >= path_size)) {
+		/* Error or output was truncated */
 		return -EINVAL;
 	}
 
@@ -62,7 +63,8 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 	int err;
 
 	err = snprintk(path, path_size, "bt/%s", key);
-	if (err < 0) {
+	if ((err < 0) || (err >= path_size)) {
+		/* Error or output was truncated */
 		return -EINVAL;
 	}
 
@@ -73,54 +75,60 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 			   const bt_addr_le_t *addr, const char *key)
 {
-	size_t len = 3;
-
-	/* path_size is less than or equal 3; strlen("bt/") */
-	if (path_size <= len) {
-		return -EINVAL;
-	}
-
 	/* Key format:
 	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
 	 */
-	strcpy(path, "bt/");
+	const char delimiter = '/';
+	const char null_term = '\0';
+	const size_t prefix_len = sizeof("bt") - 1U;
+	const size_t subsys_len = strlen(subsys);
+	const size_t addr_str_len = sizeof("554433221100t") - 1U;
+	const size_t key_len = (key == NULL) ? 0U : strlen(key);
+	const size_t total_len = prefix_len +
+				 sizeof(delimiter) + subsys_len +
+				 sizeof(delimiter) + addr_str_len +
+				 ((key == NULL) ? 0U : (sizeof(delimiter) + key_len)) +
+				 sizeof(null_term);
+	size_t offset = 0U;
+	int err;
 
-	/* Concatenate subsys as much the free space permits */
-	strncpy(&path[len], subsys, (path_size - len));
-	len += MIN(strlen(subsys), (path_size - len));
-
-	/* Postfix '/' if there is free space */
-	if (len < path_size) {
-		path[len] = '/';
-		len++;
-	}
-
-	/* Concatenate addr as much the free space permits */
-	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-		len += bin2hex(&addr->a.val[i], 1, &path[len], (path_size - len));
-	}
-
-	/* Postfix addr type if there is free space */
-	if (len < path_size) {
-		/* Type can be either BT_ADDR_LE_PUBLIC or
-		 * BT_ADDR_LE_RANDOM (value 0 or 1)
-		 */
-		path[len] = '0' + addr->type;
-		len++;
-	}
-
-	/* Postfix '/' and concatenate key as much the free space permits */
-	if ((key != NULL) && (len < path_size)) {
-		path[len] = '/';
-		len++;
-		strncpy(&path[len], key, (path_size - len));
-		len += MIN(strlen(key), (path_size - len));
-	}
-
-	/* Insufficient path_size, including null termination */
-	if (len >= path_size) {
+	if (path_size < total_len) {
 		return -EINVAL;
 	}
+
+	memcpy(path, "bt", prefix_len);
+	offset += prefix_len;
+
+	path[offset] = delimiter;
+	offset++;
+	memcpy(&path[offset], subsys, subsys_len);
+	offset += subsys_len;
+
+	path[offset] = delimiter;
+	offset++;
+	for (int8_t i = 5; i >= 0; i--) {
+		/* We supply valid 0-15 as input */
+		err = hex2char(addr->a.val[i] >> 4, &path[offset]);
+		__ASSERT_NO_MSG(err == 0);
+		offset++;
+		/* We supply valid 0-15 as input */
+		err = hex2char(addr->a.val[i] & 0xf, &path[offset]);
+		__ASSERT_NO_MSG(err == 0);
+		offset++;
+	}
+	/* We are not checking hex2char return value as we supply valid 0-1 as input */
+	err = hex2char(addr->type, &path[offset]);
+	__ASSERT_NO_MSG(err == 0);
+	offset++;
+
+	if (key != NULL) {
+		path[offset] = delimiter;
+		offset++;
+		memcpy(&path[offset], key, key_len);
+		offset += key_len;
+	}
+
+	path[offset] = null_term;
 
 	LOG_DBG("Encoded path %s", path);
 
@@ -129,26 +137,29 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 
 static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
 {
-	size_t len = 3;
-
-	/* path_size is less than or equal 3; strlen("bt/") */
-	if (path_size <= len) {
-		return -EINVAL;
-	}
-
 	/* Key format:
 	 *  "bt/<key>"
 	 */
-	strcpy(path, "bt/");
+	const char delimiter = '/';
+	const char null_term = '\0';
+	const size_t prefix_len = sizeof("bt") - 1U;
+	const size_t key_len = strlen(key);
+	const size_t total_len = prefix_len + sizeof(delimiter) + key_len + sizeof(null_term);
+	size_t offset = 0U;
 
-	/* Concatenate key as much the free space permits */
-	strncpy(&path[len], key, (path_size - len));
-	len += MIN(strlen(key), (path_size - len));
-
-	/* Insufficient path_size, including null termination */
-	if (len >= path_size) {
+	if (path_size < total_len) {
 		return -EINVAL;
 	}
+
+	memcpy(path, "bt", prefix_len);
+	offset += prefix_len;
+
+	path[offset] = delimiter;
+	offset++;
+	memcpy(&path[offset], key, key_len);
+	offset += key_len;
+
+	path[offset] = null_term;
 
 	LOG_DBG("Encoded path %s", path);
 
@@ -202,7 +213,7 @@ static int set_setting(const char *name, size_t len_rd, settings_read_cb read_cb
 
 	len = settings_name_next(name, &next);
 
-	if (!strncmp(name, "id", len)) {
+	if ((len == 2) && (memcmp(name, "id", len) == 0)) {
 		/* Any previously provided identities supersede flash */
 		if (atomic_test_bit(bt_dev.flags, BT_DEV_PRESET_ID)) {
 			LOG_WRN("Ignoring identities stored in flash");
@@ -234,7 +245,7 @@ static int set_setting(const char *name, size_t len_rd, settings_read_cb read_cb
 	}
 
 #if defined(CONFIG_BT_DEVICE_NAME_DYNAMIC)
-	if (!strncmp(name, "name", len)) {
+	if ((len == 4) && (memcmp(name, "name", len) == 0)) {
 		len = read_cb(cb_arg, &bt_dev.name, sizeof(bt_dev.name) - 1);
 		if (len < 0) {
 			LOG_ERR("Failed to read device name from storage"
@@ -249,7 +260,7 @@ static int set_setting(const char *name, size_t len_rd, settings_read_cb read_cb
 #endif
 
 #if defined(CONFIG_BT_DEVICE_APPEARANCE_DYNAMIC)
-	if (!strncmp(name, "appearance", len)) {
+	if ((len == 10) && (memcmp(name, "appearance", len) == 0)) {
 		if (len_rd != sizeof(bt_dev.appearance)) {
 			LOG_ERR("Ignoring settings entry 'bt/appearance'. Wrong length.");
 			return -EINVAL;
@@ -265,7 +276,7 @@ static int set_setting(const char *name, size_t len_rd, settings_read_cb read_cb
 #endif
 
 #if defined(CONFIG_BT_PRIVACY)
-	if (!strncmp(name, "irk", len)) {
+	if ((len == 3) && (memcmp(name, "irk", len) == 0)) {
 		len = read_cb(cb_arg, bt_dev.irk, sizeof(bt_dev.irk));
 		if (len < sizeof(bt_dev.irk[0])) {
 			if (len < 0) {
--- a/subsys/bluetooth/host/settings.c
+++ b/subsys/bluetooth/host/settings.c
@@ -24,6 +24,7 @@
 
 #include "common/bt_settings_commit.h"
 #include "common/bt_str.h"
+#include "common/assert.h"
 #include "hci_core.h"
 #include "settings.h"
 #include "sys/types.h"
@@ -75,45 +76,51 @@ int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 {
 	size_t len = 3;
 
-	/* Skip if path_size is less than 3; strlen("bt/") */
+	/* path_size is less than 3; strlen("bt/") */
+	BT_ASSERT(path_size >= len);
+
+	/* Key format:
+	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+	 */
+	strcpy(path, "bt/");
+
+	/* Concatenate subsys as much the free space permits */
+	strncpy(&path[len], subsys, (path_size - len));
+
+	/* Postfix '/' if there is free space */
+	len = strlen(path);
 	if (len < path_size) {
-		/* Key format:
-		 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+		path[len] = '/';
+		len++;
+	}
+
+	/* Concatenate addr as much the free space permits */
+	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
+		len += bin2hex(&addr->a.val[i], 1, &path[len],
+			       path_size - len);
+	}
+
+	/* Postfix addr type if there is free space */
+	if (len < path_size) {
+		/* Type can be either BT_ADDR_LE_PUBLIC or
+		 * BT_ADDR_LE_RANDOM (value 0 or 1)
 		 */
-		strcpy(path, "bt/");
-		strncpy(&path[len], subsys, path_size - len);
-		len = strlen(path);
-		if (len < path_size) {
-			path[len] = '/';
-			len++;
-		}
+		path[len] = '0' + addr->type;
+		len++;
+	}
 
-		for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-			len += bin2hex(&addr->a.val[i], 1, &path[len],
-				       path_size - len);
-		}
+	/* Postfix '/' and concatenate key as much the free space permits */
+	if ((key != NULL) && (len < path_size)) {
+		path[len] = '/';
+		len++;
+		strncpy(&path[len], key, path_size - len);
+		len += strlen(&path[len]);
+	}
 
-		if (len < path_size) {
-			/* Type can be either BT_ADDR_LE_PUBLIC or
-			 * BT_ADDR_LE_RANDOM (value 0 or 1)
-			 */
-			path[len] = '0' + addr->type;
-			len++;
-		}
-
-		if (key && len < path_size) {
-			path[len] = '/';
-			len++;
-			strncpy(&path[len], key, path_size - len);
-			len += strlen(&path[len]);
-		}
-
-		if (len >= path_size) {
-			/* Truncate string */
-			path[path_size - 1] = '\0';
-		}
-	} else if (path_size > 0) {
-		*path = '\0';
+	/* If path string is full, always null terminate at path_size */
+	if (len >= path_size) {
+		/* Truncate string */
+		path[path_size - 1] = '\0';
 	}
 
 	LOG_DBG("Encoded path %s", path);
@@ -125,26 +132,22 @@ static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const ch
 {
 	size_t len = 3;
 
-	/* Skip if path_size is less than 3; strlen("bt/") */
-	if (len < path_size) {
-		/* Key format:
-		 *  "bt/<key>", "/<key>" is optional
-		 */
-		strcpy(path, "bt/");
+	/* path_size is less than 3; strlen("bt/") */
+	BT_ASSERT(path_size >= len);
 
-		/* Concatenate key as much the free space permits */
-		strncpy(&path[len], key, path_size - len);
-		len = strlen(path);
+	/* Key format:
+	 *  "bt/<key>"
+	 */
+	strcpy(path, "bt/");
 
-		/* If path string is full, always null terminate at path_size */
-		if (len >= path_size) {
-			/* Truncate string */
-			path[path_size - 1] = '\0';
-		}
+	/* Concatenate key as much the free space permits */
+	strncpy(&path[len], key, path_size - len);
+	len = strlen(path);
 
-	} else if (path_size > 0) {
-		/* path_size not sufficient for "bt/" */
-		*path = '\0';
+	/* If path string is full, always null terminate at path_size */
+	if (len >= path_size) {
+		/* Truncate string */
+		path[path_size - 1] = '\0';
 	}
 
 	LOG_DBG("Encoded path %s", path);
@@ -375,10 +378,11 @@ int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, con
 		      size_t val_len)
 {
 	char key_str[BT_SETTINGS_KEY_MAX];
-	char id_str[4];
 	int err;
 
 	if (addr) {
+		char id_str[4];
+
 		if (id) {
 			u8_to_dec(id_str, sizeof(id_str), id);
 		}
@@ -403,10 +407,11 @@ int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, con
 int bt_settings_delete(const char *key, uint8_t id, const bt_addr_le_t *addr)
 {
 	char key_str[BT_SETTINGS_KEY_MAX];
-	char id_str[4];
 	int err;
 
 	if (addr) {
+		char id_str[4];
+
 		if (id) {
 			u8_to_dec(id_str, sizeof(id_str), id);
 		}
--- a/subsys/bluetooth/host/settings.h
+++ b/subsys/bluetooth/host/settings.h
@@ -32,7 +32,7 @@ void bt_testing_settings_store_hook(const char *key, const void *value, size_t v
 void bt_testing_settings_delete_hook(const char *key);
 
 /* Helpers for keys containing a bdaddr */
-void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
+int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
 			    const bt_addr_le_t *addr, const char *key);
 int bt_settings_decode_key(const char *key, bt_addr_le_t *addr);
 
--- a/tests/bsim/bluetooth/host/privacy/peripheral/prj_no_snprintk.conf
+++ b/tests/bsim/bluetooth/host/privacy/peripheral/prj_no_snprintk.conf
@@ -18,8 +18,8 @@ CONFIG_NVS=y
 CONFIG_SETTINGS=y
 CONFIG_BT_SETTINGS=y
 
-# Lets cover settings with use of snprintk
-CONFIG_BT_SETTINGS_USE_PRINTK=y
+# Lets cover settings without use of snprintk
+CONFIG_BT_SETTINGS_USE_PRINTK=n
 
 # Increased stack due to settings API usage
 CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
--- a/tests/bsim/bluetooth/host/privacy/peripheral/test_scripts/run_test_no_snprintk.sh
+++ b/tests/bsim/bluetooth/host/privacy/peripheral/test_scripts/run_test_no_snprintk.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# Copyright 2023 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu
+source ${ZEPHYR_BASE}/tests/bsim/sh_common.source
+
+verbosity_level=2
+simulation_id="host_privacy_peripheral_no_snprintk"
+EXECUTE_TIMEOUT=240
+
+central_exe="${BSIM_OUT_PATH}/bin/bs_${BOARD_TS}_$(guess_test_long_name)_prj_no_snprintk_conf"
+peripheral_exe="${central_exe}"
+
+cd ${BSIM_OUT_PATH}/bin
+
+Execute "$central_exe" \
+    -v=${verbosity_level} -s=${simulation_id} -d=0 -testid=central -RealEncryption=1 \
+    -flash="${simulation_id}.central.log.bin" -flash_erase
+
+Execute "$peripheral_exe" \
+    -v=${verbosity_level} -s=${simulation_id} -d=1 -testid=peripheral -RealEncryption=1 \
+    -flash="${simulation_id}.peripheral.log.bin" -flash_erase
+
+Execute ./bs_2G4_phy_v1 -v=${verbosity_level} -s=${simulation_id} \
+    -D=2 -sim_length=70e6 $@
+
+wait_for_background_jobs
+
+Execute "$central_exe" \
+    -v=${verbosity_level} -s=${simulation_id}.2 -d=0 -testid=central -RealEncryption=1 \
+    -flash="${simulation_id}.central.log.bin" -flash_rm
+
+Execute "$peripheral_exe" \
+    -v=${verbosity_level} -s=${simulation_id}.2 -d=1 -testid=peripheral -RealEncryption=1 \
+    -flash="${simulation_id}.peripheral.log.bin" -flash_rm
+
+Execute ./bs_2G4_phy_v1 -v=${verbosity_level} -s=${simulation_id}.2 \
+    -D=2 -sim_length=70e6 $@
+
+wait_for_background_jobs
--- a/tests/bsim/bluetooth/host/privacy/peripheral/testcase.yaml
+++ b/tests/bsim/bluetooth/host/privacy/peripheral/testcase.yaml
@@ -12,6 +12,11 @@ tests:
   bluetooth.host.privacy.peripheral:
     harness_config:
       bsim_exe_name: tests_bsim_bluetooth_host_privacy_peripheral_prj_conf
+  bluetooth.host.privacy.peripheral_no_snprintk:
+    harness_config:
+      bsim_exe_name: tests_bsim_bluetooth_host_privacy_peripheral_prj_no_snprintk_conf
+    extra_args:
+      EXTRA_CONF_FILE=prj_no_snprintk.conf
   bluetooth.host.privacy.peripheral_rpa_expired:
     harness_config:
       bsim_exe_name: tests_bsim_bluetooth_host_privacy_peripheral_prj_rpa_expired_conf


subsys/bluetooth/host/att.c
```diff
@@ -385,8 +384,13 @@ static void att_disconnect(struct bt_att_chan *chan)
 	 * invalid
 	 */
 
-	bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
-	LOG_DBG("ATT disconnecting device %s", addr);
+	if (IS_ENABLED(CONFIG_LOG)) {
```
alwa-nordic: Just curious: Isn't this automatically optimized like this already?

eHammarstrom: Dead code elimination (DCE) should handle this since addr is unused once LOG_DBG is a no-op.

samples/bluetooth/peripheral_hids/src/main.c
Thalley, line 1: I would actually suggest for the sample to use the logging system instead of printk, but I'd probably make the printks into LOG_INF

subsys/bluetooth/host/settings.c
```diff
@@ -33,28 +33,45 @@
...
-			 addr->a.val[5], addr->a.val[4], addr->a.val[3],
-			 addr->a.val[2], addr->a.val[1], addr->a.val[0],
-			 addr->type);
+		err = snprintk(path, path_size, "bt/%s/%02x%02x%02x%02x%02x%02x%u", subsys,
+			       addr->a.val[5], addr->a.val[4], addr->a.val[3], addr->a.val[2],
+			       addr->a.val[1], addr->a.val[0], addr->type);
+	}
+
+	if (err < 0) {
+		return -EINVAL;
```
Thalley:
```suggestion
	if (err < 0) {
		return err;
```

Wouldn't it make sense to return the (errno) error from `snprintk`?

subsys/bluetooth/host/settings.c
```diff
@@ -33,28 +33,45 @@
...
-			    const bt_addr_le_t *addr, const char *key)
+
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	int err;
+
+	err = snprintk(path, path_size, "bt/%s", key);
+	if (err < 0) {
+		return -EINVAL;
+	}
```
Thalley, line 69: Ditto

subsys/bluetooth/host/settings.c
```diff
@@ -324,21 +374,23 @@ __weak void bt_testing_settings_delete_hook(const char *key)
...
 
-		bt_settings_encode_key(key_str, sizeof(key_str), key, addr, (id ? id_str : NULL));
+		err = bt_settings_encode_key(key_str, sizeof(key_str), key, addr,
+					     (id ? id_str : NULL));
 	} else {
-		err = snprintk(key_str, sizeof(key_str), "bt/%s", key);
-		if (err < 0) {
-			return -EINVAL;
-		}
+		err = bt_settings_encode_key_no_addr(key_str, sizeof(key_str), key);
```
Thalley, line 404: I'm OK with 2 functions, but an alternative could also be to just have a single function that takes an `addr` which could be NULL and handle it there

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	size_t len = 3;
+
+	/* Skip if path_size is less than 3; strlen("bt/") */
+	if (len < path_size) {
+		/* Key format:
+		 *  "bt/<key>", "/<key>" is optional
+		 */
+		strcpy(path, "bt/");
```
Thalley: We should consider putting `"bt/"` into a macro or something; it's hardcoded so many places :(

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+	return 0;
 }
-#endif
+
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	size_t len = 3;
+
+	/* Skip if path_size is less than 3; strlen("bt/") */
+	if (len < path_size) {
```
Thalley: Shouldn't this return an error or assert instead of silently failing?

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+		/* Key format:
+		 *  "bt/<key>", "/<key>" is optional
+		 */
```
Thalley: This doesn't make sense to me. Why would we ever store just `"bt/"?` It makes sense above in `bt_settings_encode_key` since there's an address there as well, but here it doesn't really make sense IMO. Additionally the PRINTK version of `bt_settings_encode_key_no_addr` doesn't allow key to be optional, right?

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+		strcpy(path, "bt/");
+
+		/* Concatenate key as much the free space permits */
+		strncpy(&path[len], key, path_size - len);
+		len = strlen(path);
+
+		/* If path string is full, always null terminate at path_size */
+		if (len >= path_size) {
+			/* Truncate string */
```
Thalley: We should do the length check before the copy, and I don't see why we would ever truncate a key? :o Wouldn't that break things?

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+			path[path_size - 1] = '\0';
+		}
```
Thalley: If the path_size is larger than the prefix and key, shouldn't we NULL terminate right after the key, instead of this fixed index?

Thalley (changes requested):

subsys/bluetooth/host/att.c
cvinayak: Evidently that is not the case in the use of `bt_addr_le_to_str` :( ... may someone with better knowledge why the lst file and binary has the code not eliminated ?

Image

samples/bluetooth/peripheral_hids/src/main.c
cvinayak: Sample shall not use LOG subsystem, they shall use the kernel interface for system console output. Logging subsystem is optional, and not immediate for a console user interaction.

```
 * @brief Print kernel debugging message.
 *
 * This routine prints a kernel debugging message to the system console.
 * Output is send immediately, without any mutual exclusion or buffering.

```

subsys/bluetooth/host/settings.c
cvinayak: Made it consistent with old code where `snprintk` was called 
https://github.com/zephyrproject-rtos/zephyr/pull/100887/changes#diff-c0d733611772ea4600bb50136c8a70402314c69f6e20ef8f245d482382a2917fL339-L341

subsys/bluetooth/host/settings.c
cvinayak: This the old behavior.
https://github.com/zephyrproject-rtos/zephyr/pull/100887/changes#diff-c0d733611772ea4600bb50136c8a70402314c69f6e20ef8f245d482382a2917fL339-L341

subsys/bluetooth/host/settings.c
cvinayak: Kept the old behavior. Do not mind fixing if a suggestion is provided, and the purpose of the PR is maintained.

subsys/bluetooth/host/settings.c
cvinayak: Again kept the old behavior where `key` was supplied the `snprintk` unconditionally.

subsys/bluetooth/host/settings.c
cvinayak: Was my first thought too... but did not want to deviate too much from the scope and purpose of the PR.

subsys/bluetooth/host/settings.c
cvinayak: The code is taking advantage of `strncpy` and avoid added branching code.

Agree on the truncation part tho... again this file is a can of worms.

subsys/bluetooth/host/settings.c
cvinayak: Agree... will need to take this up as separate commit or PR tho.

subsys/bluetooth/host/settings.c
cvinayak: I did push a draft in this PR with both implementation, and the code size with ~ 20 bytes more due to added branching instructions to check for NULL, and to keep old behavior of optional parameters and the string output on truncation etc.

cvinayak (commented):
The scope in this PR being:
1. settings to truely respect CONFIG_BT_SETTINGS_USE_PRINTK and not use `snprintk`
  - Original PR: https://github.com/zephyrproject-rtos/zephyr/pull/19208
2. why is `snprintk` and its caller code (bt_addr_le_to_str) not eliminated when `CONFIG_LOG=n` ?

subsys/bluetooth/host/att.c
cvinayak: Use of `va_list`  in `vsnprintk` ?

image

Copilot GPT-5:
image

cvinayak:
@wopu-ot Original contribution from Oticon, will be good to get some of your reviewers :-)

samples/bluetooth/peripheral_hids/src/main.c
Thalley: > Sample shall not use LOG subsystem, they shall use the kernel interface for system console output. 

Was not aware of that. Where is that requirement documented?

subsys/bluetooth/host/settings.c
Thalley: Can you elaborate? The old behavior was `err = snprintk(key_str, sizeof(key_str), "bt/%s", key);`, so not sure how that relates to this.

subsys/bluetooth/host/settings.c
Thalley: The old behavior was `err = snprintk(key_str, sizeof(key_str), "bt/%s", key);`; if `key == NULL` then that would fail, so isn't this different than the old behavior?

subsys/bluetooth/host/settings.c
Thalley: Yeah, that change is better for a different PR IMO

subsys/bluetooth/host/settings.c
Thalley: > The code is taking advantage of strncpy and avoid added branching code.

It's not much of an advantage if the code will instead copy memory before doing the check, is it? :) Since you are already doing `strlen(path)`, then it's pretty trivial to change the first `if (len < path_size) {` to be `len + key_len < path_size`.

Additionally, keys cannot be any size, so there should also be a separate key_len check, right, to ensure it's between 7 and 16 octets when decoded (or is that done somewhere else?)

subsys/bluetooth/host/settings.c
Thalley: Why should that be in a separate PR, when the offending code is being added by this PR? 

I know the code is heavily inspired by `bt_settings_encode_key`, but I don't see any reasons to copy and keep the bugs from that (the bugs in `bt_settings_encode_key` could even be fixed as part of this PR instead)

subsys/bluetooth/host/settings.c
```diff
@@ -100,8 +117,41 @@ void bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
...
+
+	} else if (path_size > 0) {
+		/* path_size not sufficient for "bt/" */
+		*path = '\0';
+	}
```
Thalley: Should this be an assert? Seems like a programming issue, rather than a settings-content issue

Thalley (changes requested):

samples/bluetooth/peripheral_hids/src/main.c
cvinayak: Not directly, but the architecture port stating "almost a requirement" to send the console (`printk`, `printf`) output over serial line or RAM console.
https://docs.zephyrproject.org/latest/hardware/porting/arch.html#console-over-serial-line

subsys/bluetooth/host/settings.c
cvinayak: I mean with the caller calling with `path_size` of `BT_SETTINGS_KEY_MAX` (hard coded 36) bytes. And, re-using prior reviewed such implementation in `bt_settings_encode_key()`.

subsys/bluetooth/host/settings.c
cvinayak: i mean, replacement of `snprintk`  use for `bt_settings_encode_key()` has check for `key`  being `NULL`  and `bt_settings_encode_key()` stated `key` can be optional.

subsys/bluetooth/host/settings.c
cvinayak: `strncpy` will not copy memory (and already has check to ensure so, that `path_size - len > 0`).

> Additionally, keys cannot be any size, so there should also be a separate key_len check, right, to ensure it's between 7 and 16 octets when decoded (or is that done somewhere else?)

That should be in SMP?

subsys/bluetooth/host/settings.c
cvinayak: The implementation here is trying to meet: `The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.`  definition of these function.

subsys/bluetooth/host/settings.c
cvinayak: I do not have the big picture of the use of settings module how the read-back handles the values stored.

samples/bluetooth/peripheral_hids/src/main.c
Thalley: Isn't that just a matter of configuration of the logging system? I don't read the above as a requirement that `printk` shall be supported, and `LOG` shall not.

subsys/bluetooth/host/settings.c
Thalley: Interesting. You'd have the same amount of branching, so not sure why it would be 20 bytes more due to that

subsys/bluetooth/host/settings.c
Thalley: I don't approve of the argument that if an existing function has a bug or is poorly implemented, that new functions should not be properly implemented. We should improve code quality IMO, and perform the necessary checks IMO (and maybe even fix the existing code when we see these issues)

subsys/bluetooth/host/settings.c
Thalley: The key is optional in `bt_settings_encode_key` possibly because the function still encodes the address, so even when the key is `NULL` it still stores _some_ information in the settings. That's not the case here where `key` is the only value to be stored, so I don't think the behavior of the 2 functions can be compared to whether they should allow `key` to be NULL.

subsys/bluetooth/host/settings.c
Thalley: > `strncpy` will not copy memory (and already has check to ensure so, that `path_size - len > 0`).

if `strlen(key) = 100` (some invalid data), then you'd still copy 33 octets (`BT_SETTINGS_KEY_MAX - sizeof("bt/")` for no reason before actually checking if `key` has a valid length for a key. 

> That should be in SMP?

Maybe. The keysize is configurable with `CONFIG_BT_SMP_MIN_ENC_KEY_SIZE`. In the case that we have BUILD_A with `CONFIG_BT_SMP_MIN_ENC_KEY_SIZE=10` and actually store some keys of length 10, and then have BUILD_B with `CONFIG_BT_SMP_MIN_ENC_KEY_SIZE=16`, then how do we want to handle that? Do we failing init? Do we ignore those keys? 

In any case, if there are keys of size < `BT_HCI_ENCRYPTION_KEY_SIZE_MIN` or > `BT_HCI_ENCRYPTION_KEY_SIZE_MAX` then those should be considered as invalid immediately, right? But that's possibly something that SMP should be doing, if not done already, but my point is that if these exist in the settings, why even both copying them?

subsys/bluetooth/host/settings.c
Thalley: I don't think this implement meets that. 

If `strlen(key) = 16` then 
```
[0..2] = "bt/"
[3..18] = key
[19..34] = initialized data
[35] = '\0'
```

where the snprintk version would be 
```
[0..2] = "bt/"
[3..18] = key
[19] = '\0'
[20..35] = initialized data
```

Right?

subsys/bluetooth/host/settings.c
Thalley: My point is that if the `path_size` is smaller than `"bt/"` then something has gone wrong in programming, as both values are hardcoded. That has nothing to do with the actual information in settings

subsys/bluetooth/host/att.c
Thalley: If you do want to keep this guard, shouldn't it be `CONFIG_BT_ATT_LOG_LEVEL_DBG` instead?

samples/bluetooth/peripheral_hids/src/main.c
cvinayak: `printk` is the kernel feature for printing to the console.

Example: https://docs.zephyrproject.org/latest/samples/hello_world/README.html#hello_world

Logging subsystem should not be mandatory for samples when all that is needed is to print characters to the system console.

subsys/bluetooth/host/settings.c
cvinayak: The code only sets `path[path_size - 1] = '\0'`  if `strlen(path)`   is equal to `path_size`... the length is not going to exceed `path_size`  as `strncpy` is used.

subsys/bluetooth/host/att.c
cvinayak: Done.

subsys/bluetooth/host/settings.c
cvinayak: Added BT_ASSERT

subsys/bluetooth/host/settings.c
cvinayak: Updated the comment.

subsys/bluetooth/host/settings.c
cvinayak: The addition code is needed in `bt_settings_encode_key` to handle `addr` being NULL. In which case the string construction is different.

Commit for reference: https://github.com/zephyrproject-rtos/zephyr/pull/100887/changes/7d3489b251c152cbcb8e7953ad121a0229a42e4b

samples/bluetooth/peripheral_hids/src/main.c
Thalley: It's a bit odd change in any case. In no other other sample are we doing `if (IS_ENABLED(CONFIG_PRINTK)) {`, and we are also using the logging system in more than a  hundred samples

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
 	size_t len = 3;
 
-	/* Skip if path_size is less than 3; strlen("bt/") */
+	/* path_size is less than 3; strlen("bt/") */
+	BT_ASSERT(path_size >= len);
+
+	/* Key format:
+	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+	 */
+	strcpy(path, "bt/");
```
Thalley: Possibly outside the scope of this PR, but I guess we could initialize the buffer to have `"bt/", so we don't need to have 2 places where we copy it to the buffer?

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
+	len = strlen(path);
 	if (len < path_size) {
-		/* Key format:
-		 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+		path[len] = '/';
+		len++;
+	}
+
+	/* Concatenate addr as much the free space permits */
+	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
```
Thalley: Why `int8_t`?

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
+		len += bin2hex(&addr->a.val[i], 1, &path[len],
+			       path_size - len);
+	}
```
Thalley: If the idea behind this is to not use `snprintk`, shouldn't `bt_addr_to_str` be updated to do this instead in those cases, so that we can just use `bt_addr_to_str` here?

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
+		path[len] = '0' + addr->type;
+		len++;
+	}
 
-		for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-			len += bin2hex(&addr->a.val[i], 1, &path[len],
-				       path_size - len);
-		}
+	/* Postfix '/' and concatenate key as much the free space permits */
+	if (key && len < path_size) {
```
Thalley:
```suggestion
	if (key != NULL && len < path_size) {
```

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
+		 * BT_ADDR_LE_RANDOM (value 0 or 1)
 		 */
-		strcpy(path, "bt/");
-		strncpy(&path[len], subsys, path_size - len);
-		len = strlen(path);
-		if (len < path_size) {
-			path[len] = '/';
-			len++;
-		}
+		path[len] = '0' + addr->type;
```
Thalley:
```suggestion
		hex2char(addr->type, path[len]);
```

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
-			 * BT_ADDR_LE_RANDOM (value 0 or 1)
-			 */
-			path[len] = '0' + addr->type;
-			len++;
-		}
+	/* If path string is full, always null terminate at path_size */
+	if (len >= path_size) {
+		/* Truncate string */
+		path[path_size - 1] = '\0';
+	}
```
Thalley: There is where I think we need to do something different. If the buffer isn't large enough to hold the content, we shouldn't just silently truncate it

subsys/bluetooth/host/settings.c
```diff
@@ -324,21 +376,23 @@ __weak void bt_testing_settings_delete_hook(const char *key)
 int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, const void *value,
 		      size_t val_len)
 {
-	int err;
-	char id_str[4];
 	char key_str[BT_SETTINGS_KEY_MAX];
+	char id_str[4];
```
Thalley: You can move the `id_str` to the `if (addr)` clause to reduce scope

subsys/bluetooth/host/settings.c
```diff
@@ -350,21 +404,23 @@ int bt_settings_store(const char *key, uint8_t id, const bt_addr_le_t *addr, con
 
 int bt_settings_delete(const char *key, uint8_t id, const bt_addr_le_t *addr)
 {
-	int err;
-	char id_str[4];
 	char key_str[BT_SETTINGS_KEY_MAX];
+	char id_str[4];
```
Thalley: Ditto can move to reduce scope

Thalley (changes requested):

subsys/bluetooth/host/att.c
eHammarstrom: It's a tricky one, and it does not seem to stem from `strcpy` or `snprintk` as I still get `bt_conn_get_dst` compiled in even though that one looks free of side-effects.

```
// CONFIG_LOG=n
// CONFIG_BOOT_BANNER=n
// CONFIG_PRINTK=n

static int __attribute__((noinline)) aprint(void)
{
	char addr[BT_ADDR_LE_STR_LEN];
	struct bt_conn *conn = NULL; // bad bad

	bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
    return 0;
}

000093f2 <aprint.isra.0>:
static int __attribute__((noinline)) aprint(void)
{
	char addr[BT_ADDR_LE_STR_LEN];
	struct bt_conn *conn = NULL;

	bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
    93f2:	2000      	movs	r0, #0
    93f4:	f001 b8bc 	b.w	a570 <bt_conn_get_dst>

000093f8 <main>:

	return 0;
}
```

samples/bluetooth/peripheral_hids/src/main.c
cvinayak: My intention for the commit with the sample changed was to demonstrate that code elimination is not applied in all cases. I do not wish to merge the sample change. But code reduction in subsystem code is necessary where LOG is used. I will remove the sample change commit once the review is approved for other changes (subsystem ones).

samples/bluetooth/peripheral_hids/src/main.c
```diff
@@ -37,45 +37,54 @@ static const struct bt_data sd[] = {
...
 	if (err) {
-		printk("Failed to connect to %s, err 0x%02x %s\n", addr,
-		       err, bt_hci_err_to_str(err));
 		return;
 	}
 
-	printk("Connected %s\n", addr);
-
 	if (bt_conn_set_security(conn, BT_SECURITY_L2)) {
 		printk("Failed to set security\n");
```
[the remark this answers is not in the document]
cvinayak: Only implementation with code reduction opportunity is conditional compiled.

subsys/bluetooth/host/settings.c
cvinayak: we want `5`..`0` as the index and `-1` as the exit criteria. If using unsigned int, we will end always `i >= 0` true.

subsys/bluetooth/host/settings.c
cvinayak: all that is needed is `0` or `1`... calling a function is overhead here.

subsys/bluetooth/host/settings.c
cvinayak: done

subsys/bluetooth/host/settings.c
cvinayak: Should I replace with `BT_ASSERT(len < path_size)` ?

subsys/bluetooth/host/settings.c
cvinayak: done

subsys/bluetooth/host/settings.c
cvinayak: done

subsys/bluetooth/host/settings.c
cvinayak: `bt_addr_to_str`  has `:` between the octets, functionally different.

subsys/bluetooth/host/settings.c
cvinayak: This can be taken up as part of a separate PR. At this point, I can only think of a `#define BT_SETTINGS_STR "bt/"` and use the define.

subsys/bluetooth/host/settings.c
Thalley: Ah, right, we want to write it starting from val[5]

subsys/bluetooth/host/settings.c
Thalley: Fair point, although if an application is using `bt_addr_to_str` already then `snprintk` is already linked right, and we probably don't save anything from these `CONFIG_BT_SETTINGS_USE_PRINTK=n` functions, right?

subsys/bluetooth/host/settings.c
Thalley: True in terms of performance, but won't it save the `'0'` in memory?

subsys/bluetooth/host/settings.c
Thalley: Something like that. I really think that we shouldn't silently truncate data like this. Not sure if a runtime error (returning an error to the settings subsystem) or an assert is best

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,127 @@
...
+	strcpy(path, "bt/");
+
+	/* Concatenate key as much the free space permits */
+	strncpy(&path[len], key, path_size - len);
+	len = strlen(path);
+
+	/* If path string is full, always null terminate at path_size */
+	if (len >= path_size) {
+		/* Truncate string */
+		path[path_size - 1] = '\0';
```
Thalley: if `path` is uninitialized we should null terminate at `path[len]`, which seems to be missing, and we should not silently truncate data like this

subsys/bluetooth/host/settings.c
cvinayak: Correct. With the changes in this PR, applications that do not use directly/indirectly `snprintk` can use this Kconfig to reduce code size.

subsys/bluetooth/host/settings.c
cvinayak: `0` (immediate value upto 5 bits) are part of ARM thumb instruction and do not need extra code space. With  and without `hex2char` use the code space is same, but increased CPU cycle due to branching (instruction pipeline being flushed, may be branch prediction in CPUs helps?).

Assignment statement is:
```
 8355                 path[len] = '0' + addr->type;                                                       
 8356     330e:       782b            ldrb    r3, [r5, #0]                                                
 8357     3310:       3330            adds    r3, #48 ; 0x30                                              
 8358     3312:       553b            strb    r3, [r7, r4]                                                
```

With `hex2char` reused (as used in `bin2hex`):
```
 8355                 hex2char(addr->type, &path[len]);                                                   
 8356     330c:       f898 0000       ldrb.w  r0, [r8]                                                    
 8357     3310:       1931            adds    r1, r6, r4                                                  
 8358     3312:       f018 fcc4       bl      1bc9e <hex2char>                                            
```

subsys/bluetooth/host/settings.c
cvinayak: Use BT_ASSERT increases code size, replaced with returning error.

subsys/bluetooth/host/settings.c
cvinayak: `len` atmost is `path_size`. Size of `path` is `path_size`. If full, then `path_size - 1` gets the `\0`.

subsys/bluetooth/host/settings.c
Thalley: Right, but `bt_addr_str` calls `bt_addr_to_str`, and the host uses `bt_addr_str` several places for LOG_DBG and even LOG_INF, so chances are that `snprintk` is linked in any case. 

My point here is that this optimization, while valid, may not really be applied if the host, controller, or any other subsystem is enabled, uses `snprintk` and may not be visible to the application, and for this optimization to really have value (at least in a Bluetooth perspective), we kind of need to remove all the other places too. 

If we generally want to avoid `snprintk` for optimization purposes in the stack, we should probably also document that somewhere for Bluetooth at least. 

Question: if the application uses `snprintk` already for something, does disabling `CONFIG_BT_SETTINGS_USE_PRINTK` keep the same code size, or does it increase?

subsys/bluetooth/host/settings.c
Thalley: > 0 (immediate value upto 5 bits) are part of ARM thumb instruction and do not need extra code space.

What about other platforms/architectures? 

What about the CPU cycles between the `snprintk` version and the non-`snprintk` functions? If the purpose of this PR is to optimize for memory, why not stick with that approach in a general manner? 

(btw, I realize this is a pedantic comment/question, and is by no means a blocker)

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,129 @@
...
+int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
+			   const bt_addr_le_t *addr, const char *key)
 {
 	size_t len = 3;
 
-	/* Skip if path_size is less than 3; strlen("bt/") */
+	/* path_size is less than or equal 3; strlen("bt/") */
+	if (path_size <= len) {
+		return -EINVAL;
+	}
```
Thalley: Tbh I'd do something like

```suggestion
	if (path_size <= (sizeof("bt/") - 1) + strlen(path) + (sizeof("00000000000000") - 1)) {
		return -EINVAL;
	}
```

i.e. early check and termination if the path cannot fit the content. That way we avoid any unnecessary data copies, as well as skipping any needs for truncate checks later, and can probably rely on `strcpy` over `strncpy` which is probably an additional optimization. This of course assumes that `key` cannot be modified by another thread while this function is being done. 

Ditto for the other function

Thalley: The 2 sizeof could probably be defined as constants or defines for the sake of readability

subsys/bluetooth/host/settings.c
cvinayak: > Question: if the application uses `snprintk` already for something, does disabling `CONFIG_BT_SETTINGS_USE_PRINTK` keep the same code size, or does it increase?

Code size will increase marginally, 124 bytes.

`DCONFIG_PRINTK=y -DCONFIG_BT_SETTINGS_USE_PRINTK=n`  ~5KB increase.

```
           FLASH:      160076 B      1428 KB     10.95%
             RAM:       27428 B       188 KB     14.25%
        IDT_LIST:          0 GB        32 KB      0.00%
```

`-DCONFIG_PRINTK=y -DCONFIG_BT_SETTINGS_USE_PRINTK=y` ~5KB increase. Reusing `snprintk`  saves 124  bytes. 

```
           FLASH:      159952 B      1428 KB     10.94%
             RAM:       27428 B       188 KB     14.25%
        IDT_LIST:          0 GB        32 KB      0.00%
```

`-DCONFIG_PRINTK=n -DCONFIG_BT_SETTINGS_USE_PRINTK=n` ~ 2KB less compare to main.

```
           FLASH:      155012 B      1428 KB     10.60%
             RAM:       27420 B       188 KB     14.24%
        IDT_LIST:          0 GB        32 KB      0.00%
```

`-DCONFIG_PRINTK=n -DCONFIG_BT_SETTINGS_USE_PRINTK=n` on base branch:

```
           FLASH:      157012 B      1428 KB     10.74%
             RAM:       27420 B       188 KB     14.24%
        IDT_LIST:          0 GB        32 KB      0.00%
```

subsys/bluetooth/host/settings.c
cvinayak: > Tbh I'd do something like

If the failure was `likely` control path, then yes, failing early is fair. But the error checking is for `unlikely` control path, and hence, failing late with least code space used is efficient.

subsys/bluetooth/host/settings.c
cvinayak: > What about other platforms/architectures?

Original contribution was a non-ARM architecture, hence, atleast two architectures are efficient without branching in this scenario.

> why not stick with that approach in a general manner?

As seen from builds, it is significant code space increase (format strings for `snprintk` take code space in addition to `snprintk` 's own code use). Simple specific concatenation save on the heavy caller and callee code use by `snprintk`. It may be balanced if there are more module specific concatenations. For a resource constraints device, users be given the option to optimize application with less forking of the subsystems.

subsys/bluetooth/host/settings.c
Thalley: > failing late with least code space used is efficient.

Not sure I understand this statement. If we can perform a single check to ensure the rest of the function runs without errors, how is performant than doing 3 manual checks (after copying the data), plus the additional checks in `strncpy` compared to doing `strcpy` (or even `memcpy`)? Having a single check at the beginning of the function would also the reduce the amount of code to maintain

subsys/bluetooth/host/settings.c
Thalley: > Original contribution was a non-ARM architecture, hence, atleast two architectures are efficient without branching in this scenario.

Assuming that the original author actually tested and/or considered this :) 

> As seen from builds, it is significant code space increase (format strings for snprintk take code space in addition to snprintk 's own code use). Simple specific concatenation save on the heavy caller and callee code use by snprintk. It may be balanced if there are more module specific concatenations. For a resource constraints device, users be given the option to optimize application with less forking of the subsystems.

That sort of brings us back to what I think Johan or Aleksander said the other day: Why do we even have both versions then if the non-snprintk version is always smaller? 

I guess if it's smaller here, then `bt_addr_to_str` would probably also benefit from non using `snprintk` (along with a variety of other functions used by BT), but that is of course outside the scope of this PR

subsys/bluetooth/host/settings.c
Thalley: > Code size will increase marginally, 124 bytes.

So the optimization done here may increase flash if the application uses printk for anything else, including something as trivial as logging addresses with `bt_addr_to_str` which is quite commonly used. So this optimization (if enabled) will by default likely increase memory usage, and only in the case of `-DCONFIG_PRINTK=n -DCONFIG_BT_SETTINGS_USE_PRINTK=n` will provide an actual optimization, but by disabling features that used by subsystems, and may even prevent certain features to be enabled? 

For example, HAS uses `bt_addr_le_str` in `LOG_ERR`, and it would not be possible to have logging and HAS enabled with `CONFIG_PRINTK=n`, in which case the memory footprint from this PR will increase when `CONFIG_BT_SETTINGS_USE_PRINTK=n`. 

It seems to me that as long as Bluetooth uses `snprintk` (or any printk for that matter) as much as it does, then the optimization from this PR is the opposite of an optimization, and is rather increased code to maintain with a higher memory footprint. 

It looks to me that BT either needs to lean into `printk` or away from it. Optimizing a single (or 2) function like this won't do us any good IMO and I think it's actively harmful in the majority of cases.

subsys/bluetooth/host/settings.c
cvinayak: Likely control path is where the function succeeds with least number of checks that fail and finishes the concatenation. Since `strncpy` additionally has checks, an early check is redundant, unless we want to have a custom implementation of `strncpy` without the length check as we will have early checks. All other length checks in the function will be removed too.

Only settings uses `strcpy` and `strncpy`, i can make the change to not use it if you insist to add early check...

subsys/bluetooth/host/settings.c
cvinayak: >  logging addresses with bt_addr_to_str which is quite commonly used.

Only for debug build. Release builds of constraint device do not go production with logging enable unless is a product feature (backend to non-volatile memory or live cloud storage etc.).

> HAS uses bt_addr_le_str in LOG_ERR

I do not think using of LOG_ERR is a functional requirement of HAS, is it? Errors should be gracefully handled in the functional requirements, right?

> Bluetooth uses snprintk (or any printk for that matter) as much as it does, then the optimization from this PR is the opposite of an optimization

`CONFIG_BT_SETTINGS_USE_PRINTK=y` is the default. It has to be a conscious decision by user to increase code size by setting it to `n` while his application uses `snprintk`.

subsys/bluetooth/host/settings.c
cvinayak: As pointed out early `bt_add_to_str` is not simple concatenation of string buffer, but a mix of constants and variable meeting a format; best handled by `snprintk`. Additionally, use of `bt_addr_to_str` is not a functional requirement for most constraint devices with no user interface.

subsys/bluetooth/host/settings.c
Thalley: Fair points :) Ideally it'd be an optimization in any case, which would make it possible to have it as the only function, but it is what is it :)

Mesh and Classic depend on printk today regardless, and the API functions `bt_uuid_to_str` and `bt_addr_to_str` does too, so possible `CONFIG_BT_SETTINGS_USE_PRINTK` should depend on neither Classic nor Mesh being enabled (as that would just be a bad choice (or maybe just be a warning?)), and then we may want to consider if `bt_addr_to_str` and `bt_uuid_to_str` should either be changed to not depend on printk, or have 2 versions like these functions here

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,129 @@
...
+	/* Key format:
+	 *  "bt/<key>"
+	 */
+	strcpy(path, "bt/");
+
+	/* Concatenate key as much the free space permits */
+	strncpy(&path[len], key, path_size - len);
+	len = strlen(path);
+
+	/* Insufficient path_size, include null termination */
```
Thalley:
```suggestion
	/* Insufficient path_size */
```

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,129 @@
...
+	}
+
+	/* Key format:
+	 *  "bt/<key>"
+	 */
+	strcpy(path, "bt/");
+
+	/* Concatenate key as much the free space permits */
+	strncpy(&path[len], key, path_size - len);
+	len = strlen(path);
```
Thalley: I guess it's technically possible to use the return value of `strncpy` to determine if `len >= path_size` without calling `strlen` for a minor optimization (making it an O(1) check rather than a O(n)). Something like 
```c
end = strncpy(...)
if (end - path >= path_size) { ... }
```

It's OK as is (although I'd still prefer the length checks at the beginning of the function before attempting to copy anything)

subsys/bluetooth/host/settings.c
```diff
@@ -324,21 +378,24 @@ __weak void bt_testing_settings_delete_hook(const char *key)
...
 	if (addr) {
+		char id_str[4];
+
 		if (id) {
 			u8_to_dec(id_str, sizeof(id_str), id);
 		}
 
-		bt_settings_encode_key(key_str, sizeof(key_str), key, addr, (id ? id_str : NULL));
+		err = bt_settings_encode_key(key_str, sizeof(key_str), key, addr,
+					     (id ? id_str : NULL));
```
Thalley, line 402: Outside the scope of this PR perhaps, but I still think we should check the key length; i.e. the key cannot be e.g. 4 octets or 30 octets

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,129 @@
...
+	}
 
-		if (len < path_size) {
-			/* Type can be either BT_ADDR_LE_PUBLIC or
-			 * BT_ADDR_LE_RANDOM (value 0 or 1)
-			 */
-			path[len] = '0' + addr->type;
-			len++;
-		}
+	/* Insufficient path_size, include null termination */
```
Thalley:
```suggestion
	/* Insufficient path_size */
```

Or did you mean 

```suggestion
	/* Insufficient path_size, including null termination */
```

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,129 @@
...
-		for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-			len += bin2hex(&addr->a.val[i], 1, &path[len],
-				       path_size - len);
-		}
+	/* Postfix '/' and concatenate key as much the free space permits */
+	if ((key != NULL) && (len < path_size)) {
+		path[len] = '/';
+		len++;
+		strncpy(&path[len], key, path_size - len);
+		len += strlen(&path[len]);
```
Thalley:
```suggestion
		len += strlen(path);
```

Since you use `len` to check against `path_size`, you need to use the entire string.

Thalley: In the case that `strncpy` cannot copy the entire string, the resulting buffer will not contain a NULL termination, and thus we cannot perform `strlen`. You could use `strnlen` or apply the check before copying

Thalley (changes requested):

subsys/bluetooth/host/settings.c
cvinayak: Current customer's application sees it sufficient to avoid use of `snprintk` in the BLE core (GAP, GATT) subsystem to have reduced code in release builds.

I am open to research towards optimized `bt_addr_to_str` and `bt_uuid_to_str` ... This can also be a good first issue too.

subsys/bluetooth/host/settings.c
cvinayak: Good catch.

From the `strncpy` man pages:

`It's impossible to distinguish truncation by the result of the call, from a character sequence that just fits the destination buffer; truncation should be detected by comparing the length of the input string with the size of the destination buffer.` 

Hence updated to use `min(strlen(key), (path_size - len))`

cvinayak: To save on CPU cycles counting entire string, accumulating only the concatenated new length.

subsys/bluetooth/host/settings.c
cvinayak: including

subsys/bluetooth/host/settings.c
cvinayak: `strncpy()
              returns dst.`

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,128 @@
...
+	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+	 */
+	strcpy(path, "bt/");
+
+	/* Concatenate subsys as much the free space permits */
+	strncpy(&path[len], subsys, (path_size - len));
+	len += MIN(strlen(subsys), (path_size - len));
+
+	/* Postfix '/' if there is free space */
+	if (len < path_size) {
```
Thalley: If there are no free space, we should return an error. We really should not truncate data silently

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,128 @@
...
+		path[len] = '/';
+		len++;
+	}
+
+	/* Concatenate addr as much the free space permits */
```
Thalley: If there are no free space, we should return an error. We really should not truncate data silently

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,128 @@
...
+	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
+		len += bin2hex(&addr->a.val[i], 1, &path[len], (path_size - len));
+	}
+
+	/* Postfix addr type if there is free space */
```
Thalley: If there are no free space, we should return an error. We really should not truncate data silently

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +33,128 @@
...
-		}
+		path[len] = '0' + addr->type;
+		len++;
+	}
 
-		for (int8_t i = 5; i >= 0 && len < path_size; i--) {
-			len += bin2hex(&addr->a.val[i], 1, &path[len],
-				       path_size - len);
-		}
+	/* Postfix '/' and concatenate key as much the free space permits */
```
Thalley: If there are no free space, we should return an error. We really should not truncate data silently

subsys/bluetooth/host/settings.c
Thalley: Personally I would have written the function like

```c
int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
			   const bt_addr_le_t *addr, const char *key)
{
	/* Key format:
	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
	 */
	const size_t prefix_len = sizeof("bt") - 1;
	const size_t subsys_len = sizeof('/') + strlen(subsys);
	const size_t addr_str_len = sizeof('/') + sizeof("554433221100tt") - 1;
	const size_t key_len = key == NULL ? 0 : sizeof('/') + strlen(key);
	const size_t total_len = prefix_len + subsys_len + addr_str_len + key_len + sizeof('\0');
	size_t offset = 0U;

	if (path_size < total_len) {
		return -EINVAL;
	}

	memcpy(path, "bt", prefix_len);
	offset += prefix_len;

	path[offset++] = '/';
	memcpy(path + offset, subsys, subsys_len);
	offset += subsys_len;

	path[offset++] = '/';
	for (int8_t i = 5; i >= 0 && len < path_size; i--) {
		hex2char(&addr->a.val[i], &path[offset++]);
	}
	hex2char(&addr->type, &path[offset++]);

	if (key != NULL) {
		path[offset+] = '/';
		memcpy(path + offset, key, key_len);
		offset += key_len;
	}
	path[offset] = '\0';

	LOG_DBG("Encoded path %s", path);

	return 0;
}
```

I don't have any specific numbers on this, but I would expect this to not only be smaller in code size, but also more performant as `strcpy` and `strncpy` not only contains additional checks for each character copies, but each call also stores the NULL terminator. Getting the lengths and doing the length check at the beginning should not affect performance in such a high degree that your argument that it's unlikely. 

However what is the most performant solution also likely depends on the compiler, the platform/architecture and the compiler options. 

It would also be possible to use something like a net_buf_simple to increment the offset automatically instead, but not sure what the complications of performance of that would be. 

In any case, I think the performance differences we discuss here are so minimal that it doesn't really matter for something that's only done once per boot (probably). What is more important (to me) is correctness (which the current implementation (and even the version of this PR right now) struggles with (we should not truncate data...), and readability. It's not particularly clear why we have things like `len += MIN(strlen(subsys), (path_size - len));` after `strncpy`s IMO. 

I won't be blocking the PR once the truncations statements and code have been removed, as the performance metrics are quite hard to draw specific conclusions from IMO.

Thalley (dismissed):
I don't really want to discuss the optimizations anymore, as they are so minimal that it hardly matters IMO :) The only thing I want to have fixed now is proper checks for the length so that we don't silently truncate data

subsys/bluetooth/host/settings.c
Thalley: (a bit of a side track, but if I understand https://gitlab.com/MISRA/MISRA-C/MISRA-C-2012/Example-Suite/-/blob/master/R_13_03.c correct, I guess `path[offset++] = '/';` in non-compliant ?)

subsys/bluetooth/host/settings.c
Thalley: Ah, sorry, misread the `man` page and accidentally read the `return` of `stpncpy` :D

subsys/bluetooth/host/settings.c
cvinayak: > Personally I would have written the function like

That is 20 extra bytes on the program stack which cannot be detected at build time, needing runtime thread analyze to update stack size requirements of the applications.

Also, memcpy with length supplied is inefficient ~with more code executing to check overlapping memory, word alignments and length~ counting compared to strcpy which terminates on null.

example:
```
00012c3c <strcpy>:
   12c3c:  4603        mov  r3, r0
   12c3e:  f811 2b01   ldrb.w  r2, [r1], #1
   12c42:  f803 2b01   strb.w  r2, [r3], #1
   12c46:  2a00        cmp  r2, #0
   12c48:  d1f9        bne.n  12c3e <strcpy+0x2>
   12c4a:  4770        bx  lr

```

> It's not particularly clear why we have things like len += MIN(strlen(subsys), (path_size - len)); after strncpys IMO.

This is the suggestion in the `strncpy` man pages to handle truncation scenarios. `truncation should be detected by comparing the  length  of  the  input string with the size of the destination buffer.` I am sure the `strncpy` documentation has been thoroughly reviewed for this being mentioned there.

cvinayak:
> I don't really want to discuss the optimizations anymore, as they are so minimal that it hardly matters IMO :) The only thing I want to have fixed now is proper checks for the length so that we don't silently truncate data

There is the late check and hence we don't silently truncate data. Your suggestion for readability is not justified with significant increase to runtime RAM and CPU usage, potentially increasing risk of failures.

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,122 @@
...
+
+	path[offset] = '/';
+	offset++;
+	memcpy(&path[offset], subsys, subsys_len);
+	offset += subsys_len;
+
+	path[offset] = '/';
+	offset++;
+	for (int8_t i = 5; i >= 0; i--) {
+		hex2char(addr->a.val[i] >> 4, &path[offset]);
```
Thalley: Our coding guidelines actually require that we check for error responses, so we probably need to do something like 
```suggestion
		err = hex2char(addr->a.val[i] >> 4, &path[offset]);
		if (err != 0) {
		    return -EINVAL;
		}
```

Interestingly, `bin2hex` does not apply this and would silently fail on invalid input

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,122 @@
...
+	return 0;
+}
+
+static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
+{
+	/* Key format:
+	 *  "bt/<key>"
+	 */
+	const size_t prefix_len = sizeof("bt") - 1;
+	const size_t key_len = key == NULL ? 0 : sizeof('/') + strlen(key);
```
Thalley:
```suggestion
	const size_t key_len = sizeof('/') + strlen(key);
```

`key` isn't optional in this function, right?

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,122 @@
...
+	size_t offset = 0U;
+
+	if (path_size < total_len) {
+		return -EINVAL;
+	}
+
+	memcpy(path, "bt", prefix_len);
+	offset += prefix_len;
+
+	if (key != NULL) {
```
Thalley:
```suggestion
```

`key` isn't optional in this function, right?

subsys/bluetooth/host/settings.c
```diff
@@ -149,7 +195,7 @@ static int set_setting(const char *name, size_t len_rd, settings_read_cb read_cb
 
 	len = settings_name_next(name, &next);
 
-	if (!strncmp(name, "id", len)) {
+	if ((len == 2) && (memcmp(name, "id", len) == 0)) {
```
Thalley, line 216: The whole `str` vs `mem` function w.r.t. code size is interesting: 

If the application being built already uses `str` functions, it would be better to use that here too (right?), and if the application already uses the `mem` functions, it'd be better to use that here too. 

Based on your last reply, the changes here makes sense, but the optimization from it are conditional on what is already being linked, so I don't think we can make any true optimizations here w.r.t. code size. Semantically the `str` functions make sense given that we work with strings, and as you've pointed out, the `str` functions may be as good or even better than the `mem` functions (although I guess that probably also depends on the libc implementation being used).

subsys/bluetooth/host/settings.c
```diff
@@ -33,28 +33,45 @@
...
-			 addr->a.val[2], addr->a.val[1], addr->a.val[0],
-			 addr->type);
+		err = snprintk(path, path_size, "bt/%s/%02x%02x%02x%02x%02x%02x%u", subsys,
+			       addr->a.val[5], addr->a.val[4], addr->a.val[3], addr->a.val[2],
+			       addr->a.val[1], addr->a.val[0], addr->type);
+	}
+
+	if (err < 0) {
+		return -EINVAL;
 	}
```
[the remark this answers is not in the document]
Thalley: @cvinayak should we handle this one too?

Thalley (changes requested):
Minor changes requested, otherwise LGTM. 

I'm not 100% on whether `str` or `mem` functions are ideal here, and I expect there's not a one true "best" solution

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,122 @@
...
+			   const bt_addr_le_t *addr, const char *key)
+{
+	/* Key format:
+	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+	 */
+	const size_t prefix_len = sizeof("bt") - 1;
+	const size_t subsys_len = sizeof('/') + strlen(subsys);
+	const size_t addr_str_len = sizeof('/') + sizeof("554433221100t") - 1;
+	const size_t key_len = key == NULL ? 0 : sizeof('/') + strlen(key);
+	const size_t total_len = prefix_len + subsys_len + addr_str_len + key_len + sizeof('\0');
```
Thalley: Probably needs something like 

```suggestion
	const size_t prefix_len = sizeof("bt") - 1;
	const size_t subsys_len = strlen(subsys);
	const size_t addr_str_len = sizeof("554433221100t") - 1;
	const size_t key_len = key == NULL ? 0 : strlen(key);
	const size_t total_len = prefix_len + sizeof('/') + subsys_len + sizeof('/') + addr_str_len + key == NULL ? 0 : (sizeof('/') + key_len) + sizeof('\0');
```

To deal with the sonar issues. I guess we don't have any CI tests for `CONFIG_BT_SETTINGS_USE_PRINTK=n` to verify correctness of these functions?

subsys/bluetooth/host/settings.c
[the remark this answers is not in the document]
cvinayak: done

subsys/bluetooth/host/settings.c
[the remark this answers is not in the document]
cvinayak: done

subsys/bluetooth/host/settings.c
cvinayak: Added inline comment.

subsys/bluetooth/host/settings.c
cvinayak: Fixed.

subsys/bluetooth/host/settings.c
cvinayak: Fixed.

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,130 @@
...
+#else /* !CONFIG_BT_SETTINGS_USE_PRINTK */
+int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
+			   const bt_addr_le_t *addr, const char *key)
+{
+	/* Key format:
+	 *  "bt/<subsys>/<addr><type>/<key>", "/<key>" is optional
+	 */
+	const size_t prefix_len = sizeof("bt") - 1;
+	const size_t subsys_len = strlen(subsys);
+	const size_t addr_str_len = sizeof("554433221100t") - 1;
```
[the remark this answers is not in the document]
Thalley: Might be worth mentioning what `554433221100t` refers to (BD addr + type)

subsys/bluetooth/host/settings.c
Thalley: Inline comments does not make the code adhere to our coding guidelines. 
Additionally the comment itself, 

> /* We are not checking hex2char return value as we supply valid 0-15 as input */

Cannot be guaranteed as we are simply reading data from _somewhere_ without any prior validation, so we cannot know if this is actually true. 

Suggest to add an actual check (maybe just using an assert) to adhere to coding guidelines and actually validate (when assert is enabled). Alternatively we can also do a proper runtime check (might be the better solution overall)

tests/bsim/bluetooth/host/privacy/peripheral/prj.conf
```diff
@@ -18,5 +18,8 @@ CONFIG_NVS=y
 CONFIG_SETTINGS=y
 CONFIG_BT_SETTINGS=y
 
+# Lets cover settings without use of snprintk
+CONFIG_BT_SETTINGS_USE_PRINTK=n
```
Thalley: Does this test the actual content of the settings with `CONFIG_BT_SETTINGS_USE_PRINTK=n`? Perhaps it's better to add a build config for our actual settings bsim test with `CONFIG_BT_SETTINGS_USE_PRINTK=n`

subsys/bluetooth/host/settings.c
```diff
@@ -324,21 +378,24 @@ __weak void bt_testing_settings_delete_hook(const char *key)
...
 
-		bt_settings_encode_key(key_str, sizeof(key_str), key, addr, (id ? id_str : NULL));
+		err = bt_settings_encode_key(key_str, sizeof(key_str), key, addr,
+					     (id ? id_str : NULL));
 	} else {
-		err = snprintk(key_str, sizeof(key_str), "bt/%s", key);
-		if (err < 0) {
-			return -EINVAL;
-		}
+		err = bt_settings_encode_key_no_addr(key_str, sizeof(key_str), key);
```
Thalley, line 404: I'm sorry to bring this up this late, but something looks fishy here. 

for 
```c
int bt_settings_encode_key(char *path, size_t path_size, const char *subsys,
			   const bt_addr_le_t *addr, const char *key)
``` 
we supply `path=key_str`, `path_size=sizeof(key_str)`, `subsys=key`, `addr=addr`, `key=id`. 

For 
```c
static int bt_settings_encode_key_no_addr(char *path, size_t path_size, const char *key)
```

we supply `path=key_str`, `path_size=sizeof(key_str)`, `key=key`. 

Why are we supplying `key` for `subsys` in the first function, and `id` as `key` for the second function? 

I'm also not 100% sure what `id` refers to, and why the value `0` makes it optional...

subsys/bluetooth/host/settings.c
```diff
@@ -33,75 +32,130 @@
...
+		/* We are not checking hex2char return value as we supply valid 0-15 as input */
+		hex2char(addr->a.val[i] >> 4, &path[offset]);
+		offset++;
+		/* We are not checking hex2char return value as we supply valid 0-15 as input */
+		hex2char(addr->a.val[i] & 0xf, &path[offset]);
+		offset++;
+	}
+	/* We are not checking hex2char return value as we supply valid 0-1 as input */
+	hex2char(addr->type, &path[offset]);
+	offset++;
```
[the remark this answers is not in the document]
cvinayak: Type is either 0 or 1. Hence, this is ok.

subsys/bluetooth/host/settings.c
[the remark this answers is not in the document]
cvinayak: Comment above already has `<addr><type>`  mentioned.

subsys/bluetooth/host/att.c
```diff
@@ -385,8 +384,13 @@ static void att_disconnect(struct bt_att_chan *chan)
...
 
-	bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
-	LOG_DBG("ATT disconnecting device %s", addr);
+	if (IS_ENABLED(CONFIG_BT_ATT_LOG_LEVEL_DBG)) {
+		char addr[BT_ADDR_LE_STR_LEN];
+
+		bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
+
+		LOG_DBG("ATT disconnecting device %s", addr);
+	}
```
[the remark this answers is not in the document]
cvinayak: This is done to remove calls to `bt_addr_le_to_str` which is  not being removed by the compiler.

subsys/bluetooth/host/settings.c
```diff
@@ -7,7 +7,6 @@
 #include <errno.h>
 #include <stddef.h>
 #include <stdint.h>
```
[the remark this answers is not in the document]
cvinayak: adding back.... yes, strlen used.

subsys/bluetooth/host/att.c
```diff
@@ -3195,12 +3199,16 @@ static void att_chan_detach(struct bt_att_chan *chan)
...
 
-	bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
-	LOG_ERR("ATT Timeout for device %s. Disconnecting...", addr);
+	if (IS_ENABLED(CONFIG_BT_ATT_LOG_LEVEL_DBG)) {
+		char addr[BT_ADDR_LE_STR_LEN];
+
+		bt_addr_le_to_str(bt_conn_get_dst(chan->att->conn), addr, sizeof(addr));
+
+		LOG_ERR("ATT Timeout for device %s. Disconnecting...", addr);
+	}
```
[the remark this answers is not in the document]
cvinayak: Fixed

subsys/bluetooth/host/settings.c
cvinayak: @alwa-nordic do you have any comments? I am as good as anyone when it comes to history of the use here.

subsys/bluetooth/host/att.c
Thalley: I wonder if we would have the same issue if we used `bt_addr_le_str` instead of `bt_addr_le_to_str`?

subsys/bluetooth/host/settings.c
Thalley:
```suggestion
	const size_t prefix_len = sizeof("bt") - 1U;
	const size_t subsys_len = strlen(subsys);
	const size_t addr_str_len = sizeof("554433221100t") - 1U;
	const size_t key_len = (key == NULL) ? 0U : strlen(key);
	const size_t total_len = prefix_len +
				 sizeof('/') + subsys_len +
				 sizeof('/') + addr_str_len +
				 ((key == NULL) ? 0U : (sizeof('/') + key_len)) +
```

Minor (technically a coding guidelines violation)

Thalley:
```suggestion
	const size_t prefix_len = sizeof("bt") - 1U;
```

Thalley: I recently learned that `sizeof('\0')` or `sizeof('/')` is actually the same as `sizeof(int)` :s 

So we can either do `sizeof(char)` or something like 

```c
const char null_term = '\0';
sizeof(null_term);
```

Weird C behavior IMO :( So sizeof constant literal chars cannot/should not be used

Thalley: ditto with `sizeof`

Thalley (changes requested):

subsys/bluetooth/host/settings.c
Thalley: Github is being weird here; coding guidelines mandate `U` after unsigned literal values

subsys/bluetooth/host/settings.c
cvinayak: Done

subsys/bluetooth/host/settings.c
cvinayak: added `U`

cvinayak: done

cvinayak: explicit use of `/` or `\0`  is expressive as to what we are concatenating. keeping as-is.

cvinayak: cant figure where this line you refer to...

cvinayak (commented):
github changed,,,it is difficult to refer to older comments... please let me know if I missed some old ones

samples/bluetooth/peripheral_hids/src/main.c
cvinayak: The sample change commit removed and attached as patch file in the description so that the code size reduction can be reproduced using the build commandline in the description.

subsys/bluetooth/host/settings.c
Thalley: I think we need to update it; the length check is otherwise invalid. 

`total_len` will be 12 larger than the actual length needed, which means that if `path_size is 50` and the _actual_ lenght is 40, then the `if (path_size < total_len) {` check will fail because the calculation for `total_len` will be 12 larger (52 instead of 40). So while `sizeof('/')` is nice for semantic reasons (make it clear that we are taking the size of this character), due to the way `sizeof` works, it will be wrong. 

I'm surprised this isn't caught by our CI, which also kind of exposes that our CI testing for these functions are missing some verification

Thalley: `const size_t total_len = prefix_len + sizeof('/') + key_len + sizeof('\0');` has the same issue as mentioned in https://github.com/zephyrproject-rtos/zephyr/pull/100887#discussion_r2713173888

Thalley: One way that can work (just tested) is to make it `sizeof((char)'/')` which does return the expected size (1)

Thalley: (I also created https://github.com/zephyrproject-rtos/zephyr/issues/102707 to help catch such errors in CI if someone can figure that check out)

cvinayak:
> Sorry if this was already discussed and I missed it, but I'm worried that we'd now have two buildtime dependent branches of non-trivial code that will need to stay precisely in sync (generate & parse exactly the same strings). What would be the penalty of just forgetting the new Kconfig option and instead always use the snprintk-free variant of the implementation?

No worries. The code size are here: https://github.com/zephyrproject-rtos/zephyr/pull/100887#discussion_r2634389905

Use of `printk` in the said application increases the size by ~ 5KB, and using the non-snprintk variant increases it by a marginal size (this may vary more or less in the future, generally its more code to compile in tho).

This PR re-ensure `CONFIG_BT_SETTINGS_USE_PRINTK=n` truely does not use kernel's printk implementation, and hence reduce code size by ~ 2KB.

subsys/bluetooth/host/att.c
cvinayak: Updated `bt_addr_le_str` being inside the `LOG_ERR`/`LOG_BDG` call, the code is eliminated. And the implementation is cleaner.

subsys/bluetooth/host/settings.c
cvinayak: Fixed, re-review.

jhedberg (approved):

subsys/bluetooth/host/settings.c
Thalley, line 216: I assume this is avoid pulling in any of `str` functions?

Thalley (approved):
No blockers :) LGTM

subsys/bluetooth/host/settings.c
cvinayak: yes. There are some string literals in use here... there is scope for improvement to use defines for these.

PR #100890, opened 2025-12-11
west.yml: hal_stm32: update to add LPTIM pins

tpambor:
Update hal_stm32 to include pin definitions for LPTIM peripheral.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -255,7 +255,7 @@ manifest:
       groups:
         - hal
     - name: hal_stm32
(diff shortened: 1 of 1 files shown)


erwango:
@tpambor For my information, is that the only missing piece to support PWM on LPTIM ? (Is the PWM driver otherwise compatible with use of lptim devices ?)

erwango (approved):
Approved, but I feel this PR is a bit useless and https://github.com/zephyrproject-rtos/zephyr/pull/101205 could be used directly instead.

PR #100893, opened 2025-12-11
REUSE.toml: Default annotate json files

pdgendt:
It's not possible to add comments to a json file, so there's no option to add license or copyright text.

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -1,6 +1,6 @@
 version = 1
 
-# Declare default license and copyright text for files that typically do not include them.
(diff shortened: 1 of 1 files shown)


REUSE.toml
kartben:
```suggestion
# Declare default license and copyright text for files that typically do not or cannot include them.
```

kartben (dismissed):

kartben (approved):

keith-zephyr (approved):

PR #100896, opened 2025-12-11
drivers: adc: stm32: fix init issue when 2 adc are enabled

gautierg-st:
If several ADCs are used and shares a common clock property (for example ADC1/2 prescaler value on STM32U5), none of them should be enabled when the clock is set.
To that end, make sure to disable ADC at the end of the initialization, it will be enabled later when necessary anyway.

### DIFF
--- a/drivers/adc/adc_stm32.c
+++ b/drivers/adc/adc_stm32.c
@@ -1700,6 +1700,13 @@ static int adc_stm32_init(const struct device *dev)
 	LL_ADC_REG_SetTriggerSource(adc, LL_ADC_REG_TRIG_SOFTWARE);
 #endif /* HAS_CALIBRATION */
 
+	/* If several ADCs are used and share a common clock property (for example ADC1/2 prescaler
+	 * value on STM32U5), none of them should be enabled when the clock is set.
+	 * To that end, make sure to disable ADC at the end of the initialization, it will be
+	 * enabled later when necessary anyway.
+	 */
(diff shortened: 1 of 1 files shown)


drivers/adc/adc_stm32.c
```diff
@@ -1700,6 +1700,13 @@ static int adc_stm32_init(const struct device *dev)
 	LL_ADC_REG_SetTriggerSource(adc, LL_ADC_REG_TRIG_SOFTWARE);
 #endif /* HAS_CALIBRATION */
 
+	/* If several ADCs are used and share a common clock property (for example ADC1/2 prescaler
+	 * value on STM32U5), none of them should be enabled when the clock is set.
+	 * To that end, make sure to disable ADC at the end of the initialization, it will be
+	 * enabled later when necessary anyway.
+	 */
+	adc_stm32_disable(adc);
```
mathieuchopstm, line 1708: Just to be on the safe side: there are some operations that "lose" ADC calibration. `disable` isn't one of them, right?

mathieuchopstm (approved):
LGTM; one question:

drivers/adc/adc_stm32.c
gautierg-st: Fortunately no, otherwise we would be calibrating it all the time. There are plenty of times where we disable the ADCs, pretty much any time we need to set a new config. This case is just a bit special since it sets a common one.

erwango (approved):

PR #100898, opened 2025-12-11
ci: tags: Make sure openthread tag is used on OT changes

rlubos:
Use namespaced variants instead.

Note: I have purposedly skipped one symbol to see if CI triggers and catches the issue this time. Will update after getting results.

### DIFF
--- a/modules/openthread/platform/infra_if.c
+++ b/modules/openthread/platform/infra_if.c
@@ -392,21 +392,21 @@ static void remove_checksums_for_eth_offloading(uint8_t *buf, uint16_t len)
 	}
 
 	switch (ipv4_hdr->proto) {
-	case NET_IPPROTO_ICMP:
+	case IPPROTO_ICMP:
 		if ((config.chksum_support & NET_IF_CHECKSUM_IPV4_ICMP) != 0) {
 			struct net_icmp_hdr *icmp_hdr = (struct net_icmp_hdr *)pkt_cursor;
 
 			icmp_hdr->chksum = 0;
 		}
 	break;
-	case NET_IPPROTO_UDP:
+	case IPPROTO_UDP:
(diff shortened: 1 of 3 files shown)


modules/openthread/platform/infra_if.c
```diff
@@ -392,14 +392,14 @@ static void remove_checksums_for_eth_offloading(uint8_t *buf, uint16_t len)
...
-	case IPPROTO_ICMP:
+	case NET_IPPROTO_ICMP:
 		if ((config.chksum_support & NET_IF_CHECKSUM_IPV4_ICMP) != 0) {
 			struct net_icmp_hdr *icmp_hdr = (struct net_icmp_hdr *)pkt_cursor;
 
 			icmp_hdr->chksum = 0;
 		}
 	break;
-	case IPPROTO_UDP:
+	case NET_IPPROTO_UDP:
```
pdgendt: There's one more for `IPPROTO_TCP`.

aescolar:
> Note: I have purposedly skipped one symbol to see if CI triggers and catches the issue this time

@rlubos it did not :(

rlubos:
What I'm trying to achieve here is that changes in `modules/openthread` trigger twister tests with `openthread` tag so that we prevent regressions like earlier today.

I'm kind of out of ideas here, where should I specify the directories so that CI runs the twister with `openthread` tag? What I did so far doesn't seem to work with this PR. @nashif Any help here?

nashif:
> What I'm trying to achieve here is that changes in `modules/openthread` trigger twister tests with `openthread` tag so that we prevent regressions like earlier today.
> 
> I'm kind of out of ideas here, where should I specify the directories so that CI runs the twister with `openthread` tag? What I did so far doesn't seem to work with this PR. @nashif Any help here?

yeah, looking at that. Tags right now only get excluded, not included. This was limited to excluding after we had some issues with coverage at some point.

PR #100910, opened 2025-12-11
dts: arm: st: h7/h7rs: add memory-regions property to mac node

bklaric1:
## Changes
- Added the `memory-regions` property to `stm32h7.dtsi` and `stm32h7rs` with `sram2` as default, in order to explicitly locate ethernet buffer and descriptor to an sram region inside the DT. 
- Added override for `sram2` with `sram3` where possible.
- Added `#memory-region-cells` to `sram2` and `sram3` nodes, since it is needed by the `memory-regions` property.
- Added a check for memory-regions before the check for and definition of the `sram_eth_node` in `mpu_regions.c` and `sections.ld` to both h7 and h7rs SoCs.
- Added `memory-regions` property to `dts/bindings/ethernet/st,stm32h7-ethernet.yaml`.
- Added seperate compatible for h5 series, in form of `dts/bindings/ethernet/st,stm32h5-ethernet.yaml`.

## ~~Dependencies~~
~~- #99296 (due to the addition of the `memory-regions` property to `dts/bindings/ethernet/st,stm32h7-ethernet.yaml`)~~

## Notes
@etienne-lms suggested this change in #99296.

### DIFF
--- a/dts/arm/st/h5/stm32h563.dtsi
+++ b/dts/arm/st/h5/stm32h563.dtsi
@@ -27,7 +27,8 @@
 			clocks = <&rcc STM32_CLOCK(AHB1, 19)>;
 
 			mac: ethernet {
-				compatible = "st,stm32h7-ethernet", "st,stm32-ethernet";
+				compatible = "st,stm32h5-ethernet", "st,stm32h7-ethernet",
+					     "st,stm32-ethernet";
 				interrupts = <106 0>;
 				clock-names = "mac-clk-tx", "mac-clk-rx";
 				clocks = <&rcc STM32_CLOCK(AHB1, 20)>,
--- a/dts/arm/st/h7/stm32h743.dtsi
+++ b/dts/arm/st/h7/stm32h743.dtsi
@@ -25,6 +25,7 @@
 	/* System data RAM accessible over AHB bus: SRAM2 in D2 domain */
 	sram2: memory@30020000 {
 		reg = <0x30020000 DT_SIZE_K(128)>;
+		#memory-region-cells = <0>;
 	};
 
 	/* System data RAM accessible over AHB bus: SRAM3 in D2 domain */
@@ -32,5 +33,10 @@
 		compatible = "zephyr,memory-region", "mmio-sram";
 		reg = <0x30040000 DT_SIZE_K(32)>;
 		zephyr,memory-region = "SRAM3";
+		#memory-region-cells = <0>;
 	};
+};
+
+&mac {
+	memory-regions = <&sram3>;
 };
--- a/dts/arm/st/h7rs/stm32h7rs.dtsi
+++ b/dts/arm/st/h7rs/stm32h7rs.dtsi
@@ -61,6 +61,7 @@
 		reg = <0x30000000 DT_SIZE_K(16)>;
 		compatible = "zephyr,memory-region", "mmio-sram";
 		zephyr,memory-region = "SRAM1";
+		#memory-region-cells = <0>;
 		zephyr,memory-attr = <DT_MEM_ARM(ATTR_MPU_RAM)>;
 	};
 
@@ -69,6 +70,7 @@
 		compatible = "zephyr,memory-region", "mmio-sram";
 		reg = <0x30004000 DT_SIZE_K(16)>;
 		zephyr,memory-region = "SRAM2";
+		#memory-region-cells = <0>;
 		/* Disable SRAM2 by default to avoid unintended access.
 		 * To enable it, explicitly define zephyr,memory-attr
 		 * to configure MPU attributes.
@@ -937,6 +939,7 @@
 				clock-names = "mac-clk-tx", "mac-clk-rx";
 				clocks = <&rcc STM32_CLOCK(AHB1, 16)>,
 					 <&rcc STM32_CLOCK(AHB1, 17)>;
+				memory-regions = <&sram2>;
 				status = "disabled";
 			};
 
--- a/dts/bindings/ethernet/st,stm32h5-ethernet.yaml
+++ b/dts/bindings/ethernet/st,stm32h5-ethernet.yaml
@@ -1,0 +1,12 @@
+# Copyright (c) 2025, Benjamin Klaric <benjamin.klaric01@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  STM32H5 Ethernet
+
+  This binding file describes the device tree properties required to configure
+  and use the Ethernet controller on STM32H5 series microcontrollers.
+
+compatible: "st,stm32h5-ethernet"
+
+include: st,stm32-ethernet-common.yaml
--- a/dts/bindings/ethernet/st,stm32h7-ethernet.yaml
+++ b/dts/bindings/ethernet/st,stm32h7-ethernet.yaml
@@ -9,4 +9,12 @@ description: |
 
 compatible: "st,stm32h7-ethernet"
 
-include: st,stm32-ethernet-common.yaml
+include:
+  - "st,stm32-ethernet-common.yaml"
+  - "memory-region.yaml"
+
+properties:
+  memory-regions:
+    required: true
+    description: |
+      Reference to SRAM node used for Ethernet DMA buffer and descriptor.
--- a/dts/bindings/ethernet/st,stm32h7-ethernet.yaml
+++ b/dts/bindings/ethernet/st,stm32h7-ethernet.yaml
@@ -5,9 +5,7 @@ description: |
   STM32H7 Ethernet
 
   This binding file describes the device tree properties required to configure
-  and use the Ethernet controller on STM32H7 and STM32H5 series microcontrollers.
-  The `st,stm32h7-ethernet` compatible string can be used for both series,
-  ensuring flexibility and ease of configuration.
+  and use the Ethernet controller on STM32H7 series microcontrollers.
 
 compatible: "st,stm32h7-ethernet"
 
--- a/soc/st/stm32/stm32h7x/mpu_regions.c
+++ b/soc/st/stm32/stm32h7x/mpu_regions.c
@@ -7,6 +7,13 @@
 #include <zephyr/devicetree.h>
 #include <zephyr/arch/arm/mpu/arm_mpu_mem_cfg.h>
 
+#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
+
+BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
+	     "Ethernet buffer memory-region cannot be located in Zephyr system RAM.");
+#endif /* mac node enabled */
+
 static const struct arm_mpu_region mpu_regions[] = {
 	MPU_REGION_ENTRY("FLASH", CONFIG_FLASH_BASE_ADDRESS,
 					 REGION_FLASH_ATTR(REGION_FLASH_SIZE)),
@@ -21,22 +28,13 @@ static const struct arm_mpu_region mpu_regions[] = {
 					REGION_512K |
 					MPU_RASR_XN_Msk | P_RW_U_NA_Msk) }),
 
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(sram3))
-#define sram_eth_node	DT_NODELABEL(sram3)
-#else
-#define sram_eth_node	DT_NODELABEL(sram2)
-#endif
-
-#if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
+#if defined(sram_eth_node) && DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
 	MPU_REGION_ENTRY("SRAM_ETH_BUF",
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_RAM_NOCACHE_ATTR(REGION_16K)),
 	MPU_REGION_ENTRY("SRAM_ETH_DESC",
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_PPB_ATTR(REGION_256B)),
-#endif
 #endif
 };
 
--- a/soc/st/stm32/stm32h7x/sections.ld
+++ b/soc/st/stm32/stm32h7x/sections.ld
@@ -5,12 +5,7 @@
  */
 
 #if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(sram3))
-#define sram_eth_node	DT_NODELABEL(sram3)
-#else
-#define sram_eth_node	DT_NODELABEL(sram2)
-#endif
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
 
 #if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
 SECTION_DATA_PROLOGUE(eth_stm32,(NOLOAD),)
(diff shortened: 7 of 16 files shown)


juickar (commented):
@bklaric1 try to rebase on main. A fix has been pushed for the CI failed test

bklaric1:
@juickar they seem to be queued now for almost a day. Was there another fix in the meantime?

juickar:
> @juickar they seem to be queued now for almost a day. Was there another fix in the meantime?

@bklaric1 No, but from what I witnessed CI also fails to launch on other PRs

soc/st/stm32/stm32h7x/mpu_regions.c
```diff
@@ -40,6 +36,12 @@ static const struct arm_mpu_region mpu_regions[] = {
 #endif
 };
 
+#ifdef sram_eth_node
+BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
+	     "Remove property memory-regions from 'mac' node "
+	     "to locate ethernet buffers in system RAM.");
+#endif
```
erwango: Since this is not an option today, I'd suggest to remove

erwango (changes requested):
Thanks for the change, but in order to avoid impacting existing users, I'd prefer you keep the changes out of the boards.

So please assign the memory regions (`memory-regions = <&sram2/3>;`) in .dtsi files and keep existing pattern of using sram3 when available otherwise sram2.
This was, out of tree users won't see any difference.

soc/st/stm32/stm32h7x/mpu_regions.c
bklaric1: This was suggested by @etienne-lms. But I can remove it if wanted.

etienne-lms: I think we should prevent one to set `memory-regions` property to the `zephyr,sram` phandle.
I'd suggest:
```patch
 #include <zephyr/devicetree.h>
 #include <zephyr/arch/arm/mpu/arm_mpu_mem_cfg.h>
 
+#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
+
+BUILD_ASSERT(DT_NODE_HAS_PROP(DT_NODELABEL(mac), memory_regions) &&
+	     DT_NODE_HAS_STATUS_OKAY(sram_eth_node),
+	     "Ethernet MAC node must provide an enabled memory region for ethernet buffers");
+BUILD_ASSERT(!DT_SAME_NODE(sram_eth_node, DT_CHOSEN(zephyr_sram)),
+	     "Ethernet buffer memory-region cannot be located in Zephyr system RAM.");
+#endif /* mac node enabled */
+
 static const struct arm_mpu_region mpu_regions[] = {
	(...)

 #ifdef sram_eth_node
 	MPU_REGION_ENTRY("SRAM_ETH_BUF",		
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_RAM_NOCACHE_ATTR(REGION_16K)),
 	MPU_REGION_ENTRY("SRAM_ETH_DESC",
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_PPB_ATTR(REGION_256B)),
 #endif
 };

 (...)
```

I guess `DT_NODE_HAS_PROP(DT_NODELABEL(mac), memory_regions)` assertion can be removed if the property is tagged `required`.

dts/bindings/ethernet/st,stm32h7-ethernet.yaml
```diff
@@ -11,4 +11,11 @@ description: |
 
 compatible: "st,stm32h7-ethernet"
 
-include: st,stm32-ethernet-common.yaml
+include:
+  - "st,stm32-ethernet-common.yaml"
+  - "memory-region.yaml"
+
+properties:
+  memory-regions:
```
etienne-lms, line 17: @erwango, from your feedback, it seems this property should be tagged `required`.

bklaric1:
@erwango @etienne-lms I amended the commit to add the requested changes, update the property to be required, and update the commit message.
Now the `stm32h7.dtsi` defines `memory-regions` and sets it to `sram2` by default.
`stm32h7a3.dtsi` has `sram3` override for a and b series (these series don't support Ethernet as of now, but the addition of `memory-regions` to `stm32h7.dtsi` requires the addition of this).  
`stm32h743.dtsi` has `sram3` override for 43 and 53 series.  
`stm32h745.dtsi` has `sram3` override for 45, 47, 55 and 57 series.  
I hope I didn't forget anything/understood something wrong. In any case, I am open to feedback.

dts/arm/st/h5/stm32h562.dtsi
```diff
@@ -66,12 +66,14 @@
 			compatible = "zephyr,memory-region", "mmio-sram";
 			reg = <0x20040000 DT_SIZE_K(64)>;
 			zephyr,memory-region = "SRAM2";
+			#memory-region-cells = <0>;
```
mathieuchopstm: This should be declared in `zephyr,memory-region.yml` or somewhere else; IIRC `edtlib` has special casing for `#xxx-cells` properties so this works, but it should still be listed and documented as part of a binding.

dts/arm/st/h7/stm32h743.dtsi
```diff
@@ -25,12 +25,18 @@
...
 	sram3: memory@30040000 {
 		compatible = "zephyr,memory-region", "mmio-sram";
 		reg = <0x30040000 DT_SIZE_K(32)>;
 		zephyr,memory-region = "SRAM3";
+		#memory-region-cells = <0>;
 	};
 };
+
+&mac {
+	memory-regions = <&sram3>;
```
mathieuchopstm, line 41: Nit: maybe `memory-region-names` should be used?

soc/st/stm32/stm32h7x/sections.ld
```diff
@@ -5,14 +5,9 @@
...
+#define sram_eth_node DT_PHANDLE(DT_NODELABEL(mac), memory_regions)
 
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(sram3))
-#define sram_eth_node	DT_NODELABEL(sram3)
-#else
-#define sram_eth_node	DT_NODELABEL(sram2)
-#endif
-
-#if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
+#if defined(sram_eth_node) && DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
```
mathieuchopstm: `defined(sram_eth_node)` will always evaluate to 1: it is defined literally one line above

Either the check is not necessary:
```suggestion
#if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
```
Or this is a mistake and you were looking for another macro (`DT_NODE_EXISTS`? `DT_NODE_HAS_PROP` on `&mac`?)

soc/st/stm32/stm32h7x/mpu_regions.c
```diff
@@ -21,23 +28,14 @@ static const struct arm_mpu_region mpu_regions[] = {
...
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(mac))
-
-#if DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(sram3))
-#define sram_eth_node	DT_NODELABEL(sram3)
-#else
-#define sram_eth_node	DT_NODELABEL(sram2)
-#endif
-
-#if DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
+#if defined(sram_eth_node) && DT_NODE_HAS_STATUS_OKAY(sram_eth_node)
```
mathieuchopstm, line 31: ditto wrt `defined(sram_eth_node)`

soc/st/stm32/stm32h7x/mpu_regions.c
```diff
@@ -21,23 +28,14 @@ static const struct arm_mpu_region mpu_regions[] = {
...
 	MPU_REGION_ENTRY("SRAM_ETH_BUF",
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_RAM_NOCACHE_ATTR(REGION_16K)),
 	MPU_REGION_ENTRY("SRAM_ETH_DESC",
 					 DT_REG_ADDR(sram_eth_node),
 					 REGION_PPB_ATTR(REGION_256B)),
```
mathieuchopstm, line 37: Nit, outside PR scope: (ab)using `REGION_PPB` as a `REGION_STRONGLY_ORDERED` is really ugly.

bklaric1:
The failed twister-build (5) is related to missing `memory-regions` from `stm32h7rs.dtsi`. As originally stated, this PR depends on  #99296, but maybe the dependency should be the other way around.
Should I then add everything related to the `memory-regions` with this PR and then just use it in another PR?

dts/arm/st/h5/stm32h562.dtsi
bklaric1: As far as I can tell, this is needed because memory-regions is defined as phandle-array here:
https://github.com/zephyrproject-rtos/zephyr/blob/e6fd157a93db7297583d7b93f9a74c0e0977bc95/dts/bindings/reserved-memory/memory-region.yaml#L7-L10

I don't know if it's possible to define it in `memory-region.yaml` directly (as <0>), and if it's wanted there, since <0> might not always be the use case.

I hope I didn't miss the point of your comment.

dts/arm/st/h7/stm32h743.dtsi
bklaric1: `memory-region-names` is not phandle but rather string-array:
https://github.com/zephyrproject-rtos/zephyr/blob/e6fd157a93db7297583d7b93f9a74c0e0977bc95/dts/bindings/reserved-memory/memory-region.yaml#L12-L14

Since we need to pass a referance to sram node, I am unsure if it will work with `memory-region-names`, tho I might be wrong.

soc/st/stm32/stm32h7x/sections.ld
bklaric1: This was a copy&paste mistake I made from `mpu_regions.c`. The idea will be explained in the comment regarding `mpu_regions.c`. Will remove it here.

soc/st/stm32/stm32h7x/mpu_regions.c
bklaric1: The idea here was, disregarding the mistake made in `sections.ld`, to check if `sram_eth_node` has `status = "okay"`, meaning we check if the sram node passed with `memory-regions` has `status = "okay"`, to avoid the case of passing sram node which is disabled. This is the case with the current implementation of this check here:
https://github.com/zephyrproject-rtos/zephyr/blob/e6fd157a93db7297583d7b93f9a74c0e0977bc95/soc/st/stm32/stm32h7rsx/mpu_regions.c#L33-L35

`sram2` is passed but it is disabled by default in DT.

soc/st/stm32/stm32h7x/mpu_regions.c
bklaric1: As you said, it's outside the scope of PR and also beyond my current knowledge on the topic, but taking a guess here, could it be defined as `REGION_RAM_NOCACHE_ATTR()` instead? As far as I can tell, the goal is to make the region non-cacheable.

mathieuchopstm: I assume there is a reason for which `PPB` was used rather than `RAM_NOCACHE` as done above, but I don't know which. So let's leave it as is for now...

soc/st/stm32/stm32h7x/mpu_regions.c
mathieuchopstm: Ah, right, I see now that (unlike in the linker script) here `sram_eth_node` is only defined if the ETH MAC node is enabled. So indeed here it does make sense.

dts/arm/st/h7/stm32h743.dtsi
etienne-lms: Names used in `memory-region-names` are rather identifiers for the phandle indices used in a `memory-regions` property. It's useful when there can be several phandles, not a single one, see for examples `clocks = ...` and `clock-names`.
Since here we only expect one, I don't think it's really useful.

etienne-lms:
> I will also add the changes for h5 series as a seperate commit and remove them from the first (also edit the commit message body), but first we need to agree on what needs to be added for h5 series.

To address H5, talking with @erwango, the simplest may be to not set `memory-regions` as `required` in **st,stm32h7-ethernet.yaml**. Since this P-R adds the property in H7 and H7RS SoC DTSI files, the default config is the one we expect (a dedicated SRAM for eth buffers on these SoCs). Board developers on H7 and H7RS willing to locate ethernet buffers in Zephyr main RAM would still need to tweak the implementation as they currently need to, at least no regression foreseen.

bklaric1:
@etienne-lms I pushed a commit few minutes ago, adding seperate compatible for h5 series. Sorry, was too impatient :)
Could you check if this addition works as well or if I should just drop the `required: true` and leave it be, as you stated?

dts/arm/st/h5/stm32h563.dtsi
```diff
@@ -27,12 +27,13 @@
 			clocks = <&rcc STM32_CLOCK(AHB1, 19)>;
 
 			mac: ethernet {
-				compatible = "st,stm32h7-ethernet", "st,stm32-ethernet";
+				compatible = "st,stm32h5-ethernet", "st,stm32-ethernet";
```
etienne-lms: Could you preserve h7 compatible in the list. It is needed by the **eth_stm32_hal_\*.{c|h}** files.
```suggestion
				compatible = "st,stm32h5-ethernet", "st,stm32h7-ethernet", "st,stm32-ethernet";

File: dts/arm/st/h5/stm32h563.dtsi [hunk 9]
```diff
@@ -27,12 +27,13 @@
...
 				interrupts = <106 0>;
 				clock-names = "mac-clk-tx", "mac-clk-rx";
 				clocks = <&rcc STM32_CLOCK(AHB1, 20)>,
 					 <&rcc STM32_CLOCK(AHB1, 21)>;
 				status = "disabled";
+				memory-regions = <&sram2>;
```
etienne-lms: Could you remove this property. H5 is currently expected to locate ethernet buffer in Zephyr main RAM.

dts/bindings/ethernet/st,stm32h5-ethernet.yaml
```diff
@@ -0,0 +1,20 @@
...
+  The `st,stm32h5-ethernet` compatible string can be used for H5 series.
+
+compatible: "st,stm32h5-ethernet"
+
+include:
+  - "st,stm32-ethernet-common.yaml"
+  - "memory-region.yaml"
+
+properties:
+  memory-regions:
```
etienne-lms: I would currently remove this property, even if optional, since it's not supported by the implementation.

etienne-lms (commented):
I see you already propose a h5 dedicate compatible. It does the jobs. Discard my previous comment and let's see what other reviews think.

dts/arm/st/h5/stm32h563.dtsi
bklaric1: Done. Should I also remove the `#memory-region-cells = <0>;` from `sram2` and `sram3` nodes in `stm32h562.dtsi`?

dts/arm/st/h5/stm32h563.dtsi
bklaric1: Done.

dts/bindings/ethernet/st,stm32h5-ethernet.yaml
bklaric1: Makes sense. Done.

dts/arm/st/h5/stm32h563.dtsi
bklaric1: I removed it. Makes sense, since `memory-regions` isn't used and hence `#memory-region-cells = <0>;` isn't required.

dts/arm/st/h7rs/stm32h7rs.dtsi
```diff
@@ -938,6 +940,7 @@
 				clocks = <&rcc STM32_CLOCK(AHB1, 16)>,
 					 <&rcc STM32_CLOCK(AHB1, 17)>;
 				status = "disabled";
+				memory-regions = <&sram2>;
```
etienne-lms, line 942: By convention (DTS recommendation), `status` property is place last in the node. Could you swap the 2 lines above?

This comment also applies to the change in **dts/arm/st/h7/stm32h7.dtsi** in this P-R.

etienne-lms (commented):
To prevent build failures for example when playing with `git bisect`, could you place commit
*"dts: bindings: ethernet: add st,stm32h5-ethernet compatible"* before commit
*"dts: arm: st: h7: add memory-regions property to mac node"* that make `memory-regions` DT property required.

dts/arm/st/h7rs/stm32h7rs.dtsi
bklaric1: Done.
Also reordered the commits, so that the "*dts: bindings: ethernet: add st,stm32h5-ethernet compatible*" commit comes before the other two.

etienne-lms (approved):

erwango (approved):

PR #100913, opened 2025-12-11
soc: silabs: siwx91x: fix pm when bt is enabled.

Martinhoff-maker:
This PR add a workaround in order to have a functional pm when bt is activated.
Without this patch, we send to the network coprocessor (nwp) a power save request too early before the bt controller (nwp side) is ready.
We then lost the communication with the nwp which make the application crashed.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/103339

### DIFF
--- a/soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
+++ b/soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
@@ -43,7 +43,6 @@ struct siwx91x_nwp_data {
 struct siwx91x_nwp_config {
 	void (*config_irq)(const struct device *dev);
 	uint32_t stack_size;
-	uint8_t power_profile;
 	uint8_t antenna_selection;
 	bool support_1p8v;
 	bool enable_xtal_correction;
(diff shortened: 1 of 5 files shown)


jhedberg (dismissed):

jerome-pouiller (dismissed):

soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
```diff
@@ -437,16 +438,53 @@ int siwx91x_nwp_mode_switch(const struct device *dev, uint8_t oper_mode, bool hi
...
+		.profile = data->power_profile
+	};
+	sl_bt_performance_profile_t bt_performance_profile = {
+		.profile = data->power_profile
+	};
+	int ret;
+
+	if (!IS_ENABLED(CONFIG_SOC_SIWX91X_PM_BACKEND_PMGR)) {
+		/* no_op if PM is not enabled*/
+		goto out;
```
jerome-pouiller: Could be just `return 0;`

jerome-pouiller (dismissed):

soc/silabs/silabs_siwx91x/siwg917/siwx91x_nwp.c
Martinhoff-maker: done :)

jhedberg (approved):

jerome-pouiller (approved):

PR #100918, opened 2025-12-11
drivers: pwm: sctimer: Add PowerManagement support

davidmissael:
pwm_mcux_sctimer driver doesn't support power management actions. PM Actions added, as well support to save/restore for pwm channels configurations after exitig from PM3

### DIFF
--- a/drivers/pwm/pwm_mcux_sctimer.c
+++ b/drivers/pwm/pwm_mcux_sctimer.c
@@ -8,6 +8,7 @@
 #define DT_DRV_COMPAT nxp_sctimer_pwm
 
 #include <errno.h>
+#include <zephyr/pm/device.h>
 #include <zephyr/drivers/pwm.h>
 #include <fsl_sctimer.h>
 #include <fsl_clock.h>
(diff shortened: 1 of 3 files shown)


davidmissael:
This is a test on 3 pwm channels working on NXP FRDM-RW612. Channels are configured to work with different duty cycles, then entering to PM3 and after exiting, the channel's configurations are recovered and continue working:
image
image

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -38,6 +39,21 @@ struct pwm_mcux_sctimer_data {
...
+};
+
+static void mcux_sctimer_save_pwm_chn_config(uint32_t channel, uint32_t period_cycles,
+											uint32_t pulse_cycles, pwm_flags_t flags);
+
+static void mcux_sctimer_restore_pwm_chn_config(const struct device *dev);
+
+static struct pwm_mcux_sctimer_channel_config pwm_channel_config[CHANNEL_COUNT] = {0};
+
+static bool pwm_mcux_sctimer_channel_active;
```
FelixWang47831: I think "pwm_mcux_sctimer_channel_active" and "pwm_channel_config" should be moved into pwm_mcux_sctimer_config structure. 
Because there might be several SCTimer instances,  they should not share the same variables.

drivers/pwm/pwm_mcux_sctimer.c
FelixWang47831, line 1: Please fix issues in Compliance Checks

PR #100924, opened 2025-12-12
scripts: west: runners: add option for debug batch mode

hakehuang:
when in batch mode, we do not need manually 'go'.
with this, we can use `west debug` in twister run test

### DIFF
--- a/scripts/west_commands/runners/core.py
+++ b/scripts/west_commands/runners/core.py
@@ -324,6 +324,11 @@ class RunnerCaps:
                        # to allow other commands to use the rtt address
     dry_run: bool = False
     skip_load: bool = False
+    batch_debug: bool = False # In batch mode, GDB exits with status 0 after loading;
+                              # for automated debugging, add --batch with 'monitor go',
+                              # 'disconnect', and 'quit' commands (named batch_debug in west),
+                              # unlike interactive debug mode (default),
+                              # which stops and waits for user input
 
     def __post_init__(self):
         if self.mult_dev_ids and not self.dev_id:
@@ -650,6 +655,10 @@ class ZephyrBinaryRunner(abc.ABC):
                             help=("load image on target before 'west debug' session"
                                   if caps.skip_load else argparse.SUPPRESS),
                             default=True)
+
+        parser.add_argument('--batch', action=argparse.BooleanOptionalAction,
+                            help="enable west debug batch mode"
+                            if caps.batch_debug else argparse.SUPPRESS)
 
         # Runner-specific options.
         cls.do_add_parser(parser)
--- a/scripts/west_commands/runners/jlink.py
+++ b/scripts/west_commands/runners/jlink.py
@@ -55,7 +55,7 @@ class JLinkBinaryRunner(ZephyrBinaryRunner):
                  gdb_host='',
                  gdb_port=DEFAULT_JLINK_GDB_PORT,
                  rtt_port=DEFAULT_JLINK_RTT_PORT,
-                 tui=False, tool_opt=None, dev_id_type=None):
+                 tui=False, tool_opt=None, dev_id_type=None, batch=False):
         super().__init__(cfg)
         self.file = cfg.file
         self.file_type = cfg.file_type
@@ -81,6 +81,7 @@ class JLinkBinaryRunner(ZephyrBinaryRunner):
         self.loader = loader
         self.rtt_port = rtt_port
         self.dev_id_type = dev_id_type
+        self.is_batch = batch
 
         self.tool_opt = []
         if tool_opt is not None:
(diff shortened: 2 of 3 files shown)


hakehuang:
@gchwier

scripts/west_commands/runners/jlink.py
```diff
@@ -398,16 +398,22 @@ def do_run(self, command, **kwargs):
                 raise ValueError('Cannot debug; elf is missing')
             else:
                 elf_name = self.elf_name
+            is_batch = os.environ.get("ZEPHYR_TWISTER_RUN_IN_DEBUG", '')
```
pdgendt: Looks messy to have twister special handling in runners.

How about you add a runner capability and argument? It should be passed as a regular argument instead of an environment variable.

carlescufi: Yes, I also prefer this, or even if it were a JLink-specific argument I wouldn't mind.

hakehuang: let me update. Thanks

scripts/west_commands/runners/jlink.py
```diff
@@ -398,16 +400,22 @@ def do_run(self, command, **kwargs):
...
                 client_cmd += ['-ex', 'monitor halt',
                                '-ex', 'monitor reset',
                                '-ex', 'load']
-                if self.reset:
-                    client_cmd += ['-ex', 'monitor reset']
+                if self.is_batch:
+                    client_cmd += ['-ex', 'monitor go', '-ex', 'disconnect', '-ex', 'quit']
+                else:
+                    if self.reset:
+                        client_cmd += ['-ex', 'monitor reset']
```
pdgendt:
```suggestion
                elif self.reset:
                    client_cmd += ['-ex', 'monitor reset']
```

scripts/west_commands/runners/core.py
```diff
@@ -554,6 +555,10 @@ def add_parser(cls, parser):
         # using them to mean something else.
         caps = cls.capabilities()
 
+        if caps.batch_debug:
+            parser.add_argument('--batch', action=argparse.BooleanOptionalAction,
+                                help="enable west debug batch mode")
```
pdgendt: Also move it to end

```suggestion
        parser.add_argument('--batch', action=argparse.BooleanOptionalAction,
                            help="enable west debug batch mode"
                            if caps.batch_debug else argparse.SUPPRESS)
```

hakehuang: ok

scripts/west_commands/runners/core.py
```diff
@@ -324,6 +324,7 @@ class RunnerCaps:
                        # to allow other commands to use the rtt address
     dry_run: bool = False
     skip_load: bool = False
+    batch_debug: bool = False # work with debug to run without manually kick
```
pdgendt: The comment is not really clear to me, how do a "manual kick" and "batch" relate?

pdgendt (commented):
I assume there is a use in `twister` for this, is that PR also ready?

hakehuang:
> I assume there is a use in `twister` for this, is that PR also ready?

@pdgendt there is a draft version https://github.com/zephyrproject-rtos/zephyr/pull/100434, I will update according the final version of this pr.

scripts/west_commands/runners/core.py
hakehuang: according to https://man7.org/linux/man-pages/man1/gdb.1.html, if in --batch mode, the gdb will `Exit with status 0`. so in debug mode, we need add --batch and 'monitor go', 'disconnect', and 'quit'. without Exit with an error number.
So I name this in west as batch_debug.
and in current west debug mode, it will stop after `loading` and wait for user to type in the debug command.

any recommend to rephase this comment? or just remove it? @pdgendt

pdgendt: The `gdb` argument is clear, I'd expect something similar commented here, or even refer to that man page. Removing it would make it more unclear.

hakehuang: @pdgendt , I add more detailed comments in this option, please check.

pdgendt (approved):

scripts/west_commands/runners/core.py
pdgendt: Thanks!

scripts/west_commands/runners/jlink.py
```diff
@@ -398,16 +400,21 @@ def do_run(self, command, **kwargs):
...
                           self.tui_arg +
                           [elf_name] +
+                          ['-batch' if self.is_batch else ''] +
                           ['-ex', f'target remote {self.gdb_host}:{self.gdb_port}'])
             if command == 'debug':
                 client_cmd += ['-ex', 'monitor halt',
                                '-ex', 'monitor reset',
                                '-ex', 'load']
-                if self.reset:
+                if self.is_batch:
```
yvesll, line 413: So for jlink, if `--batch` given, `--reset` will be ignored?

hakehuang: yes, in most platform `west debug` is load to ram, and if we reset, every thing is gone. and even it loads to flash, the `load` operation is enough for debug.

yvesll: Make sense, thank you!

gchwier (approved):

PR #100927, opened 2025-12-12
drivers: espi: npcx: Added config to disable auto-ack and callback to handle VW Warn signals

yrajkumar439:
Added configuration option to disable auto acknowledgment for virtual wire WARN signals and added a callback to handle them.

This change introduces:

1. Support to disable automatic acknowledgment for VW WARN signals.
2. A callback mechanism to notify application firmware for explicit handling.

Behavior:

1. If auto-ack is disabled, the callback notifies the application to handle VW WARN signals.
2. If auto-ack is enabled, the driver acknowledges VW WARN signals without notifying the application.

### DIFF
--- a/drivers/espi/espi_npcx.c
+++ b/drivers/espi/espi_npcx.c
@@ -600,27 +600,34 @@ static void espi_vw_notify_host_warning(const struct device *dev,
 	espi_npcx_receive_vwire(dev, signal, &wire);
 
 	k_busy_wait(NPCX_ESPI_VWIRE_ACK_DELAY);
-	switch (signal) {
-	case ESPI_VWIRE_SIGNAL_HOST_RST_WARN:
-		espi_npcx_send_vwire(dev,
-				ESPI_VWIRE_SIGNAL_HOST_RST_ACK,
-				wire);
-		break;
-	case ESPI_VWIRE_SIGNAL_SUS_WARN:
-		espi_npcx_send_vwire(dev, ESPI_VWIRE_SIGNAL_SUS_ACK,
-				wire);
-		break;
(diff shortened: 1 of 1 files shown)


albertofloyd (approved):

ChiHuaL:
> It's actually the other way around. By default ESPI_AUTOMATIC_WARNING_ACKNOWLEDGE= y so automatically VW are send by the driver. Behavior will remain the same than before this PR.
> 
> Except when ESPI_AUTOMATIC_WARNING_ACKNOWLEDGE=n

ACK. You are right.  Sorry that I didn't notice there is a logic not in the `if` statement.

ChiHuaL (approved):

VenkatKotakonda (approved):

drivers/espi/espi_npcx.c
```diff
@@ -600,27 +600,34 @@ static void espi_vw_notify_host_warning(const struct device *dev,
...
-		espi_npcx_send_vwire(dev, ESPI_VWIRE_SIGNAL_DNX_ACK, wire);
-		break;
+		case ESPI_VWIRE_SIGNAL_DNX_WARN:
+			espi_npcx_send_vwire(dev, ESPI_VWIRE_SIGNAL_DNX_ACK, wire);
+			break;
 #endif
-	default:
-		break;
+		default:
+			break;
```
p9n, line 630: The switch block is missing a closing brace.

ChiHuaL: @p9n  
I submit a PR to fix the issue in [101249](https://github.com/zephyrproject-rtos/zephyr/pull/101249)

PR #100928, opened 2025-12-12
 drivers: microchip: interrupt controller: Add support for EIC g1 driver

mchp-asif:
Add support for EIC g1 driver
Assigns the port id to gpio immutable structures


nandojve:
Hi @mchp-asif , @NhMchp ,

I would like to have a meeting about this driver.

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
```diff
@@ -190,6 +190,7 @@
 			porta: gpio@41008000 {
 				compatible = "microchip,port-g1-gpio";
 				reg = <0x41008000 0x80>;
+				port-id = <0>;
```
nandojve: Hi @mchp-asif ,

I did not understood why this needs to be in devicetree.
Maybe we could sync about this.

nandojve (changes requested):

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
```diff
@@ -111,6 +111,31 @@
 			status = "disabled";
 		};
 
+		eic: eic@40002800 {
```
JarmouniA:
```suggestion
		eic: interrupt-controller@40002800 {
```
https://devicetree-specification.readthedocs.io/en/latest/chapter2-devicetree-basics.html#generic-names-recommendation

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) 2025-2026 Microchip Technology Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  Microchip External Interrupt Controller for g1 peripherals
+
+  Group g1 EIC driver supports following hardware peripherals:
```
JarmouniA:
```suggestion
  Group g1 EIC supports following hardware peripherals:
```

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
```diff
@@ -0,0 +1,126 @@
...
+compatible: "microchip,eic-g1-intc"
+
+include: base.yaml
+
+properties:
+  reg:
+    required: true
+    description: |
+      Specifies the base address and size of the register
+      block for the watchdog timer.
```
JarmouniA: watchdog!?

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
```diff
@@ -0,0 +1,126 @@
...
+    type: boolean
+    description: |
+      Enables low power mode for eic peripheral
+
+  porta-unsupported-pins:
+    type: int
+    default: 0
+    description: |
+      Each bit in the property denotes a pin number. This shows the pins which
+      does not support interrupt for port a.
```
JarmouniA:
```suggestion
      do not support interrupt for port a.
```
Apply elsewhere as well

JarmouniA (dismissed):

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
mchp-asif: Hi @nandojve, we should define this ID either in the driver source or in the device tree.

`port_id` is used throughout the EIC driver to uniquely identify each port. This is especially important when only a few of the ports are enabled. 

For example, if only PORTC is enabled and we rely on an index (#n), it would be 0, which loses the information that the port is actually PORTC. With an explicit port_id, we can reliably identify it as PORTC and use that consistently in the EIC driver logic.

We can have a sync-up if you need further clarification.

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
mchp-asif: Thank you @JarmouniA for your review. I have updated as suggested.

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
mchp-asif: Updated as suggested.

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
mchp-asif: Thanks for catching that. I have updated it.

dts/bindings/interrupt-controller/microchip,eic-g1-intc.yaml
mchp-asif: Updated as suggested

NhMchp (dismissed):

dts/arm/microchip/sam/sam_d5x_e5x/common/samd5xe5x.dtsi
nandojve: Ok,

But the id MUST be defined at compile time and not manually added in the devicetree.
I already had in the past same issues you are facing and manually defining in devicetree is not the answer.
If you are struggling to make this at compile time let me know and I can help you with that part.

The idea is that each port will have a fixed index like portA, porBb .. portN => 0, 1 .. N and if you enable only portB you still want to have the index 1

Is this correct?

mchp-asif: You’re right. That’s the idea.

 We could add the ID at compile time like we did in `gpio_sam0.c`, but I felt like pulling it from the device tree should reduce the amount of code in the driver and remove the need for extra node_label checks to assign the ID.

nandojve: The ports are memory-mapped at fixed 0x80 offsets under the pinctrl parent:
```
  ┌───────┬────────────┬────────────────────┬─────┐
  │ Port  │  Address   │ Offset from parent │ ID  │
  ├───────┼────────────┼────────────────────┼─────┤
  │ porta │ 0x41008000 │ 0x000              │ 0   │
  ├───────┼────────────┼────────────────────┼─────┤
  │ portb │ 0x41008080 │ 0x080              │ 1   │
  ├───────┼────────────┼────────────────────┼─────┤
  │ portc │ 0x41008100 │ 0x100              │ 2   │
  ├───────┼────────────┼────────────────────┼─────┤
  │ portd │ 0x41008180 │ 0x180              │ 3   │
  └───────┴────────────┴────────────────────┴─────┘
```
  The port ID is literally (reg_addr - parent_base) / reg_size. This can be computed entirely at compile time using existing DT macros:

```
  #define MCHP_GPIO_PORT_ID(inst) \
      ((DT_INST_REG_ADDR(inst) - \
        DT_REG_ADDR(DT_INST_PARENT(inst))) / \
       DT_INST_REG_SIZE_BY_IDX(inst, 0))
```

  Then in the config macro:
```
  #define GPIO_PORT_CONFIG(idx)                           \
      static const struct gpio_mchp_config                \
          gpio_mchp_config_##idx = {                      \
          .common = { ... },                              \
          .gpio_regs = (port_group_registers_t *)         \
              DT_INST_REG_ADDR(idx),                      \
          .gpio_port_id = MCHP_GPIO_PORT_ID(idx),        \
      };    
```

mchp-asif: That’s a solid idea, @nandjove. I’ll incorporate this change into the code and send an update. This way, we avoid the overhead of increasing file size and we also don’t need to maintain a port ID in the device tree. 

The only thing to keep in mind is that we should always ensure the pincontrol node remains the parent of the ports.

mchp-asif: @nandojve I have updated as you suggested.

nandojve (approved):

NhMchp (approved):

ArunMCHP (approved):

PR #100929, opened 2025-12-12
spi: nrfx_spim: support cross power-domain GPIOs

JordanYates:
Add support for the cross power-domain GPIO functionality on nRF54LX devices. Tested on nRF54L15DK with external flash switched from SPIM00 to SPIM20.

Test procedure is an ERASE-WRITE-READ-ERASE cycle validating data at each step. With the external flash moved to SPIM20, this testing fails without the implementation and works with it.

Relevant Nordic docs: https://docs.nordicsemi.com/bundle/ps_nrf54L15/page/chapters/pin.html#d380e188

### DIFF
--- a/drivers/spi/spi_nrfx_spim.c
+++ b/drivers/spi/spi_nrfx_spim.c
@@ -19,6 +19,7 @@
 #include <hal/nrf_clock.h>
 #endif
 #include <nrfx_spim.h>
+#include <nrf_sys_event.h>
 #include <string.h>
 #include <zephyr/linker/devicetree_regions.h>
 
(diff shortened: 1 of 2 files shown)


JordanYates:
I wasn't aware that it was previously in-tree then removed, but I don't really agree with much of the reasoning. Obviously its a Nordic driver and you can do what you want, but the suggested alternative is just bad in my opinion.

> Please see https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/app_dev/device_guides/nrf54l/pinmap.html

There is no way that it can work with subsystems like https://docs.zephyrproject.org/latest/connectivity/networking/conn_mgr/main.html (which don't have a defined "start/stop using" point in the application) apart from just permanently enabling CONSTLAT mode, which is obviously not required (as proven by the previous implementation). Even something as basic as a flash chip, where should that logic be called, keeping in mind that there are a multitude of subsystems that call into flash without going through the application main?

> The handling of quirks like this introduce uncontrolled latency, other side effects like not being able to use some peripherals from ISRs

If either of these are a problem for an application (certainly not true for every application), it can be resolved in a much more Zephyr API compliant way. Call `pm_device_runtime_get/release` in the driver that uses the bus. Transitioning the bus every transaction is already bad for throughput, i've profiled an increase from 6 Mbps to 8Mbps real throughput through a nRF7002 just by disabling that behaviour.

This is roughly the same requirement as enabling `CONSTLAT` around usage of the bus, fixes both the problems, and critically, can actually be merged into drivers. No-one is going to accept random calls to `nrf_sys_event_request_global_constlat()` in a generic driver that wants performant SPI, but calling `pm_device_runtime_get` on the bus for improved performance, sure.

> they are a nightmare to manage

From what perspective? I can agree that the previous implementation didn't look the nicest (with the `spim_has_cross_domain_connection` function), but thats not the only way to do it.

> hide the fact these configurations are highly inefficient and discouraged

As above, thats an implementation choice of the previous implementation, which also leads to the software complexity. By making it a devicetree property the user selects, there is absolutely no way it can be enabled without the user knowing the limitations.

At the end of the day, regardless of how inefficient it may or may not be, all the Nordic docs call this out as supported functionality. Noting the inefficiency sure, but not saying this in an unsupported configuration.
image
image
image

bjarki-andreasen:
> So the official line that this hardware feature is not going to be supported in software unless `main.c` knows about the SPI usage?

Correct, if a user has to make use of this quirk user must be aware of the quirk and needs to work around it.

PR #100894, opened 2025-12-11
boards: cy8proto_041tp: Enable GPIO driver and update clock driver

dharunkrithikk:
Add initial support for Infineon PSoC4 GPIO and CLOCK driver updates.

PR #100931, opened 2025-12-12
Cy8cproto_041tp SPI driver implementation

dharunkrithikk:
This PR depends on https://github.com/zephyrproject-rtos/zephyr/pull/100894


boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
```diff
@@ -0,0 +1,83 @@
...
+ * Copyright (c) 2025 Infineon Technologies AG,
+ * or an affiliate of Infineon Technologies AG.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/dt-bindings/input/input-event-codes.h>
+#include "cy8cproto_041tp-pinctrl.dtsi"
+
+uart0: &scb0 {
```
sreeramIfx, line 17: please move the changes other than SPI configuration to a separate commit with appropriate description

drivers/spi/Kconfig.infineon
```diff
@@ -39,6 +39,33 @@ config IFX_CAT1_SPI_DMA_TX_AUTO_TRIGGER
 	depends on IFX_CAT1_SPI_DMA
 	select DMA
 	help
-		Automatically trigger SPI Tx DMA after config
+	  Automatically trigger SPI Tx DMA after config
+
+config IFX_CAT1_SPI_OVERSAMPLE_DEFAULT
```
sreeramIfx: please drop cat1 from the kconfig names

drivers/spi/spi_infineon_pdl.c
```diff
@@ -25,13 +25,10 @@ LOG_MODULE_REGISTER(cat1_spi, CONFIG_SPI_LOG_LEVEL);
...
-#define IFX_CAT1_SPI_DEFAULT_OVERSAMPLE (4)
-#if defined(CONFIG_SOC_FAMILY_INFINEON_EDGE)
-#define IFX_CAT1_SPI_MIN_DATA_WIDTH (4)
-#else
-#define IFX_CAT1_SPI_MIN_DATA_WIDTH (8)
-#endif
-#define IFX_CAT1_SPI_MAX_DATA_WIDTH (32)
+#define IFX_CAT1_SPI_DEFAULT_OVERSAMPLE CONFIG_IFX_CAT1_SPI_OVERSAMPLE_DEFAULT
+
+#define IFX_CAT1_SPI_MIN_DATA_WIDTH CONFIG_IFX_CAT1_SPI_MIN_DATA_WIDTH
```
sreeramIfx: Can we use CONFIG_IFX_CAT1_SPI_MIN_DATA_WIDTH directly?

drivers/spi/spi_infineon_pdl.c
```diff
@@ -50,6 +47,12 @@ LOG_MODULE_REGISTER(cat1_spi, CONFIG_SPI_LOG_LEVEL);
 #define IFX_CAT1_SPI_ASYMM_PDL_FUNC_AVAIL
 #endif
 
+#if defined(CY_IP_MXSCB_INSTANCES)
+#define _IFX_SCB_ARRAY_SIZE (CY_IP_MXSCB_INSTANCES)
```
sreeramIfx:
```suggestion
#define IFX_SCB_ARRAY_SIZE (CY_IP_MXSCB_INSTANCES)
```

drivers/spi/spi_infineon_pdl.c
```diff
@@ -50,6 +47,12 @@ LOG_MODULE_REGISTER(cat1_spi, CONFIG_SPI_LOG_LEVEL);
...
+#elif defined(CY_IP_M0S8SCB_INSTANCES)
+#define _IFX_SCB_ARRAY_SIZE (CY_IP_M0S8SCB_INSTANCES)
```
sreeramIfx:
```suggestion
#define IFX_SCB_ARRAY_SIZE (CY_IP_M0S8SCB_INSTANCES)
```

drivers/spi/spi_infineon_pdl.c
```diff
@@ -570,6 +716,20 @@ static int ifx_cat1_spi_init(const struct device *dev)
...
 #endif
 
+/*
+ * Trigger level 7 to optimize DMA efficiency by enabling 8-byte burst transfers,
+ * which reduces overhead. Lower values would lead to frequent single-byte DMA operations,
+ * which are less efficient. This strikes a good balance between throughput and latency.
+ * (matching bare-metal implementation)
+ */
+#if defined(CONFIG_SOC_FAMILY_INFINEON_PSOC4)
+#define DEFAULT_TX_FIFO_TRIGGER_LEVEL 7
```
sreeramIfx: why is this specific to PSOC4?

sreeramIfx (changes requested):

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
dharunkrithikk: Since this was depend on previous PR it might have been an issue, now its merged so i think there wont be any issues now.

drivers/spi/Kconfig.infineon
dharunkrithikk: @sreeramIfx changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: This is was first written based on bare-metal implementation, but i think it wont affect the functionality, i have reverted it back.

sreeramIfx:
Please fix the order of the commits. Overlays are introduced before the driver support.
Also, add a separate commit for non-functional changes (renaming and format changes)

drivers/spi/spi_infineon_pdl.c
```diff
@@ -18,39 +18,37 @@ LOG_MODULE_REGISTER(cat1_spi, CONFIG_SPI_LOG_LEVEL);
...
 
-#ifdef CONFIG_IFX_CAT1_SPI_DMA
+#ifdef CONFIG_IFX_SPI_DMA
 #include <zephyr/drivers/dma.h>
 #endif
 
 #include <cy_scb_spi.h>
 #include <cy_trigmux.h>
 
-#define IFX_CAT1_SPI_DEFAULT_OVERSAMPLE (4)
```
sreeramIfx, line 30: Is it intentional to delete this code?

drivers/spi/spi_infineon_pdl.c
```diff
@@ -18,39 +18,37 @@ LOG_MODULE_REGISTER(cat1_spi, CONFIG_SPI_LOG_LEVEL);
...
+#define IFX_SPI_RSLT_TRANSFER_ERROR (-2)
+#define IFX_SPI_RSLT_CLOCK_ERROR    (-3)
 
 #if CY_SCB_DRV_VERSION_MINOR >= 20 && defined(COMPONENT_CAT1) && CY_SCB_DRV_VERSION_MAJOR == 3
-#define IFX_CAT1_SPI_ASYMM_PDL_FUNC_AVAIL
+#define IFX_SPI_ASYMM_PDL_FUNC_AVAIL
 #endif
 
-#ifdef CONFIG_IFX_CAT1_SPI_DMA
+#if defined(CY_IP_MXSCB_INSTANCES)
```
sreeramIfx, line 47: This should not be part of this commit

drivers/spi/spi_infineon_pdl.c
```diff
@@ -139,7 +137,120 @@ static void ifx_cat1_spi_cb_wrapper(const struct device *dev, uint32_t event);
 cy_rslt_t ifx_cat1_spi_transfer_async(const struct device *dev, const uint8_t *tx, size_t tx_length,
 				      uint8_t *rx, size_t rx_length);
 
-int32_t ifx_cat1_uart_get_hw_block_num(CySCB_Type *reg_addr);
+CySCB_Type *const IFX_SCB_BASE_ADDR[IFX_SCB_ARRAY_SIZE] = {
+#ifdef SCB0
+	SCB0,
```
sreeramIfx, line 145: please move to the right commit

drivers/spi/spi_infineon_pdl.c
```diff
@@ -593,6 +706,8 @@ static int ifx_cat1_spi_init(const struct device *dev)
 	PERI_INFO(n)
 #endif
 
+#define ADVANCED_SPI_FIELDS(n)
```
sreeramIfx, line 747: not the right commit

sreeramIfx (changes requested):

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: @sreeram yes, this is accessed through Kconfig, if any other board wants to change the oversample value based on user needs

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
sreeramIfx: please move this to the appropriate commit as well

dharunkrithikk: @sreeramIfx  changed

drivers/spi/Kconfig.infineon
```diff
@@ -33,4 +33,36 @@ config SPI_INFINEON_DMA
 	help
 		Enable DMA during usage of SPI driver.
 
+config IFX_SPI_DMA_TX_AUTO_TRIGGER
```
sreeramIfx: This commit still has changes other than dropping CAT1 from names. Please move to the appropriate commits

sreeramIfx (changes requested):

drivers/spi/Kconfig.infineon
dharunkrithikk: changed

sreeramIfx (dismissed):

manoj-aerlync (approved):

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
```diff
@@ -23,6 +23,16 @@ uart0: &scb0 {
 	pinctrl-names = "default";
 };
 
+spi1: &scb1 {
+	compatible = "infineon,spi";
+	status = "okay";
```
teburd: If there is nothing on the board connected to spi, the block should be disabled by default. It can be enabled in application overlays as needed.

drivers/spi/spi_infineon_pdl.c
```diff
@@ -142,7 +140,120 @@ static void ifx_cat1_spi_cb_wrapper(const struct device *dev, uint32_t event);
...
+#ifdef SCB15
+	15u,
+#endif
+};
+
+int32_t ifx_get_hw_block_num(CySCB_Type *reg_addr)
+{
+	uint32_t val;
+
+	for (val = 0u; val < IFX_SCB_ARRAY_SIZE; val++) {
```
teburd, line 249: Why didn't we stash the peripheral instance number in DT? Seems like we are doing at runtime what we could be doing at compile time here.

Non-blocking, will file a follow up on this

drivers/spi/spi_infineon_pdl.c
```diff
@@ -1013,6 +1176,15 @@ static cy_rslt_t spi_init_hw(const struct device *dev, cy_stc_scb_spi_config_t *
 		data->callback_data.callback_arg = NULL;
 		data->irq_cause = 0;
 
+#if defined(CONFIG_SOC_FAMILY_INFINEON_PSOC4)
+		/* PSOC4 slave mode: set active slave select BEFORE enabling SPI
+		 * (matching bare-metal initialization order)
+		 */
+		if (cfg->spiMode == CY_SCB_SPI_SLAVE) {
+			Cy_SCB_SPI_SetActiveSlaveSelect(config->reg_addr, CY_SCB_SPI_SLAVE_SELECT0);
```
teburd: The select line in use could be a GPIO or could be a hardware CS line, but this depends on DT. Here you have hard wired select0.

drivers/spi/spi_infineon_pdl.c
```diff
@@ -1062,6 +1233,13 @@ static void spi_irq_handler(const struct device *dev)
 
 	Cy_SCB_SPI_Interrupt(config->reg_addr, &(data->context));
 
+#if defined(CONFIG_SOC_FAMILY_INFINEON_PSOC4)
+	/* PSOC4 slave mode: completion is handled by polling in transceive() */
+	if (data->is_slave) {
```
teburd: SCB has interrupts available for this mode of operation. Why are we doing polling mode for psoc4 specifically?

samples/sensor/bme280/boards/cy8cproto_041tp.overlay
teburd, line 1: This file should not be here unless the board has a sensor. To test spi and i2c you should be using the tests, enabling the features on the board. Please submit a follow up PR to remove this and add this board to the appropriate tests for i2c and spi instead.

Non-blocking.

teburd (changes requested):
I'd say lets drop the support for peripheral (slave) mode in this PR to move forward as many of the issues that would be blockers here are around slave mode. We can revisit this mode in a future change.

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
dharunkrithikk: changed

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: This is actually based on the original implementation where the peripheral instance is obtained via a UART-specific API. To maintain driver integrity and avoid breaking the existing SPI abstraction, we've kept this approach confined to the SPI driver for now.

We agree this could be refactored in the future, but that would require broader changes to the peripheral access layer, outside the scope of this current PR.

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: Agreed,this should not be hard-wired and should depend on the DT configuration.
For now it matches the current board setup, this should be cleaned up and made configurable.

samples/sensor/bme280/boards/cy8cproto_041tp.overlay
dharunkrithikk: @teburd Will submit a follow up PR to remove this and add to zephyr tests

drivers/spi/spi_infineon_pdl.c
dharunkrithikk: We initially tried interrupts, but the slave wasn't receiving data when master transmits, so switched to polling which guarantee to receive data, following the pattern used in PSoC4 bare-metal reference code for slave configurations

boards/infineon/cy8cproto_041tp/cy8cproto_041tp-pinctrl.dtsi
```diff
@@ -23,3 +23,19 @@
 	drive-open-drain;
 	input-enable;
 };
+
+&p4_0_scb1_spi_m_select0 {
+	drive-push-pull;
```
teburd, line 28: Is there a reason we aren't doing this on the parts pinctrl?

teburd (approved):
Looks much better to me. I really don't love this instance mapping thing we have going on, but I see we need to fix some other things to correct that.

boards/infineon/cy8cproto_041tp/cy8cproto_041tp-pinctrl.dtsi
dharunkrithikk: This was actually intended for the sensor overlay so will move this to overlay with the follow up PR as mentioned

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
```diff
@@ -23,6 +23,16 @@ uart0: &scb0 {
 	pinctrl-names = "default";
 };
 
+spi1: &scb1 {
```
sreeramIfx, line 26: Why MOSI, MISO and CLK are configured here but not CS?

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
```diff
@@ -23,6 +23,16 @@ uart0: &scb0 {
...
+	compatible = "infineon,spi";
+	status = "disabled";
```
sreeramIfx, line 28: Isn't this already disabled?

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
dharunkrithikk: Those are hardware cs pins that can be muxed directly, but I'm following the standard approach using 
`cs-gpios` for software-controlled CS. Thought it might be more flexible with how most zephyr SPI devices are configured.

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
dharunkrithikk: Yes, it's already disabled. I just included the node to show the pin assignments and to have the complete configuration ready when needed.

boards/infineon/cy8cproto_041tp/cy8cproto_041tp_common.dtsi
sreeramIfx: I'm not sure if that is the standard approach. Given that these pins are not dedicated to SPI at the board level, these entries should be in the application overlay (not blocking)

sreeramIfx (approved):

PR #100935, opened 2025-12-12
Openthread border router fix compile warnings

Cristib05:
This PR aims to fix some issues that occurred when `CONFIG_NET_IPV4` or `CONFIG_OPENTHREAD_NAT64_TRANSLATOR` were not set. It also defined a dependency between those two, as there is no reason to have NAT64 enabled without IPV4 support.
Small changes have been done to OpenThread Kconfig to select more NAT64 features, if base NAT64 Translator is set.


rlubos (commented):
How about defining `CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_NAT64` where all configuration dependencies/selections could be resolved, and then use that new symbol in the code to conditionally compile NAT64 functionalities for Zephyr's OTBR?

Cristib05:
> How about defining `CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_NAT64` where all configuration dependencies/selections could be resolved, and then use that new symbol in the code to conditionally compile NAT64 functionalities for Zephyr's OTBR?

This could work as well. I will update the PR

rlubos (dismissed):

modules/openthread/Kconfig
```diff
@@ -439,6 +437,17 @@ config OPENTHREAD_ZEPHYR_BORDER_ROUTER_MSG_POOL_NUM
 	  This value represents the maximum number of messages from backbone interface
 	  that could be stored without being processed.
 
+config OPENTHREAD_ZEPHYR_BORDER_ROUTER_NAT64_TRANSLATOR
+	bool "Adds support for Border Router NAT64 translator functionality"
+	default true
```
pdgendt:
```suggestion
config OPENTHREAD_ZEPHYR_BORDER_ROUTER_NAT64_TRANSLATOR
	bool "Border Router NAT64 translator"
	default y
```

George-Stefan (dismissed):

pdgendt (dismissed):
NAK to prevent accidental fly-by merges

jukkar (dismissed):

rlubos (dismissed):

George-Stefan (dismissed):

George-Stefan (approved):

rlubos (approved):

modules/openthread/platform/infra_if.c
```diff
@@ -172,6 +175,17 @@ otError infra_if_deinit(void)
...
 		     error = OT_ERROR_FAILED);
 
+#if defined(CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_NAT64_TRANSLATOR)
+	VerifyOrExit(raw_infra_if_sock != -1);
+	VerifyOrExit(zsock_close(raw_infra_if_sock) == 0);
+
+	sockfd_raw[0].fd = -1;
+	raw_infra_if_sock = -1;
+
+	net_socket_service_register(&handle_infra_if_raw_recv, sockfd_raw,
```
M1cha, line 185: shouldn't this be `net_socket_service_unregister`? 🤔

fabiobaltieri: send a fix?

pdgendt: Not sure TBH, the register will refresh the socket descriptors?

Cristib05: I don't remember exactly, but I think I opted to use `net_socket_service_register` because output of `net sockets` was not properly updated. It's been a while since I wrote this code, if needed, I can have a look on this code.

PR #100942, opened 2025-12-12
drivers: flash: stm32 ospi: do not invalidate bet at end of loop

### LINKED ISSUE #100903: drivers: flash: stm32 ospi: detected erase type is always resetted
### Describe the bug

in the erase function, the driver tries to find the best erase command to issue. unfortunately pointer holding the found erase type is resetted at the end of the loop (line 1303) and therefor `NULL` in subsequent checks.

My guess is, most flash chips only support less than 4 erase commands. So when leaving the loop, and the assumption with less than 4 erase type is correct, the chosen erase type is of type sector and bet is `NULL`.

https://github.com/zephyrproject-rtos/zephyr/blob/15716542c20dea8d32a4e8e3072a90c6635ee589/drivers/flash/flash_stm32_ospi.c#L1276-L1317

removing the following lines seem to fix the issue. there's a comment explaining why resetting `bet` which I don't understand...
https://github.com/zephyrproject-rtos/zephyr/blob/15716542c20dea8d32a4e8e3072a90c6635ee589/drivers/flash/flash_stm32_ospi.c#L1300-L1303

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Linux
- Toolchain: Zephyr SDK
- Commit: 15716542c20dea8d32a4e8e3072a90c6635ee589

### Additional Context

_No response_

bearsh:
This modification prevents the bet pointer from being nulled at the end of the loop, which would otherwise render the search for a suitable erase type ineffective.

Without this change, the erase size defaults to one sector. If the chip defines fewer than JESD216_NUM_ERASE_TYPES (=4) erase types, this behavior still works, as the resulting command will correspond to a sector erase operation. However, if the chip defines all erase types, the resulting command will be the one specified in the erase type. But since bet is nulled, the erase size will incorrectly default to the sector size.

this fixes #100903

### DIFF
--- a/drivers/flash/flash_stm32_ospi.c
+++ b/drivers/flash/flash_stm32_ospi.c
@@ -1284,24 +1284,23 @@ static int flash_stm32_ospi_erase(const struct device *dev, off_t addr,
 				    && ((bet == NULL)
 					|| (etp->exp > bet->exp))) {
 					bet = etp;
-					cmd_erase.Instruction = bet->cmd;
-				} else if (bet == NULL) {
-					/* Use the default sector erase cmd */
-					if (dev_cfg->data_mode == OSPI_OPI_MODE) {
-						cmd_erase.Instruction = SPI_NOR_OCMD_SE;
-					} else {
-						cmd_erase.Instruction =
-							(stm32_ospi_hal_address_size(dev) ==
-							HAL_OSPI_ADDRESS_32_BITS)
-							? SPI_NOR_CMD_SE_4B
-							: SPI_NOR_CMD_SE;
-					}
 				}
-				/* Avoid using wrong erase type,
-				 * if zero entries are found in erase_types
-				 */
-				bet = NULL;
 			}
+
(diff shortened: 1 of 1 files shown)


mathieuchopstm (commented):
Wouldn't it also make sense to move L1288~L1297 outside of the `for` loop?

```c
/* pseudocode */
for (uint8_t ei = 0;
     /* ... */)
{
   if (etp->exp != 0 && ...) {
       bet = etp;
   }
}

if (bet != NULL) {
   cmd_erase.Instruction = bet->cmd;
} else if (data_mode == OPI_MODE) {
    Instruction = OCMD_SE;
} else {
    Instruction = (address_size == 32BITS) ? SE_4B : SE;
}
```

mathieuchopstm (approved):

erwango (approved):

etienne-lms (approved):

PR #100950, opened 2025-12-12
Openthread border router mdns packet limit

Cristib05:
This commit decides if an mDNS incoming packet should be processed by verifying that a number of OT message buffers are still available after a packet to OT message conversion would be perform.
In this way, it ensures that application would still be up and running, and it will not eventually end up in an assert statement raised by OT mDNS module. OT mDNS module raises an assert if there are no free available OT message buffers to encapsulate a TX message.

### DIFF
--- a/modules/openthread/Kconfig
+++ b/modules/openthread/Kconfig
@@ -257,6 +257,7 @@ config OPENTHREAD_CUSTOM_PARAMETERS
 
 config OPENTHREAD_NUM_MESSAGE_BUFFERS
 	int "The number of message buffers in the buffer pool"
+	default 512 if OPENTHREAD_ZEPHYR_BORDER_ROUTER
 	default 128
 	help
 	  "The number of message buffers in the buffer pool."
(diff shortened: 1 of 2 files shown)


modules/openthread/Kconfig
```diff
@@ -257,7 +257,8 @@ config OPENTHREAD_CUSTOM_PARAMETERS
 
 config OPENTHREAD_NUM_MESSAGE_BUFFERS
 	int "The number of message buffers in the buffer pool"
-	default 128
+	default 128 if !OPENTHREAD_ZEPHYR_BORDER_ROUTER
+	default 512 if OPENTHREAD_ZEPHYR_BORDER_ROUTER
```
rlubos: A little simpler IMHO (first successful match gets preference)
```suggestion
	default 512 if OPENTHREAD_ZEPHYR_BORDER_ROUTER
	default 128
```

modules/openthread/platform/mdns_socket.c
```diff
@@ -385,6 +387,28 @@ static void process_mdns_message(struct otbr_msg_ctx *msg_ctx_ptr)
 {
 	otMessageSettings ot_message_settings = {true, OT_MESSAGE_PRIORITY_NORMAL};
 	otMessage *ot_message = NULL;
+	otBufferInfo buffer_info;
+	uint16_t req_buff_num;
+
+	/** In large networks with high traffic, we have observed that mDNS module
+	 * might jump to assert when trying to allocate OT message buffers for a new
```
rlubos, line 394: What exactly asserts, is this on Zephyr side or OpenThread side? An assert on buffer allocation failure sounds plain wrong.

Cristib05: This is on OpenThread side, https://github.com/openthread/openthread/blob/main/src/core/net/mdns.cpp#L3654.
There is also a started discussion on OpenThread community: https://github.com/orgs/openthread/discussions/11647

rlubos (approved):

George-Stefan (approved):

PR #100953, opened 2025-12-12
cc13xx/cc26xx radio fixes

anobli:
This PR addresses some issues that happens using OpenThread joiner.
Some packet sent by the commissioner were never received by the joiner.
This improves set channel operation to only reconfigure channel (which implies many radio operations) only if this is required and this also makes sure any supported filter get applied by restarting RX operations. 

Note: This has been only tested using OpenThread but I would not expect any side effects for other radio stack based on 802.15.4.

### DIFF
--- a/drivers/ieee802154/ieee802154_cc13xx_cc26xx.c
+++ b/drivers/ieee802154/ieee802154_cc13xx_cc26xx.c
@@ -223,6 +223,7 @@ static int ieee802154_cc13xx_cc26xx_do_set_channel(const struct device *dev,
 		ret = -EIO;
 		goto out;
 	}
+	drv_data->rx_handle = cmd_handle;
 
 	ret = 0;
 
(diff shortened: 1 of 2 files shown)


drivers/ieee802154/ieee802154_cc13xx_cc26xx.c
```diff
@@ -226,6 +232,12 @@ static int ieee802154_cc13xx_cc26xx_set_channel(const struct device *dev,
 	return ret;
 }
 
+static int ieee802154_cc13xx_cc26xx_set_channel(const struct device *dev,
+						uint16_t channel)
+{
+	return ieee802154_cc13xx_cc26xx_do_set_channel(dev, channel, 0);
```
dsseng: nit

```suggestion
	return ieee802154_cc13xx_cc26xx_do_set_channel(dev, channel, false);
```

drivers/ieee802154/ieee802154_cc13xx_cc26xx.c
```diff
@@ -172,15 +172,20 @@ static inline int ieee802154_cc13xx_cc26xx_channel_to_frequency(
 	}
 }
 
-static int ieee802154_cc13xx_cc26xx_set_channel(const struct device *dev,
-						uint16_t channel)
+static int ieee802154_cc13xx_cc26xx_do_set_channel(const struct device *dev,
+						   uint16_t channel, bool force)
```
dsseng: Either skip adding extra padding, or use the same type (tabs here)

```suggestion
						uint16_t channel, bool force)
```

dsseng (commented):
Thanks, some style nitpicks, non-blocking

dsseng (commented):
looks good from the general perspective. I'm not involved with the subsys enough to have any other comments

jukkar (approved):

bogdanovs (approved):

cfriedt (approved):

vaishnavachath (approved):

PR #100959, opened 2025-12-12
dts: bindings: Move the dts clips to the examples

rruuaanng:
```
description: |
  SparkFun SerLCD Dot Character VFD Controller/Driver IC
  Example:
    &i2c1 {
      serlcd@72 {
        compatible = "sparkfun,serlcd";
        reg = <0x72>;
        columns = <16>;
        rows = <2>;
        command-delay-ms = <10>;
        special-command-delay-ms = <50>;
      };
    };
```
move to
```
examples:
  - |
    &i2c1 {
      serlcd@72 {
        compatible = "sparkfun,serlcd";
        reg = <0x72>;
        columns = <16>;
        rows = <2>;
        command-delay-ms = <10>;
        special-command-delay-ms = <50>;
      };
    };
```

### DIFF
--- a/dts/bindings/auxdisplay/gpio-7-segment.yaml
+++ b/dts/bindings/auxdisplay/gpio-7-segment.yaml
@@ -17,30 +17,6 @@ description: |
   high and active low respectively, meaning that the current flows from
   digit-gpios to segment-gpios. Vice versa for common cathode.
 
-  Example:
-
-  #include <zephyr/dt-bindings/gpio/gpio.h>
-
-  / {
-          auxdisplay_0: digi-display {
-                  compatible = "gpio-7-segment";
-                  columns = <3>;
-                  rows = <1>;
-                  segment-gpios = <&gpio0 0 GPIO_ACTIVE_LOW>, /* A */
-                                  <&gpio0 1 GPIO_ACTIVE_LOW>, /* B */
-                                  <&gpio0 2 GPIO_ACTIVE_LOW>, /* C */
-                                  <&gpio0 3 GPIO_ACTIVE_LOW>, /* D */
-                                  <&gpio0 4 GPIO_ACTIVE_LOW>, /* E */
-                                  <&gpio0 5 GPIO_ACTIVE_LOW>, /* F */
-                                  <&gpio0 6 GPIO_ACTIVE_LOW>, /* G */
-                                  <&gpio0 7 GPIO_ACTIVE_LOW>; /* DP */
-                  digit-gpios = <&gpio1 0 GPIO_ACTIVE_HIGH>,  /* DIG1 */
-                                <&gpio1 1 GPIO_ACTIVE_HIGH>,  /* DIG2 */
-                                <&gpio1 2 GPIO_ACTIVE_HIGH>;  /* DIG3 */
-                  refresh-period-ms = <1>;
-          };
-  };
-
 compatible: "gpio-7-segment"
 
 include: auxdisplay-device.yaml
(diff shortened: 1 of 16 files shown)


rruuaanng:
~Let's wait for it to be merged.~ I almost forgot, I can address it in this PR.

dts/bindings/cache/bflb,l1c.yaml
```diff
@@ -39,3 +24,14 @@ properties:
...
+examples:
+  - |
+    soc {
+        cache {
+          compatible = "bflb,l1c";
+          dcache-ways-disabled = <0>;
+        };
+    };
+
+    The cache configuration is specific to the platform, this only provides controls for it.
```
josuah: This looks like part of the description.

dts/bindings/timer/st,stm32-lptim.yaml
```diff
@@ -53,3 +47,12 @@ properties:
...
         pre-defined setting. If not, an error is raised.
+
+examples:
+  - |
+    stm32_lp_tick_source: &lptim1 {
+      status = "okay";
+    }
+
+    The lptim node to be used for counting ticks during lowpower modes
+    must be named stm32_lp_tick_source in the DTS.
```
josuah, line 56: This looks like part of the description.

josuah (commented):
The text below the examples seems more like description to me in this case. What do you think?

Thank you for the change!

dts/bindings/can/microchip,mcp251xfd.yaml
```diff
@@ -4,23 +4,7 @@
...
-          status = "okay";
-
-          spi-max-frequency = <18000000>;
-          int-gpios = <&mikrobus_header 7 GPIO_ACTIVE_LOW>;
-          reg = <0x0>;
-          osc-freq = <40000000>;
-
-      };
-  };
+  The MCP251XFD node is defined on an SPI bus.
```
henrikbrixandersen: This doesn't really provide any value as the MCP251xFD is SPI peripheral.

henrikbrixandersen (changes requested):
Today, the examples are rendered in the documentation (see e.g. https://docs.zephyrproject.org/latest/build/dts/api/bindings/can/nxp%2Cflexcan.html#dtbinding-nxp-flexcan).

With this change, the examples are no longer rendered in the documentation?

dts/bindings/timer/st,stm32-lptim.yaml
rruuaanng: These declarations seem to be specific to the example nodes. If we include them in the description, users might not understand what they refer to. What do you think?

henrikbrixandersen:
> > Today, the examples are rendered in the documentation (see e.g. https://docs.zephyrproject.org/latest/build/dts/api/bindings/can/nxp%2Cflexcan.html#dtbinding-nxp-flexcan).
> > With this change, the examples are no longer rendered in the documentation?
> 
> Yes, I believe we should keep the `Description` focused on explaining the binding, and provide an `Example` in the documentation to show a sample node. I think it should go at the very bottom of the documentation page.

My point is, we can't merge this until the `examples` section is rendered in the documentation as well - otherwise this information, which is now visible to the user, will become hidden.

dts/bindings/timer/st,stm32-lptim.yaml
kartben: then make it a dts comment please. The examples are meant to be valid DTS snippets, I'd think

dts/bindings/auxdisplay/gpio-7-segment.yaml
```diff
@@ -86,3 +62,27 @@ properties:
       This is the time between the display of each digit. The refresh period
       must be long enough to allow the segments to be driven and the digit to
       be selected.
+
+examples:
+  - |
```
kartben, line 67: those extra "|" characters all seem unnecessary (unless I am missing something)

rruuaanng: https://www.asdf-format.org/projects/asdf-standard/en/1.0.2/schemas/yaml_schema.html, In `#style`

It's just a common YAML convention for multi-line text. Just like using `description`. :)

kartben (dismissed):

kartben:
@henrikbrixandersen please revisit now that https://github.com/zephyrproject-rtos/zephyr/commit/4c8cb73198809e25e48d599ddc405fbf15e89869 has been merged

kartben:
adding myself as an assignee as this is documentation as much as it is DT and DT maintainer is currently AWOL.

dts/bindings/can/nxp,flexcan-fd.yaml
```diff
@@ -6,7 +6,13 @@ description: |
 
   This is a specialization of the NXP FlexCAN CAN controller with support for CAN FD.
 
-  Example:
+compatible: "nxp,flexcan-fd"
+
+include: ["nxp,flexcan.yaml", "can-fd-controller.yaml",
+          "pinctrl-device.yaml", "nxp,rdc-policy.yaml"]
+
```
kartben, line 19: @rruuaanng this needs fixing / likely shouldn't be there in the first place :)

henrikbrixandersen: Correct. The includes should be there.

henrikbrixandersen (changes requested):
Please address the comments regarding the unrelated changes to the NXP FlexCAN FD binding.

dts/bindings/can/nxp,flexcan-fd.yaml
rruuaanng: Sorry, it looks like I kept the conflicts :(

kartben (approved):

thedjnK (approved):
auxdisplay parts OK

henrikbrixandersen (approved):

PR #100960, opened 2025-12-12
arch: riscv: fix mstatus restore order in trap exit with userspace

sylvioalves:
When returning from a trap to user mode with CONFIG_PMP_KERNEL_MODE_DYNAMIC enabled, z_riscv_pmp_usermode_enable() is called to reconfigure PMP entries. This function modifies mstatus (clears MPRV, sets MPP).

The original code restored mstatus from the exception frame before calling z_riscv_pmp_usermode_enable(), causing the function's mstatus modifications to be lost. This resulted in incorrect PMP enforcement after returning to user mode.

Move the mstatus restore to after the userspace PMP reconfiguration block to ensure the saved mstatus value is properly restored after all PMP setup is complete.

### DIFF
--- a/arch/riscv/core/isr.S
+++ b/arch/riscv/core/isr.S
@@ -757,7 +757,6 @@ fp_trap_exit:
 #endif /* CONFIG_CLIC_SUPPORT_INTERRUPT_LEVEL */
 
 	csrw mepc, t0
-	csrw mstatus, t2
 
 #ifdef CONFIG_USERSPACE
 	/*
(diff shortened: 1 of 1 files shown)


sylvioalves:
> Won't work. The call to `z_riscv_pmp_usermode_enable()` may clobber t2. And the change to `z_riscv_pmp_usermode_enable()` is legitimate. If this was wrong user threads would have been seriously broken for over 3 years. What is the problem you're seeing?

I had issue during pmp for esp32c6 and that sounded correct but it isn't indeed. Thanks!

PR #100964, opened 2025-12-12
llext: Skip MPU region alignment if userspace is disabled.

iabdalkader:
llext aligns and rounds up memory regions to satisfy MPU requirements, which can waste significant memory by relocating read-only regions, with power-of-two alignment on some architectures. However, these MPU-aligned regions are never actually used by the MPU unless userspace (CONFIG_USERSPACE) is enabled.

I could have missed it, but as far as I can tell, there are No MPU regions created for LLEXT memory regions with CONFIG_USERSPACE=n at all. As for the MMU, not 100% sure, but it needs this at least to fix permissions.

### DIFF
--- a/subsys/llext/llext_mem.c
+++ b/subsys/llext/llext_mem.c
@@ -96,25 +96,28 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
 	 * program-accessible data (not to string tables, for example).
 	 */
 	if (region->sh_flags & SHF_ALLOC) {
-		if (IS_ENABLED(CONFIG_MPU_REQUIRES_POWER_OF_TWO_ALIGNMENT)) {
-			/* Some MPU architectures (ARMv7-M, older ARC) require regions
-			 * to be sized and aligned to the same power of two.
-			 */
-			uintptr_t block_sz = MAX(MAX(region_alloc, region_align), LLEXT_PAGE_SIZE);
-
-			block_sz = 1 << LOG2CEIL(block_sz); /* align to next power of two */
-			region_alloc = block_sz;
-			region_align = block_sz;
-		} else if (IS_ENABLED(CONFIG_ARM_MPU) || IS_ENABLED(CONFIG_ARC_MPU)) {
-			/* ARMv8-M and newer ARC MPUs use 32-byte alignment. */
-			region_alloc = ROUND_UP(region_alloc, LLEXT_PAGE_SIZE);
-			region_align = MAX(region_align, LLEXT_PAGE_SIZE);
-		} else if (IS_ENABLED(CONFIG_MMU)) {
+		if (IS_ENABLED(CONFIG_MMU)) {
 			/* MMU targets map memory in page-sized chunks. Round
 			 * the region to multiples of those.
 			 */
 			region_alloc = ROUND_UP(region_alloc, LLEXT_PAGE_SIZE);
 			region_align = MAX(region_align, LLEXT_PAGE_SIZE);
+		} else if (IS_ENABLED(CONFIG_USERSPACE)) {
+			if (IS_ENABLED(CONFIG_MPU_REQUIRES_POWER_OF_TWO_ALIGNMENT)) {
+				/* Some MPU architectures (ARMv7-M, older ARC) require regions
+				 * to be sized and aligned to the same power of two.
+				 */
+				uintptr_t block_sz =
+					MAX(MAX(region_alloc, region_align), LLEXT_PAGE_SIZE);
+
+				block_sz = 1 << LOG2CEIL(block_sz); /* align to next power of two */
+				region_alloc = block_sz;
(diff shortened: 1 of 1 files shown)


subsys/llext/llext_mem.c
```diff
@@ -96,20 +96,25 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
...
 
-			block_sz = 1 << LOG2CEIL(block_sz); /* align to next power of two */
-			region_alloc = block_sz;
-			region_align = block_sz;
-		} else if (IS_ENABLED(CONFIG_ARM_MPU) || IS_ENABLED(CONFIG_ARC_MPU)) {
-			/* ARMv8-M and newer ARC MPUs use 32-byte alignment. */
-			region_alloc = ROUND_UP(region_alloc, LLEXT_PAGE_SIZE);
-			region_align = MAX(region_align, LLEXT_PAGE_SIZE);
-		} else if (IS_ENABLED(CONFIG_MMU)) {
+		if (IS_ENABLED(CONFIG_MMU)) {
```
teburd: Why is this no longer an else if here, the logic now looks broken to me

subsys/llext/llext_mem.c
```diff
@@ -96,20 +96,25 @@ static int llext_copy_region(struct llext_loader *ldr, struct llext *ext,
 	 * program-accessible data (not to string tables, for example).
 	 */
 	if (region->sh_flags & SHF_ALLOC) {
-		if (IS_ENABLED(CONFIG_MPU_REQUIRES_POWER_OF_TWO_ALIGNMENT)) {
-			/* Some MPU architectures (ARMv7-M, older ARC) require regions
-			 * to be sized and aligned to the same power of two.
-			 */
-			uintptr_t block_sz = MAX(MAX(region_alloc, region_align), LLEXT_PAGE_SIZE);
+		if (IS_ENABLED(CONFIG_USERSPACE)) {
```
teburd: This isn't actually the thing you need to check on, you should be checking on if MPU is enabled *and* that the default MPU configuration is setup (prevents execution outside of the firmware .text)

teburd (changes requested):
I appreciate the idea that we want to save some memory here, but the changes needed aren't quite right. It's not userspace you are really looking for, its the default MPU enablement and configuration which prevents heap allocated memory from being executable. Likely need a Kconfig to decide *that* outside of userspace itself.

subsys/llext/llext_mem.c
iabdalkader: > This isn't actually the thing you need to check on,

If you look at `llext_init_mem_part` and `llext_add_domain`, in the same file, the partitions are added to `mem_parts` and then added to the domain, **only if `CONFIG_USERSPACE` is set**. So, if I understand correctly, if `CONFIG_USERSPACE` is not set, no MPU regions will be created. I don't think we need to test anything else, but I could be missing something. Am I?

Note: I will fix the if/else.

subsys/llext/llext_mem.c
iabdalkader: > Why is this no longer an else if here, the logic now looks broken to me

So for MMU targets, the region should always be aligned & rounded up to page size, regardless of `CONFIG_USERSPACE`, because the MMU maps pages. For example  `llext_adjust_mmu_permissions` expects a page-sized chunk has been allocated:
https://github.com/zephyrproject-rtos/zephyr/blob/f16c568c9924eedf245688dab1ffbaaef77aa5e7/subsys/llext/llext_mem.c#L291

For MPU targets, we only need to align & round up if userspace is used, otherwise no MPU regions are created and it's just wasted  memory. So I changed the logic now to look like this:

```C
if (IS_ENABLED(CONFIG_MMU)) {
  // Always align & roundup.
} else if (IS_ENABLED(CONFIG_USERSPACE)) {
  // For everything else, align & roundup only if userspace is enabled
 // more logic here to determine alignment/roundup size.
}
```

pillo79 (approved):
This looks good to me. It is for when `MMU` is enabled, but in a permissive enough way so that `USERSPACE`, and all the extra setup that it brings with, is not required by the app. The savings can be very important in those corner cases (and otherwise, it's a no-op).

After the update, it's also consistent with the conditionals in the rest of LLEXT.

teburd (approved):
Today the only Kconfig we have I suppose is USERSPACE to really decide on this, so I think its fine to go in.

@pillo79 ultimately though we need to differentiate between MPU usage and USERSPACE, which no one has gone and done. These are two distinct things, you can use MPU without USERSPACE just like you can use MMU without USERSPACE.

For example if you enable MPU the default configuration when you boot your mcu will prevent execution outside of the predefined .text region as far as I can remember. Did we somehow get around this here to allow for allocating executable memory that I've forgotten? Could be! I'd expect this to fail on hardware, but so be it, We can always revert.

PR #100969, opened 2025-12-12
Improve testcoverage of clang native sim

ivaniushkov:
enabled -Wunaligned-access check in native clang builds

### DIFF
--- a/boards/native/common/natsim_config.cmake
+++ b/boards/native/common/natsim_config.cmake
@@ -17,6 +17,12 @@ if("${LINKER}" STREQUAL "lld")
   target_link_options(native_simulator INTERFACE "-fuse-ld=lld")
 endif()
 
+# Enable -Wunaligned-access for Clang to catch alignment issues when running native_sim tests.
+# This warning is enabled by default for ARM embedded targets not supporting unaligned access but not for native targets.
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+  zephyr_compile_options(-Wunaligned-access)
+endif()
+
 set(nsi_config_content
   ${nsi_config_content}
   "NSI_AR:=${CMAKE_AR}"


ivaniushkov:
> Have you considered enabling this warning for all targets by enabling the flag in https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/compiler/clang/compiler_flags.cmake?

I am afraid it can create more noise and annoyance because it will force unaligned-access checks for the platforms that support unaligned access.

tpambor (approved):

boards/native/common/natsim_config.cmake
```diff
@@ -17,6 +17,12 @@ if("${LINKER}" STREQUAL "lld")
   target_link_options(native_simulator INTERFACE "-fuse-ld=lld")
 endif()
 
+# Enable -Wunaligned-access for Clang to catch alignment issues when running native_sim tests.
```
aescolar, line 20: Fundamentally:
* I'm not fond of enabling this warning *by default*, and less for one set of targets in particular, where it is not needed, but so that these targets act as a canary by default. (unaligned accesses are not a real problem in x86 and many know it)
With this change we'd start breaking tests with false positives, or, where people know they are doing unaligned accesses and are fine about it, or, they don't known and don't care, or, they don't want to be slowed down by what may be for them a pedantic warning/error.
* Background info: For native simulator based targets, users can enable today CONFIG_UBSAN which will warn them about this at runtime.
* I'd be fine adding this optional under a kconfig option, which would default to n, which then users could set for any target.
  * I would be less against having this as kconfig option which defaults to y, which users can easily set to n.

More circumstantial comments:
* This file is used to pass options to the native simulator runner build, so this would not be the place to add this.
If anything I would add something like this in the same place as we set other warnings: https://github.com/zephyrproject-rtos/zephyr/blob/main/CMakeLists.txt#L398
* Instead of if'ing in cmake for the compiler where we apply it, I guess this should be a compiler property (like https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/compiler/clang/compiler_flags.cmake#L93 ) which just defaults empty for other compilers.
* Just a note: [this warnig](https://clang.llvm.org/docs/DiagnosticsReference.html#wunaligned-access) is quite close to clang+gcc's https://clang.llvm.org/docs/DiagnosticsReference.html#waddress-of-packed-member

aescolar (changes requested):

boards/native/common/natsim_config.cmake
tpambor: > Fundamentally:
> 
> * I'm not fond of enabling this warning _by default_, and less for one set of targets in particular, where it is not needed, but so that these targets act as a canary by default. (unaligned accesses are not a real problem in x86 and many know it)
>   With this change we'd start breaking tests with false positives, or, where people know they are doing unaligned accesses and are fine about it, or, they don't known and don't care, or, they don't want to be slowed down by what may be for them a pedantic warning/error.

I agree. Nevertheless, I think we should ensure that code, that is shared by platforms that support or not unaligned accesses, does not rely on them. Currently, these issues can only be detected at runtime using UBSAN (which is not enabled in CI). This PR seems more like a workaround because AFAIK the only platform build in CI with clang is native_sim. I think a better approach would be to build some more platforms with clang (in a weekly build), e.g. some cortex-m0 platforms (which don't support unaligned access) by default and therefore already have this warning enabled by default.

https://github.com/zephyrproject-rtos/zephyr/blob/14fb2c069113ae2689bdda4bc0507e695b5fc023/.github/workflows/clang.yaml#L116-L117

ivaniushkov: >  I think a better approach would be to build some more platforms with clang (in a weekly build), e.g. some cortex-m0 platforms (which don't support unaligned access) by default and therefore already have this warning enabled by default.

I agree, I would also prefer this approach instead of a workaround suggested in this PR.
Does zephyr toolchain or CI have [ARM Toolchain for Embedded](https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded) or something similar allowing to build with clang for cortex-m targets? It is a good idea to build tests and, if possible, samples with ARM LLVM instead of enabling -Wunaligned-access in native_sim

aescolar: > Does zephyr toolchain or CI have [ARM Toolchain for Embedded](https://github.com/arm/arm-toolchain/tree/arm-software/arm-software/embedded) or something similar

I don't think so. So far we have the Zephyr SDK 0.17.4, which is gcc based (until the 1.0 sdk which will also include llvm for targetting arm and riscv AFAIR).
This may be better discussed in the infrastructure discord channel though.

PR #100971, opened 2025-12-12
soc: riscv: esp32c6: add RISCV_PMP and USERSPACE support

sylvioalves:
## Summary
- Add infrastructure for SoC-specific PMP regions via iterable sections
- Enable PMP and userspace support for ESP32-C6

## Description

This PR introduces a mechanism for RISC-V SoCs to define custom PMP regions
that are automatically programmed during PMP initialization. This is needed
for SoCs like ESP32-C6 where code execution spans multiple memory regions
(SoC ROM, IRAM, flash) that all need PMP coverage for userspace to work.

### Changes

**arch/riscv (generic infrastructure):**
- Add `struct pmp_soc_region` and `PMP_SOC_REGION_DEFINE()` macro to `pmp.h`
- Add linker snippet `pmp.ld` for iterable section collection
- Iterate over SoC-defined regions in `z_riscv_pmp_init()`

**soc/espressif/esp32c6:**
- Define PMP regions for SoC ROM (0x40000000) and IRAM text section
- Enable `PMP_NO_LOCK_GLOBAL` (required because PMP init runs from IRAM
  while the main rom region is in flash-mapped memory)
- Add userspace linker support in `default.ld`
- Add `soc_rom` device tree node for ROM region address/size


sylvioalves:
This PR can be split in 2 if mandatory to make it clear about riscv arch changes and soc itself. However, combining it here gives a better view of the SoC dependency. I gently ask the riscv maintainers to check the PMP memory changes and suggest alternatives if needed. Thanks!

fkokosinski:
> arch/riscv (generic infrastructure):

Question: have you considered using memory region with attributes? See e.g. here: https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/arch/riscv/pmp/mem-attr-entries/memattr_mapping.overlay

uLipe (dismissed):

arch/riscv/include/pmp.h
```diff
@@ -16,6 +17,45 @@
...
+ * This macro allows SoCs to register memory regions that require PMP
+ * protection. The regions are collected at link time and programmed
+ * during PMP initialization.
+ *
+ * @param name Unique identifier for this region
+ * @param _start Start address of the region (pointer or linker symbol)
+ * @param _end End address of the region (pointer or linker symbol, exclusive)
+ * @param _perm PMP permission flags (PMP_R | PMP_W | PMP_X)
+ */
+#define PMP_SOC_REGION_DEFINE(name, _start, _end, _perm)                                           \
```
fkokosinski: Shouldn't this macro be provided from `inculde/zephyr/arch/riscv`? It's meant to be used by the SoC layer

sylvioalves: Yes, thanks for spotting this. Fixed.

tests/arch/riscv/pmp/pmp-custom-regions/src/main.c
```diff
@@ -0,0 +1,163 @@
+/*
+ * Copyright (c) 2025 Espressif Systems (Shanghai) Co., Ltd.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/ztest.h>
+#include <zephyr/arch/riscv/csr.h>
+#include <zephyr/arch/riscv/pmp.h>
+#include <pmp.h>
```
fkokosinski:
```suggestion
#include <zephyr/arch/riscv/pmp.h>
```

fkokosinski (commented):
One final nit, otherwise LGTM.

tests/arch/riscv/pmp/pmp-custom-regions/src/main.c
sylvioalves: in fact, `#include <zephyr/arch/riscv/pmp.h>` is already there.
It needed `#include <kernel_internal.h>` instead. Updated.

fkokosinski (approved):
LGTM

tgorochowik (approved):

PR #89384, opened 2025-05-02
NXP drivers: pwm mcux sctimer: adds PM low power support

alrodlim:
Enables Sleep mode (PM3) in RW61x.

Tested tests/drivers/pwm/pwm_api on FRDM-RW612. Verified that RW612 goes into standby mode once all channels are inactive (pulse = 0).

PR #100973, opened 2025-12-12
drivers: pwm: sctimer: Add PowerManagement support

davidmissael:
pwm_mcux_sctimer driver doesn't support power management actions. PM Actions added, as well support to save/restore for pwm channels configurations after exitig from PM3.

### DIFF
--- a/drivers/pwm/pwm_mcux_sctimer.c
+++ b/drivers/pwm/pwm_mcux_sctimer.c
@@ -8,6 +8,7 @@
 #define DT_DRV_COMPAT nxp_sctimer_pwm
 
 #include <errno.h>
+#include <zephyr/pm/device.h>
 #include <zephyr/drivers/pwm.h>
 #include <fsl_sctimer.h>
 #include <fsl_clock.h>
@@ -27,6 +28,14 @@ LOG_MODULE_REGISTER(pwm_mcux_sctimer, CONFIG_PWM_LOG_LEVEL);
 
 /* Constant identifying that no event number has been set */
 #define EVENT_NOT_SET FSL_FEATURE_SCT_NUMBER_OF_EVENTS
+
+#ifdef CONFIG_PM_DEVICE
+typedef struct pwm_channel_config {
+	uint32_t period_cycles;
+	uint32_t duty_cycles;
+	pwm_flags_t flags;
+} pwm_channel_config_t;
+#endif /* CONFIG_PM_DEVICE */
 
 struct pwm_mcux_sctimer_config {
 	SCT_Type *base;
@@ -66,13 +75,21 @@ struct pwm_mcux_sctimer_capture_data {
 struct pwm_mcux_sctimer_data {
 	uint32_t event_number[CHANNEL_COUNT];
 	sctimer_pwm_signal_param_t channel[CHANNEL_COUNT];
+#ifdef CONFIG_PM_DEVICE
+	pwm_channel_config_t pwm_channel_config[CHANNEL_COUNT];
+#endif /* CONFIG_PM_DEVICE */
 	uint32_t match_period;
 	uint32_t configured_chan;
 #ifdef CONFIG_PWM_CAPTURE
 	uint32_t match_event;
 	struct pwm_mcux_sctimer_capture_data capture_data[CAPTURE_CHANNEL_COUNT];
 #endif /* CONFIG_PWM_CAPTURE */
+	bool pwm_channel_active;
 };
+
+#ifdef CONFIG_PM_DEVICE
+static void mcux_sctimer_restore_chn_config(const struct device *dev);
+#endif /* CONFIG_PM_DEVICE */
 
 /* Helper to setup channel that has not previously been configured for PWM */
 static int mcux_sctimer_new_channel(const struct device *dev,
@@ -167,6 +184,16 @@ static int mcux_sctimer_pwm_set_cycles(const struct device *dev,
 		LOG_ERR("Channel can not be set to inactive level");
 		return -ENOTSUP;
 	}
+
+	#ifdef CONFIG_PM_DEVICE
+	/* Save channel configuration when CONFIG_PM_DEVICE is defined
+	 * to recover it when exiting from PM3
+	 */
+	data->pwm_channel_config[channel].period_cycles = period_cycles;
+	data->pwm_channel_config[channel].duty_cycles = pulse_cycles;
+	data->pwm_channel_config[channel].flags = flags;
+	data->pwm_channel_active = true;
+	#endif /* CONFIG_PM_DEVICE */
 
 	if ((flags & PWM_POLARITY_INVERTED) == 0) {
 		data->channel[channel].level = kSCTIMER_HighTrue;
@@ -658,7 +685,27 @@ static void mcux_sctimer_isr(const struct device *dev)
 }
 #endif /* CONFIG_PWM_CAPTURE */
 
-static int mcux_sctimer_pwm_init(const struct device *dev)
+#ifdef CONFIG_PM_DEVICE
+static void mcux_sctimer_restore_chn_config(const struct device *dev)
+{
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+	uint8_t channel;
+
+	for (channel = 0; channel < CHANNEL_COUNT; channel++) {
+		/* Only restore the channels configured
+		 * before entering into a low power mode
+		 */
+		if (data->pwm_channel_config[channel].period_cycles != 0) {
+			mcux_sctimer_pwm_set_cycles(dev, channel,
+			data->pwm_channel_config[channel].period_cycles,
+			data->pwm_channel_config[channel].duty_cycles,
+			data->pwm_channel_config[channel].flags);
+		}
+	}
+}
+#endif /* CONFIG_PM_DEVICE */
+
+static int mcux_sctimer_pwm_init_common(const struct device *dev)
 {
 	const struct pwm_mcux_sctimer_config *config = dev->config;
 	struct pwm_mcux_sctimer_data *data = dev->data;
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
 	return 0;
 }
 
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
+#ifdef CONFIG_PM_DEVICE
+	const struct pwm_mcux_sctimer_config *config = dev->config;
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+	uint8_t channel;
+#endif /* CONFIG_PM_DEVICE */
+
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+#ifdef CONFIG_PM_DEVICE
+		SCTIMER_StartTimer(config->base, kSCTIMER_Counter_U);
+#endif /* CONFIG_PM_DEVICE */
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+#ifdef CONFIG_PM_DEVICE
+		/* Halt the timer counters */
+		SCTIMER_StopTimer(config->base, kSCTIMER_Counter_U);
+		/* Force the configured channels to inactive state */
+		for (channel = 0; channel < CHANNEL_COUNT; channel++) {
+			if (data->pwm_channel_config[channel].period_cycles != 0) {
+				if (data->channel[channel].level == kSCTIMER_HighTrue) {
+					config->base->OUTPUT &= ~(1U << channel);
+				} else {
+					config->base->OUTPUT |= (1U << channel);
+				}
+			}
+		}
+#endif /* CONFIG_PM_DEVICE */
+		break;
+	case PM_DEVICE_ACTION_TURN_OFF:
+		break;
+	case PM_DEVICE_ACTION_TURN_ON:
+		mcux_sctimer_pwm_init_common(dev);
+#ifdef CONFIG_PM_DEVICE
+		if (data->pwm_channel_active == true) {
+			mcux_sctimer_restore_chn_config(dev);
+		}
+#endif /* CONFIG_PM_DEVICE */
+		break;
+	default:
+		return -ENOTSUP;
+	}
+	return 0;
+}
+
+static int mcux_sctimer_pwm_init(const struct device *dev)
+{
+	/* Rest of the init is done from the PM_DEVICE_TURN_ON action
+	 * which is invoked by pm_device_driver_init().
+	 */
+	return pm_device_driver_init(dev, mcux_sctimer_pwm_pm_action);
+}
+
 static DEVICE_API(pwm, pwm_mcux_sctimer_driver_api) = {
 	.set_cycles = mcux_sctimer_pwm_set_cycles,
 	.get_cycles_per_sec = mcux_sctimer_pwm_get_cycles_per_sec,
@@ -758,10 +859,11 @@ static void mcux_sctimer_config_func_##n(const struct device *dev) \
 	static struct pwm_mcux_sctimer_data pwm_mcux_sctimer_data_##n;			\
 											\
 	SCTIMER_CONFIG_FUNC(n)								\
-	SCTIMER_INIT_CFG(n);	\
+	SCTIMER_INIT_CFG(n);								\
+	PM_DEVICE_DT_INST_DEFINE(n, mcux_sctimer_pwm_pm_action);			\
 	DEVICE_DT_INST_DEFINE(n,							\
 			      mcux_sctimer_pwm_init,					\
-			      NULL,							\
+			      PM_DEVICE_DT_INST_GET(n),					\
 			      &pwm_mcux_sctimer_data_##n,				\
 			      &pwm_mcux_sctimer_config_##n,				\
 			      POST_KERNEL, CONFIG_PWM_INIT_PRIORITY,			\


FelixWang47831 (dismissed):

Holt-Sun (dismissed):

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -97,6 +114,9 @@ static int mcux_sctimer_pwm_set_cycles(const struct device *dev,
 		return -ENOTSUP;
 	}
 
+	mcux_sctimer_save_chn_config(dev, channel, period_cycles,
+							pulse_cycles, flags);
```
zejiang0jason: is the indent right? Zephyr tab size is 8 characters, please configure your local env, thanks.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -207,7 +227,34 @@ static int mcux_sctimer_pwm_get_cycles_per_sec(const struct device *dev,
...
+	data->pwm_channel_config[channel].flags = flags;
+}
+
+static void mcux_sctimer_restore_chn_config(const struct device *dev)
+{
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+	uint8_t channel;
+
+	for (channel = 0; channel < CHANNEL_COUNT; channel++) {
+		mcux_sctimer_pwm_set_cycles(dev, channel,
```
zejiang0jason: this will configure all channels, even though some channels are not used.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -243,6 +290,37 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+		break;
+	case PM_DEVICE_ACTION_TURN_OFF:
```
zejiang0jason, line 793: is it necessary to stop the channels before turn off?

zejiang0jason (changes requested):

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: I have the tabs configured as 4 characters, but when I push the code, the tabs appears to be the double of characters I set, not sure if git is overriding this, but I haven't been able to adjust this setting. I will try to do it

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Added a check to the dutycycle configured per channel. It will only restore the previous configured channels. Thanks for the observation.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: While entering into PM3 all configurations get lost, I need to test it for the other power levels. In case the channels still active while entering into a low power mode I can add SCTIMER_Deinit in the turn off action. Thanks for the observation.

davidmissael:
I'm deinit the SCTimer and force pwm channels to output 0 when entering in a low power mode. In the deinit routine the clock source is disabled and I need to reinit the SCTime while exiting from a low power mode. Since RESUME action is called always previous the TURN ON, I'm recovering the timer and the pwm channels in the RESUME action to avoid doing it twice in the TURN ON action. For the RW612 TURN OFF action is not necessary cause it is called when entering in PM3 and there is not register retention. SUSPEND and RESUME actions will cover all Power modes.
@zejiang0jason could you please take a look on this new commit? Thanks so much.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -247,10 +247,12 @@ static void mcux_sctimer_restore_chn_config(const struct device *dev)
 	uint8_t channel;
 
 	for (channel = 0; channel < CHANNEL_COUNT; channel++) {
-		mcux_sctimer_pwm_set_cycles(dev, channel,
+		if (data->pwm_channel_config[channel].duty_cycles != 0) {
```
mmahadevan108: PLease add a comment explaining why this check was added.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -243,6 +292,44 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+		if (data->pwm_channel_active == true) {
+			mcux_sctimer_restore_chn_config(dev);
+		}
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+		if (data->pwm_channel_active == true) {
+			/* Halt the timer counters and disable the source clock */
+			SCTIMER_Deinit(config->base);
+			/* Force all outputs to inactive state */
+			config->base->OUTPUT = 0;
```
Holt-Sun: The SCTIMER_Deinit will disable clock, then the config->base->OUTPUT = 0 could cause a bus fault.
Should reorder the sequence.

drivers/pwm/pwm_mcux_sctimer.c
zejiang0jason: git won't change it, one possible problem is github may show the tag as 4 characters sometimes. You only need to make sure it is 8 characters in your local environment, and the result is right in your environment.

drivers/pwm/pwm_mcux_sctimer.c
zejiang0jason: check dutycycle 0 is not good, because the channel might be configured working with dutycycle 0. period might be better, because valid period won't be 0.

drivers/pwm/pwm_mcux_sctimer.c
zejiang0jason: SCTIMER power lost in PM3, so register content get lost, that makes sense. A small concern is, if power off the SCTIMER while it is working, will this result in some glitch? thanks.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Yeah, I changed my code editor (vscode) to 8-characters per tap. Thanks for the insight.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Since all peripheral configuration is lost while entering into PM3, the output pins go low by default. In my tests, always go low. However the user may use the power management entry callback to make sure there is not a glitch in the outputs previous to entering to a low power mode.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Thanks for the observation @Holt-Sun, writing to a register while the clock is disabled may cause a fault. 
Changing the sequence is not possible since the OUT register only can be written while the counters are halted, that's why I used the SCTimer_Denit() function because it halts the timers but I didn't realized the detail with the clock. What I did to solve this, I'm just halting the counters without disabling the clock instead of calling the SCTimer_Denit() routine. Counters will be reactivated in the mcux_sctimer_pwm_init_common() routine. Plase take a look.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Thanks @zejiang0jason, I changed the validation to check the period instead of the duty cycle.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Comments added to this check. Thanks for the observation.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -243,6 +295,44 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
 	return 0;
 }
 
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	const struct pwm_mcux_sctimer_config *config = dev->config;
+	const struct pwm_mcux_sctimer_data *data = dev->data;
```
Albort12138, line 767:
```suggestion
	struct pwm_mcux_sctimer_data *data = dev->data;
```

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -207,7 +227,39 @@ static int mcux_sctimer_pwm_get_cycles_per_sec(const struct device *dev,
...
+
+	data->pwm_channel_active = true;
+	data->pwm_channel_config[channel].period_cycles = period_cycles;
+	data->pwm_channel_config[channel].duty_cycles = pulse_cycles;
+	data->pwm_channel_config[channel].flags = flags;
+}
+
+static void mcux_sctimer_restore_chn_config(const struct device *dev)
+{
+	const struct pwm_mcux_sctimer_data *data = dev->data;
```
Albort12138, line 691:
```suggestion
	struct pwm_mcux_sctimer_data *data = dev->data;
```

zejiang0jason: why remove `const`? the content of `data` is not modified in this function.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -243,6 +295,44 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
 	return 0;
 }
 
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	const struct pwm_mcux_sctimer_config *config = dev->config;
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+
+	switch (action) {
```
zejiang0jason, line 771: Referring: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/serial/uart_mcux_flexcomm.c#L1209-L1234, a reasonable solution is place the register save/restore in TURN_ON/TURN_OFF.

davidmissael: @zejiang0jason TURN_ON/TURN_OFF and RESUME/SUSPEND are just called when entering into PM3. If system enters in PM1 or PM2 only SUSPEND and RESUME are called. We need to stops the PMW channels in any low power mode. The UART has a completely different use case, that's why only TURN_ON and TURN_OFF are used for that driver.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: @Albort12138 I will skip this comment then.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -23,6 +24,12 @@ LOG_MODULE_REGISTER(pwm_mcux_sctimer, CONFIG_PWM_LOG_LEVEL);
 /* Constant identifying that no event number has been set */
 #define EVENT_NOT_SET FSL_FEATURE_SCT_NUMBER_OF_EVENTS
 
+typedef struct pwm_channel_config {
+	uint32_t period_cycles;
+	uint32_t duty_cycles;
+	pwm_flags_t flags;
+} pwm_channel_config_t;
```
DerekSnell, line 37: This struct consumes 10 Bytes of RAM per channel, and is only used if `CONFIG_PM_DEVICE=y`.  We can wrap this with `#ifdef CONFIG_PM_DEVICE` to save RAM for apps not using PM.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -207,7 +227,39 @@ static int mcux_sctimer_pwm_get_cycles_per_sec(const struct device *dev,
 	return 0;
 }
 
-static int mcux_sctimer_pwm_init(const struct device *dev)
+static void mcux_sctimer_save_chn_config(const struct device *dev,
```
DerekSnell: Not major, and I accept this.  But personally I do not see the point of adding `mcux_sctimer_save_chn_config()` as a function.  It is only called once.  It seems simpler to add the 4 lines of code below in `mcux_sctimer_pwm_set_cycles()` instead of defining and calling this function.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -34,10 +41,20 @@ struct pwm_mcux_sctimer_config {
 struct pwm_mcux_sctimer_data {
 	uint32_t event_number[CHANNEL_COUNT];
 	sctimer_pwm_signal_param_t channel[CHANNEL_COUNT];
+	pwm_channel_config_t pwm_channel_config[CHANNEL_COUNT];
 	uint32_t match_period;
 	uint32_t configured_chan;
+	bool pwm_channel_active;
```
DerekSnell, line 87: Is this needed?  `configured_chan` will be greater than 0 if any channel was configured.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -243,6 +295,44 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+		if (data->pwm_channel_active == true) {
+			mcux_sctimer_restore_chn_config(dev);
+		}
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+		if (data->pwm_channel_active == true) {
+			/* Halt the timer counters */
+			config->base->CTRL |= (SCT_CTRL_HALT_L_MASK | SCT_CTRL_HALT_H_MASK);
+			/* Force all outputs to inactive state */
+			config->base->OUTPUT = 0;
```
DerekSnell: I'm blocking for this feedback.  In the OUTPUT register, "Writing a 0 forces the corresponding output low" for all channels.  This driver maintains the `level` for active/inactive state per channel, see [driver code](https://github.com/zephyrproject-rtos/zephyr/blob/b93fabaf25ef34fdb2f8d40d28e7337c40d14651/drivers/pwm/pwm_mcux_sctimer.c#L189).  For SUSPEND, we should drive each channel to the inactive level instead of forcing all channel outputs low.

DerekSnell (changes requested):
Hi @davidmissael ,
Thank you for contributing this.  I have some feedback, and request changing driving all outputs low in SUSPEND.

Also, please squash these commits into a single commit.  This is all adding the PM feature to this driver, and some of the commit messages no longer apply.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Hi @DerekSnell, yes we need a flag to know if there were some channels configured previous entering into PM3 to recover them after exiting PM3. We cannot use `configured_chan` because **the value is cleared** within the function `mcux_sctimer_pwm_init_common` and this function is called in PM_DEVICE_ACTION_TURN_ON. So when exiting from PM3 `configured_chan` is always 0 and is not useful to know if there were channels previously configured.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: That's a good point, we can wrap this structure with `#ifdef CONFIG_PM_DEVICE` however if we do this, we need to wrap the code within PM_DEVICE_ACTION_TURN_ON that calls `mcux_sctimer_restore_chn_config` because this function needs the structure `pwm_channel_config_t`. Does this make sense?

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Yes, we can replace the function call for the 4 code lines. Any solution works for me

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Thanks for the insight Derek. I already implemented a check for the output level to identify which is the inactive value per channel as you requested. Please take a look

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: I wrapped al the code related to PM within the `#ifdef CONFIG_PM_DEVICE` please take a look

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Hi @zejiang0jason based on our last conversation, SUSPEND/RESUME hooks only halt/run the timer, TURN_ON/TURN_OFF perform the recovery state. Please take a look on the last commit. Thanks

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: I will close this conversation since does not apply anymore.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: I will mark this conversation since as resolved since it was clarified in the conversation below.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Used SCTIMER_StopTimer to halt the counter and SCTIMER_StartTimer to reinit the counter. This won't cause the conflict with the clocks using the Deinit routine. Plase take a look.

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+{
+	#ifdef CONFIG_PM_DEVICE
+	const struct pwm_mcux_sctimer_config *config = dev->config;
+	const struct pwm_mcux_sctimer_data *data = dev->data;
+	uint8_t channel;
+	#endif /* CONFIG_PM_DEVICE */
+
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+		#ifdef CONFIG_PM_DEVICE
```
DerekSnell: This `CONFIG_PM_DEVICE` wrapper is not needed here.  The PM subsystem only delivers the `PM_DEVICE_ACTION_RESUME` when `CONFIG_PM_DEVICE=y`

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+		SCTIMER_StartTimer(config->base, kSCTIMER_Counter_U);
+		#endif /* CONFIG_PM_DEVICE */
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+		#ifdef CONFIG_PM_DEVICE
```
DerekSnell: This `CONFIG_PM_DEVICE` wrapper is not needed here.

DerekSnell (dismissed):
Hi @davidmissael ,
Thank you for your work on this.  I added some minor improvements, but I approve

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
 	return 0;
 }
 
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
+	#ifdef CONFIG_PM_DEVICE
```
mmahadevan108: Not need to indent `#ifdef` lines. Same comment for the matching #endif

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
 	return 0;
 }
 
+static int mcux_sctimer_pwm_pm_action(const struct device *dev, enum pm_device_action action)
+{
```
mmahadevan108, line 764: Can we leave SUSPEND and RESUME actions empty. Do we need to stop and start?

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Hi @DerekSnell, If we don't wrap the SUSPEND within the CONFIG_PM_DEVICE the build will fail due it is doing reference to the element "data->pwm_channel_config[channel].period_cycles" which it is wrapped around this definition in the structure "pwm_mcux_sctimer_data" as per your suggestion.

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Hi @mmahadevan108 since PM1 and PM2 does not need to recover PWM channel configurations I'm treating them independently to PM3 which needs to recover channels configuration after exiting. If we only use TURN_ON and TURN_OFF this will be the effect:
- When system uses PM1/PM2 the TURN_ON hook will call "mcux_sctimer_restore_chn_config" to restore channels configuration **which is not needed for this PM levels** and will add some latency in the exiting sequence. Also will reinitialize the SCTimer again calling the "mcux_sctimer_pwm_init_common" which will add even more latency.
- When system uses PM3 the TURN_OFF hook will stop the timer and for this PM level this is not needed.

Answering the question if we need to start/stop the sctimer: Yes we need to do it for PM1 and PM2 because halting the timer counter is the only way you can write to the output register to set the PWM channels into inactive state. I you don't do this process the PWM channels (output pins) will keep the last state (active/inactive randomly). Halting the timer is necessary to set the channel outputs to inactive state before entering into PM1 or PM2.

mmahadevan108: When we have active channels, we should probably block entering low power modes.

DerekSnell: > When we have active channels, we should probably block entering low power modes.

Sorry @mmahadevan108 , I disagree.  This topic goes back to the discussion in the original SCTimer PR https://github.com/zephyrproject-rtos/zephyr/pull/89384 .  My understanding is that the app should control the PWM functionality with PM, not this driver.  If the app wants to keep a PWM signal active and prevent SOC power modes from interfering with that PWM, then the app should call `pm_policy_device_power_lock_get()`.  The PM subsystem will then lock out any SOC power modes listed in the `zephyr,disabling-power-states` property for that device, or listed in the power domain for that device.  If the app does not call `pm_policy_device_power_lock_get()`, then the assumption is that the app is allowing these SOC low-power modes during Idle.  And then this driver uses the SUSPEND action to halt this timer and force the PWM outputs to the inactive logic level.

Do you feel differently?

mmahadevan108 (dismissed):

drivers/pwm/pwm_mcux_sctimer.c
davidmissael: Indentation removed

mmahadevan108 (approved):

drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -713,6 +760,60 @@ static int mcux_sctimer_pwm_init(const struct device *dev)
...
+	switch (action) {
+	case PM_DEVICE_ACTION_RESUME:
+#ifdef CONFIG_PM_DEVICE
+		SCTIMER_StartTimer(config->base, kSCTIMER_Counter_U);
+#endif /* CONFIG_PM_DEVICE */
+		break;
+	case PM_DEVICE_ACTION_SUSPEND:
+#ifdef CONFIG_PM_DEVICE
+		/* Halt the timer counters */
+		SCTIMER_StopTimer(config->base, kSCTIMER_Counter_U);
```
zejiang0jason, line 780: Hi, I have a small question here.

If the output pin is in active state when we stop the timer, will the pin be kept in active state? If so, I suggest place output in active state first, then stop the timer. The concern is, if the timer is stopped, and current task is switched out, then output pin will keep active for a longer time than we expect, this is not good for some critical use cases.

cc @FelixWang47831 @william-tang914  who are familiar with the timer module, about the SCTIMER behavior. Thanks.

zejiang0jason (approved):

drivers/pwm/pwm_mcux_sctimer.c
DerekSnell: Hi @zejiang0jason ,
Yes, if the SCTimer is halted, the timer will continue driving the PWM output signal at the same logic level.  So if we halt while the output is active in the SUSPEND action, the output will remain active until the RESUME action.  Which we do not want.  David's PR forces the output to the inactive logic level by writing to the OUTPUT register.  But SW can only write to the OUTPUT register AFTER the timer is halted.  

The OUTPUT register description says:
> Each SCT output has a corresponding bit in the OUTPUT register to allow software to control the output state directly or read its current state.
While the counter is running, outputs are set, cleared, or toggled only by events. However, using the OUTPUT register, software can write to any of the output registers when both counters are halted to control the outputs directly. Writing to the OUT register is only allowed when all counters (L-counter, H-counter, or unified counter) are halted (CTRL[HALT] = 1).
>
> Software can read this register at any time to sense the state of the outputs.

DerekSnell (approved):

PR #100974, opened 2025-12-12
LoRa: LBM: Fix LBM driver init and add GPIO callback API

carlocaione:
A bit of context here.

I'm working on making LBM a first-class backend citizen for LoRaWAN on Zephyr, so these couple of commits are part of a preparatory work to that.

The LBM HAL needs a bit more control over the stack than what Zephyr is currently able to guarantee or give. The `sx12xx` radio drivers in particular are doing too much during init, relinquishing the stack and configuring it at init stage. We need a way to let the HAL doing its job, configuring the radio as it please before being stepped over by Zephyr, so I'm moving the radio initialization into the `.config` hook instead of doing it at init time. The other viable option was to modify the API adding an `.init` hook but that would require an API change and I prefer to avoid that if possible.

The second commit is about adding a way for the HAL to install its own `DIO1` callback, simple as that.

### DIFF
--- a/drivers/lora/lora_basics_modem/Kconfig
+++ b/drivers/lora/lora_basics_modem/Kconfig
@@ -34,4 +34,12 @@ config LORA_BASICS_MODEM_RSSI_REPORT_TYPE_SIGNAL
 
 endchoice
 
+config LORA_BASICS_MODEM_DEFERRED_INIT
+	bool "Defer radio initialization until first use"
+	help
+	  When enabled, radio hardware initialization (reset, interrupt
+	  configuration) is deferred until the first call to lora_config().
+	  This allows applications to perform setup before the radio is
+	  initialized. When disabled, the radio is fully initialized at boot.
+
 endif
--- a/drivers/lora/lora_basics_modem/lbm_common.h
+++ b/drivers/lora/lora_basics_modem/lbm_common.h
@@ -105,3 +105,40 @@ static inline int lbm_optional_gpio_set_dt(const struct gpio_dt_spec *spec, int 
 
 /* Common LBM implementation of the LoRa API */
 extern const struct lora_driver_api lbm_lora_api;
+
+/**
+ * @brief Add a GPIO callback for DIO1 interrupts
+ *
+ * This function registers a user callback that will be invoked from interrupt
+ * context when the DIO1 interrupt fires. The user must provide a pre-allocated
+ * gpio_callback structure. The driver will initialize the callback with the
+ * provided handler and the correct pin mask for the DIO1 pin.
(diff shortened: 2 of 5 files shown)


JordanYates (commented):
I'm not a huge fan of moving all radio initialization outside of `dev_init`, since it essentially makes `device_is_ready` useless for the device.

Since LoRa and LoRaWAN don't need to co-exist, could we instead add a Kconfig that is selected by LoRaWAN which defers the initialisation, while the LoRa behaviour remains the same?
```
static int sx126x_init(const struct device *dev) {
#ifndef CONFIG_LBM_DEFERRED_INIT
    ret = lbm_driver_radio_init();
#endif
}
```

RE the GPIO callback, wouldn't it be simpler to store the callback in the data structure and call it in our own handler if it exists?

drivers/lora/lora_basics_modem/Kconfig
```diff
@@ -34,4 +34,13 @@ config LORA_BASICS_MODEM_RSSI_REPORT_TYPE_SIGNAL
 
 endchoice
 
+config LORA_BASICS_MODEM_DEFERRED_INIT
+	bool "Defer radio initialization until first use"
+	default n
```
JordanYates: `default n` is not required, it just makes it not possible to change the default in other locations

drivers/lora/lora_basics_modem/lbm_common.h
```diff
@@ -105,3 +120,40 @@ static inline int lbm_optional_gpio_set_dt(const struct gpio_dt_spec *spec, int
...
 extern const struct lora_driver_api lbm_lora_api;
+
+/**
+ * @brief Add a GPIO callback for DIO1 interrupts
+ *
+ * This function registers a user callback that will be invoked from interrupt
+ * context when the DIO1 interrupt fires. The user must provide a pre-allocated
+ * and initialized (except for pin mask) gpio_callback structure.
+ *
+ * Note: The driver will configure the pin mask of the callback structure to
```
JordanYates: Rather than making this function do something odd (half initializing the struct), I think it would be cleaner to have `lbm_driver_add_dio1_gpio_callback` handle all of the initialisation of `struct gpio_callback`, which would only require passing the callback pointer as a third argument.

JordanYates (changes requested):

drivers/lora/lora_basics_modem/Kconfig
carlocaione: ack

JordanYates (approved):

valeriosetti (approved):

PR #100977, opened 2025-12-13
drivers: lora: LDRO and bandwith additions

VynDragon:
Add missing bandwidths, set enum to be value in khz, add missing sf
use enum value in khz to calculate approximate symbol time
use symbol time to determine if ldro should be enabled
add option to always enable ldro regardless of symbol time

symbol time checked against lora calculator https://www.semtech.com/design-support/lora-calculator, formula is good enough to match without using floats.

See any modem datasheet for 16ms/16.38ms times and the symbol rate / symbol time formulas

### DIFF
--- a/include/zephyr/drivers/lora.h
+++ b/include/zephyr/drivers/lora.h
@@ -38,9 +38,21 @@ extern "C" {
  * higher data rates but typically reduce sensitivity and range.
  */
 enum lora_signal_bandwidth {
-	BW_125_KHZ = 0,	/**< 125 kHz */
-	BW_250_KHZ,	/**< 250 kHz */
-	BW_500_KHZ,	/**< 500 kHz */
+	BW_7_KHZ = 7,		/**< 7.81 kHz */
+	BW_10_KHZ = 10,		/**< 10.42 kHz */
+	BW_15_KHZ = 15,		/**< 15.63 kHz */
+	BW_20_KHZ = 20,		/**< 20.83 kHz */
+	BW_31_KHZ = 31,		/**< 31.25 kHz */
+	BW_41_KHZ = 41,		/**< 41.67 kHz */
+	BW_62_KHZ = 62,		/**< 62.5 kHz */
+	BW_125_KHZ = 125,	/**< 125 kHz */
+	BW_200_KHZ = 200,	/**< 203 kHz */
+	BW_250_KHZ = 250,	/**< 250 kHz */
+	BW_400_KHZ = 400,	/**< 406 kHz */
+	BW_500_KHZ = 500,	/**< 500 kHz */
+	BW_800_KHZ = 800,	/**< 812 kHz */
+	BW_1000_KHZ = 1000,	/**< 1000 kHz */
+	BW_1600_KHZ = 1600,	/**< 1625 kHz */
 };
 
 /**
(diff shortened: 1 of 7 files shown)


include/zephyr/drivers/lora.h
```diff
@@ -38,9 +38,21 @@ extern "C" {
...
+	BW_41_KHZ = 41,		/**< 41.67 kHz */
+	BW_62_KHZ = 62,		/**< 62.5 kHz */
+	BW_125_KHZ = 125,	/**< 125 kHz */
+	BW_200_KHZ = 200,	/**< 203 kHz */
+	BW_250_KHZ = 250,	/**< 250 kHz */
+	BW_400_KHZ = 400,	/**< 406 kHz */
+	BW_500_KHZ = 500,	/**< 500 kHz */
+	BW_800_KHZ = 800,	/**< 812 kHz */
+	BW_1000_KHZ = 1000,	/**< 1000 kHz */
+	BW_1600_KHZ = 1600,	/**< 1625 kHz */
```
JordanYates, line 55: With the extra bandwidth options, whats the behaviour when an invalid bandwidth for the modem is selected? Does LBM reject it?

VynDragon: I think it does (I think i tried it once? but LBM doesnt know about LLCC68 for example sooo), but it doesn't matter, I was working on this to solve those kind of possible issues: https://github.com/zephyrproject-rtos/zephyr/pull/100989

include/zephyr/drivers/lora.h
```diff
@@ -38,9 +38,21 @@ extern "C" {
  * higher data rates but typically reduce sensitivity and range.
  */
 enum lora_signal_bandwidth {
-	BW_125_KHZ = 0,	/**< 125 kHz */
-	BW_250_KHZ,	/**< 250 kHz */
-	BW_500_KHZ,	/**< 500 kHz */
+	BW_7_KHZ = 7,		/**< 7.81 kHz */
```
JordanYates, line 41: I don't believe there is any advantage for the enum values being the approximate frequency, it just makes the configuration struct larger (and probably makes the generated switch statements more complicated).

drivers/lora/lora_basics_modem/lbm_common.c
```diff
@@ -73,7 +81,8 @@ int lbm_lora_config(const struct device *dev, struct lora_modem_config *lora_con
 		.mod_params = {
 			.sf = lora_config->datarate,
 			.cr = lora_config->coding_rate,
-			.ldro = 0,
+			.ldro = LORA_LDRO(lora_config->datarate, lora_config->bandwidth)
```
JordanYates:
```suggestion
			.ldro = config->force_ldro ? 1 : LORA_LDRO(lora_config->datarate, lora_config->bandwidth),
```

JordanYates (changes requested):

include/zephyr/drivers/lora.h
VynDragon: Did you miss the LDRO calculation code being the direct purpose of this

VynDragon: https://github.com/zephyrproject-rtos/zephyr/pull/100977/files#diff-88cb4e955362b4a874328d053aa5c72ad84def06338fe7cd80718008fb4c0730R24
https://github.com/zephyrproject-rtos/zephyr/pull/100977/files#diff-88cb4e955362b4a874328d053aa5c72ad84def06338fe7cd80718008fb4c0730R84

JordanYates: Why does that rely on the enums being any value in particular? As far as I can tell they would work for any values, as long as they are ascending

VynDragon: Look closer, the calculation is explained and relies on the specific bandwith in KHz vs the spreading factor, you will not get correct results when using any sequential value.

JordanYates: Ok, `(((1 << sf) / bw >= 16 ? 1 : 0)` is easily misread as `bw >= 16`, I'm sure the evaluation order works out, but it should have an extra set of brackets in there for clarity.

VynDragon: Okay will add that when i get to updating my PRs

JordanYates, line 41: Is there a reason you used 7KHz as the define name for 7.81kHz instead of 8kHz? It is much closer to the later (although all the approximations in `LORA_LDRO` means the result would then be incorrect)

VynDragon: Yes, I have followed what seems to be semtech's preferred choice in LBM which made me wonder the same thing as it doesnt match the lora calculator (though I vaguely remember it being 7Khz there too? Maybe they changed it): https://github.com/Lora-net/SWL2001/blob/a8ddc544043e72807cf7db532478e1dda734ae7c/lbm_lib/smtc_modem_core/radio_drivers/lr20xx_driver/inc/lr20xx_radio_lora_types.h#L95

drivers/lora/lora_basics_modem/lbm_common.c
VynDragon: done

include/zephyr/drivers/lora.h
VynDragon: done

JordanYates (dismissed):

mniestroj (dismissed):

mniestroj (approved):

JordanYates (approved):

PR #94734, opened 2025-08-20
samples: drivers: crypto: Add Support for TI MSPM0 AES module

Girinandha-M:
TI MSPM0 has AES accelerator for encryption and decryption

- Add support for crypto AES ECB and CBC mode of operation for encryption and 
decryption for TI MSPM0 AES module.

Signed-off-by: Girinandha Manivelpandiyan girinandha@linumiz.com

PR #100981, opened 2025-12-13
include: crypto: add AES CFB and OFB modes

Girinandha-M:
This PR extends the Zephyr crypto cipher API to support AES Cipher Feedback
(CFB) and Output Feedback (OFB) modes of operation.

Changes:
- Add CFB and OFB cipher mode identifiers to `include/zephyr/crypto/cipher.h`.
- Update `include/zephyr/crypto/crypto.h` comments to describe the new modes.

### DIFF
--- a/include/zephyr/crypto/cipher.h
+++ b/include/zephyr/crypto/cipher.h
@@ -47,6 +47,8 @@ enum cipher_mode {
 	CRYPTO_CIPHER_MODE_CTR = 3, /**< Counter mode. */
 	CRYPTO_CIPHER_MODE_CCM = 4, /**< Counter with CBC-MAC mode. */
 	CRYPTO_CIPHER_MODE_GCM = 5, /**< Galois/Counter mode. */
+	CRYPTO_CIPHER_MODE_CFB = 6, /**< Cipher Feedback mode. */
+	CRYPTO_CIPHER_MODE_OFB = 7, /**< Output Feedback mode. */
 };
 
 /* Forward declarations */
@@ -124,6 +126,32 @@ typedef int (*gcm_op_t)(struct cipher_ctx *ctx, struct cipher_aead_pkt *pkt,
 			 uint8_t *nonce);
 
 /**
+ * @brief Perform a CFB cipher operation.
+ *
+ * @param ctx Cipher session context.
+ * @param pkt Packet containing input and output buffers.
+ * @param iv Initialization vector for this operation.
+ *
+ * @retval 0 Operation completed successfully.
+ * @retval -errno Negative errno code on failure.
+ */
+typedef int (*cfb_op_t)(struct cipher_ctx *ctx, struct cipher_pkt *pkt,
+			uint8_t *iv);
+
+/**
+ * @brief Perform an OFB cipher operation.
+ *
+ * @param ctx Cipher session context.
+ * @param pkt Packet containing input and output buffers.
+ * @param iv Initialization vector for this operation.
+ *
+ * @retval 0 Operation completed successfully.
+ * @retval -errno Negative errno code on failure.
+ */
+typedef int (*ofb_op_t)(struct cipher_ctx *ctx, struct cipher_pkt *pkt,
+			uint8_t *iv);
+
+/**
  * Cipher operation handlers selected for a session.
  *
  * The crypto driver populates this structure during cipher_begin_session()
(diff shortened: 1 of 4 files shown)


parthitce (changes requested):
Drop the changes related to MSPM0 and keep the API changes separated. AES driver is already tracked without CFB and OFB https://github.com/zephyrproject-rtos/zephyr/pull/94734. Support can be added at later point after the API changes are accepted.

As 4.4 is already released, this can be targeted for 4.5, update the doc.

include/zephyr/crypto/cipher.h
```diff
@@ -47,6 +47,8 @@ enum cipher_mode {
 	CRYPTO_CIPHER_MODE_CTR = 3,
 	CRYPTO_CIPHER_MODE_CCM = 4,
 	CRYPTO_CIPHER_MODE_GCM = 5,
+	CRYPTO_CIPHER_MODE_CFB = 6, /**< Cipher Feedback mode */
```
parthitce: stay with the existing struct. drop the comment.

include/zephyr/crypto/cipher.h
```diff
@@ -81,6 +89,8 @@ struct cipher_ops {
 		ctr_op_t	ctr_crypt_hndlr;
 		ccm_op_t	ccm_crypt_hndlr;
 		gcm_op_t	gcm_crypt_hndlr;
+		cfb_op_t	cfb_crypt_hndlr; /**< CFB mode crypt handler */
```
parthitce: ditto, drop the comment

parthitce (changes requested):
Minor nits, otherwise good. @ceolin does this needs to be included in Arch review?

include/zephyr/crypto/cipher.h
kartben: Public API must be fully documented using Doxygen :) so maybe a first commit can add all the missing, and hopefully trivial, comments to all the other enum fields.

parthitce:
https://github.com/zephyrproject-rtos/zephyr/pull/107901 already covers the document style. This can go in after that. @Girinandha-M use the doxygen format as in the mentioned PR.

Girinandha-M:
@valeriosetti @ceolin ping for review. Thanks.

valeriosetti (dismissed):

valeriosetti (dismissed):

ssekar15 (dismissed):

Girinandha-M:
@ceolin ping for review, Thanks.

valeriosetti (dismissed):

parthitce (dismissed):
LGTM

doc/releases/migration-guide-4.5.rst
```diff
@@ -133,6 +133,13 @@ Controller Area Network (CAN)
   are processed in the order received on the bus. Out-of-tree users may want to update any
   ``bosch,mram-cfg`` devicetree property overrides to allocate all FIFO elements to RX FIFO0.
 
+Crypto
+======
+
+* Added :c:enum:`CRYPTO_CIPHER_MODE_CFB` and :c:enum:`CRYPTO_CIPHER_MODE_OFB` cipher mode
+  identifiers to the crypto API (``include/zephyr/crypto/cipher.h``). This change is additive
+  and does not affect existing applications.
```
kartben: This doesn't belong to the migration guide. No need to mention something if it doesn't require any action from the user :)

kartben (changes requested):

valeriosetti (approved):

ssekar15 (approved):

kartben (approved):

Girinandha-M:
@ceolin ping for review, Thanks.

ceolin (approved):
Thank you. LGTM

RichardSWheatley:
@kartben

So if we decide to put crypto in at this current moment, we can do it this way? Because I was under the impression that we had to forcibly go forward with PSA only. I have multiple algorithms that I would love to put in here, but I do not have the time to convert to PSA yet.

kartben:
> @kartben
> 
> So if we decide to put crypto in at this current moment, we can do it this way? Because I was under the impression that we had to forcibly go forward with PSA only. I have multiple algorithms that I would love to put in here, but I do not have the time to convert to PSA yet.

I don't know the answer to this question, sorry. But since I've been asked the exact same one just two days ago I agree this needs clarification :) I've posted in `#security` on Discord to bring this to the attention of a larger audience.

PR #100987, opened 2025-12-13
drivers: stepper: use only driver names in Kconfigs

jilaypandya:
Drop prefixes such as STEPPER and vendor names from the Kconfigs. Use only the name of the IC for which the driver is to be activated

### DIFF
--- a/drivers/stepper/adi_tmc/CMakeLists.txt
+++ b/drivers/stepper/adi_tmc/CMakeLists.txt
@@ -4,9 +4,9 @@
 zephyr_library()
 zephyr_library_property(ALLOW_EMPTY TRUE)
 
-add_subdirectory_ifdef(CONFIG_STEPPER_ADI_TMC2209 tmc22xx)
-add_subdirectory_ifdef(CONFIG_STEPPER_ADI_TMC50XX tmc50xx)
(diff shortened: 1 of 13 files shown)


drivers/stepper/adi_tmc/CMakeLists.txt
```diff
@@ -4,9 +4,9 @@
 zephyr_library()
 zephyr_library_property(ALLOW_EMPTY TRUE)
 
-add_subdirectory_ifdef(CONFIG_STEPPER_ADI_TMC2209 tmc22xx)
-add_subdirectory_ifdef(CONFIG_STEPPER_ADI_TMC50XX tmc50xx)
-add_subdirectory_ifdef(CONFIG_STEPPER_ADI_TMC51XX tmc51xx)
```
faxe1008, line 9: Hmm i searched some other driver classes. Its really not consistent, but most of them simply have the IC name. So I guess this feels right.

faxe1008 (approved):

drivers/stepper/adi_tmc/CMakeLists.txt
MaureenHelm: That's true for sensors, but not sure it's a good example. For mfd devices we need the driver class name, so I lean toward keeping it.

PR #100992, opened 2025-12-14
modules: nrf_wifi: spi: optimise SPI RX transaction

JordanYates:
Optimise the transactions queued by `spim_xfer_rx` to only require 2 sequences started by the SPI driver, instead of 3.

Previously:
  1. TX `sizeof(hdr)`, RX nothing
  2. TX nothing, finish receiving `sizeof(hdr) + discard_bytes`
  3. TX nothing, receive `len`

Now:
 1. TX `sizeof(hdr) + discard_bytes`, RX nothing
 2. TX nothing, receive `len`

### DIFF
--- a/modules/nrf_wifi/bus/spi_if.c
+++ b/modules/nrf_wifi/bus/spi_if.c
@@ -51,33 +51,35 @@ static int spim_xfer_tx(unsigned int addr, void *data, unsigned int len)
 	return err;
 }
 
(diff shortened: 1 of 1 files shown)


krish2718 (approved):
Thanks @JordanYates  sorry for the delay.

sachinthegreen (approved):

PR #100994, opened 2025-12-14
drivers: wifi: airoc: Populate band in scan results

tchebb:
The driver gives us this information, so we just need to translate it to Zephyr's type like we already do for security.

### DIFF
--- a/drivers/wifi/infineon/airoc_wifi.c
+++ b/drivers/wifi/infineon/airoc_wifi.c
@@ -231,12 +231,33 @@ static whd_security_t convert_zephyr_security_to_whd(int security)
 	return whd_security;
 }
 
+static uint8_t convert_whd_band_to_zephyr(whd_802_11_band_t band)
+{
+	uint8_t zephyr_band = WIFI_FREQ_BAND_UNKNOWN;
+
+	switch (band) {
+	case WHD_802_11_BAND_2_4GHZ:
+		zephyr_band = WIFI_FREQ_BAND_2_4_GHZ;
+		break;
+
+	case WHD_802_11_BAND_5GHZ:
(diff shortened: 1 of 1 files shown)


drivers/wifi/infineon/airoc_wifi.c
```diff
@@ -231,12 +231,33 @@ static whd_security_t convert_zephyr_security_to_whd(int security)
 	return whd_security;
 }
 
+static uint8_t convert_whd_band_to_zephyr(whd_802_11_band_t band)
+{
+	uint8_t zephyr_band = WIFI_FREQ_BAND_UNKNOWN;
+
+	switch (band) {
```
teburd, line 238: Could this be a lookup table instead?

```c
uint8_t zephyr_band = WIFI_FREQ_BAND_UNKNOWN;

if (/* band in some range... */) {
   zephyr_band = WHD_TO_ZEPHYR[band];
}

return zephyr_band;
```

tchebb: Definitely can be. I was following the style of other similar functions in the file, like `convert_whd_security_to_zephyr()`, but I'll change it to a lookup table if you prefer.

jukkar: As the values in `enum wifi_frequency_bands` are not fully consecutive (the unknown value is "outside" of max defined value), I would not bother using an array here. It can be done of course, but feels error prone, especially if the enum values change.

teburd: Fair!

teburd (approved):

jukkar (approved):

PR #100995, opened 2025-12-14
RA8x2 ethernet driver support

duynguyenxa:
This PR introduce the support of Ethernet driver on RA8x2 devices base on eswm, rmac and rmac phy HWIP
This PR also add a new phy driver support for GPY111 chip


drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,733 @@
...
+extern fsp_err_t rmac_do_link(rmac_instance_ctrl_t *const p_instance_ctrl,
+			      const layer3_switch_magic_packet_detection_t mode);
+
+static void phy_link_cb(const struct device *phy_dev, struct phy_link_state *state, void *eth_dev)
+{
+	const struct device *dev = (struct device *)eth_dev;
+	struct eth_renesas_ra_data *data = dev->data;
+	fsp_err_t fsp_err = FSP_SUCCESS;
+
+	if (state->is_up) {
```
jukkar: I suggest following changes to avoid this "if() else" construct and extra indentation

```
if (!state->is_up) {
	if (data->phy_link_up == 1) {
		/* phy state change from up to down */
		r_rmac_disable_reception(data->fsp_ctrl);
		data->fsp_ctrl->link_establish_status = ETHER_LINK_ESTABLISH_STATUS_DOWN;
		data->phy_link_up = 0;
		LOG_DBG("link down");
		net_eth_carrier_off(data->iface);
	}

	return;
}

data->fsp_ctrl->link_establish_status = ETHER_LINK_ESTABLISH_STATUS_UP;
data->phy_link_up = 1;
...
```

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,733 @@
...
+	uint8_t *tx_frame;
+#endif
+	struct k_thread rx_thread;
+
+	K_KERNEL_STACK_MEMBER(rx_thread_stack, CONFIG_ETH_RENESAS_RA_RX_THREAD_STACK_SIZE);
+	struct k_sem rx_sem;
+#if defined(CONFIG_ETH_RENESAS_RA_USE_HW_WRITEBACK)
+	struct k_sem tx_sem;
+#endif
+	uint8_t phy_link_up;
```
jukkar: This looks to be used as a boolean in the code, so why not have `bool` type?

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,733 @@
...
+	} else if (fsp_err != FSP_SUCCESS) {
+		LOG_DBG("failed to read from FIFO");
+		ret = -EIO;
+		goto rx_end;
+	}
+
+	/* read again for remaining data */
+	k_sem_give(&data->rx_sem);
+
+	pkt = net_pkt_rx_alloc_with_buffer(data->iface, (size_t)len, AF_UNSPEC, 0, K_MSEC(100));
```
jukkar: Please use `NET_AF_UNSPEC` here

jukkar (commented):
Is the driver compile tested by the CI? If not, then I suggest you add the tests so that the code does not bit rot easily.

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,745 @@
...
+			/* phy state change from up to down */
+			r_rmac_disable_reception(data->fsp_ctrl);
+
+			data->fsp_ctrl->link_establish_status = ETHER_LINK_ESTABLISH_STATUS_DOWN;
+			data->phy_link_up = 0;
+
+			LOG_DBG("link down");
+			net_eth_carrier_off(data->iface);
+		}
+	}
```
maass-hamburg: it seems like ``phy_link_up`` is just used to ensure down doesn't happen when it is already down. Is that really needed? Would something break if it happens, if not, it can be removed

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,745 @@
...
+#endif
+
+	ret = net_recv_data(net_pkt_iface(pkt), pkt);
+	if (ret < 0) {
+		LOG_DBG("failed to push pkt to network stack");
+		net_pkt_unref(pkt);
+	}
+
+rx_end:
+	if (ret) {
```
maass-hamburg:
```suggestion
	if (ret < 0) {
```
no implicit conversion to bool

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,745 @@
...
+	}
+
+	ret = pinctrl_apply_state(config->pin_cfg, PINCTRL_STATE_DEFAULT);
+	if (ret < 0) {
+		return ret;
+	}
+
+	if (config->random_mac || !config->valid_mac) {
+		gen_random_mac(&data->mac_addr[0], RENESAS_OUI_B0, RENESAS_OUI_B1, RENESAS_OUI_B2);
+	}
```
maass-hamburg: use the new mac address configuration struct, see https://github.com/zephyrproject-rtos/zephyr/pull/100620

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,631 @@
...
+};
+
+struct maxlinear_gpy111_dev_data {
+	const struct device *dev;
+	phy_callback_t cb;
+	void *cb_data;
+	struct phy_link_state state;
+	struct k_sem sem;
+	struct k_work_delayable monitor_work;
+	bool gigabit_supported;
```
maass-hamburg: no need for checking `gigabit_supported` and the logic for it. As this is a driver for a specific phy, we already know, that it supports gigabit.

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,631 @@
...
+
+		if ((estat_reg & (MII_ESTAT_1000BASE_T_HALF | MII_ESTAT_1000BASE_T_FULL)) != 0U) {
+			*supported = true;
+			return 0;
+		}
+	}
+
+	*supported = false;
+	return 0;
+}
```
maass-hamburg: not needed, we already know it supports it

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,631 @@
...
+	.cfg_link = phy_gpy111_config_link,
+	.read = maxlinear_gpy111_read,
+	.write = maxlinear_gpy111_write,
+};
+
+#if ANY_RESET_GPIO
+#define RESET_GPIO(n)                                                                              \
+	.reset_gpio = GPIO_DT_SPEC_INST_GET_OR(n, reset_gpios, {0}),                               \
+	.reset_assert_duration_us = DT_INST_PROP_OR(n, reset_assert_duration_us, 0),               \
+	.reset_deassertion_timeout_ms = DT_INST_PROP_OR(n, reset_deassertion_timeout_ms, 0),
```
maass-hamburg: these were dt props in the phy_mii driver, because these are different on different models, but for this driver you can look them up and make them defines at the top of the file and directly use them, no need then to save them in the config struct.

dts/bindings/ethernet/phy/maxlinear,gpy111.yaml
```diff
@@ -0,0 +1,74 @@
...
+    type: string
+    description: This link is fixed and does not require PHY configuration
+    enum:
+      - "10BASE-T Half-Duplex"
+      - "10BASE-T Full-Duplex"
+      - "100BASE-T Half-Duplex"
+      - "100BASE-T Full-Duplex"
+      - "1000BASE-T Half-Duplex"
+      - "1000BASE-T Full-Duplex"
+
```
maass-hamburg: no need for this

dts/bindings/mdio/renesas,ra-mdio-rmac.yaml
```diff
@@ -0,0 +1,29 @@
...
+    required: true
+
+  phy-connection-type:
+    type: string
+    description: |
+      Specifies the interface connection type between ethernet MAC and PHY.
+    enum:
+      - "mii"
+      - "rmii"
+      - "gmii"
```
maass-hamburg: this should not be needed for a mdio controller

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
```diff
@@ -0,0 +1,195 @@
...
+	ret = (fsp_err == FSP_SUCCESS) ? 0 : -EIO;
+
+	*regval = val & UINT16_MAX;
+
+	k_mutex_unlock(&data->rw_lock);
+
+	return ret;
+}
+
+static int mdio_renesas_ra_write(const struct device *dev, uint8_t prtad, uint8_t regad,
```
maass-hamburg, line 83: I looked at the user manual with the registers and at your proposed additions of the hal. I think it would be easier to just access the 2 needed registers directly, instead of using the hal. The proposed hal code is just much too big and unneeded.

for read and write you only need the MPSM register and foir init the MPIC. 
For MPIC I would recommend not setting PIS and LSC and leave that to the ethernet driver.

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,631 @@
...
+#else
+#define INT_GPIO(n)
+#endif /* ANY_INT_GPIO */
+
+#define PHY_GPY111_CONFIG(n)                                                                       \
+	BUILD_ASSERT(PHY_INST_GENERATE_DEFAULT_SPEEDS(n) != 0,                                     \
+		     "At least one valid speed must be configured for this driver");               \
+                                                                                                   \
+	static const struct maxlinear_gpy111_dev_config maxlinear_gpy111_dev_config_##n = {        \
+		.phy_addr = 0,                                                                     \
```
maass-hamburg: this should come from the `reg` prop of the dt

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,631 @@
...
+
+#if ANY_INT_GPIO
+	if (gpio_is_ready_dt(&cfg->reset_gpio)) {
+		ret = gpio_pin_configure_dt(&cfg->int_gpio, GPIO_INPUT);
+		if (ret < 0) {
+			LOG_ERR("Failed to configure INT pin (%d)", ret);
+			return ret;
+		}
+	}
+#endif /* ANY_INT_GPIO */
```
maass-hamburg: as long as we don't use the interrupt in this driver, it can be removed.

boards/renesas/ek_ra8d2/Kconfig.defconfig
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+if BOARD_EK_RA8D2
+
+if NETWORKING
+
+config NET_L2_ETHERNET
+	default y
+
+endif # NETWORKING
```
maass-hamburg:
```suggestion
configdefault NET_L2_ETHERNET
	default y
```

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,745 @@
...
+	clock_control_get_rate(config->phy_clock, NULL, &phy_ref_rate);
+
+	if (config->phy_clock_type == ETH_PHY_REF_CLK_INTERNAL) {
+		/* internal phy clock should be 25/50 MHz */
+		if (phy_ref_rate != ETHPHYCLK_25MHZ && phy_ref_rate != ETHPHYCLK_50MHZ) {
+			LOG_DBG("internal PHY clock %d differ from 25/50 MHz", phy_ref_rate);
+		}
+	} else if (config->phy_clock_type != ETH_PHY_REF_CLK_XTAL) {
+		LOG_DBG("invalid phy clock type");
+	}
```
maass-hamburg: instead of this could we maybe have a boolean dt prop like `needs-ethphyclk` or something like that. also the check could be done inside `renesas_ra_eswm_init` instead.

maass-hamburg (changes requested):

maass-hamburg:
@duynguyenxa please address all the comments and requested changes, before re-requesting reviews.

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: @jukkar , I've changed the implementation as your suggestion

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: The phy_link_up is unnecessary so I removed it

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's changed into `NET_AF_UNSPEC `

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: @maass-hamburg , Thank you, I've removed it

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: @maass-hamburg , I've fixed the code as your suggestion

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: Thank you, I've changed it

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: All gigabit check are removed

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: All gigabit check are removed

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: I've removed the dts property and used them as macro define

dts/bindings/ethernet/phy/maxlinear,gpy111.yaml
duynguyenxa: It's removed

dts/bindings/mdio/renesas,ra-mdio-rmac.yaml
duynguyenxa: I removed this and let the RMAC ethernet driver setting the phy-connection-type

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
duynguyenxa: Thank you, I understand that the hal source code is big and many part of it are unused, but I would like to keep current implementation to keep it consistent with Renesas strategy to use FSP hal code in Zephyr shim driver layer

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: It's removed

boards/renesas/ek_ra8d2/Kconfig.defconfig
duynguyenxa: This is fixed for all RA8x2 boards

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: I changed this to get dt

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
maass-hamburg: The problem is that you are using a whole phy driver as a base for a mdio driver. While it might work, it is still a bad solution. I'm not a big fan of HALs, but they have their place and use, but it this case, the disadvantage of the HAL (lots of dead code) outweigh its benefit. Therefore please implement it directly.

Just wanting to keep the implementation consistent with Renesas strategy, is not enough for an argument here. I understand that HALs are most often used to save development time, when adding support, but in this case, it should be trivial enough with just 2 registers to do it without it.

duynguyenxa: Another reason I prefer not to change this is that the R_RMAC_PHY* functions are also invoked by the R_RMAC* and R_LAYER3_SWITCH* functions within the HAL. Therefore, R_RMAC_PHY must still be built, even if it is not used directly by the Zephyr shim driver code.

maass-hamburg: That is a problem, but one that should be fixed in the hal. Communication with the with the phy should exclusively happen over zephyrs ethernet phy api. Information about the link up state of the phy should only be shared between the ethernet driver and the phy via zephyrs phy api.

The ethernet driver **must not** communicate with the phy without using the ethernet phy api. The ethernet driver should also only use `phy_link_callback_set()` to interact with the phy. No raw commands over mdio (the phy driver can do it, but not the ethernet driver).

The ethernet driver should even not know what the phy driver uses to communicate with the phy. While most phys use MDIO, there are phys that use spi (not in zephyr tree), these also have to work. Or if the "ethernet-phy-fixed-link" phy is used, i don't want to see any communication on the mdio bus.

duynguyenxa: Yes, I understand. I have already made several modifications to the HAL driver so that all PHY communication is handled exclusively via Zephyr’s Ethernet PHY API. However, we also want to keep the HAL changes as minimal as possible to reduce maintenance overhead when updating to newer HAL versions, and this approach satisfies that requirement.

maass-hamburg: Simply use 
```
#ifndef __ZEPHYR__
  // non zephyr code
#endif /* __ZEPHYR__ */
```
in the hal code to "remove" the unneeded parts, this shouldn't make it too difficult, when there are updates of the HAL. You could even upstream these parts internally.

But in general using a HAL, especially for ethernet, often leads to sub-optimal performance, as data is often copied more often then necessary. At least if the hal don't uses zephyrs `struct net_buf`

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: I moved this check to the `renesas_ra_eswm_init ` and changed to check only when internal clock is used

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
duynguyenxa: @maass-hamburg , I changed the implement of mdio driver, please help to check

boards/renesas/mck_ra8t2/board.c
```diff
@@ -0,0 +1,46 @@
...
+
+		gpio_pin_set_dt(&phy_reset_gpios, 0);
+		k_msleep(phy_reset_deassert);
+	}
+
+	return 0;
+}
+
+SYS_INIT(phy_init, POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_DEFAULT);
+#endif
```
maass-hamburg, line 52: the "ethernet-phy" has support for reseting the phy durring init using gpios, you just need to set the dt props "reset-gpios", "reset-assert-duration-us" and "reset-deassertion-timeout-ms".

If the reset pin is for both phys and you there you want to do it here, you can hard code them here as defines in this file, as this is board specific.

include/zephyr/drivers/ethernet/eth_renesas_ra_rmac.h
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @defgroup renesas_ra_eth_rmac_defs Renesas RA Ethernet Definitions
+ * @brief Common constants for Renesas RA Ethernet drivers.
+ * @{
+ */
+
+#define RENESAS_OUI_B0 0x74 /**< OUI byte 0 */
+#define RENESAS_OUI_B1 0x90 /**< OUI byte 1 */
+#define RENESAS_OUI_B2 0x50 /**< OUI byte 2 */
```
maass-hamburg: not used

drivers/ethernet/eth_renesas_ra_rmac.h
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) 2025 Renesas Electronics Corporation
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
```
maass-hamburg, line 6: this file can be moved inside `drivers/ethernet`.

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+	const struct device *dev = (const struct device *)arg1;
+	struct eth_renesas_ra_data *data = dev->data;
+
+	ARG_UNUSED(unused1);
+	ARG_UNUSED(unused2);
+
+	while (1) {
+		if (k_sem_take(&data->rx_sem, K_MSEC(100))) {
+			continue;
+		}
```
maass-hamburg: why not use a K_FOREVER here?

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+		k_sem_reset(&data->rx_sem);
+		return;
+	} else if (fsp_err != FSP_SUCCESS) {
+		LOG_DBG("Failed to read from FIFO");
+		ret = -EIO;
+		goto rx_end;
+	}
+
+	/* read again for remaining data */
+	k_sem_give(&data->rx_sem);
```
maass-hamburg: make this a loop to eliminate the `k_sem_give`. you can also make the function return a bool, and then do something like 
```
while(renesas_ra_eth_rx()) {
}
```

maass-hamburg (changes requested):

boards/renesas/mck_ra8t2/board.c
duynguyenxa: @maass-hamburg , I changed it into hard code

drivers/ethernet/eth_renesas_ra_rmac.h
duynguyenxa: This file is used by both mdio and ethernet driver source

maass-hamburg: yes and mdio is now in `drivers/ethernet/mdio` and ethernet in `drivers/ethernet`

From the mdio driver you can do `#include "../eth_renesas_ra_rmac.h".

include/zephyr/drivers/ethernet/eth_renesas_ra_rmac.h
duynguyenxa: Thank you, I forgot to remove this when change to new MAC address struct, it's removed

drivers/ethernet/eth_renesas_ra_rmac.h
duynguyenxa: oh I thought Zephyr coding rule prevent this, it's fixed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: I changed it into K_FOREVER

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: I changed the implementation, please help to check, thank you

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+		/* Nothing to receive, all desc in queue was read */
+		k_sem_reset(&data->rx_sem);
+		return false;
+	} else if (fsp_err != FSP_SUCCESS) {
+		LOG_DBG("Failed to read from FIFO");
+		ret = -EIO;
+		goto rx_end;
+	}
+
+	pkt = net_pkt_rx_alloc_with_buffer(data->iface, (size_t)len, NET_AF_UNSPEC, 0, K_MSEC(100));
```
maass-hamburg: make the timeout a define

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+	ARG_UNUSED(unused1);
+	ARG_UNUSED(unused2);
+
+	while (1) {
+		if (k_sem_take(&data->rx_sem, K_FOREVER)) {
+			continue;
+		}
+
+		while(renesas_ra_eth_rx(dev)) {
+			k_yield();
```
maass-hamburg: is this yielding necessary?

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+	const struct device *dev = (const struct device *)arg1;
+	struct eth_renesas_ra_data *data = dev->data;
+
+	ARG_UNUSED(unused1);
+	ARG_UNUSED(unused2);
+
+	while (1) {
+		if (k_sem_take(&data->rx_sem, K_FOREVER)) {
+			continue;
+		}
```
maass-hamburg: when used with `K_FOREVER`, k_sem_take will always return `0`, therefore we can ignore the return value here

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+		ret = (fsp_err == FSP_ERR_ETHER_ERROR_TRANSMIT_BUFFER_FULL) ? -ENOBUFS : -EIO;
+		LOG_DBG("Write to FIFO failed, err=%d", fsp_err);
+	}
+
+tx_end:
+
+#if defined(CONFIG_ETH_RENESAS_RA_USE_HW_WRITEBACK)
+	if (fsp_err == FSP_SUCCESS) {
+		/* wait descriptor write back complete */
+		ret = k_sem_take(&data->tx_sem, K_MSEC(1));
```
maass-hamburg: timeout should be a define

drivers/ethernet/eth_renesas_ra_rmac.h
maass-hamburg: it's still there, forgot to push?

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+	int ret = 0;
+
+	ret = pinctrl_apply_state(config->pin_cfg, PINCTRL_STATE_DEFAULT);
+	if (ret < 0) {
+		return ret;
+	}
+
+	(void)net_eth_mac_load(&config->mcfg, data->mac_addr);
+
+	data->fsp_cfg->p_mac_address = &data->mac_addr[0];
```
maass-hamburg:
```suggestion
	data->fsp_cfg->p_mac_address = data->mac_addr;
```

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+		.zerocopy = ETH_RENESAS_RA_DATA_BUF_MODE,                                          \
+		.multicast = ETHER_MULTICAST_ENABLE,                                               \
+		.promiscuous = ETHER_PROMISCUOUS_DISABLE,                                          \
+		.flow_control = ETHER_FLOW_CONTROL_DISABLE,                                        \
+		.p_callback = &eth_rmac_cb,                                                        \
+		.p_context = (void *)DEVICE_DT_INST_GET(n),                                        \
+		.p_extend = &eth##n##_ext_cfg,                                                     \
+	};                                                                                         \
+	static struct eth_renesas_ra_data eth##n##_renesas_ra_data = {                             \
+		.mac_addr = DT_INST_PROP_OR(n, local_mac_address, {0}),                            \
```
maass-hamburg: this line is no longer needed, already handled by NET_ETH_MAC_DT_INST_CONFIG_INIT

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+		ETH_RENESAS_RA_DATA_BUF_PROP_DECLARE(n)};                                          \
+	static struct eth_renesas_ra_config eth##n##_renesas_ra_config = {                         \
+		.mcfg = NET_ETH_MAC_DT_INST_CONFIG_INIT(n),                                        \
+		.phy_conn_type = DT_INST_ENUM_IDX(n, phy_connection_type),                         \
+		.pin_cfg = PINCTRL_DT_INST_DEV_CONFIG_GET(n),                                      \
+		.phy_dev = DEVICE_DT_GET(DT_INST_PHANDLE(n, phy_handle)),                          \
+	};                                                                                         \
+	ETH_NET_DEVICE_DT_INST_DEFINE(n, renesas_ra_eth_init, NULL, &eth##n##_renesas_ra_data,     \
+				      &eth##n##_renesas_ra_config, CONFIG_ETH_INIT_PRIORITY,       \
+				      &eth_renesas_ra_api, NET_ETH_MTU /*MTU*/);
```
maass-hamburg:
```suggestion
				      &eth_renesas_ra_api, NET_ETH_MTU);
```
this comment looks off

maass-hamburg (changes requested):

drivers/ethernet/mdio/Kconfig.renesas_ra
```diff
@@ -9,3 +9,30 @@ config MDIO_RENESAS_RA
 	select USE_RA_FSP_ETHER_PHY
 	help
 	  Enable RENESAS MDIO support.
+
+config MDIO_RENESAS_RA_RMAC
+	bool "RENESAS MDIO RMAC controller driver"
+	default y
+	depends on DT_HAS_RENESAS_RA_MDIO_RMAC_ENABLED
+	select PINCTRL
+	select USE_RA_FSP_RMAC_PHY
```
maass-hamburg: is this still needed?

boards/renesas/ek_ra8m2/ek_ra8m2_r7ka8m2jflcac_cm85.dts
```diff
@@ -183,3 +184,35 @@
 mikrobus_serial: &uart7 {};
 
 mikrobus_spi: &spi1 {};
+
+&eswm {
```
maass-hamburg, line 188: please add
```
  - netif:eth
```

to the `supported` section in the boards `.yaml` files.
This way twister knows that these boards support ethernet and build tests them.

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
```diff
@@ -0,0 +1,188 @@
...
+		.mdio_capture_time = 0,                                                            \
+	};                                                                                         \
+	static const struct mdio_renesas_ra_config renesas_ra_mdio##n##_cfg = {                    \
+		.eswclk_dev = DEVICE_DT_GET(DT_CLOCKS_CTLR_BY_NAME(DT_INST_PARENT(n), eswclk)),    \
+		.pincfg = PINCTRL_DT_INST_DEV_CONFIG_GET(n),                                       \
+		.mdc_clock_rate = DT_INST_PROP(n, clock_frequency),                                \
+		.reg =                                                                             \
+			{                                                                          \
+				.mpsm = (uint32_t *)DT_REG_ADDR_BY_NAME(DT_DRV_INST(n), MPSM),     \
+				.mpic = (uint32_t *)DT_REG_ADDR_BY_NAME(DT_DRV_INST(n), MPIC),     \
```
maass-hamburg: you can use `DT_INST_REG_ADDR_BY_NAME`, also make register name lowercase.

drivers/ethernet/Kconfig.renesas_ra_rmac
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2025 Renesas Electronics Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+config ETH_RENESAS_RA_RMAC
+	bool "Renesas RA ethernet RMAC driver"
+	default y
+	depends on DT_HAS_RENESAS_RA_ETHERNET_RMAC_ENABLED
+	select USE_RA_FSP_RMAC
+	select USE_RA_FSP_LAYER3_SWITCH
+	select MDIO
```
maass-hamburg: the phy selects mdio, not the eth driver, remove line

drivers/ethernet/Kconfig.renesas_ra_rmac
```diff
@@ -0,0 +1,48 @@
...
+	  Enable this option to wait for Ethernet descriptor writeback, ensuring
+	  that the TX frame is written successfully. Note that enabling this option
+	  may cause ethernet write operations to take slightly longer.
+
+config ESWM_RENESAS_RA_INIT_PRIORITY
+	int "Ethernet driver init priority"
+	default 55
+	help
+	  Ethernet switch module ESWM device driver initialization priority.
+
```
maass-hamburg: as this is the parent of the ethernet driver, you can use `CONFIG_ETH_INIT_PRIORITY` for that too, as the ordinance/dependency from the dt is used as a second prio and parents init before their children.

dts/bindings/ethernet/phy/maxlinear,gpy111.yaml
```diff
@@ -0,0 +1,63 @@
...
+    type: string
+    enum:
+      - "tx_skew_0n0"
+      - "tx_skew_0n5"
+      - "tx_skew_1n0"
+      - "tx_skew_1n5"
+      - "tx_skew_2n0"
+      - "tx_skew_2n5"
+      - "tx_skew_3n0"
+      - "tx_skew_3n5"
```
maass-hamburg: please remove the `tx_skew_` prefix `1.5` and `0` would be better

maass-hamburg (changes requested):

drivers/ethernet/eth_renesas_ra_rmac.h
duynguyenxa: ah sorry, I forgot to push, please help to check again

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: I changed it into define

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: Not really necessary, I just refer to some other implement, anyway I've removed it

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: The condition check is removed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: this is changed to a define

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: Thank you, it's fixed as your suggestion

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's removed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's removed

drivers/ethernet/mdio/Kconfig.renesas_ra
duynguyenxa: I've just removed it

boards/renesas/ek_ra8m2/ek_ra8m2_r7ka8m2jflcac_cm85.dts
duynguyenxa: All related boards update are added

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
duynguyenxa: It's changed to use DT_INST_REG_ADDR_BY_NAME

drivers/ethernet/Kconfig.renesas_ra_rmac
duynguyenxa: I've removed it

drivers/ethernet/Kconfig.renesas_ra_rmac
duynguyenxa: It's changed to `CONFIG_ETH_INIT_PRIORITY`

dts/bindings/ethernet/phy/maxlinear,gpy111.yaml
duynguyenxa: I changed the enum value

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+	.pclk_subsys.mstp = (uint32_t)DT_INST_CLOCKS_CELL_BY_NAME(0, pclk, mstp),
+	.pclk_subsys.stop_bit = DT_INST_CLOCKS_CELL_BY_NAME(0, pclk, stop_bit),
+	.ethphyclk_subsys.mstp = (uint32_t)DT_INST_CLOCKS_CELL_BY_NAME(0, ethphyclk, mstp),
+	.ethphyclk_subsys.stop_bit = DT_INST_CLOCKS_CELL_BY_NAME(0, ethphyclk, stop_bit),
+	.ethphyclk_enable = DT_INST_FOREACH_CHILD_STATUS_OKAY_SEP(0, ETH_USE_INTERNAL_PHY_CLK, (+)),
+	.pin_cfg = PINCTRL_DT_INST_DEV_CONFIG_GET(0),
+	.en_irq = &renesas_ra_eswm_init_irq,
+};
+DEVICE_DT_INST_DEFINE(0, renesas_ra_eswm_init, NULL, &eswm_data, &eswm_config, POST_KERNEL,
+		      CONFIG_ESWM_RENESAS_RA_INIT_PRIORITY, NULL);
```
maass-hamburg:
```suggestion
DEVICE_DT_INST_DEFINE(0, renesas_ra_eswm_init, NULL, &eswm_data, &eswm_config, POST_KERNEL,
		      CONFIG_ETH_INIT_PRIORITY, NULL);
```

you could directly use CONFIG_ETH_INIT_PRIORITY here or is there a use case where CONFIG_ESWM_RENESAS_RA_INIT_PRIORITY would be different to CONFIG_ETH_INIT_PRIORITY

duynguyenxa: No, for now it just need to be initialize before the rmac and the mdio, I changed it to use  `CONFIG_ETH_INIT_PRIORITY` and remove `CONFIG_ESWM_RENESAS_RA_INIT_PRIORITY`

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
```diff
@@ -0,0 +1,188 @@
...
+
+	ret = clock_control_get_rate(config->eswclk_dev, NULL, &eswclk);
+	if (ret < 0) {
+		return ret;
+	}
+
+	data->mdio_hold_time = DIV_ROUND_UP(eswclk * CONFIG_MDIO_RENESAS_RA_RMAC_MDIO_HOLD_NS, 1e9);
+	data->mdio_capture_time =
+		DIV_ROUND_UP(eswclk * CONFIG_MDIO_RENESAS_RA_RMAC_MDIO_CAPTURE_NS, 1e9);
+
```
maass-hamburg: these values are only used during init, they don't need to be saved in the data struct

drivers/ethernet/mdio/Kconfig.renesas_ra
```diff
@@ -9,3 +9,29 @@ config MDIO_RENESAS_RA
...
+	help
+	  Time for changing MDO after rising edge of MDC in nanoseconds.
+	  This value depends on PHY chip AC Characteristics. Use 0 to disable.
+
+config MDIO_RENESAS_RA_RMAC_MDIO_CAPTURE_NS
+	int "MDIO capture time"
+	default 0
+	help
+	  Time for capturing MDI after rising edge of MDC in nanoseconds.
+	  This value depends on PHY chip AC Characteristics. Use 0 to disable.
```
maass-hamburg: if these are phy specific, i think they would make more sense as dt props of the mdio controller.

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,562 @@
...
+	if (cfg->no_reset == false) {
+		ret = reset(dev);
+		if (ret < 0) {
+			LOG_ERR("Failed to reset PHY (%d): %d", cfg->phy_addr, ret);
+			return ret;
+		}
+	}
+
+	if (get_id(dev, &phy_id) == 0) {
+		if (phy_id == MII_INVALID_PHY_ID) {
```
maass-hamburg: as this is just for one phy model, the phy id is in the datasheet, you can test for that and not just if it is valid or not

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,562 @@
...
+}
+
+static int maxlinear_gpy111_init(const struct device *dev)
+{
+	const struct maxlinear_gpy111_dev_config *const cfg = dev->config;
+	struct maxlinear_gpy111_dev_data *const data = dev->data;
+	uint32_t phy_id;
+	int ret = 0;
+
+	data->state.is_up = false;
```
maass-hamburg: line can be removed, false/0 already is the initial value

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+	k_sem_init(&data->rx_sem, 0, K_SEM_MAX_LIMIT);
+#if defined(CONFIG_ETH_RENESAS_RA_USE_HW_WRITEBACK)
+	k_sem_init(&data->tx_sem, 0, 1);
+#endif
+
+	k_thread_create(&data->rx_thread, data->rx_thread_stack,
+			K_KERNEL_STACK_SIZEOF(data->rx_thread_stack), eth_rx_thread, (void *)dev,
+			NULL, NULL, K_PRIO_COOP(CONFIG_ETH_RENESAS_RA_RX_THREAD_PRIORITY), 0,
+			K_NO_WAIT);
+	k_thread_name_set(&data->rx_thread, "eth_renesas_ra_rx");
```
maass-hamburg: in the thread, you are using `data->iface`, it will only be set in the iface init. move the creation of this thread there too.

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+	.gwcaclk_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_NAME(0, gwcaclk)),
+	.pclk_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_NAME(0, pclk)),
+	.eswclk_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_NAME(0, eswclk)),
+	.eswphyclk_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_NAME(0, eswphyclk)),
+	.ethphyclk_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_NAME(0, ethphyclk)),
+	.pclk_subsys.mstp = (uint32_t)DT_INST_CLOCKS_CELL_BY_NAME(0, pclk, mstp),
+	.pclk_subsys.stop_bit = DT_INST_CLOCKS_CELL_BY_NAME(0, pclk, stop_bit),
+	.ethphyclk_subsys.mstp = (uint32_t)DT_INST_CLOCKS_CELL_BY_NAME(0, ethphyclk, mstp),
+	.ethphyclk_subsys.stop_bit = DT_INST_CLOCKS_CELL_BY_NAME(0, ethphyclk, stop_bit),
+	.ethphyclk_enable = DT_INST_FOREACH_CHILD_STATUS_OKAY_SEP(0, ETH_USE_INTERNAL_PHY_CLK, (+)),
```
maass-hamburg:
```suggestion
	.ethphyclk_enable = DT_INST_FOREACH_CHILD_STATUS_OKAY_SEP(0, ETH_USE_INTERNAL_PHY_CLK, (||)),
```

also `ethphyclk_enable` can then be a bool

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+
+/* Queues declare */
+#define ETH_TX_QUEUE_DECLARE(idx, n)                                                               \
+	{                                                                                          \
+		.queue_cfg =                                                                       \
+			{                                                                          \
+				.array_length = ETH_TX_QUEUE_LEN(n),                               \
+				.p_descriptor_array = eth##n##_tx_desc_array##idx,                 \
+				.p_ts_descriptor_array = NULL,                                     \
+				.ports = 1 << DT_INST_PROP(n, channel),                            \
```
maass-hamburg: use BIT() macro

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+			},                                                                         \
+	}
+#define ETH_RX_QUEUE_DECLARE(idx, n)                                                               \
+	{                                                                                          \
+		.queue_cfg =                                                                       \
+			{                                                                          \
+				.array_length = ETH_RX_QUEUE_LEN(n),                               \
+				.p_descriptor_array = eth##n##_rx_desc_array##idx,                 \
+				.p_ts_descriptor_array = NULL,                                     \
+				.ports = 1 << DT_INST_PROP(n, channel),                            \
```
maass-hamburg: here too

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+	K_KERNEL_STACK_MEMBER(rx_thread_stack, CONFIG_ETH_RENESAS_RA_RX_THREAD_STACK_SIZE);
+	struct k_sem rx_sem;
+#if defined(CONFIG_ETH_RENESAS_RA_USE_HW_WRITEBACK)
+	struct k_sem tx_sem;
+#endif
+#if defined(CONFIG_NET_STATISTICS_ETHERNET)
+	struct net_stats_eth stats;
+#endif
+	rmac_instance_ctrl_t *fsp_ctrl;
+	ether_cfg_t *fsp_cfg;
```
maass-hamburg: if you already have to save them too, why don't you save them in this struct?

```suggestion
	rmac_instance_ctrl_t fsp_ctrl;
	ether_cfg_t fsp_cfg;
```

maass-hamburg (changes requested):

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
duynguyenxa: I moved them to config struct and get value from dts property

drivers/ethernet/mdio/Kconfig.renesas_ra
duynguyenxa: I moved them to config struct and get value from dts property

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: Thank you, it's changed

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: it's removed

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
```diff
@@ -0,0 +1,187 @@
...
+static DEVICE_API(mdio, mdio_renesas_ra_api) = {
+	.read = mdio_renesas_ra_read,
+	.write = mdio_renesas_ra_write,
+};
+
+#define RENSAS_RA_MDIO_RMAC_DEFINE(n)                                                              \
+	BUILD_ASSERT(DT_INST_CHILD_NUM(n) <= BSP_FEATURE_ETHER_NUM_CHANNELS);                      \
+	PINCTRL_DT_INST_DEFINE(n);                                                                 \
+                                                                                                   \
+	static struct mdio_renesas_ra_data renesas_ra_mdio##n##_data = {0};                        \
```
maass-hamburg:
```suggestion
	static struct mdio_renesas_ra_data renesas_ra_mdio##n##_data;                        \
```
0 is already the default for static vars.

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,561 @@
...
+		if (ret < 0) {
+			LOG_ERR("Failed to reset PHY (%d): %d", cfg->phy_addr, ret);
+			return ret;
+		}
+	}
+
+	if (get_id(dev, &phy_id) == 0) {
+		if (phy_id != GPY111_PHY_ID) {
+			LOG_ERR("PHY (%d) ID 0x%X not as expected", cfg->phy_addr, phy_id);
+			LOG_ERR("Expected PHY ID: 0x%X", GPY111_PHY_ID);
```
maass-hamburg: combine it in one message

drivers/ethernet/phy/phy_maxlinear_gpy111.c
```diff
@@ -0,0 +1,561 @@
...
+
+			return -EINVAL;
+		}
+
+		LOG_INF("PHY (%d) ID %X", cfg->phy_addr, phy_id);
```
maass-hamburg: i don't think this is anymore needed, it will always be the same

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,753 @@
...
+		R_ESWM->MIIRR =
+			R_ESWM->MIIRR | (1 << (R_ESWM_MIIRR_RGRST_Pos + data->fsp_cfg.channel));
+		break;
+	}
+
+	default: {
+		/* MII or GMII. */
+		*p_miicr_register = 0;
+		break;
+	}
```
maass-hamburg: the enclosing `{}` for each case can be removed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: I moved it into the iface init function

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's changed to bool

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's changed to BIT()

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's changed to BIT()

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: It's fixed

drivers/ethernet/eth_renesas_ra_rmac.c
```diff
@@ -0,0 +1,754 @@
...
+	.get_capabilities = eth_renesas_ra_get_capabilities,
+	.get_phy = eth_renesas_ra_get_phy,
+	.send = eth_renesas_ra_tx,
+};
+
+#define DT_DRV_COMPAT renesas_ra_eswm
+
+#define ETH_USE_INTERNAL_PHY_CLK(id)                                                               \
+	COND_CODE_1(UTIL_AND(DT_NODE_HAS_COMPAT(id, renesas_ra_ethernet_rmac),                     \
+		 DT_ENUM_HAS_VALUE(id, phy_clock_type, internal)), (1), (0))
```
maass-hamburg:
```suggestion
	UTIL_AND(DT_NODE_HAS_COMPAT(id, renesas_ra_ethernet_rmac),                     \
		 DT_ENUM_HAS_VALUE(id, phy_clock_type, internal)
```
no need for the COND_CODE_1 here

maass-hamburg (changes requested):

drivers/ethernet/mdio/mdio_renesas_ra_rmac.c
duynguyenxa: it's removed

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: Log is combined

drivers/ethernet/phy/phy_maxlinear_gpy111.c
duynguyenxa: it's removed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: Removed

drivers/ethernet/eth_renesas_ra_rmac.c
duynguyenxa: COND_CODE_1 is removed

maass-hamburg (approved):

khoa-nguyen-18 (approved):

PR #100996, opened 2025-12-14
drivers/soc: nxp: add CACHE64 init-only driver for RT5xx/RW6xx

Holt-Sun:
## Summary

Add CACHE64 devicetree binding and init-only driver for NXP RT5xx/RW6xx SoCs.

Replaces hard-coded cache initialization with DT-based configuration, allowing board-specific cache policy tuning.

## Changes
- New `nxp,cache64` binding and driver
- Enable CACHE64 on frdm_rw612, rd_rw612_bga, mimxrt595_evk  
- Remove legacy soc.c cache init
- Update cache documentation

Tested on RW612/RT595 EVK boards.


Holt-Sun:
@Albort12138

Holt-Sun:
@Albort12138

drivers/cache/Kconfig.mcux
Albort12138, line 1: Instead of creating a separate Kconfig file for each cache implementation, we could introduce a single Kconfig.mcux to include all MCUX cache implementations—similar to what we already do in other device drivers: https://github.com/zephyrproject-rtos/zephyr/blob/cb93f97bd63207211d92f7c5a0ffd8d5e270dd42/drivers/adc/Kconfig.mcux

dts/bindings/cache/nxp,cache64.yaml
```diff
@@ -0,0 +1,77 @@
...
+      Boundaries are specified in the FlexSPI physical address domain.
+      The number of regions is determined by the array length (N boundaries = N+1 regions).
+
+      Note: The last region extends to the end of addressable space. While its
+      boundary_addr is not used by the hardware for configuration, it should still
+      specify the actual end address for documentation and potential validation purposes.
+
+      Example 1: Single boundary (2 regions)
+        regions = <0x01000000 2>;
+        - Region 0: 0x00000000-0x00FFFFFF -> Write-back (16MB)
```
Albort12138: The meaning of the first item is confusing—it seems to represent the offset of the region’s end address.
I suggest changing the format to:
`<startAddr size policy>`

Holt-Sun: I prefer to keep the boundary-based format. It matches the hardware (monotonic boundary registers) and easy to validate, and changing to <start size policy> would be a breaking binding change and can imply non-monotonic or gappy regions that the HW can’t represent.

I agree that the confusion stems from description word, so I update the description.

drivers/cache/Kconfig.mcux
Holt-Sun: Done.

Albort12138 (approved):

PR #100997, opened 2025-12-14
drivers: flash: mcux_flexspi: fix unused variable warnings

Holt-Sun:
Fix compiler warnings when CONFIG_HAS_MCUX_CACHE is not defined:
- Add pointer validity checks before cache invalidation operations
- Remove unnecessary intermediate 'size' variable and use 'len' 
  directly in DCACHE_InvalidateByRange() to eliminate unused 
  variable when CONFIG_HAS_MCUX_CACHE is disabled

### DIFF
--- a/drivers/flash/flash_mcux_flexspi_nor.c
+++ b/drivers/flash/flash_mcux_flexspi_nor.c
@@ -357,7 +357,9 @@ static int flash_flexspi_nor_write(const struct device *dev, off_t offset,
 		const void *buffer, size_t len)
 {
 	struct flash_flexspi_nor_data *data = dev->data;
-
+#ifdef CONFIG_HAS_MCUX_CACHE
+	size_t size = len;
+#endif
 	if (!buffer) {
 		return -EINVAL;
 	}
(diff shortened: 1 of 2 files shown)


drivers/flash/flash_mcux_flexspi_nor.c
```diff
@@ -420,7 +423,7 @@ static int flash_flexspi_nor_write(const struct device *dev, off_t offset,
 	}
 
 #ifdef CONFIG_HAS_MCUX_CACHE
-	DCACHE_InvalidateByRange((uint32_t) dst, size);
+	DCACHE_InvalidateByRange((uint32_t) dst, len);
```
JimmyChen-NXP: This change is not aligned with the commit log

JimmyChen-NXP (approved):

drivers/flash/flash_mcux_flexspi_nor.c
zejiang0jason: the `len` has been changed previously in this function.

Holt-Sun: Good catch, thank you.

zejiang0jason (approved):

ZhaoxiangJin (approved):

PR #100999, opened 2025-12-14
drivers: sensor: adxl345: implement proper interrupt configuration

### LINKED ISSUE #100891: drivers/sensor/adi/adxl345: Overwriting of previous initialization
### Describe the bug

I have been using a Nucleo-F411RE board with a adxl345 breakout board, connected via I2C.

The commit that causes the issue is 9cbcdda: In adxl345_init a second value is written to the INT_MAP register (0x2F), overwriting the initially set value.

### Regression

- [x] This is a regression.

### Steps to reproduce

I built the zephyr/samples/sensor/accel_trig sample. When running the sample, no sensor reading appears in the log and no level change on the interrupt pin is observed.

After commenting out the lines with the second write and re-running the sample I could see the sensor readings in the log and the level change on the interrupt appears as expected.

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

kaessert:
Fix interrupt enable/disable sequence per ADXL345 datasheet recommendations.

The ADXL345 datasheet (page 19) states: "When initially configuring the interrupt pins, it is recommended that the functions and interrupt mapping be done before enabling the interrupts."

Main changes:

1. Remove adxl345_interrupt_config() which incorrectly enabled interrupts during driver initialization without proper sequencing. It's not clear to me what was the purpose of this function as all the required logic (setting gpio interrups, adxl mapping and enabling happens already in tigger_set). I checked other drivers and this seems to be the common pattern.

2. Implement proper INT_ENABLE disable/enable sequence in adxl345_submit_stream() when reconfiguring watermark interrupts:
   - Disable watermark interrupt in INT_ENABLE
   - Configure interrupt mapping in INT_MAP
   - Flush FIFO
   - Re-enable watermark interrupt in INT_ENABLE

3. Add datasheet-based comments in adxl345_trigger.c explaining non-obvious register operations:
   - AC/DC coupling modes (ACT_INACT_CTL register)
   - INT_MAP routing logic (0=INT1, 1=INT2)
   - Interrupt clearing by reading data registers

4. Fix bug in adxl345_trigger_set() where ADXL345_INT_MAP_DATA_RDY_MSK was hardcoded instead of using the correct int_mask variable.

5. Enhance accel_trig sample with activity detection support and improved error reporting for testing interrupt functionality.

Fixes #100891

Tested on ebyte_e73_tbb with samples `accel_polling` and `accel_trig`


drivers/sensor/adi/adxl345/adxl345_trigger.c
```diff
@@ -135,7 +139,7 @@ int adxl345_trigger_set(const struct device *dev,
 	}
 
 	if (handler) {
-		int_en = ADXL345_INT_MAP_DATA_RDY_MSK;
+		int_en = int_mask;
```
stemschmidt: This is the modification that prevents the accel_trig sample to run with my given config. Your implementation is correct, but it revealed some issues which were covered by setting a hard coded ADXL345_INT_MAP_DATA_RDY_MSK.

With my configuration a 0x20 is written to FIFO_CTL. This register has a field "Samples" which is set to "0" when 0x20 is written to FIFO_CTL. The description of "Samples" has this remark: "Entering a value of 0 in the samples bits immediately sets the watermark status bit in the INT_SOURCE register, regardless of which FIFO mode is selected."
The one bit that is set with 0x20 is the "Trigger Bit", which links the trigger event to either INT1 ("0") or INT2 ("1").
0x20 -> set watermark status bit in INT_SOURCE register, link trigger event to INT2

You modification does not set only ADXL345_INT_MAP_DATA_RDY_MSK, but also DXL345_INT_MAP_OVERRUN_MSK  and ADXL345_INT_MAP_WATERMARK_MSK (line 117-119 in this file). This int_mask is then used to map those events to INT1 (line 152) and enable the corresponding interrupts (line 173).
 
When I start the application I see that the interrupt line is immediately asserted, the INT_SOURCE register is read and gives 0x02. But as the watermark status bit cannot be cleared (see the description above) the interrupt line remains asserted and no further interrupt is detected as it is an edge interrupt.

I have no good advice on how to solve this. I would remove DXL345_INT_MAP_OVERRUN_MSK and ADXL345_INT_MAP_WATERMARK_MSK from being set to int_mask as this variable is not forwarded somewhere else (so the change stays local), but I would double-check with Luis who had added a comment just above these lines.

ttmut:
Could you please fix the conflicts? Thanks.

samples/sensor/accel_trig/src/main.c
ubieda, line 1: May we put the changes to the sample in another patch?

drivers/sensor/adi/adxl345/adxl345_trigger.c
ubieda: @stemschmidt are these concerns still valid or have they been addressed?

ubieda (commented):
I'd like some clarity on the issue of interrupts not being cleared. Other than that and the sample-changes it looks good.

samples/sensor/accel_trig/src/main.c
kaessert: What's the issue with the changes?

ubieda: The changes to the sample are not directly related to "drivers: sensor: adxl345: implement proper interrupt configuration", and would be better served in a separate patch/commit.

kaessert: i disagree because they serve as a test too. If you want i can make a separate PR but i personally feel its very much related

ubieda: I'm not asking for a separate PR. I'm asking for a separate commit in the same PR.

kaessert: Ah sure, sorry my bad i read that wrong. There you go :)

drivers/sensor/adi/adxl345/adxl345_trigger.c
kaessert: @stemschmidt I did a little bit of further digging and here is how i see it:
```
 Per the ADXL345 datasheet: "Entering a value of 0 in the samples bits immediately sets the watermark status bit in the INT_SOURCE register, regardless of which FIFO mode is selected."

  So from boot onward, INT_SOURCE bit 1 (watermark) is permanently asserted — the condition FIFO_entries >= 0 is always true and can never be cleared.

  Step 2 — trigger_set: WATERMARK interrupt gets enabled

  When the accel_trig sample calls sensor_trigger_set() with SENSOR_TRIG_DATA_READY, adxl345_trigger_set() at adxl345_trigger.c:117-119 builds:
  int_mask = ADXL345_INT_MAP_DATA_RDY_MSK |   // BIT(7) = 0x80
             ADXL345_INT_MAP_OVERRUN_MSK |     // BIT(0) = 0x01
             ADXL345_INT_MAP_WATERMARK_MSK;    // BIT(1) = 0x02
  // int_mask = 0x83

  The adxl345_configure_fifo(dev, BYPASSED, INT2, 0) call at line 148 is inside #ifdef CONFIG_ADXL345_STREAM — skipped in this config.

  At line 182, all three interrupts are enabled:
  adxl345_reg_write_mask(dev, ADXL345_INT_ENABLE, 0x83, 0x83);

  Step 3 — Immediate interrupt assertion

  The moment WATERMARK is enabled in INT_ENABLE, the INT pin goes HIGH because INT_SOURCE bit 1 was already asserted (from Samples=0 at boot). The GPIO fires once with GPIO_INT_EDGE_TO_ACTIVE.

  Step 4 — Interrupt handler reads INT_SOURCE = 0x02

  adxl345_thread_cb() at adxl345_trigger.c:28 calls adxl345_get_status(), which reads INT_SOURCE → 0x02 (only watermark bit set, no DATA_RDY yet).

  The handler checks:
  - ADXL345_STATUS_DATA_RDY(0x02) → (0x02 >> 7) & 0x1 = 0 — no DRDY handler called
  - ADXL345_STATUS_ACTIVITY(0x02) → (0x02 >> 4) & 0x1 = 0 — no activity handler called

  Nothing is dispatched. The watermark bit cannot be cleared because FIFO_entries >= 0 is always true.

  Step 5 — GPIO re-enabled but pin is stuck HIGH

  At adxl345_trigger.c:42-43, GPIO is re-enabled for edge detection. But the INT pin is still HIGH because the watermark bit never cleared. No 0→1 edge transition will ever occur → dead lock. No further
  interrupts are ever received.

  Why the Old Hardcoded Code "Worked"

  In the original code, int_en was hardcoded to ADXL345_INT_MAP_DATA_RDY_MSK (0x80). So the final INT_ENABLE write was:
  adxl345_reg_write_mask(dev, ADXL345_INT_ENABLE, 0x83, 0x80);
  This only enabled DATA_RDY (bit 7). WATERMARK (bit 1) and OVERRUN (bit 0) were masked in INT_MAP but never enabled in INT_ENABLE. The permanently-asserted watermark bit in INT_SOURCE was invisible because it
  wasn't routed to the INT pin.

  The PR fix (int_en = int_mask) correctly enables all three bits — but now the stuck watermark becomes visible.

  Three Issues Intersecting

  1. adxl345_init() writes Samples=0 to FIFO_CTL even when STREAM is disabled (adxl345.c:500-504). This permanently asserts the watermark bit in INT_SOURCE.
  2. int_mask for DATA_READY includes WATERMARK (adxl345_trigger.c:117-119). The comment says "Enabling DRDY means not using Watermark interrupt" — the intent is to disable it, but int_en = int_mask then enables
   all three bits including watermark.
  3. adxl345_thread_cb() doesn't read data registers (adxl345_trigger.c:20-45). Even if watermark could theoretically clear, the handler only reads INT_SOURCE (which doesn't clear watermark/DRDY bits per the
  datasheet).
```

I see three potential ways for a fix

```
  Option A — Fix int_mask to only enable DATA_RDY (minimal, reviewer-suggested):
  Remove WATERMARK and OVERRUN from int_mask when enabling DRDY trigger:
  // For SENSOR_TRIG_DATA_READY:
  int_mask = ADXL345_INT_MAP_DATA_RDY_MSK;  // Only enable DATA_RDY
  The comment about "not using Watermark" already expresses this intent. Watermark and overrun would remain disabled.

  Option B — Fix FIFO_CTL to use Samples=1 at boot:
  In adxl345_init(), pass fifo_samples=1 instead of cfg->fifo_config.fifo_samples (which is 0 when no DT watermark is configured):
  adxl345_configure_fifo(dev, ADXL345_FIFO_BYPASSED, ADXL345_INT2,
                         MAX(cfg->fifo_config.fifo_samples, 1));
  This prevents the permanent watermark assertion.

  Option C — Split int_mask and int_en:
  Use int_mask for the bits to manipulate in INT_MAP, but only enable DATA_RDY in INT_ENABLE:
  int_mask = ADXL345_INT_MAP_DATA_RDY_MSK |
             ADXL345_INT_MAP_OVERRUN_MSK |
             ADXL345_INT_MAP_WATERMARK_MSK;
  int_en = ADXL345_INT_MAP_DATA_RDY_MSK;  // Only enable DATA_RDY in INT_ENABLE
  This keeps watermark/overrun mapped to INT1 (so they don't interfere on INT2) but only enables DATA_RDY to actually fire. This is closest to what the old hardcoded behavior achieved, but now intentional.
```

I think option C is the most correct and implemented that one. Could you please verify it on your end?

stemschmidt: I will have a look at this tomorrow and run it on my setup

stemschmidt: Hi @kaessert, @ubieda, the sample runs now out of the box! @kaessert, thank you for your fast response!

kaessert: That was a tough one 😅 Thank you so much for verifying @stemschmidt! @ubieda We're good then as it seems.

ubieda (approved):

MaureenHelm (approved):

PR #101004, opened 2025-12-14
lib: cobs: implement streaming cobs variant

mkschreder:
## Summary
This PR adds a streaming (incremental) COBS encode/decode API to Zephyr’s COBS utility, along with comprehensive unit tests.

## Why
Many transports deliver data in small, arbitrary chunks (for example, UART RX buffers). In these cases, treating COBS encoding/decoding as a single “whole buffer in / whole buffer out” operation is awkward. This paves the way for implementing new cobs based uart slip network interface which I will push in a separate PR. 

## What’s added
- **New state-machine based streaming APIs** for COBS:
  - Initialize/reset encoder/decoder state
  - Encode to an output buffer across multiple calls
  - Decode incrementally until a frame delimiter is encountered
- **Tests covering streaming behavior**, including:
  - Chunked/fragmented input handling
  - Boundary conditions (notably delimiter-at-boundary cases)
  - Roundtrip encode/decode validation

## Compatibility
- The existing COBS APIs remain intact; this change adds new capabilities without removing or redefining existing behavior.

## Testing
- Extended `tests/lib/cobs` with streaming-focused test coverage.

### DIFF
--- a/include/zephyr/data/cobs.h
+++ b/include/zephyr/data/cobs.h
@@ -123,30 +123,16 @@ struct cobs_decode_state {
 /**
  * @brief Initialize a COBS streaming encoder context
  *
- * @param self        Pointer to the COBS encoder state machine context
+ * @param self        Pointer to the COBS encoder state machine context (must not be NULL)
  */
 void cobs_encode_init(struct cobs_encode_state *self);
 
 /**
- * @brief Reset a COBS streaming encoder context
- *
- * @param self        Pointer to the COBS encoder state machine context
- */
-void cobs_encode_reset(struct cobs_encode_state *self);
-
-/**
  * @brief Initialize a COBS streaming decoder context
  *
- * @param self        Pointer to the COBS decoder state machine context
+ * @param self        Pointer to the COBS decoder state machine context (must not be NULL)
  */
 void cobs_decode_init(struct cobs_decode_state *self);
-
-/**
- * @brief Reset a COBS streaming decoder context
- *
- * @param self        Pointer to the COBS decoder state machine context
- */
-void cobs_decode_reset(struct cobs_decode_state *self);
 
 /**
  * @brief Encode data using a COBS streaming state machine
@@ -159,12 +145,13 @@ void cobs_decode_reset(struct cobs_decode_state *self);
  * @param src         Source buffer to encode (data is pulled from it)
  * @param dst         Destination buffer for encoded data
  * @param dst_len     On input: capacity of dst; On output: bytes written
+ * @param delimiter   Delimiter byte value (typically COBS_DEFAULT_DELIMITER)
  *
  * @retval 0        Success
- * @retval -EINVAL  Invalid parameters
+ * @retval -EINVAL  Invalid parameters or delimiter conflicts with overhead byte
  */
 int cobs_encode_stream(struct cobs_encode_state *self, struct net_buf *src,
-			uint8_t *dst, size_t *dst_len);
+			uint8_t *dst, size_t *dst_len, uint8_t delimiter);
 
 /**
  * @brief Finalize COBS encoding and flush remaining data
@@ -174,12 +161,14 @@ int cobs_encode_stream(struct cobs_encode_state *self, struct net_buf *src,
  * @param self        Pointer to the COBS encoder state machine context
  * @param dst         Destination buffer for encoded data
  * @param dst_len     On input: capacity of dst; On output: bytes written
+ * @param delimiter   Delimiter byte value (typically COBS_DEFAULT_DELIMITER)
  *
  * @retval 0        Success (encoder is reset)
  * @retval -ENOMEM  Insufficient destination space
  * @retval -EINVAL  Invalid parameters
  */
-int cobs_encode_finalize(struct cobs_encode_state *self, uint8_t *dst, size_t *dst_len);
+int cobs_encode_finalize(struct cobs_encode_state *self, uint8_t *dst, size_t *dst_len,
+			 uint8_t delimiter);
 
 /**
  * @brief Decode data using a COBS streaming state machine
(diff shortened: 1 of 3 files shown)


include/zephyr/data/cobs.h
```diff
@@ -86,8 +86,109 @@ int cobs_encode(struct net_buf *src, struct net_buf *dst, uint32_t flags);
...
 
-/** @} */
+/**
+ * @brief COBS streaming encoder state machine context
+ * 
+ * Encodes from a source buffer across multiple output buffer calls.
+ * Non-destructive: tracks read position without modifying source.
+ */
+struct cobs_encode_state {
+	struct net_buf *src_frag;   /* Current source fragment being read */
```
jukkar: I already commented about using Doxygen comments here in the other PR, same comment applies to this one.

mkschreder: Fixed.

lib/utils/cobs.c
```diff
@@ -1,104 +1,638 @@
...
+{
+	if (!self) {
+		return;
+	}
+
+	self->src_frag = NULL;
+	self->src_offset = 0;
+	self->block_code = 0;
+	self->block_pos = 0;
+}
```
pdgendt: `init` and `reset` are identical, keep one. The argument shouldn't be NULL (as the doxygen suggests, so change that to an assert.

```suggestion
	__ASSERT_NO_MSG(self != NULL);

	self->src_frag = NULL;
	self->src_offset = 0;
	self->block_code = 0;
	self->block_pos = 0;
}
```

lib/utils/cobs.c
```diff
@@ -1,104 +1,638 @@
...
+
+/* ========================================================================
+ * Source Navigation Helpers
+ * ======================================================================== */
+
+/* Skip empty fragments to find next available data */
+static inline void skip_empty_fragments(struct cobs_encode_state *self)
+{
+	while (self->src_frag && self->src_offset >= self->src_frag->len) {
```
pdgendt: No implicit boolean conversions (same for others)

```suggestion
	while (self->src_frag != NULL && self->src_offset >= self->src_frag->len) {
```

lib/utils/cobs.c
```diff
@@ -1,104 +1,638 @@
...
+/* ========================================================================
+ * Block-based Decoding Helpers
+ * ======================================================================== */
+
+/* Validate and remove trailing delimiter if present */
+static inline int validate_trailing_delimiter(struct net_buf *src, 
+					      const uint32_t flags,
+					      const uint8_t delimiter)
+{
+	if (!(flags & COBS_FLAG_TRAILING_DELIMITER)) {
```
pdgendt:
```suggestion
	if ((flags & COBS_FLAG_TRAILING_DELIMITER) == 0U) {
```

pdgendt (changes requested):
We shouldn't have (or need to maintain) different implementations for the streaming mechanism vs block based.

Ideally the `cobs_decode/cobs_encode` functions simply use the streaming with a stack allocated state.

The streaming decoder/encoders also don't have support for custom delimiters...

lib/utils/cobs.c
```diff
@@ -1,104 +1,638 @@
...
 	}
+	
+	return 0;
+}
+
+/* Handle delimiter found during encoding */
+static inline void handle_delimiter(uint8_t **code_ptr, uint8_t *code, 
+				    struct net_buf *dst)
+{
+	**code_ptr = *code;
```
pdgendt: Looking back at this, I'm starting to think the custom delimiter is faulty here, as there is no check/handling if the overhead byte itself would be equal to the custom delimiter.

For the default `0x00` delimiter the overhead byte is never zero and this is automatically true. I also didn't find a reference on how a custom delimiter should be implemented.

CC @hlord2000

mkschreder: I'm not sure either why there is a need for custom delimiter. I think 0x00 is obvious and robust

mkschreder: There is nothing anywhere about any kind of custom delimiter. I propose we remove that.

lib/utils/cobs.c
mkschreder: I think I fixed all of them.

lib/utils/cobs.c
mkschreder: Fixed as well

lib/utils/cobs.c
mkschreder: It's just a pain to maintain.

lib/utils/cobs.c
mkschreder: Solved

lib/utils/cobs.c
```diff
@@ -1,104 +1,625 @@
...
+ * @return 0 if valid (0x00), -EINVAL if potentially problematic
+ */
+static inline int validate_delimiter(uint8_t delimiter)
+{
+	if (delimiter != COBS_DEFAULT_DELIMITER) {
+		/* Custom delimiters (0x01-0xFF) can conflict with overhead bytes */
+		return -EINVAL;
+	}
+	return 0;
+}
```
pdgendt: Not sure we should take away the feature, it's still useful in some contexts. The [Wikipedia page](https://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing) also mentions that zero is the _typical_ value, but not the only.

For example this [rust library](https://docs.rs/cobs/latest/cobs/) has support for a custom delimiter or sentinel value, and I've updated [my own go library](https://github.com/pdgendt/cobs) to do the same, see https://github.com/pdgendt/cobs/pull/30

lib/utils/cobs.c
```diff
@@ -1,104 +1,625 @@
...
 #include <zephyr/data/cobs.h>
 
+/* ========================================================================
+ * Delimiter Validation
+ * ======================================================================== */
+
+/**
+ * @brief Validate delimiter value
+ *
+ * COBS overhead bytes range from 0x01 to 0xFF. Only 0x00 is guaranteed
```
hlord2000: there is no requirement for the delimiter to be zero, this is just what is commonly done. it can be any value.

lib/utils/cobs.c
mkschreder: Ok lets keep it.

lib/utils/cobs.c
mkschreder: Ok i will keep support for custom delimiter

PR #101005, opened 2025-12-14
drivers: net: cobs l2 network interface with async uart transport

mkschreder:
### Summary 
This PR adds a new way to run Zephyr IP networking over a simple TX/RX UART link using cobs for framing. Zephyr currently does not have a proper serial line network interface with multi instance support because the existing one uses uart pipe. 

The interface I'm adding in this PR uses predictable overhead COBS framing that is much better than slip and I'm also adding this as an interface of which you can have many instances. This interface is primarily meant for reliable board to board networking and I'm also adding throughput samples for both UDP and TCP so you can test how this works. 

**NOTE**: this is a big PR and still needs to be cleaned up. Also I'm still working on refactoring the samples and my goal is to make them bare yet robust. I also plan to update docs and describe in detail how this works.

- **A new COBS UART async network driver** that exposes a standard Zephyr network interface over UART.
- **A small COBS-based framing layer for serial links** so the UART link has clear packet boundaries.
- **New samples** to make bring-up and evaluation easy:
  - `samples/net/cobs_uart_async` (basic interface + net shell)
  - `samples/net/cobs_throughput_client` / `samples/net/cobs_throughput_server` (UDP echo-based throughput/latency)
  - `samples/net/cobs_throughput_single` and `samples/net/cobs_throughput_tcp` (single-board, cross-wired UART testing)
- **New tests** covering both “bytes on the wire” and “sockets on top”:
  - `tests/drivers/cobs_uart_async_loopback`
  - `tests/drivers/cobs_uart_async_sockets`
- **Documentation + devicetree binding** to make configuration straightforward.

### Why this is useful
- **Reliable serial networking**: enables IP connectivity on systems that only have a basic UART link available (TX/RX + GND), without requiring hardware flow control.
- **Easier evaluation and regression coverage**: provides ready-to-run samples for throughput testing and adds automated tests to keep the full stack (driver → network stack → BSD sockets) working over time.
- **Cleaner framing behavior**: COBS framing is both efficient and necessary in this case for packet based communication on a stream-based transport like UART.

### What’s included (high level)
- **Driver**: `drivers/net/cobs_uart_async.c` (+ Kconfig + build integration)
- **Serial framing support**: new COBS Serial L2 + net_pkt COBS helper APIs
- **Devicetree**: binding for declaring one or more COBS UART interfaces
- **Samples + tests**: end-to-end examples and validation coverage for real hardware and simulation scenarios

### Dependencies

- https://github.com/zephyrproject-rtos/zephyr/pull/101004: COBS library extensions.

### Testing
- Added unit test and a socket integration samples to validate end-to-end behavior over COBS framed packet link.


include/zephyr/data/cobs.h
```diff
@@ -86,8 +86,109 @@ int cobs_encode(struct net_buf *src, struct net_buf *dst, uint32_t flags);
...
 
-/** @} */
+/**
+ * @brief COBS streaming encoder state machine context
+ * 
+ * Encodes from a source buffer across multiple output buffer calls.
+ * Non-destructive: tracks read position without modifying source.
+ */
+struct cobs_encode_state {
+	struct net_buf *src_frag;   /* Current source fragment being read */
```
jukkar: Doxygen requires `/**< Current source fragment being read */` style comment in order it to generate the documentation properly.

doc/connectivity/networking/api/cobs_uart_async.rst
```diff
@@ -0,0 +1,210 @@
...
+
+   /* Get the SLIP interface */
+   iface = net_if_get_first_by_type(&NET_L2_GET_NAME(COBS_SERIAL));
+
+   /* Bring it up */
+   net_if_up(iface);
+
+   /* Configure IP address */
+   struct in_addr addr;
+   net_addr_pton(AF_INET, "192.168.7.1", &addr);
```
jukkar: Please use IP address range reserved for documentation, so here you could use `192.0.2.1` for example

drivers/net/slip_uart_async.c
```diff
@@ -0,0 +1,441 @@
...
+ *
+ * SLIP driver using UART async API with COBS framing.
+ * This driver provides a network interface over a UART using COBS encoding.
+ */
+
+#define LOG_MODULE_NAME slip_uart_async
+#define LOG_LEVEL CONFIG_SLIP_UART_ASYNC_LOG_LEVEL
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(LOG_MODULE_NAME);
```
jukkar: Instead of this, you could just write
```
LOG_MODULE_REGISTER(slip_uart_async, CONFIG_SLIP_UART_ASYNC_LOG_LEVEL);
```
which is the style we are mainly using in zephyr.

drivers/net/slip_uart_async.c
```diff
@@ -0,0 +1,441 @@
...
+{
+	int ret;
+	static int scan_count = 0;
+
+	switch (ctx->rx_state) {
+	case STATE_SCANNING_FOR_FRAME:
+		scan_count++;
+		if (byte == COBS_DELIMITER) {
+			/* Start of frame */
+			LOG_INF("✓ Found COBS delimiter (0x00), starting frame (scanned %d bytes)", scan_count);
```
jukkar: I am not sure if it is a good idea to print non ascii chars here, it does not give much benefit and might cause some garbage to be printed.
Also this looks like a debug print than an info message. Usually there is no need to print this kind of messages unless you are debugging things.

drivers/net/cobs_uart_async.c
```diff
@@ -0,0 +1,441 @@
...
+	ctx->init_done = true;
+	ctx->iface = iface;
+
+	/* Generate MAC address (00-00-5E-00-53-xx per RFC 7042) */
+	ctx->mac_addr[0] = 0x00;
+	ctx->mac_addr[1] = 0x00;
+	ctx->mac_addr[2] = 0x5E;
+	ctx->mac_addr[3] = 0x00;
+	ctx->mac_addr[4] = 0x53;
+	ctx->mac_addr[5] = sys_rand8_get();
```
jukkar, line 483: I am not sure we should use this MAC address space here, the values specified in the RFC are meant for documentation but as this is a driver that might be used in real products, then this is perhaps not a good thing. Could we specify that mac address in DT?

drivers/net/cobs_uart_async.c
```diff
@@ -0,0 +1,441 @@
...
+	uint8_t tx_buf[UART_TX_BUF_LEN];
+	struct k_sem tx_sem;
+
+	/* Worker for processing received data */
+	struct k_work rx_work;
+	struct k_work_q rx_workq;
+	K_KERNEL_STACK_MEMBER(rx_stack, CONFIG_SLIP_UART_ASYNC_RX_STACK_SIZE);
+
+	/* MAC address */
+	uint8_t mac_addr[6];
```
jukkar, line 74: There is `NET_ETH_ADDR_LEN` that could be used here

samples/net/cobs_throughput_client/boards/native_sim.overlay
```diff
@@ -0,0 +1,17 @@
+/*
```
jukkar: About the samples, could you please place them in a separate commit. Also we could create `samples/net/slip` and place the related samples there.

samples/net/cobs_throughput_client/src/main.c
```diff
@@ -0,0 +1,439 @@
...
+#define CONFIG_SLIP_THROUGHPUT_PACKET_SIZE 512
+#endif
+
+#ifndef CONFIG_SLIP_THROUGHPUT_SERVER_IP
+#define CONFIG_SLIP_THROUGHPUT_SERVER_IP "192.168.7.1"
+#endif
+
+#ifndef CONFIG_SLIP_THROUGHPUT_DURATION_SEC
+#define CONFIG_SLIP_THROUGHPUT_DURATION_SEC 60
+#endif
```
jukkar: I don't recommend re-defining the config variables but do it like this instead
```
#if defined(CONFIG_SLIP_THROUGHPUT_PACKET_SIZE)
#define SLIP_THROUGHPUT_PACKET_SIZE CONFIG_SLIP_THROUGHPUT_PACKET_SIZE
#else
#define SLIP_THROUGHPUT_PACKET_SIZE 512
#endif
```

Why the Kconfig variable would be undefined, why do we even need to check this?

samples/net/cobs_throughput_client/src/main.c
```diff
@@ -0,0 +1,439 @@
...
+	if (sock < 0) {
+		LOG_ERR("Failed to create socket: %d", errno);
+		test_params.running = false;
+		return -errno;
+	}
+
+	/* Set receive timeout */
+	timeout.tv_sec = 1;
+	timeout.tv_usec = 0;
+	ret = zsock_setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout));
```
jukkar: Please check the code and do not mix POSIX and network namespaced symbols. We recently merged a PR that adds a prefix to various network symbols in order to avoid conflicts with POSIX.
So here use `ZSOCK_SOL_SOCKET` and `ZSOCK_SO_RCVTIMEO`
If you want to check the usage of the correct APIs, you can set `CONFIG_NET_NAMESPACE_COMPAT_MODE=n` to disable the compatibility mode and make sure correct symbols are used.

samples/net/cobs_throughput_client/Kconfig
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2024 Nordic Semiconductor ASA
+# SPDX-License-Identifier: Apache-2.0
+
+mainmenu "SLIP Throughput Client Configuration"
+
+source "Kconfig.zephyr"
+
+config SLIP_THROUGHPUT_PACKET_SIZE
```
jukkar: We have been trying to namespace the network sample specific Kconfig symbols so that it is clear where they come from. So the symbols should be called
```
config NET_SAMPLE_SLIP_THROUGHPUT_PACKET_SIZE
```
etc

samples/net/cobs_throughput_client/prj.conf
```diff
@@ -0,0 +1,46 @@
...
+CONFIG_NET_STATISTICS=y
+CONFIG_NET_STATISTICS_USER_API=y
+
+# Random number generator (needed by networking stack)
+CONFIG_ENTROPY_GENERATOR=y
+CONFIG_TEST_RANDOM_GENERATOR=y
+
+# Packet size configuration (can be overridden)
+CONFIG_SLIP_THROUGHPUT_PACKET_SIZE=512
+CONFIG_SLIP_THROUGHPUT_SERVER_IP="192.168.7.1"
```
jukkar: Please use the IP address space we use in the configs by default. So the address in this case should be `192.0.2.1`
User can always change this in command line when compiling the application if needed.

samples/net/cobs_throughput_client/README.rst
```diff
@@ -0,0 +1,156 @@
...
+           status = "okay";
+       };
+   };
+
+The client automatically configures the SLIP interface with:
+- IPv4: 192.168.7.2/24
+- IPv6: fe80::2
+
+Default test parameters (can be overridden via Kconfig or shell):
+- Server IP: 192.168.7.1
```
jukkar: Please change this

jukkar (commented):
Nice work! There is a lot of code and I will continue the review later, this was just a first glance of the PR.
Couple of thoughts about this:
* It might be confusing if this is called SLIP as it typically means certain framing for the packets as seen in https://en.wikipedia.org/wiki/Serial_Line_Internet_Protocol. Should we mention the COBS there or something like that?
* Would it be possible to create a Linux counter part of the protocol so that this could be used from qemu or even from native_sim? So similar way the SLIP can be used.

include/zephyr/data/cobs.h
mkschreder: I will update all comments to follow doxygen format

doc/connectivity/networking/api/cobs_uart_async.rst
mkschreder: Sure. I think I will update them to be ipv6 addresses though.

samples/net/cobs_throughput_client/src/main.c
mkschreder: These samples will be cleaned up as I mentioned in the note in PR description.

include/zephyr/data/cobs.h
mkschreder: This will disappear once the cobs MR is merged.

drivers/net/slip_uart_async.c
mkschreder: I think this is gone now.

dts/bindings/net/zephyr,cobs-uart-async.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+    &uart1 {
+      status = "okay";
+      current-speed = <115200>;
+    };
+
+    &uart2 {
+      status = "okay";
+      current-speed = <115200>;
+    };
```
rruuaanng, line 35: You can use `examples` keyword, and place those examples there.
```
examples:
  - |
    / {
       ....
    };
```

dts/bindings/net/zephyr,cobs-uart-async.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2025 Martin Schröder
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  COBS network interface using UART async API with COBS framing.
+
+  This binding creates a network interface instance that uses a UART
```
rruuaanng, line 7: This is well known, so it's best to remove explanations such as the `this binding`.

dts/bindings/net/zephyr,cobs-uart-async.yaml
```diff
@@ -0,0 +1,57 @@
...
+  device for transport. Multiple instances can be created, each using
+  a different UART.
```
rruuaanng, line 9:
```suggestion
title: COBS network interface using UART async API with COBS framing

description: |
  This binding creates a network interface instance that uses a UART
  device for transport. Multiple instances can be created, each using
  a different UART.
```

dts/bindings/net/zephyr,cobs-uart-async.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+properties:
+  reg:
+    type: int
+    required: true
+    description: |
+      Instance identifier. Must match the unit address.
+      Parent node must have #size-cells = <0>.
+
+  uart:
```
rruuaanng, line 47:
```suggestion
  cobs-uart:
```
Would this be better? :-----)

PR #101009, opened 2025-12-14
doc: clarified description of k_uptime_delta

### LINKED ISSUE #100697: Improve k_uptime_delta() documentation
### Summary

In k_uptime_delta(int64 *reftime), the documented functionality is the following:
`This routine computes the elapsed time between the current system uptime and an earlier reference time, in milliseconds.`

However, if you don't read the information on the parameter, you will miss the following note:
`Pointer to a reference time, **which is updated to the current uptime upon return**.`

Adjust the description of the function to include that the pointer is changed on function call. Such as:

`This routine computes the elapsed time between the current system uptime and an earlier reference time, in milliseconds, then updates the reference time.`

### Alternatives

_No response_

### Additional Context

_No response_

nicbra:
Description added in docs, and updated the header to clarify that the function modifies the argument.

Resolves #100697

### DIFF
--- a/doc/kernel/services/timing/clocks.rst
+++ b/doc/kernel/services/timing/clocks.rst
@@ -81,6 +81,11 @@ Apps with precise timing requirements (that are willing to do their
 own conversions to portable real time units) may access this with
 :c:func:`k_uptime_ticks`.
 
+:c:func:`k_uptime_delta` can be used to get the time elapsed between a reference
+and the current time. The referenced time will be updated to the current uptime,
+to easily calculate the next elapsed time.
+
+
 Timeouts
 ========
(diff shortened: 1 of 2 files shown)


peter-mitsis:
I like the content of the proposed change. Once the compliance check failure is fixed, I will be happy to give this a +1.

nicbra:
> I like the content of the proposed change. Once the compliance check failure is fixed, I will be happy to give this a +1.

It should now pass the compliance check

peter-mitsis (approved):

ceolin (approved):

aescolar (approved):

PR #95362, opened 2025-09-03
lora: optional packet CRC and `lora_airtime`

JordanYates:
Add the option to disable the builtin 16 bit CRC on the LoRa payload. The conditional is inverted (`packet_crc_disable` vs `packet_crc_enable`) to maintain the behavior of current applications that do not assign to the variable.

Add a simple function that exposes the airtime of a packet of a given length.

PR #101011, opened 2025-12-14
drivers: lora: ability to set custom sync word for LBM

VynDragon:
Change way the sync word is set for LoraBasicModem so as to permit using a custom sync word/byte
Modify samples in consequence, reduce default TX output for safety in case of mistake, set SF default to something LLCC68 will accept.

Article about the sync word compatibility: https://blog.classycode.com/lora-sync-word-compatibility-between-sx127x-and-sx126x-460324d1787a
otherwise looked through lbm to see what it did with them as you would expect.

cross checked with 433mhz sx1278 and llcc68.

### DIFF
--- a/drivers/lora/lora_basics_modem/lbm_common.c
+++ b/drivers/lora/lora_basics_modem/lbm_common.c
@@ -84,8 +84,7 @@ int lbm_lora_config(const struct device *dev, struct lora_modem_config *lora_con
 		},
 		.rf_freq_in_hz = lora_config->frequency,
 		.output_pwr_in_dbm = lora_config->tx_power,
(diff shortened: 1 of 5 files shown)


JordanYates (commented):
I think this makes the transition from LoRaMAC node to LBM harder than it needs to be, while also making "standard" LBM usage more difficult. `.public` works fine for selecting between the two most common SYNC words, and changing the config struct for different backends makes user code (and samples) less ideal.

I actually already implemented this in my fork (https://github.com/Embeint/zephyr/commit/4b1678f94c55d451af3689ddd5d1b26775cd86d8), but was waiting for some other PRs to be merged before submitting it (#99884, #95362)

VynDragon:
Well I considered implementing it exactly like in your code, but decided to go this way instead which made more sense to me.

PR #101022, opened 2025-12-15
kernel: only use _k_neg_eagain in test

TaiJuWu:
Replace the _k_neg_eagain global variable with assembly-safe errno macros to save memory and improve performance.

Previously, errno.h could not be included in assembly files due to C-specific syntax (function declarations, typedefs). This required maintaining a global variable _k_neg_eagain, which wasted 4 bytes and required memory access instructions.

### DIFF
--- a/kernel/errno.c
+++ b/kernel/errno.c
@@ -18,9 +18,10 @@
 /*
  * Define _k_neg_eagain for use in assembly files as errno.h is
  * not assembly language safe.
- * FIXME: wastes 4 bytes
  */
+#if defined(CONFIG_TEST)
 const int _k_neg_eagain = -EAGAIN;
+#endif
 
 #ifdef CONFIG_ERRNO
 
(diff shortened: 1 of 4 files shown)


kernel/errno.c
```diff
@@ -18,9 +18,10 @@
 /*
  * Define _k_neg_eagain for use in assembly files as errno.h is
  * not assembly language safe.
- * FIXME: wastes 4 bytes
  */
+#if defined(CONFIG_TEST)
```
edersondisouza, line 22: On the commit message you mention "update test test_write_kernro to use system_handler to do
read only modification test", but here you keep `_k_neg_eagain` for tests - am I missing something here?

edersondisouza (commented):
This indeed solves the "wastes 4 bytes" issue, but considering that only x86 (32) and mips use `_k_neg_eagain`, they seem to be the odd one out. Have you considered changing them to _not_ use `_k_neg_eagain` to start with?

PR #101024, opened 2025-12-15
Fix pwm ctimer prescaler error

### LINKED ISSUE #101000: mcux_ctimer_pwm_get_cycles_per_sec: Incorrect cycles vale if prescaler is > 0
### Describe the bug

Hi.
In the `pwm_mcux_ctimer.c` driver, the function `mcux_ctimer_pwm_get_cycles_per_sec` incorrectly calcualtes the `cyles` value if the `prescaler` (form the DT) if not equal to `0`.

The problem is here:
`if (config->prescale > 0) {
	*cycles /= config->prescale;
}`

The prescaler is implemented as a counter, therefore its division ratio is equal to its top value (the prescale value form the DT) + 1.

The simplest fix is to  remove the `if` and modify the calculation: `*cycles /= config->prescale + 1;`

I'm attaching a patch file for this fix.

Its not super critical, since its a 32bit counter almost nobody uses the prescaler functionality.

[pwm_mcux_ctimer.patch](https://github.com/user-attachments/files/24149772/pwm_mcux_ctimer.patch)

### Regression

- [ ] This is a regression.

### Steps to reproduce

Make use of the pwm_mcux_ctimer driver. In the device tree set the prescaler value to != 0, eg. 1:

> &ctimer0 {
	status = "okay";
	compatible = "nxp,ctimer-pwm";
 	#pwm-cells = <3>;
 	pinctrl-0 = <&pinmux_ctimer0_pwm>;
 	pinctrl-names = "default";
	prescaler = <1>;
};

When called, the `mcux_ctimer_pwm_get_cycles_per_sec` will give the result twice as large.

### Relevant log output

```shell

```

### Impact

### Environment

Tested on NXP's FRDM-MCXA156 board, Zephyr 4.3.0, SDK 0.16.1.

### Additional Context

_No response_

FelixWang47831:
The prescaler divides the clock by (prescaler + 1), not by the prescaler value directly. Update the calculation to correctly account for this.

### DIFF
--- a/drivers/pwm/pwm_mcux_ctimer.c
+++ b/drivers/pwm/pwm_mcux_ctimer.c
@@ -1,5 +1,6 @@
 /*
  * (c) Meta Platforms, Inc. and affiliates.
+ * Copyright 2025 NXP
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -206,9 +207,9 @@ static int mcux_ctimer_pwm_get_cycles_per_sec(const struct device *dev, uint32_t
 		return err;
 	}
 
-	if (config->prescale > 0) {
-		*cycles /= config->prescale;
-	}
+	*cycles /= (uint64_t)config->prescale + 1;
+	__ASSERT((*cycles) > 0, "Invalid PWM frequency: cycles per second is 0(check clock rate
+		and prescaler)");
 
 	return err;
 }
--- a/dts/bindings/pwm/nxp,ctimer-pwm.yaml
+++ b/dts/bindings/pwm/nxp,ctimer-pwm.yaml
@@ -14,7 +14,9 @@ properties:
   prescaler:
     type: int
     default: 1
-    description: prescaling value
+    description: |
+      Specifies the prescale value. Valid range: 0x0 - 0xFFFFFFFF.
+      The clock is divided by (prescaler + 1).
 
   clk-source:
     type: int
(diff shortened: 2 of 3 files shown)


drivers/pwm/pwm_mcux_ctimer.c
```diff
@@ -206,9 +207,9 @@ static int mcux_ctimer_pwm_get_cycles_per_sec(const struct device *dev, uint32_t
 		return err;
 	}
 
-	if (config->prescale > 0) {
-		*cycles /= config->prescale;
-	}
+	*cycles /= config->prescale + 1;
```
Holt-Sun: Technically, shall do overflow check for prescale + 1.

dts/bindings/pwm/nxp,ctimer-pwm.yaml
```diff
@@ -14,7 +14,9 @@ properties:
   prescaler:
     type: int
     default: 1
-    description: prescaling value
+    description: |
+      Specifies the prescale value. Valid range: 0x0 - 0xFFFFFFFF.
```
Holt-Sun, line 18: Could it be 0xffffffff which will cause overflow for prescaler + 1?

Holt-Sun (changes requested):

dts/bindings/pwm/nxp,ctimer-pwm.yaml
FelixWang47831: No, it's set directly 
```
	pwm_config.prescale = config->prescale;

	CTIMER_Init(config->base, &pwm_config);
```
And PRIVAL is a 32 bits register, so the range is 0x0 - 0xFFFFFFFF

image

Holt-Sun (approved):

henrikbrixandersen (approved):

mmahadevan108 (approved):

henrikbrixandersen:
Retriggered CI.

PR #101027, opened 2025-12-15
support nxp wifi with custom host platform

fengming-ye:
manifest: update hal_nxp revision to support custom host platform for WiFi
    
    Add CONFIG_NXP_WIFI_CUSTOM_HOST to customize host platform for
    NXP Wi-Fi card.

wifi: nxp: support nxp wifi with custom host platform
    
    Add CONFIG_NXP_WIFI_CUSTOM_HOST to build with custom
    host platform.

### DIFF
--- a/drivers/wifi/nxp/Kconfig.nxp
+++ b/drivers/wifi/nxp/Kconfig.nxp
@@ -26,6 +26,22 @@ config HEAP_MEM_POOL_ADD_SIZE_NXP_WIFI
 config WIFI_MGMT_SCAN_CHAN_MAX_MANUAL
 	default 50
 
+config NXP_MONOLITHIC_WIFI
+	bool "WiFi firmware monolithic build"
+	default y
+	depends on !NXP_WIFI_CUSTOM
+	help
+	   If enabled, the WiFi firmware used by the device will be linked with the
+	   application directly.
+
+config NXP_WIFI_CUSTOM_HOST
+	bool "Custom host platform for NXP Wi-Fi"
+	default y if BOARD_NATIVE_SIM
+	default y if !HAS_MCUX && !HAS_IMX_HAL && !HAS_NXP_S32_HAL
+	depends on !NXP_RW610
+	help
+	  Customize host platform eg. non NXP host platform for NXP Wi-Fi.
+
 config NXP_WIFI_CUSTOM
 	bool "Custom NXP Wi-Fi part"
 	help
@@ -830,8 +846,8 @@ config NXP_WIFI_CSI_AMI
 config NXP_WIFI_RESET
 	bool "Wi-Fi reset"
 	default y
-	imply NXP_WIFI_IND_DNLD if NXP_IW610
-	imply NXP_WIFI_IND_RESET if NXP_IW610
+	imply NXP_WIFI_IND_DNLD if NXP_IW610 && !NXP_WIFI_CUSTOM_HOST
+	imply NXP_WIFI_IND_RESET if NXP_IW610 && !NXP_WIFI_CUSTOM_HOST
 	help
 	  This option is used to enable/disable/reset Wi-Fi.
 
--- a/drivers/wifi/nxp/nxp_wifi_drv.c
+++ b/drivers/wifi/nxp/nxp_wifi_drv.c
@@ -111,6 +111,7 @@ int nxp_wifi_wlan_event_callback(enum wlan_event_reason reason, void *data)
 	struct in_addr base_addr;
 	struct in_addr netmask_addr;
 	struct wifi_ap_sta_info ap_sta_info = { 0 };
+	sta_node *con_sta_info;
 #endif
 
 	LOG_DBG("WLAN: received event %d", reason);
@@ -293,7 +294,7 @@ int nxp_wifi_wlan_event_callback(enum wlan_event_reason reason, void *data)
 #endif
 		break;
 	case WLAN_REASON_UAP_CLIENT_ASSOC:
-		sta_node *con_sta_info = (sta_node *)data;
+		con_sta_info = (sta_node *)data;
 
 		if (con_sta_info->is_11n_enabled) {
 			ap_sta_info.link_mode = WIFI_4;
--- a/modules/hal_nxp/wifi/CMakeLists.txt
+++ b/modules/hal_nxp/wifi/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+add_subdirectory(
+  ${ZEPHYR_CURRENT_MODULE_DIR}/mcux/middleware/wifi_nxp
+  ${CMAKE_CURRENT_BINARY_DIR}/middleware/wifi_nxp
+)
+
+add_subdirectory(
+  ${ZEPHYR_CURRENT_MODULE_DIR}/mcux/mcux-sdk-ng/components/osa
+  ${CMAKE_CURRENT_BINARY_DIR}/components/osa
+)
+
+add_subdirectory(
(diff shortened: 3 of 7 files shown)


modules/hal_nxp/wifi/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright 2025-2026 NXP
```
ZhaoxiangJin: 2025 is enough?

modules/hal_nxp/wifi/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+add_subdirectory(
+  ${ZEPHYR_CURRENT_MODULE_DIR}/mcux/middleware/wifi_nxp
+  ${CMAKE_CURRENT_BINARY_DIR}/middleware/wifi_nxp
+)
```
ZhaoxiangJin, line 7: Use empty lines to separate logically distinct sections within a CMake file, reference: https://docs.zephyrproject.org/latest/contribute/style/cmake.html#cmake-style

modules/hal_nxp/wifi/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+add_subdirectory(
+  ${ZEPHYR_CURRENT_MODULE_DIR}/mcux/mcux-sdk-ng/components/osa
+  ${CMAKE_CURRENT_BINARY_DIR}/components/osa
+)
+add_subdirectory(
+  ${ZEPHYR_CURRENT_MODULE_DIR}/mcux/mcux-sdk-ng/components/wifi_bt_module
+  ${CMAKE_CURRENT_BINARY_DIR}/components/wifi_bt_module
+)
+
+zephyr_compile_definitions_ifdef(CONFIG_NXP_WIFI_CUSTOM_HOST
```
ZhaoxiangJin, line 19: use `zephyr_library_compile_definitions_ifdef `instead

west.yml
ZhaoxiangJin, line 1: This commit (manifest change) should first

modules/hal_nxp/wifi/CMakeLists.txt
fengming-ye: yes updated

modules/hal_nxp/wifi/CMakeLists.txt
fengming-ye: updated

modules/hal_nxp/wifi/CMakeLists.txt
fengming-ye: driver/wifi/nxp lib also needs this macro

west.yml
fengming-ye: ok updated

drivers/wifi/nxp/nxp_wifi_drv.c
dhruvkakadiya, line 1: Within this file, another patch is required too since `nxp_wifi_internal_tx()` API has been changed.
```
@@ -1945,7 +1946,7 @@ static NXP_WIFI_SET_FUNC_ATTR int nxp_wifi_send(const struct device *dev, struct
 #endif
 
        /* Enqueue packet for transmission */
-       if (nxp_wifi_internal_tx(dev, pkt) != WM_SUCCESS) {
+       if (nxp_wifi_internal_tx(dev, pkt, 1) != WM_SUCCESS) {
                goto out;
        }
```

drivers/wifi/nxp/Kconfig.nxp
dhruvkakadiya, line 1: Within this file another patch is required if let say one is trying to use IW610 module instead RW610. Maybe more modules can be added as dependency if relevant [wlcmgr_mon_task()](https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/middleware/wifi_nxp/wlcmgr/wlan.c#L445) needs to be supported in other cases otherwise, we can get into compilation errors.
```
--- a/drivers/wifi/nxp/Kconfig.nxp
+++ b/drivers/wifi/nxp/Kconfig.nxp
@@ -418,7 +418,7 @@ menu "Wi-Fi driver Stack configurations"
 
 config NXP_WIFI_MON_TASK_STACK_SIZE
        int "Mon thread stack size"
-       depends on NXP_RW610
+       depends on NXP_RW610 || NXP_IW610
        default 3072
        help
          This option specifies the size of the stack used by the mon task.
@@ -460,7 +460,7 @@ menu "Wi-Fi thread priority configurations"
 
 config NXP_WIFI_MON_TASK_PRIO
        int "Mon task priority"
-       depends on NXP_RW610
+       depends on NXP_RW610 || NXP_IW610
        default 4
        help
          This option specifies the priority of the mon task.
```

dhruvkakadiya: Also another patch would be good to add is below, such that when `NXP_WIFI_CUSTOM_HOST` is enabled then it doesn't end up turning on `NXP_WIFI_IND_DNLD` and `NXP_WIFI_IND_RESET` features as on custom host relevant `fsl_gpio` drivers won't be found and it will end up into compilation issues.
```
@@ -827,8 +827,8 @@ config NXP_WIFI_CSI
 config NXP_WIFI_RESET
        bool "Wi-Fi reset"
        default y
-       imply NXP_WIFI_IND_DNLD if NXP_IW610
-       imply NXP_WIFI_IND_RESET if NXP_IW610
+       imply NXP_WIFI_IND_DNLD if NXP_IW610 && !NXP_WIFI_CUSTOM_HOST
+       imply NXP_WIFI_IND_RESET if NXP_IW610 && !NXP_WIFI_CUSTOM_HOST
        help
          This option is used to enable/disable/reset Wi-Fi.
 ```

That brings up another question that, if we would like to have those features enabled in `NXP_WIFI_CUSTOM_HOST` case then seems like more changes required at `hal_nxp` layer too?

drivers/wifi/nxp/nxp_wifi_drv.c
fengming-ye: current code base is
if (nxp_wifi_internal_tx(dev, pkt, 0) != WM_SUCCESS) {
maybe you have a revision mismatch with hal_nxp and zephyr. In that case you can update either module.

drivers/wifi/nxp/Kconfig.nxp
fengming-ye: NXP_WIFI_MON_TASK_PRIO and NXP_WIFI_MON_TASK_STACK_SIZE have matched dependency in latest code base
depends on NXP_RW610 || NXP_IW610 || NXP_IW61X || NXP_IW416.
So you can update zephyr revision to get these changes.

fengming-ye: updated to fix NXP_WIFI_RESET dependency.
And yes, we need more investigation to support NXP_WIFI_IND_DNLD and NXP_WIFI_IND_RESET on non NXP host platform.

dhruvkakadiya (approved):
Changes in this PR, allowing to have successful build when using IW610 module with custom non-NXP chipset, though can't verify actual functionality yet.

It would be good to add following things.
1. Sample example documentation where it's been documented that how to use IW610 module with non-NXP host, this would help documenting required minimal Kconfig settings
2. NXP_WIFI_IND_DNLD and NXP_WIFI_IND_RESET features working in case of NXP_WIFI_CUSTOM_HOST
3. CI perspective, build/compilation gets exercised when NXP_WIFI_CUSTOM_HOST is enabled which help ensuring proper isolation from MCUX toolchain dependencies which could easily sneak back in for long term stable solution

MaochenWang1 (approved):

ZhaoxiangJin (approved):

PR #100143, opened 2025-11-27
modules: mbedtls: improve mbedtls testing

valeriosetti:
Most of the Mbed TLS testing currently present in Zephyr has been introduced several years ago by copying what was present back then in Mbed TLS and it hasn't been much updated over time. The result is that all of this testing relies on legacy crypto, some of which has also been removed from upstream Mbed TLS.

This PR tries to modernize testing by removing usage of legacy crypto and testing only PSA Crypto API. This will be the only available option starting from Mbed TLS 4.0/TF-PSA-Crypto 1.0.

PR #101033, opened 2025-12-15
tests: crypto: mbedtls_psa: enable malloc arena for Mbed TLS

M-Moawad:
The PSA Crypto implementation in Mbed TLS requires some heap to operate.
Provide a small malloc arena to ensure this test remains compatible with `MINIMAL_LIBC` configurations
where the arena is disabled by default.

The test added recently in this PR: https://github.com/zephyrproject-rtos/zephyr/pull/100143
The fix is similar to: https://github.com/zephyrproject-rtos/zephyr/pull/82099


tomi-font (approved):

ruuddw (approved):
Should this be moved to the mbedtls module configuration, instead of repeating it for individual tests? If it's a generic requirement for using mbedtls with minimal libc, and not related to specific tests that would be a better place to avoid more failing apps and tests in the future.

ceolin (approved):

PR #101035, opened 2025-12-15
drivers: flash: flexspi_nor: Add RDSR2 command to quad enable LUT

### LINKED ISSUE #101032: drivers: flash: flash_mcux_flexspi_nor.c: FlexSPI NOR Quad Enable Bug Triggers CMP Bit
### Describe the bug

When enabling quad mode for the internal Winbond FlexSPI NOR flash on some MIMXRT1064 (and potentially other) chips, the Complement Protection Bit (CMP) in the flash status register is inadvertently set. This enables full-chip write protection, blocking re-programming—even via the serial downloader.
Recovery requires a custom RAM program with a specialized LUT to reconfigure the flash status registers to a valid state.

Root Cause
In flash_mcux_flexspi_nor.c, the flash_flexspi_nor_quad_enable() function contains flawed logic for handling JESD216_DW15_QER_VAL_S2B1v4:
The code attempts a read-modify-write of Status1 and Status2 registers by reading 2 bytes (assuming reading past Status1’s 1-byte boundary automatically reads Status2) and writing back 2 bytes.
This violates Winbond’s flash specification (documented in Figure 8 of the part’s manual): Reading >1 byte from any status register repeats the same register’s value (instead of advancing to the next status register).
In contrast, writing >1 byte does advance to subsequent status registers (Figure 9b).
The incorrect read captures invalid data that overlaps with Status2’s CMP bit. When written back, this incorrectly sets the CMP bit, enabling full-chip write protection.

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

Albort12138:
Fixes: #101032

When qer is S2B1v1 or S2B1v4, reading Status Register 1 and Status Register 2 requires different opcodes, and writing Status Register 1/2 requires two bytes in a single write. Add the Read Status Register 2 command sequence to the LUT used by the quad-enable function, and ensure the write operation sends two bytes (SR1|SR2) when setting the QE bit.

### DIFF
--- a/drivers/flash/flash_mcux_flexspi_nor.c
+++ b/drivers/flash/flash_mcux_flexspi_nor.c
@@ -583,6 +583,15 @@ static int flash_flexspi_nor_quad_enable(struct flash_flexspi_nor_data *data,
 		flexspi_lut[SCRATCH_CMD][0] = FLEXSPI_LUT_SEQ(
 				kFLEXSPI_Command_SDR, kFLEXSPI_1PAD, SPI_NOR_CMD_RDSR,
 				kFLEXSPI_Command_READ_SDR, kFLEXSPI_1PAD, 0x1);
+		flexspi_lut[SCRATCH_CMD][1] = FLEXSPI_LUT_SEQ(
+				kFLEXSPI_Command_JUMP_ON_CS, kFLEXSPI_1PAD, 0x2,
+				kFLEXSPI_Command_JUMP_ON_CS, kFLEXSPI_1PAD, 0x2);
+		flexspi_lut[SCRATCH_CMD][2] = FLEXSPI_LUT_SEQ(
+				kFLEXSPI_Command_SDR, kFLEXSPI_1PAD, SPI_NOR_CMD_RDSR2,
+				kFLEXSPI_Command_READ_SDR, kFLEXSPI_1PAD, 0x1);
+		flexspi_lut[SCRATCH_CMD][3] = FLEXSPI_LUT_SEQ(
+				kFLEXSPI_Command_STOP, kFLEXSPI_1PAD, 0x0,
+				kFLEXSPI_Command_STOP, kFLEXSPI_1PAD, 0x0);
 		flexspi_lut[SCRATCH_CMD2][0] = FLEXSPI_LUT_SEQ(
 				kFLEXSPI_Command_SDR, kFLEXSPI_1PAD, SPI_NOR_CMD_WRSR,
 				kFLEXSPI_Command_WRITE_SDR, kFLEXSPI_1PAD, 0x1);


nmGit:
Hi @Albort12138,
It looks like STOP commands were added post SPI_NOR_CMD_RDSR, before RDSR2. According to the IMX reference manual, a stop command ends the execution of the entire sequence, resetting the sequence engine instruction pointer to 0. Wouldn't this mean that the sequence stops before RDSR2 is ever read?

nmGit:
> Hi @Albort12138, It looks like STOP commands were added post SPI_NOR_CMD_RDSR, before RDSR2. According to the IMX reference manual, a stop command ends the execution of the entire sequence, resetting the sequence engine instruction pointer to 0. Wouldn't this mean that the sequence stops before RDSR2 is ever read?

@Albort12138, I tested the implemented in this PR and it unfortunately doesn't seem to fix the issue.

In a similar vein, though, using the following patched code does fix the issue:
```
...	
case JESD216_DW15_QER_VAL_S2B1v4:
		/* Install read and write status command */
		flexspi_lut[SCRATCH_CMD][0] = FLEXSPI_LUT_SEQ(
				kFLEXSPI_Command_SDR, kFLEXSPI_1PAD, SPI_NOR_CMD_RDSR2,
				kFLEXSPI_Command_READ_SDR, kFLEXSPI_1PAD, 0x1);
		flexspi_lut[SCRATCH_CMD2][0] = FLEXSPI_LUT_SEQ(
				kFLEXSPI_Command_SDR, kFLEXSPI_1PAD, SPI_NOR_CMD_WRSR2,
				kFLEXSPI_Command_WRITE_SDR, kFLEXSPI_1PAD, 0x1);

		/* Set bit 1 of status register 2 */
		bit = BIT(1); 
		rd_size = 1;
		wr_size = 1;
		break;
...
```
The concern was brought up that this method might not adhere to the SFDP standard, though.

hakehuang:
> Hi @hakehuang, please help to verify test cases regarding of flash: flexspi_nor, especially on mimxrt1064_evk board.

@Albort12138 , I do not have mimxrt1064 boards but I find mimxrt1024_evk use the same spi-nor flash, and regression test with your pr seems not regression, but @nmGit 's comments needs your attention

Albort12138:
> Hi @Albort12138, It looks like STOP commands were added post SPI_NOR_CMD_RDSR, before RDSR2. According to the IMX reference manual, a stop command ends the execution of the entire sequence, resetting the sequence engine instruction pointer to 0. Wouldn't this mean that the sequence stops before RDSR2 is ever read?

Sorry, my mistake — it should be JUMP_ON_CS. I’m trying to keep the changes minimal, so I attempted to execute two different read commands within the same sequence.
I’ve already updated the PR. If you have time, please help test it again. Apologies for any inconvenience caused. I’m also working on finding a board that can reproduce the issue you encountered.

For the details of QER, in short in case of 0x1, write 1 byte will clear SR2, please take look of the snapshot. 
image

zejiang0jason (approved):

Holt-Sun (approved):

ofirshe:
We also have a custom board with an RT1061 and two Winbond NOR flashes, and we’ve encountered many cases where the CMP protection bit gets set. Since I didn’t have time to fully understand and debug the root cause, I added a Zephyr patch that always clears this bit at boot.
[0002-flexspi-clear-cmp-bit.patch](https://github.com/user-attachments/files/24332549/0002-flexspi-clear-cmp-bit.patch)
@Albort12138, are you sure your fix will completely eliminate the issue? What happens in the case of a power failure in the middle of an ongoing transaction, where the system could enter a boot loop? These are the main scenarios in which I’ve seen this bit being set.
I’m very concerned because this happened once in a customer’s field unit, and there was no way to upgrade the firmware. The only workaround was to load a program into RAM that clears this bit. This is a real nightmare scenario.
Also, could you please backport the fix to Zephyr 4.3?

Thanks

Albort12138:
> Also, could you please backport the fix to Zephyr 4.3?

Yes, I will

Albort12138:
> @Albort12138, are you sure your fix will completely eliminate the issue? What happens in the case of a power failure in the middle of an ongoing transaction, where the system could enter a boot loop? These are the main scenarios in which I’ve seen this bit being set.
> I’m very concerned because this happened once in a customer’s field unit, and there was no way to upgrade the firmware. The only workaround was to load a program into RAM that clears this bit. This is a real nightmare scenario.

For case JESD216_DW15_QER_VAL_S2B1v1 and JESD216_DW15_QER_VAL_S2B1v4, current patch will completely fix the issue

Albort12138:
> I noticed that this driver puts sets the write enable flag on the flash and leaves it set. Perhaps a better way would be to check to see if status registers need modifying before setting the write enable latch? Then unsetting it to prevent accidental writes.

Good point, will update in the separate patch.

lucien-nxp (approved):

mmahadevan108 (approved):

de-nordic (approved):

PR #101042, opened 2025-12-15
boards: cc1352p7_lp: Add a way to reset radio core using JTAG

anobli:
The radio driver hangs when the application core is reset by JTAG. We can't use the system reset because it also reset the JTAG itself and cause a lot of error in OpenOCD.
This updates OpenOCD script to reset the radio core and makes possible using JTAG for debugging radio.

### DIFF
--- a/boards/ti/cc1352p7_launchpad/support/openocd.cfg
+++ b/boards/ti/cc1352p7_launchpad/support/openocd.cfg
@@ -1,3 +1,13 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+#
+# TI CC13x2P7 LaunchPad Evaluation Kit
+#
+
+source [find mem_helper.tcl]
+set  PRCM_BASE_ADDR    0x40082000
+set  PDCTL0RFC         [expr       {$PRCM_BASE_ADDR     +  0x00000130}]
+
 # Serial could be found using the following command:
 # lsusb -d 0451:bef3 -v | grep -i iserial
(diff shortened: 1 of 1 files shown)


vaishnavachath (approved):

boards/ti/cc1352p7_launchpad/support/openocd.cfg
```diff
@@ -1,7 +1,25 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
```
Ayush1325, line 1: Why are you changing the file license.

Ayush1325 (commented):
Shouldn't these changes go to upstream openocd `board/ti_cc26x2x7_launchpad.cfg`?

boards/ti/cc1352p7_launchpad/support/openocd.cfg
kartben, line 1: No GPL code in zephyr please - is this copied from Linux or something?

kartben (changes requested):

boards/ti/cc1352p7_launchpad/support/openocd.cfg
anobli: Yes and no! Actually, I have fork of openocd in which I have added and updated support of cc1352p7.
But I have never pushed this specific change to openocd because I would like to test it on other SoC from the same family and also because I think I could do something better.
So, I am the original author of that change, and, I will fix the license!

PR #101053, opened 2025-12-15
drivers: led: Add TI TLC59116 driver

markus-becker-tridonic-com:
Add led_strip I2C driver for TI TLC 59116 (https://www.ti.com/product/de-de/TLC59116).


drivers/led_strip/tlc59116.c
```diff
@@ -0,0 +1,113 @@
...
+
+	uint8_t buf[16];
+
+	for (size_t i = 0; i < num_pixels && i < 16; i++) {
+		buf[i] = pixels[i].r;
+	}
+
+	int ret = i2c_burst_write_dt(&config->i2c, AUTO_INCREMENT_ALL | TLC59116_PWM0, buf,
+					 num_pixels);
+
```
simonguinot: Hi @markus-becker-tridonic-com,

Thanks for this driver.

Can multiple TLC59116 devices be daisy chained ? I can see here that the LED registers can be updated independently, and that each TLC59116 device can have its own I2C address. So I'd say that the TLC59116 is a "classic" LED controller and that its Zephyr driver should implement the LED API instead of LED strip.

In addition this driver should support any kind of DT representation:

- 3x RGBW LEDs;
```
led0: led_0 {
        label = "LED 0";
        color-mapping = <LED_COLOR_ID_RED>,
                        <LED_COLOR_ID_GREEN>,
                        <LED_COLOR_ID_BLUE>,
                        <LED_COLOR_ID_WHITE>;
};

...

led3: led_3 {
        label = "LED 3";
        color-mapping = <LED_COLOR_ID_RED>,
                        <LED_COLOR_ID_GREEN>,
                        <LED_COLOR_ID_BLUE>,
                        <LED_COLOR_ID_WHITE>;
};
```
- 16x monochromatic LEDs:
```
led0: led_0 {
        label = "LED 0";
        color-mapping = <LED_COLOR_ID_BLUE>;
};

...

led15: led_15 {
        label = "LED 15";
        color-mapping = <LED_COLOR_ID_BLUE>;
};
```
- Or any other custom configuration mixing different types of LEDs.

markus-becker-tridonic-com: The chip has 16 output channels and multiple chips can be on the same I2C line.

@simonguinot Should it then be similar to `drivers/led/lp5569.c`?

What should be led_strip then? Chained LEDs where the complete chain needs to be updated together?

simonguinot: > The chip has 16 output channels and multiple chips can be on the same I2C line.
> 
> @simonguinot Should it then be similar to `drivers/led/lp5569.c`?

Yes. Maybe `lp50xx` is a better model. Note that there is also a `tlc59108` driver. Maybe it can be extended to suppor the TLC59116 ?

>
> What should be led_strip then? Chained LEDs where the complete chain needs to be updated together?

These are daisy chained LED modules, where the signal is transmitted from one LED to the next. And indeed, they cannot be updated individually.

markus-becker-tridonic-com: @simonguinot OK, I have adapted to the led API. I have kept it separate from tlc59108 for now. Do you have directions on how to generalize, i.e. how to configure for the different number of LEDs?

Also I have implemented the write_channels API for the 59116.

simonguinot: > @simonguinot OK, I have adapted to the led API. I have kept it separate from tlc59108 for now. Do you have directions on how to generalize, i.e. how to configure for the different number of LEDs?
> 
> Also I have implemented the write_channels API for the 59116.

Hi @markus-becker-tridonic-com,

First of all, I apologize for the huge delay in this answer. Don't hesitate to ping me if you feel that I don't answer fast enough.

About how to handle different types of LEDs (i.e. monochromatic, multi-color, mixed - if it is the question), you can have a look at the lp50xx or is31fl319x drivers.

First you have to extract the configuration from DT: https://github.com/zephyrproject-rtos/zephyr/blob/afcfbf6dee02ec40e1690132228da1c5faa7d5fd/drivers/led/is31fl319x.c#L387

And after you have to perform basic checks in `->set_color`, `->set_brightness` and `init_fn()`:
- https://github.com/zephyrproject-rtos/zephyr/blob/afcfbf6dee02ec40e1690132228da1c5faa7d5fd/drivers/led/is31fl319x.c#L217
- https://github.com/zephyrproject-rtos/zephyr/blob/afcfbf6dee02ec40e1690132228da1c5faa7d5fd/drivers/led/is31fl319x.c#L243
- https://github.com/zephyrproject-rtos/zephyr/blob/afcfbf6dee02ec40e1690132228da1c5faa7d5fd/drivers/led/is31fl319x.c#L272

Or you can also limit this driver to monochromatic LEDs, and let the remaining work to someone else who needs it :) In that case, you just have to state the limitations in the Kconfig and DT binding files (and the commit message as well).

The next step for this PR is to merge the tlc59108 and tlc59116 LED drivers (as they are under Linux).

samples/drivers/led/tlc59116/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(led_tlc59116)
+
+FILE(GLOB app_sources src/*.c)
+target_sources(app PRIVATE ${app_sources})
```
thedjnK, line 9: do not use a glob for 1 .c file

PR #101066, opened 2025-12-15
stm32: tests: drivers: app_development: vector_table_relocation: exclude STM32F4 series

djiatsaf-st:
After commit 2d4feea8cea7c93d0b16b9544110ad0adc90e350, CCM RAM was treated as DTCM in the device tree, causing the test to incorrectly attempt relocation to CCM RAM. This results in failure because the CPU cannot fetch vectors from CCM RAM.

Based on RM0090 Rev 21 Page 60&69 

- The Cortex-M4 core always fetches the reset vector using the ICode bus, which only accesses the code region (typically Flash memory).

- The CCM RAM  on STM32F4 series is connected via the D-bus, not the ICode bus, and therefore cannot serve as an executable boot space for the vector table.

Additionally, let's add support for the STM32F3 disco (which also has a CCM RAM) to run this test.

### DIFF
--- a/tests/application_development/vector_table_relocation/testcase.yaml
+++ b/tests/application_development/vector_table_relocation/testcase.yaml
@@ -14,7 +14,11 @@ tests:
 
   application_development.vector_table_relocation.dtcm:
     arch_allow: arm
+    # Exclude STM32F4 series due to architectural constraints. STM32F4 CCM is only accessible
+    # over the D-bus but exception vectors are fetched over the I-bus.
+    # As such, the vector table can't be relocated to CCM.
     filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,dtcm")
+            and not CONFIG_SOC_SERIES_STM32F4X
     extra_configs:
       - CONFIG_ARM_VECTOR_TABLE_DTCM=y
 
--- a/boards/st/stm32f3_disco/stm32f3_disco.dts
+++ b/boards/st/stm32f3_disco/stm32f3_disco.dts
@@ -19,6 +19,7 @@
 		zephyr,sram = &sram0;
 		zephyr,flash = &flash0;
 		zephyr,canbus = &can1;
+		zephyr,dtcm = &ccm0;
 	};
 
 	leds {


tests/application_development/vector_table_relocation/testcase.yaml
```diff
@@ -17,6 +17,8 @@ tests:
     filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,dtcm")
     extra_configs:
       - CONFIG_ARM_VECTOR_TABLE_DTCM=y
+    platform_exclude:
+      - nucleo_f429zi
```
JarmouniA: It would probably be better to exclude the F3 & F4 series instead of individual boards.

djiatsaf-st: Thanks,  updated

tests/application_development/vector_table_relocation/testcase.yaml
```diff
@@ -15,6 +15,7 @@ tests:
   application_development.vector_table_relocation.dtcm:
     arch_allow: arm
     filter: CONFIG_CPU_CORTEX_M_HAS_VTOR and dt_chosen_enabled("zephyr,dtcm")
+            and not CONFIG_SOC_SERIES_STM32F4X
```
mathieuchopstm, line 21: Nit: add a comment about why F4 is banned? having to look at history is annoying

e.g.:
```
STM32F4 CCM is only accessible over the DCode bus
but exception vectors are fetched over the ICode bus.
As such, the vector table can't be relocated to CCM.
```

djiatsaf-st: updated

mathieuchopstm (dismissed):

JarmouniA (dismissed):

nordicjm (dismissed):

tests/application_development/vector_table_relocation/testcase.yaml
```diff
@@ -14,7 +14,11 @@ tests:
 
   application_development.vector_table_relocation.dtcm:
     arch_allow: arm
+    # Exclude STM32F4 series due to architectural constraints. STM32F4 CCM is only accessible
+    # over the DCode bus but exception vectors are fetched over the ICode bus.
```
etienne-lms: To prevent confusion:
```suggestion
    # over the D-bus but exception vectors are fetched over the I-bus.
```
DCode/ICode busses connect the interconnect to the flash memory while D-bus/I-bus busses connect the CPU to the interconnect (and CCM for D-bus only).

Suggestion for the commit message:
```patch
-not the ICode bus, and therefore cannot serve as an executable boot
+not the I-bus, and therefore cannot serve as an executable boot

**Review by etienne-lms — CHANGES_REQUESTED** [review]
Minor comment on inline comment but worth to be fixed IMHO.

File: tests/application_development/vector_table_relocation/testcase.yaml (hunk 3 above)
Comment (djiatsaf-st) [reply tests/application_development/vector_table_relocation/testcase.yaml]: Updated

**Review by etienne-lms — APPROVED** [review]

File: tests/application_development/vector_table_relocation/testcase.yaml (hunk 3 above)
Comment (mathieuchopstm) [reply tests/application_development/vector_table_relocation/testcase.yaml]: > DCode/ICode busses connect the interconnect to the flash memory while D-bus/I-bus busses connect the CPU to the interconnect (and CCM for D-bus only).

You are mixing `DCode`/`ICode` (the name of buses - or rather bus interfaces - on the Cortex-M4) and `DCODE`/`ICODE` (the name given by ST to slave interfaces of the flash interface - or rather an arbiter)

At no point in the Cortex-M4 TRM will you find `I-bus` or `D-bus`, this is an unofficial nickname at best; in fact, it is so unofficial that it only appears once in the entirety of RM0090 whereas `ICode` (or `I-Code` - again unofficial, but almost identical to ARM's naming) is used everywhere else.

And again quoting the RM:
> The Cortex®-M4 with FPU CPU always fetches the reset vector on the **ICode** bus

File: tests/application_development/vector_table_relocation/testcase.yaml [hunk 4]
```diff
@@ -14,7 +14,11 @@ tests:
 
   application_development.vector_table_relocation.dtcm:
     arch_allow: arm
+    # Exclude STM32F4 series due to architectural constraints. STM32F4 CCM is only accessible
+    # over the D-bus but exception vectors are fetched over the I-bus.
+    # As such, the vector table can't be relocated to CCM.
```
mathieuchopstm, line 19: Nit: flatten the message
```suggestion
    # Exclude STM32F4 series due to architectural constraints. STM32F4 CCM is only
    # accessible over the D-bus but exception vectors are fetched over the I-bus.
    # As such, the vector table can't be relocated to CCM.
```

mathieuchopstm (approved):
won't block but I don't like the updated wording

tejlmand (approved):

PR #101067, opened 2025-12-15
net_buf: Fixed race condition in net_buf_ref() / net_buf_unref() functions

dwyoung43:
which could cause an incorrect buffer ref count when the same buffer was referenced/dereferenced from two different threads, causing a memory leak.

### DIFF
--- a/include/zephyr/sys/atomic_types.h
+++ b/include/zephyr/sys/atomic_types.h
@@ -8,6 +8,13 @@
 #ifndef ZEPHYR_INCLUDE_SYS_ATOMIC_TYPES_H_
 #define ZEPHYR_INCLUDE_SYS_ATOMIC_TYPES_H_
 
+/* Detect C11 stdatomic support (C mode only, not C++) */
+#if !defined(__cplusplus) && defined(__STDC_VERSION__) && \
+	(__STDC_VERSION__ >= 201112L) && !defined(__STDC_NO_ATOMICS__)
+#define ZEPHYR_HAS_C11_STDATOMIC 1
+#include <stdatomic.h>
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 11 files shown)


jhedberg (changes requested):
I'm not opposed to adding such a feature (yes, it's a feature - the current design is a known compromise that was acceptable when the use cases for net_buf were limited and all known).

That said, it's possible that net_buf will at some point do all this using Zephyr's atomic API, meaning `atomic_t`, `atomic_dec()` and `atomic_inc()`. I think it would therefore make sense to create local helpers which mimic those `dec()` & `inc()` APIs, to ease the transition to `atomic_t` (if this ever happens it might also be a compile-time option to let people use the more compact representation we have today).

Zephyr has an existing C-only implementation for `atomic_t` for architectures which don't offer any native implementation of this. I think you can likely just copy that implementation except the underlying type is the current `uint8_t`:
https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/atomic_c.c

lib/net_buf/buf.c
```diff
@@ -445,19 +445,14 @@ void net_buf_unref(struct net_buf *buf)
...
-		__ASSERT(buf->ref, "buf %p double free", buf);
-		if (!buf->ref) {
-#if defined(CONFIG_NET_BUF_LOG)
-			NET_BUF_ERR("%s():%d: buf %p double free", func, line,
-				    buf);
-#endif
-			return;
-		}
-		NET_BUF_DBG("buf %p ref %u pool_id %u frags %p", buf, buf->ref,
+		__ASSERT(atomic_get(&buf->ref), "buf %p double free", buf);
```
pdgendt: No implicit boolean conversion please.

```suggestion
		__ASSERT(atomic_get(&buf->ref) > 0, "buf %p double free", buf);
```

dwyoung43: > No implicit boolean conversion please.

Thanks for the feedback, addressed this and the Copilot feedback.

lib/net_buf/buf.c
```diff
@@ -485,9 +485,9 @@ struct net_buf *net_buf_ref(struct net_buf *buf)
 {
 	__ASSERT_NO_MSG(buf);
 
-	NET_BUF_DBG("buf %p (old) ref %u pool_id %u",
-		    buf, buf->ref, buf->pool_id);
-	buf->ref++;
+	NET_BUF_DBG("buf %p (old) ref %ld pool_id %u",
+		    buf, atomic_get(&buf->ref), buf->pool_id);
+	atomic_inc(&buf->ref);
```
jhedberg: Isn't there some assert or NULL return missing here if the previous value of `buf->ref` was 0?

include/zephyr/sys/atomic_types.h
```diff
@@ -17,6 +24,33 @@ typedef atomic_t atomic_val_t;
...
+ * - Tier 1 (C11): When C11 stdatomic is available, uses atomic_uchar
+ *   from <stdatomic.h> for true atomic semantics.
+ * - Tier 2 (GCC Builtins): In C++ mode or without C11 support, uses
+ *   unsigned char with GCC __atomic_* builtins.
+ */
+#if defined(CONFIG_RISCV) || defined(CONFIG_XTENSA) || \
+	defined(CONFIG_CPU_CORTEX_M0) || defined(CONFIG_CPU_CORTEX_M0PLUS) || \
+	defined(CONFIG_RX) || defined(CONFIG_ARC)
+/* Tier 3: These architectures lack native byte-sized atomics, use word type */
+typedef atomic_t atomic_uchar_t;
```
jhedberg, line 45: Wouldn't this violate the C spec, which states that the `atomic_uchar` type should resolve to `_Atomic unsigned char`? E.g. I think it should be valid for code do depend on `sizeof(atomic_uchar) == sizeof(unsigned char)`.

dwyoung43: Note the code is using a new non standard type, atomic_uchar_t. So here atomic_uchar_t, not atomic_unchar, is getting typedef'd to atomic_t. It was done this way because I had a lot of trouble getting the atomic_uchar API's working on every single platform. Some platforms support C11 / byte size atomics, while other platforms support atomics but not byte-sized atomics. C++ doesn't support atomics at all. Creating the intermediate type made it easier to abstract this complexity away from client code. In this file, if the platform support C11 byte-sized atomics, atomic_uchar_t is typedef'd to atomic_uchar. Otherwise it's typedef'd to something else which will work on the respective platform.

jhedberg: Understood. Wouldn't you still have had the option of introducing spinlocks or similar for the platforms that don't support byte size atomics, or did you already consider that and decided that it's better to increase the struct size (RAM) rather than code size (ROM)?

dwyoung43: I did think about using spinlocks but was concerned about disabling interrupts every time a buffer was ref'd or unref'd - that can sometimes cause undesirable side effects, particularly if it's happening often. I thought it was better to take the RAM hit on the platforms that don't support byte-sized atomics - seemed lower risk and at compile time developers will know whether they've got a problem or not.

PR #101069, opened 2025-12-15
intel_adsp: common: gdbstub: Fix compilation CONFIG_INTEL_ADSP_DEBUG_…

ujfalusi:
…SLOT_MANAGER=n

The GDB is not compiled by default and a recent change in the slot manager series contained a type in ifdef and missed by not compiling the GDB support.

Fixes: ebb5625bee8a ("intel_adsp: Add debug slot manager")

### DIFF
--- a/soc/intel/intel_adsp/common/gdbstub_backend_sram.c
+++ b/soc/intel/intel_adsp/common/gdbstub_backend_sram.c
@@ -60,7 +60,7 @@ static inline int ring_have_data(const volatile struct gdb_sram_ring *ring)
 static volatile struct gdb_sram_ring *rx;
 static volatile struct gdb_sram_ring *tx;
 
-#ifndef CONFIG_INTEL_ADSP_DEBUG_SLOT_MANAGER
+#ifdef CONFIG_INTEL_ADSP_DEBUG_SLOT_MANAGER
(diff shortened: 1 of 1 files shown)


dcpleung (dismissed):

nashif (dismissed):

ujfalusi:
Changes since v1:
- there were a missing `#else` as well, now added and triple compile and boot tested the firmware.

lyakh (approved):
will probably need another round of approvals after fixing the CI

tmleman (approved):
Thanks for the quick fix! But can you correct the commit title (it's too long) so it wouldn't hang in review until the new year.

soc/intel/intel_adsp/common/gdbstub_backend_sram.c
```diff
@@ -60,7 +60,7 @@ static inline int ring_have_data(const volatile struct gdb_sram_ring *ring)
 static volatile struct gdb_sram_ring *rx;
```
kv2019i, line 60: Murphy strikes... "type in ifdef" :D

kv2019i (approved):

dcpleung (approved):

PR #101076, opened 2025-12-15
drivers: wifi: siwx91x: Disable automatic rejoin in NWP

silabs-MuzaffarA:
Disable rejoin by setting max_retry_attempts to 1. This is done to keep the NWP state in sync with Zephyr Host, given that Zephyr does not have a "rejoin ongoing" state.

### DIFF
--- a/drivers/wifi/siwx91x/siwx91x_wifi.c
+++ b/drivers/wifi/siwx91x/siwx91x_wifi.c
@@ -454,6 +454,12 @@ static void siwx91x_iface_init(struct net_if *iface)
 {
 	const struct siwx91x_config *siwx91x_cfg = iface->if_dev->dev->config;
 	struct siwx91x_dev *sidev = iface->if_dev->dev->data;
+	sl_wifi_advanced_client_configuration_t client_config = {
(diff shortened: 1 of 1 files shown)


drivers/wifi/siwx91x/siwx91x_wifi.c
```diff
@@ -454,6 +454,12 @@ static void siwx91x_iface_init(struct net_if *iface)
 {
 	const struct siwx91x_config *siwx91x_cfg = iface->if_dev->dev->config;
 	struct siwx91x_dev *sidev = iface->if_dev->dev->data;
+	sl_wifi_advanced_client_configuration_t client_config = {
```
jerome-pouiller, line 457:
```
/* Despite the name, all the field in this struct impact are related to the
 * automatic re-join mechanism in the NWP. Disable this behavior to keep Zephyr
 * aware about the real status of the Wifi connection.
 */
```

jerome-pouiller (approved):
You may want to add the comment I suggest. Otherwise, LGTM.

jhedberg (approved):

PR #101077, opened 2025-12-15
drivers: clock_control: stm32h7: fix PLL rate with fractional PLL

etienne-lms:
Take PLL fractional part into account when computing PLL output rate for stm32h7xx and stm32h7rs based targets.

### DIFF
--- a/drivers/clock_control/clock_stm32_ll_h7.c
+++ b/drivers/clock_control/clock_stm32_ll_h7.c
@@ -48,6 +48,16 @@
 #define z_apb5_prescaler(v) LL_RCC_APB5_DIV_ ## v
 #define apb5_prescaler(v) z_apb5_prescaler(v)
 
+/* PLLx fractional ratio is 2^13 */
+#define PLL_FRACN_DIVISOR 8192
+
+/* Given source clock and dividers, compute the output frequency of PLLx_P */
+#define PLLX_VCO_FREQ(pllsrc_freq, divm, divn, fracn) \
+	((((pllsrc_freq) / (divm)) * (((divn) * PLL_FRACN_DIVISOR) + (fracn))) / PLL_FRACN_DIVISOR)
+
+#define PLLX_OUT_FREQ(pllsrc_freq, divm, divn, fracn, out_div) \
+	(PLLX_VCO_FREQ((pllsrc_freq), (divm), (divn), (fracn)) / (out_div))
+
 /* Macro to check for clock feasibility */
 /* It is Cortex M7's responsibility to setup clock tree */
 /* This check should only be performed for the M7 core code */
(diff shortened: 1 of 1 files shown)


drivers/clock_control/clock_stm32_ll_h7.c
```diff
@@ -205,13 +209,17 @@ static uint32_t get_bus_clock(uint32_t clock, uint32_t prescaler)
...
+				     unsigned int pllout_div)
 {
-	__ASSERT_NO_MSG(pllm_div && pllout_div);
+	__ASSERT_NO_MSG(pllm_div != 0 && pllout_div != 0);
+
+	uint32_t f_vco = (pllsrc_freq / pllm_div) *
+			 ((uint64_t)plln_mul * PLL_FRACN_DIVISOR + plln_frac) / PLL_FRACN_DIVISOR;
 
-	return (pllsrc_freq / pllm_div) * plln_mul / pllout_div;
+	return f_vco / pllout_div;
```
mathieuchopstm: There is a macro doing both of these operations: maybe it should be reused?

drivers/clock_control/clock_stm32_ll_h7.c
mathieuchopstm, line 1: Have you checked that this optimizes out to the same code as previously when `fracn == 0` (i.e., fractional mode is not enabled - I think it *is* optional?)

etienne-lms: Tested on Zephyr SDK's GCC: yes, the same code footprint is generated when no `fracn` properties are defined for PLLs.

drivers/clock_control/clock_stm32_ll_h7.c
etienne-lms: Fair enough.

etienne-lms:
Fixed h7xx//m4 variants.

drivers/clock_control/clock_stm32_ll_h7.c
etienne-lms: Addressed.

drivers/clock_control/clock_stm32_ll_h7.c
```diff
@@ -205,13 +209,16 @@ static uint32_t get_bus_clock(uint32_t clock, uint32_t prescaler)
...
+				     unsigned int plln_frac,
+				     unsigned int pllout_div)
 {
-	__ASSERT_NO_MSG(pllm_div && pllout_div);
+	uint64_t plln_mul_u64 = plln_mul;
+
+	__ASSERT_NO_MSG(pllm_div != 0 && pllout_div != 0);
 
-	return (pllsrc_freq / pllm_div) * plln_mul / pllout_div;
+	return PLLX_P_FREQ(pllsrc_freq, pllm_div, plln_mul_u64, plln_frac, pllout_div);
```
erwango: Function is not dedicated to P output, it is also called for Q/S/R outputs.
Probably we should avoid `_P_` here

erwango (changes requested):

drivers/clock_control/clock_stm32_ll_h7.c
mathieuchopstm: \+1 - `PLLX_P_FREQ` macro has little benefits, just divide directly

etienne-lms: > `PLLX_P_FREQ` macro has little benefits, just divide directly

I've added the `PLLX_VCO_FREQ` helper macro to ease reading unless what line escapes and indentation make the division operations not obvious, e.g.:
```c
#define PLLX_OUT_FREQ(pllsrc_freq, divm, divn, fracn, out_div) \
	(((((pllsrc_freq) / (divm)) * (((divn) * PLL_FRACN_DIVISOR) + (fracn))) / \
	  PLL_FRACN_DIVISOR) / (out_div))
```

mathieuchopstm (approved):

erwango (approved):

PR #101088, opened 2025-12-16
drivers: spi: spi_context.h compiler warning

billwatersiii:
Add (void) to all spi_context_cs_* macros to avoid this warning.
warning: statement with no effect [-Wunused-value]

This was already do for two of the five macros.

### DIFF
--- a/drivers/spi/spi_context.h
+++ b/drivers/spi/spi_context.h
@@ -302,13 +302,6 @@ static inline void spi_context_complete(struct spi_context *ctx,
 #endif /* CONFIG_SPI_ASYNC */
 }
 
-#ifdef DT_SPI_CTX_HAS_NO_CS_GPIOS
-#define spi_context_cs_configure_all(...) 0
-#define spi_context_cs_get_all(...) 0
-#define spi_context_cs_put_all(...) 0
-#define _spi_context_cs_control(...) (void) 0
-#define spi_context_cs_control(...) (void) 0
-#else /* DT_SPI_CTX_HAS_NO_CS_GPIOS */
 /*
  * This function initializes all the chip select GPIOs associated with a spi controller.
  * The context first must be initialized using the SPI_CONTEXT_CS_GPIOS_INITIALIZE macro.
@@ -319,6 +312,7 @@ static inline void spi_context_complete(struct spi_context *ctx,
  */
 static inline int spi_context_cs_configure_all(struct spi_context *ctx)
 {
+#ifndef DT_SPI_CTX_HAS_NO_CS_GPIOS
 	int ret;
 	const struct gpio_dt_spec *cs_gpio;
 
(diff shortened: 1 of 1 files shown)


JarmouniA (dismissed):

drivers/spi/spi_context.h
```diff
@@ -303,9 +303,9 @@ static inline void spi_context_complete(struct spi_context *ctx,
 }
 
 #ifdef DT_SPI_CTX_HAS_NO_CS_GPIOS
-#define spi_context_cs_configure_all(...) 0
-#define spi_context_cs_get_all(...) 0
-#define spi_context_cs_put_all(...) 0
+#define spi_context_cs_configure_all(...) (void) 0
+#define spi_context_cs_get_all(...) (void) 0
+#define spi_context_cs_put_all(...) (void) 0
```
pdgendt: This is not possible, the return type of the functions is `int`.

I think a better solution is to move the `ifdefry` to inside the (already `static inline`) functions.

For example

```c
static inline int spi_context_cs_get_all(struct spi_context *ctx)
{
#ifndef DT_SPI_CTX_HAS_NO_CS_GPIOS
	return _spi_context_cs_pm_all(ctx, true);
#else
	ARG_UNUSED(ctx);
	return 0;
#endif
}
```

And similar for the others.

pdgendt (changes requested):

drivers/spi/spi_context.h
billwatersiii: Ah.  Got it.  I will use your update instead.  Thanks.

pdgendt (approved):

teburd (approved):

billwatersiii:
@tbursztyka , do you have an opinion about the comments from @decsny and @pdgendt ?  Can you approve this pull request?

maass-hamburg (approved):

drivers/spi/spi_context.h
```diff
@@ -302,13 +302,6 @@ static inline void spi_context_complete(struct spi_context *ctx,
 #endif /* CONFIG_SPI_ASYNC */
```
tbursztyka, line 302: a little bit more descriptive commit title would be nice

billwatersiii: @tbursztyka , I updated the commit title

drivers/spi/spi_context.h
```diff
@@ -334,10 +328,14 @@ static inline int spi_context_cs_configure_all(struct spi_context *ctx)
 			return ret;
 		}
 	}
+#else
+	ARG_UNUSED(ctx);
```
dkalowsk, line 332: not a blocking comment, but this kind of behavior can be mitigated with the `__maybe_unused` on the function arg as well.

billwatersiii: Oh.  Thanks.  Like this?

`static inline int spi_context_cs_configure_all(__maybe_unused struct spi_context *ctx)`

And then remove lines 315, 331-333, and 338?

teburd:
@tbursztyka Any chance you could take a look? I'm going to re-assign to myself as a collaborator as this has sat for some time. But I would really appreciate your review as maintainer.

dkalowsk (approved):
Not a huge fan of preprocessor, but that is the pattern in use here.

PR #101096, opened 2025-12-16
drivers: spi: esp32: fix CS handling when using GPIO chip select

sylvioalves:
The driver was unconditionally setting hal_dev->cs_pin_id to the target number, which activates hardware CS lines even when GPIO-based chip select (cs-gpios) is used. This caused issues when using hardware CS via pinctrl with reg > 0.

Also update the binding documentation to clarify the interaction between cs-gpios and hardware CS via pinctrl.

Fixes #100069


wmrsouza (approved):

JarmouniA (approved):

sylvioalves:
> @sylvioalves Can this fix be backported to v4.3 & v4.2?

I'll check.

PR #101098, opened 2025-12-16
drivers: usb: uhc: fix unused variable warning for phy_config

AidenHu:
Wrap phy_config declaration with CONFIG_DT_HAS_NXP_USBPHY_ENABLED to avoid compiler warning when USB PHY is not enabled.

### DIFF
--- a/drivers/usb/uhc/uhc_mcux_ehci.c
+++ b/drivers/usb/uhc/uhc_mcux_ehci.c
@@ -93,8 +93,7 @@ static usb_status_t mcux_host_callback(usb_device_handle deviceHandle,
 
 static int uhc_mcux_init(const struct device *dev)
 {
-	const struct uhc_mcux_config *config = dev->config;
-	usb_phy_config_struct_t *phy_config;
+	const struct uhc_mcux_ehci_config *config = dev->config;
 	struct uhc_mcux_data *priv = uhc_get_private(dev);
 	k_thread_entry_t thread_entry = NULL;
 	usb_status_t status;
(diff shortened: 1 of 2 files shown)


Holt-Sun (dismissed):

drivers/usb/uhc/uhc_mcux_ehci.c
```diff
@@ -94,7 +94,9 @@ static usb_status_t mcux_host_callback(usb_device_handle deviceHandle,
 static int uhc_mcux_init(const struct device *dev)
 {
 	const struct uhc_mcux_config *config = dev->config;
+#ifdef CONFIG_DT_HAS_NXP_USBPHY_ENABLED
 	usb_phy_config_struct_t *phy_config;
+#endif
```
jfischer-no: Change it to
```c
static int uhc_mcux_init(const struct device *dev)
{

	const struct uhc_mcux_config *config = dev->config;
	usb_phy_config_struct_t *phy_config = config->phy_config;
```

Remove L112.

jfischer-no (changes requested):

drivers/usb/uhc/uhc_mcux_ehci.c
MarkWangChinese, line 113: It seems that the code can use `config->phy_config` directly here.
```
	if (config->phy_config != NULL) {
		USB_EhciPhyInit(priv->controller_id, 0u, config->phy_config);
	}
```

drivers/usb/uhc/uhc_mcux_ehci.c
AidenHu: @jfischer-no 
Have updated to be like this:
`	const struct uhc_mcux_ehci_config *config = dev->config;`

MarkWangChinese (approved):

AidenHu:
Hi @jfischer-no 
Could you help to check if it is ok to merge this PR? Thanks.

mmahadevan108 (approved):

jfischer-no (approved):

PR #101105, opened 2025-12-16
subsys: logging: backends: fix mqtt log backend on client disconnect

### LINKED ISSUE #101064: MQTT Logging backend can cause logging thread to take over idle time
### Describe the bug

Im using the mqtt logging backend. Every time I lose connection to the network the logging thread seems to take over my systems idle time.

### Regression

- [ ] This is a regression.

### Steps to reproduce

1. Use the mqtt logging backend. 
2. lose network connection
3. Log via the mqtt backend
4. observe logging thread runtime

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Windows
- Toolchain: Zephyr SDK + Nordic SDK
- Last Commit to log_backend_mqtt.c: https://github.com/zephyrproject-rtos/zephyr/commit/44b4fff1640b2d035bacf4f4034e4af4d37bf07d

### Additional Context

The function _log_output_func_ in  [_log_backend_mqtt.c_](https://github.com/zephyrproject-rtos/zephyr/blob/44b4fff1640b2d035bacf4f4034e4af4d37bf07d/subsys/logging/backends/log_backend_mqtt.c) returns the _mqtt_publish_ error code when it fails. The description for _log_output_func_t_ states:

 ```C
...
 *
 * @note If the log output function cannot process all of the data, it is
 *       its responsibility to mark them as dropped or discarded by returning
 *       the corresponding number of bytes dropped or discarded to the caller.
 */
typedef int (*log_output_func_t)(uint8_t *buf, size_t size, void *ctx);
```

So i assume this function should return _length_ in all case? Not returning _length_ would cause the logging thread to attempt to send the log again?

Im currently ignoring the return value of the _mqtt_publish_ and always returning _length_ (thus dropping the msg). In my case, this behavior matches my system requirements. If it is desiredfor the logs to be resent once the mqtt connection is restored, there should be some kind of timeout or backoff between send attempts to prevent the logging thread from fully taking over the idle time?

The net and ble backends also just return _length_ on error.

Am i missing something here? Are there other workarounds for this?

rlux-bluezone:
The MQTT logging backend can cause the logging thread to take over idle time when the used client disconnects/fails to send the log messages. Fixed by always returning the length of the log in every case. 

Fixes zephyrproject-rtos/zephyr#101064

Signed-off-by: Roland Lux <roland.lux@blue-zone.at>

### DIFF
--- a/subsys/logging/backends/log_backend_mqtt.c
+++ b/subsys/logging/backends/log_backend_mqtt.c
@@ -44,11 +44,7 @@ static int log_output_func(uint8_t *data, size_t length, void *output_ctx)
 	param.message_id = sys_rand32_get();
 #endif
 
-	int ret = mqtt_publish(client, &param);
-
-	if (ret != 0) {
-		return ret;
-	}
+	(void)mqtt_publish(client, &param);
 
 	return length;
 }


rlubos (commented):
The change looks good but why 3 commits? @rlux-bluezone  please squash them, and use `logging: backend: mqtt: ` for the commit title prefix.

rlubos (approved):

kartben (approved):

PR #101110, opened 2025-12-16
modules: Make uoscore-uedhoc module external

rlubos:
The uoscore-uedhoc module was introduced to Zephyr as a potential solution to a problem of NAT timeouts with DTLS sessions for protocols like LwM2M. In a meantime however, a viable alternative was presented in the core DTLS itself (DTLS Connection ID), which is currently used. In result, the uoscore-uedhoc library is not currently used in core Zephyr components, beside tests.

Moreover, the upstream project doesn't seem actively maintained anymore, so keeping it as an internal module is just a maintenance burden for the Zephyr project.

Therefore I propose to make the uoscore-uedhoc module external.


jukkar (dismissed):

valeriosetti (dismissed):

carlescufi (dismissed):

doc/develop/manifest/external/uoscore-uedhoc.rst
kartben, line 1: the file is misnamed s/udhoc/uedhoc/

kartben (changes requested):

str4t0m:
Didn't check the details yet, but #101126 seems to use the module with the coap server.

doc/develop/manifest/external/uoscore-uedhoc.rst
rlubos: Fixed

pdgendt (dismissed):

rlubos:
> This would be a good use case I think. I agree that there is no point in maintaining the module if nothing uses it. The PR is still a draft because at the moment, it is only a proof of concept, but it is generally working. A proper CoAP integration (i.e., with EDHOC or DTLS key exchange, tests, documentation) would be a valuable addition to Zephyr. This is something that I can work on in the near future.

I'd say we'd need to put the module removal on hold then and see where we get with the OSCORE/EDHOC integration in CoAP server.

PR #101117, opened 2025-12-16
net: zperf: Fix net API use and enable in tests/net/all

aescolar:
net: zperf: Fix net API use
    
    In d45cd6716bbab3a805e3a5fd461934f0dcdc13e5 the mayority of the
    Zephyr codebased was changed to use the Zephyr native net_ prefixed
    types, but some were forgotten.
    Without this fix/change the code still builds as we are by now setting
    CONFIG_NET_NAMESPACE_COMPAT_MODE. But when this is not set, things
    fail to build.
 
-------
   
    samples: net: zperf: Do not enable the POSIX_API
    
    This sample does not need it, and having it enabled masks uses of the
    non native API by this code.
 
-----

    tests: net: all: Do not disable sys_getopt users
    
    Since the shell (and its use of sys_getopt) does not have any dependency
    on the POSIX_API anymore, we do not need to disable it when we disable
    the POSIX_API compatibility layer.
    
-------

    tests: net: all: Also build zperf
    
    Also build the zperf common and server code.


jukkar (approved):
Nice catch! Thanks Alberto for fixing these, much appreciated :bow: :+1:

rlubos (approved):

PR #101121, opened 2025-12-16
kernel: poll: patch recursive lock in z_vrfy_k_poll

bjarki-andreasen:
In z_vrfy_k_poll, there is a memory access check
`K_SYSCALL_MEMORY_WRITE()` which is wrapped in a spinlock, the same spinlock used in `z_handle_obj_poll_events()` which is called from `k_sem_give()` for example.

The `K_SYSCALL_MEMORY_WRITE()` macro conditionally calls `LOG_ERR()` which may call the UART console, which may call an API like `k_sem_give()`. This will cause a deadlock since the locked spinlock will be relocked, and a recursive lock if `SPINLOCK_VALIDATE` and `ASSERTS` are enabled as the validation will fail, causing a `printk()`, causing a `k_sem_give()` causing a relock... until stack overflows.

To solve the issue, only protect the copy of `events` to `events_copy` with the spinlock, the content of `events` is not actually checked, and `bound` is not shared, so there is no need to do this validation in a critical section. The contents of `events` is shared so that must be copied in atomically.

### DIFF
--- a/kernel/poll.c
+++ b/kernel/poll.c
@@ -376,11 +376,11 @@ static inline int z_vrfy_k_poll(struct k_poll_event *events,
 		goto out;
 	}
 
-	key = k_spin_lock(&lock);
 	if (K_SYSCALL_MEMORY_WRITE(events, bounds)) {
-		k_spin_unlock(&lock, key);
 		goto oops_free;
 	}
+
+	key = k_spin_lock(&lock);
 	(void)memcpy(events_copy, events, bounds);
 	k_spin_unlock(&lock, key);
 


peter-mitsis:
What a tricky thing that must have been to track down.

If I am following the intended logic correctly, it looks like the original lock placement was to ensure that the buffer validation AND copy was atomic--that there is threat that the buffer may become invalid after it was validated, but before the copying was done. *If* that is true, then it also looks like there should be an atomic validation and copy after the call to k_poll() where we copy the data out of the kernel.

Based on this line of thinking, I am thinking that we need a version of K_SYSCALL_MEMORY_WRITE() that does not lead to this lock-up. That is, one that does not lead to calls to LOG_ERR() or other such logging utilities.

cfriedt (approved):

peter-mitsis (approved):
> Could you provide an example as to how the buffer would become invalid? K_SYSCALL_MEMORY_WRITE just checks that the usermode thread has write access to that area in memory, that can not change during a syscall unless the kernel explodes I think.

Sorry for the delayed response--I was on holidays. When I throw on my "what's the worst that can happen" hat, here is what I come up with .... If there is no lock around K_SYSCALL_MEMORY_WRITE + memcpy(), then there is nothing to prevent to prevent an ISR or a context switch to another thread between the two parts. Could either an ISR or another thread change the access permissions of that chunk of memory between the calls?

As I continue to think about that, I have to ask myself "Is that something we can control or should worry about?" The same type of concern could be applied to another CPU on an SMP system and the locking, like McBain's goggles, would do nothing. The best we can do is make a reasonable check, which is exactly what you are proposing.

I think I have convinced myself that this is OK.

PR #101124, opened 2025-12-16
drivers: flash: Implement readout protection for STM32U5 / STM32L5 

Johannes-Schuler:
These changes are adding readout protection for the stm32u5 and stm32l5 series.

Tested on Nucleo U575ZI Q with tests/drivers/flash/stm32/ with CONFIG_FLASH_STM32_READOUT_PROTECTION=y
Screenshot 2025-12-02 084042


FRASTM (dismissed):

florianschuele (approved):

drivers/flash/Kconfig.stm32
```diff
@@ -54,7 +54,8 @@ config FLASH_STM32_READOUT_PROTECTION
 		   SOC_SERIES_STM32G0X || \
 		   SOC_SERIES_STM32G4X || \
 		   SOC_SERIES_STM32H7X || \
-		   SOC_SERIES_STM32L4X
+		   SOC_SERIES_STM32L4X || \
+		   SOC_SERIES_STM32U5X
```
tpambor, line 59: I think `SOC_SERIES_STM32L5X` is missing here. Note, STM32H5 is using the same flash controller, so that `SOC_SERIES_STM32H5X` could be added here as well.

Johannes-Schuler: Added them. Thank you.

Johannes-Schuler: Unfortunately, I had to remove `SOC_SERIES_STM32H5X` again, because `flash_stm32_option_bytes_write()` uses register definitions which are unknown to this series.

drivers/flash/Kconfig.stm32
```diff
@@ -54,7 +54,9 @@ config FLASH_STM32_READOUT_PROTECTION
 		   SOC_SERIES_STM32G0X || \
 		   SOC_SERIES_STM32G4X || \
 		   SOC_SERIES_STM32H7X || \
-		   SOC_SERIES_STM32L4X
+		   SOC_SERIES_STM32L4X || \
+		   SOC_SERIES_STM32U5X || \
+		   SOC_SERIES_STM32L5X
```
etienne-lms: Nitpicking: prefer an alphabetical order. Would ease maintenance.
```suggestion
		   SOC_SERIES_STM32L4X || \
		   SOC_SERIES_STM32L5X || \
		   SOC_SERIES_STM32U5X

File: drivers/flash/flash_stm32l5x.c [hunk 3]
```diff
@@ -275,6 +275,43 @@ int flash_stm32_write_range(const struct device *dev, unsigned int offset,
...
+	FLASH_TypeDef *regs = FLASH_STM32_REGS(dev);
+	int rc;
+
+	if ((regs->OPTR & mask) == value) {
+		return 0;
+	}
+
+	regs->OPTR = (regs->OPTR & ~mask) | value;
+	regs->NSCR |= FLASH_NSCR_OPTSTRT;
+
```
etienne-lms, line 291:
```
	/* Make sure previous write is completed. */
	barrier_dsync_fence_full();

File: drivers/flash/Kconfig.stm32 (hunk 2 above)
Comment (Johannes-Schuler) [reply drivers/flash/Kconfig.stm32]: Yes, makes sense. Changed it.

File: drivers/flash/flash_stm32l5x.c (hunk 3 above)
Comment (Johannes-Schuler) [reply drivers/flash/flash_stm32l5x.c:291]: Thanks. Added it.

**Comment by Johannes-Schuler** [comment]
ping @de-nordic

**Review by etienne-lms — APPROVED** [review]

**Review by erwango — APPROVED** [review]

File: drivers/flash/flash_stm32l5x.c [hunk 4]
```diff
@@ -275,6 +276,46 @@ int flash_stm32_write_range(const struct device *dev, unsigned int offset,
...
+	barrier_dsync_fence_full();
+
+	rc = flash_stm32_wait_flash_idle(dev);
+	if (rc < 0) {
+		return rc;
+	}
+
+	regs->NSCR |= FLASH_NSCR_OBL_LAUNCH;
+
+	return 0;
```
lindblandro, line 302: Shouldn't this check the values of `OPTWERR` and `PGSERR` bits?

Johannes-Schuler: You might be right. I based my approach on other implementations that don't perform these checks. Perhaps someone from STM could explain why these checks haven't been done before?

lindblandro: If it's in line with other drivers then it might be fine, but might be good to know if it's by design or accidentally left out.

etienne-lms: `OPTWERR` and `PGSERR` error status bits are already checked by `flash_stm32_wait_flash_idle()`.

On L5 **and U5**, setting `OBL_LAUNCH` does restart the system. Note that this is also true on other STM32 series that embed the `OBL_LAUNCH` control bit. `return -EIO` could apply here, but not guaranteed to execute. `CODE_UNREACHABLE` would be valid too I think.

drivers/flash/flash_stm32l5x.c
```diff
@@ -275,6 +276,46 @@ int flash_stm32_write_range(const struct device *dev, unsigned int offset,
 	return rc;
 }
 
+#if defined(CONFIG_FLASH_STM32_READOUT_PROTECTION)
```
mrk-its, line 279: Can you do conditional compilation only for `flash_stm32_get_rdp_level` / `flash_stm32_set_rdp_level`, like other drivers do? The `flash_stm32_option_bytes_write` will be opThe `flash_stm32_option_bytes_write` will be optimized out by compiler when not referenced anywaytimized out by compiler when not referenced anyway.

It would be also great to expose `flash_stm32_option_bytes_write` (and `flash_stm32_option_bytes_read`) via `flash_stm32_ex_op` interface

drivers/flash/flash_stm32l5x.c
```diff
@@ -275,6 +276,46 @@ int flash_stm32_write_range(const struct device *dev, unsigned int offset,
...
+int flash_stm32_option_bytes_write(const struct device *dev, uint32_t mask, uint32_t value)
+{
+	FLASH_TypeDef *regs = FLASH_STM32_REGS(dev);
+	int rc;
+
```
mrk-its, line 284: Other drivers typically check here if option bytes are unlocked. It can be done this way:
```
if (regs->NSCR & FLASH_NSCR_OPTLOCK) {
        return -EIO;
}
```

drivers/flash/flash_stm32l5x.c
```diff
@@ -275,6 +276,46 @@ int flash_stm32_write_range(const struct device *dev, unsigned int offset,
...
+	if ((regs->OPTR & mask) == value) {
+		return 0;
+	}
+
```
mrk-its, line 288: other drivers do here:
```
rc = flash_stm32_wait_flash_idle(dev);
if (rc < 0) {
       return rc;
}
```

mrk-its (commented):
I have similar patch implementing `flash_stm32_option_bytes_write` and `flash_stm32_option_bytes_read` and exposing them via `flash_stm32_ex_op`, I noticed few differences between our implementations.

de-nordic (approved):

PR #100242, opened 2025-11-29
STM32H7RS: DCMIPP: enabling of camera pipe with ST-B-CAMS-OMV shield

avolmat-st:
This PR enables the DCMIPP based camera pipeline on the STM32H7RS. This is being tested by usage of the ST-B-CAMS-OMV shield (DVP) on the STM32H7S78-DK.

Tested ok via:
```
west build -p -b stm32h7s78_dk//ext_flash_app --shield st_b_cams_omv_mb1683 samples/drivers/video/capture --sysbuild
```

PR #101131, opened 2025-12-16
soc: st: stm32n6: Add DCMIPP CSI IRQ

not-na:
Currently, the DCMIPP driver for the STM32N6 only implements the DCMIPP's own IRQ. However, the CSI part of the DCMIPP has its own IRQ, which may be useful in some applications.

This PR adds the CSI IRQ to the STM32N6 devicetree and routes it to the ST HAL, just like the main DCMIPP ISR. Applications and libraries can then implement the various `HAL_DCMIPP_CSI_*Callback` weak methods they are interested in.

To demonstrate this, simply add something like below to the `drivers/video/capture` sample to log every CSI SOF:
```diff
diff --git a/samples/drivers/video/capture/src/main.c b/samples/drivers/video/capture/src/main.c
index f1b54aae110..bf5d86fc4c4 100644
--- a/samples/drivers/video/capture/src/main.c
+++ b/samples/drivers/video/capture/src/main.c
@@ -19,6 +19,11 @@ LOG_MODULE_REGISTER(main, CONFIG_LOG_DEFAULT_LEVEL);
 #error No camera chosen in devicetree. Missing "--shield" or "--snippet video-sw-generator" flag?
 #endif
 
+void HAL_DCMIPP_CSI_StartOfFrameEventCallback(DCMIPP_HandleTypeDef *hdcmipp, uint32_t VirtualChannel)
+{
+       LOG_INF("DCMIPP CSI IRQ!");
+}
+
 static inline int app_setup_display(const struct device *const display_dev, const uint32_t pixfmt)
 {
        struct display_capabilities capabilities;
```

As far as I can tell, only the N6 has the DCMIPP with CSI, so I only added the new IRQ to this SoC.

Please let me know if anything should be changed or more specific APIs for the various CSI events and errors are wanted.

### DIFF
--- a/drivers/video/Kconfig.stm32_dcmipp
+++ b/drivers/video/Kconfig.stm32_dcmipp
@@ -42,4 +42,11 @@ config VIDEO_STM32_DCMIPP_SENSOR_PIXEL_FORMAT
 	help
 	  Pixel format of the sensor video frame.
 
+config VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ
+	bool "CSI Interrupt"
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_DCMIPP),interrupt-names,dcmipp_csi)
+	help
+	  Enable the secondary DCMIPP interrupt dedicated to the CSI host. HAL callbacks
+	  HAL_DCMIPP_CSI_*Callback will be invoked and can be implemented by applications.
+
 endif
--- a/drivers/video/video_stm32_dcmipp.c
+++ b/drivers/video/video_stm32_dcmipp.c
@@ -1828,6 +1828,15 @@ static void stm32_dcmipp_isr(const struct device *dev)
 	HAL_DCMIPP_IRQHandler(&dcmipp->hdcmipp);
 }
 
+#if defined(STM32_DCMIPP_HAS_CSI) && defined(CONFIG_VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ)
+static void stm32_dcmipp_csi_isr(const struct device *dev)
+{
+	struct stm32_dcmipp_data *dcmipp = dev->data;
+
+	HAL_DCMIPP_CSI_IRQHandler(&dcmipp->hdcmipp);
+}
+#endif
+
 #define SOURCE_DEV(inst) DEVICE_DT_GET(DT_NODE_REMOTE_DEVICE(DT_INST_ENDPOINT_BY_ID(inst, 0, 0)))
 
 #define DCMIPP_PIPE_INIT_DEFINE(node_id, inst)						\
@@ -1856,8 +1865,20 @@ static void stm32_dcmipp_isr(const struct device *dev)
 					    (0),						\
 					    (DT_PROP_BY_IDX(DT_INST_ENDPOINT_BY_ID(inst, 0, 0),	\
 							    data_lanes, 1))),
+
+#if defined(CONFIG_VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ)
+#define STM32_DCMIPP_CSI_IRQ_INIT(inst)								\
+		IRQ_CONNECT(DT_INST_IRQ_BY_NAME(inst, dcmipp_csi, irq),				\
+			    DT_INST_IRQ_BY_NAME(inst, dcmipp_csi, priority),			\
+			    stm32_dcmipp_csi_isr, DEVICE_DT_INST_GET(inst), 0);			\
+		irq_enable(DT_INST_IRQ_BY_NAME(inst, dcmipp_csi, irq));
+#else
+#define STM32_DCMIPP_CSI_IRQ_INIT(inst)
+#endif
+
 #else
 #define STM32_DCMIPP_CSI_DT_PARAMS(inst)
+#define STM32_DCMIPP_CSI_IRQ_INIT(inst)
 #endif
 
 #if defined(STM32_DCMIPP_HAS_PIXEL_PIPES)
@@ -1871,9 +1892,11 @@ static void stm32_dcmipp_isr(const struct device *dev)
 #define STM32_DCMIPP_INIT(inst)									\
 	static void stm32_dcmipp_irq_config_##inst(const struct device *dev)			\
 	{											\
-		IRQ_CONNECT(DT_INST_IRQN(inst), DT_INST_IRQ(inst, priority),			\
+		IRQ_CONNECT(DT_INST_IRQ_BY_NAME(inst, dcmipp, irq),				\
+			    DT_INST_IRQ_BY_NAME(inst, dcmipp, priority),			\
 			    stm32_dcmipp_isr, DEVICE_DT_INST_GET(inst), 0);			\
-		irq_enable(DT_INST_IRQN(inst));							\
+		irq_enable(DT_INST_IRQ_BY_NAME(inst, dcmipp, irq));				\
+		STM32_DCMIPP_CSI_IRQ_INIT(inst)							\
 	}											\
 												\
 	static struct stm32_dcmipp_data stm32_dcmipp_data_##inst = {				\
--- a/dts/arm/st/n6/stm32n6.dtsi
+++ b/dts/arm/st/n6/stm32n6.dtsi
@@ -776,7 +776,8 @@
 			clocks = <&rcc STM32_CLOCK(APB5, 2)>,
 				 <&rcc STM32_SRC_IC17 DCMIPP_SEL(2)>,
 				 <&rcc STM32_CLOCK(APB5, 6)>;
-			interrupts = <48 0>;
+			interrupts = <48 0>, <47 0>;
+			interrupt-names = "dcmipp", "dcmipp_csi";
 			resets = <&rctl STM32_RESET(APB5, 2)>,
 				 <&rctl STM32_RESET(APB5, 6)>;
 			status = "disabled";
--- a/dts/bindings/video/st,stm32-dcmipp.yaml
+++ b/dts/bindings/video/st,stm32-dcmipp.yaml
@@ -49,6 +49,10 @@ properties:
   interrupts:
     required: true
 
+  interrupt-names:
+    required: true
+    enum: ["dcmipp", "dcmipp_csi"]
+
   resets:
     required: true
 
--- a/dts/arm/st/h7rs/stm32h7rs.dtsi
+++ b/dts/arm/st/h7rs/stm32h7rs.dtsi
@@ -1026,6 +1026,7 @@
 				 <&rcc STM32_SRC_PCLK5 NO_SEL>;
 			clock-names = "dcmipp", "dcmipp-ker";
 			interrupts = <95 0>;
+			interrupt-names = "dcmipp";
 			resets = <&rctl STM32_RESET(APB5, 2)>;
 			status = "disabled";
 
--- a/dts/arm/st/mp13/stm32mp135.dtsi
+++ b/dts/arm/st/mp13/stm32mp135.dtsi
@@ -15,6 +15,7 @@
 			compatible = "st,stm32-dcmipp";
 			reg = <0x5a000000 0x400>;
 			interrupts = <GIC_SPI 79 IRQ_TYPE_LEVEL IRQ_DEFAULT_PRIORITY>;
+			interrupt-names = "dcmipp";
 			clocks = <&rcc STM32_CLOCK(APB4, 1)>,
 				 <&rcc STM32_SRC_PLL2_Q DCMIPP_SEL(1)>;
 			clock-names = "dcmipp", "dcmipp-ker";


drivers/video/video_stm32_dcmipp.c
```diff
@@ -1847,9 +1862,10 @@ static void stm32_dcmipp_isr(const struct device *dev)
 #define STM32_DCMIPP_INIT(inst)									\
 	static void stm32_dcmipp_irq_config_##inst(const struct device *dev)			\
 	{											\
-		IRQ_CONNECT(DT_INST_IRQN(inst), DT_INST_IRQ(inst, priority),			\
+		IRQ_CONNECT(DT_INST_IRQN_BY_IDX(inst, 0), DT_INST_IRQ_BY_IDX(inst, 0, priority),\
```
etienne-lms: If several interrupts are used, prefer using `interrupt-names` property.

drivers/video/video_stm32_dcmipp.c
```diff
@@ -1832,8 +1841,14 @@ static void stm32_dcmipp_isr(const struct device *dev)
 					    (0),						\
 					    (DT_PROP_BY_IDX(DT_INST_ENDPOINT_BY_ID(inst, 0, 0),	\
 							    data_lanes, 1))),
+
+#define STM32_DCMIPP_CSI_IRQ_INIT(inst)								\
+	IRQ_CONNECT(DT_INST_IRQN_BY_IDX(inst, 1), DT_INST_IRQ_BY_IDX(inst, 0, priority),	\
+		stm32_dcmipp_csi_isr, DEVICE_DT_INST_GET(inst), 0);				\
```
etienne-lms: Nitcpikcing: prefer indent to the opening parenthesis.

dts/arm/st/n6/stm32n6.dtsi
```diff
@@ -756,7 +756,7 @@
 			clocks = <&rcc STM32_CLOCK(APB5, 2)>,
 				 <&rcc STM32_SRC_IC17 DCMIPP_SEL(2)>,
 				 <&rcc STM32_CLOCK(APB5, 6)>;
-			interrupts = <48 0>;
+			interrupts = <48 0>, <47 0>;
```
etienne-lms, line 779: If several interrupts are used, prefer using `interrupt-names` property to identify them.
(and use `DT_DRV_IRQ_BY_NAME()` in `STM32_DCMIPP_INIT()` and `STM32_DCMIPP_CSI_IRQ_INIT()`)

etienne-lms (commented):
Could you refine the commit message?
```patch
 soc: st: stm32n6: add DCMIPP CSI IRQ
 
-The STM32N6 devicetree is missing the CSI IRQ of the DCMIPP. Add it and pass it through to the ST HAL.
+Add DCMIPP CSI IRQ support for STM32N6.
 
 Signed-off-by: Anton Puppe <anton-noel-flynn.puppe@zeiss.com>
-on-behalf-of: @ZEISS anton-noel-flynn.puppe@zeiss.com

File: dts/arm/st/n6/stm32n6.dtsi (hunk 3 above)
Comment (not-na) [reply dts/arm/st/n6/stm32n6.dtsi:779]: I did not know about interrupt-names, it seems useful here. If I change the main DCMIPP IRQ to e.g. `DT_INST_IRQN_BY_NAME(inst, dcmipp)`, I also have to add `interrupt-names` to all other DTs using the DCMIPP driver, right? So that would be the STM32MP135.

Alternatively, I would just leave the main IRQ unchanged, basically assuming it's always the first. From what I can tell, the STM32 LTDC driver seems to do it like that.

Comment (mathieuchopstm) [reply dts/arm/st/n6/stm32n6.dtsi:779]: If `interrupt-names` is used - and it should be used here - then it must be used for all interrupts.

Comment (etienne-lms) [reply dts/arm/st/n6/stm32n6.dtsi:779]: I agree, including in *stm32mp13.dtsi*. `interrupt-names` property should be required in **st,stm32-dcmipp.yaml**.

Comment (not-na) [reply dts/arm/st/n6/stm32n6.dtsi:779]: OK. I've added `interrupt-names` to *stm32mp135.dtsi* and made it mandatory in *st,stm32-dcmipp.yaml*.

**Review by etienne-lms — COMMENTED** [review]
Nitpicking: could you remove `on-behalf-of: @ZEISS anton-noel-flynn.puppe@zeiss.com| from the commit message. The commit message body is expected to end with valid tag which `on-behalf-of` is not AFAICT.
Otherewise LGTM.

**Comment by not-na** [comment]
OK. I was following the [GitHub docs for this](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-on-behalf-of-an-organization), but I've removed it, since its the same E-Mail address anyways.

**Review by etienne-lms — DISMISSED** [review]
> OK. I was following the [GitHub docs for this](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-on-behalf-of-an-organization), but I've removed it, since its the same E-Mail address anyways.

My mistake. I did not know about it. I don't think it's used in Zephyr scope.

**Comment by avolmat-st** [comment]
> > Hi @not-na,
> > thanks for this patch. Could you give us some example why you need to get notifications of the CSI events ? I tend to go check them usually when there are issues with the sensor, phy configuration and want to see if there are activty at the CSI level but apart of that I haven't used it much so far.
> 
> @avolmat-st Getting notifications of CSI events allows more precise timing alignment to the sensor scan itself. This can be useful if you want to synchronize e.g. a flash to the camera. It could also be done based on the DCMIPP IRQ, but that has more jitter and depends on memory access speeds etc. Given that the CSI IRQ is easy enough to use, it seems like the logical choice for such tasks.

@not-na thanks. I understand, indeed, getting the SOF / EOF as early as possible in the pipeline seems a good choice. By the way, since you refer to memory access speed, I think you were referring probably to DCMIPP end of pipeline events when a frame has been completed, but the DCMIPP itself has also the VSYNC / HSYNC events per pipe so that could also be a possibility to capture those.
Anyway, capturing them from CSI is fine too I think.

However, I am wondering if such interrupt should always be enabled / connected, I mean, while this can be useful for some use-cases, probably not all use-cases needs to get those SOF / EOF events, especially since this leads to getting 2 extra interrupts per frame.
For example maybe we could think about making this 2nd interrupt optional, allowing to only add it into the node if the application is really interested in receiving those events ?

**Comment by not-na** [comment]
> However, I am wondering if such interrupt should always be enabled / connected, I mean, while this can be useful for some use-cases, probably not all use-cases needs to get those SOF / EOF events, especially since this leads to getting 2 extra interrupts per frame.
For example maybe we could think about making this 2nd interrupt optional, allowing to only add it into the node if the application is really interested in receiving those events ?

@avolmat-st I agree, extra interrupts that aren't needed should be avoided. What do you think the best approach to making this configurable is? I could remove it from the DT by default and detect if the extra IRQ is present, but that would require users to figure out the IRQ number if they want to use it. I could also keep the `IRQ_CONNECT` but leave out the `irq_enable`, so users would just have to call that.

Though I think the easiest and most user-friendly way is probably via Kconfig, e.g. have a `VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ` or similar.

What do you think?

**Comment by avolmat-st** [comment]
> > However, I am wondering if such interrupt should always be enabled / connected, I mean, while this can be useful for some use-cases, probably not all use-cases needs to get those SOF / EOF events, especially since this leads to getting 2 extra interrupts per frame.
> > For example maybe we could think about making this 2nd interrupt optional, allowing to only add it into the node if the application is really interested in receiving those events ?
> 
> @avolmat-st I agree, extra interrupts that aren't needed should be avoided. What do you think the best approach to making this configurable is? I could remove it from the DT by default and detect if the extra IRQ is present, but that would require users to figure out the IRQ number if they want to use it. I could also keep the `IRQ_CONNECT` but leave out the `irq_enable`, so users would just have to call that.
> 
> Though I think the easiest and most user-friendly way is probably via Kconfig, e.g. have a `VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ` or similar.
> 
> What do you think?

Yeah I originally thought about the same thing, to have the 2nd interrupt optional but as you said it is not very user friendly to have to figure out the number of the 2nd IT.
Good idea to have an child Kconfig option of the DCMIPP I think, very user-friendly, Thanks.

File: drivers/video/Kconfig.stm32_dcmipp [hunk 4]
```diff
@@ -42,4 +42,13 @@ config VIDEO_STM32_DCMIPP_SENSOR_PIXEL_FORMAT
 	help
 	  Pixel format of the sensor video frame.
 
+# Needs to be manually declared since its not auto-generated
+DT_COMPAT_ST_STM32N6_DCMIPP := st,stm32n6-dcmipp
```
erwango: This should be avoided.
Actual issue here is that `st,stm32n6-dcmipp` is used but not defined. It should rather be removed.

Then, what you actually need to do is to check `dcmipp_csi` irq presence name.
This won't be easier, but that's the usual way to proceed.

erwango (changes requested):

drivers/video/Kconfig.stm32_dcmipp
avolmat-st: > This should be avoided. Actual issue here is that `st,stm32n6-dcmipp` is used but not defined. It should rather be removed.

Yes, sorry about that, this is part of another PR where soc specifc compatible for the dcmipp are removed and only st,stm32-dcmipp is kept.

not-na: OK. I did it like that because the DCMIPP driver already uses the `st,stm32n6-dcmipp` compatible for detection if CSI is supported:
https://github.com/zephyrproject-rtos/zephyr/blob/458e6f8ae3df3a182782de2349e832be8bdf1cd1/drivers/video/video_stm32_dcmipp.c#L33-L36

However, I'm not sure how to get the proper condition for showing the Kconfig option otherwise, as the [Kconfig preprocessor functions](https://docs.zephyrproject.org/latest/build/kconfig/preprocessor-functions.html#devicetree-related-functions) don't seem to have a way to check if a string value is present in a string array. I'm not too familiar with the internals of the Kconfig processing pipeline, so maybe there is some other way to get access to this condition?

not-na: I looked into this a bit more and am now pretty sure that this would need a new Kconfig preprocessor function. Probably something like `dt_compat_any_array_prop_has_value` which returns `y` if any compatible has an array prop containing a specific value (in this case, the string `dcmipp_csi`).

Any thoughts on the name or some simpler approach? I'll try to add this in the next few days.

avolmat-st: While this isn't yet reviewed, I changed this code as part of PR #100242

cf commit https://github.com/zephyrproject-rtos/zephyr/pull/100242/commits/ab86194264836448de6e3188c1cb2cbf70063f43

This approach makes that there is now only a unique compatible for the dcmipp and the driver knows what has to be enabled by looking at what is exposed by lower layers.

The other approach would be to keep and describe the stm32n6 specific compatible and introduced new ones for all socs which have dcmipp.

not-na: OK. So there will only be a `st,stm32-dcmipp` compatible, regardless of whether the SoC has CSI support, correct?

What do you mean by looking at what is exposed by lower layers? The HAL API / Macros?

The problem I have here is that only a few, very specific devicetree properties are accessible from Kconfig to control visibility of the new Kconfig option (which should only be visible if the DCMIPP supports CSI). Ideally I would want to hook into something like the `STM32_DCMIPP_HAS_CSI` macro defined [here](https://github.com/zephyrproject-rtos/zephyr/pull/100242/changes#diff-41261b8df72847d560971c21aec3e0431912df38e5ed5a791a1b9633ac38cbc0R33), but I don't think it's possible or intended to get access to macros defined in C code from Kconfig. Where does the `DCMIPP_SERIAL_MODE` just above come from? I couldn't find where it is supposed to be defined.

Alternatively, I could go the route of a new Kconfig preprocessor function as described above, but that also requires all DTs using the DCMIPP with CSI to give the correct name for their interrupt names. I can ensure this for this PR, but am not sure how to enforce this for any future SoCs with DCMIPP+CSI support.

erwango: `DCMIPP_SERIAL_MODE` comes from HAL.

not-na: I've changed this to use `dt_compat_any_has_prop`, which seems to have support for array props. This removes the dependency on the removed `st,stm32n6-dcmipp` compatible.

drivers/video/Kconfig.stm32_dcmipp
```diff
@@ -42,4 +42,10 @@ config VIDEO_STM32_DCMIPP_SENSOR_PIXEL_FORMAT
 	help
 	  Pixel format of the sensor video frame.
 
+config VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ
+	bool "CSI Interrupt"
+	depends on $(dt_compat_any_has_prop,$(DT_COMPAT_ST_STM32_DCMIPP),interrupt-names,dcmipp_csi)
+	help
+	  Enable the secondary DCMIPP interrupt dedicated to the CSI host.
```
mathieuchopstm: Maybe indicate what this implies? (e.g., `HAL callbacks HAL_DCMIPP_CSI_xxx will be invoked and can be implemented by application`?)

mathieuchopstm (commented):
otherwise LGTM

drivers/video/Kconfig.stm32_dcmipp
not-na: Makes sense. Added in latest push.

mathieuchopstm (dismissed):

avolmat-st (dismissed):
LGTM. Thanks !!

dts/bindings/video/st,stm32-dcmipp.yaml
```diff
@@ -49,6 +49,9 @@ properties:
   interrupts:
     required: true
 
+  interrupt-names:
+    required: true
```
etienne-lms, line 53: Since names are expected to be `dcmipp` and `dcmipp_csi`, I would suggest:
```
    enum: ["dcmipp", "dcmipp_csi"]

**Review by etienne-lms — DISMISSED** [review]
LGTM with a comment

File: drivers/video/Kconfig.stm32_dcmipp [hunk 7]
```diff
@@ -42,4 +42,11 @@ config VIDEO_STM32_DCMIPP_SENSOR_PIXEL_FORMAT
 	help
 	  Pixel format of the sensor video frame.
 
+config VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ
+	bool "CSI Interrupt"
```
erwango, line 46: Question here:

Do we expect user to be able to modify this symbol at will ?
For instance use `CONFIG_VIDEO_STM32_DCMIPP_CSI_ENABLE_IRQ=n` at application/board level.

If not expected prompt (`"CSI Interrupt"`) should be removed.

erwango (commented):
Otherwise LGTM

drivers/video/Kconfig.stm32_dcmipp
not-na: Yes, it's expected to be modified. The default is `n`, so the symbol will never activate by itself. I would expect most modifications to occur at the application level, where there may be uses for the extra interrupt.

dts/bindings/video/st,stm32-dcmipp.yaml
not-na: Ah, I didn't know about that feature. I've added it in the latest push.

etienne-lms (approved):

avolmat-st (approved):

erwango (approved):

dts/bindings/video/st,stm32-dcmipp.yaml
```diff
@@ -49,6 +49,10 @@ properties:
...
+    enum: ["dcmipp", "dcmipp_csi"]
```
mathieuchopstm, line 54:
```suggestion
    enum: ["dcmipp", "dcmipp-csi"]
```
\+ should update DTSI files accordingly

mathieuchopstm (approved):
Overlooked this during last review but IRQ name should use `-` instead of `_`... I guess we can avoid another approval round and live with the underscore though.

PR #101137, opened 2025-12-16
drivers: mspi_dw: Edit op-mode setting to be after ctrl0 clearing

Hi-Im-David:
The op-mode mode ctrlr0 setting was being cleared
in the start_next_packet function. This would
only affect MSPI_DW peripherals that can be
configured as a controller or peripheral while
being set as a controller.


swift-tk (approved):

anangl (changes requested):
This PR does not solve the actual problem. See #101213.

PR #101138, opened 2025-12-16
drivers: wifi: Fix net API use

aescolar:
In 55c49cdb8f7e93b018f908fc8356212cc0e42da8 wifi drivers were changed to use the Zephyr native net_ prefixed
types, but some were forgotten.
Without this fix/change the code still builds as we are by now setting CONFIG_NET_NAMESPACE_COMPAT_MODE. But when this is not set, things fail to build.

Notes: 
* The esp_at , eswifi and simplelink drivers have been test built with samples/net/wifi/shell/ setting `CONFIG_NET_NAMESPACE_COMPAT_MODE=n`.
* For the NXP driver, the HAL also uses the old/non-native names, so the HAL would also need changing to be able to build with `CONFIG_NET_NAMESPACE_COMPAT_MODE=n`
* For the winc1500 the Atmel API has not been touched. Just the API towards Zephyr.

### DIFF
--- a/drivers/wifi/esp_at/esp.c
+++ b/drivers/wifi/esp_at/esp.c
@@ -575,7 +575,7 @@ MODEM_CMD_DEFINE(on_cmd_cipdns)
 		}
 
 		addrs[i].sin_family = NET_AF_INET;
-		addrs[i].sin_port = htons(53);
+		addrs[i].sin_port = net_htons(53);
 
 		valid_servers++;
 	}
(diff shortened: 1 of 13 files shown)


loicpoulain (approved):
LGTM

krish2718 (approved):

rajb9 (approved):

MaochenWang1 (approved):

jukkar (approved):

PR #101139, opened 2025-12-16
drivers: entropy: virtio

julek-wolfssl:
The device is not able to place the entropy into stack or heap allocated buffers. This uses a bounce buffer to be able to use any buffer with virtio entropy.

### DIFF
--- a/drivers/entropy/entropy_virtio.c
+++ b/drivers/entropy/entropy_virtio.c
@@ -19,9 +19,13 @@ struct entropy_virtio_config {
 	const struct device *vdev;
 };
 
+#define ENTROPY_BUFFER_MAX_LEN 256
+
 struct entropy_virtio_data {
 	struct k_sem sem;
 	uint32_t received_len;
+	struct k_mutex buf_mutex;
+	uint8_t buf[ENTROPY_BUFFER_MAX_LEN];
 };
 
 static void entropy_virtio_virtq_recv_cb(void *priv, uint32_t len)
@@ -43,7 +47,6 @@ static uint16_t entropy_virtio_enum_queues_cb(uint16_t q_index, uint16_t q_size_
 
 static int entropy_virtio_get_entropy(const struct device *dev, uint8_t *buffer, uint16_t length)
 {
-	struct virtq_buf buf[] = {{.addr = buffer, .len = length}};
 	const struct entropy_virtio_config *cfg = dev->config;
 	struct entropy_virtio_data *data = dev->data;
 	struct virtq *vq = virtio_get_virtqueue(cfg->vdev, VIRTIO_ENTROPY_QUEUE_IDX);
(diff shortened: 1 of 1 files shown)


drivers/entropy/entropy_virtio.c
```diff
@@ -53,21 +56,33 @@ static int entropy_virtio_get_entropy(const struct device *dev, uint8_t *buffer,
...
+    	if (data->received_len != buf[0].len) {
+    		LOG_ERR("insufficient number of values: %d/%d", data->received_len, buf[0].len);
+    		k_mutex_unlock(&data->buf_mutex);
+    		return -EIO;
+    	}
+    
+    	memcpy(buffer, data->buf, data->received_len);
+    	k_mutex_unlock(&data->buf_mutex);
+        buffer += data->received_len;
+        length -= data->received_len;
```
tomi-font: indentation is wrong

tomi-font (dismissed):
Please fix the commit message line length and make the commit title more explicit (`drivers: entropy: virtio: <short summary>`).

tomi-font (commented):
You still need to fix all the compliance issues: https://github.com/zephyrproject-rtos/zephyr/actions/runs/20342714010/job/58481495133?pr=101139

drivers/entropy/entropy_virtio.c
```diff
@@ -54,20 +57,35 @@ static int entropy_virtio_get_entropy(const struct device *dev, uint8_t *buffer,
...
-	k_sem_take(&data->sem, K_FOREVER);
-
-	if (data->received_len != length) {
-		LOG_ERR("insufficient number of values: %d/%d", data->received_len, length);
-		return -EIO;
+	while (length > 0) {
+		struct virtq_buf buf[] = {
+			{.addr = data->buf, .len = MIN(length, ENTROPY_BUFFER_MAX_LEN)}};
+
+		data->received_len = 0;
```
ceolin: You should lock the mutex before this assignment and release it after line 88 otherwise possible overflow can happens if `data->received_len` is assigned to zero in between lines 87 and 88.

ceolin (changes requested):
It looks good to me, just a comment regarding mutex usage. Please check if it makes sense.

drivers/entropy/entropy_virtio.c
julek-wolfssl: I think you're right. I pushed a fix to hold the mutex while `data->received_len` is being used.

ceolin (approved):
Thank you !

drivers/entropy/entropy_virtio.c
```diff
@@ -83,9 +83,9 @@ static int entropy_virtio_get_entropy(const struct device *dev, uint8_t *buffer,
 		}
 
 		memcpy(buffer, data->buf, data->received_len);
-		k_mutex_unlock(&data->buf_mutex);
 		buffer += data->received_len;
 		length -= data->received_len;
+		k_mutex_unlock(&data->buf_mutex);
```
tomi-font, line 88: IMO you should just squash this commit into the first one, what you're fixing here is introduced by the previous commit.

julek-wolfssl: Squashed and force-pushed.

tomi-font (approved):

PR #101141, opened 2025-12-16
riscv: pmp: expose pmp_decode_region in public header

fsammoura1980:
Moves pmp_decode_region from pmp.c to pmp.h and removes the conditional static scoping to allow use by other subsystems.


dcpleung:
I think it would be a good idea to prefix the function with `arch_riscv_` since it is now in a public header.

PR #101143, opened 2025-12-16
drivers: pinctrl: sam: Add PIO alternate functions

### LINKED ISSUE #95353: atmel: samv71q21b pinctrl_sam improperly handles pins with "extra" or "alternate" function selections
### Describe the bug

I've been working on a project which involves bring up of a custom board which leverages the samv71q21b soc and during basic IO functionality testing it was discovered that two IO's which were configured as analog inputs appeared to be driving line, seen here in the following scope captures:

AFEC1_AD8 appeared as a being driven low (except for when the adc is being acquired):
Image

And AFEC0_AD5 appears driven high:

Image

This is despite the fact that I applied the following overlay configuration for my samv71 xult xplained board:

```dts
&pinctrl {
    afec0_default: afec0_default {
        group1 {
            pinmux = <PB2X_AFE0_AD5>;   // IO13
        };
    };

    afec1_default: afec1_default {
        group1 {
            // NOTE these are validated against the atmel module in
            // /atmel/include/dt-bindings-pinctrl/samv71q-pinctrl
            pinmux =<PC27X_AFE1_AD8>;  // IO0
        };
    };

};

&afec0 {
    status = "okay";
    #address-cells = <1>;
    #size-cells = <0>;
    pinctrl-0 = <&afec0_default>;
    pinctrl-names = "default";

    channel@5 {
        reg = <5>;
        zephyr,gain = "ADC_GAIN_1";
        zephyr,reference = "ADC_REF_EXTERNAL0";
        zephyr,vref-mv = <3300>;
        zephyr,acquisition-time = <ADC_ACQ_TIME_DEFAULT>;
        zephyr,resolution = <12>;
    };

};

&afec1 {
    status = "okay";
    #address-cells = <1>;
    #size-cells = <0>;
    pinctrl-0 = <&afec1_default>;
    pinctrl-names = "default";

    channel@8 {
        reg = <0x8>;
        zephyr,gain = "ADC_GAIN_1";
        zephyr,reference = "ADC_REF_EXTERNAL0";
        zephyr,vref-mv = <3300>;
        zephyr,acquisition-time = <ADC_ACQ_TIME_DEFAULT>;
        zephyr,resolution = <12>;
    };

};
```

I was able to validate that the adc channels were reading voltages accurately by applying a known voltage source, at this time it was suspected that the pin may be improperly mux'd to select some peripheral functionality that was driving the line. To investigate this I referred to [the datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-E70-S70-V70-V71-Family-Data-Sheet-DS60001527.pdf) and inspected various status registers most notably the lines enabling output/pullups/pulldowns as seen in the following logic diagram:

Image

I was able to determine that according to the PIO controller that the pin was not configured as an output and peripheral control was not enabled. 

Note that this is pretty concerning because the software itself provides no indication that this is an issue and if left uncheck this could result in damaging sensors by unexpectedly driving them high.

### Regression

- [ ] This is a regression.

### Steps to reproduce

In order to reliably reproduce the error and validate it's origin as software I created a simple build which read from AFEC0_AD5 and deployed to both the atmel samv71 xult xplained dev board  and a custom board with the same soc. Furthermore I was able to produce a software build with Atmel's Mlab X IDE that read from the AFEC0_AD5 without driving the pin high.

### Relevant log output

```shell

```

### Impact

### Environment

- OS: Linux (Building in a docker image running Ubuntu 22.04.5 LTS)
- Zephyr version:

```yaml
manifest:
  version: 0.12
  self:
    path: .
  projects:
    - name: zephyr
      # This ensures west has access to zephyrs extensions
      west-commands: scripts/west-commands.yml
      url: https://github.com/zephyrproject-rtos/zephyr
      revision: v4.1-branch
      clone-depth: 1
      import:
        path-prefix: ./deps
        name-allowlist:
          - hal_atmel
          - cmsis
          - cmsis_6
```

### Additional Context

Note that I traced the source of the error to [soc_gpio_configure](https://github.com/zephyrproject-rtos/zephyr/blob/413b789deb391d3a37d06b463288a5fe765ee57e/soc/atmel/sam/common/soc_gpio.c#L116)

Specifically the line appears high after the common attributes are applied and the ```SOC_GPIO_FUNC_A``` case is selected and the ```PIO_PDR ``` mask is set.

This led me to compare how [pinctrl_configure_pin](https://github.com/zephyrproject-rtos/zephyr/blob/413b789deb391d3a37d06b463288a5fe765ee57e/drivers/pinctrl/pinctrl_sam.c#L53) was demuxing the vendor provided [pinctrl_cfg for AFEC0_AD5](https://github.com/zephyrproject-rtos/hal_atmel/blob/065e57c5013051c8b7f2256271349c6942bd9344/include/dt-bindings/pinctrl/samv71q-pinctrl.h#L730)

As compared to the logical description of the related pin:

Image

Image

Image

Image

This to me indicates that the source of the error is how we are interpreting the pin configuration for this desired functionality.

The data sheet indicates that the EBI takes over control of this pin when the "extra" or "alternate" functionality is desired and so its assumed that the only configuration relevant to pinctrl is the enabling/disabling of pull-ups or pull-downs. While this is essentially correct and the current implementation results in enabling the AFEC functionality there exists a side effect in that because no mux selection is specified by ```PB2X_AFE0_AD5``` the assumed selection is "A" which in this case will result in the routing of the pin to the ```CANTX0``` line which is pull-up high (note this is by default at a hardware level even when can is not built in).

This means that whenever we are not leveraging the EBI to utilize the pin's ```AFEC``` functionality it appears driven high for PB2, similarly it appears driven low for PC27 (which by default is a dac output).

Note that I've implemented [a workaround available on my fork ](https://github.com/ReggieMarr/zephyr/commit/cef88da40d2b21d88a1e90f1d94b2122cdd2e717) which sets a pio functionality selection that is electrically compatible with the pin in the case that it is one with its "extra" or "alternate"  functionality selected.

This unblocks me for my current task however is likely not to be considered a generalized solution as it involves some brittle/hard coded logic. Ideally this would be fixed either at the driver level or by otherwise parsing the pinctrl config with more context.

zebo9x:
The PIO alternate function is not correctly muxed. More peripheral function has been introduced (E, F, G, H) to enable the extra (E), system (F), lpm (G), and wakeup (H) alternate function. This approach has been choosen to be consistent with the current implementation. Tested the extra function (AFEC0, AFEC1) with SAM V71B: working.

Fixes zephyrproject-rtos/zephyr#95353

The system, lpm, and wakeup alternate functions behaviour are assumed to be like the extra function.

### DIFF
--- a/drivers/pinctrl/pinctrl_sam.c
+++ b/drivers/pinctrl/pinctrl_sam.c
@@ -68,9 +68,24 @@ static void pinctrl_configure_pin(pinctrl_soc_pin_t pin)
 	soc_pin.mask = 1 << SAM_PINMUX_PIN_GET(pin);
 	soc_pin.flags = SAM_PINCTRL_FLAGS_GET(pin) << SOC_GPIO_FLAGS_POS;
 
-	if (port_func == SAM_PINMUX_FUNC_periph) {
-		soc_pin.flags |= (SAM_PINMUX_PERIPH_GET(pin)
-				  << SOC_GPIO_FUNC_POS);
+	switch(port_func)
+	{
+		case SAM_PINMUX_FUNC_periph:
+			soc_pin.flags |= (SAM_PINMUX_PERIPH_GET(pin)
+					<< SOC_GPIO_FUNC_POS);
+			break;
+		case SAM_PINMUX_FUNC_extra:
+			soc_pin.flags |= SOC_GPIO_FUNC_E;
+			break;
+		case SAM_PINMUX_FUNC_system:
+			soc_pin.flags |= SOC_GPIO_FUNC_F;
+			break;
+		case SAM_PINMUX_FUNC_lpm:
+			soc_pin.flags |= SOC_GPIO_FUNC_G;
+			break;
+		case SAM_PINMUX_FUNC_wakeup:
+			soc_pin.flags |= SOC_GPIO_FUNC_H;
+			break;
 	}
 
 	soc_gpio_configure(&soc_pin);
--- a/soc/atmel/sam/common/soc_gpio.c
+++ b/soc/atmel/sam/common/soc_gpio.c
@@ -163,6 +163,26 @@ void soc_gpio_configure(const struct soc_gpio_pin *pin)
 		break;
 #endif
 
+	case SOC_GPIO_FUNC_E: // extra
+	case SOC_GPIO_FUNC_F: // system
+	case SOC_GPIO_FUNC_G: // lpm
+	case SOC_GPIO_FUNC_H: // wakeup
+		/* Disable the interrupt. */
+		pio->PIO_IDR = mask;
+		/* Disable pull-up. */
+		pio->PIO_PUDR = mask;
+#if defined(CONFIG_SOC_SERIES_SAM4S) || \
+	defined(CONFIG_SOC_SERIES_SAM4E) || \
+	defined(CONFIG_SOC_SERIES_SAMX7X)
+		/* Disable pull-down. */
+		pio->PIO_PPDDR = mask;
+#endif
+		/* Let the PIO control the pin (instead of a peripheral). */
+		pio->PIO_PER = mask;
+		/* Disable output. */
+		pio->PIO_ODR = mask;
+		break;
+
 	case SOC_GPIO_FUNC_IN:
 		/* Enable module's clock */
 		soc_pmc_peripheral_enable(periph_id);


zebo9x:
I do not use AFEC and PIO interrupt for now, so I'm not sure about this line:
https://github.com/zebo9x/zephyr/commit/acf0f46b1f4c8889c51cf351f2f6124a49bcd98c#diff-8fb735ee5c6fd8dc9aeeacfd7f8176d69d39cece6b1b7bbd419863c015578e03R171

nandojve:
Hi @zebo9x ,

Thank you for your proposal but unfortunately this go on a wrong direction. We can not mix peripheral functions with possible properties. I'm still waiting an answer in the issue. So, if my proposal is OK, the only thing that we need to do is add the correct property in the pin on devicetree.

Let me know your inputs here:
https://github.com/zephyrproject-rtos/zephyr/issues/95353#issuecomment-3591789335

PR #101146, opened 2025-12-16
modules: hal_infineon: CMakeLists.txt

billwatersiii:
Only compile the mtb-hal-cat1 module's source when CONFIG_USE_INFINEON_LEGACY_HAL is set.

### DIFF
--- a/modules/hal_infineon/serial-flash/CMakeLists.txt
+++ b/modules/hal_infineon/serial-flash/CMakeLists.txt
@@ -7,4 +7,4 @@ set(serial_flash_dir ${ZEPHYR_HAL_INFINEON_MODULE_DIR}/serial-flash)
 
 zephyr_include_directories(${serial_flash_dir})
 
-zephyr_library_sources(${serial_flash_dir}/cy_serial_flash_qspi.c)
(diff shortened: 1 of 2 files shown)


modules/hal_infineon/serial-flash/CMakeLists.txt
```diff
@@ -7,4 +7,4 @@ set(serial_flash_dir ${ZEPHYR_HAL_INFINEON_MODULE_DIR}/serial-flash)
 
 zephyr_include_directories(${serial_flash_dir})
 
-zephyr_library_sources(${serial_flash_dir}/cy_serial_flash_qspi.c)
+zephyr_library_sources_ifdef(CONFIG_USE_INFINEON_SMIF ${serial_flash_dir}/cy_serial_flash_qspi.c)
```
teburd, line 10: change here seems unrelated to the commit message, fine to include but deserves a note in the message

billwatersiii: I thought about that, and then forgot to add it... will do...

teburd (approved):

sreeramIfx (approved):

PR #101148, opened 2025-12-16
ztest: Benchmarking framework

### LINKED ISSUE #101149: ztest benchmarking framework
### Problem Description

Zephyr's testsuite (ztest) lacks the ability to measure code performance (benchmarking).
Existing performance measurement often relies on ad-hoc timing solutions, which lack
consistency, precision, and standard statistical analysis (mean, standard deviation, standard error).

This deficiency makes it difficult for the zephyr community to:
* Measure the impact of code changes on performance.
* Compare the performance of different implementations & hardware rigorously.
* Track performance advancements/regressions over time.
* Get statistically meaningful data, especially on systems with variable cycle counts, interrupts,
    and architectural differences.
* Have meaningful discussions about performance optimizations within the Zephyr community.

A dedicated benchmarking framework is needed to elevate performance analysis within the Zephyr
ecosystem and provide developers with the tools necessary to make informed decisions based on
reliable performance data.

### Proposed Change (Summary)

To introduce a ztest benchmarking framework that leverages Zephyrs timing subsystem
(`zephyr/timing/timing.h`) to accurately measure the execution time (in CPU cycles) of clearly
defined code blocks.

This framework will provide:
* Standardized API: Macros for grouping and defining benchmark tests.
* Statistical Analysis: Built-in computation of common metrics, including Mean, Standard Deviation,
    Standard Error, and minimum/maximum values.
* Clear Output: Formatted console output of benchmark results, making performance data easy to interpret.

## Macros

* `ZTEST_BENCHMARK_TESTSUITE(suite_name, setup_fn, teardown_fn)`: Defines a benchmark test suite
    that can contain multiple benchmark test cases. The optional `setup_fn` and `teardown_fn` are
    functions that run once before and after all tests in the suite, respectively.

* `ZTEST_BENCHMARK_SETUP_TEARDOWN(suite_name, test_name, iterations, setup_fn, teardown_fn)`:
    Defines a benchmark test case with setup and teardown code blocks that run before and
    after each iteration, respectively.

* `ZTEST_BENCHMARK(suite_name, test_name, iterations)`: short-hand version of the above macro
    without setup and teardown functions.

### Sample usage

The sample is taken from the accompanying draft PR(#101148)[https://github.com/zephyrproject-rtos/zephyr/pull/101148].
```c
static void suit_setup(void)
{
	LOG_DBG("Setting up math benchmarks suite");
}

static void suit_teardown(void)
{
	LOG_DBG("Tearing down math benchmarks suite");
}

ZTEST_BENCHMARK_TESTSUITE(math_benchmarks, suit_setup, suit_teardown);

ZTEST_BENCHMARK(math_benchmarks, void_test, 100)
{
	/* Intentionally left blank */
}

ZTEST_BENCHMARK(math_benchmarks, benchmark_addition, 100)
{
	int a = 5;
	int b = 10;
	int c = 0;

	c += a + b;
}

struct container {
	int a;
	int b;
	int res;
};

static struct container data;

static void setup_fn(void)
{
	struct container *c = (struct container *)&data;
	c->a = 20;
	c->b = 15;
	c->res = 0;
}

static void teardown_fn(void)
{
	struct container *c = (struct container *)&data;

	memset(c, 0, sizeof(*c));
}

ZTEST_BENCHMARK_SETUP_TEARDOWN(math_benchmarks, benchmark_subtraction, 100, setup_fn, teardown_fn)
{
	struct container *c = (struct container *)&data;
	c->res = c->a - c->b;
}

ZTEST_BENCHMARK(math_benchmarks, benchmark_pure_asm_add, 100)
{
	__asm__ volatile (
#if defined(CONFIG_X86)
		"add %eax, %eax \n\t"
#elif defined(CONFIG_ARM)
		"add r0, r0, r0 \n\t"
#elif defined(CONFIG_RISCV)
		"add a0, a0, a0 \n\t"
#else
	#error "Unsupported architecture for pure assembly benchmark"
#endif
	);
}
```

### sample output

This output was generated from the above sample code when running on an nrf5340dk board without any
compiler optimizations (`CONFIG_NO_OPTIMIZATIONS=y`) enabled for demonstration purposes.

The actual cycle counts will vary based on the target hardware and optimization level.

```txt
math_benchmarks ############################################
[00:00:00.288,787] <dbg> test_benchmarks: suit_setup: Setting up math benchmarks suite
benchmark_addition =========================================
	Sample size:100, total cycles: 1621
	Mean(μ̂): 16.202
	Standard deviation(s): 2.102
	Standard Error(SE): 0.210
	Min: 16 (run #4)
	Max: 37 (run #1)
benchmark_pure_asm_add =====================================
	Sample size:100, total cycles: 122
	Mean(μ̂): 1.212
	Standard deviation(s): 2.399
	Standard Error(SE): 0.239
	Min: 1 (run #2)
	Max: 25 (run #1)
benchmark_subtraction ======================================
	Sample size:100, total cycles: 1631
	Mean(μ̂): 16.302
	Standard deviation(s): 3.300
	Standard Error(SE): 0.330
	Min: 16 (run #2)
	Max: 49 (run #1)
void_test ==================================================
	Sample size:100, total cycles: 30
	Mean(μ̂): 0.292
	Standard deviation(s): 2.810
	Standard Error(SE): 0.281
	Min: 0 (run #4)
	Max: 28 (run #1)
[00:00:00.392,150] <dbg> test_benchmarks: suit_teardown: Tearing down math benchmarks suite
```

### Dependencies

* Zephyr Timing Subsystem: The framework relies on the `<zephyr/timing/timing.h>` API for
    accurate, low-overhead cycle counting. The timing subsystem must be configured and available in
    the target.

* Standard Math Library: Use of `<math.h>` is required for statistical calculations like standard
    deviation and error.

* Overhead of Benchmarking: The framework must ensure that the overhead introduced by timing and
    statistical calculations is minimal and does not skew results significantly. This has been
    partly mitigated using a control test to measure the overhead of the benchmarking framework
    itself. However, even with this mitigation, very fast-executing code blocks may still be
    affected by overhead generated by stack allocations etc. as seen in the `void_test` benchmark &
    the assembly benchmark in the sample output.

* Cache warming: In the proposed implementation, there is no explicit cache warming step before each
    benchmark iteration. This could lead to a slightly higher variance in the results for small sample sizes.
    I've chosen to omit this due to concerns for test integrity, but I welcome feedback on this
    decision.

* Depending on what statistical measures are needed, additional computational overhead may be introduced.
    However, the current set of metrics (mean, standard deviation, standard error) are relatively
    lightweight to compute. If more complex statistics are required in the future, this concern may need
    to be revisited. Since more complex statistics would most likely require *much* more memory to keep
    samples in memory instead of being updated on-the-fly using Welfords algorithm.

* Naming Convention: It is said that the only two hard things in computer science are cache
    invalidation and naming things.
    The proposed macro names (`ZTEST_BENCHMARK`, `ZTEST_BENCHMARK_TESTSUITE`, etc.) are designed to align with existing ztest conventions while clearly indicating their purpose for benchmarking. However, alternative naming schemes could be considered based on community feedback.

* The sample output format could be changed to a more compact version, but I opted for the more
    verbose for the sake of clarity. Feedback on the output format is encouraged as it should be
    easily read by humans and easily parsed by scripts.

* Integration with Twister: While the benchmarking framework is designed to work independently,
    future integration with Zephyrs Twister test runner could enhance the community's ability to track
    performance over time.
    However, this integration is not addressed in this proposal and remains an open question for
    future exploration.

* The current implementation will force developers to either rely on global variables or introduce
    noise in the benchmarked code block by managing setup of stack variables in the measured code
    block itself. I started exploring ways to pass user data pointers to the setup/teardown
    functions and benchmark code blocks, but this resulted in the hidden variables in the test scope
    as the macros pre-expansion would not show the function signatures clearly.
    If there is major pushback from the community on this, I'm open to revisiting this decision.
    However, as I'm writing this, I believe that the current approach wins this trade-off.

### Alternatives Considered

Integrating benchmarking capabilities into the current ztest framework was initially explored, but
this was discarded due to the architectural complexity and potential disruption to existing unit
tests. I concluded that a dedicated, separate benchmarking framework is necessary to maintain a
clear separation of concerns. This allows both ztest and the new benchmark framework to be developed
and iterated upon independently, yielding a cleaner & more maintainable implementation.

Mattemagikern:
Zephyr benchmarking framework. 

closes #101149

### DIFF
--- a/doc/develop/test/benchmark.rst
+++ b/doc/develop/test/benchmark.rst
@@ -1,0 +1,212 @@
+.. _ztest_benchmarking:
+
+Benchmarking Framework
+######################
+
+The Zephyr benchmarking framework provides cycle-accurate performance
+measurements. It automates data collection and statistical calculation, offering
+a standardized way to evaluate execution metrics across the Zephyr ecosystem.
+
+Overview
+********
+
+This framework helps identify regressions and optimize critical paths by
+providing:
+
+* **Standardized API**: Macros that align with existing ``ztest`` conventions.
+* **Statistical Analysis**: Calculations of Mean, Standard Deviation, Standard
+  Error, and Min/Max values.
+* **Overhead Compensation**: Inclusion of a control test to account for the
+  benchmarking frameworks own execution time.
+
+Configuration
+*************
+
+To use the benchmarking framework, you must enable the following Kconfig
+options:
+
+.. code-block:: cfg
+
+   CONFIG_ZTEST=y
+   CONFIG_ZTEST_BENCHMARK=y
+
+Usage
+*****
+
+A benchmark suite is defined similarly to the normal ztest testsuite by first
+defining the suite with ``ZTEST_BENCHMARK_SUITE`` and then adding individual
+benchmark tests to the suite using either ``ZTEST_BENCHMARK`` or ``ZTEST_BENCHMARK_TIMED`` macros.
+
+.. code-block:: c
+
+   #include <zephyr/ztest.h>
+
+   ZTEST_BENCHMARK_SUITE(<test suite name>, <setup_fn>, <teardown_fn>);
+
+
+Standard Benchmarks
+===================
+
+Standard benchmarks are sample-based, meaning they execute the test a specified number
+of times and measure the total cycles taken. This is useful for benchmarking critical paths where
+you want to understand the raw CPU performance in terms of cycles. It provides insights into the
+efficiency of the code and helps identify bottlenecks in terms of CPU usage.
+This benchmarking method is suitable for code that has consistent execution times and is not heavily
+influenced by external factors such as I/O operations or context switches.
+
+.. code-block:: c
+
+   #include <zephyr/ztest.h>
+
+   ZTEST_BENCHMARK_SUITE(<suite name>, NULL, NULL);
+
+   ZTEST_BENCHMARK(<suite name>, <benchmark name>, <number of samples>)
+   {
+       /* Code to benchmark */
+   }
+
+A standard benchmark follows a flow where the setup function is called before each sample, the test
+function is executed for the specified number of samples, and the teardown function is called after
+each sample.
+
+Timed Benchmarks
+================
+
+Timed benchmarks in contrast to the standard benchmarks measures execution time of the code
+instead of cycles. This is useful for benchmarking code that may have variable execution
+times or when you want to measure the actual time taken rather than just CPU cycles of a critical
+path. It provides a broader view of performance characteristics, especially for code that involves
+I/O operations, context switches, or other factors that can influence execution time beyond raw CPU
+performance.
+
+.. code-block:: c
+
+   ZTEST_BENCHMARK_TIMED(<suite name>, <benchmark name>, <time in milliseconds>)
+   {
+         /* Code to benchmark */
+   }
+
+
+Unlike standard benchmarks that favor isolation, Timed Benchmarks executes setup and teardown
+functions only once allowing the test function to run hot within a dedicated time window.
+This will give a more realistic measurements as it includes the overhead of the system as it would
+be in a real-world scenario, such as interrupts, context switches, and other background tasks.
+
+Understanding Results
+*********************
+
+Standard Benchmarking Results
+=============================
+
+.. code-block:: console
+
+   <suite name> ###############################################
+   <benchmark name> ===========================================
+      Sample size:<number of samples>, total cycles: <total amount of cycles>
+      Mean(u): <mean cycles per sample>
+      Standard deviation(s): <cycles>
+      Standard Error(SE): <cycles>
+      Min: <cycles> (run #<sample number>)
+      Max: <cycles> (run #<sample number>)
+
+
+Statistical Metrics
+"""""""""""""""""""
+
+* **Mean (u)**: The average number of cycles taken per sample. It provides a
+  central value representing the expected cost of execution.
+
+* **Standard Deviation (s)**: Measures the amount of variation or fluctuation of
+  the execution cost from the mean. A low standard deviation indicates that the
+  behavior is deterministic and consistent.
+
+* **Standard Error (SE)**: Estimates how far the sample mean is likely to be
+  from the "true" mean of the system. It provides insight into the statistical
+  reliability of the test. A lower SE indicates higher confidence in the
+  result.
+
+* **Min/Max**: The minimum and maximum cycle counts observed, along with which
+  sample they occurred on.
+
+Plainly speaking, lower values are better for all metrics.
+Lower mean, min, and max values indicates better raw performance.
+Lower standard deviation and standard error values indicate more consistent and
+therefore reliable performance.
+
+Timed Benchmarking Results
+==========================
+
+.. code-block:: console
+
+   <benchmark name> ===============================================
+      Samples: <Number of samples executed during the benchmark>
+      Total Time: <Gross execution time>
+      Work Time: <Net execution time> ns (Net)
+      Ops/Sec: <average operations per second>
+      Cycles/Op: <average cycles per operation>
+
+Statistical Metrics
+"""""""""""""""""""
+* **Total Time**: The total time taken for all samples of the benchmark, including overhead.
+
+* **Work Time**: The total time taken for the code under test, excluding the overhead of the
+  benchmarking framework itself. This provides a more accurate measure of the actual performance of
+  the code being benchmarked.
+
+* **Ops/Sec**: The number of operations (samples) that can be performed per second, calculated by
+  dividing the number of samples by the net work time (in seconds). This metric is particularly
+  useful for understanding the throughput of the code being benchmarked.
+
+* **Cycles/Op**: The average number of CPU cycles taken per operation, calculated by dividing the
+  net number of cycles by the number of samples. This metric provides insight into the
+  efficiency of the code in terms of CPU usage.
+
+In general higher values for Ops/Sec are better as it indicates higher throughput, just as lower
+values for Cycles/Op are better. As the Cycles/Op and Ops/Sec metrics are derived from the same
+underlying data, they are reflecting the same performance characteristics from different
+perspectives. A high Ops/Sec should correspond to a low Cycles/Op, and vice versa.
+
+
+Benchmark Output Options
+========================
+
+The benchmarking framework provides several options for outputting results of benchmarking data. By
+default, results are printed in a verbose human-readable format that is easy to understand and
+interpret.
+Alterantively, you can enable the ``CONFIG_ZTEST_BENCHMARK_OUTPUT_CSV`` Kconfig option to output
+results in a CSV format that can be easily imported by scripts for further analysis.
+The CSV output includes all the same metrics as the verbose output, but in a
+format that is more conducive to automated analysis and reporting.
+
+The standard benchmark csv output format is as follows:
+.. code-block:: console
+
+   S,<suite name>,<benchmark name>,<sample size>,<total cycles>,<mean>,<stddev>,<stderr>,<min>,<min sample>,<max>,<max sample>
+
+The timed benchmark csv output format is as follows:
+.. code-block:: console
+
+   T,<suite name>,<benchmark name>,<samples>,<total time>,<work time>,<ops/sec>,<cycles/op>
+
+
+Important Considerations
+************************
+
+* **Noise**: Benchmarking is inherently sensitive to system noise. To obtain as
+  accurate results as possible, disable unnecessary background tasks and
+  interrupts that may interfere with timing.
+
+* **Cache Warming**: The first sample of a benchmark may often be slower
+  due to cache misses and this can skew results slightly on small sample sizes,
+  choose a sufficiently large number of samples to mitigate this effect.
+
+* **Use setup/teardown functions**: It is *highly* encouraged to utilize the
+  setup and teardown functions to isolate the code under test as much as
+  possible. Excessive setup/teardown code in the benchmark code can introduce
+  noise and skew results *significantly* for small critical paths.
+
+
+API Reference
+*************
+
+.. doxygengroup:: ztest_benchmark
--- a/samples/subsys/testsuite/benchmark/prj.conf
+++ b/samples/subsys/testsuite/benchmark/prj.conf
@@ -1,0 +1,9 @@
+# Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+CONFIG_MAIN_STACK_SIZE=4096
+CONFIG_LOG=y
+CONFIG_ZTEST=y
+CONFIG_ZTEST_BENCHMARK=y
+CONFIG_TEST_RANDOM_GENERATOR=y
+CONFIG_NO_OPTIMIZATIONS=y
--- a/samples/subsys/testsuite/benchmark/src/main.c
+++ b/samples/subsys/testsuite/benchmark/src/main.c
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <zephyr/ztest.h>
+
+#include <zephyr/timing/timing.h>
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(test_benchmarks, LOG_LEVEL_DBG);
+
+static void suit_setup(void)
+{
+	LOG_DBG("Setting up math benchmarks suite");
+}
+
+static void suit_teardown(void)
+{
+	LOG_DBG("Tearing down math benchmarks suite");
+}
+
+ZTEST_BENCHMARK_SUITE(math_benchmarks, suit_setup, suit_teardown);
+
+ZTEST_BENCHMARK(math_benchmarks, void_test, 100)
+{
+	/* Intentionally left blank */
+}
+
+ZTEST_BENCHMARK(math_benchmarks, benchmark_stack_add, 100)
+{
+	int a = 5;
+	int b = 10;
+	int c = 0;
+
+	c += a + b;
+}
+
+ZTEST_BENCHMARK(math_benchmarks, benchmark_stack_sub, 100)
+{
+	int a = 10;
+	int b = 5;
+	int c = 0;
+
+	c += a - b;
+}
+
+struct container {
+	int a;
+	int b;
+	int res;
+};
+
+static struct container data;
+
+static void setup_fn(void)
+{
+	data.a = 20;
+	data.b = 15;
+	data.res = 0;
+}
+
+static void teardown_fn(void)
+{
+	memset(&data, 0, sizeof(data));
+}
+
+ZTEST_BENCHMARK_SETUP_TEARDOWN(math_benchmarks, global_subtraction, 100, setup_fn, teardown_fn)
+{
+	data.res = data.a - data.b;
+}
+
+ZTEST_BENCHMARK_SETUP_TEARDOWN(math_benchmarks, global_addition, 100, setup_fn, teardown_fn)
+{
+	data.res = data.a + data.b;
+}
+
+ZTEST_BENCHMARK(math_benchmarks, benchmark_pure_asm_add, 100)
+{
+	__asm__ volatile (
+#if defined(CONFIG_X86)
+		"add %eax, %eax\n\t"
+#elif defined(CONFIG_ARM)
+		"add r0, r0, r0\n\t"
+#elif defined(CONFIG_RISCV)
+		"add a0, a0, a0\n\t"
+#else
+	#error "Unsupported architecture for pure assembly benchmark"
+#endif
+	);
+}
+
+static uint8_t buffer[32];
+#include <zephyr/random/random.h>
+ZTEST_BENCHMARK(math_benchmarks, csrand, 100)
+{
+	sys_csrand_get(buffer, sizeof(buffer));
+}
+
+ZTEST_BENCHMARK_TIMED(math_benchmarks, csrand_timed, 1000)
+{
+	sys_csrand_get(buffer, sizeof(buffer));
+}
+
+ZTEST_BENCHMARK_TIMED(math_benchmarks, add_timed, 1000)
+{
+	int a = 10;
+	int b = 5;
+	int c = 0;
+
+	c += a + b;
+}
+
+ZTEST_BENCHMARK_TIMED(math_benchmarks, benchmark_pure_asm_add_timed, 1000)
+{
+	__asm__ volatile (
+#if defined(CONFIG_X86)
+		"add %eax, %eax\n\t"
+#elif defined(CONFIG_ARM)
+		"add r0, r0, r0\n\t"
+#elif defined(CONFIG_RISCV)
+		"add a0, a0, a0\n\t"
+#else
+	#error "Unsupported architecture for pure assembly benchmark"
+#endif
+	);
+}
--- a/subsys/testsuite/ztest/benchmark/CMakeLists.txt
+++ b/subsys/testsuite/ztest/benchmark/CMakeLists.txt
@@ -1,0 +1,12 @@
+# Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+cmake_minimum_required(VERSION 3.20.0)
+
+zephyr_library_sources_ifdef(CONFIG_ZTEST_BENCHMARK
+  ${CMAKE_CURRENT_LIST_DIR}/src/benchmark.c
+)
+
+zephyr_linker_sources(DATA_SECTIONS benchmark.ld)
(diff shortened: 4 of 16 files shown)


samples/subsys/testsuite/benchmark/src/main.c
```diff
@@ -0,0 +1,82 @@
...
+	int a;
+	int b;
+	int res;
+};
+
+static struct container data;
+
+static void setup_fn(void)
+{
+	struct container *c = (struct container *)&data;
```
faxe1008: Casts not needed, just access directly.

samples/subsys/testsuite/benchmark/src/main.c
```diff
@@ -0,0 +1,82 @@
...
+	c->a = 20;
+	c->b = 15;
+	c->res = 0;
+}
+
+static void teardown_fn(void)
+{
+	struct container *c = (struct container *)&data;
+
+	memset(c, 0, sizeof(*c));
```
faxe1008: Memset directly.

samples/subsys/testsuite/benchmark/src/main.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+ZTEST_BENCHMARK_SETUP_TEARDOWN(math_benchmarks, benchmark_subtraction, 100, setup_fn, teardown_fn)
+{
+	struct container *c = (struct container *)&data;
+
+	c->res = c->a - c->b;
```
faxe1008: No aquisition of a pointer needed, access directly.

faxe1008:
Since I currently have a place where I want to use it I tested it a bit:
```
render_text ================================================
	Sample size:100, total cycles: 2801847200
	Mean(u): 4294967295.4294967295
	Standard deviation(s): 29173.076
	Standard Error(SE): 2917.307
	Min: 4294966847 (run #0)
	Max: 3292173216 (run #100)
````
There seems to be an issue with the max/min stat and during my testing the printed mean value was always this exact number despite some really really big performance changes.

File: subsys/testsuite/ztest/benchmark/src/benchmark.c [hunk 4]
```diff
@@ -0,0 +1,159 @@
...
+	.iterations = 1000,
+	.stats = &ctrl_stats,
+};
+
+void benchmark_main(void)
+{
+	timing_init();
+	timing_start();
+
+	k_msleep(100); /* system stabilization */
```
Mattemagikern: I can't explain this without major guess work. 

When I replaced the sleep with combinations from `<zephyr/sys/barrier.h>` it didn't yield the expected results, the control tests was noticably higher than it should have been and it was on avrage slower than the asm_add benchmark in the sample. 

If anyone can explain this with a level of confidence I'm all ears. Because I loath arbitrary sleeps in code, and yet here I am... 

Edit: 
This was tested rigorously on nrf53_cpuapp as a hardware target. It is only visable when running on real hardware, not when benchmarking with qemu.

JarmouniA:
> Sorry, fumbeled in the GH app. :^)

The worst app ever :)

subsys/testsuite/ztest/benchmark/benchmark.ld
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
```
faxe1008, line 2: Copyrights need update

subsys/testsuite/ztest/benchmark/src/benchmark.c
bjarki-andreasen: It can be many things, most obvious would be some background event happening, like the UART being suspended, or since this is an nrf soc, the LFXO or HFXO being ready. I ran into exactly that here https://github.com/zephyrproject-rtos/zephyr/blob/7367513d916a56531b9660f3ff1df82acd52f277/tests/drivers/spi/spi_loopback/src/spi.c#L362-L367 since the UART was not done printing the previous test output, in this case it could be the boot prompt or something.

subsys/testsuite/ztest/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+cmake_minimum_required(VERSION 3.28)
+zephyr_library_sources_ifdef(CONFIG_ZTEST_BENCHMARK
+  ${CMAKE_CURRENT_LIST_DIR}/src/benchmark.c
+)
+zephyr_linker_sources(DATA_SECTIONS benchmark.ld)
```
bjarki-andreasen, line 9: Missing commands for linker script generator:
```
zephyr_iterable_section(NAME ztest_benchmark GROUP DATA_REGION ${XIP_ALIGN_WITH_INPUT})
zephyr_iterable_section(NAME ztest_benchmark_timed GROUP DATA_REGION ${XIP_ALIGN_WITH_INPUT})
zephyr_iterable_section(NAME ztest_benchmark_suite GROUP DATA_REGION ${XIP_ALIGN_WITH_INPUT})
```

subsys/testsuite/ztest/benchmark/src/benchmark.c
```diff
@@ -0,0 +1,249 @@
...
+	printk("\n");
+}
+
+static void print_double(const char *str, double value)
+{
+	int64_t digit = (int64_t)value;
+	int64_t frac = (int64_t)((value - digit) * 1e3);
+
+	printk("\t%s: %lld.%03lld\n", str, digit, frac);
+}
```
bjarki-andreasen: How about
```
CONFIG_REQUIRES_FULL_LIBC=y
CONFIG_REQUIRES_FLOAT_PRINTF=y
```
and
```
printk("\t%s: %.3f\n", str, double_value);
```

subsys/testsuite/ztest/benchmark/src/benchmark.c
```diff
@@ -0,0 +1,249 @@
...
+#include <zephyr/timing/timing.h>
+
+#define RUN_10(x) x; x; x; x; x; x; x; x; x; x;
+#define RUN_100(x) RUN_10(x) RUN_10(x) RUN_10(x) RUN_10(x) RUN_10(x) \
+			RUN_10(x) RUN_10(x) RUN_10(x) RUN_10(x) RUN_10(x)
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(ztest_benchmark, LOG_LEVEL_WRN);
+
+static void printk_line(const char *fn, char sep_char)
```
bjarki-andreasen, line 20: I would suggest using the same format as the "normal" ZTEST
```
START - benchmark_addition
  DONE - benchmark_addition
  SAMPLE SIZE - 100
  TOTAL CYCLES - 1621
  STD - 2.102
  SE - 0.210
  MAX - 16 - #4
  MIN - 37 - #1
TESTSUITE benchmark_addition complete.
```
Most importantly to make the output machine parseable, no special chars, strict key value format, spaces for indentation, and to avoid duplicate code.

bjarki-andreasen: I could imagine having pytest or something read the output and validate the result against some expected values, similar to power tests and of course "normal" pass/fail ones.

bjarki-andreasen (changes requested):
Cool feature, until now I have measured CPU performance by running some "basic" app and looking at the CPU idle times, which is still a useful method for testing, but this approach is definitely a bit more systematic for more discrete code.

subsys/testsuite/ztest/benchmark/CMakeLists.txt
Mattemagikern: I thought this was managed by the: 
```
ITERABLE_SECTION_RAM(ztest_benchmark, 4)
ITERABLE_SECTION_RAM(ztest_benchmark_timed, 4)
ITERABLE_SECTION_RAM(ztest_benchmark_suite, 4)
```
When I tried to mimic what "regular" ztest does I found this [doc](https://docs.zephyrproject.org/latest/kernel/iterable_sections/index.html) for the section_itterable macros which I thought was really neat. But the docs left out the `zephyr_iterable_section` CMake function. 

As I'm looking around I see: `subsys/net/ip/CMakeLists.txt`  which uses both, and that the `zephyr_iterable_section` is missing from the "regular" ztest CMake. I also found the: `Z_LINK_ITERABLE_SUBALIGN` which I sholud change the alignment to. 

Could you explain this to me? and should I update the docs for the iterable_section?

subsys/testsuite/ztest/benchmark/src/benchmark.c
Mattemagikern: Hahah that is absolutely the better way :+1:

samples/subsys/testsuite/benchmark/prj.conf
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+CONFIG_LOG=y
+CONFIG_LOG_MODE_IMMEDIATE=y
```
faxe1008: Any particular reason for that one?

doc/develop/test/benchmark.rst
```diff
@@ -0,0 +1,187 @@
...
+Timed benchmarks incontrast to the standard benchmarks measures execution time of the code
+instead of cycles. This is useful for benchmarking code that may have variable execution
+times or when you want to measure the actual time taken rather than just CPU cycles of a critical
+path. It provides a broader view of performance characteristics, especially for code that involves
+I/O operations, context switches, or other factors that can influence execution time beyond raw CPU
+performance.
+
+.. code-block:: c
+
+   ZTEST_BENCHMARK_TIMED(measurements, timed_critical_path, 1000)
```
faxe1008: imo it could be clearer in the explanation that this test is run for 1000ms

subsys/testsuite/ztest/include/zephyr/benchmark.h
```diff
@@ -0,0 +1,154 @@
...
+
+/** @cond INTERNAL_HIDDEN */
+typedef void (*ztest_benchmark_fn_t)(void);
+struct ztest_benchmark_suite {
+	const char *name;
+	ztest_benchmark_fn_t setup;
+	ztest_benchmark_fn_t teardown;
+};
+
+struct ztest_extream_value {
```
faxe1008: Typo ?

subsys/testsuite/ztest/benchmark/src/benchmark.c
Mattemagikern: > since the UART was not done printing the previous test output, in this case it could be the boot prompt or something.

That was indeed it :+1:, If I moved to deffered logging the output became much better but that poses the risk of dropping measurements. 
As I was fixing the review comments I found out, to my dismay, that the measurements differ some clock cycles depending on the logging and depending on if I were logging machine-output or human readable output etc. on the nrf53 (qemu did not suffer the same behaviour). We're talking sub 10 CCs that could be interference from the ideal and I'm not sure if it is any point in trying to address it as I've found no way of knowing when the target board reaches a steady-state. Ideally I would wrapp a test-case in a steady-state barrier, but I've found no way of doing this.

samples/subsys/testsuite/benchmark/prj.conf
Mattemagikern: The logging interferes a little on real-targets, we're talking about sub 1-5 CCs on the mean depending on the critical path. When I was developing the framework I found the LOG_MODE_IMMEDIATE to decrease the deviation the most but it did not completly remove it.

I'm going to remove it as it isn't strictly neccesary :+1:

subsys/testsuite/ztest/include/zephyr/benchmark.h
Mattemagikern: it is indeed, spelling isn't my thing haha. 
Thanks for making me aware, would not have noticed otherwise :)

subsys/testsuite/ztest/benchmark/benchmark.ld
Mattemagikern: I'm not sure how to update it. 
Is it just to update to 2026 or should I put a range? 2025-2026? 
The docs was not specific when it comes to the year and the significance of it.

subsys/testsuite/ztest/benchmark/src/benchmark.c
Mattemagikern: I made a Kconfig option that produces a CSV for each test which should be easy to parse for a script :) 

What do you think?

subsys/testsuite/ztest/benchmark/src/benchmark.c
```diff
@@ -0,0 +1,249 @@
...
+	if (stats->samples > 1) {
+		sample_variance = stats->m2 / (double)(stats->samples - 1);
+		stddev = sqrt(sample_variance);
+		std_error = stddev / sqrt(stats->samples);
+	}
+
+	printk_line(benchmark->name, '=');
+	printk("\tSample size:%lld, total cycles: %lld\n", stats->samples,
+			discrete_noise_correction(stats->total, ctrl * stats->samples));
+	print_double("Mean(u)", noise_correction(stats->mean, ctrl));
```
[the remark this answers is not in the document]
Mattemagikern: If I change this a code-compliance check triggers, it was part of the original commit. :)

subsys/testsuite/ztest/benchmark/benchmark.ld
faxe1008: imo since its mostly written 2025 should be fine. Objections @kartben ?

subsys/testsuite/ztest/benchmark/Kconfig
```diff
@@ -0,0 +1,21 @@
...
+config ZTEST_BENCHMARK
+	bool "ztest benchmarking features"
+	select TIMING_FUNCTIONS
+	select REQUIRES_FULL_LIBC
+	select REQUIRES_FLOAT_PRINTF
+	help
+		Enable features for benchmarking purposes.
+
+if ZTEST_BENCHMARK
+config ZTEST_BENCHMARK_MACHINE_OUTPUT
```
faxe1008: Maybe a choice symbol if we want different implementations later on

subsys/testsuite/ztest/benchmark/Kconfig
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2025 Måns Ansgariusson <mansgariusson@gmail.com>
+# SPDX-License-Identifier: Apache-2.0
+
+menu "benchmarking options"
+config ZTEST_BENCHMARK
+	bool "ztest benchmarking features"
+	select TIMING_FUNCTIONS
+	select REQUIRES_FULL_LIBC
+	select REQUIRES_FLOAT_PRINTF
+	help
```
faxe1008, line 11: Maybe mark as experimental as there might still be some shifting API wise

subsys/testsuite/ztest/benchmark/benchmark.ld
kartben: no objection at all - you could even drop the year altogether for what it's worth :)

faxe1008 (dismissed):

subsys/testsuite/ztest/benchmark/CMakeLists.txt
bjarki-andreasen: To explain it, a linker is a command line utility like any other, it needs to know what files to include, how and where to define and place symbols, like iterable sections, .text, you name it. Many compilers, like GCC, support the .LD linker script format, which can be used instead of a very long list of arguments. Not all linkers use this format, so trying to pass a .ld file to a llvm based linker will simply not work. To support any linker (GCC included), we have our own zephyr specific linker script format based on CMAKE. It can generate a .LD file from for example `zephyr_iterable_section(...)` to then feed to a GCC based linker. That same macro can also generate whatever the equivalent for IAR or LLVM or other would need to create that same linker section.

bjarki-andreasen: So in short, it will work as long as the toolchain used supports LD files, it will not work otherwise without the zephyr_iterable_section macros

samples/subsys/testsuite/benchmark/src/main.c
```diff
@@ -0,0 +1,126 @@
...
+
+ZTEST_BENCHMARK_TIMED(math_benchmarks, add_timed, 1000)
+{
+	int a = 10;
+	int b = 5;
+	int c = 0;
+
+	c += a + b;
+}
+
```
bjarki-andreasen, line 112: Add some math here, like expf or logf or something, that would be cool

bjarki-andreasen (dismissed):
The CSV output is a fantastic idea

nashif:
just ran the sample on qemu_x86 and getting negative numbers...:

```
*** Booting Zephyr OS build v4.3.0-8464-g82878bc2fbe6 ***
math_benchmarks ############################################
D: Setting up math benchmarks suite
benchmark_pure_asm_add =====================================
	Sample size:100, total cycles: -8972
	Mean(u): -89.728
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: -89 (run #1)
	Max: -89 (run #1)
```

subsys/testsuite/ztest/benchmark/CMakeLists.txt
Mattemagikern: Thanks for the explanation :)

samples/subsys/testsuite/benchmark/src/main.c
Mattemagikern: I think it is a good idea, a "real" maths benchmark testsuite would be great. Especially when compairing hardware targets and testing Kconfig options (like with or without FPU). 

However, I think that the sample should just showcase the ztest-benchmarks capabilities and it tests almost all functionallity in its current form (the only thing missing I can think of right now is the different output formats).

Mattemagikern:
@nashif 
It is possible to get negative values on a real hardware target (such as the nrf53 I've been testing on) and that comes down to system-noise during the initial control tests. The samples are extreamly sensitive to noise since the critical-path is so small.

But I've not seen that on qemu_x86 before, qemu_x86 has been deterministic on my machine, always 32 cycles for an asm add, always 0 standard deviation etc.

```bash 
             RAM:      225336 B        31 MB      0.69%
        IDT_LIST:          0 GB         2 KB      0.00%
Generating files from /workspace/build/zephyr/zephyr.elf for board: qemu_x86/atom
[145/146] To exit from QEMU enter: 'CTRL+a, x'[QEMU] CPU: qemu32,+nx,+pae
SeaBIOS (version zephyr-v1.16.3-2-0-gf8da0cc-zephyr)
Booting from ROM..
*** Booting Zephyr OS build v4.3.0-8436-geccdc41ef1c0 ***
math_benchmarks ############################################
D: Setting up math benchmarks suite
benchmark_pure_asm_add =====================================
	Sample size:100, total cycles: 3200
	Mean(u): 32.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 32 (run #1)
	Max: 32 (run #1)
benchmark_stack_add ========================================
	Sample size:100, total cycles: 28800
	Mean(u): 288.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 288 (run #1)
	Max: 288 (run #1)
benchmark_stack_sub ========================================
	Sample size:100, total cycles: 25600
	Mean(u): 256.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 256 (run #1)
	Max: 256 (run #1)
csrand =====================================================
	Sample size:100, total cycles: 6336064
	Mean(u): 63360.640
	Standard deviation(s): 12166.400
	Standard Error(SE): 1216.640
	Min: 62144 (run #1)
	Max: 183808 (run #5)
global_addition ============================================
	Sample size:100, total cycles: 12800
	Mean(u): 128.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 128 (run #1)
	Max: 128 (run #1)
global_subtraction =========================================
	Sample size:100, total cycles: 12800
	Mean(u): 128.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 128 (run #1)
	Max: 128 (run #1)
void_test ==================================================
	Sample size:100, total cycles: 0
	Mean(u): 0.000
	Standard deviation(s): 0.000
	Standard Error(SE): 0.000
	Min: 0 (run #1)
	Max: 0 (run #1)
add_timed ==================================================
	Iterations: 1463500
	Total Time: 997999399 ns (Gross)
	Work Time: 427965249 ns (Net)
	Ops/Sec: 3419670.180
	Cycles/Ops: 292.426
benchmark_pure_asm_add_timed ===============================
	Iterations: 2349100
	Total Time: 996854055 ns (Gross)
	Work Time: 81878160 ns (Net)
	Ops/Sec: 28690190.400
	Cycles/Ops: 34.855
csrand_timed ===============================================
	Iterations: 15900
	Total Time: 1000384424 ns (Gross)
	Work Time: 994191363 ns (Net)
	Ops/Sec: 15992.897
	Cycles/Ops: 62527.746
D: Tearing down math benchmarks suite
```

faxe1008 (approved):

nashif (approved):

PR #101161, opened 2025-12-17
soc: nxp: imxrt118x: Change to use __rom_region_start symbol

zejiang0jason:
The boot container was using a non-standard symbol name `__rom_start_address` for the ROM start address. This symbol is not defined when build with `-DCONFIG_CMAKE_LINKER_GENERATOR=y`.

This commit use the symbol `__rom_region_start` which is available for both cases.

### DIFF
--- a/soc/nxp/imxrt/imxrt118x/soc.c
+++ b/soc/nxp/imxrt/imxrt118x/soc.c
@@ -31,7 +31,7 @@ LOG_MODULE_REGISTER(soc, CONFIG_SOC_LOG_LEVEL);
 
 extern char __start[];
 extern char _flash_used[];
-extern char __rom_start_address[];
+extern char __rom_region_start[];
 const __imx_boot_container_section container boot_header = {
 	.hdr = {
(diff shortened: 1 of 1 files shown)


lucien-nxp (approved):

ZhaoxiangJin (approved):
`non-standard` => `non-GNU-ld non-compatible`

zejiang0jason:
> `__rom_start_address` is also used in:
> 
> * zephyr\soc\nxp\imxrt\boot_header.ld
> * soc\nxp\mcx\mcxe\mcxe31x\boot_header.ld
> * zephyr\soc\nxp\rw\boot_header.ld
> 
> Should they be fixed too?

Hi @butok , not needed, `__rom_start_address` is available for ld-script, 
https://github.com/zephyrproject-rtos/zephyr/blob/main/arch/common/rom_start_address.ld#L11

The problem only occurs when `-DCONFIG_CMAKE_LINKER_GENERATOR=y`.

butok (approved):

PR #101039, opened 2025-12-15
cmake: linker: arm: ARM cmake linker generator improvement

zejiang0jason:
1. Add symbol _flash_used

PR #101165, opened 2025-12-17
nxp: linker: Add cmake linker generator support

zejiang0jason:
Support cmake linker generator for some NXP boards, use build option `-DCONFIG_CMAKE_LINKER_GENERATOR=y` to enable the feature

RT500/RT600/RT700/MCXE31 not supported yet, because the these SOCs use code relocation feature, which need to improve cmake linker generator common part.

Depends on: https://github.com/zephyrproject-rtos/zephyr/pull/101039

### DIFF
--- a/soc/nxp/kinetis/CMakeLists.txt
+++ b/soc/nxp/kinetis/CMakeLists.txt
@@ -13,6 +13,15 @@ zephyr_linker_sources_ifdef(CONFIG_KINETIS_FLASH_CONFIG
   SORT_KEY ${CONFIG_KINETIS_FLASH_CONFIG_OFFSET}
   flash_config.ld
   )
+if(CONFIG_KINETIS_FLASH_CONFIG)
+  zephyr_linker_section_configure(
+    SECTION .rom_start
+    INPUT ".kinetis_flash_config"
+          ".kinetis_flash_config.*"
+    OFFSET ${CONFIG_KINETIS_FLASH_CONFIG_OFFSET}
+    KEEP
+    )
+endif()
 
 # CMSIS SystemInit will disable watchdog unless instructed not to.
 # Add a compiler definition here to leave watchdog untouched
--- a/soc/nxp/rw/CMakeLists.txt
+++ b/soc/nxp/rw/CMakeLists.txt
@@ -18,6 +18,24 @@ zephyr_sources_ifdef(CONFIG_PM
 zephyr_linker_sources_ifdef(CONFIG_NXP_RW6XX_BOOT_HEADER
   ROM_START SORT_KEY 0 boot_header.ld)
 
+if(CONFIG_NXP_RW6XX_BOOT_HEADER)
+  zephyr_linker_section_configure(
+    SECTION .rom_start
+    INPUT ".flash_conf"
+    OFFSET ${CONFIG_FLASH_CONFIG_OFFSET}
+    KEEP
+    PRIO 10
+  )
+
+  zephyr_linker_section_configure(
+    SECTION .rom_start
+    INPUT ".boot_hdr.ivt"
+    OFFSET ${CONFIG_IMAGE_VECTOR_TABLE_OFFSET}
+    KEEP
+    PRIO 11
+  )
+endif()
+
 set(SOC_LINKER_SCRIPT ${ZEPHYR_BASE}/include/zephyr/arch/arm/cortex_m/scripts/linker.ld CACHE INTERNAL "")
 
 zephyr_include_directories(.)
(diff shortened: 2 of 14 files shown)


Albort12138 (dismissed):

soc/nxp/kinetis/CMakeLists.txt
```diff
@@ -13,6 +13,15 @@ zephyr_linker_sources_ifdef(CONFIG_KINETIS_FLASH_CONFIG
   SORT_KEY ${CONFIG_KINETIS_FLASH_CONFIG_OFFSET}
   flash_config.ld
   )
+if(CONFIG_KINETIS_FLASH_CONFIG)
```
ZhaoxiangJin, line 16: shoule be: `if(CONFIG_CMAKE_LINKER_GENERATOR AND CONFIG_KINETIS_FLASH_CONFIG)`

Holt-Sun (dismissed):

ZhaoxiangJin:
> You have added linker sections, and nothing has been deleted/moved. Just wondering, how did it work before?

Previously, we focused on Zephyr SDK, so relying on templates + linker snippets (.ld) was fine. Now, to extend support for IAR/ARM Clang, which don't recognize .ld files, we need to add CMake linker generator support.

ZhaoxiangJin:
> > > You have added linker sections, and nothing has been deleted/moved. Just wondering, how did it work before?
> > 
> > 
> > Previously, we focused on Zephyr SDK, so relying on templates + linker snippets (.ld) was fine. Now, to extend support for IAR/ARM Clang, which don't recognize .ld files, we need to add CMake linker generator support.
> 
> So this is a workaround for IAR/Clang limitations. Should we report this to IAR? I had thought that IAR was adapting Zephyr, and we could continue to rely on common templates/snippets/code.

It's hard to say this is a limitation of IAR/Clang, because each toolchain has its own linker format. If we want our code to be compiled correctly by these toolchains, we should use the functions provided by the CMake linker generator to properly place these sections. IAR has already document this. https://docs.zephyrproject.org/latest/develop/toolchains/iar_arm_toolchain.html

hakehuang:
I will do a regression on all NXP platforms with IAR compiler after https://github.com/zephyrproject-rtos/zephyr/pull/101039 get merged

butok:
> It's hard to say this is a limitation of IAR/Clang, because each toolchain has its own linker format. If we want our code to be compiled correctly by these toolchains, we should use the functions provided by the CMake linker generator to properly place these sections. IAR has already document this. https://docs.zephyrproject.org/latest/develop/toolchains/iar_arm_toolchain.html

Hi @RobinKastberg 

There is the known IAR limitation:
`"The IAR Toolchain uses ilink for linking and depends on Zephyr’s CMAKE_LINKER_GENERATOR. ilink is incompatible with Zephyr’s linker script template, which works with GNU ld."`

Is IAR planning to solve it? 
Or is this something that Zephyr projects must adjust in its own way for each platform?

zejiang0jason:
rebase to include latest change in pr: https://github.com/zephyrproject-rtos/zephyr/pull/101039

soc/nxp/rw/CMakeLists.txt
```diff
@@ -18,11 +18,108 @@ zephyr_sources_ifdef(CONFIG_PM
...
 zephyr_linker_sources(RAM_SECTIONS sections.ld)
+
+set(TXQ23_SIZE 0x1080)
+dt_nodelabel(smu1_data_node NODELABEL "smu1_data")
+dt_reg_size(SMU1_SIZE PATH ${smu1_data_node})
+dt_nodelabel(smu2_data_node NODELABEL "smu2_data")
+dt_reg_size(SMU2_SIZE PATH ${smu2_data_node})
+
+zephyr_linker_section(
+  NAME .SMU1
```
zejiang0jason: Hi @RobinKastberg ,

I found that when using IAR, the section name can't be dot + region name. For example, this code place `.SMU1` in region `SMU1`.
https://github.com/zephyrproject-rtos/zephyr/blob/f16c568c9924eedf245688dab1ffbaaef77aa5e7/soc/nxp/rw/sections.ld#L12-L23

When I adding such code for IAR, there is build error that block SMU1 is redefined.
```
    block SMU1_winput with alignment=4, fixed order { section .SMU1, section .SMU1.* },
          ^
"/home/debian/work/zephyr/zephyr/build/zephyr/linker_zephyr_pre0.cmd",1534  Error[Lc011]: already defined

  define block SMU1 with fixed order, alignment=4, minimum size=0x7f800
               ^
"/home/debian/work/zephyr/zephyr/build/zephyr/linker_zephyr_pre0.cmd",1532  Error[Lc011]: already defined

  ".SMU1": place in SMU1 { block SMU1 };
                           ^
"/home/debian/work/zephyr/zephyr/build/zephyr/linker_zephyr_pre0.cmd",1539  Error[Lc039]: "SMU1" already placed in region or block
            (at line 1530 of "/home/debian/work/zephyr/zephyr/build/zephyr/linker_zephyr_pre0.cmd")

  define image symbol __SMU1_load_start = LOADADDR(SMU1);
```
I got such from the linker_zephyr_pre0.cmd:
```
define block SMU1 with fixed order, alignment=4
{
  block SMU1_winput with alignment=4, fixed order { section SMU1, section SMU1.* }
};
"SMU1": place in SMU1 { block SMU1 };

define block SMU1 with fixed order, alignment=4, minimum size=0x7f800
{
  block SMU1_winput with alignment=4, fixed order { section .SMU1, section .SMU1.* },
  block SMU1_100 with alignment = 4, fixed order {  section .smu_cpu13_mbox,   section .smu_cpu31_txq }
};
keep { block SMU1_100, section .smu_cpu13_mbox, section .smu_cpu31_txq };
```
Could you please confirm whether this is the desired behavior, thanks.

RobinKastberg: This is probably not intended, and has little to do with the "." I think. 
Since Regions from device tree generate a default block, and your output section defines a block with the same name, there is a collision.
I will have to take a closer look how to resolve this. But my first reaction is that you can probably change the output section name to work around this. Annoying but should be harmless unless there are tools consuming the output section info.

zejiang0jason: Thanks, the step to reproduce the issue is:
1. Using latest code in this PR.
2. run command: `west build -p always samples/hello_world/ -b  frdm_rw612`

Thanks.

tejlmand (commented):
LGTM, didn't spot anything suspicious code-wise, but the section naming collision must be solved before giving a +1

zejiang0jason:
convert to draft because the PR depends on is not merged yet: https://github.com/zephyrproject-rtos/zephyr/pull/101039.

soc/nxp/rw/CMakeLists.txt
zejiang0jason: I applied workaround, use name `.smu1` instead of `.SMU1`

tejlmand:
> Hi @tejlmand , this PR depends on #101039, the non-NXP specific code changes are in that PR, please help review to see whether it can be merged first, thanks.

which again seems to have spawned https://github.com/zephyrproject-rtos/zephyr/pull/102472 that needs merging first.
Guess we should get the cleanup handled first :slightly_smiling_face:

zejiang0jason:
Because there are many platforms to support, I plan to use split PRs to update family by family.

> @zejiang0jason there are more platforms need work see below summary

butok:
https://github.com/zephyrproject-rtos/zephyr/pull/101039 it depends on has been closed.

hakehuang:
@zejiang0jason can you rebase ?

PR #91933, opened 2025-06-20
soc: dts: drivers: add SoCs for amebad series

zjian-zhang:
The [amebad](https://www.realmcu.com/en/Home/Products/RTL872xCS-RTL872xD-Series) microcontroller (MCU) is an advanced system-on-chip (SoC) featuring an Arm® Cortex®-M33 and Cortex®-M23, large Flash and SRAM memories, and supports dual-band Wi-Fi 4(2.4GHz + 5GHz) and BLE5.0 specifications.

The initial support for the board includes GPIO, PINCTRL and Loguart(serial).

PR #101168, opened 2025-12-17
soc: dts: drivers: add SoCs for amebaG2 series

zjian-zhang:
The amebaG2 microcontroller (MCU) is an advanced system-on-chip (SoC) featuring two Arm® Cortex®-M55, large Flash and SRAM memories, and supports dual-band Wi-Fi 4(2.4GHz + 5GHz) and Bluetooth 5.0 LE specifications.

The initial support for the board includes GPIO, PINCTRL and Loguart(serial).


soc/realtek/ameba/amebaG2/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+zephyr_sources(soc.c)
+
+zephyr_linker_sources(SECTIONS boot_section.ld)
+set(SOC_LINKER_SCRIPT ${ZEPHYR_BASE}/include/zephyr/arch/arm/cortex_m/scripts/linker.ld CACHE INTERNAL "")
+
+# west sign
+file(GLOB BIN_FILES "${ZEPHYR_HAL_REALTEK_MODULE_DIR}/zephyr/blobs/${CONFIG_SOC_SERIES}/bin/*.bin")
+
+if(BIN_FILES)
```
nordicjm: as per previous PR

nordicjm (changes requested):

soc/realtek/ameba/amebaG2/CMakeLists.txt
zjian-zhang: Yes, I'll remove it. Since this PR overlaps partially with #91933, I've converted this PR to a draft and will rebase the latest changes after #91933 is merged.
Thanks for your review and feedback!

zjian-zhang: done

west.yml
```diff
@@ -225,8 +225,8 @@ manifest:
       groups:
         - hal
     - name: hal_realtek
+      revision: 3daa83b545eb3a50188e9285fc9226acf4a55026
       path: modules/hal/realtek
-      revision: 4f8703eb110220e3a98c7e883f672a575679b9e8
```
nordicjm, line 229: put on same line it was before

soc/realtek/ameba/Kconfig
```diff
@@ -9,4 +9,16 @@ if SOC_FAMILY_REALTEK_AMEBA
 
 rsource "*/Kconfig"
 
+choice
```
nordicjm: needs a name, also needs a "none" option which is selected by default which does not include anything

soc/realtek/ameba/Kconfig
```diff
@@ -9,4 +9,16 @@ if SOC_FAMILY_REALTEK_AMEBA
...
+	prompt "Ameba NP prebuilt images"
+	help
+	  Choose the prebuilt image to be executed on the NP core of the Ameba multi-core MCU.
+	  The image is responsible for booting the AP on the device.
+
+config SOC_AMEBA_NP_IMAGE
```
nordicjm, line 24: there is no reference to this in the whole PR other than this?

nordicjm (changes requested):

west.yml
zjian-zhang: Done. The previous modification was intended to keep it consistent with the one above.

soc/realtek/ameba/Kconfig
zjian-zhang: Done. The previous choice syntax was based on the example you provided earlier:
https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/infineon/cat1a/Kconfig#L41

soc/realtek/ameba/Kconfig
zjian-zhang: Actually, there is a reference in the PR discussion. As mentioned in https://github.com/zephyrproject-rtos/zephyr/pull/91933:
> "I am not seeing what gets added here, it should be one symbol for a specific driver = one blob and should be in code in the module folder not soc folder since blobs are modules."

Following that guidance, I’ve moved the use of this option into the hal_realtek repository:
https://github.com/zephyrproject-rtos/hal_realtek/blob/main/ameba/amebaG2/CMakeLists.txt#L54

zjian-zhang: And, following the example at https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/infineon/cy8ckit_062s4/cy8ckit_062s4_defconfig#L11, I have also enabled `CONFIG_SOC_AMEBA_NP_IMAGE=y` by default in the board's defconfig now.

nordicjm: by module folder I mean 'zephyr/modules', including blob images should not be done in your repo, it needs to be disabled by default and it needs TSC approval to even be accepted in the first place

zjian-zhang: The Ameba SoC family includes multiple different CPUs. Only one of them runs Zephyr, and it does not require any binary blobs. The other CPU (which does not run Zephyr) requires a precompiled binary (.bin) to function.
As suggested, the build logic for this optional feature has been placed in the `zephyr/modules/hal_realtek` module stub. We have disabled the inclusion and handling of this binary by default. The corresponding Kconfig option is now off by default, so the non-Zephyr CPU remains inactive unless explicitly enabled by the user.

west.yml
```diff
@@ -226,7 +226,7 @@ manifest:
         - hal
     - name: hal_realtek
       path: modules/hal/realtek
-      revision: 4f8703eb110220e3a98c7e883f672a575679b9e8
+      revision: pull/6/head
```
fabiobaltieri: `89085709fd6a14326f6b107b82f58ccf686e1fda`

zjian-zhang: done.

modules/hal_realtek/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2026 Realtek Semiconductor Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+if(ZEPHYR_HAL_REALTEK_MODULE_DIR)
+  add_subdirectory(${ZEPHYR_HAL_REALTEK_MODULE_DIR} hal_realtek)
+endif()
+
+add_subdirectory_ifdef(CONFIG_SOC_FAMILY_REALTEK_AMEBA ameba)
```
nordicjm, line 8:
```suggestion
if(CONFIG_SOC_FAMILY_REALTEK_AMEBA)
  if(ZEPHYR_HAL_REALTEK_MODULE_DIR)
    add_subdirectory(${ZEPHYR_HAL_REALTEK_MODULE_DIR} hal_realtek)
  endif()

  add_subdirectory(ameba)
endif()
```

modules/hal_realtek/ameba/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+# west sign
+if(CONFIG_SOC_AMEBA_NP_IMAGE)
+  set(hal_blob_dir ${ZEPHYR_HAL_REALTEK_MODULE_DIR}/zephyr/blobs/${CONFIG_SOC_SERIES}/bin/)
+
+  if(CONFIG_SOC_SERIES_AMEBAD)
+    set(blob_bin_file ${hal_blob_dir}/km0_image2_all.bin)
+  endif()
+
+  if(CONFIG_SOC_SERIES_AMEBADPLUS)
```
nordicjm: (non-blocking) cmake has `elseif()` for these

soc/realtek/ameba/Kconfig
```diff
@@ -9,4 +9,22 @@ if SOC_FAMILY_REALTEK_AMEBA
...
+config SOC_AMEBA_NP_DISABLED
+	bool "None"
+	help
+	  Do not include any prebuilt image for the NP core.
+
+config SOC_AMEBA_NP_IMAGE
+	bool "DeepSleep"
+	help
+	  DeepSleep prebuilt image is executed on the NP core of the Ameba multi-core MCU.
+endchoice
```
nordicjm, line 29: missing newline line 28

modules/hal_realtek/ameba/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+    add_custom_command(
+      OUTPUT ${ZEPHYR_BINARY_DIR}/${KERNEL_NAME}.raw.map
+      COMMAND ${CMAKE_NM} ${ZEPHYR_BINARY_DIR}/${KERNEL_ELF_NAME} | sort > ${ZEPHYR_BINARY_DIR}/${KERNEL_NAME}.raw.map
+      COMMAND ${CMAKE_OBJDUMP} -d ${ZEPHYR_BINARY_DIR}/${KERNEL_ELF_NAME} > ${ZEPHYR_BINARY_DIR}/${KERNEL_NAME}.asm
+      COMMAND ${PYTHON_EXECUTABLE} ${ZEPHYR_HAL_REALTEK_MODULE_DIR}/ameba/scripts/merge_bin.py ${python_script_args}
+      DEPENDS ${ZEPHYR_BINARY_DIR}/${KERNEL_ELF_NAME} ${blob_bin_file} ${ZEPHYR_HAL_REALTEK_MODULE_DIR}/ameba/${CONFIG_SOC_SERIES}/manifest.json5
+    )
+  else()
+    message(WARNING "Missing Realtek blobs. Run: `west blobs fetch hal_realtek`")
```
nordicjm: look into `zephyr_blobs_verify` instead of this, note you need to account for `CONFIG_BUILD_ONLY_NO_BLOBS`

soc/realtek/ameba/amebaG2/soc.c
```diff
@@ -0,0 +1,68 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/cache.h>
+
+void z_arm_reset(void);
+
+IMAGE2_ENTRY_SECTION
+RAM_START_FUNCTION Img2EntryFun0 = {z_arm_reset, NULL, /* BOOT_RAM_WakeFromPG, */
+				    (uint32_t)RomVectorTable};
+
+void SOC_OSC131_Enable(void)
```
nordicjm: non compliant with zephyr coding standards, also should not be global

nordicjm (changes requested):

modules/hal_realtek/CMakeLists.txt
zjian-zhang: Thanks for the review!
Realtek has multiple product families (e.g., Ameba, Bee, etc.). The recently submitted PR #102691 (which adds support for the RTL87x2G "Bee" series) also relies on `add_subdirectory( $ {ZEPHYR_HAL_REALTEK_MODULE_DIR} hal_realtek)`.
Because this HAL module is shared across different Realtek SoC families—not just Ameba—we cannot restrict its inclusion to `if(CONFIG_SOC_FAMILY_REALTEK_AMEBA)` alone. Using a broader condition ensures all Realtek-based platforms (current and future) can properly access the common HAL.
Let me know if you'd suggest an alternative approach!

modules/hal_realtek/ameba/CMakeLists.txt
zjian-zhang: done.

soc/realtek/ameba/Kconfig
zjian-zhang: done

modules/hal_realtek/CMakeLists.txt
nordicjm: ah OK, leave it as it was then

modules/hal_realtek/ameba/CMakeLists.txt
zjian-zhang: done.

soc/realtek/ameba/amebaG2/soc.c
zjian-zhang: fixed.

soc/realtek/ameba/amebaG2/soc.c
```diff
@@ -0,0 +1,73 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/cache.h>
+
+extern void z_arm_reset(void);
+
+struct RAM_START_FUNCTION {
+	void (*RamStartFun)(void);
+	void (*RamWakeupFun)(void);
+	uint32_t *VectorNS;
+};
```
nordicjm: this is not camel case, lower case letters and underscores - applies to the struct name too

zjian-zhang: Thanks for the feedback! Could you please take another look to see if there are any remaining issues?

boards/realtek/rtl8721f_evb/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. zephyr-app-commands::
+   :zephyr-app: samples/hello_world
+   :board: rtl8721f_evb
+   :goals: build
+
+Flashing
+********
+
+When the build finishes, downloading images into flash by `AmebaImageTool`_:
```
JarmouniA: Shoud be added to https://docs.zephyrproject.org/latest/develop/flash_debug/host-tools.html#flash-debug-host-tools

boards/realtek/rtl8721f_evb/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+
+See the ApplicationNote chapter Image Tool from documentation links for more details.
+
+#. Environment Requirements: EX. WinXP, Win 7 or later, Microsoft .NET Framework 4.0.
+#. Connect chip and PC with USB wire.
+#. Choose the Device profiles according to the chip you use.
+#. Select the corresponding serial port and transmission baud rate. The default baud rate is 1500000.
+#. Select the images to be programmed and set the start address and end address according to the flash layout, refer to [ameba_flashcfg.c/Flash_layout].
+#. Click the Download button and start. The progress bar will show the download progress of each image and the log window will show the operation status.
```
JarmouniA: Does the flashing tool not have a CLI mode?

boards/realtek/rtl8721f_evb/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. note::
+
+   For an empty chip, the bootloader and app image shall be downloaded.
```
JarmouniA: What bootloader?

boards/realtek/rtl8721f_evb/doc/index.rst
```diff
@@ -0,0 +1,108 @@
...
+*********
+
+Using SWD through PA18(SWD_CLK) and PA19(SWD_DAT).
+
+References
+**********
+
+.. _`RTL8721F-EVB`: https://www.realmcu.com/en/Home/Product/add965ea-d661-4a63-9514-d18b6912f8ab#
+.. _`RTL8721F Series`: https://www.realmcu.com
+.. _`AmebaImageTool`: https://github.com/Ameba-AIoT/ameba-rtos/blob/master/tools/ameba/ImageTool/AmebaImageTool.exe
```
JarmouniA: Put a link to the page instead of the executable.

modules/hal_realtek/ameba/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2026 Realtek Semiconductor Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+# west sign
+
+if(CONFIG_BUILD_ONLY_NO_BLOBS OR CONFIG_SOC_AMEBA_NP_DISABLED)
+  message(WARNING "
+  ---------------------------------------------------------------------------
+  Building only the Ameba SoC without non-processor binary blobs.
```
JarmouniA: non-precessor?

modules/hal_realtek/ameba/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+  This is only for building (CI) purposes and will not work on a real device.
```
JarmouniA:
```suggestion
  This is only meant for building images (CI) without running them on a real device.
```

soc/realtek/ameba/amebag2/boot_section.ld
JarmouniA, line 1: (non-blocking) Consider defining the necessary sections in CMake for better portability.

soc/realtek/ameba/amebaG2/soc.c
```diff
@@ -0,0 +1,73 @@
...
+struct ram_start_function {
+	void (*ram_start_func)(void);
+	void (*ram_wakeup_func)(void);
+	uint32_t *vector_ns;
+};
+
+IMAGE2_ENTRY_SECTION
+struct ram_start_function img2_entry_fun0 = {z_arm_reset, NULL, NULL};
+
+static void SOC_OSC131_Enable(void)
```
JarmouniA: Add a comment clarifying that this is an HAL-declared function (if that's the case).

JarmouniA (dismissed):

boards/realtek/rtl8721f_evb/doc/index.rst
zjian-zhang: I've switched to the `amebaflash` runner provided by `hal_realtek`, so `AmebaImageTool` is no longer needed. Therefore, there's no need to update it on that website. Thanks for the reminder!

boards/realtek/rtl8721f_evb/doc/index.rst
zjian-zhang: Yes, already switched to the `amebaflash` runner, which only supports CLI mode.

boards/realtek/rtl8721f_evb/doc/index.rst
zjian-zhang: A secondary bootloader handles loading the non-Zephyr CPU.

boards/realtek/rtl8721f_evb/doc/index.rst
zjian-zhang: `AmebaImageTool` is no longer used; it has been replaced with `amebaflash.py`.

modules/hal_realtek/ameba/CMakeLists.txt
zjian-zhang: changed to non-Zephyr CPU.

modules/hal_realtek/ameba/CMakeLists.txt
zjian-zhang: done

soc/realtek/ameba/amebag2/boot_section.ld
zjian-zhang: OK, Removed unnecessary assignments.

soc/realtek/ameba/amebaG2/soc.c
zjian-zhang: done.

scripts/west_commands/runners/amebaflash.py
JarmouniA, line 1: New runner should be added in a separate PR.

soc/realtek/ameba/amebag2/boot_section.ld
JarmouniA: > OK, Removed unnecessary assignments.

Not sure what you mean.

scripts/west_commands/runners/amebaflash.py
fabiobaltieri: why though? I typically see the opposite, new platform with no runner and reviewer asking "how is this supposed to be usable"

JarmouniA: > why though?

To get a proper review!? And reduce the size of this PR that's already touching many areas.

soc/realtek/ameba/amebag2/boot_section.ld
zjian-zhang: just removed `__image2_backtrace_start__ = .;`, `__image2_backtrace_end__ = .;`, `__ipc_table_start__ = .;` and `__ipc_table_end__ = .;`, others are all needed.

scripts/west_commands/runners/amebaflash.py
zjian-zhang: the new runner needs to be submitted together with this change, because the `board.cmake` file relies on it to enable `west flash` support.
By the way, I noticed that in [PR #102691](https://github.com/zephyrproject-rtos/zephyr/pull/102691), the new runner was also included in the same pull request — which seems to be the established pattern for adding flash support.
Let me know if you'd still prefer to split them, but we believe bundling them ensures the board is fully functional upon merge.

west.yml
```diff
@@ -226,7 +226,7 @@ manifest:
         - hal
     - name: hal_realtek
       path: modules/hal/realtek
-      revision: 4f8703eb110220e3a98c7e883f672a575679b9e8
```
nordicjm, line 229: must come before things that need it i.e. the blobs

soc/realtek/ameba/Kconfig
nordicjm: Not fixed in correct commit

boards/realtek/rtl872xda_evb/board.cmake
```diff
@@ -1,6 +1,17 @@
 # Copyright (c) 2024 Realtek Semiconductor Corp.
 # SPDX-License-Identifier: Apache-2.0
 
+dt_chosen(SHELLUART PROPERTY zephyr,shell-uart)
```
nordicjm: lowercase variables for locals in cmake

boards/realtek/rtl8721f_evb/board.cmake
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2024 Realtek Semiconductor Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+dt_chosen(SHELLUART PROPERTY zephyr,shell-uart)
+if(SHELLUART)
```
nordicjm: as above

west.yml
zjian-zhang: done

soc/realtek/ameba/Kconfig
zjian-zhang: It was my mistake, and I truly appreciate your careful review. Great catch!

boards/realtek/rtl872xda_evb/board.cmake
zjian-zhang: done.

boards/realtek/rtl8721f_evb/board.cmake
zjian-zhang: done

scripts/west_commands/runners/amebaflash.py
fabiobaltieri: > To get a proper review!? And reduce the size of this PR that's already touching many areas.

It's in its own commit that's good enough for the west maintainer to comfortably do a proper review in isolation, the request is for new platform to have the minimum to run hello world, hard to do that if you can't flash it.

Look I could understand if we were talking about non basic drivers or a massively complex runner but that's not the case here and juggling multiple related PRs is a huge pain, poor contributor would have to deal with local merged branches all the time to test anything, was there myself many times you end up screwing things up and then we introduce bugs and waste even more time. Let's be reasonable.

soc/realtek/ameba/amebaG2/Kconfig.soc
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2024 Realtek Semiconductor Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+config SOC_SERIES_AMEBAG2
+	bool
+	select SOC_FAMILY_REALTEK_AMEBA
+
+config SOC_SERIES
+	default "amebaG2" if SOC_SERIES_AMEBAG2
```
nordicjm:
```suggestion
	default "amebag2" if SOC_SERIES_AMEBAG2
```

soc/realtek/ameba/soc.yml
```diff
@@ -7,3 +7,6 @@ family:
       - name: amebadplus
         socs:
           - name: rtl8721dx
+      - name: amebaG2
```
nordicjm:
```suggestion
      - name: amebag2
```

nordicjm (changes requested):
final issue

soc/realtek/ameba/amebaG2/Kconfig.soc
zjian-zhang: done.

soc/realtek/ameba/soc.yml
zjian-zhang: done.

boards/realtek/rtl872xda_evb/board.cmake
```diff
@@ -1,6 +1,17 @@
 # Copyright (c) 2024 Realtek Semiconductor Corp.
 # SPDX-License-Identifier: Apache-2.0
 
+dt_chosen(shelluart PROPERTY "zephyr,shell-uart")
+if(shelluart)
+  dt_prop(shelluart_baudrate PATH ${shelluart} PROPERTY "current-speed")
+  board_runner_args(amebaflash "--baudrate=${shelluart_baudrate}")
+endif()
```
fabiobaltieri, line 8: Interesting, never seen it done in other boards, nothing wrong with it I suppose but does it really make sense? I mean picking the bitrate from the shell, I imagine amebaflash talks with a bootloader and that one has an hardcoded bitrate.

include/zephyr/dt-bindings/pinctrl/amebag2-pinctrl.h
fabiobaltieri, line 1: this file probably belongs to its own commit, and needs some doxygen comments

boards/realtek/rtl872xda_evb/board.cmake
zjian-zhang: Thanks for the observation! You're right—this code isn't needed for normal operation (`amebaflash` works fine with the bootloader's hardcoded baud rate, e.g., 1.5 Mbps).
However, it's currently required for Zephyr Twister tests, which shall run at 115200 baud to avoid log loss (which occurs at 1.5 Mbps; we plan to submit a follow-up PR to resolve this underlying issue). Without shell-configurable baud rate, `amebaflash` fails to send the reboot command at 115200, breaking the test flow.

include/zephyr/dt-bindings/pinctrl/amebag2-pinctrl.h
zjian-zhang: I've pushed this as a separate commit. Regarding the Doxygen comments, it seems a bit strange: I tried using the grouping syntax (`/** ... @{ */` ... `/** @} */`) to cover these macros collectively, but the CI is still flagging them as missing comments.

boards/realtek/rtl872xda_evb/board.cmake
fabiobaltieri: oh is it like a special sequence recognized by the hardware? the boards are all set to 115200, I would have just hardcoded 115200, but hey I guess no harm picking from dt just in case

include/zephyr/dt-bindings/pinctrl/amebag2-pinctrl.h
fabiobaltieri: yeah the check has been recently introduced, we may reconsider in the future but for the moment you need to comment them all, though in this case since they are dt headers I think it makes sense, then the documentation can add a pretty comment next to the definition

scripts/west_commands/runners/amebaflash.py
```diff
@@ -0,0 +1,78 @@
...
+from zephyr_ext_common import ZEPHYR_BASE
+
+
+def _get_flash_module():
+    """
+    Lazy load the external flash.py module.
+    This prevents FileNotFoundError during CI tests when the HAL is missing.
+    """
+    flash_path = (
+        Path(ZEPHYR_BASE) / "../" / "modules" / "hal" / "realtek" / "ameba" / "scripts" / "flash.py"
```
fabiobaltieri: just `".."`? (without the trailing slash)

zjian-zhang: Ah, good point — you're right that using ".." (without the trailing slash) is the idiomatic way with pathlib.Path. I have update it to ".." for consistency and clarity. Thanks for the suggestion!

boards/realtek/rtl8721f_evb/rtl8721f_evb.dts
```diff
@@ -0,0 +1,42 @@
...
+	chosen {
+		zephyr,console = &loguart;
+		zephyr,shell-uart = &loguart;
+		zephyr,sram = &sram0;
+		zephyr,flash = &flash0;
+	};
+};
+
+&flash0 {
+	reg = <0x04000020 DT_SIZE_M(4)>;
```
nordicjm, line 26: missing ranges property

scripts/west_commands/runners/amebaflash.py
fabiobaltieri: right, actually I think you could even do

`Path(ZEPHYR_BASE).parent / "modules" / ...`

zjian-zhang: Thanks, modified.

boards/realtek/rtl8721f_evb/rtl8721f_evb.dts
zjian-zhang: Thanks, modified.

nordicjm (approved):

fabiobaltieri (approved):

PR #101171, opened 2025-12-17
scripts: dts: gen_dts_cmake: Output compatibles for no prop. nodes

nordicjm:
Seemingly partitions (fixed-partitions and fixed-subpartitions) have no properties when the edt file is loaded, work around this issue by outputting compatibles for nodes that have them but have no properties

### DIFF
--- a/scripts/dts/gen_dts_cmake.py
+++ b/scripts/dts/gen_dts_cmake.py
@@ -118,21 +118,29 @@ def main():
         for label in node.labels:
             cmake_props.append(f'"DT_NODELABEL|{label}" "{node.path}"')
 
-        for item in node.props:
-            # We currently do not support phandles for edt -> cmake conversion.
-            if "phandle" not in node.props[item].type:
-                if "array" in node.props[item].type:
-                    # Convert array to CMake list
-                    cmake_value = ''
-                    for val in node.props[item].val:
-                        cmake_value = f'{cmake_value}{val};'
-                else:
-                    cmake_value = node.props[item].val
+        if node.props:
+            for item in node.props:
+                # We currently do not support phandles for edt -> cmake conversion.
+                if "phandle" not in node.props[item].type:
+                    if "array" in node.props[item].type:
+                        # Convert array to CMake list
+                        cmake_value = ''
+                        for val in node.props[item].val:
+                            cmake_value = f'{cmake_value}{val};'
+                    else:
+                        cmake_value = node.props[item].val
 
-                # Encode node's property 'item' as a CMake target property
-                # with a name like 'DT_PROP|<path>|<property>'.
-                cmake_prop = f'DT_PROP|{node.path}|{item}'
-                cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
+                    # Encode node's property 'item' as a CMake target property
+                    # with a name like 'DT_PROP|<path>|<property>'.
+                    cmake_prop = f'DT_PROP|{node.path}|{item}'
+                    cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
+        elif node.compats:
+            # Manually output compatibles for nodes that have no properties
+            cmake_value = ''
+            for val in node.compats:
+                cmake_value = f'{cmake_value}{val};'
+            cmake_prop = f'DT_PROP|{node.path}|compatible'
+            cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
 
         for comp in node.compats:
             compatible2paths[comp].append(node.path)


scripts/dts/gen_dts_cmake.py
```diff
@@ -118,21 +118,31 @@ def main():
...
+                            cmake_value = f'{cmake_value}{val};'
+                    else:
+                        cmake_value = node.props[item].val
+
+                    # Encode node's property 'item' as a CMake target property
+                    # with a name like 'DT_PROP|<path>|<property>'.
+                    cmake_prop = f'DT_PROP|{node.path}|{item}'
+                    cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
+        elif node.compats and ('fixed-partitions' in node.compats or \
+             'fixed-subpartitions' in node.compats):
```
pdgendt: I don't think we should be special casing some compatibles. Can't we just enable for all `node.compats`?

nordicjm: We already do, see https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/dts/python-devicetree/src/devicetree/edtlib.py#L1503 and https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/dts/gen_defines.py#L570

rruuaanng:
```suggestion
        elif (node.compats
              and ('fixed-partitions' in node.compats
              or 'fixed-subpartitions' in node.compats)):
```
(Using parentheses on a new line would be better.)

scripts/dts/gen_dts_cmake.py
```diff
@@ -118,21 +118,31 @@ def main():
...
+            # Manually output compatibles for fixed partition/subpartition nodes as they have no
+            # properties
+            cmake_value = ''
+            for val in node.compats:
+                cmake_value = f'{cmake_value}{val};'
```
rruuaanng, line 141:
```
def to_cmake_list(items: list) -> str:
    return ''.join(f'{item};' for item in items)
```
```suggestion
            cmake_value = to_cmake_list(node.compats)
```
It might be more readable.

scripts/dts/gen_dts_cmake.py
```diff
@@ -118,21 +118,31 @@ def main():
...
-                cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
+        if node.props:
+            for item in node.props:
+                # We currently do not support phandles for edt -> cmake conversion.
+                if "phandle" not in node.props[item].type:
+                    if "array" in node.props[item].type:
+                        # Convert array to CMake list
+                        cmake_value = ''
+                        for val in node.props[item].val:
+                            cmake_value = f'{cmake_value}{val};'
```
rruuaanng, line 129:
```suggestion
                        cmake_value = to_cmake_list(node.props[item].val)
```

scripts/dts/gen_dts_cmake.py
nordicjm: Updated

scripts/dts/gen_dts_cmake.py
nordicjm: This would be changing existing code which this PR does not do, it just further indents it, I do not want to make changes like this to a bug fix PR (they can be done in another optimisation PR)

scripts/dts/gen_dts_cmake.py
nordicjm: as above

scripts/dts/gen_dts_cmake.py
```diff
@@ -118,21 +118,31 @@ def main():
...
+                    else:
+                        cmake_value = node.props[item].val
+
+                    # Encode node's property 'item' as a CMake target property
+                    # with a name like 'DT_PROP|<path>|<property>'.
+                    cmake_prop = f'DT_PROP|{node.path}|{item}'
+                    cmake_props.append(f'"{cmake_prop}" "{escape(cmake_value)}"')
+        elif node.compats and \
+             ('fixed-partitions' in node.compats or 'fixed-subpartitions' in node.compats):
```
tejlmand: do we need to test specifically for partitions ?

We are already in the `elseif node.compats` of the `if node.props` meaning node matching `node.compats` has failed `node.props` test, so wouldn't it be nice to get all those available in CMake, partitions or not ?

tejlmand (commented):
Perhaps do the `elseif` for all nodes having  `compats` but not `props` set.

I see no reason to only do it for partitions, unless of course there are some performance penalties or similar.
Also it would remove the need for future adding `'xyz' in node.compats` when next node needing special handling comes along.

scripts/dts/gen_dts_cmake.py
nordicjm: Have updated

tejlmand (approved):

rruuaanng (approved):

pdgendt (approved):

de-nordic (approved):

PR #101172, opened 2025-12-17
net: dns: resolve: Enable discovery of all service instances

Ayush1325:
The DNS-based Service Discovery has the following abstract:

>    This document specifies how DNS resource records are named and
>    structured to facilitate service discovery.  Given a type of service
>    that a client is looking for, and a domain in which the client is
>    looking for that service, this mechanism allows clients to discover
>    a list of named instances of that desired service, using standard
>    DNS queries.  This mechanism is referred to as DNS-based Service
>    Discovery, or DNS-SD.

As is stated here, DNS-based service discovery is designed to find all instances implementing a service, not just the first one.

Currently, zephyr `dns_resolve_service` will call the callback for only the first client instance found. It still does receive the responses from other instances, but those are dropped since the query is already marked as finished. The seems incorrect behavior.

With that said, the changes in this commit are API breaking, since even the docs already state that only the first response is currently used. So I am fine with creating a new function for discovering all instances if that is more acceptable.

Tested with 3 BeagleConnect Freedoms (2 running mdns_responder and 1 running dns_resolve sample).

### DIFF
--- a/include/zephyr/net/dns_resolve.h
+++ b/include/zephyr/net/dns_resolve.h
@@ -516,6 +516,9 @@ struct dns_resolve_context {
 		 * cannot be used to find correct pending query.
 		 */
 		uint16_t query_hash;
+
+		/** Flag to indicate that the callback has been called at least once. */
+		bool cb_called;
 	} queries[DNS_NUM_CONCUR_QUERIES];
 
 	/** Is this context in use */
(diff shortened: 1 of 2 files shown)


jukkar (approved):
This looks reasonable although it is difficult to see all the possible corner cases just by looking the resolver source code.

rlubos (approved):

PR #101173, opened 2025-12-17
rtio: executor: release mempool buffer before resetting it

wouterh:
`rtio_executor_handle_multishot` releases the RX buffer in case of cancellation or error. However, in case of an sqe that uses a mempool, it reset the buffer before that, making the release a NOOP.  Fix that by moving the reset of the buffer to right before pushing the sqe back onto the queue.


ubieda (approved):
Thanks!

teburd (approved):

PR #101176, opened 2025-12-17
soc: nxp: imx943: m33: Define soc_clock_enable() only when needed

aescolar:
Define soc_clock_enable() with the same condition with which it is used to avoid the build warning `'soc_clock_enable' defined but not used.`

The issue was introduced in c520b3da1ae4c5ffc2433c3ab5c9357e4aeb921d

Can be seen failing in main in 
https://github.com/zephyrproject-rtos/zephyr/actions/runs/20274946738/job/58220894138#step:12:1049
and in PRs


aescolar:
CC @biwenli

PR #101177, opened 2025-12-17
Bluetooth: Mesh: clang unaligned access suppress warning

ivaniushkov:
1. added TOOLCHAIN_WARNING_UNALIGNED_ACCESS macro to allow suppressing -Wunaligned-access warnings
2. suppressed warnings in mesh net.c packed struct net_val declaration

Fixes #101144


pdgendt (commented):
Don't use `#101144` reference in the commit message (it has no meaning in clones). Use full links.

jhedberg (approved):

pdgendt (commented):
I wonder if the following need to use the `UNALIGNED_GET` helper macros?

https://github.com/zephyrproject-rtos/zephyr/blob/7c67dea76a9081f790766391cec1ac1302eec842/subsys/bluetooth/mesh/net.c#L1033-L1036

alxelax (approved):

PavelVPV (approved):

PavelVPV (changes requested):
I think title of the second commit should be changed to something like:
```
Bluetooth: mesh: suppress clang warning in net.c
```
to match area (there is no toolchain area in bluetooth subsystem).

PavelVPV (approved):

tejlmand (approved):

PR #101178, opened 2025-12-17
edtlib: Relax 'lacks binding' error for nodes used in a specifier

57300:
Suppose that node A is referenced by node B's `foos` specifier property.
Consequently, node A is required to set a `#foo-cells = <n>` property.

If n > 0, then edtlib needs to look up the binding of node A for a list of names under the `foo-cells` key.

If n = 0, then `foo-cells` is optional and, if absent from the binding, implied to be an empty list. However, edtlib would still expect node A to have a binding, even though it serves no purpose in this context.

Therefore, let's only raise the 'lacks binding' error when n > 0, and update an existing test case to demonstrate this scenario.

### DIFF
--- a/scripts/dts/python-devicetree/src/devicetree/edtlib.py
+++ b/scripts/dts/python-devicetree/src/devicetree/edtlib.py
@@ -1946,6 +1946,10 @@ class Node:
         # byte array.
 
         if not controller._binding:
+            if not data:
+                # As long as the 'controller' has '#<basename>-cells = <0>',
+                # it's not required to have a binding with '<basename>-cells:'.
+                return {}
             _err(f"{basename} controller {controller._node!r} "
                  f"for {self._node!r} lacks binding")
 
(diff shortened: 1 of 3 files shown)


mbolivar (changes requested):
I am against merging this PR. With the benefit of history, I think that attempting to guess things that are missing in bindings is a mistake.

I can't tell you how many times I've seen missing bindings for things in private repositories, just because people just wanted to stick a `reg` into the node and get back to work without taking the time to properly make the binding. I don't think we should encourage that type of behavior like is being done here.

Furthermore, I think this would silently hide an issue where there is a binding, but there is a typo in the compatible (or the file has been renamed but its `compatible` line was not updated, etc.)

mbolivar:
> There are several cases of that in edtlib today, but it sounds like you would support moving away from that direction in the near future.

I think we're stuck with the ones we have, but I don't want to add more, yeah.

mbolivar:
@57300 I'm going to close this for now to get it off my to-do list on the dashboard, but please feel free to reopen it at any time if you want to continue discussion.

PR #101183, opened 2025-12-17
drivers: rtc: add max31331 driver

gabz-arsenal:
This PR brings support for the MAX31331 Automotive RTC. The following features are supported:
- Alarm 1
- Trickle charger
- CLKO frequency output
- Compensation


bjarki-andreasen (commented):
Looks good, please run it through the the [rtc test suite](https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/drivers/rtc/rtc_api) and add it to the [build_all test suite](https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/drivers/build_all/rtc)

bjarki-andreasen (commented):
please squash the first 4 commits, otherwise looks good

drivers/rtc/rtc_max31331.c
```diff
@@ -0,0 +1,522 @@
...
+	uint8_t regs[2];
+	int16_t comp;
+
+	if (calibration < MAX31331_MIN_COMP || calibration > MAX31331_MAX_COMP) {
+		LOG_ERR("Calibration value %d ppb out of range! [min: %d ppb max: %d ppb]",
+			calibration, MAX31331_MIN_COMP, MAX31331_MAX_COMP);
+		return -EINVAL;
+	}
+
+	comp = nearbyintf((float)calibration / MAX31331_COMP_PPB);
```
bjarki-andreasen: There is no gain in using floating point arithmetic here, and it is generally not done in drivers due to the overhead in both size and execution time introduced. Instead, use
```
comp = DIV_ROUND_CLOSEST(calibration, MAX31331_COMP_PPB);
```
which is done entirely with fixed point math. Then `#include <math.h>` may be removed as well

bjarki-andreasen (changes requested):
Nice, last change request I think

drivers/rtc/rtc_max31331.c
gabz-arsenal: I was looking for something like this, thanks! Fixed

bjarki-andreasen:
> I just noticed that @ttmut pointed out another PR for the same device. Guess I'm at fault since it was opened before mine. What to do now @bjarki-andreasen? 

Oh, this happens every once in a while, it has happened specifically with another rtc, the guidelines state the first PR takes precedense, but that misses the point a bit. Best is to review each others PRs and combine the best parts, you are now THE most qualified reviewer of the other PR

PR #101184, opened 2025-12-17
modem: cmux: Add retry counter for SABM, CLD and DISC control messages

SeppoTakalo:
Add retry counter for opening and closing CMUX as well as opening and closing the DLCI channel.

Use same retry counter for both DLCI control messages as well as CMUX control messages as there is very minimal room for race condition. DLCI messages are only send when CMUX control channel is open.

Where relevant, use disconnect(cmux) for all closing calls. So we have one entry point for cleaning.

Similarly refactor modem_cmux_on_dlci_frame_dm() to dlci_close() as this is the single function to close and clean up a channel.

### DIFF
--- a/include/zephyr/modem/cmux.h
+++ b/include/zephyr/modem/cmux.h
@@ -177,6 +177,7 @@ struct modem_cmux {
 
 	/* State */
 	enum modem_cmux_state state;
+	uint8_t retry_count;
 	bool flow_control_on : 1;
 	bool initiator : 1;
 
--- a/subsys/modem/modem_cmux.c
+++ b/subsys/modem/modem_cmux.c
@@ -1024,7 +1024,8 @@ static void modem_cmux_on_control_frame(struct modem_cmux *cmux)
 	modem_cmux_log_received_frame(&cmux->frame);
 
 	if (is_connected(cmux) && cmux->frame.cr == cmux->initiator) {
-		LOG_DBG("Received a response frame");
+		LOG_DBG("Drop a response frame");
+		return;
 	}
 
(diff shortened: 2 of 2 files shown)


rerickson1 (dismissed):

SeppoTakalo:
Rebased to re-trigger the tests.
The failure `sample.kernel.synchronization on cpkcor_ra8d1b/r7fa8d1bhecbd error (CMake build failure - devicetree error)` is not relevant to this change.

tests/subsys/modem/modem_cmux/src/main.c
```diff
@@ -357,6 +357,8 @@ static void test_modem_cmux_before(void *f)
 
 	/* Reset mock pipes */
 	modem_backend_mock_reset(&bus_mock);
+	cmux.state = MODEM_CMUX_STATE_CONNECTED;
+	k_event_set(&cmux.event, BIT(cmux.state));
```
bjarki-andreasen: this looks like something went wrong, we should not need to directly change instance data of the cmux

SeppoTakalo: You were right.

It was a previous test-case that broke this, but somehow it got unnoticed before. There was a case, where event was not generated, as the state transition was not expected.


I added https://github.com/zephyrproject-rtos/zephyr/pull/101184/commits/a44e90db524d6c779d7d3551ed24dbfae73865cd to fix the testcase.

bjarki-andreasen (approved):

rerickson1 (approved):

PR #101187, opened 2025-12-17
modem: cmux: Change default timeouts to match 3GPP TS 27.010

SeppoTakalo:
3GPP TS 27.010 defines default timeouts as follows:
* Acknowledgement Timer (T1), min 10ms, max 2.5s default 100ms
* Response Timer for multiplexer control channel (T2), min 20ms, max 2.5s, default 300ms

The Response Timer for wake-up procedure(T3) is already a Kconfig and its default match the specification.


tests/subsys/modem/modem_cmux/src/main.c
```diff
@@ -370,7 +372,7 @@ ZTEST(modem_cmux, test_modem_cmux_receive_dlci2_at)
 	modem_backend_mock_put(&bus_mock, cmux_frame_dlci2_at_newline,
 			       sizeof(cmux_frame_dlci2_at_newline));
 
-	k_msleep(100);
+	k_msleep(10);
```
rerickson1: I know this is just a test, but this sleep time is used a lot. Probably best to have a define for it so its more clear as to it purpose.

SeppoTakalo:
> Could we make them Kconfigs? I think I remember using defaults from a Quectel modem when defining these, users may want to keep them (or raise them above the default level)

Yes. I made them Kconfig as the T3 timer was already Kconfig. Better to allow these tuned for specific modems.

tests/subsys/modem/modem_cmux/src/main.c
SeppoTakalo: Made a definition for this "TRANSMISSION_DELAY"

bjarki-andreasen (approved):
cool

rerickson1 (approved):

PR #101188, opened 2025-12-17
scripts: ci: Add ignorecase option to zephyr-keep-sorted

pdgendt:
Allow making the zephyr-keep-sorted check ignore cases, for example:

```rst
.. zephyr-keep-sorted-start re(^\w) ignorecase
Shell
=====
  * foo
STM32
=====
  * bar
.. zephyr-keep-sorted-stop
```

### DIFF
--- a/scripts/ci/check_compliance.py
+++ b/scripts/ci/check_compliance.py
@@ -2362,7 +2362,7 @@ class KeepSorted(ComplianceTest):
 
     MARKER = "zephyr-keep-sorted"
 
-    def block_check_sorted(self, block_data, *, regex, strip, fold):
+    def block_check_sorted(self, block_data, *, regex, strip, fold, icase):
         def _test_indent(txt: str):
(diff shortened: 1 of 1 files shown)


fabiobaltieri (approved):

mathieuchopstm (approved):

henrikbrixandersen (approved):

mathieuchopstm:
> Do we have a specific use-case in mind? I mean a file where this would be desired?

Documentation pages - a good candidate is the Migration Guide which currently has the `STM32` section is sorted above `Shell` because `'T' < 'h'`, even though `t` comes after `h` in the alphabet (causing confusion for someone expecting alphabetical ordering)

(when this is merged, I'll open a PR to fix the ordering in that file)

fabiobaltieri:
Let's move on with this, it's easy enough, does not affect the infra itself etc

PR #101196, opened 2025-12-17
net: ocpp: several OCPP fixes

markushx:
OCPP implementation had a few issues:
* Logging had a lot of trailing \n
* BootNotification did not transport Serial Number and other information, because the NULL check is wrong.
* statically allocated string variables are free'd
* zero-length NULL payload is tried to be printed in websocket, which crashes.


subsys/net/lib/ocpp/key_mgmt.c
```diff
@@ -48,74 +48,56 @@ struct ocpp_cfg_info {
...
+	FILL_KEY_TABLE(KEY_TYPE_INT, 1, CFG_HEARTBEAT_INTERVAL, "HeartbeatInterval", 0),
+	FILL_KEY_TABLE(KEY_TYPE_INT, 1, CFG_LIGHT_INTENSITY, "LightIntensity", 0),
+	FILL_KEY_TABLE(KEY_TYPE_BOOL, 1, CFG_LOCAL_AUTH_OFFLINE, "LocalAuthorizeOffline", 0),
+	FILL_KEY_TABLE(KEY_TYPE_BOOL, 1, CFG_LOCAL_PREAUTH, "LocalPreAuthorize", 0),
+	FILL_KEY_TABLE(KEY_TYPE_INT, 1, CFG_MAX_ENERGYON_INVL_ID, "MaxEnergyOnInvalidId", 0),
+	FILL_KEY_TABLE(KEY_TYPE_CSL, 1, CFG_MTR_VAL_ALGIN_DATA, "MeterValuesAlignedData",
+		       .str = NULL),
+	FILL_KEY_TABLE(KEY_TYPE_INT, 0, CFG_MTR_VAL_ALGIN_DATA_MAXLEN,
+		       "MeterValuesAlignedDataMaxLength", 1),
+	FILL_KEY_TABLE(KEY_TYPE_CSL, 1, CFG_MTR_VAL_SAMPLED_DATA, "MeterValuesSampledData",
```
rlubos: I have a hard time finding out what actually changed here, due to all the code style refactoring going on. Please **do not** introduce unrelated changes in commits. If you want to do a style cleanup, that's fine, but please do it in a separate commit.

markushx: @rlubos Indeed this is hard to review. I unfortunately amended my commit with the clang-format fixes that were required to get the CI green.

The main fixes are `.str = "0"`  to `.str = NULL` for `KEY_TYPE_CSL` (MeterValuesAlignedData, MeterValuesSampledData, ConnectorPhaseRotation, StopTxnAlignedData) which are free'd if written to. 

SupportedFeatureProfiles is not writable and thus not initialised with NULL.

markushx: I have dropped the unrelated formatting.

subsys/net/lib/websocket/websocket.c
```diff
@@ -697,6 +697,11 @@ int websocket_send_msg(int ws_sock, const uint8_t *payload, size_t payload_len,
 	NET_DBG("[%p] Len %zd %s/%d/%s", ctx, payload_len, opcode2str(opcode),
 		mask, final ? "final" : "more");
 
+	if (payload_len == 0 || payload == NULL) {
+		NET_DBG("DATA len 0 or NULL");
+	} else {
+		NET_DBG("DATA %.*s", payload_len, payload);
+	}
```
rlubos: I don't understand this change, the commit message says `OCPP pongs do not carry payload and cannot be NET_DBG'd without crash.`, but you didn't fix any existing debug message, just added new ones.

Also, I don't think the actual payload printout is correct, it assumes that the payload is an ascii string, while in fact it could just be any binary data.

rlubos (commented):
I think websocket change is a mistake, otherwise looks ok.

subsys/net/lib/websocket/websocket.c
markushx: My bad. It was my own debug message which was crashing this. Will drop the commit.

markushx:
> @markushx It seems you've dropped the wrong commit?

@rlubos Now the formatting should be really gone.

rlubos (approved):

ssekar15 (approved):

PR #101197, opened 2025-12-17
doc: contribute: guidelines: Document KeepSorted and options

pdgendt:
Add a documentation about the `KeepSorted` compliance check and which options it provides.

Depends on https://github.com/zephyrproject-rtos/zephyr/pull/101188

### DIFF
--- a/doc/contribute/guidelines.rst
+++ b/doc/contribute/guidelines.rst
@@ -616,6 +616,88 @@ before opening a new Pull Request:
    Set the default app to Strawberry Perl. By default the executable is
    installed at ``C:\Strawberry\perl\bin\perl.exe``.
 
+KeepSorted Check
+^^^^^^^^^^^^^^^^
+
+The KeepSorted check ensures that specified blocks of code, configuration, or
+documentation remain in sorted order.
+
+To use the KeepSorted check, wrap your sorted content between dedicated lines
+containing start and stop markers, typically using comments:
+
+.. code-block:: c
+
+   // zephyr-keep-sorted-start
+   option_a
+   option_b
+   option_c
+   // zephyr-keep-sorted-stop
+
+KeepSorted Marker Options
+"""""""""""""""""""""""""
+
+The sorting behavior of each block can be customized in several ways.
+To do this, one or more of the following parameters can be added on
+the same line as the start marker itself:
+
+**re(regex_pattern)**
+   Enables regex mode where only lines matching the specified regular expression
+   are checked for sorting. Other lines are ignored.
+
+   Example checking sorted properties in a yaml file:
+
+   .. code-block:: yaml
+
+      # zephyr-keep-sorted-start re(^\s+\- name:)
+      projects:
+        - name: application
+          revision: main
+        - name: library1
+          revision: feature-branch
+        - name: library2
+          revision: main
+      # zephyr-keep-sorted-stop
+
+**strip(characters)**
+   Strips the specified characters from lines before performing the sort comparison.
+   This is useful when lines have optional prefixes or suffixes that should be
+   ignored during sorting.
+
+   Example stripping quotes from yaml dictionary keys:
+
+   .. code-block:: yaml
+
+      # zephyr-keep-sorted-start strip(":)
+      ACPI:
+        status: odd fixes
+      "West project: acpica":
+        status: odd fixes
+      # zephyr-keep-sorted-stop
+
+**nofold**
+   Disables line folding. By default, indented lines following a main line are
+   concatenated (folded) together for sorting comparison. The ``nofold`` option
+   disables this behavior and ignores indented lines.
+
+**ignorecase**
+   Enables case-insensitive sorting using Python's `str.casefold`_. This allows
+   mixing uppercase and lowercase items in sorted blocks without causing sort
+   order violations. Defaults to Python's string ordering if omitted.
+
+.. _str.casefold: https://docs.python.org/3/library/stdtypes.html#str.casefold
+
+Multiple options can be combined on the same marker line:
+
+.. code-block:: rst
+
+   .. zephyr-keep-sorted-start re(^\* \w) ignorecase
+   * Shell
+     Some important message about the shell.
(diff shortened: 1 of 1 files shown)


doc/contribute/guidelines.rst
```diff
@@ -616,6 +616,85 @@ before opening a new Pull Request:
...
+**nofold**
+   Disables line folding. By default, indented lines following a main line are
+   concatenated (folded) together for sorting comparison. The ``nofold`` option
+   disables this behavior and ignores indented lines.
+
+**ignorecase**
+   Enables case-insensitive sorting using Python's
+   `casefold https://docs.python.org/3/library/stdtypes.html#str.casefold`_. This allows
+   mixing uppercase and lowercase items in sorted blocks without causing sort
+   order violations.
```
mathieuchopstm: Maybe indicate the default sort order if not present (i.e., `default Python string sort order`)

doc/contribute/guidelines.rst
```diff
@@ -616,6 +616,85 @@ before opening a new Pull Request:
    Set the default app to Strawberry Perl. By default the executable is
    installed at ``C:\Strawberry\perl\bin\perl.exe``.
 
+KeepSorted Check
+^^^^^^^^^^^^^^^^
+
+The KeepSorted check ensures that specified blocks of code, configuration, or
+documentation remain in sorted order.
+
+To use the KeepSorted check, wrap your sorted content with marker comments:
```
mathieuchopstm: Could be a bit more explicit about `zephyr-keep-sorted-{start,stop}` being the magic marker, and the "comment" just being needed so that the file remains valid.

Something like:

```suggestion
To use the KeepSorted check, wrap your sorted content between a dedicated line containing
start marker ``zephyr-keep-sorted-start`` and a dedicated line containing end marker
``zephyr-keep-sorted-end`` (usually, these marker lines will be comments):
```

doc/contribute/guidelines.rst
```diff
@@ -616,6 +616,85 @@ before opening a new Pull Request:
...
+   Disables line folding. By default, indented lines following a main line are
+   concatenated (folded) together for sorting comparison. The ``nofold`` option
+   disables this behavior and ignores indented lines.
+
+**ignorecase**
+   Enables case-insensitive sorting using Python's
+   `casefold https://docs.python.org/3/library/stdtypes.html#str.casefold`_. This allows
+   mixing uppercase and lowercase items in sorted blocks without causing sort
+   order violations.
+
```
mathieuchopstm, line 688: I'd provide a simple example with only one option:
```
Options are provided on the same line as the start marker. For example, the following
<ensures only lines matching regexp will be sorted>:

    zephyr-keep-sorted-start re(........)
    xxxx
    yyyy
    zephyr-keep-sorted-end
```

doc/contribute/guidelines.rst
pdgendt: We can't put these inline as they would be picked up by the compliance script 😆

doc/contribute/guidelines.rst
pdgendt: The regex option has an example already?

doc/contribute/guidelines.rst
mathieuchopstm: oh right :sob: - but maybe we can keep something like this without the actual markers then?

doc/contribute/guidelines.rst
mathieuchopstm: I took regexp as an example but it could be anything. As you said, there are examples already, so this is a bit silly... but the current wording feels a bit rough:

>  The KeepSorted start marker supports several optional parameters to
customize the sorting behavior:
> *<list of options>*

See new review comment for alternative proposal.

doc/contribute/guidelines.rst
```diff
@@ -616,6 +616,87 @@ before opening a new Pull Request:
...
+   option_a
+   option_b
+   option_c
+   // zephyr-keep-sorted-stop
+
+KeepSorted Marker Options
+"""""""""""""""""""""""""
+
+The KeepSorted start marker supports several optional parameters to
+customize the sorting behavior:
```
mathieuchopstm: cont. of https://github.com/zephyrproject-rtos/zephyr/pull/101197#discussion_r2627777557

```suggestion
The sorting behavior of each block can be customized in several ways.
To do this, one or more of the following parameters can be added on
the same line as the start marker itself:
```

doc/contribute/guidelines.rst
```diff
@@ -616,6 +616,88 @@ before opening a new Pull Request:
...
+   mixing uppercase and lowercase items in sorted blocks without causing sort
+   order violations. Defaults to Python's string ordering if omitted.
+
+.. _str.casefold: https://docs.python.org/3/library/stdtypes.html#str.casefold
+
+Multiple options can be combined on the same marker line:
+
+.. code-block:: rst
+
+   .. zephyr-keep-sorted-start re(^\* \w) ignorecase
```
mathieuchopstm, line 693: Nit: the migration guide uses `re(^\w)` :slightly_smiling_face:

mathieuchopstm (approved):

doc/contribute/guidelines.rst
pdgendt: This sample has bullet points 🔫

fabiobaltieri (approved):
Not super enthusiastic about having out of band documentation for stuff like this to be totally honest, it's one more bit of text that we have to somehow sync with the implementation that happens in another file somewhere in the repository. Thinking maybe if this was its own utility we could have an helptext with the documentation that lives in the same file as the code and then the doc here could expand it from there automatically. Not practical in this case though since it's part of the compliance script, which should probably be refactored into modules at this point. Ah well.

doc/contribute/guidelines.rst
mathieuchopstm: oh right :facepalm:

keith-zephyr (approved):

PR #101200, opened 2025-12-17
xtensa: mmu: (cosmetic) clean up pointer types

lyakh:
Use void * cleanly: avoid needless type-casts and use void * for generic pointers instead of uint8_t *.


lyakh:
> IIRC, this was done because xt-clang was complaining about pointer arithmetic. Is it not complaining anymore?

@dcpleung I guess you mean changes in mm_drv_common.c. I don't see any locations there where affected variables would be used for pointer arithmetic. I only see them used as function arguments.

dcpleung (approved):

kv2019i (approved):

PR #101203, opened 2025-12-17
net: sockets: tls: Add support for multiple client sessions in DTLS server socket

### LINKED ISSUE #64954: DTLS server socket support multiple clients
The current DTLS server socket implementation only supports a single client peer, because it has a single TLS context per socket file descriptor.
A side effect of this limitation is that, if the same host client initializes multiple handshakes on different ports, the server responds to all requests on the initial port.

 An improvement would be to have a TLS context for each peer (`host:port` combination).

An idea I briefly looked into, is to split the `subsys/net/lib/sockets/sockets_tls.c` file into a TLS and a DTLS part, but nothing concrete.

**Additional context**
N/A

rlubos:
Implement support for multiple client sessions per DTLS server socket. This includes some major refactoring of the DTLS server socket functionality, including splitting the TLS session context from the TLS socket context. DTLS server socket will allocate new session context for packets coming from unknown clients. TLS client/server and DTLS client functionality shouldn't be affected, as those scenarios only support a single session per socket.

Resolves #64954


jukkar (approved):

subsys/net/lib/sockets/sockets_tls.c
```diff
@@ -890,29 +1055,259 @@ static int dtls_rx(void *ctx, unsigned char *buf, size_t len)
...
+	}
+
+	dtls_peer_address_set(session_ctx, addr, addrlen);
+
+	ret = mbedtls_ssl_set_client_transport_id(&session_ctx->ssl,
+						  (const unsigned char *)addr,
+						  addrlen);
+	if (ret < 0) {
+		tls_session_free(session_ctx);
+		return -ENOMEM;
```
pdgendt, line 1132: Should `ret` be returned here?

pdgendt (approved):
Nice work!

Tested using a CoAP server with multiple clients.

subsys/net/lib/sockets/sockets_tls.c
rlubos: `ret` will contain an error code from mbed TLS error space (result of `mbedtls_ssl_set_client_transport_id()`) so we need to translate to errno.

PR #101205, opened 2025-12-17
drivers: pwm: add STM32 LPTIM PWM driver

### LINKED ISSUE #57316: lptim pwm support for stm32
**Is your feature request related to a problem? Please describe.**
Currently it is not possible to use the lptimers of a stm32 as a pwm source. 

Using lptimers the same way as normal timers for pwm generation.

Using the HAL.

**Additional context**
The devictree binding "st,stm32-lptim" has the st,prescaler in it's property blocklist. I think this value is required  by the pwm interface. Furthermore the lptim2 is not defined in the devicetree of the stm32u5xx platform.

tpambor:
This PR introduces a PWM driver for the STM32 low-power timer (LPTIM) peripheral on STM32 MCUs.

This driver is loosely based on the PWM driver for standard timers on STM32. For maintainability and readability, I split this into a new driver instead of integrating with the existing STM PWM driver, as the supported features differ and LPTIMs use a different low-level API `LL_LPTIM_*`.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/57316.


drivers/pwm/pwm_stm32_lptim.c
```diff
@@ -0,0 +1,249 @@
...
+static uint32_t get_polarity(pwm_flags_t flags)
+{
+	/* PWM_POLARITY_NORMAL (active-high) pulses correspond to
+	 * LL_LPTIM_OUTPUT_POLARITY_INVERSE and vice versa.
+	 */
+	if ((flags & PWM_POLARITY_MASK) == PWM_POLARITY_NORMAL) {
+		return LL_LPTIM_OUTPUT_POLARITY_INVERSE;
+	}
+
+	return LL_LPTIM_OUTPUT_POLARITY_REGULAR;
```
gautierg-st:
```suggestion
	return (flags & PWM_POLARITY_MASK) == PWM_POLARITY_NORMAL ?
	        LL_LPTIM_OUTPUT_POLARITY_INVERSE : LL_LPTIM_OUTPUT_POLARITY_REGULAR;
```

drivers/pwm/pwm_stm32_lptim.c
```diff
@@ -0,0 +1,249 @@
...
+
+	LL_LPTIM_Enable(timer);
+
+	/* Start the LPTIM counter in continuous mode */
+	LL_LPTIM_StartCounter(timer, LL_LPTIM_OPERATING_MODE_CONTINUOUS);
+
+	return 0;
+}
+
+#define PWM(index) DT_DRV_INST(index)
```
gautierg-st, line 215: Ideally, LPTIM PWM driver should be as close as possible to TIM PWM driver. So there should be a `pwm` child node to the `lptim` instances in the dtsi files.
And this line, like it's done in the TIM PWM driver, should be
```suggestion
#define PWM(index) DT_INST_PARENT(index)
```

samples/drivers/led/pwm/boards/stm32h573i_dk.overlay
gautierg-st, line 1: Please split the board overlay in a different commit.

gautierg-st (changes requested):
Thanks for this. Driver side looks ok to me.
As I said in one comment, a `pwm` child node should be added in the `lptim` nodes of the dtsi, like it is done for the `timers` nodes. This will require some minor changes to the driver and the overlay.
And make that 3 commits:
- Driver
- Dtsi update
- Tests/samples

drivers/pwm/pwm_stm32_lptim.c
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		r = clock_control_get_rate(clk, (clock_control_subsys_t)&cfg->pclken[1], &tim_clk);
+		if (r < 0) {
+			LOG_ERR("Timer clock rate get error (%d)", r);
+			return r;
+		}
+	} else {
+		LOG_ERR("Timer clock source is not specified");
+		return -EINVAL;
```
etienne-lms: Maybe `PWM_DEVICE_INIT()` could add `BUILD_ASSERT(DT_NUM_CLOCKS(PWM(index)) == 2)` to trap the error at build time instead of at run time?

drivers/pwm/pwm_stm32_lptim.c
```diff
@@ -0,0 +1,249 @@
...
+	if (r < 0) {
+		LOG_ERR("PWM pinctrl setup failed (%d)", r);
+		return r;
+	}
+
+	LOG_INF("Clock rate: %u Hz, Prescaler: %u", data->tim_clk, cfg->prescaler);
+
+	/* Initialize timer */
+	LL_LPTIM_SetClockSource(timer, LL_LPTIM_CLK_SOURCE_INTERNAL);
+	LL_LPTIM_SetPrescaler(timer, (__CLZ(__RBIT(cfg->prescaler)) << LPTIM_CFGR_PRESC_Pos));
```
etienne-lms: Nitpicking: could remove an inner parentheses pair:
```suggestion
	LL_LPTIM_SetPrescaler(timer, __CLZ(__RBIT(cfg->prescaler)) << LPTIM_CFGR_PRESC_Pos);

File: drivers/pwm/pwm_stm32_lptim.c (hunk 2 above)
Comment (tpambor) [reply drivers/pwm/pwm_stm32_lptim.c:215]: I generally agree, but this would require major refactoring and introduce breaking changes. The current top-level `st,stm32-lptim` binding is tailored specifically for using LPTIM as a system tick source. Therefore, I see the following steps:
1. Introduce a new top-level binding: `st,stm32-lptimers`, essentially based on `st,stm32-lptim` but without the tick-source-specific `st,timeout` property.
2. Add a new binding for tick source functionality and include it as a subnode within LPTIM instances.
3. Update the current LPTIM tick source driver to align with the new structure.
4. Update samples and tests to reflect these changes.
5. Update the binding of this PWM driver so that it can be added as a subnode to LPTIM instances that support PWM.
6. Update this PWM driver

All users using LPTIM as systick source would then need to update their device tree to adapt to these changes.

Comment (gautierg-st) [reply drivers/pwm/pwm_stm32_lptim.c:215]: You're right, I didn't have the full picture in mind.
Even though it's a lot more work, I still think it's a better way to go. Changing the compatible value of a node doesn't seem like a good idea for readability and maintenance.
Besides, it's better if the dtsi can describe all available features.
And this could also be used to add support for other potential features of the LPTIM in the future.
But before diving into it, let's have other opinions: @erwango, @etienne-lms ?

File: samples/drivers/led/pwm/boards/stm32h573i_dk.overlay [hunk 5]
```diff
@@ -0,0 +1,34 @@
...
+		red_pwm_led: red_pwm_led {
+			pwms = <&lptim5 2 PWM_MSEC(20) PWM_POLARITY_INVERTED>;
+			label = "LD3 - red led";
+		};
+	};
+};
+
+&lptim5 {
+	compatible = "st,stm32-lptim-pwm";
+	clocks = <&rcc STM32_CLOCK(APB3, 14)>,
```
JarmouniA, line 24: I see no point in inventing another compatible for the same hw device (https://github.com/zephyrproject-rtos/zephyr/blob/72360f821a610b64fda3bd300b6ca4577f6abe8c/dts/arm/st/h5/stm32h562.dtsi#L106) just to use it with another driver class.
Add a child PWM node and use it with the PWM driver.

JarmouniA (changes requested):

drivers/pwm/pwm_stm32_lptim.c
JarmouniA: > Introduce a new top-level binding: st,stm32-lptimers, essentially based on st,stm32-lptim but without the tick-source-specific st,timeout property.

Why is that? I must be missing something, the property is not required, so just don't use it in the PWM driver.
https://github.com/zephyrproject-rtos/zephyr/blob/72360f821a610b64fda3bd300b6ca4577f6abe8c/dts/bindings/timer/st%2Cstm32-lptim.yaml#L48

samples/drivers/led/pwm/boards/stm32h573i_dk.overlay
JarmouniA: Related #105249

drivers/pwm/pwm_stm32_lptim.c
FRASTM: > > Introduce a new top-level binding: st,stm32-lptimers, essentially based on st,stm32-lptim but without the tick-source-specific st,timeout property.
> 
> Why is that? I must be missing something, the property is not required, so just don't use it in the PWM driver.
> 
> [zephyr/dts/bindings/timer/st,stm32-lptim.yaml](https://github.com/zephyrproject-rtos/zephyr/blob/72360f821a610b64fda3bd300b6ca4577f6abe8c/dts/bindings/timer/st%2Cstm32-lptim.yaml#L48)
> 
> Line 48 in [72360f8](/zephyrproject-rtos/zephyr/commit/72360f821a610b64fda3bd300b6ca4577f6abe8c)
> 
>  st,timeout:

As @JarmouniA says, the `st,timeout` is not a "required" property, 
Then [‎dts/bindings/pwm/st,stm32-lptim-pwm.yaml](https://github.com/zephyrproject-rtos/zephyr/pull/101205/changes/6caabeef934bbe77a112e8af85c22ccbae713772#diff-e5805b235f97413bbd49bb0ed490942b9fc23fb646b2da7d664090587bb2a6be) is not needed, a compatible: "st,stm32-pwm"  is fine

FRASTM (changes requested):

FRASTM:
@tpambor do you plan to continue working on that PR ?

tpambor:
> @tpambor do you plan to continue working on that PR ?

I am still interested but I currently have no plans in short term to work on this.

PR #101218, opened 2025-12-17
Sctimer pwm capture

FelixWang47831:
1.Add mcux_sctimer_pwm_update_polarity() function to properly reconfigure PWM output polarity when duty cycle is updated

2.Add input-channels and inputmux-connections properties to the
    nxp,sctimer-pwm device tree binding to support SCTimer input
    capture configuration.

3.Implement PWM input capture functionality for the NXP SCTimer
    driver with support for pulse width and period measurement
    in single-shot and continuous modes

4.Enable test on lpcxpresso55s36

### DIFF
--- a/drivers/pwm/pwm_mcux_sctimer.c
+++ b/drivers/pwm/pwm_mcux_sctimer.c
@@ -13,12 +13,17 @@
 #include <fsl_clock.h>
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/drivers/clock_control.h>
-
+#ifdef CONFIG_PWM_CAPTURE
+#include <zephyr/irq.h>
+#include <fsl_inputmux.h>
+#endif
 #include <zephyr/logging/log.h>
 
 LOG_MODULE_REGISTER(pwm_mcux_sctimer, CONFIG_PWM_LOG_LEVEL);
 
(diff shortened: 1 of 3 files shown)


drivers/pwm/pwm_mcux_sctimer.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, NXP
+ * Copyright (c) 2021, 2025 NXP
```
yvesll, line 2: Can squash commit 3 to commit 1?

FelixWang47831: Fixed.

yvesll: I mean merge the two commits as both of them modify the same file.

FelixWang47831: They're two different code change, one for PWM set cycles(Generate PWM wave) bug fix, one for capture(Measure input signal). That's why use two patches

Holt-Sun (dismissed):

Holt-Sun (approved):

mmahadevan108 (approved):

PR #101220, opened 2025-12-17
boards: cyw920829m2evk: flash configuration moved to board

MerinMGeorge:
- CYW920829 configuration from the flash driver is moved to modules hal_infineon zephyr-ifx-cycfg folder


modules/hal_infineon/CMakeLists.txt
```diff
@@ -113,3 +113,7 @@ if(CONFIG_SOC_FAMILY_INFINEON_PSOC4)
   ## Add abstraction-rtos sources
   add_subdirectory(abstraction-rtos)
 endif()
+
+if(CONFIG_SOC_SERIES_CYW20829)
+  add_subdirectory(zephyr-ifx-cycfg)
```
sreeramIfx: given that we are checking for the series in the modules/hal_infineon/zephyr-ifx-cycfg/CMakeLists.txt, Can we include the sub-folder here unconditionally?

MerinMGeorge: That is fine. But this is the approach that i see in this CMakeLists.txt for other soc's. ( CONFIG_SOC_SERIES_PSC3 and CONFIG_SOC_FAMILY_INFINEON_EDGE). If we are going ahead based on your proposal, i think that we should change the other's as well. Let me know what you think

sreeramIfx: yes, we should remove the other instances and have one add_subdirectory(zephyr-ifx-cycfg) in this file

MerinMGeorge: Fixed

teburd (dismissed):

teburd:
To be clear, the changes are good, needs to pass CI still to get approved and merged.

teburd (commented):
Commit message needs a little work. The title should match the addition not the subtraction. The message should cover the what and the why.

For example something like...

```
boards: cyw920829m2evk: Moves flash configuration from driver to board

Flash configuration was previously being done in the driver selected by Kconfig. Moves flash configuration to the board removing the need for ifdefs in the driver. This will allow boards to configure specifics related to the on board (not part of the SoC) flash.

Signed-off-by: Merin George <merin.george@infineon.com>
```

We ideally want to see in the commit message what you are changing and why.

jsbatch (approved):

sreeramIfx (approved):

PR #101221, opened 2025-12-17
drivers: sdhc: stm32: skip semaphore wait on HAL error

iabdalkader:
When HAL_SDIO_WriteExtended_DMA or HAL_SDIO_ReadExtended_DMA returns an error, interrupts are not enabled so the semaphore would never be given, and it will always timeout.

Some drivers expect certain functions to fail during normal operation. For example, WHD calls whd_kso_enable during sleep/wakeup cycles which can fail. This would waste 200ms (CONFIG_SD_CMD_TIMEOUT default) on every such failure.

For reference, the logic of `HAL_SDIO_WriteExtended_DMA` and `HAL_SDIO_ReadExtended_DMA` looks like this:

  ```c
  HAL_StatusTypeDef HAL_SDIO_WriteExtended_DMA(...) {
      if (params == NULL)
          return HAL_ERROR;              // No ISR enabled

      if (state != READY)
          return HAL_BUSY;               // No ISR enabled

      // ... setup DMA, configure DPSM ...

      errorstate = SDMMC_SDIO_CmdReadWriteExtended(...);
      if (errorstate != HAL_SDIO_ERROR_NONE) {
          hsdio->ErrorCode |= errorstate;
          if (errorstate != (ADDR_OUT_OF_RANGE | ILLEGAL_CMD | COM_CRC_FAILED | GENERAL_UNKNOWN_ERR)) {
              // cleanup
              return HAL_ERROR;          // No ISR enabled
          }
      }

      // Enable interrupts - only reached if no early return
      __HAL_SDIO_ENABLE_IT(hsdio, (DCRCFAIL | DTIMEOUT | TXUNDERR | DATAEND));

      return HAL_OK;
  }
```

### DIFF
--- a/drivers/sdhc/sdhc_stm32.c
+++ b/drivers/sdhc/sdhc_stm32.c
@@ -334,6 +334,12 @@ static int sdhc_stm32_rw_extended(const struct device *dev, struct sdhc_command 
 	}
 
 	if (!IS_ENABLED(CONFIG_SDHC_STM32_POLLING_SUPPORT)) {
+		/* Only wait on semaphore if HAL function succeeded */
+		if (res != HAL_OK) {
+			k_free(dev_data->sdio_dma_buf);
+			return -EIO;
+		}
+
 		/* Wait for whole transfer to complete */
 		if (k_sem_take(&dev_data->device_sync_sem, K_MSEC(CONFIG_SD_CMD_TIMEOUT)) != 0) {
 			k_free(dev_data->sdio_dma_buf);

### CODE REVIEW

File: drivers/sdhc/sdhc_stm32.c
Comment (mathieuchopstm) [inline drivers/sdhc/sdhc_stm32.c:1]: How about the following instead? avoids an early return
```diff

	if (!IS_ENABLED(CONFIG_SDHC_STM32_POLLING_SUPPORT)) {
-		/* Wait for whole transfer to complete */
-		if (k_sem_take(&dev_data->device_sync_sem, K_MSEC(CONFIG_SD_CMD_TIMEOUT)) != 0) {
-			k_free(dev_data->sdio_dma_buf);
-			return -ETIMEDOUT;
-		}
-
-		if (direction == SDIO_IO_READ) {
-			sys_cache_data_invd_range(dev_data->sdio_dma_buf,
						dev_data->total_transfer_bytes);
-			memcpy(data->data, dev_data->sdio_dma_buf, data->block_size * data->blocks);
-		}
+		if (res == HAL_OK) {
+			/* Wait for whole transfer to complete */
+			if (k_sem_take(&dev_data->device_sync_sem, K_MSEC(CONFIG_SD_CMD_TIMEOUT)) != 0) {
+				k_free(dev_data->sdio_dma_buf);
+				return -ETIMEDOUT;
+			}
+
+			if (direction == SDIO_IO_READ) {
+				sys_cache_data_invd_range(dev_data->sdio_dma_buf,
+							dev_data->total_transfer_bytes);
+				memcpy(data->data, dev_data->sdio_dma_buf, data->block_size * data->blocks);
+			}
+		}

		k_free(dev_data->sdio_dma_buf);
	}
```

iabdalkader: I considered the same, but it's more indentation and might be harder to follow. I don't mind using that though.

ExaltZephyr:
> @ExaltZephyr Can you have a look ?

@erwango  yes we are ok with this change and we have tested it on Arduino Portenta, Nicla Vision, and Giga R1.

ExaltZephyr (approved):

erwango (approved):

etienne-lms (approved):

PR #101237, opened 2025-12-18
toolchain: iar: fix some potential issues for the IAR build tool

thenguyenyf:
This PR introduces fixes for potential issues that could cause build errors or warnings, or result in failed kernel test suites, when using the IAR build tool.

### DIFF
--- a/cmake/compiler/iar/target.cmake
+++ b/cmake/compiler/iar/target.cmake
@@ -160,3 +160,10 @@ endforeach()
 foreach(F ${IAR_ASM_FLAGS})
   list(APPEND TOOLCHAIN_C_FLAGS $<$<COMPILE_LANGUAGE:ASM>:${F}>)
 endforeach()
+
+# --------------------------------------------------------------------
+# Disable GCC-specific runtime library detection for IAR
+# --------------------------------------------------------------------
+function(compiler_set_linker_properties)
+  message(STATUS "Skipping GCC-specific runtime lib detection for IAR toolchain")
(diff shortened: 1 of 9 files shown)


tejlmand (dismissed):

cmake/compiler/iar/target.cmake
```diff
@@ -160,3 +160,10 @@ endforeach()
 foreach(F ${IAR_ASM_FLAGS})
   list(APPEND TOOLCHAIN_C_FLAGS $<$<COMPILE_LANGUAGE:ASM>:${F}>)
 endforeach()
+
+# --------------------------------------------------------------------
+# Disable GCC-specific runtime library detection for IAR
+# --------------------------------------------------------------------
+function(compiler_set_linker_properties)
+  message(STATUS "Skipping GCC-specific runtime lib detection for IAR toolchain")
```
tejlmand, line 168: No reason for the status message.
```suggestion
```

tejlmand (approved):
lgtm, but would like @RobinKastberg to also take a look.

RobinKastberg (approved):
Thanks for your work!

If you find a particular warning too noisy, please consider adding it to 
https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/compiler/iar/compiler_flags.cmake instead. Failing to inline is more of a compiler shortcoming than an actual problem with the users code, it can probably be suppressed globally or downgraded to an remark, or what do you say @LoveKarlsson ?

PR #101239, opened 2025-12-18
linker: ld: correct GNU linker generator output for block linking

thenguyenyf:
Fix the issue that occurs when building with GNU and CONFIG_LINKER_GENERATOR=y.

Currently, the GROUP {region} directive does not function properly with GNU. The input section cannot link to the memory group defined by "zephyr,memory-region" in the final output linker script.

This commit updates the GNU section name parser script, allowing it to look up both group and region for output block generation in the final output linker script.


tejlmand (approved):

57300 (approved):
LGTM

PR #101245, opened 2025-12-18
Fixed struct lll_scan *aux_scan_lll spell mistake in ull_scan.c

jackjinxb:
For uint8_t ull_scan_disable(uint8_t handle, struct ll_scan_set *scan) in line 673,in ull_scan.c file. I think it should be struct lll_scan *aux_scan_lll; because struct ll_scan_aux_set parent is lll_scan or lll_sync.


subsys/bluetooth/controller/ll_sw/ull_scan.c
```diff
@@ -670,7 +670,7 @@ uint8_t ull_scan_disable(uint8_t handle, struct ll_scan_set *scan)
 	/* Find and stop associated auxiliary scan contexts */
 	for (uint8_t aux_handle = 0; aux_handle < CONFIG_BT_CTLR_SCAN_AUX_SET;
 	     aux_handle++) {
-		struct lll_scan_aux *aux_scan_lll;
+		struct lll_scan *aux_scan_lll;
```
cvinayak, line 673:
```suggestion
		struct lll_scan_aux *aux_scan_lll;
```

Mostly likely `aux_scan_lll = aux->parent;`  is `struct lll_scan_aux *` under ULL scheduled reception of Auxiliary PDU reception. Actually, we are compare only pointer values at line 684, hence any pointer type should be ok here.

cvinayak (changes requested):

subsys/bluetooth/controller/ll_sw/ull_scan.c
jackjinxb: Yes,it work fine.So your suggestion donot modify?

cvinayak: no change.

PR #101254, opened 2025-12-18
Enable touch driver tma525b

KATE-WANG-NXP:
1.Add support for the Parade Tech TMA525B capacitive touch controller. The driver supports both interrupt-driven and polling modes, and can handle up to 4 simultaneous touch points.
2.Add support for the TMA525B touch controller to the  zc143ac72mipi display shield.
3.Adds the zc143ac72mipi shield configuration for the mimxrt700_evk platform to the LVGL demos sample test platforms


drivers/input/kconfig.tma525b
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig INPUT_TMA525B
+	bool "TMA525B capacitive touch controller"
+	default y
+	depends on DT_HAS_PARADE_TMA525B_ENABLED
+	select I2C
+    select INPUT_TOUCH
```
pdgendt:
```suggestion
	select INPUT_TOUCH
```

pdgendt (commented):
Fix compliance issues, please

boards/shields/zc143ac72mipi/zc143ac72mipi.overlay
```diff
@@ -15,6 +16,27 @@
 		enable-gpios = <&nxp_mipi_connector 32 GPIO_ACTIVE_HIGH>;
 		regulator-boot-on;
 	};
+
+	lvgl_pointer {
```
faxe1008: Not needed since the introduction of  `CONFIG_LV_Z_POINTER_FROM_CHOSEN_TOUCH` adding a pseudo device to the board is not needed anymore since you have set `zephyr,touch`.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+#define TMA525B_TOUCH_DATA_LEN_BYTES 2U
+
+/* TMA525B subaddress size */
+#define TMA525B_SUBADDR_SIZE 1U
+
+/* TMA525B report ID for touch data */
+#define TMA525B_REPORT_ID_TOUCH 0x01U
+
+/* Touch event types - matching fsl_tma525b.h */
+typedef enum _touch_event {
```
faxe1008: enums are usually not aliased and members are capitalized, without a `k` prefix.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+	/* 2 MSB: report counter, 3 LSB: noise effects. Not used in current driver model. */
+	uint8_t report_noise;
+	struct tma525b_touch_point touch[TMA525B_MAX_TOUCHES];
+} __packed;
+
+/* Macros for extracting touch data - matching SDK implementation */
+#define TMA525B_TOUCH_DATA_GET_ID(td) ((td).report_id)
+#define TMA525B_TOUCH_DATA_GET_LENGTH(td)                                                          \
+	(uint16_t)((((uint16_t)(td).length_h) << 8) | (uint16_t)(td).length_l)
+#define TMA525B_TOUCH_DATA_GET_COUNT(td) ((td).num_touch)
```
faxe1008: This macro has one usage - this is not needed.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+	uint8_t timestamp_100us_h;
+	uint8_t num_touch; /* Number of touch points detected */
+	/* 2 MSB: report counter, 3 LSB: noise effects. Not used in current driver model. */
+	uint8_t report_noise;
+	struct tma525b_touch_point touch[TMA525B_MAX_TOUCHES];
+} __packed;
+
+/* Macros for extracting touch data - matching SDK implementation */
+#define TMA525B_TOUCH_DATA_GET_ID(td) ((td).report_id)
+#define TMA525B_TOUCH_DATA_GET_LENGTH(td)                                                          \
```
faxe1008: Same here

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+	/* Timestamp in 100us units high byte. Not used in current driver model. */
+	uint8_t timestamp_100us_h;
+	uint8_t num_touch; /* Number of touch points detected */
+	/* 2 MSB: report counter, 3 LSB: noise effects. Not used in current driver model. */
+	uint8_t report_noise;
+	struct tma525b_touch_point touch[TMA525B_MAX_TOUCHES];
+} __packed;
+
+/* Macros for extracting touch data - matching SDK implementation */
+#define TMA525B_TOUCH_DATA_GET_ID(td) ((td).report_id)
```
faxe1008: Not needed

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+#define TMA525B_TOUCH_DATA_GET_LENGTH(td)                                                          \
+	(uint16_t)((((uint16_t)(td).length_h) << 8) | (uint16_t)(td).length_l)
+#define TMA525B_TOUCH_DATA_GET_COUNT(td) ((td).num_touch)
+
+#define TMA525B_TOUCH_POINT_GET_ID(tp)    ((tp).event_id & 0x1F)
+#define TMA525B_TOUCH_POINT_GET_EVENT(tp) ((touch_event_t)(uint8_t)(((tp).event_id & 0x60) >> 5U))
```
faxe1008: This one might be ok to leave but you do not need to cast to a uint8_t first

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+	struct gpio_dt_spec pwr_gpio;
+	struct gpio_dt_spec rst_gpio;
+	struct gpio_dt_spec int_gpio;
+};
+
+/* TMA525B runtime data */
+struct tma525b_data {
+	const struct device *dev;
+	struct k_work work;
+	uint8_t touch_buf[TMA525B_TOUCH_DATA_LEN];
```
faxe1008, line 86: Why not just have `struct tma525b_touch_data` instead of a raw buffer?

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+
+	/* Process current touch points */
+	for (i = 0; i < touch_count; i++) {
+		touch_id = TMA525B_TOUCH_POINT_GET_ID(touch_data->touch[i]);
+		touch_event = TMA525B_TOUCH_POINT_GET_EVENT(touch_data->touch[i]);
+		x = TMA525B_TOUCH_POINT_GET_X(touch_data->touch[i]);
+		y = TMA525B_TOUCH_POINT_GET_Y(touch_data->touch[i]);
+
+		/* Update coordinates only if there is valid touch event */
+		if (touch_event != kTouch_Reserved) {
```
faxe1008: Invert logic to save one indent level.

```
if (touch_event == kTouch_Reserved)
   continue

File: drivers/input/input_tma525b.c [hunk 10]
```diff
@@ -0,0 +1,458 @@
...
+		if (ret < 0) {
+			LOG_ERR("Failed to enable power: %d", ret);
+			return ret;
+		}
+		k_sleep(K_MSEC(10));
+	}
+
+	/* Reset the controller */
+	if (config->rst_gpio.port != NULL) {
+		gpio_pin_set_dt(&config->rst_gpio, 0);
```
faxe1008: I think the `rst` line should be active low and the logic inverted, so that rst line assertion means `1`.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,458 @@
...
+
+	/* Enter application mode - check if ready */
+	while (!app_mode && retry < 0xFF) {
+		ret = i2c_write_read_dt(&config->bus, &reg_addr, TMA525B_SUBADDR_SIZE, read_buf,
+					sizeof(read_buf));
+		if (ret == 0) {
+			/* Check for application mode signature */
+			if ((read_buf[0] == 0x02U && read_buf[1] == 0x00U) ||
+			    read_buf[1] == 0xFFU) {
+				app_mode = true;
```
faxe1008: Drop `app_mode`, return early with 0 here.
This way you can still have your error print further down without this temporary var.

faxe1008 (changes requested):

drivers/input/input_tma525b.c
KATE-WANG-NXP: The first 2 bytes of each data packet represent the length information of the touch report. Each time driver reads first 2 data to determine the required length, and then read the data again according to this length information(max length is 256 including the 2 byte of length). If the data read count is smaller than this length, the touch data packet will not be updated. tma525b_touch_data is only 47 bytes long not enough to hold the possible max length report.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+					sizeof(read_buf));
+		if (ret == 0) {
+			/* Check for application mode signature */
+			if ((read_buf[0] == 0x02U && read_buf[1] == 0x00U) ||
+			    read_buf[1] == 0xFFU) {
+				LOG_INF("TMA525B entered application mode");
+				break;
+			}
+		}
+		k_sleep(K_MSEC(300));
```
fabiobaltieri: that's a long wait, 300msec * 255, does it need to wait for so long? I'd imagine if it does not respond after 300ms it's just not there anymore

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+		if (!gpio_is_ready_dt(&config->rst_gpio)) {
+			LOG_ERR("Reset GPIO controller not ready");
+			return -ENODEV;
+		}
+		ret = gpio_pin_configure_dt(&config->rst_gpio, GPIO_OUTPUT_INACTIVE);
+		if (ret < 0) {
+			LOG_ERR("Failed to configure reset GPIO: %d", ret);
+			return ret;
+		}
+	}
```
fabiobaltieri, line 319: how about moving these to `tma525b_chip_init` and starting as `GPIO_OUTPUT_ACTIVE`? you'd end up reinitializing on resume but I think that's alright, simplify the code a bit

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+	uint8_t i, j;
+	int x, y;
+	uint8_t touch_id;
+	enum touch_event event;
+	bool touch_found;
+	uint8_t touch_count;
+
+	/* Read first two bytes to get the data length */
+	ret = i2c_write_read_dt(&config->bus, &reg_addr, TMA525B_SUBADDR_SIZE, data->touch_buf,
+				TMA525B_TOUCH_DATA_LEN_BYTES);
```
fabiobaltieri: use `i2c_burst_read_dt`, drop `TMA525B_SUBADDR_SIZE`

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+	if (ret < 0) {
+		LOG_ERR("Failed to read data length: %d", ret);
+		return ret;
+	}
+
+	touch_data = (struct tma525b_touch_data *)data->touch_buf;
+	touch_data_len = ((uint16_t)touch_data->length_h << 8U) | touch_data->length_l;
```
fabiobaltieri: can you use a uint16_t for length and then convert with `sys_be16_to_cpu`?

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+	uint8_t timestamp_100us_h;
+	uint8_t num_touch; /* Number of touch points detected */
+	/* 2 MSB: report counter, 3 LSB: noise effects. Not used in current driver model. */
+	uint8_t report_noise;
+	struct tma525b_touch_point touch[TMA525B_MAX_TOUCHES];
+} __packed;
+
+/* Macros for extracting touch data. */
+#define TMA525B_TOUCH_POINT_GET_ID(tp)    ((tp).event_id & 0x1F)
+#define TMA525B_TOUCH_POINT_GET_EVENT(tp) (((tp).event_id & 0x60) >> 5U)
```
fabiobaltieri: don't reference the field struct in this, just pass tp.event_id as argument to the macro

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,453 @@
...
+#define TMA525B_TOUCH_POINT_GET_X(tp)                                                              \
+	(int)(uint16_t)((((uint16_t)(tp).xh & 0x0FU) << 8) | (uint16_t)(tp).xl)
+#define TMA525B_TOUCH_POINT_GET_Y(tp)                                                              \
+	(int)(uint16_t)((((uint16_t)(tp).yh & 0x0FU) << 8) | (uint16_t)(tp).yl)
```
fabiobaltieri: use  `sys_be16_to_cpu` for these

drivers/input/input_tma525b.c
KATE-WANG-NXP: actually I think maybe better to keep the tma525b_chip_init to handle only the init sequence, these pin configuration code only appear once after all

boards/shields/g1120b0mipi/boards/mimxrt595_evk_mimxrt595s_cm33.conf
```diff
@@ -1,2 +1,3 @@
 # Enable DMA, so that DSI MCUX will use SMARTDMA
 CONFIG_DMA=y
+CONFIG_LV_Z_POINTER_FROM_CHOSEN_TOUCH=y
```
faxe1008: This is on by default, so can be removed

drivers/input/input_tma525b.c
KATE-WANG-NXP: I added a kconfig item to configure the wait times, and set the default value to 10. During test I find this is an acceptable value. Also about the delay time, the sample code provided by vendor is 300ms, but the app note says 120ms, so I changed to 120ms and it worked so far.

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,438 @@
...
+
+	/* Only when report ID is 0x01 the touch data is valid */
+	if (touch_data->report_id != TMA525B_REPORT_ID_TOUCH) {
+		LOG_DBG("Invalid report ID: 0x%02x", touch_data->report_id);
+		return -EINVAL;
+	}
+
+	/* Get number of touches */
+	touch_count = touch_data->num_touch;
+	if (touch_count > CONFIG_INPUT_TMA525B_MAX_TOUCH_POINTS) {
```
faxe1008: Maybe use MIN

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,438 @@
...
+
+	/* Reset the controller */
+	if (config->rst_gpio.port != NULL) {
+		gpio_pin_set_dt(&config->rst_gpio, 1);
+		k_sleep(K_MSEC(5));
+		gpio_pin_set_dt(&config->rst_gpio, 0);
+	}
+
+	/* Wait 300ms for boot */
+	k_sleep(K_MSEC(120));
```
faxe1008: Comment does not match

faxe1008 (dismissed):

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,436 @@
...
+
+/* TMA525B touch point structure */
+struct tma525b_touch_point {
+	uint16_t touch_type : 8U; /* 0 for standard finger/glove, 1 for proximity. Not used*/
+	uint16_t event_id : 8U;   /* Bit 0-4: touch ID, bit 5-6: touch event */
+	uint16_t x;
+	uint16_t y;
+	uint16_t pressure : 8U;       /* Touch intensity. Not used in current driver model. */
+	uint16_t major_axis_len_mm : 8U; /* Major axis length. Not used in current driver model. */
+	uint16_t minor_axis_len_mm : 8U; /* Minor axis length. Not used in current driver model. */
```
fabiobaltieri: hey I'm not entirely sure I follow, what's the bit packing doing here? can't you just have uint8_t and some dummy entries? may be easier to follow, not a fan of bit packed struct, they tend to do unexpected things

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,436 @@
...
+} __packed;
+
+/* TMA525B touch data structure */
+struct tma525b_touch_data {
+	uint16_t length; /* Packet length. */
+	uint8_t report_id;
+	/* Timestamp in 100us units low byte. Not used in current driver model. */
+	uint8_t timestamp_100us_l;
+	/* Timestamp in 100us units high byte. Not used in current driver model. */
+	uint8_t timestamp_100us_h;
```
fabiobaltieri: use a single `uint16_t`? it's unused anyway

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,436 @@
...
+	uint8_t num_touch; /* Number of touch points detected */
+	/* 2 MSB: report counter, 3 LSB: noise effects. Not used in current driver model. */
+	uint8_t report_noise;
+	struct tma525b_touch_point touch[TMA525B_MAX_TOUCHES];
+} __packed;
+
+/* Macros for extracting touch data. */
+#define TMA525B_TOUCH_POINT_GET_ID(event_id)    (event_id & 0x1F)
+#define TMA525B_TOUCH_POINT_GET_EVENT(event_id) (((event_id) & 0x60) >> 5U)
```
fabiobaltieri: drop the parenthesis from around `event_id`? you don't have them in the line above anyway

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,436 @@
...
+
+		/* Store current touch for tracking */
+		data->prev_touches[i].id = touch_id;
+		data->prev_touches[i].x = x;
+		data->prev_touches[i].y = y;
+	}
+
+	/* Handle release events for touches that disappeared */
+	for (i = 0; i < data->prev_touch_count; i++) {
+		touch_found = false;
```
fabiobaltieri: set a local variable for `data->prev_touches[i]`, should make things a bit more compact

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	/* Reset the controller */
+	if (config->rst_gpio.port != NULL) {
+		gpio_pin_set_dt(&config->rst_gpio, 1);
+		k_sleep(K_MSEC(5));
+		gpio_pin_set_dt(&config->rst_gpio, 0);
+	}
+
+	/* Wait 120ms for boot */
```
fabiobaltieri: sorry I know you changed this already, just drop the time constant, leave "Wait for the device to boot", just comment what we can't infer from the code.

even better would be to add a define for 120, call it like `BOOT_DELAY_MS` and then you don't need the comment at all as the code would be clear enough by itself, yeah do that please :-)

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+if SHIELD_ZC143AC72MIPI
+
+if LVGL
+# Enable input subsystem for TMA525b driver
+config INPUT
+	default y
+
+if MIPI_DSI_MCUX_2L
```
JarmouniA: board-specific, should be moved to board defconfig.

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,39 @@
...
+# Enable color swap in driver
+
+config MIPI_DSI_MCUX_2L_SWAP16
+	default y
+
+endif # MIPI_DSI_MCUX_2L
+
+# Swap 16 bit color setting for LVGL, to send high byte first
+configdefault LV_COLOR_16_SWAP
+	default y if !MIPI_DSI_MCUX_2L_SWAP16
```
JarmouniA: ditto

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,39 @@
...
+
+config LV_Z_VDB_SIZE
+	default 16
```
JarmouniA, line 13: ditto

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,39 @@
+# Copyright 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+if SHIELD_ZC143AC72MIPI
+
+if LVGL
```
JarmouniA:
```suggestion
```

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,39 @@
...
+	default y
+
+if MIPI_DSI_MCUX_2L
+# Enable color swap in driver
+
+config MIPI_DSI_MCUX_2L_SWAP16
+	default y
+
+endif # MIPI_DSI_MCUX_2L
+
```
JarmouniA:
```suggestion

if LVGL

```

dts/bindings/input/parade,tma525b.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2025 NXP
+# SPDX-License-Identifier: Apache-2.0
+
+description: TMA525b capacitive touch panels
```
JarmouniA:
```suggestion
description: TMA525b capacitive touch controller
```

JarmouniA (dismissed):

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+#include <zephyr/pm/device.h>
+#include <zephyr/pm/device_runtime.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(tma525b, CONFIG_INPUT_LOG_LEVEL);
+
+#define TMA525B_BOOT_DELAY_MS 120U
+
+/* TMA525B I2C address */
+#define TMA525B_I2C_ADDRESS 0x24U
```
fabiobaltieri: unused, drop

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+};
+
+INPUT_TOUCH_STRUCT_CHECK(struct tma525b_config);
+
+static int tma525b_process(const struct device *dev)
+{
+	const struct tma525b_config *config = dev->config;
+	struct tma525b_data *data = dev->data;
+	struct tma525b_touch_data *touch_data;
+	uint8_t reg_addr = TMA525B_TOUCH_DATA_SUBADDR;
```
fabiobaltieri: you only use this variable for this constant, just use the constant directly so one does not have to figure out what the variable points to to understand what i2c reg is been read

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+	int ret;
+	uint8_t i, j;
+	int x, y;
```
fabiobaltieri: declare these two in the cycle where they are used, just trying to narrow the scope, the automatic variable block here is huge, alternatively you could split the function but moving this would help

```c
	for (i = 0; i < touch_count; i++) {
                int x, y;
...
```

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+INPUT_TOUCH_STRUCT_CHECK(struct tma525b_config);
+
+static int tma525b_process(const struct device *dev)
+{
+	const struct tma525b_config *config = dev->config;
+	struct tma525b_data *data = dev->data;
+	struct tma525b_touch_data *touch_data;
+	uint8_t reg_addr = TMA525B_TOUCH_DATA_SUBADDR;
+	int ret;
+	uint8_t i, j;
```
fabiobaltieri: declare this in the loop

```c
	for (uint8_t i = 0; i < touch_count; i++) {
```

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+	int x, y;
+	uint8_t touch_id;
+	enum touch_event event;
+	bool touch_found;
```
fabiobaltieri: same, declare it in the loop where it's used

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+	struct tma525b_data *data = CONTAINER_OF(timer, struct tma525b_data, timer);
+
+	k_work_submit(&data->work);
+}
+#endif
+
+static int tma525b_chip_init(const struct device *dev)
+{
+	const struct tma525b_config *config = dev->config;
+	uint8_t reg_addr = TMA525B_TOUCH_DATA_SUBADDR;
```
fabiobaltieri: same, use the value directly

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+	if (ret < 0) {
+		LOG_ERR("Failed to add GPIO callback: %d", ret);
+		return ret;
+	}
+	LOG_INF("TMA525B using interrupt mode");
+#else
+	k_timer_init(&data->timer, tma525b_timer_handler, NULL);
+	k_timer_start(&data->timer, K_MSEC(CONFIG_INPUT_TMA525B_PERIOD_MS),
+		      K_MSEC(CONFIG_INPUT_TMA525B_PERIOD_MS));
+	LOG_INF("TMA525B using polling mode");
```
fabiobaltieri: drop or change to `LOG_DBG`, normally driver code is silent if there are no warnings or errors, same above

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+#endif
+
+	ret = pm_device_runtime_enable(dev);
+	if (ret < 0 && ret != -ENOTSUP) {
+		LOG_ERR("Failed to enable runtime power management");
+		return ret;
+	}
+
+	LOG_INF("TMA525B initialized successfully");
```
fabiobaltieri: drop

drivers/input/input_tma525b.c
```diff
@@ -0,0 +1,434 @@
...
+#else
+	k_timer_init(&data->timer, tma525b_timer_handler, NULL);
+	k_timer_start(&data->timer, K_MSEC(CONFIG_INPUT_TMA525B_PERIOD_MS),
+		      K_MSEC(CONFIG_INPUT_TMA525B_PERIOD_MS));
+	LOG_INF("TMA525B using polling mode");
+#endif
+
+	ret = pm_device_runtime_enable(dev);
+	if (ret < 0 && ret != -ENOTSUP) {
+		LOG_ERR("Failed to enable runtime power management");
```
fabiobaltieri: log the error code

boards/shields/zc143ac72mipi/Kconfig.defconfig
fabiobaltieri: has this been addressed? not sure what the original comment was

JarmouniA: > not sure what the original comment was

https://github.com/zephyrproject-rtos/zephyr/pull/101254#discussion_r2649382008

drivers/input/input_tma525b.c
JarmouniA: Also one var declaration/def per line.
(non-blocking)

fabiobaltieri: Do we have a guideline that says that? I don't mind both in the same line if they are correlated.

JarmouniA: Yes it's in the Linux kernel coding guidelines.

fabiobaltieri: Do you have a pointer? Been working a fair bit on a Linux driver these days have plenty of `int row, col`, reviewers were happy with it (it was probably the only thing they were happy with).

fabiobaltieri: it's VERY common in `linux/drivers/input` at least

JarmouniA: I failed to find the reference, maybe it doesn't exist, but in my opinion still a good recommendation, less chance of git conflicts, more readable, the downside is the extra lines.

fabiobaltieri: The only reference I found was about documentation, like in struct where you may want to document each variable with a comment before, does not apply here, I'd suggest keeping it in this case since the two have the same meaning, keep it a bit more compact. I also normally lay on the side of one variable per line but for cases like this, `x, y` or `row, col`, I prefer them on the same line, funny I had to check my own code and realize that that what I've been doing too, semi consciously I suppose.

Ultimately since it's driver code I'd leave it up to the author but my preference would be as is, just in the loop.

uLipe (dismissed):

boards/shields/zc143ac72mipi/Kconfig.defconfig
```diff
@@ -0,0 +1,28 @@
...
+if SHIELD_ZC143AC72MIPI
+
+# Enable input subsystem for TMA525b driver
+config INPUT
+	default y
+
+if LVGL
+
+config LV_Z_VDB_SIZE
```
JarmouniA, line 12: board-specific, should be moved to board defconfig.

KATE-WANG-NXP: I think it is panel-specific? Related to the display size
image

faxe1008 (approved):

boards/shields/zc143ac72mipi/Kconfig.defconfig
JarmouniA: Also depends on how much memory the board has.

uLipe (approved):

fabiobaltieri (approved):

PR #101255, opened 2025-12-18
linker: section_tags: remove deprecated STM32 CCM section tags

mathieuchopstm:
The ST-specific CCM mechanism was replaced by DTCM in Zephyr release v4.4. Remove backwards compatibility macros as part of this release, as planned and documented in the v4.4 migration guide and on the macros themselves - c.f. #100590

### DIFF
--- a/doc/releases/migration-guide-4.5.rst
+++ b/doc/releases/migration-guide-4.5.rst
@@ -54,6 +54,12 @@ Clock Control
   ``loop-div = clock-mult * 2`` and ``post-div = clock-div``.
 
 
+STM32
+=====
+
+* The attribute macros ``__ccm_data_section``, ``__ccm_data_section`` and ``__ccm_data_section``
+  retained in Zephyr 4.4 for backwards compatibility have been removed. (:github:`101255`)
+
 .. zephyr-keep-sorted-stop
 
 Bluetooth
--- a/include/zephyr/linker/section_tags.h
+++ b/include/zephyr/linker/section_tags.h
@@ -43,14 +43,6 @@
 #define __imx_boot_dcd_section Z_GENERIC_SECTION(_IMX_BOOT_DCD_SECTION_NAME)
 #define __imx_boot_container_section Z_GENERIC_SECTION(_IMX_BOOT_CONTAINER_SECTION_NAME)
 #define __stm32_backup_sram_section Z_GENERIC_SECTION(_STM32_BACKUP_SRAM_SECTION_NAME)
-
-/*
- * Deprecated aliases, provided for backwards compatibility.
- * These aliases will be removed in Zephyr v4.5.
- */
-#define __ccm_data_section __dtcm_data_section __DEPRECATED_MACRO
-#define __ccm_bss_section __dtcm_bss_section __DEPRECATED_MACRO
-#define __ccm_noinit_section __dtcm_noinit_section __DEPRECATED_MACRO
 #endif /* CONFIG_ARM */
 
 #if defined(CONFIG_NOCACHE_MEMORY)


mathieuchopstm:
@MaureenHelm @stephanosio could you take a look please?

MaureenHelm (commented):
If these macros were deprecated in 4.4, then it's too early to remove them per the deprecation policy:

> Deprecation Time (stable APIs): 2 Releases The API needs to be marked as deprecated in at least two full releases. For example, if an API was first deprecated in release 4.0, it will be ready to be removed in 4.2 at the earliest.

https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html#deprecated

mathieuchopstm:
> If these macros were deprecated in 4.4, then it's too early to remove them per the deprecation policy:
> 
> > Deprecation Time (stable APIs): 2 Releases The API needs to be marked as deprecated in at least two full releases. For example, if an API was first deprecated in release 4.0, it will be ready to be removed in 4.2 at the earliest.
> 
> https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html#deprecated

As far as I can tell, these macros are not part of any "proper" API so I don't think the API Lifecycle policy applies here.

PR #100590 was *explicitly* non-compliant with the API Lifecycle policy: it removed CCM section symbols without backwards-compatible replacements (and also breaks images if you haven't migrated, even though CMake could have made this a clean build error - past me wasn't very thoughtful on this...)

The backwards compatibility aliases provided in #100590 were intended as merely "vendor-specific courtesy" and **not** as compliance with the API Lifecycle policy's "mandatory backwards-compatibility" clause. It was also *explicitly* documented in #100590 that the macros would be removed in v4.5 - and no one objected to this clause in the PR :slightly_smiling_face:.


For all these reasons, I think deprecation in v4.5 is perfectly acceptable. If you strongly disagree with my rationale, then we can postpone to v4.6 though...

mathieuchopstm:
@MaureenHelm / @stephanosio could you please take a look?

MaureenHelm:
> > If these macros were deprecated in 4.4, then it's too early to remove them per the deprecation policy:
> > > Deprecation Time (stable APIs): 2 Releases The API needs to be marked as deprecated in at least two full releases. For example, if an API was first deprecated in release 4.0, it will be ready to be removed in 4.2 at the earliest.
> > 
> > 
> > [docs.zephyrproject.org/latest/develop/api/api_lifecycle.html#deprecated](https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html#deprecated)
> 
> As far as I can tell, these macros are not part of any "proper" API so I don't think the API Lifecycle policy applies here.
> 
> PR #100590 was _explicitly_ non-compliant with the API Lifecycle policy: it removed CCM section symbols without backwards-compatible replacements (and also breaks images if you haven't migrated, even though CMake could have made this a clean build error - past me wasn't very thoughtful on this...)
> 
> The backwards compatibility aliases provided in #100590 were intended as merely "vendor-specific courtesy" and **not** as compliance with the API Lifecycle policy's "mandatory backwards-compatibility" clause. It was also _explicitly_ documented in #100590 that the macros would be removed in v4.5 - and no one objected to this clause in the PR 🙂.
> 
> For all these reasons, I think deprecation in v4.5 is perfectly acceptable. If you strongly disagree with my rationale, then we can postpone to v4.6 though...

I think it dilutes the meaning of "deprecated" to have "vendor-specific courtesy" deprecations that don't follow the deprecation policy. The clause about removing in v4.5 shouldn't have been accepted in #100590.

mathieuchopstm:
> [...]
> 
> I think it dilutes the meaning of "deprecated" to have "vendor-specific courtesy" deprecations that don't follow the deprecation policy. The clause about removing in v4.5 shouldn't have been accepted in #100590.

I don't disagree in principle but the policy today is explicitly scoped to *Public* APIs, which these macros were not part of as far as I can tell; maybe a proper policy for the lifecycle of non-Public APIs is necessary (which would be less constraining than the Public APIs policy but still impose some restrictions).

In this particular instance, I don't mind postponing to v4.6 so I'll change the milestone.

PR #35001, opened 2021-05-08
Bluetooth: Mesh: Check the CID field before opcode compare

LingaoM:
Bluetooth Mesh Vendor model hava company id field.

Accordin MeshPRFV1.0.1 3.7.3.1 Operation codes.

The 3-octet opcodes are used for manufacturer-specific opcodes.
The company identifiers are 16-bit values defined by the
Bluetooth SIG and are coded into the second and third octets of
the 3-octet opcodes.

Therefore, we can speed up the search process by checking whether
CID fields match, rather than comparing opcodes one by one.

Signed-off-by: Lingao Meng <menglingao@xiaomi.com>

PR #101262, opened 2025-12-18
Bluetooth: Mesh: Let vendor models receive messages with SIG opcodes

ludvigsj:
This fixes a limitation where the stack assumed that vendor models could
not specify handlers for SIG-defined (1- or 2-byte) opcodes. This
assumption does not exist in the specification. In fact, the
specification expects this very use-case to be possible, in MeshPRT 1.1,
section 3.8.3:

> To exchange messages with a Bluetooth SIG adopted model, a Vendor Model
> shall use the Access message defined for the Bluetooth SIG model.

In addition, it includes a commit to add a BSIM test to check this
behvior, a commit to fix a compile warning when not forcing CID to match
between model and opcode handlers, and a commit to deprecate the forced CID
match (and related optimization) option and disable it by default. See individual commit messages
for details.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -445,6 +445,12 @@ Bluetooth Audio
   :c:func:`bt_bap_broadcast_assistant_read_recv_state` to read the existing receive states, if any,
   prior to performing any operations. (:github:`91587`)
 
+Bluetooth Mesh
+==============
+
+* :kconfig:option:`CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE` has been deprecated. Enabling it no
+  longer has any effect on message handling performance.
+
 Networking
 **********
 
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -785,11 +785,12 @@ config BT_MESH_ACCESS_LAYER_MSG
 	  might wear out persistent memory.
 
 config BT_MESH_MODEL_VND_MSG_CID_FORCE
-	bool "Force vendor model to use the corresponding CID field message"
-	default y
+	bool "Force vendor model to use opcodes with matching CID [DEPRECATED]"
+	default n
+	select DEPRECATED
 	help
-	  This option forces vendor model to use messages for the
-	  corresponding CID field.
+	  This option forces vendor models to only use opcodes with a CID field
+	  matching the models CID. It is deprecated and should no longer be used.
 
 config BT_MESH_MODEL_EXTENSIONS
 	bool "Support for Model extensions"
--- a/subsys/bluetooth/mesh/access.c
+++ b/subsys/bluetooth/mesh/access.c
@@ -1413,12 +1413,6 @@ static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
 
 		const struct bt_mesh_model_op *op;
 
-		if (IS_ENABLED(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE) &&
-		     cid != UINT32_MAX &&
-		     cid != models[i].vnd.company) {
-			continue;
-		}
-
 		*model = &models[i];
 
 		for (op = (*model)->op; op->func; op++) {
--- a/subsys/bluetooth/mesh/access.c
+++ b/subsys/bluetooth/mesh/access.c
@@ -998,7 +998,7 @@ static int bt_mesh_vnd_mod_msg_cid_check(const struct bt_mesh_model *mod)
 	for (op = mod->op; op->func; op++) {
 		cid = (uint16_t)(op->opcode & 0xffff);
 
-		if (cid == mod->vnd.company) {
+		if (BT_MESH_MODEL_OP_LEN(op->opcode) == 3 && cid == mod->vnd.company) {
 			continue;
 		}
 
@@ -1386,31 +1386,11 @@ static bool model_has_dst(const struct bt_mesh_model *mod, uint16_t dst, const u
 	return mod->rt->elem_idx == 0;
 }
 
-static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
-					      uint32_t opcode, const struct bt_mesh_model **model)
+static const struct bt_mesh_model_op *find_op_in_list(const struct bt_mesh_model *models,
+						      uint8_t model_count, uint32_t opcode,
+						      const struct bt_mesh_model **model)
 {
-	uint8_t i;
-	uint8_t count;
-	/* This value shall not be used in shipping end products. */
-	uint32_t cid = UINT32_MAX;
-	const struct bt_mesh_model *models;
-
-	/* SIG models cannot contain 3-byte (vendor) OpCodes, and
-	 * vendor models cannot contain SIG (1- or 2-byte) OpCodes, so
-	 * we only need to do the lookup in one of the model lists.
-	 */
-	if (BT_MESH_MODEL_OP_LEN(opcode) < 3) {
-		models = elem->models;
-		count = elem->model_count;
-	} else {
-		models = elem->vnd_models;
-		count = elem->vnd_model_count;
-
-		cid = (uint16_t)(opcode & 0xffff);
-	}
-
-	for (i = 0U; i < count; i++) {
-
+	for (uint8_t i = 0U; i < model_count; i++) {
 		const struct bt_mesh_model_op *op;
 
 		*model = &models[i];
@@ -1424,6 +1404,21 @@ static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
 
 	*model = NULL;
 	return NULL;
+}
+
+static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
+					      uint32_t opcode, const struct bt_mesh_model **model)
+{
+	const struct bt_mesh_model_op *op;
+
+	op = find_op_in_list(elem->vnd_models, elem->vnd_model_count, opcode, model);
+
+	if (op != NULL || BT_MESH_MODEL_OP_LEN(opcode) == 3) {
+		/* Return early if LEN == 3, since SIG models cannot contain 3-byte opcodes. */
+		return op;
+	}
+
+	return find_op_in_list(elem->models, elem->model_count, opcode, model);
 }
 
 static int get_opcode(struct net_buf_simple *buf, uint32_t *opcode)
--- a/subsys/bluetooth/mesh/access.c
+++ b/subsys/bluetooth/mesh/access.c
@@ -990,7 +990,6 @@ const struct bt_mesh_model *bt_mesh_model_get(bool vnd, uint8_t elem_idx, uint8_
 	}
 }
 
-#if defined(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE)
 static int bt_mesh_vnd_mod_msg_cid_check(const struct bt_mesh_model *mod)
 {
 	uint16_t cid;
@@ -1012,7 +1011,6 @@ static int bt_mesh_vnd_mod_msg_cid_check(const struct bt_mesh_model *mod)
 
 	return 0;
 }
-#endif
 
 static void mod_init(const struct bt_mesh_model *mod, const struct bt_mesh_elem *elem,
 		     bool vnd, bool primary, void *user_data)
--- a/tests/bsim/bluetooth/mesh/src/test_access.c
+++ b/tests/bsim/bluetooth/mesh/src/test_access.c
@@ -29,6 +29,7 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME, LOG_LEVEL_INF);
 #define TEST_MESSAGE_OP_3  BT_MESH_MODEL_OP_1(0x13)
 #define TEST_MESSAGE_OP_4  BT_MESH_MODEL_OP_1(0x14)
 #define TEST_MESSAGE_OP_5  BT_MESH_MODEL_OP_1(0x15)
+#define TEST_MESSAGE_OP_E  BT_MESH_MODEL_OP_1(0x1E)
 #define TEST_MESSAGE_OP_F  BT_MESH_MODEL_OP_1(0x1F)
 
 #define PUB_PERIOD_COUNT 3
@@ -40,10 +41,14 @@ static int model2_init(const struct bt_mesh_model *model);
 static int model3_init(const struct bt_mesh_model *model);
 static int model4_init(const struct bt_mesh_model *model);
 static int model5_init(const struct bt_mesh_model *model);
+static int vnd_model_init(const struct bt_mesh_model *model);
 static int test_msg_handler(const struct bt_mesh_model *model,
 			struct bt_mesh_msg_ctx *ctx,
 			struct net_buf_simple *buf);
 static int test_msg_ne_handler(const struct bt_mesh_model *model,
+			struct bt_mesh_msg_ctx *ctx,
+			struct net_buf_simple *buf);
+static int test_vnd_msg_handler(const struct bt_mesh_model *model,
 			struct bt_mesh_msg_ctx *ctx,
 			struct net_buf_simple *buf);
 
@@ -97,7 +102,8 @@ static const struct {
 	},
 };
 
-static struct k_sem publish_sem;
+static K_SEM_DEFINE(publish_sem, 0, 1);
+static K_SEM_DEFINE(vnd_rx_sem, 0, 1);
 static bool publish_allow;
 
 static int model1_update(const struct bt_mesh_model *model)
@@ -155,6 +161,10 @@ static const struct bt_mesh_model_cb test_model5_cb = {
 	.init = model5_init,
 };
 
+static const struct bt_mesh_model_cb vnd_model_cb = {
+	.init = vnd_model_init,
+};
+
 static const struct bt_mesh_model_op model_op1[] = {
 	{ TEST_MESSAGE_OP_1, 0, test_msg_handler },
 	{ TEST_MESSAGE_OP_F, 0, test_msgf_handler },
@@ -178,6 +188,11 @@ static const struct bt_mesh_model_op model_op4[] = {
 
 static const struct bt_mesh_model_op model_op5[] = {
 	{ TEST_MESSAGE_OP_5, 0, test_msg_handler },
+	BT_MESH_MODEL_OP_END
+};
+
+static const struct bt_mesh_model_op vnd_model_op[] = {
+	{ TEST_MESSAGE_OP_E, 0, test_vnd_msg_handler },
 	BT_MESH_MODEL_OP_END
 };
 
@@ -228,11 +243,16 @@ static const struct bt_mesh_model models_ne[] = {
 	BT_MESH_MODEL_CB(TEST_MODEL_ID_5, model_ne_op5, NULL, NULL, &test_model5_cb),
 };
 
-static const struct bt_mesh_model vnd_models[] = {};
+static const struct bt_mesh_model vnd_models[] = {
+	BT_MESH_MODEL_VND_CB(TEST_VND_COMPANY_ID, TEST_VND_MOD_ID, vnd_model_op, NULL, NULL,
+			     &vnd_model_cb)
+};
+
+static const struct bt_mesh_model vnd_models_ne[] = {};
 
 static const struct bt_mesh_elem elems[] = {
 	BT_MESH_ELEM(0, models, vnd_models),
-	BT_MESH_ELEM(1, models_ne, vnd_models),
+	BT_MESH_ELEM(1, models_ne, vnd_models_ne),
 };
 
 const struct bt_mesh_comp local_comp = {
@@ -287,6 +307,11 @@ static int model5_init(const struct bt_mesh_model *model)
 	return 0;
 }
 
+static int vnd_model_init(const struct bt_mesh_model *model)
+{
+	return 0;
+}
+
 static int test_msg_handler(const struct bt_mesh_model *model,
 			struct bt_mesh_msg_ctx *ctx,
 			struct net_buf_simple *buf)
@@ -302,6 +327,16 @@ static int test_msg_ne_handler(const struct bt_mesh_model *model,
 			struct net_buf_simple *buf)
 {
 	FAIL("Model %#4x on neighbor element received msg", model->id);
+
+	return 0;
+}
+
+static int test_vnd_msg_handler(const struct bt_mesh_model *model,
+			struct bt_mesh_msg_ctx *ctx,
+			struct net_buf_simple *buf)
+{
+	LOG_DBG("vnd msg rx");
+	k_sem_give(&vnd_rx_sem);
 
 	return 0;
 }
@@ -344,6 +379,13 @@ static void common_configure(uint16_t addr)
 					model_ids[i], err, status);
 			return;
 		}
+	}
+
+	err = bt_mesh_cfg_cli_mod_app_bind_vnd(0, addr, addr, 0, TEST_VND_MOD_ID,
+					       TEST_VND_COMPANY_ID, &status);
+	if (err || status) {
+		FAIL("Vendor model bind failed (err %d, status %u)", err, status);
+		return;
 	}
 
 	err = bt_mesh_cfg_cli_net_transmit_set(0, addr, BT_MESH_TRANSMIT(2, 20), &status);
@@ -689,8 +731,6 @@ static void tx_period(bool delayable)
 	provision(UNICAST_ADDR1);
 	common_configure(UNICAST_ADDR1);
 
-	k_sem_init(&publish_sem, 0, 1);
-
 	model->pub->delayable = delayable;
 
 	for (size_t i = 0; i < ARRAY_SIZE(test_period); i++) {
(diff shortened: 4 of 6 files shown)


subsys/bluetooth/mesh/Kconfig
```diff
@@ -785,11 +785,12 @@ config BT_MESH_ACCESS_LAYER_MSG
...
-	bool "Force vendor model to use the corresponding CID field message"
-	default y
+	bool "Force vendor model to use opcodes with matching CID"
+	default n
 	help
-	  This option forces vendor model to use messages for the
-	  corresponding CID field.
+	  This option forces vendor models to only use opcodes with a CID field
+	  matching the models CID, making it possible for the stack to slightly
+	  optimize message handling in applications that use vendor models.
```
omkar3141:
```suggestion
	  optimize message handling in applications that use vendor models in a way that these vendor models deal with opcodes only specific to them and no other opcodes.
```

subsys/bluetooth/mesh/access.c
```diff
@@ -1388,36 +1387,16 @@ static bool model_has_dst(const struct bt_mesh_model *mod, uint16_t dst, const u
 	return mod->rt->elem_idx == 0;
 }
 
-static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
-					      uint32_t opcode, const struct bt_mesh_model **model)
+static const struct bt_mesh_model_op *find_op_in_list(const struct bt_mesh_model *models,
+						      uint8_t count, uint32_t opcode,
```
omkar3141: minor
```suggestion
						      uint8_t mdl_count, uint32_t opcode,
```
?

subsys/bluetooth/mesh/access.c
```diff
@@ -1434,6 +1413,31 @@ static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
...
+	if (BT_MESH_MODEL_OP_LEN(opcode) == 3) {
+		return find_op_in_list(elem->vnd_models, elem->vnd_model_count, opcode, model);
+	}
+
+	op = find_op_in_list(elem->models, elem->model_count, opcode, model);
+
+	/* Return (NULL) already if we force vendor model opcode CID match. SIG opcodes per
+	 * definition cannot match since they don't have a CID.
+	 */
+	if (op != NULL || IS_ENABLED(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE)) {
```
omkar3141: Looks like this "|| IS_ENABLED(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE)" is not needed here. Because, if opcode is 3 byte long, what to do with that opcode in the context of this KCONFIG, this decision will be taken in "find_op_in_list()".

ludvigsj: `find_op_in_list` will loop through every vendor model before making this decision. Since `CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE` is supposed to optimize I thought it would be better to just return early here.

ludvigsj: Of course, I might be doing the work of the compiler here ...

ludvigsj: Oh, and "if opcode is 3 byte long", no, this line only runs when opcode is 1-2 bytes long. The `if` in the beginning of the function is the only code that runs if it's 3 bytes long.

ludvigsj: If we don't keep this, then IMO we can simplify the entire thing further and just leave this:

```c
static const struct bt_mesh_model_op *find_op(const struct bt_mesh_elem *elem,
					      uint32_t opcode, const struct bt_mesh_model **model)
{
	const struct bt_mesh_model_op *op;

	op = find_op_in_list(elem->models, elem->model_count, opcode, model);
	if (op != NULL) {
		return op;
	}

	return find_op_in_list(elem->vnd_models, elem->vnd_model_count, opcode, model);
}
```

omkar3141: We should just remove this KConfig option in future. This option is not really useful and it just adds to code when this option is enabled with very little gain.

omkar3141 (dismissed):

subsys/bluetooth/mesh/Kconfig
```diff
@@ -785,11 +785,14 @@ config BT_MESH_ACCESS_LAYER_MSG
 	  might wear out persistent memory.
 
 config BT_MESH_MODEL_VND_MSG_CID_FORCE
-	bool "Force vendor model to use the corresponding CID field message"
-	default y
+	bool "Force vendor model to use opcodes with matching CID"
+	default n
```
PavelVPV, line 789: This should be mentioned in the migration guide as it changes the default behavior.

But why should we change the default behavior? Why can't those who doesn't need the check, disable it?

subsys/bluetooth/mesh/Kconfig
```diff
@@ -785,11 +785,14 @@ config BT_MESH_ACCESS_LAYER_MSG
...
 	help
-	  This option forces vendor model to use messages for the
-	  corresponding CID field.
+	  This option forces vendor models to only use opcodes with a CID field
+	  matching the models CID, making it possible for the stack to slightly
+	  optimize message handling in applications that use vendor models, where
```
PavelVPV: What `slightly` means is not clear to anybody I think. This needs to be either removed or rephrased. What this options did previously:
- Didn't allow to initialize stack if any vendor model had an opcode handler which CID didn't match with model's CID
- Skipped searching for opcode if CID of message doesn't match with model's CID.

I believe it would be more useful for a user if we could explain this in the help message.

tests/bsim/bluetooth/mesh/tests_scripts/access/access_vnd_rcv_sig_msg.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Copyright 2025 Nordic Semiconductor
+# SPDX-License-Identifier: Apache-2.0
+
+source $(dirname "${BASH_SOURCE[0]}")/../../_mesh_test.sh
+
+RunTest mesh_access_vnd_rcv_sig_msg \
```
PavelVPV, line 17: Needs test scenario

subsys/bluetooth/mesh/access.c
```diff
@@ -1037,13 +1037,12 @@ static void mod_init(const struct bt_mesh_model *mod, const struct bt_mesh_elem
 	if (vnd) {
 		mod->rt->mod_idx = mod - elem->vnd_models;
 
-		if (IS_ENABLED(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE)) {
-			*err = bt_mesh_vnd_mod_msg_cid_check(mod);
-			if (*err) {
-				return;
-			}
+#if defined(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE)
```
PavelVPV: I'd prefer the other way around: `IS_ENABLED` is in `bt_mesh_vnd_mod_msg_cid_check` since there is nothing in `bt_mesh_vnd_mod_msg_cid_check` that prevents it to be working with `IS_ENABLED`.

subsys/bluetooth/mesh/access.c
```diff
@@ -1388,36 +1387,16 @@ static bool model_has_dst(const struct bt_mesh_model *mod, uint16_t dst, const u
...
-	for (i = 0U; i < count; i++) {
-
+	for (uint8_t i = 0U; i < model_count; i++) {
 		const struct bt_mesh_model_op *op;
 
 		if (IS_ENABLED(CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE) &&
-		     cid != UINT32_MAX &&
-		     cid != models[i].vnd.company) {
+		    BT_MESH_MODEL_OP_LEN(opcode) == 3 &&
+		    (uint16_t)(opcode & 0xffff) != models[i].vnd.company) {
```
PavelVPV: I don't like this tbh. Passing here sig models as `models` but accessing `company` field of vendor models. Even though it is currently safe because [company](https://github.com/zephyrproject-rtos/zephyr/blob/c3190d34aa912d39cd082db822f6ebb332806a2b/include/zephyr/bluetooth/mesh/access.h#L707) and [id](https://github.com/zephyrproject-rtos/zephyr/blob/c3190d34aa912d39cd082db822f6ebb332806a2b/include/zephyr/bluetooth/mesh/access.h#L716) are the same types at the same location, it may change and then this code causes undefined behavior. Neither logically I understand this code.

tests/bsim/bluetooth/mesh/src/test_access.c
```diff
@@ -98,6 +103,7 @@ static const struct {
 };
 
 static struct k_sem publish_sem;
+static struct k_sem vnd_rx_sem;
```
PavelVPV: I think both semaphores should be statically initialized, otherwise it will be undefined behavior if some test use wrong message triggering a sempahore which is not initialized.

subsys/bluetooth/mesh/access.c
```diff
@@ -999,7 +999,7 @@ static int bt_mesh_vnd_mod_msg_cid_check(const struct bt_mesh_model *mod)
 	for (op = mod->op; op->func; op++) {
 		cid = (uint16_t)(op->opcode & 0xffff);
 
-		if (cid == mod->vnd.company) {
+		if (BT_MESH_MODEL_OP_LEN(op->opcode) == 3 && cid == mod->vnd.company) {
```
PavelVPV: I don't think this is correct since the force option is to exactly ensure that a vendor model doesn't serve messages opcode with different Company ID.

Since the spec clearly allows a vendor model serve 1-/2-octets opcodes, this should only check matching Company ID for 3-octets opcode messages. Therefore, I think the check should be like this:
```
if (BT_MESH_MODEL_OP_LEN(op->opcode) == 3 && cid != mod->vnd.company) {
    LOG_ERR("...");
    return -EINVAL;
}
```

subsys/bluetooth/mesh/access.c
PavelVPV: I'd suggest to make it more human readable by putting `(uint16_t)(opcode & 0xffff)` into a variable `cid`.

PavelVPV (changes requested):

subsys/bluetooth/mesh/Kconfig
ludvigsj: Because

1. being able to use opcodes that don't match the CID of the model is a very normal behavior, and having this fail is already confusing users who don't know about this option. It is even required by the spec to be able to communicate with other models, see commit message.
2. The change in behavior is slight, I would be surprised if anyone was consciously relying on this optimization. Note, every application that compiles and runs properly will still do so after this change, the _only_ consequence will be a slight performance decrease for applications with multiple vendor models using at least two different CIDs, which is such a niche use case combined with a tiny consequence (or should we note in the changelog every time CPU performance of an API changes slightly?) that I decided that it didn't warrant a migration guide.
3. Since the corner case where this optimization can do anything at all (at least two different vendor models with at least two different CIDs in the network), and the benefit of it when it _does_ apply (saving a tiny handful of comparisons per opcode defined in vendor models on a node) is tiny in "normal" cases, I'd prefer to just remove the entire thing, but at the very least make it default to off

ludvigsj: Regarding 2), I'm not even sure if this is true; my extra `if`s and function calls might negate that already, meaning even without changing the deafult they get decreased performance.

ludvigsj: We discussed this in the Mesh team, and we really don't like this defaulting to y and confusing users who want to just follow what the spec says is allowed (and sometimes required!) and use other opcodes in their models.

subsys/bluetooth/mesh/Kconfig
ludvigsj: As far as I can tell, this option does the first thing only to be able to do the second thing. How is "Skipped searching for opcode if CID of message doesn't match with model's CID" more clear? I don't know how computationally complex "searching for opcode" is so the benefit could be anything even in this case. Or are you suggesting people are using this option as a safeguard for themselves to ensure they don't type the wrong CIDs?

subsys/bluetooth/mesh/access.c
ludvigsj: TBH, we don't even need that, do we? If the only call site is inside an `if(IS_ENABLED)` this should be trivially removed by the compiler already.

subsys/bluetooth/mesh/Kconfig
PavelVPV: > As far as I can tell, this option does the first thing only to be able to do the second thing. 

It could do both independently really. In fact, the check at initialization just adds extra code for nothing (if you don't have such vendor models).

> Or are you suggesting people are using this option as a safeguard for themselves to ensure they don't type the wrong CIDs?

I'm suggesting to rephrase the help message to a more meaningful text that is understandable. The added clarification doesn't help me to understand whether I need to use the Kconfig or not. In fact, it adds more code so I don't know what optimization it adds really.

ludvigsj: IMO, the only reason why anyone should enable this is because they a) want an optimization of message handling (this is the motivation) and b) they are not using "foreign" opcodes in any of their vendor models (this is a constraint)

ludvigsj: I agree, I'm not sure if it does optimize anything. We are probably going to deprecate and remove this entire thing if we find the time, it looks a bit useless and complicates the code.

subsys/bluetooth/mesh/access.c
jhedberg: Even a `BT_MESH_MODEL_OP_CID(op)` macro could be justified

subsys/bluetooth/mesh/Kconfig
PavelVPV: Your comments are correct with the assumption of the new change from this PR. This is not correct if the change is such that this Kconfig option only verifies 3-octet opcodes with CID (like before) leaving 1-/2-octet opcodes ignored, thus making the code spec compliant even if it is enabled. Right? I guess the initial intention of the option was exactly to have opcode.CID with model.CID match, disabling which you can build a hacked model.

subsys/bluetooth/mesh/access.c
ludvigsj: @jhedberg seeing as we are considering removing this entire check in the future and (AFAIK) the stack doesn't expose incoming opcodes to the application anywhere, I don't see the reason for this when there is all of one (potentially soon-to-be-removed) instance of this extraction in the stack. I did what Pavel suggested for now, which matches what this module did before my changes.

jhedberg: @ludvigsj that sounds fine to me

subsys/bluetooth/mesh/Kconfig
PavelVPV: > IMO, the only reason why anyone should enable this is because they a) want an optimization of message handling (this is the motivation) and b) they are not using "foreign" opcodes in any of their vendor models (this is a constraint)

Opposite, by default they get spec compliant stack which doesn't allow to assign opcodes to models which CIDs don't match. If one wants to make it non-compliant -> disable this option: https://github.com/zephyrproject-rtos/zephyr/pull/35001#discussion_r630757534

> I agree, I'm not sure if it does optimize anything. We are probably going to deprecate and remove this entire thing if we find the time, it looks a bit useless and complicates the code.

I guess if it was introduced, it optimizes something. I guess the problem is when you have more than 5-10 messages per model.

subsys/bluetooth/mesh/Kconfig
alxelax: This is not `hacked model`. This is `shall` statement that the existing implementation violates. Then people who implement the vendor model cannot receive SIG commands due to this violation.

PavelVPV: > This is not `hacked model`. This is `shall` statement that the existing implementation violates. Then people who implement the vendor model cannot receive SIG commands due to this violation.

You didn't get me. The hacked model I meant is a model which CID doesn't match with opcode's CID.

In my mind I had exactly the opposite: don't touch this option for 1-/2-octet opcodes for vendor models as it makes not sense for this option.

ludvigsj: To recap:

Per spec, SIG models only (so far, and I assume for the foreseeable future) only handle messages with 1 and 2 byte opcodes. Vendor models can handle any opcodes they want, this is not a "hacked" model, and vendor models are _required_ to do this if they want to communicate with SIG models (MeshPRT 1.1, section 3.8.3):

> To exchange messages with a Bluetooth SIG adopted model, a Vendor Model
shall use the Access message defined for the Bluetooth SIG model.

Before my changes
-----------------

**_VND_MSG_CID_FORCE=y**: Vendor models could only ever receive messages using 3-byte opcodes where the CID matches the model-CID. Specifying handlers for 3-byte opcodes with a CID different from the model-CID would cause an error at model_init time. Specifying handlers for 1- or 2-byte opcodes would just not work (meaning, you could do it, but would not receive any messages). Both of these things is perfectly legal to do per spec and not a "hack", and sometimes required. This extra constraint (not from spec) allows the stack to perform a tiny optimization in certain cases:
  * In the case where all vendor models and opcodes use the same CID, it cannot do anything, since all opcode handler need to be checked in every case.
  * In the case where some opcodes uses a different CID than some models, it can skip a small amount of integer comparisons in a loop.

As far as I'm concerned, you would only ever want to enable this if you need the tiny optimization.

**_VND_MSG_CID_FORCE=n**: Vendor models could only ever receive messages using 3-byte opcodes. Specifying handlers for 1- or 2-byte opcodes would just not work (meaning, you could do it, but would not receive any messages). Doing this is perfectly legal to do per spec and not a "hack", and sometimes required.

After my changes
----------------

**_VND_MSG_CID_FORCE=y**: Vendor models can only ever receive messages using 3-byte opcodes where the CID matches the model-CID. Specifying handlers for 3-byte opcodes with a CID different from the model-CID would cause an error at model_init time, as will specifying handlers for 1- or 2-byte opcodes. Both of these things is perfectly legal to do per spec and not a "hack", and sometimes required. This extra constraint (not from spec) allows the stack to perform a tiny optimization in certain cases:
  * In the case where all vendor models and opcodes use the same CID, it cannot do anything, since all opcode handler need to be checked in every case.
  * In the case where some opcodes uses a different CID than some models, it can skip a small amount of integer comparisons in a loop.

**_VND_MSG_CID_FORCE=n**: Vendor models can receive any opcode, allowing you to do everything the spec allows you to.

subsys/bluetooth/mesh/Kconfig
ludvigsj: The spec does not prohibit you from using opcodes with CID1 in a model with CID2. This is an artificial limitation we introduce.

ludvigsj: > I guess the problem is when you have more than 5-10 messages per model.

... _and_ use multiple CIDs in your network. Even then, scanning 5-10 extra handlers is not enough for me to justify the added complexity here.

subsys/bluetooth/mesh/Kconfig
PavelVPV: Going back to the initial comment: 

> This should be mentioned in the migration guide as it changes the default behavior.

This needs a note in the migration guide.

I don't understand why this option is not simply deprecated now? What do we need to wait for?

omkar3141: > I don't understand why this option is not simply deprecated now?

This is exactly what we want to do, after this PR. To not drag this. This PR saves people from doing costly hacks on their side to get SIG opcodes working with vendor models.

omkar3141 (dismissed):

subsys/bluetooth/mesh/Kconfig
PavelVPV: If the goal is to deprecate this option, why not to do it right away and get the spec compliant code without juggling Kconfig option, changing defaults, adding more text to try to explain the option behavior which is even more tangled now?

subsys/bluetooth/mesh/Kconfig
```diff
@@ -785,11 +785,15 @@ config BT_MESH_ACCESS_LAYER_MSG
...
+	default n
+	select DEPRECATED
 	help
-	  This option forces vendor model to use messages for the
-	  corresponding CID field.
+	  This option forces vendor models to only use opcodes with a CID field
+	  matching the models CID. This was previously used to enable a small
+	  optimization in networks with multiple vendor models using multiple CIDs,
+	  where the stack could skip searching the opcode handlers in models with a
+	  CID different from the one in the message.
```
PavelVPV: Again. The term “small optimization” is subjective. It’s unclear whether additional clarification is needed here, since this option is deprecated and no longer has any effect.

tests/bsim/bluetooth/mesh/src/test_access.c
```diff
@@ -97,7 +102,8 @@ static const struct {
 	},
 };
 
-static struct k_sem publish_sem;
+K_SEM_DEFINE(publish_sem, 0, 1);
+K_SEM_DEFINE(vnd_rx_sem, 0, 1);
```
PavelVPV:
```suggestion
static K_SEM_DEFINE(publish_sem, 0, 1);
static K_SEM_DEFINE(vnd_rx_sem, 0, 1);
```

tests/bsim/bluetooth/mesh/tests_scripts/access/access_vnd_rcv_sig_msg.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright 2025 Nordic Semiconductor
+# SPDX-License-Identifier: Apache-2.0
+
+source $(dirname "${BASH_SOURCE[0]}")/../../_mesh_test.sh
+
+# Test scenario:
```
PavelVPV, line 11: Please add purpose of the test.

PavelVPV (changes requested):

PavelVPV (dismissed):

doc/releases/migration-guide-4.4.rst
```diff
@@ -437,6 +280,12 @@ Bluetooth Host
   this method will be downgraded to unauthenticated when loaded from persistent storage, resulting
   in a lower security level.
 
+Bluetooth Mesh
+==============
+
+* :kconfig:option:`CONFIG_BT_MESH_MODEL_VND_MSG_CID_FORCE` has been deprecated. Enabling it no
+  longer has any effect on message handling performance.
```
omkar3141, line 452: This is not fully correct, unless it is a mistake. Enabling this option will stop the initialization of the vendor model if one model tries to use opcode from another. 

So, I think following lines should be deleted as well.
https://github.com/zephyrproject-rtos/zephyr/pull/101262/changes#diff-fced093b1243729a3f60760434e674f6c075f19c42b1e71aa44b8c1eb63721eeR1038-R1042

Also, please check in the model initialization API or documentation, if we say by mistake that different vendor models cannot use each others opcodes. If we do, we need to remove those senstences as well.

omkar3141 (changes requested):

PavelVPV:
> Hi @ludvigsj, @PavelVPV, could you elaborate why was deprecated option mentioned in migration guide but not in the release notes?
> 
> There is already mentioned deprecated API related to OOB authentication: https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/releases/release-notes-4.4.rst?plain=1#L72-L79
> 
> Seems kind of inconsistent to me. To figure out full set of deprecated functionality customer needs to read 2 documents.

https://docs.zephyrproject.org/latest/releases/index.html#migration-guides

I think it makes sense to mention the deprecation in the migration guide since the option has no effect anymore and the second sentence clarifies that. If you disagree, request a change with a reason and suggestion.

I agree that it should be mentioned in the release notes as well due to this:
https://github.com/zephyrproject-rtos/zephyr/blob/7963b69fa451c6b611040597e6d99e0dc3c61be2/doc/releases/release-notes-4.4.rst?plain=1#L4

doc/releases/migration-guide-4.4.rst
ludvigsj: It no longer has any effect on _handling performance_. I kept the check during initialization, since it doesn't do any harm and I didn't want to change more behavior than necessary until we remove the option. If you'd like to just have this option do absolutely nothing for two releases, then I'm fine with that as well, I just didn't see the point to remove this check yet.

omkar3141: Ok. But 'enabling it will cause vendor model initialization to fail, should two different vendor models share the opcodes'. This fact will become known only at run time when user does this and it is not mentioned anywhere. It is okay to keep the code, but can you mention this fact somewhere? I think it is better to mention here, because anyone dependent on this option will read this part of migration guide and become aware of consequences?

ludvigsj: There is no mention of any opcode constraints anywhere in the docs as far as I (and Copilot) could find.

alxelax (changes requested):
@ludvigsj, please update release notes with deprecation option too (see reference in my initial message).

doc/releases/migration-guide-4.4.rst
ludvigsj: @omkar3141 this fact is mentioned in the Kconfig help text. I assume people who are enabling options read the description of the option they are enabling.

omkar3141 (dismissed):

PavelVPV (approved):

alxelax (approved):

omkar3141 (approved):

PR #101264, opened 2025-12-18
[RFC] Scope-based Cleanup Helpers

pdgendt:
### Problem description

Manual resource cleanup is error-prone in C. When you acquire a resource (lock, memory, file handle), you must remember to release it on every exit path - normal returns, early returns, error conditions, etc. Missing even one cleanup leads to resource leaks or deadlocks.

### Proposed Change

This PR introduces scope-based cleanup helpers similar to the [Linux kernel Scope-based cleanup helpers](https://docs.kernel.org/core-api/cleanup.html). A RAII-like pattern for C - Brings the safety of C++ RAII, Go's defer, or Java's try-with-resources to C.

The Zephyr project can provide these helper constructs for users to write safer **_application code_**, if supported by the compiler.

[doc preview](https://builds.zephyrproject.io/zephyr/pr/101264/docs/kernel/cleanup.html)

See the following sample of how a mutex lock guard simplifies the code:
```c
/* Before: Manual unlock needed */
static int foo(void)
{
	int ret = 0;

	k_mutex_lock(&lock, K_FOREVER);
	/* code body without return statements and goto jumps */

exit_label:
	k_mutex_unlock(&lock);
	return ret;
}
```
```c
/* After: Automatic unlock when scope ends */
static int bar(void)
{
	scope_guard(k_mutex)(&lock);
	/* code body that allows early returns */

	return 0;
}
/* Alternative: Defer an unlock */
static int bar_alt(void)
{
	int ret = k_mutex_lock(&lock, K_NO_WAIT);
	if (ret < 0) {
		return ret;
	}
	scope_defer(k_mutex_unlock)(&lock);
	/* code body that allows early returns */

	return 0;
}
```

Automatic cleanup of allocated resources using `DEFER(k_free)`:

```c
static void baz(void)
{
	void *ptr = k_malloc(42);
	scope_defer(k_free)(ptr); // <- safe even if ptr is NULL

	/* code body */
}
```

- Compiler dependent, so this can only be used in application code, not in the kernel or subsystems!
- Macro names (uppercase vs lowercase `guard`/`defer`)
- Put it behind a Kconfig option?

### Alternatives Considered

There's no alternative AFAIK for calling a cleanup function at the end of a scope. The closest alternative would be some macro abusing a `for`-loop scope with a custom break statement (no returns!).


include/zephyr/cleanup.h
```diff
@@ -0,0 +1,272 @@
...
+ * @brief Use a guard with automatic variable name
+ *
+ * This macro creates a guard variable with an automatically generated unique name.
+ * The guard will acquire the lock on initialization and release it when going out of scope.
+ *
+ * @param _name Name of the guard type (defined by DEFINE_GUARD or DEFINE_CLASS with guard_ prefix)
+ *
+ * Usage: GUARD(k_mutex)(&my_mutex);
+ */
+#define GUARD(_name) CLASS(guard_##_name, __UNIQUE_ID(guard))
```
fabiobaltieri: I suggest doing it lowercase as it is in Linux, the reasoning behind it is that it can only be used in function, so it very much be named like a function, the fact that it's implemented with macros is an irrelevant details, it's not like when we alias functions with macros we make them uppercase

and doing container_of uppercase was a mistake

pdgendt: The issue I had with this, is the "inconsistency" linux has itself `CLASS()` vs `guard()`. I figured doing all uppercase for these macros would at least keep it straightforward.
I can imagine `class` might be a keyword somewhere in the future?

include/zephyr/toolchain/gcc.h
```diff
@@ -317,6 +317,10 @@ do {                                                                    \
 #define __attribute_nonnull(...) __attribute__((nonnull(__VA_ARGS__)))
 #endif
 
+#ifndef __cleanup
```
keith-packard, line 324: Should this be guarded with a `#if __has_attribute(__cleanup__)`?

pdgendt: This isn't done anywhere in Zephyr? 🤔

cfriedt:
Cool concept. It's a bit tricky to enforce their usage only in application code. Can the preprocessor be used to generate a build assertion if they are being used as part of the kernel?

fabiobaltieri:
> I was contemplating introducing a Kconfig option to make it an opt-in feature, but I think this might be overkill.

I don't think it's a bad idea thugh, simple enough, better than sprinkling define around the code

pdgendt:
> > I was contemplating introducing a Kconfig option to make it an opt-in feature, but I think this might be overkill.
> 
> I don't think it's a bad idea thugh, simple enough, better than sprinkling define around the code

I already had it listed under "Concerns and Unresolved Questions", so if this is the consensus, I can add it.

pdgendt:
> > I already had it listed under "Concerns and Unresolved Questions", so if this is the consensus, I can add it.
> 
> Actually what I would really like is for some of this code to sneak in the main code base and see if anyone notices really, easiest way to find how not supported is it.

It was introduced in GCC 3 almost 25 years ago, but I didn't find a similar feature in IAR, for example.
CC @RobinKastberg

fabiobaltieri (dismissed):
Spiritual LGTM, I think this is cool.

RobinKastberg:
@pdgendt we implemented support for attribute cleanup in IAR, since we found it in zephyr kernel codebase and missed the `__GNUC__ ` guard here: 
https://github.com/zephyrproject-rtos/zephyr/blob/ed5bfabf2ed016f3260318d9f64e36df78e0be26/include/zephyr/spinlock.h#L377

But I think leaving it off by default is fine for now. It should probably be disabled in MISRA mode as well

pdgendt:
> @pdgendt - another option that is really quite similar is to use (something like) `pthread_cleanup_push()` and `pthread_cleanup_pop()`.
> 
> [pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cleanup_pop.htm](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_cleanup_pop.htm)
> 
> That would work with or without explicit compiler support for the `__cleanup__` attribute and would work equivalently in both in kernel and user threads.
> 
> The cleanup action is RAII-like, so when a scope is exited, the cleanup handler is invoked.
> 
> I just happened to be rewriting `pthread_cleanup_push()` and `pthread_cleanup_pop()` today so that they would call in to `k_thread_cleanup_push()` and `k_thread_cleanup_pop()` instead and realized it's mostly equivalent to what the `__cleanup__` attribute would do.
> 
> The [implementation in Zephyr](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/posix/pthread.h?rgh-link-date=2026-01-08T17%3A00%3A58Z#L471) is pretty trivial.

Isn't it what I described in the "Alternatives considered" in my description? It looks like abusing a loop, where you can't use early returns?

include/zephyr/cleanup.h
aescolar: Please do keep it uppercase. a) this is not a function (It is also instantiating a variable); b) keeping all macros uppercase helps developers know what they are looking at; c) Using C++ reserved keywords or something that looks like a keyword would be a bad idea as it is likely to confuse developers.

pdgendt:
So to continue the discussion from the Arch meeting, I think only the naming part needs a resolution?

- Cleanup class vs cleanup helper
- UPPERCASE vs lowercase

Thanks for the positive feedback everyone!

include/zephyr/kernel.h
```diff
@@ -6799,6 +6800,37 @@ void k_sys_runtime_stats_enable(void);
  */
 void k_sys_runtime_stats_disable(void);
 
+#if defined(CONFIG_CLEANUP_CLASSES)
```
aescolar: I think it would be better to add these in `cleanup.h` instead:

`kernel.h` is included from too many places, while the actual users of these, would be including `cleanup.h`. So, instead of always pulling `cleanup.h` and these definitions even for those who don't use this or where they don't use it.

pdgendt: Or how about creating `include/zephyr/cleanup/kernel.h`?

aescolar: > Or how about creating include/zephyr/cleanup/kernel.h?

Seems like overkill to me (though maybe I'm missing something?), but I would not oppose it :)

pdgendt: Note that these are added for convenience, as I think these would be commonly used. But we can omit them and let users define the cleanup classes/helpers themselves.

aescolar: I think it is very fine to add these, it is just about the place and about pulling cleanup.h from kernel.h

pdgendt:
Updated proposal (after some pondering about clear naming):

- Rename the feature to "Scope-based cleanup helpers" with a focus on "scope"
- Use `scope_`/`SCOPE_` prefixes to "namespace" the macros.
- Uppercase for the `_DEFINE` macros used outside of functions
- Lowercase for the variable declaration macros
- Moved kernel specific pre-defined macros to `include/zephyr/cleanup/kernel.h`
- Use init/exit terminology instead of constructor/destructor

doc/services/cleanup.rst
```diff
@@ -0,0 +1,227 @@
...
+* Automatic unlocking of mutexes and semaphores
+* Automatic freeing of dynamically allocated memory
+* Ensuring cleanup actions occur on all code paths (including early returns)
+* Reducing boilerplate cleanup code
+
+.. warning::
+
+   The cleanup mechanism is implemented using the ``__cleanup`` attribute.
+   If the toolchain doesn't support this attribute, the macros expand to empty
+   definitions and the API is not available.
```
aescolar:
```suggestion
   If the toolchain doesn't support this attribute, the API is not available.
```
Minor: They are not defined at all in that case, or?

include/zephyr/cleanup.h
```diff
@@ -0,0 +1,263 @@
...
+		    (Z_SCOPE_DEFER_DEFINE_N(_func, __VA_ARGS__)))
+
+/** @cond INTERNAL_HIDDEN */
+
+/**
+ * @brief Helper macro to generate a unique identifier with prefix
+ * @param _prefix Prefix for the unique identifier
+ * @internal
+ */
+#define Z_UNIQUE_ID(_prefix) _CONCAT(_CONCAT(Z_UNIQUE_ID_, _prefix), __COUNTER__)
```
aescolar, line 176: minor: maybe this and Z_COMMA would belong better in sys/util.h or the like?

aescolar (dismissed):
Thanks @pdgendt this addresses my main concerns.
(The SCOPE_ prefix achieves the same as the sys_ one.
Though I would have kept the macros that instantiate the variables in capital as mentioned).

fabiobaltieri (dismissed):

aescolar (dismissed):

carlescufi (dismissed):

doc/kernel/cleanup.rst
```diff
@@ -0,0 +1,226 @@
+.. _cleanup_api:
+
+Scope-based Cleanup Helpers
```
nashif, line 3: does not feel like a service to me, feels more like something that should show up here https://docs.zephyrproject.org/latest/kernel/index.html tbh.

pdgendt: Moved the doc page to `kernel`.

aescolar (approved):

fabiobaltieri (approved):

peter-mitsis (approved):

Thalley:
@pdgendt I was looking into using this but got curious: Do we have any benchmarks (memory and/or performance) on using these automatic helpers compared to manually cleanup? 

e.g. does

```c
/* Before: Manual unlock needed */
static int foo(void)
{
	int ret = 0;

	k_mutex_lock(&lock, K_FOREVER);
	/* code body without return statements and goto jumps */

	bar();

exit_label:
	k_mutex_unlock(&lock);
	return ret;
}
```

Have a higher, lower or same footprint as

```c
static int bar(void)
{
	scope_guard(k_mutex)(&lock);
	/* code body that allows early returns */

	bar();

	return 0;
}
```

Additionally, since the example given for scope_guard 

```c
 * SCOPE_GUARD_DEFINE(k_mutex, struct k_mutex *,
 *                    (void)k_mutex_lock(_T, K_FOREVER),
 *                    (void)k_mutex_unlock(_T));
```

is non-compliant with our coding guidelines, how would you suggest to handle any errors with locking the mutex? Instead of `(void)k_mutex_lock(_T, K_FOREVER)` then do `__maybe_unused int _err = k_mutex_lock(_T, K_FOREVER); __ASSERT_NO_MSG(_err == 0)`?

pdgendt:
> @pdgendt I was looking into using this but got curious: Do we have any benchmarks (memory and/or performance) on using these automatic helpers compared to manually cleanup?

We don't have comparison benchmarks, there is some overhead like I [mentioned in the architecture working group meeting](https://github.com/zephyrproject-rtos/zephyr/pull/101264#issuecomment-3773912156).
It's trading readability/coding ergonomics for raw performance, that's why it's also opt-in.

Some insights from Claude:

---

Code generation comparison (Arm Cortex-M, -Os)

Manual pattern at the unlock site:
```asm
bl k_mutex_unlock       ; direct call with &lock already in r0
pop {pc}
```

scope_guard pattern at each return:
```asm
ldr r0, [sp, #offset]  ; reload &lock from stack slot
cbz r0, .skip          ; NULL check (guard's extra branch)
bl  k_mutex_unlock
.skip:
pop {pc}
```

Overhead breakdown
```
┌─────────────────────────┬────────┬────────────────────────────┐
│          Cost           │ Manual │        scope_guard         │
├─────────────────────────┼────────┼────────────────────────────┤
│ Extra stack word        │ —      │ +1 pointer (4/8 bytes)     │
├─────────────────────────┼────────┼────────────────────────────┤
│ NULL check + branch     │ —      │ +2 instructions per return │
├─────────────────────────┼────────┼────────────────────────────┤
│ Extra load              │ —      │ +1 instruction per return  │
├─────────────────────────┼────────┼────────────────────────────┤
│ Pointer indirection     │ —      │ +1 dereference             │
├─────────────────────────┼────────┼────────────────────────────┤
│ Inline wrapper overhead │ —      │ none (static inline)       │
└─────────────────────────┴────────┴────────────────────────────┘
```

Verdict: slightly higher footprint

The `__cleanup__` attribute is zero-overhead for the happy path calling convention — the unlock call itself happens exactly once either way. But the guard adds:

1. One extra stack slot — the stored struct k_mutex * pointer
2. ~3 extra instructions per return point — a load, a NULL check, and a conditional branch
3. The `NULL` check is never true for `scope_guard` (the pointer is always non-null after a successful lock), but the compiler typically cannot prove this and emits it anyway

With multiple early returns the overhead stays constant (same extra instructions at each exit, just more of them), while the manual goto exit_label approach adds only a branch.

---

> Additionally, since the example given for scope_guard
> 
> ```c
>  * SCOPE_GUARD_DEFINE(k_mutex, struct k_mutex *,
>  *                    (void)k_mutex_lock(_T, K_FOREVER),
>  *                    (void)k_mutex_unlock(_T));
> ```
> 
> is non-compliant with our coding guidelines, how would you suggest to handle any errors with locking the mutex? Instead of `(void)k_mutex_lock(_T, K_FOREVER)` then do `__maybe_unused int _err = k_mutex_lock(_T, K_FOREVER); __ASSERT_NO_MSG(_err == 0)`?

Why is it non-compliant? I explicitly cast the result to `void`, same as `ARG_UNUSED`. We could add the assert if desired.

Thalley:
> Why is it non-compliant? I explicitly cast the result to `void`, same as `ARG_UNUSED`. We could add the assert if desired.

[Coding guidelines](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html) [rule 9](https://gitlab.com/MISRA/MISRA-C/MISRA-C-2012/Example-Suite/-/blob/master/D_04_07.c), `If a function returns error information, then that error information shall be tested` does not allow for simply casting an error return to `(void)` :) But if these macros aren't allowed in-tree due to the toolchain requirements, then it's not really a problem in any case

pdgendt:
> > Why is it non-compliant? I explicitly cast the result to `void`, same as `ARG_UNUSED`. We could add the assert if desired.
> 
> [Coding guidelines](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html) [rule 9](https://gitlab.com/MISRA/MISRA-C/MISRA-C-2012/Example-Suite/-/blob/master/D_04_07.c), `If a function returns error information, then that error information shall be tested` does not allow for simply casting an error return to `(void)` :) But if these macros aren't allowed in-tree due to the toolchain requirements, then it's not really a problem in any case

Well, there's no generic way to handle it, except asserting, but that is compiled out if disabled. Users can create custom lock/unlock guards if desired.

```c
SCOPE_GUARD_DEFINE(
	k_mutex_log, struct k_mutex *,
	{
		int ret = k_mutex_lock(_T, K_FOREVER);
		if (ret < 0) {
			LOG_ERR("lock failed");
		}
	},
	{
		int ret = k_mutex_unlock(_T);
		if (ret < 0) {
			LOG_ERR("unlock failed");
		}
	});

// ....

static void some_func(void)
{
	scope_guard(k_mutex_log)(&lock);
}

```

PR #101266, opened 2025-12-18
drivers: intc: vim: disable all interrupts at init

natto1784:
Disable all interrupts when the VIM interrupt controller is initialized so that it doesn't encounter any stray interrupts that were not enabled on Zephyr.

### DIFF
--- a/drivers/interrupt_controller/intc_vim.c
+++ b/drivers/interrupt_controller/intc_vim.c
@@ -65,6 +65,11 @@ void z_vim_irq_init(void)
 		 "(%" PRIu32 ") interrupts",
 		 CONFIG_NUM_IRQS, num_of_irqs);
 	LOG_DBG("VIM: Number of IRQs = %u\n", num_of_irqs);
+
+	/* Start with all interrupts masked */
+	for (int i = 0; i <= VIM_MAX_GROUP_NUM; i++) {
+		sys_write32(VIM_GRP_INTR_EN_CLR_MSK_MASK, VIM_INTR_EN_CLR(i));
+	}
 }
 
 void z_vim_irq_priority_set(unsigned int irq, unsigned int prio, uint32_t flags)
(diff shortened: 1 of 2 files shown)


include/zephyr/drivers/interrupt_controller/intc_vim.h
```diff
@@ -48,73 +48,73 @@
 
 /* RAW */
 
-#define VIM_GRP_RAW_STS_MASK                 (BIT_MASK(32))
+#define VIM_GRP_RAW_STS_MASK                 (BIT64_MASK(32))
```
glneo, line 50: The mask changes should all go in its own patch.

include/zephyr/drivers/interrupt_controller/intc_vim.h
```diff
@@ -48,73 +48,73 @@
...
 #define VIM_GRP_RAW_STS_SHIFT                (0x00000000U)
 #define VIM_GRP_RAW_STS_RESETVAL             (0x00000000U)
-#define VIM_GRP_RAW_STS_MAX                  (BIT_MASK(32))
+#define VIM_GRP_RAW_STS_MAX                  (BIT64_MASK(32))
```
glneo: These `_MAX` definitions are never used, just remove them. Same for `_RESETVAL` which is always 0 and `_SHIFT` which can be easily derived from the least set bit in the `_MASK` value.

natto1784: Agreed, will do.

drivers/interrupt_controller/intc_vim.c
```diff
@@ -65,6 +65,10 @@ void z_vim_irq_init(void)
 		 "(%" PRIu32 ") interrupts",
 		 CONFIG_NUM_IRQS, num_of_irqs);
 	LOG_DBG("VIM: Number of IRQs = %u\n", num_of_irqs);
+
```
glneo, line 68: I'd add a quick comment here, something like:

`/* Start with all interrupts masked */`

Other than that, LGTM.

glneo (dismissed):

glneo (approved):

dcpleung (approved):

henrikbrixandersen (approved):

PR #101271, opened 2025-12-18
drivers: watchdog: wdt_mcux_wwdt: fix window calculation and warning interrupt handling

Holt-Sun:
This PR fixes several issues in the NXP WWDT driver and adds board-specific configuration for mimxrt700_evk:

### Summary of Changes

1. **Fix WWDT window value calculation** (Patch 2)
   - Corrected window calculation formula: `windowValue = timeoutValue - min_window_ticks`
   - Added proper range validation for timeout values (MIN_TIMEOUT to MAX_TIMEOUT)
   - Previous implementation incorrectly set window value to minimum timeout directly

2. **Reject callback without warning interrupt** (Patch 3)
   - Return `-ENOTSUP` when callback is provided but `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG` is 0
   - Added error logging to clarify configuration requirements
   - Prevents silent no-op when callback would never fire

3. **Configure ITRC for WWDT reset handling on mimxrt700_evk** (Patch 4)
   - Reconfigure ITRC OUT3_SEL0/OUT4_SEL0 registers to 0xAAAAAA0A
   - Fixes issue where ROM configuration causes ITRC to reset system faster than WWDT
   - Allows RSTCTL to properly handle WWDT reset events and warning interrupts
   - Enables warning interrupt in board test configuration

4. **Enable write-through cache for mimxrt700 watchdog tests** (Patch 1)
   - Adds `CONFIG_ARM_MPU_SRAM_WRITE_THROUGH=y` to ensure timely data flush to SRAM


Albort12138 (approved):

yvesll (approved):

ZhaoxiangJin (dismissed):

hakehuang:
board test pass on v4.3.0-940-gc9b01e8257db

tests/drivers/watchdog/wdt_basic_api/testcase.yaml
```diff
@@ -169,3 +169,9 @@ tests:
     extra_configs:
       - CONFIG_PM=y
       - CONFIG_PM_DEVICE=y
+  drivers.watchdog.basic_api.dcache:
+    platform_allow:
```
hakehuang: if you add a case for this you need add a filter see https://github.com/zephyrproject-rtos/zephyr/blob/c3190d34aa912d39cd082db822f6ebb332806a2b/tests/drivers/uart/uart_async_api/testcase.yaml#L66

also you need filter out other cases for rt700.

Holt-Sun: Since the testcase.yaml update has been reverted. Comment resolved.

hakehuang (dismissed):

drivers/watchdog/wdt_mcux_wwdt.c
```diff
@@ -107,15 +109,22 @@ static int mcux_wwdt_install_timeout(const struct device *dev,
...
-	      data->wwdt_config.windowValue))) {
+	/*
+	 * Window value is specified as time before the timeout when feeding
+	 * is allowed. Therefore, calculate the window value as:
+	 * windowValue = timeoutValue - min_window_ticks(maybe 0)
+	 */
+	data->wwdt_config.windowValue =
+		data->wwdt_config.timeoutValue - MSEC_TO_WWDT_TICKS(clock_freq, cfg->window.min);
+
+	if (data->wwdt_config.windowValue > MAX_WINDOW) {
```
yvesll: The check is redundant as it will not exceed `timeoutValue`, which was checked in line 112.

drivers/watchdog/wdt_mcux_wwdt.c
```diff
@@ -107,15 +109,22 @@ static int mcux_wwdt_install_timeout(const struct device *dev,
...
+		LOG_ERR("Timeout value out of range");
+		return -EINVAL;
 	}
 
-	if ((data->wwdt_config.timeoutValue < MIN_TIMEOUT) ||
-	    ((data->wwdt_config.windowValue != 0xFFFFFFU) &&
-	     (data->wwdt_config.timeoutValue <
-	      data->wwdt_config.windowValue))) {
+	/*
+	 * Window value is specified as time before the timeout when feeding
```
yvesll: The description from RM may be more clear:
> Specifies the highest timer value in which a watchdog feed can occur.

And here or in other comments, it shall tell user WWDT uses a count-down counter

drivers/watchdog/wdt_mcux_wwdt.c
```diff
@@ -22,6 +22,8 @@
 LOG_MODULE_REGISTER(wdt_mcux_wwdt);
 
 #define MIN_TIMEOUT 0xFF
+#define MAX_TIMEOUT WWDT_TC_COUNT_MASK
+#define MAX_WINDOW  WWDT_WINDOW_WINDOW_MASK
```
yvesll: it's redundant

drivers/watchdog/wdt_mcux_wwdt.c
Holt-Sun: 112 make sure that data->wwdt_config.timeoutValue > MAX_TIMEOUT. The MAX_WINDOW is not MAX_TIMEOUT although in most cases they are same, but they are different.

drivers/watchdog/wdt_mcux_wwdt.c
Holt-Sun: They are used for validation. It is easy to give a max time bigger than allowed.

drivers/watchdog/wdt_mcux_wwdt.c
yvesll: They are same, both to the counter register valuable width. So line 123 ensures `windowValue` always smaller than `timeoutValue`

Holt-Sun: For now, they are all same, I will keep one. 
Update the codes.

yvesll (approved):

Holt-Sun:

ZhaoxiangJin (dismissed):

drivers/watchdog/wdt_mcux_wwdt.c
```diff
@@ -107,31 +108,37 @@ static int mcux_wwdt_install_timeout(const struct device *dev,
...
-		return -ENOTSUP;
+	if (cfg->callback) {
+		if (CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG > 0) {
+			data->callback = cfg->callback;
+			data->wwdt_config.warningValue = CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG;
+		} else {
+			LOG_ERR("Warning interrupt callback requires "
+				"CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG > 0");
+			return -ENOTSUP;
+		}
```
zejiang0jason, line 146: suggestions:
1. Seems the callback change is not related with the WWDT window improvement, suggest use a separate patch.
2. Add some comment to explain why do like this.

zejiang0jason (approved):

ZhaoxiangJin (approved):

HyperionCoding:
This change (and the previous one that added `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG` requirement) made `WDT_FLAG_RESET_NONE` unusable. For debugging I used "callback only" timeout by setting flags to `WDT_FLAG_RESET_NONE` and callback to my callback that prints the current thread etc. Having to set `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG` means that every timeout is now two timeouts. I could work around it by setting the actual watchdog timeout to the warning length but the kconfig `range 0 1023` limits it lower than I want.

It is an opinionated question whether Zephyr should make the callback warning only or also the timeout but to restore the previous (and full) functionality setting callback without warning should be allowed.

Holt-Sun:
> This change (and the previous one that added `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG` requirement) made `WDT_FLAG_RESET_NONE` unusable. For debugging I used "callback only" timeout by setting flags to `WDT_FLAG_RESET_NONE` and callback to my callback that prints the current thread etc. Having to set `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG` means that every timeout is now two timeouts. I could work around it by setting the actual watchdog timeout to the warning length but the kconfig `range 0 1023` limits it lower than I want.
> 
> It is an opinionated question whether Zephyr should make the callback warning only or also the timeout but to restore the previous (and full) functionality setting callback without warning should be allowed.

@HyperionCoding, thanks, I think the concern on the reset-none path is valid, but it predates PR #101271. The `-ENOTSUP` restriction was introduced earlier in dbbdd3c6cfe; #101271 mainly kept that behavior and clarified the logging.

For `WDT_FLAG_RESET_NONE`, allowing a callback with `CONFIG_WDT_MCUX_WWDT_WARNING_INTERRUPT_CFG=0` restores the regular callback-at-expiry flow without introducing an extra warning stage. For reset-enabled modes, keeping the warning requirement still makes sense so the callback can run before reset.

I’ve prepared that change in https://github.com/zephyrproject-rtos/zephyr/pull/107188

PR #101272, opened 2025-12-18
samples: openthread: add esp platforms to the border_router (eth)

wmrsouza:
Add the following esp platforms to the openthread border_router sample (eth) using rcp with hdlc trhough uart:
- esp32s3_devkitc/esp32s3/procpu
- esp_threadbr/esp32s3/procpu

### DIFF
--- a/samples/net/openthread/border_router/boards/esp32s3_devkitc_procpu.overlay
+++ b/samples/net/openthread/border_router/boards/esp32s3_devkitc_procpu.overlay
@@ -34,6 +34,35 @@
 	pinctrl-names = "default";
 };
 
+&spim2_default {
+	group1 {
+		pinmux = <SPIM2_MISO_GPIO2>,
+			 <SPIM2_SCLK_GPIO6>,
+			 <SPIM2_CSEL_GPIO10>;
+	};
+
+	group2 {
+		pinmux = <SPIM2_MOSI_GPIO7>;
+		output-low;
+	};
+};
+
+&spi2 {
+	pinctrl-0 = <&spim2_default>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	eth_w5500: eth-w5500@0 {
+		compatible = "wiznet,w5500";
+		status = "okay";
+		reg = <0x0>;
+		spi-max-frequency = <DT_FREQ_M(20)>;
+		int-gpios = <&gpio0 15 GPIO_ACTIVE_LOW>;
+		reset-gpios = <&gpio0 5 GPIO_ACTIVE_LOW>;
+		zephyr,random-mac-address;
+	};
+};
+
 &psram0 {
 	size = <DT_SIZE_M(2)>;
 };
--- a/samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf
+++ b/samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf
@@ -1,0 +1,58 @@
+#
+# Copyright (c) 2026 Espressif Systems (Shanghai) Co., Ltd.
+#
+
+CONFIG_ESP_SPIRAM=y
+CONFIG_ESP_SPIRAM_BSS_RELOC_LIBS_AND_OBJS="libopenthread-ftd.a libopenthread-cli-ftd.a"
+
+# Generic network options
+CONFIG_SPI=y
+CONFIG_ETH_DRIVER=y
+CONFIG_NET_L2_ETHERNET=y
+CONFIG_NET_L2_ETHERNET_MGMT=y
+
+CONFIG_NET_CONFIG_AUTO_INIT=y
+CONFIG_NET_CONFIG_NEED_IPV4=y
+CONFIG_NET_CONFIG_SETTINGS=y
+
+# lib & os
+CONFIG_EVENTS=y
+
+# CPP library
+CONFIG_CPP=y
+
+# Use NVS as settings backend
+CONFIG_NVS=y
+
+# Shell
+CONFIG_SHELL_ARGC_MAX=26
+CONFIG_SHELL_CMD_BUFF_SIZE=512
+CONFIG_SHELL_STACK_SIZE=5120
+
+# Network shell
+CONFIG_NET_SHELL=y
+CONFIG_SHELL=y
+CONFIG_SHELL_GETOPT=y
+CONFIG_SHELL_TAB=y
+CONFIG_SHELL_TAB_AUTOCOMPLETION=y
+
+# Enable Openthread RCP host interface
+CONFIG_IEEE802154=n
+CONFIG_HDLC_RCP_IF=y
+CONFIG_OPENTHREAD_MANUAL_START=y
+CONFIG_OPENTHREAD_RCP_RESTORATION_MAX_COUNT=10
+
+# Border Router mDNS host name and service base name
+CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_VENDOR_NAME="Espressif Systems"
+CONFIG_OPENTHREAD_ZEPHYR_BORDER_ROUTER_BASE_SERVICE_NAME="ESP-BorderRouter"
+
+# As per Thread v1.2.0 Conformance, a Thread Border Router MUST be able to hold a Multicast Listener
+# Table in memory with at least 75 entries.
+# 75 entries for conformance, 10 extra for other purposes, if any.
+CONFIG_NET_MAX_MCAST_ROUTES=85
+CONFIG_NET_IF_MCAST_IPV6_ADDR_COUNT=85
+
+# Network diagnostics
+CONFIG_OPENTHREAD_NET_DIAG_VENDOR_NAME="Espressif Systems"
+CONFIG_OPENTHREAD_NET_DIAG_VENDOR_MODEL="ESP-BR"
+CONFIG_OPENTHREAD_NET_DIAG_VENDOR_SW_VERSION="v1.0.0"
--- a/samples/net/openthread/border_router/sample.yaml
+++ b/samples/net/openthread/border_router/sample.yaml
@@ -25,3 +25,16 @@ tests:
     platform_allow:
       - esp32s3_devkitc/esp32s3/procpu
       - esp_threadbr/esp32s3/procpu
+  sample.net.openthread.border_router.esp_rcp_host_eth.esp_devkits:
+    build_only: true
+    extra_args:
+      - EXTRA_CONF_FILE="overlay-ot-rcp-host-eth-esp.conf"
+    platform_allow:
+      - esp32s3_devkitc/esp32s3/procpu
+  sample.net.openthread.border_router.esp_rcp_host_eth.esp_threadbr:
+    build_only: true
+    extra_args:
+      - SHIELD="esp_threadbr_ethernet"
+      - EXTRA_CONF_FILE="overlay-ot-rcp-host-eth-esp.conf"
+    platform_allow:
+      - esp_threadbr/esp32s3/procpu


modules/openthread/platform/udp.c
```diff
@@ -9,6 +10,7 @@
 #include "openthread/platform/udp.h"
 #include "openthread_border_router.h"
 #include "sockets_internal.h"
+#include <assert.h>
```
sylvioalves, line 12: Is this file change really needed?

sylvioalves: Ok, `udp.c` uses `assert` call and for some reason has indirect usage of `assert.h`. So it is needed.

wmrsouza: I think so because without this include I get a lot of
```implicit declaration of function 'assert'```

Comment (rlubos) [inline modules/openthread/platform/udp.c:12]: Such a change should be in a separate commit though, it has nothing to do with the sample itself (which the commit claims to update).

File: samples/net/openthread/border_router/sample.yaml [hunk 2]
```diff
@@ -18,3 +18,16 @@ tests:
...
+      - EXTRA_CONF_FILE="overlay-ot-rcp-host-eth-esp32.conf"
+    platform_allow:
+      - esp32s3_devkitc/esp32s3/procpu
+  sample.net.openthread.border_router.esp32_eth_openthread_rcp_host.esp_threadbr:
+    build_only: true
+    extra_args:
+      - SHIELD="esp_threadbr_ethernet"
+      - EXTRA_CONF_FILE="overlay-ot-rcp-host-eth-esp32.conf"
+    platform_allow:
+      - esp_thread_br/esp32s3/procpu
```
sylvioalves: `esp_threadbr/esp32s3/procpu`

samples/net/openthread/border_router/boards/esp32s3_devkitc_procpu.overlay
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2026 Espressif Systems (Shanghai) CO LTD
```
sylvioalves: `Co., Ltd.`

soc/espressif/common/generate_esp_spiram_relocate_bss.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+  list(REMOVE_DUPLICATES relocate_objects)
+
+  foreach(obj IN LISTS relocate_objects)
+    string(STRIP "${obj}" obj)
+    if(NOT obj STREQUAL "")
+      # EXCLUDE_FILE pattern: *libname.a:* or *file.o:*
+      string(APPEND exclude_patterns "*${obj}:* ")
+      # .ext_ram.data: KEEP(*libname.a:(.bss .bss.*))
+      string(APPEND extram_lines "    KEEP(*${obj}:(.bss .bss.*))\n")
```
sylvioalves: I think this is regular BSS data, `KEEP` should no be there:
`string(APPEND extram_lines "    *${obj}:(.bss .bss.*)\n")`.

Is there a reason to use that?

modules/openthread/platform/udp.c
wmrsouza: It is alredy implemented in a separated commit. Only waiting more changes to submit new commits to this PR

soc/espressif/common/generate_esp_spiram_relocate_bss.cmake
wmrsouza: I'll check

wmrsouza: That's the rationale for this fragment:

What it does
- string(APPEND extram_lines ...) – Appends the given string to the variable extram_lines (initialized as "" on line 33).
- " KEEP(*${obj}:(.bss .bss.*))\n" – For each library/object in relocate_objects (e.g. libopenthread-ftd.a), it adds one line of linker script content to extram_lines.

Meaning of the linker syntax
- KEEP(...) – Tells the linker to keep these sections even if they look unused (so they are not dropped and their symbols stay).
- *libopenthread-ftd.a:(.bss .bss.*) – “From the input file matching *libopenthread-ftd.a, take the sections .bss and .bss.*.”
- So for each obj, the script generates a line that explicitly pulls that library’s BSS (and .bss.*) into whatever output section will use extram_lines.

Where it’s used
- extram_lines is written into the generated linker script and included in the .ext_ram.data (or .ext_ram) output section.
- That section is placed in SPIRAM (external RAM).

Net effect
- Together with the EXCLUDE_FILE logic (line 56), this line does the “relocate BSS to SPIRAM” part:
  - The other line excludes those libraries’ BSS from the normal .bss (internal RAM).
  - This line includes the same libraries’ .bss / .bss.* into the .ext_ram.data section in SPIRAM.
So: for each obj in relocate_objects, it appends a KEEP(*obj:(.bss .bss.*)) line so that library’s BSS is placed in external RAM instead of internal RAM.

Those are relevant fragments from the linker script generated for the concrete case:
```
...
  /* Shared RAM */
  .dram0.bss (NOLOAD) :
  {
...
#if defined(CONFIG_ESP_SPIRAM) && ESP_SPIRAM_RELOCATE_BSS_ACTIVE
    *(EXCLUDE_FILE(
*libopenthread-ftd.a:* *libopenthread-cli-ftd.a:*
    ) .bss)
    *(EXCLUDE_FILE(
*libopenthread-ftd.a:* *libopenthread-cli-ftd.a:*
    ) .bss.*)
#else
    *(.bss)
    *(.bss.*)
#endif /* CONFIG_ESP_SPIRAM && ESP_SPIRAM_RELOCATE_BSS_ACTIVE */
...
  } GROUP_LINK_IN(RAMABLE_REGION)
...
#if defined(CONFIG_ESP_SPIRAM)
...
  /* This section holds .ext_ram.bss data, and will be put in PSRAM */
  .ext_ram.data (NOLOAD) :
  {
...
#if defined(CONFIG_ESP_SPIRAM) && ESP_SPIRAM_RELOCATE_BSS_ACTIVE
    KEEP(*libopenthread-ftd.a:(.bss .bss.*))
    KEEP(*libopenthread-cli-ftd.a:(.bss .bss.*))

#endif
...
  } GROUP_LINK_IN(EXT_DRAM_REGION)
#endif /* CONFIG_ESP_SPIRAM */
...

That's the result I was expecting when I prepared this commit.
Let me know if you think we should have a different result.

File: modules/openthread/platform/udp.c (hunk 1 above)
Comment (sylvioalves) [reply modules/openthread/platform/udp.c:12]: You can remove this from this PR then.

File: samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf
Comment (uLipe) [inline samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf:1]: nit: Since this is an esp32 config specific file, why it is outside from the boards folder?

Comment (wmrsouza) [reply samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf:1]: This .conf is outside the boards folder because it is an extra conf file common to all Espressif Systems SoCs in the host role and Ethernet communication (I'm using the same nomenclature em file placement used by the other vendors)

File: soc/espressif/common/generate_esp_spiram_relocate_bss.cmake (hunk 4 above)
Comment (sylvioalves) [reply soc/espressif/common/generate_esp_spiram_relocate_bss.cmake]: The explanation is correct but the reasoning isn't it I think. KEEP prevents gc sections..Do this:

Build once with KEEP() and once without, then compare the map files:

1) with KEEP()
west build -p -b esp32s3_devkitc/esp32s3/procpu ... 
cp build/zephyr/zephyr.map map_with_keep.txt

2) Without KEEP()
west build -p -b esp32s3_devkitc/esp32s3/procpu ...
cp build/zephyr/zephyr.map map_without_keep.txt

Then compare the `ext_ram` section in those files.

If no diff (which I expect), it proves KEEP() has no effect here because those .bss symbols are actively referenced by OT code and the linker keeps them regardless.

KEEP() only matters for sections with `no direct code references` like vector tables or buffers accessed only via linker symbols. OT BSS is regular global data referenced by compiled code and the linker already knows it's needed.

Comment (wmrsouza) [reply soc/espressif/common/generate_esp_spiram_relocate_bss.cmake]: Sorry, now I see: you are only against the use of KEEP
You are totally right

Comment (wmrsouza) [reply soc/espressif/common/generate_esp_spiram_relocate_bss.cmake]: Removing KEEP() worked, @sylvioalves, since there was  nothing to be gc

File: samples/net/openthread/border_router/boards/esp_threadbr_procpu.conf [hunk 5]
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright 2026, Espressif Systems
```
raffarost: `Copyright (c) 2026 Espressif Systems (Shanghai) Co., Ltd.` would look best (standard)

samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp32.conf
```diff
@@ -0,0 +1,58 @@
+#
+# Copyright 2026, Espressif Systems
```
raffarost: same

soc/espressif/common/CMakeLists.txt
```diff
@@ -89,7 +89,47 @@ elseif(CONFIG_SOC_ESP32_APPCPU OR CONFIG_SOC_ESP32S3_APPCPU)
...
+        set(ESP_SPIRAM_RELOCATE_BSS_OBJECTS_NONEMPTY TRUE)
+      endif()
+    endif()
+  endif()
+
+  if(ESP_SPIRAM_RELOCATE_BSS_OBJECTS_NONEMPTY)
+    set(ESP_SPIRAM_LD_TEMPLATE "${CMAKE_CURRENT_SOURCE_DIR}/../${CONFIG_SOC}/default.ld")
+    set(ESP_SPIRAM_LD_OUTPUT "${PROJECT_BINARY_DIR}/zephyr/linker_esp_spiram.ld")
+    file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/zephyr")
+    execute_process(
```
marekmatej: Included function and use is way faster then spawning the separate process.

soc/espressif/common/CMakeLists.txt
```diff
@@ -89,7 +89,47 @@ elseif(CONFIG_SOC_ESP32_APPCPU OR CONFIG_SOC_ESP32S3_APPCPU)
   set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/../${CONFIG_SOC}/default_appcpu.ld CACHE INTERNAL "")
 elseif(CONFIG_SOC_ESP32C5_LPCORE OR CONFIG_SOC_ESP32C6_LPCORE)
   set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/../${CONFIG_SOC}/default_lpcore.ld CACHE INTERNAL "")
-else()
+elseif(CONFIG_ESP_SPIRAM)
+  # Generate combined linker script only when CONFIG_ESP_SPIRAM and
+  # CONFIG_ESP_SPIRAM_RELOCATE_BSS_OBJECTS is non-empty (same condition as in the linker script).
+  # Otherwise use source default.ld; its #if ESP_SPIRAM_RELOCATE_BSS_ACTIVE is false so normal .bss.
+  set(ESP_SPIRAM_RELOCATE_BSS_CONFIG "${PROJECT_BINARY_DIR}/.config")
+  if(NOT EXISTS "${ESP_SPIRAM_RELOCATE_BSS_CONFIG}")
```
marekmatej: you can use constant location `${CMAKE_BINARY_DIR}` instead of trying

soc/espressif/common/Kconfig.spiram
```diff
@@ -229,6 +229,14 @@ config SPIRAM_SPIWP_SD3_PIN
 
 endif # SOC_SERIES_ESP32
 
+config ESP_SPIRAM_RELOCATE_BSS_OBJECTS
```
marekmatej: I like the idea of generating the object list. This makes the use of `EXCLUDE_FILE` well organized. But it deserves a separate PR and wider approach. Worth considering already existing linker fragments.

soc/espressif/esp32/default.ld
```diff
@@ -787,8 +787,19 @@ SECTIONS
     *(.sbss2.*)
     *(.gnu.linkonce.sb2.*)
     *(.dynbss)
+
+#if defined(CONFIG_ESP_SPIRAM) && ESP_SPIRAM_RELOCATE_BSS_ACTIVE
+    *(EXCLUDE_FILE(
+#include "generated/esp_spiram_relocate_bss_exclude.ld.inc"
```
marekmatej: you can use `#include <esp_spiram_relocate_bss.ld>` where other generated files are located

soc/espressif/esp32s2/default.ld
```diff
@@ -787,8 +787,19 @@ SECTIONS
     *(.sbss2.*)
     *(.gnu.linkonce.sb2.*)
     *(.dynbss)
+
+#if defined(CONFIG_ESP_SPIRAM) && ESP_SPIRAM_RELOCATE_BSS_ACTIVE
+    *(EXCLUDE_FILE(
+#include "generated/esp_spiram_relocate_bss_exclude.ld.inc"
```
marekmatej: dtto

soc/espressif/common/generate_esp_spiram_relocate_bss.cmake
```diff
@@ -0,0 +1,90 @@
...
+# Mode B (combined script in build dir): -DTEMPLATE_LD=<path> -DOUT_LINKER_SCRIPT=<path>
+#   Reads TEMPLATE_LD, inlines the fragment content in place of the two
+#   #include "generated/..." lines, writes to OUT_LINKER_SCRIPT (no files in source tree).
+#
+# Usage: cmake -DBSS_CONFIG_FILE=<path>
+#             [-DOUT_EXCLUDE=<path> -DOUT_EXTRAM=<path>]
+#             [-DTEMPLATE_LD=<path> -DOUT_LINKER_SCRIPT=<path>]
+#             -P generate_esp_spiram_relocate_bss.cmake
+
+if(NOT BSS_CONFIG_FILE)
```
marekmatej: this should be the cmake function to be included where used.

marekmatej (changes requested):
The `ot` related commit LGTM. The CMake script, however, needs more attention. In addition to the latter comments, it can be simplified to gather information (possibly using existing functions such as `import_kconfig(${KCONFIG_NAMESPACE} ${DOTCONFIG})`) and generate a linker snippet to be included from an existing include path.

soc/espressif/common/CMakeLists.txt
```diff
@@ -93,4 +93,36 @@ else()
   set(SOC_LINKER_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/../${CONFIG_SOC}/default.ld CACHE INTERNAL "")
 endif()
 
+# Generate linker snippets for BSS relocation to PSRAM.
+# The snippet files are always created so that #include directives in the
+# linker scripts resolve regardless of the current configuration.
+set(_bss_inc "${PROJECT_BINARY_DIR}/include/generated/esp_spiram_relocate_bss.ld.inc")
```
marekmatej: maybe single object list can be used ?

soc/espressif/common/CMakeLists.txt
```diff
@@ -93,4 +93,36 @@ else()
...
+set(_extram_inc "${PROJECT_BINARY_DIR}/include/generated/esp_spiram_relocate_bss_ext_ram.ld.inc")
+
+if(CONFIG_ESP_SPIRAM AND NOT "${CONFIG_ESP_SPIRAM_RELOCATE_BSS_OBJECTS}" STREQUAL "")
+  string(REPLACE " " ";" _relocate_objects "${CONFIG_ESP_SPIRAM_RELOCATE_BSS_OBJECTS}")
```
marekmatej: at this place I suggest to do somethink like:
```
.
.
# populate working list of bss relocate objects with prefix
import_kconfig(CONFIG_ESP_SPIRAM_BSS_RELOC_ADD_ ${DOTCONFIG} bss_reloc_objs)

foreach(obj ${bss_reloc_objs})
  # add object to list
  .
  .
endforeach()
.
.
```

soc/espressif/common/Kconfig.spiram
```diff
@@ -228,6 +228,18 @@ config SPIRAM_SPIWP_SD3_PIN
 
 endif # SOC_SERIES_ESP32
 
+menu "Relocate BSS to PSRAM"
```
uLipe, line 231: question, is this related to the actual goal of the PR?

wmrsouza: without this change it was not possible make the application fits soc memory - sram overflowed  around 150KB (I have tried a lot of strategies to fix that)
This mechanism is a generic way to relocate the .bss segment of any lib or object to psram without need of future change on linker scripts.
Are you suggesting that I should submit this change as a separate PR instead of only a separate commit of this PR?

samples/net/openthread/border_router/overlay-ot-rcp-host-eth-esp.conf
```diff
@@ -0,0 +1,58 @@
...
+CONFIG_SHELL_TAB=y
+CONFIG_SHELL_TAB_AUTOCOMPLETION=y
+
+# Enable Openthread RCP host interface
+CONFIG_IEEE802154=n
+CONFIG_HDLC_RCP_IF=y
+CONFIG_OPENTHREAD_MANUAL_START=y
+CONFIG_OPENTHREAD_RCP_RESTORATION_MAX_COUNT=10
+
+#Border Router mDNS host name and service base name
```
raffarost: Just a nit, but the 4 next comments are `#Comment` instead of `# Comment`. For readability.

marekmatej (approved):

sylvioalves (approved):

rlubos (approved):

PR #101274, opened 2025-12-18
soc: st: stm32: handle debug power mode in common code for WBA series

mathieuchopstm:
Perform call to `LL_DBGMCU_{Dis,En}ableDBGStandbyMode()` for STM32WBA series in the common code, as done with other series. While at it, also add missing call to `LL_DBGMCU_{Dis,En}ableDBGStopMode()`.

This does change the dependency from `CONFIG_DEBUG` to `CONFIG_STM32_ENABLE_DEBUG_SLEEP_STOP` but the latter is `default y if CONFIG_DEBUG` so this should not bring any behavior change(?)


etienne-lms (approved):
LGTM. 
Minor comment on commit message:  `While at it, also add
missing call to LL_DBGMCU_{Dis,En}ableDBGStopMode().` could be removed, it was implemented in the default `/* all other parts */` `#else` case.

erwango (approved):

PR #91914, opened 2025-06-19
soc : ti: Soc layer power domain and clock nodes init

Kronosblaster:
Power domains are added to the device tree.
The enabled nodes with power domains assigned to them have them turn on in the SOC layer.
Init for enabled clock nodes.
This PR waits on #90216 to be merged.

PR #101280, opened 2025-12-18
TISCI clock and power initialization, changes and bugfixes

natto1784:
Dependency: https://github.com/zephyrproject-rtos/zephyr/pull/100642

Supersedes: https://github.com/zephyrproject-rtos/zephyr/pull/91914

# Changes

### drivers: firmware:
Restricted the TISCI driver to a single instance within the device tree.

### drivers: mbox:
Introduced a vendor-specific polling API for ti_secproxy to support blocking RX message retrieval during PRE_KERNEL.

### drivers: firmware
Modified TISCI to use polling instead of semaphores during PRE_KERNEL stages to allow early mailbox communication.

### drivers: power_domain
Optimized TISCI PD initialization to only target PDs used by "okay" nodes.

### soc/ti/k3/am6x
- Implemented runtime TISCI clock frequency configuration via SYS_INIT and added early UART console initialization.
- Added runtime power domain support using SYS_INIT hooks for both general peripherals and the early UART console.

###  dts
Added TISCI clock controller nodes and peripheral clock/power properties for TI AM64x variants.

### am243x_evm
Enabled TISCI, clock, and power management Kconfig options for R5FSS0_0 and cleaned up peripheral status in DT.


natto1784:
not stale

PR #101288, opened 2025-12-18
include: display: add doxygen docs for device-specific Display API extensions

kartben:
Properly document ssd16xx display API extension so that it shows up nicely in the online docs and is easier for folks to discover

https://builds.zephyrproject.io/zephyr/pr/101288/docs/doxygen/html/group__display__interface__ext.html
https://builds.zephyrproject.io/zephyr/pr/101288/api-coverage/zephyr/display/index.html

### DIFF
--- a/include/zephyr/drivers/display.h
+++ b/include/zephyr/drivers/display.h
@@ -20,6 +20,11 @@
  * @version 0.8.0
  * @ingroup io_interfaces
  * @{
+ *
+ * @defgroup display_interface_ext Device-specific Display API extensions
+ * @{
+ * @}
+ *
  */
 
 #include <zephyr/device.h>
--- a/include/zephyr/display/ssd16xx.h
+++ b/include/zephyr/display/ssd16xx.h
@@ -4,13 +4,28 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/**
+ * @file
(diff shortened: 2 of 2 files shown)


include/zephyr/drivers/display.h
```diff
@@ -530,4 +530,11 @@ static inline int display_set_orientation(const struct device *dev,
  * @}
  */
 
+/**
+ * @defgroup display_interface_ext Device-specific Display API extensions
```
VynDragon: Could you explain this one? There is nothing there, would other headers queue after or is nothing there yet?

kartben: that's the "hook" group for other extended headers (like e.g in the second commit) to mount themselves "under"

https://github.com/zephyrproject-rtos/zephyr/pull/101288/changes#diff-a12e34eecdffa10d68744b15f7b9c6069bc62b37fac2339b9227f54e19676e51R19

actually let me do it in a slightly less confusing way

kartben: @VynDragon hopefully slightly less confusing now - let me know!

VynDragon: Well I dont understand doxygen stuffs, so i guess it must be right.

VynDragon (approved):

JarmouniA (approved):
Thanks!

PR #101290, opened 2025-12-18
drivers: sensor: adi: adxl345: Unit tests for conversion function

stemschmidt:
This PR adds two unit tests with the required mocks to test the conversion function in the adxl345 for right justified data, in 2G and 4G mode.

To run this test invoke twister from zephyr root:
"twister -T zephyr/drivers/sensor/adi/adxl345/"

### DIFF
--- a/drivers/sensor/adi/adxl345/adxl345_features.h
+++ b/drivers/sensor/adi/adxl345/adxl345_features.h
@@ -1,0 +1,17 @@
+/*
+ * SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_DRIVERS_SENSOR_ADX345_ADX345_FEATURES_H_
+#define ZEPHYR_DRIVERS_SENSOR_ADX345_ADX345_FEATURES_H_
+
+#define ADXL345_RANGE_2G  0x0
+#define ADXL345_RANGE_4G  0x1
+#define ADXL345_RANGE_8G  0x2
+#define ADXL345_RANGE_16G 0x3
+
+#define ADXL345_COMPLEMENT 0xfc00
+
+#endif /* ZEPHYR_DRIVERS_SENSOR_ADX345_ADX345_FEATURES_H_ */
--- a/tests/drivers/sensor/adxl345/CMakeLists.txt
+++ b/tests/drivers/sensor/adxl345/CMakeLists.txt
@@ -1,0 +1,22 @@
+# Set minimum CMake version
+cmake_minimum_required(VERSION 3.20.0)
+
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+find_package(Zephyr COMPONENTS unittest REQUIRED HINTS $ENV{ZEPHYR_BASE})
+
+project(test_adxl345_accel_convert)
+
+target_sources(testbinary
+    PRIVATE
+        src/main.c
+        $ENV{ZEPHYR_BASE}/drivers/sensor/adi/adxl345/adxl345_convert.c
+)
+
+# make sure we include prepared sensor.h and have access to adxl345_features.h
+target_include_directories(testbinary
+    BEFORE
+    PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/include
+        $ENV{ZEPHYR_BASE}/drivers/sensor/adi/adxl345
+)
--- a/tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
+++ b/tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
@@ -1,0 +1,25 @@
+/*
+ * SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef ZEPHYR_INCLUDE_DRIVERS_SENSOR_H_STRIPPED_
+#define ZEPHYR_INCLUDE_DRIVERS_SENSOR_H_STRIPPED_
+
+#include <stdint.h>
+#include <zephyr/sys/util_macro.h>
+
+#ifndef SENSOR_G
+#define SENSOR_G 9806650LL
+#endif
+
+/* Define struct sensor_value here to avoid including include/zepyhr/drivers/sensor.h */
+struct sensor_value {
+	/** Integer part of the value. */
+	int32_t val1;
+	/** Fractional part of the value (in one-millionth parts). */
+	int32_t val2;
+};
+
+#endif /* ZEPHYR_INCLUDE_DRIVERS_SENSOR_H_STRIPPED_ */
--- a/tests/drivers/sensor/adxl345/src/main.c
+++ b/tests/drivers/sensor/adxl345/src/main.c
@@ -1,0 +1,61 @@
+/*
+ * SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/ztest.h>
+
+/*
+ * Include the stripped sensor.h from the include directory of the test folder!
+ * The directory /zephyr/tests/drivers/sensor/adxl345 is injected in CMakeLists.txt
+ */
+#include <zephyr/drivers/sensor.h>
+
+#include "adxl345_features.h"
+
+extern void adxl345_accel_convert(struct sensor_value *val, int16_t sample, uint8_t selected_range);
+
+ZTEST(adxl345_accel_convert, test_convert_10bit_right_justified_2g_mode)
+{
+	struct sensor_value data[3] = {0};
+	int16_t samples[3] = {0, 100, -100};
+
+	/* Convert and check x-axis. */
+	adxl345_accel_convert(&data[0], samples[0], ADXL345_RANGE_2G);
+	zassert_equal(data[0].val1, 0, "expected 0, was %d", data[0].val1);
+	zassert_equal(data[0].val2, 0, "expected 0, was %d", data[0].val2);
+
+	/* Convert and check y-axis. */
+	adxl345_accel_convert(&data[1], samples[1], ADXL345_RANGE_2G);
+	zassert_equal(data[1].val1, 3, "expected 3, was %d", data[1].val1);
+	zassert_equal(data[1].val2, 830700, "expected 830700, was %d", data[1].val2);
+
+	/* Convert and check z-axis. */
+	adxl345_accel_convert(&data[2], samples[2], ADXL345_RANGE_2G);
+	zassert_equal(data[2].val1, -3, "expected -3, was %d", data[2].val1);
+	zassert_equal(data[2].val2, -830700, "expected -830700, was %d", data[2].val2);
+}
+
+ZTEST(adxl345_accel_convert, test_convert_10bit_right_justified_4g_mode)
+{
+	struct sensor_value data[3] = {0};
+	int16_t samples[3] = {0, 100, -100};
+
+	/* Convert and check x-axis. */
+	adxl345_accel_convert(&data[0], samples[0], ADXL345_RANGE_4G);
+	zassert_equal(data[0].val1, 0, "expected 0, was %d", data[0].val1);
+	zassert_equal(data[0].val2, 0, "expected 0, was %d", data[0].val2);
+
+	/* Convert and check y-axis. */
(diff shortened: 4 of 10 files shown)


stemschmidt:
I have considered it, but my intention is to have fast tests that test the functions registered in the device api. And test those in isolation. I used the conversion function as a start to demonstrate the approach, the PR is far from complete. 

I detected an issue in the conversion for the Fetch and Get API that was caused by changes in the Read and Decode API. It gets more and more complicated when functionality is added to drivers to ensure that all other parts of the driver still perform as expected. A fast unit test suite can help with that.

I have mocking the bus in mind for unit testing the init function: Depending on the Kconfigs I would check for the expected initialization values being "written" to the device. But still then I would like to avoid device tree compilation and building as I am again only interested in the invocation of the single "init" function.

What would you consider the advantage of your approach with mocking the bus?

stemschmidt:
Thanks for pointing me to the drivers, I was not aware of these split approach. I will modify my PR accordingly.

PerMac (changes requested):
all tests should be located in zephyr/tests directory

include/zephyr/drivers/sensor.h
```diff
@@ -35,30 +35,12 @@
...
- *     -1.0: val1 = -1, val2 =  0
- *     -1.5: val1 = -1, val2 = -500000
- */
-struct sensor_value {
-	/** Integer part of the value. */
-	int32_t val1;
-	/** Fractional part of the value (in one-millionth parts). */
-	int32_t val2;
-};
-
```
bjarki-andreasen, line 61: I don't believe we should start splitting headers because someone wants to only include a tiny part of it, this is unscalable and cluttery. Redefine the type where you need it, or even better, consider why you can't include the header and maybe reconsider the approach.

bjarki-andreasen (changes requested):
The location is fine, the approach is not as it requires changing zephyr APIs, just build a normal zephyr application...

stemschmidt:
Hi Bjarki, sorry, I do not get your point: Where does it require a change in a Zephyr API?

include/zephyr/drivers/sensor.h
MaureenHelm: I agree. And while this has been addressed at the `sensor.h` level, it's still present in the adxl345 driver. I don't think having individual `drivers/sensor/adi/adxl345/adxl345_convert.h` will scale

stemschmidt: Hi @MaureenHelm, what would be your prefered way to test the conversion functions? So far I have only included the function for the fetch-and-get method, my intention was to also include the conversion functions for the read-and-decode approach.

stemschmidt: Hi @teburd, @MaureenHelm, @bjarki-andreasen, could you please let me know how to proceed? I am happy to continue, but it would also be fine to close this PR. Although I think it would be great to add automatic bug detection to the drivers code base.

vai-helin: Could you just compile the conversion functions to a static library and test that as normal library code? The problem would then be mostly CMake ergonomics* and you wouldn't need the brittle include paths in the test code as those would come from the library target.

*) You'd need a nice way to define the conversion library and some way of referring to it in the tests. Most likely you'd want to use something like `zephyr_link_driver_lib(my_test_target driver_name)` to hide the build system details. This differs from how the code and tests are normally built, so you might need to think this a bit more thoroughly.

stemschmidt: I can put all conversion function(s) into a c file and add this file to 
a) the Zephyr driver CMakeLists.txt
b) the unit test CMakeLists.txt
With this approach the conversion function will be in the global namespace which might be ok as the name is unique (adxl345_accel_convert).

vai-helin: I was thinking that you would add a separate library for the driver utilities. In plain CMake it would look something like this:

```
add_library(zephyr_driver_adxl345_accel_conv STATIC adxl345_accel_conv.c)
target_include_directories(zephyr_driver_adxl345_accel_conv PUBLIC ${CMAKE_CURRENT_LIST_DIR})
target_link_libraries(zephyr_driver_lib PRIVATE zephyr_driver_adxl345_accel_conv)

# in unit test
target_link_libraries(mytest PRIVATE zephyr_driver_adxl345_accel_conv)
```
This of course requires that you have some reliable way of getting the name of the conversion library target in the unit tests -> create a CMake interface that hides the library naming conventions and other such details. I have a feeling Kconfig should also be consulted at some point.

Point being that (IMO) to do this correctly you'll need a new build system feature first which you can then use to add tests to driver utility functions, such as the ones for adxl345. For example, how would you test the other adxl-chips that seem to have very similar conversion functions (and structure). Would this approach work for other sensor drivers with different designs? How about testing just the decoder interface?

stemschmidt: I have moved the conversion function into a c file and added it to the CMakeLists.txt for the Zephyr driver and the unit test.

The unit tests will be created and build for each driver separately. They can be selected individually via the twister command line or run as part of a complete unit test run.

tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
vai-helin, line 17: Why do you want to avoid this?

tests/drivers/sensor/adxl345/CMakeLists.txt
vai-helin, line 17: This is better than the relative include path in the test, but ideally you should just refer to a target name `adxl345_convert` (or something like that) and the include paths would come from interface / public include directories.

I'm not really sure if the existing `<ZEPHYR_BASE>/tests/` organization is good for these kinds of unit tests which aim to stress the internals of particular driver. More logical / better structured organization might be to have a test directory in e.g. `<ZEPHYR_BASE>/drivers/sensor/adi/adxl345/tests`. This again is completely new territory and I think it warrants a broader discussion about different types of tests written for drivers.

tests/drivers/sensor/adxl345/src/main.c
vai-helin, line 13: This is very misleading with the current compile flags. Are you sure which file gets included here?

drivers/sensor/adi/adxl345/adxl345_features.h
vai-helin, line 1: How are these features distinct from the ones in `adxl345.h`? Are you trying to separate the defines from the main header which includes Zephyr specifics or what's the purpose of this file?

tests/drivers/sensor/adxl345/src/main.c
stemschmidt: It has to be the same name as it is also included in adxl345_convert.c.

In CmakeLists.txt in the unit test directory I ensure that sensor.h from the unit test directory is seen first and used for unit testing

tests/drivers/sensor/adxl345/CMakeLists.txt
stemschmidt: I had to move the unit tests from the drivers/sensor/adi/adxl345 directory to tests, see a previous comment in the discussion.

I would also prefer to have the unit tests closer to the code, but having them under tests/ seems to be the Zephyr way.

tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
stemschmidt: Because it pulls in all Zephyr and devicetree related structures and defines. And I only want to test a simple C function

vai-helin: But does it prevent the test from compiling or what? I don't really understand why including those prevents you to write tests for the conversion functions.

tests/drivers/sensor/adxl345/CMakeLists.txt
vai-helin: Yeah, I sort of assumed that tests/ would be the preferred place but didn't notice you had originally considered alternatives.

BTW you can get a link to a specific comment behind the three dots.

tests/drivers/sensor/adxl345/src/main.c
vai-helin: Yes it has to be same, but it makes the `BEFORE` in `target_include_directories()` pretty critical and you have to _really_ read the whole test project carefully to spot that (not just the test file). 

I would urge you to reconsider the design and if this is truly required, justify clearly why it _has_ to be this way. I don't think "I want to limit what gets included" is good enough reason unless there's a technical rationale behind it.

tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
stemschmidt: Yes, it would not compile.

The unit tests are build for the unit_testing platform (https://docs.zephyrproject.org/latest/develop/test/ztest.html#quick-start-unit-testing). The resulting binary is an executable for your host.

I include sensor.h only for SENSOR_G and struct sensor_value.

In the original Zephyr headers there is no separation of those hardware independent defines and structures and the Zephyr related hardware elements, so if I would not replace it for the tests I would get all the magic ingredients of Zephyr which I do not need to run my simple test and which will not compile because I am not running the devicetree configuration.

My simple test is: Convert a given value to an expected result. That is all. No driver invocation, no devicetree dependency.

tests/drivers/sensor/adxl345/src/main.c
stemschmidt: Mmmh, I agree that I could add some hint to it. But the whole unit test project is made up of 5 files, with the standard unit test layout. And mocking or injecting prepared files is what I do all the time for unit testing. So I am pretty happy with the approach.

drivers/sensor/adi/adxl345/adxl345_features.h
stemschmidt: Correct, I am splitting device specific features from the general Zephyr specifics

tests/drivers/sensor/adxl345/include/zephyr/drivers/sensor.h
vai-helin: I tested the PR and yup, it's the missing some syscalls and `devicetree_generated.h` that are causing problems. I tried tinkering with `unittest.cmake` and compiler flags (gcc-15 has stricter warnings than gcc-12 in the sdk) but `devicetree_generated.h` is still problematic. You'd probably need to add mocks to the generated `kobject.h` as well. Gets messy quite fast.

The way the code's currently written makes it hard to decouple from the parts that cannot be compiled for unit tests. Refactoring the conversion code (severing dependencies to Zephyr types and constants) might help, but is it truly worth the effort? The way I see it, there's an infrastructure issue (enabling developers to write convenient unit test for the host architecture) and the specific issue of writing tests for these conversion functions. Until the bigger issue is somehow resolved I'd personally write these by mocking the SPI interface and testing the whole driver. Tests can always be improved later and some tests are better than no tests. :)

stemschmidt: My motivation to write these unit tests was a regression bug in the adxl345 driver in the official release of Zephyr. So I thought it is pretty easy to prevent this kind of bugs sneak in by setting up a unit test for the conversion. The driver consists of two parts: The mechanics to get the value out of the device (via I2C or SPI) and the conversion of this raw value to something with a unit (m/s^2). As the problem was in the conversion part I wrote unit tests for this part. There might be a range of options how to approach this. I am coming from TDD where the tests are focused, short and fast to execute. And "unit" for means the minimal executable unit.

I have been through the mocking hell trying to mock all the devicetree objects in a project and it costed a lot of time and money.

You can do mocking of the SPI interface, or the I2C interface, or both? And then do a test. But that requires to build the whole driver with devicetree and what not. If you build it with SPI, are you sure the I2C version is correct? If there is an error in the unit test, where should you start searching? Is it the mechanics to get the value? On SPI? On I2C?

I fully understand that the current approach with copying defines and structs from the "real" header is not good, but that is what I came up with to unit test something that has been developed without having unit testing in mind. It is not optimal, but it works and the risk is handable: If it fails (e.g. because the structure has changed in the "real" header) it is the unit test that fails, not the real driver code.

I hate "Tests can always be improved later...". With this mindset they never will.

vai-helin: Regarding SPI/I2C: your goal is to test the conversions, so presumably either would work and get you some working tests.

> If it fails (e.g. because the structure has changed in the "real" header) it is the unit test that fails, not the real driver code.

That still requires someone to look into the failing tests and increases maintenance burden.

I understand your point in wanting small, simple unit tests, but IMO you should work on building the infrastructure first. This might take a while as headers might require modifications which are probably handled in a working group for the relevant subsystem. Sorry if improving things in small increments isn't your cup of tea.

PR #101294, opened 2025-12-18
Add DNS Additional record support

Ayush1325:
- As described in Section 12 of RFC 6763 [0], DNS-SD responses can contain SRV, TXT and A/AAAA under additional records. This is done even in the zephyr DNS service advertisement. So these additional records should be passed to the user for handling.
- If required, we can add an enum to `dns_addrinfo` struct to signify if the record is an answer, an authority or an additional record.

[0]: https://datatracker.ietf.org/doc/html/rfc6763.html#section-12

### DIFF
--- a/subsys/net/lib/dns/resolve.c
+++ b/subsys/net/lib/dns/resolve.c
@@ -1089,6 +1089,137 @@ quit:
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


jukkar (approved):
LGTM

rlubos (approved):

PR #20665, opened 2019-11-13
kernel/sched: Fix edge case in MetaIRQ preemption of cooperative threads

andyross:
When a MetaIRQ preempts a cooperative thread, that thread would be
added back to the generic run queue.  When the MetaIRQ is done, the
highest priority thread will be selected to run, which may obviously
be a cooperative thread of a higher priority than the one that was
preempted.

But that's wrong, because the original thread was promised that it
would NOT be preempted until it reached a scheduling point on its own
(that's the whole point of a cooperative thread, of course).

We need to track the thread that got preempted (one per CPU) and
return to it instead of whatever else the scheduler might have found.

Fixes #20255

Signed-off-by: Andy Ross <andrew.j.ross@intel.com>

PR #101297, opened 2025-12-18
kernel: Re-instate metaIRQ z_is_thread_ready() check

### LINKED ISSUE #101296: kernel: Recent metaIRQ change broke bluetooth sample on Silabs
### Describe the bug

From a discussion in #99349.

> Heads up that commit https://github.com/zephyrproject-rtos/zephyr/commit/5dd36854fde7e6497124b791243c2e8fbe747210 from this PR seems to break silabs bluetooth HCI stack, which uses a Meta-IRQ priority:
> https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/bluetooth/hci/hci_silabs_efr32.c#L250
> 
> Reverting just this commit fixes the issue, so is there a path we missed here?

### Regression

- [ ] This is a regression.

### Steps to reproduce

west build -p -b xg24_dk2601b samples/bluetooth/peripheral_hr

### Relevant log output

```shell
@asmellby reported ...

I had a look at this, and I'm observing the following:

The device crashes in sys_dlist_dequeue due to a null pointer dereference

The list node in question is bt_tx_processor_workq
Both the prev and next pointers are NULL, so the dereferencing in prev->next and next->prev cause the crash
The call stack is
sys_list_dequeue
z_priq_simple_yield
z_impl_k_yield
work_queue_main
Starting from the changed logic in next_up before the crash:

thread is 0 (idle thread) and mirqp is bt_tx_processor_workq
the workqueue thread is not ready
So it seems like the new logic chooses to return to the workqueue despite it not being ready, while the old logic goes to the idle thread in this case. I don't know the kernel well to understand the implications of selecting a not-ready thread and if that directly causes the issue when that thread later tries to yield.

I reproduced this by doing the following:

Build a BT sample for a Silicon Labs board, I used west build -p -b xg24_dk2601b samples/bluetooth/peripheral_hr
Flash the sample
Use a mobile app such as Si Connect or nRF Connect to connect to the device
The connection will immediately drop due to the crash
```

### Impact

### Environment

_No response_

### Additional Context

The problem has been traced to the following lines in the identified commit ...

```
-		if (z_is_thread_ready(mirqp)) {
-			thread = mirqp;
-		} else {
-			_current_cpu->metairq_preempted = NULL;
-		}
+		thread = mirqp;
```

Recommendation: Revert this part of the commit.

peter-mitsis:
Re-instate a z_is_thread_ready() check on the preempted metaIRQ thread before selecting it as the preferred next thread to schedule.

Fixes #101296

### DIFF
--- a/kernel/sched.c
+++ b/kernel/sched.c
@@ -193,7 +193,11 @@ static ALWAYS_INLINE struct k_thread *next_up(void)
 	struct k_thread *mirqp = _current_cpu->metairq_preempted;
 
 	if (mirqp != NULL && (thread == NULL || !thread_is_metairq(thread))) {
-		thread = mirqp;
+		if (z_is_thread_ready(mirqp)) {
+			thread = mirqp;
+		} else {
+			_current_cpu->metairq_preempted = NULL;
+		}
 	}
 #endif /* CONFIG_NUM_METAIRQ_PRIORITIES > 0 */
 


asmellby (approved):
Verified on `xg24_dk2601b` and `sltb010a`, Bluetooth no longer crashes with this fix.

JarmouniA:
Can we link to @andyross's [comment](https://github.com/zephyrproject-rtos/zephyr/pull/20665#issuecomment-553997571) in the commit message, or quote the relevant part of it verbatim? Seems very helpful in understanding the need for the check.

henrikbrixandersen (approved):

JarmouniA (approved):

jhedberg (approved):

TaiJuWu (approved):

PR #101298, opened 2025-12-18
drivers: i2c: bflb: Fix I2C again

VynDragon:
Found more corner cases where GCC seems to reliably fail on this function for unknown reasons, would need to add even more volatiles.
Take nuclear option to handling compiler failing at compiling

ASM output was looked at directly and it was determined to be wrong and doing reordering that is breaking the driver.


josuah (approved):
That's also fewer `volatile` scattered across the driver this way.

teburd (approved):

PR #101299, opened 2025-12-18
include: Bluetooth: Host: add doxygen docs for GAP timer macros

kartben:
This adds proper doxygen documentation for the GAP timers macros.

https://builds.zephyrproject.io/zephyr/pr/101299/docs/doxygen/html/gap_8h.html#define-members-1

### DIFF
--- a/include/zephyr/bluetooth/gap.h
+++ b/include/zephyr/bluetooth/gap.h
@@ -36,30 +36,52 @@ extern "C" {
  */
 
 /**
- * @name Defined GAP timers
+ * @name Defined GAP timers.
+ * As per Bluetooth Core Specification 6.2, Vol 3, Part C, Appendix A.
  * @{
  */
-#define BT_GAP_SCAN_FAST_INTERVAL_MIN           0x0030  /* 30 ms    */
-#define BT_GAP_SCAN_FAST_INTERVAL               0x0060  /* 60 ms    */
-#define BT_GAP_SCAN_FAST_WINDOW                 0x0030  /* 30 ms    */
-#define BT_GAP_SCAN_SLOW_INTERVAL_1             0x0800  /* 1.28 s   */
-#define BT_GAP_SCAN_SLOW_WINDOW_1               0x0012  /* 11.25 ms */
-#define BT_GAP_SCAN_SLOW_INTERVAL_2             0x1000  /* 2.56 s   */
-#define BT_GAP_SCAN_SLOW_WINDOW_2               0x0012  /* 11.25 ms */
-#define BT_GAP_ADV_FAST_INT_MIN_1               0x0030  /* 30 ms    */
-#define BT_GAP_ADV_FAST_INT_MAX_1               0x0060  /* 60 ms    */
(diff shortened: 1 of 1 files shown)


Thalley (dismissed):

include/zephyr/bluetooth/gap.h
```diff
@@ -39,27 +39,48 @@ extern "C" {
...
+/** Recommended maximum periodic advertising interval for fast advertising (mode 2). 150 ms. */
+#define BT_GAP_PER_ADV_FAST_INT_MAX_2           0x0078
+/** Recommended minimum periodic advertising interval for slow advertising. 1 s. */
+#define BT_GAP_PER_ADV_SLOW_INT_MIN             0x0320
+/** Recommended maximum periodic advertising interval for slow advertising. 1.2 s. */
+#define BT_GAP_PER_ADV_SLOW_INT_MAX             0x03C0
+/** Recommended minimum initial connection interval. 30 ms. */
+#define BT_GAP_INIT_CONN_INT_MIN                0x0018
+/** Recommended maximum initial connection interval. 50 ms. */
+#define BT_GAP_INIT_CONN_INT_MAX                0x0028
```
cvinayak, line 84: If `Recommended`  then please refer to Bluetooth Specification sections and the define names as in the specification.

cvinayak (changes requested):

include/zephyr/bluetooth/gap.h
jhedberg: FWIW, these can be found here: https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-62/out/en/host/generic-access-profile.html#UUID-847b49e4-1999-9402-fa2d-74231c86a3e9

Thalley: Probably a good idea to only use a section reference rather than a link; BT SIG has a tendency to move things around without proper forwards, so it's fairly common that their URL dies in my experience

kartben: please check updated commit - thanks for the feedback :)
let me know if it should say Bluetooth 6.0? ... 6.2?  ... 5.4 ? :)

jhedberg: I'd reference the latest one, so 6.2 (that's what my link was for also)

kartben: > I'd reference the latest one, so 6.2 (that's what my link was for also)

sure thing! fwiw I had prepared the patch before seeing you guys had commented further in this thread :)

Thalley: > I'd reference the latest one, so 6.2 (that's what my link was for also)

I'd do the same, although I do recall the the sections used by the Bluetooth Core Spec are never removed, so arguably it may make sense to provide the section from the version of the spec that added them, e.g. "Bluetooth Core version 4.0 or newer", but it's also completely fair to use the newest version from today's date

Thalley (approved):

cvinayak (approved):

jhedberg (approved):

PR #101304, opened 2025-12-18
drivers: pwm: pwm_nrfx: fix suspend to avoid undefined behavior

bjarki-andreasen:
When suspending the pwm device, the entire data structure was erased. This includes the nrfx instance data, which should not be cleared. Update driver to not clear the nrfx instance data.

Additionally remove a redundant call to pinctrl as pm_device_driver_init calls this automatically.


anangl (approved):

bjarki-andreasen:
duplicate of https://github.com/zephyrproject-rtos/zephyr/pull/100309

PR #101306, opened 2025-12-18
drivers: display: Improve SSD1306 error handling and reporting

VynDragon:
returning -1 doesn't constitute proper error reporting.

### DIFF
--- a/drivers/display/ssd1306.c
+++ b/drivers/display/ssd1306.c
@@ -29,6 +29,8 @@ LOG_MODULE_REGISTER(ssd1306, CONFIG_DISPLAY_LOG_LEVEL);
 #define SSD1306_ADDRESSING_MODE		(SSD1306_SET_MEM_ADDRESSING_HORIZONTAL)
 #endif
 
+#define SSD1306_PPB_SHIFT	3
+
 union ssd1306_bus {
 	struct i2c_dt_spec i2c;
 	struct spi_dt_spec spi;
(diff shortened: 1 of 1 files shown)


drivers/display/ssd1306.c
```diff
@@ -402,7 +409,7 @@ static int ssd1306_set_pixel_format(const struct device *dev,
 	}
 
 	ret = ssd1306_write_bus(dev, &cmd, 1, true);
-	if (ret) {
+	if (ret < 0) {
 		return ret;
```
mathieuchopstm, line 455: Maybe `LOG_ERR("Failed to set pixel format")`?

drivers/display/ssd1306.c
```diff
@@ -296,28 +298,33 @@ static int ssd1306_write_sh1106(const struct device *dev, const uint16_t x, cons
...
+			LOG_ERR("Failed to write pixel data");
+			return ret;
 		}
 
 		buf_ptr = buf_ptr + desc->width;
 		if (buf_ptr > ((uint8_t *)buf + buf_len)) {
 			LOG_ERR("Exceeded buffer length");
-			return -1;
+			return -EINVAL;
 		}
```
mathieuchopstm: Can't this be determined in advance (before the loop)?
```c
if (((desc->height / 8) * desc->width) > buf_len) {
  LOG_ERR("buffer too small");
  return -EINVAL;
}
```
(this is a bit outside PR scope though)

mathieuchopstm (commented):
LGTM otherwise

drivers/display/ssd1306.c
VynDragon: Looking at this I dont think it's even a possible condition? Reengineering things a little bit anyway now you've prompted me for it lol ( found ch1116 has a 'breathing' function and my 'ssd1306' (sh1106) 1.3 is actually ch1116 and not sh1106).

drivers/display/ssd1306.c
VynDragon: done

drivers/display/ssd1306.c
VynDragon: done

drivers/display/ssd1306.c
```diff
@@ -438,43 +451,57 @@ static int ssd1306_init_device(const struct device *dev)
...
+	if (ret < 0) {
+		LOG_ERR("Failed to suspend: %d", ret);
+		return ret;
 	}
 
-	if (ssd1306_set_timing_setting(dev)) {
-		return -EIO;
+	ret = ssd1306_set_timing_setting(dev);
+	if (ret < 0) {
+		LOG_ERR("Failed to set timings: %d", ret);
```
danieldegrasse, line 499: Suggestion here: we could save some flash space by logging like this:
```
LOG_ERR("Failed to set %s: %d", "timings", ret);
LOG_ERR("Failed to set %s: %d", "hardware configuration", ret);
```

VynDragon: By deduplication?

JarmouniA (approved):

drivers/display/ssd1306.c
```diff
@@ -29,6 +29,8 @@ LOG_MODULE_REGISTER(ssd1306, CONFIG_DISPLAY_LOG_LEVEL);
 #define SSD1306_ADDRESSING_MODE		(SSD1306_SET_MEM_ADDRESSING_HORIZONTAL)
 #endif
 
+#define SSD1306_PPB_SHIFT	3
```
mathieuchopstm, line 32: FWIW, compilers will automagically optimize `a / 8` into `a >> 3` so the direct and more explicit approach could be used instead.

mathieuchopstm (approved):
Tiny nit

PR #101309, opened 2025-12-19
cpu_freq: pressure: Introduce pressure policy

seankyer:
Add a pressure policy to cpu_freq subsystem, which calculates 'load' by accumulating the priorities of threads waiting to be executed in the ready queue. As more threads are set to 'ready', the system pressure increases and the cpu_freq subsystem may then choose a more aggressive P-state to run in.

System pressure is calculated as:

$$
P_{sys} = \frac{\sum_{t \in R} (P_{min} - prio_t + 1)}
                {\sum_{t \in T} (P_{min} - prio_t + 1)} \times 100
$$

Where:

- \(R\) is the set of runnable threads
- \(T\) is the set of all threads considered for pressure
- \(prio_t\) is the numeric priority of thread \(t\)
- \(P_{min}\) is the configured minimum priority used for weighting (but this is really the highest numerical value)

The +1 is to make sure we don't divide by 0.

Essentially, sum up the priority of all ready threads and divide them by the sum of the priorities of all possible threads to determine the normalized pressure that the queue is under. In code this looks like:
```c
static void thread_eval_cb(const struct k_thread *thread, void *user_data)
{
	struct pressure_stats *pressure = user_data;

	LOG_DBG("Evaluating thread: %p with prio: %d status: %d", thread, thread->base.prio, thread->base.thread_state);

	int weight = CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO - thread->base.prio + 1;
	if (weight < 0) {
		return;
	}

	pressure->max_pressure += weight;

	if (is_runnable(thread)) {
		pressure->pressure_acum += weight;
	}
}

static int get_normalized_sys_pressure(void) {
	struct pressure_stats sys_pressure = {
		.max_pressure = 0,
		.pressure_acum = 0
	};

	k_thread_foreach(thread_eval_cb, &sys_pressure);

	int normalized_pressure = (sys_pressure.pressure_acum * 100) / sys_pressure.max_pressure;

	LOG_DBG("System pressure is: %d%% (raw: %d / max: %d)", normalized_pressure, sys_pressure.pressure_acum, sys_pressure.max_pressure);

	return normalized_pressure;
}
```

This was developed on the frdm_mxcn236 which recently had support added for this subsystem.

Board logs from the added sample:
```
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: thread_eval_cb: Evaluating thread: 0x30000188 with prio: 1 status: 16
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: thread_eval_cb: Evaluating thread: 0x30000238 with prio: 5 status: 128
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: thread_eval_cb: Evaluating thread: 0x300002e8 with prio: 10 status: 128
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: thread_eval_cb: Evaluating thread: 0x300003a8 with prio: 15 status: 0
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: thread_eval_cb: Evaluating thread: 0x30000458 with prio: 0 status: 4
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: get_normalized_sys_pressure: System pressure is: 38% (raw: 87 / max: 224)
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: cpu_freq_policy_select_pstate: CPU0 Pressure: 38%
[00:00:02.500,000] <dbg> cpu_freq_policy_pressure: cpu_freq_policy_select_pstate: Pressure Policy: Selected P-state 1 with load_threshold=30% 
```

This PR also includes some minor updates to things I noticed that could be improved while developing this.


subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+static void thread_eval_cb(const struct k_thread *thread, void *user_data)
+{
+	struct pressure_stats *pressure = user_data;
+
+	LOG_DBG("Evaluating thread: %p with prio: %d status: %d", thread, thread->base.prio,
+		thread->base.thread_state);
+
+	int weight = CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO - thread->base.prio + 1;
```
peter-mitsis, line 70: I suspect that it would be good to limit the lowest pressure priority to the idle thread's priority. I've not done the math, but wouldn't allowing a value that is significantly higher (say 1,000,000)  introduce a significant bias towards higher pressures.

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,180 @@
...
+#endif /* CONFIG_SMP && (CONFIG_MP_MAX_NUM_CPUS > 1) && !CONFIG_CPU_FREQ_PER_CPU_SCALING */
+
+struct pressure_stats {
+	int pressure_acum;
+	int max_pressure;
+};
+
+static bool is_runnable(const struct k_thread *thread)
+{
+	return (bool)(thread->base.thread_state == _THREAD_QUEUED);
```
peter-mitsis: Depending upon whether the SMP or non-SMP scheduling is selected, this is going to lead to different results. In SMP, a CPU's current thread is not in the ready queue and thus not marked (even if MP_MAX_NUM_CPUS=1). In non-SMP, the thread the CPU's current thread is on the ready queue and thus marked.

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,180 @@
...
+	if (is_runnable(thread)) {
+		pressure->pressure_acum += weight;
+	}
+}
+
+static int get_normalized_sys_pressure(void)
+{
+	struct pressure_stats sys_pressure = {.max_pressure = 0, .pressure_acum = 0};
+
+	k_thread_foreach(thread_eval_cb, &sys_pressure);
```
peter-mitsis, line 90: Is this feature intended for use with CONFIG_SCHED_CPU_MASK_PIN_ONLY=y? (Each CPU has its own readyQ.) If so, that would be a potential use case for per-CPU P-states and the use of k_thread_foreach_filter_by_cpu() ?

peter-mitsis (commented):
I'm on holidays, so I won't be able to give a full proper review until I get back into the office (Jan 5).

This is a very interesting idea! :) A couple of things did jump out at me that I thought were worth asking about.

subsys/cpu_freq/Kconfig
```diff
@@ -51,6 +51,22 @@ config CPU_FREQ_POLICY_ON_DEMAND
 	bool "On-demand Policy"
 	select CPU_LOAD_METRIC
 
+config CPU_FREQ_POLICY_PRESSURE
+	bool "Pressure based Policy"
+	select TRACING
```
ttmut, line 56: Does this feature require tracing to be enabled? Will that introduce any overhead?

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: I didn't think of that, I think setting it to the idle thread priority makes sense, I'll update that

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: Ah I see.. tbh I just copied what you had done from SMP in the on demand policy naively thinking it would work here too. Maybe I should remove this SMP stuff for now and set kconfig to !SMP? Unless you think there's an easy way around it.

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: I think it could. So you suggest I use that macro instead and then this will "just work" on a per CPU basis?

subsys/cpu_freq/Kconfig
seankyer: I have to double check if it's required but I think when prototyping, tracing must be selected in order to inspect the readyQ. This does introduce overhead of course. I could make a note about this in the docs.. at the very least.

subsys/cpu_freq/cpu_freq.c
```diff
@@ -45,6 +49,14 @@ static void cpu_freq_next_pstate(void)
 		return;
 	}
 
+	if (pstate_next == pstate_last) {
```
peter-mitsis, line 55: I think a little extra tweaking is required for the CONFIG_CPU_FREQ_PER_CPU_SCALING=y case. It could be disabled for that case by wrapping the relevant code chunks in an #ifndef CONFIG_CPU_FREQ_PER_CPU_SCALING/#endif block.

Alternatively, the #if/#endif block could be extended so that there is a #else case for when it is enabled. That could define a pstate_last[] array -- one item per CPU, and tweak the code as necessary.

subsys/cpu_freq/policies/pressure/pressure.c
peter-mitsis: Something kind of like the following should work.

```
bool rv = false;

if (thread->base.thread_state == _THREAD_QUEUED) {
    return true;
}

#ifdef CONFIG_SMP
    uint8_t cpu = thread->base.cpu;

    rv = (_kernel.cpus[cpu].current == thread);
#endif

return rv;
```

subsys/cpu_freq/policies/pressure/pressure.c
peter-mitsis: A system that uses CONFIG_SCHED_CPU_MASK_PIN_ONLY=y would probably be a good candidate for enabling CONFIG_CPU_FREQ_PER_CPU_SCALING assuming the hardware supports the latter. But for our purposes, we would probably want to leverage the following somewhere (not intended as a solution at this point, but more as a signpost pointing the way) ...

```
#ifdef CONFIG_CPU_FREQ_PER_CPU_SCALING
    k_thread_foreach_filter_by_cpu(...)
#else
    k_thread_foreach(...)
#endif
```

peter-mitsis (commented):
I have a little more to do on the reviewing, but I'm getting a nasty headache.

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+#if defined(CONFIG_SMP)
+	/* The caller has already ensured that the CPU is fixed */
+	cpu_id = arch_curr_cpu()->id;
+#endif
+
+	k_sched_lock();
+	sys_pressure = get_normalized_sys_pressure();
+	k_sched_unlock();
```
peter-mitsis: I don't think that we need the `k_sched_lock()` nor the `k_sched_unlock()` as the caller has already ensured that the CPU is fixed. Furthermore, we may want to pass the `cpu_id` to `get_normalized_sys_pressure()` so that it has the option of deciding whether to use `k_thread_foreach()` or `k_thread_foreach_filter_by_cpu()`.

peter-mitsis (commented):
That is all I have for now. And no, the headache was not from the code--I've been trying to cut out caffeine and it was from withdrawal.

subsys/cpu_freq/Kconfig
seankyer: It is required. The runnable() APIs are part of the tracing subsystem. I added a note to the docs about potential overheads.

subsys/cpu_freq/cpu_freq.c
```diff
@@ -45,6 +49,16 @@ static void cpu_freq_next_pstate(void)
...
 
+#ifndef CONFIG_CPU_FREQ_PER_CPU_SCALING
+	if (pstate_next == pstate_last) {
+		LOG_DBG("pstate_next: %p == pstate_last: %p - pstate does not need to be updated",
+			pstate_next, pstate_last);
+		return;
+	}
+#endif /* CONFIG_CPU_FREQ_PER_CPU_SCALING */
+
+	pstate_last = pstate_next;
```
peter-mitsis: It appears that this assignment is superfluous for the _!CONFIG_CPU_FREQ_PER_CPU_SCALING_ case.  It could be moved into the preceding _#ifdef_ block and its definition around line 16 could also be made conditional to save a few bytes.

subsys/cpu_freq/Kconfig
```diff
@@ -51,6 +51,23 @@ config CPU_FREQ_POLICY_ON_DEMAND
...
+	  Pressure based policy
+
+if CPU_FREQ_POLICY_PRESSURE
+
+config CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO
+	int "The lowest priority level (highest numerically) to accumulate"
+	default 50
+	help
+	  The lowest priority level (highest numerically) to accumulate. This should be your lowest priority thread.
+	  This priority must not be lower than the idle threads priority.
```
peter-mitsis: thread -> thread's

peter-mitsis (dismissed):

subsys/cpu_freq/cpu_freq.c
```diff
@@ -45,6 +49,16 @@ static void cpu_freq_next_pstate(void)
 		return;
 	}
 
+#ifndef CONFIG_CPU_FREQ_PER_CPU_SCALING
```
ZhaoxiangJin: I understand that in the shared frequency scenario (`!CONFIG_CPU_FREQ_PER_CPU_SCALING`), the policy layer typically relies on all CPUs calling `cpu_freq_policy_pstate_set()`, and only the last CPU actually triggers `cpu_freq_pstate_set()`. But now the early-return you've added here will cause some CPUs to exit directly, which may result in `num_unprocessed_cpus` never reaching zero, and ultimately no one actually sets the frequency, right?

subsys/cpu_freq/cpu_freq.c
```diff
@@ -12,16 +12,20 @@
 
 LOG_MODULE_REGISTER(cpu_freq, CONFIG_CPU_FREQ_LOG_LEVEL);
 
+/* Track the last pstate that was applied to avoid unncessary pstate transitions */
+static const struct pstate *pstate_last;
```
ZhaoxiangJin, line 17: Under SMP, multiple CPUs may execute `cpu_freq_next_pstate()` simultaneously and read/write `pstate_last`. Is there a data race issue here?

subsys/cpu_freq/cpu_freq.c
ZhaoxiangJin: Currently, it's: first `pstate_last = pstate_next`; then call `cpu_freq_policy_pstate_set(pstate_next)`. But `cpu_freq_policy_pstate_set()` might not successfully set it, for example, returning `NULL`. In this case, 'last already updated' would cause subsequent cycles to mistakenly think **it has already been set**, thus skipping the necessary set, am I right?

subsys/cpu_freq/Kconfig
```diff
@@ -51,6 +51,23 @@ config CPU_FREQ_POLICY_ON_DEMAND
 	bool "On-demand Policy"
 	select CPU_LOAD_METRIC
 
+config CPU_FREQ_POLICY_PRESSURE
```
ZhaoxiangJin, line 54: It seems that `k_thread_foreach` and `k_thread_foreach_filter_by_cpu`depend on [CONFIG_THREAD_MONITOR](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/kernel.h#L139)? Should we select `CONFIG_THREAD_MONITOR` here?

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,197 @@
...
+struct pressure_stats {
+	int pressure_acum;
+	int max_pressure;
+};
+
+static bool is_runnable(const struct k_thread *thread)
+{
+	bool rv = false;
+
+	if (thread->base.thread_state == _THREAD_QUEUED) {
```
ZhaoxiangJin: `thread_state` is defined in [bitmask](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/kernel_structs.h#L43), so using `==` here is not the best choice, we should check whether the bit is set.

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/*
+ * The pressure policy iterates through the threads currently sitting in the ready queue
+ * at the time of evaluation and accumulates the sum of their priorities, normalizing them
+ * around CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO, configured by the user. The policy
+ * then iterates through the list of available P-states and selects the first P-state where
+ * the current normalized system pressure is greater than or equal to the load threshold of
+ * the P-state. If the calculated pressure is below all available P-state thresholds, then
+ * the last P-state in the array will be selected. P-states must be defined in increasing
```
ZhaoxiangJin: shouldn't it be decreasing?

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,197 @@
...
+
+#include <zephyr/kernel.h>
+#include <zephyr/kernel_structs.h>
+#include <zephyr/init.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/cpu_freq/policy.h>
+#include <zephyr/cpu_freq/cpu_freq.h>
+
+BUILD_ASSERT(CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO > K_LOWEST_THREAD_PRIO,
+	     "CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO must be lower priority than idle");
```
ZhaoxiangJin: shouldn't `CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO` must be **higher** priority than idle?

subsys/cpu_freq/Kconfig
```diff
@@ -51,6 +51,23 @@ config CPU_FREQ_POLICY_ON_DEMAND
...
+	bool "Pressure based Policy"
+	select TRACING
+	help
+	  Pressure based policy
+
+if CPU_FREQ_POLICY_PRESSURE
+
+config CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO
+	int "The lowest priority level (highest numerically) to accumulate"
+	default 50
```
ZhaoxiangJin: seems 50 is an uncommon priority level. can we set to `NUM_PREEMPT_PRIORITIES`, since [#define K_LOWEST_THREAD_PRIO CONFIG_NUM_PREEMPT_PRIORITIES](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/kernel.h#L58). However, this requires explicitly excluding the idle thread in the pressure policy.

ZhaoxiangJin (commented):
ideal is pretty good, I have some comments.

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: okay, I removed it.

subsys/cpu_freq/cpu_freq.c
seankyer: Good catch. I have also wrapped it in `#ifndef` blocks

subsys/cpu_freq/cpu_freq.c
seankyer: thank you for spotting this. With the current setup, this optimization I've added really only works for the single-core scenario anyways. So in that case, it should just be disabled for SMP entirely. 

Side note: I think that logic of `num_unprocessed_cpu--` should be moved to the main `cpu_freq.c` file, it will be a lot of duplication later.

subsys/cpu_freq/cpu_freq.c
seankyer: See previous comment, I am disabling it for SMP so this shouldn't be a worry.

subsys/cpu_freq/cpu_freq.c
seankyer: yeah you are right. So that should only be set until after the call to `cpu_freq_policy_pstate_set` an only if it's non-null.

subsys/cpu_freq/Kconfig
seankyer: It seems to be building without it. I think Tracing includes it.

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: Thanks, updated it.

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: thats right, this was copy-pasted mostly from on-demand, so it will need to be fixed there too, but in a separate PR.

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: Yeah, confusing myself with the inverse of priority number to priority in the system. I reversed the check, updated the default Kconfig and fixed the comment.

subsys/cpu_freq/Kconfig
seankyer: I am not sure if there is a way to put a #define as a default in Kconfig. I've instead changed the default value to `1` and it's a parameter that the user is expected to configure to their system, since they define their own priorities.

subsys/cpu_freq/policies/pressure/pressure.c
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/kernel_structs.h>
+#include <zephyr/init.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/cpu_freq/policy.h>
+#include <zephyr/cpu_freq/cpu_freq.h>
+
+BUILD_ASSERT(CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO <= K_LOWEST_THREAD_PRIO,
```
peter-mitsis, line 15: Suggestion (not blocking): If CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO is configured to not be at a higher priority level than K_LOWEST_THREAD_PRIO, we could silently fix it. Something like ...

#define CPU_FREQ_POLICY_PRESSURE_THRESHOLD ((CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO <= K_LOWEST_THREAD_PRIO) ? CONFIG_CPU_FREQ_POLICY_PRESSURE_LOWEST_PRIO : K_LOWEST_THREAD_PRIO)

And then we can just use the new macro (CPU_FREQ_POLICY_PRESSURE_THRESHOLD).

peter-mitsis (approved):

subsys/cpu_freq/policies/pressure/pressure.c
seankyer: I like the suggestion. I'll add it as an enhancement in a subsequent PR unless another change is requested in this pull request. Thanks.

ZhaoxiangJin (approved):

PR #101311, opened 2025-12-19
soc/ite: it8xxx2: Add SOC_IT8XXX2_RAM_CODE_NOINLINE option

GTLin08:
Add the SOC_IT8XXX2_RAM_CODE_NOINLINE Kconfig option to prevent functions marked with __soc_ram_code from being inlined when LTO is enabled.

This ensures RAM code functions remain in the RAM section instead of being merged into callers by the compiler.

### DIFF
--- a/soc/ite/ec/it8xxx2/Kconfig
+++ b/soc/ite/ec/it8xxx2/Kconfig
@@ -245,6 +245,15 @@ config SOC_IT8XXX2_ZEPHYR_IN_RAM
 	help
 	  Place zephyr handling code in ILM. This can significantly improve performance.
 
+config SOC_IT8XXX2_RAM_CODE_NOINLINE
+	bool "Force RAM code functions to be non-inlined"
+	depends on LTO
+	help
+	  When enabled, functions marked with __soc_ram_code will use the 'noinline'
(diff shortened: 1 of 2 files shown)


fabiobaltieri (approved):

soc/ite/ec/it8xxx2/Kconfig
```diff
@@ -245,6 +245,15 @@ config SOC_IT8XXX2_ZEPHYR_IN_RAM
 	help
 	  Place zephyr handling code in ILM. This can significantly improve performance.
 
+config SOC_IT8XXX2_RAM_CODE_NOINLINE
```
keith-zephyr, line 248: Does this need to be configurable?  I think it makes sense to always use `noinline` for RAM code.

At a minimum, this option should probably default to Y if `SOC_IT8XXX2_USE_ILM`

GTLin08: I made this configurable because of downstream projects RAM constraints. For some projects with very limited RAM, always forcing `noinline` for RAM code can increase RAM usage and potentially cause build failures.

keith-zephyr (approved):

Dino-Li (approved):

PR #101313, opened 2025-12-19
drivers: sdhc: infineon: add support for pse84

ChanChihHsiang:
Add SDHC driver support
1. support SDIO function on PSE84 board and cy8cproto_062_4343w board
2. support SDMMC on PSE84 - verified Samsung SD card
3. added related dts, kconfig files for PSE84/cy8cproto_062_4343w  board
4. added overlay file and conf. file
    - SDMMC related samples and tests
    - WIFI Shell samples 
5. Removed infineon_sdio.c (HAL based SDIO solution)


drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+#define IFX_SD_HOST_SD_ACMD_OFFSET        (0x40UL)
+
+struct sdhc_infineon_config {
+	const struct pinctrl_dev_config *pincfg;
+	SDHC_Type *reg_addr;
+	uint8_t irq_priority;
+	IRQn_Type irq;
+};
+
+struct __aligned(CONFIG_SDHC_BUFFER_ALIGNMENT) sdhc_infineon_data {
```
danieldegrasse, line 78: Why does this need to be aligned? I assume for the adma descriptors?

`CONFIG_SDHC_BUFFER_ALIGNMENT` should describe the required alignment of DMA _buffers_ for SD data, not the ADMA descriptors. That alignment is specific to this hardware (and could be set with a driver specific Kconfig or hardcoded)

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+}
+
+static void infineon_sdhc_normal_reset(SDHC_Type *base)
+{
+	uint32_t intNormal; /* The normal events mask. */
+
+	intNormal = Cy_SD_Host_GetNormalInterruptStatus(base);
+
+	/* Check the normal event. */
+	if (0UL < intNormal) {
```
danieldegrasse: Why are you using a `UL` suffix here? Shouldn't the `uint32_t` just need a `U` suffix?

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+
+	Cy_SD_Host_SoftwareReset(config->reg_addr, CY_SD_HOST_RESET_DATALINE);
+	Cy_SD_Host_SoftwareReset(config->reg_addr, CY_SD_HOST_RESET_CMD_LINE);
+
+	while ((config->reg_addr->CORE.SW_RST_R != 0) && (timeout_us-- != 0)) {
+		k_busy_wait(1);
+	}
+
+	/* Reset was not cleared by SDHC IP Block. Something wrong. Are clocks enabled? */
+	CY_ASSERT(timeout_us != 0);
```
danieldegrasse: Why use a custom assert here instead of Zephyr's macro?

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_host_poll_transfer_complete(SDHC_Type *base)
+{
+	cy_en_sd_host_status_t ret = CY_SD_HOST_ERROR_TIMEOUT;
+	uint32_t retry = IFX_SD_HOST_RETRY_TIME;
+
+	while (retry > 0UL) {
+		/* Transfer Complete */
+		if (true == _FLD2BOOL(SDHC_CORE_NORMAL_INT_STAT_R_XFER_COMPLETE,
```
danieldegrasse: No need to compare `true` here

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_poll_buf_read_ready(SDHC_Type *base)
+{
+	cy_en_sd_host_status_t ret = CY_SD_HOST_ERROR_TIMEOUT;
+	uint32_t retry = IFX_SD_HOST_RETRY_TIME;
+
+	while (retry > 0UL) {
+		/* Check the Buffer Read ready */
+		if (true == _FLD2BOOL(SDHC_CORE_NORMAL_INT_STAT_R_BUF_RD_READY,
```
danieldegrasse: Same as above, don't compare `true`

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+
+	blkCnt = pcmd->numberOfBlock;
+	blkSize = pcmd->blockSize;
+
+	while (blkCnt > 0UL) {
+		/* Wait for the Buffer Read ready. */
+		ret = infineon_sdhc_poll_buf_read_ready(base);
+
+		if (ret != CY_SD_HOST_SUCCESS) {
+			break;
```
danieldegrasse, line 320: How would this return ever propagate? Wouldn't it be overwritten by `infineon_sdhc_host_poll_transfer_complete`?

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+
+			/* Read data from the Host buffer. */
+			*pcmd->data = Cy_SD_Host_BufferRead(base);
+			pcmd->data++;
+		}
+		blkCnt--;
+	}
+
+	/* Wait for the Transfer Complete. */
+	ret = infineon_sdhc_host_poll_transfer_complete(base);
```
danieldegrasse:
```suggestion
	return infineon_sdhc_host_poll_transfer_complete(base);
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+	bool large_response;
+	int result = 0;
+
+	LOG_DBG("Opcode: %d", cmd->opcode);
+
+	/* Reset semaphore */
+	k_sem_reset(&sdhc_data->transfer_sem);
+	result = k_sem_take(&sdhc_data->thread_lock, K_FOREVER);
+	if (result < 0) {
+		LOG_ERR("Can not take sem!");
```
danieldegrasse:
```suggestion
		LOG_ERR("Cannot take sem!");
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+		sdhc_data->app_cmd = true;
+	} else {
+		sdhc_data->app_cmd = false;
+	}
+	k_sem_give(&sdhc_data->thread_lock);
+
+	return result;
+}
+
+static cy_rslt_t infineon_sdhc_find_best_div(uint32_t hz_src, uint32_t desired_hz,
```
danieldegrasse: Why include a return here if we always return success?

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+	} else if (config->reg_addr == SDHC1) {
+		clk_idx = PCLK_SDHC1_CLK_HF;
+	} else {
+		return CY_SD_HOST_ERROR_INVALID_PARAMETER;
+	}
+
+	source_freq =
+		ifx_cat1_utils_peri_pclk_get_frequency((en_clk_dst_t)clk_idx, &sdhc_data->clock);
+
+	if (result == CY_RSLT_SUCCESS) {
```
danieldegrasse: Rather than use this pattern, could we do something like:

```
result = infineon_sdhc_find_best_div(source_freq, *frequency, &most_suitable_div);
if (result != CY_RSLT_SUCCESS) {
	return result;
}
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+
+		sdhc_data->signal_voltage = ios->signal_voltage;
+	}
+
+	return ret;
+}
+
+static int sdhc_infineon_get_card_present(const struct device *dev)
+{
+	return true;
```
danieldegrasse: Is there no card detect circuit on this system? Also return `1`, not `true`.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,993 @@
...
+static int sdhc_infineon_get_host_props(const struct device *dev, struct sdhc_host_props *props)
+{
+	struct sdhc_infineon_data *sdhc_data = dev->data;
+
+	memcpy(props, &sdhc_data->props, sizeof(struct sdhc_host_props));
+
+	return 0;
+}
+
+bool Cy_SD_Host_IsCardConnected(SDHC_Type const *base)
```
danieldegrasse: Is this function used by the HAL layer? Not clear where it is being called. If the HAL uses it, please add a comment making that clear

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Corrected all UL to U if using uint32_t

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: changed it to __ASSERT

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: removed

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: removed

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I added LOG_WRN if infineon_sdhc_poll_buf_read_ready returns non-zero. And removed some unnecessary ret variables.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: changed it to void

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: based on above change (infineon_sdhc_find_best_div return void), did corresponding changes.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: my bad, it was testing code, removed it

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: This is for adma descriptor, and it is saved in RAM. So we have to clean/flush dcache before kicking DMA.
CONFIG_SDHC_BUFFER_ALIGNMENT is defined in our drivers/sdhc/Kconfig.infineon file.

 I just prevent one rare scenario, when dcache is enabled (ex: cache line is 32 bytes, 0x34000000 ~ 0x34000020 is one cache line), if adma_descriptor_tbl is located at 0x34000010 ~ 0x34000018 and 0x34000000 ~ 0x34000010 is used by other DMA, clean 0x34000000 ~ 0x34000020 cache line could cause issue on 0x34000000 ~ 0x34000010.

Therefore this is reason

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: changed it to 1

boards/infineon/kit_pse84_eval/kconfig.defconfig
```diff
@@ -0,0 +1,31 @@
...
+# Enable L2 Ethernet
+config NET_L2_ETHERNET
+	default y
+
+endif # WIFI
+
+# Heap Pool Size
+config HEAP_MEM_POOL_ADD_SIZE_BOARD
+	int
+	default 15000 if WIFI
```
teburd: Do we really need a heap without WIFI? What for?

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	/* Check the normal event. */
+	if (intNormal > 0U) {
+		/* Clear the normal event. */
+		Cy_SD_Host_ClearNormalInterruptStatus(base, intNormal);
+	}
+}
+
+static void infineon_sdhc_error_reset(SDHC_Type *base)
+{
+	uint32_t intError; /* The error events mask. */
```
teburd:
```suggestion
	uint32_t err_status; /* The error events mask. */
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+};
+
+static void sdhc_infineon_irq_handler(const struct device *dev)
+{
+	const struct sdhc_infineon_config *const config = dev->config;
+	struct sdhc_infineon_data *const sdhc_data = dev->data;
+
+	/* Clear all interrupts that could have been configured */
+	SDHC_Type *base = config->reg_addr;
+	uint32_t interruptStatus = Cy_SD_Host_GetNormalInterruptStatus(base);
```
teburd:
```suggestion
	uint32_t int_status = Cy_SD_Host_GetNormalInterruptStatus(base);
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		/* Disabling transfer complete interrupt mask */
+		Cy_SD_Host_SetNormalInterruptMask(base,
+						  Cy_SD_Host_GetNormalInterruptMask(base) &
+							  (uint32_t)~CY_SD_HOST_XFER_COMPLETE);
+	}
+}
+
+static void infineon_sdhc_normal_reset(SDHC_Type *base)
+{
+	uint32_t intNormal; /* The normal events mask. */
```
teburd:
```suggestion
	uint32_t int_normal; /* The normal events mask. */
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+
+static int sdhc_infineon_reset(const struct device *dev)
+{
+	const struct sdhc_infineon_config *config = dev->config;
+	uint32_t timeout_us = 1000;
+
+	Cy_SD_Host_SoftwareReset(config->reg_addr, CY_SD_HOST_RESET_DATALINE);
+	Cy_SD_Host_SoftwareReset(config->reg_addr, CY_SD_HOST_RESET_CMD_LINE);
+
+	while ((config->reg_addr->CORE.SW_RST_R != 0) && (timeout_us-- != 0)) {
```
teburd: Use a WAIT_FOR macro which takes care of polling and timing out correctly.

https://docs.zephyrproject.org/latest/doxygen/html/group__sys-util.html#ga68eb35df6b4715714312df90209accee

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		k_busy_wait(1);
+	}
+
+	/* Reset was not cleared by SDHC IP Block. Something wrong. Are clocks enabled? */
+	__ASSERT(timeout_us != 0, "Reset was not cleared by SDHC IP Block");
```
teburd: WAIT_FOR returns the final polled value of the expression (SW_RST_R != 0) which can be used in place of timeout_us != 0.

sdhc_hw_reset has a -ETIMEDOUT return documented, we should likely return that here instead of asserting, with a nice LOG_ERR() message perhaps suggesting the clocks may not be configured.

https://docs.zephyrproject.org/latest/doxygen/html/group__sdhc__interface.html#gad6b082a75f1272620b79a7d3a08862f7

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	case SD_SWITCH:
+	case SD_SEND_STATUS:
+	case (SD_APP_SEND_SCR + IFX_SD_HOST_SD_ACMD_OFFSET):
+		return false;
+	default:
+		return true;
+	}
+}
+
+static cy_en_sd_host_auto_cmd_t infineon_sdhc_autocommand(cy_stc_sd_host_cmd_config_t *cmdConfig,
```
teburd:
```suggestion
static cy_en_sd_host_auto_cmd_t infineon_sdhc_autocommand(cy_stc_sd_host_cmd_config_t *cmd_config,
```

I know the camel case/snake case will be forever confusing with the way we do things, but Zephyr uses snake case and we should try our best to follow.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+							  struct sdhc_data *data)
+{
+	if (cmdConfig->commandIndex == SDIO_RW_EXTENDED) {
+		return CY_SD_HOST_AUTO_CMD_NONE;
+	}
+
+	return (data->blocks > 1U) ? CY_SD_HOST_AUTO_CMD_AUTO : CY_SD_HOST_AUTO_CMD_NONE;
+}
+
+static cy_en_sd_host_auto_cmd_t infineon_sdhc_intatblockgap(cy_stc_sd_host_cmd_config_t *cmdConfig,
```
teburd: because these functions are static you do not need to prefix them with infineon_sdhc

intatblockgap is not reader friendly, recommend renaming the function to better match what its doing (little helper for cmd config it looks like).

Rename cmdConfig to cmd_Config

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+							    struct sdhc_data *data)
+{
+	if (cmdConfig->commandIndex == SDIO_RW_EXTENDED) {
+		return false;
+	}
+
+	return (data->blocks > 1U) ? true : false;
```
teburd: This is already a boolean expression no need for ternary.

```suggestion
	return data->blocks > 1U;
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_pollcmdcomplete(const struct device *dev,
+							    uint32_t *time_used_ms)
+{
+	const struct sdhc_infineon_config *config = dev->config;
+	cy_en_sd_host_status_t result = CY_SD_HOST_ERROR_TIMEOUT;
+	uint32_t retry = IFX_SDHC_RETRY_TIMES * IFX_CMD_CMPLT_DELAY_US;
+
+	while (retry > 0U) {
```
teburd: I would use WAIT_FOR here with a reasonable timeout and a small busy_wait.

The two defines can go away then and a simple inline timeout in microseconds is just fine.

You currently have 5 milliseconds of timeout effectively, easily read with WAIT_FOR.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		}
+
+		k_busy_wait(1);
+		retry--;
+	}
+
+	return result;
+}
+
+static cy_rslt_t infineon_sdhc_prepare_for_transfer(const struct device *dev)
```
teburd: This seems to only be used in one place, recommend moving the Cy_SD_Host_SetNormalInterruptMask(...) call to be inline where its being used.

True for any other one line type helper as well, the function indirect isn't necessarily making it easier to read in these cases.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		/* Transfer Complete */
+		if (_FLD2BOOL(SDHC_CORE_NORMAL_INT_STAT_R_XFER_COMPLETE,
+			      SDHC_CORE_NORMAL_INT_STAT_R(base))) {
+			/* Clear the interrupt flag */
+			SDHC_CORE_NORMAL_INT_STAT_R(base) = CY_SD_HOST_XFER_COMPLETE;
+
+			return CY_SD_HOST_SUCCESS;
+		}
+
+		k_busy_wait(IFX_SD_HOST_WRITE_TIMEOUT_MS);
```
teburd: We should not busy wait for milliseconds, use k_sleep

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+						  CY_SD_HOST_XFER_COMPLETE);
+
+	return CY_RSLT_SUCCESS;
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_host_poll_transfer_complete(SDHC_Type *base)
+{
+	uint32_t retry = IFX_SD_HOST_RETRY_TIME;
+
+	while (retry > 0U) {
```
teburd: WAIT_FOR on the _FLD2BOOL(...) status check with a timeout and a reasonable k_sleep(K_MSEC()) as the delay.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	return CY_SD_HOST_ERROR_TIMEOUT;
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_poll_buf_read_ready(SDHC_Type *base)
+{
+	uint32_t retry = IFX_SD_HOST_RETRY_TIME;
+
+	while (retry > 0U) {
+		/* Check the Buffer Read ready */
+		if (_FLD2BOOL(SDHC_CORE_NORMAL_INT_STAT_R_BUF_RD_READY,
```
teburd: Much like the poll on transfer complete comments, should use WAIT_FOR with k_sleep

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+			SDHC_CORE_NORMAL_INT_STAT_R(base) = CY_SD_HOST_BUF_RD_READY;
+
+			return CY_SD_HOST_SUCCESS;
+		}
+
+		k_busy_wait(IFX_SD_HOST_BUFFER_RDY_TIMEOUT_MS);
+		retry--;
+	}
+
+	return CY_SD_HOST_ERROR_TIMEOUT;
```
teburd: This is Zephyr code and should use standard errno's, -ETIMEDOUT and 0 for success.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+			SDHC_CORE_NORMAL_INT_STAT_R(base) = CY_SD_HOST_XFER_COMPLETE;
+
+			return CY_SD_HOST_SUCCESS;
+		}
+
+		k_busy_wait(IFX_SD_HOST_WRITE_TIMEOUT_MS);
+		retry--;
+	}
+
+	return CY_SD_HOST_ERROR_TIMEOUT;
```
teburd: This is Zephyr code and should use standard errno's, -ETIMEDOUT and 0 for success.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		/* Command complete */
+		if (CY_SD_HOST_CMD_COMPLETE ==
+		    (CY_SD_HOST_CMD_COMPLETE &
+		     Cy_SD_Host_GetNormalInterruptStatus(config->reg_addr))) {
+
+			/* Clear interrupt flag */
+			Cy_SD_Host_ClearNormalInterruptStatus(config->reg_addr,
+							      CY_SD_HOST_CMD_COMPLETE);
+
+			result = CY_SD_HOST_SUCCESS;
```
teburd: This is Zephyr code and should use standard errno's, e.g. -ETIMEDOUT and 0 for success.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		/* Wait for the Buffer Read ready. */
+		if (infineon_sdhc_poll_buf_read_ready(base) != CY_SD_HOST_SUCCESS) {
+			LOG_WRN("%s Buffer read is not ready", __func__);
+			break;
+		}
+
+		for (i = blkSize >> 2U; i != 0U; i--) {
+			/* Wait if valid data exists in the Host buffer. */
+			retry = IFX_SD_HOST_RETRY_TIME;
+			while ((false == _FLD2BOOL(SDHC_CORE_PSTATE_REG_BUF_RD_ENABLE,
```
teburd: WAIT_FOR here as well, if the timeout is over ~1 millisecond then please use k_sleep rather than k_busy_wait to allow for thread rescheduling/idling

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	}
+
+	return CY_SD_HOST_ERROR_TIMEOUT;
+}
+
+static cy_en_sd_host_status_t infineon_sdhc_cmd_rx_data(SDHC_Type *base,
+							cy_stc_sd_host_data_config_t *pcmd)
+{
+	uint32_t blkSize;
+	uint32_t blkCnt;
```
teburd:
```suggestion
	uint32_t blk_size = pcmd->blockSize;
	uint32_t blk_cnt = pcmd->numberOfBlock;
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+		blkCnt--;
+	}
+
+	/* Wait for the Transfer Complete. */
+	return infineon_sdhc_host_poll_transfer_complete(base);
+}
+
+static cy_rslt_t infineon_sdhc_config_data_transfer(const struct device *dev,
+						    struct sdhc_data *data,
+						    cy_stc_sd_host_data_config_t *dataConfig)
```
teburd:
```suggestion
						    cy_stc_sd_host_data_config_t *data_config)
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+{
+	const struct sdhc_infineon_config *config = dev->config;
+	struct sdhc_infineon_data *const sdhc_data = dev->data;
+	bool large_response;
+	int result = 0;
+
+	LOG_DBG("Opcode: %d", cmd->opcode);
+
+	/* Reset semaphore */
+	k_sem_reset(&sdhc_data->transfer_sem);
```
teburd, line 543: We should not be resetting the semaphore outside of the lock

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	case SD_ALL_SEND_CID:
+	case SD_SEND_CSD:
+		cmdConfig.enableCrcCheck = true;
+		cmdConfig.enableIdxCheck = false;
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		break;
+
+	case SD_SEND_STATUS:
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		k_busy_wait(500);
```
teburd: This busy wait deserves a comment on why this command (among all commands) needs a delay

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	case SD_ERASE_BLOCK_END:
+	case SD_ERASE_BLOCK_OPERATION:
+	case SDIO_RW_DIRECT:
+	case SD_APP_CMD:
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		break;
+
+	case SD_VOL_SWITCH:
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		k_msleep(200);
```
teburd: This sleep deserves some comments as to why it is needed.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+			result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		} else {
+			result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, true);
+		}
+		break;
+
+	case SDIO_RW_EXTENDED:
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data,
+						!(cmd->arg & BIT(SDIO_CMD_ARG_RW_SHIFT)));
+		k_busy_wait(500);
```
teburd: Please add some inline commentary as to why this sleep is needed, as a reader its unclear.

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+	case SD_WRITE_MULTIPLE_BLOCK:
+		result = infineon_sdhc_send_cmd(dev, &cmdConfig, data, false);
+		break;
+
+	default:
+		result = -ENOTSUP;
+	}
+
+	if (cmdConfig.respType != CY_SD_HOST_RESPONSE_NONE) {
+		large_response = (cmdConfig.respType == CY_SD_HOST_RESPONSE_LEN_136) ? true : false;
```
teburd:
```suggestion
		large_response = (cmdConfig.respType == CY_SD_HOST_RESPONSE_LEN_136);
```

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+
+		sdhc_data->signal_voltage = ios->signal_voltage;
+	}
+
+	return ret;
+}
+
+static int sdhc_infineon_get_card_present(const struct device *dev)
+{
+	return 1;
```
teburd: We should support a chip detect pin perhaps? ST has a reasonable card_present implementation, I recommend taking a look at this

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+
+	return 0;
+}
+
+static DEVICE_API(sdhc, sdhc_infineon_api) = {
+	.reset = sdhc_infineon_reset,
+	.request = sdhc_infineon_request,
+	.set_io = sdhc_infineon_set_io,
+	.get_card_present = sdhc_infineon_get_card_present,
+	.execute_tuning = sdhc_infineon_execute_tuning,
```
teburd, line 1040: This is optional, if we don't implement it do not assign it. Can drop the associated empty execute_tuning function

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,965 @@
...
+{
+	return 0;
+}
+
+static int sdhc_infineon_card_busy(const struct device *dev)
+{
+	const struct sdhc_infineon_config *config = dev->config;
+	int busy_status = 0;
+	/* Check DAT Line Active */
+	uint32_t pState = Cy_SD_Host_GetPresentState(config->reg_addr);
```
teburd:
```suggestion
	uint32_t state = Cy_SD_Host_GetPresentState(config->reg_addr);
```

teburd (changes requested):

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: https://elixir.bootlin.com/zephyr/v4.3.0/source/subsys/sd/sdmmc.c#L567
-> if support UHS, this hooks is must. Otherwise initialization fails

teburd: Thanks for clarifying! @kartben I wonder how your opt/req flagging might work for things like this where some APIs are required under certain circumstances but not others... I don't have an answer but its something to keep in mind perhaps.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Thanks for your suggestion. Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: OK. Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I changed name from infineon_sdhc_intatblockgap to infineon_sdhc_int_at_blockgap.

Speaking of static function naming, all functions are static in sdhc_infineon.c. I just follow sdhc_renesas_ra.c logic file.
if the APIs are DEVICE_API(), which means they are ops/hooks, I use sdhc_infineon_xxx()
if not, I use infineon_sdhc_xxx() instead.

Do you have any suggestion?

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I changed this function to static inline.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: This MACRO is microsecond. I changed the MACRO name from XXX_MS to XXX_US if it is microsecond.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Thanks for your opinions. Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Sure.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Moved it into thread_lock

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I use MACRO instead

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I use MACRO instead

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: I use MACRO instead

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Done

drivers/sdhc/sdhc_infineon.c
teburd: You can name them however you like since they are not global but maybe for simplicity ifx in place of infineon will shorten things quite a bit.

E.g. all functions maybe prefixed

sdhc_ifx is fine, no need to differentiate the ones for the API struct and little helpers like this one.

This is a non-blocking comment though as you've fixed the naming issue at hand.

drivers/sdhc/Kconfig.infineon
```diff
@@ -1,23 +1,40 @@
...
-if SDHC_INFINEON_CAT1
+if (SDHC_INFINEON_CAT1 || SDHC_INFINEON_CAT1_PDL)
 
 config SDHC_INIT_PRIORITY
 	default 70
 
+config SDHC_BUFFER_ALIGNMENT
+	default DCACHE_LINE_SIZE if DCACHE
+	default 1
+	help
```
teburd, line 26: This shouldn't have a help if its meant to be non-selectable by users/applications though menuconfig. Instead add a comment above it describing what its intent is.

drivers/sdhc/sdhc_infineon.c
teburd: I think this makes sense, and I understand not wanting to have a pointer to something holding two words just to get alignment. But as noted here the SDHC_BUFFER_ALIGNMENT name is a bit odd in this context, its really descriptor alignment to a cache line.

Add some comments around why this is needed here and maybe rework the Kconfig name to better suite the intent. I think that would likely resolve the issue (at least for me!) but ultimately still up to @danieldegrasse

drivers/sdhc/sdhc_infineon.c
```diff
@@ -0,0 +1,935 @@
...
+				 struct sdhc_data *data)
+{
+	const struct sdhc_infineon_config *config = dev->config;
+	struct sdhc_infineon_data *const sdhc_data = dev->data;
+	bool large_response;
+	int result = 0;
+
+	LOG_DBG("Opcode: %d", cmd->opcode);
+
+	result = k_sem_take(&sdhc_data->thread_lock, K_FOREVER);
```
teburd: k_sem_take with K_FOREVER is infallible https://docs.zephyrproject.org/latest/doxygen/html/group__semaphore__apis.html#gac71e2383c1920dddc45a561cacfef090 so no need to check the return here

drivers/sdhc/Kconfig.infineon
ChanChihHsiang: Not sure I understood what you mean. currently it is an invisible symbol (An invisible symbol is a symbol without a prompt.)
You are supposed SDHC_BUFFER_ALIGNMENT should be invisible, isn't it?

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Got it. Done. Thanks.

drivers/sdhc/sdhc_infineon.c
ChanChihHsiang: Based on PSE84 HW, it has a GPIO pin to do this SD card detection. So I added this logic for SD card. if SDIO, it always return true.

boards/infineon/kit_pse84_eval/kconfig.defconfig
ChanChihHsiang: I have no strong reason, I just copy from https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/infineon/cy8cproto_062_4343w/Kconfig.defconfig.
I moved it into WIFI MACRO

ChanChihHsiang:
@de-nordic Can you take a look when you have time or re-assign assignees? Thank you

teburd (dismissed):

drivers/sdhc/sdhc_infineon.c
teburd: Could still use some commentary here, please do add

ChanChihHsiang: Added commentary above "struct sdhc_infineon_data".

teburd (dismissed):

teburd (dismissed):
Refreshing +1, all my comments have been addressed.

jukkar (dismissed):
+1 to wifi changes

teburd (dismissed):
Refresh +1

teburd (approved):

sreeramIfx (approved):

PR #101315, opened 2025-12-19
dts/arm/st: Add cryp to STM32H753

jacobschloss:
STM32H753 has a cryp block like the STM32H755, but it is missing from the stm32h753.dtsi file.

Adds an entry to stm32h753.dtsi essentially copied from H755.

Untested, but confirmed that the reg address, reg size, AHB clock enable number, interrupt number match the H753 datasheet.

### DIFF
--- a/boards/st/nucleo_h753zi/nucleo_h753zi.dts
+++ b/boards/st/nucleo_h753zi/nucleo_h753zi.dts
@@ -275,3 +275,7 @@ zephyr_udc0: &usbotg_fs {
 &iwdg1 {
 	status = "okay";
 };
+
+&cryp {
+	status = "okay";
+};
--- a/dts/arm/st/h7/stm32h753.dtsi
+++ b/dts/arm/st/h7/stm32h753.dtsi
@@ -9,5 +9,14 @@
 / {
 	soc {
 		compatible = "st,stm32h753", "st,stm32h7", "simple-bus";
+
+		cryp: cryp@48021000 {
+			compatible = "st,stm32-cryp";
+			reg = <0x48021000 0x400>;
+			clocks = <&rcc STM32_CLOCK(AHB2, 4)>;
+			resets = <&rctl STM32_RESET(AHB2, 4)>;
+			interrupts = <79 0>;
+			status = "disabled";
+		};
 	};
 };
--- a/dts/arm/st/h7/stm32h755.dtsi
+++ b/dts/arm/st/h7/stm32h755.dtsi
@@ -15,6 +15,7 @@
 			compatible = "st,stm32-cryp";
 			reg = <0x48021000 0x400>;
 			clocks = <&rcc STM32_CLOCK(AHB2, 4)>;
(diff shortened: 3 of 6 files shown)


jacobschloss:
Looks like H755 and H757 are missing the resets property that is marked as required, should I add those in this PR too?

erwango (changes requested):
Thanks for this addition, but ideally I'd like to go a bit further on this, if you're ready for it:
- fix the missing reset property globally on other h7.dtsi files (use a dedicated commit for this)
- Enable node on a board in order to ensure driver is built in CI (through samples/drivers/crypto). Use a dedicated commit for this as well

jacobschloss:
Running on nucleo_h753zi seems to work:

```
*** Booting Zephyr OS build v4.3.0-2699-g8988ffc326d1 ***
I: Cipher Sample
I: ECB Mode
I: Output length (encryption): 16
I: ECB mode ENCRYPT - Match
I: Output length (decryption): 16
I: ECB mode DECRYPT - Match
I: CBC Mode
I: Output length (encryption): 80
I: CBC mode ENCRYPT - Match
I: Output length (decryption): 64
I: CBC mode DECRYPT - Match
I: CTR Mode
I: Output length (encryption): 64
I: CTR mode ENCRYPT - Match
I: Output length (decryption): 64
I: CTR mode DECRYPT - Match
I: CCM Mode
E: Unsupported mode
I: GCM Mode
E: Unsupported mode
```

erwango (changes requested):
Thanks for this extra mile.

One request would be to squash commits 1 & 2 (adding the crypt node with complete description (ie including `reset` prop) in one shot.

Then, you probably need to fix your sign off by adding `<>` around the email address to pass the compliance checks

etienne-lms (commented):
Could you refine the commit message header lines?
1st commit:
```patch
-dts/arm/st: Add cryp to STM32H753
+dts: arm: st: h7: Add cryp to STM32H753/H755
```
2nd commit:
```patch
-dts/arm/st: Add cryp to STM32H753
+dts: arm: st: h7: Add missing cryp reset controllers
```
3rd commit:
```patch
-dts/arm/st: Add cryp to STM32H753
+boards: st: nucleo_h753zi: enable cryp node
```
For that last commit, if enabling `cryp` for this board, should it be added in `tests/crypto/crypto_aes/testcase.yaml`? (together with a `CRYPTO_DEV_COMPAT` change in the test **main.c**)

jacobschloss:
I ran the crypto tests locally with

`./scripts/twister -p nucleo_h753zi --device-testing --device-serial /dev/ttyACM0 -T tests/crypto`

and the crypto.aes suite seems to run with 10 passing and 4 skipped for me (gcm and ccm skipped).

etienne-lms (approved):
Looks all good the me.

erwango (approved):

valeriosetti (approved):

jacobschloss:
I think I missed that H750 also has cryp hardware, as pointed out in #102224. ok if I add that one too?

PR #101320, opened 2025-12-19
scripts: check_compliance: read text files as UTF-8 on Windows

ZhaoxiangJin:
KeepSorted can fail on Windows when the active code page is not UTF-8 (e.g. GBK), because text files in the Zephyr tree are UTF-8. Detect file type via libmagic and only process text/* files, avoiding attempts to decode binary content. Open text files via GIT_TOP / <path> with encoding="utf-8" (and errors="surrogateescape") to prevent UnicodeDecodeError during local compliance runs without changing the KeepSorted sorting rules.

### DIFF
--- a/scripts/ci/check_compliance.py
+++ b/scripts/ci/check_compliance.py
@@ -2404,11 +2404,6 @@ class KeepSorted(ComplianceTest):
         return -1
 
     def check_file(self, file, fp):
-        mime_type = magic.from_file(os.fspath(file), mime=True)
-
-        if not mime_type.startswith("text/"):
-            return
-
         block_data = ""
         in_block = False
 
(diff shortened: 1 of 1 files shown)


scripts/ci/check_compliance.py
```diff
@@ -2424,46 +2426,52 @@ def check_file(self, file, fp):
...
+                    block_data += line
 
         if in_block:
             self.failure(f"unterminated {start_marker} in {file}")
 
     def run(self):
         for file in get_files(filter="d"):
-            with open(file) as fp:
-                self.check_file(file, fp)
+            self.check_file(file)
```
pdgendt: Can we keep the `open` call here? It reverts the need to indent the `check_file` method like in this PR.

The `GIT_TOP` can be added here too.

yvesll: And as [PEP-686](https://peps.python.org/pep-0686/) will make [Python UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode) default since Python 3.15. We can add a hint in doc to suggest Windows users set `PYTHONUTF8` and `PYTHONIOENCODING` in env.

pdgendt: I think that's a hard bargain, I'm afraid we need to make it explicit ourselves..

yvesll: Yes, developer shall open file with explicit encoding. So it's just a hint to avoid issue in third party code.

pdgendt (approved):

yvesll (approved):

fabiobaltieri (approved):

nashif (approved):

PR #101324, opened 2025-12-19
drivers: usb: Adjust log level

alexsven:
- Change all LOG_INF to LOG_DBG
- Change most LOG_ERR to LOG_WRN

- Viewpoint for making this change: Drivers should never have LOG_INF, they should just operate silently. If a fault is not breaking the application it should throw a warning and not an error. Error during setup is a clear LOG_ERR though.

Disclaimer: I do not know the inner workings of the USB drivers, so there might be some places where I've set the wrong level. The general idea is to limit the amount of logs from the drivers and reserver LOG_ERRs for actual breaking faults that should be caught in CI

### DIFF
--- a/subsys/usb/device_next/usbd_ch9.c
+++ b/subsys/usb/device_next/usbd_ch9.c
@@ -130,7 +130,7 @@ static int sreq_set_configuration(struct usbd_context *const uds_ctx)
 	const enum usbd_speed speed = usbd_bus_speed(uds_ctx);
 	int ret;
 
-	LOG_INF("Set Configuration Request value %u", setup->wValue);
+	LOG_DBG("Set Configuration Request value %u", setup->wValue);
 
 	/* Not specified if wLength is non-zero, treat as error */
 	if (setup->wValue > UINT8_MAX || setup->wLength) {
@@ -207,7 +207,7 @@ static int sreq_set_interface(struct usbd_context *const uds_ctx)
 
 	ret = usbd_interface_set(uds_ctx, setup->wIndex, setup->wValue);
 	if (ret == -ENOENT) {
-		LOG_INF("Interface or alternate does not exist");
+		LOG_DBG("Interface or alternate does not exist");
 		errno = ret;
 		ret = 0;
 	}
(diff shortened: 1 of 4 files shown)


subsys/usb/device_next/usbd_ch9.c
```diff
@@ -207,7 +207,7 @@ static int sreq_set_interface(struct usbd_context *const uds_ctx)
 
 	ret = usbd_interface_set(uds_ctx, setup->wIndex, setup->wValue);
 	if (ret == -ENOENT) {
-		LOG_INF("Interface or alternate does not exist");
+		LOG_DBG("Interface or alternate does not exist");
```
koffes, line 210: Unsure if this should actually be a WRN

subsys/usb/device_next/usbd_ch9.c
```diff
@@ -1040,14 +1040,13 @@ static int handle_setup_request(struct usbd_context *const uds_ctx,
 	}
 
 	if (errno) {
-		LOG_INF("protocol error:");
-		LOG_HEXDUMP_INF(setup, sizeof(*setup), "setup:");
+		LOG_DBG("protocol error:");
```
koffes, line 1043: LOG_WRN?

subsys/usb/device_next/usbd_ch9.c
```diff
@@ -1137,39 +1136,38 @@ int usbd_handle_ctrl_xfer(struct usbd_context *const uds_ctx,
...
 	if (USB_EP_GET_IDX(bi->ep)) {
-		LOG_ERR("Can only handle control requests");
+		LOG_WRN("Can only handle control requests");
 		return -EIO;
 	}
 
 	if (err && err != -ENOMEM && !bi->setup) {
 		if (err == -ECONNABORTED) {
-			LOG_INF("Transfer 0x%02x aborted (bus reset?)", bi->ep);
+			LOG_DBG("Transfer 0x%02x aborted (bus reset?)", bi->ep);
```
koffes, line 1145: WRN?

henrikbrixandersen: I think DBG is okay here. Aborted transfers are okay in certain situations.

koffes (approved):

jfischer-no (changes requested):
Neither the changes, nor the commit message, nor doing all the different changes in a single commit message seems to be acceptable to me. Please do not use bullet points of changes in the commit message, they are also obvious from the diff. Each change of error logging to warning requires own justification.

> Drivers should never have LOG_INF,

Where did you get that from?

> If a fault is not breaking the application it should throw a warning and not an error.

Maybe that is what you do in your application. In USB drivers and the stack, I prefer to decide what error should be logged at what level.

PR #101326, opened 2025-12-19
cmake: toolchain: xcc: prioritize board-specific toolchain version

anilozrenk:
## Description

This PR changes the toolchain version resolution order for the Xtensa XCC toolchain to check for board-specific `TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET}` first, falling back to the generic `TOOLCHAIN_VER` if not defined.

## Motivation

If `TOOLCHAIN_VER` is defined, cmake won't check other. It makes it hard to maintain various boards that use different toolchains. E.g. if only one board use different version, but other 100 board which same version, we had to define all of them individually. Now we can define `TOOLCHAIN_VER*` variables in ease.

## Impact
- No breaking changes
- Enables more flexible per-board toolchain version configuration


cmake/toolchain/xcc/common.cmake
```diff
@@ -7,13 +7,13 @@ if(NOT EXISTS ${XTENSA_TOOLCHAIN_PATH})
...
+zephyr_get(TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET})
+if(DEFINED TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET})
+  set(XTENSA_TOOLCHAIN_VER ${TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET}})
 else()
-  zephyr_get(TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET})
-  if(DEFINED TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET})
-    set(XTENSA_TOOLCHAIN_VER ${TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET}})
+  zephyr_get(TOOLCHAIN_VER)
+  if(DEFINED TOOLCHAIN_VER)
+    set(XTENSA_TOOLCHAIN_VER ${TOOLCHAIN_VER})
```
tejlmand: all of this logic with `if` / `else` and `set` can be simplified to:
```
zephyr_get(XTENSA_TOOLCHAIN_VER VAR TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET} TOOLCHAIN_VER)
```

which will set the value of `XTENSA_TOOLCHAIN_VER` to `TOOLCHAIN_VER_${NORMALIZED_BOARD_TARGET}` if that one is defined, or fallback to use `TOOLCHAIN_VER`

tejlmand (changes requested):
would be nice to simplify.

cmake/toolchain/xcc/common.cmake
anilozrenk: Oh, thanks for advice @tejlmand.

tejlmand (approved):

MaureenHelm (approved):

PR #101327, opened 2025-12-19
kernel: thread monitor - Assert reusing TCB in use

ChidvilasYQC:
Add a check in thread monitor to assert if a TCB is reused while it is still present in the kernel thread list.
This prevents undefined behavior due to TID reuse before proper cleanup.

tests: Add test for thread tcb reuse

### DIFF
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -873,6 +873,28 @@ static inline void init_thread_usage(struct k_thread *thread)
 #endif /* CONFIG_SCHED_THREAD_USAGE */
 }
 
+static inline void check_if_tid_is_reused(struct k_thread *new_thread)
+{
+#ifdef CONFIG_THREAD_MONITOR
+#ifdef CONFIG_ASSERT
+	k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);
+	bool reused = false;
+
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		if (t == new_thread) {
+			reused = true;
+			break;
+		}
+	}
+
+	k_spin_unlock(&z_thread_monitor_lock, key);
+
+	__ASSERT(!reused, "thread %p is already in the running list", new_thread);
+#endif
+#endif
+}
+
 /*
  * The provided stack_size value is presumed to be either the result of
  * K_THREAD_STACK_SIZEOF(stack), or the size value passed to the instance
@@ -887,6 +909,8 @@ char *z_setup_new_thread(struct k_thread *new_thread,
 	char *stack_ptr;
 
 	Z_ASSERT_VALID_PRIO(prio, entry);
+
+	check_if_tid_is_reused(new_thread);
 
 	thread_abort_cleanup_check_reuse(new_thread);
 	init_thread_obj_core(new_thread);
--- a/kernel/thread.c
+++ b/kernel/thread.c
@@ -873,6 +873,30 @@ static inline void init_thread_usage(struct k_thread *thread)
 #endif /* CONFIG_SCHED_THREAD_USAGE */
 }
 
+/* Assert the thread wasn't already linked into the monitor list. */
+static inline void assert_thread_not_reused(struct k_thread *new_thread)
+{
+#ifdef CONFIG_THREAD_MONITOR
+#ifdef CONFIG_ASSERT
+	k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);
+	bool reused = false;
+
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		if (t == new_thread) {
+			reused = true;
+			break;
+		}
+	}
+
+	k_spin_unlock(&z_thread_monitor_lock, key);
+
+	__ASSERT(!reused, "thread %p is already in the running list", new_thread);
+#endif /* CONFIG_ASSERT */
+#endif /* CONFIG_THREAD_MONITOR */
+	ARG_UNUSED(new_thread);
+}
+
 /*
  * The provided stack_size value is presumed to be either the result of
  * K_THREAD_STACK_SIZEOF(stack), or the size value passed to the instance
@@ -887,6 +911,8 @@ char *z_setup_new_thread(struct k_thread *new_thread,
 	char *stack_ptr;
 
 	Z_ASSERT_VALID_PRIO(prio, entry);
+
+	assert_thread_not_reused(new_thread);
 
 	thread_abort_cleanup_check_reuse(new_thread);
 	init_thread_obj_core(new_thread);
--- a/tests/kernel/threads/thread_error_case/src/main.c
+++ b/tests/kernel/threads/thread_error_case/src/main.c
@@ -14,8 +14,11 @@ static ZTEST_DMEM int case_type;
 
 static K_THREAD_STACK_DEFINE(tstack, STACK_SIZE);
 static K_THREAD_STACK_DEFINE(test_stack, STACK_SIZE);
+static K_THREAD_STACK_DEFINE(reuse_stack1, STACK_SIZE);
+static K_THREAD_STACK_DEFINE(reuse_stack2, STACK_SIZE);
 static struct k_thread tdata;
 static struct k_thread test_tdata;
+static struct k_thread reuse_tdata;
 
 /* identifiers for each negative test scenario */
 enum {
@@ -32,12 +35,19 @@ enum {
 	THREAD_PRIORITY_GET_NULL,
 	THREAD_WAKEUP_NULL,
 	THREAD_CREATE_SUPERVISOR,
-	THREAD_CREATE_ESSENTIAL
+	THREAD_CREATE_ESSENTIAL,
+	THREAD_REUSE_ACTIVE_THREAD_ID
 } neg_case;
 
 static void test_thread(void *p1, void *p2, void *p3)
 {
 	/* intentionally empty target thread */
+}
+
+static void long_running_thread(void *p1, void *p2, void *p3)
+{
+	/* Sleep for a long time to keep thread active */
+	k_sleep(K_FOREVER);
 }
 
 static void tThread_entry_negative(void *p1, void *p2, void *p3)
@@ -148,6 +158,19 @@ static void tThread_entry_negative(void *p1, void *p2, void *p3)
 			test_thread, NULL, NULL, NULL,
 			K_PRIO_PREEMPT(THREAD_TEST_PRIORITY),
 			K_USER | K_ESSENTIAL, K_NO_WAIT);
+		break;
+	case THREAD_REUSE_ACTIVE_THREAD_ID:
+		ztest_set_fault_valid(true);
+		if (k_is_user_context()) {
+			perm = perm | K_USER;
+		}
+		/* Create first thread that will run for a while */
+		k_thread_create(&reuse_tdata, reuse_stack1, STACK_SIZE, long_running_thread, NULL,
+				NULL, NULL, K_PRIO_PREEMPT(THREAD_TEST_PRIORITY + 1), perm,
+				K_NO_WAIT);
+		/* Try to reuse the same thread structure while it's still active */
+		k_thread_create(&reuse_tdata, reuse_stack2, STACK_SIZE, test_thread, NULL, NULL,
+				NULL, K_PRIO_PREEMPT(THREAD_TEST_PRIORITY), perm, K_NO_WAIT);
 		break;
 	default:
 		TC_PRINT("should not be here!\n");
@@ -289,6 +312,22 @@ ZTEST_USER(thread_error_case, test_thread_create_stack_overflow)
 }
 
 /**
+ * @brief Test that k_thread_create() asserts if a thread pointer is
+ *        reused, while the thread pointer is in use
+ *
+ * Verifies that thread pointer that is passed to thread create is
+ * not the only which is already created and in use
+ *
+ * @ingroup kernel_thread_tests
+ *
+ * @see k_thread_create()
+ */
+ZTEST_USER(thread_error_case, test_thread_reuse_active_thread_id)
+{
+	create_negative_test_thread(THREAD_REUSE_ACTIVE_THREAD_ID);
+}
+
+/**
  * @brief Test that k_thread_suspend() rejects a NULL thread pointer
  *
  * Verifies that passing a NULL pointer to k_thread_suspend() triggers
@@ -402,7 +441,8 @@ ZTEST_USER(thread_error_case, test_thread_create_essential)
 /* Grant the test thread access to all shared kernel objects. */
 void *thread_grant_setup(void)
 {
-	k_thread_access_grant(k_current_get(), &tdata, &tstack, &test_tdata, &test_stack);
+	k_thread_access_grant(k_current_get(), &tdata, &tstack, &test_tdata, &test_stack,
+			      &reuse_tdata, &reuse_stack1, &reuse_stack2);
 
 	return NULL;
 }
--- a/tests/kernel/threads/thread_error_case/src/main.c
+++ b/tests/kernel/threads/thread_error_case/src/main.c
@@ -14,8 +14,11 @@ static ZTEST_DMEM int case_type;
 
 static K_THREAD_STACK_DEFINE(tstack, STACK_SIZE);
 static K_THREAD_STACK_DEFINE(test_stack, STACK_SIZE);
+static K_THREAD_STACK_DEFINE(reuse_stack1, STACK_SIZE);
+static K_THREAD_STACK_DEFINE(reuse_stack2, STACK_SIZE);
 static struct k_thread tdata;
 static struct k_thread test_tdata;
+static struct k_thread reuse_tdata;
 
 /* identifiers for each negative test scenario */
 enum {
@@ -32,12 +35,19 @@ enum {
 	THREAD_PRIORITY_GET_NULL,
 	THREAD_WAKEUP_NULL,
 	THREAD_CREATE_SUPERVISOR,
-	THREAD_CREATE_ESSENTIAL
+	THREAD_CREATE_ESSENTIAL,
+	THREAD_REUSE_ACTIVE_THREAD_ID
 } neg_case;
 
 static void test_thread(void *p1, void *p2, void *p3)
 {
 	/* intentionally empty target thread */
+}
+
+static void long_running_thread(void *p1, void *p2, void *p3)
+{
+	/* Sleep for a long time to keep thread active */
+	k_sleep(K_FOREVER);
 }
 
 static void tThread_entry_negative(void *p1, void *p2, void *p3)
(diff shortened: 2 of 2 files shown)


tests/kernel/threads/reuse_tid/src/main.c
```diff
@@ -0,0 +1,106 @@
...
+			       K_KERNEL_STACK_SIZEOF(test_thread_reused_stack),
+			       reused_thread_waitforever, NULL, NULL, NULL,
+			       TEST_THREAD_REUSED_PRIORITY, 0, K_NO_WAIT);
+
+	printf("Giving test sem\n");
+	k_sem_give(&test_sem);
+	printf("Giving test sem again\n");
+	k_sem_give(&test_sem);
+
+	k_thread_join(tid1, K_FOREVER);
```
peter-mitsis: What purpose do these lines serve?  If the test behaves as expected, then we never reach them. If it fails, then the failure should be caught by the subsequent ztest_test_fail().

ChidvilasYQC: My initial approach was to give semaphore twice to make both the faulty threads run.

But this logic may not give faults as expected. So, update the test function to use atomic counter and raise fault.

kernel/thread.c
```diff
@@ -657,6 +657,11 @@ char *z_setup_new_thread(struct k_thread *new_thread,
 	new_thread->no_wake_on_timeout = false;
 #endif /* CONFIG_EVENTS */
 #ifdef CONFIG_THREAD_MONITOR
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		__ASSERT(t != new_thread, "thread already in list");
+	}
```
peter-mitsis: A few notes about this block ...

1. This section will need to be locked using the `z_thread_monitor_lock` spin lock. This is to ensure that the linked list being traversed does not get modified while the searching the list.
2. The search should be moved up to the start of the routine. Otherwise if there is reuse, then we run the risk of corrupting data before we find out that it was in use.
3. Although the search could be done using `k_thread_foreach()` (which would automatically take care of the locking), it may be worthwhile make the inclusion of the search dependent upon both `CONFIG_THREAD_MONITOR` and the definition of `__ASSERT`. I am thinking that if the `__ASSERT` macro is empty, we should avoid an otherwise unnecessary linear search. However, if the compiler is smart enough to realize that an empty macro means the linear search can be skipped, then that simplifies things (though a comment about that would be warranted).

ChidvilasYQC: Thanks for pointing that this loop is not protected under the lock.
1. Updated code to use `z_thread_monitor_lock`
2. This search is moved to the beginning of the `k_thread_create` implementation.
3. Since `z_thread_monitor_lock` must be unlocked before calling `__ASSERT`, I added an if condition to check for TID reuse. Inside this if block, the spinlock is released and the assertion is triggered.
Because the compiler will not optimize out `k_spin_unlock` inside the if block when assertions are disabled, I placed this search under `CONFIG_ASSERT`.

ChidvilasYQC:
> @ChidvilasYQC - Instead of appending new commits that address the review comments, please squash them into the relevant originating commit(s).

Squashed all the commits and rebased main branch.

kernel/thread.c
peter-mitsis: This must be moved up to near the start of the routine (within that spin-locked section) to prevent a race condition.

ChidvilasYQC: @peter-mitsis, thanks for pointing it, moved this block also into the top, where reuse check is.

peter-mitsis (dismissed):

nashif (changes requested):
is there a real problem this change is solving? what is the bug and when does this happen and why? 
asserts should not be enable in production, so this does not solve any problem really..

tests/kernel/threads/reuse_tid/src/main.c
```diff
@@ -0,0 +1,113 @@
+/*
```
nashif: can this test be folded into some existing thread test suite? Having  a dedicated application for this is a bit extreme.

nashif (dismissed):

kernel/thread.c
```diff
@@ -586,6 +586,28 @@ char *z_setup_new_thread(struct k_thread *new_thread,
...
 	Z_ASSERT_VALID_PRIO(prio, entry);
 
+#ifdef CONFIG_THREAD_MONITOR
+	k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);
+
+#ifdef CONFIG_ASSERT
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		if (t == new_thread) {
+			k_spin_unlock(&z_thread_monitor_lock, key);
```
andyross: This line will produce a double-unlock in the situation where the app has a custom assert handler that doesn't panic.  And obviously, in the default case where assert does panic, you don't care about lock hygiene.  Asserts and panics work just fine when you're holding a spinlock.

Remove, basically.

andyross (commented):
One slightly pedantic note about the lock.  Reviewed mostly just the kernel code, not the test.

Generally this looks reasonable to me.  It's a legitimate (and easy) mistake to make, and this test probably provides value.  There's a performance worry about adding the O(N) search of the global thread list on every thread creation, but in practice there aren't that many threads in a running system (the size of the stacks needed keeps these things down into the "few dozen" or so for realistic apps).

It's also been discussed that we could solve this in a heavier/less-compatible way with thread struct state: maybe require that the thread struct have a magic "has been destroyed" cookie set on it somewhere.  That would require the use of a new API like `k_thread_struct_init()` or whatever though.  But we could also grandfather in a rule like "zero-filled structs are OK" and probably make it for for 99% of existing code.

tests/kernel/threads/reuse_tid/src/main.c
ChidvilasYQC: yes, moved this test code into `tests/kernel/threads/thread_error_case` test case

kernel/thread.c
ChidvilasYQC: got it, I have removed spin unlock.

tests/kernel/threads/thread_error_case/src/main.c
```diff
@@ -25,14 +28,21 @@ enum {
 	TIMEOUT_EXPIRES_TICKS,
 	THREAD_CREATE_NEWTHREAD_NULL,
 	THREAD_CREATE_STACK_NULL,
-	THREAD_CTEATE_STACK_SIZE_OVERFLOW
+	THREAD_CTEATE_STACK_SIZE_OVERFLOW,
```
peter-mitsis: Typo: THREAD_CTEATE_STACK_SIZE_OVERFLOW -> THREAD_CTRATE_STACK_SIZE_OVERFLOW

ChidvilasYQC: typo fixed.

ChidvilasYQC:
@peter-mitsis, @nashif and @andyross can you please review

peter-mitsis (approved):

ChidvilasYQC:
Hi @peter-mitsis, shall I close this PR and create a new one to fix issues with PR checks?

peter-mitsis:
@ChidvilasYQC - Near as I can tell, the CI failures are unrelated to the changes. My recommendation at this point would be ...
1. If you know how to fix the cause of the failures, submit a separate PR to fix them.
2. When fixed (or if already fixed), rebase and re-push to re-run CI.

peter-mitsis (dismissed):
Note to any other reviewers: The CI failures are unrelated to this change set.

kernel/thread.c
```diff
@@ -589,6 +589,27 @@ char *z_setup_new_thread(struct k_thread *new_thread,
...
+		}
+	}
+#endif
+	new_thread->entry.pEntry = entry;
+	new_thread->entry.parameter1 = p1;
+	new_thread->entry.parameter2 = p2;
+	new_thread->entry.parameter3 = p3;
+
+	new_thread->next_thread = _kernel.threads;
+	_kernel.threads = new_thread;
```
npitre: With this block moved to the top of `z_setup_new_thread()`, the new thread is linked into `_kernel.threads` **before** `z_waitq_init()`, `z_init_thread_base()`, `setup_thread_stack()`, `arch_new_thread()`, and the rest of the initialization run further down. Anything iterating the list on another CPU — `k_thread_foreach`, `k_thread_foreach_filter_by_cpu`, a debug shell, the thread analyzer — will now observe a thread with uninitialized `base` (no valid `thread_state`/`prio`/`cpu_mask`), NULL/stale `switch_handle`, etc. The previous code only published the thread after full init for exactly this reason.

Suggest keeping the insertion at the original (end-of-routine) location, and doing the check + insert atomically under the same lock there:

```c
#ifdef CONFIG_THREAD_MONITOR
    k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);

#ifdef CONFIG_ASSERT
    for (struct k_thread *t = _kernel.threads; t != NULL; t = t->next_thread) {
        __ASSERT(t != new_thread,
                 "thread %p is already in the running list", new_thread);
    }
#endif

    new_thread->entry.pEntry     = entry;
    new_thread->entry.parameter1 = p1;
    new_thread->entry.parameter2 = p2;
    new_thread->entry.parameter3 = p3;

    new_thread->next_thread = _kernel.threads;
    _kernel.threads = new_thread;
    k_spin_unlock(&z_thread_monitor_lock, key);
#endif
```

Same atomicity guarantee, but no partially-initialized thread is ever visible in the global list. If you specifically want to fail fast on reuse *before* clobbering the caller's TCB, do a separate check-only walk under the lock at the top and keep the insert at the bottom — but the assert is best-effort either way, since two CPUs racing on the same TCB will already be clobbering each other before either's check runs.

kernel/thread.c
```diff
@@ -589,6 +589,27 @@ char *z_setup_new_thread(struct k_thread *new_thread,
...
 	Z_ASSERT_VALID_PRIO(prio, entry);
 
+#ifdef CONFIG_THREAD_MONITOR
+	k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);
+
+#ifdef CONFIG_ASSERT
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		if (t == new_thread) {
+			__ASSERT(0, "thread %p is already in the running list\n", new_thread);
```
npitre: The `if (...) __ASSERT(0, ...)` form is unusual; the idiom elsewhere in this file is to fold the condition into the assert itself:

```c
#ifdef CONFIG_ASSERT
    for (struct k_thread *t = _kernel.threads; t != NULL; t = t->next_thread) {
        __ASSERT(t != new_thread,
                 "thread %p is already in the running list", new_thread);
    }
#endif
```

Keep the `#ifdef CONFIG_ASSERT` (or `if (IS_ENABLED(CONFIG_ASSERT))`) wrapper — the compiler won't eliminate the list walk on its own, since it reads through `_kernel.threads` / `next_thread`.

Minor: the trailing `\n` in the message isn't the style used by other `__ASSERT` calls in this file.

tests/kernel/threads/thread_error_case/src/main.c
```diff
@@ -95,6 +105,24 @@ static void tThread_entry_negative(void *p1, void *p2, void *p3)
...
+		if (k_is_user_context()) {
+			perm = perm | K_USER;
+		}
+		/* Create first thread that will run for a while */
+		k_thread_create(&reuse_tdata, reuse_stack1, STACK_SIZE,
+			long_running_thread, NULL, NULL, NULL,
+			K_PRIO_PREEMPT(THREAD_TEST_PRIORITY + 1),
+			perm, K_NO_WAIT);
+		/* Give the thread time to start */
+		k_sleep(K_MSEC(10));
```
npitre: Is the `k_sleep(K_MSEC(10))` needed here? `k_thread_create()` adds the new thread to `_kernel.threads` synchronously (kernel/thread.c around the monitor-lock block), and the first thread's lower priority (`THREAD_TEST_PRIORITY + 1`) means the caller isn't preempted — so the second `k_thread_create` finds `reuse_tdata` on the list and asserts whether or not the first thread has ever been scheduled in. `k_sleep` doesn't remove a thread from the monitor list either; only `z_thread_monitor_exit` does, on abort.

If that's right, dropping the sleep and making `long_running_thread` just `k_sleep(K_FOREVER)` would remove a timing dependency from the test.

npitre (changes requested):
Approach is reasonable, but the new monitor-list insertion at the top of `z_setup_new_thread()` publishes the thread to `_kernel.threads` before its `base`, stack and arch state are initialized — see inline comment. A couple of style/test nits too.

kernel/thread.c
ChidvilasYQC: Right, I was having spin unlock before assert, so I had to use if condition check.

But after this suggestion from @andyross, I have removed the spin unlock. But kept the if condition. Since, there is only one assert statement now, I will remove if condition.
>This line will produce a double-unlock in the situation where the app has a custom assert handler that doesn't panic. And obviously, in the default case where assert does panic, you don't care about lock hygiene. Asserts and panics work just fine when you're holding a spinlock.
Remove, basically.

kernel/thread.c
ChidvilasYQC: @npitre, this code is protected by the `z_thread_monitor_lock`, hence other utilities like shell thread commands must use this lock before reading thread list.

Also, this suggestion from @peter-mitsis makes sense. We cannot let `new_thread` values to be overwritten.
>The search should be moved up to the start of the routine. Otherwise if there is reuse, then we run the risk of corrupting data before we find out that it was in use.

tests/kernel/threads/thread_error_case/src/main.c
ChidvilasYQC: updated with for ever sleep

kernel/thread.c
npitre: The lock isn't the issue I'm raising. `z_thread_monitor_lock` serializes access to the list *pointers*, so a reader won't trip over a half-linked `next_thread` — agreed. But it doesn't order the initialization of `new_thread`'s own fields.

With both the check and the insert (`_kernel.threads = new_thread`) at the top, the thread is published into the global list before `z_init_thread_base()`, `arch_new_thread()`, and the rest run below. A `k_thread_foreach` / thread-analyzer / shell walk on another CPU that correctly takes the lock will find `new_thread` and then read `base.thread_state`, `base.prio`, `base.cpu_mask`, `switch_handle` — all still garbage at that point. The lock is held correctly on both sides; the data simply isn't initialized yet.

To get both properties — fail-fast on reuse *before* clobbering the caller's TCB, and no partially-initialized thread visible in the list — split it: a check-only walk at the top under the lock (no insert), and keep the insert at the bottom after the thread is fully built, under the lock again. The reuse `__ASSERT` fires from the top walk; publication still happens last.

ChidvilasYQC:
@npitre, moved reuse tid check into a new function and it happens at the start of z_setup_new_thread

kernel/thread.c
```diff
@@ -873,6 +873,26 @@ static inline void init_thread_usage(struct k_thread *thread)
...
+static inline void check_if_tid_is_reused(struct k_thread *new_thread)
+{
+#ifdef CONFIG_THREAD_MONITOR
+#ifdef CONFIG_ASSERT
+	k_spinlock_key_t key = k_spin_lock(&z_thread_monitor_lock);
+
+	/* Check if the thread is already in the list */
+	for (struct k_thread *t = _kernel.threads; t; t = t->next_thread) {
+		if (t == new_thread) {
+			k_spin_unlock(&z_thread_monitor_lock, key);
```
npitre: This unlock inside the `if`, together with the unconditional `k_spin_unlock` at the end of the function, double-unlocks `z_thread_monitor_lock` on reuse when the assert handler returns (a custom, non-panicking handler): unlock here, `__ASSERT` returns, the loop falls through to the trailing unlock. This is the block andyross asked to drop and you replied you would remove. Delete the `if (t == new_thread)` unlock and keep only the folded `__ASSERT` plus the single trailing unlock.

npitre (changes requested):
The double-unlock andyross flagged is back (see inline): the `if (t == new_thread) { k_spin_unlock(...); }` inside the loop plus the new unconditional unlock at the end of the function release the lock twice when reuse is detected and the assert handler returns.

The publish-before-init concern is resolved now that the walk is check-only and the insert stays in z_thread_monitor_init(), and the assert-message idiom is fixed. Minor: the K_MSEC(10) sleep in THREAD_REUSE_ACTIVE_THREAD_ID is still there (only long_running_thread changed to K_FOREVER) and is unnecessary since the reuse check is synchronous.

kernel/thread.c
ChidvilasYQC: Fixed both points from the reivew:
 - Double-unlock: removed the k_spin_unlock inside the if (t == new_thread) branch. check_if_tid_is_reused() now just sets a reused flag and breaks out of the loop, unlocks once unconditionally after the loop, then asserts on !reused. So the lock is released exactly once regardless of whether reuse is detected.
- K_MSEC(10) sleep: removed - the reuse check in z_setup_new_thread() runs synchronously on the second k_thread_create() call, so there's nothing to wait for.

tests/kernel/threads/thread_error_case/src/main.c
```diff
@@ -288,6 +311,22 @@ ZTEST_USER(thread_error_case, test_thread_create_stack_overflow)
...
+ *        reused, while the thread pointer is in use
+ *
+ * Verifies that thread pointer that is passed to thread create is
+ * not the only which is already created and in use
+ *
+ * @ingroup kernel_thread_tests
+ *
+ * @see k_thread_create()
+ */
+ZTEST_USER(thread_error_case, test_thread_reuse_active_thread_id)
```
npitre: Use `ZTEST` rather than `ZTEST_USER` here so the worker runs in supervisor mode and the create actually reaches `z_setup_new_thread()`. The `if (k_is_user_context())` block at lines 164-166 then becomes dead and can go, as can the three `reuse_*` arguments added to `k_thread_access_grant()` in `thread_grant_setup()`.

kernel/thread.c
```diff
@@ -873,6 +873,28 @@ static inline void init_thread_usage(struct k_thread *thread)
 #endif /* CONFIG_SCHED_THREAD_USAGE */
 }
 
+static inline void check_if_tid_is_reused(struct k_thread *new_thread)
```
npitre: Naming: the helpers around this one are `init_thread_*()` / `assert_*()` (see `assert_thread_coherence()`, `assert_switch_handle()`), so `assert_thread_not_reused()` would fit better, and `tid` is not a term used in this file. The two `#endif`s also want their `/* CONFIG_... */` comments like every other helper in this block, and the feature-disabled path needs an `ARG_UNUSED(new_thread)`.

npitre (commented):
The added test never reaches the added assert. `thread_error_case` is filtered on `CONFIG_ARCH_HAS_USERSPACE` and sets `CONFIG_TEST_USERSPACE=y`, which force-enables `CONFIG_USERSPACE` (`TEST_ENABLE_USERSPACE` selects it), so `test_thread_reuse_active_thread_id` always runs in user mode. The second `k_thread_create()` is then rejected by `K_SYSCALL_OBJ_NEVER_INIT()` in `z_vrfy_k_thread_create()`, because the first create already ran `k_object_init()` on `reuse_tdata`, and `z_setup_new_thread()` is never entered. Running the suite on qemu_x86 at 87420c9:

```
START - test_thread_reuse_active_thread_id
E: 0x137000 k_thread in use
E: syscall z_vrfy_k_thread_create failed check: access denied
```

That is the same verifier check `test_thread_create_uninit` already covers, so the new kernel code ends up with no coverage in any configuration CI builds. Making the case supervisor-mode does reach it:

```
ASSERTION FAIL [!reused] @ ZEPHYR_BASE/kernel/thread.c:893
	thread 0x137000 is already in the running list
```

and the suite still passes, so `ztest_set_fault_valid(true)` already handles the assert panic and no assert hook is needed.

The double-unlock and the `K_MSEC(10)` sleep are both resolved.

tests/kernel/threads/thread_error_case/src/main.c
ChidvilasYQC: Switched to `ZTEST` and dropped `reuse_tdata`/`reuse_stack1`/`reuse_stack2` from `k_thread_access_grant()` in `thread_grant_setup()`. Also removing the now-dead if `(k_is_user_context())` block.

kernel/thread.c
ChidvilasYQC: Renamed to `assert_thread_not_reused` to match the `assert_*` helpers, added the `/* CONFIG_ASSERT */` / `/* CONFIG_THREAD_MONITOR */` labels on the `#endif`s, and added `ARG_UNUSED(new_thread)` for the disabled path.

npitre (dismissed):
Needs a rebase before it can merge

ChidvilasYQC:
@npitre, I have rebased main, please review and approve.

npitre (approved):

mbolivar (approved):

PR #101331, opened 2025-12-19
drivers: spi: stm32: add RTIO-DMA support

juickar:
This PR adds the support of RTIO SPI transactions using DMA, it also adds a test case for the loopback test.

Tested on our test bench using `H7` and `non-H7` SPI compatible boards

### DIFF
--- a/drivers/spi/spi_stm32.c
+++ b/drivers/spi/spi_stm32.c
@@ -194,15 +194,56 @@ static uint8_t bits2bytes(spi_operation_t operation)
  */
 static __aligned(32) uint32_t dummy_rx_tx_buffer __nocache;
 
+#ifdef CONFIG_SPI_RTIO
+static void spi_stm32_iodev_complete(const struct device *dev, int status);
+
+static void spi_rtio_dma_complete_transaction(const struct device *dev)
+{
+	struct spi_stm32_data *data = dev->data;
+	const struct spi_stm32_config *cfg = dev->config;
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	if (LL_SPI_IsActiveFlag_EOT(cfg->spi)) {
+		LL_SPI_DisableIT_EOT(cfg->spi);
+		LL_SPI_ClearFlag_EOT(cfg->spi);
+	}
+	/* Clear the interrupt flags */
+	if (cfg->fifo_enabled) {
+		LL_SPI_ClearFlag_TXTF(cfg->spi);
+		LL_SPI_ClearFlag_OVR(cfg->spi);
+	}
+#else
+	LL_SPI_DisableDMAReq_TX(cfg->spi);
+	LL_SPI_DisableDMAReq_RX(cfg->spi);
+#endif /* st_stm32h7_spi */
+
+	int err;
+
+	err = dma_stop(data->dma_rx.dma_dev, data->dma_rx.channel);
+	if (err != 0) {
+		LOG_DBG("Rx dma_stop failed with error %d", err);
+	}
+
+	err = dma_stop(data->dma_tx.dma_dev, data->dma_tx.channel);
+	if (err != 0) {
+		LOG_DBG("Tx dma_stop failed with error %d", err);
+	}
+
+	spi_stm32_iodev_complete(dev, 0);
+}
+
+#endif /* CONFIG_SPI_RTIO */
+
 /* This function is executed in the interrupt context */
 static void dma_callback(const struct device *dma_dev, void *arg, uint32_t channel, int status)
 {
 	ARG_UNUSED(dma_dev);
 
-	/* arg holds SPI DMA data
+	/* arg holds the SPI device
 	 * Passed in spi_stm32_dma_tx/rx_load()
 	 */
-	struct spi_stm32_data *data = arg;
+	const struct device *spi_dev = arg;
+	struct spi_stm32_data *data = spi_dev->data;
 
 	if (status < 0) {
 		LOG_ERR("DMA callback error with channel %d.", channel);
@@ -221,6 +262,43 @@ static void dma_callback(const struct device *dma_dev, void *arg, uint32_t chann
 		}
 	}
 
+#ifdef CONFIG_SPI_RTIO
+	/* If both TX and RX are done, complete the transfer */
+	if ((data->status_flags & SPI_STM32_DMA_TX_DONE_FLAG) != 0 &&
+	    (data->status_flags & SPI_STM32_DMA_RX_DONE_FLAG) != 0) {
+
+		const struct spi_stm32_config *cfg = spi_dev->config;
+
+		if ((data->status_flags & SPI_STM32_DMA_ERROR_FLAG) != 0U) {
+			spi_stm32_iodev_complete(spi_dev, -EIO);
+			return;
+		}
+
+#ifdef SPI_SR_FTLVL
+		while (LL_SPI_GetTxFIFOLevel(cfg->spi) > 0) {
+		}
+#endif /* SPI_SR_FTLVL */
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		if (cfg->fifo_enabled && !LL_SPI_IsActiveFlag_EOT(cfg->spi)) {
+			LL_SPI_EnableIT_EOT(cfg->spi);
+			return;
+		}
+#else
+#ifdef CONFIG_SPI_STM32_ERRATA_BUSY
+		WAIT_FOR(!ll_spi_dma_busy(cfg->spi), CONFIG_SPI_STM32_BUSY_FLAG_TIMEOUT, k_yield());
+#else
+		while (ll_spi_dma_busy(cfg->spi) && LL_SPI_IsEnabled(cfg->spi)) {
+			if (LL_SPI_HALF_DUPLEX_RX == LL_SPI_GetTransferDirection(cfg->spi)) {
+				ll_disable_spi(cfg->spi);
+			}
+		}
+#endif /* CONFIG_SPI_STM32_ERRATA_BUSY */
+#endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)*/
+
+		spi_rtio_dma_complete_transaction(spi_dev);
+	}
+#endif /* CONFIG_SPI_RTIO */
 	k_sem_give(&data->status_sem);
 }
 
@@ -271,8 +349,8 @@ static int spi_stm32_dma_tx_load(const struct device *dev, const uint8_t *buf, s
 
 	/* direction is given by the DT */
 	stream->dma_cfg.head_block = blk_cfg;
-	/* give the dma channel data as arg, as the callback comes from the dma */
-	stream->dma_cfg.user_data = data;
+	/* Pass SPI device as callback context */
+	stream->dma_cfg.user_data = (void *)dev;
 	/* pass our client origin to the dma: data->dma_tx.dma_channel */
 	ret = dma_config(data->dma_tx.dma_dev, data->dma_tx.channel, &stream->dma_cfg);
 	/* the channel is the actual stream from 0 */
@@ -327,7 +405,7 @@ static int spi_stm32_dma_rx_load(const struct device *dev, uint8_t *buf, size_t 
 
 	/* direction is given by the DT */
 	stream->dma_cfg.head_block = blk_cfg;
-	stream->dma_cfg.user_data = data;
+	stream->dma_cfg.user_data = (void *)dev;
 
 
 	/* pass our client origin to the dma: data->dma_rx.channel */
@@ -666,6 +744,8 @@ static void spi_stm32_cs_control(const struct device *dev, bool on __maybe_unuse
 #endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32_spi_subghz) */
 }
 
+#if !(defined(CONFIG_SPI_RTIO) && defined(CONFIG_SPI_STM32_DMA))
+
 static void spi_stm32_msg_start(const struct device *dev, bool is_rx_empty)
 {
 	const struct spi_stm32_config *cfg = dev->config;
@@ -725,6 +805,7 @@ static void spi_stm32_msg_start(const struct device *dev, bool is_rx_empty)
 #endif /* CONFIG_SOC_SERIES_STM32H7X */
 #endif /* CONFIG_SPI_STM32_INTERRUPT */
 }
+#endif /* !(CONFIG_SPI_RTIO && CONFIG_SPI_STM32_DMA) */
 
 #ifdef CONFIG_SPI_RTIO
 /* Forward declaration for RTIO handlers conveniance */
@@ -743,7 +824,8 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 				      const uint8_t *tx_buf, uint8_t *rx_buf, uint32_t buf_len)
 {
 	struct spi_stm32_data *data = dev->data;
-	uint32_t size = buf_len / bits2bytes(config->operation);
+	const uint8_t dfs = bits2bytes(config->operation);
+	const uint32_t size = buf_len / dfs;
 
 	const struct spi_buf current_tx = {.buf = NULL, .len = size};
 	const struct spi_buf current_rx = {.buf = NULL, .len = size};
@@ -764,9 +846,10 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 	ctx->recv_frames = 0;
 #endif /* CONFIG_SPI_SLAVE */
 
+	__maybe_unused const struct spi_stm32_config *cfg = dev->config;
+	__maybe_unused SPI_TypeDef *spi = cfg->spi;
+
 #if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
-	const struct spi_stm32_config *cfg = dev->config;
-	SPI_TypeDef *spi = cfg->spi;
 
 	if (cfg->fifo_enabled && SPI_OP_MODE_GET(config->operation) == SPI_OP_MODE_MASTER) {
 		if (LL_SPI_IsEnabled(spi)) {
@@ -780,7 +863,77 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 	}
 #endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
 
+#ifdef CONFIG_SPI_STM32_DMA
+
+#ifdef CONFIG_DCACHE
+
+	if ((tx_buf != NULL && !stm32_buf_in_nocache((uintptr_t)tx_buf, buf_len)) ||
+	    (rx_buf != NULL && !stm32_buf_in_nocache((uintptr_t)rx_buf, buf_len))) {
+		LOG_ERR("SPI DMA transfers not supported on cached memory");
+		spi_stm32_iodev_complete(dev, -EINVAL);
+		return;
+	}
+#endif /* CONFIG_DCACHE */
+
+	size_t dma_len;
+	int ret = 0;
+
+	struct dma_config *rx_cfg = &data->dma_rx.dma_cfg;
+	struct dma_config *tx_cfg = &data->dma_tx.dma_cfg;
+
+	rx_cfg->source_data_size = rx_cfg->source_burst_length = dfs;
+	rx_cfg->dest_data_size   = rx_cfg->dest_burst_length   = dfs;
+	tx_cfg->source_data_size = tx_cfg->source_burst_length = dfs;
+	tx_cfg->dest_data_size   = tx_cfg->dest_burst_length   = dfs;
+
+	uint32_t transfer_dir = LL_SPI_GetTransferDirection(spi);
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	/* set request before enabling (else SPI CFG1 reg is write protected) */
+	spi_dma_enable_requests(spi);
+
+	LL_SPI_Enable(spi);
+
+	/* In half-duplex rx mode, start transfer after
+	 * setting DMA configurations
+	 */
+	if (transfer_dir != LL_SPI_HALF_DUPLEX_RX && LL_SPI_GetMode(spi) == LL_SPI_MODE_MASTER) {
+		LL_SPI_StartMasterTransfer(spi);
+	}
+#else
+	LL_SPI_Enable(spi);
+#endif /* st_stm32h7_spi */
+
+	/* Assert CS before enabling transfer */
+	spi_stm32_cs_control(dev, true);
+	data->status_flags = 0;
+
+	if (transfer_dir == LL_SPI_FULL_DUPLEX) {
+		dma_len = spi_context_max_continuous_chunk(&data->ctx);
+		ret = spi_dma_move_buffers(dev, dma_len);
+	} else if (transfer_dir == LL_SPI_HALF_DUPLEX_TX) {
+		dma_len = data->ctx.tx_len;
+		ret = spi_dma_move_tx_buffers(dev, dma_len);
+	} else {
+		dma_len = data->ctx.rx_len;
+		ret = spi_dma_move_rx_buffers(dev, dma_len);
+	}
+	if (ret != 0) {
+		LOG_ERR("Failed to start SPI DMA transfer");
+		spi_stm32_iodev_complete(dev, -EIO);
+	}
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	if (transfer_dir == LL_SPI_HALF_DUPLEX_RX && LL_SPI_GetMode(spi) == LL_SPI_MODE_MASTER) {
+		LL_SPI_StartMasterTransfer(spi);
+	}
+#else
+	/* toggle the DMA request to restart the transfer */
+	spi_dma_enable_requests(spi);
+
+#endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
+#else
 	spi_stm32_msg_start(dev, rx_buf == NULL);
+#endif /* CONFIG_SPI_STM32_DMA */
 }
 
 static void spi_stm32_iodev_start(const struct device *dev)
@@ -1000,6 +1153,15 @@ static void spi_stm32_isr(const struct device *dev)
 			return;
 		}
 	}
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) && defined(CONFIG_SPI_RTIO) &&                       \
+	defined(CONFIG_SPI_STM32_DMA)
+	if (LL_SPI_IsActiveFlag_EOT(spi)) {
+		spi_rtio_dma_complete_transaction(dev);
+		k_sem_give(&data->status_sem);
+		return;
+	}
+#endif /* st_stm32h7_spi && CONFIG_SPI_RTIO && CONFIG_SPI_STM32_DMA */
 
 	err = spi_stm32_get_err(spi);
 	if (err) {
@@ -1464,7 +1626,7 @@ end:
 	return ret;
 }
 
-#ifdef CONFIG_SPI_STM32_DMA
+#if defined(CONFIG_SPI_STM32_DMA) && !defined(CONFIG_SPI_RTIO)
 
 static int wait_dma_rx_tx_done(const struct device *dev)
 {
@@ -1762,21 +1924,21 @@ end:
 
 	return ret;
 }
-#endif /* CONFIG_SPI_STM32_DMA */
+#endif /* CONFIG_SPI_STM32_DMA && !CONFIG_SPI_RTIO */
 
 static int spi_stm32_transceive(const struct device *dev,
 				const struct spi_config *config,
 				const struct spi_buf_set *tx_bufs,
 				const struct spi_buf_set *rx_bufs)
 {
-#ifdef CONFIG_SPI_STM32_DMA
+#if defined(CONFIG_SPI_STM32_DMA) && !defined(CONFIG_SPI_RTIO)
 	struct spi_stm32_data *data = dev->data;
 
 	if ((data->dma_tx.dma_dev != NULL) && (data->dma_rx.dma_dev != NULL)) {
 		return transceive_dma(dev, config, tx_bufs, rx_bufs,
 				      false, NULL, NULL);
 	}
-#endif /* CONFIG_SPI_STM32_DMA */
+#endif /* CONFIG_SPI_STM32_DMA && !CONFIG_SPI_RTIO */
 	return transceive(dev, config, tx_bufs, rx_bufs, false, NULL, NULL);
 }
 
--- a/drivers/spi/Kconfig.stm32
+++ b/drivers/spi/Kconfig.stm32
@@ -26,6 +26,8 @@ config SPI_STM32_DMA
 	help
 	  Enable the SPI DMA mode for SPI instances
 	  that enable dma channels in their device tree node.
+	  This can be used with the SPI_RTIO config to have SPI
+	  transactions using DMA + RTIO
 
 config SPI_STM32_ERRATA_BUSY
 	bool
--- a/tests/drivers/spi/spi_loopback/overlay-stm32-spi-rtio-dma.conf
+++ b/tests/drivers/spi/spi_loopback/overlay-stm32-spi-rtio-dma.conf
@@ -1,0 +1,8 @@
+#
+# Copyright (c) 2026 STMicroelectronics
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+CONFIG_SPI_RTIO=y
+CONFIG_SPI_STM32_DMA=y
+CONFIG_SPI_ASYNC=n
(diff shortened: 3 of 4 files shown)


teburd (commented):
Suppose we toggle power here. Is the spi fifo pair flushed on the completion or is only DMA done?

drivers/spi/spi_stm32.c
```diff
@@ -215,6 +220,24 @@ static void dma_callback(const struct device *dma_dev, void *arg,
 		}
 	}
 
+#ifdef CONFIG_SPI_RTIO
+	/* If both TX and RX are done, complete the transfer */
+	if ((spi_dma_data->status_flags & SPI_STM32_DMA_TX_DONE_FLAG) &&
+		(spi_dma_data->status_flags & SPI_STM32_DMA_RX_DONE_FLAG)) {
+
```
etienne-lms: For consistency in indentation + explicit boolean tests:
```suggestion
	if ((spi_dma_data->status_flags & SPI_STM32_DMA_TX_DONE_FLAG) != 0 &&
	    (spi_dma_data->status_flags & SPI_STM32_DMA_RX_DONE_FLAG) != 0) {
```

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +711,69 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 	}
 #endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
 
+#ifdef CONFIG_SPI_STM32_DMA
+	int ret = 0;
+
+#if !DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	const struct spi_stm32_config *cfg = dev->config;
+	SPI_TypeDef *spi = cfg->spi;
+#endif /* !DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
```
etienne-lms: `cfg` and `spi` being defined at 2 places looks a bit weird.
I would rather suggest to always define them at function entry with a `__maybe_unused ` attribute.

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +711,69 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+	ret = spi_stm32_dma_tx_load(dev, tx_buf, buf_len);
+	if (ret != 0) {
+		LOG_DBG("DMA TX load failed with error %d", ret);
+		spi_stm32_cs_control(dev, false);
+		return;
+	}
+	/* Assert CS before enabling transfer */
+	spi_stm32_cs_control(dev, true);
+
+
```
etienne-lms: Could you remove this empty extra line?

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +711,69 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+	LL_SPI_Enable(spi);
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+
+	/* Start master transfer if required */
+	if (LL_SPI_GetMode(spi) == LL_SPI_MODE_MASTER) {
+		LL_SPI_StartMasterTransfer(spi);
+		while (!LL_SPI_IsActiveMasterTransfer(spi)) {
+		/* NOP */
```
etienne-lms: Nitpicking on indentation:
```suggestion
			/* NOP */
```
or even, remove this line.

wouterh:
I have tested this for my use case:

- custom board (L4 based) with ST iis3dwb accelerometer
- device runtime power management enabled

It works *sometimes* and if it works, I see nice DMA transfers in my logic analyzer and the code is able to keep up with the accelerometer (3-axis data @ 26kHz).

However, it crashes most of the time. I think the crashes have something to with power management.

A couple of crashes, I've seen:

```
[00:00:07.856,000] <err> os: ***** BUS FAULT *****
[00:00:07.856,000] <err> os:   Precise data bus error
[00:00:07.856,000] <err> os:   BFAR Address: 0xf7ff9805
[00:00:07.856,000] <err> os: r0/a1:  0xf7ff9805  r1/a2:  0x08010f11  r2/a3:  0x2001f778
[00:00:07.856,000] <err> os: r3/a4:  0x2001f778 r12/ip:  0x00000000 r14/lr:  0x080086db
[00:00:07.856,000] <err> os:  xpsr:  0x01000200
[00:00:07.856,000] <err> os: s[ 0]:  0x080032a0  s[ 1]:  0x080032a0  s[ 2]:  0x000f423f  s[ 3]:  0x00000000
[00:00:07.856,000] <err> os: s[ 4]:  0x00000000  s[ 5]:  0x00000000  s[ 6]:  0x080132a0  s[ 7]:  0x00000003
[00:00:07.856,000] <err> os: s[ 8]:  0x00000000  s[ 9]:  0x200009c0  s[10]:  0x200009e8  s[11]:  0x00000000
[00:00:07.856,000] <err> os: s[12]:  0x20000080  s[13]:  0x0800fb9d  s[14]:  0x080132a0  s[15]:  0x00000002
[00:00:07.856,000] <err> os: fpscr:  0x00000000
[00:00:07.856,000] <err> os: Faulting instruction address (r15/pc): 0x08011faa
[00:00:07.856,000] <err> os: >>> ZEPHYR FATAL ERROR 25: Unknown error on CPU 0
```
Faulting instruction points to `include/zephyr/device.h:733`

```
[00:00:58.732,000] <err> os: ***** USAGE FAULT *****
[00:00:58.732,000] <err> os:   Illegal use of the EPSR
[00:00:58.732,000] <err> os: r0/a1:  0x0003bbb6  r1/a2:  0x00000000  r2/a3:  0x40007c00
[00:00:58.732,000] <err> os: r3/a4:  0x00000000 r12/ip:  0x00000000 r14/lr:  0x08009bf1
[00:00:58.732,000] <err> os:  xpsr:  0x00000000
[00:00:58.732,000] <err> os: s[ 0]:  0x00000000  s[ 1]:  0x00000000  s[ 2]:  0x2001f770  s[ 3]:  0x080124f1
[00:00:58.732,000] <err> os: s[ 4]:  0x01000003  s[ 5]:  0x0801f2bb  s[ 6]:  0x0801f320  s[ 7]:  0x08000200
[00:00:58.732,000] <err> os: s[ 8]:  0x00000000  s[ 9]:  0x00000000  s[10]:  0x080132a0  s[11]:  0x00000000
[00:00:58.732,000] <err> os: s[12]:  0x00000000  s[13]:  0x200009c0  s[14]:  0x200009e8  s[15]:  0x00000000
[00:00:58.732,000] <err> os: fpscr:  0x20000080
[00:00:58.732,000] <err> os: Faulting instruction address (r15/pc): 0x080132a0
[00:00:58.732,000] <err> os: >>> ZEPHYR FATAL ERROR 35: Unknown error on CPU 0
```
Faulting instruction points to `drivers/power_domain/power_domain_gpio.c:139`

```
[00:03:10.752,000] <err> os: ***** BUS FAULT *****
[00:03:10.752,000] <err> os:   Instruction bus error
[00:03:10.752,000] <err> os: r0/a1:  0x2001f760  r1/a2:  0x0000000e  r2/a3:  0x00000000
[00:03:10.752,000] <err> os: r3/a4:  0x2001f71c r12/ip:  0x08007cdd r14/lr:  0x0a404fed
[00:03:10.752,000] <err> os:  xpsr:  0x61000000
[00:03:10.752,000] <err> os: s[ 0]:  0x0000000a  s[ 1]:  0x2001f714  s[ 2]:  0x00000004  s[ 3]:  0x00000000
[00:03:10.752,000] <err> os: s[ 4]:  0x00007ae2  s[ 5]:  0x00000000  s[ 6]:  0x08013830  s[ 7]:  0x0000000e
[00:03:10.752,000] <err> os: s[ 8]:  0x080131ec  s[ 9]:  0x2001f7b8  s[10]:  0x0000000a  s[11]:  0x00000000
[00:03:10.752,000] <err> os: s[12]:  0x00000000  s[13]:  0x00000000  s[14]:  0x00000000  s[15]:  0xe000ed00
[00:03:10.752,000] <err> os: fpscr:  0x20002838
[00:03:10.752,000] <err> os: Faulting instruction address (r15/pc): 0x0a404420
```
`arm-none-eabi-addr2line` can't figure out that one.

Could this be related to the subtle races that @teburd mentioned?

I also have a quick and dirty RTIO+DMA implementation (see https://github.com/versasense/zephyr/tree/rtio-dma) that I only tested on my board. I haven't seen any crashes with that. In that implementation, I do wait for TXE before marking completion.

petejohanson-adi:
> > > Suppose we toggle power here. Is the spi fifo pair flushed on the completion or is only DMA done?
> > 
> > 
> > Only DMA is done
> 
> This should _really_ only be marking the completion when the bus peripheral says its done if possible. Otherwise there could be subtle races on power/clock gating happening before the data actually was shifted out.

Yeah, that was my expectation of this as well. I happened to be working on this in parallel before this PR was up, and my implementation uses a timer, if needed, to check for TX fifo levels before reporting any TX operation as completed, to ensure we are actually "done" before continuing the next operation in the sequence. The timer is needed, AFAIK, since some STM32 SoCs don't have an interrupt for TX FIFO *empty* and the TXE, if there, is just to indicate there's some amount of space in the TX FIFO for more to be added.

You can see my version (which still needs cleaning up, and is more PoC level at this point) at https://github.com/zephyrproject-rtos/zephyr/commit/c4bc4ad82f8bc422d29d332362be1aca73a869d7

petejohanson-adi:
With DMA enabled, my sensor streaming testing just fails to init:

```
[00:00:00.005,000] <err> ADXL362: wrong part_id: 0
uart:~$ 
```

If I disable DMA, and just use streaming with just interrupts, it instead faults:

```
uart:~$ sensor stream adxl362@0 on fifo_wm incl 
Enabling stream...
Trigger (1 / data_ready) detected
Trigger (10 / fifo_wm) detected
[00:15:24.986,000] <err> os: ***** USAGE FAULT *****
[00:15:24.986,000] <err> os:   Illegal use of the EPSR
[00:15:24.986,000] <err> os: r0/a1:  0x00000000  r1/a2:  0x24004060  r2/a3:  0x24004064
[00:15:24.986,000] <err> os: r3/a4:  0x00000000 r12/ip:  0x0000d6d8 r14/lr:  0x080011a9
[00:15:24.986,000] <err> os:  xpsr:  0x60000000
[00:15:24.986,000] <err> os: Faulting instruction address (r15/pc): 0x00000000
[00:15:24.986,000] <err> os: >>> ZEPHYR FATAL ERROR 35: Unknown error on CPU 0
[00:15:24.986,000] <err> os: Current thread: 0x24001868 (sensor_shell_processing_tid)
[00:15:25.007,000] <err> os: Halting system
``` 

Both setups are working fine in the implementation I had linked a few minutes ago, with the same configs.

petejohanson-adi:
> I also have a quick and dirty RTIO+DMA implementation (see https://github.com/versasense/zephyr/tree/rtio-dma) that I only tested on my board. I haven't seen any crashes with that. In that implementation, I do wait for TXE before marking completion.

Do note, that TXE doesn't really mean empty, despite that naming. From https://www.st.com/resource/en/reference_manual/rm0434-multiprotocol-wireless-32bit-mcu-armbased-cortexm4-with-fpu-bluetooth-lowenergy-and-802154-radio-solution-stmicroelectronics.pdf section 35.4.10, for instance:

> Tx buffer empty flag (TXE)
> The TXE flag is set when transmission TXFIFO has enough space to store data to send.
TXE flag is linked to the TXFIFO level. The flag goes high and stays high until the TXFIFO
level is lower or equal to 1/2 of the FIFO depth. An interrupt can be generated if the TXEIE
bit in the SPIx_CR2 register is set. The bit is cleared automatically when the TXFIFO level
becomes greater than 1/2.
Rx buffer not empty (RXNE)
The RXNE flag is set depending on the FRXTH bit value 

Which is why my version is actually checking if the TX FIFO is actually empty, using an additional timer if needed, since TXE *isn't* actually enough to be sure a given RTIO TX operation has been fully sent.

AFAICT, that's the only way to be 100% sure that after a DMA TX is complete, that the SPI peripheral has actually sent *all* the data

juickar:
> With DMA enabled, my sensor streaming testing just fails to init:
> 
> ```
> [00:00:00.005,000] <err> ADXL362: wrong part_id: 0
> uart:~$ 
> ```
> 
> If I disable DMA, and just use streaming with just interrupts, it instead faults:
> 
> ```
> uart:~$ sensor stream adxl362@0 on fifo_wm incl 
> Enabling stream...
> Trigger (1 / data_ready) detected
> Trigger (10 / fifo_wm) detected
> [00:15:24.986,000] <err> os: ***** USAGE FAULT *****
> [00:15:24.986,000] <err> os:   Illegal use of the EPSR
> [00:15:24.986,000] <err> os: r0/a1:  0x00000000  r1/a2:  0x24004060  r2/a3:  0x24004064
> [00:15:24.986,000] <err> os: r3/a4:  0x00000000 r12/ip:  0x0000d6d8 r14/lr:  0x080011a9
> [00:15:24.986,000] <err> os:  xpsr:  0x60000000
> [00:15:24.986,000] <err> os: Faulting instruction address (r15/pc): 0x00000000
> [00:15:24.986,000] <err> os: >>> ZEPHYR FATAL ERROR 35: Unknown error on CPU 0
> [00:15:24.986,000] <err> os: Current thread: 0x24001868 (sensor_shell_processing_tid)
> [00:15:25.007,000] <err> os: Halting system
> ```
> 
> Both setups are working fine in the implementation I had linked a few minutes ago, with the same configs.

I'm currently working on adding a check to make sure TX Fifo is empty. These errors happened with my changes ?

drivers/spi/spi_stm32.c
juickar: done

drivers/spi/spi_stm32.c
juickar: done

drivers/spi/spi_stm32.c
```diff
@@ -187,16 +187,21 @@ static uint8_t bits2bytes(spi_operation_t operation)
  */
 static __aligned(32) uint32_t dummy_rx_tx_buffer __nocache;
 
+#if defined(CONFIG_SPI_STM32_DMA) && defined(CONFIG_SPI_RTIO)
```
erwango: Update description of `SPI_STM32_DMA` symbol explaining that it is compatible with `SPI_RTIO`.

drivers/spi/spi_stm32.c
```diff
@@ -215,6 +220,60 @@ static void dma_callback(const struct device *dma_dev, void *arg,
...
+		int err = dma_stop(spi_dma_data->dma_rx.dma_dev, spi_dma_data->dma_rx.channel);
+
+		if (err != 0) {
+			LOG_DBG("Rx dma_stop failed with error %d", err);
+		}
+		err = dma_stop(spi_dma_data->dma_tx.dma_dev, spi_dma_data->dma_tx.channel);
+		if (err != 0) {
+			LOG_DBG("Tx dma_stop failed with error %d", err);
+		}
+		spi_stm32_iodev_complete(spi_dev, status);
```
erwango:
```suggestion
		int err = dma_stop(spi_dma_data->dma_rx.dma_dev, spi_dma_data->dma_rx.channel);
		if (err != 0) {
			LOG_DBG("Rx dma_stop failed with error %d", err);
		}

				err = dma_stop(spi_dma_data->dma_tx.dma_dev, spi_dma_data->dma_tx.channel);
		if (err != 0) {
			LOG_DBG("Tx dma_stop failed with error %d", err);
		}
		spi_stm32_iodev_complete(spi_dev, status);
```

drivers/spi/spi_stm32.c
```diff
@@ -215,6 +220,60 @@ static void dma_callback(const struct device *dma_dev, void *arg,
...
+#ifdef CONFIG_SPI_RTIO
+	/* If both TX and RX are done, complete the transfer */
+	if ((spi_dma_data->status_flags & SPI_STM32_DMA_TX_DONE_FLAG) != 0 &&
+	    (spi_dma_data->status_flags & SPI_STM32_DMA_RX_DONE_FLAG) != 0) {
+
+		if ((spi_dma_data->status_flags & SPI_STM32_DMA_ERROR_FLAG) != 0U) {
+			spi_stm32_iodev_complete(spi_dev, -EIO);
+			return;
+		}
+		const struct spi_stm32_config *cfg = spi_dev->config;
```
erwango: Break a line before this

erwango (changes requested):

drivers/spi/spi_stm32.c
```diff
@@ -215,6 +220,60 @@ static void dma_callback(const struct device *dma_dev, void *arg,
...
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+			uint32_t width = SPI_WORD_SIZE_GET(spi_dma_data->ctx.config->operation);
+			/* The TXC flag is not raised at the end of 9, 17 or 25
+			 * bit transfer, so disable the SPI in these cases to avoid being stuck.
+			 */
+			if ((width == 9U) || (width == 17U) || (width == 25U)) {
+				k_busy_wait(1000);
+				ll_func_disable_spi(cfg->spi);
+			}
+#endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
```
petejohanson-adi: So... I really have to question the design here.. The DMA callback occurs in an interrupt context, and these busy loops seem like a really bad idea to just duplicate in here. You can see in the implementation I have in #101971 that instead I use a secondary timer that will fire and check for the FIFO levels, DMA being done, etc. to ensure we can not block, but also not complete the transaction until we're really ready to.

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +746,88 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 	}
 #endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
 
+#ifdef CONFIG_SPI_STM32_DMA
```
petejohanson-adi, line 866: This whole block is duplicated, roughly, from elsewhere in the code. Ideally should be extracted and re-used.

This is also missing checks for the buffers being in cached memory, so this silently fails if they are, instead of checking/warning/exiting earlier.

I *was* able to track that down and get this working as well for my sensor streaming test case, once I disabled DCACHE, but it would be good to have this check (which properly catches this case in `tranceive_dma`.

drivers/spi/spi_stm32.c
teburd: Unfortunate if the hardware does not signal fifo empty through an interrupt, but in that case yes... to make it asynchronous and not block in the ISR context your options are basically going to be...

* Schedule a timer to check this in the next tick or whatever you think is appropriate to poll for the completion
* Offload to a work queue/thread to poll for completion

wouterh: There is the TXE interrupt that I use in [my implementation](https://github.com/versasense/zephyr/commit/509d6033623b006475f6701c68884a3443093ac0). I still have the busy loop in there because I didn't have time to figure out what the BSY errata is about. It might not be needed anymore.

@petejohanson-adi regarding your timer-based implementation. I see a timer timeout of 100µs in there. Isn't that a long time for a bus running at several MHz?
I am a bit tight on time this week, but I'll see if I can squeeze in a test of your code on my use case...

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +746,88 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+		LL_SPI_EnableDMAReq_TX(cfg->spi);
+	} else {
+		LL_SPI_EnableDMAReq_RX(cfg->spi);
+	}
+
+	LL_SPI_Enable(cfg->spi);
+
+	/* In half-duplex rx mode, start transfer after setting DMA configurations */
+	if (transfer_dir != LL_SPI_HALF_DUPLEX_RX &&
+		LL_SPI_GetMode(cfg->spi) == LL_SPI_MODE_MASTER) {
```
etienne-lms:
```suggestion
	    LL_SPI_GetMode(cfg->spi) == LL_SPI_MODE_MASTER) {

File: drivers/spi/spi_stm32.c [hunk 11]
```diff
@@ -682,7 +746,88 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+
+	if (ret != 0) {
+		spi_stm32_cs_control(dev, false);
+		spi_stm32_iodev_complete(dev, ret);
+		return;
+	}
+
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	if (transfer_dir == LL_SPI_HALF_DUPLEX_RX &&
+		LL_SPI_GetMode(spi) == LL_SPI_MODE_MASTER) {
```
etienne-lms:
```suggestion
	    LL_SPI_GetMode(spi) == LL_SPI_MODE_MASTER) {

File: drivers/spi/spi_stm32.c (hunk 8 above)
Comment (juickar) [reply drivers/spi/spi_stm32.c]: and maybe for the h7 compatible boards we could use the EOT interrupts:
see [at 50.11.6](https://www.st.com/resource/en/reference_manual/dm00314099.pdf#page=2225)
> 3 EOT: end of transfer
EOT is set by hardware as soon as a full transfer is complete, that is when TSIZE number of
data have been transmitted and/or received on the SPI. EOT is cleared by software write 1 to
EOTC bit at SPI_IFCR.
EOT flag triggers an interrupt if EOTIE bit is set.
If DXP flag is used until TXTF flag is set and DXPIE is cleared, EOT can be used to
download the last packets contained into RxFIFO in one-shot.
0: transfer is on-going or not started
1: transfer complete
In master, EOT event terminates the data transaction and handles SS output optionally. When
CRC is applied, the EOT event is extended over the CRC frame transaction.

since the TXE interrupt does not exist for these boards.

**Comment by JarmouniA** [comment]
@juickar You may want to take a look at #102014, I don't know if it will have a direct impact on this PR though.

File: drivers/spi/spi_stm32.c (hunk 8 above)
Comment (petejohanson-adi) [reply drivers/spi/spi_stm32.c]: > There is the TXE interrupt that I use in [my implementation](https://github.com/versasense/zephyr/commit/509d6033623b006475f6701c68884a3443093ac0). I still have the busy loop in there because I didn't have time to figure out what the BSY errata is about. It might not be needed anymore.

See my comment here: https://github.com/zephyrproject-rtos/zephyr/pull/101331#issuecomment-3715928981

TXE on the couple reference manuals I checked for different STM32 SoCs is a misnomer, and doesn't actually mean completely empty, just "TX FiFO has room for more", so IMHO isn't enough to mark the transaction as completed.

On platforms where OET is available, that *does* seem to match expectations, and could be used directly without any timer/polling approaches.

> 
> @petejohanson-adi regarding your timer-based implementation. I see a timer timeout of 100µs in there. Isn't that a long time for a bus running at several MHz? I am a bit tight on time this week, but I'll see if I can squeeze in a test of your code on my use case...

Yes, this definitely needs better tuning, which is why I opened this just as a draft for discussion/comparison. On the sensor WG call today, we discussed implications for this, how it might be tuned, possible nicer integration via the ideas discussed in #86503 etc.

File: drivers/spi/spi_stm32.c (hunk 9 above)
Comment (juickar) [reply drivers/spi/spi_stm32.c:866]: Added helper functions and the DCACHE test

File: drivers/spi/spi_stm32.c (hunk 7 above)
Comment (juickar) [reply drivers/spi/spi_stm32.c]: done

File: drivers/spi/spi_stm32.c (hunk 6 above)
Comment (juickar) [reply drivers/spi/spi_stm32.c]: done

File: drivers/spi/spi_stm32.c (hunk 5 above)
Comment (juickar) [reply drivers/spi/spi_stm32.c]: done

**Comment by juickar** [comment]
The comments has been addressed, but the TX completion check still needs to be reworked.

File: drivers/spi/spi_stm32.c [hunk 12]
```diff
@@ -187,16 +187,77 @@ static uint8_t bits2bytes(spi_operation_t operation)
  */
 static __aligned(32) uint32_t dummy_rx_tx_buffer __nocache;
 
+static int spi_dma_move_buffers(const struct device *dev, uint32_t transfer_dir);
+static int spi_dma_move_tx_buffers(const struct device *dev, size_t dma_len);
+static int spi_dma_move_rx_buffers(const struct device *dev, size_t dma_len);
+
+static void spi_stm32_enable_dma_transfer( SPI_TypeDef *spi, uint32_t transfer_dir){
+	#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
```
etienne-lms:
```suggestion
#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)

File: drivers/spi/spi_stm32.c [hunk 13]
```diff
@@ -215,6 +276,62 @@ static void dma_callback(const struct device *dma_dev, void *arg,
...
+
+#ifdef CONFIG_SPI_STM32_ERRATA_BUSY
+		WAIT_FOR(!ll_func_spi_dma_busy(cfg->spi),
+				CONFIG_SPI_STM32_BUSY_FLAG_TIMEOUT,
+				k_yield());
+#else
+		/* wait until spi is no more busy (spi TX fifo is really empty) */
+		while (ll_func_spi_dma_busy(cfg->spi) && LL_SPI_IsEnabled(cfg->spi)) {
+#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+			uint32_t width = SPI_WORD_SIZE_GET(spi_dma_data->ctx.config->operation);
```
etienne-lms: Could you also add an empty line below this local variable definition?

drivers/spi/spi_stm32.c
```diff
@@ -215,6 +276,62 @@ static void dma_callback(const struct device *dma_dev, void *arg,
...
+		}
+#endif /* CONFIG_SPI_STM32_ERRATA_BUSY */
+
+#if !DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+		/* toggle the DMA transfer request */
+		LL_SPI_DisableDMAReq_TX(cfg->spi);
+		LL_SPI_DisableDMAReq_RX(cfg->spi);
+#endif /* ! st_stm32h7_spi */
+
+		int err = dma_stop(spi_dma_data->dma_rx.dma_dev, spi_dma_data->dma_rx.channel);
```
etienne-lms: Here also, add an empty line below. Or alternatively:
```c
		int err;

		err = dma_stop(spi_dma_data->dma_rx.dma_dev, spi_dma_data->dma_rx.channel);
		(...)

File: drivers/spi/spi_stm32.c [hunk 15]
```diff
@@ -636,8 +755,26 @@ static void spi_stm32_msg_start(const struct device *dev, bool is_rx_empty)
...
+	return buf->buf == NULL;
+}
+
+static bool spi_buf_set_in_nocache(const struct spi_buf *buf)
+{
+	if (!(is_dummy_buffer(buf)) &&
+		!stm32_buf_in_nocache((uintptr_t)buf->buf, buf->len)) {
+		return false;
+	}
+	return true;
```
etienne-lms:
```suggestion
	return (buf->buf == NULL) || stm32_buf_in_nocache((uintptr_t)buf->buf, buf->len);
```
That said, these helper functions may not be needed, see below.

drivers/spi/spi_stm32.c
```diff
@@ -682,7 +821,64 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
 	}
 #endif /* DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi) */
 
+#ifdef CONFIG_SPI_STM32_DMA
+
+#ifdef CONFIG_DCACHE
+const struct spi_buf tx = {.buf = (void *)tx_buf, .len = buf_len};
+const struct spi_buf rx = {.buf = (void *)rx_buf, .len = buf_len};
```
etienne-lms:
```suggestion
	const struct spi_buf tx = {.buf = (void *)tx_buf, .len = buf_len};
	const struct spi_buf rx = {.buf = (void *)rx_buf, .len = buf_len};

File: drivers/spi/spi_stm32.c [hunk 17]
```diff
@@ -682,7 +821,64 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+
+	if ((tx_buf != NULL && !spi_buf_set_in_nocache(&tx)) ||
+	    (rx_buf != NULL && !spi_buf_set_in_nocache(&rx))) {
```
etienne-lms: `spi_buf_set_in_nocache()` already addresses NULL pointers.
```suggestion
	if (!spi_buf_set_in_nocache(&tx)) || !spi_buf_set_in_nocache(&rx))) {
```
Alternatively, maybe call `stm32_buf_in_nocache()` striaght?
```c
	if ((tx_buf != NULL && !stm32_buf_in_nocache((uintptr_t)tx_buf, buf_len)) ||
	    (rx_buf != NULL && !stm32_buf_in_nocache((uintptr_t)rx_buf, buf_len))) {

File: drivers/spi/spi_stm32.c [hunk 18]
```diff
@@ -682,7 +821,64 @@ static void spi_stm32_iodev_msg_start(const struct device *dev, struct spi_confi
...
+	tx_cfg->dest_data_size   = tx_cfg->dest_burst_length   = word_size_bytes;
+
+	uint32_t transfer_dir = LL_SPI_GetTransferDirection(spi);
+
+	spi_stm32_enable_dma_transfer(spi, transfer_dir);
+
+	/* Assert CS before enabling transfer */
+	spi_stm32_cs_control(dev, true);
+
+	spi_stm32_move_dma_buffers(dev, transfer_dir);
```
etienne-lms:
```c
	ret = spi_stm32_move_dma_buffers(dev, transfer_dir);
	if (ret != 0) {
		...
	}
```

drivers/spi/spi_stm32.c
```diff
@@ -187,16 +187,77 @@ static uint8_t bits2bytes(spi_operation_t operation)
...
+	LL_SPI_Enable(spi);
+#endif /* st_stm32h7_spi */
+}
+
+static int spi_stm32_move_dma_buffers(const struct device *dev, uint32_t transfer_dir)
+{
+	struct spi_stm32_data *data = dev->data;
+	size_t dma_len;
+	int ret;
+	data->status_flags = 0;
```
etienne-lms: Add an empty line above to separate from local variable definitions.

drivers/spi/spi_stm32.c
```diff
@@ -187,16 +187,77 @@ static uint8_t bits2bytes(spi_operation_t operation)
...
+	#if DT_HAS_COMPAT_STATUS_OKAY(st_stm32h7_spi)
+	/* set request before enabling (else SPI CFG1 reg is write protected) */
+	if (transfer_dir == LL_SPI_FULL_DUPLEX) {
+		LL_SPI_EnableDMAReq_RX(spi);
+		LL_SPI_EnableDMAReq_TX(spi);
+	} else if (transfer_dir == LL_SPI_HALF_DUPLEX_TX) {
+		LL_SPI_EnableDMAReq_TX(spi);
+	} else {
+		LL_SPI_EnableDMAReq_RX(spi);
+	}
```
etienne-lms: This sequence is repeated 3 times. Maybe worth to factorize in a helper function (or not):
```c
static void enable_ll_spi_dma_request(SPI_TypeDef *spi, uint32_t transfer_dir)
{
	if (transfer_dir == LL_SPI_FULL_DUPLEX) {
		LL_SPI_EnableDMAReq_RX(spi);
		LL_SPI_EnableDMAReq_TX(spi);
	} else if (transfer_dir == LL_SPI_HALF_DUPLEX_TX) {
		LL_SPI_EnableDMAReq_TX(spi);
	} else {
		LL_SPI_EnableDMAReq_RX(spi);
	}
}

petejohanson-adi:
@juickar This does need a rebase to account for recent change in `main`.

wouterh:
> Despite my best efforts to disabling logging, etc to reduce any latency, I am *not* ever seeing our DMA callback getting invoked *before* the peripheral has managed to complete the actual data transfer. This certainly could use more testing on additional targets to *really* verify (which perhaps @juickar has done, and if so, excellent!), but my initial hesitation about the current "RTIO completion" logic is significantly reduced.
> 

These are full duplex transfers, I assume? I've been thinking about it and the RX DMA transfer can't be completed before the SPI transfer is completed and all data is actually received. The code checks completion of both TX and RX DMA transfers, which explains what you see, I guess? 

Half duplex transfers might need other logic?

petejohanson-adi:
> > Despite my best efforts to disabling logging, etc to reduce any latency, I am _not_ ever seeing our DMA callback getting invoked _before_ the peripheral has managed to complete the actual data transfer. This certainly could use more testing on additional targets to _really_ verify (which perhaps @juickar has done, and if so, excellent!), but my initial hesitation about the current "RTIO completion" logic is significantly reduced.
> 
> These are full duplex transfers, I assume? I've been thinking about it and the RX DMA transfer can't be completed before the SPI transfer is completed and all data is actually received. The code checks completion of both TX and RX DMA transfers, which explains what you see, I guess?


There's actually a mix here. And I agree on the RX side, that's simpler since it's impossible, AFAICT, for the DMA transfer to complete there without if first being completed on the SPI peripheral FIFO. But yes, the code is waiting for both DMA events to be completed, which *may* be superfluous for some RTIO msg types.

> 
> Half duplex transfers might need other logic?

It seems to work fine, but in theory may be able to be optimized to trigger slightly earlier for half duplex transfers.

juickar:
- Rebased
- Comments adressed
- Changed the check completion for the H7 compatible boards to use the EOT interrupt

petejohanson-adi (approved):
LGTM, and testing here with ADXL362 streaming in the shell works as expected.

erwango (approved):

tbursztyka (approved):

PR #101338, opened 2025-12-19
drivers: firmware: tisci: Add secure path support

Kronosblaster:
**Summary**
Add support for secure TISCI communication paths to enable usage of secure threads for communication.

**Changes:**
- Add is_secure flag to tisci_config for secure/non-secure path selection
- Implement tisci_msg_requires_secure_path() to identify messages requiring secure path
- Add secure header wrapping for outbound messages on secure path
- Add secure header parsing for inbound messages on secure path
- Improve error handling with proper semaphore cleanup in error paths
- Add ti,is-secure device tree property to indicate secure host entities

**Testing**
Change the host-id and thread ids in the core specific dt file and run the /tests/drivers/tisci

### DIFF
--- a/drivers/firmware/tisci/tisci.c
+++ b/drivers/firmware/tisci/tisci.c
@@ -75,6 +75,7 @@ static struct tisci_xfer *tisci_setup_one_xfer(const struct device *dev, uint16_
 	if (rx_message_size > config->max_msg_size || tx_message_size > config->max_msg_size ||
 	    (rx_message_size > 0 && rx_message_size < sizeof(*hdr)) ||
 	    tx_message_size < sizeof(*hdr)) {
+		k_sem_give(&data->data_sem);
 		return NULL;
 	}
 
@@ -125,17 +126,20 @@ static int tisci_get_response(const struct device *dev, struct tisci_xfer *xfer)
 
 	if (!xfer->rx_message.buf) {
 		LOG_ERR("No response buffer provided");
+		k_sem_give(&data->data_sem);
 		return -EINVAL;
 	}
 
 	if (k_sem_take(data->rx_message.response_ready_sem, K_MSEC(config->max_rx_timeout_ms)) !=
 	    0) {
 		LOG_ERR("Timeout waiting for response");
+		k_sem_give(&data->data_sem);
 		return -ETIMEDOUT;
 	}
 
 	if (xfer->rx_message.size > config->max_msg_size) {
 		LOG_ERR("rx_message.size [ %d ] > max_msg_size\n", xfer->rx_message.size);
+		k_sem_give(&data->data_sem);
 		return -EINVAL;
 	}
 
@@ -147,6 +151,7 @@ static int tisci_get_response(const struct device *dev, struct tisci_xfer *xfer)
 	if (data->rx_message.size < xfer->rx_message.size) {
 		LOG_ERR("rx_message.size [ %zu ] < xfer->rx_message.size [ %zu ]\n",
 			data->rx_message.size, xfer->rx_message.size);
+		k_sem_give(&data->data_sem);
 		return -EINVAL;
 	}
 
@@ -164,6 +169,7 @@ static int tisci_get_response(const struct device *dev, struct tisci_xfer *xfer)
 	/* Sanity check for message response */
 	if (hdr->seq != data->seq) {
 		LOG_ERR("HDR seq != data seq [%d != %d]\n", hdr->seq, data->seq);
+		k_sem_give(&data->data_sem);
 		return -EINVAL;
 	}
 
@@ -215,6 +221,7 @@ static int tisci_do_xfer(const struct device *dev, struct tisci_xfer *xfer)
 	if (ret < 0) {
 		LOG_ERR("Could not send on %s path\n",
 			config->is_secure ? "secure" : "non-secure");
+		k_sem_give(&data->data_sem);
 		return ret;
 	}
 
@@ -226,8 +233,12 @@ static int tisci_do_xfer(const struct device *dev, struct tisci_xfer *xfer)
 		}
 		if (!tisci_is_response_ack(xfer->rx_message.buf)) {
 			LOG_ERR("TISCI Response in NACK\n");
+			k_sem_give(&data->data_sem);
 			return -ENODEV;
 		}
+	} else {
+		/* No response requested, release semaphore */
+		k_sem_give(&data->data_sem);
 	}
 
 	return 0;
--- a/drivers/firmware/tisci/tisci.c
+++ b/drivers/firmware/tisci/tisci.c
@@ -31,6 +31,7 @@ struct tisci_config {
 	uint32_t host_id;
 	int max_msg_size;
 	int max_rx_timeout_ms;
+	bool is_secure;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


drivers/firmware/tisci/tisci.c
```diff
@@ -58,16 +59,45 @@ struct tisci_data {
...
 /* Core/Setup Functions */
+/**
+ * @brief Determine if a message type requires the secure path
+ * @param msg_type: TISCI message type
+ * @return true if secure path required, false for non-secure path
+ *
+ * Board configuration, security, and OTP messages must use the secure path.
+ * All other messages use the non-secure path by default.
+ *
+ * Reference: MCU+SDK sciclient.c Sciclient_getCurrentContext()
```
soumya-TI: This is no need I guess, referencing a different code base with no proper access.

drivers/firmware/tisci/tisci.c
```diff
@@ -58,16 +59,45 @@ struct tisci_data {
...
+ */
+static bool tisci_msg_requires_secure_path(uint16_t msg_type)
+{
+	switch (msg_type) {
+	/* Board configuration messages (0x000x range) */
+	case TISCI_MSG_BOARD_CONFIG:             /* 0x000B */
+	case TISCI_MSG_BOARD_CONFIG_RM:          /* 0x000C */
```
soumya-TI: RM and PM are not secure messages. There are several secure messages, TISCI_MSG_SEC_HANDOVER,TISCI_MSG_BOOT_NOTIFICATION, TISCI_MSG_SA2UL_SET_DKEK,TISCI_MSG_SA2UL_GET_DKEK and more. I think messages are missing

drivers/firmware/tisci/tisci.c
```diff
@@ -58,16 +59,45 @@ struct tisci_data {
...
+
 static struct tisci_xfer *tisci_setup_one_xfer(const struct device *dev, uint16_t msg_type,
 					       uint32_t msg_flags, void *req_buf,
 					       size_t tx_message_size, void *resp_buf,
 					       size_t rx_message_size)
 {
 	struct tisci_data *data = dev->data;
+	const struct tisci_config *config = dev->config;
+
+	if (!config->is_secure && tisci_msg_requires_secure_path(msg_type)) {
```
soumya-TI: Why are we having a block here, not to send a message on a non-secure path, its job of server to nack it. isn't it? Client driver should be dumb enough to send the message, if that so, then all the messages should go as its is there shouldn't be a need to differentiate secure and non-secure messages. If we are differentiating we are building a intelligence, that can be added to support both secure and non-secure pathways on the same driver. Isn't it?

drivers/firmware/tisci/tisci.c
Kronosblaster: updated

drivers/firmware/tisci/tisci.c
Kronosblaster: updated

drivers/firmware/tisci/tisci.c
soumya-TI: Resolving this comment, from a host perspective it can be either secure or non-secure, if non-secure we should not have secure messages sent out. The block seems okay in that perspective

glneo: I agree with your original assessment, a host can either be secure or non-secure. If secure then all messages go over the secure channel, non-secure -> all over non-secure. The TI-SCI client should not have to know details about which messages are "secure" messages, if they are and the host is not, then TI-SCI firmware will simply reject the message and we will return the appropriate error.

Suggest removing patch 2 from this series.

Kronosblaster: TISCI firmware doesnt reject the message. It silently drops it. This might lead to confusion if not blocked at send

glneo: TISCI firmware not responding will result in a timeout waiting for the result which is treated as an error the same as a reject message.

drivers/firmware/tisci/tisci.h
```diff
@@ -21,24 +21,56 @@
...
 
-/* Device requests */
-#define TISCI_MSG_SET_DEVICE_STATE  0x0200
-#define TISCI_MSG_GET_DEVICE_STATE  0x0201
-#define TISCI_MSG_SET_DEVICE_RESETS 0x0202
+/* Low Power Mode Messages */
+#define TISCI_MSG_ENTER_SLEEP 0x0301
 
-/* Clock requests */
+/* Security Messages */
```
glneo: Why are these message ID's put up here, why not down below with other high numbered messages (0x9000 messages should go almost at the bottom of the list)

drivers/firmware/tisci/tisci.c
```diff
@@ -58,6 +58,44 @@ struct tisci_data {
...
+	case TISCI_MSG_READ_OTP_MMR:
+	case TISCI_MSG_LOCK_OTP_ROW:
+	case TISCI_MSG_GET_OTP_ROW_LOCK_STATUS:
+	case TISCI_MSG_SOFT_LOCK_OTP_WRITE_GLOBAL:
+	case TISCI_MSG_READ_SWREV:
+	case TISCI_MSG_WRITE_SWREV:
+	case TISCI_MSG_READ_KEYCNT_KEYREV:
+	case TISCI_MSG_KEY_WRITER:
+	case TISCI_MSG_WRITE_KEYREV:
+	case TISCI_MSG_ENTER_SLEEP:
```
glneo: If a message is secure the core should not have any reason to use that message to begin with. This list is info that shouldn't need to be stored over here in applications using TI-SCI. I'd suggest just dropping this check for now as it is holding up this series that otherwise would almost ready.

drivers/firmware/tisci/tisci.h
glneo: These are still all out of order, why keep updating this series without first addressing my simple request to keep the definitions in numerical order?

Kronosblaster: This is updated. Aside from TISCI_MSG_OPEN_DEBUG_FWLS and TISCI_MSG_SA2UL_GET_DKEK which are grouped by function all other messages have been numerically ordered

drivers/firmware/tisci/tisci.c
```diff
@@ -213,7 +213,9 @@ static int tisci_do_xfer(const struct device *dev, struct tisci_xfer *xfer)
 
 	ret = mbox_send_dt(&config->mbox_tx, msg);
 	if (ret < 0) {
-		LOG_ERR("Could not send (%d)\n", ret);
+		LOG_ERR("Could not send on %s path\n",
+			config->is_secure ? "secure" : "non-secure");
```
glneo, line 223: Move this change into the patch that added secure data path in the first place.

drivers/firmware/tisci/tisci.c
```diff
@@ -139,12 +139,26 @@ static int tisci_get_response(const struct device *dev, struct tisci_xfer *xfer)
...
+		LOG_ERR("rx_message.size [ %zu ] < expected_rx_size [ %zu ]\n",
+			data->rx_message.size, expected_rx_size);
 		return -EINVAL;
 	}
 
-	memcpy(xfer->rx_message.buf, data->rx_message.buf, xfer->rx_message.size);
+	/* Skip secure header (if present) and copy tisci_msg_hdr + payload */
+	memcpy(xfer->rx_message.buf,
+	       (uint8_t *)data->rx_message.buf + rx_offset,
+	       xfer->rx_message.size);
```
glneo: Why not just do this data move/shift up inside the `if (config->is_secure)` block, then you don't need to keep track of `rx_offset`. You could also just subtract out the secure header bytes from the `xfer->rx_message.size` at that point too so you can keep the old size check as it was (less churn).

glneo (approved):

LaurentiuM1234 (approved):

PR #101343, opened 2025-12-19
lib: cbprintf: fix printing high-bit characters

rbrscoleman:
The `cbprintf_cb` callback function taken by `cbvprintf()` may return a negative number to indicate an I/O error; in the success case, it is supposed to return the (positive) input. The `str_out()` implementation can never error (it just copies the input to a buffer), but it always returns its literal input. When called with a signed character with its MSB set (e.g., UTF-8-encoded text), the integer value is sign-extended, and the function returns a negative number. This is interpreted as an error by `outs()` (in `cbprintf_complete.c`), and so printing concludes after a single character.

For example, prior to this change, the assertion in this snippet fails:

```c
char buf[16];
int len = snprintfcb(buf, sizeof(buf), "\xED\xA0\xBD\xED\xB1\x8D");
assert(len == 6);
```

This can be seen when compiling with `CONFIG_CBPRINTF_LIBC_SUBSTS=y` and `CONFIG_CBPRINTF_NANO=n`, as covered by these unit tests:

```sh
west build --pristine --board unit_testing/unit_testing tests/unit/cbprintf/ --test-item utilities.prf.m32v101
# or
west build --pristine --board unit_testing/unit_testing tests/unit/cbprintf/ --test-item utilities.prf.m64v101
```

Then try:

```sh
./build/testbinary
```

...without and with the change to `lib/os/cbprintf.c`.

### DIFF
--- a/lib/os/cbprintf.c
+++ b/lib/os/cbprintf.c
@@ -46,7 +46,7 @@ static int str_out(int c,
 		*(scp->dp++) = c;
 	}
 
-	return c;
+	return (int) (unsigned char) c;
 }
 
 int fprintfcb(FILE *stream, const char *format, ...)
--- a/tests/unit/cbprintf/main.c
+++ b/tests/unit/cbprintf/main.c
@@ -93,6 +93,8 @@
 #define PKG_ALIGN_OFFSET (size_t)0
 #endif
 
+static const char *utf8_str = "\xF0\x9F\xAA\x81";
+
 /* We can't determine at build-time whether int is 64-bit, so assume
  * it is.  If not the values are truncated at build time, and the str
  * pointers will be updated during test initialization.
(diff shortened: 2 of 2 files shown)


nordic-krch (approved):

lib/os/cbprintf.c
```diff
@@ -45,7 +45,7 @@ static int str_out(int c,
 		*(scp->dp++) = c;
 	}
 
-	return c;
+	return 0;
```
nordic-krch: Actually it should return `c` but casted to uint8_t and then to int: https://github.com/zephyrproject-rtos/zephyr/blob/3638fc57328a7a0e6988b79b0bdedd2b9e27e0e1/include/zephyr/sys/cbprintf.h#L289

nordic-krch (changes requested):

lib/os/cbprintf.c
rbrscoleman: Ah, thank you, I missed that! Fixed now (although I used `unsigned char` rather than `uint8_t` – can change if you'd prefer).

rbrscoleman:
Not stale, just awaiting review again.

@nordic-krch, please do let me know if there's anything I can do to make review easier.

nordic-krch:
Unit tests are run on `unit_testing` virtual board. You can run it like: ` west build -p -b unit_testing/unit_testing tests/unit/cbprintf/ -T utilities.prf.m32v00` and then `./build/testbinary` to run the executable.

rbrscoleman:
Thanks, it was the `-T, --test-item` flag I was missing.

I've added coverage for this issue to the `prf.test_libc_substs` unit test, and I've manually checked that the test fails without the patch, and passes with.

nordic-krch (approved):

nordic-krch:
you can split it into 2 commits.

dcpleung (approved):

Issue #100948, opened 2025-12-12
zms: zms_clear() unusable when ZMS partition is corrupt


antevir:
### Describe the bug

We have a scenario where we changed the block size used in the ZMS partition. When upgrading from old FW this result in an expected error when calling `zms_mount()`. We detect this in our fw and call `zms_clear()` to just wipe the partition. 
However, if you call `zms_clear()` after `zms_mount()` has failed `zms_clear()` will do nothing and return `-EACCES` due to these lines:
```c
	if (!fs->ready) {
		LOG_ERR("zms not initialized");
		return -EACCES;
	}
```
`fs->ready` will be set to `true` only if `zms_mount()` succeeds completely. Hence, `zms_clear()` can currently not be used for recovering from a corrupt ZMS partition.

The workaround is of course to manually set `fs->ready = true` before calling `zms_clear()`.

I can make a simple patch just removing the if statement, but I'm guessing there maybe was a specific reason why it is there in the first place so I want to check here first.

### Regression

- [ ] This is a regression.

### Relevant log output

```shell

```

### Impact

Functional Limitation – Some features not working as expected, but system usable.

### Environment

SHA: we use ff8f0c579, but current main cb2109e has the same lines in zms_clear()

antevir:
Hi @rghaddab
thanks for the feedback! I had completely missed the future feature doc.

I made a quick implementation from your desc that I have verified on my board and through the twister tests. I pushed it from my working account as you can see above. Will go on vacation in about an hour so we can discuss exact implementation when I get back.

PR #101348, opened 2025-12-19
fs: zms: Add zms_mount_force()

ante-ublox:
Add a `wipe_on_failure` parameter to zms_mount() that automatically wipes and reinitializes the partition if the initial mount fails.

A corresponding Kconfig option `SETTINGS_ZMS_WIPE_ON_MOUNT_FAILURE` is added to the settings subsystem to enable this feature when using ZMS as the storage backend.

Relates to #100948

### DIFF
--- a/subsys/settings/Kconfig
+++ b/subsys/settings/Kconfig
@@ -92,6 +92,12 @@ config SETTINGS_ZMS_LL_CACHE_SIZE
 	help
 	  Number of entries in Settings ZMS linked list cache.
 
+config SETTINGS_ZMS_FORCE_MOUNT
+	bool "ZMS force mount"
+	help
+	  Enable forcing of ZMS storage mount, wiping partition if necessary to recover from
+	  corrupted data.
+
 endif # SETTINGS_ZMS
 
 config SETTINGS_FCB
--- a/tests/subsys/kvss/zms/src/main.c
+++ b/tests/subsys/kvss/zms/src/main.c
@@ -1318,3 +1318,78 @@ ZTEST_F(zms, test_zms_free_space_5sectors)
 	zassert_equal(free_space_total, zms_calc_free_space(&fixture->fs),
 		      "total free space did not match sum of gc'd sectors");
 }
+
+/*
+ * Test mounting recovery using zms_mount_force().
+ * Verifies that corrupted ZMS partition triggers mount failure when using zms_mount(),
+ * but auto-recovery when using zms_mount_force().
+ */
+ZTEST_F(zms, test_zms_mount_force)
+{
+	int err;
+	struct zms_ate close_ate;
+	struct zms_ate empty_ate;
+	uint32_t test_data = 0xDEADBEEFU;
+	ssize_t len;
+
+	/* Construct valid ZMS ATEs with UNSUPPORTED VERSION (version=2).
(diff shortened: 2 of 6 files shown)


subsys/fs/zms/zms.c
```diff
@@ -1459,7 +1469,7 @@ static int zms_init(struct zms_fs *fs)
 	return rc;
 }
 
-int zms_mount(struct zms_fs *fs)
+int zms_mount(struct zms_fs *fs, bool wipe_on_failure)
```
rghaddab: @ante-ublox To avoid breaking the API and making all the users change their code I would suggest the following.
Add a zms_mount_force(struct zms_fs *fs) public function
This function will call zms_mount and if it returns an error (other than -EINVAL) it will wipe_out all sectors.
Keep what you already did for zms_wipe_partition()

ante-ublox: Back from holidays now

@rghaddab: Great, I'll update accordingly.  Should we name the kconfig `SETTINGS_ZMS_FORCE_MOUNT` then or do you have any other suggestion?

subsys/settings/Kconfig
```diff
@@ -92,6 +92,13 @@ config SETTINGS_ZMS_LL_CACHE_SIZE
 	help
 	  Number of entries in Settings ZMS linked list cache.
 
+config SETTINGS_ZMS_FORCE_MOUNT
+	bool "ZMS force mount"
+	default n
```
pdgendt: Omit `default n`

```suggestion
```

ante-ublox: Fixed

subsys/fs/zms/zms.c
ante-ublox: I created a shared `zms_mount_internal(..., bool wipe_on_failure)` that is used by both `zms_mount()` and `zms_mount_force()` since if I would have called `zms_mount()` from `zms_mount_force()` then I need to check the specific return code - as I don't think we want to wipe the partition if one of the early `-EINVAL` returns get called. And since I am not certain if there are any code paths in `zms_init()` that may return `-EINVAL` when partition is corrupted, I though it could be safer to do the wipe only (and always) when `zms_init()` fails.

But just let me know if you want to change so that `zms_mount_force()` directly calls `zms_mount()` instead

tests/subsys/kvss/zms/src/main.c
```diff
@@ -1318,3 +1318,78 @@ ZTEST_F(zms, test_zms_free_space_5sectors)
...
+ * Test mounting recovery using zms_mount_force().
+ * Verifies that corrupted ZMS partition triggers mount failure when using zms_mount(),
+ * but auto-recovery when using zms_mount_force().
+ */
+ZTEST_F(zms, test_zms_mount_force)
+{
+	int err;
+	struct zms_ate close_ate;
+	struct zms_ate empty_ate;
+	uint32_t test_data = 0xDEADBEEF;
```
rghaddab:
```suggestion
	uint32_t test_data = 0xDEADBEEFU;
```

ante-ublox: Done!

rghaddab (approved):

tpambor (approved):

PR #101349, opened 2025-12-19
tests: Bluetooth: tester: Fix Central Address Resolution chr. read

m-alperen-sener:
Addressing PTS v8.11.1b6 updates for PTS test cases:
GAP/CONN/DCON/BV-04-C
GAP/CONN/DCON/BV-05-C

Required  AutoPTS Changes https://github.com/auto-pts/auto-pts/pull/1660

CAR characteristic read process was incomplete, CAR status was not actually stored.

Added a new BTP_GAP_EV_PEER_CAR_RECEIVED to tester. This is to prevent disconnect events before IUT successfully reads CAR from central.

Updated start_directed_advertising() so that if the central has no CAR support we must not send directed advertisements. IUT might enter another connectable mode according to specification, thus IUT starts sending connectable undirected advertisements with resovable address.

### DIFF
--- a/tests/bluetooth/tester/src/btp_gap.c
+++ b/tests/bluetooth/tester/src/btp_gap.c
@@ -66,17 +66,48 @@ static struct bt_le_oob oob_sc_remote = { 0 };
 #define REJECT_LATENCY 0x0000
 #define REJECT_SUPERVISION_TIMEOUT 0x0C80
 
+/**
+ * Used to store CAR support status of bonded peer devices not every connected peer.
+ * So it is sufficient to have storage enough for CONFIG_BT_MAX_PAIRED.
+ * CAR status is only requested from peers that are bonded, because GATT CARR read is
+ * triggered in auth_pairing_complete callback.
+ * Thus GATT read callback calls add_to_peers_with_car and bond_deleted callback resets the
+ * CAR status when bond is deleted.
+ */
 static struct {
 	bt_addr_le_t addr;
 	bool supported;
-} cars[CONFIG_BT_MAX_PAIRED];
+} peers_with_car[CONFIG_BT_MAX_PAIRED];
+
+static void add_to_peers_with_car(const bt_addr_le_t *addr, bool supported)
+{
+	/* Check if the peer is already in the list, update the supported state */
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (bt_addr_le_eq(addr, &peers_with_car[i].addr)) {
+			peers_with_car[i].supported = supported;
+			return;
+		}
+	}
+
+	/* If the peer is not in the list, add it */
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (bt_addr_le_eq(&peers_with_car[i].addr, &bt_addr_le_any)) {
+			peers_with_car[i].supported = supported;
+			bt_addr_le_copy(&peers_with_car[i].addr, addr);
+			return;
+		}
+	}
+}
 
 static uint8_t read_car_cb(struct bt_conn *conn, uint8_t err,
 			  struct bt_gatt_read_params *params, const void *data,
 			  uint16_t length)
 {
+	struct btp_gap_peer_car_status_ev ev;
 	struct bt_conn_info info;
 	bool supported = false;
+
+	ev.car = 0x00;
 
 	if (!err && data && length == 1) {
 		const uint8_t *tmp = data;
@@ -84,17 +115,16 @@ static uint8_t read_car_cb(struct bt_conn *conn, uint8_t err,
 		/* only 0 or 1 are valid values */
 		if (tmp[0] == 1) {
 			supported = true;
+			ev.car = 0x01;
 		}
 	}
 
 	bt_conn_get_info(conn, &info);
 
-	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-		if (bt_addr_le_eq(info.le.dst, &cars[i].addr)) {
-			cars[i].supported = supported;
-			break;
-		}
-	}
+	add_to_peers_with_car(info.le.dst, supported);
+
+	bt_addr_le_copy(&ev.address, info.le.dst);
+	tester_event(BTP_SERVICE_ID_GAP, BTP_GAP_EV_PEER_CAR_RECEIVED, &ev, sizeof(ev));
 
 	return BT_GATT_ITER_STOP;
 }
@@ -967,30 +997,46 @@ static uint8_t start_advertising(const void *cmd, uint16_t cmd_len,
 	return BTP_STATUS_SUCCESS;
 }
 
+/**
+ * Start directed advertising with a peer address with or without RPA.
+ * If privacy is enabled and the peer does not support Central Address Resolution,
+ * the advertisement will be started as undirected with RPA.
+ */
 static uint8_t start_directed_advertising(const void *cmd, uint16_t cmd_len,
 					  void *rsp, uint16_t *rsp_len)
 {
 	const struct btp_gap_start_directed_adv_cmd *cp = cmd;
 	struct btp_gap_start_directed_adv_rp *rp = rsp;
-	struct bt_le_adv_param adv_param;
+	struct bt_le_adv_param adv_param = BT_LE_ADV_PARAM_INIT(
+		BT_LE_ADV_OPT_CONN, BT_GAP_ADV_FAST_INT_MIN_2, BT_GAP_ADV_FAST_INT_MAX_2, NULL);
 	uint16_t options = sys_le16_to_cpu(cp->options);
 
-	adv_param = *BT_LE_ADV_CONN_DIR(&cp->address);
-
-	if (!(options & BTP_GAP_START_DIRECTED_ADV_HD)) {
-		adv_param.options |= BT_LE_ADV_OPT_DIR_MODE_LOW_DUTY;
-		adv_param.interval_max = BT_GAP_ADV_FAST_INT_MAX_2;
-		adv_param.interval_min = BT_GAP_ADV_FAST_INT_MIN_2;
+	if (bt_addr_le_eq(&cp->address, &bt_addr_le_any)) {
+		LOG_ERR("Invalid peer address");
+		return BTP_STATUS_FAILED;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_PRIVACY) && (options & BTP_GAP_START_DIRECTED_ADV_PEER_RPA)) {
-		/* check if peer supports Central Address Resolution */
+		/**
+		 * In accordance with the test spec for test case GAP/CONN/DCON/BV-05-C, if the peer
+		 * does not support Central Address Resolution, the advertisement will be started
+		 * as undirected.
+		 */
 		for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-			if (bt_addr_le_eq(&cp->address, &cars[i].addr)) {
-				if (cars[i].supported) {
-					adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
+			if (bt_addr_le_eq(&cp->address, &peers_with_car[i].addr) &&
+			    peers_with_car[i].supported) {
+				adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
+				adv_param.peer = &cp->address;
+				if ((options & BTP_GAP_START_DIRECTED_ADV_HD) == 0U) {
+					adv_param.options |= BT_LE_ADV_OPT_DIR_MODE_LOW_DUTY;
 				}
+				break;
 			}
+		}
+	} else {
+		adv_param.peer = &cp->address;
+		if ((options & BTP_GAP_START_DIRECTED_ADV_HD) == 0U) {
+			adv_param.options |= BT_LE_ADV_OPT_DIR_MODE_LOW_DUTY;
 		}
 	}
 
@@ -1575,9 +1621,21 @@ static void auth_pairing_complete(struct bt_conn *conn, bool bonded)
 	}
 }
 
+static void bond_deleted(uint8_t id, const bt_addr_le_t *peer)
+{
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (bt_addr_le_eq(peer, &peers_with_car[i].addr)) {
+			peers_with_car[i].supported = false;
+			bt_addr_le_copy(&peers_with_car[i].addr, &bt_addr_le_any);
+			return;
+		}
+	}
+}
+
 static struct bt_conn_auth_info_cb auth_info_cb = {
 	.pairing_failed = auth_pairing_failed,
 	.pairing_complete = auth_pairing_complete,
+	.bond_deleted = bond_deleted,
 };
 
 #if defined(CONFIG_BT_CLASSIC)
--- a/tests/bluetooth/tester/src/btp/btp_gap.h
+++ b/tests/bluetooth/tester/src/btp/btp_gap.h
@@ -619,6 +619,12 @@ struct btp_gap_periodic_biginfo_ev {
 	uint8_t encryption;
 } __packed;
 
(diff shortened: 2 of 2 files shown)


tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -972,26 +1001,33 @@ static uint8_t start_directed_advertising(const void *cmd, uint16_t cmd_len,
...
+						adv_param.options |=
+							BT_LE_ADV_OPT_DIR_MODE_LOW_DUTY;
+					}
+					break;
 				}
 			}
 		}
+	} else {
+		adv_param.peer = &cp->address;
+		if (!(options & BTP_GAP_START_DIRECTED_ADV_HD)) {
```
Thalley:
```suggestion
		if ((options & BTP_GAP_START_DIRECTED_ADV_HD) == 0U) {
```

Thalley (changes requested):
One minor coding guideline violation, otherwise it looks good, once the Copilot review comments have been handled

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,63 @@ static struct bt_le_oob oob_sc_remote = { 0 };
...
+{
+
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (!peers_with_car[i].paired) {
+			peers_with_car[i].paired = true;
+			peers_with_car[i].supported = supported;
+			bt_addr_le_copy(&peers_with_car[i].addr, addr);
+			return;
+		}
+	}
```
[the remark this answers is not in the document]
m-alperen-sener: not needed, Host is limited by CONFIG_BT_MAX_PAIRED.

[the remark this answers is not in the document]
Thalley: This assumes that entries in `peers_with_car` are also removed once a bond is removed. As far as I can tell `remove_from_peers_with_car` is called when a peer disconnects, and `add_to_peers_with_car` seems to be called when a connected peer does a read. Should the size of `peers_with_car` be related to `CONFIG_BT_MAX_CONN` instead? 

I think either `peers_with_car` need to use `CONFIG_BT_MAX_CONN` if it depends on the connected state (add on read, remove on disconnect), or it should continue to use `CONFIG_BT_MAX_PAIRED` but then `addr` should be set when a device bonds and remove when a bond is deleted. 

`CONFIG_BT_MAX_CONN` can be larger than `CONFIG_BT_MAX_PAIRED` so the comment here from Copilot is correct

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,63 @@ static struct bt_le_oob oob_sc_remote = { 0 };
...
+}
```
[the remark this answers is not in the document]
Thalley: @m-alperen-sener This one also seems valid. `read_car_cb` may be triggered more than once, and we don't want to add duplicates, especially since that we can no longer rely on `CONFIG_BT_MAX_CONN` or `CONFIG_BT_MAX_PAIRED` to determine the maximum size of this array

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,61 @@ static struct bt_le_oob oob_sc_remote = { 0 };
...
+
+static void remove_from_peers_with_car(const bt_addr_le_t *addr)
+{
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (peers_with_car[i].paired && bt_addr_le_cmp(addr, &peers_with_car[i].addr) == 0) {
```
Thalley:
```suggestion
		if (peers_with_car[i].paired && bt_addr_le_eq(addr, &peers_with_car[i].addr)) {
```

tests/bluetooth/tester/src/btp_gap.c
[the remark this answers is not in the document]
m-alperen-sener: CAR read is only triggered in  auth_pairing_complete, not all connected devices are asked for CAR but only paired/bonded devices. Even though the CAR itself does not require authentication or authorization.

https://github.com/m-alperen-sener/zephyr/blob/5ee52419c4f62554a468f0ec5832f49d4be2d338/tests/bluetooth/tester/src/btp_gap.c#L1609

That was why I though CONFIG_BT_MAX_PAIRED should be enough.

I'll check the part when should we remove, do we need it to remove at all?; it is not related to disconnect ( anyway, you are right) but to device being unpaired etc, instead.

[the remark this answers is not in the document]
m-alperen-sener: About removing the CAR status; un-pairing etc. eventually ends up triggering disconnect event and removing CAR status, and that seems satisfactory to address functionality under being tested and properly holding the CAR status. Besides the CAR status is not persistent and overwritten for each auth connection. Adding an address check should prevent a failure in `add_to_peers_with_car()`

tests/bluetooth/tester/src/btp_gap.c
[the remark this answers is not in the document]
m-alperen-sener: added the check

tests/bluetooth/tester/src/btp_gap.c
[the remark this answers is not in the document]
Thalley: > CAR read is only triggered in auth_pairing_complete, not all connected devices are asked for CAR but only paired/bonded devices. Even though the CAR itself does not require authentication or authorization.

Ah, right, so it should be OK. `auth_pairing_complete` can be called multiple times for a single connection, so it does still need a check to avoid adding the same device multiple times. 

> About removing the CAR status; un-pairing etc. eventually ends up triggering disconnect event and removing CAR status, and that seems satisfactory to address functionality under being tested and properly holding the CAR status. Besides the CAR status is not persistent and overwritten for each auth connection. Adding an address check should prevent a failure in add_to_peers_with_car()

That is also true, and if we do not need to store the value after disconnect then the current behavior here is valid

Thalley (dismissed):

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -972,26 +1008,33 @@ static uint8_t start_directed_advertising(const void *cmd, uint16_t cmd_len,
...
-		/* check if peer supports Central Address Resolution */
+		/*
+		 * Check if peer supports Central Address Resolution.
+		 * If not supported the advertisement will be started as undirected
+		 */
 		for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-			if (bt_addr_le_eq(&cp->address, &cars[i].addr)) {
-				if (cars[i].supported) {
+			if (bt_addr_le_eq(&cp->address, &peers_with_car[i].addr)) {
+				if (peers_with_car[i].supported && peers_with_car[i].paired) {
```
PavelVPV: IMHO, if `BTP_GAP_START_DIRECTED_ADV_PEER_RPA` has been requested and peer does not support CAR, this function shall return error as it will otherwise behave differently from what was expected.

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -972,26 +1008,33 @@ static uint8_t start_directed_advertising(const void *cmd, uint16_t cmd_len,
 {
 	const struct btp_gap_start_directed_adv_cmd *cp = cmd;
 	struct btp_gap_start_directed_adv_rp *rp = rsp;
-	struct bt_le_adv_param adv_param;
+	struct bt_le_adv_param adv_param = BT_LE_ADV_PARAM_INIT(
```
PavelVPV, line 1010: Nitpick. Why this part of the code needs to be changed? I see the only difference is the adv interval change. Why not just pull the interval assignment out of the if-statement at line 980?

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,70 @@ static struct bt_le_oob oob_sc_remote = { 0 };
 #define REJECT_LATENCY 0x0000
 #define REJECT_SUPERVISION_TIMEOUT 0x0C80
 
+/* Used to store CAR support status of peer devices. */
 static struct {
 	bt_addr_le_t addr;
+	bool paired;
```
PavelVPV: Seems like `paired` is only used to find an empty entry in the array. Can't we use `bt_addr_le_any`? https://github.com/zephyrproject-rtos/zephyr/blob/7963b69fa451c6b611040597e6d99e0dc3c61be2/subsys/bluetooth/common/addr.c#L11

tests/bluetooth/tester/src/btp_gap.c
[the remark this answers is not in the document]
PavelVPV: > CAR read is only triggered in auth_pairing_complete, not all connected devices are asked for CAR but only paired/bonded devices. Even though the CAR itself does not require authentication or authorization.

I think it makes sense to add a comment regarding this implicit relationship somewhere in this function.

tests/bluetooth/tester/src/btp_gap.c
m-alperen-sener: nice, good point. then I can remove paired flag.

tests/bluetooth/tester/src/btp_gap.c
m-alperen-sener: Also a valid behavior to return error and do nothing. Referring to test spec GAP.TS.p49, in test case GAP/CONN/DCON/BV-05-C test step 7
> The IUT refuses the order. The IUT might enter another connectable mode and establish the connection this way.

So as you suggested (1)refusing the order and doing nothing is valid behavior, (2)starting to send a different type of connectable adv is also a valid behavior. I have chosen the (2) because PTS application fails to verify test case by (1).

The correct behavior is to not sent directed adv.

tests/bluetooth/tester/src/btp_gap.c
m-alperen-sener: related to: https://github.com/zephyrproject-rtos/zephyr/pull/101349/changes#r2664908231

In case the central device does not support CAR and Privacy enabled, it defaults to "non-directed" connectable adv type.

I will add info for this WID API to clarify its behavior.

tests/bluetooth/tester/src/btp_gap.c
PavelVPV: In this test case yes, but what if another test case uses this command? autopts requests to use RPA, but tester doesn't follow the order. IMHO it would be better if autopts controls this and tester just either does what it has been told or fails and not silently does something else.

This can be a bigger change though, so I am fine with the current change.

tests/bluetooth/tester/src/btp_gap.c
PavelVPV: Yeah, exactly what in my opinion is not good. autopts should know whether peer supports CAR or not and behave based on this and not tester. Even the function in autopts is called `gap_direct_adv_on` but apparently it may initiate undirected advs...

m-alperen-sener:
So sorry I have mistakenly closed the PR.

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,76 @@ static struct bt_le_oob oob_sc_remote = { 0 };
...
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (bt_addr_le_eq(addr, &peers_with_car[i].addr)) {
+			peers_with_car[i].supported = supported;
+			return;
+		}
+	}
+
+	/* If the peer is not in the list, add it */
+	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
+		if (bt_addr_le_eq(addr, &bt_addr_le_any)) {
```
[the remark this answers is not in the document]
PavelVPV: This is valid

tests/bluetooth/tester/src/btp_gap.c
PavelVPV: I'm now thinking it was a bad suggestion, sorry :( The fact that `bt_addr_le_any` is zero-ed struct is hidden from user and we can't rely on it. If it changes, the code will behave incorrectly. Unless addresses in `peers_with_car` are initialized with `bt_addr_le_any`.

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -967,31 +1009,42 @@ static uint8_t start_advertising(const void *cmd, uint16_t cmd_len,
...
+		 */
 		for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-			if (bt_addr_le_eq(&cp->address, &cars[i].addr)) {
-				if (cars[i].supported) {
-					adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
+			if (bt_addr_le_eq(&cp->address, &peers_with_car[i].addr) &&
+			    peers_with_car[i].supported) {
+				adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
+				adv_param.peer = &cp->address;
+				if (!(options & BTP_GAP_START_DIRECTED_ADV_HD)) {
```
Thalley:
```suggestion
				if ((options & BTP_GAP_START_DIRECTED_ADV_HD) == 0U) {
```

Thalley (changes requested):
Minor coding style violation, otherwise LGTM

PavelVPV (dismissed):

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -967,31 +997,42 @@ static uint8_t start_advertising(const void *cmd, uint16_t cmd_len,
...
+		/**
+		 * In accordance with the test spec for test case GAP/CONN/DCON/BV-05-C, if the peer
+		 * does not support Central Address Resolution, the advertisement will be started
+		 * as undirected.
+		 */
 		for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-			if (bt_addr_le_eq(&cp->address, &cars[i].addr)) {
-				if (cars[i].supported) {
-					adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
+			if (bt_addr_le_eq(&cp->address, &peers_with_car[i].addr) &&
```
Thalley, line 1026: Minor: Given that we now use `bt_addr_le_any` to make "empty" entries in the array, we probably should verify that `cp->address != bt_addr_le_any` as that could cause false positive or unexpected behavior

tests/bluetooth/tester/src/btp_gap.c
```diff
@@ -66,35 +66,65 @@ static struct bt_le_oob oob_sc_remote = { 0 };
...
 
 	bt_conn_get_info(conn, &info);
 
-	for (int i = 0; i < CONFIG_BT_MAX_PAIRED; i++) {
-		if (bt_addr_le_eq(info.le.dst, &cars[i].addr)) {
-			cars[i].supported = supported;
-			break;
-		}
-	}
+	add_to_peers_with_car(info.le.dst, supported);
```
Thalley, line 124: Ideally we'd actually check if `conn == NULL` (indicates a disconnect while reading) that would cause a fatal error in `bt_conn_get_info`, but arguably outside the scope of this PR

Thalley (dismissed):
Two minor comments, but no blockers as they are quite unlikely

PavelVPV (dismissed):

tests/bluetooth/tester/src/btp_gap.c
m-alperen-sener: Well I don't see this necessary considering we we are in read_cb already and got we already wanted. Also the err is checked before setting the ```supported``` and ```ev.car``` flags

tests/bluetooth/tester/src/btp_gap.c
m-alperen-sener: added the check (Y), not pushed yet. Running tests locallly.

m-alperen-sener: done

Thalley (approved):

PavelVPV (approved):

sjanc (approved):

PR #101350, opened 2025-12-19
modules: hal_nordic: nrfx: Move defines to header file

nordicjm:
Moves defines to a header file to keep the locally defined to files that need them which are in or use files in the HAL rather than globally applying them to every file

### DIFF
--- a/modules/hal_nordic/nrfx/CMakeLists.txt
+++ b/modules/hal_nordic/nrfx/CMakeLists.txt
@@ -36,75 +36,75 @@ zephyr_include_directories(.)
 include(${BSP_DIR}/zephyr/nrfx.cmake OPTIONAL)
 
 # Define MDK defines globally
-zephyr_compile_definitions_ifdef(CONFIG_SOC_SERIES_NRF51X       NRF51)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF51822_QFAA       NRF51422_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF51822_QFAB       NRF51422_XXAB)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF51822_QFAC       NRF51422_XXAC)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52805            NRF52805_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52810            NRF52810_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52811            NRF52811_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52820            NRF52820_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52832            NRF52832_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_COMPATIBLE_NRF52833 NRF52833_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52840            NRF52840_XXAA)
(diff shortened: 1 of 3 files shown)


modules/hal_nordic/nrfx/CMakeLists.txt
```diff
@@ -36,75 +36,75 @@ zephyr_include_directories(.)
...
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52810            NRF52810_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52811            NRF52811_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52820            NRF52820_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52832            NRF52832_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_COMPATIBLE_NRF52833 NRF52833_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF52840            NRF52840_XXAA)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_COMPATIBLE_NRF5340_CPUAPP NRF5340_XXAA_APPLICATION)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_COMPATIBLE_NRF5340_CPUNET NRF5340_XXAA_NETWORK)
-zephyr_compile_definitions_ifdef(CONFIG_SOC_NRF54H20_CPUAPP     NRF54H20_XXAA
+zephyr_library_compile_definitions_ifdef(CONFIG_SOC_SERIES_NRF51X       NRF51)
```
aescolar, line 39: Note that these macros are not just used by the nrfx HAL but are also used/needed by the MDK. So they need to be set for whoever uses the MDK even if they don't use the HAL.

boards/native/common/natsim_config.cmake
```diff
@@ -31,6 +31,7 @@ set(nsi_config_content
   "NSI_EXTRA_LIBS:=$<JOIN:$<TARGET_PROPERTY:native_simulator,RUNNER_LINK_LIBRARIES>,\ >"
   "NSI_PATH:=${NSI_DIR}/"
   "NSI_N_CPUS:=${CONFIG_NATIVE_SIMULATOR_NUMBER_MCUS}"
+  "NSI_EXTRA_INCLUDES:=-I${ZEPHYR_BASE}/modules/hal_nordic/nrfx"
```
aescolar: Note this cmake file is common for all native simulator builds. It is not used just for the `nrf*bsim` targets. So this would not be the place.
For setting options from the Zephyr build we have the `native_simulator` library.
So if we wanted to set from Zephyr's cmake files this we could do  `target_compile_options(native_simulator INTERFACE "-I balblalbal")`. For the nRF HW models related settings we have it in https://github.com/BabbleSim/ext_nRF_hw_models/blob/main/CMakeLists.txt#L88

aescolar (dismissed):

modules/hal_nordic/nrfx/CMakeLists.txt
nordicjm: They are in the https://github.com/zephyrproject-rtos/zephyr/pull/101350/files#diff-bd8eb7360aa761aff5dbe9f107d15e992ca5b9b9672fb8791b10bc040ce1bf12 file

aescolar: Ok, but note that `nrfx_kconfig.h` is included by Zephyr's variant of `nrfx_config.h` which is included by `nrfx.h`. So they will be defined if `nrfx.h` is included. But not if somebody includes MDK headers without including `nrfx.h`.
Meaning, this is just a remark: "this change will break things for some users"

PR #101352, opened 2025-12-19
net: doc: t1s: Add documentation and configuration files for nucleo_g474re to run T1S stress tests (LAN8651)

lmajewski:
This patch set provides info regarding T1S stress testing with using:
- zperf
- echo_server with udp-perf (https://gitlab.com/network851614/udp-perf)

Proper documentation entry (i.e. rst file) has been provided to facilitate testing T1S driver before submission.


samples/net/ethernet/t1s/README.rst
```diff
@@ -0,0 +1,153 @@
+.. zephyr:code-sample:: stm32_t1s
+   :name: 10BASE-T1S tests (OA_TC6)
+
+   Setup tests to examine long term stability of 10BASE-T1S communication
```
erwango, line 4: Are we sure this is the correct location then ?
This has only little to do with STM.

What about moving this under net ?

lmajewski: I was not sure where to place it..... Just wanted to make it looking correct with `make html-fast`.
I can move it to `net` - no problem.

lmajewski: @erwango I've updated "Labels" and moved the test setup description to new localization.

samples/net/zperf/boards/nucleo_g474re.overlay
```diff
@@ -0,0 +1,51 @@
...
+	status = "okay";
+	pinctrl-0 = <&spi1_sck_pa5 &spi1_miso_pa6 &spi1_mosi_pa7>;
+	pinctrl-names = "default";
+	cs-gpios = <&gpiob 6 GPIO_ACTIVE_LOW>;
+
+	spe: lan865x@0 {
+		compatible = "microchip,lan865x";
+		reg = <0>;
+
+		// pa9 -> D8 label on nucleo
```
erwango: Use C comments style (`/*  */`)

erwango (changes requested):
Nitpick, otherwise looks better to me, but net maintainers need to have a look.

samples/net/sockets/echo_server/boards/nucleo_g474re.conf
```diff
@@ -0,0 +1,110 @@
...
+CONFIG_SHELL_ARGC_MAX=5
+CONFIG_SENSOR_SHELL=n
+CONFIG_SHELL_HISTORY_BUFFER=128
+
+# Network application options and configuration
+CONFIG_NET_CONFIG_SETTINGS=y
+CONFIG_NET_CONFIG_NEED_IPV4=y
+# When more than one board is used - adjust the IPv{46}
+# addresses - e.g. 3.3 -> 3.5
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.168.3.3"
```
jukkar: Please don't use IP addresses that are specific to your use case. Remove these and let the user set them when compiling the app

samples/net/sockets/echo_server/boards/nucleo_g474re.conf
```diff
@@ -0,0 +1,110 @@
...
+# The attempt to fix the initialization routines
+# call sequence - default values of):
+# CONFIG_ETH_INIT_PRIORITY - 80
+# CONFIG_MDIO_INIT_PRIORITY - 60
+# CONFIG_PHY_INIT_PRIORITY - 70
+# are causing dependency error.
+# The 'solution' is to lower the default ETH
+# priority to allow first PHY initialization, then
+# MDIO and on the very end ETH
+CONFIG_ETH_INIT_PRIORITY=59
```
jukkar: There is just too much data in the conf file. Please place only data here that is relevant to this board. All generic application related data needs to go into prj.conf or kept out of tree

samples/net/ethernet/t1s/README.rst
```diff
@@ -0,0 +1,155 @@
+.. zephyr:code-sample:: t1s
```
jukkar, line 1: Please place this file to `samples/net/ethernet/t1s` directory

jukkar (changes requested):
This needs to be a bit more generic

samples/net/ethernet/t1s/README.rst
```diff
@@ -0,0 +1,177 @@
...
+
+   uname -a # Linux raspberrypi 6.1.21
+   cd ~/work/Zephyr/lan867x-linux-driver-2v2/linux-v6.1.21-support
+   ./load.sh
+
+Add default route to Nucleo (on RPI4):
+
+.. code-block:: shell
+
+   ip -6 route add 2004::3 dev eth0
```
jukkar: Shouldn't this be `2001:db8::3`?

samples/net/ethernet/t1s/README.rst
```diff
@@ -0,0 +1,177 @@
...
+
+* Enable config :code:`UDPPERF` in the config file for   :code:`echo_server`
+  or :code:`zperf`
+
+* Extend ``boards/nucleo_g474re.conf`` for per setup IP configuration:
+
+.. code-block:: shell
+
+  CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.168.3.3"
+  CONFIG_NET_CONFIG_PEER_IPV4_ADDR="192.168.3.1"
```
jukkar: Could this be changed to use the IPv4 documentation addresses like in all other places in the networking samples. So `192.0.2.1` for zephyr node and `192.0.2.3` for the Linux host, this way this is more consistent with other samples. In this case the IPv6 addresses could be changed too to `2001:db8::1` for zephyr and `2001:db8::3` for Linux host.

samples/net/ethernet/t1s/README.rst
lmajewski: No problem - I will adjust it.

samples/net/ethernet/t1s/README.rst
lmajewski: Ok, I will adjust IPv{46} and then resend those patches.

jukkar (approved):
Thanks, LGTM.

erwango (approved):
Thanks for the changes

PR #101353, opened 2025-12-19
WIP - optimise gen_driver_kconfig_dts.py performance

gudnimg:
**This is still just a work in progress... but I am curious to see if it passes Zephyr's CI**

----

This change reduces execution time by around 50%

Instead of loading all YAML files into memory and parsing the entire file in a node tree. Simply parse each YAML file and go through the parsing events until the 'compatible' key and its value is found. This speeds up the parsing process significantly.


gudnimg:
Not sure what this means. The function is too many lines...? Hopefully not a hard requirement.

image

PR #101355, opened 2025-12-19
modem: backend: tty: Support building with any C library

aescolar:
Ensure we call into the host C library open/close/read/write independently of which C library the code is built with.

We do this by:
a) Using the native simulator nsi_host* trampolines when we just want to call straight into the host libC.
b) Building in the native_simulator runner context (and therefore with the host C library) two functions which we call from the embedded side.

&

tests/subsys/modem/backends/tty: Ensure it is built with the proper C lib

This test itself (not the code under test) needs to be build with the host C library as it calls into its open/close/read/write APIs.
So let's ensure we build it with it, independently of what the default C library may be.

Note there is a minor functional change, the code before would block during polls during 100ms and then wait 100ms until the next poll. Now it does not block during polls (it just checks and returns immediately) and then waits 100ms.


tests/subsys/modem/backends/tty/prj.conf
```diff
@@ -6,6 +6,8 @@ CONFIG_NO_OPTIMIZATIONS=y
 CONFIG_MODEM_MODULES=y
 CONFIG_MODEM_BACKEND_TTY=y
 
+CONFIG_EXTERNAL_LIBC=y
```
tpambor, line 9: Maybe I am missing something but shouldn't this already be the default for native_sim? https://github.com/zephyrproject-rtos/zephyr/blob/72360f821a610b64fda3bd300b6ca4577f6abe8c/lib/libc/Kconfig#L80

tpambor (approved):

tests/subsys/modem/backends/tty/prj.conf
aescolar: It is. I just wanted to ensure it would continue working even if that default would change at some point.

bjarki-andreasen (approved):

SeppoTakalo (approved):

PR #101370, opened 2025-12-19
include: tracing: add missing k_thread_heap_assign tracing hook

kartben:
Add sys_port_trace_k_thread_heap_assign macro which was defined in all backend implementations but missing from the main tracing.h header.

This macro traces when a resource pool is assigned to a thread via k_thread_heap_assign().


nashif:
`k_thread_heap_assign` is not being traced and there should not be a hook for that in tracing to start with, will submit a patch removing it.

PR #101372, opened 2025-12-19
kernel: O(1) search for threads among CPUs

peter-mitsis:
Instead of performing a linear search to determine if a given thread is running on another CPU, or if it is marked as being preempted by a metaIRQ on any CPU do this in O(1) time.

On SMP systems, Zephyr already tracks the CPU on which a thread executes (or lasted executed). This information is leveraged to do the search in O(1) time.

### DIFF
--- a/kernel/sched.c
+++ b/kernel/sched.c
@@ -304,23 +304,24 @@ static ALWAYS_INLINE void update_cache(int preempt_ok)
 #endif /* CONFIG_SMP */
 }
 
+/**
+ * Returns pointer to _cpu if the thread is currently running on
+ * another CPU.
+ */
 static struct _cpu *thread_active_elsewhere(struct k_thread *thread)
 {
-	/* Returns pointer to _cpu if the thread is currently running on
-	 * another CPU. There are more scalable designs to answer this
-	 * question in constant time, but this is fine for now.
-	 */
 #ifdef CONFIG_SMP
-	int currcpu = _current_cpu->id;
+	int thread_cpu_id = thread->base.cpu;
+	struct _cpu *thread_cpu;
 
-	unsigned int num_cpus = arch_num_cpus();
+	__ASSERT_NO_MSG((thread_cpu_id >= 0) &&
+			(thread_cpu_id < arch_num_cpus()));
 
-	for (int i = 0; i < num_cpus; i++) {
-		if ((i != currcpu) &&
-		    (_kernel.cpus[i].current == thread)) {
-			return &_kernel.cpus[i];
-		}
+	thread_cpu = &_kernel.cpus[thread_cpu_id];
+	if ((thread_cpu->current == thread) && (thread_cpu != _current_cpu)) {
+		return thread_cpu;
 	}
+
 #endif /* CONFIG_SMP */
 	ARG_UNUSED(thread);
 	return NULL;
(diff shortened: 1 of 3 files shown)


kernel/thread.c
```diff
@@ -890,6 +890,7 @@ void z_init_thread_base(struct _thread_base *thread_base, int priority,
 
 #ifdef CONFIG_SMP
 	thread_base->is_idle = 0;
+	thread_base->cpu = 0;
```
TaiJuWu, line 898: It means the run on cpu 0 last time, should we use a magic number to indicate edge case for thread does not run before?

Additional, there is comment from `_thread_base`, the semantic is changed now.
```
/* CPU index on which thread was last run */
uint8_t cpu;
```

By the way, I think it is used to cpu affinity before but I am not sure we have this feature or not.

kernel/sched.c
```diff
@@ -300,23 +300,24 @@ static ALWAYS_INLINE void update_cache(int preempt_ok)
...
 
-	unsigned int num_cpus = arch_num_cpus();
+	__ASSERT_NO_MSG((thread_cpu_id >= 0) &&
+			(thread_cpu_id < arch_num_cpus()));
 
-	for (int i = 0; i < num_cpus; i++) {
-		if ((i != currcpu) &&
-		    (_kernel.cpus[i].current == thread)) {
-			return &_kernel.cpus[i];
-		}
```
JarmouniA, line 322: Would it make sense to keep the linear search as a safe fallback if `thread_cpu_id >= arch_num_cpus()` is true?

kernel/sched.c
```diff
@@ -386,11 +387,13 @@ static void thread_halt_spin(struct k_thread *thread, k_spinlock_key_t key)
 static ALWAYS_INLINE void z_metairq_preempted_clear(struct k_thread *thread)
 {
 #if (CONFIG_NUM_METAIRQ_PRIORITIES > 0)
-	for (unsigned int i = 0; i < CONFIG_MP_MAX_NUM_CPUS; i++) {
-		if (_kernel.cpus[i].metairq_preempted == thread) {
-			_kernel.cpus[i].metairq_preempted = NULL;
-			break;
```
JarmouniA, line 396: ditto

kernel/thread.c
peter-mitsis: I don't _think_ that we need a special value to indicate that the thread has not yet executed. At least with how it is being used in this patch, this will at least ensure that the field is initialized to a value that is within the valid range.

I will update that comment.

Regarding the CPU affinity, I think that is cpu_mask, but I will double check. That is a good catch.

kernel/sched.c
peter-mitsis: It was the linear search that I wanted to eliminate. As far as I can tell, thread_cpu_id should always be less than arch_num_cpus().  With the exception of Xtensa, arch_num_cpus() always returns CONFIG_MP_MAX_NUM_CPUS. In that Xtensa case, it returns either CONFIG_MP_MAX_NUM_CPUS or a value from the SoC (which can not be greater than CONFIG_MP_MAX_NUM_CPUS).

At the time of PR submission, the only place that I could think of where base.cpu could be outside the range was before the thread was scheduled. However, initializing that to 0 should take care of that, and this field is updated when it is scheduled in via do_swap() and determined to be the next thread to execute in z_get_next_switch_handle().

That being said, I will be doing a double check to ensure that there is nowhere else that it is will be set to an out-of-range value.

JarmouniA: I too don't think there is any "normal" case where that condition would be true, but I just thought of it as a kind of defensive programming that would handle the odd case of `base.cpu` containing garbage.

TaiJuWu (approved):
LGTM, thanks for the nice improvement.

teburd (approved):

PR #101373, opened 2025-12-19
drivers: dma: stm32: the source and destination sizes must only match with v1

benedekkupper:
Commit e7f222a fixed the bug where a data size mismatch produces unexpected behavior with the DMA on V1 devices. However, this limitation is only valid for the STM32 series with V1 DMA, V2 doesn't have this. (Perhaps due to the use of FIFOs on V1.) The STM32CubeMX configuration tool correctly implements this limitation in the UI, it is a good way to cross-check.

### DIFF
--- a/drivers/dma/dma_stm32.c
+++ b/drivers/dma/dma_stm32.c
@@ -344,18 +344,25 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
 			dev->name);
 		return -ENOTSUP;
 	}
-#endif /* CONFIG_DMA_STM32_V1 */
-
 	/* Support only the same data width for source and dest */
-	if ((config->dest_data_size != config->source_data_size)) {
+	if (config->dest_data_size != config->source_data_size) {
 		LOG_ERR("source and dest data size differ.");
 		return -EINVAL;
 	}
+#else /* CONFIG_DMA_STM32_V1 */
+	if (config->dest_data_size != 4U &&
+	    config->dest_data_size != 2U &&
+	    config->dest_data_size != 1U) {
+		LOG_ERR("invalid dest unit size: %d",
+			config->dest_data_size);
+		return -EINVAL;
+	}
+#endif /* CONFIG_DMA_STM32_V1 */
 
 	if (config->source_data_size != 4U &&
 	    config->source_data_size != 2U &&
 	    config->source_data_size != 1U) {
-		LOG_ERR("source and dest unit size error, %d",
+		LOG_ERR("invalid source unit size: %d",
 			config->source_data_size);
 		return -EINVAL;
 	}
@@ -457,7 +464,7 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
 
 	stream->source_periph = (stream->direction == PERIPHERAL_TO_MEMORY);
 
-	/* set the data width, when source_data_size equals dest_data_size */
+	/* set the data widths */
 	int index = find_lsb_set(config->source_data_size) - 1;
 	DMA_InitStruct.PeriphOrM2MSrcDataSize = table_p_size[index];
 	index = find_lsb_set(config->dest_data_size) - 1;


djiatsaf-st (dismissed):

drivers/dma/dma_stm32.c
```diff
@@ -344,18 +344,25 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
...
-	if ((config->dest_data_size != config->source_data_size)) {
+	if (config->dest_data_size != config->source_data_size) {
 		LOG_ERR("source and dest data size differ.");
 		return -EINVAL;
 	}
+#else
+	if (config->dest_data_size != 4U &&
+	    config->dest_data_size != 2U &&
+	    config->dest_data_size != 1U) {
+		LOG_ERR("dest unit size error, %d",
```
JarmouniA:
```suggestion
		LOG_ERR("Invalid dest unit size %d",
```

drivers/dma/dma_stm32.c
```diff
@@ -344,18 +344,25 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
...
+			config->dest_data_size);
+		return -EINVAL;
+	}
+#endif /* CONFIG_DMA_STM32_V1 */
 
 	if (config->source_data_size != 4U &&
 	    config->source_data_size != 2U &&
 	    config->source_data_size != 1U) {
-		LOG_ERR("source and dest unit size error, %d",
+		LOG_ERR("source unit size error, %d",
```
JarmouniA:
```suggestion
		LOG_ERR("Invalid source unit size %d",
```

drivers/dma/dma_stm32.c
benedekkupper: Updated

drivers/dma/dma_stm32.c
benedekkupper: Updated

djiatsaf-st (dismissed):

erwango (dismissed):

drivers/dma/dma_stm32.c
```diff
@@ -344,18 +344,25 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
...
 	}
-#endif /* CONFIG_DMA_STM32_V1 */
-
 	/* Support only the same data width for source and dest */
-	if ((config->dest_data_size != config->source_data_size)) {
+	if (config->dest_data_size != config->source_data_size) {
 		LOG_ERR("source and dest data size differ.");
 		return -EINVAL;
 	}
+#else
```
mathieuchopstm:
```suggestion
#else /* CONFIG_DMA_STM32_V1 */
```

drivers/dma/dma_stm32.c
```diff
@@ -344,18 +344,25 @@ DMA_STM32_EXPORT_API int dma_stm32_configure(const struct device *dev,
...
+	if (config->dest_data_size != 4U &&
+	    config->dest_data_size != 2U &&
+	    config->dest_data_size != 1U) {
+		LOG_ERR("Invalid dest unit size: %d",
```
mathieuchopstm: Nit: for consistency with other messages nearby
```suggestion
		LOG_ERR("invalid dest unit size: %d",
```

drivers/dma/dma_stm32.c
etienne-lms, line 1: Also update the inline comment added in `dma_stm32_configure()` by the referred commit?
```patch
-	/* set the data width, when source_data_size equals dest_data_size */
	int index = find_lsb_set(config->source_data_size) - 1;
+
	DMA_InitStruct.PeriphOrM2MSrcDataSize = table_p_size[index];
	index = find_lsb_set(config->dest_data_size) - 1;
	DMA_InitStruct.MemoryOrM2MDstDataSize = table_m_size[index];

benedekkupper:
Let me also add that the current implementation is mixing up the data sizes when the transfer direction is Memory to Peripheral. I found this out when implementing a led strip driver with STM32 TIM PWM and DMA, streaming memory bytes into the TIM CC register. With `source_data_size = sizeof(uint8_t), dest_data_size = sizeof(uint16_t)`, the PSIZE was set to 1 bytes, and MSIZE to 2 bytes. I had to swap them to `source_data_size = sizeof(uint16_t), dest_data_size = sizeof(uint8_t)` to make it work. Shall I report this in a separate issue / PR, or add a fix for that here?

etienne-lms (approved):

erwango (approved):

PR #101374, opened 2025-12-19
drivers: can: stm32: fdcan: enable hardware RX timestamping

skullbox305:
Enables hardware RX timestamping for the STM32 FDCAN driver by leveraging the M_CAN timestamping interface.

Introduces a new `timestamp-counter` property in the device tree binding. If `CONFIG_CAN_RX_TIMESTAMP` is enabled and this property is defined, the driver configures the M_CAN Timestamp Select to use the external timestamp counter value.

This allows linking a Zephyr Counter device to the FDCAN instance to provide precise packet timing.

### DIFF
--- a/drivers/can/can_stm32_fdcan.c
+++ b/drivers/can/can_stm32_fdcan.c
@@ -9,6 +9,7 @@
 #include <zephyr/drivers/can/can_mcan.h>
 #include <zephyr/drivers/clock_control/stm32_clock_control.h>
 #include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/counter.h>
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/kernel.h>
 #include <zephyr/sys/__assert.h>
@@ -173,6 +174,9 @@ struct can_stm32fd_config {
 	void (*config_irq)(void);
 	const struct pinctrl_dev_config *pcfg;
 	uint8_t clock_divider;
+#ifdef CONFIG_CAN_RX_TIMESTAMP
+	const struct device *external_timestamp_counter_dev;
+#endif
 };
 
 static inline uint16_t can_stm32fd_remap_reg(uint16_t reg)
@@ -520,6 +524,29 @@ static int can_stm32fd_init(const struct device *dev)
 		return ret;
 	}
 
+
+#ifdef CONFIG_CAN_RX_TIMESTAMP
+	if (stm32fd_cfg->external_timestamp_counter_dev != NULL) {
+		if (!device_is_ready(stm32fd_cfg->external_timestamp_counter_dev)) {
+			LOG_ERR("Timestamp counter device not ready");
+			return -ENODEV;
+		}
+
+		ret = counter_start(stm32fd_cfg->external_timestamp_counter_dev);
+		if (ret < 0) {
+			LOG_ERR("Failed to start timestamp counter (%d)", ret);
+			return ret;
+		}
+
+		/* Use External Timestamp counter (TSS=2) */
+		ret = can_mcan_write_reg(dev, CAN_MCAN_TSCC, FIELD_PREP(CAN_MCAN_TSCC_TSS, 0x2));
+		if (ret != 0) {
+			LOG_ERR("Failed to write TSCC register");
+			return ret;
+		}
+	}
+#endif /* CONFIG_CAN_RX_TIMESTAMP */
+
 	stm32fd_cfg->config_irq();
 
 	return ret;
@@ -609,7 +636,10 @@ static const struct can_mcan_ops can_stm32fd_ops = {
 		.pclk_len = DT_INST_NUM_CLOCKS(inst),				\
 		.config_irq = config_can_##inst##_irq,				\
 		.pcfg = PINCTRL_DT_INST_DEV_CONFIG_GET(inst),			\
-		.clock_divider = DT_INST_PROP_OR(inst, clk_divider, 0)		\
+		.clock_divider = DT_INST_PROP_OR(inst, clk_divider, 0),		\
+		IF_ENABLED(CONFIG_CAN_RX_TIMESTAMP,				\
+			   (.external_timestamp_counter_dev = DEVICE_DT_GET_OR_NULL( \
+					DT_INST_PHANDLE(inst, external_timestamp_counter)),)) \
 	};									\
 										\
 	static const struct can_mcan_config can_mcan_cfg_##inst =		\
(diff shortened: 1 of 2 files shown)


drivers/can/can_stm32_fdcan.c
```diff
@@ -495,6 +499,33 @@ static int can_stm32fd_init(const struct device *dev)
 
 	can_mcan_enable_configuration_change(dev);
 
+#ifdef CONFIG_CAN_RX_TIMESTAMP
+	/* Internal Timestamp counter (TSS=1) with Prescaler /1 (TCP=0) */
+	uint32_t tscc_val = FIELD_PREP(CAN_MCAN_TSCC_TCP, 0) | FIELD_PREP(CAN_MCAN_TSCC_TSS, 0x1);
```
henrikbrixandersen: This is handled for all M_CAN backend driver in https://github.com/zephyrproject-rtos/zephyr/pull/101396

drivers/can/can_stm32_fdcan.c
```diff
@@ -495,6 +499,33 @@ static int can_stm32fd_init(const struct device *dev)
...
+
+	if (stm32fd_cfg->timestamp_counter_dev) {
```
henrikbrixandersen:
```suggestion
	if (stm32fd_cfg->timestamp_counter_dev != NULL) {
```

drivers/can/can_stm32_fdcan.c
```diff
@@ -495,6 +499,33 @@ static int can_stm32fd_init(const struct device *dev)
...
+		}
+
+		ret = counter_start(stm32fd_cfg->timestamp_counter_dev);
+		if (ret < 0) {
+			LOG_ERR("Failed to start timestamp counter (%d)", ret);
+			return ret;
+		}
+
+		/* Use External Timestamp counter (TSS=2) */
+		tscc_val = FIELD_PREP(CAN_MCAN_TSCC_TCP, 0) | FIELD_PREP(CAN_MCAN_TSCC_TSS, 0x2);
```
henrikbrixandersen: No need for bitwise OR'ing with 0 here.

```suggestion
		tscc_val = FIELD_PREP(CAN_MCAN_TSCC_TSS, 0x2);
```

drivers/can/can_stm32_fdcan.c
henrikbrixandersen, line 525: Please move the external timestamp configuration after the call to `can_mcan_init()` to be compatible with https://github.com/zephyrproject-rtos/zephyr/pull/101396

drivers/can/can_stm32_fdcan.c
```diff
@@ -610,6 +641,11 @@ static const struct can_mcan_ops can_stm32fd_ops = {
 		.config_irq = config_can_##inst##_irq,				\
 		.pcfg = PINCTRL_DT_INST_DEV_CONFIG_GET(inst),			\
 		.clock_divider = DT_INST_PROP_OR(inst, clk_divider, 0)		\
+		IF_ENABLED(CONFIG_CAN_RX_TIMESTAMP,				\
+			   (.timestamp_counter_dev = COND_CODE_1(		\
+				    DT_INST_NODE_HAS_PROP(inst, timestamp_counter), \
+				    (DEVICE_DT_GET(DT_INST_PHANDLE(inst, timestamp_counter))), \
+				    (NULL)),)) \
```
henrikbrixandersen: You can use a conditional `DEVICE_DT_GET_OR_NULL()` here instead of the `COND_CODE_1()` macro.

dts/bindings/can/st,stm32-fdcan.yaml
```diff
@@ -43,3 +43,14 @@ properties:
       Note that the divisor is common to all 'st,stm32-fdcan' instances.
       Divide by 1 is the peripherals reset value and remains set unless
       this property is configured.
+
+  timestamp-counter:
```
henrikbrixandersen: I'd suggest naming the property `external-timestamp-counter` to avoid ambiguity.

dts/bindings/can/st,stm32-fdcan.yaml
```diff
@@ -43,3 +43,14 @@ properties:
...
+    type: phandle
+    description: |
+      Phandle to the counter device to be used as the external RX timestamp source.
+
+      When this property is defined, the driver configures the FDCAN Timestamp
+      Counter Configuration (TSCC) to use the External Timestamp (TSS=0x2).
+
+      Hardware Constraint: The selected timer must be physically routed to the
+      fdcan_ts input in the SoC interconnect. (e.g. TIM3 on STM32G4).
```
henrikbrixandersen: Perhaps add a hint as to how this can be done?

henrikbrixandersen (changes requested):
Great addition, thanks! Coincidentally, I had been meaning to fix internal timestamp counter support for the M_CAN backend in general - and add support for the external timestamp counter of the NXP LPC5500 family.

I have submitted this as https://github.com/zephyrproject-rtos/zephyr/pull/101396 but it should integrate nicely with this PR if the SoC-specific external timestamp counter configuration code is moved to after the call to `can_mcan_init()`, as it can then overwrite the default configuration set by the M_CAN backend.

dts/bindings/can/st,stm32-fdcan.yaml
skullbox305: The used timer is hard wired inside the stm32, which means the user has to look in the reference manual of the specific SoC to determine which timer to use.  I added a hint mentioning that one has to look into the reference manual

drivers/can/can_stm32_fdcan.c
```diff
@@ -520,6 +524,33 @@ static int can_stm32fd_init(const struct device *dev)
...
+
+		/* Use External Timestamp counter (TSS=2) */
+		tscc_val = FIELD_PREP(CAN_MCAN_TSCC_TSS, 0x2);
+	}
+
+	ret = can_mcan_write_reg(dev, CAN_MCAN_TSCC, tscc_val);
+	if (ret != 0) {
+		LOG_ERR("Failed to write TSCC register");
+		return ret;
+	}
```
henrikbrixandersen, line 551: You need to move this under the `if (stm32fd_cfg->external_timestamp_counter_dev != NULL)`, otherwise you'll overwrite the TSCC register value set here: https://github.com/zephyrproject-rtos/zephyr/blob/76dd50833fabb56a8de65f01629f205de6c4f155/drivers/can/can_mcan.c#L1473

henrikbrixandersen (changes requested):

drivers/can/can_stm32_fdcan.c
skullbox305: Ok done :)

henrikbrixandersen (approved):

mathieuchopstm (approved):

PR #101379, opened 2025-12-20
drivers: ethernet: dwmac: Fix Kconfig for stm32h7

tpambor:
The DWMAC driver was no longer selectable for STM32H7 series because of `CONFIG_ETH_DWMAC`'s dependency on `DT_HAS_SNPS_DESIGNWARE_ETHERNET_MMU_ENABLED`, which is only used for MMU based platforms.

This PR restructures the Kconfig to have user-selectable options for the platform-specific DWMAC drivers and select the common DWMAC driver accordingly. This way platform-specific dependencies can be tracked on these options, simplifying the logic.

Additionally, the DWMAC driver has been included in the Ethernet build-all test suite to improve pipeline coverage and help prevent future regressions.

### DIFF
--- a/tests/drivers/build_all/ethernet/testcase.yaml
+++ b/tests/drivers/build_all/ethernet/testcase.yaml
@@ -23,3 +23,12 @@ tests:
       - CONFIG_ETH_STM32_MULTICAST_FILTER=y
     platform_allow:
       - stm32h573i_dk
+
+  net.ethernet.build.stm32_dwmac:
+    extra_configs:
+      - CONFIG_ETH_STM32_HAL=n
+      - CONFIG_MDIO_ST_STM32_HAL=n
+      - CONFIG_ETH_DWMAC_STM32H7X=y
+      - CONFIG_ETH_PHY_DRIVER=n
+    platform_allow:
+      - stm32h735g_disco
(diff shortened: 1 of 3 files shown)


tests/drivers/build_all/ethernet/testcase.yaml
```diff
@@ -23,3 +23,12 @@ tests:
       - CONFIG_ETH_STM32_MULTICAST_FILTER=y
     platform_allow:
       - stm32h573i_dk
+
+  net.ethernet.build.stm32_dwmac:
+    filter: dt_compat_enabled("st,stm32-ethernet")
```
JarmouniA:
```suggestion
```
Not needed with the presence of "platform_allow".

maass-hamburg:
Before this "snps,designware-ethernet" had to be added to the compatible of the stm ethernet controller to work. With this no longer



> because of `CONFIG_ETH_DWMAC`'s dependency on `DT_HAS_SNPS_DESIGNWARE_ETHERNET_MMU_ENABLED`

where is `DT_HAS_SNPS_DESIGNWARE_ETHERNET_MMU_ENABLED`?

maass-hamburg (approved):
Ah so it fixes a 3,5 years old regression

pdgendt (approved):

PR #101387, opened 2025-12-20
video: ov5640: fix out of bounds control initialization by setting the control array first.

### LINKED ISSUE #90508: [Coverity CID: 520281] Out-of-bounds access in drivers/video/ov5640.c

https://github.com/zephyrproject-rtos/zephyr/tree/265cfb45a818f39cf03c3527b5d1b21027d3cac1/drivers/video/ov5640.c

Category: Memory - corruptions
Function: `ov5640_init_controls`
Component: Drivers
CID: [520281](https://scan9.scan.coverity.com/#/project-view/29271/12996?selectedIssue=520281)

Details:

https://github.com/zephyrproject-rtos/zephyr/blob/265cfb45a818f39cf03c3527b5d1b21027d3cac1/drivers/video/ov5640.c#L1258

https://scan9.scan.coverity.com/#/project-view/29271/12996?selectedIssue=520281

For more information about the violation, check the [Coverity Reference](https://scan9.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_OVERRUN). ([CWE-119](http://cwe.mitre.org/data/definitions/119.html))

uLipe:
fixes: #90508

### DIFF
--- a/drivers/video/ov5640.c
+++ b/drivers/video/ov5640.c
@@ -1168,6 +1168,7 @@ static int ov5640_init_controls(const struct device *dev)
 	int ret;
 	struct ov5640_data *drv_data = dev->data;
 	struct ov5640_ctrls *ctrls = &drv_data->ctrls;
+	struct video_ctrl *auto_ctrls = &ctrls->auto_gain;
 
 	ret = video_init_ctrl(&ctrls->auto_gain, dev, VIDEO_CID_AUTOGAIN,
 			      (struct video_ctrl_range){.min = 0, .max = 1, .step = 1, .def = 1});
@@ -1182,7 +1183,7 @@ static int ov5640_init_controls(const struct device *dev)
 		return ret;
 	}
 
-	video_auto_cluster_ctrl(&ctrls->auto_gain, 2, true);
+	video_auto_cluster_ctrl(auto_ctrls, 2, true);
 
 	ret = video_init_ctrl(
 		&ctrls->brightness, dev, VIDEO_CID_BRIGHTNESS,


drivers/video/ov5640.c
josuah, line 159: The alternative would have been to name the struct:

```suggestion
	struct {
		struct video_ctrl auto;
		struct video_ctrl manual;
	} gain;
```

Then `gain` effectively holds two `struct video_ctrl`, but then it needs a `(struct video_ctrl *)` cast.

josuah (approved):
This "address past struct" is how V4L2 builds Linux API and how Zephyr Video APIs are built, considering a struct of only one type as an array.

If you frown upon this, a proposed alternative is proposed, but current workaround LGTM.

Thank you!

uLipe:
Thank you @josuah , I just sent because my local Zephyr that I use in a private project was failling, happy this is helpful

drivers/video/ov5640.c
```diff
@@ -1182,7 +1183,7 @@ static int ov5640_init_controls(const struct device *dev)
 		return ret;
 	}
 
-	video_auto_cluster_ctrl(&ctrls->auto_gain, 2, true);
+	video_auto_cluster_ctrl(auto_ctrls, 2, true);
```
ngphibang, line 1186: Sorry, could you clarify what is the issue on your side so that this fix is needed ?

Also, I can't see the differences between the code before and after the fix. They seems exactly the same except an extra step of using an intermediate pointer for `&ctrls->auto_gain`.

drivers/video/ov5640.c
ngphibang: The anoymous struct is just to group the two ctrl struct together and has no other effect. It is equivalent to the array of two items. At line 1185, a pointer to `struct video_ctrl auto_gain` is passed into the function which is already a correct type (the address of the 1st element is equivalent to the address of the array). So, I don't understand what the issue is.

josuah: I think it is a matter of getting the compiler/static analysis tools happy. https://github.com/zephyrproject-rtos/zephyr/issues/90508

```c
video_auto_cluster_ctrl(&ctrls->auto_gain, 2, true);
```

The pointer `&ctrls->auto_gain` is passed, but `video_auto_cluster_ctrl()` also accesses `&ctrls->gain` after it. Which is on purpose and how the API is designed, but Coverity complains.

At least that is how I understood it.

drivers/video/ov5640.c
josuah: Probably just to fool Coverity.

ngphibang: If this really made Coverity silent so that's fine. But did you run Coverity or have some check to see if this actually fixes the Coverity issue ?

I think Coverity complains because it is afraid that we passed a single pointer but used it as an array in the function (but actually it's not a problem because we are sure that it's an array of two items). IMHO, we don't need to fix.

nashif (approved):

PR #90872, opened 2025-05-30
Add board from SCDZ with SOC from wch ch32v307 

SoftwareArchitekt:
Add a board from SCDZ named ch32v307evt
this board has a SOC from wch named ch32v307
inside SOC is a Qinke_v4f Core
There are some diffences to the wch-ch32v307 board, which I will add in the next step: Core Speed, LED, Switches

As it is my first PR for this project, forgive me if I had made mistakes

Issue #70721, opened 2024-03-26
hwmv2: Move "virtual" CPUs (e.g. ARM Trustzone) to soc.yml files


nordicjm:
At present, "virtual" CPUs (which execute a different image but are fundamentally the same CPU executing a main image) are not defined as part of hwmv2 and these are implemented at the board level with variants, generally a `ns` variant. These should instead be moved to soc.yml as this feature is a soc feature, not a board feature, and allows for additional filtering of these

PR #101390, opened 2025-12-20
Drivers: Ethernet: ch32v30x SoC

BOJIT:
This Draft PR adds Ethernet support for the CH32V307 SOC.

This has been fully tested on the CH32V307 dev board with the internal 10M PHY. Support for external PHY options has been added but I don't have the hardware to test it.

This currently depends on addition of a custom `arch_cpu_idle` implementation to fix for a hardware bug. This cannot be merged until the approach for this has been decided (https://github.com/zephyrproject-rtos/zephyr/pull/101391)


drivers/ethernet/Kconfig.wch
```diff
@@ -0,0 +1,51 @@
...
+	  This option allows to configure the priority of the RX thread that
+	  handles incoming Ethernet packets.
+	  Switching between preemptive and cooperative scheduling can be done by
+	  NET_TC_THREAD_PREEMPTIVE.
+	  Preemptive scheduling can lead to more responsive handling of network traffic,
+	  especially under high load.
+
+config ETH_WCH_HW_CHECKSUM
+	bool "Use TX and RX hardware checksum"
+	default n
```
pdgendt: Omit `default n`, not needed.

drivers/ethernet/Kconfig.wch
```diff
@@ -0,0 +1,51 @@
...
+	help
+	  Enable receive and transmit checksum offload to enhance throughput
+	  performances.
+
+config ETH_WCH_MULTICAST_FILTER
+	bool "Multicast hash filter support"
+	default n
```
pdgendt: Ditto

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+#include <zephyr/sys/printk.h>
+#include <zephyr/sys/util.h>
+#include <ethernet/eth_stats.h>
+
+#include <hal_ch32fun.h>
+
+#include "eth.h"
+
+/* Internal 10BASE-T PHY 50R*4 pull-up resistance enable or disable */
+#define ETH_Internal_Pull_Up_Res_Enable  ((uint32_t)0x00100000)
```
pdgendt: Bit flags are typically set using the `BIT` macro, and defines use uppercase.

```c
#define ETH_INTERNAL_PULL_UP_RES_ENABLE BIT(20)
```

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	}
+
+	k_mutex_lock(&data->tx_mutex, K_FOREVER); /* Only one packet can be transmitted at once */
+	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
+	k_sem_reset(&data->tx_int_sem);
+
+	size_t bytes_remaining = total_len;
+
+	do {
+		if (dma_tx_desc_current->Status & ETH_DMATxDesc_OWN) {
```
pdgendt: No implicit boolean conversions please (same for other checks)

```c
if ((dma_tx_desc_current->Status & ETH_DMATXDESC_OWN) != 0U) {
```

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	static const struct eth_wch_config eth_wch_config_##inst = {                               \
+		.regs = (ETH_TypeDef *)DT_REG_ADDR(DT_INST_PARENT(inst)),                          \
+		.phy_dev = DEVICE_DT_GET(DT_INST_PHANDLE(inst, phy_handle)),                       \
+		.clk_dev = DEVICE_DT_GET(DT_CLOCKS_CTLR(DT_INST_PARENT(inst))),                    \
+		.clk_id = DT_CLOCKS_CELL(DT_INST_PARENT(inst), id),                                \
+		.clk_tx_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_IDX(inst, 0)),                  \
+		.clk_tx_id = DT_INST_CLOCKS_CELL_BY_IDX(inst, 0, id),                              \
+		.clk_rx_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_IDX(inst, 1)),                  \
+		.clk_rx_id = DT_INST_CLOCKS_CELL_BY_IDX(inst, 1, id),                              \
+		.use_random_mac = DT_INST_PROP(inst, zephyr_random_mac_address),                   \
```
pdgendt: Use the MAC address configuration as described here: https://docs.zephyrproject.org/latest/connectivity/networking/api/mac_config.html

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+}
+#endif /* CONFIG_NET_STATISTICS_ETHERNET */
+
+static int eth_wch_init(const struct device *dev)
+{
+	struct eth_wch_data *data = dev->data;
+	const struct eth_wch_config *config = dev->config;
+	clock_control_subsys_t clock_sys;
+
+	int ret = 0;
```
pdgendt: No need to init as it's assigned first.

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+
+	/* enable clocks */
+	clock_sys = (clock_control_subsys_t *)(uintptr_t)config->clk_id;
+	ret = clock_control_on(config->clk_dev, clock_sys);
+	clock_sys = (clock_control_subsys_t *)(uintptr_t)config->clk_tx_id;
+	ret |= clock_control_on(config->clk_tx_dev, clock_sys);
+	clock_sys = (clock_control_subsys_t *)(uintptr_t)config->clk_rx_id;
+	ret |= clock_control_on(config->clk_rx_dev, clock_sys);
+
+	if (ret) {
```
pdgendt:
```c
if (ret != 0) {
```

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+static void setup_multicast_filter(const struct device *dev, const struct ethernet_filter *filter)
+{
+	/* TODO implement */
+}
+#endif /* CONFIG_ETH_WCH_MULTICAST_FILTER */
+
+static void setup_mac_filter(ETH_TypeDef *eth)
+{
+	/* TODO implement */
+}
```
maass-hamburg, line 167: either implement it in that commit or remove it. no placeholders

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	eth_mac_init(dev);
+	setup_mac_filter(config->regs);
+
+	net_if_carrier_off(iface);
+	net_lldp_set_lldpdu(iface);
+
+	/* Initialize semaphores */
+	k_mutex_init(&data->tx_mutex);
+	k_sem_init(&data->rx_int_sem, 0, K_SEM_MAX_LIMIT);
+	k_sem_init(&data->tx_int_sem, 0, K_SEM_MAX_LIMIT);
```
maass-hamburg: move to regular init (not iface)

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	if (device_is_ready(config->phy_dev)) {
+		phy_link_callback_set(config->phy_dev, phy_link_state_changed, (void *)dev);
+	} else {
+		LOG_ERR("PHY device not ready");
+	}
+
+	if (is_first_init) {
+		/* Now that the iface is setup, we are safe to enable IRQs. */
+		__ASSERT_NO_MSG(config->irq_config_func != NULL);
+		config->irq_config_func(dev);
```
maass-hamburg: irq init should be done in the regular init too

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	ETH_TypeDef *eth = dev_config->regs;
+
+	switch (type) {
+	case ETHERNET_CONFIG_TYPE_MAC_ADDRESS:
+		memcpy(data->mac_addr, config->mac_address.addr, 6);
+		eth->MACA0HR = (data->mac_addr[5] << 8) | data->mac_addr[4];
+		eth->MACA0LR = (data->mac_addr[3] << 24) | (data->mac_addr[2] << 16) |
+			       (data->mac_addr[1] << 8) | data->mac_addr[0];
+		net_if_set_link_addr(data->iface, data->mac_addr, sizeof(data->mac_addr),
+				     NET_LINK_ETHERNET);
```
maass-hamburg: move into seperate function and use here and in iface init

drivers/ethernet/Kconfig.wch
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2025 James Bennion-Pedley <james@bojit.org>
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig ETH_WCH
+	bool "WCH Ethernet driver"
+	default y
+	depends on DT_HAS_WCH_CH32_ETHERNET_ENABLED
+	select PM
```
maass-hamburg: PM should be optional and not selected by a driver.

maass-hamburg (changes requested):

drivers/ethernet/Kconfig.wch
BOJIT: Just to check my understanding, if `CONFIG_PM`, is not enabled, can the device never drop down from `PM_STATE_ACTIVE`? Due to a hardware quirk on the CH32s any `WFI` instruction will break ongoing DMA transactions, hence the custom `arch_cpu_idle` implementation.
(https://github.com/zephyrproject-rtos/zephyr/pull/101391)

I presume should I put the [PM locks](https://github.com/zephyrproject-rtos/zephyr/pull/101390/files#diff-40dcc11ffd66ee17758c6ee07a8a42237409da3d49f52a5fc6d28000f52e661dR332) inside `#ifdef CONFIG_PM` guards? then by default if `CONFIG_PM` is not enabled the device just busy-loops in idle.

dts/riscv/wch/ch32v307/ch32v307.dtsi
```diff
@@ -218,6 +229,30 @@
...
+			reg = <0x40028000 0x2000>;
+			compatible = "wch,ch32-ethernet-controller";
+			clocks = <&rcc CH32V20X_V30X_CLOCK_ETHMAC>;
+
+			mac: ethernet {
+				compatible = "wch,ch32-ethernet";
+				interrupt-parent = <&pfic>;
+				interrupts = <77 78>;
+				clocks = <&rcc CH32V20X_V30X_CLOCK_ETHMACTX>,
+					 <&rcc CH32V20X_V30X_CLOCK_ETHMACRX>;
```
maass-hamburg: probably better to have all clocks on the parent.

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,904 @@
...
+	ETH_WCH_IRQ_HANDLER_DECL(inst)                                                             \
+	static const struct eth_wch_config eth_wch_config_##inst = {                               \
+		.regs = (ETH_TypeDef *)DT_REG_ADDR(DT_INST_PARENT(inst)),                          \
+		.phy_dev = DEVICE_DT_GET(DT_INST_PHANDLE(inst, phy_handle)),                       \
+		.clk_dev = DEVICE_DT_GET(DT_CLOCKS_CTLR(DT_INST_PARENT(inst))),                    \
+		.clk_id = DT_CLOCKS_CELL(DT_INST_PARENT(inst), id),                                \
+		.clk_tx_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_IDX(inst, 0)),                  \
+		.clk_tx_id = DT_INST_CLOCKS_CELL_BY_IDX(inst, 0, id),                              \
+		.clk_rx_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR_BY_IDX(inst, 1)),                  \
+		.clk_rx_id = DT_INST_CLOCKS_CELL_BY_IDX(inst, 1, id),                              \
```
maass-hamburg: propably better to name the clocks and refer them by name

drivers/ethernet/Kconfig.wch
maass-hamburg: > I presume should I put the [PM locks](https://github.com/zephyrproject-rtos/zephyr/pull/101390/files#diff-40dcc11ffd66ee17758c6ee07a8a42237409da3d49f52a5fc6d28000f52e661dR332) inside `#ifdef CONFIG_PM` guards? then by default if `CONFIG_PM` is not enabled the device just busy-loops in idle.

no need for that, it is already handled that CONFIG_PM is deactivated in include/zephyr/pm/policy.h

u only need to remove the `select PM` then.

BOJIT:
> > As for the H417, I can't find the reference manual
> 
> Only in chinese last i checked, i did the work for it on ch32fun side but not even started porting it to zephyr.
> 
> > Yep, it should work, the register maps are the same.
> 
> Great 👍 !

@VynDragon sorry, had another look at the reference manual and it turns out that the V203/208 Ethernet peripheral is completely different register maps, so won't be compatible :(. This driver will only support the V307, V317 and F207.

BOJIT:
> You need one of the zephyr badge for the ci to run without someone manually triggering it

Ah, I see. Is that this one? [Claim Your Zephyr Badge](https://forms.gle/oCw9iAPLhUsHTapc8)? Or the [nomination template](https://github.com/zephyrproject-rtos/zephyr/issues/new?assignees=&labels=Role+Nomination&template=006_nomination.yml)?

I have merged a handful of patches in the past.

VynDragon:
No it's the github one, possibly 'contributor' badge you get after one merged PR (if you merged one and didnt get it you probably should ask for it), and also automatically if you are put in MAINTAINERS.yml

VynDragon:
I asked on discord and they said it should've been triggering... Probably will fix itself in the next PR or something. Either way contirbutor badge is good to have.

drivers/ethernet/Kconfig.wch
```diff
@@ -0,0 +1,49 @@
...
+
+config ETH_WCH_HW_CHECKSUM
+	bool "Use TX and RX hardware checksum"
+	help
+	  Enable receive and transmit checksum offload to enhance throughput
+	  performances.
+
+config ETH_WCH_MULTICAST_FILTER
+	bool "Multicast hash filter support"
+	imply CRC
```
maass-hamburg:
```suggestion
	select CRC
```
looks like it is needed, therefore it has to be slected

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+	uint8_t internal_phy_prediv;
+
+	const struct pinctrl_dev_config *pin_cfg;
+	void (*irq_config_func)(const struct device *dev);
+};
+
+struct eth_wch_data {
+	struct net_if *iface;
+	uint8_t mac_addr[NET_ETH_ADDR_LEN];
+	struct k_mutex tx_mutex;
```
maass-hamburg: tx mutex is not needed, the net subsystem already provides one. That's why it also was removed from the other ethernet drivers.

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+
+	/* Certain initialisation is only done once per interface */
+	bool is_first_init = false;
+
+	if (data->iface == NULL) {
+		data->iface = iface;
+		is_first_init = true;
+	}
+
+	if (is_first_init) {
```
maass-hamburg: combine this if with the if before, not need for this extra bool var.

drivers/ethernet/Kconfig.wch
```diff
@@ -0,0 +1,49 @@
...
+	default 2
+	help
+	  This option allows to configure the priority of the RX thread that
+	  handles incoming Ethernet packets.
+	  Switching between preemptive and cooperative scheduling can be done by
+	  NET_TC_THREAD_PREEMPTIVE.
+	  Preemptive scheduling can lead to more responsive handling of network traffic,
+	  especially under high load.
+
+config ETH_WCH_HW_CHECKSUM
```
maass-hamburg, line 31: enable it by default. Or why not?

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+struct eth_wch_config {
+	ETH_TypeDef *regs;
+	const struct device *phy_dev;
+
+	const struct device *clk_dev;
+	uint8_t clk_id;
+	const struct device *clk_tx_dev;
+	uint8_t clk_tx_id;
+	const struct device *clk_rx_dev;
+	uint8_t clk_rx_id;
```
maass-hamburg, line 47: group the structs and the uint8_t together for a more efficient use

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+		eth_stats_update_errors_tx(data->iface);
+		LOG_ERR("Packet spans all available descriptors");
+		return -ENOBUFS;
+	}
+
+	k_mutex_lock(&data->tx_mutex, K_FOREVER); /* Only one packet can be transmitted at once */
+	pm_policy_state_lock_get(PM_STATE_RUNTIME_IDLE, PM_ALL_SUBSTATES);
+	k_sem_reset(&data->tx_int_sem);
+
+	size_t bytes_remaining = total_len;
```
maass-hamburg: Var declaration should be at the beginning of the function

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+				(void *)dev, NULL, NULL,
+				K_PRIO_COOP(CONFIG_ETH_WCH_HAL_RX_THREAD_PRIO), 0, K_NO_WAIT);
+
+		k_thread_name_set(&data->rx_thread, dev->name);
+	}
+
+	/* Initialise interface with relevant hardware settings */
+	ethernet_init(iface);
+	eth_mac_init(dev);
+	set_mac_addr(config->regs, data->mac_addr, iface);
```
maass-hamburg, line 542: Do setting the mac address and init of the mac before ethernet_init.

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+
+	eth->MACCR &= ~(ETH_MACCR_TE | ETH_MACCR_RE);
+	eth->DMAOMR &= ~(ETH_DMAOMR_ST | ETH_DMAOMR_SR);
+
+	return 0;
+}
+
+static int get_hw_mac(uint8_t *mac_addr)
+{
+	uint8_t *mac_base = (uint8_t *)(ROM_CFG_USERADR_ID);
```
maass-hamburg, line 428: Is it documented, that a mac address is saved there?

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+	int res;
+
+	ARG_UNUSED(unused1);
+	ARG_UNUSED(unused2);
+
+	while (1) {
+		res = k_sem_take(&data->rx_int_sem, K_FOREVER);
+		if (res == 0) {
+			while ((pkt = eth_rx(dev)) != NULL) {
+				iface = net_pkt_iface(pkt);
```
maass-hamburg: Can't you use the iface from the dev data struct and assign it before the while loop?

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,786 @@
...
+				res = net_recv_data(iface, pkt);
+				if (res < 0) {
+					eth_stats_update_errors_rx(net_pkt_iface(pkt));
```
maass-hamburg: Use the iface from data

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,158 @@
...
+	tmpreg = config->regs->MACMIIAR;
+	tmpreg &= ~MACMIIAR_CR_MASK;
+	tmpreg |= (((uint32_t)prtad << ETH_MACMIIAR_PA_POS) & ETH_MACMIIAR_PA);
+	tmpreg |= (((uint32_t)regad << ETH_MACMIIAR_MR_POS) & ETH_MACMIIAR_MR);
+	tmpreg |= ETH_MACMIIAR_MW;
+	tmpreg |= ETH_MACMIIAR_MB;
+	config->regs->MACMIIDR = data;
+	config->regs->MACMIIAR = tmpreg;
+
+	uint32_t read_start = k_uptime_get_32();
```
maass-hamburg: There is a timepoint* api use that.

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,158 @@
...
+
+	do {
+		if ((k_uptime_get_32() - read_start) > PHY_RESPONSE_TIMEOUT_MS) {
+			k_sem_give(&dev_data->sem);
+			LOG_ERR("Write: PHY Timeout!");
+			return -EIO;
+		}
+
+		k_yield();
```
maass-hamburg: use a k_sleep here instead with a time. Maybe you can find out how long these typically take, as mdio is a pretty slow bus

dts/riscv/wch/ch32v307/ch32v307.dtsi
```diff
@@ -198,6 +202,13 @@
 			status = "okay";
 		};
 
+		rng: rng@40023c00 {
+			compatible = "wch,rng";
+			reg = <0x40023c00 12>;
+			clocks = <&rcc CH32V20X_V30X_CLOCK_RNG>;
```
maass-hamburg, line 263: Don't disable it, if it is added to the chosen.

dts/riscv/wch/ch32v303/ch32v303.dtsi
```diff
@@ -198,6 +198,13 @@
 			status = "okay";
 		};
 
+		rng: rng@40023c00 {
+			compatible = "wch,rng";
+			reg = <0x40023c00 12>;
+			clocks = <&rcc CH32V20X_V30X_CLOCK_RNG>;
+			status = "disabled";
```
maass-hamburg, line 249: Should be enabled by default, as there are probably no external dependencies

maass-hamburg (changes requested):

dts/riscv/wch/ch32v307/ch32v307.dtsi
BOJIT: Can I set it to status = okay by default if its parent clock is disabled?
Or should I rely on the board DTS to add it to the chosen?
Looks like the ST DTSI does the same thing: https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/st/f7/stm32f7.dtsi#L914

drivers/ethernet/eth_wch.c
BOJIT: Yes - this does point to an IANA assigned MAC address. As is the case with all the WCH resources, the english documentation is very sparse.
A note in MounRiver Studio states that the USERADR contains a valid MAC. WCH make Ethernet ICs and have 8 OUIs, so don't seem to mind putting a unique MAC in every MCU.

drivers/ethernet/Kconfig.wch
BOJIT: I've been seeing some strange intermittent IP checksum failures on on the TX path when enabling TCP checksum offload.
I will change the capabilities to only include RX checksum offload with this config option, as that is very reliable.

drivers/mdio/mdio_wch.c
BOJIT: Have changed over to timepoint. On my board + PHY a transaction takes approx. 20ms. But I think it can vary depending on how the clock is configured. A 10ms checking interval is probably appropriate.

dts/riscv/wch/ch32v307/ch32v307.dtsi
maass-hamburg: okay then leave it like that

dts/riscv/wch/ch32v303/ch32v303.dtsi
maass-hamburg: okay then leave it like that

drivers/ethernet/eth_wch.c
maass-hamburg: then it is fine

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,158 @@
...
+
+	do {
+		if (sys_timepoint_expired(timeout)) {
+			k_sem_give(&dev_data->sem);
+			LOG_ERR("Write: PHY Timeout!");
+			return -EIO;
+		}
+
+		k_msleep(10);
+	} while (config->regs->MACMIIAR & ETH_MACMIIAR_MB);
```
maass-hamburg: it seems like this is also used in write, make it a shared function

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,145 @@
...
+		tmpreg |= ETH_MACMIIAR_MW;
+		config->regs->MACMIIDR = *data;
+	} else {
+		tmpreg &= ~ETH_MACMIIAR_MW;
+	}
+
+	tmpreg |= ETH_MACMIIAR_MB;
+	config->regs->MACMIIAR = tmpreg;
+
+	k_timepoint_t timeout = sys_timepoint_calc(K_MSEC(PHY_RESPONSE_TIMEOUT_MS));
```
maass-hamburg: move declaration to the bedinning

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,145 @@
...
+
+static int mdio_wch_transfer(const struct device *dev, uint8_t prtad, uint8_t regad, uint16_t *data,
+			     bool write)
+{
+	struct mdio_wch_data *const dev_data = dev->data;
+	const struct mdio_wch_config *const config = dev->config;
+
+	k_sem_take(&dev_data->sem, K_FOREVER);
+
+	uint32_t tmpreg = 0;
```
maass-hamburg: here too

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,145 @@
...
+
+#include <stdint.h>
+#include <errno.h>
+#include <zephyr/device.h>
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/drivers/mdio.h>
+#include <zephyr/net/ethernet.h>
+#include <zephyr/net/mdio.h>
```
maass-hamburg: remove, you don't use these

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright (c) 2025 James Bennion-Pedley <james@bojit.org>
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT wch_mdio
+
+#define LOG_LEVEL CONFIG_MDIO_LOG_LEVEL
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(mdio_wch, LOG_LEVEL);
```
maass-hamburg:
```suggestion
#include <zephyr/logging/log.h>
LOG_MODULE_REGISTER(mdio_wch, CONFIG_MDIO_LOG_LEVEL);
```

drivers/mdio/mdio_wch.c
```diff
@@ -0,0 +1,145 @@
...
+
+#include <stdint.h>
+#include <errno.h>
```
maass-hamburg: is already included with #include <zephyr/kernel.h>

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,775 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define DT_DRV_COMPAT wch_ethernet
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(ethernet_wch, CONFIG_ETHERNET_LOG_LEVEL);
+
+#include <errno.h>
+#include <stdbool.h>
```
maass-hamburg: already included by zephyr/kernel.h

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,775 @@
...
+#include <zephyr/device.h>
+#include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/hwinfo.h>
```
maass-hamburg: not used here

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,775 @@
...
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/kernel.h>
+#include <zephyr/irq.h>
```
maass-hamburg: already included by zephyr/kernel.h

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,775 @@
...
+#include <zephyr/net/ethernet.h>
+#include <zephyr/net/lldp.h>
+#include <zephyr/net/net_if.h>
+#include <zephyr/net/net_pkt.h>
```
maass-hamburg: part of zephyr/net/ethernet.h

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,775 @@
...
+#include <zephyr/net/phy.h>
+#include <zephyr/pm/policy.h>
+#include <zephyr/sys/crc.h>
+#include <zephyr/sys/printk.h>
```
maass-hamburg: unused

maass-hamburg (changes requested):

dts/riscv/wch/ch32v307/ch32v307.dtsi
Colt-Ma, line 1: The ch32v208 also has a 10M Ethernet interface and a built-in PHY. It would be great if this code could be adapted to the ch32v208.

BOJIT: Hi @Colt-Ma ,
I'm afraid the CH32V203/208 Ethernet is a completely different peripheral.
The CH32F207/V307/V317 uses this synopsys IP: https://www.synopsys.com/designware-ip/interface-ip/ethernet/ether-mac10-100-universal.html (pretty common, is virtually identical to the STM32F4/F7 MAC)

The CH32F208/V203/V208 MCUs use a much more limited peripheral that is probably best managed as a separate driver rather than trying to make the same driver work for both.

Colt-Ma: @BOJIT 
Is there a plan for an Ethernet driver specifically for the CH32V208?

BOJIT: Hi @Colt-Ma,
At some point, yes, but it's not something I'm planning to work on at the moment. I'm keen to get better coverage of SPI/I2C/Timer/USB support across the WCH family before looking at this. The main reason I'm working on CH32V3x7 Ethernet is to utilise the PTP Timestamping feature of the MAC, which isn't present on the CH32V208 MAC.

maass-hamburg (changes requested):
> This has been fully tested on the CH32V307 dev board with the internal 10M PHY

I suppose that is the CH32V307V-EVT-R1?

Can you add that? Because I currently don't see there being a board in zephyr with the CH32V307 and without that the ethernet driver will not be build tested by the ci. Every ethernet driver has to be build tested by the ci.

dts/riscv/wch/ch32v307/ch32v307.dtsi
Colt-Ma: @BOJIT That's great, looking forward to your wonderful work.

BOJIT:
> > This has been fully tested on the CH32V307 dev board with the internal 10M PHY
> 
> I suppose that is the CH32V307V-EVT-R1?
> 
> Can you add that? Because I currently don't see there being a board in zephyr with the CH32V307 and without that the ethernet driver will not be build tested by the ci. Every ethernet driver has to be build tested by the ci.

I am using an SCDZ dev kit, virtually identical to the official EVT. The board was attempted to be added here by someone else (https://github.com/zephyrproject-rtos/zephyr/pull/90872) but the PR stalled.

I am happy to add a tidied version of this board, but can I request to get these two supporting PRs merged first?
- https://github.com/zephyrproject-rtos/zephyr/pull/101688
- https://github.com/zephyrproject-rtos/zephyr/pull/101391

Currently I'm already trying to juggle multiple unmerged interdependent patches and it's become a bit hard to manage.

loblik:
> I suppose that is the CH32V307V-EVT-R1?
> 
> Can you add that? Because I currently don't see there being a board in zephyr with the CH32V307

I just opened a PR for that one https://github.com/zephyrproject-rtos/zephyr/pull/103651 .

jpedley:
> > I suppose that is the CH32V307V-EVT-R1?
> > Can you add that? Because I currently don't see there being a board in zephyr with the CH32V307
> 
> I just opened a PR for that one #103651 .

@loblik would you mind testing Ethernet on this board if I add the DTS changes to your board PR?

Also FYI a lot of the Ethernet examples need the RAM/FLASH size configured larger to run on the WCH boards ([details here](https://blog.dan.drown.org/ch32v307-dev-board-part-5/))

I'm still wondering what the best way to handle this in Zephyr is: either board variants or perhaps snippets like [espressif](https://docs.zephyrproject.org/latest/boards/espressif/common/board-variants.html) uses? Not sure how these would integrate with the CI tests though.

loblik:
> @loblik would you mind testing Ethernet on this board if I add the DTS changes to your board PR?

Sure.

maass-hamburg:
> @maass-hamburg is this kind of thing accepatble?

I don't think so. Kconfigs can't be used in the dt. see https://docs.zephyrproject.org/latest/build/cmake/index.html#configuration-phase the dt is processed before the Kconfig.

BOJIT:
> Do I really need to do some pin configuration, when using internal 10Base PHY?

Yep, as the output pins of the PHY can apparently be remapped to do other things (think this is not the case on the V317 100M PHY though)

image

My board support looks like this:

```dts
// Pinctrl
&pinctrl {
	mac_default: mac_default {
		mac_internal_rxp {
			pinmux = <CH32V20X_V30X_PINMUX_DEFINE(PC, 6, ETH, 0)>;
			slew-rate = "max-speed-50mhz";
		};

		mac_internal_rxn {
			pinmux = <CH32V20X_V30X_PINMUX_DEFINE(PC, 7, ETH, 0)>;
			slew-rate = "max-speed-50mhz";
		};

		mac_internal_txp {
			pinmux = <CH32V20X_V30X_PINMUX_DEFINE(PC, 8, ETH, 0)>;
			slew-rate = "max-speed-50mhz";
		};

		mac_internal_txn {
			pinmux = <CH32V20X_V30X_PINMUX_DEFINE(PC, 9, ETH, 0)>;
			slew-rate = "max-speed-50mhz";
		};
	};

	mdio_default: mdio_default {
		/* No MDIO pinmux required for internal PHY */
	};
};

// Main DTS
&mac {
	status = "okay";
	pinctrl-0 = <&mac_default>;
	pinctrl-names = "default";
	phy-connection-type = "internal";
	phy-handle = <&eth_phy>;
};

&mdio {
	status = "okay";
	pinctrl-0 = <&mdio_default>;
	pinctrl-names = "default";

	eth_phy: ethernet-phy@0 {
		compatible = "ethernet-phy";
		reg = <0x00>;
	};
};
```

Note that the MDIO driver treats the internal PHY like an external PHY. It still uses the same physical bus, it's just it's muxed to the internal PHY when it's enabled.

Hang tight! I'm planning to add ETH support to your board as part of this PR this weekend.

BOJIT:
@loblik I've added the additions to your board on the end of this PR.

The first commit adds the Ethernet peripheral to the board, the second adds variants for the different Flash/RAM splits on the SOC.

A lot of the network samples overflow the `32k` default RAM config. You can build this others with: `-b ch32v307v_evt_r1/ch32v307/ram_96k`, etc...

It doesn't seem ideal doing this at the board level, but that was the approach suggested to me here: https://github.com/zephyrproject-rtos/zephyr/issues/70721

BOJIT:
@maass-hamburg This should now have all the changes required. Unless @loblik has any concerns with the board additions I think this is ready for review.

boards/wch/ch32v307v_evt_r1/ch32v307v_evt_r1_defconfig
```diff
@@ -6,3 +6,4 @@ CONFIG_GPIO=y
 CONFIG_SERIAL=y
 CONFIG_CONSOLE=y
 CONFIG_UART_CONSOLE=y
+CONFIG_NET_L2_ETHERNET=y
```
maass-hamburg: not needed, already covered by the Kconfig.defconfig. Also we don't want the activated when we don't have Networking enabled.

boards/wch/ch32v307v_evt_r1/ch32v307v_evt_r1.yaml
```diff
@@ -6,7 +6,11 @@ toolchain:
...
 ram: 32
-flash: 256
+flash: 480
 supported:
   - gpio
   - i2c
+  - netif:eth
+  - uart
+  - spi
+vendor: wch
```
maass-hamburg, line 16: I think you need this file for every variant https://docs.zephyrproject.org/latest/develop/test/twister.html#board-configuration

drivers/ethernet/mdio/mdio_wch.c
```diff
@@ -0,0 +1,138 @@
...
+		tmpreg &= ~ETH_MACMIIAR_MW;
+	}
+
+	tmpreg |= ETH_MACMIIAR_MB;
+	config->regs->MACMIIAR = tmpreg;
+
+	do {
+		if (sys_timepoint_expired(timeout)) {
+			k_sem_give(&dev_data->sem);
+			LOG_ERR("Write: PHY Timeout!");
```
maass-hamburg: is read and write, also it is mdio, not phy. Make it something like "MDIO Timeout"

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,764 @@
...
+
+	tmpreg &= ~ETH_MACCR_DM;
+	if (PHY_LINK_IS_FULL_DUPLEX(state->speed)) {
+		tmpreg |= ETH_MACCR_DM;
+	}
+
+	tmpreg &= ~ETH_MACCR_FES_MASK;
+	tmpreg |= PHY_LINK_IS_SPEED_1000M(state->speed)  ? ETH_MACCR_FES_1000M
+		  : PHY_LINK_IS_SPEED_100M(state->speed) ? ETH_MACCR_FES_100M
+							 : ETH_MACCR_FES_10M;
```
maass-hamburg, line 453: shouldn't we write the value of `tmpreg` back to `eth->MACCR` to set it? Otherwise this seeems pretty useless

dts/bindings/ethernet/wch,ethernet.yaml
```diff
@@ -0,0 +1,63 @@
...
+    required: true
+  phy-handle:
+    required: true
+  internal-phy-pllmul:
+    type: int
+    description: |
+      PLL Multiplier of the Internal PHY PLL (PLL3). Adjust so PHY clocks at 60 MHz
+    enum:
+      - 2.5
+      - 12.5
```
maass-hamburg: these are not integers. but you could make these strings.

Also it seems like, you don't need the value, but just the index of the enum, you can use `DT_INST_ENUM_IDX()`. Just keep the values here in the right order.

Then you can also remove these lines:
```
static const uint8_t phy_pllmul_lut[] = {2, 0, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 20};
static const uint8_t phy_prediv_lut[] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16};
```

drivers/ethernet/eth_wch.c
```diff
@@ -0,0 +1,764 @@
...
+			    DEVICE_DT_INST_GET(idx), 0);                                           \
+		irq_enable(DT_INST_IRQN(idx));                                                     \
+	}
+
+#define ETH_WCH_DEVICE(inst)                                                                       \
+	BUILD_ASSERT(DT_INST_ENUM_HAS_VALUE(inst, phy_connection_type, mii) ||                     \
+			     DT_INST_ENUM_HAS_VALUE(inst, phy_connection_type, rmii) ||            \
+			     DT_INST_ENUM_HAS_VALUE(inst, phy_connection_type, rgmii) ||           \
+			     DT_INST_ENUM_HAS_VALUE(inst, phy_connection_type, internal),          \
+		     "Unsupported PHY connection type");                                           \
```
maass-hamburg, line 729: it seems like we only need to know, if the `phy_connection_type` is `internal` or not. Maybe leave that prop optional.

dts/riscv/wch/ch32v307/ch32v307.dtsi
```diff
@@ -273,6 +284,31 @@
...
+			compatible = "wch,ethernet-controller";
+			clocks = <&rcc CH32V20X_V30X_CLOCK_ETHMAC>,
+				 <&rcc CH32V20X_V30X_CLOCK_ETHMACTX>,
+				 <&rcc CH32V20X_V30X_CLOCK_ETHMACRX>;
+			clock-names = "mac", "tx", "rx";
+
+			mac: ethernet {
+				compatible = "wch,ethernet";
+				interrupt-parent = <&pfic>;
+				interrupts = <77 78>;
```
maass-hamburg:
```suggestion
				interrupts = <77>, <78>;
```
be consistent

maass-hamburg (changes requested):

drivers/ethernet/eth_wch.c
BOJIT: Turns out this was an oversight on my part. We do need to know if the interface is RMII/RGMII, there's an additional EXTEN register than needs to be set for Gigabit, I'll add this.

(Note that the Gigabit mode is largely untested, I don't have any dev boards that have the Gigabit PHY).

drivers/ethernet/eth_wch.c
BOJIT: Thanks for spotting, this got lost during a refactor!
Doesn't really show up as a problem when I test as I'm usually using PHY autonegotiation.

drivers/ethernet/eth_wch.c
BOJIT: Have now replaced this with a string prop for the connection type, similar to logic in the ESP32 ethernet driver. I've also added the register inits for Gigabit mode, but these are untested.

BOJIT:
@maass-hamburg if you get a chance could you take another look? Ideally trying to get this merged before starting some wch-platform-wide refactors.

maass-hamburg (approved):

BOJIT:
@pdgendt is it possible for you to have a look at this PR?
Sorry to keep pestering, but currently this PR is blocking https://github.com/zephyrproject-rtos/zephyr/pull/105534 along with a few other PTP-related features that I'm trying to get added to the WCH platform.

ceolin (approved):
+1 for the entropy changes.

I removed myself as the assignee because there are bigger changes than entropy in this pr.

PR #94286, opened 2025-08-08
drivers: usb: udc: add wch usbfs driver.

miggazElquez:
This is a draft PR for adding a a driver for the usbfs device.

There is some small things left to fix, the main problem is that I need to put a sleep in some places to get the driver to works: otherwise it seems like the device is only sending 0s. I've indicated these place with a `/* FIXME */` comment.

PR #101391, opened 2025-12-20
soc: wch: Add idle gating functionality to fix hardware bug.

BOJIT:
In CH32V chips, entering idle via `wfi` will break ongoing DMA transactions by gating the clocks of certain MCU subsystems.

This is described very briefly in section 2.3.2 of the reference manual and as far as I can tell it cannot be disabled for the `Quinke V4` processor core.

image

This PR adds an API to let specific drivers prevent `wfi` idle during an ongoing transaction. It is modelled on the similar behaviour that has been implemented on the ITE Embedded Controllers: https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/ite/ec/it51xxx/soc.c#L91

This does mean defining `CONFIG_ARCH_HAS_CUSTOM_CPU_IDLE` for the CH32V family, but the `chip_idle` code is only enabled if the symbol `SOC_CH32V_CPU_IDLE_GATING` is set. The ITE chips manage this by only setting this symbol in drivers that require it ([ITE example](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/i2c/Kconfig.it8xxx2#L46)).

This work needs to be done for high speed peripherals to be properly supported:
Ethernet Driver PR: https://github.com/zephyrproject-rtos/zephyr/pull/101390
USB FS PR: https://github.com/zephyrproject-rtos/zephyr/pull/94286

It would be good to get some WCH maintainers' opinion on this approach.


bjarki-andreasen (commented):
Implement a PM sleep state for idle, then lock that state using pm policy state lock from drivers which break if WFI is called while enabled. This scales to more sleep states and more quirks than this PR.

peter-mitsis (dismissed):

bjarki-andreasen (dismissed):

maass-hamburg (dismissed):

maass-hamburg (approved):

fkokosinski (approved):

PR #101392, opened 2025-12-21
counter: mcux_lptmr: Fix prescaler/glitch filter configuration

Holt-Sun:
and frequency calculation

- Change prescale-glitch-filter-bypass from boolean to int enum [0, 1] to align with hardware register representation
- Fix prescaler division calculation in time counter mode to use 2^(value + 1) instead of 2^value
- Fix glitch filter calculation in pulse counter mode to use 2^value
- Introduce MCUX_LPTMR_EFFECTIVE_FREQ macro to correctly calculate counter frequency based on mode and bypass settings
- Remove incorrect prescaler_glitch offset calculation that added timer_mode_sel - 1
- Update prescale-glitch-filter enum range from [0-16] to [0-15] as pulse mode cannot support value 16
- Improve BUILD_ASSERT error message for prescale-glitch-filter range
- Update devicetree binding documentation to clarify prescaler and glitch filter behavior

This fixes incorrect counter frequency reporting and ensures the prescaler/glitch filter values are correctly mapped to hardware register settings according to the NXP LPTMR reference manual.

### DIFF
--- a/doc/releases/migration-guide-4.4.rst
+++ b/doc/releases/migration-guide-4.4.rst
@@ -93,6 +93,139 @@ Controller Area Network (CAN)
   * :kconfig:option:`CONFIG_CAN_STM32_FDCAN_MAX_STD_ID_FILTERS` for :dtcompatible:`st,stm32-fdcan`
   * :kconfig:option:`CONFIG_CAN_XMC4XXX_MAX_FILTERS` for :dtcompatible:`infineon,xmc4xxx-can-node`
 
+Counter
+=======
+
+* The NXP LPTMR driver (:dtcompatible:`nxp,lptmr`) has been updated to fix incorrect
+  prescaler and glitch filter configuration:
+
+  * The ``prescale-glitch-filter`` property valid range changed from ``[0-16]`` to ``[0-15]``.
+    The value ``16`` was invalid for pulse counter mode and has been removed. Device trees using
+    value ``16`` must be updated to use values in the range ``[0-15]``.
+
+  * A new boolean property ``prescale-glitch-filter-bypass`` has been introduced to explicitly
+    control prescaler/glitch filter bypass. Previously, setting ``prescale-glitch-filter = <0>``
+    implicitly enabled bypass mode, which was ambiguous.
+
+    In v4.4 and later, bypass is controlled only by the presence of
+    ``prescale-glitch-filter-bypass``. If the property is absent, the prescaler/glitch filter is
+    active and ``prescale-glitch-filter`` is applied.
+
+  * The prescaler/glitch filter behavior has been clarified:
+
+    * In Time Counter mode: prescaler divides the clock by ``2^(prescale-glitch-filter + 1)``
+    * In Pulse Counter mode: glitch filter recognizes change after ``2^prescale-glitch-filter``
+      rising edges (value 0 is not supported for glitch filtering)
+
+  * All in-tree device tree nodes have been updated to use ``prescale-glitch-filter-bypass;``
+    instead of ``prescale-glitch-filter = <0>;``. Out-of-tree boards should be updated
+    accordingly.
+
+  * If both ``prescale-glitch-filter-bypass`` and ``prescale-glitch-filter`` are set,
+    bypass mode takes precedence and the ``prescale-glitch-filter`` value is ignored.
+
+  Example migration:
+
+  .. code-block:: devicetree
+
+     /* Old (deprecated) */
+     lptmr0: counter@40040000 {
+         compatible = "nxp,lptmr";
+         /* Implicitly bypassed */
+         prescale-glitch-filter = <0>;
+     };
+
+     /* New (correct) */
+     lptmr0: counter@40040000 {
+         compatible = "nxp,lptmr";
+         /* Explicitly bypassed */
+         prescale-glitch-filter-bypass;
+     };
+
+  .. rubric:: Examples of using ``prescale-glitch-filter``
+
+  .. note::
+
+     ``prescale-glitch-filter-bypass`` is a boolean. If present, bypass is enabled. If absent,
+     bypass is disabled and ``prescale-glitch-filter`` is applied.
+
+     In Pulse Counter mode, ``prescale-glitch-filter = <0>`` is not a supported glitch filter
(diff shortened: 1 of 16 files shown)


dts/bindings/counter/nxp,lptmr.yaml
```diff
@@ -52,21 +52,29 @@ properties:
     required: true
     description: Represents the width of the counter in bits.
 
+  prescale-glitch-filter-bypass:
```
FelixWang47831, line 55: Should be boolean type.
In init function, lptmr_config.bypassPrescaler = config->bypass_prescaler_glitch;
The "bypassPrescaler" is bool data.

Holt-Sun: Update to bool type.

FelixWang47831 (approved):

ZhaoxiangJin (dismissed):

Holt-Sun:

dts/bindings/counter/nxp,lptmr.yaml
```diff
@@ -52,21 +52,27 @@ properties:
...
-        from the input-pin value.
-      Note, that the pulse mode cannot be 2 ^ 16.
+      Configures the size of the prescaler in Time Counter mode and the width of the
+      glitch filter in Pulse Counter mode.
+      When in prescaler mode, prescaler divides the prescaler clock by
+        2^(prescaler-glitch-filter + 1)
+      When in pulse mode, if prescale-glitch-filter is 0, the glitch filter does not
+        support it. If prescale-glitch-filter is greater than 0, glitch filter
+        recognizes change on input pin after 2^prescale-glitch-filter rising clock
+        edges
```
zejiang0jason: is it better to mention this change in release notes?

Holt-Sun: Added

doc/releases/migration-guide-4.4.rst
```diff
@@ -93,6 +93,46 @@ Controller Area Network (CAN)
...
+     /* Old (deprecated) */
+     lptmr0: counter@40040000 {
+         compatible = "nxp,lpc-lptmr";
+         prescale-glitch-filter = <0>;  /* Implicitly bypassed */
+     };
+
+     /* New (correct) */
+     lptmr0: counter@40040000 {
+         compatible = "nxp,lpc-lptmr";
+         prescale-glitch-filter-bypass; /* Explicitly bypassed */
```
lucien-nxp: I have a question here: if user define the prescale-glitch-filter-bypass and  prescale-glitch-filter two parameter at the same time, should we generate a error log for users about the incorrect configuration? Because I bypassed this prescale-glitch-filter at the SOC level.

Holt-Sun: I don’t think we should add a runtime “error log” for this, because devicetree configuration is resolved at build time and the driver can’t meaningfully “recover” at runtime. It’s also not a functional misconfiguration: when prescale-glitch-filter-bypass is set, the HW bypasses the prescaler/glitch filter, so prescale-glitch-filter is simply unused.

I have updated the nxp,lptmr.yaml to clarify that prescale-glitch-filter is ignored when bypass is set and add a note in migration-guide-4.4.rst stating that if both are set, bypass takes precedence and the value is ignored.

lucien-nxp: Then could we provide a few examples to show how to use prescale-glitch-filter?  It seems that we have bypassed the   prescaler/glitch filte default on all the platforms. These driver upgrades are actually incompatible with the previous user's usage habits. After upgrading the Zephyr version, they may actually have an impact on users.

Holt-Sun: Here are the updates:
1. Kept the BUILD_ASSERT in counter_mcux_lptmr.c so any out-of-tree DTS that uses Pulse mode with prescale-glitch-filter = <0> and no prescale-glitch-filter-bypass will fail the build and must be migrated. It makes sense becase the RM explicitly states: glitch filter does not support this configuration(pulse mode, prescale-glitch-filter = <0>)
2. Ensured bypass default remains false when the boolean property is absent (so prescale-glitch-filter is actually usable by default).
3. Fixed the RST indentation and kept the concrete DTS examples in migration-guide-4.4.rst so they render properly and clearly state that Pulse mode 0 is unsupported and you must use prescale-glitch-filter-bypass;.
4. Updated the binding text to be explicit about Pulse mode 0 and the required bypass property in nxp,lptmr.yaml.

ZhaoxiangJin (dismissed):

doc/releases/migration-guide-4.4.rst
```diff
@@ -93,6 +93,139 @@ Controller Area Network (CAN)
   * :kconfig:option:`CONFIG_CAN_STM32_FDCAN_MAX_STD_ID_FILTERS` for :dtcompatible:`st,stm32-fdcan`
   * :kconfig:option:`CONFIG_CAN_XMC4XXX_MAX_FILTERS` for :dtcompatible:`infineon,xmc4xxx-can-node`
 
+Counter
+=======
+
+* The NXP LPTMR driver (:dtcompatible:`nxp,lpc-lptmr`) has been updated to fix incorrect
```
zejiang0jason: `nxp,lptmr`

doc/releases/migration-guide-4.4.rst
```diff
@@ -93,6 +93,139 @@ Controller Area Network (CAN)
...
+           prescale-glitch-filter = <3>;
+           resolution = <16>;
+       };
+
+  * Time Counter mode: explicit bypass (no division)
+
+    .. code-block:: devicetree
+
+       lptmr0: counter@40040000 {
+           compatible = "nxp,lpc-lptmr";
```
zejiang0jason: `nxp,lptmr`

zejiang0jason (changes requested):

doc/releases/migration-guide-4.4.rst
Holt-Sun: Good catch. Thank you.

zejiang0jason (approved):

ZhaoxiangJin (approved):

PR #101393, opened 2025-12-21
drivers: sensor: add adxl345 range attribute setter

walidbadar:
Add adxl345_attr_set_range() to allow runtime config of the accelerometer's measurement range via the
SENSOR_ATTR_FULL_SCALE attribute.

The function maps the requested range to the nearest supported ADXL345 range (±2g, ±4g, ±8g, or ±16g) and updates the device configuration accordingly.

### DIFF
--- a/drivers/sensor/adi/adxl345/Kconfig
+++ b/drivers/sensor/adi/adxl345/Kconfig
@@ -15,6 +15,9 @@ config ADXL345
 
 if ADXL345
 
+config ADXL345_ACCEL_RANGE_RUNTIME
+	bool "Select range at runtime"
+
 choice ADXL345_TRIGGER_MODE
 	prompt "Trigger mode"
 	default ADXL345_TRIGGER_NONE
--- a/drivers/sensor/adi/adxl345/adxl345.c
+++ b/drivers/sensor/adi/adxl345/adxl345.c
@@ -264,6 +264,63 @@ static int adxl345_attr_set_odr(const struct device *dev,
 	return ret;
 }
 
+static int adxl345_set_range(const struct device *dev, uint8_t range)
+{
+	int rc;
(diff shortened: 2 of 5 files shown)


drivers/sensor/adi/adxl345/Kconfig
```diff
@@ -15,6 +15,27 @@ config ADXL345
...
+	bool "2G"
+
+config ADXL345_ACCEL_RANGE_4G
+	bool "4G"
+
+config ADXL345_ACCEL_RANGE_8G
+	bool "8G"
+
+config ADXL345_ACCEL_RANGE_16G
+	bool "16G"
```
MaureenHelm: These don't appear to be used

MaureenHelm (changes requested):

drivers/sensor/adi/adxl345/Kconfig
walidbadar: I intended to use it for build-time configuration but must have forgotten. Please review the updated commit.

drivers/sensor/adi/adxl345/Kconfig
```diff
@@ -15,6 +15,20 @@ config ADXL345
...
+config ADXL345_ACCEL_RANGE
+	int "ADXL345 accelerometer range"
+	range 0 3
+	default 2
+	help
+	  Select accelerometer range:
+	    0 = ±2g
+	    1 = ±4g
+	    2 = ±8g
+	    3 = ±16g
```
MaureenHelm: This should be a devicetree property so it can be configured differently for each instance.

walidbadar: I have moved those Kconfig to devicetree property.

MaureenHelm (approved):

ttmut (approved):

PR #101396, opened 2025-12-21
drivers: can: mcan: add timestamp counter support

henrikbrixandersen:
- Add support for configuring and enabling the internal timestamp counter of the Bosch M_CAN IP core. Frontend drivers can overwrite this configuration for using a SoC-specific, external timestamp counter.
- Enable support for configuring and enabling the external timestamp counter of the NXP LPC MCAN.

### DIFF
--- a/drivers/can/can_mcan.c
+++ b/drivers/can/can_mcan.c
@@ -1469,6 +1469,23 @@ int can_mcan_init(const struct device *dev)
 		return err;
 	}
 
+#ifdef CONFIG_CAN_RX_TIMESTAMP
+	/*
+	 * Enable the internal timestamp counter by default. SoC-specific driver frontends can
+	 * overwrite this if configured for using a SoC-specific, external timestamp counter.
+	 */
+	reg = FIELD_PREP(CAN_MCAN_TSCC_TCP, config->timestamp_prescaler - 1U) |
+		FIELD_PREP(CAN_MCAN_TSCC_TSS, 1U);
+#else /* CONFIG_CAN_RX_TIMESTAMP */
+	/* Disable timestamp counter */
+	reg = 0U;
+#endif /* !CONFIG_CAN_RX_TIMESTAMP */
+
+	err = can_mcan_write_reg(dev, CAN_MCAN_TSCC, reg);
+	if (err != 0) {
+		return err;
+	}
+
 	err = can_mcan_read_reg(dev, CAN_MCAN_TEST, &reg);
 	if (err != 0) {
 		return err;
--- a/drivers/can/can_mcux_mcan.c
+++ b/drivers/can/can_mcux_mcan.c
@@ -18,8 +18,13 @@ LOG_MODULE_REGISTER(can_mcux_mcan, CONFIG_CAN_LOG_LEVEL);
 #define DT_DRV_COMPAT nxp_lpc_mcan
 
 /* Message RAM Base Address register */
-#define MCUX_MCAN_MRBA	 0x200
+#define MCUX_MCAN_MRBA    0x200
 #define MCUX_MCAN_MRBA_BA GENMASK(31, 16)
+
+/* External timestamp counter configuration register */
+#define MCUX_MCAN_ETSCC      0x400
+#define MCUX_MCAN_ETSCC_ETCP GENMASK(10, 0)
+#define MCUX_MCAN_ETSCC_ETCE BIT(31)
 
 struct mcux_mcan_config {
 	mm_reg_t base;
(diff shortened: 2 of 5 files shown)


dts/bindings/can/nxp,lpc-mcan.yaml
```diff
@@ -16,3 +16,15 @@ properties:
...
+  use-external-timestamp-counter:
+    type: boolean
+    description: |
+      Use external 16-bit timestamp counter clocked by the CPUCLK.
+
+  external-timestamp-counter-prescaler:
+    type: int
+    default: 1
+    description: |
+      Configure the prescaler for the external timestamp counter. Valid range is 1 to 2049. Default
```
zejiang0jason: Should the valid range be 1 to **2048**? because the register field is 11-bit with maximum value 2047.

henrikbrixandersen: Yes, thank you! That was a typo. Fixed!

zejiang0jason (approved):

drivers/can/can_mcan.c
```diff
@@ -1469,6 +1469,23 @@ int can_mcan_init(const struct device *dev)
 		return err;
 	}
 
+#ifdef CONFIG_CAN_RX_TIMESTAMP
+	/*
+	 * Enable the internal timestamp counter by default. SoC-specific driver frontends can
+	 * overwrite this if configured for using a SoC-specific, external timestamp counter.
+	 */
+	reg = FIELD_PREP(CAN_MCAN_TSCC_TCP, config->timestamp_prescaler - 1U) |
+		FIELD_PREP(CAN_MCAN_TSCC_TSS, 1U);
```
str4t0m: You could consider adding defines for the TSCV source selected with TSCC.TSS, as multiple platform drivers will change it to the external counter, but not blocking.

william-tang914 (approved):

drivers/can/can_mcan.c
mathieuchopstm, line 1478: Nit: I guess `1U` here means `INTERNAL_TSC` and `2U` in MCUx driver means `EXTERNAL_TSC` - maybe macros could be provided? (idk what is the common practice for this driver/device - ignore if this is not)

dts/bindings/can/nxp,lpc-mcan.yaml
```diff
@@ -16,3 +16,15 @@ properties:
 
   clocks:
     required: true
+
+  use-external-timestamp-counter:
```
mathieuchopstm, line 20: Nit: maybe `nxp,` since this seems NXP specific.

drivers/can/can_mcux_mcan.c
```diff
@@ -130,6 +139,26 @@ static int mcux_mcan_init(const struct device *dev)
...
+		err = can_mcan_write_reg(dev, MCUX_MCAN_ETSCC, reg);
+		if (err != 0) {
+			return -EIO;
+		}
+
+		/* Use external timestamp counter */
+		err = can_mcan_write_reg(dev, CAN_MCAN_TSCC, FIELD_PREP(CAN_MCAN_TSCC_TSS, 2U));
+		if (err != 0) {
+			return -EIO;
+		}
```
mathieuchopstm, line 158: I assume this is done here because it is invalid to set `TSS=2` if the external TSC is not ready?

If it is allowed, I would set `TSS=2` in common code and move the `use-external-tsc` property to the `bosch,m_can-base` binding since it would be common to all devices (if not supported, one can `property-blocklist` it... which in the immediate term means changing all bindings - your solution is less intrusive so better as-is, but it's something that can be kept in mind)

mathieuchopstm (approved):
LGTM functionally. Some remarks (though may be silly as I don't know much about CAN):

dts/bindings/can/nxp,lpc-mcan.yaml
henrikbrixandersen: It _is_ NXP specific, but so is the binding. There is no need for using a vendor prefix on a property for binding already having the same vendor prefix.

Vendor-prefixed properties are only needed for otherwise vendor-independent bindings.

drivers/can/can_mcan.c
henrikbrixandersen: I don't have a strong opinion on this. When part of an already defined macro (CAN_MCAN_TSCC_TSS), I think setting a verbatim value is okay.

drivers/can/can_mcux_mcan.c
henrikbrixandersen: > I assume this is done here because it is invalid to set `TSS=2` if the external TSC is not ready?

Yes.

> If it is allowed, I would set `TSS=2` in common code and move the `use-external-tsc` property to the `bosch,m_can-base` binding since it would be common to all devices (if not supported, one can `property-blocklist` it... which in the immediate term means changing all bindings - your solution is less intrusive so better as-is, but it's something that can be kept in mind)

The `use-external-tsc` property is specific to the NXP LPC MCAN frontend. Other frontends will have other implementations of external time counters using different properties (see e.g. the proposal in https://github.com/zephyrproject-rtos/zephyr/pull/101374, which uses a phandle to a counter devicetree node for enabling an external TSC).

Until we have more examples of how this vendor-specific integration of an external TSC is handled, I prefer to keep it out of the common code.

PR #101400, opened 2025-12-21
net: ipv4: Drop ipv4 mcast packets for non joined groups

fabiobaltieri:
Add a check to drop ipv4 multicast packets that are not matching any currently registered multicast group for the interface that received them.


subsys/net/ip/ipv4.c
```diff
@@ -362,6 +362,17 @@ enum net_verdict net_ipv4_input(struct net_pkt *pkt)
 		goto drop;
 	}
 
+	if (net_ipv4_is_addr_mcast_raw(hdr->dst)) {
```
jukkar, line 365: Usually there is no need to do this because the network device driver is suppose to filter these out i.e., the driver should monitor `ETHERNET_CONFIG_TYPE_FILTER` config option and program the chip according to this. But if the device does not support `ETHERNET_HW_FILTERING` then something like this is needed.
Could you update also the IPv6 side with similar changes?

jukkar (commented):
Could you split the tests into separate commit?

subsys/net/ip/ipv4.c
fabiobaltieri: Hey Jukka, thanks for the pointer, I was indeed running on an STM32 with `CONFIG_ETH_STM32_MULTICAST_FILTER=n`, did not realize that there was an option for it, wondering what's the point of leaving it off by default. That said I also tried an nxp board using the `nxp,enet-qos` and it had the same behavior, so I've figured the stack should probably be doing it, and I guess since the stm32 check is hash based and the nxp does not have it at all, it still make sense to have.

Anyway, the ipv6 side seems to already to this, that's what I used as reference

https://github.com/zephyrproject-rtos/zephyr/blob/c960d0ed0f9ad1b3f11372e271c8f80b0d1e221e/subsys/net/ip/ipv6.c#L652-L658

Haven't tested it though, should I try it? Was aiming to try and switch the multicast part of this project on ipv6 anyway so I may as well. Will give it a try later tonight.

fabiobaltieri: Ok tested, the ipv6 code is already doing this, works fine.

subsys/net/ip/ipv4.c
```diff
@@ -362,6 +362,17 @@ enum net_verdict net_ipv4_input(struct net_pkt *pkt)
...
+		struct net_if *iface = net_pkt_iface(pkt);
+		struct net_if_mcast_addr *if_mcast_addr = net_if_ipv4_maddr_lookup(
+				(struct net_in_addr *)hdr->dst, &iface);
+		if (!if_mcast_addr ||
+		    !net_if_ipv4_maddr_is_joined(if_mcast_addr)) {
```
pdgendt: Nit
```suggestion
		if (if_mcast_addr == NULL ||
		    !net_if_ipv4_maddr_is_joined(if_mcast_addr)) {
```

fabiobaltieri: fixed

subsys/net/ip/ipv4.c
```diff
@@ -362,6 +362,17 @@ enum net_verdict net_ipv4_input(struct net_pkt *pkt)
 		goto drop;
 	}
 
+	if (net_ipv4_is_addr_mcast_raw(hdr->dst)) {
+		struct net_if *iface = net_pkt_iface(pkt);
+		struct net_if_mcast_addr *if_mcast_addr = net_if_ipv4_maddr_lookup(
+				(struct net_in_addr *)hdr->dst, &iface);
+		if (if_mcast_addr == NULL ||
+		    !net_if_ipv4_maddr_is_joined(if_mcast_addr)) {
```
pdgendt: Just noticed, but `net_if_ipv4_maddr_is_joined` returns false on `NULL`.

```suggestion
		if (!net_if_ipv4_maddr_is_joined(if_mcast_addr)) {
```

fabiobaltieri: uh, nice! applied

pdgendt (approved):

rlubos (approved):

PR #101408, opened 2025-12-22
firmware: scmi: lock all pm states in interrupt flow

yongxu-wang15:
Prevent system from entering low power modes during SCMI message transmission to avoid channel conflicts.

When PM is enabled, the system may enter idle mode while waiting for SCMI reply. If the idle transition itself requires SCMI communication, it causes concurrent channel access and corrupts the channel state.

Lock all PM states before sending SCMI messages and unlock after completion to ensure exclusive channel access during the entire transaction.

This issue only affects interrupt-driven SCMI communication. Polling mode is not affected as it already disables interrupts.

### DIFF
--- a/drivers/firmware/scmi/core.c
+++ b/drivers/firmware/scmi/core.c
@@ -115,6 +115,13 @@ static int scmi_send_message_polling(struct scmi_protocol *proto,
 	int ret;
 	int status;
 
+	/* wait for channel to be free */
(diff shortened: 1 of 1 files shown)


bjarki-andreasen (dismissed):

bjarki-andreasen (approved):

LaurentiuM1234 (approved):
Back from holiday PTO - sorry for the long wait time :(

LGTM, thanks for your contribution!

PR #101411, opened 2025-12-22
Renesas Ethernet driver enhance

duynguyenxa:
This PR add several of enhancement for Renesas Ethernet driver:
+ Fix driver code to adapt with `ethernet-phy-fixed-link` phy
+ Enable Ethernet zerocopy mode and enable caching support for RA8 devices


drivers/ethernet/eth_renesas_ra.c
```diff
@@ -159,8 +175,17 @@ void renesas_ra_eth_callback(ether_callback_args_t *p_args)
 	struct device *dev = (struct device *)p_args->p_context;
 	struct renesas_ra_eth_context *ctx = dev->data;
 
-	if (p_args->event == ETHER_EVENT_RX_COMPLETE) {
+	switch (p_args->event) {
+	case ETHER_EVENT_RX_COMPLETE:
+		__fallthrough;
```
maass-hamburg:
```suggestion
```
This line is not needed, if it is otherwise empty

drivers/ethernet/eth_renesas_ra.c
```diff
@@ -274,39 +299,63 @@ static void renesas_ra_eth_initialize(struct net_if *iface)
...
 	err = R_ETHER_Write(&ctx->ctrl, tx_buf, len);
 	if (err != FSP_SUCCESS) {
-		goto error;
+		LOG_DBG("R_ETHER_Write failed (%d)", err);
+		return -EIO;
 	}
 
-	return 0;
+	/* Wait for TX completion with timeout (e.g., 100 ms) */
+	if (k_sem_take(&ctx->tx_sem, K_MSEC(100)) != 0) {
```
maass-hamburg: We don't like magic numbers deep in the code, this should at least be a define

boards/renesas/mck_ra8t1/mck_ra8t1_defconfig
```diff
@@ -9,3 +9,5 @@ CONFIG_SERIAL=y
 CONFIG_UART_CONSOLE=y
 CONFIG_UART_INTERRUPT_DRIVEN=y
 CONFIG_CONSOLE=y
+
+CONFIG_ARM_MPU=y
```
maass-hamburg: Is this really just board dependent or should it be enabled from the soc?

maass-hamburg (changes requested):

drivers/ethernet/eth_renesas_ra.c
```diff
@@ -274,39 +299,63 @@ static void renesas_ra_eth_initialize(struct net_if *iface)
...
+	/* Start transmission */
 	err = R_ETHER_Write(&ctx->ctrl, tx_buf, len);
 	if (err != FSP_SUCCESS) {
-		goto error;
+		LOG_DBG("R_ETHER_Write failed (%d)", err);
+		return -EIO;
 	}
 
-	return 0;
+	/* Wait for TX completion with timeout */
```
maass-hamburg: if we are waiting for tx to finish, it probably wouldn't make sense to have ``ETH_RENESAS_TX_BUF_NUM`` higher than one, as we could only send one at a time.

Also is `R_ETHER_TxStatusGet` really needed? In which cases would it fail?
If it isn't that important, maybe this don't has to be blocking and use all of the buffers instead.

drivers/ethernet/Kconfig.renesas_ra
```diff
@@ -38,4 +39,11 @@ config ETH_RENESAS_RA_USE_NS_BUF
 	depends on CPU_HAS_TEE
 	depends on !ARM_SECURE_FIRMWARE && !ARM_NONSECURE_FIRMWARE
 
+config ETH_RENESAS_RA_USE_ZERO_COPY
+	bool "Use zero copy feature"
+	default y
```
maass-hamburg: Are there reasons, why someone should deactivate it? If there are none, why not making it mandatory? so not both ways have to be maintained

soc/renesas/ra/ra8t1/Kconfig.defconfig
```diff
@@ -12,6 +12,12 @@ config BUILD_OUTPUT_HEX
 config CLOCK_CONTROL
 	default y
 
+config DCACHE
+	default n
+
+config CACHE_MANAGEMENT
+	default n
```
maass-hamburg: this is already the default

soc/renesas/ra/ra8m1/Kconfig.defconfig
```diff
@@ -12,6 +12,12 @@ config BUILD_OUTPUT_HEX
 config CLOCK_CONTROL
 	default y
 
+config DCACHE
+	default n
+
+config CACHE_MANAGEMENT
+	default n
```
maass-hamburg: same

drivers/ethernet/eth_renesas_ra.c
duynguyenxa: @maass-hamburg It's removed

drivers/ethernet/eth_renesas_ra.c
duynguyenxa: @maass-hamburg , I changed to use macro define

boards/renesas/mck_ra8t1/mck_ra8t1_defconfig
duynguyenxa: all of them are moved into soc layer

soc/renesas/ra/ra8m1/Kconfig.defconfig
```diff
@@ -12,6 +12,12 @@ config BUILD_OUTPUT_HEX
 config CLOCK_CONTROL
 	default y
 
+config DCACHE
+	default n
```
maass-hamburg: why is that disabled by default?

is because of the ethernet driver?
if yes, you should add support for the ethernet driver to work with the dcache by using the cache management functions

drivers/ethernet/eth_renesas_ra.c
duynguyenxa: @maass-hamburg , I removed the R_ETHER_TxStatusGet, semaphore is changed to init base on CONFIG_ETH_RENESAS_TX_BUF_NUM

drivers/ethernet/Kconfig.renesas_ra
duynguyenxa: No, I've removed this config option and changed code to use the zero copy implement

DBY047:
Is this enhancement applicable to the RA8T2 ?

bricle:
> Is this enhancement applicable to the RA8T2 ?

No, actually RA8T2  uses drivers/ethernet/eth_renesas_ra_rmac.c, it's a different driver.

@duynguyenxa Hi, is there any plan to continue this work? Thanks a lot.

duynguyenxa:
> > Is this enhancement applicable to the RA8T2 ?
> 
> No, actually RA8T2 uses drivers/ethernet/eth_renesas_ra_rmac.c, it's a different driver.
> 
> @duynguyenxa Hi, is there any plan to continue this work? Thanks a lot.

Hi @bricle , this commit is trying to support the Ethernet buffer allocation when cache is use and also apply the no copy config in FSP hal to reduce 1 memcpy process when write and read, these are already support in the RMAC driver.

bricle:
> > > Is this enhancement applicable to the RA8T2 ?
> > 
> > 
> > No, actually RA8T2 uses drivers/ethernet/eth_renesas_ra_rmac.c, it's a different driver.
> > @duynguyenxa Hi, is there any plan to continue this work? Thanks a lot.
> 
> Hi @bricle , this commit is trying to support the Ethernet buffer allocation when cache is use and also apply the no copy config in FSP hal to reduce 1 memcpy process when write and read, these are already support in the RMAC driver.

Thanks, good to hear that.

drivers/ethernet/eth_renesas_ra.c
```diff
@@ -32,47 +33,63 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME, LOG_LEVEL);
 #define ETHER_BUF_SIZE         1536
 #define ETHER_PADDING_OFFSET   1
 #define ETHER_BROADCAST_FILTER 0
+#define ETHER_TX_TIMEOUT_MS    100
 #define ETHER_TOTAL_BUF_NUM    (CONFIG_ETH_RENESAS_TX_BUF_NUM + CONFIG_ETH_RENESAS_RX_BUF_NUM)
 #define ETHER_EE_RECEIVE_EVENT_MASK                                                                \
 	(ETHER_EESR_EVENT_MASK_RFOF | ETHER_EESR_EVENT_MASK_RDE | ETHER_EESR_EVENT_MASK_FR |       \
-	 ETHER_EESR_EVENT_MASK_RFCOF)
+	 ETHER_EESR_EVENT_MASK_TFUF | ETHER_EESR_EVENT_MASK_TDE | ETHER_EESR_EVENT_MASK_TC | 0U)
```
maass-hamburg:
```suggestion
	 ETHER_EESR_EVENT_MASK_TFUF | ETHER_EESR_EVENT_MASK_TDE | ETHER_EESR_EVENT_MASK_TC)
```

drivers/ethernet/eth_renesas_ra.c
```diff
@@ -113,17 +130,9 @@ const ether_extended_cfg_t g_ether0_extended_cfg_t = {
...
-ether_phy_instance_ctrl_t g_ether_phy0_ctrl;
-
-const ether_phy_instance_t g_ether_phy0 = {.p_ctrl = &g_ether_phy0_ctrl,
-					   .p_cfg = &g_ether_phy0_cfg,
-					   .p_api = &g_ether_phy_on_ether_phy};
-
 const ether_cfg_t g_ether0_cfg = {
 	.channel = ETHER_CHANNEL0,
-	.zerocopy = ETHER_ZEROCOPY_DISABLE,
+	.zerocopy = ETHER_ZEROCOPY_ENABLE,
```
maass-hamburg, line 133: you still have the Kconfig CONFIG_ETH_RENESAS_RA_USE_ZERO_COPY, can remove that Kconfig and just keep the "zero copy" part. Or is there any reason why somebody should deactivate CONFIG_ETH_RENESAS_RA_USE_ZERO_COPY?

drivers/ethernet/eth_renesas_ra.c
```diff
@@ -32,47 +33,63 @@ LOG_MODULE_REGISTER(LOG_MODULE_NAME, LOG_LEVEL);
...
 	struct k_sem rx_sem;
+	struct k_sem tx_sem;
 	ether_instance_ctrl_t ctrl;
 	/** pinctrl configs */
 	const struct pinctrl_dev_config *pcfg;
 };
 
 struct renesas_ra_eth_config {
 	const ether_cfg_t *p_cfg;
+	const struct pinctrl_dev_config *pincfg;
```
maass-hamburg, line 75: there is also a ``const struct pinctrl_dev_config *pcfg;`` in the data struct. that should then be removed.

maass-hamburg (changes requested):

drivers/ethernet/eth_renesas_ra.c
duynguyenxa: It's removed

drivers/ethernet/eth_renesas_ra.c
duynguyenxa: Sorry, the change for will be done in https://github.com/zephyrproject-rtos/zephyr/pull/109511, I drop the RMAC changes.

soc/renesas/ra/ra8m1/Kconfig.defconfig
duynguyenxa: I removed the default config. Ethernet buffer is put in the non-cache memory

soc/renesas/ra/ra8t1/Kconfig.defconfig
duynguyenxa: Removed

soc/renesas/ra/ra8m1/Kconfig.defconfig
duynguyenxa: Removed

maass-hamburg (approved):

thenguyenyf (approved):

KhiemNguyenT (approved):
LGTM

aescolar:
This PR seems to have broken drivers/ethernet/eth_renesas_ra.c
see 
https://github.com/zephyrproject-rtos/zephyr/actions/runs/30524429835/job/90812123086#step:12:1272

Note the change in ee980e98a0de4bb32f956cf24e84a5cd971506f6 removed pcfg from renesas_ra_eth_context but not setting it.

maass-hamburg:
> This PR seems to have broken drivers/ethernet/eth_renesas_ra.c see https://github.com/zephyrproject-rtos/zephyr/actions/runs/30524429835/job/90812123086#step:12:1272
> 
> Note the change in [ee980e9](https://github.com/zephyrproject-rtos/zephyr/commit/ee980e98a0de4bb32f956cf24e84a5cd971506f6) removed pcfg from renesas_ra_eth_context but not setting it.

this should fix it: https://github.com/zephyrproject-rtos/zephyr/pull/114945

PR #101412, opened 2025-12-22
subsys/mgmt: Declare PROCESSOR_NAME for native_sim

JaagupAverin:
I don't think this value can be elegantly selected in Kconfig like it is for actual boards, so falling back to selecting it by compiler definitions should be required.

This was intentionally put outside of the regular #elif block because it follows a different pattern and this block should be removed if in the future the CONFIG_architecture can properly be selected by the Kconfig/Cmake system somehow.

### DIFF
--- a/subsys/mgmt/mcumgr/grp/os_mgmt/include/os_mgmt_processor.h
+++ b/subsys/mgmt/mcumgr/grp/os_mgmt/include/os_mgmt_processor.h
@@ -184,6 +184,22 @@ extern "C" {
 #define PROCESSOR_NAME "sparc"
 #endif
 
+#if defined(CONFIG_ARCH_POSIX)
+#if defined(PROCESSOR_NAME)
+#error "Processor name already selected; compile-time check should be removed."
+#endif
+
+#if defined(__x86_64__)
+#define PROCESSOR_NAME "x86_64"
+#elif defined(__i386__)
+#define PROCESSOR_NAME "x86"
+#elif defined(__aarch64__)
+#define PROCESSOR_NAME "aarch64"
+#elif defined(__arm__)
+#define PROCESSOR_NAME "arm"
+#endif
+#endif
+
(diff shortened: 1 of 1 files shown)


subsys/mgmt/mcumgr/grp/os_mgmt/include/os_mgmt_processor.h
```diff
@@ -184,6 +184,22 @@ extern "C" {
 #define PROCESSOR_NAME "sparc"
 #endif
 
+#if defined(CONFIG_BOARD_NATIVE_SIM)
```
nordicjm: @aescolar for comment

aescolar: (Background: I know nothing about what you are using this for)
Note that if you want you can just use the arch kconfig option like for the other architectures (`CONFIG_ARCH_POSIX`) so you are not limited to only the native_sim target.

JaagupAverin: Force pushed changing `CONFIG_BOARD_NATIVE_SIM` to `CONFIG_ARCH_POSIX` for more generic handling.
The purpose of this is purely to have the mcumgr's application info command return a valid value, and to stop the compile-time #warning from being generated when building for native_sim.

nordicjm (approved):

subsys/mgmt/mcumgr/grp/os_mgmt/include/os_mgmt_processor.h
JaagupAverin: The core annoyance being the compile-time warning. For all I care the PROCESSOR_NAME can just have the value "native" in such cases but since we have the compiler keywords for detecting the correct arch, might as well use them.

de-nordic (approved):

PR #101418, opened 2025-12-22
soc: nxp: mcxw23: Integrate low power with ble

yeaissa:
Adding support for deep sleep/ power down on BLE apps.
Change exit-latency on power down to match SDK one.

### DIFF
--- a/dts/arm/nxp/nxp_mcxw23x_common.dtsi
+++ b/dts/arm/nxp/nxp_mcxw23x_common.dtsi
@@ -63,7 +63,7 @@
 				compatible = "zephyr,power-state";
 				power-state-name = "standby";
 				min-residency-us = <2000>;
-				exit-latency-us = <1345>;
+				exit-latency-us = <1558>;
 			};
 		};
 	};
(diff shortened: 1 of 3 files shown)


dts/arm/nxp/nxp_mcxw23x_common.dtsi
```diff
@@ -63,7 +63,7 @@
 				compatible = "zephyr,power-state";
 				power-state-name = "standby";
 				min-residency-us = <2000>;
-				exit-latency-us = <1345>;
+				exit-latency-us = <1558>;
```
mmahadevan108, line 66: Can you share why this was changed?

yeaissa: on the SDK the default delay on power down, is set to 918us (`POWER_DOWN_WITH_CPU_RETENTION_EXIT_LATENCY_IN_US`), and there is also a delay of 640us for DCDC restart on some configurations (DCDC bypass) which we are using here on Zephyr power down.

mmahadevan108 (approved):

NeilChen93 (approved):

zejiang0jason (approved):

PR #90681, opened 2025-05-27
doc: _extensions: kconfig: add :kconfig:option-regex: role

kartben:
Allow users to create links to Kconfig regex searches. The new role generates links to the Kconfig search page with the regex pattern as the search query.

Fixes zephyrproject-rtos/zephyr#90571

doc for the new role and demo of it in action: https://builds.zephyrproject.io/zephyr/pr/90681/docs/contribute/documentation/guidelines.html#role-kconfig-option-regex

image

PR #101419, opened 2025-12-22
Replace usage of legacy Mbed TLS crypto Kconfigs with PSA Crypto API

valeriosetti:
Just replace usage of legacy Kconfigs related to Mbed TLS crypto with corresponding PSA Crypto API.


SeppoTakalo (dismissed):

subsys/net/lib/sockets/Kconfig
```diff
@@ -131,12 +131,9 @@ config NET_SOCKETS_SOCKOPT_TLS
 	imply TLS_CREDENTIALS
 	select MBEDTLS if NET_NATIVE
 	imply MBEDTLS_SSL_PROTO_TLS1_2 if !NET_L2_OPENTHREAD
-	imply MBEDTLS_MD_C if !NET_L2_OPENTHREAD
-	imply MBEDTLS_RSA_C if !NET_L2_OPENTHREAD
-	imply MBEDTLS_PKCS1_V15 if !NET_L2_OPENTHREAD
-	imply MBEDTLS_PKCS1_V21 if !NET_L2_OPENTHREAD
+	imply PSA_WANT_ALG_RSA_PKCS1V15_SIGN if !NET_L2_OPENTHREAD
+	imply PSA_WANT_ALG_RSA_PSS if !NET_L2_OPENTHREAD
 	imply MBEDTLS_KEY_EXCHANGE_RSA_ENABLED if !NET_L2_OPENTHREAD
```
ceolin: Is this still available in MBEDTLS 4.0 ?

ceolin (dismissed):

subsys/net/lib/sockets/Kconfig
valeriosetti: Oops, good catch, indeed this [won't be supported](https://github.com/Mbed-TLS/mbedtls/blob/f7173e4389e172589fb9f1d7b74853962023d4fa/ChangeLog#L74) anymore

tomi-font (dismissed):

subsys/net/lib/sockets/Kconfig
valeriosetti: Mmmm replacing `MBEDTLS_KEY_EXCHANGE_RSA_ENABLED` with some other key exchange seems to create a lot of other problems in various samples. If you don't mind I will create a specific PR for this.

tomi-font (dismissed):

ceolin (dismissed):

modules/mbedtls/Kconfig.ciphersuites
tomi-font, line 1: Why use `imply` and not `select`? If you are enabling a ciphersuite, you want all the Kconfig options to be unconditionally enabled, no?

modules/mbedtls/Kconfig.ciphersuites
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2026 BayLibre SAS
+# SPDX-License-Identifier: Apache-2.0
+
+config MBEDTLS_CIPHERSUITE_TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+	bool "Ciphersuite TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
+	imply MBEDTLS_SSL_PROTO_TLS1_2
+	imply MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
+	imply PSA_WANT_ALG_SHA_256
```
tomi-font: Group the algs and key types together? (applies to the rest as well)

modules/mbedtls/Kconfig.ciphersuites
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2026 BayLibre SAS
+# SPDX-License-Identifier: Apache-2.0
+
+config MBEDTLS_CIPHERSUITE_TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+	bool "Ciphersuite TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
+	imply MBEDTLS_SSL_PROTO_TLS1_2
```
tomi-font: Are all the ciphersuites in this file available only with TLS 1.2 (not 1.3)?

modules/mbedtls/Kconfig.ciphersuites
```diff
@@ -0,0 +1,87 @@
...
+	imply PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_EXPORT
+	imply PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_GENERATE
+	imply PSA_WANT_ALG_RSA_PKCS1V15_CRYPT
+	imply PSA_WANT_ALG_RSA_PKCS1V15_SIGN
+	imply PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT
+	imply PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT
+	imply PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE
+	imply PSA_WANT_ECC_SECP_R1_256
+	imply PSA_WANT_ALG_ECDH
+	imply PSA_WANT_ALG_ECDSA
```
tomi-font: Is this one really needed?

modules/mbedtls/Kconfig.ciphersuites
```diff
@@ -0,0 +1,87 @@
...
+	imply MBEDTLS_SSL_PROTO_TLS1_2
+	imply MBEDTLS_KEY_EXCHANGE_PSK_ENABLED
+	imply PSA_WANT_KEY_TYPE_AES
+	imply PSA_WANT_ALG_GCM
+	imply PSA_WANT_ALG_SHA_256
+	imply PSA_WANT_ALG_TLS12_PSK_TO_MS
+	imply PSA_WANT_ALG_TLS12_PRF
+	imply PSA_WANT_KEY_TYPE_DERIVE
+
+config MBEDTLS_CIPHERSUITE_TLS_RSA_WITH_AES_256_CBC_SHA256
```
tomi-font, line 60: Hmm, I see some of these introduced helpers have only one or a few differences between them. How about implementing some (possibly promptless) middle helper that they would commonly select?

doc/releases/migration-guide-4.4.rst
```diff
@@ -633,6 +633,10 @@ Networking
   handle this status in the handler callback to properly reset resource state after successful
   response transmission.
 
+* Automatic selection of crypto Kconfigs has been removed from :kconfig:option:`NET_SOCKETS_SOCKOPT_TLS`
+  as these strongly depends on the final application's needs. As a consequence desired TLS protocol
+  version and ciphersuite(s) must be explicitly selected.
```
tomi-font, line 712: Advertise the new helpers? using https://github.com/zephyrproject-rtos/zephyr/pull/90681

west.yml
```diff
@@ -326,7 +328,8 @@ manifest:
       groups:
         - crypto
     - name: mcuboot
-      revision: 9ac72969f281491d677e669d053281fc2d538ed4
+      remote: mcuboot-upstream
+      revision: 31fcf095a0c22e14c143b28e909c85f9309157ae
```
tomi-font: Is there a PR doing the update in MCUboot?

doc/releases/migration-guide-4.4.rst
```diff
@@ -633,6 +633,10 @@ Networking
   handle this status in the handler callback to properly reset resource state after successful
   response transmission.
 
+* Automatic selection of crypto Kconfigs has been removed from :kconfig:option:`NET_SOCKETS_SOCKOPT_TLS`
+  as these strongly depends on the final application's needs. As a consequence desired TLS protocol
```
tomi-font:
```suggestion
  as they strongly depend on the final application's needs. As a consequence the desired TLS protocol
```

subsys/net/lib/sockets/Kconfig
valeriosetti: Update that I forgot to mention before: I ended up removing usage of that key exchange since there were failures in the CI related to that anyway.

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: Yes, you are right, I've been to conservative this time :)

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: Yes. I know [this is not](https://github.com/zephyrproject-rtos/mbedtls/blob/be1dd5b11d8f59af0ab0995985e348a429999013/library/ssl_ciphersuites.c#L280) a very user-friendly kind of documentation, but TLS 1.3 has other ciphersuites.

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: [Nope](https://github.com/zephyrproject-rtos/mbedtls/blob/be1dd5b11d8f59af0ab0995985e348a429999013/include/mbedtls/mbedtls_config.h#L1011), good catch!

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: I'll leave `imply` only where it really makes sense to change that for the end user. For example `secp256r1` curve for ECDH/ECDSA is only a possible choice.

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: IDK, I would prefer not to create some other intermediate Kconfig that wraps too many things (like what's currently happening with key exchanges Kconfigs).
Moreover if you look at the ciphersuite structure that I linked above, you'll see that there are really a lot of of options available. I just added the few ones that I needed to make the CI happy, but many other can be added and I don´t know if these hidden Kconfigs would be helpful then.

west.yml
valeriosetti: IIRC the last time I faced a similar situation I was told that Zephyr's fork of McuBoot is periodically synced with the upstream version and that no PR should be opened on Zephyr's fork. @de-nordic Am I right? If this is the case, can we bump the version now so that I can include the commit I need for this PR?

de-nordic: > IIRC the last time I faced a similar situation I was told that Zephyr's fork of McuBoot is periodically synced with the upstream version and that no PR should be opened on Zephyr's fork. @de-nordic Am I right?

You are right.

modules/mbedtls/Kconfig.ciphersuites
tomi-font: Perfect!

modules/mbedtls/Kconfig.ciphersuites
tomi-font: Would it then make sense to have `TLS1_2` in the name of these Kconfig options (instead of just `TLS`?

modules/mbedtls/Kconfig.ciphersuites
```diff
@@ -0,0 +1,86 @@
...
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE
+	imply PSA_WANT_ECC_SECP_R1_256
+
+config MBEDTLS_CIPHERSUITE_TLS_PSK_WITH_AES_256_CBC_SHA384
+	bool "Ciphersuite TLS_PSK_WITH_AES_256_CBC_SHA384"
+	select MBEDTLS_SSL_PROTO_TLS1_2
+	select MBEDTLS_KEY_EXCHANGE_PSK_ENABLED
+	select PSA_WANT_ALG_CBC_NO_PADDING
```
tomi-font: Put this line a bit below for consistency to have the `ALG_TLS12` ones first?

doc/releases/migration-guide-4.4.rst
```diff
@@ -707,6 +707,12 @@ Networking
 * The protocol version passed to :c:func:`zsock_socket` when creating a secure socket is now
   enforced as the minimum TLS version to use for the TLS session.
 
+* Automatic selection of crypto Kconfigs has been removed from :kconfig:option:`NET_SOCKETS_SOCKOPT_TLS`
+  as they strongly depend on the final application's needs. As a consequence the desired TLS protocol
+  version and ciphersuite(s) must be explicitly selected.
+  :kconfig:option-regex:`MBEDTLS_CIPHERSUITE_TLS_.*` gives a list of the currently available
```
tomi-font:
```suggestion
  :kconfig:option-regex:`CONFIG_MBEDTLS_CIPHERSUITE_TLS_.*` gives a list of the currently available
```

doc/releases/migration-guide-4.4.rst
```diff
@@ -707,6 +707,12 @@ Networking
...
+  ciphersuites. More can be added following the same pattern.
```
tomi-font: This to me can be interpreted like these ciphersuites are the only ones supported at the moment, which isn't quite true (they are supported in Mbed TLS).
Maybe rather something in the lines of
```suggestion
  Available :kconfig:option-regex:`CONFIG_MBEDTLS_CIPHERSUITE_TLS_.*` Kconfig helpers can be used to automatically enable all the dependencies of a given ciphersuite, and more can be added as needed following the same pattern.
```
?

doc/releases/migration-guide-4.4.rst
valeriosetti: Done as per the following suggestion

modules/mbedtls/Kconfig.ciphersuites
valeriosetti: Well, even though [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446#appendix-B.4) uses the same names for TLS 1.2 and TLS 1.3, Mbed TLS only differentiate with "TLS" vs "TLS1_3".
The names of the Kconfigs that I used here are literally a copy of those [defined in Mbed TLS](https://github.com/zephyrproject-rtos/mbedtls/blob/be1dd5b11d8f59af0ab0995985e348a429999013/include/mbedtls/ssl_ciphersuites.h#L27). For example:
`CONFIG_MBEDTLS_CIPHERSUITE_TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`
is made of:
- `CONFIG_MBEDTLS_CIPHERSUITE_` for a bit of namespacing
- `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` which matches exactly with the `MBEDTLS_TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` defined in Mbed TLS.

Following this pattern if we were to add a TLS 1.3 ciphersuite, it would be:
`CONFIG_MBEDTLS_CIPHERSUITE_TLS1_3_AES_128_GCM_SHA256`

west.yml
valeriosetti: @nordicjm in order to include the [McuBoot's PR](https://github.com/zephyrproject-rtos/mcuboot/commit/31fcf095a0c22e14c143b28e909c85f9309157ae) necessary for this PR, I updated the manifest to point to the HEAD of `upstream-sync` branch (`main` branch doesn't include that fix). Unfortunately this made the [CI failing](https://github.com/zephyrproject-rtos/zephyr/actions/runs/21292842919/job/61295504895?pr=101419) on some ESP32 boards due to McuBoot issues with macro `MCUBOOT_WATCHDOG_SETUP`.
Do you know if there is already a fix for this problem?

valeriosetti: Update: I just saw a message from @de-nordic mentioning that you are already [working](https://github.com/zephyrproject-rtos/zephyr/pull/102517) on this issue. I subscribed to your PR so when it's resolved I will run the CI here again

nordicjm: I pulled in their changes to fix it so it should pass now

modules/mbedtls/Kconfig.ciphersuites
tomi-font: Okay, makes sense to me as is, thanks for the explanation!

west.yml
```diff
@@ -326,7 +326,7 @@ manifest:
       groups:
         - crypto
     - name: mcuboot
-      revision: 9ac72969f281491d677e669d053281fc2d538ed4
+      revision: cb6e4ceaacd4dfe1e3cdeb92c7f5680f62967837
```
tomi-font, line 329: FYI MCUboot is the correct spelling

tomi-font (dismissed):

rlubos (dismissed):

rlubos (approved):

ceolin (approved):

PR #101421, opened 2025-12-22
led: check for empty led_dt_spec on _dt calls

fabiobaltieri:
Add a check for NULL dev on _dt calls, this allows using LED_DT_SPEC_GET_OR(..., {0}) to initialize led structure and not having to add check to each calls if the code runs on multiple boards where the LED may just not be there.


henrikbrixandersen (changes requested):
I don't think we should add special error checking to the LED DT APIs.

fabiobaltieri:
Alright let's not do it.

PR #101426, opened 2025-12-22
drivers: wifi: siwx91x: Fix multicast filtering on IGMP leave

Devika-Raju-Z:
Device continued receiving multicast data after leaving IGMP group because the driver did not configure the SiWx917 firmware to stop filtering for the multicast MAC address.

Registered a Zephyr multicast monitor callback (net_if_mcast_mon_register) that calls sl_wifi_configure_multicast_filter() to add or remove multicast MAC filters in the firmware when groups are joined or left.

### DIFF
--- a/drivers/wifi/siwx91x/siwx91x_wifi.c
+++ b/drivers/wifi/siwx91x/siwx91x_wifi.c
@@ -301,6 +301,56 @@ unref:
 	return SL_STATUS_FAIL;
 }
 
+static enum ethernet_hw_caps siwx91x_get_capabilities(const struct device *dev)
+{
+	ARG_UNUSED(dev);
+
+	return ETHERNET_HW_FILTERING;
+}
+
+static int siwx91x_set_config(const struct device *dev,
+			      enum ethernet_config_type type,
+			      const struct ethernet_config *config)
+{
+	sl_wifi_multicast_filter_info_t filter_info = {};
+	sl_status_t status;
+
+	ARG_UNUSED(dev);
+
+	switch (type) {
+	case ETHERNET_CONFIG_TYPE_FILTER:
+		memcpy(filter_info.mac_address.octet, config->filter.mac_address.addr,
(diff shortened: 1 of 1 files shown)


jhedberg (dismissed):

drivers/wifi/siwx91x/siwx91x_wifi.c
```diff
@@ -6,6 +6,7 @@
 #define DT_DRV_COMPAT silabs_siwx91x_wifi
 
 #include <zephyr/version.h>
+#include <zephyr/net/ethernet.h>
```
jukkar: The commits could be squashed together, the 2nd commit does not need to be separate.

drivers/wifi/siwx91x/siwx91x_wifi.c
```diff
@@ -26,6 +26,10 @@ LOG_MODULE_REGISTER(siwx91x_wifi);
 
 NET_BUF_POOL_FIXED_DEFINE(siwx91x_tx_pool, 1, _NET_ETH_MAX_FRAME_SIZE, 0, NULL);
 
+#if defined(CONFIG_NET_NATIVE_IPV4) || defined(CONFIG_NET_NATIVE_IPV6)
+static struct net_if_mcast_monitor mcast_monitor;
```
jukkar: This is certainly possible but as the `subsys/net/l2/ethernet.c` already does all this, we could save some memory and register a `set_config` callback in Ethernet API and catch `ETHERNET_CONFIG_TYPE_FILTER` event. See how Ethernet drivers in `drivers/ethernet` are doing this (also Airoc WiFi driver does this in the recommended way).

jerome-pouiller (commented):
For some reason, the commit log is duplicated. Can you fix that?

Otherwise, code looks good to me.

jerome-pouiller (approved):

jukkar (approved):

jhedberg (approved):

PR #101427, opened 2025-12-22
drivers: ethernet: nxp: Increase rx frame length for dsa tail tagging

flowctl:
Description:
Fix frame reception errors when DSA tail tagging is enabled on NXP ethernet driver.

Problem:
When the KSZ8836 DSA switch IC is used, it appends a tail tag byte to every
received frame. Maximum-sized Ethernet frames (1518 bytes) become 1519 bytes
after tagging, which exceeds the NXP driver's configured maximum frame size.
This causes `kStatus_ENET_RxFrameError` and frames are dropped.

Root Cause:
- Standard Ethernet max frame: 1518 bytes
- KSZ8863 DSA tail tag: +1 byte
- NXP driver max frame size: 1518 bytes (`ENET_FRAME_MAX_FRAMELEN`)
- Result: 1519 bytes > 1518 bytes configured max → frame dropped

Solution:
Increase `ENET_FRAME_MAX_FRAMELEN` to accommodate DSA tail tagging overhead,
allowing frames with DSA tail tag based on `CONFIG_DSA_TAG_SIZE`.

### DIFF
--- a/drivers/ethernet/eth_nxp_enet.c
+++ b/drivers/ethernet/eth_nxp_enet.c
@@ -727,6 +727,15 @@ static int eth_nxp_enet_init(const struct device *dev)
 
 	ENET_GetDefaultConfig(&enet_config);
 
+	/* Adjust rxMaxFrameLen to account for DSA tail tag if enabled.
+	 * The FSL SDK defaults to ENET_FRAME_MAX_FRAMELEN (1518), but when DSA
+	 * tail tagging is active, the switch adds an extra tag that must fit within
+	 * the frame length.
+	 */
+#if defined(CONFIG_DSA_TAG_SIZE) && (CONFIG_DSA_TAG_SIZE > 0)
+	enet_config.rxMaxFrameLen = ENET_FRAME_MAX_FRAMELEN + CONFIG_DSA_TAG_SIZE;
+#endif
+
 	if (IS_ENABLED(CONFIG_NET_PROMISCUOUS_MODE)) {
 		enet_config.macSpecialConfig |= kENET_ControlPromiscuousEnable;
 	}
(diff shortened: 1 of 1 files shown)


pdgendt:
@flowctl Please don't use fixup commits, squash into a single commit and force push.

drivers/ethernet/eth_nxp_enet.c
```diff
@@ -727,6 +727,15 @@ static int eth_nxp_enet_init(const struct device *dev)
 
 	ENET_GetDefaultConfig(&enet_config);
 
+	/* Adjust rxMaxFrameLen to account for DSA tail tag if enabled.
+	 * The FSL SDK defaults to ENET_FRAME_MAX_FRAMELEN (1518), but when DSA
+	 * tail tagging is active, the switch adds an extra tag that must fit within
+	 * the frame length.
+	 */
+#if defined(CONFIG_DSA_TAG_SIZE) && (CONFIG_DSA_TAG_SIZE > 0)
+	enet_config.rxMaxFrameLen = ENET_FRAME_MAX_FRAMELEN + CONFIG_DSA_TAG_SIZE;
```
stanislav-poboril, line 736: If `CONFIG_NET_VLAN` is enabled, `kENET_ControlVLANTagEnable` flag is set and the MCUX SDK ENET driver ignores `rxMaxFrameLen`:
https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/drivers/enet/fsl_enet.c#L562
https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/drivers/enet/fsl_enet.c#L569

So we must either change this behavior in MCUX SDK driver and NXP HAL or prohibit usage of both VLAN and DSA tags at the same time (is it a common scenario?).

drivers/ethernet/eth_nxp_enet.c
```diff
@@ -875,20 +884,28 @@ static const struct ethernet_api api_funcs = {
...
+#define _ETH_NXP_ENET_DSA_TAG_LEN 0
+#endif /* CONFIG_DSA_TAG_SIZE */
+
 #if defined(CONFIG_NET_VLAN)
-#if !defined(ENET_FRAME_MAX_VLANFRAMELEN)
-#define ENET_FRAME_MAX_VLANFRAMELEN (ENET_FRAME_MAX_FRAMELEN + 4)
-#endif
-#define ETH_NXP_ENET_BUFFER_SIZE \
-		ROUND_UP(ENET_FRAME_MAX_VLANFRAMELEN, ENET_BUFF_ALIGNMENT)
+#define _ETH_NXP_ENET_VLAN_LEN 4
```
stanislav-poboril, line 900: According to MCUX SDK ENET driver, if ENET instance has AVB feature, VLAN tag size should be added twice:

https://github.com/zephyrproject-rtos/hal_nxp/blob/master/mcux/mcux-sdk-ng/drivers/enet/fsl_enet.c#L569

PR #101428, opened 2025-12-22
Xtensa ace bug

### LINKED ISSUE #101339: xtensa: ace 3.0: Unchecked error leads to exception
### Describe the bug

Allocation at https://github.com/zephyrproject-rtos/zephyr/blob/c3190d34aa912d39cd082db822f6ebb332806a2b/arch/xtensa/core/ptables.c#L1081 can fail, which then leads to an exception. The result should be checked and handled instead, at least with an assertion.

Further, if a new domain with a new L1 page table has been allocated with no mappings, then memory gets mapped so that that new domain is also updated, this line https://github.com/zephyrproject-rtos/zephyr/blob/c3190d34aa912d39cd082db822f6ebb332806a2b/arch/xtensa/core/ptables.c#L1072 will return `NULL` which then also will lead to an exception. When that is fixed by
```
@@ -1136,6 +1136,8 @@ static void dup_l2_table_if_needed(uint32_t *l1_table, uint32_t l1_pos)
 	uint32_t *l2_table, *src_l2_table;
 
 	src_l2_table = (uint32_t *)PTE_PPN_GET(l1_table[l1_pos]);
+	if (!src_l2_table)
+		return;
 
 	if (l2_page_tables_counter[l2_table_to_counter_pos(src_l2_table)] == 1) {
 		/* Only one user of L2 table, no need to duplicate. */
```
another exception occurs later when unmapping that memory

### Regression

- [ ] This is a regression.

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Impact

### Environment

_No response_

### Additional Context

_No response_

Majistor:
Ensure `dup_l2_table_if_needed()` safely handles cases where the
source L2 page table pointer is NULL and where `dup_l2_table()`
may fail due to allocation failure. Previously, dereferencing
`src_l2_table` without a NULL check could lead to a fault, and no
check was present for the return value of `dup_l2_table()`. This
change avoids dereferencing NULL and returns early when duplication
fails, with a kernel assert in debug builds to catch failures during
development/testing.

Fixes: #101339

### DIFF
--- a/arch/xtensa/core/ptables.c
+++ b/arch/xtensa/core/ptables.c
@@ -1068,7 +1068,9 @@ static void dup_l2_table_if_needed(uint32_t *l1_table, uint32_t l1_pos, enum dup
 	k_spinlock_key_t key;
 
 	src_l2_table = (uint32_t *)PTE_PPN_GET(l1_table[l1_pos]);
-
+	if (!src_l2_table) {
+		return;
+	}
 	key = k_spin_lock(&xtensa_counter_lock);
 	if (l2_page_tables_counter[l2_table_to_counter_pos(src_l2_table)] == 1) {
 		/* Only one user of L2 table, no need to duplicate. */
@@ -1077,6 +1079,7 @@ static void dup_l2_table_if_needed(uint32_t *l1_table, uint32_t l1_pos, enum dup
 	}
 
 	l2_table = dup_l2_table(src_l2_table, action);
+	__ASSERT(l2_table, "dup_l2_table() failed");
 
 	/* The page table is using kernel ASID because we don't
 	 * user thread manipulate it.


arch/xtensa/core/ptables.c
```diff
@@ -1068,7 +1068,9 @@ static void dup_l2_table_if_needed(uint32_t *l1_table, uint32_t l1_pos, enum dup
 	k_spinlock_key_t key;
 
 	src_l2_table = (uint32_t *)PTE_PPN_GET(l1_table[l1_pos]);
-
+	if (!src_l2_table) {
+		return;
+	}
```
dcpleung, line 1073: This is not correct as it is possible to map page with physical address 0x0 to a virtual address.

dcpleung (changes requested):

arch/xtensa/core/ptables.c
Majistor: Would you like me to use the assert?

dcpleung: The check for NULL here is entirely wrong. It is not a question of whether to use return or assert.

Majistor: I should remove this check and keep the other one, right?

dcpleung: That's not enough for the issue. See https://github.com/zephyrproject-rtos/zephyr/pull/101441

Majistor: Got it, you will be fixing that issue, right?
But you could have guided me instead.

PR #101435, opened 2025-12-22
boards: silabs: siwx91x: Add ADC input dt-bindings and add io-channel in boards dtsi

Martinhoff-maker:
The goal of this PR is to have a better interface to configure an ADC channel on SiWx91x SoC with ADC dt-bindings.
Moreover, It enables samples for all SiWx91x boards by defining an ADC io-channel in board dtsi.


include/zephyr/dt-bindings/adc/silabs-siwx91x-adc.h
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2025 Silicon Laboratories Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
```
JarmouniA, line 7: needs some doc.

Martinhoff-maker: I’ve added some documentation, but I’m not sure if this is what you meant by “doc.” I can also add a reference to the SoC reference manual with the register descriptions if you think that’s useful, but I don’t personally think it is.

JarmouniA: > I’ve added some documentation, but I’m not sure if this is what you meant by “doc.”

It is.

JarmouniA (approved):

jhedberg (approved):

PR #101436, opened 2025-12-22
drivers: timer: Fix radio timer issue on STM32WB07/STM32WB05

HoZHel:
- Change the initialization priority in `entropy_stm32.c` when there is a `IRQLESS_TRNG` SoC.  According to the documentation, kernel services should not be used during device configuration with PRE_KERNEL_1.
In "entropy_stm32_rng_init", there is "start_pool_filling" which calls
"k_work_schedule". Moreover, "k_work_schedule" requires system timer which
is initialized within PRE_KERNEL_2. Therefore, the SoC gets stuck.

- Due to the hardware limitation, the time between the CPU wake-up IRQ fire
(IRQ 23) and the next call of LL_RADIO_TIMER_SetCPUWakeupTime should not
be less than 16 MTU (Machine Time Unit), i.e. approximately 30us.
otherwise, the next CPU wake-up doesn't happen unless the timer wraps.

- Lock IRQs while sys_clock_set_timeout is being executed.

Supplementary PR[#337](https://github.com/zephyrproject-rtos/hal_stm32/pull/337)

### DIFF
--- a/drivers/entropy/entropy_stm32.h
+++ b/drivers/entropy/entropy_stm32.h
@@ -18,10 +18,18 @@
 #if IRQLESS_TRNG
 #define DT_DRV_COMPAT st_stm32_rng_noirq
 #define TRNG_GENERATION_DELAY	K_NSEC(DT_INST_PROP_OR(0, generation_delay_ns, 0))
+
+/* In IRQ-less hardware, the system workqueue is used to poll the TRNG
+ * during driver initialization. Since the workqueue service is not available
+ * in PRE_KERNEL_1/PRE_KERNEL_2, the init routine must run at POST_KERNEL.
+ */
+#define STM32_TRNG_INIT_LEVEL POST_KERNEL
+
 #else /* !IRQLESS_TRNG */
 #define DT_DRV_COMPAT st_stm32_rng
 #define IRQN		DT_INST_IRQN(0)
 #define IRQ_PRIO	DT_INST_IRQ(0, priority)
+#define STM32_TRNG_INIT_LEVEL PRE_KERNEL_1
 #endif /* IRQLESS_TRNG */
 
 /* Cross-series LL compatibility wrappers */
--- a/drivers/timer/stm32wb0_radio_timer.c
+++ b/drivers/timer/stm32wb0_radio_timer.c
@@ -45,6 +45,7 @@ static const uint32_t calibration_data_freq1 = 0x0028F5C2;
 uint32_t blue_unit_conversion(uint32_t time, uint32_t period_freq, uint32_t thr);
 
 static uint64_t announced_cycles;
+static uint32_t last_cpu_wakeup_time;
 
 static void radio_timer_error_isr(void *args)
 {
@@ -69,7 +70,16 @@ static void radio_timer_cpu_wkup_isr(void *args)
 
 	ARG_UNUSED(args);
 
-	HAL_RADIO_TIMER_TimeoutCallback();
+	last_cpu_wakeup_time = LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP);
+
+	/* Clear the interrupt */
+	LL_RADIO_TIMER_ClearFlag_CPUWakeup(WAKEUP);
+
+	/* Read back the register to ensure that the previous
+	 * write operation is completed.
+	 */
+	LL_RADIO_TIMER_IsActiveFlag_CPUWakeup(WAKEUP);
+
 	if (IS_ENABLED(CONFIG_TICKLESS_KERNEL)) {
 		diff_cycles = HAL_RADIO_TIMER_GetCurrentSysTime() - announced_cycles;
 		dticks = (int32_t)k_cyc_to_ticks_near64(diff_cycles);
@@ -85,7 +95,17 @@ static void radio_timer_txrx_wkup_isr(void *args)
 {
 	ARG_UNUSED(args);
 
-	HAL_RADIO_TIMER_WakeUpCallback();
+	/* The callback body will be properly implemented in the future
+	 * while providing PM support for WB06/WB07.
+	 */
+
+	/* Clear the interrupt */
+	LL_RADIO_TIMER_ClearFlag_BLEWakeup(WAKEUP);
+
+	/* Read back the register to ensure that the previous
+	 * write operation is completed.
+	 */
+	LL_RADIO_TIMER_IsActiveFlag_BLEWakeup(WAKEUP);
 }
 #endif /* CONFIG_SOC_STM32WB06XX || CONFIG_SOC_STM32WB07XX */
 
@@ -99,6 +119,7 @@ void sys_clock_set_timeout(int32_t ticks, bool idle)
 
 	if (IS_ENABLED(CONFIG_TICKLESS_KERNEL)) {
 		uint32_t current_time, delay;
+		uint32_t key;
 
 		ticks = MAX(1, ticks);
 		delay = blue_unit_conversion(k_ticks_to_cyc_near32(ticks), calibration_data_freq1,
@@ -108,9 +129,20 @@ void sys_clock_set_timeout(int32_t ticks, bool idle)
 		} else {
 			delay = MAX(MIN_ALLOWED_DELAY, delay);
 		}
-		current_time = LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP);
+		key = irq_lock();
+
+		/* Due to a hardware limitation, the radio timer wake-up time
+		 * must not be updated until at least 16 Machine Time Units
+		 * have elapsed since the interrupt was triggered; otherwise,
+		 * the next wake-up will only occur after the timer wraps around.
+		 */
+		do {
+			current_time = LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP);
+		} while ((current_time - last_cpu_wakeup_time) < 16U);
+
 		LL_RADIO_TIMER_SetCPUWakeupTime(WAKEUP, current_time + delay + TIMER_ROUNDING);
 		LL_RADIO_TIMER_EnableCPUWakeupTimer(WAKEUP);
+		irq_unlock(key);
 	}
 }
 
--- a/drivers/timer/stm32wb0_radio_timer.c
+++ b/drivers/timer/stm32wb0_radio_timer.c
@@ -32,7 +32,7 @@ LOG_MODULE_REGISTER(radio_timer_driver);
 BUILD_ASSERT(DT_NODE_HAS_STATUS(DT_NODELABEL(clk_lsi), disabled),
 	     "LSI is not supported yet");
 
-#if (defined(CONFIG_SOC_STM32WB06XX) || defined(CONFIG_SOC_STM32WB06XX)) && defined(CONFIG_PM)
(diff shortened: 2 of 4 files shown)


msmttchr (approved):

drivers/entropy/entropy_stm32.h
```diff
@@ -18,10 +18,12 @@
 #if IRQLESS_TRNG
 #define DT_DRV_COMPAT st_stm32_rng_noirq
 #define TRNG_GENERATION_DELAY	K_NSEC(DT_INST_PROP_OR(0, generation_delay_ns, 0))
+#define KERNEL_PRIORITY POST_KERNEL
```
mathieuchopstm: `INIT_PRIORITY` or `RNG_INIT_PRIORITY` would be a more understandable name

A comment explaining why this is required would also be nice. Proposal:
```c
/* 
 * On IRQ-less hardware, the system workqueue is
 * used to poll the TRNG. The driver init routine
 * must run at POST_KERNEL because the workqueue
 * service is not available in PRE_KERNEL.
 */
```

drivers/timer/stm32wb0_radio_timer.c
```diff
@@ -85,7 +91,9 @@ static void radio_timer_txrx_wkup_isr(void *args)
 {
 	ARG_UNUSED(args);
 
-	HAL_RADIO_TIMER_WakeUpCallback();
+	/* It will be properly implemented in the future. */
```
mathieuchopstm: Could you elaborate on what needs to be implemented? (i.e., what is `It`)

drivers/entropy/entropy_stm32.h
etienne-lms: Could you rename the macro: `LEVEL` instead of `PRIORITY`,? And maybe a `STM32_` prefix, for consistency.
E.g. `STM32_TRNG_INIT_LEVEL`.

drivers/timer/stm32wb0_radio_timer.c
```diff
@@ -108,9 +117,12 @@ void sys_clock_set_timeout(int32_t ticks, bool idle)
 		} else {
 			delay = MAX(MIN_ALLOWED_DELAY, delay);
 		}
+		key = irq_lock();
+		while (((uint32_t)(LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP) - last_cpu_wakeup_time)) < 16U);
 		current_time = LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP);
 		LL_RADIO_TIMER_SetCPUWakeupTime(WAKEUP, current_time + delay + TIMER_ROUNDING);
 		LL_RADIO_TIMER_EnableCPUWakeupTimer(WAKEUP);
+		irq_unlock(key);
```
mathieuchopstm, line 145: Nit: re-ordering the loop could avoid an additional read of the timer
```suggestion
		key = irq_lock();
		do {
		    current_time = LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP);
		} while ((uint32_t)current_time - last_cpu_wakeup_time) < 16U);
		LL_RADIO_TIMER_SetCPUWakeupTime(WAKEUP, current_time + delay + TIMER_ROUNDING);
		LL_RADIO_TIMER_EnableCPUWakeupTimer(WAKEUP);
		irq_unlock(key);
```
(but maybe the additional read is intended/necessary)

drivers/timer/stm32wb0_radio_timer.c
```diff
@@ -108,9 +117,12 @@ void sys_clock_set_timeout(int32_t ticks, bool idle)
 		} else {
 			delay = MAX(MIN_ALLOWED_DELAY, delay);
 		}
+		key = irq_lock();
+		while (((uint32_t)(LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP) - last_cpu_wakeup_time)) < 16U);
```
mathieuchopstm: Add a comment mirroring your explanation. Feel free to add a reference to Errata Sheet/RefMan if there is one:
```suggestion
                /*
                 * Due to a hardware limitation, the radio timer wake-up time
                 * must not be updated until at least 16 Machine Time Units
                 * have elapsed since the interrupt was triggered; otherwise,
                 * the next wake-up will only occur after timer wraparound.
                 */
		while (((uint32_t)(LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP) - last_cpu_wakeup_time)) < 16U);
```

drivers/timer/stm32wb0_radio_timer.c
```diff
@@ -45,6 +45,7 @@ static const uint32_t calibration_data_freq1 = 0x0028F5C2;
 uint32_t blue_unit_conversion(uint32_t time, uint32_t period_freq, uint32_t thr);
 
 static uint64_t announced_cycles;
+static uint32_t last_cpu_wakeup_time;
```
etienne-lms, line 48: According to the commit message header line, the issue is related to WB07 only. Does it apply also to other WB0x series?

asm5878 (dismissed):

drivers/timer/stm32wb0_radio_timer.c
etienne-lms: For coding style (+ cast to `uint32_t` is not needed):
```suggestion
		while ((LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP) - last_cpu_wakeup_time) < 16U) {
		}
```
I think an inline comment would be welcome here, e.g.:
```patch
+		/* Hardware requires at least 16 Machine Time Units between calls to
+		 * LL_RADIO_TIMER_SetCPUWakeupTime().
+		 */
 		while ((LL_RADIO_TIMER_GetAbsoluteTime(WAKEUP) - last_cpu_wakeup_time) < 16U) {
 		}
```

drivers/timer/stm32wb0_radio_timer.c
HoZHel: "It" refers to the body of the ISR. When I provide PM support for WB06/WB07, it'll be updated.

drivers/entropy/entropy_stm32.h
HoZHel: Ok, I'll change the name. Since this is a STM32 header file, is it really necessary to add the prefix?

etienne-lms: I think it would be less ambiguous, but it's not a string opinion. That said, better not define  a private macro which name could conflict with a generic macro. Using subsystem and vendor prefixes helps.

HoZHel: Done

drivers/timer/stm32wb0_radio_timer.c
HoZHel: Done

drivers/timer/stm32wb0_radio_timer.c
HoZHel: Done

drivers/entropy/entropy_stm32.h
HoZHel: Done

drivers/timer/stm32wb0_radio_timer.c
HoZHel: It's also related to WB05. I updated the commit message.

drivers/timer/stm32wb0_radio_timer.c
HoZHel: Done

drivers/timer/stm32wb0_radio_timer.c
```diff
@@ -85,7 +91,9 @@ static void radio_timer_txrx_wkup_isr(void *args)
 {
 	ARG_UNUSED(args);
 
-	HAL_RADIO_TIMER_WakeUpCallback();
+	/* It will be properly implemented in the future. */
+	LL_RADIO_TIMER_ClearFlag_BLEWakeup(WAKEUP);
+	LL_RADIO_TIMER_IsActiveFlag_BLEWakeup(WAKEUP);
```
etienne-lms, line 108: `LL_RADIO_TIMER_IsActiveFlag_BLEWakeup()` only reads WAKEUP_BLE_IR_STATUS[0] and returns whether it is set. I think this call should be removed.
Also applies to `radio_timer_cpu_wkup_isr()`.

drivers/timer/stm32wb0_radio_timer.c
etienne-lms:
```suggestion
	/* PM is not yet supported. Only clear the interrupt,
	 * it will be properly implemented in the future.
	 */
```
Also applies to the inline comment added in `radio_timer_cpu_wkup_isr()`.

drivers/timer/stm32wb0_radio_timer.c
etienne-lms: Is is needed for WB6 and WB9? If not, maybe condition the changes (mainly the wait loop in `sys_clock_set_timeout()`) to
`defined(CONFIG_SOC_STM32WB05XX) && defined(CONFIg_SOC_STM32WB07XX)`.
Otherwise, change the commit message:
```patch
-drivers: timer: Fix radio timer issue on STM32WB07/STM32WB05
+drivers: timer: Fix radio timer issue on STM32WB0x
```

HoZHel: There is no board based on WB06 SoC in Zephyr, so no need to mention it. However, the functionality is similar to WB07.
For WB09, the changes are compatible and don't harm.
Your suggested commit message implies that the issue is related to all the WB0 SoCs, which isn't true.

drivers/timer/stm32wb0_radio_timer.c
HoZHel: Why do you think it applies to `radio_timer_cpu_wkup_isr`?

drivers/timer/stm32wb0_radio_timer.c
HoZHel: From a software point of view, it may seem redundant, but it makes sense from a hardware perspective. This extra read is linked to an HW issue when the system clock is set to 64MHz. The peripheral clock and the system clock are different, the peripheral clock can be set to 16MHz or 32MHz. When a write is executed with this setting, it may happen that the operation will be pending until the next clock cycle, reading the same bus after a write operation forces the bus to apply the pending changes immediately. Therefore, this "read" ensures that when the callback exits, the flag has already been cleared.

etienne-lms: Ok. Would you mind adding an inline comment to explain this, even i few words, unless what someone could think it's useless and be tempted to remove the instrcution. 
Maybe something like `/* Read back SoC register to ensure previous write operation is completed */`.

drivers/timer/stm32wb0_radio_timer.c
etienne-lms: My mistake. Indeed it does not apply.

drivers/timer/stm32wb0_radio_timer.c
HoZHel: Done

drivers/timer/stm32wb0_radio_timer.c
etienne-lms: Ok

etienne-lms (approved):

erwango (approved):

PR #101440, opened 2025-12-22
drivers: gpio: emul: validate pins in single-pin APIs and  mask in port APIs

jayvikdesai:
Add input validation to single-pin GPIO APIs (e.g., gpio_pin_configure) and emulator-specific helpers to return -EINVAL for invalid pins. This ensures tests correctly fail when targeting non-existent pins.

For standard port-wide APIs (e.g., gpio_port_set_bits_raw), align behavior with hardware drivers by masking out invalid pins instead of returning an error. This mimics hardware behavior where writing to unimplemented bits is ignored, and ensures compatibility with generic tests that use full 32-bit masks (e.g., 0xFFFFFFFF).

### DIFF
--- a/drivers/gpio/gpio_emul.c
+++ b/drivers/gpio/gpio_emul.c
@@ -333,10 +333,21 @@ int gpio_emul_input_set_masked(const struct device *port, gpio_port_pins_t mask,
 			      gpio_port_value_t values)
 {
 	struct gpio_emul_data *drv_data = (struct gpio_emul_data *)port->data;
+	const struct gpio_emul_config *config = (const struct gpio_emul_config *)port->config;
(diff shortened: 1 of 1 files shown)


drivers/gpio/gpio_emul.c
keith-zephyr: NIT: I think this is easier to read and it matches the other checks done in this file.
```suggestion
	if ((config->common.port_pin_mask & mask) == 0) {
```

jayvikdesai: Acknowledged

keith-zephyr (dismissed):

keith-zephyr (approved):

henrikbrixandersen (approved):

PR #101441, opened 2025-12-22
xtensa: mmu: add assertions and system halting point if there are not enough free L2 tables to be allocated

### LINKED ISSUE #101339: xtensa: ace 3.0: Unchecked error leads to exception

dcpleung:
This adds some assertions when there are not enough free L2 page tables to be allocated. And adds system halting point after these assertions in case assertion is disabled via kconfigs.

Also adds debug logging on page table pool usage during de-/allocation so that there are some visual clues on whether there are enough tables in the pool.

Fixes #101339

### DIFF
--- a/arch/xtensa/core/ptables.c
+++ b/arch/xtensa/core/ptables.c
@@ -968,8 +968,14 @@ static uint32_t *dup_l2_table(uint32_t *src_l2_table, enum dup_action action)
 	uint32_t *l2_table;
 
 	l2_table = alloc_l2_table();
+
+	/* Duplicating L2 tables is a must-have and must-success operation.
+	 * If we are running out of free L2 tables to be allocated, we cannot
+	 * continue.
+	 */
+	__ASSERT_NO_MSG(l2_table != NULL);
 	if (l2_table == NULL) {
-		return NULL;
+		arch_system_halt(K_ERR_KERNEL_PANIC);
 	}
 
 	switch (action) {
--- a/arch/xtensa/core/ptables.c
+++ b/arch/xtensa/core/ptables.c
@@ -407,6 +407,15 @@ static void map_memory_range(const uint32_t start, const uint32_t end,
 			__ASSERT(l2_table != NULL,
 				 "There is no l2 page table available to map 0x%08x\n", page);
 
+			if (l2_table == NULL) {
+				/* This function is called during boot. If this cannot
+				 * properly map all predefined memory regions, it is very
+				 * unlikely for anything to run correctly. So forcibly
+				 * halt the system in case assertion has been turned off.
+				 */
+				arch_system_halt(K_ERR_KERNEL_PANIC);
+			}
+
 			init_page_table(l2_table, L2_PAGE_TABLE_NUM_ENTRIES, PTE_L2_ILLEGAL);
 
 			xtensa_kernel_ptables[l1_pos] =
--- a/arch/xtensa/core/ptables.c
+++ b/arch/xtensa/core/ptables.c
@@ -326,6 +326,26 @@ static void init_page_table(uint32_t *ptable, size_t num_entries, uint32_t val)
 	}
 }
 
+static void calc_l2_page_tables_usage(void)
+{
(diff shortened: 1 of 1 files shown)


arch/xtensa/core/ptables.c
```diff
@@ -968,8 +968,14 @@ static uint32_t *dup_l2_table(uint32_t *src_l2_table, enum dup_action action)
 	uint32_t *l2_table;
 
 	l2_table = alloc_l2_table();
+
+	/* Duplicating L2 tables is a must-have and must-success operation.
+	 * If we are running out of free L2 tables to be allocated, we cannot
+	 * continue.
+	 */
+	__ASSERT_NO_MSG(l2_table != NULL);
```
lgirdwood, line 976: There may be occasion where this will be non fatal for the application. i.e. user request to allocate resources when there are no resources free should be survivable and allow user/upper level application to handle.

arch/xtensa/core/ptables.c
```diff
@@ -407,6 +407,15 @@ static void map_memory_range(const uint32_t start, const uint32_t end,
 			__ASSERT(l2_table != NULL,
 				 "There is no l2 page table available to map 0x%08x\n", page);
 
+			if (l2_table == NULL) {
+				/* This function is called during boot. If this cannot
+				 * properly map all predefined memory regions, it is very
+				 * unlikely for anything to run correctly. So forcibly
+				 * halt the system in case assertion has been turned off.
+				 */
+				arch_system_halt(K_ERR_KERNEL_PANIC);
```
lgirdwood, line 416: I assume this will show as a panic in the logs ?

lgirdwood (commented):
Just a few opens. @lyakh pls review

arch/xtensa/core/ptables.c
lyakh: in principle `init_page_table()` below will trigger an exception anyway, but doesn't hurt having it explicit either

arch/xtensa/core/ptables.c
dcpleung: Currently, it is not possible to recover fully as we do not have the mechanism to undo any changes to page tables up to the failure point. The already modified PTEs would not be reverted to previous states which may have security implications. So it is better to halt the system instead.

ceolin (approved):

arch/xtensa/core/ptables.c
dcpleung: We can only reach this if you have assertion turned off. Besides, this is the last line of defense, and we usually should have bailed out earlier.

peter-mitsis (approved):

edersondisouza (approved):

PR #101442, opened 2025-12-22
github: Add copilot-instructions.md

Thalley:
Add the copilot-instructions.md that Github Copilot will use when performing e.g. code review, as defined by https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review

The content of the file is generated by https://github.com/copilot based on a few inputs, and can be modified/extended as we want.

The purpose of adding the file is to improve the quality and process of using Github Copilot for reviews, and help catch more "trivial" issues and help users to more easily follow the contribution and coding guidelines.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,20 @@
+# GitHub Copilot Instructions
+
+This project follows the coding and contribution guidelines outlined in the [Zephyr Project Documentation](https://docs.zephyrproject.org/latest/). GitHub Copilot should follow these principles when providing suggestions for code completion and reviews:
+
+## Coding Style and Guidelines
+- Adhere to the coding style defined in the [Coding Guidelines](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html).
+  - Ensure code is formatted to meet these style requirements.
+  - Prefer functions and constructs defined in the guidelines.
+
+## Contribution Guidelines
+- Follow the contribution process described in the [Contribution Guide](https://docs.zephyrproject.org/latest/contribute/index.html).
+  - Maintain consistency with existing code.
+  - Require commits to follow the format specified in the guidelines.
+
+## Key Practices for Copilot
+- Prefer idiomatic Zephyr APIs and patterns where applicable.
+- Use documentation examples as the default reference for common tasks.
+- Avoid recommendations that diverge from the principles outlined in the Zephyr documentation.
+
+For detailed standards, contributors are expected to review the latest version of the Zephyr Project documentation linked above.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,20 @@
+# GitHub Copilot Instructions
+
+This project follows the coding and contribution guidelines outlined in the [Zephyr Project Documentation](https://docs.zephyrproject.org/latest/). GitHub Copilot should follow these principles when providing suggestions for code completion and reviews:
+
+## Coding Style and Guidelines
+- Adhere to the coding style defined in the [Coding Guidelines](https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html).
```
omkar3141, line 6: Did you check for sure that this online copilot reviewbot actually fetches the guidelines from index.html? 

Specially, see this: https://github.blog/ai-and-ml/unlocking-the-full-power-of-copilot-code-review-master-your-instructions-files/#:~:text=Trying%20to%20change%20the%20UX,into%20your%20instructions%20files%20instead.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,20 @@
...
+  - Ensure code is formatted to meet these style requirements.
+  - Prefer functions and constructs defined in the guidelines.
+
+## Contribution Guidelines
+- Follow the contribution process described in the [Contribution Guide](https://docs.zephyrproject.org/latest/contribute/index.html).
+  - Maintain consistency with existing code.
+  - Require commits to follow the format specified in the guidelines.
+
+## Key Practices for Copilot
+- Prefer idiomatic Zephyr APIs and patterns where applicable.
```
omkar3141, line 16: This may not work as you think it will work. Because for an LLM, depending on how the backend tooling is done, it is very hard to infer what does 'idiomatic' and 'pattern' mean in the context of 'zephyr'. It is best to insert verbatim examples here (or links, if you can be very sure that those links are fetched by the copilot bot and added to input context while performing review).

.github/copilot-instructions.md
```diff
@@ -0,0 +1,20 @@
+# GitHub Copilot Instructions
```
omkar3141, line 1: I think we should include certain coding best practices for zephyr's codebase directly as verbatim input here.

.github/copilot-instructions.md
Thalley: > Did you check for sure that this online copilot reviewbot actually fetches the guidelines from index.html?

Nope - I couldn't find a way to actually verify this. 

> Specially, see this: https://github.blog/ai-and-ml/unlocking-the-full-power-of-copilot-code-review-master-your-instructions-files/#:~:text=Trying%20to%20change%20the%20UX,into%20your%20instructions%20files%20instead.

Was that page written by Copilot? :D 

I assume that the part you are referring to is (under "What not to do") 

> Including external links. Copilot won’t follow them. You should copy relevant info into your instructions files instead.

but then in their example below they do

> - Use [Jest](https://jestjs.io/) for all testing.

As mentioned in the PR description, I did use Copilot to generate this file (but that doesn't mean that it's done correctly!). 

From https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review they do also provide an example with

> When performing a code review, apply the checks in the `/security/security-checklist.md` file.

So maybe instead of URLs we can/should provide relative paths to the files themselves? 

But you make some good points, and thanks for raising them. Ideally we'd have a way to verify the usefulness of this file before merging.

.github/copilot-instructions.md
Thalley: > depending on how the backend tooling is done

Can you elaborate on what backend tooling you refer to here? 

> It is best to insert verbatim examples here (or links, if you can be very sure that those links are fetched by the copilot bot and added to input context while performing review).

I'd actually not have this file, and have verbatim example in this that we need to update to ensure correctness. If we refer to some example, then we are implicitly telling Copilot "the right way of doing it" or "the best way of doing it", neither of which is something we define anywhere in our coding guidelines for most parts. We have ~100 or so coding guidelines, each containing several examples of "right" and "wrong" (in the context of the guidelines) plus more coding style guides. I don't want to add several hundred examples for it to have the wanted effect, nor do I want this file to contain any sort of "best practices" unless they are specifically written in our documentation, in which case we should refer to those (somehow)

.github/copilot-instructions.md
Thalley: As per my above comments you can probably infer that I disagree with this :)

.github/copilot-instructions.md
omkar3141: > So maybe instead of URLs we can/should provide relative paths to the files themselves?

This might work. I think checking this experimentally is the best way to find out or unless someone from zephyr community is also someone working with creation of this tool in microsoft.ai. We can quickly derive a significant value from this tool if we insert condensed version of zephyr coding guidelines in this file. 

By condensed I mean: Remove everything that tool can anyway infer from local context (e.g. All #else, #elif and #endif preprocessor directives shall reside in the same file with related directives, etc.). Remove all generic instructions which are well known and unverifiable by copilot tool (e.g. compilation check, code shall be traceable to documented requirements, don't use stdio API, etc.).

I think it will be interesting to check if the trick with relative path works. 

It may be possible to create a copy of the main branch in this repo, and ask appropriate authority to enable copilot review on that branch (if possible), and then experiment there if main code base should not be touched.

.github/copilot-instructions.md
omkar3141: > Can you elaborate on what backend tooling you refer to here?

Depending on the implementation, LLM based review bots can have ability to fetch links or additional documents mentioned in the rule files and incorporate them as their input context.

Issue #25663, opened 2020-05-27
Documentation and features of IEEE802.15.4 drivers


markus-becker-tridonic-com:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

The documentation at https://docs.zephyrproject.org/latest is missing information about the available IEEE802.15.4 drivers, which features they support, and which platforms/boards they are supported on.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context or graphics (drag-and-drop an image) about the feature request here.

kartben:
@fgrandel  fyi

fgrandel:
@nandojve Point 2 of the above list has been implemented by @rlubos in the meantime (thanks!). Are you still considering implementing (parts of) the Thread requirements in your driver? If so, this should be possible now.

fgrandel:
Marked as "good first contribution" as this is a rather simple documentation task that requires some initial invest to get going and can be taken forward from there by myself or driver maintainers.

PR #101443, opened 2025-12-22
doc: add initial IEEE 802.15.4 drivers overview

iliaskoro:
This PR adds an initial IEEE 802.15.4 drivers overview section to the
documentation.

It provides a minimal, high-level list of available in-tree radio drivers
as a first step towards documenting driver support and characteristics.

Related to: #25663

### DIFF
--- a/doc/connectivity/networking/api/ieee802154.rst
+++ b/doc/connectivity/networking/api/ieee802154.rst
@@ -37,6 +37,31 @@ Zephyr supports both, native IEEE 802.15.4 and Thread, with 6LoWPAN. Zephyr's
 <https://openthread.io/>`_. The IPv6 header compression in 6LoWPAN is used for
 native IEEE 802.15.4.
 
+IEEE 802.15.4 Drivers Overview
+******************************
+
+This section provides a high-level overview of the IEEE 802.15.4 radio drivers
+available in Zephyr. It is intended as an initial reference to help users
+identify supported drivers and their general characteristics. Detailed feature
+support and board compatibility may vary per driver and are documented in the
+corresponding driver sources.
+
+.. list-table:: Available IEEE 802.15.4 radio drivers (initial list)
+   :header-rows: 1
+
+   * - Driver
+     - Vendor / SoC family
+     - Notes
+   * - nrf5
+     - Nordic Semiconductor nRF52 / nRF53
+     - Widely used 2.4 GHz IEEE 802.15.4 radio driver
+   * - kw41z
+     - NXP KW41Z
+     - IEEE 802.15.4 radio driver for NXP KW series
+   * - cc13xx_cc26xx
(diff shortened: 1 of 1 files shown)


doc/connectivity/networking/api/ieee802154.rst
```diff
@@ -37,6 +37,31 @@ Zephyr supports both, native IEEE 802.15.4 and Thread, with 6LoWPAN. Zephyr's
 <https://openthread.io/>`_. The IPv6 header compression in 6LoWPAN is used for
 native IEEE 802.15.4.
 
+IEEE 802.15.4 Drivers Overview
+*****************************
```
jukkar: This needs to be same length as the title (it is missing one `*`)

iliaskoro: Fixed, thanks.

jukkar (approved):

doc/connectivity/networking/api/ieee802154.rst
```diff
@@ -37,6 +37,31 @@ Zephyr supports both, native IEEE 802.15.4 and Thread, with 6LoWPAN. Zephyr's
...
+corresponding driver sources.
+
+.. list-table:: Available IEEE 802.15.4 radio drivers (initial list)
+   :header-rows: 1
+
+   * - Driver
+     - Vendor / SoC family
+     - Notes
+   * - nrf5
+     - Nordic Semiconductor nRF52 / nRF53
```
rlubos, line 56: It's also for `nRF54L`, but I think simply stating `nRF5` (as the driver does itself) is enough, no need to enumerate families.

doc/connectivity/networking/api/ieee802154.rst
```diff
@@ -37,6 +37,31 @@ Zephyr supports both, native IEEE 802.15.4 and Thread, with 6LoWPAN. Zephyr's
...
+     - Widely used 2.4 GHz IEEE 802.15.4 radio driver
+   * - kw41z
+     - NXP KW41Z
+     - IEEE 802.15.4 radio driver for NXP KW series
+   * - cc13xx_cc26xx
+     - Texas Instruments CC13xx / CC26xx
+     - Supports 2.4 GHz and Sub-GHz configurations
+
```
rlubos, line 64: The list doesn't seem exhaustive though? There are many more drivers listed in https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/ieee802154/CMakeLists.txt.

I think that if we want to document supported drivers in such a form, we should at least make the table complete (as for today).

rlubos (commented):
Please add a signoff entry and some description in the commit body:
> 1: UC2 Commit message does not contain a 'Signed-off-by:' line
1: UC6 Commit message body is empty, should at least have 1 line(s).

PR #101448, opened 2025-12-22
Add Espressif RMT driver

### LINKED ISSUE #72546: ESP32/ESP32S3 RMT
**Is your feature request related to a problem? Please describe.**
In general, when applying an LED strip, we like to use SPI to transmit the signal to it.Sometimes, however, the transmitted signal is interrupted by other interruptions. In addition, it also occupies additional SPI ports.

RMT does a better job of driving a small number of LED strips, it is not interrupted by other interrupts, and it does not need to occupy DMA.

There is no better way to drive the LED strip than the RMT, when the number of led strips is less than 16, the FIFO using RMT can drive the led strip perfectly.

joelguittet:
Add initial version of the Espressif RMT driver.

A sample application is created to demonstrate the usage.

This pull-request depends on an update of hal_espressif: https://github.com/zephyrproject-rtos/hal_espressif/pull/510

Fixes: https://github.com/zephyrproject-rtos/zephyr/issues/72546


include/zephyr/drivers/misc/espressif_rmt/rmt_encoder.h
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * @param[in] config Bytes encoder configuration
+ * @param[out] ret_encoder Returned encoder handle
+ * @return
+ *      - ESP_OK: Create RMT bytes encoder successfully
+ *      - ESP_ERR_INVALID_ARG: Create RMT bytes encoder failed because of invalid argument
+ *      - ESP_ERR_NO_MEM: Create RMT bytes encoder failed because out of memory
+ *      - ESP_FAIL: Create RMT bytes encoder failed because of other error
+ */
+esp_err_t rmt_new_bytes_encoder(const rmt_bytes_encoder_config_t *config, rmt_encoder_handle_t *ret_encoder);
```
kartben: You're writing a zephyr driver, please return zephyr error codes :)
Also use `@retval` for documenting the discrete error values

dts/bindings/misc/espressif,esp32-rmt.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2025 Joel Guittet (joelguittet@gmail.com)
+# SPDX-License-Identifier: Apache-2.0
+
+description: ESP32 RMT
```
kartben: Add a title, spell out the acronym according to the guidelines, and consider adding an actual description :) (bonus point for including an example of a dts snippet as well)

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,253 @@
...
+}
+
+esp_err_t rmt_select_periph_clock(rmt_channel_handle_t chan, rmt_clock_source_t clk_src)
+{
+    esp_err_t ret = ESP_OK;
+    rmt_group_t *group = chan->group;
+    int channel_id = chan->channel_id;
+    uint32_t periph_src_clk_hz = 0;
+    bool clock_selection_conflict = false;
+    // check if we need to update the group clock source, group clock source is shared by all channels
```
kartben: `/*` comments please

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,21 @@
...
+
+menuconfig ESPRESSIF_RMT
+	bool "Espressif RMT driver"
+	help
+	  Enable config options for Espressif RMT driver.
+
+if ESPRESSIF_RMT
+
+config ESPRESSIF_RMT_INIT_PRIORITY
+	int "Espressir RMT init priority"
```
kartben: Espressif

kartben (changes requested):

fabiobaltieri:
Sure it's fair to have it in misc if you are willing to put to work down to fix it up, problem though is that if it gets updated in idf it'll be impossible to track the changes. Anyway, your call on that.

For the rest, yeah prob fair to have it in misc to be used without a specific behavior but then I don't think we should merge a sample that makes it do a functionality that we have a subsystem for if it does not use the subsystem apis. But at that point it should be pretty straightforward to implement a led_strip driver on top of it.

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2025 Joel Guittet (joelguittet@gmail.com)
+# SPDX-License-Identifier: Apache-2.0
+
+menuconfig ESPRESSIF_RMT
+	bool "Espressif RMT driver"
+	help
+	  Enable config options for Espressif RMT driver.
```
wmrsouza: My sugestion:
```c
config RMT_ESP32
	bool "ESP32 RMT driver"
	default y
	depends on DT_HAS_ESPRESSIF_ESP32_RMT_ENABLED
	select DMA if DT_HAS_ESPRESSIF_ESP32_GDMA_ENABLED
	help
	  Enables the ESP32 Remote Control Transceiver driver.
```

joelguittet: My comment on this proposition:
- Renaming the name of the config: ok for me.
- Why defaulting to y ? We may want to just not include it in general. => EDIT: with default y then we just need to activate rmt in the dts of course :-) I add this one.
- select DMA should be added later when I will integrate DMA API for sure. However, this is only for ESP32S3 .so an additional condition should be added  I guess.

joelguittet: @wmrsouza For renaming the config, I cannot. The CI is checking for proper order and consistency of drivers in misc directory. I should be espressif_rmt top have vendor first, and then ESPRESSIF_RMT for the config.

wmrsouza:
@joelguittet I think the driver could have register_rcv_callback() api instead of read()

dts/bindings/misc/espressif,esp32-rmt-device.yaml
```diff
@@ -0,0 +1,28 @@
...
+  pinctrl-0:
+    required: true
+
+  pinctrl-names:
+    required: true
+
+  with-dma:
+    type: boolean
+    description: |
+      If set, the driver will allocate an RMT channel with DMA capability.
```
wmrsouza: There is no need for this property. You can check if dmas and dma-names properties are present in device tree. For reference:
- dma_dev = UTIL_AND(DT_INST_DMAS_HAS_NAME(index, dir), DEVICE_DT_GET(DT_INST_DMAS_CTLR_BY_NAME(index, dir)))
- dma_channel = UTIL_AND(DT_INST_DMAS_HAS_NAME(index, dir), DT_INST_DMAS_CELL_BY_NAME(index, dir, channel))

By the way only channel 3 on esp32s3 is DMA capable (channel 3 memory can only be accessed using DMA)
https://documentation.espressif.com/esp32-s3_technical_reference_manual_en.pdf / 37.3.2.3 RAM Access / DMA Mode

dts/bindings/misc/espressif,esp32-rmt-device.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+  io-loop-back:
+    type: boolean
+    description: |
+      The signal output from the GPIO will be fed to the input path as well.
```
wmrsouza: There is no need for this property. You can configure loopback making input signal as output and output signal as input through the same pin on device tree.

joelguittet: Agree i don't see specific interest, removing this one.

dts/bindings/misc/espressif,esp32-rmt-device.yaml
joelguittet: It is required. Thre are only one "tx" and one "rx" channels available on ESP32S3.
In case multiple RMT devices are used, only one for tx and one for rx can activate DMA. So this flag should be set only for 1 device.

Something like that for example (I taken two led strips but of course can be something else):
```

&rmt {
	status = "okay";

	dmas = < &dma 9 >;
	dma-names = "tx";

	led_strip0: ws2812@0 {
		compatible = "worldsemi,ws2812-espressif-rmt";
		pinctrl-0 = <&led_strip0_rmt_default>;
		pinctrl-names = "default";
		reg = <0>;
		chain-length = <1>;
		color-mapping = <LED_COLOR_ID_GREEN
				 LED_COLOR_ID_RED
				 LED_COLOR_ID_BLUE>;
		reset-delay = <500>;
		with-dma; // This one with DMA so it will take the tx channel defined above
	};

	led_strip1: ws2812@0 {
		compatible = "worldsemi,ws2812-espressif-rmt";
		pinctrl-0 = <&led_strip1_rmt_default>;
		pinctrl-names = "default";
		reg = <0>;
		chain-length = <1>;
		color-mapping = <LED_COLOR_ID_GREEN
				 LED_COLOR_ID_RED
				 LED_COLOR_ID_BLUE>;
		reset-delay = <500>;
		// This one without DMA
	};
};

&dma {
	status = "okay";
};
```

joelguittet: I'm adding this information in the bindings.

dts/bindings/misc/espressif,esp32-rmt-device.yaml
wmrsouza: for reference:
https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/drivers/uart/uart_async_api/socs/esp32c3.overlay

dts/bindings/misc/espressif,esp32-rmt-device.yaml
wmrsouza: @joelguittet I think you are mixing up the information regarding rmt device driver with that regarding to led_strip driver for ws2812 using rmt back end, but lets see (may be I missed something)

dts/bindings/misc/espressif,esp32-rmt-device.yaml
wmrsouza: In my humble opinion, you should separate rmt driver PR from led_strip rmt backend PR, provide a test (with loopback for the RMT driver; with and without DMA) and after that follow with the led_strip rmt backend PR.

I'm saying that because it could exist some application that wants to use rmt with DMA (using esp32s3 rmt channel 3) and this could be a device without a device driver so no compatible as "worldsemi,ws2812-espressif-rm"

dts/bindings/misc/espressif,esp32-rmt-device.yaml
joelguittet: @wmrsouza maybe it's me, do you suggest the dma declaration should be done inside the led strip node like below? Could make more sense and avoid the with-dma flag for sure !

```
&rmt {
	status = "okay";

	led_strip0: ws2812@0 {
		compatible = "worldsemi,ws2812-espressif-rmt";
		pinctrl-0 = <&led_strip0_rmt_default>;
		pinctrl-names = "default";
		reg = <0>;
		chain-length = <1>;
		color-mapping = <LED_COLOR_ID_GREEN
				 LED_COLOR_ID_RED
				 LED_COLOR_ID_BLUE>;
		reset-delay = <500>;
		// This one with DMA so it will take the tx channel defined above
		dmas = < &dma 9 >;
		dma-names = "tx";
	};

	led_strip1: ws2812@0 {
		compatible = "worldsemi,ws2812-espressif-rmt";
		pinctrl-0 = <&led_strip1_rmt_default>;
		pinctrl-names = "default";
		reg = <0>;
		chain-length = <1>;
		color-mapping = <LED_COLOR_ID_GREEN
				 LED_COLOR_ID_RED
				 LED_COLOR_ID_BLUE>;
		reset-delay = <500>;
		// This one without DMA
	};
};

&dma {
	status = "okay";
};
```

joelguittet:
@wmrsouza 'rmt' added to all relevant espressif boards.
For the default pinctrl I don't know. How many channels ? Which GPIO to select ? There is no preferences to provide I guess. I see also the default pinctrl is for uart, spi, i2s. The other peripherals have no default pinctrl or it's somewhere else ?

wmrsouza:
> @wmrsouza 'rmt' added to all relevant espressif boards. For the default pinctrl I don't know. How many channels ? Which GPIO to select ? There is no preferences to provide I guess. I see also the default pinctrl is for uart, spi, i2s. The other peripherals have no default pinctrl or it's somewhere else ?

- at leat one channel 
- preferencialy GPIO near each other, and not used by other peripherals or overlaping with GPIO used by peripheral with `status = disabled;`

joelguittet:
> at leat one channel
preferencialy GPIO near each other, and not used by other peripherals or overlaping with GPIO used by peripheral with status = disabled;

So this is just dead code at the end? @wmrsouza

wmrsouza:
> > at leat one channel
> > preferencialy GPIO near each other, and not used by other peripherals or overlaping with GPIO used by peripheral with status = disabled;
> 
> So this is just dead code at the end? @wmrsouza

no, I'm NOT saying to make rmt `status="disabled";`. What I'm saying is if the board doesn't have "free" pins then ovelap those from peripherals having `status="disabled";`

wmrsouza (commented):
@joelguittet Suggestion for commit title and message body
- title:
manifest: esp32: update hal_espressif
- message body
Update hal to cover latest features and fixes.

samples/drivers/misc/espressif_rmt/src/main.c
```diff
@@ -0,0 +1,329 @@
...
+
+rmt_encoder_handle_t tx_encoder;
+
+rmt_symbol_word_t raw_symbols[64];
+rmt_receive_config_t receive_config = {
+	.signal_range_min_ns = 125,
+	.signal_range_max_ns = 2500000,
+};
+
+static size_t example_encoder_encode(rmt_encoder_t *encoder,
```
wmrsouza: Is the example prefix necessary? (nitpick)
The same question applies to all other occurrences.

wmrsouza:
@joelguittet Overral is not bad. In fact is very good code but would not be better treat rmt channels as nodes with addresses (channel index under rmt peripheral) under device tree and a setup function to configure channels (configurations that chan change and are not informet in device threem, rx callback function receive a device struct as a parameter, ...) or some more Zephyr's similar kind of api function (I know rmt is Espressif only peripheral) 

@sylvioalves @marekmatej @raffarost @LucasTambor let me know what you think about this. Might be I'm being too strict again?

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,77 @@
...
+		return rc;
+	}
+
+	return 0;
+}
+
+/* Retrieve DMA device if defined, else default to NULL */
+#define ESPRESSIF_RMT_DT_INST_DMA_CTLR(idx)                                         \
+	COND_CODE_1(DT_INST_NODE_HAS_PROP(idx, dmas),                                   \
+		(DEVICE_DT_GET(DT_INST_DMAS_CTLR(0))),                                      \
```
sylvioalves: DMa is hardcoded with `0`. Use `idx` instead:
`(DEVICE_DT_GET(DT_INST_DMAS_CTLR(idx))),`

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,77 @@
...
+/* Retrieve DMA channel if defined, else default to ESPRESSIF_RMT_DMA_CHANNEL_UNDEFINED */
+#define ESPRESSIF_RMT_DT_INST_DMA_CELL(idx, name, cell)                             \
+	COND_CODE_1(DT_INST_NODE_HAS_PROP(idx, dmas),                                   \
+		(COND_CODE_1(DT_INST_DMAS_HAS_NAME(idx, name),                              \
+		(DT_INST_DMAS_CELL_BY_NAME(idx, name, cell)),                               \
+		(ESPRESSIF_RMT_DMA_CHANNEL_UNDEFINED))),                                    \
+		(ESPRESSIF_RMT_DMA_CHANNEL_UNDEFINED))
+
+#define ESPRESSIF_RMT_INIT(idx)                                                     \
+	PINCTRL_DT_DEFINE(DT_NODELABEL(rmt));                                           \
```
sylvioalves:
```
+	PINCTRL_DT_INST_DEFINE(idx);                                                    \
+	static const struct espressif_rmt_config espressif_rmt_cfg_##idx = {            \
+		.pcfg = PINCTRL_DT_INST_DEV_CONFIG_GET(idx),                                \
 		.dma_dev = ESPRESSIF_RMT_DT_INST_DMA_CTLR(idx),                             \
-		.tx_dma_channel = ESPRESSIF_RMT_DT_INST_DMA_CELL(0, tx, channel),           \
-		.rx_dma_channel = ESPRESSIF_RMT_DT_INST_DMA_CELL(0, rx, channel),           \
+		.tx_dma_channel = ESPRESSIF_RMT_DT_INST_DMA_CELL(idx, tx, channel),         \
+		.rx_dma_channel = ESPRESSIF_RMT_DT_INST_DMA_CELL(idx, rx, channel),         \
```

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,77 @@
...
+		.irq_priority = DT_INST_IRQ_BY_IDX(idx, 0, priority),                       \
+		.irq_flags = DT_INST_IRQ_BY_IDX(idx, 0, flags),                             \
+	};                                                                              \
+	static struct espressif_rmt_data espressif_rmt_data_##idx = {                   \
+		.hal = {                                                                    \
+			.regs = (rmt_soc_handle_t)DT_INST_REG_ADDR(idx),                        \
+		},                                                                          \
+	};                                                                              \
+	DEVICE_DT_INST_DEFINE(idx, rmt_init, NULL, &espressif_rmt_data_##idx,           \
+		&espressif_rmt_cfg_##idx, PRE_KERNEL_1, CONFIG_ESPRESSIF_RMT_INIT_PRIORITY, \
```
sylvioalves: I guess it should be `POST_KERNEL`?

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,37 @@
...
+	depends on DT_HAS_ESPRESSIF_ESP32_RMT_ENABLED
+	select DMA if DT_HAS_ESPRESSIF_ESP32_GDMA_ENABLED && SOC_SERIES_ESP32S3
+	help
+	  Enables the ESP32 Remote Control Transceiver driver.
+
+if ESPRESSIF_RMT
+
+config ESPRESSIF_RMT_INIT_PRIORITY
+	int "ESP32 RMT driver init priority"
+	default 90
```
sylvioalves: `default 50`

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,77 @@
...
+#error "DMA peripheral is not enabled!"
+#endif
+
+LOG_MODULE_REGISTER(espressif_rmt, CONFIG_ESPRESSIF_RMT_LOG_LEVEL);
+
+static int rmt_init(const struct device *dev)
+{
+	const struct espressif_rmt_config *config = dev->config;
+	int rc;
+
```
sylvioalves, line 30: Missing common driver init checks:

```
if (!device_is_ready(config->clock_dev)) {
	LOG_ERR("Clock device not ready");
	return -ENODEV;
}

#if SOC_RMT_SUPPORT_DMA
	if (config->dma_dev && !device_is_ready(config->dma_dev)) {
		LOG_ERR("DMA device not ready");
		return -ENODEV;
}
#endif
```

include/zephyr/drivers/misc/espressif_rmt/rmt_common.h
```diff
@@ -0,0 +1,106 @@
...
+#endif
+
+/**
+ * @brief RMT carrier wave configuration (for either modulation or demodulation)
+ */
+typedef struct {
+	/** Carrier wave frequency, in Hz, 0 means disabling the carrier */
+	uint32_t frequency_hz;
+	/** Carrier wave duty cycle (0~100%) */
+	float duty_cycle;
```
sylvioalves: I think having this as float isn't helpful due to FPU and SoC dependency. I suggest modifying to:
`uint16_t duty_cycle;`
`uint32_t high_ticks = total_ticks * config->duty_cycle / 1000;`
or if you follow Zephyr's PWM API:
```
+	uint32_t period_ticks;
+	uint32_t pulse_ticks;
```

sylvioalves:
1. Would you also add a test case? `tests/drivers/misc/espressif_rmt/ with loopback TX→RX`

dts/bindings/misc/espressif,esp32-rmt.yaml
```diff
@@ -0,0 +1,76 @@
...
+    Espressif Remote Control Transceiver (RMT)
+
+    The RMT (Remote Control Transceiver) peripheral was designed to act as an infrared transceiver.
+    However, due to the flexibility of its data format, RMT can be extended to a versatile and
+    general-purpose transceiver, transmitting or receiving many other types of signals.
+
+    pinctrl is used to configure RMT channels and GPIO to be used.
+    When using ESP32 or ESP32S2 series, each RMT channel should be used once (eg. RMT_OUT0_GPIOxx
+    and RMT_IN0_GPIOxx can't be used at the same time, etc.). An error is displayed when trying to
+    create channels whith such configuration.
```
sylvioalves: `create channels with such configuration`

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,37 @@
...
+	help
+	  ESP32 RMT driver initialization priority in POST_KERNEL.
+
+config ESPRESSIF_RMT_ISR_IRAM_SAFE
+	bool "RMT ISR IRAM-Safe"
+	default n
```
sylvioalves: `default n` can go as it is default.

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,37 @@
...
+	help
+	  Ensure the RMT interrupt is IRAM-Safe by allowing the interrupt handler to be
+	  executable when the cache is disabled (e.g. SPI Flash write).
+
+config ESPRESSIF_RMT_PM
+	bool "Support for power management"
+	default n
```
sylvioalves: same

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1321 @@
...
+#include "soc/rtc.h"
+#include "hal/rmt_ll.h"
+#include "rmt_private.h"
+#include "esp_memory_utils.h"
+
+#include <zephyr/drivers/dma.h>
+#include <zephyr/drivers/dma/dma_esp32.h>
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/drivers/pinctrl/pinctrl_esp32_common.h>
+#include <zephyr/logging/log.h>
```
sylvioalves, line 20: would you move the LOG module not to be between #includes? same on `rmt_tx.c` file.

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,77 @@
...
+#include <zephyr/drivers/misc/espressif_rmt/rmt.h>
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/logging/log.h>
+#include <esp_attr.h>
+
+#if SOC_RMT_SUPPORT_DMA && !DT_HAS_COMPAT_STATUS_OKAY(espressif_esp32_gdma)
+#error "DMA peripheral is not enabled!"
+#endif
+
+LOG_MODULE_REGISTER(espressif_rmt, CONFIG_ESPRESSIF_RMT_LOG_LEVEL);
```
sylvioalves, line 24: This is correct as it is registering the RMT logging.
But in all other .c files, it should only `DECLARE` it, not register again. So please update other .c files to use `LOG_MODULE_DECLARE(espressif_rmt, CONFIG_ESPRESSIF_RMT_LOG_LEVEL);` instead.

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,923 @@
...
+		LOG_WRN("channel resolution loss, real=%"PRIu32, rx_channel->base.resolution_hz);
+	}
+
+	rx_channel->filter_clock_resolution_hz = group->resolution_hz;
+	/*
+	 * On esp32 and esp32s2, the counting clock used by the RX filter always comes from
+	 * APB clock no matter what the clock source is used by the RMT channel as the "core"
+	 * clock
+	 */
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
```
sylvioalves: `#if defined(CONFIG_SOC_SERIES_ESP32) || defined(CONFIG_SOC_SERIES_ESP32S2)`

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1321 @@
...
+
+#if SOC_RMT_SUPPORT_DMA
+	if (channel->dma_dev) {
+		rc = dma_start(channel->dma_dev, channel->dma_channel);
+		if (rc) {
+			LOG_ERR("Starting DMA channel failed");
+			return;
+		}
+		/* delay a while, wait for DMA data going to RMT memory block */
+		esp_rom_delay_us(1);
```
sylvioalves: Use Zephyr API instead: `k_busy_wait(1);`.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1321 @@
...
+			rmt_tx_mark_eof(tx_chan);
+			encoded_symbols += 1;
+		}
+	}
+
+	/* for loop transaction, the memory block should accommodate all encoded RMT symbols */
+	if (t->loop_count != 0) {
+		if (unlikely(encoded_symbols > tx_chan->base.mem_block_num
+			* SOC_RMT_MEM_WORDS_PER_CHANNEL)) {
+			ESP_DRAM_LOGE("rmt",
```
sylvioalves: `LOG_ERR`

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,923 @@
...
+			& RMT_LL_EVENT_RX_ERROR(channel_id)) {
+		key = k_spin_lock(&channel->spinlock);
+		rmt_ll_rx_reset_pointer(hal->regs, channel_id);
+		k_spin_unlock(&channel->spinlock, key);
+		/*
+		 * this clear operation can only take effect after we copy out the received
+		 * data and reset the pointer
+		 */
+		rmt_ll_clear_interrupt_status(hal->regs, RMT_LL_EVENT_RX_ERROR(channel_id));
+		ESP_DRAM_LOGE("rmt", "hw buffer too small, received symbols truncated");
```
sylvioalves: `LOG_ERR`. Check all files please.

drivers/misc/espressif_rmt/rmt.c
joelguittet: Done

drivers/misc/espressif_rmt/rmt.c
joelguittet: Done

drivers/misc/espressif_rmt/rmt.c
joelguittet: I copied on other ESP32 drivers, but okay to change to POST_KERNEL. Done.

drivers/misc/espressif_rmt/Kconfig
joelguittet: Probably better with `config KERNEL_INIT_PRIORITY_DEVICE` (which default to 50).

drivers/misc/espressif_rmt/rmt.c
joelguittet: Done

include/zephyr/drivers/misc/espressif_rmt/rmt_common.h
joelguittet: Done for the first proposal. Comment describing the `duty_cycle` variable updated to indicate the unit is 0.1%

dts/bindings/misc/espressif,esp32-rmt.yaml
joelguittet: Good catch! Fixed.

drivers/misc/espressif_rmt/Kconfig
joelguittet: Correct, removed.

drivers/misc/espressif_rmt/Kconfig
joelguittet: Removed too.

joelguittet:
> 1. Would you also add a test case? `tests/drivers/misc/espressif_rmt/ with loopback TX→RX`

Examples in `samples/drivers/misc/espressif_rmt` directory implement a loopback. This will mean duplicate a sample to go in the tests ?

drivers/misc/espressif_rmt/rmt.c
joelguittet: Correct, fixed.

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Fixed

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Fixed. Note: you have a few of them in i2c and sdhc driver + hw_init. Let me know if you would like a PR for them (if relevant, which I'm not sure, but maybe preferable to change using k_sleep with K_USEC).

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Done.

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Done all.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Done

dts/bindings/misc/espressif,esp32-rmt.yaml
```diff
@@ -0,0 +1,76 @@
...
+        reg = <0x60016000 0x1000>;
+        pinctrl-0 = <&rmt_default>;
+        pinctrl-names = "default";
+        dmas = <&dma 8>, <&dma 9>;
+        dma-names = "rx", "tx";
+        interrupts = <RMT_INTR_SOURCE IRQ_DEFAULT_PRIORITY 0>;
+        interrupt-parent = <&intc>;
+        clocks = <&clock ESP32_RMT_MODULE>;
+        status = "okay";
+    };
```
kartben, line 51: use `examples:` for this

dts/bindings/misc/espressif,esp32-rmt.yaml
```diff
@@ -0,0 +1,76 @@
+# SPDX-FileCopyrightText: Copyright The Zephyr Project Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+    Espressif Remote Control Transceiver (RMT)
+
```
kartben, line 6:
```suggestion
title: Espressif RMT (Remote Control Transceiver)

description: |
```

include/zephyr/drivers/misc/espressif_rmt/rmt.h
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * @file
+ * @brief Espressif RMT public API
+ */
+
+#ifndef ZEPHYR_DRIVERS_MISC_ESPRESSIF_RMT_RMT_H_
+#define ZEPHYR_DRIVERS_MISC_ESPRESSIF_RMT_RMT_H_
+
+/**
+ * @brief Espressif RMT.
```
kartben, line 16:
```suggestion
 * @brief Interfaces for Espressif RMT (Remote Control Transceiver).
```

drivers/misc/espressif_rmt/Kconfig
```diff
@@ -0,0 +1,41 @@
...
+	  Ensure the RMT interrupt is Cache-Safe by allowing the interrupt handler to be
+	  executable when the cache is disabled (e.g. SPI Flash write).
+
+config ESPRESSIF_RMT_RX_ISR_CACHE_SAFE
+	bool "RMT RX ISR Cache Safe"
+	help
+	  Ensure the RMT interrupt is Cache-Safe by allowing the interrupt handler to be
+	  executable when the cache is disabled (e.g. SPI Flash write).
+
+config ESPRESSIF_RMT_PM
```
raffarost: This would be "replaced" (removed) by `CONFIG_PM` (as reference), and the locks need to be implemented following Zephyr's PM system throughout the driver.

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,1015 @@
...
+#include <zephyr/dt-bindings/pinctrl/esp32c5-gpio-sigmap.h>
+#elif defined(CONFIG_SOC_SERIES_ESP32C6)
+#include <zephyr/dt-bindings/pinctrl/esp32c6-gpio-sigmap.h>
+#elif defined(CONFIG_SOC_SERIES_ESP32H2)
+#include <zephyr/dt-bindings/pinctrl/esp32h2-gpio-sigmap.h>
+#endif
+
+LOG_MODULE_DECLARE(espressif_rmt);
+
+#define ALIGN_UP(num, align) (((num) + ((align) - 1)) & ~((align) - 1))
```
raffarost: is this used? you can likely remove it or replace for `ROUND_UP(x, align)`

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,1015 @@
...
+		goto err;
+	}
+
+	rmt_group_t *group = rx_channel->base.group;
+	rmt_hal_context_t *hal = &group->hal;
+	int channel_id = rx_channel->base.channel_id;
+
+#if SOC_RMT_SUPPORT_SLEEP_RETENTION
+	if (config->flags.allow_pd != 0) {
+		rmt_create_retention_module(group);
```
raffarost: This is also related to power management (sleep), so you can likely remove all references to `SOC_RMT_SUPPORT_SLEEP_RETENTION` in the code, or adapt to Zephyr standard (as discussed).
Maybe would become then (if PM will be supported now):

`#if CONFIG_PM && SOC_RMT_SUPPORT_SLEEP_RETENTION`

The driver should then compile without those two PM related files (from HAL).

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,473 @@
...
+	bool do_deinitialize = false;
+	rmt_hal_context_t *hal = &group->hal;
+
+	s_platform.key = irq_lock();
+	s_platform.group_ref_counts[group->group_id]--;
+	if (s_platform.group_ref_counts[group->group_id] == 0) {
+		do_deinitialize = true;
+		s_platform.groups[group->group_id] = NULL;
+		/* disable core clock */
+		PERIPH_RCC_ATOMIC() {
```
raffarost: I think you might want to check the usage of this macro (if it is really necessary).
Not sure how (and if) it translates in Zephyr.

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,473 @@
...
+#endif
+	default:
+		break;
+	}
+
+	if (do_deinitialize) {
+#if SOC_RMT_SUPPORT_SLEEP_RETENTION
+		sleep_retention_module_t module = rmt_reg_retention_info[group->group_id].module;
+
+		if (sleep_retention_is_module_created(module)) {
```
raffarost: Sleep related stuff, as discussed already.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1323 @@
...
+#include <zephyr/dt-bindings/pinctrl/esp32c6-gpio-sigmap.h>
+#elif defined(CONFIG_SOC_SERIES_ESP32H2)
+#include <zephyr/dt-bindings/pinctrl/esp32h2-gpio-sigmap.h>
+#endif
+
+LOG_MODULE_DECLARE(espressif_rmt);
+
+/* RMT TX Channel ID from pinmux configuration */
+#define RMT_TX_CHANNEL_ID(pinmux) \
+	ESP32_PIN_SIGO(pinmux) - ESP_RMT_SIG_OUT0
```
raffarost: Usually the channel ID is configured via device tree, and Zephyr calls the API informing which channel is being operated. Deriving from the IO (even though it's a clever idea), is not very standard :-)
Maybe take a look at other Espressif drivers as reference.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1323 @@
...
+		 */
+		rmt_ll_enable_interrupt(group->hal.regs, RMT_LL_EVENT_TX_THRES(channel->channel_id),
+			false);
+		k_spin_unlock(&group->spinlock, key);
+	}
+
+	return need_eof_marker ? 1 : 0;
+}
+
+static size_t IRAM_ATTR rmt_encode_check_result(rmt_tx_channel_t *tx_chan, rmt_tx_trans_desc_t *t)
```
raffarost, line 820: should the next two functions be unconditionally allocated in IRAM?
maybe allow this optimization (IRAM) only when adding `RMT_TX_ISR_HANDLER_IN_IRAM`

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1323 @@
...
+		LOG_DBG("enable carrier modulation for channel(%d,%d), freq=%"PRIu32"Hz",
+			group->group_id, channel->channel_id, real_frequency);
+	} else {
+		LOG_DBG("disable carrier modulation for channel(%d,%d)", group->group_id,
+			channel->channel_id);
+	}
+	return 0;
+}
+
+static bool IRAM_ATTR rmt_isr_handle_tx_threshold(rmt_tx_channel_t *tx_chan)
```
raffarost: There are more functions in IRAM, so the total amount of memory might be significant indeed.

boards/espressif/esp32_devkitc/esp32_devkitc-pinctrl.dtsi
```diff
@@ -120,4 +120,14 @@
 				 <I2S1_I_SD_GPIO11>;
 		};
 	};
+
+	rmt_default: rmt_default {
+		group1 {
+			pinmux = <RMT_OUT0_GPIO13>;
```
raffarost, line 102: I wouldn't create these entries on all boards, but instead move the `pinctrl` config to the sample as these are not fixed resources of the boards.

joelguittet: @wmrsouza ask me like that :D 
So we need to decide, both solution works for me.

drivers/misc/espressif_rmt/Kconfig
joelguittet: On it!

joelguittet: See also my answer in the HAL PR.

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Good catch, not used at all. Removing.

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Good point. Some ESP MCUs have not `SOC_RMT_SUPPORT_SLEEP_RETENTION` so it should be `#if CONFIG_PM && SOC_RMT_SUPPORT_SLEEP_RETENTION`. Updating.

drivers/misc/espressif_rmt/rmt_common.c
joelguittet: Yep!

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Thanks ahahah
Yes not very standard probably, but can you elaborate on the possibility to retrieve it from dts? Actually I have only the pins (see rmt_default). Moreover pins and channel ID are linked so duplicating the information in the dts can introduce errors. What do you think?

drivers/misc/espressif_rmt/rmt_common.c
joelguittet: It translate in the HAL to an atomic counter and an irq lock.
The logic comes from the ESP-IDF RMT driver.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: I have not `RMT_TX_ISR_HANDLER_IN_IRAM` and I don't see similar logic in other esp drivers.
The IRAM_ATTR here come from the ESP-IDF RMT driver, I just kept it.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Agree. Logic from ESP-IDF RMT driver but I'm open to suggestions.
If i read correctly it seems ISR and related functions called from ISR are in IRAM, which make sense, and correspond to t he logic being used by other esp driver as well. Isn't it?

joelguittet:
Hello guys! I know there is 4.4.0 soon and scope is frozen, just a ping to remind you this stuff :-)
@raffarost any concern about the answers i provided to your above comments?
@sylvioalves how is it on your side? Additional review to do or is it almost done?

drivers/misc/espressif_rmt/rmt_rx.c
raffarost: I'm working on sleep retention support for a number of drivers.
I'll soon post a PR and will reference it here, for you to take a look.

drivers/misc/espressif_rmt/rmt_tx.c
raffarost: I think at least in latest HAL/IDF, the function does not define IRAM_ATTR directly, as it has a linking script to do the job of reading the optimization functions in Kconfig (such as `RMT_TX_ISR_HANDLER_IN_IRAM`) and allocate accordingly. Take a look at `components/esp_driver_rmt/linker.lf`. In Zephyr we'll need the Kconfigs and the conditional statements are coded in `default.ld` liker file.

joelguittet:
@sylvioalves you told me previously about tests. I looked a bit at existing tests generally. Seems API testing is a minimum to do. I should probably add this, correct? Do you have any recommendations or expectations on this side at Espressif?

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: I see this in the esp-idf driver. But for zephyr what should we do now?
I mean, other esp32 drivers already in Zephyr also use IRAM_ATTR (uart, i2c, etc)

drivers/misc/espressif_rmt/rmt_rx.c
raffarost: Take a look at #107344 for reference about sleep retention (would be nice to follow the same style).
This PR will enable testing the feature on your driver, if you're using a C6.

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,93 @@
...
+	/* Check clock device is ready */
+	if (!device_is_ready(config->clock_dev)) {
+		LOG_ERR("DMA device not ready");
+		return -ENODEV;
+	}
+
+#if SOC_RMT_SUPPORT_DMA
+	/* Check DMA device is ready (if used) */
+	if (config->dma_dev) {
+		if (!device_is_ready(config->dma_dev)) {
```
sylvioalves, line 40: You could make this 1 single line: `if (config->dma_dev && !device_is_ready(config->dma_dev)) {`

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,93 @@
...
+			LOG_ERR("Clock device not ready");
```
sylvioalves, line 41: `DMA not ready`

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,1023 @@
...
+#include <zephyr/dt-bindings/pinctrl/esp32c6-gpio-sigmap.h>
+#elif defined(CONFIG_SOC_SERIES_ESP32H2)
+#include <zephyr/dt-bindings/pinctrl/esp32h2-gpio-sigmap.h>
+#endif
+
+LOG_MODULE_DECLARE(espressif_rmt);
+
+/* RMT RX Channel ID from pinmux configuration */
+#define RMT_RX_CHANNEL_ID(pinmux) \
+	ESP32_PIN_SIGI(pinmux) - ESP_RMT_SIG_IN0
```
sylvioalves, line 47: Missing parentheses: 
`(ESP32_PIN_SIGI(pinmux) - ESP_RMT_SIG_IN0)`

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1328 @@
...
+	}
+	for (size_t i = 0; i < config->array_size; i++) {
+		synchro->tx_channel_array[i] = config->tx_channel_array[i];
+	}
+	synchro->array_size = config->array_size;
+
+	/* acquire group handle, increase reference count */
+	rmt_group_t *group = rmt_acquire_group_handle(config->tx_channel_array[0]->group->group_id);
+	/* sanity check */
+	assert(group);
```
sylvioalves, line 519: PLease convert all `assert` to `__ASSERT_NO_MSG` instead.

drivers/misc/espressif_rmt/rmt.c
```diff
@@ -0,0 +1,93 @@
...
+LOG_MODULE_REGISTER(espressif_rmt, CONFIG_ESPRESSIF_RMT_LOG_LEVEL);
+
+static int rmt_init(const struct device *dev)
+{
+	const struct espressif_rmt_config *config = dev->config;
+	int rc;
+
+	/* Check clock device is ready */
+	if (!device_is_ready(config->clock_dev)) {
+		LOG_ERR("DMA device not ready");
```
sylvioalves, line 33: RMT clock not ready

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,1023 @@
...
+		LOG_ERR("set group clock failed");
+		goto err;
+	}
+	rx_channel->filter_clock_resolution_hz = group->resolution_hz;
+	/*
+	 * On esp32 and esp32s2, the counting clock used by the RX filter always comes from
+	 * APB clock no matter what the clock source is used by the RMT channel as the "core"
+	 * clock
+	 */
+#if defined(CONFIG_IDF_TARGET_ESP32) || defined(CONFIG_IDF_TARGET_ESP32S2)
```
sylvioalves, line 393: Please replace all CONFIG_IDF_xxx to proper Zephyr-like `CONFIG_SOC_ESPxxxx`.

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,452 @@
...
+#include "esp_private/sleep_retention.h"
+#endif
+
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_DECLARE(espressif_rmt);
+
+typedef struct rmt_platform_t {
+	/**< Platform level mutex lock */
+	unsigned int key;
```
sylvioalves, line 22: There is no need to global locking, this can go.

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,452 @@
...
+static int rmt_create_sleep_retention_link_cb(void *arg);
+#endif
+
+rmt_group_t *rmt_acquire_group_handle(int group_id)
+{
+	bool new_group = false;
+	rmt_group_t *group = NULL;
+
+	/* prevent install rmt group concurrently */
+	s_platform.key = irq_lock();
```
sylvioalves, line 41: Use local locking/unlocking instead.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1328 @@
...
+	tx_chan->mem_end = tx_chan->ping_pong_symbols * 3 - tx_chan->mem_end;
+
+	return false;
+}
+
+static bool IRAM_ATTR rmt_isr_handle_tx_done(rmt_tx_channel_t *tx_chan)
+{
+	rmt_channel_t *channel = &tx_chan->base;
+	rmt_tx_trans_desc_t *trans_desc = NULL;
+	bool need_yield = false;
```
sylvioalves, line 1171: Zephyr does not need to yield. 
1. Remove all `need_yield` from the code
2. Replace `bool` function signatures to `void` that only returns/uses the `need_yield` flag.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1328 @@
...
+			encoded_symbols += rmt_tx_mark_eof(tx_chan,
+				!(encode_state & RMT_ENCODING_WITH_EOF));
+		}
+	}
+
+	/* for loop transaction, the memory block should accommodate all encoded RMT symbols */
+	if (t->loop_count != 0) {
+		if (unlikely(encoded_symbols > tx_chan->base.mem_block_num
+			* SOC_RMT_MEM_WORDS_PER_CHANNEL)) {
+			LOG_ERR("encoding artifacts can't exceed hw memory block for loop "
```
sylvioalves, line 849: I think we should avoid/remove `LOG_ERR` from ISR path.

drivers/misc/espressif_rmt/rmt_encoder_simple.c
```diff
@@ -0,0 +1,253 @@
...
+					 */
+					simple_encoder->callback_done = true;
+				} else {
+					if (enc_size == 0) {
+						/*
+						 * According to the callback docs, this is illegal.
+						 * Report this. EARLY_LOGE as we're running from an
+						 * ISR.
+						 */
+						LOG_ERR("rmt_encoder_simple: encoder callback "
```
sylvioalves, line 146: Same here, can you remove the `LOG_ERR` from ISR path?

drivers/misc/espressif_rmt/rmt_encoder_simple.c
```diff
@@ -0,0 +1,253 @@
...
+
+	simple_encoder->last_symbol_index = 0;
+	simple_encoder->ovf_buf_fill_len = 0;
+	simple_encoder->ovf_buf_parsed_pos = 0;
+	simple_encoder->callback_done = false;
+
+	return 0;
+}
+
+static size_t rmt_encode_simple(rmt_encoder_t *encoder, rmt_channel_handle_t channel,
```
sylvioalves, line 48: Add `IRAM_ATTR`.

drivers/misc/espressif_rmt/rmt_encoder_simple.c
```diff
@@ -0,0 +1,253 @@
...
+	const void *data, size_t data_size, rmt_encode_state_t *ret_state)
```
sylvioalves, line 49: Would you fix those sonarqube issues?

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,452 @@
...
+			 * Group interrupt priority is shared between all channels,
+			 * it will be set when allocate the first channel.
+			 */
+			group->intr_priority = RMT_GROUP_INTR_PRIORITY_UNINITIALIZED;
+			/* enable the bus clock for the RMT peripheral */
+			PERIPH_RCC_ATOMIC() {
+				rmt_ll_enable_bus_clock(group_id, true);
+				rmt_ll_reset_register(group_id);
+			}
+#if CONFIG_PM && SOC_RMT_SUPPORT_SLEEP_RETENTION
```
raffarost, line 65: We should enable sleep retention only when `CONFIG_PM_POWER_DOWN_PERIPHERAL_IN_LIGHT_SLEEP=y`. You could add to `rmt_private.h` something like:

```c
#if CONFIG_PM_POWER_DOWN_PERIPHERAL_IN_LIGHT_SLEEP && SOC_RMT_SUPPORT_SLEEP_RETENTION
#define RMT_SLEEP_RETENTION_ENABLED 1
#else
#define RMT_SLEEP_RETENTION_ENABLED 0
#endif
```

And use it elsewhere in the code for a clean read.

drivers/misc/espressif_rmt/rmt_common.c
raffarost: I just realized, maybe you removed the PM code from the driver.
If that's the case, all things related to sleep retention could be sacked as well, as it wouldn't make sense to only add those.

drivers/misc/espressif_rmt/rmt.c
joelguittet: Good catch, and the above for clock is with a DMA in the log. Fixing it too.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Done for all assert.

drivers/misc/espressif_rmt/rmt.c
joelguittet: Sure, done

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Done. And also for rmt_tx.c which has a similar definition

drivers/misc/espressif_rmt/rmt.c
joelguittet: Catched together with the previous log comment.

drivers/misc/espressif_rmt/rmt_rx.c
joelguittet: Done

drivers/misc/espressif_rmt/rmt_common.c
joelguittet: True. Initially copied from ESP-IDF logic.

drivers/misc/espressif_rmt/rmt_common.c
joelguittet: Done

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: I know that, I was not sure what to do with it. So removing this need_yield.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: Seems that many things are printed on ISR, that would mean much less information in case of error. Is there an alternative?

drivers/misc/espressif_rmt/rmt_encoder_simple.c
joelguittet: Do you mean the cognitive complexity? Seems acceptable according to me, it's not so much. 25 is the default from SonarQube that I generally consider to low in C. Good comments being the key to reduce the complexity, which is not considered in the calculation. Additionally there are about 1200 cognitive complexity issue in Sonar report actually, and some are much higher that this one...
Tell me if this is strictly required here. If so, I will do it.

drivers/misc/espressif_rmt/rmt_tx.c
joelguittet: @sylvioalves note the doc here https://docs.zephyrproject.org/latest/services/logging/index.html#default-frontend indicates "the logging API can be called in an interrupt".

drivers/misc/espressif_rmt/rmt_common.c
joelguittet: @raffarost it seems `CONFIG_PM_POWER_DOWN_PERIPHERAL_IN_LIGHT_SLEEP` doesn't exist in Zephyr, but I see it several times in hal_espressif. The CI also complains here about this missing option. What is wrong there? Maybe it has been removed recently?

joelguittet:
@sylvioalves @raffarost several of your comments above fixed, please advise. Thanks!!

drivers/misc/espressif_rmt/rmt_encoder_simple.c
joelguittet: Note: seems it has gone on the latest analysis... (without changing anything)

sylvioalves:
> @sylvioalves @raffarost several of your comments above fixed, please advise. Thanks!!

Checking again, thanks for updating.

sylvioalves:
> @sylvioalves you told me previously about tests. I looked a bit at existing tests generally. Seems API testing is a minimum to do. I should probably add this, correct? Do you have any recommendations or expectations on this side at Espressif?

Would be good if some `tests/drivers/misc/espressif_rmt/` could be added. Use gpio loopback to test the rmt driver and provide pass/fail test cases. That will be pulled into internal CI to verify in real board. We can work this out in another PR as well.

joelguittet:
> > @sylvioalves you told me previously about tests. I looked a bit at existing tests generally. Seems API testing is a minimum to do. I should probably add this, correct? Do you have any recommendations or expectations on this side at Espressif?
> 
> Would be good if some `tests/drivers/misc/espressif_rmt/` could be added. Use gpio loopback to test the rmt driver and provide pass/fail test cases. That will be pulled into internal CI to verify in real board. We can work this out in another PR as well.

Yes i would definitively have such test but maybe it's better on a separated PR. But I will try to work on it in parallel of current reviews here.

drivers/misc/espressif_rmt/rmt_common.c
raffarost: You're correct, on `main` this option is still not included. The PR #107344 is the one enabling its use since other changes are also needed for this hardware feature to work correctly. I'd suggest you remove anything related to PM for now, and we can add that in a follow up PR.

joelguittet: @raffarost I moved the sleep retention stuff to a draft PR waiting yours to be merged so we are not blocked for the current PR: https://github.com/zephyrproject-rtos/zephyr/pull/108146

joelguittet:
@sylvioalves I have added here a "loopback" test application, and executed it on my esp32s3-devkitc is fine. It's checking expected lines in the console. It's based on the bytes encoder sample. Tell me if this is suitable, this is the very first time i do that in Zephyr repo :-)

drivers/misc/espressif_rmt/rmt_rx.c
```diff
@@ -0,0 +1,998 @@
...
+	rmt_rx_channel_t *rx_channel = NULL;
+#if SOC_RMT_SUPPORT_DMA
+	bool with_dma;
+	size_t num_dma_nodes = 0;
+#endif
+	k_spinlock_key_t key;
+	esp_err_t ret;
+	int rc;
+
+	/* Check if priority is valid */
```
sylvioalves, line 267: I think you can check for NULL pointer before:
```c
if (!(config && ret_chan && config->resolution_hz && config->trans_queue_depth)) {
 		LOG_ERR("Invalid argument");
 		return -EINVAL;
 	}
```

sylvioalves: Same thing on rmt_rx.c file.

drivers/misc/espressif_rmt/rmt_tx.c
```diff
@@ -0,0 +1,1300 @@
...
+	}
+	for (size_t i = 0; i < config->array_size; i++) {
+		synchro->tx_channel_array[i] = config->tx_channel_array[i];
+	}
+	synchro->array_size = config->array_size;
+
+	/* acquire group handle, increase reference count */
+	rmt_group_t *group = rmt_acquire_group_handle(config->tx_channel_array[0]->group->group_id);
+	/* sanity check */
+	__ASSERT_NO_MSG(group);
```
sylvioalves, line 511: Perhaps in this case it should return -ENOMEM instead of asserting? What you think?
```c
	if (!group) {
		LOG_ERR("Unable to acquire RMT group");
		k_free(synchro);
		return -ENOMEM;
	}
```

drivers/misc/espressif_rmt/rmt_common.c
```diff
@@ -0,0 +1,380 @@
...
+} rmt_platform_t;
+
+static rmt_platform_t s_platform; /* singleton platform */
+
+rmt_group_t *rmt_acquire_group_handle(int group_id)
+{
+	unsigned int key;
+	bool new_group = false;
+	rmt_group_t *group = NULL;
+
```
sylvioalves, line 31: Please add null check as well:
```c
	if (group_id < 0 || group_id >= RMT_LL_GET(INST_NUM)) {
		return NULL;
	}
```

drivers/misc/espressif_rmt/rmt_encoder_copy.c
```diff
@@ -0,0 +1,145 @@
...
+static size_t IRAM_ATTR rmt_encode_copy(rmt_encoder_t *encoder, rmt_channel_handle_t channel,
+	const void *input_symbols, size_t data_size, rmt_encode_state_t *ret_state)
+{
+	rmt_copy_encoder_t *copy_encoder = __containerof(encoder, rmt_copy_encoder_t, base);
+	rmt_tx_channel_t *tx_chan = __containerof(channel, rmt_tx_channel_t, base);
+	rmt_symbol_word_t *symbols = (rmt_symbol_word_t *)input_symbols;
+	rmt_encode_state_t state = RMT_ENCODING_RESET;
+	size_t symbol_index = copy_encoder->last_symbol_index;
+	/* how many symbols will be copied by the encoder */
+	size_t mem_want = (data_size / 4 - symbol_index);
```
sylvioalves, line 40: would you replace the magic 4 to ` sizeof(rmt_symbol_word_t)` instead?

drivers/misc/espressif_rmt/rmt_encoder_simple.c
```diff
@@ -0,0 +1,253 @@
...
+
+	size_t min_chunk_size = config->min_chunk_size;
+
+	if (min_chunk_size == 0) {
+		min_chunk_size = 64;
+	}
+	encoder->ovf_buf = rmt_alloc_encoder_mem(min_chunk_size * sizeof(rmt_symbol_word_t));
+	if (!encoder->ovf_buf) {
+		LOG_ERR("Unable to allocate memory for overflow buffer");
+		if (encoder) {
```
sylvioalves, line 239: I think this check can go. If it got here, it is not NULL and `k_free` is ok.

samples/drivers/misc/espressif_rmt/simple_encoder/prj.conf
```diff
@@ -0,0 +1 @@
+# nothing here
```
kartben, line 1: this seems off, it should explicitly set `CONFIG_ESPRESSIF_RMT=y` as the subsystem shoudn't be `default y`

joelguittet:
@sylvioalves @kartben I have seen your comments, globally agree with you, just expect a few days delay on my side to fix them. But be sure, i'm still on it :-)

carlescufi:
Architecture WG meeting:

- RMT is a peripheral to generate infrared signals for remote control
- @joelguittet explains how the signal is generated: it's a multi-purpose signal generator without any byte code interpretation
- @bjarki-andreasen says there is a similar peripheral in Nordic (PWM) that can also do arbitrary binary pulse encoding
-

sylvioalves:
> Architecture WG meeting:
> 
> * RMT is a peripheral to generate infrared signals for remote control
> * @joelguittet explains how the signal is generated: it's a multi-purpose signal generator without any byte code interpretation
> * @bjarki-andreasen says there is a similar peripheral in Nordic (PWM) that can also do arbitrary binary pulse encoding

A little bit more info about RMT capabilities:
1. IR remote control — transmit/receive NEC, RC5, Sony SIRC, and other IR protocols
2. WS2812 / NeoPixel LEDs — generate the precise timing needed for addressable RGB LED strips (most popular use)
3. 1-Wire protocols — bit-banging DS18B20 temperature sensors, etc.
4. Stepper motor control — generating step pulses with precise timing
5. Servo PWM — though LEDC/MCPWM are more typical (already existing Espressif driver's in Zephyr)
6. Custom digital protocols — any application that needs accurate, hardware-timed pulse sequences without CPU intervention
7. Signal generation/capture — arbitrary waveforms, pulse measurement, frequency counting

bjarki-andreasen:
Architecture meeting notes to add to https://github.com/zephyrproject-rtos/zephyr/pull/101448#issuecomment-4489344390 (from memory right after meeting ended):
* @nashif Mentions public headers in misc contain HAL includes which is not allowed
* @sylvioalves Mentions the RMT has waveform capture capabilties. (note Nordic hardware does not have this capability)
* @bjarki-andreasen Proposes creating a driver class that groups similar hardware, though with vendor specific APIs, like we do for drivers/firmware and drivers/interrupt controllers to some extent.
* No hard objections to the above

sylvioalves:
> Architecture meeting notes to add to [#101448 (comment)](https://github.com/zephyrproject-rtos/zephyr/pull/101448#issuecomment-4489344390) (from memory right after meeting ended):
> 
> * @nashif Mentions public headers in misc contain HAL includes which is not allowed
> * @sylvioalves Mentions the RMT has waveform capture capabilties. (note Nordic hardware does not have this capability)
> * @bjarki-andreasen Proposes creating a driver class that groups similar hardware, though with vendor specific APIs, like we do for drivers/firmware and drivers/interrupt controllers to some extent.
> * No hard objections to the above

@carlescufi @bjarki-andreasen https://github.com/zephyrproject-rtos/zephyr/issues/109586

marekmatej:
I think the `lpcore` target tests can be filtered out.

PR #101456, opened 2025-12-23
shell: sensor: treat single axis channels as 3-axis

aperea-cc:
`sensor_three_axis_data` is not only used by 3-axis data, but also by the individual axis channels, according to [its documentation](https://docs.zephyrproject.org/latest/doxygen/html/structsensor__three__axis__data.html). This commit updates sensor_shell so that it treats the data of these channels correctly.


aperea-cc:
@MaureenHelm Could I please get an update on this (Or who should I ping instead)?  Thanks!

ubieda (changes requested):
Thanks for your contribution!

This has been brought up on #106850. As a result, documentation will be updated to treat single-channel axis data as `sensor_q31_data` https://github.com/zephyrproject-rtos/zephyr/issues/106850#issuecomment-4204721398

CC: @avisconti @teburd

avisconti:
> @MaureenHelm @teburd @avisconti Could this change please get reviewed? Thanks! :)

As mentioned by @ubieda it has been decided to change documentation instead. This has been proposed and merged in #107036

PR #101457, opened 2025-12-23
scripts: west_commands: build: support --cmake-opt

thorsten-klein:
## Why?
When using west aliases, it is helpful for users to give cmake arguments, e.g. to make building specific variants as easy as possible. Currently `--` must be used, e.g.
```
[alias]
build-llvm = build -- -DZEPHYR_TOOLCHAIN_VARIANT=llvm
build-codechecker = build -- -DZEPHYR_SCA_VARIANT=codechecker
build-llvm-codechecker = build -- -DZEPHYR_TOOLCHAIN_VARIANT=llvm -DZEPHYR_SCA_VARIANT=codechecker
```
With this approach it is not possible anymore to specify west arguments when the alias is used (e.g. board `-b native_sim`), as the argument comes behind `--` and therefore it is handled as cmake argument.

## Solution
Support for `--cmake-opt` argument is added which allows specifiying cmake options via `west build --cmake-opt=<arg>`, in order to avoid `--` in alias commands. This argument can be specified multiple times, e.g.
```
[alias]
build-llvm = build --cmake-opt=-DZEPHYR_TOOLCHAIN_VARIANT=llvm
build-codechecker = build --cmake-opt=-DZEPHYR_SCA_VARIANT=codechecker
build-llvm-codechecker = build --cmake-opt=-DZEPHYR_TOOLCHAIN_VARIANT=llvm --cmake-opt=-DZEPHYR_SCA_VARIANT=codechecker

# or even easier: build-llvm can be reused
build-llvm-codechecker = build-llvm --cmake-opt=-DZEPHYR_SCA_VARIANT=codechecker
```

### DIFF
--- a/scripts/west_commands/build.py
+++ b/scripts/west_commands/build.py
@@ -263,7 +263,7 @@ class Build(Forceable):
             else:
                 self.die("test item path does not exist")
 
-        self._run_cmake(board, origin, self.args.cmake_opts)
+        self._run_cmake(board, origin)
         if args.cmake_only:
             return
 
@@ -625,7 +625,7 @@ class Build(Forceable):
                 self.source_dir = self._find_source_dir()
                 self._sanity_check_source_dir()
 
-    def _run_cmake(self, board, origin, cmake_opts):
+    def _run_cmake(self, board, origin):
         if board is None and config_getboolean('board_warn', True):
             self.wrn('This looks like a fresh build and BOARD is unknown;',
                     "so it probably won't work. To fix, use",
--- a/scripts/west_commands/build.py
+++ b/scripts/west_commands/build.py
@@ -30,7 +30,7 @@ BUILD_USAGE = '''\
 west build [-h] [-b BOARD[@REV]]] [-d BUILD_DIR]
            [-S SNIPPET] [--shield SHIELD]
            [-t TARGET] [-p {auto, always, never}] [-c] [--cmake-only]
-           [-n] [-o BUILD_OPT] [-f]
+           [--cmake-opt CMAKE_OPT] [-n] [-o BUILD_OPT] [-f]
            [--sysbuild | --no-sysbuild] [--domain DOMAIN]
            [--extra-conf FILE.conf]
            [--extra-dtc-overlay FILE.overlay]
@@ -120,6 +120,13 @@ class Build(Forceable):
         group = parser.add_argument_group('cmake and build tool')
         group.add_argument('-c', '--cmake', action='store_true',
                            help='force a cmake run')
+        group.add_argument('--cmake-opt', action='append',
+                           dest="cmake_opts", default=[],
+                           help='''same as using '-- cmake_opt' but avoid the
+                           end-of-options marker '--' (e.g. in alias commands);
+                           those options are passed to cmake first, so they can
+                           be overridden via '-- cmake_opt';
+                           may be given multiple times.''')
         group.add_argument('--cmake-only', action='store_true',
                            help="just run cmake; don't build (implies -c)")
         group.add_argument('--domain', action='append',
@@ -294,7 +301,7 @@ class Build(Forceable):
 
     def _parse_remainder(self, remainder):
         self.args.source_dir = None
-        self.args.cmake_opts = None
+        self.args.cmake_opts = getattr(self.args, 'cmake_opts', [])
 
         try:
             # Only one source_dir is allowed, as the first positional arg
@@ -306,7 +313,7 @@ class Build(Forceable):
             if remainder[0] == _ARG_SEPARATOR:
                 remainder = remainder[1:]
             if remainder:
-                self.args.cmake_opts = remainder
+                self.args.cmake_opts.extend(remainder)
         except IndexError:
             pass
 
@@ -408,10 +415,7 @@ class Build(Forceable):
                         required_snippets.extend(arg_list)
                         continue
 
-                    if self.args.cmake_opts:
-                        self.args.cmake_opts.extend(args)
-                    else:
-                        self.args.cmake_opts = args
+                    self.args.cmake_opts.extend(args)
 
             self.args.sysbuild = sysbuild
 
@@ -432,10 +436,7 @@ class Build(Forceable):
             # Build the final argument list
             args_expanded = ["-D{}".format(a.replace('"', '')) for a in args]
 
-            if self.args.cmake_opts:
-                self.args.cmake_opts.extend(args_expanded)
-            else:
-                self.args.cmake_opts = args_expanded
+            self.args.cmake_opts.extend(args_expanded)
 
         return found_test_metadata
 
--- a/scripts/west_commands/tests/test_build.py
+++ b/scripts/west_commands/tests/test_build.py
@@ -13,11 +13,11 @@ import pytest
 
 TEST_CASES = [
     {'r': [],
-     's': None, 'c': None},
+     's': None, 'c': []},
     {'r': ['source_dir'],
-     's': 'source_dir', 'c': None},
+     's': 'source_dir', 'c': []},
     {'r': ['source_dir', '--'],
-     's': 'source_dir', 'c': None},
+     's': 'source_dir', 'c': []},
     {'r': ['source_dir', '--', 'cmake_opt'],
      's': 'source_dir', 'c': ['cmake_opt']},
     {'r': ['source_dir', '--', 'cmake_opt', 'cmake_opt2'],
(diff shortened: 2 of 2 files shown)


scripts/west_commands/build.py
```diff
@@ -119,6 +119,10 @@ def do_add_parser(self, parser_adder):
         group = parser.add_argument_group('cmake and build tool')
         group.add_argument('-c', '--cmake', action='store_true',
                            help='force a cmake run')
+        group.add_argument('--cmake-opt', action='append',
+                           help='''same as '-- cmake_opt', but it prevents
+                           using '--' (e.g. for usage in alias commands);
+                           may be given multiple times.''')
```
pdgendt, line 129: Don't use both `cmake_opt` and `cmake_opts`, it is confusing. Set a default instead and extend in `_parse_remainder`

```suggestion
        group.add_argument('--cmake-opt', action='append',
                           dest="cmake_opts", default=[],
                           help='''same as '-- cmake_opt', but it prevents
                           using '--' (e.g. for usage in alias commands);
                           may be given multiple times.''')
```

(There's a debug print for `cmake_opts`)

pdgendt (changes requested):

scripts/west_commands/build.py
```diff
@@ -293,7 +298,7 @@ def _find_board(self):
 
     def _parse_remainder(self, remainder):
         self.args.source_dir = None
-        self.args.cmake_opts = None
+        self.args.cmake_opts = getattr(self.args, 'cmake_opts', None)
```
pdgendt:
```suggestion
        self.args.cmake_opts = getattr(self.args, 'cmake_opts', [])
```

scripts/west_commands/build.py
```diff
@@ -119,6 +119,11 @@ def do_add_parser(self, parser_adder):
         group = parser.add_argument_group('cmake and build tool')
         group.add_argument('-c', '--cmake', action='store_true',
                            help='force a cmake run')
+        group.add_argument('--cmake-opt', action='append',
+                           dest="cmake_opts", default=[],
```
pdgendt, line 124:
```suggestion
                           dest="cmake_opts",
```

scripts/west_commands/build.py
```diff
@@ -305,7 +310,8 @@ def _parse_remainder(self, remainder):
             if remainder[0] == _ARG_SEPARATOR:
                 remainder = remainder[1:]
             if remainder:
-                self.args.cmake_opts = remainder
+                self.args.cmake_opts = self.args.cmake_opts or []
```
pdgendt:
```suggestion
```

pdgendt (dismissed):

pdgendt:
> Or, you could just use shell functions like this one:
> 
> ```shell
> wb_llvm() { west build "$@" -- -DZEPHYR_TOOLCHAIN_VARIANT=llvm; }
> wb_llvm -b native_sim ...
> ```
> 
> This is exactly the type of problems shells are designed to solve. They can solve this one simply and can do a lot more.
> 
> So this looks like "feature creep" and unnecessary maintenance and complexity to me.

I have to disagree here, adding shell aliases/functions scatters the commands to different places in the OS/workspace. I think this is a useful feature, and adds little maintenance overhead.

scripts/west_commands/build.py
```diff
@@ -407,10 +412,7 @@ def _parse_test_item(self, test_item, board):
                         required_snippets.extend(arg_list)
                         continue
 
-                    if self.args.cmake_opts:
-                        self.args.cmake_opts.extend(args)
-                    else:
-                        self.args.cmake_opts = args
+                    self.args.cmake_opts.extend(args)
```
pdgendt, line 418: Can't `self.args.cmake_opts` still be `None` here? (same for change further down the function)

thorsten-klein: Yes, you are right after I have removed the default `[]` again.
I have now set default to `[]` again and adapted 4 existing test cases to use the new default (`[]` instead of `None`).

pdgendt (dismissed):

scripts/west_commands/build.py
```diff
@@ -119,6 +119,11 @@ def do_add_parser(self, parser_adder):
         group = parser.add_argument_group('cmake and build tool')
         group.add_argument('-c', '--cmake', action='store_true',
                            help='force a cmake run')
+        group.add_argument('--cmake-opt', action='append',
+                           dest="cmake_opts", default=[],
+                           help='''same as '-- cmake_opt', but it prevents
+                           using '--' (e.g. for usage in alias commands);
+                           may be given multiple times.''')
```
marc-hb, line 129: What is the order when you use both? I think it should be explicit here.

pdgendt (approved):

tejlmand (dismissed):
PR is not needed.

> Currently -- must be used,

Help text for `west build` is imprecise on this.

First positional argument to `west build` is treated as the source dir, remaining positional arguments are passed to CMake.

See also: 
https://github.com/zephyrproject-rtos/zephyr/blob/a600317285ff82884055c2135a2ec39af7d42191/scripts/west_commands/build.py#L301-L309

So as long as you use `west build -b <board> <source>` then you don't need `--` and can just do: `west build -b <board> <source> -DFOO=bar`.
Only when omitting source dir, you need `--`, such as in this case:
```
$ cd <app-folder>
$ west build -b <board> -- -DFOO=bar 
```

to avoid using `--` in this case, one can just use a `.`, like this:
```
$ cd <app-folder>
$ west build -b <board> . -DFOO=bar 
```

Other places where source-dir can be omitted is when running an incremental, non-pristine, build, but that to can be handled easily if goal is to allow passing arguments to CMake without the use of `--`.

also see: 
https://github.com/zephyrproject-rtos/west/blob/11302e1ee5af96c9a0cf8c7436992936b1bbd848/src/west/commands.py#L196
https://github.com/zephyrproject-rtos/west/blob/11302e1ee5af96c9a0cf8c7436992936b1bbd848/src/west/commands.py#L214

scripts/west_commands/build.py
```diff
@@ -120,6 +120,13 @@ def do_add_parser(self, parser_adder):
         group = parser.add_argument_group('cmake and build tool')
         group.add_argument('-c', '--cmake', action='store_true',
                            help='force a cmake run')
+        group.add_argument('--cmake-opt', action='append',
+                           dest="cmake_opts", default=[],
+                           help='''same as using '-- cmake_opt' but avoid the
+                           end-of-options marker '--' (e.g. in alias commands);
+                           those options are passed to cmake first, so they can
+                           be overridden via '-- cmake_opt';
```
tejlmand, line 128: > so they can be overridden via '-- cmake_opt';

I guess they can also be overwritten with a later `--cmake-opt` argument, correct ?

Thus I think it's better to just described that cmake options later on command line takes precedence.

Then it's for the user to decide whether to use `-- <arg>` or `--cmake-opt=<arg>` .

tejlmand (approved):
I share the same concerns as @marc-hb here: https://github.com/zephyrproject-rtos/zephyr/pull/101457#issuecomment-3712611876

I do see the usefulness and reason for wanting this in context of `west alias`, but  i'm concerned we can end up having to not allow positional arguments in west extension commands, simply because positional arguments in alias command are never going to be compatible with alias.

I'm not approving this PR because of the wish for having `--cmake-opt` as keyword argument addition to positional argument for the sake of alias.

I'm approving the PR because having `--cmake-opt` in addition aligns it with the build argument `--build-opt`.
Allowing both `--cmake-opt` and `--build-opt` gives a more consistent interface to `west build`. 

Implementation itself looks ok, just a small comment regarding help text.

scripts/west_commands/build.py
marc-hb: `-- cmake_opt` is visually too similar to `--cmake-opt`, that's a bit confusing. Just say "... options after --" if needed.

Also, don't say "overriden" because this is assuming last always wins with CMake.  I bet it's often true but I also bet there are also some CMake options that cannot be repeated more than once: no "overriding" possible. Only describe the order in which the options are passed to CMake = what `west build` does and that's it. I mean don't describe what CMake does, that's the job of CMake documentation. It may even depend on the CMake version.

marc-hb: @henrikbrixandersen just merged this. I'll submit a small fixup and we can discuss there.

henrikbrixandersen: Sorry, merged just before you commented.

marc-hb: No worries, it wasn't blocking at all and I didn't "take the lock" :-)

marc-hb: >  I'll submit a small fixup and we can discuss there.

- #102419

PR #101458, opened 2025-12-23
soc: nxp: ke1xz: Enhance power management with STOP modes

Holt-Sun:
Implement proper STOP mode support with partial stop options, add power state timing characteristics, and enable LPTMR-based wake-up for KE1xZ SoC family.

- Fix STOP mode entry with correct SLEEPDEEP and PSTOPO configuration
- Add min-residency-us/exit-latency-us to all power states
- Use LPTMR as cortex-m-idle-timer for low-power operation
- Add power_mgmt_soc test support for FRDM-KE17Z

Depends on following PRs:
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/101392
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/101573
- [x] https://github.com/zephyrproject-rtos/zephyr/pull/101705


boards/nxp/frdm_ke17z/frdm_ke17z.dts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 NXP
+ * Copyright 2024, 2025 NXP
```
ZhaoxiangJin: 2024-2025, also other files.

tests/subsys/pm/power_mgmt_soc/boards/frdm_ke17z.conf
```diff
@@ -0,0 +1,8 @@
+# Copyright 2025 NXP
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Keep LPTMR as wake/idle timer via counter API
+CONFIG_COUNTER=y
+CONFIG_COUNTER_MCUX_LPTMR=y
```
ZhaoxiangJin, line 7: `CONFIG_COUNTER_MCUX_LPTMR `not needed manually set to y here guess.

soc/nxp/kinetis/ke1xz/Kconfig.defconfig
```diff
@@ -1,16 +1,13 @@
...
-# Copyright 2024 NXP
+# Copyright 2024, 2025 NXP
 # SPDX-License-Identifier: Apache-2.0
 
 if SOC_SERIES_KE1XZ
 
-config MCUX_LPTMR_TIMER
-	default y if PM
-
 config CORTEX_M_SYSTICK
```
ZhaoxiangJin: Maybe we can change it to `configdefault CORTEX_M_SYSTICK` at the same time?

dts/arm/nxp/nxp_ke1xz.dtsi
```diff
@@ -33,24 +33,33 @@
 			idle: idle {
 				compatible = "zephyr,power-state";
 				power-state-name = "runtime-idle";
+		        	min-residency-us = <10>;
```
ZhaoxiangJin: What is the clock frequency of the LPTMR? I think the `min-residency-us` value needs careful consideration, as 10us may not be sufficient for even one hardware cycle of the LPTMR.

Holt-Sun:
@Albort12138 @ZhaoxiangJin , please help review.

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -22,22 +23,48 @@ __ramfunc static void wait_for_flash_prefetch_and_idle(void)
 
 	k_cpu_idle();
 }
+#endif /* CONFIG_XIP */
 
 void pm_state_set(enum pm_state state, uint8_t substate_id)
 {
+	__enable_irq();
+
```
bjarki-andreasen: `__enable_irq()`? usually IRQs are disabled before sleep to ensure any pending IRQ is left pending until the CPU enters sleep. Otherwise a race may occur if an interrupt which is expected to wake up the CPU is handled before the CPU enters sleep, thus the CPU enters sleep to never wake up again.

bjarki-andreasen (changes requested):
Looks good, but I'm suspicious of the `__enable_irq()`

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -22,22 +23,48 @@ __ramfunc static void wait_for_flash_prefetch_and_idle(void)
 
 	k_cpu_idle();
 }
+#endif /* CONFIG_XIP */
 
 void pm_state_set(enum pm_state state, uint8_t substate_id)
 {
+	__enable_irq();
```
Albort12138:
```suggestion
	__disable_irq();
```

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -22,22 +23,48 @@ __ramfunc static void wait_for_flash_prefetch_and_idle(void)
...
+		 * The devicetree substate-id maps to the PSTOPO field (partial stop option).
+		 * Valid values are SoC-specific; for KE1xZ this is typically 0..2.
+		 */
+		if (substate_id > 2U) {
+			LOG_WRN("Unsupported substate-id %u, using 0", substate_id);
+			substate_id = 0U;
+		}
+
+		/* Ensure STOPM is set to normal STOP (not VLPS/VLLS families), if present. */
+		SMC->PMCTRL = (SMC->PMCTRL & ~SMC_PMCTRL_STOPM_MASK) | SMC_PMCTRL_STOPM(0U);
```
Albort12138:
```suggestion
		SMC->PMCTRL &= ~SMC_PMCTRL_STOPM_MASK;
		
```

dts/arm/nxp/nxp_ke1xz.dtsi
Albort12138:
```suggestion
			cpu-power-states = <&idle &pstop2 &pstop1 &stop>;
```
Based on this order, to find the appropriate mode more efficiently

dts/arm/nxp/nxp_ke1xz.dtsi
```diff
@@ -33,24 +33,33 @@
...
 				substate-id = <0>;
+				min-residency-us = <140>;
+				exit-latency-us = <14>;
 			};
 
 			pstop1: pstop1 {
 				compatible = "zephyr,power-state";
 				power-state-name = "suspend-to-idle";
 				substate-id = <1>;
+				min-residency-us = <60>;
```
Albort12138: Where residency come from?

soc/nxp/kinetis/ke1xz/power.c
ZhaoxiangJin: If IRQ is not enabled, the wake-up source may not be able to wake up the MCU properly. (Of course, some MCUs can be woken up as long as there is a pending interrupt.)

dts/arm/nxp/nxp_ke1xz.dtsi
```diff
@@ -33,24 +33,33 @@
 			idle: idle {
 				compatible = "zephyr,power-state";
 				power-state-name = "runtime-idle";
+				min-residency-us = <10>;
+				/* 0.7 µs rounded */
+				exit-latency-us = <1>;
```
ZhaoxiangJin, line 38: The exit-latency-us value should be incremented across different power modes.

tests/subsys/pm/power_mgmt_soc/boards/frdm_ke17z.conf
Holt-Sun: Yes, CONFIG_COUNTER_MCUX_LPTMR is the dependency of COUNTER_MCUX_LPTMR_ALARM.

I will keep it for the configurations should be explicit about dependencies, anyone reading this file immediately knows:
- This test requires COUNTER subsystem
- This test specifically needs LPTMR counter driver
- This test uses LPTMR alarm functionality

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: It should be __disable_irq().  __disable_irq() won't stop wake-up source wake up the CPU. It stops irq handler with interrupts happen waking up the CPU.
When entering power_state_set, the irq is always disabled, so remove it.

dts/arm/nxp/nxp_ke1xz.dtsi
Holt-Sun: This is got from RM. In the power_mgmt_soc, provide anther set working.

soc/nxp/kinetis/ke1xz/soc.c
```diff
@@ -186,6 +187,9 @@ void soc_early_init_hook(void)
 {
 	/* Initialize system clocks and PLL */
 	clk_init();
+#ifdef CONFIG_PM
+	SMC_SetPowerModeProtection(SMC, kSMC_AllowPowerModeAll);
```
ZhaoxiangJin, line 191: I'd like to know why allow enter low power mode here?

dts/arm/nxp/nxp_ke1xz.dtsi
Holt-Sun: From the RM.
Min Residency = 10×exit-latency which is a rough estimate. Developers/customers need to adjust it according to practice.

Holt-Sun:
> To be honest, I'm not sure if your MCU has actually gone to low power mode. Look [here](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/pm/pm.c#L207): LPTMR can only be correctly configured as the wake-up source when `exit_latency_ticks` is greater than 0. And looking at the power mode nodes in the devicetree, their `min-residency-us` values ​​are very low, possibly less than one kernel tick under the default transition strategy.

Your concern is valid. Previous systick pending interrupt forbid the CPU to stop down. Already update the code clear pending interrupt before entering stop. Now it is ok.
Here is more detailed log:
```
I: About to enter deep Sleep
PM >
LPM_ENTRY: max_lpm_time_us=1099600 ticks=4398
From isr wrapper.
In pm_system_resume: calling sys_clock_idle_exit.
LPM_EXIT: idle_timer_pre=0 idle_timer_post=4432
LPM_EXIT: pre=0 post=4432 pending=1 wrap=1 diff=8831 us=2207750
LPM_EXIT: systick_us=13205 idle_timer_us=2207750
LPM_EXIT: missed_cycles=105338160 dticks=22567
ISR Wrapper: irq_number=29
In mcux_lptmr_isr.
In pm_system_resume.
I: Wake from Deep Sleep
PM <
I: PM sleep residency 2.261 seconds
I: PM dummy single-thread completed
```

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -12,32 +12,86 @@
...
+	__ISB();
+	__WFI();
 }
+#endif /* CONFIG_XIP */
 
 void pm_state_set(enum pm_state state, uint8_t substate_id)
 {
+	/* Set PRIMASK */
+	__disable_irq();
+	/* Set BASEPRI to 0 */
```
ZhaoxiangJin: If I remember correctly, the KE1xZ ([CM0P](https://github.com/zephyrproject-rtos/zephyr/blob/main/soc/nxp/kinetis/ke1xz/Kconfig#L9)) has no BASEPRI ?

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -12,32 +12,86 @@
...
-			k_cpu_idle();
+#if defined(CONFIG_CORTEX_M_SYSTICK)
+		/* If SysTick stays enabled, it can pend and immediately wake WFI. */
+		uint32_t syst_ctrl_saved = SysTick->CTRL;
+
+		/* Disable SysTick interrupt during STOP entry */
+		SysTick->CTRL = syst_ctrl_saved & ~SysTick_CTRL_TICKINT_Msk;
+
+		/* Clear pending core exceptions that can cause immediate WFI return */
+		SCB->ICSR = SCB_ICSR_PENDSTCLR_Msk | SCB_ICSR_PENDSVCLR_Msk;
```
ZhaoxiangJin: I'd like to know why PendSV also need to be cleared?
Does the systick pending state here prevent the system from entering low power mode? Currently, the systick driver allows disabling systick by enabling [CORTEX_M_SYSTICK_RESET_BY_LPM](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.cortex_m_systick#L99). We might be able to select `CORTEX_M_SYSTICK_RESET_BY_LPM `and enhance the systick logic (by adding operations to clear the systick pending state)?

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: Good catch, fixed.

soc/nxp/kinetis/ke1xz/soc.c
Holt-Sun: If the CONFIG_PM is enabled, it is better to enable all power mode. It is one time job. soc.c is better than power.c.

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: 1. Clear PendSV is overhead, removed.
2. The systick pending state does prevent the system enter low power mode for KE1x device with no wake up management.
3. [CORTEX_M_SYSTICK_RESET_BY_LPM](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.cortex_m_systick#L99) is for the case that systick is reset during the some low power modes in some SOC, in KE1X, the systick continuously run in stop mode which is not suitable with [CORTEX_M_SYSTICK_RESET_BY_LPM](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.cortex_m_systick#L99). Introduce [CORTEX_M_SYSTICK_STOP_IN_LPM](https://github.com/nxp-upstream/zephyr/blob/c2972636c4daa1c040c9fbd7e8045105ad84c262/drivers/timer/Kconfig.cortex_m_systick#L115) to support such case: intentionally stop systick during low power enter process.

Code has been updated, please review again.

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: Here is the detailed run log with updated code:
```
pm_state_set: Entering STOP mode (substate_id=0)

[14:34:36.555]_isr_wrapper: Enter pm_system_resume in _kernel.idle.
pm_system_resume: entry
pm_system_resume: post_ops_required=1
pm_system_resume: doing post ops.
pm_state_exit_post_ops: entered.
pm_state_exit_post_ops: exited power state 2
pm_state_exit_post_ops: exiting
pm_system_resume: calling sys_clock_idle_exit.
sys_clock_idle_exit: entered
sys_clock_idle_exit: calling z_cms_lptim_hook_on_lpm_exit
z_cms_lptim_hook_on_lpm_exit: entered
z_cms_lptim_hook_on_lpm_exit: exiting
sys_clock_idle_exit: systick_us=664400 idle_timer_us=2249750
sys_clock_idle_exit: missed_cycles=76096800 dticks=23410
sys_clock_idle_exit: exiting
_isr_wrapper: Back from pm_system_resume in _kernel.idle.
_isr_wrapper: calling irq_number=29
Run mcux_lptmr_isr.
pm_state_set: Woke from STOP mode
pm_state_set: exiting
pm_system_resume: entry
pm_system_resume: post_ops_required=0
I: Wake from Deep Sleep
```

tests/subsys/pm/power_mgmt_soc/boards/frdm_ke17z.overlay
```diff
@@ -0,0 +1,77 @@
...
+	exit-latency-us = <300>;
+};
+
+&pstop2 {
+	exit-latency-us = <200>;
+};
+
+/* Disable other potential wake sources/peripherals */
+&adc0 {
+	status = "disabled";
```
ZhaoxiangJin: This is not a recommend way, you should use [power-domain-soc-state-change](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/bindings/power-domain/power-domain-soc-state-change.yaml) to turn-on and turn-off peripherals. for example: [MCXW23](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/nxp/nxp_mcxw23x_common.dtsi#L73) and [RW61x ](https://github.com/zephyrproject-rtos/zephyr/blob/main/dts/arm/nxp/nxp_rw6xx_common.dtsi#L87)

Holt-Sun: Updated as suggested.

Albort12138: I have a few points I don't understand:

Is it because we want to obtain the lowest power consumption data that we need to turn off unused peripherals? Or would these peripherals unexpectedly wake up the MCU if not turned off?

If the goal is to obtain the lowest power consumption data, then setting them to disable in the overlay seems reasonable to me. Why introduce a power domain that doesn't actually exist in the MCU, which is only used to manually suspend/turn off devices when CONFIG_PM_DEVICE_SYSTEM_MANAGED is set to 1, instead of simply not initializing them from the beginning? This is just a test case after all.

If it's because not turning off these peripherals would cause unexpected wake-ups, then this is a software bug, not something that should merely be disabled.

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: Will use [CORTEX_M_SYSTICK_RESET_BY_LPM](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/Kconfig.cortex_m_systick#L99) but extend its functionality semantically: https://github.com/zephyrproject-rtos/zephyr/pull/101705

tests/subsys/pm/power_mgmt_soc/boards/frdm_ke17z.overlay
Holt-Sun: Good question. The original goal is to obtain lower power consumption but it is not necessary. 
There is no software bug needed by turning off them.
Remove such code.

Holt-Sun: This is the implementation for system power management mechanism, don't mix device power management requirement which is in progress in another branch.

bjarki-andreasen (dismissed):

soc/nxp/kinetis/ke1xz/Kconfig.defconfig
```diff
@@ -1,17 +1,14 @@
...
 
 # Copyright (c) 2019-2021 Vestas Wind Systems A/S
-# Copyright 2024 NXP
+# Copyright 2024-2025 NXP
 # SPDX-License-Identifier: Apache-2.0
 
 if SOC_SERIES_KE1XZ
 
-config MCUX_LPTMR_TIMER
+configdefault CORTEX_M_SYSTICK
```
ZhaoxiangJin: this code seems not necessarily?

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -12,35 +12,63 @@
...
+		if (substate_id > 2U) {
+			LOG_WRN("Unsupported substate-id %u, using 0", substate_id);
+			substate_id = 0U;
 		}
+		/* Ensure STOPM is set to normal STOP (not VLPS/VLLS families), if present. */
+		SMC->PMCTRL &= ~SMC_PMCTRL_STOPM_MASK;
+
+#if (defined(FSL_FEATURE_SMC_HAS_PSTOPO) && FSL_FEATURE_SMC_HAS_PSTOPO)
+		SMC->STOPCTRL = (SMC->STOPCTRL & ~SMC_STOPCTRL_PSTOPO_MASK) |
+				SMC_STOPCTRL_PSTOPO(substate_id);
```
ZhaoxiangJin, line 55: should be aligned with the indentation of the previous row.

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -12,35 +12,63 @@
...
 {
 	switch (state) {
 	case PM_STATE_RUNTIME_IDLE:
-		k_cpu_idle();
+		/* Normal WAIT: WFI with SLEEPDEEP cleared. */
+		SCB->SCR &= ~SCB_SCR_SLEEPDEEP_Msk;
+		__DSB();
+		__ISB();
+		__WFI();
+		irq_unlock(0);
```
ZhaoxiangJin, line 43: move this code to `pm_state_exit_post_ops`. After wakeup, the `pm_state_exit_post_ops` will execute fist, we should unlock IRQ as soon as possible.

soc/nxp/kinetis/ke1xz/power.c
```diff
@@ -12,35 +12,63 @@
...
+		__ISB();
+#ifdef CONFIG_XIP
+		wait_for_flash_prefetch_and_wfi();
+#else
+		__WFI();
+#endif
 		if (SMC->PMCTRL & SMC_PMCTRL_STOPA_MASK) {
 			LOG_DBG("partial stop aborted");
 		}
+		irq_unlock(0);
```
ZhaoxiangJin, line 71: same as above

ZhaoxiangJin (commented):
we should split the PR into different commits

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: The irq shall be unlock here, this is the process sequence which is expected.
```
(scenario 1)
pm_state_set: entered
pm_state_set: enable irq
pm_state_set: No pending core exceptions before STOP/WFI
pm_state_set: Entering STOP mode (substate_id=0)

[14:34:36.555]_isr_wrapper: Enter pm_system_resume in _kernel.idle.
pm_system_resume: entry
pm_system_resume: post_ops_required=1
pm_system_resume: doing post ops.
pm_state_exit_post_ops: entered.
pm_state_exit_post_ops: exited power state 2
pm_state_exit_post_ops: exiting
pm_system_resume: calling sys_clock_idle_exit.
sys_clock_idle_exit: entered
sys_clock_idle_exit: calling z_cms_lptim_hook_on_lpm_exit
z_cms_lptim_hook_on_lpm_exit: entered
z_cms_lptim_hook_on_lpm_exit: directly reading idle timer value due to pending interrupt
z_cms_lptim_hook_on_lpm_exit: idle_timer_pre=0 idle_timer_post=4600 pending=1 wrap=1 diff=8999 us=2249750
z_cms_lptim_hook_on_lpm_exit: exiting
sys_clock_idle_exit: systick_us=664400 idle_timer_us=2249750
sys_clock_idle_exit: missed_cycles=76096800 dticks=23410
sys_clock_idle_exit: exiting
_isr_wrapper: Back from pm_system_resume in _kernel.idle.
_isr_wrapper: calling irq_number=29
Run mcux_lptmr_isr.
pm_state_set: Woke from STOP mode
pm_state_set: exiting
pm_system_resume: entry
pm_system_resume: post_ops_required=0
I: Wake from Deep Sleep
```

If the irq_locked is removed here and added in pm_state_exit_post_ops, then the process sequence is:
```
(scenario 2)
pm_state_set: entered
pm_state_set: disable irq
pm_state_set: No pending core exceptions before STOP/WFI
pm_state_set: Entering STOP mode (substate_id=0)

[14:31:20.872]Pending core exception: ISR pending (VECTPENDING=45)
pm_state_set: Woke from STOP mode
pm_state_set: exiting
pm_system_resume: entry
pm_system_resume: post_ops_required=1
pm_system_resume: doing post ops.
pm_state_exit_post_ops: entered.
pm_state_exit_post_ops: exited power state 2
pm_state_exit_post_ops: reenable irq
_isr_wrapper: Enter pm_system_resume in _kernel.idle.
pm_system_resume: entry
pm_system_resume: post_ops_required=0
_isr_wrapper: Back from pm_system_resume in _kernel.idle.
_isr_wrapper: calling irq_number=29
Run mcux_lptmr_isr.
pm_state_exit_post_ops: exiting
pm_system_resume: calling sys_clock_idle_exit.
sys_clock_idle_exit: entered
sys_clock_idle_exit: calling z_cms_lptim_hook_on_lpm_exit
z_cms_lptim_hook_on_lpm_exit: entered
z_cms_lptim_hook_on_lpm_exit: directly using idle_timer_alarm_fired_ticks since interrupt isr has run
z_cms_lptim_hook_on_lpm_exit: idle_timer_pre=0 idle_timer_post=4612 pending=0 wrap=0 diff=4612 us=1153000
z_cms_lptim_hook_on_lpm_exit: exiting
sys_clock_idle_exit: systick_us=362166 idle_timer_us=1153000
sys_clock_idle_exit: missed_cycles=37960032 dticks=12461
sys_clock_idle_exit: exiting
I: Wake from Deep Sleep
```
There is recursively entering pm_system_resume which is not prefered.
(That is also the reason I added https://github.com/zephyrproject-rtos/zephyr/pull/101570)

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: See above.

soc/nxp/kinetis/ke1xz/power.c
ZhaoxiangJin: > ```
> z_cms_lptim_hook_on_lpm_exit: idle_timer_pre=0 idle_timer_post=4600 pending=1 wrap=1 diff=8999 us=2249750
> z_cms_lptim_hook_on_lpm_exit: exiting
> sys_clock_idle_exit: systick_us=664400 idle_timer_us=2249750
> ```
I have a question to this log, the sleep time is 2249750us in STOP mode (SoC idle mode)? In the device tree, the `min-residency-us` of idle mode is 10.

> ```
> pm_state_set: entered
> pm_state_set: disable irq
> pm_state_set: No pending core exceptions before STOP/WFI
> pm_state_set: Entering STOP mode (substate_id=0)
> [14:31:20.872]Pending core exception: ISR pending (VECTPENDING=45)
> pm_state_set: Woke from STOP mode
> ```

From the log, it appears there was a premature wakeup. The irq_unlock(0); occurred after WFI (Wait For Interrupt). I don't understand why this situation would happen. This is worth a deeper analysis, but for now, let's leave it as is.

Holt-Sun: `I have a question to this log, the sleep time is 2249750us in STOP mode (SoC idle mode)? In the device tree, the min-residency-us of idle mode is 10.` This is a good catch. The issue may not be in this PR, but in https://github.com/zephyrproject-rtos/zephyr/blob/cdb7debfb726569440fc22ce6673de2b86c728f7/drivers/timer/cortex_m_systick.c#L176-L208 : 
lin179 gets the idle_timer_post which is same as idle_timer_scheduled_sleep_ticks , line181 gets the top, actually, they are same for most cases(lptmr frequency is usually much lower than CPU frequency), in line 197-199, the idle_timer_int_pending is true(which may be an issue), then it is always wrap. so the calculated the sleep time is wrong, if we fix it, then the sleep time is normal:
`sys_clock_idle_exit: systick_us=0 idle_timer_us=1099500 systick_diff=4294967296 missed_cycles=52776000 dticks=11689`
I would think it is more like issue in [cortex_m_systick.c](https://github.com/zephyrproject-rtos/zephyr/blob/cdb7debfb726569440fc22ce6673de2b86c728f7/drivers/timer/cortex_m_systick.c).

If you see scenario 2, the idle time is expected which I get counter from [idle_timer_alarm_stub](https://github.com/zephyrproject-rtos/zephyr/blob/cdb7debfb726569440fc22ce6673de2b86c728f7/drivers/timer/cortex_m_systick.c#L126). I will try to fix this in PR https://github.com/zephyrproject-rtos/zephyr/pull/101705

`From the log, it appears there was a premature wakeup. The irq_unlock(0); occurred after WFI (Wait For Interrupt). I don't understand why this situation would happen. This is worth a deeper analysis, but for now, let's leave it as is.`
The irq_unlock(0) occurred after WFI, but not immediately, the LPTMR interrupt(no matter irq lock or not) wakes up the CPU when set alarm time hits, the PC will run irq_unlock(0) to open the irq, then PC will run into [_isr_wrapper](https://github.com/zephyrproject-rtos/zephyr/blob/cdb7debfb726569440fc22ce6673de2b86c728f7/arch/arm/core/cortex_m/isr_wrapper.c#L33), the time duration is idle_timer_pre + idle_timer_post = 4600. There is also a "[14:31:20.872]" to tip that there is time passed. So there is no prematrue wakeup here.

soc/nxp/kinetis/ke1xz/power.c
Holt-Sun: I used clang-format to format this file, it is this result.

soc/nxp/kinetis/ke1xz/Kconfig.defconfig
Holt-Sun: Removed.

soc/nxp/kinetis/ke1xz/power.c
ZhaoxiangJin: After look closer, the mcu core is cm0p, the `irq_unlock(0);` should be here.

ZhaoxiangJin (approved):

Holt-Sun:
@bjarki-andreasen , all dependencies and issues have been fixed, please review. Thank you.

bjarki-andreasen (approved):

PR #101459, opened 2025-12-23
shell: adc_shell: fix read values for 16-bit adcs

aperea-cc:
The ADC shell was showing incorrect values for ADCs with 16-bit resolution, as the 16th bit was being misinterpreted as the sign bit.

This PR aims to fix that by changing the type of the sample buffer to unsigned 16-bit values.

### DIFF
--- a/drivers/adc/adc_shell.c
+++ b/drivers/adc/adc_shell.c
@@ -379,7 +379,7 @@ static int cmd_adc_read(const struct shell *sh, size_t argc, char **argv)
 	uint8_t adc_channel_id = strtol(argv[1], NULL, 10);
 	/* -1 index of adc label name */
 	struct adc_hdl *adc = get_adc(argv[-1]);
-	int16_t m_sample_buffer[BUFFER_SIZE];
+	uint16_t m_sample_buffer[BUFFER_SIZE];
 	int retval;
 
 	if (!device_is_ready(adc->dev)) {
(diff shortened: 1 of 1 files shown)


drivers/adc/adc_shell.c
```diff
@@ -379,7 +379,7 @@ static int cmd_adc_read(const struct shell *sh, size_t argc, char **argv)
 	uint8_t adc_channel_id = strtol(argv[1], NULL, 10);
 	/* -1 index of adc label name */
 	struct adc_hdl *adc = get_adc(argv[-1]);
-	int16_t m_sample_buffer[BUFFER_SIZE];
+	uint16_t m_sample_buffer[BUFFER_SIZE];
```
ZhaoxiangJin, line 382: Thank for this patch.
But this fix introduces a new risk: ADC supports differential sampling (`channel_config.differential=1`), and differential results can inherently be negative values. Changing the buffer to `uint16_t` at this point would make it impossible to represent negative values.
I suggest keep `int16_t` here and change the shell_print to:
```
	if (retval >= 0) {
		if (adc->channel_config.differential) {
			shell_print(sh, "read: %d", (int)m_sample_buffer[0]);
		} else {
			shell_print(sh, "read: %u",
				   (unsigned int)(uint16_t)m_sample_buffer[0]);
		}
	} 
```

can you try this with this [PR](https://github.com/zephyrproject-rtos/zephyr/pull/103118)? This works fine on my local.

PR #101460, opened 2025-12-23
drivers: usb: udc: numaker: support m333x hsusbd

ccli8:
This adds support for Nuvoton NuMaker m333x HSUSBD and makes other fixes as below:

1. Add config for disabling USB on unplug for soc supporting this feature
2. Fix control out failure with FIFO flush timing in HSUSBD
3. Fix compile error on m2l31x
4. Fix submit condition to meet updated msc class driver which requests submit data to be multiple of sector size


fabiobaltieri (dismissed):

dts/bindings/usb/nuvoton,numaker-hsusbd.yaml
```diff
@@ -25,3 +25,9 @@ properties:
     required: true
     description: |
       Size of DMA buffer in bytes
+
+  allow-disable-usb-on-unplug:
+    type: boolean
+    description: |
+      Some soc series allow USB function to be disabled temporarily on USB unplug. It can enable
+      back on USB re-plug.
```
josuah: > Some soc series allow USB **function** to be disabled temporarily on USB unplug. It can enable back on USB re-plug.

If I remember right, an "USB function" is something well-defined in the USB  standard, so can be confusing (i.e. disable an USB class?).

What about "USB functionality" or "USB peripheral"?

dts/bindings/usb/nuvoton,numaker-usbd.yaml
```diff
@@ -31,3 +31,9 @@ properties:
     description: |
       Some soc series don't allow Isochronous IN/OUT endpoints to be assigned the same numbers,
       for example, 0x82 (for Isochronous IN) and 0x02 (for Isochronous OUT) are disallowed.
+
+  allow-disable-usb-on-unplug:
+    type: boolean
+    description: |
+      Some soc series allow USB function to be disabled temporarily on USB unplug. It can enable
+      back on USB re-plug.
```
josuah: Same feedback as https://github.com/zephyrproject-rtos/zephyr/pull/101460/changes#r2669016088

drivers/usb/udc/udc_numaker.c
```diff
@@ -2223,7 +2458,18 @@ static int numaker_usbd_msg_handle_out(const struct device *dev, struct numaker_
...
 		}
-	} else {
+	} else if ((net_buf_tailroom(buf) == 0) || (data_len < ep_cfg->mps) ||
+		   (ep_type == USB_EP_TYPE_ISO)) {
+		/* Fix submit condition for non-control transfer
+		 *
+		 * Do submit when any of the following conditions is met:
+		 * 1. Transfer buffer (net_buf) is full
+		 * 2. Last packet size is less than mps
+		 * 3. Isochronous transfer
```
josuah, line 2249: I assume that INTERRUPT transfer is in the same case as BULK transfers?
I am only asking out of just for curiosity and proof-reading purpose.
Thank you.

josuah (dismissed):
I do not know this hardware, but best I can say is I did not find issues regarding the USB device driver API.

dts/bindings/usb/nuvoton,numaker-hsusbd.yaml
ccli8: Thanks for correcting the term. Change to "functionality" in DT binding files and commit message.

dts/bindings/usb/nuvoton,numaker-usbd.yaml
ccli8: Same as above

drivers/usb/udc/udc_numaker.c
ccli8: Yes, 1. and 2. apply to both INTERRUPT transfer and BULK transfer.

fabiobaltieri (dismissed):

josuah (dismissed):

drivers/usb/udc/udc_numaker.c
```diff
@@ -2116,7 +2116,7 @@ static int numaker_usbd_msg_handle_setup(const struct device *dev, struct numake
 
 	udc_ep_buf_set_setup(buf);
 	data_ptr = net_buf_tail(buf);
-	numaker_usbd_setup_copy_to_user(dev, data_ptr);
+	memcpy(data_ptr, msg->setup.packet, 8);
```
tmon-nordic, line 2133: Is there possibility for getting part of the setup data from previous control transfer and part from new one in case of host timeout with very precise timing (i.e. when there are two SETUP transfers without the data/status stage in between)?

drivers/usb/udc/udc_numaker.c
```diff
@@ -2100,7 +2094,11 @@ static int numaker_usbd_msg_handle_setup(const struct device *dev, struct numake
 	/* Abort previous CTRL OUT/IN */
 	if (config->is_hsusbd) {
 		/* For HSUSBD, there is timing concern between FIFO flush and
-		 * immediately following data transaction. Do in ISR for in time.
+		 * immediately following Data OUT transaction. Even though FIFO flush
+		 * is done in Setup token ISR (HSUSBD_CEPINTSTS_SETUPTKIF_Msk),
+		 * it can still be not timely. For this, error recovery with FIFO
+		 * is not done in-place here and rely on USB reset handler to do it
+		 * as catch-all.
```
tmon-nordic, line 2112: This sounds like some serious hardware limitation. Is there really not a register access sequence that would guarantee proper operation for a Control Write (with Data OUT stage) transfer regardless of "software timing"? It sounds strange that there would be inevitable race condition, where it would have to rely on USB reset handler (I assume host issues reset after multiple control transfer timeouts) to recover.

drivers/usb/udc/udc_numaker.c
```diff
@@ -2456,7 +2458,18 @@ static int numaker_usbd_msg_handle_out(const struct device *dev, struct numaker_
 				return err;
 			}
 		}
-	} else {
+	} else if ((net_buf_tailroom(buf) == 0) || (data_len < ep_cfg->mps) ||
+		   (ep_type == USB_EP_TYPE_ISO)) {
+		/* Fix submit condition for non-control transfer
```
tmon-nordic, line 2244: What do C1 and C2 in commit message "USB mass may request C1 for sector size (512 bytes)-aligned and C2 for CBW (13 bytes)." refer to?

Enqueued MSC bulk OUT buffer from stack will always be capable of holding wMaxPacketSize while CBW is short. The short packet ending transfer is actually specified in USB 2.0 Specification.

dts/bindings/usb/nuvoton,numaker-hsusbd.yaml
```diff
@@ -25,3 +25,9 @@ properties:
     required: true
     description: |
       Size of DMA buffer in bytes
+
+  allow-disable-usb-on-unplug:
+    type: boolean
+    description: |
+      Some soc series allow USB functionality to be disabled temporarily on USB unplug.
+      It can enable back on USB re-plug.
```
tmon-nordic: I think this is not quite the format expected in devicetree binding descriptions, but I am not sure what the proper text would be. I think something more like "USB controller can be disabled when USB cable is unplugged." would fit better, but don't take my word on it.

drivers/usb/udc/udc_numaker.c
ccli8: Yes, there is the possibility. Reply the same concern below together.

drivers/usb/udc/udc_numaker.c
ccli8: For performance consideration, the design allows Control Write  without NACK after Setup stage, but doesn't reserve a flag to disable it. This limitation may cause trouble when previous Control transfer is incomplete. In this case, it may rely on USB reset to recover.

drivers/usb/udc/udc_numaker.c
ccli8: In previous implementation, BULK OUT data is submitted to stack immediately on receive. For MSC class driver, the data size can be 31 bytes (CBW) or 64 bytes (sector data), assuming wMaxPacketSize is 64 bytes. This will fail with updated MSC , in which sector data must collect to multiple of 512 bytes (net_buf being full) before submit. So there are the submit conditions C1 and C2 for MSC.

I am not sure if stack has definition on size conditions to submit DATA OUT data. It is appreciated if someone could hint.

dts/bindings/usb/nuvoton,numaker-hsusbd.yaml
ccli8: Rephrase the description: "Supports USB controller can be disabled temporarily when USB cable is unplugged and then enabled back when the cable is re-plugged."

fabiobaltieri (dismissed):

drivers/usb/udc/udc_numaker.c
ccli8: About this and above concerns on consecutive Control transfers, due to the hardware limitation, there is no proper register control to address them, and keep  current implementation.

drivers/usb/udc/udc_numaker.c
```diff
@@ -71,6 +71,143 @@ LOG_MODULE_REGISTER(udc_numaker, CONFIG_UDC_DRIVER_LOG_LEVEL);
...
 
+/* Dummy HSUSBD device definition
+ *
+ * This is used to pass compile for targets not supporting this device.
+ * Related code should be unreachable.
+ */
+#if defined(CONFIG_SOC_SERIES_M2L31X)
+
+BUILD_ASSERT(DT_NUM_INST_STATUS_OKAY(nuvoton_numaker_hsusbd) == 0,
+	     "The SoC series should have no HSUSBD");
```
jfischer-no: What is the point of this BUILD_ASSERT?

drivers/usb/udc/udc_numaker.c
```diff
@@ -71,6 +71,143 @@ LOG_MODULE_REGISTER(udc_numaker, CONFIG_UDC_DRIVER_LOG_LEVEL);
 #endif
 #endif
 
+/* Dummy HSUSBD device definition
```
jfischer-no: Can you always include hsusbd_reg.h instead?

ccli8: If the target soc has hsusbd, `hsusbd_reg.h` would have been included via `#include <soc.h>`. However, the target soc m2l31 has only usbd but no hsusbd, so it has `usbd_reg.h` but no  `hsusbd_reg.h`. This is why one dummy hsusbd is created here only to pass compile.

drivers/usb/udc/udc_numaker.c
ccli8: Guard from devicetree error. The target soc m2l31 has no hsusbd, so its devicetree should have no `nuvoton,numaker-hsusbd` compatible.

cyliangtw (dismissed):

drivers/usb/udc/udc_numaker.c
jfischer-no: Well, that is why I asked. Or perhaps you can add a common header file for this controller type to avoid 140 lines of dummy defines?

jfischer-no: You could add drivers/usb/udc/udc_numaker.h and move 
`#include <soc.h>`, `#include <NuMicro.h>` and these dummy defines there. @fabiobaltieri, what are your thoughts on this?

ccli8: Move `#include <soc.h>` and dummy usbd/hsusbd defines to `udc_numaker.h` so that `udc_numaker.c` can be more clear. `NuMicro.h` is included via `soc.h`, so `#include <NuMicro.h>` is removed.

ccli8: Ping @fabiobaltieri

fabiobaltieri: yeah I think it's a good idea to have those defines off this file

ccli8: Note the above suggestion has done in https://github.com/zephyrproject-rtos/zephyr/pull/101460/changes/ef3db832264bdf66570f994ffbcfff9cf4effa0c

fabiobaltieri (approved):

cyliangtw (approved):

jfischer-no (approved):

PR #101461, opened 2025-12-23
samples: usb: support numaker_pfm_m467/numaker_m55m1 for hsusbd

ccli8:
This supports hsusbd on numaker_pfm_m467 and numaker_m55m1 boards by changing default udc driver from usbd to hsusbd. The support samples include:
- cdc_acm
- hid-mouse
- mass
- uvc


ccli8:
Ping

tmon-nordic (commented):
Why hsusbd cannot be default udc driver?

ccli8:
> Why hsusbd cannot be default udc driver?

HSUSBD needs an external oscillator HXT, so it is not preferred as default.

jfischer-no (changes requested):
Same as for the other similar RPs. I do not accept this kind of board/soc specific overlays in USB samples. A board should configure the one controller as default with the best properties. If that is not possible for whatever reason, there is nothing that forces a sample to support your board and there are likely other ways to fix it.

PR #101462, opened 2025-12-23
kernel: timer: Add timer observer hooks for extensibility

VijsharQC:
Introduce lifecycle observer callbacks (init, start, stop, expiry) for k_timer using Zephyr's iterable sections pattern. This enables external modules to extend timer functionality without modifying kernel internals.


include/zephyr/kernel.h
```diff
@@ -1801,6 +1801,16 @@ struct k_timer {
 	 */
 };
 
+#ifdef CONFIG_TIMER_OBSERVER
+struct k_timer_observer {
+	void (*on_init)(struct k_timer *timer);
+	void (*on_start)(struct k_timer *timer, k_timeout_t duration,
+			k_timeout_t period);
```
pdgendt: Nit indentation

```suggestion
	void (*on_start)(struct k_timer *timer, k_timeout_t duration,
			 k_timeout_t period);
```

include/zephyr/kernel.h
```diff
@@ -6645,6 +6691,56 @@ static inline void k_cpu_atomic_idle(unsigned int key)
...
+ * Invoked by the kernel after k_timer_start() completes. Observers can
+ * use this to apply policies, log telemetry, or adjust behavior.
+ *
+ * @param timer    Pointer to the timer being started.
+ * @param duration Initial timeout before the first expiry.
+ * @param period   Period for subsequent expiries.
+ */
+
+void z_timer_observer_on_start(struct k_timer *timer, k_timeout_t duration,
+			  k_timeout_t period);
```
pdgendt:
```suggestion
void z_timer_observer_on_start(struct k_timer *timer, k_timeout_t duration,
			       k_timeout_t period);
```

include/zephyr/linker/common-ram.ld
```diff
@@ -151,3 +151,7 @@ ITERABLE_SECTION_RAM(scmi_protocol, Z_LINK_ITERABLE_SUBALIGN)
 #ifdef CONFIG_USERSPACE
 	PLACE_SYMBOL_HERE(_static_kernel_objects_end);
 #endif
+
+#if defined(CONFIG_TIMER_OBSERVER)
+	ITERABLE_SECTION_RAM(k_timer_observer, Z_LINK_ITERABLE_SUBALIGN)
+#endif /* CONFIG_TIMER_OBSERVER */
```
pdgendt: Why put these in `RAM` instead of `ROM`?

kernel/Kconfig
```diff
@@ -717,6 +717,14 @@ config TIMESLICE_PER_THREAD
 	  a per-thread basis, with an application callback invoked when
 	  a thread reaches the end of its timeslice.
 
+config TIMER_OBSERVER
+	bool "Support extension points for k_timer lifecycle"
+	default n
```
pdgendt: Omit `default n`

```suggestion
```

include/zephyr/kernel.h
```diff
@@ -6645,6 +6691,56 @@ static inline void k_cpu_atomic_idle(unsigned int key)
...
+/**
+ * @brief Notify registered timer observers after k_timer initialization.
+ *
+ * Called by the kernel at the end of k_timer_init() to allow external
+ * modules to perform additional setup for the timer.
+ *
+ * @param timer Pointer to the timer being initialized.
+ */
+
+void z_timer_observer_on_init(struct k_timer *timer);
```
pdgendt: These functions should be surrounded in the header file:

```c
#if defined(CONFIG_TIMER_OBSERVER) || defined(__DOXYGEN__)
// ...
#endif
```

and omit the `static inline` functions in `timer.c`.

pdgendt (commented):
Can't comment on the usefulness of the feature, but I do have some implementation suggestions.

include/zephyr/kernel.h
```diff
@@ -6645,6 +6691,57 @@ static inline void k_cpu_atomic_idle(unsigned int key)
  * @internal
  */
 void z_timer_expiration_handler(struct _timeout *timeout);
+
+#if defined(CONFIG_TIMER_OBSERVER)
```
pdgendt: Come to think of it, do we need to put any function in the header at all? Make these `static inline` in the `timer.c` file and reduce the scope to that file.

VijsharQC: Yes, we can do that way.

kernel/timer.c
```diff
@@ -19,6 +19,45 @@ static struct k_spinlock lock;
 static struct k_obj_type obj_type_timer;
 #endif /* CONFIG_OBJ_CORE_TIMER */
 
+#if defined(CONFIG_TIMER_OBSERVER)
+static inline void z_timer_observer_on_init(struct k_timer *timer)
+{
+	STRUCT_SECTION_FOREACH(k_timer_observer, obs) {
+		if (obs != NULL && obs->on_init) {
```
pdgendt: I don't think the `obs` can be `NULL`? Also no implicit boolean conversion for the function pointer. (same for the other functions)

```suggestion
		if (obs->on_init != NULL) {
```

kernel/timer.c
```diff
@@ -19,6 +19,45 @@ static struct k_spinlock lock;
...
+			obs->on_init(timer);
+		}
+	}
+}
+
+static inline void z_timer_observer_on_start(struct k_timer *timer, k_timeout_t duration,
+					      k_timeout_t period)
```
pdgendt:
```suggestion
static inline void z_timer_observer_on_start(struct k_timer *timer, k_timeout_t duration,
					     k_timeout_t period)
```

pdgendt (commented):
I think this needs tests.

kernel/timer.c
pdgendt: Make the `!= NULL` explicit.

VijsharQC:
There are few cases where timers may need to operate only in specific modes. This approach allows vendors to add custom functionality to manage those timers without making significant changes to Zephyr’s core timer implementation.

include/zephyr/linker/common-ram.ld
VijsharQC: There can be chances when someone wants to have changes in the callbacks based on the feature requirements.

pdgendt: This should be handled in the observer callback functions, instead of updating the iterable struct instances themselves IMO.

VijsharQC: "changes in the callbacks based on the feature requirements ", I mean to say if there are change in the callback function itself(ex: on_init callback is changed from x_cb to y_cb).

pdgendt: I got what you were saying, but you can have logic in your callback function for that

```c
static void some_cb(struct k_timer *timer)
{
	if (some_condition) {
		x_cb(timer);
	} else {
		y_cb(timer);
	}
}

static K_TIMER_OBSERVER_DEFINE(some, NULL, NULL, NULL, some_cb);
```

This saves the need to store your observers both in ROM and RAM.

VijsharQC: Yes, this can be one way. And I see that the STRUCT_SECTION_ITERABLE objects should be present in RAM/ROM linker to make STRUCT_SECTION_FOREACH work. Still do not want to risk it moving into ROM, what if in the future the need arise to make it mutable.

pdgendt: What risk? Most iterable sections with function handlers in Zephyr are stored in ROM. I don't see the added value of wasting RAM, especially when users can easily work around the need to update the callback handlers.

This is already a very niche feature, where I don't think there will be a lot of users. All statements so far about usage has been vague; to the point I'm doubting its usefulness. Please provide samples where this would be used.

VijsharQC: Sure, I’ll make the changes to store them in ROM. 

In one of our use cases, we need to pause and resume timers based on specific events (X and Y). Currently, there’s no built-in way to iterate over the active timers. With these callbacks, we can track which timers are running and, when event X occurs, stop the relevant timers. Later, when event Y triggers, we restart those timers with their remaining expiry time. 
Since we cannot move all of this logic into Zephyr’s core, we adopted this approach to extend timer functionality without major changes.

include/zephyr/kernel.h
```diff
@@ -1872,6 +1882,42 @@ typedef void (*k_timer_stop_t)(struct k_timer *timer);
...
+ *
+ * @param name Name of the k_timer_observer variable.
+ * @param init Pointer to initialization callback (or NULL).
+ * @param start Pointer to start callback (or NULL).
+ * @param stop Pointer to stop callback (or NULL).
+ * @param expiry Pointer to expiry callback (or NULL).
+ */
+#define K_TIMER_OBSERVER_DEFINE(name, init, start, stop, expiry) \
+	STRUCT_SECTION_ITERABLE(k_timer_observer, name) = \
+		Z_TIMER_OBSERVER_INITIALIZER(name, init, start, stop, expiry)
```
pdgendt, line 1924:
```suggestion
#define K_TIMER_OBSERVER_DEFINE(name, init, start, stop, expiry) \
	static const STRUCT_SECTION_ITERABLE(k_timer_observer, name) = \
		Z_TIMER_OBSERVER_INITIALIZER(name, init, start, stop, expiry)
```

include/zephyr/kernel.h
```diff
@@ -1872,6 +1882,42 @@ typedef void (*k_timer_stop_t)(struct k_timer *timer);
...
+/**
+ * INTERNAL_HIDDEN @endcond
+ */
+
+/**
+ * @brief Statically define and initialize a timer observer.
+ *
+ * The observer can be accessed outside the module where it is defined using:
+ *
+ * @code extern struct k_timer_observer <name>; @endcode
```
pdgendt: I don't think this useful with observers stored in ROM.

```suggestion
```

pdgendt (commented):
Still missing some sample/test, it's currently not being built in CI.

Ideally a sample is added with a use case to prove the usefulness of the feature.

VijsharQC:
After adding the test code, it is failing with below error:

INFO    - 962/962 qemu_x86_tiny/atom        kernel.timer.timer_observer                        FAILED Timeout (qemu 119.955s <zephyr>)
INFO    - /__w/zephyr/zephyr/twister-out/qemu_x86_tiny_atom/zephyr/tests/kernel/timer/timer_observer/kernel.timer.timer_observer/handler.log
ERROR   - SeaBIOS (version zephyr-v1.16.3-2-0-gf8da0cc-zephyr)
Booting from ROM..
ASSERTION FAIL [page_frames_initialized] @ WEST_TOPDIR/zephyr/kernel/mmu.c:1625
	page fault at 0x4011b015 happened too early
E: EAX: 0x4010a805, EBX: 0x4011b015, ECX: 0x4010a805, EDX: 0x40117ee0
E: ESI: 0x4011b015, EDI: 0x00000000, EBP: 0x40117f08, ESP: 0x40117ed4
E: EFLAGS: 0x00000012 CS: 0x0008 CR3: 0x00110000
E: EIP: 0x40105bcd
E: call trace:
E:      0: 0x40104818
E:      1: 0x40105add
E:      2: 0x4010070e
E:      3: 0x4010404a
E:      4: 0x40100411
E:      5: 0x40106c94
E: >>> ZEPHYR FATAL ERROR 4: Kernel panic on CPU 0
E: Current thread: 0x40111000 (main)
E: Halting system

tests/kernel/timer/timer_observer/src/main.c
```diff
@@ -0,0 +1,89 @@
...
+	k_busy_wait((dur_ms + (per_ms * 3)) * 1000);
+
+	/* Now explicitly stop; observer should see on_stop */
+	k_timer_stop(&test_periodic_timer);
+
+	/* Small delay to ensure stop path completes */
+	k_busy_wait(10 * 1000);
+
+	/* Verify timer on_init was called once */
+	zassert_equal(obs.init_cnt, 1, "obs init count mismatch");
```
peter-mitsis, line 88: This test assumes that there will only be a single timer in the system--an assumption that is unlikely to hold for all boards. A quick grep in the drivers/ directory shows that there are many drivers that use k_timer; without delving deep into their use cases, it seems reasonable that they may interfere with the expected counts. I'm not sure what the best way to resolve that would be. (It might be to just document it as a comment here noting that those boards should be excluded from the test as they show up).

pdgendt: Or only increment the counters if the timer instance matches the test timer

```c
static void obs_on_expiry(struct k_timer *timer)
{
	if (timer == &test_periodic_timer) {
		obs.expiry_cnt++;
	}
}
```

VijsharQC: Hi @pdgendt , @peter-mitsis - thank you for the suggestions.

I did try updating the observer callbacks so that they only increment counts when the timer instance matches test_periodic_timer. Unfortunately, the tests are still failing with the same issue.

As mentioned by @peter-mitsis, I’ve excluded qemu_x86_tiny for this test.

VijsharQC:
Thank you @peter-mitsis, adding __boot_func to obs_on_* routines solves this.

VijsharQC:
Hi @peter-mitsis / @andyross / @pdgendt, requesting review for this PR.

peter-mitsis (dismissed):

TaiJuWu (dismissed):

kernel/Kconfig
```diff
@@ -717,6 +717,13 @@ config TIMESLICE_PER_THREAD
 	  a per-thread basis, with an application callback invoked when
 	  a thread reaches the end of its timeslice.
 
+config TIMER_OBSERVER
+	bool "Support extension points for k_timer lifecycle"
+	help
+	  This enables an iterable-section based observer interface for k_timer
+	  lifecycle events (init/start/stop/expiry). External modules can
+	  register additional functionality without modifying kernel internals.
+	  Runs expiry hooks in ISR context.
```
mathieuchopstm: I think this would be more appropriate as a comment in the header.

kernel/timer.c
```diff
@@ -19,6 +19,45 @@ static struct k_spinlock lock;
...
+
+static inline void z_timer_observer_on_expiry(struct k_timer *timer)
+{
+	STRUCT_SECTION_FOREACH(k_timer_observer, obs) {
+		if (obs->on_expiry != NULL) {
+			obs->on_expiry(timer);
+		}
+	}
+}
+#endif /* CONFIG_TIMER_OBSERVER */
```
mathieuchopstm, line 64: A more common(I think?), and in my opinion better, pattern for this kind of thing is:
```c
#ifdef CONFIG_TIMER_OBSERVER
/* ...the implementation...*/
#else
#define z_timer_observer_on_init(timer)
#define z_timer_observer_on_start(timer, duration, period)
#define z_timer_observer_on_stop(timer)
#define z_timer_observer_on_expiry(timer)
#endif
```

Then just call the functions unconditionally (no `#ifdef`).

kernel/timer.c
```diff
@@ -80,6 +119,13 @@ void z_timer_expiration_handler(struct _timeout *t)
 	/* update timer's status */
 	timer->status += 1U;
 
+#if defined(CONFIG_TIMER_OBSERVER)
+	/* Notify observers - keep this minimal as we're in ISR context
+	 * Consider: Only notify critical observers here, defer others to work queue
+	 */
```
mathieuchopstm: This is a comment "nobody will see" - it should be in the header instead.

kernel/timer.c
```diff
@@ -134,6 +180,10 @@ void k_timer_init(struct k_timer *timer,
 
 	timer->user_data = NULL;
 
+#if defined(CONFIG_TIMER_OBSERVER)
+	z_timer_observer_on_init(timer);
+#endif /* CONFIG_TIMER_OBSERVER */
```
mathieuchopstm: Shouldn't this be called after the timer has been fully initialized (i.e., after `k_object_init` and `k_obj_core_init_and_link`)?

include/zephyr/kernel.h
```diff
@@ -1801,6 +1801,16 @@ struct k_timer {
...
 };
 
+#ifdef CONFIG_TIMER_OBSERVER
+struct k_timer_observer {
+	void (*on_init)(struct k_timer *timer);
+	void (*on_start)(struct k_timer *timer, k_timeout_t duration,
+			 k_timeout_t period);
+	void (*on_stop)(struct k_timer *timer);
+	void (*on_expiry)(struct k_timer *timer);
+};
```
mathieuchopstm, line 1818: This lacks documentation, which is particularly critical for `on_init` (receives a partially init'ed object only?), `on_stop` (called even if the timer is inactive) and `on_expiry` (runs in ISR context).

kartben (dismissed):
going out on a limb here and making @mathieuchopstm's very valid comments an actual request for changes :) thanks!

include/zephyr/kernel.h
```diff
@@ -1872,6 +1889,43 @@ typedef void (*k_timer_stop_t)(struct k_timer *timer);
...
+ * INTERNAL_HIDDEN @endcond
+ */
+
+/**
+ * @brief Statically define and initialize a timer observer.
+ *
+ * Iterable-section based observer interface for k_timer lifecycle
+ * events (init/start/stop/expiry). External modules can register
+ * additional functionality without modifying kernel internals.
+ * Runs expiry hooks in ISR context.
```
mathieuchopstm: Not necessary here
```suggestion
```

mathieuchopstm (commented):
LGTM otherwise

kernel/timer.c
```diff
@@ -19,6 +19,50 @@ static struct k_spinlock lock;
...
+		if (obs->on_expiry != NULL) {
+			obs->on_expiry(timer);
+		}
+	}
+}
+#else
+#define z_timer_observer_on_init(timer)
+#define z_timer_observer_on_start(timer, duration, period)
+#define z_timer_observer_on_stop(timer)
+#define z_timer_observer_on_expiry(timer)
```
pdgendt: Nit, to prevent empty statement (because of the `;` character).

```suggestion
#define z_timer_observer_on_init(...) (void)0
#define z_timer_observer_on_start(...) (void)0
#define z_timer_observer_on_stop(...) (void)0
#define z_timer_observer_on_expiry(...) (void)0
```

mathieuchopstm (approved):

peter-mitsis (approved):

PR #101464, opened 2025-12-23
drivers: sdhc: imx_usdhc: add retry for DAT3 card detection

MaochenWang1:
DAT3-based card detection may fail for the first attempt on RT1060evkc + NXP_IW610x_v1 card, caused by transient signal states during initialization. Add a short retry loop (by default 10 attempts, 1 ms delay) to improve detection reliability.

### DIFF
--- a/boards/shields/nxp_m2_wifi_bt/Kconfig.defconfig
+++ b/boards/shields/nxp_m2_wifi_bt/Kconfig.defconfig
@@ -124,6 +124,9 @@ endif # WIFI_NM_HOSTAPD_AP
 
 endif # WIFI
 
+configdefault IMX_USDHC_DAT3_DETECT_RETRY
+	default 10
+
 endif # SHIELD_NXP_M2_2LL_WIFI_BT
 
 if (BT_NXP_IW416) || (BT_NXP_NW612) || (BT_NXP_IW610) || (NXP_IW416) || (NXP_IW61X) || (NXP_IW610)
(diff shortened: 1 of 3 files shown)


drivers/sdhc/imx_usdhc.c
```diff
@@ -42,6 +42,10 @@ enum transfer_callback_status {
 #define IMX_USDHC_STANDARD_TUNING_COUNTER (60U)
 /* Default transfer timeout in ms for tuning */
 #define IMX_USDHC_DEFAULT_TIMEOUT         (5000U)
+/* USDHC DAT3 detect attempts */
+#define IMX_USDHC_DAT3_DETECT_RETRY_MAX   (10U)
```
danieldegrasse: Would it make sense to make this a Kconfig, and only set it for boards that need it?

JiafeiPan: Agree,  a Kconfig item for retry count, the default value could be 1,  and could be changed by boards

JiafeiPan (changes requested):

drivers/sdhc/imx_usdhc.c
MaochenWang1: please review again.

boards/shields/nxp_m2_wifi_bt/Kconfig.defconfig
```diff
@@ -124,6 +124,9 @@ endif # WIFI_NM_HOSTAPD_AP
 
 endif # WIFI
 
+config IMX_USDHC_DAT3_DETECT_RETRY
```
JiafeiPan: use configdefault

JiafeiPan (approved):

Zhiqiang-Hou (approved):

PR #101467, opened 2025-12-23
include: video: rollback video api version to 0.8.0

avolmat-st:
The video API is currently set to v1.1.0 after having been promoted to version v1.0.0 (leading to making the API stable) after its introduction in Zephyr version 2.1.

The API is still going though extensive changes, such as ControlFramework, video buffer pool handling, RTIO based buffer managements updates in order to allow making it fit with intended use-cases.

Considering its current adoption situation in Zephyr, platforms relying on it and tests, it should rather be considered as Unstable instead of Stable. As such, set the version to 0.8.0.

### DIFF
--- a/include/zephyr/drivers/video.h
+++ b/include/zephyr/drivers/video.h
@@ -19,7 +19,7 @@
  * @brief Interfaces for video devices.
  * @defgroup video_interface Video
  * @since 2.1
- * @version 1.1.0
+ * @version 0.8.0
  * @ingroup io_interfaces
  * @{
  */


avolmat-st:
This PR follows a point raised in PR #100628 (https://github.com/zephyrproject-rtos/zephyr/pull/100628#discussion_r2642721714) in which video api is modified. As stated in the comment, it turns out that the video API is currently set to version v1.1.0, making it automatically a stable api.

The API is still under going rather big updates now and I think it shouldn't be considered as stable for the moment and challenge the fact that it was promoted to stable previously.

I understand that rollback of a version number is not really proper, please consider this as a temporary solution in order to raise this issue and figure out a way to discuss about this point.

teburd:
If we've had a release with 1.x out already then its probably too late to go back.

Instead the right approach might be moving forward with a new 2.x type API that is unstable.

include/zephyr/drivers/video.h
```diff
@@ -19,7 +19,7 @@
  * @brief Interfaces for video devices.
  * @defgroup video_interface Video
  * @since 2.1
- * @version 1.1.0
+ * @version 0.8.0
```
JarmouniA, line 22: Not possible, since it has been released.
There is no such thing as rollback in the process documented in https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html

JarmouniA (changes requested):

avolmat-st:
Thank you all for your comments. Indeed as I wrote in my second comment, I also think that a rollback in the version is not appropriate.
The only reason why I mentioned this was, first to get some feedback and moreover, based on the api_lifecycle document, Stable means api version > v1.0.0, hence experimental or unstable means version has to be below < 1.0.0.

> I would like to hear from the two video maintainers before moving on with this, but given that the video API has been in the tree since v2.1 in 2019, I don't think marking it stable in v3.7 in 2023 was premature.

Maybe it is not that important to focus on the date now but in my understanding, this has been marked as Stable in 2020, not 2023, but I might have misunderstood the commit.

@ngphibang, @josuah, could you comment about the point above ?
(@ngphibang made a comment on the related PR here: https://github.com/zephyrproject-rtos/zephyr/pull/100628#discussion_r2642951347)

PR #101468, opened 2025-12-23
drivers: usb: uhc: handle stall status in transfer callback of mcux

AidenHu:
Return -EPIPE when status is kStatus_USB_TransferStall.

### DIFF
--- a/drivers/usb/uhc/uhc_mcux_ehci.c
+++ b/drivers/usb/uhc/uhc_mcux_ehci.c
@@ -162,6 +162,10 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 		}
 	}
 
+	if (status == kStatus_USB_TransferStall) {
+		err = -EPIPE;
+	}
+
 #if defined(CONFIG_NOCACHE_MEMORY)
 	if (transfer->setupPacket != NULL) {
 		uhc_mcux_nocache_free(transfer->setupPacket);
--- a/drivers/usb/uhc/uhc_mcux_ip3516hs.c
+++ b/drivers/usb/uhc/uhc_mcux_ip3516hs.c
@@ -137,6 +137,10 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


drivers/usb/uhc/uhc_mcux_ehci.c
MarkWangChinese, line 140: It seems to be better that creating one error codes converting function like `uhc_mucx_covert_mcux_err` set `err`'s value. But it is not for this pr, we can do it using a new pr.

drivers/usb/uhc/uhc_mcux_ehci.c
```diff
@@ -160,6 +160,8 @@ static void uhc_mcux_transfer_callback(void *param, usb_host_transfer_t *transfe
 				mcux_ep->update_addr = 1U;
 			}
 		}
+	} else if (status == kStatus_USB_TransferStall) {
```
MarkWangChinese: Forget to do the similar code changes for `uhc_mcux_khci.c` and `uhc_mcux_ohci.c`.

mmahadevan108:
@AidenHu , can you address the comment from @MarkWangChinese

drivers/usb/uhc/uhc_mcux_ehci.c
AidenHu: Done.

mmahadevan108 (dismissed):

drivers/usb/uhc/uhc_mcux_ehci.c
jfischer-no: Unnecessary `else if`
```c

	if (status == kStatus_USB_Success) {
	...
	}

	if (status == kStatus_USB_TransferStall) {
		err = -EPIPE;
	}
```

jfischer-no (changes requested):

drivers/usb/uhc/uhc_mcux_ehci.c
AidenHu: Yes, updated.

MarkWangChinese (approved):

mmahadevan108 (approved):

jfischer-no (approved):

PR #100998, opened 2025-12-14
Enable nxp low power comparator driver

ZhaoxiangJin:
The comparator now has its own driver API. We are planning to remove the sensor-based `mcux_lpcmp `driver and introduce the comparator-based `comparator_nxp_lpcmp `driver. In this PR, we:

1. Removed the `mcux_lpcmp `sample.
2. Marked the `mcux_lpcmp `driver as **deprecated** (will be deprecated in Zephyr 4.6).
3. Implemented the `comparator_nxp_lpcmp `driver.
4. Migrated all boards/SoCs that previously supported the `mcux_lpcmp `driver and sample to support the `comparator_nxp_lpcmp `driver and `gpio_loopback `test.

PR #101469, opened 2025-12-23
Update sensor lpcmp compatible name

ZhaoxiangJin:
1. Removed `nxp,sensor-lpcmp.yaml` and merged its properties into `nxp,lpcmp.yaml`. Now a single dt-binding (compatible: `nxp,lpcmp`) is shared by two drivers: the sensor API-based LPCMP driver and the comparator API-based LPCMP driver.
2. Updated the driver and samples accordingly. This change reverts the dt compatible modification introduced by https://github.com/zephyrproject-rtos/zephyr/pull/100998
3. Simplified the LPCMP driver migration notes by removing outdated information about the devicetree compatible renaming. The compatible `nxp,lpcmp` is now shared between both the sensor API-based and comparator API-based LPCMP drivers, so the previous note about renaming to `nxp,sensor-lpcmp` is no longer accurate.


henrikbrixandersen (changes requested):
As already discussed on Discord, a compatible should not include the name of the SDK/HAL used for implementing the driver.

dts/bindings/sensor/nxp,mcx-lpcmp.yaml
```diff
@@ -3,7 +3,7 @@
 
 description: NXP low-power analog comparator (LPCMP)
 
-compatible: "nxp,sensor-lpcmp"
+compatible: "nxp,mcx-lpcmp"
```
henrikbrixandersen: This device is not - and never was - limited to the MCX families.

butok:
> As already discussed on Discord, a compatible should not include the name of the SDK/HAL used for implementing the driver.

Hi @henrikbrixandersen 
I understand your point, but this is already established practice:
- Zephyr NXP Shim drivers based on MCUXpresso drivers (NXP HAL) contain "mcux" in their name and binding.
- Zephyr NXP Native drivers (not shims) do not use "mcux".

butok:
> > Hi @henrikbrixandersen I understand your point, but this is already established practice:
> > 
> > * Zephyr NXP Shim drivers based on MCUXpresso drivers (NXP HAL) contain "mcux" in their name and binding.
> > * Zephyr NXP Native drivers (not shims) do not use "mcux".
> 
> That may well be, but that practice is wrong. Devicetree compatibles describe the hardwares programming _model_, not the driver, not the SDK used for implementing that driver.

I agree with your point. 
But if this is so terrible, why has nobody stopped or noticed this in all these ten years?

henrikbrixandersen:
> I agree with your point.

Super.

> But if this is so terrible, why has nobody stopped or noticed this in all these ten years?

I don't know that, but to be fair we do not have that many DTS bindings with "mcux" in their compatible string:

```
$ git grep -E 'compatible:.*mcux.*' dts/bindings/|wc -l
10
```

We do, however, have a lot of driver implementation files with "mcux" in their name, denoting that it's a shim driver on top of the MCUXpresso SDK (which is fine):

```
$ git ls-files 'drivers/***/*mcux*.c'|wc -l
127
```

drivers/sensor/nxp/mcux_lpcmp/mcux_lpcmp.c
```diff
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#define DT_DRV_COMPAT nxp_sensor_lpcmp
+#define DT_DRV_COMPAT nxp_mcx_lpcmp
```
JarmouniA: The whole point here is to revert the renaming from the other PR and only have one compatible for the underlying IP.
If the 2 APIs this IP is used with cannot share the same binding, create a child node (with a child-binding or a separate binding) for the API that needs extra special properties, and leave the shared properties in the parent node.

Adding mcx (the SoC family name) to the compatible name is fine, but this is not the place for it, and should be done (and agreed upon internally) for all NXP peripherals' compatibles for consistency.

JarmouniA (dismissed):

JarmouniA:
> But if this is so terrible, why has nobody stopped or noticed this in all these ten years?

Honestly, I always thought it was a SoC family name (confusion with MCX), similar to STM32, nRF...

drivers/sensor/nxp/mcux_lpcmp/mcux_lpcmp.c
ZhaoxiangJin: single dt-binding for different API based driver is feasible in this case, updated in the PR, please review.

dts/bindings/comparator/nxp,lpcmp.yaml
```diff
@@ -120,3 +120,56 @@ properties:
       Negative input selection. "DAC" routes the internal DAC output.
       NOTE: For MCXN/A devices, IN6 and IN7 are not available, only
       IN0-IN5 and DAC.
+
+  enable-output-pin:
+    type: boolean
+    description: |
+      Decide whether to enable the comparator is available in selected pin.
```
JarmouniA:
```suggestion
      Enable the comparator in selected pin.
```

dts/bindings/comparator/nxp,lpcmp.yaml
```diff
@@ -120,3 +120,56 @@ properties:
...
+
+  enable-output-invert:
+    type: boolean
+    description: |
+      Decide whether to invert the comparator output.
```
JarmouniA:
```suggestion
      Invert the comparator output.
```

henrikbrixandersen (dismissed):

doc/releases/release-notes-4.4.rst
```diff
@@ -93,10 +93,7 @@ Deprecated APIs and options
     * The new ``comparator_nxp_lpcmp`` driver (:zephyr_file:`drivers/comparator/comparator_nxp_lpcmp.c`) has
       been introduced to support NXP LPCMP peripheral. SoCs and boards that previously supported the ``mcux_lpcmp``
       driver and its associated sample have now migrated to use the ``comparator_nxp_lpcmp`` driver along with the
-      ``gpio_loopback`` test case (:zephyr_file:`tests/drivers/comparator/gpio_loopback/src/test.c`).
-    * The original :dtcompatible:`nxp,lpcmp` has been renamed to :dtcompatible:`nxp,sensor-lpcmp`, applications
-      using the old compatible must update their devicetree nodes to use the ``mcux_lpcmp`` driver. The
-      :dtcompatible:`nxp,lpcmp` now is used for the ``comparator_nxp_lpcmp`` driver. (:github:`100998`).
+      ``gpio_loopback`` test case (:zephyr_file:`tests/drivers/comparator/gpio_loopback/src/test.c`). (:github:`100998`).
```
kartben: Why is this bullet point in the deprecated section? This seems wrong, looks like something for the migration guide?

kartben (dismissed):

doc/releases/release-notes-4.4.rst
JarmouniA: > looks like something for the migration guide?

It is a test, the user is not really expected to take any migration action.

But also, I don't see why we have to mention it at all (in the migration guide or release notes).

ZhaoxiangJin: Test-related information has been removed, please review.

henrikbrixandersen (dismissed):

drivers/sensor/nxp/mcux_lpcmp/Kconfig
```diff
@@ -5,19 +5,12 @@
 
 config MCUX_LPCMP
 	bool "NXP LPCMP driver [DEPRECATED]"
-	default y
-	depends on DT_HAS_NXP_SENSOR_LPCMP_ENABLED
+	depends on DT_HAS_NXP_LPCMP_ENABLED
```
JarmouniA, line 8:
```suggestion
	depends on DT_HAS_NXP_LPCMP_ENABLED
	depends on !COMPARATOR_NXP_LPCMP 
```
To avoid having both drivers enabled at the same time.

ZhaoxiangJin: added `depends on !COMPARATOR_NXP_LPCMP`.

ZhaoxiangJin:
> > I prefer to rebase to the latest main branch to ensure that the features in the PR pass tests on the most up-to-date codebase, avoiding any functional issues caused by time lags during the commit integration process.
> 
> CI rebases the commits on the latest main each time it runs.

that is good, but it can not check the run result, right?

JarmouniA (approved):

henrikbrixandersen (approved):

MaureenHelm (approved):

PR #101474, opened 2025-12-23
stm32: drivers: flash: ospi: fix erase regression in OPI mode by using OPI opcode instead of SFDP SPI opcode

djiatsaf-st:
Fix a regression introduced after c42c8a4da4 where, in OPI mode, the driver uses

`bet->cmd` (SPI opcode) for erase once a best erase type is found via SFDP.

In Octal OPI, this opcode is invalid and leads to erase failures (observed on
`b_u585i_iot02a` at offset 0xff000). 

The patch forces the OPI sector erase opcode
(SPI_NOR_OCMD_SE) in OPI mode when bet != NULL, restoring correct behavior.

### DIFF
--- a/drivers/flash/flash_stm32_ospi.c
+++ b/drivers/flash/flash_stm32_ospi.c
@@ -1288,7 +1288,12 @@ static int flash_stm32_ospi_erase(const struct device *dev, off_t addr,
 			}
 
 			if (bet != NULL) {
-				cmd_erase.Instruction = bet->cmd;
+				/* In OPI mode, do not use the SFDP SPI opcode (bet->cmd). */
+				if (dev_cfg->data_mode == OSPI_OPI_MODE) {
+					cmd_erase.Instruction = SPI_NOR_OCMD_SE;
+				} else {
+					cmd_erase.Instruction = bet->cmd;
+				}
 			} else {
 				/* Use the default sector erase cmd */
 				if (dev_cfg->data_mode == OSPI_OPI_MODE) {


FRASTM (commented):
LGTM

drivers/flash/flash_stm32_ospi.c
```diff
@@ -1288,7 +1288,12 @@ static int flash_stm32_ospi_erase(const struct device *dev, off_t addr,
 			}
 
 			if (bet != NULL) {
-				cmd_erase.Instruction = bet->cmd;
+				/* In OPI mode, do not use the SFDP SPI opcode (bet->cmd). */
+				if (dev_cfg->data_mode == OSPI_OPI_MODE) {
+					cmd_erase.Instruction = SPI_NOR_OCMD_SE;
+				} else {
+					cmd_erase.Instruction = bet->cmd;
+				}
```
etienne-lms: Suggestion:
```c
			if (dev_cfg->data_mode == OSPI_OPI_MODE) {
				cmd_erase.Instruction = SPI_NOR_OCMD_SE;
￼			} else if (bet != NULL) {
				cmd_erase.Instruction = bet->cmd;
			} else {
				cmd_erase.Instruction = (stm32_ospi_hal_address_size(dev) ==
							 HAL_OSPI_ADDRESS_32_BITS)
						? SPI_NOR_CMD_SE_4B
						: SPI_NOR_CMD_SE;
			}

**Review by erwango — APPROVED** [review]

**Review by FRASTM — APPROVED** [review]

File: drivers/flash/flash_stm32_ospi.c (hunk 1 above)
Comment (mathieuchopstm) [inline drivers/flash/flash_stm32_ospi.c:1296]: Since both sides of `if (bet != NULL)` do the same, it would make more sense to refactor this into:
```c
if (dev_cfg->data_mode == OSPI_OPI_MODE) {
	/* Always use OSPI erase command in OPI mode */
	cmd_erase.instruction = SPI_NOR_OCMD_SE;
} else if (bet != NULL) {
        /* Use command from best erase type if found */
	cmd_erase.Instruction = bet->cmd;
} else {
	/* Use default sector command otherwise */
	cmd_erase.Instruction = (stm32_ospi_hal_address_size(dev) ==
						HAL_OSPI_ADDRESS_32_BITS)
						? SPI_NOR_CMD_SE_4B
						: SPI_NOR_CMD_SE;
}
```

mathieuchopstm: Also note just below the:
```c
			if (bet != NULL) {
				addr += BIT(bet->exp);
				size -= BIT(bet->exp);
			} else {
				addr += SPI_NOR_SECTOR_SIZE;
				size -= SPI_NOR_SECTOR_SIZE;
			}
```
Shouldn't the `else` branch always be taken in case we're in OPI mode?

PR #101477, opened 2025-12-23
tests: net: use zsock APIs instead of POSIX APIs

cfriedt:
Use zsock APIs rather than POSIX APIs in the net subsys testsuites.

It's convention to use native zephyr APIs in the testsuites and POSIX APIs in samples.

### DIFF
--- a/include/zephyr/net/net_compat.h
+++ b/include/zephyr/net/net_compat.h
@@ -298,10 +298,6 @@ extern "C" {
 #define IN6_IS_ADDR_MC_SITELOCAL(addr)  ZSOCK_IN6_IS_ADDR_MC_SITELOCAL(addr)
 #define IN6_IS_ADDR_MC_ORGLOCAL(addr)   ZSOCK_IN6_IS_ADDR_MC_ORGLOCAL(addr)
 
-#if !defined(ARPHRD_ETHER)
-#define ARPHRD_ETHER NET_ARPHRD_ETHER
-#endif
-
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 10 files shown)


cfriedt:
* removed a number of references to fcntl and posix constants

include/zephyr/net/net_compat.h
```diff
@@ -298,10 +298,6 @@ extern "C" {
 #define IN6_IS_ADDR_MC_SITELOCAL(addr)  ZSOCK_IN6_IS_ADDR_MC_SITELOCAL(addr)
 #define IN6_IS_ADDR_MC_ORGLOCAL(addr)   ZSOCK_IN6_IS_ADDR_MC_ORGLOCAL(addr)
 
-#if !defined(ARPHRD_ETHER)
-#define ARPHRD_ETHER NET_ARPHRD_ETHER
-#endif
-
```
cfriedt, line 304: I was getting an warning (promoted to error) about this being redefined elsewhere, and this seemed to fix the issue.

jukkar: This change is not related to other ones, could you create a separate commit for it.

jukkar (commented):
No objections to the changes, most of those are cosmetic anyway.

include/zephyr/net/net_compat.h
cfriedt: Fixed!

cfriedt:
@jukkar - should be fixed now. I added the commit after - not sure if you prefer it to be added before. Happy to readjust however you need.

jukkar (approved):

pdgendt (approved):

PR #101479, opened 2025-12-23
lib: heap: Add Address Sanitizer memory poisoning support

tmleman:
This PR introduces ASAN memory poisoning support for the heap allocator through the following modifications:

## Core Implementation

- **lib/heap/heap.c**: Integrated ASAN poisoning/unpoisoning calls at key heap operations:
  - Poison entire heap memory on initialization
  - Poison freed memory to detect use-after-free
  - Unpoison memory when allocated to allow normal access
  - Poison free chunk bodies while preserving free-list metadata
  - Handle poisoning during realloc operations (both shrink and expand)

- **lib/heap/heap.h**: Added ASAN-specific macros and attributes:
  - `HEAP_NO_SANITIZE_ADDRESS` attribute for chunk field accessors
  - `ASAN_POISON_HEAP_MEMORY()` and `ASAN_UNPOISON_HEAP_MEMORY()` wrapper macros
  - Conditional compilation based on `CONFIG_SYS_HEAP_ASAN_POISONING`

- **lib/heap/Kconfig**: Added `CONFIG_SYS_HEAP_ASAN_POISONING` configuration option
  - Depends on `ASAN && ARCH_POSIX`
  - Available only on POSIX-based simulation platforms (e.g., native_sim)

## Memory Error Detection

When enabled, ASAN can detect:
- **Use-after-free**: Accessing memory after it has been freed
- **Unallocated access**: Accessing heap memory that was never allocated
- **Buffer overflows**: Writing beyond allocated boundaries into adjacent freed/unallocated regions
- **Use-after-realloc**: Accessing old pointer after realloc() relocates memory

## Implementation Details

The implementation maintains heap functionality while adding error detection:
- Metadata regions (heap structure, buckets, chunk headers) remain unpoisoned for heap operations
- Free-list pointers (first 16 bytes of free chunks) remain unpoisoned for list maintenance
- Rest of free chunk memory is poisoned to detect invalid access

### DIFF
--- a/lib/heap/Kconfig
+++ b/lib/heap/Kconfig
@@ -164,6 +164,16 @@ config SYS_HEAP_CANARIES_RANDOM
 	  Without this, canaries are still effective at detecting
 	  accidental corruption but are predictable.
 
+config SYS_HEAP_SANITIZER_HOOKS
+	bool
+	help
+	  Hidden symbol selected by any sys_heap sanitizer backend. When set,
+	  heap.c emits the generic poison/unpoison hook calls
+	  (heap_sanitizer_on_init / _alloc / _free) at the public-API boundary. The
+	  active backend implements those three entry points. Keeping the call
+	  sites gated on this single symbol means heap.c carries no
+	  backend-specific conditionals.
+
 module = SYS_HEAP
 module-str = sys_heap
 source "subsys/logging/Kconfig.template.log_config"
--- a/lib/heap/Kconfig
+++ b/lib/heap/Kconfig
@@ -174,6 +174,38 @@ config SYS_HEAP_SANITIZER_HOOKS
 	  sites gated on this single symbol means heap.c carries no
 	  backend-specific conditionals.
 
+config SYS_HEAP_SANITIZER_ASAN
+	bool "Full Address Sanitizer (ASAN) heap poisoning (native_sim)"
+	depends on ASAN && ARCH_POSIX
+	depends on SYS_HEAP_BIG_ONLY || (64BIT && !SYS_HEAP_SMALL_ONLY)
+	select SYS_HEAP_SANITIZER_HOOKS
+	help
+	  Enables integration with the toolchain's Address Sanitizer (ASAN)
+	  runtime to implement memory poisoning for the sys_heap allocator.
+
+	  The whole heap buffer is poisoned at initialization; an allocation
+	  unpoisons the user region it returns and a free re-poisons it. This
+	  lets ASAN report accesses to freed, never-allocated and out-of-bounds
+	  heap memory with a full backtrace and shadow dump.
+
+	  When enabled, the following classes of memory error are detected:
+	  - Use-after-free: accessing memory after it has been freed
+	  - Unallocated access: accessing memory that was never allocated
+	  - Buffer overruns: writing beyond the requested allocation size
+
+	  This feature is only available on native simulator based platforms
+	  where ASAN is supported. It works with any toolchain that provides
+	  the ASAN runtime and the manual poisoning API, e.g. GCC (libasan) or
+	  Clang/LLVM (compiler-rt).
+
+	  The big-heap chunk format is required so the user pointer and the end
+	  of the usable region are 8-byte aligned, matching ASAN's shadow
+	  granularity. On 64-bit this is implied; on 32-bit SYS_HEAP_BIG_ONLY
+	  must be chosen explicitly, which means the small-heap chunk format is
+	  not exercised under this configuration.
+
+	  Use for debugging and validation only.
+
 module = SYS_HEAP
 module-str = sys_heap
 source "subsys/logging/Kconfig.template.log_config"
--- a/lib/heap/heap.c
+++ b/lib/heap/heap.c
@@ -688,7 +688,7 @@ void *sys_heap_realloc(struct sys_heap *heap, void *ptr, size_t bytes)
 	if (ptr2 != NULL) {
 		size_t prev_size = sys_heap_usable_size(heap, ptr);
 
-		memcpy(ptr2, ptr, min(prev_size, bytes));
+		memcpy(ptr2, ptr, MIN(prev_size, bytes));
 		sys_heap_free(heap, ptr);
 	}
 	return ptr2;
@@ -722,7 +722,7 @@ void *sys_heap_aligned_realloc(struct sys_heap *heap, void *ptr,
 	if (ptr2 != NULL) {
 		size_t prev_size = sys_heap_usable_size(heap, ptr);
 
-		memcpy(ptr2, ptr, min(prev_size, bytes));
+		memcpy(ptr2, ptr, MIN(prev_size, bytes));
 		sys_heap_free(heap, ptr);
 	}
 	return ptr2;
--- a/lib/heap/heap.c
+++ b/lib/heap/heap.c
@@ -21,6 +21,10 @@ LOG_MODULE_REGISTER(os_heap, CONFIG_SYS_HEAP_LOG_LEVEL);
 #include <zephyr/random/random.h>
 #endif
 
+#ifdef CONFIG_SYS_HEAP_SANITIZER_HOOKS
+#include "heap_sanitizer.h"
+#endif
+
 #ifdef CONFIG_SYS_HEAP_RUNTIME_STATS
 static inline void increase_allocated_bytes(struct z_heap *h, size_t num_bytes)
 {
@@ -345,6 +349,9 @@ void sys_heap_free(struct sys_heap *heap, void *mem)
 				  chunk_usable_bytes(h, c) - mem_align_gap(h, mem));
 #endif
 
+	IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+		   (heap_sanitizer_on_free(heap, mem,
+				      chunk_usable_bytes(h, c) - mem_align_gap(h, mem))));
 	free_chunk(h, c);
 }
 
@@ -457,6 +464,7 @@ void *sys_heap_alloc(struct sys_heap *heap, size_t bytes)
 				   chunk_usable_bytes(h, c));
 #endif
 
+	IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS, (heap_sanitizer_on_alloc(heap, mem, bytes)));
 	IF_ENABLED(CONFIG_MSAN, (__msan_allocated_memory(mem, bytes)));
 	return mem;
 }
@@ -546,6 +554,7 @@ void *sys_heap_aligned_alloc(struct sys_heap *heap, size_t align, size_t bytes)
 #endif
 
 	IF_ENABLED(CONFIG_MSAN, (__msan_allocated_memory(mem, bytes)));
+	IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS, (heap_sanitizer_on_alloc(heap, mem, bytes)));
 	return mem;
 }
 
@@ -678,7 +687,15 @@ void *sys_heap_realloc(struct sys_heap *heap, void *ptr, size_t bytes)
 		return NULL;
 	}
 
+#ifdef CONFIG_SYS_HEAP_SANITIZER_HOOKS
+	size_t old_usable = sys_heap_usable_size(heap, ptr);
+#endif
+
 	if (inplace_realloc(heap, ptr, bytes)) {
+		IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+			   (heap_sanitizer_on_free(heap, ptr, old_usable)));
+		IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+			   (heap_sanitizer_on_alloc(heap, ptr, bytes)));
 		return ptr;
 	}
 
@@ -688,6 +705,14 @@ void *sys_heap_realloc(struct sys_heap *heap, void *ptr, size_t bytes)
 	if (ptr2 != NULL) {
 		size_t prev_size = sys_heap_usable_size(heap, ptr);
 
+		/*
+		 * The copy reads the source block's whole usable region, which
+		 * a sanitizer backend keeps poisoned past the user-requested
+		 * size. Re-grant access to the full region for the duration of
+		 * the copy; the free below re-poisons it.
+		 */
+		IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+			   (heap_sanitizer_on_alloc(heap, ptr, prev_size)));
 		memcpy(ptr2, ptr, MIN(prev_size, bytes));
 		sys_heap_free(heap, ptr);
 	}
@@ -708,8 +733,16 @@ void *sys_heap_aligned_realloc(struct sys_heap *heap, void *ptr,
 
 	__ASSERT((align & (align - 1)) == 0, "align must be a power of 2");
 
+#ifdef CONFIG_SYS_HEAP_SANITIZER_HOOKS
+	size_t old_usable = sys_heap_usable_size(heap, ptr);
+#endif
+
 	if ((align == 0 || ((uintptr_t)ptr & (align - 1)) == 0) &&
 	    inplace_realloc(heap, ptr, bytes)) {
+		IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+			   (heap_sanitizer_on_free(heap, ptr, old_usable)));
+		IF_ENABLED(CONFIG_SYS_HEAP_SANITIZER_HOOKS,
+			   (heap_sanitizer_on_alloc(heap, ptr, bytes)));
 		return ptr;
 	}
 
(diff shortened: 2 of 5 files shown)


npitre (commented):
This is nice work. Some comments:

- I think you are making it too complicated.   If chunk_field() and chunk_set() are marked with HEAP_NO_SANITIZE_ADDRESS, they can already access poisoned memory without triggering ASAN errors. The functio
  ns that manipulate free-list pointers:

  - `prev_free_chunk(h, c)` -> `chunk_field(h, c, FREE_PREV)`
  - `next_free_chunk(h, c)` -> `chunk_field(h, c, FREE_NEXT)`
  - `set_prev_free_chunk(...)` -> `chunk_set(h, c, FREE_PREV, ...)`
  - `set_next_free_chunk(...)` -> `chunk_set(h, c, FREE_NEXT, ...)`

  All go through the no_sanitize functions, so they can read/write the
  `FREE_PREV` and `FREE_NEXT` fields even when poisoned. Therefore the "Leave
  the free-list prev/next words unpoisoned so list maintenance can run" code is
  redundant. In fact it is perhaps even wrong as we don't want any code other
  than the allocator messing with those pointers.

- You are poisoning freed memory in `sys_heap_free()` then doing it again in
  `free_list_add_bidx()` for what is essentially the same memory.

- The metadata (`struct z_heap` and the buckets array) is internal heap
  implementation data that should never be accessed by user code. If user
  code accidentally writes to this region and corrupts it, ASAN should
  catch that bug. So I think the entire heap should be poisoned by default,
  more functions should be marked with `HEAP_NO_SANITIZE_ADDRESS` and only
  the strictly user allowed area be unpoisoned before returning from
  `sys_heap_alloc()`, etc. and repoisoning it all in `sys_free()`.

- Finally I think you should simply exclude the entire heap.c from ASAN
  scrutiny. The purpose of this ASAN poisoning feature is to catch user
  code bugs, not heap implementation bugs. The heap implementation already
  has its own validation with `CONFIG_SYS_HEAP_VALIDATE`.

tmleman:
Removed redundant poisoning from `sys_heap_free()` and rebased on main.

npitre (changes requested):
Sorry if my previous comment wasn't clear — let me try to lay out what I had 
in mind more concretely.

The purpose of ASAN poisoning in the heap is to catch bugs in user code 
(use-after-free, buffer overflows into freed regions, etc.), not to validate 
the heap implementation itself. The heap already has its own internal 
validation for that (`CONFIG_SYS_HEAP_VALIDATE`, and now the 
`SYS_HEAP_HARDENING` levels in PR #103611). So the heap implementation 
should be free to touch any heap memory it wants without ASAN getting in 
the way.

The simplest way to achieve this is to exclude heap.c from ASAN 
instrumentation entirely at the CMake level:

```
if(CONFIG_ASAN)
  set_source_files_properties(heap.c PROPERTIES COMPILE_OPTIONS -fno-sanitize=address)
endif()
```

The ASAN runtime API calls (`__asan_poison_memory_region` etc.) are regular 
function calls that still work in non-instrumented code. This eliminates 
the need for any `HEAP_NO_SANITIZE_ADDRESS` attributes and all the selective 
unpoisoning of free-list pointers and metadata.

Then the implementation becomes a mirror of the existing MSAN hooks already 
in heap.c - just a few one-liners at the public API boundary:

- `sys_heap_init()`: `ASAN_POISON_MEMORY_REGION(mem, bytes)` - poison the 
  whole heap
- `sys_heap_alloc()` / `sys_heap_aligned_alloc()`: 
  `ASAN_UNPOISON_MEMORY_REGION(mem, bytes)` - unpoison the user region on alloc
- `sys_heap_free()`: `ASAN_POISON_MEMORY_REGION(mem, usable_bytes)` - re-poison 
  on free to catch use-after-free

That's it. No per-function attributes, no carve-outs for internal metadata, 
no redundant poisoning steps.

tmleman:
First of all, thank you @npitre for the review, and sorry it took me so long to respond. I didn't have time to finish this, and since it was more of a nice-to-have feature I had to set it aside for a while.

> Sorry if my previous comment wasn't clear — let me try to lay out what I had
in mind more concretely.


I understood your comment, but I wanted to first improve my original implementation (fixing the issues you pointed out) and only then try implementing it the way you suggested. I had trouble excluding the heap code from instrumentation because `-fno-sanitize=address` was not being passed down correctly, and it took me longer than it should have to figure out how to do it. But eventually I implemented it as you wished (at least I think so), and I have to say: your way is better.

In short, the heap implementation is now excluded from ASAN instrumentation at the CMake level, and the poisoning is reduced to a few calls at the public API boundary (init/alloc/free/realloc), mirroring the existing MSAN hooks.

Please re-review and let me know what you think.

npitre (changes requested):
Thanks for coming back to this — the structure is right now, and good call excluding the `heap_*.c` helpers too, not just `heap.c`.

I checked the realloc shadow tracking on all paths (shrink, expand, no-change, move fallback) and it's correct. Two things worth changing, one of which connects to a parallel PR.

**1. Don't clear poison on the chunk header — keep it as a permanent redzone.**

Right now `asan_unpoison_chunk()` unpoisons from the chunk *base*, so the header is unpoisoned for the life of the allocation. That means the header only acts as a redzone while the neighbor is *free*: once the right neighbor is allocated, its header is unpoisoned and an overflow from the current allocation into it is no longer caught. Underflow into a chunk's own header is likewise missed.

The fix is to unpoison `[mem, mem + usable)` instead of the whole chunk, leaving the header (poisoned from init) untouched. The only obstacle is alignment: ASan's shadow is 8-byte-granular and partial-addressability is prefix-only ("first N bytes good"), so you can't keep the low half of a cell poisoned while opening the high half. That requires `mem` to be 8-aligned — i.e. an 8-byte header. So:

- `select SYS_HEAP_BIG_ONLY` from `SYS_HEAP_ASAN_POISONING`. On 64-bit this is already implied (`sizeof(void*) > 4` forces big_heap); it only bites 32-bit, where it costs 4 bytes/chunk of header — irrelevant for a native_sim debug feature. Worth a one-line Kconfig help note that the small-heap chunk format isn't exercised under this config.
- Then unpoison `[mem, mem+usable)`, keep the header poisoned.

This makes the header a real redzone on every target, catches overflow into a live neighbor and underflow into the header, and — because ASan's trailing partial-cell encoding is byte-precise — gives byte-granular tail-overflow detection for free.

**2. Converge on a shared hook interface with #110001.**

There's a parallel PR (#110001, "lightweight heap write address sanitizer") that adds a *second* heap poisoning backend — a custom shadow map that runs on real hardware instead of host libasan. It instruments the exact same heap.c sites you do (init / alloc / free / realloc), touches the same Kconfig and CMakeLists, and excludes heap.c the same way. As written, the two would collide and we'd end up with two parallel `#ifdef` ladders at the same call sites.

These are genuinely complementary, not competing: yours reuses real libasan (catches reads too, byte-precise, native_sim only); #110001's custom shadow is write-only and coarser but runs on-target. The clean outcome is one set of hook points in heap.c with two mutually-exclusive backends behind them. Proposed contract (a shared internal `lib/heap/heap_asan.h` included by heap.c):

```c
void heap_asan_on_init (struct sys_heap *heap, void *mem, size_t bytes);  /* poison whole heap   */
void heap_asan_on_alloc(struct sys_heap *heap, void *mem, size_t bytes);  /* unpoison [mem,+bytes) */
void heap_asan_on_free (struct sys_heap *heap, void *mem, size_t bytes);  /* re-poison [mem,+bytes) */
```

heap.c passes: whole-heap size on init; the user-requested `bytes` on alloc; `sys_heap_usable_size()` on free (the original request isn't retained). In-place realloc is then just `on_free(mem, old_usable)` followed by `on_alloc(mem, new_bytes)` around the resize — which, nicely, **lets you delete `asan_repoison_inplace()` entirely**: poison-old-then-unpoison-new produces the correct shadow for both shrink and expand. heap.c carries no backend `#ifdef` at the call sites, just the three hook calls (no-op stubs when neither feature is on). Your libasan backend becomes the three functions wrapping `__asan_{,un}poison_memory_region`.

One behavior note to put on the record: unpoisoning only `[mem, mem+requested]` (not the full usable region) means writing into the slack between the requested size and the usable size is now flagged as an overflow. That's the correct ASan semantic — it catches code assuming more than it asked for — but any code that does `malloc(n)` and then writes up to `usable_size()` will trip it.

Whoever lands first just establishes `heap_asan.h` and the call sites; the other adds its backend behind the same hooks. No ordering dependency either way — I'll put the same note on #110001 so we converge on one seam.

**Non-blocking nits:**
- The CMakeLists `if/else` duplicates the full source list across both branches — add a future `heap_foo.c` to one only and it silently diverges. Factor the list once (a variable, or a single listing + a grouped compile-option on the heap-internal set).
- `min` → `MIN` is an unrelated cosmetic change; better to keep the diff focused.
- Hoisting `prev_usable` to the top computes it on the in-place-success path where it's unused — only `prev_chunk_bytes` actually needed hoisting (and that goes away with the realloc-as-free-then-alloc model anyway).

lib/heap/Kconfig
```diff
@@ -51,6 +51,32 @@ config SYS_HEAP_RUNTIME_STATS
 	help
 	  Gather system heap runtime statistics.
 
+config SYS_HEAP_ASAN_POISONING
+	bool "Address Sanitizer (ASAN) memory poisoning"
+	depends on ASAN && ARCH_POSIX
+	help
+	  Enables integration with Clang's Address Sanitizer (ASAN)
```
aescolar: Nit:
```suggestion
	  Enables integration with the host Address Sanitizer (ASAN) libasan
```
gcc has had libasan support for quite a while now. So I'd phrase it in a way as to avoid implying only clang can be used

lib/heap/Kconfig
```diff
@@ -51,6 +51,32 @@ config SYS_HEAP_RUNTIME_STATS
...
+	  When enabled, the following types of memory errors can be detected:
+	  - Use-after-free: Accessing memory after it has been freed
+	  - Unallocated access: Accessing memory that was never allocated
+	  - Buffer overruns: Writing beyond allocated boundaries
+
+	  This feature is only available on POSIX-based simulation platforms
+	  (e.g., native_sim) where ASAN is supported.
+
+	  Use for debugging and validation only.
+	  Depends on compiler support for ASAN (typically Clang).
```
aescolar: Nit:
```suggestion
	  Depends on compiler support for ASAN.
```
Same as above

lib/heap/Kconfig
```diff
@@ -51,6 +51,32 @@ config SYS_HEAP_RUNTIME_STATS
...
+	  - Poisoning unallocated memory to detect access to never-allocated regions
+	  - Poisoning freed memory to detect use-after-free errors
+	  - Maintaining boundary protection between allocated and free regions
+
+	  When enabled, the following types of memory errors can be detected:
+	  - Use-after-free: Accessing memory after it has been freed
+	  - Unallocated access: Accessing memory that was never allocated
+	  - Buffer overruns: Writing beyond allocated boundaries
+
+	  This feature is only available on POSIX-based simulation platforms
```
aescolar: Nit:
```suggestion
	  This feature is only available on native simulator based platforms
```
(I try to avoid calling them POSIX based due to the confusion with the POSIX API compatibility shim component on top of Zephyr)

tests/lib/heap_asan_demo/testcase.yaml
```diff
@@ -0,0 +1,9 @@
+tests:
+  libraries.heap.asan_demo:
+    tags: heap asan
+    platform_allow: native_sim
```
aescolar: Minor:
```suggestion
    arch_allow: posix
```
Other posix arch based targets would also work

tests/lib/heap_asan_demo/testcase.yaml
```diff
@@ -0,0 +1,9 @@
+tests:
+  libraries.heap.asan_demo:
+    tags: heap asan
```
aescolar: Minor:
```suggestion
```
Why add these tags?

tests/lib/heap_asan_demo/testcase.yaml
```diff
@@ -0,0 +1,9 @@
...
+    platform_allow: native_sim
+    toolchain_allow:
+      - host/gnu
+      - host/llvm
```
aescolar: Nit/silly thing: These 2 are "all" :), so it is a bit redundant. I do not know if you are intentionally trying to be explicit about both being supported, but technically it is not doing anything.

aescolar (commented):
Thanks @tmleman 
A few minors/nits in areas @npitre did not already comment on

tmleman:
> Thanks @tmleman A few minors/nits in areas @npitre did not already comment on

@aescolar  I introduced changes addressing the comments regarding Kconfig, please take a look if this is OK. I left the DEMO part unchanged, this code is not worth reviewing for style etc. I don't plan to push it to the version I would like to merge. I just wanted the reviewer, interested person to have the possibility to test the changes.

tmleman:
@npitre

> Don't clear poison on the chunk header — keep it as a permanent redzone.

- Done, please check

> Converge on a shared hook interface with #110001.

- Working on it.

> The CMakeLists if/else duplicates the full source list across both branches — add a future heap_foo.c to one only and it silently diverges. Factor the list once (a variable, or a single listing + a grouped compile-option on the heap-internal set).


- I thought about this earlier and decided to leave it this way, this approach is not elegant but it's simple. I looked for other solutions, AI for example proposed something like this to me:

```cmake
zephyr_library()
zephyr_library_sources(heap.c)
zephyr_library_sources_ifdef(CONFIG_SYS_HEAP_RUNTIME_STATS heap_stats.c)
zephyr_library_sources_ifdef(CONFIG_SYS_HEAP_INFO          heap_info.c)
zephyr_library_sources_ifdef(CONFIG_SYS_HEAP_VALIDATE      heap_validate.c)
zephyr_library_sources_ifdef(CONFIG_SYS_HEAP_STRESS        heap_stress.c)
zephyr_library_sources_ifdef(CONFIG_SYS_HEAP_ARRAY_SIZE    heap_array.c)

if(CONFIG_SYS_HEAP_ASAN_POISONING)
  zephyr_library_compile_options(-fno-sanitize=address)
endif()
```

Trade-off: this builds the heap as its own library for all builds, not just ASAN ones.

aescolar:
> Trade-off: this builds the heap as its own library for all builds, not just ASAN ones.

That does not sound like a problem..

lib/heap/Kconfig
```diff
@@ -51,6 +51,36 @@ config SYS_HEAP_RUNTIME_STATS
 	help
 	  Gather system heap runtime statistics.
 
+config SYS_HEAP_ASAN_POISONING
+	bool "Address Sanitizer (ASAN) memory poisoning"
+	depends on ASAN && ARCH_POSIX && !SYS_HEAP_SMALL_ONLY
```
npitre: Header redzone change looks right.

`!SYS_HEAP_SMALL_ONLY` doesn't guarantee the 8-byte header the comment relies on: on 32-bit `native_sim`, an auto heap under 32K chunks is still a small (4-byte-header) heap, so `mem` is 4-aligned and the unpoison rounds across the header. Use `select SYS_HEAP_BIG_ONLY` instead.

npitre: Correction: `SYS_HEAP_BIG_ONLY` is a `choice` member, so it can't be `select`ed. Use a dependency instead:

```
depends on ASAN && ARCH_POSIX && (SYS_HEAP_BIG_ONLY || (64BIT && !SYS_HEAP_SMALL_ONLY))
```

64-bit AUTO already satisfies it (no change for the common native_sim case); 32-bit then requires explicitly choosing BIG_ONLY, which is the right place to make that call.

tmleman:
Update with latest push:
- extracted the replacement of min() with MIN() into a separate commit,
- extracted into a separate common API for multiple backends the functions (hooks: heap_sanitizer_on_init, heap_sanitizer_on_alloc, heap_sanitizer_on_free),
- fixes in dependency for kernel config.

npitre (dismissed):
Looks good. Just drop the `[DNM]` demo commit before merging.

npitre (approved):

tpambor (approved):

lib/heap/heap.c
```diff
@@ -688,7 +688,7 @@ void *sys_heap_realloc(struct sys_heap *heap, void *ptr, size_t bytes)
 	if (ptr2 != NULL) {
 		size_t prev_size = sys_heap_usable_size(heap, ptr);
 
-		memcpy(ptr2, ptr, min(prev_size, bytes));
+		memcpy(ptr2, ptr, MIN(prev_size, bytes));
```
fabiobaltieri: Hey @tmleman, these are not really equivalent, lowercase is single evaluation and should be the one to use in code, uppercase is double evaluation but can be used in macro. They are both safe in this case but the uppercase one can cause issues if the argument are functions, I did change these to the lowercase ones in 700a1a5a28a5c5ab58e80c06a3c298ab87122a0d, filing a revert of this change hope you don't mind.

And yeah they are both Zephyr function there's no inconsistency, lowercase in code uppercase in macro, same as Linux.

tmleman: I wasn't aware of that. I made the change quite a long time ago during a rebase of an earlier version. Clearly I misread the change. Thank you for catching it quickly.

PR #101483, opened 2025-12-23
drivers: bluetooth: esp32: add vendor-specific HCI commands

sylvioalves:
Add vendor-specific HCI command support for ESP32 BLE controllers:

- Read/Write TX power level with per-handle support
- Read version info (platform, variant, firmware version)
- Read supported commands and features
- Read static addresses from eFuse (C2/C6/H2)
- Read build info (controller version string)

Also adds Espressif platform/variant definitions to hci_vs.h and
sets appropriate BLE HCI buffer defaults for ESP32 SoCs.


wmrsouza (dismissed):

include/zephyr/bluetooth/hci_vs.h
jhedberg, line 1: This commit raises a couple of questions:

 - You make a reference to the "Zephyr software controller" as a justification for removing NXP & Intel IDs, however I think there's a misunderstanding here: `hci_vs.h` isn't strictly about the Zephyr upstream controller implementation - these definitions are available to be implemented by any Link Layer. The formal specification is https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/connectivity/bluetooth/api/hci.txt
 - Do you actually have an ESP32 Lower Link Layer implementation for the Zephyr LL? If so, that's really cool. @cvinayak FYI.

sylvioalves: Hi @jhedberg,

> You make a reference to the "Zephyr software controller" as a justification for removing NXP & Intel IDs, however I think there's a misunderstanding here: hci_vs.h isn't strictly about the Zephyr upstream controller implementation - these definitions are available to be implemented by any Link Layer. The formal specification is https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/connectivity/bluetooth/api/hci.txt

Ok, I removed because it is not being used anywhere. Perhaps I shouldn't have touched. I'll restore it accordingly.

> Do you actually have an ESP32 Lower Link Layer implementation for the Zephyr LL? If so, that's really cool. 

No, there is not. LL+PHY is all in closed blobs. The goal of this PR was to enabled ESP32 to be able to have dynamic power control using Zephyr's VS and those attributes. That is all.

hermabe: Yes, please keep the old numbers and just allocate a new one for Espressif. Changing the numbers here would mean that our out-of-tree Nordic controller that implements these commands would be identified as an Espressif controller: https://github.com/nrfconnect/sdk-nrfxlib/blob/c6627e79bef09b28cb8e56ad0be2e4b7aa9db760/softdevice_controller/include/sdc_hci_vs.h#L343.

sylvioalves: @hermabe fixed.

hermabe: Thanks! Should probably update `hci.txt` as well, otherwise looks good to me

wmrsouza (dismissed):

uLipe (approved):

wmrsouza (approved):

jhedberg (approved):

PR #101492, opened 2025-12-24
modules: hostap: Add new configs in Kconfig for wpa_supplicant

nxf58150:
and hostapd cleanup interval

Both wpa_supplicant and hostapd register cleanup interval to eloop and the intervals are all 10 seconds by default. In order to update the interval in configuration files, add new configs in Kconfig and update corresponding interval values in CMake file for hostap module.


MaochenWang1 (dismissed):

modules/hostap/CMakeLists.txt
```diff
@@ -737,6 +737,12 @@ zephyr_library_sources_ifdef(CONFIG_WIFI_NM_WPA_SUPPLICANT_CRYPTO_TEST
 )
 endif()
 
+zephyr_compile_definitions(WPA_SUPPLICANT_CLEANUP_INTERVAL=${CONFIG_WIFI_NM_WPA_SUPPLICANT_CLEANUP_INTERVAL})
+
+if(CONFIG_WIFI_NM_HOSTAPD_AP)
+zephyr_compile_definitions(HOSTAPD_CLEANUP_INTERVAL=${CONFIG_WIFI_NM_HOSTAPD_CLEANUP_INTERVAL})
```
jukkar: Please use `zephyr_compile_definitions_ifdef` instead so we can avoid having a separate `if()`

nxf58150: Updated. Thanks!

MaochenWang1 (dismissed):

modules/hostap/Kconfig
```diff
@@ -481,6 +481,19 @@ config WIFI_NM_WPA_SUPPLICANT_BGSCAN_LEARN
 
 endif # WIFI_NM_WPA_SUPPLICANT_BGSCAN
 
+config WIFI_NM_WPA_SUPPLICANT_CLEANUP_INTERVAL
+	int "WPA supplicant cleanup interval (seconds)"
+	default 10
+	help
+	  WPA supplicant cleanup interval.
```
krish2718: Update help with the tradeoff: low power vs. delayed cleanup

nxf58150: Updated.

krish2718 (approved):

MaochenWang1 (approved):

jukkar (approved):

PR #101493, opened 2025-12-24
drivers: i2c: Add support i2c_target for Renesas RA I2C driver

khoa-nguyen-18:
- Add support I2C target mode for Renesas RA I2C driver
- Update I2C clock source for all Renesas RA I2C node
- Add `max-bitrate-supported` property for Renesas RA I2C binding
- Add support test app `i2c_target_api` for Renesas RA devices

### DIFF
--- a/drivers/i2c/i2c_renesas_ra_iic.c
+++ b/drivers/i2c/i2c_renesas_ra_iic.c
@@ -34,6 +34,7 @@ struct i2c_ra_iic_config {
 	double rise_time_s;
 	double fall_time_s;
 	uint32_t duty_cycle_percent;
+	uint32_t max_bitrate_supported;
 };
 
 struct i2c_ra_iic_data {
@@ -67,7 +68,9 @@ struct ra_iic_master_bitrate {
 
 static int i2c_ra_iic_configure(const struct device *dev, uint32_t dev_config)
 {
+	const struct i2c_ra_iic_config *config = dev->config;
 	struct i2c_ra_iic_data *data = (struct i2c_ra_iic_data *const)dev->data;
+	uint32_t desire_bitrate;
 
 	if (!(dev_config & I2C_MODE_CONTROLLER)) {
 		LOG_ERR("Only I2C Master mode supported.");
@@ -76,16 +79,24 @@ static int i2c_ra_iic_configure(const struct device *dev, uint32_t dev_config)
 
 	switch (I2C_SPEED_GET(dev_config)) {
 	case I2C_SPEED_STANDARD:
-		data->fsp_config.rate = I2C_MASTER_RATE_STANDARD;
+		desire_bitrate = I2C_MASTER_RATE_STANDARD;
 		break;
 	case I2C_SPEED_FAST:
-		data->fsp_config.rate = I2C_MASTER_RATE_FAST;
+		desire_bitrate = I2C_MASTER_RATE_FAST;
 		break;
 	case I2C_SPEED_FAST_PLUS:
-		data->fsp_config.rate = I2C_MASTER_RATE_FASTPLUS;
+		desire_bitrate = I2C_MASTER_RATE_FASTPLUS;
 		break;
 	default:
 		LOG_ERR("%s: Invalid I2C speed rate flag: %d", __func__, I2C_SPEED_GET(dev_config));
+		return -EIO;
+	}
+
+	if (desire_bitrate <= config->max_bitrate_supported) {
+		data->fsp_config.rate = desire_bitrate;
+	} else{
+		LOG_ERR("%s: Requested bitrate %u exceeds max supported bitrate %u",
+			__func__, desire_bitrate, config->max_bitrate_supported);
 		return -EIO;
 	}
 
@@ -546,6 +557,10 @@ static DEVICE_API(i2c, i2c_ra_iic_driver_api) = {
 		irq_enable(DT_INST_IRQ_BY_NAME(index, eri, irq));                                  \
 	}                                                                                          \
                                                                                                    \
+	BUILD_ASSERT(DT_INST_PROP(index, clock_frequency) <=                                        \
+			     DT_INST_PROP(index, max_bitrate_supported),                           \
+		     "The desire clock-frequency in devicetree exceeds max-bitrate-supported");    \
+                                                                                                   \
 	static const struct i2c_ra_iic_config i2c_ra_iic_config_##index = {                        \
 		.pcfg = PINCTRL_DT_INST_DEV_CONFIG_GET(index),                                     \
 		.clock_dev = DEVICE_DT_GET(DT_INST_CLOCKS_CTLR(index)),                            \
@@ -559,6 +574,7 @@ static DEVICE_API(i2c, i2c_ra_iic_driver_api) = {
 		.rise_time_s = DT_INST_PROP(index, rise_time_ns) / RA_IIC_MASTER_DIV_TIME_NS,      \
 		.fall_time_s = DT_INST_PROP(index, fall_time_ns) / RA_IIC_MASTER_DIV_TIME_NS,      \
 		.duty_cycle_percent = DT_INST_PROP(index, duty_cycle_percent),                     \
+		.max_bitrate_supported = DT_INST_PROP(index, max_bitrate_supported),               \
 	};                                                                                         \
                                                                                                    \
 	static struct i2c_ra_iic_data i2c_ra_iic_data_##index = {                                  \
--- a/drivers/i2c/i2c_renesas_ra_iic.c
+++ b/drivers/i2c/i2c_renesas_ra_iic.c
@@ -6,50 +6,63 @@
 
 #define DT_DRV_COMPAT renesas_ra_iic
 
-#include <math.h>
 #include <zephyr/devicetree.h>
-#include <zephyr/drivers/clock_control.h>
 #include <zephyr/drivers/i2c.h>
 #include <zephyr/drivers/pinctrl.h>
 #include <zephyr/irq.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/drivers/clock_control/renesas_ra_cgc.h>
 
-#include "r_iic_master.h"
+#include <r_iic_master.h>
+#include <r_iic_slave.h>
 #include <errno.h>
 #include <soc.h>
 
 #include <zephyr/logging/log.h>
-LOG_MODULE_REGISTER(renesas_ra_iic);
+LOG_MODULE_REGISTER(renesas_ra_iic, CONFIG_I2C_LOG_LEVEL);
 
-typedef void (*init_func_t)(const struct device *dev);
-static const double RA_IIC_MASTER_DIV_TIME_NS = 1000000000;
+#define OPERATION(msg) (((struct i2c_msg *)msg)->flags & I2C_MSG_RW_MASK)
 
 struct i2c_ra_iic_config {
-	void (*irq_config_func)(const struct device *dev);
 	const struct pinctrl_dev_config *pcfg;
 	const struct device *clock_dev;
 	const struct clock_control_ra_subsys_cfg clock_subsys;
-	uint32_t noise_filter_stage;
-	double rise_time_s;
-	double fall_time_s;
-	uint32_t duty_cycle_percent;
+	uint32_t ctrl_rise_time_ns;
+	uint32_t ctrl_fall_time_ns;
+	uint32_t ctrl_noise_filter_stage;
+	uint32_t ctrl_duty_cycle_percent;
 	uint32_t max_bitrate_supported;
 };
 
 struct i2c_ra_iic_data {
-	iic_master_instance_ctrl_t ctrl;
-	i2c_master_cfg_t fsp_config;
+	iic_master_instance_ctrl_t control_ctrl;
+	i2c_master_cfg_t ctrl_fconfig;
+	i2c_master_event_t ctrl_event;
+	iic_master_extended_cfg_t iic_ctrl_ext_cfg;
+#ifdef CONFIG_I2C_TARGET
+	iic_slave_instance_ctrl_t target_ctrl;
+	i2c_slave_cfg_t target_fconfig;
+	iic_slave_extended_cfg_t iic_target_ext_cfg;
+	struct i2c_target_config *target_cfg;
+	bool transaction_completed;
+#ifdef CONFIG_I2C_TARGET_BUFFER_MODE
+	uint8_t target_buf[CONFIG_I2C_TARGET_RENESAS_RA_IIC_BUFFER_SIZE];
+#else
+	uint8_t target_buf;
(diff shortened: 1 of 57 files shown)


khoa-nguyen-18:
Update to fix the CI failures

khoa-nguyen-18:
- Update `hal_renesas` revision
- Update some typos in the driver

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,157 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+		}
+		break;
+	case I2C_SLAVE_EVENT_RX_REQUEST:
+		err = slave_cb->write_requested(data->target_cfg);
+		if (err == 0) {
+			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf,
+						    sizeof(data->slave_buf));
+			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+		} else {
+			R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf, 0);
```
thenguyenyf: Should there be an error verification similar to what is found in other places?

```suggestion
			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf, 0);
			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,157 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+					"remaining bytes");
+			}
+		} else {
+			LOG_DBG("Error is returned when calling buf_read_requested, Master will "
+				"read 0xFF for the remaining bytes");
+		}
+		data->transaction_completed = false;
+		break;
+	case I2C_SLAVE_EVENT_RX_MORE_REQUEST:
+		R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf, 0);
```
thenguyenyf:
```suggestion
		fsp_err  = R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf, 0);
		__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
```

thenguyenyf: The same update could be applied to the entire file.

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,157 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+	case I2C_SLAVE_EVENT_RX_REQUEST:
+		err = slave_cb->write_requested(data->target_cfg);
+		if (err == 0) {
+			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf,
+						    sizeof(data->slave_buf));
+			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+		} else {
+			R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf, 0);
+			LOG_DBG("Error is returned when calling write_received, send NACK to "
+				"master");
```
thenguyenyf: It can be updated to make it more contextually appropriate.

```suggestion
			LOG_DBG("The I2C target does not want to receive data."
				"Send a NACK to master to terminate the transaction.");
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,157 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+			if (buf != NULL && len != 0) {
+				fsp_err = R_IIC_SLAVE_Write(&data->slave_ctrl, buf, len);
+				__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+			} else {
+				LOG_ERR("buf is NULL or len is 0, Master will read 0xFF for the "
+					"remaining bytes");
+			}
+		} else {
+			LOG_DBG("Error is returned when calling buf_read_requested, Master will "
+				"read 0xFF for the remaining bytes");
```
thenguyenyf:
```suggestion
			LOG_DBG("I2C target doesn't provide new data."
					"The I2C bus will be left floating, MASTER will read the value 0xFF for the remaining bytes.");
```

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: I've updated it. Thank you

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: This also. Thank you

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: Also. Thank you

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,160 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+		break;
+	case I2C_SLAVE_EVENT_RX_REQUEST:
+		err = slave_cb->write_requested(data->target_cfg);
+		if (err == 0) {
+			/* Continue to receive data */
+			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 1);
+			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+		} else {
+			/* NACK the received data */
+			R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 0);
```
thenguyenyf:
```suggestion
			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 0);
			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,160 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+			LOG_DBG("Error is returned when calling write_received, send NACK to "
+				"master");
```
thenguyenyf:
```suggestion
			LOG_DBG("I2C target does not want to receive data."
				"Send a NACK to master to terminate the transaction.");
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,160 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+		data->transaction_completed = false;
+		break;
+	case I2C_SLAVE_EVENT_TX_REQUEST:
+		err = slave_cb->read_requested(data->target_cfg, &data->slave_buf);
+		if (err == 0) {
+			fsp_err = R_IIC_SLAVE_Write(&data->slave_ctrl, &data->slave_buf, 1);
+			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+		} else {
+			LOG_DBG("Error is returned when calling buf_read_requested, Master will "
+				"read 0xFF for the remaining bytes");
```
thenguyenyf:
```suggestion
			LOG_DBG("I2C target doesn't provide new data. The I2C bus will be left "
				"floating, MASTER will read the value 0xFF.");
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,160 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+		break;
+	case I2C_SLAVE_EVENT_RX_MORE_REQUEST:
+		err = slave_cb->write_received(data->target_cfg, data->slave_buf);
+		if (err == 0) {
+			/* Continue to receive data */
+			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 1);
+			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
+		} else {
+			/* NACK the received data */
+			R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 0);
```
thenguyenyf:
```suggestion
			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, &data->slave_buf, 0);
			__ASSERT_NO_MSG(fsp_err == FSP_SUCCESS);
```

thenguyenyf (dismissed):
nits

thenguyenyf (dismissed):

khoa-nguyen-18:
Hi @teburd, could you please take a look at this PR?

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -8,51 +8,77 @@
...
 struct i2c_ra_iic_config {
-	void (*irq_config_func)(const struct device *dev);
 	const struct pinctrl_dev_config *pcfg;
-	uint32_t noise_filter_stage;
-	double rise_time_s;
-	double fall_time_s;
-	uint32_t duty_cycle_percent;
+	const struct device *clock_dev;
+	const struct clock_control_ra_subsys_cfg clock_subsys;
+	double master_rise_time_s;
```
teburd: Using double in a driver seems problematic unless you've enabled fpu sharing and register stashing globally on renesas? It the arm core in use here implies FPU_SHARING, but its not explicitly depended on by the driver in Kconfig. This might be a reasonable thing to do given the use of not just float but doubles here which if it ever went back to software float ops would probably be quite slow.

Basically, I recommend adding a depends on FPU_SHARING for this driver to ensure you don't end up with software float ops.

khoa-nguyen-18:
rebase main to solve the conflict

thenguyenyf (dismissed):

nhutnguyenkc (dismissed):
LGTM

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -8,51 +8,77 @@
...
+	iic_master_extended_cfg_t iic_master_ext_cfg;
+#ifdef CONFIG_I2C_TARGET
+	iic_slave_instance_ctrl_t slave_ctrl;
+	i2c_slave_cfg_t slave_fconfig;
+	i2c_slave_event_t slave_event;
+	iic_slave_extended_cfg_t iic_slave_ext_cfg;
+	struct i2c_target_config *target_cfg;
+	bool transaction_completed;
+#ifdef CONFIG_I2C_TARGET_BUFFER_MODE
+	uint8_t slave_buf[CONFIG_I2C_TARGET_RENESAS_RA_IIC_BUFFER_SIZE];
```
teburd:
```suggestion
	uint8_t target_buf[CONFIG_I2C_TARGET_RENESAS_RA_IIC_BUFFER_SIZE];
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -8,51 +8,77 @@
...
+#else
+	uint8_t slave_buf;
```
teburd:
```suggestion
	uint8_t target_buf;
```

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,164 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+						     p_args->bytes);
+		}
+		__fallthrough;
+	case I2C_SLAVE_EVENT_TX_COMPLETE:
+		if (data->transaction_completed) {
+			slave_cb->stop(data->target_cfg);
+		}
+		break;
+	case I2C_SLAVE_EVENT_RX_REQUEST:
+		err = slave_cb->write_requested(data->target_cfg);
```
teburd: This seems ok, there's I guess a read request (RX_REQUEST) so in terms of the target its meant to write here...

drivers/i2c/i2c_renesas_ra_iic.c
```diff
@@ -253,21 +301,164 @@ static int i2c_ra_iic_transfer(const struct device *dev, struct i2c_msg *msgs, u
...
+		if (err == 0) {
+			fsp_err = R_IIC_SLAVE_Read(&data->slave_ctrl, data->slave_buf,
```
teburd: this should be calling write received repeatedly for each byte that comes in, I'm not seeing this here. How does this work without the buffer API which is the one most samples use in the tree today?

If it only works with the optional buffer API then its likely not going to work with the samples and tests in the tree (eeprom for example).

I guess alternatively if this isn't sensible for this driver...

* Add a depends on/select for the TARGET_BUFFER_API
* Update the test in the tree to account for this (it only depends on write_recieved) today.

teburd (commented):
Some questions that could use answering I think, and some minor nits. Had been stacking review comments over the last few days here but kept getting pulled away. Was trying to answer some of these myself but given the rush to review here we are :-)

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: Thanks for your concern. I agree that using float or double calculations in the driver can make it more complicated. Since the driver requires some minor math, I have updated it to use integer calculations while still maintaining good accuracy.

This update aims to make the IIC driver work across all devices, including those that do not support `FPU_SHARING`, while also reducing code size

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: Thank you. Based on Zephyr conventions, I've updated all the parameters defined in the driver to use the correct standard naming

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: The `I2C_SLAVE_EVENT_RX_REQUEST` is the event where Controller write, Target read, and it also seems correct to me in this case.

drivers/i2c/i2c_renesas_ra_iic.c
khoa-nguyen-18: In `i2c_ra_iic_target_callback` function, I separated the cases for using and not using `CONFIG_I2C_TARGET_BUFFER_MODE` with `#ifdef`. When the buffer API is not used, the driver calls `write_received` for each byte it receives.

image

Please let me know if this addresses your concern. Thank you

teburd (dismissed):

teburd (approved):

duynguyenxa (approved):

PR #101496, opened 2025-12-24
include: hwspinlock.h: Optimize hw_spin_trylock function

ZhaoxiangJin:
If `api->trylock(dev, id)` fails (typically because a cross-core HW lock is held by another core and returns `-EBUSY`, or returns `-EINVAL`, etc.), the function will immediately return the error code, but the local `ctx->lock` is not released, and interrupts also remain in the locked state from the trylock. This may cause IRQ lock state leakage. If there's no opportunity in the call path to reach `k_spin_unlock()`, the system may exhibit unpredictable "freeze-like" behavior.

### DIFF
--- a/include/zephyr/drivers/hwspinlock.h
+++ b/include/zephyr/drivers/hwspinlock.h
@@ -266,7 +266,15 @@ static inline int hw_spin_trylock(const struct device *dev, hwspinlock_ctx_t *ct
 	if (ret) {
 		return ret;
 	}
-	return api->trylock(dev, id);
+
+	ret = api->trylock(dev, id);
+	if (ret) {
+		/* HW trylock failed: release local lock before returning. */
+		k_spin_unlock(&ctx->lock, *key);
+		return ret;
+	}
+
+	return 0;
 }
 
 /**


ZhaoxiangJin:
cc @Fymyte

ZhaoxiangJin:
I noticed that this domain doesn't have a maintainer/collaborator, so I added some people who I think can do a good job reviewing PR quality. If this is disruptive, feel free to remove yourself.

Fymyte:
Ah yes, that's quite a mistake 😅, thank you for the catch

ZhaoxiangJin:
> Ah yes, that's quite a mistake 😅, thank you for the catch


I saw through the history records that you have done a lot of work in this domain. I'm not sure if this is a correct suggestion: I suggest you apply to become the maintainer/collaborator of this domain (if you are willing and if the TSC  agrees as well), so at least there will be someone with relevant experience to review the related PRs.

fabiobaltieri (approved):

bjarki-andreasen (approved):

PR #101498, opened 2025-12-24
boards: esp32s3_touch_lcd_1_28: remove cs-gpios to use hardware CS

faxe1008:
The board has hardware CS configured via pinctrl (SPIM2_CSEL_GPIO9) but also defines cs-gpios for the same pin. After commit 5d1e443bdf8e, the driver disables hardware CS when cs-gpios is present, breaking display communication. Remove cs-gpios to rely on hardware CS via pinctrl.


JarmouniA (dismissed):

faxe1008:
@sylvioalves I tested your solution and its indeed fixed and better then my approach. Thanks for your help :^)

PR #101499, opened 2025-12-24
Enable sema42 driver based on hwspinlock API

ZhaoxiangJin:
1. implemented nxp (Semaphores2) sema42 peripheral based on zephyr hwspinlock device driver API.
2. enabled sema42 device node for `frdm_mcxn947` and `mcx_n9xx_evk` boards.
3. add` vnd,hwspinlock-consumer.yaml` for test purpose.
4. enabled hwspinlock sample, This sample demonstrate cross-core synchronization using the HWSPINLOCK API together with the MBOX API.
5. enabled hwspinlock sample for NXP `frdm_mcxn947` and `mcx_n9xx_evk` platform.
<details>
<summary>Logs</summary>

```
HOST:
*** Booting Zephyr OS build v4.3.0-2894-g3ac0cef356bc ***
Hello from HOST - frdm_mcxn947/mcxn947/cpu0
Maximum RX channels: 4
Maximum TX channels: 4
HWSPINLOCK max id: 15
Maximum bytes of data in the TX message: 3
Demo: HOST locks gate, sends ping while holding it; REMOTE blocks on same gate then replies.
HOST: locked gate id=0, sending ping on channel 1
HOST: unlocked gate, waiting for pong...
HOST: received pong
HOST: locked gate id=0, sending ping on channel 1
HOST: unlocked gate, waiting for pong...
HOST: received pong
HOST: locked gate id=0, sending ping on channel 1
HOST: unlocked gate, waiting for pong...
HOST: received pong
HOST: locked gate id=0, sending ping on channel 1
HOST: unlocked gate, waiting for pong...
HOST: received pong
```

```
REMOTE:
*** Booting Zephyr OS build v4.3.0-2894-g3ac0cef356bc ***
Hello from REMOTE - frdm_mcxn947/mcxn947/cpu1
Maximum RX channels: 4
Maximum TX channels: 4
HWSPINLOCK max id: 15
Maximum bytes of data in the TX message: 3
REMOTE: got ping (rx_sem), checking gate contention...
REMOTE: trylock returned -16 (expected if HOST holds the lock)
REMOTE: blocking on hw_spin_lock_dt()...
REMOTE: acquired gate id=0
REMOTE: locked gate id=0, sending pong on channel 0
REMOTE: got ping (rx_sem), checking gate contention...
REMOTE: trylock returned -16 (expected if HOST holds the lock)
REMOTE: blocking on hw_spin_lock_dt()...
REMOTE: acquired gate id=0
REMOTE: locked gate id=0, sending pong on channel 0
```

</details>


dts/bindings/hwspinlock/nxp,sema42.yaml
```diff
@@ -0,0 +1,36 @@
...
+  clocks:
+    type: phandle-array
+    description: |
+      Clock controller specifier for the SEMA42 peripheral clock gate.
+
+  domain-id:
+    type: int
+    required: true
+    description: |
+      Domain ID (AHB master ID) used by this CPU/domain when locking/unlocking
```
zejiang0jason: Domain ID is not always AHB master ID, for some platforms, it could be dynamically assigned using some domain controller, such as XRDC.

samples/drivers/hwspinlock/remote/src/main.c
```diff
@@ -0,0 +1,102 @@
...
+int main(void)
+{
+	int ret;
+
+	printk("Hello from REMOTE - %s\n", CONFIG_BOARD_TARGET);
+
+	const struct mbox_dt_spec rx_channel = MBOX_DT_SPEC_GET(DT_PATH(mbox_consumer), rx);
+	const struct mbox_dt_spec tx_channel = MBOX_DT_SPEC_GET(DT_PATH(mbox_consumer), tx);
+	struct hwspinlock_dt_spec gate =
+			HWSPINLOCK_DT_SPEC_GET_BY_NAME(DT_PATH(hwspinlock_consumer), gate0);
```
zejiang0jason, line 38: should use `hwlock-names` instead of `gate0`?

drivers/hwspinlock/hwspinlock_nxp_sema42.c
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		int ret = clock_control_on(cfg->clock_dev, cfg->clock_subsys);
+
+		if (ret) {
+			LOG_ERR("Device clock turn on failed");
+			return ret;
+		}
+	}
+
```
zejiang0jason, line 123: is it necessary to reset all locks at the beginning?

drivers/hwspinlock/hwspinlock_nxp_sema42.c
```diff
@@ -0,0 +1,151 @@
...
+	while (true) {
+		sys_write8(lock_val, gate_addr);
+		gate_val = sys_read8(gate_addr) & 0x0FU;
+
+		if (gate_val == lock_val) {
+			return;
+		}
+
+		arch_spin_relax();
```
zejiang0jason, line 86: why need this?

samples/drivers/hwspinlock/remote/src/main.c
```diff
@@ -0,0 +1,102 @@
...
+
+		/* Demonstrate contention: trylock should fail while HOST still holds the gate. */
+		k_spinlock_key_t key;
+
+		ret = hw_spin_trylock_dt(&gate, &key);
+		if (ret == 0) {
+			/* Unexpected: trylock succeeded (HOST must have released early). */
+			printk("REMOTE: trylock unexpectedly succeeded, releasing immediately\n");
+			hw_spin_unlock_dt(&gate, key);
+			continue;
```
zejiang0jason, line 90: this sample runs infinite loops, if some error happens, could the two sides continue running?

samples/drivers/hwspinlock/remote/src/main.c
ZhaoxiangJin: should be `gate0`, `hwlock-names` is the property name, `gate0 `is the property value.

drivers/hwspinlock/hwspinlock_nxp_sema42.c
ZhaoxiangJin: Assuming core0 executes `nxp_sema42_lock()` and doesn't release the lock, when core1 attempts to acquire the lock, it will enter a while(true) loop, continuously writing/reading registers. Adding `arch_spin_relax `can relax/ease the pressure on cpu1. Just like https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/spinlock.h#L206

samples/drivers/hwspinlock/remote/src/main.c
ZhaoxiangJin: I'm not sure what error you're referring to? This sample just demonstrates cross-core synchronization using the HWSPINLOCK API together with the MBOX API.

dts/bindings/hwspinlock/nxp,sema42.yaml
ZhaoxiangJin: updated

drivers/hwspinlock/hwspinlock_nxp_sema42.c
ZhaoxiangJin: added

zejiang0jason: we should consider the corporation between multiple cores, the sema42 only need one time reset, not every core need to reset them.

drivers/hwspinlock/hwspinlock_nxp_sema42.c
```diff
@@ -0,0 +1,167 @@
...
+ */
+static inline mem_addr_t nxp_sema42_gate_addr(const struct device *dev, uint32_t id)
+{
+	return (mem_addr_t)(DEVICE_MMIO_GET(dev) + ((id ^ 3U) & 0x0FU));
+}
+
+static inline uint8_t nxp_sema42_lock_value(const struct device *dev)
+{
+	/* HW expects (domain_id + 1), where domain_id is 0..14 */
+	const struct nxp_sema42_config *cfg = (const struct nxp_sema42_config *)dev->config;
```
zejiang0jason, line 50: For MCXN947, it is OK to use static value defined in dts. For improvement, need consider the case that the ID is dynamically allocated.

samples/drivers/hwspinlock/remote/src/main.c
zejiang0jason: This project needs corporation between two cores. The code here continues to next loop when `ret =0`, I was asking could it continue running? because the peer core is in an wrong state.
This is not a blocking item, because this is just an sample, the error may never happen.

dts/bindings/test/vnd,hwspinlock-consumer.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+description: VND HW spinlock consumer
+
+compatible: "vnd,hwspinlock-consumer"
+
+include: base.yaml
+
+properties:
+  hwlocks:
+    type: phandle-array
```
Fymyte: You should not have to re-specify the type for hwlocks. It is already defined in base.yaml

drivers/hwspinlock/hwspinlock_nxp_sema42.c
ZhaoxiangJin: You mean that `domain_id` is dynamically allocated to the domain controller at `runtime`?

drivers/hwspinlock/hwspinlock_nxp_sema42.c
ZhaoxiangJin: In Zephyr multi-core setups, each core runs its own device init sequence. If we unconditionally perform a `reset all gates` sequence in `nxp_sema42_init()`, a later-booting core could clear gates while another core is actively holding a lock, effectively breaking the critical-section protection and allowing concurrent access to shared resources.

I have removed reset operation, per the SEMA42 RM, the module does not require initialization and gates are expected to be cleared by POR/system reset. The secure reset mechanism is mainly for exceptional recovery cases, not normal driver init.

samples/drivers/hwspinlock/remote/src/main.c
ZhaoxiangJin: Good point. The previous continue on `trylock==0` could desynchronize the ping/pong flow (HOST waits for pong but REMOTE skips it). I updated the sample to treat `trylock==0` as a valid acquisition and continue the normal flow. Also replaced the blocking `hw_spin_lock_dt()` with bounded trylock polling + timeout so the sample doesn’t hang forever if the peer core gets stuck holding the gate; it now prints diagnostics and continues looping.

dts/bindings/test/vnd,hwspinlock-consumer.yaml
ZhaoxiangJin: removed `type: phandle-array`

Fymyte (approved):

ZhaoxiangJin:
update:
Added comments for the `MCUX_SEMA42_CLK` macro to resolve a CI issue. It appears that doxygen check recently added macro comments check. Adding comments to some macros doesn't make much sense, but there may not be a good way to distinguish which macros actually need comments.

TomasGalbickaNXP (approved):

butok (approved):

Holt-Sun (approved):

PR #101500, opened 2025-12-24
Bluetooth: Classic: HFP: CLCC report NULL when finish

expliyh:
The CLCC API does not report when it finishes, so the app cannot
determine whether the CLCC procedure has ended. This commit modifies the
behavior of the CLCC API to call the callback with a NULL Call pointer to
indicate that the CLCC procedure has completed.


gzh-terry (approved):

lylezhu2012:
Regardless of whether there is a `+CLCC` response, there will always be a `query_call(hf, NULL)` callback in the end. This is a bit strange.
Have you ever thought about adding a timeout mechanism directly at the application layer?

PR #101503, opened 2025-12-24
drivers: spi: esp32: move pinctrl to init

sylvioalves:
Move pinctrl_apply_state() from configure to init to ensure pinctrl is applied once at boot rather than on every transaction.

When cs-gpios is defined, spi_context_cs_configure_all() runs after pinctrl and reconfigures the CS pin as GPIO, properly overriding any CS routing set by pinctrl.


JarmouniA (approved):
Makes sense.

marekmatej (approved):

wmrsouza (approved):

PR #101506, opened 2025-12-24
drivers: dma: emul: check complete_callback_en flag

osamakader:
The code was using 'if (true)' which always called the completion callback, ignoring the complete_callback_en flag. Fix by checking the flag before calling the callback, matching the intended behavior and the pattern used for error callbacks.

### DIFF
--- a/drivers/dma/dma_emul.c
+++ b/drivers/dma/dma_emul.c
@@ -268,8 +268,7 @@ static void dma_emul_work_handler(struct k_work *work)
 		dma_emul_set_channel_state(dev, channel, DMA_EMUL_CHANNEL_STOPPED);
 		k_spin_unlock(&data->lock, key);
 
-		/* FIXME: tests/drivers/dma/chan_blen_transfer/ does not set complete_callback_en */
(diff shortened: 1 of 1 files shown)


osamakader:
Please note that SonarCloud error is not related to this PR change.

drivers/dma/dma_emul.c
```diff
@@ -268,8 +268,7 @@ static void dma_emul_work_handler(struct k_work *work)
 		dma_emul_set_channel_state(dev, channel, DMA_EMUL_CHANNEL_STOPPED);
 		k_spin_unlock(&data->lock, key);
 
-		/* FIXME: tests/drivers/dma/chan_blen_transfer/ does not set complete_callback_en */
-		if (true) {
+		if (xfer_config.complete_callback_en) {
```
teburd, line 271: This isn't quite right, see https://docs.zephyrproject.org/latest/doxygen/html/structdma__config.html#a39526177f03017d245651b4f51820161

teburd (changes requested):

PR #101509, opened 2025-12-25
tests: adc: update adc samples and tests for nxp platform

hakehuang:
update adc samples platform allow for NXP platforms

### DIFF
--- a/boards/nxp/frdm_mcxa344/frdm_mcxa344.yaml
+++ b/boards/nxp/frdm_mcxa344/frdm_mcxa344.yaml
@@ -14,6 +14,7 @@ toolchain:
   - zephyr
   - gnuarmemb
 supported:
+  - adc
   - arduino_gpio
   - counter
   - flash
--- a/boards/nxp/frdm_mcxe247/frdm_mcxe247.yaml
+++ b/boards/nxp/frdm_mcxe247/frdm_mcxe247.yaml
@@ -14,6 +14,7 @@ toolchain:
   - zephyr
   - gnuarmemb
 supported:
+  - adc
   - uart
   - i2c
   - spi
--- a/boards/nxp/mimxrt595_evk/mimxrt595_evk_mimxrt595s_cm33.yaml
+++ b/boards/nxp/mimxrt595_evk/mimxrt595_evk_mimxrt595s_cm33.yaml
@@ -14,6 +14,7 @@ toolchain:
 ram: 4608
 flash: 65536
 supported:
+  - adc
   - arduino_gpio
   - arduino_i2c
   - arduino_serial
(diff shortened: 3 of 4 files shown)


samples/drivers/adc/adc_dt/sample.yaml
```diff
@@ -41,15 +41,22 @@ tests:
...
       - mcx_n9xx_evk/mcxn947/cpu0
       - frdm_mcxc242
+      - frdm_mcxc444
       - ucans32k1sic
       - siwx917_rb4342a
       - slwrb4180a
       - xg27_rb4194a
       - xg29_rb4412a
+      - frdm_mcxa153
+      - frdm_mcxa156
```
JarmouniA: One board for each SoC series is sufficient, no!?

JarmouniA (changes requested):

samples/drivers/adc/adc_dt/sample.yaml
hakehuang: as this has been added to the boards folder, it would be strange not in the support list. @JarmouniA  is this a must follow rule here?

JarmouniA: > as this has been added to the boards folder

That's nrelated.

> it would be strange not in the support list

That's not a support list, rather what's will be built in CI, which means multiple boards of the same SoC series is completely unnecessary for a Sample & will waste valuable CI time for no benefit at all.

samples/drivers/adc/adc_sequence/sample.yaml
```diff
@@ -34,6 +34,12 @@ tests:
       - xg29_rb4412a
       - raytac_an54lq_db_15/nrf54l15/cpuapp
       - mck_ra4t1
+      - mimxrt595_evk/mimxrt595s/cm33
+      - mimxrt1040_evk
+      - frdm_mcxc242
+      - frdm_mcxe247
+      - frdm_mcxc444
+      - frdm_mcxe31b
```
JarmouniA: ditto

samples/drivers/adc/adc_dt/sample.yaml
kartben: @hakehuang yes it's a must follow rule :)

Only what is documented in the readme should be covered in test cases. More is unnecessary and usually abusing samples to treat them as tests

"The tests should verify the default configuration of the sample and any optional features documented in the sample’s README file."
 https://docs.zephyrproject.org/latest/samples/sample_definition_and_criteria.html#:~:text=The%20tests%20should%20verify%20the%20default%20configuration%20of%20the%20sample%20and%20any%20optional%20features%20documented%20in%20the%20sample%E2%80%99s%20README%20file.

samples/drivers/adc/adc_sequence/sample.yaml
hakehuang: I remove the sample.yaml changes and only add platform adc support, and leave test enhancement to https://github.com/zephyrproject-rtos/zephyr/pull/101521

kartben (approved):

JarmouniA (approved):

PR #101518, opened 2025-12-25
drivers: led_strip: ws2812_rpi_pico_pio: Add DMA support

TerryGeng:
so that time-consuming I/O takes place in the background. This is especially important when writing to multiple long LED strips.

~~The new Kconfig option that enables this is
`CONFIG_WS2812_STRIP_RPI_PICO_PIO_DMA`.~~ DMA is enabled by the global `CONFIG_DMA` flag and properly populating the `tx` DMA channel in the DT (see below DT example).

This commit mostly mirrors what was done in the pl022 driver and the rpi_pico_pio_spi driver, including the needed DT configuration. If no proper `tx` channel is configured for a device, DMA will not be used for that device (the same behavior as the pl022 driver).

The DT overlay I used when testing:
```
&pio0 {
	status = "okay";

	pio-ws2812 {
		compatible = "worldsemi,ws2812-rpi_pico-pio";
		status = "okay";
		pinctrl-0 = <&ws2812_pio0_default>;
		pinctrl-names = "default";
		bit-waveform = <3>, <3>, <4>;

                /* This device has no tx channel configured so DMA will NOT
                 * be used. */
		ws2812_1: ws2812_1 {
			status = "okay";
			gpios = <&gpio0 23 GPIO_ACTIVE_HIGH>;
			chain-length = <1>;
			color-mapping = <LED_COLOR_ID_GREEN
                                         LED_COLOR_ID_RED
                                         LED_COLOR_ID_BLUE>;
			reset-delay = <280>;
			frequency = <800000>;
		};

                /* This device has proper tx channel configured so DMA will
                 * be used. */
		ws2812_2: ws2812_2 {
			status = "okay";
			gpios = <&gpio0 19 GPIO_ACTIVE_HIGH>;
			chain-length = <64>;
			color-mapping = <LED_COLOR_ID_RED
                                         LED_COLOR_ID_GREEN
                                         LED_COLOR_ID_BLUE>;
			reset-delay = <280>;
			frequency = <800000>;

			dmas = <&dma 1 0 0>;
                        /* DMA slot will be
                         * determined by the driver so what's in here doesn't
                         * matter. */
			dma-names = "tx";
		};
	};
};

&dma {
	status = "okay";
};
```

### DIFF
--- a/drivers/led_strip/ws2812_rpi_pico_pio.c
+++ b/drivers/led_strip/ws2812_rpi_pico_pio.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2023 TOKITA Hiroshi
+ * Copyright (c) 2025 The Zephyr Project Contributors
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -18,6 +19,7 @@ LOG_MODULE_REGISTER(ws2812_rpi_pico_pio, CONFIG_LED_STRIP_LOG_LEVEL);
 
 struct ws2812_led_strip_data {
 	uint32_t sm;
+	struct k_timer reset_on_complete_timer;
 };
 
 struct ws2812_led_strip_config {
@@ -65,20 +67,17 @@ static int ws2812_led_strip_sm_init(const struct device *dev)
 	return sm;
 }
 
-/*
- * Latch current color values on strip and reset its state machines.
- */
-static inline void ws2812_led_strip_reset_delay(uint16_t delay)
-{
-	k_usleep(delay);
-}
-
 static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb *pixels,
 				       size_t num_pixels)
 {
 	const struct ws2812_led_strip_config *config = dev->config;
 	struct ws2812_led_strip_data *data = dev->data;
 	PIO pio = pio_rpi_pico_get_pio(config->piodev);
+
+	/* Wait for the delay needed to latch current color values on
+	 * WS2812 and reset its state machine.
+	 */
+	k_timer_status_sync(&data->reset_on_complete_timer);
 
 	for (size_t i = 0; i < num_pixels; i++) {
 		uint32_t color = 0;
@@ -104,7 +103,8 @@ static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb 
 		pio_sm_put_blocking(pio, data->sm, color << (config->num_colors == 4 ? 0 : 8));
 	}
 
-	ws2812_led_strip_reset_delay(config->reset_delay);
+	k_timer_start(&data->reset_on_complete_timer, K_USEC(config->reset_delay),
+			K_NO_WAIT);
 
 	return 0;
 }
@@ -157,6 +157,8 @@ static int ws2812_led_strip_init(const struct device *dev)
 	}
 
 	data->sm = sm;
+
+	k_timer_init(&data->reset_on_complete_timer, NULL, NULL);
 
 	return 0;
 }
--- a/drivers/led_strip/ws2812_rpi_pico_pio.c
+++ b/drivers/led_strip/ws2812_rpi_pico_pio.c
@@ -15,11 +15,41 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(ws2812_rpi_pico_pio, CONFIG_LED_STRIP_LOG_LEVEL);
 
+#if defined(CONFIG_DMA)
+
+#include <zephyr/drivers/dma.h>
+#if defined(CONFIG_SOC_SERIES_RP2040)
+#include <zephyr/dt-bindings/dma/rpi-pico-dma-rp2040.h>
+#elif defined(CONFIG_SOC_SERIES_RP2350)
+#include <zephyr/dt-bindings/dma/rpi-pico-dma-rp2350.h>
+#endif
+
+#include "hardware/dma.h"
+
+#endif
+
 #define DT_DRV_COMPAT worldsemi_ws2812_rpi_pico_pio
+
+#if defined(CONFIG_DMA)
+struct ws2812_led_strip_dma_data {
+	struct dma_config dma_config;
+	struct dma_block_config dma_block;
+	struct k_sem complete_sem;
+	uint32_t *pixel_buf;
+};
+
+struct ws2812_led_strip_dma_config {
+	const struct device *dev;
+	uint32_t tx_channel;
+};
+#endif
 
 struct ws2812_led_strip_data {
 	uint32_t sm;
 	struct k_timer reset_on_complete_timer;
+#if defined(CONFIG_DMA)
+	struct ws2812_led_strip_dma_data *dma_data;
+#endif
 };
 
 struct ws2812_led_strip_config {
@@ -31,6 +61,9 @@ struct ws2812_led_strip_config {
 	const uint8_t *const color_mapping;
 	uint16_t reset_delay;
 	uint32_t cycles_per_bit;
+#if defined(CONFIG_DMA)
+	const struct ws2812_led_strip_dma_config dma_config;
+#endif
 };
 
 struct ws2812_rpi_pico_pio_config {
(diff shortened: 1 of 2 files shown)


drivers/led_strip/Kconfig.ws2812
```diff
@@ -121,3 +121,13 @@ config WS2812_STRIP_RPI_PICO_PIO
...
+
+config WS2812_STRIP_RPI_PICO_PIO_DMA
+	bool "Use DMA for WS2812 LED strip Raspberry Pi Pico PIO driver"
+	default n
+	select DMA
+	depends on WS2812_STRIP_RPI_PICO_PIO
+	help
+	  Enable DMA for WS2812 LED strip Raspberry Pi Pico PIO driver. I/O to the LED strip will take place in
+	  the background and the function that updates the pixels will return immediately. Note that this
+	  option will be ignored if no DMA `tx` channel is configured for this device in the devicetree.
```
thedjnK:
```suggestion
```
No Kconfig is needed, if the option is there in dts, use DMA for it

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -80,31 +228,39 @@ static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb
...
+			      K_NO_WAIT);
 	}
-
-	ws2812_led_strip_reset_delay(config->reset_delay);
+#if defined(CONFIG_WS2812_STRIP_RPI_PICO_PIO_DMA)
+	else {
+		for (size_t i = 0; i < num_pixels; i++) {
+			data->pixel_buf[i] = ws2812_led_strip_map_color(dev, &pixels[i]);
+		}
+		ws2812_led_strip_start_dma_put(dev);
```
thedjnK: this should be blocking until it has finished (use a semaphore)

TerryGeng: Gotcha. Thanks for the really quick review during the holiday season!

I see the point that I shouldn't deviate from the current expected blocking behavior of the APIs, but do you consider adding some APIs like what we have for async SPI transfer (this will definitely not be part of this PR)?

drivers/led_strip/Kconfig.ws2812
TerryGeng: At this moment, whether to declare DMA-related members in `ws2812_led_strip_data` depends on this Kconfig. Without this Kconfig, those members will be declared even if DMA is not configured and those members are never going to be used. So there is some implication for the binary size. This approach has been used in the spi_pl022 driver. If this is not a concern here, I'm totally okay to drop this Kconfig item.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -65,12 +95,130 @@ static int ws2812_led_strip_sm_init(const struct device *dev)
...
+	const struct device *dev = (const struct device *)arg;
+	struct ws2812_led_strip_data *data = dev->data;
+	const struct ws2812_led_strip_config *dev_cfg = dev->config;
+
+	if (dma_dev != dev_cfg->dma_config.dev) {
+		/* rpi pico has only one dma controller so far
+		 * so this branch will never run in theory.
+		 */
+		return;
+	}
```
ajf58, line 171: This feels like an `ASSERT` would capture this better?

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -65,12 +95,130 @@ static int ws2812_led_strip_sm_init(const struct device *dev)
...
-	k_usleep(delay);
+	const struct ws2812_led_strip_config *config = dev->config;
+	uint32_t color = 0;
+
+	for (size_t j = 0; j < config->num_colors; j++) {
+		switch (config->color_mapping[j]) {
+		/* White channel is not supported by LED strip API. */
+		case LED_COLOR_ID_WHITE:
+			color |= 0;
+			break;
```
ajf58: Other values are defined in `include/zephyr/dt-bindings/led/led.h` and are also not supported, correct? If so, would a `default` catch-all (for "not red, or green, or blue") be better?

ajf58 (dismissed):

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: I like the idea of a default branch and will do that. The reason that `LED_COLOR_ID_WHITE ` is treated specially is that there are two variants of WS2812, one takes RGB, and one takes RGBW (which is not fully supported). Other color values have no proper reason to appear here and should be rejected even at compile time.

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: I'm purely speculating now... Maybe RPI will have chips with multiple DMA controllers in the future?

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: Done.

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: Done.

drivers/led_strip/Kconfig.ws2812
TerryGeng: Done. Regardless of my questions, I have followed this suggestion.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2023 TOKITA Hiroshi
+ * Copyright (c) 2025 The Zephyr Project Contributors
```
soburi, line 3: AI generated?
Please put your name here.

TerryGeng: Haha no... I'm too primitive to operate the modern AI machine. Per @kartben I believe this is the preferred copyright notice nowadays.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -14,10 +15,39 @@
...
+#endif
+
+#include "hardware/dma.h"
+
+#endif
+
 #define DT_DRV_COMPAT worldsemi_ws2812_rpi_pico_pio
 
+struct ws2812_led_strip_dma_data {
+#if defined(CONFIG_DMA)
```
simonguinot, line 33: I am not a fan of having a useless `dma_data` pointer in `struct ws2812_led_strip_data` when `CONFIG_DMA` is not enabled. It costs memory.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -14,10 +15,39 @@
...
+	struct dma_config dma_config;
+	struct dma_block_config dma_block;
+	struct k_sem complete_sem;
+	uint32_t *pixel_buf;
+#endif
+};
+
 struct ws2812_led_strip_data {
 	uint32_t sm;
+	struct k_timer reset_on_complete_timer;
```
simonguinot, line 49: This timer updates the non DMA path as well. It would probably be better/clearer to introduce this in a dedicated commit (with a proper message), separate from DMA support.

simonguinot (commented):
Hi @TerryGeng,

Thanks for this patch. Overall, it seems OK to me. But since I am not familiar with this device, I'll let @soburi handle the review.

Please find a couple of comments below.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -80,31 +226,35 @@ static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb
...
-				color |= pixels[i].g << (8 * (2 - j));
-				break;
-			case LED_COLOR_ID_BLUE:
-				color |= pixels[i].b << (8 * (2 - j));
-				break;
-			}
+	/* Wait for the delay needed to latch current color values on
+	 * WS2812 and reset its state machine.
+	 */
+	k_timer_status_sync(&data->reset_on_complete_timer);
```
fabiobaltieri, line 255: this is to avoid two `ws2812_led_strip_update_rgb` firing too close to each other right? it's a bit odd to use a timer for a single shot delay, would it make sense to use a `k_work_delayable` instead?

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -80,31 +226,35 @@ static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb
...
+
+	if (COND_CODE_1(CONFIG_DMA, (!config->dma_config.dev), (true))) {
```
fabiobaltieri: this is really hard to follow, how about refactoring it to have something like

```
if (use_dma(...)) {
        return transfer_with_dma();
}

<normal transfer>
```

then you can have a single ifdef block that defines `use_dma` to `return false`, the other as a noop and everything gets optimized out when not needed and the two logic are well isolated and you have less conditionals kicking around the normal code path

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: Thanks for the review. You are correct that this timer (or before this PR, a `k_usleep`) is to wait for the state of WS2812 to reset before updating it again. This reset time, per the datasheet, is several hundred microseconds.

The reason for moving from using `k_usleep` to the timer is: It is unlikely that the LED update function is called twice within the reset time, and calling `k_usleep` (i.e. triggering a context switch) regardless of the actual time between the two function calls to the `update` function is likely wasteful. So it is better to note down the time when the last update takes place and, when on the next call to update, compare the current time to the last update time to decide whether to sleep or not to sleep. I think a timer is a very good tool to guarantee the passage of a certain amount of time (and block the thread only if the wait time hasn't fully elapsed), so I went for it.

On the other hand, `k_work_delayable` is agnostic about when the last update takes place, so I couldn't see how to easily achieve the above.

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: Yes, I can see how I can improve the readability.

drivers/led_strip/ws2812_rpi_pico_pio.c
fabiobaltieri: ok, I'll be honest I think the sleep would work well enough in practice, but sure let's keep the timer

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -31,6 +61,7 @@ struct ws2812_led_strip_config {
 	const uint8_t *const color_mapping;
 	uint16_t reset_delay;
 	uint32_t cycles_per_bit;
+	const struct ws2812_led_strip_dma_config dma_config;
```
fabiobaltieri, line 65: shouldn't this have a `CONFIG_DMA` guard too?

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -79,32 +244,30 @@ static int ws2812_led_strip_update_rgb(const struct device *dev, struct led_rgb
...
-				color |= pixels[i].r << (8 * (2 - j));
-				break;
-			case LED_COLOR_ID_GREEN:
-				color |= pixels[i].g << (8 * (2 - j));
-				break;
-			case LED_COLOR_ID_BLUE:
-				color |= pixels[i].b << (8 * (2 - j));
-				break;
-			}
+	if (!ws2812_led_strip_use_dma(dev)) {
```
fabiobaltieri: swap the condition and keep the non-dma exectuion inline, then on the dma branch do the dma stuff and put a `return 0` at the end, that way you'll avoid that split `} else {`

also my suggestion was to move the whole dma path in its own function so you don't even have a `if defined(CONFIG_DMA)` at all in this function,  so that's still an option, like this could be a

```c
if (ws2812_led_strip_use_dma() {
   return ws2812_led_strip_start_dma_put();
}

<normal non dma code>
```

and then those two  functions could be no-op in the non dma case, but I think


```c
#if defined(CONFIG_DMA)
if (ws2812_led_strip_use_dma() {
   <do the dma thing>
   return 0;
}
#endif
```

is also valid, if you go with that you can remove the condition from`ws2812_led_strip_use_dma` and just undefine it in the non-dma case

am I making sense?

TerryGeng: If you want me to populate the pixel buf in that `ws2812_led_strip_start_dma_put` function, I can also do that.

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -15,11 +15,41 @@
...
 	struct k_timer reset_on_complete_timer;
+#if defined(CONFIG_DMA)
+	struct ws2812_led_strip_dma_data *dma_data;
+#endif
+};
+
+struct ws2812_led_strip_dma_config {
+	const struct device *dev;
+	uint32_t tx_channel;
 };
```
fabiobaltieri: ok so this can go up next to `ws2812_led_strip_dma_data` and only get declared if DMA is enabled

drivers/led_strip/ws2812_rpi_pico_pio.c
fabiobaltieri: nah no need, though I was tempted to suggest moving that loop out before the `if` and do it for both dma and non dma case (and have `pio_sm_put_blocking` work on the data already altered), what do you think?

drivers/led_strip/ws2812_rpi_pico_pio.c
```diff
@@ -158,6 +317,14 @@ static int ws2812_led_strip_init(const struct device *dev)
 
 	data->sm = sm;
 
+	k_timer_init(&data->reset_on_complete_timer, NULL, NULL);
+
+#if defined(CONFIG_DMA)
+	if (config->dma_config.dev) {
```
fabiobaltieri: you can use `ws2812_led_strip_use_dma` here too

drivers/led_strip/ws2812_rpi_pico_pio.c
TerryGeng: I recall the ws2812 SPI driver was written in the way you suggested (since SPI drivers can possibly do async or DMA IO depending on DT). I tend not to do this here since creating a pixel buffer takes quite a bit of memory, and doesn't offer much convenience in the non-DMA case since pixels are written to ws2812 one by one.

fabiobaltieri (approved):
good stuff

TerryGeng:
@fabiobaltieri Thanks for your rapid response and patience!

simonguinot:
@soburi is that good for you ?

soburi (approved):

simonguinot (approved):
Thanks @TerryGeng !

PR #101523, opened 2025-12-26
samples: drivers: crc: fix twister regex and run in CI

Holt-Sun:
1. Update the CRC sample’s Twister console regex to match the current log output.
2. Add an integration_platforms entry (ek_ra8m1) so the sample runs in PR CI.
3. Make CRC verification unconditional so the expected “verification succeeded” log is always emitted and mismatches fail the test.

### DIFF
--- a/samples/drivers/crc/sample.yaml
+++ b/samples/drivers/crc/sample.yaml
@@ -6,8 +6,10 @@ tests:
     tags:
       - drivers
       - crc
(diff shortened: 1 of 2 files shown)


thenguyenyf (approved):
LGTM!

zejiang0jason (approved):

ZhaoxiangJin (approved):

